PR #6036, opened 2021-04-30
Ensure MFT digits are decoded in chipID increasing order

shahor02:
@bovulpes @rpezzi here is it. I see 1% overhead due to the sorting (tested on 100 TFs with 70 pbpb events).


bovulpes (approved):

bovulpes:
Thanks @shahor02 it works fine now.

PR #6038, opened 2021-05-01
 Raw decoding optimized; Made Mapping singleton; Semigauss raw fitter added;

peressounko:
Raw decoding optimized; Made Mapping singleton; Semigauss raw fitter added;


Detectors/PHOS/base/include/PHOSBase/Mapping.h
```diff
@@ -50,17 +50,19 @@ class Mapping
...
   ~Mapping() = default;
 
+  //Getters for unique instance of Mapping
+  static Mapping* Instance();
+  static Mapping* Instance(std::basic_string_view<char> path);
+
   /// \brief convert hardware address to absId and caloFlag
   ErrorStatus hwToAbsId(short ddl, short hw, short& absId, CaloFlag& caloFlag);
   /// \brief convert absId and caloflag to hardware address and ddl
   ErrorStatus absIdTohw(short absId, short caloFlag, short& ddl, short& hwAddr);
```
shahor02: const methods ?

Detectors/PHOS/base/src/Mapping.cxx
```diff
@@ -18,12 +18,38 @@
...
+      return sMapping;
+    } else {
+      delete sMapping;
+    }
+  }
+  sMapping = new Mapping(path);
+  sMapping->setMapping();
+  return sMapping;
+}
+//_______________________________________________________
```
shahor02, line 52: Please don't use in https://github.com/AliceO2Group/AliceO2/pull/6038/files#diff-4f5317238307bf32c475ee3ca25b3ca8e1b59324833e0e6a9c64da7b05009ee6R136-R138 the sprintf, moreover that you don't know a priory what is the length of $O2_ROOT. Use instead e.g.
````
std::string fname = fmt::format("{:s}/Mod{:d}RCU{:d}.data",p,m,i);
````
Also, using of `new` (btw, you don't delete the fIn, close() just closes the file) is disadvised, you can simply use the local instance `std::ifstream fIn(fname)` which will be deleted automatically on exit.

File: Detectors/PHOS/reconstruction/include/PHOSReconstruction/AltroDecoder.h [hunk 3]
```diff
@@ -97,35 +99,41 @@ class AltroDecoder
...
-  void checkRCUTrailer();
-
-  RawReaderMemory& mRawReader;       ///< underlying raw reader
-  RCUTrailer mRCUTrailer;            ///< RCU trailer
-  std::vector<Channel> mChannels;    ///< vector of channels in the raw stream
-  bool mChannelsInitialized = false; ///< check whether the channels are initialized
+  //check and convert HW address to absId and caloFlag
+  bool hwToAbsAddress(short hwaddress, short& absId, Mapping::CaloFlag& caloFlag);
+  //read trigger digits
+  void readTRUDigits(short absId, int payloadSize, std::vector<o2::phos::Cell>& truContainer);
```
shahor02: const method?

Detectors/PHOS/reconstruction/include/PHOSReconstruction/CaloRawFitter.h
```diff
@@ -65,44 +65,44 @@ class CaloRawFitter
...
-  float getChi2(int is) const { return mChi2[is]; }
+  float getChi2() const { return mChi2; }
 
   /// \brief time in last fitted sample
-  float getTime(int is) { return mTime[is]; }
+  float getTime() { return mTime; }
 
   /// \brief is last fitted sample has overflow
-  bool isOverflow(int is) { return mOverflow[is]; }
+  bool isOverflow() { return mOverflow; }
```
shahor02: make getters const methods, also below

Detectors/PHOS/reconstruction/src/CaloRawFitter.cxx
```diff
@@ -285,13 +242,13 @@ CaloRawFitter::FitStatus CaloRawFitter::fitGamma2(const Bunch& b)
...
       uint16_t si = *it;
-      float ti = tsart + i - t0;
+      float ti = i - t0;
       it++;
       i++;
-      if (mOverflow.back() && si == mMaxSample) { //do not fit saturated samples
+      if (mOverflow && si == mMaxSample) { //do not fit saturated samples
         continue;
       }
       float fi = ti * ti * exp(-ti * alpha);
```
shahor02: you can save operations by precalculating `(si - A * fi)`

Detectors/PHOS/reconstruction/src/CaloRawFitterGS.cxx
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  mStatus = kNotEvaluated;
+  // Extract amplitude and time
+  mStatus = evalFit(signal);
+  return mStatus;
+}
+
+CaloRawFitterGS::FitStatus CaloRawFitterGS::evalFit(gsl::span<short unsigned int> signal)
+{
```
shahor02, line 80: here too, I guess this method is used a lot, you can save quite some multiplications by precalculating some terms or using Horner form for polynomials

Detectors/PHOS/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -68,35 +59,42 @@ void RawToCellConverterSpec::init(framework::InitContext& ctx)
   if (fitmethod == "default") {
     LOG(INFO) << "Using default raw fitter";
     mRawFitter = std::unique_ptr<o2::phos::CaloRawFitter>(new o2::phos::CaloRawFitter);
-    //TODO: Configure parameters of fitter from options
-    // mRawFitter->setAmpCut(mNoiseThreshold);
-    // mRawFitter->setL1Phase(0.);
+  }
+  if (fitmethod == "semigaus") {
+    LOG(INFO) << "Using SemiGauss raw fitter";
+    mRawFitter = std::unique_ptr<o2::phos::CaloRawFitter>(new o2::phos::CaloRawFitterGS);
```
shahor02, line 65: mRawFitter is a pointer on the non-virtual CaloRawFitter class, which means that e.g. `evaluate` method of CaloRawFitter will be used, even if you cast the pointer from CaloRawFitterGS.
You should either use virtual methods or use templates.

shahor02 (changes requested):

Detectors/PHOS/workflow/src/RawToCellConverterSpec.cxx
peressounko: Dear @shahor02, this function will be called a lot. What would be more efficient: putting semiGS method to the base class CaloRawFitter and switch between methods with "if()" statement; Template or virtual member function?

shahor02: well, using a virtual methods has some small overhead, but given that the method itself is quite compute-intensive, I don't think you notice any difference.

peressounko:
dear @shahor02 , I implemented all your suggestions now. Dmitri

Detectors/PHOS/reconstruction/include/PHOSReconstruction/CaloRawFitterGS.h
```diff
@@ -0,0 +1,67 @@
...
+ public:
+  static constexpr int NMAXSAMPLES = 40; ///< maximal expected number of samples per bunch
+  /// \brief Constructor
+  CaloRawFitterGS();
+
+  /// \brief Destructor
+  ~CaloRawFitterGS() = default;
+
+  /// \brief Evaluation Amplitude and TOF
+  FitStatus evaluate(gsl::span<short unsigned int> signal);
```
shahor02: please define it as final, otherwise the checker will complain

Detectors/PHOS/reconstruction/include/PHOSReconstruction/CaloRawFitter.h
```diff
@@ -45,7 +45,7 @@ class CaloRawFitter
 
  public:
   /// \brief Constructor
-  CaloRawFitter() = default;
+  CaloRawFitter();
 
   /// \brief Destructor
   ~CaloRawFitter() = default;
```
shahor02: destructor also should be virtual

Detectors/PHOS/reconstruction/include/PHOSReconstruction/CaloRawFitterGS.h
```diff
@@ -0,0 +1,67 @@
...
+class CaloRawFitterGS : public CaloRawFitter
+{
+
+ public:
+  static constexpr int NMAXSAMPLES = 40; ///< maximal expected number of samples per bunch
+  /// \brief Constructor
+  CaloRawFitterGS();
+
+  /// \brief Destructor
+  ~CaloRawFitterGS() = default;
```
shahor02: define it as `default final`

shahor02 (changes requested):

shahor02 (approved):

PR #6045, opened 2021-05-02
Adapt ITS-TPC matching combined or separate ITS-(TPC,TPC-TOF) input

shahor02:
````o2-tpcits-match-workflow --track-sources "ITS,TPC,TPC-TOF"```` 
will use both `TPC` and `TPC-TOF` tracks (accounting for redundancy) to match to ITS (it can be skipped in the `track-sources`). 
With --track-sources "TPC-TOF"` only TOF-constrained TPC tracks will be used.


shahor02:
@davidrohr @TimoWilken any idea what could be the reason for FullCI  showing `task timeout reached .. killing all processes`?
Locally I get the same ~30s for `reco_NOGPU` stage, same as before.

TimoWilken:
I've restarted the test in case it was just a random failure (but left the log URL intact in case anyone wants to investigate in the meantime).

davidrohr:
> @davidrohr @TimoWilken any idea what could be the reason for FullCI showing `task timeout reached .. killing all processes`?
> Locally I get the same ~30s for `reco_NOGPU` stage, same as before.

No idea, to me it looks like the workflow didn't actually start to run.
I see from the file reader that it reads the file, but all other processors just report that they are ready.
I have seen this problem that it doesn't start when there was some bogus shm segments were left from a previously failing tests.

@TimoWilken : Actually, will 2 CI jobs if they run on the same server run inside the same docker container? Then we might actually have a problem, since the FST uses the fixed `default` session name. So 2 running in parallel could collide.

TimoWilken:
> Actually, will 2 CI jobs if they run on the same server run inside the same docker container? Then we might actually have a problem, since the FST uses the fixed `default` session name. So 2 running in parallel could collide.

No, they would run in separate docker containers.

PR #6047, opened 2021-05-03
Use InputHelper in global workflows, avoid global DataRequest + related fixes

shahor02:


davidrohr (dismissed):
Moving the specs task classes from the header into the cxx is in my opinion not strictly needed, but in some way also cleaner since it reduces the bloat inside the header, and the spec task class is by definition only used in that single cxx, so for me this is totrarlly fine.

davidrohr:
Obviously you have to add the GlobalTrackingWorkflowHelpers as dependency to the globaltrackingworkflow. I am wondering how it can have worked locally for you.

In general, we have to avoid cyclic dependencies. I was thinking about moving all the readers from the detectors into the GlobalTrackingWorkflowReaders library. (Which is separate from this input helper library. The idea is that it contains only the readers, and it dependes only on the data formats.)

Then, the detector workflows (and the global tracking workflow) would only depend on this reader library (if they instantiate the reader directly) and possibly on the library with the inputhelper if needed.

What I don't like is mixing up the readers of different detectors, but having one reader library per detector is probably overkill?
Alternatively, I would create one `reader` subdirectory inside each detectors' workflow directory, and create a reader library there. That would be a bit more fine grained.

What is your opinion on this?

shahor02:
> Obviously you have to add the GlobalTrackingWorkflowHelpers as dependency to the globaltrackingworkflow. I am wondering how it can have worked locally for you.

I also have no idea, locally it worked that's why I did not notice that the dependency is missing

I would be more in favor of single readers directory, though I am not sure that indeed all detectors have their dataformats in DataFormats (should be imposed). Then, eventually the same should be done for writers.

davidrohr:
Hi Ruben,

ok, if you prefer single libraries than I will do that. In case sume stuff is not in data formats, I'll see if I can easily move it there.

For the writers: I am not sure whether it is necessary to extract it into a separate library, since I think the writer will usually run together with the reconstruction workflow. But ok, if the writer will only depend on the data formats, we can move it to the reader library later if needed.

davidrohr:
@TimoWilken : The fullCI has been reporting `Queued (1 ahead) on mesosci/ci_alisw_slc8-gpu-builder_latest/0` for several hours now. Can you check whether it is stuck?

TimoWilken:
That builder has just finished PR 6026; it's running this one now.

PR #6026, opened 2021-04-29
TRD tracking adaptions in view of GPU processing

martenole:
- tracklets are already sorted by HCID
- tracklets don't need to be loaded individually anymore

PR #5912, opened 2021-04-12
TRD raw data reconstruction, compressed and not.

bazinski:
- read cru on flp
- compress
- read raw and compressed data on epn.
- data is self consistant with the mc 2 raw generation.
- zero suppressed digits (mc is only done with these)
- real raw data needs the byte swap option described in the docs.
- add event record which stores a trd event.
Documentation added to the readme as well.

PR #6056, opened 2021-05-03
Move TRD readers into independent lib

martenole:
To reduce the number of dependencies which are added when including the readers.
Also cleaned up a bit leftover unused variables and includes which are no longer needed.
Ping @davidrohr


martenole:
@jolopezl I noticed that the digit reader includes `TRDBase/MCLabel.h`. In this header there is only an alias defined, basically `using o2::trd::MCLabel = o2::MCCompLabel;`. Is this alias needed? If so I can also move that to `DataFormatsTRD`. Because otherwise the new library which I created for the TRD workflow readers would depend again on TRDBase and we need to reduce the number of dependencies for the readers as much as possible.

davidrohr:
Hi @martenole : please move the header to DataFormatsTRD, there is no reason that this resides in TRDBase.

jolopezl:
> @jolopezl I noticed that the digit reader includes `TRDBase/MCLabel.h`. In this header there is only an alias defined, basically `using o2::trd::MCLabel = o2::MCCompLabel;`. Is this alias needed? If so I can also move that to `DataFormatsTRD`. Because otherwise the new library which I created for the TRD workflow readers would depend again on TRDBase and we need to reduce the number of dependencies for the readers as much as possible.

@martenole, the alias isn't needed. It has only historical meaning and stood there waiting for a specialization. So, it can be removed, moved, left there, or – maybe better – have new readers depend on `o2::MCCompLabel` only, as you already did.

martenole:
Ah ok, thanks. Then I'll remove it again :)

bazinski (dismissed):

martenole:
With the last commit I also moved the digit and tracklet writer specs into the new library (now called `TRDWorkflowIO` instead of `TRDWorkflowReaders`). The only additional dependency is `DPLUtils` for the ROOT tree writer.
@bazinski I could have also moved `TRDTrapRawWriterSpec`, but I noticed it is not added into a workflow yet. This probably comes with your https://github.com/AliceO2Group/AliceO2/pull/5912 ? Anyway I did not touch it yet to not interfere, but this can be trivially moved as well later.
@davidrohr There is also a `TRDTrackWriterSpec` which I did not move, since this would pull some GPU dependencies for the `GPUTRDTrack` type. But when I add a track reader then this will anyways be needed, so it might come later when the GPU data types are in a separate library.

bazinski:
@martenole no TRDTrapRawWriterSpec was removed when the raw data generation was removed from the trapsim, so it used to produce digits/tracklets/triggerrecords/rawstream. The latter is no longer there. The logic that sat there was removed to trap2raw so all the raw data generation is logically contained in 1 place instead of 2.

davidrohr:
@martenole : Please don't pull in the GPU dependency to the reader library, let's leave the writer in the normal TRDWorkflow then. I have just merged your #6026, so you have to rebase and solve a minor conflict :)

davidrohr:
Fix for the CI failures in in #6059 
@TimoWilken : Could you restart the failed CIs?

davidrohr:
superseded by #6057

PR #6058, opened 2021-05-04
Rename option in TOF MC QA

njacazio:


njacazio:
Hi, this fix is needed because of a conflict between option naming when running the task with the TOF PID task (which is always needed).

iarsene (approved):

jgrosseo:
I merged it, but aren't options per workflow?

njacazio:
I had the same questions myself, but indeed I could not make it work together, with this change the error is gone

PR #6060, opened 2021-05-04
MCH: adapted page reader output spec to three-letters standard

aferrero2707:
The origin string in the MCH page reader output spec is shortened to three letter, as the previous four-letters version is now rising an exception due to excessive length.


aphecetche:
@aferrero2707 fine with me, but where is this 3 letters standard documented ?  (I was not aware of it)

aphecetche (approved):

PR #6062, opened 2021-05-04
DPL: add a DataTakingContext to be filled with run related constants

ktf:


ktf:
@Barthelemy you can get the run number from here now.

PR #6066, opened 2021-05-04
TRD track readers + requested changes from PR6057

martenole:
Incorporates most changes requested in https://github.com/AliceO2Group/AliceO2/pull/6057
Missing is the disentangling of the GPUTRDInterface header (will be added tomorrow) and the proper logic to avoid loading the same track twice (is only done in a simple stupid way at the moment). The latter will probably come in a separate PR

### DIFF
--- a/GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
+++ b/GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
@@ -201,13 +201,13 @@ void GPUTRDTracker_t<TRDTRK, PROP>::Reset()
 }
 
 template <class TRDTRK, class PROP>
-void GPUTRDTracker_t<TRDTRK, PROP>::DoTracking(GPUChainTracking* chainTracking)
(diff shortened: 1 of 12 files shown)


GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
```diff
@@ -200,6 +200,54 @@ void GPUTRDTracker_t<TRDTRK, PROP>::Reset()
   mNTracks = 0;
 }
 
+template <class TRDTRK, class PROP>
```
davidrohr, line 203: Thx, I would propose to call the preparefunction from the DoTracking to avoid duplicating the code.

martenole:
The output format (TRD track interface for O2 base track class) is moved to `GPUTracking/DataTypes`. This is ready now (at least for review)

davidrohr:
Looks OK, will check later when the CI is through.

davidrohr:
seems to work so far

PR #6069, opened 2021-05-05
Tested with the Entropy Encoder. OK

fapfap69:


davidrohr:
I have tried this with the latest O2 dev + this PR and seems to work in the full system test.
For the next time, please use a more reasonable title for the commit and for the PR, no problem for now, I will change it.

And then, I see that you added a second decoding class. Without checking the details, in general I would try to avoid code duplication, so I do not understand why we need 2 HMPID decoders. I think this could be followed up and cleaned up a bit. But anyway, thx for fixing the issues, merging this since it is an improvement.

fapfap69:
@davidrohr  yes after we close the transition, i'll provide to remove the old version of coder/decoder and drop down the code duplication

PR #6074, opened 2021-05-05
[WIP] TRD CCDB object for calibration parameters

aberdnikova:


martenole:
Isn't this simply a copy of the class which we already have `ChamberCalibrations`? Why do we need it twice?

bazinski:
@martenole high availability ;-)

Detectors/TRD/base/include/TRDBase/CalVdriftExB.h
```diff
@@ -0,0 +1,56 @@
...
+  void setVDrift(int p, int c, int s, float vdrift) { mVDrift[o2::trd::Geometry::getDetector(p, c, s)] = vdrift; };
+  void setVDrift(int roc, float vdrift) { mVDrift[roc] = vdrift; };
+  void setExB(int p, int c, int s, float exb) { mExB[Geometry::getDetector(p, c, s)] = exb; };
+  void setExB(int roc, float exb) { mExB[roc] = exb; };
+  //bulk gets ?
+  bool init(int run2run = 0);
+
+ protected:
+  std::array<float, constants::MAXCHAMBER> mVDrift{};     // mean drift velocity per chamber.
+  std::array<float, constants::MAXCHAMBER> mGainFactor{}; // mean gas gain per chamber
```
tdietel, line 50: The parameters mGainFactor and mT0 should be removed from this class. This is because they will be created by a different calibration device, and stored separately.

tdietel (changes requested):
Sean raised a good point here, and we should get the calibration classes into a good shape before we actually move on.

As I see it, we have two overall options to move forward:

- Use a generic (1 float or int per chamber) class for each parameter. One calibration device would then write 2 or 3 classes.

- Create a specific class for each of the time-slot based calibrations ( vDrift,ExB / 4x4 gain factors / T0,real drift velocity, alt gain). 

I think the advantage of the first method is that there is no code duplication. The second approach with dedicated classes would have some duplication, but this seems to be minimal: just a few calls to `o2::trd::Geometry::getDetector(p, c, s)`.

At the moment, I am leaning towards the second option. Thoughts?

martenole:
Hi @tdietel 
the first approach is as far as I can see already implemented for pad calibrations (see https://github.com/AliceO2Group/AliceO2/blob/1138de579d1fd9b75a51aa6944d7a03555ff75ad/Detectors/TRD/base/include/TRDBase/PadCalibrations.h#L35 and for example the pad noise https://github.com/AliceO2Group/AliceO2/blob/1138de579d1fd9b75a51aa6944d7a03555ff75ad/Detectors/TRD/base/include/TRDBase/PadNoise.h#L32)

Could we not simply do the same for chamber based calibrations?

Detectors/TRD/base/include/TRDBase/CalVdriftExB.h
```diff
@@ -0,0 +1,56 @@
...
+{
+namespace trd
+{
+class CalVdriftExB
+{
+ public:
+  CalVdriftExB() = default;
+  ~CalVdriftExB() = default;
+  //
+  float getVDrift(int p, int c, int s) const { return mVDrift[o2::trd::Geometry::getDetector(p, c, s)]; };
```
martenole, line 37: Why would we need all these setters and getters at all for values which are only stored per chamber? Whereever the calibration parameters are queried they are queried for a full chamber, are they not?

aberdnikova: i think in the line just below there is a getter to get a value by i_det, and this is an alternative option to get it by layer, stack, sector number? same for setters

martenole: @aberdnikova I see but my point was that I don't think this functionality is needed by this class

tdietel:
> Hi @tdietel
> the first approach is as far as I can see already implemented for pad calibrations (see
> 
> https://github.com/AliceO2Group/AliceO2/blob/1138de579d1fd9b75a51aa6944d7a03555ff75ad/Detectors/TRD/base/include/TRDBase/PadCalibrations.h#L35
> and for example the pad noise
> 
> https://github.com/AliceO2Group/AliceO2/blob/1138de579d1fd9b75a51aa6944d7a03555ff75ad/Detectors/TRD/base/include/TRDBase/PadNoise.h#L32
> )
> 
> Could we not simply do the same for chamber based calibrations?

I find the approach for the pad parameters quite elegant, and the same could be done easily for the chamber parameters. It would be a different approach to what discussed during the week, but it would give us a fair bit of flexibility.

One could even think about a unified class for arbitrary pad groups - if you specify the number of groups in pad and row direction as template parameters.

What is the best way to proceed with this discussion? Here on GitHub, or is a short Zoom call more efficient?

Detectors/TRD/base/include/TRDBase/CalVdriftExB.h
bazinski: either way the p, c, and s should at least be expanded. The person that originally wrote this with the p, c, and s should be keel hauled ;-)

martenole: Agree with @bazinski 
But on topic: all we need are 4 arrays with one value per chamber. This could be a simple `struct` which then can go into `DataFormatsTRD`. No need for any dependency on simulation parameters or geometry. No need for any setters or getters. If one would need to go from p, c, s (whatever these values mean) to detector number then this can be done at the place where these values are queried. Because at those places one also knows about the TRD geometry (for example in the TRD digitizer)

bazinski: i concur with @martenole about removing the p,c,s version of getVDrift and thereby removing dependencies.

tdietel:
Hello @aberdnikova,

Is there any news on on the calibration objects for the database? There are a few requests/suggestions, and I think we should be able to merge this soon.

Cheers, Tom.

aberdnikova:
hi @tdietel, actually i think we should close this PR, because these changes are already a part of a new bigger future PR based on time slot workflow. what do you think @martenole ?

martenole:
fine with me @aberdnikova

PR #6075, opened 2021-05-05
[O2-2018] Drop support for Microsoft GSL v2

aphecetche:
Only use V3 from now on.
We can thus drop our custom `Findms_gsl.cmake`.
Also using the "official" target name of `Microsoft.GSL::GSL` instead of
our own `ms_gsl::ms_gsl`.


aphecetche:
drafting this PR as some work is needed on the QC side first...

aphecetche:
@Barthelemy I should probably know this, but is the tagging of QC kind of regular or done "on demand" ?

Barthelemy:
@aphecetche we do it before the O2Suite releases but also on demand

aphecetche:
@vkucera  regarding AliceO2Group/AliceO2#6201 is there some documentation people could use to know how to deal with a failing Space Checker CI test ? (like in this PR for instance :wink: )

vkucera:
> @vkucera regarding #6201 is there some documentation people could use to know how to deal with a failing Space Checker CI test ? (like in this PR for instance wink )

Hi @aphecetche, The [guidelines](https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html?showone=Spaces_vs._Tabs#Spaces_vs._Tabs) advise: "You should set your editor to emit spaces when you hit the tab key." So I don't think any further instructions are necessary.
Any decent editor should be able to replace tabs and remove trailing spaces automatically when the file is saved.
Another way is to use `sed -e "s/${old}/${new}/g" ${file}` for example.

aphecetche:
@vkucera well, tabs are one thing, but if I read correctly the errors I get there, most of them are trailing whitespaces. And while I can certainly instruct my text editor to be careful with those as well, I'm pretty sure those will creep in many files for many people (in particular in files where they are already present), so it'd be good to advertise how to deal with those as well.

vkucera:
@aphecetche I see 21 lines with trailing spaces and way more lines with tabs.
But anyway, I don't understand where you see the difficulty. If a file appears in a PR, it means that the author opened it in an editor, modified it and saved it. If the editor is configured as advised by the guidelines, all the tabs and trailing spaces in the changed files are fixed automatically. And if they use an editor that is not able to do this, I am sure they can use the replacing functions.

aphecetche:
I may be pessimistic but this space checker is yet a another thing that can delay PR merging (with, IMO, very little added value for the code quality) and I'm afraid people won't be very patient with it, even more so if they get no guidance on how to deal with it.

aphecetche:
Actually I would go even further and say that this change should be accompanied with an automated procedure to help fix the errors (e.g. automated PR request to the original branch like what's done with the clang-format errors ?)
Sorry for the rant.

vkucera:
In my opinion, removing trailing spaces is a task that is so trivial that it does not deserve dedicated instructions. Trailing spaces worsen readability of diffs which does matter to me since I am reviewing pretty large PRs on a daily basis.
It takes 10 seconds to fix it in any editor I know and if the editor is configured properly, there is zero delay for the author and for the PR. So I am pretty sure people will spend much less time with it than with `clang-format` for example.
But feel free to add instructions in the guidelines if you are convinced that people will not figure it out easily.
The best would be to let the GitHub actions push fixes in the PR branches automatically. It used to be the case for `clang-format` for a while but then it was removed for some technical problems with GitHub.

ktf:
As far as I know the clang-format automatic fix should still be there. If not, @TimoWilken could you look into it?

aphecetche:
(as a side note, for the diffs, the whitespace changes can be hidden in GitHub for instance)
Maybe I'm just having a bad week, but frankly I don't enjoy the idea of opening again the 50 files of this PR just for whitespacing them, and don't really want to spend my time there. That should be done by a (existing) script.

vkucera:
@aphecetche For the tabs in CMake files, `cmake-format` fixes that, so once we agree on the format and the PR is merged, tabs will not be an issue.
For the trailing spaces, I already suggested several ways which literally take seconds to apply to the changed files.
But I agree that having a tool that fixes the PR automatically would be better.

davidrohr:
> Trailing spaces worsen readability of diffs which does matter to me since I am reviewing pretty large PRs on a daily basis.

In general I agree, but you can also just do diffs that ignore whitespace changes.

aphecetche:
@Barthelemy @TimoWilken @ktf this one is still red because of the QC. But QC has been tagged (v1.18 yesterday) with the relevant changes that should fix the issue. But as far as I can tell the CI is still using 1.17. Is there something particular to be done so that the latest QC tag is used in the CI ?

Barthelemy:
@aphecetche I don't think that the failures are related to QC any more. The `macOS` build that was affected is actually green.

Barthelemy:
My bad, it is indeed related to the QC but it is not the same. It is something new. QC v1.18 has been released but the PR in alidist has not been merged yet. I don't know what is going on.

Barthelemy:
ok, understood, sorry, still sleepy. QC v1.18 was tagged but the PR in alidist was not yet merged (done now). Thus the error here. Could you trigger a new build (with a useless commit) ?

aphecetche:
@ktf @davidrohr @ihrivnac this one is finally green and ready to be reviewed and/or merged.

davidrohr:
gladly!

PR #6081, opened 2021-05-06
TOF calibration with cosmics + improvements for TimeCalibration framework

chiarazampolli:
With @noferini .


noferini (dismissed):

chiarazampolli:
Hello @shahor02 , 

It is not urgent, but maybe you can take a look before the calib review next week? There are changes in the TimeSlotCalibration also in this PR.

Cheers,

Chiara

chiarazampolli:
Many errors in the CI, they look not coming from the changes in this PR to me - but there are many, not sure I overlooked some related to this PR.

DataFormats/Detectors/TOF/include/DataFormatsTOF/CosmicInfo.h
```diff
@@ -58,6 +58,29 @@ class CosmicInfo
 
   ClassDefNV(CosmicInfo, 2);
 };
+
+class CalibInfoTrackCl
+{
```
shahor02, line 63: This does not relates directly to CosmicInfo and I guess can be used independently, better define in separate file.

Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
```diff
@@ -112,27 +128,67 @@ template <typename Input, typename Container>
...
+
+  constexpr uint64_t INFINITE_TF = 0xffffffffffffffff;
+  constexpr int64_t INFINITE_TF_int64 = std::numeric_limits<long>::max() - 1; // this is used to define the end
+                                                                              // of the slot in case it is "std::numeric_limits<long>::max()"
+                                                                              // long (so we need to subtract 1)
+
+  // if we have one slot only which is INFINITE_TF_int64 long, and we are not at the end of run (tf != INFINITE_TF),
+  // we need to check if we got enough statistics, and if so, redefine the slot
+  if (mSlots.size() == 1 && mSlots[0].getTFEnd() == INFINITE_TF_int64) {
+    if (tf >= mUpdateInterval + mLastCheckedTF || tf == INFINITE_TF) {
```
shahor02: not sure this is correct: `mLastCheckedTF` will be updated to the new tf and the next check will be postponed by another mUpdateInterval TFs. Should not you use the lowest checked TF (should be introduced)?

Detectors/TOF/calibration/include/TOFCalibration/TOFChannelCalibrator.h
```diff
@@ -105,16 +121,335 @@ class TOFChannelCalibrator final : public o2::calibration::TimeSlotCalibration<o
...
-  static const int NCHANNELSXSECTOR = o2::tof::Geo::NCHANNELS / o2::tof::Geo::NSECTORS;
+  static double FuncDeltaOffset(double* x, double* params)
+  {
+    int i1 = int(x[0]) % 96;
+    int i2 = int(x[0]) / 96 ? (i1 + 48) : (i1 + 1);
+
+    if (i1 < 0)
+      return 0;
+    if (i2 >= Geo::NPADS)
+      return 0;
```
shahor02: braces.

Detectors/TOF/calibration/include/TOFCalibration/TOFChannelCalibrator.h
```diff
@@ -105,16 +121,335 @@ class TOFChannelCalibrator final : public o2::calibration::TimeSlotCalibration<o
...
+      //for (int sector = 0; sector < 1; sector++) {
+      int offsetsector = sector * Geo::NSTRIPXSECTOR * Geo::NPADS;
+      for (int istrip = 0; istrip < Geo::NSTRIPXSECTOR; istrip++) {
+        //for (int istrip = 0; istrip < 38; istrip++) {
+        int offsetstrip = istrip * Geo::NPADS + offsetsector;
+        int goodpoints = 0;
+
+        for (int ichLocal = 0; ichLocal < Geo::NPADS; ichLocal++) {
+          fracUnderPeak[ichLocal] = 0.0;
+        }
```
shahor02: faster to use `memset(&fracUnderPeak[0],0, sizeof(fracUnderPeak))`

Detectors/TOF/calibration/include/TOFCalibration/TOFChannelCalibrator.h
```diff
@@ -105,16 +121,335 @@ class TOFChannelCalibrator final : public o2::calibration::TimeSlotCalibration<o
...
+            intmax = -mRange;
+          }
+          if (intmin > mRange) {
+            intmin = mRange;
+          }
+          if (intmax > mRange) {
+            intmax = mRange;
+          }
+
+          /*
```
shahor02: if commented code is dead, please remove it.

Detectors/TOF/calibration/include/TOFCalibration/TOFChannelCalibrator.h
```diff
@@ -105,16 +121,335 @@ class TOFChannelCalibrator final : public o2::calibration::TimeSlotCalibration<o
...
+      int sector = ich / Geo::NPADSXSECTOR;
+      int chinsector = ich % Geo::NPADSXSECTOR;
+      auto entriesInChannel = c->integral(ich);
+      if (entriesInChannel < mMinEntries) {
+        LOG(DEBUG) << "channel " << ich << " will not be calibrated since it has only " << entriesInChannel << " entries (min = " << mMinEntries << ")";
+        continue;
+      }
+      std::vector<float> fitValues;
+      std::vector<float> histoValues;
+      /* //less efficient way
```
shahor02: idem

Detectors/TOF/calibration/include/TOFCalibration/TOFChannelCalibrator.h
```diff
@@ -105,16 +121,335 @@ class TOFChannelCalibrator final : public o2::calibration::TimeSlotCalibration<o
...
+        intmax = -mRange;
+      }
+      if (intmin > mRange) {
+        intmin = mRange;
+      }
+      if (intmax > mRange) {
+        intmax = mRange;
+      }
+
+      /*
```
shahor02: idem

shahor02 (commented):
Hi @chiarazampolli 
see below, I did not closely follow the logic of changes in TOF calib.

DataFormats/Detectors/TOF/include/DataFormatsTOF/CosmicInfo.h
noferini: Hi Ruben, it is true but in practice these infos (cosmic tracks with many TOF strips involved)  are extracted and re-aranged from CosmicInfo. In this way the combinatorial check for coincidence was run only once.
For this reason I prefered to put them together.

Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
noferini: This is true only if HasEnoughData returns true.
If not this will trigger the next check after N TF from the last check.
For instance with 10000 update time and infinite_TF slot the check is performed every 5 minuts (with 256 TF size) and the finalize is called only when enough statistics was accumulated. We tested it and it do exactly what I mentioned

DataFormats/Detectors/TOF/include/DataFormatsTOF/CosmicInfo.h
shahor02: but would not the same method of neighbours work also with collision data?

Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
chiarazampolli: Ciao @noferini , I think that Ruben understood correctly: what I want is that the next time the check happens is only after `mUpdateInterval`, from the last check, so it checks every `mUpdateInterval` TFs (more or less, since they are async). I don't understand what one would gain using the lowest checked TF. Note also that checking at every TF does not make much sense: this was introduce to optimize the check.

shahor02: Exactly, I thought if you consider that after 5 min it is worth to check the stat., then from that moment on you should keep checking. 5 min is an easy case, but if e.g. you put the mUpdateInterval to 1h, then the next check will happen only in 2h.
Anyway, this is a matter of conventions, can be left also like this. But the names are somewhat misleading, I would rename 
`mUpdateInterval` to `mCheckIntervalInfSlot` and `mLastCheckedTFInfSlot`, to indicate that this applies to infinite single slot processing case.

chiarazampolli: Ok, I can change the naming. When I tested, checking at every TF was not making sense, but maybe one could check with a difference frequency if the first check did not pass. This would require a new variable, but it is not an issue. What do you think? E.g. a fraction of the update interval.

shahor02: I meant checking at every TF starting from `mUpdateInterval` filled.

chiarazampolli: Yes, I understood, but if your mUpdateInterval is just a guess, it will check at every TF for long. This was my first implementation, and I did not like it.

shahor02: OK, lets keep it as it is at the moment. Though, checking hasEnoughStat is not supposed to take more time that evaluation the `if (tf >= mUpdateInterval + mLastCheckedTF || tf == INFINITE_TF)` itself :)

noferini: This was a my proposal since I didn't understand what is the gain to check only after a given threshold.
I started assuming that the problem was the overhead introduced by the check. For that I think that the best solution is to dilute such an operation (even if you expect to have enough statistics after 1h you can aslo check every 5 minutes since it is not a problem in terms of CPU).
If there is another reason to have such a threshold we can re-arrange name in a different way. 
If the overhead expected for the check is not a problem we can restore as it was before.

DataFormats/Detectors/TOF/include/DataFormatsTOF/CosmicInfo.h
noferini: In principle yes. In practice I don't know if we will need it.
We started to look at these tracks for two reasons:
- to study MRPC efficiency with cosmics
- to study if we can have something to align locally the strips
With magnetic field ON I think that muon coincidence cannot be used without track reconstruction.
In principle TOF tracks can be still used but it is not yet clear to me if we really have a benefit.

Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
chiarazampolli: The code initially checked at every TF. I changed it because if one expects that anyway it takes some time to reach enough statistics, then he can use this knowledge and tune the code. 
hasEnoughData should take longer than the `if's` above, e.g. for TOF we check the value of ~240K elements of an array. It is not super expensive, but why doing it if you know there is anyway no chance? In addition, setting this value to 1 would check at every TF, so that type of behavior is there.

If you think this does not make sense, I will remove it, but I think it could be beneficial. And: one can go with a solution in between: check ones after mUpdateInterval and if not successful check again with a different frequency. 

Let me know what you prefer, I tried to be flexible, but I will not insist (a check every time frame was bothering me very much in debugging phase, but this is a special case of course).

shahor02: Hi @chiarazampolli 

It does not hurt, since with default value mUpdateInterval to 1 it is equivalent to checking every TF. 
Concerning your use case of hasEnoughData in TOF: checking 240k elements is quite expensive, but why do you need to do this? When you fill the channel you can increment the counter of "completed channels" if it has reached the target stat., and in the hasEnoughData you just check this counter.

chiarazampolli: > since with default value mUpdateInterval to 1 it is equivalent to checking every TF

Yes, this is also what I mentioned. I had already thought about an "intermediate" setting, maybe I will add it. 
Someone might use it. 

> When you fill the channel you can increment the counter of "completed channels" if it has reached the target stat., and in the hasEnoughData you just check this counter.

Good suggestion, will do it on Monday!

chiarazampolli: Hi @shahor02 , @noferini ,
I introduced the possibility to retry the calibration at a different frequency, if the previous check failed. 
For now, I did not change the `hasEnoughData`, since the current code does not only count the number of entries, but also defines a mean over the channels that have at least 1 entry. So, if I move this part in the `fill`, I will need to do several `if`'s there, which might not be optimal. E.g.:
```
for (int i = data.size(); i--;) {
auto ch = data[i].getTOFChIndex();
...
if (mEntries[ch] == 0) mNValid++;
++mMean;
mEntries[ch] += 1;
}
```
Francesco, the occupancy at TOF in PbPb is how much? I think we usually said 100 useful tracks/collisions. Then, if we want to have 500 tracks per channel for beam data, we need (500 * 160000)/100 events, which, with IR = 30 kHz (average), means ~26 seconds. This is then ~2700 TFs. So, before getting enough statistics, I would do 2700 * 100 times `if (mEntries[ch] == 0) mNValid++;`. Right? 
At the moment, the `hasNoughtData`has some `if`s inside a loop over the number of channels (which is 160K in case of beam data, and 240K in case of cosmics), but we could even remove them as they are not really used. Then we would be left with the loop over the number of channels, which I think is less expensive than moving everything in the `fill` method. What do you think? 
@noferini , if needed, I can explain you what I mean in "virtual" person.

Chiara

Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
```diff
@@ -112,27 +135,72 @@ template <typename Input, typename Container>
...
+  constexpr int64_t INFINITE_TF_int64 = std::numeric_limits<long>::max() - 1; // this is used to define the end
+                                                                              // of the slot in case it is "std::numeric_limits<long>::max()"
+                                                                              // long (so we need to subtract 1)
+
+  // if we have one slot only which is INFINITE_TF_int64 long, and we are not at the end of run (tf != INFINITE_TF),
+  // we need to check if we got enough statistics, and if so, redefine the slot
+  if (mSlots.size() == 1 && mSlots[0].getTFEnd() == INFINITE_TF_int64) {
+    uint64_t checkInterval = mCheckIntervalInfiniteSlot + mLastCheckedTFInfiniteSlot;
+    if (mWasCheckedInfiniteSlot)
+      checkInterval = mCheckDeltaIntervalInfiniteSlot + mLastCheckedTFInfiniteSlot;
```
shahor02, line 150: braces...

shahor02 (commented):
By checking the mEntries in the fill, you can even avoid the filling if the stat. for the channel is ok and safe much more time than 1 if.
BTW, I guess you will have some dead channels, right? Then current `TOFChannelData::hasEnoughData` will always fail.

chiarazampolli:
> By checking the mEntries in the fill, you can even avoid the filling if the stat. for the channel is ok and safe much more time than 1 if.
> BTW, I guess you will have some dead channels, right? Then current `TOFChannelData::hasEnoughData` will always fail.

I am not sure I understand your comments. 
If I fill only after the check, then it means that the previous round already had the `hasEnoughStat` = true, so nothing will change, will it? It should not change because I have already calibrated and reset everything. 

What we do now to account for dead channels if to exclude those that have 0 entries, then calculate the mean number of entries in all other channels, and if this number if higher than a certain threshold, we calibrate. It is not true that it fails, I even have the proof that it works in the tests. Probably I have misunderstood your comment, then.

shahor02:
I meant checking in the fill method itself if the channel has already reached the target, then not filling it.

> It is not true that it fails, I even have the proof that it works in the tests

Ah, ok, sorry, I was reading your dev code: I was referring to this code: https://github.com/AliceO2Group/AliceO2/blob/01b8b2e983907665c2b8c1c2b1055f00963e7fe5/Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx#L81-L84
Yes, if you give up per channel check this should work. BTW, in case you have a noisy channel, would not it invalidate your mean calculation?

chiarazampolli:
> Yes, if you give up per channel check this should work. BTW, in case you have a noisy channel, would not it invalidate your mean calculation?

Honestly, I hope that noisy channels will be masked, and won't arrive at all. If suddenly something gets noisy, then yes, it will create an issue in the calculate of the mean (depending how many channels are noisy over the total available ones). @noferini , should we remove the entries that are above a certain threshold from the mean? 

@shahor02 , we had also a question. In the MathUtils/fit.h, is there a way to get the error on the fit parameters?

Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
noferini: Hi Chiara, we should discuss about that (I agree with you), how to implement the check in HasEnoughData properly. Now I would focus on the chain in order to have something which we can test on EPNs with cosmics to see where bottlenecks could be.

noferini:
> Honestly, I hope that noisy channels will be masked, and won't arrive at all. If suddenly something gets noisy, then yes, it will create an issue in the calculate of the mean (depending how many channels are noisy over the total available ones). @noferini , should we remove the entries that are above a certain threshold from the mean?

Yes, we can. The problem with noisy channels will be still there (I don't trust completely the masking) but we will find a solution. With tracks the problem should be almost negligible since we require a good match. With cosmics (no tracks) we can apply a noise filter (I have already something like that as on option in the clusterizer)

Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
chiarazampolli: Sure, if also for Ruben it is fine, we can leave it for the next PR (where we will also split the fitting from the accumulation of the data - this I mention for Ruben).

shahor02:
Fine for me, will merge once CI passed.
@TimoWilken The o2 is again killed by compiler, could you restart please?

chiarazampolli:
Hi @shahor02 , 

The checks are fine, can we merge?

Chiara

shahor02 (approved):

PR #6083, opened 2021-05-06
Log only 1st instances of Alpide/GBT errors in ITS/MFT, summarize on EOS

shahor02:
* LOG only 1st instance of new Alpide decoding error, show full report in the end.
Direct logging of decoding errors from AlpideCoder is changed to DEBUG level. Instead, each time a new Alpide decoding error type on the given GBT link is encountered for the 1st time, a LOG(ERROR) will be produced.
The decoded->printReport in the EOS of STFDecoder will be now called with per link decoding statistics
with skipEmpy option, showing statistics and errors only for links which registered some error.

* Log only 1st instance of new GBT link decoding error.
Error message will be produces at 1st encounter of given error type on given GBT link.
The full error summary with error counts will be printed on EOS.

### DIFF
--- a/Detectors/ITSMFT/common/reconstruction/include/ITSMFTReconstruction/AlpideCoder.h
+++ b/Detectors/ITSMFT/common/reconstruction/include/ITSMFTReconstruction/AlpideCoder.h
@@ -273,8 +273,7 @@ class AlpideCoder
 #ifdef ALPIDE_DECODING_STAT
           chipData.setError(ChipStat::NoDataFound);
 #endif
-          LOG(ERROR) << "Expected DataShort or DataLong mask, got : " << dataS;
-          return Error;
+          return unexpectedEOF(fmt::format("Expected DataShort or DataLong mask, got {:x}", dataS));
         }
         expectInp = ExpectChipTrailer | ExpectData | ExpectRegion;
         continue; // end of DATA(SHORT or LONG) processing
@@ -415,7 +414,11 @@ class AlpideCoder
   void resetMap();
 
   ///< error message on unexpected EOF
-  static int unexpectedEOF(const std::string& message);
+  static int unexpectedEOF(const std::string& message)
+  {
+    LOG(DEBUG) << message;
+    return Error;
+  }
 
   // =====================================================================
   //
--- a/Detectors/ITSMFT/common/reconstruction/include/ITSMFTReconstruction/DecodingStat.h
+++ b/Detectors/ITSMFT/common/reconstruction/include/ITSMFTReconstruction/DecodingStat.h
@@ -57,11 +57,11 @@ struct ChipStat {
     "Unknow word"                                   // UnknownWord
   };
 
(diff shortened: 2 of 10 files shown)


rpezzi (dismissed):
Thanks, Ruben.

iouribelikov (approved):

shahor02:
merging since the FullCI failure is not related

@davidrohr @sawenzel @ktf : fullCI again fails with `[ERROR] error while setting up workflow: boost::filesystem::status: Operation not permitted: "o2sim"`.
Presumably this happens when checking if `o2sim` is a directory in https://github.com/AliceO2Group/AliceO2/blob/cdb4752294a1bb1dfcec3e1344abc6866b47bc41/Common/Utils/src/StringUtils.cxx#L42
Recently I've changed this part to boost from `sys/stat`. 
If we are sure that the problem is not with permissions on the build server but in boost, perhaps we should revert back to system calls. Then, I guess, this should be done in all places where `boost::filesystem` is used.

davidrohr:
I would actually prefer to understand why it fails in the CI. Running in the CI container locally, I cannot reproduce this, I tried several times. And also in the CI, it fails only sometimes, e.g. here it passed: https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/AliceO2/6078/fe05840a922399d2401056fcd439976162bc06b9/build_O2_fullCI/pretty.html . This irreproducible CI behavior has been a headache for quite some time now, and we we do not understand this, these issues will keep popping up.

shahor02:
@davidrohr OK, but how do think this can be investigated? In this case it fails on querying the status of apparently non-existing directory. This can be either permissions problem or boost problem. If we see that with sys/stat call the problem persists or gone, we can blame either the 1st or the 2nd as a culprit.

davidrohr:
>  `return boost::filesystem::is_directory(std::string{p});`

How are you actually sure it is exactly this line that fails?

davidrohr:
OK, perhaps it is the problem also seen here: https://stackoverflow.com/questions/67256984/boostfilesystemexists-throws-operation-not-permitted-exception-on-centos-8-d

Actually we should just switch to std::filesystem, now that we have gcc 10. I think that should be done anyway, I would not use boost too much for something where there is something in the STL.

shahor02:
The exception says `boost::filesystem::status: Operation not permitted: "o2sim"`, which means some `boost::filesystem` method was queried with this path. This is exactly what happens (and I think this is the only place where it does) when we query NameConf with user provided string and try to figure out if the full path was requested or a just prefix used to build the path according to some conventions.

davidrohr:
ok, probably you are right. I thought there might be other places where boost:filesystem is used. In any case, let's switch to std::filesystem. That is anyway better, and apparently there is a problem with boost filesystem in docker containers without `--privileged` flag. Can you take care of that?

shahor02:
yes, will eliminate all entries of boost:filesystem

PR #6085, opened 2021-05-06
[MCH] add optional protections against high-occupancy DE and/or event

pillot:
In run 1&2 we observed from time to time a high level of electronic noise in some Detection Elements, sometimes correlated across chambers. The current clustering do not really filter out this noise, resulting in a lot of clusters and thus a lot of possible combinations to tests in the tracking, in particular in case of correlated noise across several chambers, which can make the tracking very slow.

These protections allow to discard noisy DE and/or events if too many DE are affected, with the sole purpose to speed up the reconstruction if needed (at the prize of slightly lower efficiency and higher systematics).

The correlated noise is expected to be reduced in run3, and the new clustering might better deal with the remaining noise, so I kept these protections optional.

### DIFF
--- a/Detectors/MUON/MCH/PreClustering/include/MCHPreClustering/PreClusterFinder.h
+++ b/Detectors/MUON/MCH/PreClustering/include/MCHPreClustering/PreClusterFinder.h
@@ -49,6 +49,8 @@ class PreClusterFinder
   void loadDigits(gsl::span<const Digit> digits);
   void loadDigit(const Digit& digit);
 
+  int discardHighOccupancy(bool perDE, bool perEvent);
+
   int run();
 
   void getPreClusters(std::vector<o2::mch::PreCluster>& preClusters, std::vector<Digit>& digits);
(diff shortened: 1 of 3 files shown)


aphecetche:
A very cosmetic comment : I would use all lower-case option name `discard-high-occupancy-des` (generally find less confusing all lower instead of assuming people know what /where to capitalize)

Less cosmetic : you hardcode the "thresholds" of the protections. Wouldn't it be better to e.g. use configurableParam to make them tweakable ?  Or is there really only one set of values that is relevant ?

pillot:
I hardcoded these cuts because I think it is not trivial neither desirable to play with them at the moment. It took me quite some time studying the noise in details to be able to tune this protection (the cuts but also the logic) and I think it is the optimum settings for now.

Since not only the value of these cuts but also the logic and even the need of this protection and the way to address this kind of noise (if any) in run 3 is not yet clear, I preferred to keep it simple for the moment and avoid adding too many parameters until proven to be needed.

Actually I hope this protection won't be needed, because it affects the efficiency in a way that is hard to reproduce in MC (without embedding) and would need dedicated systematic studies. (By the way, it would be nice to be able to propagate some flags in the analysis objects to identify events that trigger such protections in the reconstruction).

aphecetche (approved):

PR #6087, opened 2021-05-07
DPL GUI: improve profiler / debugger support

ktf:
On macOS:

* do not open two terminals when launching debugger / profiler
* avoid spurious opening the wrong profiler stacktrace
* Use xcrun for compatibility with new XCode


ktf:
@aalkin this should fix a couple of things you reported a while ago. Could you check?

ktf:
Now with the correct commit. :-/

ktf:
Only affects __APPLE__ tested on my box. Merging.

PR #6088, opened 2021-05-07
Fix dy calculation in TRD TrackletTransformer

jbarrella:
This PR fixes the calculation and calibration of the deflection length in `TrackletTransformer`. The previous calculations anticipated pre-calibrated incoming tracklets rather than raw ones. With the incoming tracklets being raw, the calculations are much simpler as no reversal of the previous calibration is required. 

@aschmah, @aberdnikova I have three questions that need to be answered to ensure that the new calculation is correct please:
1. What is the distance in local x-direction over which the Lorentz angle is measured? My expectation would be the 3cm of the drift region, but I recall that we used to use 3.35cm which was the height of the anode plane. I'm not certain if either of those are incorrect to use, but I know for sure that we need to be consistent with regard to this.
2. What sign convention do we use for the Lorentz angle? Alex, I sent you a picture of fig. 4.17 from Bailhache's 2009 thesis. My question is basically whether this angle as it is drawn in the picture would be a positive or negative Lorentz angle in the new code. (This is probably something that could easily be verified later on with residuals)
3. I recall that the vDrift value that is calculated in the calibrations is the component of the velocity in x-direction. Is this still the case? Just wanted to confirm.

Please also let me know if I can help in clearing up any of these questions. Thanks!

### DIFF
--- a/Detectors/TRD/base/src/TrackletTransformer.cxx
+++ b/Detectors/TRD/base/src/TrackletTransformer.cxx
@@ -24,18 +24,15 @@ TrackletTransformer::TrackletTransformer()
 
   // 3 cm
   mXCathode = mGeo->cdrHght();
-  // 2.221
-  // mXAnode = mGeo->anodePos();
   // 3.35
   mXAnode = mGeo->cdrHght() + mGeo->camHght() / 2;
-  // 2.5
+  // 5mm below cathode plane to reduce error propogation from tracklet fit and driftV
   mXDrift = mGeo->cdrHght() - 0.5;
   mXtb0 = -100;
 
   // dummy values for testing. This will change in the future when values are pulled from CCDB
   mt0Correction = -0.279;
-  mOldLorentzAngle = 0.16;
-  mLorentzAngle = -0.14;
+  mLorentzAngle = 0.14;
   mDriftVRatio = 1.1;
 }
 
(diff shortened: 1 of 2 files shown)


Detectors/TRD/base/src/TrackletTransformer.cxx
```diff
@@ -77,50 +76,34 @@ float TrackletTransformer::calculateZ(int padrow)
...
+  // NOTE: confirm the direction in which vDrift is measured/determined. Is it in x or in direction of drift?
+  double lorentzCorrection = TMath::Tan(lorentzAngle) * driftHeight;
 
-  // LOG(info) << "ORIGINAL: " << calibratedDy;
-  // LOG(info) << "ALTERNATIVE: " << rawDy + calibrationShift;
+  // assuming angle in Bailhache, fig. 4.17 would be positive in our calibration code
+  double calibratedDy = rawDy - lorentzCorrection;
 
-  return rawDy; // OS: temporary until calibratedDy is checked. Currently it is too far off from rawDy
+  return calibratedDy;
```
martenole: Please don't yet send the `calibratedDy` but the `rawDy` instead until we actually have proper calibration

martenole: Or did you already compare calibrated and raw value and check whether they are in sufficient agreement?

jbarrella: Well I have a Lorentz angle of 8 degrees hardcoded in there which equates to an offset of about -0.4cm over 3cm. I don't really know whether this would be better or worse for your tracking. I think 8 degrees is a reasonable estimate and hopefully the sign is correct...

martenole: Maybe you could just run a small simulation and actually compare both values for `dY` which you get?

jbarrella: Yeah, I did a quick test and all rawDy values are just reduced by ~0.42cm

martenole: Given that dy ranges from -0.9 cm to 0.9 cm that is not a negligible change.. So the tracker would not find any matches if dy is shifted by that much. Only when we have our actual FEE configuration and the TrapSimulator produces the compatible output as the FEE only then we can tune the calibration of the tracklets and use the calibrated values. Until then we have to stick with the raw values. At the moment the TrapSimulator still does some correction to the tracklets

jbarrella:
@aschmah, @aberdnikova just bumping this for your attention. Not a big rush, but would like to merge when possible

jbarrella:
After looking at the calibration code that @aberdnikova shared with me, the answers to questions 1 and 3 seem to be clear and have not changed since I last looked at the code.

1.  It looks like the anode plane height of 3.35 cm is still used. I noticed that Klein fig. 38 seems to measure the Lorentz angle from the cathode plane and Bailhache fig. 4.17 seems to measure from the pad plane. I think the tracklet can only be approximated as a straight line where the velocity is constant which is only in the drift region. I don't think it actually matters, as long as the `TrackletTransformer` is consistent when it applies the correction, but it might be safer to just use 3 cm. We could also check this at a later stage in the residuals maybe.
2. Let's check in the residuals.
3. It seems clear that this is still the case.

bazinski (dismissed):
Approved but as the ci is failing on the full system test which does execute this it should probably pass that before merging. Even though the failure is mem issues of the digitizer in the test, and hence not related.

bazinski (dismissed):

Detectors/TRD/base/src/TrackletTransformer.cxx
```diff
@@ -77,50 +74,33 @@ float TrackletTransformer::calculateZ(int padrow)
...
-  // double x_anode_hit = driftDistance*driftVRatio/cmSlope;
-  // double y_anode_hit = driftDistance*driftVRatio;
-
-  // double x_Lorentz_drift_hit = TMath::Tan(oldLorentzAngle)*driftDistance*driftVRatio - TMath::Tan(lorentzAngle)*driftDistance;
-  // double y_Lorentz_drift_hit = driftDistance*driftVRatio - driftDistance;
-
-  // double Delta_x_Lorentz_drift_hit = x_anode_hit - x_Lorentz_drift_hit;
-  // double Delta_y_Lorentz_drift_hit = y_anode_hit - y_Lorentz_drift_hit;
-  // double impact_angle_rec = TMath::ATan2(Delta_y_Lorentz_drift_hit,Delta_x_Lorentz_drift_hit);
+  double rawDy = slopeSigned * ((mXAnode / vDrift) * 10.) * padWidth * GRANULARITYTRKLSLOPE;
```
martenole: This should be `mXCathode` instead of `mXAnode` I think

martenole (changes requested):

Detectors/TRD/base/src/TrackletTransformer.cxx
jbarrella: I can't recall if you were in the meeting when we talked about this, but we decided to go with the anode height of 3.35 cm for now. This will be consistent with what is used in the calibration code. However, we will likely revisit this issue in the future since there is still disagreement about 3.35 cm being the most sensible value here, but the change needs to happen in calibration code as well.

tdietel (dismissed):
Looks good to me, including the comments on mXAnode vs mXDrift. We can merge this as is, and revisit when we have a better handle on tracklet resolution and residuals.

We should still check the CI issue raised by Sean.

Detectors/TRD/base/src/TrackletTransformer.cxx
martenole: Well, but if you change it here then this will affect the tracking. Couldn't you change this to be a correction factor for `calibratedDy`? For me `rawDy` should be the deflection that is calculated in the TRAP and there the drift length of 3cm is used and not 3.35 cm.

jbarrella: Should be resolved now

martenole: Looks good, thanks

martenole (approved):

jbarrella:
@TimoWilken Hi, can this be merged? I do not think the errors are caused by me.

TimoWilken:
Hi @jbarrella, would it be alright to wait for another fullCI run when the updated copyright notice template is merged into the o2checkcode check? The macOS error is a known problem at the moment, but fullCI should at least get through the builds and unit tests.

martenole:
@davidrohr could you kindly merge this please? fullCI is green and the two errors from macOS and o2 build are not related.
A followup on this PR is needed in any case. And I checked the TRD reco with the changes from here and it is not affected.

PR #6090, opened 2021-05-07
Fix file name building

fcatalan92:
@sawenzel I found an error when trying to load the geometry with `o2::base::GeometryManager::loadGeometry(...)` passing only the directory path
```
[INFO] Loading geometry FAIRGeom from .//o2sim_geometryroot
Error in <TFile::TFile>: file .//o2sim_geometryroot does not exist
[FATAL] Failed to open file .//o2sim_geometryroot
```
This PR should fix the filename creation.

### DIFF
--- a/DataFormats/Detectors/Common/src/NameConf.cxx
+++ b/DataFormats/Detectors/Common/src/NameConf.cxx
@@ -21,7 +21,7 @@ std::string NameConf::buildFileName(const std::string_view prefix, const std::st
                                     const std::string_view extension, const std::string_view optDir)
 {
   if (o2::utils::Str::pathIsDirectory(prefix)) { // if path is directory, just add to default name, ignoring optional directory optDir argument
-    return o2::utils::Str::concat_string(prefix, "/", defPrefix, delimiter, defName, extension);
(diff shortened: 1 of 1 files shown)


shahor02 (approved):
@fcatalan92 thanks, actually it was my bug, merging as the fix is trivial.

shahor02:
PS: for the record, you actually don't need to call `loadGeometry("./")`, the `cwd` is assumed by default.
In the workflows, a non-cwd geometry path can be imposed via `--configKeyValues "mDirGeom=<dirname>"`.
`dirname` = `"none"` is also interpreted as `cwd`. The full list of parameters whose directory can be imposed in the same way is https://github.com/AliceO2Group/AliceO2/blob/cdb4752294a1bb1dfcec3e1344abc6866b47bc41/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/NameConf.h#L126-L128

fcatalan92:
> thanks, actually it was my bug, merging as the fix is trivial.

I checked the git blame and not the history, sorry.

> PS: for the record, you actually don't need to call `loadGeometry("./")`, the cwd is assumed by default.

Thank you for the information, I will update the macro I'm using

PR #6091, opened 2021-05-07
Introduce Resource management

ktf:
In order to limit the amount of resources used by a DPL workflow,
DPL now has the concept of a resource quota that the task can use.
The default behavior is to ignore quota, but this will change to a mode
where a task can run only if enough CPU quota is available.

This should also allow improving the behavior of the shared memory based rate
limiting, by letting the reader know how much it is allowed to read ahead
before pausing to reduce memory pressure (right now there is a race
condition due to the fact that the "memory used" memory might come too late,
for the driver to be able to stop the reader).

### DIFF
--- a/Framework/Core/CMakeLists.txt
+++ b/Framework/Core/CMakeLists.txt
@@ -31,6 +31,7 @@ o2_add_library(Framework
                        src/CommonMessageBackends.cxx
                        src/CompletionPolicy.cxx
                        src/CompletionPolicyHelpers.cxx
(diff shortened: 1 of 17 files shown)


ktf:
@TimoWilken I think there is something wrong with the builders the o2-dataflow one is taking over 4 hours to build.

TimoWilken:
I've been restarting the slc7 builders this evening to move to the new container, I think that'll be why.

ktf:
Ok, thanks. PR updated with actual ResourcePolicy customisation mechanism.

PR #6092, opened 2021-05-07
Multithread version of populateCCDB used at MW2

shahor02:
The comment in https://alice.its.cern.ch/jira/browse/O2-2053 about multi-threaded CCDB upload refers to this fix.


shahor02:
FullCI failure is unrelated, merging

PR #6099, opened 2021-05-07
Add tab detection to CI tests.

vkucera:
A simple GitHub action to find tabs in modified text files

### DIFF
--- a/.github/workflows/tab-finder.yml
+++ b/.github/workflows/tab-finder.yml
@@ -11,20 +11,22 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v2
         with:
-          fetch-depth: 0 # needed to get the full history and pull from upstream
-      - name: Pull from upstream
+          fetch-depth: 0 # needed to get the full history
+      - name: Fetch upstream
         run: |
-          # Update from upstream to avoid including missing commits in the diff
+          # Fetch the main upstream branch to find the common ancestor
           git config --global user.name "Nemo" # required on some servers
-          git config pull.rebase false
           git remote add upstream https://github.com/AliceO2Group/AliceO2.git || exit 1
-          git pull upstream ${{ env.MAIN_BRANCH }} || exit 1
+          git fetch upstream ${{ env.MAIN_BRANCH }} || exit 1
       - name: Find tabs
         run: |
           # Find tabs in modified text files and show where they are
           status=0
+          # Get the common ancestor of the current branch and the main upstream branch
+          BASE_COMMIT=$(git merge-base HEAD upstream/${{ env.MAIN_BRANCH }})
           # loop over changed files
-          for f in $(git diff --name-only upstream/${{ env.MAIN_BRANCH }}); do
+          echo "Diffing against: $BASE_COMMIT"
+          for f in $(git diff --diff-filter d --name-only $BASE_COMMIT); do
             # ignore binary files
             file -bi "$f" | grep -q "charset=binary" && continue
             # find tabs in file
--- a/.github/workflows/tab-finder.yml
+++ b/.github/workflows/tab-finder.yml
@@ -1,0 +1,39 @@
+# Find tabs in modified text files
+name: Tab finder
+on: [push, pull_request]
+env:
+  MAIN_BRANCH: dev
+jobs:
+  build:
+    name: Tab finder
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # needed to get the full history and pull from upstream
+      - name: Pull from upstream
+        run: |
+          # Update from upstream to avoid including missing commits in the diff
+          git config --global user.name "Nemo" # required on some servers
+          git config pull.rebase false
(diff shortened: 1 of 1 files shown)


ktf:
@TimoWilken looks like we have something again deleting /tmp.

.github/workflows/tab-finder.yml
```diff
@@ -0,0 +1,39 @@
...
+          git config --global user.name "Nemo" # required on some servers
+          git config pull.rebase false
+          git remote add upstream https://github.com/AliceO2Group/AliceO2.git || exit 1
+          git pull upstream ${{ env.MAIN_BRANCH }} || exit 1
+      - name: Find tabs
+        run: |
+          # Find tabs in modified text files and show where they are
+          status=0
+          # loop over changed files
+          for f in $(git diff --name-only upstream/${{ env.MAIN_BRANCH }}); do
```
TimoWilken: You should probably use `git diff --diff-filter d` here, to avoid checking deleted files.

I'd also use `"$(git merge-base HEAD ${{ github.event.pull_request.base.sha }})"` instead of `upstream/${{ env.MAIN_BRANCH }}`, so you don't end up checking unrelated changes in dev, which weren't introduced by the PR in question. (See also: [code-formatting.yml](https://github.com/AliceO2Group/AliceO2/blob/d08978eabc0702203c6a6141bdb9650f11cd1a8b/.github/workflows/code-formatting.yml#L44-L48))

.github/workflows/tab-finder.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # needed to get the full history and pull from upstream
+      - name: Pull from upstream
+        run: |
+          # Update from upstream to avoid including missing commits in the diff
+          git config --global user.name "Nemo" # required on some servers
+          git config pull.rebase false
+          git remote add upstream https://github.com/AliceO2Group/AliceO2.git || exit 1
```
TimoWilken, line 19: Is this necessary? Don't we get the remote from the clone step above already?

.github/workflows/tab-finder.yml
```diff
@@ -0,0 +1,39 @@
...
+          git pull upstream ${{ env.MAIN_BRANCH }} || exit 1
```
TimoWilken: Would it be easier to just `git fetch ${{ github.event.pull_request.base.sha }}`? Then we don't need to change the `pull.rebase` setting and deal with any other magic that `git pull` might do.

(Also, wouldn't this try to merge the current `dev` into the PR branch? I can't see why we'd want that.)

TimoWilken (changes requested):

.github/workflows/tab-finder.yml
vkucera: `github.event.pull_request` is available only in pull requests. This way the action runs for a push as well.

.github/workflows/tab-finder.yml
vkucera: Doesn't the checkout action just clone the fork remote? If the main branch in the fork remote is not up to date, diff against upstream would include unrelated files.

.github/workflows/tab-finder.yml
vkucera: Again, `github.event.pull_request` is not available for a push event. This is the simplest way I could figure out to get files that are different from upstream/dev.

.github/workflows/tab-finder.yml
TimoWilken: Ah, OK. How about using the merge-base of dev and the HEAD of the PR or push? I still think that'd be the correct thing to do for PRs, but does that cause problems for plain pushes?

.github/workflows/tab-finder.yml
TimoWilken: OK. In that case, I'd use `git fetch upstream ${{env.MAIN_BRANCH}}` to avoid the auto-merging behaviour of `git pull`. That gives you `upstream/${{env.MAIN_BRANCH}}` locally.

.github/workflows/tab-finder.yml
TimoWilken: Good point, I'm not sure about that.

.github/workflows/tab-finder.yml
vkucera: I had `git fetch` in the first version. I cannot remember now why I changed to `git pull`... :-)

vkucera: Probably because of the unrelated files that will appear in the diff if the missing commits are not merged in the current branch.

vkucera: But if you prefer, I can replace the `pull` with `fetch` + `merge`. I don't know much the details about the extra magic that happens with `pull`.

TimoWilken: > Probably because of the unrelated files that will appear in the diff if the missing commits are not merged in the current branch.

Yes, that's why I suggested diffing against the merge-base instead of the current dev.

.github/workflows/tab-finder.yml
vkucera: Done

TimoWilken (approved):

PR #6102, opened 2021-05-08
boost::filesystem -> std::filesystem

shahor02:


davidrohr:
I can reproduce the issue in the full system test with this PR locally, the failure is actually already during the SYNC stage which does not start up and doesn't produce output. I have opened a Jira to follow up the independent problem that this did not lead to a stop in the FST: https://alice.its.cern.ch/jira/browse/O2-2315

davidrohr:
@shahor02 : I ran this locally several times now. What happens is that after this PR the raw readers can sometimes not create their output folder, e.g. this is the output from the FDD raw reader:
```
Running: TIME="#walltime %e" /home/qon/docker/sw/slc8_x86-64/O2/dev-local2/share/scripts/monitor-mem.sh /usr/bin/time --output=fddraw.log_time './fddraw.log_tmp.sh'
/usr/bin/time --output=fddraw.log_time ./fddraw.log_tmp.sh
[FATAL] could not create output directory raw/FDD/
For later analysis we write a core dump to core_dump_98396
./fddraw.log_tmp.sh: line 1: 98396 Aborted                 (core dumped) o2-fdd-digit2raw --file-per-link -o raw/FDD
TASK-EXIT-CODE: 0
Job finished; Exiting 
PROCESS MAX MEM = 1293.73
PROCESS AVG MEM = 663.93
PROCESS MAX FMQ_SHM = 2.23065e-29
```
However, it is not deterministic, sometimes the raw/FDD folder is created and filled correctly. And even in this case where it reports it cannot create the folder, the folder is actually created but empty.

As far as I see the situation is the same for the other detectors, though I didn't check whether all are affected.

@sawenzel : It seems the taskwrapper did also not catch these errors in the raw writer, could you have a look?

shahor02:
@davidrohr it was a stupid bug of mine, due to the succes return value of `fs::create_directories` and `gSystem->mkdir` (now changed to fs) having opposite conventions... Will post a fix after some checks.

davidrohr:
ok, thx, but it was actually still good to see that the errors during the FST were not catched correctly, so we should follow up the JIRAs I have created anyway.

davidrohr:
@TimoWilken : Could you restart O2/o2?

davidrohr:
Now it got past the memory issue, but this seem to be a real issue created by this PR:
```
In file included from input_line_106:1:
/mnt/mesos/sandbox/sandbox/o2/O2/Detectors/Base/test/buildMatBudLUT.C:52:25: error: no member named 'exists' in namespace 'std::filesystem'
  if (!std::filesystem::exists(geomName)) { // if needed, create geometry
```
Actually, I am not sure if / how much of std::filesystem is supported by ROOT cling. I think cling is currently still at LLVM version 5.

davidrohr:
Full CI has passed before, this finally seems to work, merging.

Barthelemy:
It broke the `o2-dataflow` build (also here it is red).

awegrzyn:
@ktf This is perfect example why we we don't want to use `O2@dev` in alidist

davidrohr:
Actually, the problem is rather that the CI was not working correctly, and o2-dataflow was red all the time without log.
And sorry, I didn't recheck when I finally merged it. Indeed, it had a real error then.
Another point is that o2-dataflow is not marked as required for O2 PRs currently...

But so what do we do?
The problem is that o2-dataflow uses the old GCC, std::filesystem is not available in GCC 7.
Due to some bug with boost / docker / cc8, the old implementation broke the full system test in docker containers. This PR fixed that. So if we revert this, we break the full CI again.
Basically in the long run we have to bump to GCC 10 for o2 dataflow anyway. Even if you move to CS8, the system compiler will not be GCC 10, so I think there is no way around using the alidist GCC-Toolchain in the long run.

So what stops us from using GCC 10 in o2-dataflow?
We can revert this here temporarily and discuss tomorrow in the CWG3 what to do, but keeping this reverted is no fix in the long run.

davidrohr:
ping @awegrzyn @Barthelemy in case you are not subscribed to this thread.

Barthelemy:
@davidrohr I understand and agree with your analysis. Let me discuss with my colleagues and I come back to you. On the long term it is clear that we will move to GCC 10, we had agreed to it as `o2-dataflow` binary tarballs are now produced. However, we had in mind the next FLP Suite (?).

sawenzel:
I think using 2 largely different compiler versions should be absolutely avoided in a common software project. I would be in favour of using the same GCC everywhere, or else we will run into these kind of troubles over and over again.

davidrohr:
@Barthelemy : ok, thx, shall we already open an alidist pr, just to check whether the bump to gcc10 for o2-dataflow would actually work?

When will the next FLP suite be out? I think reverting this for few days will be annoying since the FullCI will be red, but should be ok. If it would take longer, I would prefer to try to bump the GCC :)
Just let me know.

Barthelemy:
@davidrohr Adam @awegrzyn is preparing the PR. We can survive as it is, just a check failing QC that we will ignore. Let's see how it goes !

Barthelemy:
@sawenzel yes, we tried to avoid it :) the problem is that sometimes our schedules are difficult to align.

ktf:
@awegrzyn no one is forcing you to use O2@dev. You have one branch per day of the week which you can use and fork and base your own CI on. This is actually exactly why we should use dev for the CI, so that we spot issues early.

@Barthelemy for the binary tarballs, we have them, provided you use the default centos:7 installation which comes with python as default (python3 ones are coming).

PR #6105, opened 2021-05-08
jobutils: Fix propagation of return codes

sawenzel:
A recently introduced bug is fixed. Unfortunately,
$? was not escaped correctly, leading to a premature expansion
if is value.


sawenzel:
@davidrohr

davidrohr:
I ran this locally together with #6102, and now we get propper errors reported.
Thanks, merging.

PR #6107, opened 2021-05-08
Cleanup for TRD tracking

martenole:
- `GPUTRDTracker` does not do any MC processing anymore which was used for a simplified performance analysis in AliRoot. For O2 this analysis has to be done in a separate step after the tracking (to be added later)
- Cleanup for GPUTRDTrack type
- TRD tracks without any tracklets attached are dropped


Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -124,7 +124,7 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
   for (int iTrk = 0; iTrk < mChainTracking->mIOPtrs.nTracksTPCITSO2; ++iTrk) {
     const auto& trkITSTPC = mChainTracking->mIOPtrs.tracksTPCITSO2[iTrk];
     GPUTRDTrack trkLoad(trkITSTPC, mTPCVdrift);
-    if (mTracker->LoadTrack(trkLoad, -1, nullptr, -1, iTrk)) {
+    if (mTracker->LoadTrack(trkLoad, trkITSTPC.getRefTPC())) {
```
davidrohr: Actually, I would not use getRefTPC() here, this should be the id of the reference track used for the matching, and if that was a TPC-ITS track, I would point to that one.

davidrohr (commented):
Without checking in detail, looks OK, just one comment:

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
martenole: Not sure if that is better. This track will end up as `MATCHTRD_GLO`, i.e. as ITS-TPC-TRD track and therefore it is clear that it was from a TPC-ITS seed. If I change it to
```suggestion
    if (mTracker->LoadTrack(trkLoad, iTrk)) {
```
then I don't know after the tracking is done which seeding type each TRD track had. Although in `GPUTRDTrack::mIsFindable` there is an unused bit which I could use to distinguish between TPC-only and ITS-TPC. Eventually I think I will use the RecoContainer logic for that.

davidrohr: But for the further processing one needs to know which TPCITS track it belongs to. And the mapping it TPCITS --> TPC, i.e. if you store the TPC id, that would require a revert search.

I think we should use the GlobalTrackID. We can still store it as the 32 bit integer in the GPUTRDTrack, since GlobalTrackID is an abstractRef, which is just 32 bit. You can just add a getter and setter in the TrackTRD, which converts it forth and back. (with a `static_assert(sizeof(GlobalTrackId) == sizeof(whatever we use))')

Then you can get the seedType via the GlobalTrackId::getSource()

davidrohr: And you would not need to cast, actually, just use GlobalTrackId::getRaw() and setRaw()

martenole: Sounds good, I will try that. I just realized now that since `class GlobalTrackID : public AbstractRef<25, 5, 2>` the `AbstractRef::Base_t` is a 25 + 5 + 2 = 32 bit integer. That is not super obvious when one only takes a very quick look at those classes

martenole: But now the `TrackTPCITS` has two `GlobalTrackID`s, one for the ITS track and one for the TPC track. There is no `GlobalTrackID` for the ITS-TPC match, or is there?

davidrohr: You just use `GlobalTrackId{iTrk, GlobalTrackID::ITSTPC}` to construct one, this will refer to the ITS-TPC track. If the user wants to go to the TPC track, he has to follow the ids.

davidrohr: Same for the TPC track, just `GlobalTrackId{iTrk, GloballTrackID::TPC}`

martenole: Is https://github.com/AliceO2Group/AliceO2/pull/6107/commits/aa3b2a368bbfef8473bc29e0d9946f2f9d31e909  what you meant?

GPU/GPUTracking/DataTypes/GPUTRDTrack.h
```diff
@@ -71,57 +78,55 @@ class GPUTRDTrack_t : public T
...
-  GPUd() int GetNmissingConsecLayers(int iLayer) const;
-
-  GPUd() void AddTracklet(int iLayer, int idx)
+  // attach a tracklet to this track; this overwrites the mIsFindable flag to true for this layer
+  GPUd() void addTracklet(int iLayer, int idx) { mAttachedTracklets[iLayer] = idx; }
+
+  // getters
+  GPUd() int getNlayersFindable() const;
+  GPUd() int getTracklet(int iLayer) const { return mAttachedTracklets[iLayer]; }
+#ifdef HAVE_O2HEADERS
```
davidrohr: This is not so save, since the HAVE_O2HEADERS is not in the PUBLIC definitions.
I would put this function inside of `TrackTRD.h`.

And I would rename it into getReferenceTrackGlobalId(), since it is not the global Id of this track, but of the TPC or TPC-ITS reference. You can also put a comment `// return GlobalID of reference track in its original container`.

davidrohr: OK, thinking about that, it is not so simple, since `TrackTRD.h` contains only an alias using the `using` keyword.
What you can do is: You place only the declaration of the function here in GPUTRDTrack.h. You might need to forward-declare GlobalTrackId. Then you can put the function body as inline function inside TrackTRD.h. This way you make sure it is always available for O2, but not otherwise.

davidrohr (changes requested):

GPU/GPUTracking/DataTypes/GPUTRDTrack.h
martenole: But if I do that then how can I define the function for `o2::gpu::GPUTRDTrack_t<o2::gpu::trackInterface<o2::gpu::GPUTPCGMTrackParam> >`?

And also it does not work for the O2 track type for some reason:

`/home/oschmidt/alice/sw/SOURCES/O2/gputrdtracking/0/DataFormats/Detectors/TRD/include/DataFormatsTRD/TrackTRD.h:30:48: error: definition of 'GTrackID o2::gpu::GPUTRDTrack_t<o2::gpu::trackInterface<o2::gpu::GPUTRDO2BaseTrack> >::getReferenceTrackGlobalId() const' is not in namespace enclosing 'o2::gpu::GPUTRDTrack_t<o2::gpu::trackInterface<o2::gpu::GPUTRDO2BaseTrack> >' [-fpermissive]
   30 | GTrackID TrackTRD::getReferenceTrackGlobalId() const
      |                                                ^~~~~
/home/oschmidt/alice/sw/SOURCES/O2/gputrdtracking/0/DataFormats/Detectors/TRD/include/DataFormatsTRD/TrackTRD.h:30:10: error: no declaration matches 'GTrackID o2::gpu::GPUTRDTrack_t<o2::gpu::trackInterface<o2::gpu::GPUTRDO2BaseTrack> >::getReferenceTrackGlobalId() const'
   30 | GTrackID TrackTRD::getReferenceTrackGlobalId() const
      |          ^~~~~~~~
In file included from /home/oschmidt/alice/sw/SOURCES/O2/gputrdtracking/0/DataFormats/Detectors/TRD/include/DataFormatsTRD/TrackTRD.h:17,
                 from /home/oschmidt/alice/sw/SOURCES/O2/gputrdtracking/0/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h:27,
                 from /home/oschmidt/alice/sw/SOURCES/O2/gputrdtracking/0/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx:17:
/home/oschmidt/alice/sw/SOURCES/O2/gputrdtracking/0/GPU/GPUTracking/DataTypes/GPUTRDTrack.h:91:41: note: candidate is: 'o2::dataformats::GlobalTrackID o2::gpu::GPUTRDTrack_t<T>::getReferenceTrackGlobalId() const [with T = o2::gpu::trackInterface<o2::gpu::GPUTRDO2BaseTrack>]'
   91 |   GPUd() o2::dataformats::GlobalTrackID getReferenceTrackGlobalId() const;
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~
/home/oschmidt/alice/sw/SOURCES/O2/gputrdtracking/0/GPU/GPUTracking/DataTypes/GPUTRDTrack.h:59:7: note: 'class o2::gpu::GPUTRDTrack_t<o2::gpu::trackInterface<o2::gpu::GPUTRDO2BaseTrack> >' defined here
   59 | class GPUTRDTrack_t : public T
      |       ^~~~~~~~~~~~~`

And when I change the definition inside `TrackTRD` to `o2::dataformats::GlobalTrackID o2::gpu::GPUTRDTrack_t<T>::getReferenceTrackGlobalId() const` then I still get a compiler error:
`/home/oschmidt/alice/sw/SOURCES/O2/gputrdtracking/0/DataFormats/Detectors/TRD/include/DataFormatsTRD/TrackTRD.h:30:65: error: definition of 'GTrackID o2::gpu::GPUTRDTrack_t<T>::getReferenceTrackGlobalId() const' is not in namespace enclosing 'o2::gpu::GPUTRDTrack_t<T>' [-fpermissive]
   30 | GTrackID o2::gpu::GPUTRDTrack_t<T>::getReferenceTrackGlobalId() const
      |                                                                 ^~~~~
ninja: build stopped: subcommand failed.`

My declaration inside `GPUTRDTrack.h` looks as follows:
`GPUd() o2::dataformats::GlobalTrackID getReferenceTrackGlobalId() const;`
Is the problem the `GPUd()` macro?

davidrohr:
Without checking in too much detail, it should work. The syntax would be like what I did in
https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/GPUTracking/DataTypes/GPUTRDInterfaceO2Track.h#L88
and
https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/GPUTracking/TRDTracking/GPUTRDInterfaces.h#L147

martenole:
@davidrohr I could not define `getReferenceTrackGlobalId()` inside `TrackTRD`, because that function is also required by the `GPUTRDTracker` and that does not now `TrackTRD`. So I put the definition in `GPUTRDTrackO2.cxx` instead. And for `GPUTRDTrack_t<trackInterface<GPUTPCGMTrackParam>>` I defined it in `GPUTRDTrack.cxx`. For O2 and for AliRoot this works. I have not tested the GPU version, but I will see what the CI says.

davidrohr:
Hi @martenole , I think I have to apologize, with my suggestion we created some kind of mess. Also this PR becomes convoluted I would propose we clean up the global ID thing finally, and then we merge it, and continue in a follow-up PR.

So what I do not like are the varying definitions depending on the defines, and the fact that we use a global id, which is a combination of source and trackid, as single id. I think we also misunderstood each other a bit. I'll try to explain here a bit better what I think that we can do, but we can also have a zoom meeting tomorrow to discuss it (just to avoid that you waste time solving compile issues :))

(As is now, I see still some things I do not like, e.g.
```
mTRDTrackIds[mIOPtrs->trdTracks[i].getReferenceTrackGlobalId()] = i;
```
in the Display class uses the global id as index. Then
```
#ifdef GPUCA_ALIROOT_LIB
  GPUd() int getReferenceTrackGlobalId() const
  {
    return mRefGlobalTrackId;
  }
#else
  // returns GlobalTrackID of the reference track in its original container
  GPUd() o2::dataformats::GlobalTrackID getReferenceTrackGlobalId() const;
#endif
```
breaks the standalone benchmark for Run 2 data, since the `GPUCA_ALIROOT_LIB` is not set. And then we need the additional version of the `getReferenceTrackGlobalId` for `GPUTRDO2BaseTrack`, which I would also try to avoid.
```
GPUdi() o2::dataformats::GlobalTrackID GPUTRDTrack_t<trackInterface<GPUTRDO2BaseTrack>>::getReferenceTrackGlobalId() const
```
)

Then 2 minor comments:
- currently sometimes it is called TrackGlobalId and sometimes GlobalTrackId, I would make that consistent.
- I am not sure what the `>= 0` check here does in `LoadTrack(...)`, but I would remove it:
```
mTracks[mNTracks].setReferenceGlobalTrackId(tpcTrackId >= 0 ? tpcTrackId : mNTracks);
```
and also remove the default parameter for the `tpcTrackId = -1`, but force the user to provide one.

Ok, but to conclude on the main topic with the global id: I would disentangle the integer id and the global id. I would put 2 setters and getters each, say `getReferenceIdRaw` / `setReferenceIdRaw` / `getReferenceIdGlobal` / `setReferenceIdGlobal`. The raw versions would work with integers, the global versions with GlobalTrackId. All would be declared in `GPUTRDTrack.h`, but only the raw versions would be defined in `GPUTRDTrack.h`. The global versions I would only define in `TrackTRD.h` as I said earlier. Consequently it could only be used if the user uses the `o2::trd::TrackTRD`, not in the GPU code. But isn't that sufficient? The global id would only be set by the TRD workflow initially, and then at the end it is read again in the TRD workflow to decide whether it was a TPC or a TPC-ITS seed. Everywhere else it is not used (except for debug output), so there (like in the copy constructors), one can just use the raw integer version.

We cannot properly hide the `getReferenceIdGlobal` `setReferenceIdGlobal` from the other trd track types (well we could, but that would again be annoying template magic), but if we just only declare them in the GPUTRDTrack.h, but then we define it in `TrackTRD.h`, this will give a compile error when it is used. We can leave a comment in GPUTRDTrack.h that it is only intended to be used with `TrackTRD`, then it should be clear and save. And in this way (I think) we can get rid of the #ifdefs.

martenole:
Hi @davidrohr ok thanks a lot, now that also makes a lot more sense to me. I have implemented all the changes and named the ID consistently to GlobalTrackId, since that is also the name for the O2 class. To be precise there it is GlobalTrackID, but I thought `getRefGlobalTrackIdRaw` looked better than `getRefGlobalTrackIDRaw`.
If this runs through the CI I think the last two commits should be squashed before merging. Should I do that? I thought it might be helpful not to do it immediately so that you can easier see what I have changed now.

GPU/GPUTracking/DataTypes/GPUTRDTrack.h
```diff
@@ -81,15 +81,9 @@ class GPUTRDTrack_t : public T
...
-  GPUd() int getReferenceTrackGlobalId() const
-  {
-    return mRefGlobalTrackId;
-  }
-#else
-  // returns GlobalTrackID of the reference track in its original container
-  GPUd() o2::dataformats::GlobalTrackID getReferenceTrackGlobalId() const;
-#endif
+  GPUd() unsigned int getRefGlobalTrackIdRaw() const { return mRefGlobalTrackId; }
+  // This method is only define in TrackTRD.h and is intended to be used only with that TRD track type
```
davidrohr: typo: defined

davidrohr (dismissed):
looks much better, please squash the commits

davidrohr:
@martenole : Could you please also add this patch:
```
diff --git a/GPU/Workflow/helper/src/GPUWorkflowHelper.cxx b/GPU/Workflow/helper/src/GPUWorkflowHelper.cxx
index 4520dc3fb08..684fc089e89 100644
--- a/GPU/Workflow/helper/src/GPUWorkflowHelper.cxx
+++ b/GPU/Workflow/helper/src/GPUWorkflowHelper.cxx
@@ -187,10 +187,10 @@ std::shared_ptr<const GPUWorkflowHelper::tmpDataContainer> GPUWorkflowHelper::fi
   recoCont.createTracksVariadic(creator);
   if (maskTrk[GID::TPC]) {
     for (unsigned int i = 0; i < ioPtr.nTRDTracksTPCTRD; i++) { // TODO: This should be handled by the createTracks logic, but so far it lacks the TRD tracks
-      retVal->tpcLinkTRD[ioPtr.trdTracksTPCTRD[i].GetTPCtrackId()] = i;
+      retVal->tpcLinkTRD[ioPtr.trdTracksTPCTRD[i].getRefGlobalTrackId().getIndex()] = i;
     }
     for (unsigned int i = 0; i < ioPtr.nTRDTracksITSTPCTRD; i++) {
-      retVal->tpcLinkTRD[ioPtr.tracksTPCITSO2[ioPtr.trdTracksITSTPCTRD[i].GetTPCtrackId()].getRefTPC().getIndex()] = i | 0x40000000;
+      retVal->tpcLinkTRD[ioPtr.tracksTPCITSO2[ioPtr.trdTracksITSTPCTRD[i].getRefGlobalTrackId().getIndex()].getRefTPC().getIndex()] = i | 0x40000000;
     }
   }
   ioPtr.globalTracks = retVal->globalTracks.data();
```
Otherwise my reader crashes trying to parse the new IDs incorrectly.

martenole:
Sorry I missed that. It's fixed now.

shahor02:
Hi @martenole 

Should not we merge TPC-TRD and ITS-TPC-TRD outputs to 1 vector? I don't really see advantage of splitting in 2.
https://github.com/AliceO2Group/AliceO2/blob/d6cfe5268ac7570a9cd1fdf1ddcbfb1c21ec5e1d/Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx#L213-L218
Also, TrackTRD does not contain time info and I don't see where do you push trigger info (although you define it in the specs).

davidrohr:
> Hi @martenole
> 
> Should not we merge TPC-TRD and ITS-TPC-TRD outputs to 1 vector? I don't really see advantage of splitting in 2.
> 
> https://github.com/AliceO2Group/AliceO2/blob/d6cfe5268ac7570a9cd1fdf1ddcbfb1c21ec5e1d/Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx#L213-L218
> 
> Also, TrackTRD does not contain time info and I don't see where do you push trigger info (although you define it in the specs).

Indeed. Actually now that we have the GlobalTrackId in, I agree we can just merge the output. Although I am not fully sure what to do in case we run first tpc-trd matching, then tpc-its, and then trd-global matching. Do we want to overwrite the root file, or then use a different file name? Let's discuss this beforehand. Also, when requesting the inputs in the RecoContainer, if you merge it to one stream you can only request all at once. In any case, I'd go ahead with the PR as it is now, and then continue in a new one.

davidrohr:
@martenole : There is actually one thing I totally forgot about, but also for the next PR! Since we copy the tracks in the end anyway (at least currently, we should see if we merge the output, then we could just always write in the output buffer directly), we could as well do the filter for the ntracklets in the output step. That would avoid the additional copy before.

shahor02:
I would never overwrite the root files, just write to different names, switched automatically depending if ITS is present in the input or not.

davidrohr:
But then, if the reader want to read the file with TPC-TRD matches, those could be either in the filename for without its, or in the name with its? Or we should distinguish the file names according to phase, strict phase tpc-its for constraining in 1 file, global matches in other file (containing both TPC-TRD and ITS-TPC-TRD), is that what you mean? We would still need some logic in the reader.
And actually, the same holds for TOF matching, doesn't it?

shahor02:
Yes, difference file names will need to be given for the input of 1st and second stage. Whether we want to have for that some programmed algo or rely on scripts, we can discuss. But this is actually does not depend on split vs joint output question. In any case the TPC-TRD files at second stage are not the same as in 1st one.
Yes, this problem is there also for TOF which now writes different files depending on TPC-only or TPC-ITS input

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

martenole:
Hi @shahor02 

> Also, TrackTRD does not contain time info and I don't see where do you push trigger info (although you define it in the specs).

I can add the time as a member to `GPUTRDTrack` which would be set to the TRD trigger time. In principle one could also load the trigger records and then from the tracklet indices attached to the TRD tracks determine the interaction. But it is probably more efficient to just add it as a member to the track itself. Do I really define the trigger records output in the specs? I think it is only used as input. The trigger records will be provided by the TRD raw reader.

davidrohr:
Hm, but getting the trigger record from the tracklet is a bit annoying I believe, since the tracklet doesn't point to the record. Instead the record points to a tracklet range. So one would have to find out which particular range the traklet belongs to by searching, or do I miss something?
I am not really against adding the time to the TRD track. In order to save some memory, I would store it as number of BC since the beginning of the TF, I wouldn't store the full interaction record.
Alternatively, we could store the index of the interaction record in the interaction-record container in the track. Then one could get to the correct record correctly, but one would still need to load the records. Is there anything else we might need from the record besides the time?

martenole:
Concerning the split output of the TRD tracks and copying of the tracks:
The input tracks are currently copied to the TRD tracker one by one via the `GPUTRDTracker::LoadTrack()` method. After the tracking is done the tracks are filtered and there is some copying involved. This I assume you mean can be avoided if we do the filtering in the `TRDGlobalTrackingSpec` right @davidrohr? Then I will add this (in a followup PR?).
And I would then put all tracks in a single vector and change the TRD track writer so that in case both ITS-TPC and TPC or only ITS-TPC is presend as input I write it to a file called `trdmatches_itstpc.root` and otherwise in case only TPC tracks are provided as input I write it to `trdmatches_tpc.root`. How to distinguish between first and second stage is another question then

martenole:
> So one would have to find out which particular range the traklet belongs to by searching, or do I miss something?

No, that is how I had in mind that one can find the time.

> Is there anything else we might need from the record besides the time?

I don't see anything, because next to the time there is only the tracklet range and this is not needed if the time is known.

davidrohr:
> This I assume you mean can be avoided if we do the filtering in the `TRDGlobalTrackingSpec` right @davidrohr? Then I will add this (in a followup PR?).

Yes, this is what I meant. 

But before we change the outputs, I would try to understand a bit better how the tracks shall be fetched by the RecoContainer and by the InputHelper. If we write both TPC-TRD matches and ITS-TPC-TRD matches from the second stage to trdmatches_itstpc.root, then how do we fetch them. Currently we do this by the source mask. So will then GID:TPC-TRD and GID:ITS-TPC-TRD both fetch the same container from the same file. But then, how is getTracksTPCTRD and getTracksITSTPCTRD supposed to work. They are returning a container currently. We could introduce a new source TRDGlobal and TOFGlobal, for the global matches for TOF and TRD respectively in the final stage? The one could get the container getTracksTRDGlobal, and that would contain all matches in the final stage? Or @shahor02 how did you foresee this?

davidrohr:
> I don't see anything, because next to the time there is only the tracklet range and this is not needed if the time is known.

OK, then my preference is to add the time to the TRD track, but each option would work.

shahor02:
Hi, 

> Do I really define the trigger records output in the specs? I think it is only used as input. The trigger records will be provided by the TRD raw reader.

I've indeed confused the input and output, but the trigger record for tracks is anyway needed, we don't want to loop over the tracks to pick this belonging to the given trigger.
I am not sure the time data member is needed in the tracks: as for the ITStracks which are referred by the readout-frame-record, the TriggerRecord would be fully enough: here is how the time is accessed for the ITS: https://github.com/AliceO2Group/AliceO2/blob/5104f129879520fdda46b60ff8b7047a17c35a80/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L145-L150

> But then, how is getTracksTPCTRD and getTracksITSTPCTRD supposed to work. 

In principle, we don't need to have them separated and can merge these 2 IDs to a single one, but ok, let's at the moment keep it as it is and just add a time info (at least the TriggerRecord with RangeReference to tracks).

davidrohr:
> (at least the TriggerRecord with RangeReference to tracks)

Ah, that is actually also an option, to have the reference the other way around. But then we would need to sort the TRD tracks after trigger records, which is in principle OK I think. Since we have the filter / copy in the output anyway, we could sort an index array, and the copy according to the sorted indizes. But it that case, still one could not easily get the time of a particular track.

shahor02:
> But it that case, still one could not easily get the time of a particular track.

This is true and so far it works for the ITS: they are accessed either in the loop over the ROFRecord or the CreateTracks... provides their time.

davidrohr:
> This is true and so far it works for the ITS: they are accessed either in the loop over the ROFRecord or the CreateTracks... provides their time.

OK, if we converge that this is sufficient, I think it is the best option, since it will need the least memory and also since it will coincide with how ITS does it.

martenole:
Ok, so in summary what I will do is:

- create a trigger record for the TRD tracks with two members: 1) interaction records for the timing information and 2) a range reference for the TRD tracks
- this `TrackTriggerRecord` will be created and output by the `TRDGlobalTrackingSpec` with the help of an index array sort
- the TRD track does not get an additional member with timing information

Is that correct?

davidrohr:
Sounds good to me, + my comment here: https://github.com/AliceO2Group/AliceO2/pull/6114#pullrequestreview-655071329

shahor02:
Yes, looks fine for me too.

martenole:
Thinking a bit more about this. In order to assign BC information to the TRD tracks I want to use the BC information from the TRD tracklets which is stored in the trigger records. In order to avoid having to search for the collision index based on the tracklets attached to the TRD tracks I would like to add a `short mCollisionId` to the `GPUTRDTrack` which is set when a tracklet is added to the track by the GPUTRDTracker. At that point the collision ID is available and this can then be used in the global tracking spec to first easily sort the tracks and second copy the relevant BC information for that collision from the trigger records to the new `TrackTriggerRecord`.
Would that be ok to add another short member to the TRD track?

davidrohr:
I do not see a problem with this. The question is then whether you want a short for the index in the trigger records, or just an in with BCs since start of TF.

martenole:
I would prefer the index, since that is directly available in the GPUTRDTracker already

shahor02:
I also would prefer a `short` index within the TF (referring on the TriggerRecord entry) on top of the TriggerRecords.

PR #6108, opened 2021-05-08
Use createTracksVariadic instead of createTracksWithMatchingTimeInfo

shahor02:
* Add some isXXXTrack template aliases to RecoContainerCreateTracksVariadic
  + fix to use isTPCTrack alias in GPUWorkflowHelper

*  Use createTracksVariadic instead of createTracksWithMatchingTimeInfo
    TPC-ITS matcher
    Cosmics matcher
    PVertexer
    PVertex - track matcher

### DIFF
--- a/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
+++ b/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
@@ -43,7 +43,7 @@ void o2::globaltracking::RecoContainer::createTracksVariadic(T creator) const
   // which depends on the track time:
   // 1) For track types containing TimeStampWithError ts it is ts.getTimeStamp(), getTimeStampError()
   // 2) For tracks with asymmetric time uncertainty, e.g. TPC: as mean time of t0-errBwd,t+errFwd and 0.5(errBwd+errFwd), all in TPC time bins
-  // 3) For tracks whose timing is provided as RO frame: as time in \mus for RO frame start since the start of TF, half-duration of RO window and 0.
+  // 3) For tracks whose timing is provided as RO frame: as time in \mus for RO frame start since the start of TF, half-duration of RO window.
 
   auto start_time = std::chrono::high_resolution_clock::now();
   constexpr float PS2MUS = 1e-6;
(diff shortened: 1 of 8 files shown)


DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
```diff
@@ -162,3 +162,45 @@ void o2::globaltracking::RecoContainer::createTracksVariadic(T creator) const
   auto current_time = std::chrono::high_resolution_clock::now();
   LOG(INFO) << "RecoContainer::createTracks took " << std::chrono::duration_cast<std::chrono::microseconds>(current_time - start_time).count() * 1e-6 << " CPU s.";
 }
+
+template <class T>
+inline constexpr auto isITSTrack(const T&)
+{
+  return std::is_same_v<T, o2::its::TrackITS>;
+}
+
```
davidrohr: This makes sense!

davidrohr (dismissed):

davidrohr:
Ah, unfortunately this fails with
```
/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6108/0/GPU/Workflow/helper/src/GPUWorkflowHelper.cxx:149:5: error: 'trk' is not a constant expression
  149 |     if constexpr (isTPCTrack(trk)) {
```
I have no really good idea how to solve this. One could use
`template <class T> isTPCTrack() ...` and then `isTPCTrack<decltype(trk)>()`, that should work safely. I would propose to add a `std::decay_t<...>` in that case, for references, ptrs, etc.

shahor02:
I am also scratching my head, somehow on ubuntu this worked, although should not... A trivial solution would be to add in 
RecoContainerCreateTracksVariadic a macro and redefine the aliases 
````
#define GetType(t)  std::decay_t<decltype(t)>

template <class T>
inline constexpr auto isTPCTrack()
{
  return std::is_same_v<T, o2::tpc::TrackTPC>;
}
````
Then to call in the creator
````
if constexpr (isTPCTrack<GetType(trk)>()) {
   time = trk.getTime0();
}
````

More rigorous solution would be to equip each track types by corresponding isXXX getter, e.g. for TrackTPC add a method 
````
inline static constexpr auto isTrackTPC() { return true; }
````
I think this should work with sfinae. Then these aliases in RecoContainerCreateTracksVariadic are not needed at all.
What do you think?

**Comment by davidrohr** [comment]
I would not use a macro with a common name such as GetType, since `GetType` is certainly used somewhere in O2. Defining that in a header is quite dangerous in my opinion. Actually I meant putting the dacay_t inside the function. I.e.

```
class RecoContainer {
[...]
template <class T> inline constexpr bool isTPCTrack() { return std::is_same_v<std::decay_t<T>, o2::tpc::TrackTPC>; }
};

[...]
auto creator = [...] {
  if constexpr (isTPCTrack<decltype(trk)>()) { [...] }
}
```
It is actually similar compared to what you have, but without the macro.

shahor02:
yes, but it still requires adding decltype in all creators. In't per track `constexp auto isXXX...` better?

davidrohr:
OK, you can try, but I don't see how it will work out of the box with sfinae. sfinae will usually mean that a template instantiation is dropped, don't know if that can apply to an `if constexpr`. One could of course define all those constexpr `isXXXTrack()` with false in TrackParCov, but I don't like that too much either.

shahor02:
o2 CI failures are not related, looks like boost::fs  problems in a few places.
Merging

PR #6109, opened 2021-05-08
MCH: introduce RoFRecord in decoder

aferrero2707:
The MCH RoF are designed to group all SAMPA digits that have exactly the same time stamp.
Further grouping of the digits into time clusters is delegated to a later stage of the processing pipeline.


Detectors/MUON/MCH/Raw/Decoder/src/UserLogicElinkDecoder.h
```diff
@@ -186,9 +186,15 @@ bool UserLogicElinkDecoder<CHARGESUM>::append10(uint10_t data10)
         if (isSync(mSampaHeader.uint64())) {
           reset();
         } else if (mSampaHeader.packetType() == SampaPacketType::HeartBeat) {
-          sendHBPacket();
-          transition(State::WaitingHeader);
-          result = true;
+          if (mSampaHeader.isHeartbeat()) {
```
aphecetche: this change is unrelated to this PR, isn't it ?

Detectors/MUON/MCH/Workflow/src/DataDecoderSpec.cxx
```diff
@@ -158,30 +174,44 @@ class DataDecoderTask
...
+    if (mDebug) {
+      std::cout << "digitsSize " << digitsSize << "  rofsSize " << rofsSize << "  orbitsSize " << orbitsSize << std::endl;
+    }
+
     // create the output message
     auto freefct = [](void* data, void*) { free(data); };
-    pc.outputs().adoptChunk(Output{"MCH", "DIGITS", 0}, digitsBuffer, digitsSize, freefct, nullptr);
-    pc.outputs().adoptChunk(Output{"MCH", "ORBITS", 0}, orbitsBuffer, orbitsSize, freefct, nullptr);
+    pc.outputs().adoptChunk(Output{header::gDataOriginMCH, "DIGITS", 0}, digitsBuffer, digitsSize, freefct, nullptr);
+    pc.outputs().adoptChunk(Output{header::gDataOriginMCH, "ROFS", 0}, rofsBuffer, rofsSize, freefct, nullptr);
```
aphecetche: other workflows expect the digit rofs spec to be named "DIGITROFS" (because there's also CLUSTEROFs, etc...)

aferrero2707: Ok, I will change the output spec.

Detectors/MUON/MCH/Raw/Decoder/src/UserLogicElinkDecoder.h
aferrero2707: Yes, indeed. I will remove the unrelated changes and prepare separate PR(s) for them.

aferrero2707:
I just pushed some changes/simplifications to the ROF finder code. In particular:

- the time-based sorting of the digits is now applied on a vector containing the indexes of the digits in the input vector (while the old code was using a multiset of digit pointers). For simplicity the digits are not split into different orbits, and only a single global vector is sorted.
- the ROFFinder object does not need to be persistent, it is therefore instantiated each time the decoding  is completed. The ROFFinder constructor now takes a reference to the decoded digits vector and the value of the first orbit in the TF, and outputs a vector of ROFRecords and a vector of digit indexes sorted in time-ascending order. Two helper functions allow to store the ROF and sorted digits vectors into memory buffers that can then be passed to the DPL outputs.
- it is possible that some digits at the beginning of a given Time Frame are arriving late, and that their time stamp actually corresponds to the previous TF. At this stage those digits are kept and the corresponding ROFs are created with an orbit number smaller than the first orbit of the current TF. For the moment the proper handling of such out-of-TF digits is deferred to a later stage in the workflow...

Detectors/MUON/MCH/Raw/Decoder/src/ROFFinder.cxx
```diff
@@ -0,0 +1,370 @@
...
+{
+  // number of bunch crossings in one orbit
+  static const int32_t BCINORBIT = o2::constants::lhc::LHCMaxBunches;
+  auto rofDiff = [](const o2::mch::ROFRecord& rof1, const o2::mch::ROFRecord& rof2) -> int64_t {
+    int64_t dOrbit = static_cast<int64_t>(rof2.getBCData().orbit) - static_cast<int64_t>(rof1.getBCData().orbit);
+    int64_t dBC = static_cast<int64_t>(rof2.getBCData().bc) - static_cast<int64_t>(rof1.getBCData().bc);
+
+    return (dOrbit * BCINORBIT + dBC);
+  };
+
```
pillot, line 259: you could use the function InteractionRecord::differenceInBC(...) instead of this lambda.

Detectors/MUON/MCH/Raw/Decoder/src/ROFFinder.cxx
```diff
@@ -0,0 +1,370 @@
...
+    std::cout << "Inserting digit: "
+              << "pad " << digit.getDetID() << "," << digit.getPadID() << " "
+              << digit.getOrbit() << " " << digit.getTime() << " " << digit.getBXTime() << std::endl;
+#endif
+
+    mOrderedDigits.emplace_back(i);
+  }
+
+  mRawDigitIdComp.digits = &mInputDigits;
+  std::sort(mOrderedDigits.begin(), mOrderedDigits.end(), mRawDigitIdComp);
```
pillot: I think you can use a lambda function instead of creating this structure

pillot (commented):
The new logic looks fine to me. I just have have a couple of minor suggestions.

Detectors/MUON/MCH/Raw/Decoder/src/ROFFinder.cxx
aferrero2707: Done.

Detectors/MUON/MCH/Raw/Decoder/src/ROFFinder.cxx
aferrero2707: Good suggestion. I  have implemented it and it will appear in the next commit.

aphecetche (dismissed):
Just a remark (and certainly not a showstopper for this PR), but please note that the rofs output is in principle already included in the `o2-mch-digits-file-writer-workflow` (see https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/MUON/MCH/DevIO/Digits). The idea being to remove the digits-sink and other related  "debug-only" from the workflow directory (and keep there only the "production" code) (at least once we confirm that the ones in DevIO library are working ok).

aferrero2707:
The CI system seems to have gone bananas again... @ktf any idea?

aphecetche:
@ktf @davidrohr @TimoWilken what's the situation with the CI ?

ktf:
See https://alice.its.cern.ch/jira/browse/O2-2356 for something which is at least partially related.

aphecetche (approved):

ktf:
Simulation issue unrelated. Merging.

PR #6121, opened 2021-05-09
Modify internals of RecoContainer, add fwd-declared getters + support for S-vertices (also in InputHelper)

shahor02:
@rpezzi this may create some conflicts with your PR, but better to cleanup it now.

### DIFF
--- a/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
+++ b/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
@@ -17,6 +17,7 @@
 
 #include "CommonDataFormat/InteractionRecord.h"
 #include "ReconstructionDataFormats/GlobalTrackAccessor.h"
+#include "CommonDataFormat/RangeReference.h"
 #include "ReconstructionDataFormats/GlobalTrackID.h"
 #include "CommonDataFormat/AbstractRefAccessor.h"
 #include "SimulationDataFormat/MCCompLabel.h"
@@ -29,6 +30,7 @@
 // Please include headers for TPC Hits or TRD tracklets directly (DataFormatsTPC/WorkflowHelper.h / DataFormatsTRD/RecoInputContainer.h)
 namespace o2::tpc
 {
+class TrackTPC;
 using TPCClRefElem = uint32_t;
 struct ClusterNativeAccess;
 namespace internal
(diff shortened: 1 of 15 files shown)


davidrohr:
I didn't try it yet, but looks good. I'll try later with the GPU workflow and gpu display if all inputs still work as before. Then I would merge it ASAP and clean up the TRD stuff later.

davidrohr:
OK, actually I did a quick test just now, and in the GPU display I now see less matches to TOF. Something has changed. Will have a look later, but now really have to proceed with my CHEP slides...

To redroduce: Run
```
--read-from-files --display-clusters ITS,TPC,TRD,TOF --display-tracks TPC,ITS,ITS-TPC,TPC-TRD,ITS-TPC-TRD,TPC-TOF,ITS-TPC-TOF --configKeyValues "GPU_global.continuousMaxTimeBin=2000"
```
before and after this PR and press 8 to show the tracks.

shahor02:
Is there a way to visualize TOF-related tracks only? I've tried with 
````
o2-gpu-display --read-from-files --display-clusters "TOF"  --display-tracks "TPC-TOF,ITS-TPC-TOF" --configKeyValues "GPU_global.continuousMaxTimeBin=2000"
```` 
but it shows TPC clusters only...

**Comment by davidrohr** [comment]
> Is there a way to visualize TOF-related tracks only?

I assume you meant "it shows TOF clusters only"?

Currently, I can show only tracks when there are also TPC tracks.
And requesting tracks without the corresponding cluster might actually crash, I'll have to fix that. Also higher tracks without the lower tracks might crash, i.e. ITS-TPC-TOF tracks currently require also ITS tracks.
What you can do is
```
o2-gpu-display --read-from-files --display-clusters "TPC,TOF"  --display-tracks "TPC,TPC-TOF" --configKeyValues "GPU_global.continuousMaxTimeBin=2000"
```

shahor02:
@davidrohr indeed, ITSTPCTOF was not correctly attached, thanks for spotting.
I noticed that in https://github.com/AliceO2Group/AliceO2/blob/01b8b2e983907665c2b8c1c2b1055f00963e7fe5/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L460-L463 you are dereferencing the pointer which will crash if there is no MC loaded. Why not returning just a pointer on MCTruthContainer? This is what I did for ITS and TOF clusters.

davidrohr:
> you are dereferencing the pointer which will crash if there is no MC loaded. Why not returning just a pointer on MCTruthContainer?

Yes sure, we can return a pointer. (Did I actually add this getter? Don't remember...)
How is it for the TPC? Might be the same. The TPC input container might also have a nullptr to the labels.

Are you changing it here or shall I do it?

shahor02:
The TPC clusters MClabels are actually not attached to RecoContainer, not sure they will be used in the global workflows, but will add, just in case, and change the ref to pointer for the TRD MC.

davidrohr:
Actually they are already loaded, just not exposed directly. You can get the pointer to the container via `inputsTPCclusters->clusterIndex.clustersMCTruth`

davidrohr (approved):

davidrohr:
#6128, which contains this one, passed the FullCi, so this one must be ok. merging, to go ahead.

PR #6122, opened 2021-05-09
Add --loop and --delay options to o2-ctf-reader-workflow

shahor02:


davidrohr:
@TimoWilken : Could you check what happened here in build/O2/o2:
```
go: creating work dir: stat /tmp: no such file or directory
[1;33mgo mod vendor[0m
go: creating work dir: stat /tmp: no such file or directory
make: *** [vendor] Error 1
```
That looks quite suspicious. Perhaps there is really something wrong with the filesystem or /tmp in the container.

TimoWilken:
Indeed, `/tmp` was completely gone in that builder. I've recreated it. A fresh `alisw/slc7-builder:latest` image has a `/tmp` directory, and earlier checks find `/tmp`. Something must have removed it, but I can't find why in the logs. (Every other CI builder also still has its `/tmp`. Strange.)

davidrohr:
OK, so I think that is one more point that speaks for isolating the CI a bit more, and using a new container for each PR. Perhaps we can discuss a bit tomorrow in WP3.

PR #6125, opened 2021-05-10
Move TRD track reader/writer to TRDWorkflowIO library

martenole:
@davidrohr in order to avoid a too crowded PR I start here with two smaller changes

1. fix TRD event time calculation if the first orbit of the TF is not zero analog to how ITS is doing it
2. move TRD track reader and writer to IO library

Usage of createTracks method and TRD track trigger records will come in a followup PR


martenole:
Closing, superseded by https://github.com/AliceO2Group/AliceO2/pull/6130

PR #6127, opened 2021-05-10
MID raw data checker workflow

dstocco:
Workflow to check the consistency of the raw data. Useful for debugging and commissioning


aphecetche (approved):
Would be nice to add a some usage doc for this workflow somewhere.

PR #6129, opened 2021-05-10
Cluster finder for ITS3

mpuccio:
Dear @shahor02 ,
this is the second round of the modifications needed for running the ITS2 CF on ITS3. We had to duplicate other code to avoid major modifications to mission critical ITS2 components. We tried to avoid breaking ITS2 as much as possible but maybe you spot something that is suboptimal (one thing might be the modification to PixelData).
Let us know if there are any comments,
Alex and Max


Detectors/ITSMFT/common/reconstruction/include/ITSMFTReconstruction/PixelData.h
```diff
@@ -84,8 +84,13 @@ class PixelData
 
  private:
   void sanityCheck() const;
+#ifdef ENABLE_UPGRADES
+  static constexpr int RowMask = 0x7FFF; ///< 32768 rows are supported
+  static constexpr int MaskBit = 0x8000; ///< 16-th bit is used to flag masked pixel
+#else
```
shahor02: Since everything else is a copy-paste from current ITS, commenting only here: I would simply remove the ifdef and leave IT3 masks, it is anyway there to account for the masking bit. As far as both ITS2 and ITS3 rows fit to 15 bits, we don't need to differentiate them

mpuccio: OK! Fixed, thanks!

mpuccio:
Green CI finally! :D

PR #6130, opened 2021-05-10
TRD tracking developments

martenole:
This sits on top of https://github.com/AliceO2Group/AliceO2/pull/6125 and https://github.com/AliceO2Group/AliceO2/pull/6128

In addition the following changes are implemented:
- the logic implemented in `GPUWorkflowHelper` is used to request the input tracks and to avoid loading the same track twice if a TPC track is present and also its match to an ITS track
- A track trigger record is added for the TRD
- track filtering is not done in `GPUTRDTracker` anymore. Instead, if a TRD track has the `mCollisionId == -1` it does not have a tracklet attached is skipped
- output is only one TRD track vector

What I am not 100% sure about if that is what you meant @shahor02 with having only one output vector of TRD tracks. At the moment tracks can still be written to two different files, depending if only TPC tracks were used as input or TPC and ITS-TPC tracks. In the latter case both tracks are written to a single file.


shahor02:
Hi @martenole 
I am also not sure anymore, I am afraid merging 2 types creates more problems than offers benefits (easier accounting of the redundancy). I was thinking we can change the `GlobalTrackID::TPCTRD` and `GlobalTrackID::ITSTPCTRD` to a single TPCTRD (for the strict stage) and GLOTRD (for the final stage), but this will invalidate some methods, like decaying ID mask to contributing detectors, now this would require checking the contributor IDs of every tracks. And many similar problems. 
What is worse, the InputHelper now inits 2 instances of the reader triggered by the the request of GTrackID::Source::ITSTPC or GTrackID::Source::ITSTPC assuming that this distinguish between the "strict" and "final" phases, but we request the TPCTRD also in the final phase assuming that we get final TPCTRD tracks.
I am afraid, to have it clean, we have to do the opposite: not only to produce 2 outputs (each with its own trigger record) in case of "final" global phase, but also write them to separate files.
Then, the InputHelper will ship on demand exactly what is needed. And the name of TPCTRD tracks file should be different depending whether we are in "strict" or "final" phase. 
Also, I think we will need to pass to the InputHelper a flag, saying if we are in the "initial" or "final" phase, which should be passed to corresponding getXXXReaderSpecs (and this should be done for TOF also which will have even 6 outputs: 2 "initial" for TPC-TOF and TPC-TRD-TOF and 3 "final" TPC-TOF, TPC-TRD-TOF, ITS-TPC-TRD-TOF and ITS-TPC-TOF. This switch should not only trigger the default file-name change of the reader but also trigger different DataDescription name e.g. with 0 added in the end for "initial" stage.

I am sorry that this cause more work. I would propose that here you simply restore separate outputs for TPCTRD and ITSTPCTRD, each one with its own TriggerRecord, and the rest we do in the following PR.
The InputHelpe

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
```diff
@@ -280,8 +280,6 @@ void GPUTRDTracker_t<TRDTRK, PROP>::DoTracking(GPUChainTracking* chainTracking)
 #endif
   }
 
-  FilterOutTracks();
```
davidrohr, line 283: I might miss something, but where do you filter for the mNTracklets = 0?

davidrohr (changes requested):
I agree with Ruben that the output in the single container produces more headache than it solves. Fortunately you have nicely separated the commits, you can easily leave that out :)
And I'd again do further changes in a follow-up PR.

@shahor02 : I don't like the idea of using different data types too much. This is in my opinion a deficit of the framework that it just matches inputs / outputs purely on type. But perhaps we can just use the subspecification for "iteration count", ie. first stage would be subspec 0, after ITS-TPC would be subspec 1. Then we could apply the same scheme in TOF, and it is easily extendable to a 3rd stage, and one could also subscribe to all of them at once if needed.

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -155,20 +157,37 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
...
-    if (trackGID.includesDet(GTrackID::Source::ITS)) {
-      // this track is from an ITS-TPC seed
-      tracksOutITSTPC.push_back(trdTrack);
-    } else {
-      // this track is from a TPC-only seed
-      tracksOutTPC.push_back(trdTrack);
+    const auto& trdTrack = mTracker->Tracks()[trackIdxArray[iTrk]];
+    if (trdTrack.getCollisionId() < 0) {
+      // skip tracks without TRD tracklets
+      continue;
```
martenole, line 193: @davidrohr filtering for nTracklets == 0 happens here. I will make the comment more clear. The collision ID for tracks is initialized to -1 and only if a tracklet is attached to a track the collision ID is set to the collision ID of the tracklet.

martenole:
I have restored writing out two vectors of tracks each with its own trigger record. For now this PR should be ready

shahor02:
Hi @davidrohr 
Yes, after writing the last message I also thought about using the subspec to tag different stages instead of modifying the DataDescription. But the workflows will still need to get a CL option telling at which stage they operated: e.g. the readers will need to know which subspec to add etc.

davidrohr:
@ktf @TimoWilken : Do you have an idea why the CI is so slow currently? O2/o2 for this PR is pending since 24 hours.

ktf:
I see the same. Actually it seems that since around beginning of May the time taken by the builders is noticeably higher http://alisw.github.io/infrastructure-pr-testing. I also see a bunch of zombies on the build machines, either in o2-sim or in git related helpers. I suspect some update of docker installation but I have not verified.

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
martenole: See comment below

martenole:
@davidrohr finally all checks are passed. Could you please merge this when you get a chance?

PR #6131, opened 2021-05-11
Check/enforce that all links got AddData call in every IR it processed

shahor02:
RawFileWriter will check for every IR for which at least 1 link was filled by detector
if all other registered links also got `addData` call. If not, it will be enforced internally
with 0 payload and `trigger` and `detField` RDH fields provided in the 1st `addData` call of this IR.
This may be pose a problem in case detector needs to insert some header even for 0-size payloads.
In this case it is detector responsibility to make sure that all links receive their proper
`addData` call for every IR.

For the record: currently the following detectors have missing addData calls (numbers for 5 pbpb events + QED from FST):
````
grep 'RawFileWriter forced' *log
emcraw.log:[WARN] RawFileWriter forced 29 dummy addData calls in 3 IRs for links which did not receive data
ft0raw.log:[WARN] RawFileWriter forced 17561 dummy addData calls in 1087 IRs for links which did not receive data
fv0raw.log:[WARN] RawFileWriter forced 54 dummy addData calls in 1193 IRs for links which did not receive data
zdcraw.log:[WARN] RawFileWriter forced 300 dummy addData calls in 20 IRs for links which did not receive data
````

### DIFF
--- a/Detectors/Raw/README.md
+++ b/Detectors/Raw/README.md
@@ -168,6 +168,11 @@ It proveds the ``RDH`` of the page for which it is called, information if the pr
 Some detectors signal the end of the HBF by adding an empty CRU page containing just a header with ``RDH.stop=1`` while others may simply set the ``RDH.stop=1`` in the last CRU page of the HBF (it may appear to be also the 1st and the only page of the HBF and may or may not contain the payload).
 This behaviour is steered by ``writer.setAddSeparateHBFStopPage(bool v)`` switch. By default end of the HBF is signaled on separate page.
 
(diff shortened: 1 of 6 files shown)


shahor02:
@davidrohr now should work.

Pinging to @AllaMaevskaya @mslupeck @mfasDa @cortesep to make sure this automatic data completion is ok for them.

shahor02:
@TimoWilken it looks like the fullCI is stuck here, it even does not tell how many jobs are ahead. Could you check, please?

PR #6134, opened 2021-05-11
DPL: allow more than one plugin in a shared object

ktf:


ktf:
@vascobarroso @awegrzyn

PR #6135, opened 2021-05-11
DPL: introduce accumulated metric for total bytes processed (O2-2324)

ktf:


ktf:
@vascobarroso @awegrzyn this is the correct one.

ktf:
@TimoWilken is it normal that the queues are so long?

ktf:
full CI passed, no real test available.

PR #6136, opened 2021-05-11
TPC: add device to decode IDCs per CRU, laser vdrift calibration, trigger info decoding, unreated fixes

wiechula:
This commit contains various unrelated things
* decoder for integrated digital currents
* trigger information decoding in LinkZS, trigger info usage to be finalized, for the moment only skip trigger information to allow for decoding of LinksZS data
* v-drift calibration using laser tracks. Still work in progress.
* some more unrelated fixes


wiechula:
@davidrohr , I think the failed test is unrelated `task timeout reached .. killing all processes`. Can this be merged?

PR #6137, opened 2021-05-11
DPL GUI: use id rather than device name to identify metrics

ktf:


ktf:
Not tested in the full system test.

PR #6138, opened 2021-05-12
TrackPar::setPID will not alter the charge

shahor02:


shahor02:
@ddobrigk this removes the charge overriding by PID in the constructor with cartesian momenta.

Trivial fix, merging

PR #6140, opened 2021-05-12
DPL: improve signaling logic

ktf:
The fact that lastSignal was global introduced a race condition
for the second (third...) reader. This now makes it per device so
that we do not skip signals for different devices which come too close.

Also add metrics to debug the time between different signals.


ktf:
Merging this as it passes all tests and runs fine by hand on hyperloop. Will create a daily build and check with the full setup and revert in case of issues.

PR #6141, opened 2021-05-12
Fix in TPC-ITS matching: cleaunup of prev TF reco buffers was incomplete

shahor02:


shahor02:
Fix is trivial and tested locally, merging

PR #6145, opened 2021-05-12
make sure MFT digi2raw creates 1 file per link if requested

shahor02:


shahor02:
trivial fix, tested locally, merging.

PR #6147, opened 2021-05-12
[EMCAL-699] Add RCU trailer also for links without payload

mfasDa:
In case a link does not contain data for a given
event, still a RCU trailer with the SRU configuration,
L1 phase and payload size 0 needs to be added
according to the implementation in the SRU
firmware.

### DIFF
--- a/Detectors/EMCAL/simulation/src/RawWriter.cxx
+++ b/Detectors/EMCAL/simulation/src/RawWriter.cxx
@@ -177,8 +177,9 @@ bool RawWriter::processTrigger(const o2::emcal::TriggerRecord& trg)
     }
 
     if (!payload.size()) {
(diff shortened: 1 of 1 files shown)


mfasDa:
@shahor02 @davidrohr In case of data we sent an RCU trailer with the SRU configuration, L1 phase and 0 payload size in case  no payload was found in the SRU for a given trigger. I adapted this now in simulation. Apparently it is sufficient to remove a continue in the payload size check, because in this case the RCU trailer is automatically added to an empty payload for all active links.

shahor02 (commented):
@mfasDa could you check if it work as expected? With https://github.com/AliceO2Group/AliceO2/pull/6131 + your PR you are not expected to see in the log the message
````
emcraw.log:[WARN] RawFileWriter forced 29 dummy addData calls in 3 IRs for links which did not receive data
````

PR #6151, opened 2021-05-13
New MFT Power cables and services update

frmanso:
- Addition of the power cables for the MFT disks and the Power Supply Unit (in HalfCone.cxx and .h)
- Addition of flags to control the creation of the readout cables and the services (in MFTBaseParam.h)
- Modification of the cooling pipes to avoid overlaps with the power cables and the beam support (in HeatExchanger.cxx).

### DIFF
--- a/Detectors/ITSMFT/MFT/base/src/HeatExchanger.cxx
+++ b/Detectors/ITSMFT/MFT/base/src/HeatExchanger.cxx
@@ -3738,14 +3738,16 @@ void HeatExchanger::createCoolingPipes(Int_t half, Int_t disk)
   //-----------------------------------------------------------------
   if (disk == 0 || disk == 1 || disk == 2) {
     auto* mCoolingPipe1 = new TGeoVolumeAssembly(Form("cooling_pipe1_H%d", half));
+    auto* mCoolingPipeRear1 = new TGeoVolumeAssembly(Form("cooling_pipeRear1_H%d", half));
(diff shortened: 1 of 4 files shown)


Detectors/ITSMFT/MFT/base/src/HeatExchanger.cxx
```diff
@@ -3738,14 +3738,16 @@ void HeatExchanger::createCoolingPipes(Int_t half, Int_t disk)
   //-----------------------------------------------------------------
   if (disk == 0 || disk == 1 || disk == 2) {
     auto* mCoolingPipe1 = new TGeoVolumeAssembly(Form("cooling_pipe1_H%d", half));
+    auto* mCoolingPipeRear1 = new TGeoVolumeAssembly(Form("cooling_pipeRear1_H%d", half));
+    auto* mCoolingPipeRear2 = new TGeoVolumeAssembly(Form("cooling_pipeRear2_H%d", half));
     if (disk == 0) {
-      length1 = 5.3;
+      length1 = 1.5; //5.3;
```
rpezzi: Hi Franck. Thanks.
Is it necessary to keep previous values like this as comments?

frmanso: No sure, I'm going to clean that.

rpezzi (dismissed):

rpezzi (approved):

PR #6157, opened 2021-05-14
Add MaxDXYIni cut on DCAFitterN seed XY DCA

shahor02:
@aimeric-landou here is it, at the moment I set 4cm distance cut, might be narrowed, probably.


shahor02:
for some reason clang-format check did not run, but locally it is ok, merging.

PR #6166, opened 2021-05-16
Changes for Root v6-24

pzhristov:
- Test some macros in compiled mode
- Corrected include files
- Define the ms_gsl version

### DIFF
--- a/cmake/O2AddTestWrapper.cmake
+++ b/cmake/O2AddTestWrapper.cmake
@@ -100,7 +100,8 @@ function(o2_add_test_wrapper)
 
   # For now, we enforce 3 max attempts for all tests.
   # No need to ignore time out, since we have 3 attempts
-  #PH set(A_MAX_ATTEMPTS 3)
+  # PH The line below should be removed - no need to force three attempts
+  set(A_MAX_ATTEMPTS 3)
   set(A_DONT_FAIL_ON_TIMEOUT "")
 
   math(EXPR ctestTimeout "(20 + ${A_TIMEOUT}) * ${A_MAX_ATTEMPTS}")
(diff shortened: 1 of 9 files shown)


cmake/O2AddTestWrapper.cmake
```diff
@@ -100,7 +100,7 @@ function(o2_add_test_wrapper)
 
   # For now, we enforce 3 max attempts for all tests.
   # No need to ignore time out, since we have 3 attempts
-  set(A_MAX_ATTEMPTS 3)
+  #PH set(A_MAX_ATTEMPTS 3)
```
davidrohr: I do not understand what is the relation between the number of ctest attempts and ROOT 6.24?

davidrohr: I mean, are we sure that now all tests are stable? We added the 3 attempts since some tests were failing randomly and the CI was red because of that quite often.

pzhristov: It is very boring to wait 3 times for the same test to fail and to get the same long output three times. If a test is not stable, it should be fixed or disabled. I think there is no need to waste time with three attempts.

davidrohr: ok, I agree with that, but then I think we need to be aggressive to immediately disable failing tests.
And then I think there are some geant tests with fail randomly, which we cannot fix, so perhaps at least for them we should maintain 3 attempts. @sawenzel can comment.

pzhristov: If we want specific number of attempts, we should specify MAX_ATTEMPTS in the o2_add_test call and not change the value in the wrapper.

aphecetche: Unless there is some urgency to fix the max_attempt right now, there is a PR in progress to get rid of some of our custom test cooking (https://github.com/AliceO2Group/AliceO2/pull/5916) (to be honest I've not been able to work on it lately, but I'll get back to it), so I guess the test policy change should not happen in this Root 6.24 PR.

pzhristov: Ok, changed back with a reminder to take care later

PR #6168, opened 2021-05-16
TRD TimeSlot based calib based on global tracking

martenole:
Hi @aberdnikova, @aschmah and @tdietel 
this is my first try to create a time-slot based calibration routine for the TRD. The class `TrackBasedCalib` receives the output from the TRD matching and the TRD tracklets and creates the histograms with angular residuals (the class to collect the data is called `AngularResidHistos`). I think `TrackBasedCalib` would be the natural place to add also a method to collect the TRD tracklet ADC vs. TPC track dEdx information needed for the gain calibration. I added a function prototype, but not yet the implementation for that. After the `o2-trd-global-tracking` workflow the filled angular residual histograms are written to a file.

The DPL device which will run on the aggregator node is called `VdAndExBCalibDevice`. This steers the time-slot based calibration in `CalibratorVdExB`. In the latter one needs to add the fitting of the histograms and thus the determination of vD and ExB inside the method `finalizeSlot()`. Also the CCDB objects need to be created by that class. I think Nastia you started with this inside the `CalibVDrift` class. You can probably simply move your methods to `CalibratorVdExB`.

Currently the `CalibratorVdExB` is a `o2::calibration::TimeSlotCalibration<o2::trd::AngularResidHistos, o2::trd::AngularResidHistos>`, so the input and the container for the TimeSlotCalibration are the same. This I did because the `AngularResidHistos` have a fixed memory size and can easily be merged together. But we could also change that and create a custom container if that would be more convenient for the calibration.

We could also discuss this a bit in a dedicated TRD O2 meeting.

Cheers,
Ole


martenole:
@aberdnikova I would like to merge this now. As a reminder the calibration can be run with
`o2-trd-global-tracking -b --disable-root-output | o2-calibration-trd-vdrift-exb -b`

martenole:
Added reader for calibration data so that the tracking doesn't have to be run each time in order to test the calibration. Should also be helpful for testing the calibration with manually merged input data. To run from file:
`o2-calibration-trd-vdrift-exb -b --enable-root-input`

martenole:
I don't understand the error from dataflow, it says:
/mnt/mesos/sandbox/sandbox/o2suite-dataflow/O2/macro/run_trac_its.C:173:17: error: no viable constructor or deduction guide for deduction of template arguments of 'span'
I don't think this can be related to the changes of this PR, right?

Also, what is the best way to deal with failed builds probably caused by too large memory consumption of the build process? Should I bump the CI again with a dummy commit? Are those checks retried automatically?

shahor02:
@martenole actually, also the macOS complains about the `run_trac_its.C`, it might be that indeed there is some confusion in the headers. Are you able to do `.L $O2_ROOT/share/macro/run_trac_its.C` and  `.L $O2_ROOT/share/macro/run_trac_its.C+` with your branch?
In general, all tests are broken for different reasons, it seems there is no point in retrying until they are fixed.

martenole:
Hi @shahor02 
this is weird. Indeed I cannot load the `run_trac_its.C` macro, neither in normal nor in compiled mode. I get a seg fault in both cases. And before the seg fault there is the error that it cannot construct a span, because it fails to deduct the template arguments. But it does not complain that a header cannot be found.

martenole:
With `dev` the macro is loaded fine, so it looks as if I broke something here

Detectors/TRD/base/include/TRDBase/RecoParam.h
```diff
@@ -0,0 +1,57 @@
...
+  /// Get tracklet z correction coefficient for track-eta based corraction
+  float getZCorrCoeffNRC() const { return mZCorrCoefNRC; }
+
+ private:
+  // tracklet error parameterization depends on the magnetic field
+  float mA2{1.f};            ///< parameterization for tracklet position resolution
+  float mB{0.f};             ///< parameterization for tracklet position resolution
+  float mC2{0.f};            ///< parameterization for tracklet position resolution
+  float mZCorrCoefNRC{1.4f}; ///< tracklet z-position depends linearly on track dip angle
+};
```
shahor02, line 58: not sure this causes the macro test problem, but since you require a dictionary for this class in the CMakeLists and LinkDef, you should have here `ClassDefNV...`

shahor02 (changes requested):

Detectors/TRD/base/include/TRDBase/RecoParam.h
martenole: Thanks for spotting! I have added the `ClassDefNV` and also added RecoParam to the LinkDef (this I think I forgot before), but still no luck. I cannot load the macro locally and don't really know how to debug the problem

Detectors/TRD/workflow/CMakeLists.txt
```diff
@@ -21,10 +21,11 @@ o2_add_library(TRDWorkflow
                        src/TRDTrackingWorkflow.cxx
                        src/EntropyDecoderSpec.cxx
                        src/EntropyEncoderSpec.cxx
+                       src/TrackBasedCalibSpec.cxx
                PUBLIC_LINK_LIBRARIES O2::Framework O2::DPLUtils O2::Steer O2::Algorithm O2::DataFormatsTRD O2::TRDSimulation O2::TRDReconstruction O2::DetectorsBase O2::SimulationDataFormat O2::TRDBase O2::TRDCalibration O2::GPUTracking O2::GlobalTrackingWorkflowReaders O2::GPUWorkflowHelper O2::ReconstructionDataFormats O2::TPCWorkflow O2::TRDWorkflowIO)
 
-                   #o2_target_root_dictionary(TRDWorkflow
-                   # HEADERS include/TRDWorkflow/TRDTrapSimulatorSpec.h)
+o2_target_root_dictionary(TRDWorkflow
```
martenole: Found my own stupid mistake here. Don't know why I tried to create a root dictionary here. Probably because VdAndExBCalibSpec only has a header and therefore I did not add it to the sources above. I changed that now and that fixes the issue with the macro test

martenole:
In the fullCI I see an error from PHOS:
[28711:PHOSDigitizer]: [16:31:45][ERROR] impossible parameterization 0
But I don't think it is related to this PR.
@tdietel would it be fine for you to merge this?

davidrohr:
I am not sure if this PHOS error is critical. What I see is that the digitizer times out,
In this case it was running for a really long time > 10 minutes. No idea why, somehow that server must have been under heavy load. @TimoWilken : Can you reset the fullCI for this PR?

shahor02:
> In the fullCI I see an error from PHOS:
> [28711:PHOSDigitizer]: [16:31:45][ERROR] impossible parameterization 0

It is not critical but should be fixed:
https://github.com/AliceO2Group/AliceO2/pull/6287#issuecomment-853076378

DataFormats/Detectors/TRD/include/DataFormatsTRD/AngularResidHistos.h
```diff
@@ -0,0 +1,56 @@
...
+  size_t getNEntries() const { return mNEntriesTotal; }
+
+  void fill(const gsl::span<const AngularResidHistos> input);
+  void merge(const AngularResidHistos* prev);
+  void print();
+
+ private:
+  // TODO use NCHAMBER instead of MAXCHAMBER and indirection array?
+  static constexpr float INVBINWIDTH = constants::NBINSANGLEDIFF / (2.f * constants::MAXIMPACTANGLE);
+  // the additional bin is for storing under-/overflow entries. Is it needed?
```
tdietel: I would either go with separate over-/under-flow bins, or none at all. 

The over-/underflow might be useful to confirm that the range is ok. Alternatively, we can have one bin on either side that we exclude from the fit.

DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
```diff
@@ -56,7 +56,8 @@ constexpr float GRANULARITYTRKLSLOPE = 1.f / PADGRANULARITYTRKLSLOPE; // granula
 
 // OS: Should this not be flexible for example in case of Kr calib?
 constexpr int TIMEBINS = 30; // the number of time bins
-constexpr int NBINSANGLEDIFF = 26; // the number of bins for the track angle used for the vDrift and ExB calibration based on the tracking (last bin is for under-/overflow entries)
+constexpr float MAXIMPACTANGLE = 25.f; // the maximum impact angle for tracks relative to the TRD detector plane to be considered for vDrift and ExB calibration
```
tdietel, line 59: Is there a reason for the maximum impact angle to be 25 (degrees?)?
And is this a data format thing?

Detectors/TRD/base/include/TRDBase/RecoParam.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <array>
+#include "Rtypes.h"
+
+namespace o2
+{
+namespace trd
+{
+
+class RecoParam
```
tdietel, line 26: Do we have some more documentation for this class? Maybe the relevant variables are defined in your thesis, Ole?

tdietel (dismissed):
Looks good to me. My comments are not critical for the merge.

DataFormats/Detectors/TRD/include/DataFormatsTRD/AngularResidHistos.h
martenole: @aschmah and @aberdnikova could you please comment on this? Not sure if it is needed to keep track of the entries in the over-/underflow bin of the delta Alpha histograms or if those entries should be discarded?

DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
martenole: Also here I would ask @aschmah and @aberdnikova to comment. I think you studied the ranges for the angular residuals and probably +-25 degrees gives a sufficient handle for the fit, right? 25 degrees over the 3cm TRD drift volume corresponds to ~1.4cm deflection which is already out of range for the online tracklets in Run 2 format though.
The maximum deflection and number of bins define the exchange format `AngularResidHistos`. If needed I can also change it to be configurable and not a fixed size array. Just let me know.

Detectors/TRD/base/include/TRDBase/RecoParam.h
martenole: Yes, this is documented in my thesis. I could add a pointer here to the relevant chapter or just write a more detailed comment on how the tracklet resolution is determined

martenole:
After discussing with Alex and Nastia I removed the under/overflow bins and added some comments to the `RecoParam` class. After the checks are through this can be merged

martenole:
I cannot see an error in the fullCI that could be related to this PR. @shahor02 or @davidrohr could you merge this please?

PR #6170, opened 2021-05-17
Switch to RecoContainer for information fetching 

nburmaso:


Detectors/AOD/src/aod-producer-workflow.cxx
```diff
@@ -8,18 +8,28 @@
...
 void customize(std::vector<ConfigParamSpec>& workflowOptions)
 {
   // option allowing to set parameters
   std::vector<o2::framework::ConfigParamSpec> options{
-    {"ignore-aod-writer", VariantType::Int, 0, {"Ignore DPL AOD writer and write tables directly into a file. 0 -- off, != 0 -- on"}},
-    {"configKeyValues", VariantType::String, "", {"Semicolon separated key=value strings"}}};
+    {"disable-root-input", o2::framework::VariantType::Bool, false, {"disable root-files input reader"}},
+    {"disable-root-output", o2::framework::VariantType::Bool, false, {"disable root-files output writer"}},
+    {"disable-mc", o2::framework::VariantType::Bool, false, {"disable mc info"}},
+    {"vertexing-sources", VariantType::String, std::string{GID::ALL}, {"comma-separated list of sources"}},
```
shahor02: not needed here

Detectors/AOD/src/aod-producer-workflow.cxx
```diff
@@ -29,7 +39,20 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
...
+
+  bool useMC = true;
+
+  GID::mask_t src = GID::getSourcesMask("ITS,TPC,ITS-TPC,ITS-TPC-TOF,TPC-TOF");
+  GID::mask_t dummy, srcClus = GID::includesDet(DetID::TOF, src) ? GID::getSourceMask(GID::TOF) : dummy;
+
+  WorkflowSpec specs;
+
+  specs.emplace_back(o2::ft0::getDigitReaderSpec(useMC));
+  specs.emplace_back(o2::ft0::getReconstructionSpec(useMC));
```
shahor02: why is this here?

Detectors/AOD/src/aod-producer-workflow.cxx
```diff
@@ -29,7 +39,20 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
...
+  specs.emplace_back(o2::aodproducer::getAODProducerWorkflowSpec(src));
+
+  o2::globaltracking::InputHelper::addInputSpecs(configcontext, specs, srcClus, src, src, useMC, srcClus);
+  o2::globaltracking::InputHelper::addInputSpecsPVertex(configcontext, specs, useMC);
+
```
shahor02, line 55: you would need also `o2::globaltracking::InputHelper::addInputSpecsPVertex(configcontext, specs);`, perhaps conditioned by the option to process secondary vertices.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -910,46 +810,40 @@ void AODProducerWorkflowDPL::endOfStream(EndOfStreamContext& ec)
...
+DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src)
 {
-  std::vector<InputSpec> inputs;
   std::vector<OutputSpec> outputs;
+  auto dataRequest = std::make_shared<DataRequest>();
+
+  bool useMC = true;
+  dataRequest->requestTracks(src, useMC);
+  dataRequest->requestPrimaryVertertices(useMC);
+
```
shahor02: also `dataRequest->requestSecondaryVertertices(useMC);` (note that the useMC here is dummy, the v0s and cascades don't provide MC truth.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -730,6 +664,8 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
                      truncateFloatFraction(timeStamp.getTimeStampError() * 1E3, mCollisionPositionCov),
                      collisionTimeMask);
 
+    LOG(INFO) << "vertex.getNContributors() = " << vertex.getNContributors();
+
     auto trackRef = primVer2TRefs[collisionID];
     int start = trackRef.getFirstEntryOfSource(0);
     int ntracks = trackRef.getEntriesOfSource(0);
```
shahor02: In order to account for eventual redundancies among the tracks, I would suggest instead to do
````
using GIDV = o2::dataformats::VtxTrackIndex;
for (int src=GIDV::NSources;src--;) 
{
  int from = trackRef.getFirstEntryOfSource(src), to = from + trackRef.getEntriesOfSource(src);
  for (int it=from;it<to;it++) {
    auto tID = primVerGIs[it];
    if (src == ....) // processing for each type.

  }
}
````
The tID, apart from giving the source and the index within the source, also provides info on the track relation to vertex:
If `tID.isPVContributor()` is `true`, then this track was used in this p.vertex fit and as such, will not be encountered in the the track references of other vertices (even if time-wise it is compatible to them).
If `tID.isAmbiguous()` is `true`, then this track was matched (time-wise) to at least 1 other vertex.
If both flags are false, then the track was matched (time-wise) to given vertex only.

Note that with such a loop when processing e.g. the `src = ITS or TPC`, you will not get those tracks which were contributing to ITS-TPC matches, and in a same fashion, when processing `src = ITS-TPC` you will not get those tracks which were contributing to e.g. `ITS-TPC-TRD` or `ITS-TPC-TOF` or `ITS-TPC-TRD-TOF`.
Therefore, when processing, e.g. `src=ITS-TPC-TRD-TOF`, you can access its contributors on all levels by method e.g.
````
auto contributorsGID = recoCont.getSingleDetectorRefs(tID);
for (int isrc=GIDV::NSources;isrc--;) 
{
  if (contributorsGID[isrc].isIndexSet()) { // this source contributes to this more global track
     // process info of this source.
  }
}
````

File: Detectors/AOD/src/aod-producer-workflow.cxx (hunk 2 above)
Comment (nburmaso) [reply Detectors/AOD/src/aod-producer-workflow.cxx]: Hi @shahor02. If I understand correctly, FT0's [RecPointReader](https://github.com/AliceO2Group/AliceO2/blob/9f0a3892d7457c693a43a20c88cda000c3d2ed12/Detectors/FIT/FT0/workflow/src/RecPointReaderSpec.cxx#L78-L94) does not return channel data, which is needed for FT0 table. So for now I subscribe to [FT0 reconstructor](https://github.com/AliceO2Group/AliceO2/blob/9f0a3892d7457c693a43a20c88cda000c3d2ed12/Detectors/FIT/FT0/workflow/src/ReconstructionSpec.cxx#L83-L102), which does return it. Is there a better way to access channel data?

Comment (shahor02) [reply Detectors/AOD/src/aod-producer-workflow.cxx]: Well, the RecPointReader should be fixed, and then I'll add the channel data to RecoContainer. 
But one should not instantiate from the aod-producer other processing devices, exactly the same input data should be obtained either from already running upstream devices or from the readers.

**Comment by nburmaso** [comment]
Hi @shahor02. I have tried fetching FT0 information using RecoContainer (see the last commit), but it seems that I am doing something wrong here. When I try running the converter, it crashes on ft0-recpoint-reader workflow -- see file attached [ft0_reader_crash.log](https://github.com/AliceO2Group/AliceO2/files/6511149/ft0_reader_crash.log). 

Also I have rather strange situation: as you mentioned before, there are some flags in customize(), which are not actually needed (such as "disable-root-input" etc.), but when I try removing them, I get a message like `[ERROR] error while setting up workflow: missing option: disable-mc (No such node (disable-mc))`. I am very confused because of it.

Could you please give an advice on these issues?

Cheers,
Nazar

**Comment by nburmaso** [comment]
> Hi @shahor02. I have tried fetching FT0 information using RecoContainer (see the last commit), but it seems that I am doing something wrong here. When I try running the converter, it crashes on ft0-recpoint-reader workflow -- see file attached [ft0_reader_crash.log](https://github.com/AliceO2Group/AliceO2/files/6511149/ft0_reader_crash.log).
> 
> Also I have rather strange situation: as you mentioned before, there are some flags in customize(), which are not actually needed (such as "disable-root-input" etc.), but when I try removing them, I get a message like `[ERROR] error while setting up workflow: missing option: disable-mc (No such node (disable-mc))`. I am very confused because of it.
> 
> Could you please give an advice on these issues?
> 
> Cheers,
> Nazar

@shahor02, I think I have found a problem in FT0 RecPointReader. It seems that there is a typo, which causes segfault error https://github.com/AliceO2Group/AliceO2/blob/b3fc036fe473fbc95fe6cbfbb839790c48c17e07/Detectors/FIT/FT0/workflow/src/RecPointReaderSpec.cxx#L68-L69

**Comment by shahor02** [comment]
Hi @nburmaso 
Thanks for spotting! To avoid other workflows crashing, I would prefer to merge your patch directly in other PR, coulld you then revert it in your PR once I do this?

For the options: I was commenting on the "vertexing-sources" option, which seems to be the copy-paste from the pvertexer workflow. In opposite, I think all workflows should have "disable-root-input", "disable-root-output" and "disable-mc" options. Anyway, frin the way you describe the problem it looks like you have suppressed the option when building `std::vector<o2::framework::ConfigParamSpec>` but forgot to suppress its query, i.e. `configcontext.options().get<...`.

**Comment by nburmaso** [comment]
> Thanks for spotting! To avoid other workflows crashing, I would prefer to merge your patch directly in other PR, coulld you then revert it in your PR once I do this?

Yes, sure. 

> For the options: I was commenting on the "vertexing-sources" option, which seems to be the copy-paste from the pvertexer workflow. In opposite, I think all workflows should have "disable-root-input", "disable-root-output" and "disable-mc" options. Anyway, frin the way you describe the problem it looks like you have suppressed the option when building `std::vector<o2::framework::ConfigParamSpec>` but forgot to suppress its query, i.e. `configcontext.options().get<...`.

Oh, I understand now. I will fix this too.

**Comment by nburmaso** [comment]
This PR becomes a bit bulky, so I think it is better to see if it is ready to be merged, and then continue with updates in the next PR

File: Detectors/AOD/src/AODProducerWorkflowSpec.cxx [hunk 6]
```diff
@@ -83,6 +87,61 @@ void AODProducerWorkflowDPL::findMinMaxBc(gsl::span<const o2::ft0::RecPoints>& f
...
+    if (bcsMap.find(bc) == bcsMap.end()) {
+      bcsMap[bc] = 1;
+    }
+  }
+
+  for (auto& vertex : primVertices) {
+    auto& timeStamp = vertex.getTimeStamp();
+    Double_t tsTimeStamp = timeStamp.getTimeStamp() * 1E3; // mus to ns
+    uint64_t globalBC = std::round(tsTimeStamp / o2::constants::lhc::LHCBunchSpacingNS);
+    tsTimeStamp = globalBC * o2::constants::lhc::LHCBunchSpacingNS - tsTimeStamp;
```
shahor02: why is this needed?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -83,6 +87,61 @@ void AODProducerWorkflowDPL::findMinMaxBc(gsl::span<const o2::ft0::RecPoints>& f
...
+      bc = maxGlBC - minGlBC;
+    }
+    if (bcsMap.find(bc) == bcsMap.end()) {
+      bcsMap[bc] = 1;
+    }
+  }
+
+  for (auto& ft0RecPoint : ft0RecPoints) {
+    uint64_t globalBC = ft0RecPoint.getInteractionRecord().orbit * o2::constants::lhc::LHCMaxBunches + ft0RecPoint.getInteractionRecord().bc;
+    uint64_t bc = globalBC - minGlBC;
```
shahor02: This is not directly related to this PR, but why do you introduce this minBC, if you have the 1orbit of TF, e.g. 
`o2::InteractionRecord ir =  recoContainer.startIR; `
Note that in all objects provided by reconstruction the time information is provided  (or should be, if it is still not the case) wrt this start of TF (except the cases when it is explicitly provided as an InteractionRecord).

Also note that the `o2::InteractionRecord` has plenty of methods to for comparison, subtraction, addition etc., e.g.
`auto bcdif = ft0RecPoint.getInteractionRecord().differenceInBC(  recoContainer.startIR );`

shahor02 (commented):
Hi @nburmaso 

Usage of the RecoContainer looks fine, did you try to run the `prodtests/sim_challenge.sh` to make sure there are no problems?

I've commented below on some other points, which, if needed, can be sorted out later.

nburmaso:
Hi @shahor02. Yes, I tried running `sim_challenge.sh -s pp -r 10 -n 100`, and it seems to work fine.

Concerning your latest comments: 
* I have introduced minGlBC before first BC of an orbit was propagated, so it is not needed now indeed.
* These manipulations with timestamps are (were?) needed, because there was a discussion during one of wp4+14 meetings where it was decided that we need to assign a BC to each collision and store a timestamp wrt to the beginning of this BC. This discussion was several month ago, and there were some slides by Jan Fiete with the idea of which BC to assign to a collision based on a filling scheme, but I can't find these slides.

shahor02:
@nburmaso my comment on https://github.com/AliceO2Group/AliceO2/blob/e3a7b824133a59db8d8a11912cf83a0f639828a7/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L127 was that you do some manipulation on `tsTimeStamp` which is never used.
As for the manipulations with BC in general, I was just hinting that the `InteractionRecord` has plenty of methods defined, particularly for manipulating the orbit/bc as contiguous BC counter.

shahor02:
@nburmaso please remove [WIP] to trigger a CI check

nburmaso:
> @nburmaso please remove [WIP] to trigger a CI check

Done. Thanks

Detectors/AOD/src/aod-producer-workflow.cxx
```diff
@@ -29,7 +36,19 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
...
 {
   o2::conf::ConfigurableParam::updateFromString(configcontext.options().get<std::string>("configKeyValues"));
-  //  o2::conf::ConfigurableParam::updateFromString(configcontext.options().get<std::string>("ignore-aod-writer"));
-  int ignoreWriter = configcontext.options().get<int>("ignore-aod-writer");
-  return std::move(o2::aodproducer::getAODProducerWorkflow(ignoreWriter));
+
+  bool useMC = true;
+
+  GID::mask_t src = GID::getSourcesMask("ITS,MFT,TPC,ITS-TPC,ITS-TPC-TOF,TPC-TOF");
+  src |= GID::getSourceMask(GID::FT0);
```
shahor02: you can have all this in 1 line:
`GID::mask_t src = GID::getSourcesMask("ITS,MFT,TPC,ITS-TPC,ITS-TPC-TOF,TPC-TOF,FT0");`

Detectors/AOD/src/aod-producer-workflow.cxx
shahor02: I think SVertices also should be added.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -121,122 +179,6 @@ uint64_t AODProducerWorkflowDPL::getTFNumber(uint64_t firstVtxGlBC, int runNumbe
   return ts;
```
shahor02, line 136: No sure this is correct. 1st of all, what you extract is not the TFID but the timestamp.
But even assuming that this is what you want to get: the ts initially is the timestamp of SOR in \mus. Then you add to it the time difference between the 1st vertex and 1st seen BC, both withing the same TF (which can be anywhere within the run). This time difference might be 0 and limited in time by the duration of the TF. So, you will always get the timestamp of "SOR + something<TF duration".
I think you should have something like:
````
// to get the timestamp
ts += (firstRec - initialOrbit).bc2ns() / 1000000; 
````
or
````
// to get the TFid since the beginning of run
auto tfID = firstRec.differenceInBC(initialOrbit) / o2::constants::lhc::LHCMaxBunches / nHBFperTF;
````

File: Detectors/AOD/src/AODProducerWorkflowSpec.cxx [hunk 10]
```diff
@@ -625,26 +520,26 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
   int index = 0;
   for (auto& rec : mcRecords) {
     auto time = rec.getTimeNS();
-    globalBC = rec.bc + rec.orbit * o2::constants::lhc::LHCMaxBunches;
-    BCid = globalBC - minGlBC;
-    if (BCid < 0) {
-      BCid = 0;
-    } else if (BCid > maxGlBC - minGlBC) {
-      BCid = maxGlBC - minGlBC;
+    uint64_t globalBC = rec.bc + rec.orbit * o2::constants::lhc::LHCMaxBunches;
```
shahor02: note that there is an equivalent `rec.toLong()` method

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -786,6 +677,9 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
           vCollRefsMFT[trackIndex.getIndex()] = collisionID;
         } else if (src == GIndex::Source::ITSTPC) {
           vCollRefsTPCITS[trackIndex.getIndex()] = collisionID;
+          auto& track = tracksITSTPC[trackIndex.getIndex()];
+          vToFillITS[track.getRefITS()] = false;
+          vToFillTPC[track.getRefTPC()] = false;
```
shahor02: The vertex to tracks references table already accounts for the redundancy, i.e. the ITS or TPC tracks referred by `track.getRefITS()` and `track.getRefTPC()` will never appear in this table.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -786,6 +677,9 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
           vCollRefsMFT[trackIndex.getIndex()] = collisionID;
         } else if (src == GIndex::Source::ITSTPC) {
           vCollRefsTPCITS[trackIndex.getIndex()] = collisionID;
```
shahor02: The same track may appear in the references of different vertices (if it is not a contributor but is compatible in time). 
I don't see how this is accounted: here you will account the last matching vertex (collisionID) which is not even necessarily the best match.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -794,46 +688,272 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+    }
+    prevItem = vCollRefsTPCITS.begin();
+    item = std::find(prevItem, vCollRefsTPCITS.end(), iCollision);
+    while (item != vCollRefsTPCITS.end()) {
+      int idx = item - vCollRefsTPCITS.begin();
+      vTracksIdx.push_back({idx, GIndex::Source::ITSTPC});
+      prevItem = item;
+      item = std::find(prevItem + 1, vCollRefsTPCITS.end(), iCollision);
+    }
+  }
```
shahor02: I guess the abundance of `find` calls will make this Vertex->tracks to tracks->Vertex inversion quite inefficient.
You can see how I do such inversion here: https://github.com/AliceO2Group/AliceO2/blob/e4e46e95ee16c3d6b1a276187daccd11cc22899b/Detectors/Vertexing/src/SVertexer.cxx#L147 (rejecting primary track contributors in https://github.com/AliceO2Group/AliceO2/blob/e4e46e95ee16c3d6b1a276187daccd11cc22899b/Detectors/Vertexing/src/SVertexer.cxx#L154-L156, which you don't need).

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -846,9 +966,14 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   uint32_t labelTPC;
   uint16_t labelMask;
 
+  // filling track labels
```
shahor02: I think we need more complete treatment of labels. E.g. ITS-TPC track label will be that of the TPC track, with the isFake()==true of the ITStrack label != TPCtrack label. Don't we want to store the labels of these contributors? This is the only way to know if e.g. ITS or TPC track is fake (i.e. has wrong clusters attached).

nburmaso:
The latest commit includes updates:
* Track table is filled collision by collision, starting from unassigned tracks
* Tracks MC labels are now sorted as well
* The issues related to first BC of TF min. BC and max. BC are addressed.
* Secondary vertices are now requested by the converter, although the table is not added

todo:
* MFT tracks labels need to be moved to a separate table, as it leads to different sizes of tracks and MC track labels tables

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -25,65 +26,71 @@
...
-  for (auto& ft0RecPoint : ft0RecPoints) {
-    uint64_t bc = ft0RecPoint.getInteractionRecord().orbit * o2::constants::lhc::LHCMaxBunches + ft0RecPoint.getInteractionRecord().bc;
-    if (minGlBC > bc) {
-      minGlBC = bc;
-    }
-    if (maxGlBC < bc) {
-      maxGlBC = bc;
+  // collecting non-empty BCs and enumerating them
+  for (auto& rec : mcRecords) {
+    auto time = rec.getTimeNS();
```
shahor02: `time` is not used

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -25,65 +26,71 @@
...
+    uint64_t globalBC = rec.toLong();
+    if (bcsMap.find(globalBC) == bcsMap.end()) {
+      bcsMap[globalBC] = 1;
```
shahor02: why do you need this `find` then `bcsMap[]` assignment? In this way you are doing `find` twice if the assignment to be done. Do directly `bcsMap[globalBC] = 1;`
The same comment for entries below.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -25,65 +26,71 @@
...
-    uint64_t colBCMin = colIRMin.orbit * o2::constants::lhc::LHCMaxBunches + colIRMin.bc;
-    uint64_t colBCMax = colIRMax.orbit * o2::constants::lhc::LHCMaxBunches + colIRMax.bc;
-    if (minGlBC > colBCMin) {
-      minGlBC = colBCMin;
-    }
-    if (maxGlBC < colBCMax) {
-      maxGlBC = colBCMax;
+  for (auto& ft0RecPoint : ft0RecPoints) {
+    uint64_t globalBC = ft0RecPoint.getInteractionRecord().toLong();
+    uint64_t bc = globalBC;
```
shahor02: bc is not used

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -689,69 +531,284 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+      float tofChi2 = -999.f;
+      float tpcSignal = -999.f;
+      float trdSignal = -999.f;
+      float tofSignal = -999.f;
+      float length = -999.f;
+      float tofExpMom = -999.f;
+      float trackEtaEMCAL = -999.f;
+      float trackPhiEMCAL = -999.f;
+      auto& trackIndex = primVerGIs[ti];
+      if (src != GIndex::Source::MFT) {
```
shahor02: why this check is needed? You anyway explicitly check the source for each track type

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -689,69 +531,284 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+        if (src == GIndex::Source::ITS && mFillTracksITS) {
```
shahor02: Perhaps later we should think about rearranging this routine, for sure we will have some changes and with the same code repeating for different types of tracks it will be a headache to introduce them.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -767,73 +824,266 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+                            (Char_t)(128. * track.getSigmaSnpZ() / track.getSigmaSnp2() / track.getSigmaZ2()),
+                            (Char_t)(128. * track.getSigmaTglY() / track.getSigmaTgl2() / track.getSigmaY2()),
+                            (Char_t)(128. * track.getSigmaTglZ() / track.getSigmaTgl2() / track.getSigmaZ2()),
+                            (Char_t)(128. * track.getSigmaTglSnp() / track.getSigmaTgl2() / track.getSigmaSnp2()),
+                            (Char_t)(128. * track.getSigma1PtY() / track.getSigma1Pt2() / track.getSigmaY2()),
+                            (Char_t)(128. * track.getSigma1PtZ() / track.getSigma1Pt2() / track.getSigmaZ2()),
+                            (Char_t)(128. * track.getSigma1PtSnp() / track.getSigma1Pt2() / track.getSigmaSnp2()),
+                            (Char_t)(128. * track.getSigma1PtTgl() / track.getSigma1Pt2() / track.getSigmaTgl2()));
+          }
+          if (src == GIndex::Source::ITSTPCTOF && mFillTracksITSTPC) {
```
shahor02: I think we should also store the label of the TOF cluster

shahor02 (commented):
Hi @nburmaso 
In general fine, but given the amount of different track types, I would separate storing each type to dedicated routines calling the common part. 
I left a few comments below.

nburmaso:
Changes in the latest commit:
* Filling tracks tables was moved to separate subroutines
* Code cleanups and fixes for typos

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Hi @shahor02. Sorry, I don't quite understand: I think there is no such column in the tracks tables in the data model (but I may be wrong)

shahor02: Yes, but I think it should be introduced, this is a MC label like others.
Btw, not there are also TPC-TRD and ITS-TPC-TRD tracks are added to the reconstruction (and RecoContainer), although at the moment w/o the MC labels.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -115,13 +129,90 @@ uint64_t AODProducerWorkflowDPL::getTFNumber(const o2::InteractionRecord& tfStar
 
   uint32_t initialOrbit = mapStartOrbit->at(runNumber);
   uint16_t firstRecBC = tfStartIR.bc;
-  uint32_t firstRecOrbit = tfStartIR.orbit;
+  uint32_t firstRecOrbit = initialOrbit + tfStartIR.orbit;
```
shahor02, line 132: The `initialOrbit` is the abs. orbit of the run start, tfStartIR is the abs orbit of the TF, why summing them? I think the prev. version is correct.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -52,37 +53,50 @@ using GID = o2::dataformats::GlobalTrackID;
...
+float tpcChi2NCl = -999.f;
+float trdChi2 = -999.f;
+float tofChi2 = -999.f;
+float tpcSignal = -999.f;
+float trdSignal = -999.f;
+float tofSignal = -999.f;
+float length = -999.f;
+float tofExpMom = -999.f;
+float trackEtaEMCAL = -999.f;
+float trackPhiEMCAL = -999.f;
```
shahor02, line 75: Instead of out-of-class global variables, better you use a an exchange structure (can be nested to this class),
e.g.
struct TrackInfo {
  float tpcInnerParam = 0.f;
  ...
};
which would created locally and passed to addToTracksExtraTable

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: I thought that tfStartIR is defined wrt beginnig of run, so I added them. Thanks, I will fix this

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: I see, apparently this way is better, so I will switch to struct with parameters

shahor02:
Hi @nburmaso 
Shall I merge this and we see if the [problem](https://alice.its.cern.ch/jira/browse/O2-2380) still persists? The cleanup of out-class variables can be done in the following PR.

nburmaso:
> Hi @nburmaso
> Shall I merge this and we see if the [problem](https://alice.its.cern.ch/jira/browse/O2-2380) still persists? The cleanup of out-class variables can be done in the following PR.

Hi @shahor02. Sure, we can continue working on updates in the next PR

shahor02 (approved):
Merging since the [problem](https://alice.its.cern.ch/jira/browse/O2-2380?focusedCommentId=269072&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-269072) is on the FT0 reconstruction side.

PR #6171, opened 2021-05-17
Fix filter

jgrosseo:


jgrosseo:
@pbuehler It doesn't work otherwise. Did it work for you?
`[47933:histograms-tutorial_C]: [09:00:12][ERROR] Exception caught: Failed to create filter: ExpressionValidationError: Function bool greater_than(float, double) not supported yet.  `

PR #6189, opened 2021-05-18
Fix in composition of ITS/MFT dictionary name

shahor02:


bovulpes (approved):

shahor02:
merging as the fix is almost trivial and most checks are passed

PR #6193, opened 2021-05-18
MCH: improved time frame grouping in CRU page reader

aferrero2707:
This PR improves the grouping of orbits into TimeFrames in the CRU page reader workflow.

The code is not intended as a replacement for the TFBuilder, however it allows to group all the orbits belonging to the same TF into a single buffer when reading data that was directly written to disk by the readout process.
This is useful for testing the MCH processing workflows against real data recoded on the FLPs.

### DIFF
--- a/Detectors/MUON/MCH/Workflow/src/cru-page-reader-workflow.cxx
+++ b/Detectors/MUON/MCH/Workflow/src/cru-page-reader-workflow.cxx
@@ -168,6 +168,7 @@ class FileReaderTask
     /// Get the input file and other options from the context
     LOG(INFO) << "initializing file reader";
     mFrameMax = ic.options().get<int>("nframes");
+    mTimeFrameMax = ic.options().get<int>("max-time-frame");
     mPrint = ic.options().get<bool>("print");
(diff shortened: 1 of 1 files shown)


Detectors/MUON/MCH/Workflow/src/cru-page-reader-workflow.cxx
```diff
@@ -168,6 +168,7 @@ class FileReaderTask
     /// Get the input file and other options from the context
     LOG(INFO) << "initializing file reader";
     mFrameMax = ic.options().get<int>("nframes");
+    mTimeFrameMax = ic.options().get<int>("max-time-frame");
```
aphecetche, line 171: more a note to self than a comment to this PR : at some point we should uniformize (at least within muon) the names of some workflow options. e.g. for digit [reader/writer](https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/MUON/MCH/DevIO/Digits) I've used `--first-tf` and `--max-nof-tfs` for a similar purpose

aphecetche (approved):

PR #6194, opened 2021-05-18
Update CompressorTask.cxx

preghenella:
output buffer size will be the sum of the command-line-specified buffer size and the value automatically calculated.
unless a negative value is used as the command-line-specified buffer size, in which case a fixed buffer size is used


noferini (approved):

preghenella:
@ktf @shahor02 
can you please merge this?
thanks

PR #6196, opened 2021-05-18
[O2-2075] Forwarding EOS in the DPL input proxy

matthiasrichter:
This implements broadcast of `SourceInfoHeader` received in the `ExternalFairMQDeviceProxy` to all downstream channels in order to propagate end-of-stream message. Downstream devices can now react properly on EOS.

The communication with the driver has been reestablished by dropping the `OnData` callback and using a normal DPL ProcessingCallback. The proxy can now terminate on EOS.

Behind this fix: Internal DPL communication between driver and devices goes via websockets, which are  set up from the `ConditionalRun` overload implemented in `DataProcessingDevice`. The proxy has been using an `OnData` callback right from the beginning, but in the meantime, device control has been completely rewritten.

In addition, the DPL output proxy has been updated to send a special {DPL/EOS} message on the out-of-band channel. The output proxy methods need to be consolidated, now there are two methods with some code duplication. I need to check for what reason the second function was added, and ideally merge the two.

### DIFF
--- a/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
+++ b/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
@@ -18,6 +18,7 @@
 #include "Framework/ProcessingContext.h"
 #include "Framework/RawDeviceService.h"
 #include "Framework/CallbackService.h"
+#include "Framework/SourceInfoHeader.h"
 #include "Headers/DataHeader.h"
 #include "Headers/Stack.h"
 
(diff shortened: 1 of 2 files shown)


Framework/Core/src/DataProcessingHelpers.cxx
```diff
@@ -19,18 +19,18 @@
 
 namespace o2::framework
 {
-void DataProcessingHelpers::sendEndOfStream(FairMQDevice& device, OutputChannelSpec const& channel)
+void DataProcessingHelpers::sendEndOfStream(FairMQDevice& device, std::string const& channelName)
```
ktf: Is it really required to change it to a string? I personally like the forward declaration better...

ktf: Moreover (even if not critical in this case), using a proper type would allow us to cache the pointer to the channel (rather than doing the lookup by string).

ktf:
Thank you for this! No objections on my side, just a small comment on the string usage.

Framework/Core/src/DataProcessingHelpers.cxx
matthiasrichter: ok, I will check

davidrohr (dismissed):
This fixes the problem for the full system test.
I would go ahead and merge this soon, since it is an actual improvement. What do we do about the string parameter @ktf commented on? Can we clean this up, or do we do it in another PR?

chiarazampolli:
Hello @matthiasrichter ,

I tried this PR for the aggregator case, and also for that I confirm that the aggregator receives the EOS. 

Cheers,

Chiara

Framework/Core/src/DataProcessingHelpers.cxx
matthiasrichter: Getting back to that. I share your opinion to keep the tree of includes small. Since forward declaration is not possible for std::string we could either switch to `const char*` or introduce a specific class `ChannelName` holding the string and providing initialization and conversion function. But since we want to pass the argument by ref to allow the forward declaration, in the caller we would need to specify `ChannelName(...)`.

matthiasrichter: Ah, maybe I do not need to change the signature of this helper at all. I'm getting the output routes from the device spec, but I can equally good use the output channels.
```
struct DeviceSpec {
  ...
  std::vector<OutputChannelSpec> outputChannels;
  ...
  std::vector<OutputRoute> outputs;
  std::vector<ForwardRoute> forwards;
  ...
}
```
Now I'm thinking its better anyhow to use the `outputChannels` vector and then I have `OutputChannelSpec` as argument.

wiechula:
@matthiasrichter , there are some checks failing. I think it would be nice if this could be merged soon.

Framework/Core/src/DataProcessingHelpers.cxx
matthiasrichter: I have updated the PR to use `outputChannels` instead of `outputRoutes`

matthiasrichter:
I have pushed an update taking the comment of @ktf into account. Test failures have been unrelated, lets see how it goes with the updated PR.

matthiasrichter:
There are again build failures, but the servers seem to be overloaded, the processes are killed
```
```

davidrohr:
@matthiasrichter : Could you solve the merge conflict again :(

davidrohr:
@TimoWilken : Could you restart the macos and the O2/o2 CI?

matthiasrichter:
macOS and o2 Ci are failing unrelated to this PR in `o2sim_G4_checklogs`, also fullCI seems to be stuck, the PR is number 3 in the queue for a long, long time already

ktf:
@TimoWilken could you check what is going on with the fullCI stuff?

matthiasrichter:
How is the merging policy now, are we waiting for the unrelated failures to be fixed?
```
255/366 Test  #80: CCDB/test/testCcdbApi.cxx ......................................................***Failed  300.14 sec
256/366 Test  #81: CCDB/test/testCcdbApi_alien.cxx ................................................***Failed  300.30 sec
257/366 Test  #82: CCDB/test/testBasicCCDBManager.cxx .............................................***Failed  300.70 sec
258/366 Test  #91: Detectors/Base/test/testMatBudLUT.cxx ..........................................***Failed  300.74 sec
322/366 Test #327: o2sim_G4_checklogs .............................................................***Failed    1.19 sec
364/366 Test #323: o2sim_G3 .......................................................................***Failed  Required regular expression not found. Regex=[SIMULATION RETURNED SUCCESFULLY
```

PR #6197, opened 2021-05-18
Have to use FATAL message to catch failure condition in workflow unit…

matthiasrichter:
… tests

At some point the logic in the DPL driver has changed and the workflow
does not terminate with an error code if a device has messages of severity
ERROR.

### DIFF
--- a/Framework/Core/test/test_DataAllocator.cxx
+++ b/Framework/Core/test/test_DataAllocator.cxx
@@ -35,7 +35,7 @@ using namespace o2::framework;
 
 #define ASSERT_ERROR(condition)                                   \
   if ((condition) == false) {                                     \
-    LOG(ERROR) << R"(Test condition ")" #condition R"(" failed)"; \
+    LOG(FATAL) << R"(Test condition ")" #condition R"(" failed)"; \
   }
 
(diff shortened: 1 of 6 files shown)


davidrohr:
For reference: the usage of returnCode=1 in case there was an ERROR but no FATAL was somewhat misleading, since some scripts running multiple workflows in a row are checking the error code and exit if it is non-0.
That led to the case that the single workflow finished, but the collection of workflows didn't.
And even worse, it could spend a lot of CPU time finishing a workflow that had already thrown an error.

matthiasrichter:
The two build failures doesn't seem to be related to the changes in this PR, the builds are keeping on failing, is this a known problem?

davidrohr:
The FullCI problem happens everywhere, no idea what is going on. I investigated it a bit in #6229 but didn't come to any conclusion.
It seems O2/o2 fails since it runs out of memory in the CI container. This is a reocuring problem. @TimoWilken @ktf  : It happens during the compilation of the Analysis. I am not sure whether reducing the number of jobs further would help. We need to fix the memory consumption for analysis.

davidrohr (approved):

matthiasrichter:
ok, thanks for the info

PR #6198, opened 2021-05-18
Fix in ITS digits->raw conversion

shahor02:
Due to the problem in the cables mapping the digits of half-stave 0 on the ITS
layers 3 and 4 were not encoded/decoded


iouribelikov (approved):
@shahor02  Thank you, Ruben.  It seems now to work perfectly fine.

Do I understand correctly, that these changes do not affect anything in the decoding of real raw data ?

shahor02:
@iouribelikov yes, the problem was only in the mapping part used for digit->raw.

PR #6202, opened 2021-05-19
rANS: allow encoders/decoders with empty dictionaries

shahor02:
@MichaelLettrich in some rare cases the certain columns might not get any entries to build dictionaries. Currently in such cases the rANS crashes both in decoding and encoding, instead of using literals only. 
This PR fixes it (tested on data where I impose zero stat on the dictionary using a hack). I am pretty sure the fix is correct for the encoder, but please check the decoder part, the logic there is less trivial

### DIFF
--- a/Utilities/rANS/include/rANS/internal/ReverseSymbolLookupTable.h
+++ b/Utilities/rANS/include/rANS/internal/ReverseSymbolLookupTable.h
@@ -38,11 +38,6 @@ class ReverseSymbolLookupTable
   {
     LOG(trace) << "start building reverse symbol lookup table";
 
-    if (stats.size() == 1) {
-      LOG(warning) << "SymbolStatistics of empty message passed to " << __func__;
-      return;
-    }
-
     mLut.resize(bitsToRange(probabilityBits));
     // go over all symbols
     for (auto symbolIT = std::begin(stats); symbolIT != std::end(stats); ++symbolIT) {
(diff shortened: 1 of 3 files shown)


Utilities/rANS/include/rANS/internal/ReverseSymbolLookupTable.h
```diff
@@ -38,11 +38,6 @@ class ReverseSymbolLookupTable
   {
     LOG(trace) << "start building reverse symbol lookup table";
 
-    if (stats.size() == 1) {
-      LOG(warning) << "SymbolStatistics of empty message passed to " << __func__;
-      return;
-    }
-
     mLut.resize(bitsToRange(probabilityBits));
```
shahor02, line 41: @MichaelLettrich please check here. For empty dictionary (stats.size()==1) we still create LUT of `1<<probabilityBits` size filled by 0s, and this LUT is queried here: https://github.com/shahor02/AliceO2/blob/139aec32c41a94e386c6f21c3dd87e5c85f877d3/Utilities/rANS/include/rANS/LiteralDecoder.h#L70-L73 to check the `isRareSymbol`. So the `streamSymbol` will be always 0 and on top of that with empty dict. any symbol is rare.
So I don't fully understand the logic here.

shahor02:
@MichaelLettrich I am merging this to fix the crash when using dictionary from one data set for compression of other one. The test we have discussed can be done on top of this.

PR #6203, opened 2021-05-19
Fix CMake format and add CI test

vkucera:
As agreed in WP3.
* Obsolete format of `.cmake-format.py` updated.
* Specification of `additional_commands` fixed and updated based on functions defined in `cmake/*.cmake`.
* Adjusted settings to prefer vertical format of `SOURCES`, `PUBLIC_LINK_LIBRARIES`, etc.
* Added a GitHub action to check `CMakeLists.txt` and `*.cmake` files with `cmake-format` and `cmake-lint`.
* Formatting rules applied to all `CMakeLists.txt` and `*.cmake` files.
* Fixed duplicate keyword arguments and trailing whitespaces.

`cmake-lint` still reports many problems, but these have to be fixed manually by the respective code owners or other contributors.

Instructions added in the guidelines: https://github.com/AliceO2Group/CodingGuidelines/pull/26


ktf (changes requested):
Didn't we agree that in any case we want to do incremental changes for this kind of cosmetic stuff?

Unless the checks are green after this PR is merged, I think there is little point in doing a one time merge which creates conflicts and will mark all PRs red afterwards.

Also, why do we need to have an extra line per statement? What is wrong with the current alignment?

CMakeLists.txt
```diff
@@ -10,41 +10,68 @@
...
-  set(_total_analysis_jobs 1)
+set(ALIGPU_BUILD_TYPE
+    "O2")
+
+set(CMAKE_MODULE_PATH
+    ${CMAKE_MODULE_PATH}
+    ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+set_property(
+  GLOBAL
+  PROPERTY REPORT_UNDEFINED_PROPERTIES)
```
ktf: Sorry, but these (and many other) changes are completely unreadable. Who decided this was a good idea?

davidrohr (changes requested):
Do we really want a neline in every `set`?
I.e.
```
set(FOO
    1)
```
is in my opinion much less readable then
```
set(FOO 1)
```

vkucera:
> Do we really want a neline in every `set`?
> I.e.
> 
> ```
> set(FOO
>     1)
> ```
> 
> is in my opinion much less readable then
> 
> ```
> set(FOO 1)
> ```

Fixed in https://github.com/AliceO2Group/AliceO2/pull/6203/commits/22c0f8ee854b52b6920c9a71484e0f6772dc9be1

CMakeLists.txt
vkucera: Fixed in https://github.com/AliceO2Group/AliceO2/pull/6203/commits/22c0f8ee854b52b6920c9a71484e0f6772dc9be1
The only change in the configuration now is `max_pargs_hwrap = 5` -> `max_pargs_hwrap = 1` in order to preserve the current vertical style of `SOURCES`.

Algorithm/CMakeLists.txt
```diff
@@ -10,51 +10,64 @@
...
-
-
-o2_add_test(FlattenRestore
-            SOURCES test/test_FlattenRestore.cxx
-            COMPONENT_NAME Algorithm
-            LABELS algorithm)
+o2_target_man_page(
+  Algorithm
+  NAME Algorithm
+  SECTION 3)
```
ktf: I think the old indentation is still more readable and since the coding convention say that lines up to 120 chars are allowed, what are we trying to fix with this, exactly?

vkucera:
> Didn't we agree that in any case we want to do incremental changes for this kind of cosmetic stuff?

I remember quite clearly that I was explicitly asked at WP3 to apply the format on the CMake files in a single PR and that it was confirmed later when I asked to double-check. The formatting is actually the main point of this PR. The CI tests and fixes are extra.

> Unless the checks are green after this PR is merged, I think there is little point in doing a one time merge which creates conflicts and will mark all PRs red afterwards.

The checks would be green if only `cmake-format` was applied. They are red because the new CI test also runs `cmake-lint` which actually revealed a lot of genuine syntax problems that should be fixed. I fixed the duplicate keywords arguments but I really cannot fix the rest by myself. This will have to be done by others manually later in other PRs.

Algorithm/CMakeLists.txt
vkucera: I suppose you are referring to the [C++ formatting rules](https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html?showone=Line_Length#Line_Length). Those are enforced by `clang-format`.
I am not aware of any document that would describe CMake formatting rules besides `.cmake-format.py` which defines `line_width = 80`

davidrohr:
I can confirm that it was said in CWG to do this in one PR, since it will affect less files than the clang-format for the cxx files.
Still, I'd keep this open for few days to let people dig through the changes, and if there are annoying defaults in the cmake format like the one you just fixed for the `set`, we should clean that up first. We could also separate the pure formatting commit from the rest in separate PRs.

Algorithm/CMakeLists.txt
ktf: 80 would also be fine, but there you have 20...

To me:

```
o2_target_man_page(Algorithm NAME Algorithm SECTION 3)
```

looks infinitely better, frankly. 

IMHO, it's fine to break long lines or to enforce it for things like SOURCES when there is more than a couple of files, but doing everywhere indiscriminately reduces readability.

davidrohr: totally agree, I would keep it as 1-liner where possible. I would only cut ridiculously long lines, (certainly not 80, it is 2021 with 5k screens :)), or some special selected arguments like `SOURCES` if there is more than 1 parameter for that argument.

ktf:
Well 343 files is not exactly few files, but ok, then.

That said, I still think that the indentation choice is a step backwards to what we currently have. 

Can we:

* make sure that the targets stay on the first line.
* not split lists where only a few elements are present

?

aphecetche:
I would also vote for keeping the current indentation (which I find, in most cases, quite adequate) as much as possible. Just using cmake-format to uniformise the syntax of our cmake files should be the goal here IMO.

Honestly, I've not looked at Cmake-format since my initial attempts a while ago (when migrating O2 to "modern" cmake). At the time I decided that Cmake-format was not up to the task, because it lacked the level of flexibility I wanted (or at least I was not able to make it work to my liking). I see the tool has evolved quite a bit so I hope we can now configure it to our needs (and not the reverse, ie. align our syntax to the abilities of the tool)

Algorithm/CMakeLists.txt
ktf: I think 120 was indeed a good compromise between those with big screens and those with thick glasses... ;-)

vkucera:
@ktf @aphecetche @davidrohr I adjusted the line length and splitting thresholds. Please have a look.
In case you want to suggest further changes, please specify the settings you propose.

ktf:
@vkucera It looks to me like before. What changed exactly? I would vote for disabling completely the line splitting and keeping the old indentation. While I can see why one would want `if(something)` to become `if (something)` all the indentation / line splitting changes seem to make the readability much worse.

vkucera:
@ktf `o2_target_man_page(Algorithm NAME Algorithm SECTION 3)` and similar statements are now on a single line, as you wanted.
Can you point to an example of a change you don't like?
Which settings do you propose?
I don't know what idea you have about disabling the splitting, but I suppose that would mean that all `SOURCES` end up on the same line.

ktf:
I think:

```
o2_add_dpl_workflow(alice3-trackselection
                    SOURCES src/alice3-trackselection.cxx
                    PUBLIC_LINK_LIBRARIES O2::Framework O2::AnalysisDataModel O2::AnalysisCore 
                    COMPONENT_NAME Analysis)
```

is much more readable than:

```
o2_add_dpl_workflow(
  alice3-trackselection
  SOURCES src/alice3-trackselection.cxx
  PUBLIC_LINK_LIBRARIES
    O2::Framework
    O2::AnalysisDataModel
    O2::AnalysisCore 
  COMPONENT_NAME Analysis)
```

so my request would be:

* do not have the name of the target on a separate line
* do not split PUBLIC_LINK_LIBRARIES (or sources, as a matter of fact) if it still fits 120 chars.

I have no idea how to do that with `cmake-format`, but it's not me trying to impose the use of the tool, so I think it's not up to me to come up with a solution.

davidrohr:
I agree with @ktf that the previous style was more readable. I have no idea how configurable cmake-format is and whether we can easily achieve what is suggested, but then I would not apply a format which we do not really want to use only because the tool doesn't support what we actually want.

vkucera:
@ktf I am not trying to impose the use of the tool either.
Use of `cmake-format` is endorsed in the O2 docs and it has been [confirmed in WP3](https://alice.its.cern.ch/jira/browse/O2-2086).

> Formatting CMake files - cmake-format should be used, but it is not enforced; see doc instructions in https://github.com/AliceO2Group/AliceO2/blob/dev/doc/CMakeInstructions.md

My proposal and my task (as agreed in WP3) include applying a style on the existing code, enforcing the style by a CI test, and providing instructions on how to apply the style automatically to pass the test.

I am not advocating for any particular style or tool. I took the recommended tool, used it with the existing configuration and adjusted the parameters based on your feedback.
If you think the configuration should be adjusted further or you know a better formatting tool, please suggest a solution.

ktf:
I am happy if someone provides a **better** style and a tool to enforce it in the CI. Here, the solution you propose results in a not so readable style and I personally would prefer no automation if the result is the one proposed.
If this PR is the only possible solution using `cmake-format`, I think we should simply go back to the drawing board and revaluate the tool and the task itself.

vkucera:
not stale

PR #6215, opened 2021-05-20
DPL: hotfix for resource manager

ktf:
* Demote some messages to debug.
* Properly handle more than one reader.


ktf:
Currently broken for more than one reader. Tested separate on my mac and on the train machine.

PR #6216, opened 2021-05-20
sim_challenge: fix in VZ generation and ITS reco options in pp

shahor02:


shahor02:
trivial fixes in the script, not tested by CI, merging.

PR #6218, opened 2021-05-20
DPL: properly handle SIGTERM

ktf:


ktf:
@Barthelemy indeed this is missing to properly handle sigterm. Somehow in the past it was creating troubles, but now seems to be working just fine.

Barthelemy:
@ktf  Thanks for the very fast response.

ktf:
@Barthelemy it's actually ages that I needed to fix this, sorry about that...

PR #6222, opened 2021-05-20
dcs-config-proxy and sample consumer workflow

shahor02:


ktf:
Is this ready to be tested?

shahor02:
Yes, modulo that it contains the fix for parsing `--channle-config` with multiple channels, which we could delegate to FMQ if the problem we were discussing privately could be solved. Anyway, one does not exclude another, removing the WIP.

ktf (approved):

shahor02:
the failures are unrelated to PR, merging. The further investigation of " "-concatenated channels problem can be continued with dev.

PR #6223, opened 2021-05-21
Sort tracks table by collisionID

nburmaso:


sawenzel (dismissed):

sawenzel (dismissed):

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
```diff
@@ -193,12 +197,20 @@ class AODProducerWorkflowDPL : public Task
...
-  void fillTracksTable(const TTracks& tracks, std::vector<int>& vCollRefs, const TTracksCursor& tracksCursor,
-                       const TTracksCovCursor& tracksCovCursor, const TTracksExtraCursor& tracksExtraCursor, int trackType);
-
-  template <typename TTracks, typename TTracksCursor>
-  void fillMFTTracksTable(const TTracks& tracks, std::vector<int>& vCollRefs, const TTracksCursor& tracksCursor);
+  template <typename TMFTCursor, typename TTracksCursor, typename TTracksCovCursor, typename TTracksExtraCursor>
+  void fillTracksTable(gsl::span<const o2::its::TrackITS>& tracksITS,
+                       gsl::span<const o2::mft::TrackMFT>& tracksMFT,
+                       gsl::span<const o2::dataformats::TrackTPCITS>& tracksITSTPC,
+                       gsl::span<const o2::tpc::TrackTPC> tracksTPC,
```
shahor02, line 204: note that we will have ~15 different track types and they will be all available from the reco container as e.g. `recoContainer::getSpan<o2::its::TrackITS>(GlobalTrackID::ITS)` or via aliases `getITSTracks()` etc., so better to pass directly the RecoContainer.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -121,119 +121,214 @@ uint64_t AODProducerWorkflowDPL::getTFNumber(uint64_t firstVtxGlBC, int runNumbe
...
+  float tofChi2 = -999.f;
+  float tpcSignal = -999.f;
+  float trdSignal = -999.f;
+  float tofSignal = -999.f;
+  float length = -999.f;
+  float tofExpMom = -999.f;
+  float trackEtaEMCAL = -999.f;
+  float trackPhiEMCAL = -999.f;
+
+  for (int i = 0; i < vTracksIdx.size(); i++) {
```
shahor02, line 139: Might be my ignorance, but how the tracks redundancy (the fact that e.g. ITS tracks is already accounted in the ITS-TPC) are accounted.I don't see the references of "more global" tracks to contributing simpler tracks.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -121,119 +121,214 @@ uint64_t AODProducerWorkflowDPL::getTFNumber(uint64_t firstVtxGlBC, int runNumbe
...
+                          truncateFloatFraction(tofExpMom, mTrack1Pt),
+                          truncateFloatFraction(trackEtaEMCAL, mTrackPosEMCAL),
+                          truncateFloatFraction(trackPhiEMCAL, mTrackPosEMCAL));
+      }
+      if (trackType == GIndex::Source::TPC) {
+        auto& track = tracksTPC[vTracksIdx[i][0]];
+        collisionID = vCollRefsTPC[vTracksIdx[i][0]];
+        tpcChi2NCl = track.getNClusters() ? track.getChi2() / track.getNClusters() : 0;
+        tpcSignal = track.getdEdx().dEdxTotTPC;
+        tpcNClsFindable = track.getNClusters();
```
shahor02, line 232: just for the record: `tpcNClsFindable` is not the same as `track.getNClusters()`, I assume this placeholder at the moment.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -121,119 +121,214 @@ uint64_t AODProducerWorkflowDPL::getTFNumber(uint64_t firstVtxGlBC, int runNumbe
...
+                        (Char_t)(128. * track.getSigmaTglZ() / track.getSigmaTgl2() / track.getSigmaZ2()),
+                        (Char_t)(128. * track.getSigmaTglSnp() / track.getSigmaTgl2() / track.getSigmaSnp2()),
+                        (Char_t)(128. * track.getSigma1PtY() / track.getSigma1Pt2() / track.getSigmaY2()),
+                        (Char_t)(128. * track.getSigma1PtZ() / track.getSigma1Pt2() / track.getSigmaZ2()),
+                        (Char_t)(128. * track.getSigma1PtSnp() / track.getSigma1Pt2() / track.getSigmaSnp2()),
+                        (Char_t)(128. * track.getSigma1PtTgl() / track.getSigma1Pt2() / track.getSigmaTgl2()));
+        tracksExtraCursor(0,
+                          truncateFloatFraction(tpcInnerParam, mTrack1Pt),
+                          flags,
+                          itsClusterMap,
```
shahor02, line 262: you are using here the last value set for the unrelated ITS track?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -121,119 +121,214 @@ uint64_t AODProducerWorkflowDPL::getTFNumber(uint64_t firstVtxGlBC, int runNumbe
...
+                          tpcNClsFindable,
+                          tpcNClsFindableMinusFound,
+                          tpcNClsFindableMinusCrossedRows,
+                          tpcNClsShared,
```
shahor02, line 316: same here: these variable will  have the values of last unrelated track which set it.

shahor02 (changes requested):

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
nburmaso: Hi @shahor02. Yes, I understand. I have made this PR only to implement track sorting and to test it separately. I can fix this after #6170 is merged.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Yes, redundant tracks are not accounted for in this code. Same as in previous comment: I think I need RecoContainer for this to get only more 'global' tracks. So I can implement it after #6170 is merged, where I added RecoContainer to the producer.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Yes, this is kind of a placeholder now

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Thank you for noticing, I have fixed this in the latest commit

sawenzel:
@nburmaso : Did you check if the mcHistogram analysis working after this PR? If yes, I would be in favour of merging (the transition to reco-container is done in a separate development I think).

nburmaso:
Changes from here are now in #6170, so I close this PR

PR #6150, opened 2021-05-13
CCDB populator will match payload and wrapper according to DataDescription

chiarazampolli:
This allows to send to the CCDB populator multiple CCDB objects at the same time. The pairing paylod <--> wrapper will be done by the framework based on the Data Description.

PR #6230, opened 2021-05-21
MFT Noise calibration

mcoquet642:
Implementing noise scan calibration for MFT following the lines of ITS writing a single NoiseMap object in the CCDB. Additional features include an option **--useDigits** to make the device run in mode based on digits rather than clusters (default mode).


Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx
```diff
@@ -0,0 +1,181 @@
...
+  LOG(INFO) << "Sending object " << info.getPath() << "/" << info.getFileName()
+            << " of size " << image->size()
+            << " bytes, valid for " << info.getStartValidityTimestamp()
+            << " : " << info.getEndValidityTimestamp();
+
+  using clbUtils = o2::calibration::Utils;
+  output.snapshot(
+    Output{clbUtils::gDataOriginCLB, clbUtils::gDataDescriptionCLBPayload, 0}, *image.get());
+  output.snapshot(
+    Output{clbUtils::gDataOriginCLB, clbUtils::gDataDescriptionCLBInfo, 0}, info);
```
shahor02: Please note that the convention for the messages sent to CCDB populator was changed since https://github.com/AliceO2Group/AliceO2/pull/6150 in order to avoid clashes between the inputs from different calibrators. The documentation was not fully updated, just pushed https://github.com/AliceO2Group/AliceO2/pull/6234.
Now it should be  e.g.
````
output.snapshot(Output{clbUtils::gDataOriginCDBPayload, "MFT_NoiseMap", 0}, *image.get());
output.snapshot(Output{clbUtils::gDataOriginCDBWrapper, "MFT_NoiseMap", 0}, info);
````

File: Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx [hunk 2]
```diff
@@ -0,0 +1,181 @@
...
+    inputs.emplace_back("patterns", detOrig, "PATTERNS", 0, Lifetime::Timeframe);
+    inputs.emplace_back("ROframes", detOrig, "CLUSTERSROF", 0, Lifetime::Timeframe);
+  }
+
+  using clbUtils = o2::calibration::Utils;
+  std::vector<OutputSpec> outputs;
+  outputs.emplace_back(
+    ConcreteDataTypeMatcher{clbUtils::gDataOriginCLB, clbUtils::gDataDescriptionCLBPayload});
+  outputs.emplace_back(
+    ConcreteDataTypeMatcher{clbUtils::gDataOriginCLB, clbUtils::gDataDescriptionCLBInfo});
```
shahor02: Respectively, in the getNoiseCalibratorSpec the outputs should be declared as 
````
outputs.emplace_back(ConcreteDataTypeMatcher{clbUtils::gDataOriginCDBPayload, "MFT_NoiseMap"});
outputs.emplace_back(ConcreteDataTypeMatcher{clbUtils::gDataOriginCDBWrapper, "MFT_NoiseMap"});
````

File: Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx [hunk 3]
```diff
@@ -0,0 +1,181 @@
...
+
+  return DataProcessorSpec{
+    "mft-noise-calibrator",
+    inputs,
+    outputs,
+    AlgorithmSpec{adaptFromTask<NoiseCalibratorSpec>(useDigits)},
+    Options{
+      {"1pix-only", VariantType::Bool, false, {"Fast 1-pixel calibration only"}},
```
shahor02: What is the purpose of this option? It will miss the noisy channels which happen to be neighbors (e.g. noisy column).

Detectors/ITSMFT/MFT/calibration/src/NoiseSlotCalibrator.cxx
```diff
@@ -0,0 +1,147 @@
...
+namespace o2
+{
+using Slot = calibration::TimeSlot<o2::itsmft::NoiseMap>;
+
+namespace mft
+{
+bool NoiseSlotCalibrator::processTimeFrame(gsl::span<const o2::itsmft::Digit> const& digits,
+                                           gsl::span<const o2::itsmft::ROFRecord> const& rofs)
+{
+  calibration::TFType nTF = rofs[0].getBCData().orbit / mHBFperTF;
```
shahor02: The calibration may start from any orbit, so the nTF in general will not reflect the TFID. In other time-slot-based calibrations we are using `auto tfcounter = o2::header::get<o2::framework::DataProcessingHeader*>(pc.inputs().get("input").header)->startTime;`. This is also not really correct, but at least, once we come to agreement on the common TF counter definition, we can change all centrally. I would suggest to use the same: extract the TF counter in the CalibrationSpec and pass it to the Calibrator, e.g. like in `https://github.com/AliceO2Group/AliceO2/blob/c2f765d42f324beb1e1ecda9cf5ba4acf29704cd/Detectors/TOF/calibration/testWorkflow/LHCClockCalibratorSpec.h#L51-L54`
Then, you also will not need the `mHBFperTF`.
Actually, for you the specific slot should be irrelevant, since you have just 1 with 0 - infinity span.

Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx
```diff
@@ -0,0 +1,181 @@
...
+  LOG(INFO) << "Setting the probability threshold to " << probT;
+  auto HBperTF = ic.options().get<int>("hb-per-tf");
+  LOG(INFO) << "Nb of HBF per TF used : " << HBperTF;
+
+  mPath = ic.options().get<std::string>("path");
+  mMeta = ic.options().get<std::string>("meta");
+  mStart = ic.options().get<int64_t>("tstart");
+  mEnd = ic.options().get<int64_t>("tend");
+
+  mCalibrator = std::make_unique<CALIBRATOR>(onepix, probT, HBperTF);
```
shahor02: You should set the length of the calibration slot, i.e. in your case infinite (e.g. `std::numeric_limits<TFType>::max`) since you have just 1 slot.

shahor02 (changes requested):
Hi @mcoquet642 , please see some comments below.
This version of time-slot based calibration is essentially similar to that of ITS, where the slot receives the digits or clusters fills the map. As such, it can aggregate the data both from the EPNs and from FLPs (if the decoding/clusterization is done there).
An alternative, as we discussed at the meeting, might be the FLP processing which builds the map for the chips seen on each FLP, then the time-slot-calibration receives the maps from different FLPs, merges them and sends to the CCDB populator.

Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx
mcoquet642: Hi Ruben, thanks for the comments. This is now corrected.

Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx
mcoquet642: This was a remanent of what was in ITS code, now removed.

Detectors/ITSMFT/MFT/calibration/src/NoiseSlotCalibrator.cxx
mcoquet642: Thanks for pointing this out. Indeed I had some troubles with this. I think the same line is in the ITS code.

Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx
mcoquet642: Just a comment : by default this PR doesn't use the NoiseSlotCalibrator class but the NoiseCalibrator which doesn't use TimeSlots, as Jouri mentioned during the discussion that in the end ITS doesn't use the former. So I included the appropriate line but commented it out.

Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx
```diff
@@ -0,0 +1,174 @@
...
+  auto probT = ic.options().get<float>("prob-threshold");
+  LOG(INFO) << "Setting the probability threshold to " << probT;
+
+  mPath = ic.options().get<std::string>("path");
+  mMeta = ic.options().get<std::string>("meta");
+  mStart = ic.options().get<int64_t>("tstart");
+  mEnd = ic.options().get<int64_t>("tend");
+
+  mCalibrator = std::make_unique<CALIBRATOR>(probT);
+  //  mCalibrator->setSlotLength(std::numeric_limits<TFType>::max); //For TimeSlot calibration
```
shahor02: Since you never need multiple slots, why don't you set it in the NoiseSlotCalibrator constructor?

Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx
```diff
@@ -0,0 +1,174 @@
...
+
+    return pairs;
+  };
+  std::map<std::string, std::string> meta;
+  auto keyvalues = toKeyValPairs(Str::tokenize(mMeta, ';', true));
+
+  // fill meta map
+  for (auto& p : keyvalues) {
+    meta[p.first] = p.second;
+  }
```
shahor02: Why don't you do directly:
````
std::map<std::string, std::string> meta;
auto toKeyValPairs = [&meta](std::vector<std::string> const& tokens) {
  for (auto&amp; token : tokens) {
     auto keyval = Str::tokenize(token, '=', false);
     if (keyval.size() != 2) {
        LOG(ERROR) << "Illegal command-line key/value string: " << token;
        continue;
     }
     Str::trim(keyval[1]);
     meta[keyval[0]] = keyval[1];
  }
};
toKeyValPairs(Str::tokenize(mMeta, ';', true));
````

Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx
mcoquet642: Should be done now.

Detectors/ITSMFT/MFT/calibration/src/NoiseCalibratorSpec.cxx
mcoquet642: Thanks, indeed. Now fixed.

shahor02 (approved):
approving to trigger CI check

bovulpes (approved):

shahor02:
Squashing/merging, CI failures are not related to the PR.

@TimoWilken all 3 required tests are regularly failing: macos does not start, in o2 the compiler is killed, and fullCI fails at random stages, I suspect due to the insufficient memory.

PR #6233, opened 2021-05-23
Suppress leftover PHOSCalib directory [fix for PR6187]

shahor02:
@peressounko there was and obsolete `BadChannelMap.h` in PHOSCalib dir., which should be suppressed, on macos CI it was leading to `ninja: error: 'stage/lib/G__O2PHOSCalib_rdict.pcm', needed by 'Detectors/PHOS/reconstruction/G__O2PHOSReconstruction.cxx', missing and no known rule to make it`


shahor02:
merging as the fix is trivial and w/o it macOS CI is broken.

peressounko:
This file should be kept untill QC will be updated to point to new include of BadMap.h. I am working on QC update now

shahor02:
@peressounko but then you should probably rename the new BadChannelMap.h, otherwise you have 2 classes with same signature in the same namespace, which confuses the rootcling on the macos.
E.g. you can rename the BadChannelMap in the DataFormats to `BadChannelsMap` (and use it in the patch for the QC) and revert the this PR. Then, once the QC PR will be merged, again suppress this PHOSCalib directory.

peressounko:
But how should be proceed now? I expect that QC will not compile with present code. I can prepare new PR with renamed BadMaps in DataFormats, is it OK?

peressounko:
@shahor02, may be it is easier to configure makefile to copy header to include/PHOSCalib/BadChannelsMap.h in addition to normal path? But I do not know how to do this.

shahor02:
I think it is not the problem of the makefile but the peculiarity of the rootcling: when generating the rootmap for the O2PHOSReconstruction it detects that there is a BadChannelMap in the rootmap of the PHOSCalib which is not included. The easiest would be to change the name of the BadChannelMap in the DataFormats.

PR #6241, opened 2021-05-25
Fix parsing for InputSpec with DataOrigin only

shahor02:
@ktf this fixes the problem reported by @chiarazampolli and @peressounko : addition of the fake `CLP/PEDDIFFLG` input to CCDB populator.


ktf:
What was the actual problem? having the subspecification?

chiarazampolli:
Yes, and two conditions were fulfilled.

shahor02:
Actually, after creation of the inputSpec for the DataOrigin only, because of the lack of the `else` it was creating extra inputs with DataDescription of the previous iteration (although the hasDescription is false).

chiarazampolli:
Yes, this is what I meant, that 2 InputSpecs were created since 2 conditions were fulfilled.

PR #6251, opened 2021-05-26
[MFT] Add a geometry misaligner 

robincaron13:
This PR contains the geometry misaligner class, which is able to misalign the existing MFT geometry. Basic shifts and rotations are allowed on: half-cone, disk, ladder, sensor. 
These functions have been tested through simple simulations (o2-sim-serial -m MFT -e TGeant3 -n 1). The resulting misaligned geometry can be available as a simple output root file (just after closing the geometry with 'gGeoManager->RefreshPhysicalNodes()' after applying misalignments).

### DIFF
--- a/Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
+++ b/Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
@@ -18,10 +18,8 @@
 
 #include "MFTBase/Geometry.h"
 #include "MFTBase/GeometryTGeo.h"
-#include "MFTBase/MFTBaseParam.h"
 
 #include "MFTSimulation/Detector.h"
-#include "MFTSimulation/GeometryMisAligner.h"
 
 #include "Field/MagneticField.h"
 #include "SimulationDataFormat/Stack.h"
@@ -707,42 +705,6 @@ void Detector::addAlignableVolumesChip(Int_t hf, Int_t dk, Int_t lr, Int_t ms,
   if (!gGeoManager->SetAlignableEntry(sname, path.Data(), uid)) {
     LOG(FATAL) << "Unable to set alignable entry ! " << sname << " : " << path;
   }
-}
-
-//_____________________________________________________________________________
-void Detector::Misaligner() const
-{
-  // produce misalignment parameters stored into local file or CCDB
-  auto& mftBaseParam = MFTBaseParam::Instance();
-
-  const std::string& ccdbHost = "http://ccdb-test.cern.ch:8080";
-  long tmin = 0;
-  long tmax = -1;
-  const std::string& objectPath = "";
-  const std::string& fileName = "MFTAlignment.root";
-
-  // Initialize the misaligner
-  o2::mft::GeometryMisAligner aGMA;
-
-  if (mftBaseParam.misalignHalf) {
-    aGMA.SetHalfCartMisAlig(0., mftBaseParam.xHalf, 0., mftBaseParam.yHalf, 0., mftBaseParam.zHalf);
-    aGMA.SetHalfAngMisAlig(0., mftBaseParam.psiHalf, 0., mftBaseParam.thetaHalf, 0., mftBaseParam.phiHalf);
-  }
-
-  if (mftBaseParam.misalignDisk) {
-    aGMA.SetDiskCartMisAlig(0., mftBaseParam.xDisk, 0., mftBaseParam.yDisk, 0., mftBaseParam.zDisk);
-    aGMA.SetDiskAngMisAlig(0., mftBaseParam.psiDisk, 0., mftBaseParam.thetaDisk, 0., mftBaseParam.phiDisk);
-  }
-  if (mftBaseParam.misalignLadder) {
-    aGMA.SetLadderCartMisAlig(0., mftBaseParam.xLadder, 0., mftBaseParam.yLadder, 0., mftBaseParam.zLadder);
-    aGMA.SetLadderAngMisAlig(0., mftBaseParam.psiLadder, 0., mftBaseParam.thetaLadder, 0., mftBaseParam.phiLadder);
-  }
-  if (mftBaseParam.misalignSensor) {
-    aGMA.SetSensorCartMisAlig(0., mftBaseParam.xSensor, 0., mftBaseParam.ySensor, 0., mftBaseParam.zSensor);
-    aGMA.SetSensorAngMisAlig(0., mftBaseParam.psiSensor, 0., mftBaseParam.thetaSensor, 0., mftBaseParam.phiSensor);
-  }
-
-  aGMA.MisAlign(false, ccdbHost, tmin, tmax, objectPath, fileName);
 }
 
 //_____________________________________________________________________________
--- a/Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
+++ b/Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
@@ -20,6 +20,7 @@
 #include "MFTBase/GeometryTGeo.h"
 
 #include "MFTSimulation/Detector.h"
+#include "MFTSimulation/GeometryMisAligner.h"
 
 #include "Field/MagneticField.h"
 #include "SimulationDataFormat/Stack.h"
@@ -709,6 +710,36 @@ void Detector::addAlignableVolumesChip(Int_t hf, Int_t dk, Int_t lr, Int_t ms,
   if (!gGeoManager->SetAlignableEntry(sname, path.Data(), uid)) {
     LOG(FATAL) << "Unable to set alignable entry ! " << sname << " : " << path;
   }
+}
+
+//_____________________________________________________________________________
+void Detector::MisalignGeometry() const
+{
+  // Function to misalign the MFT geometry
+
+  // Initialize the misaligner
+  o2::mft::GeometryMisAligner aGMA;
+
+  aGMA.SetHalfCartMisAlig(0., 0., 0., 0., 0., 0.); // half-MFT translated on X, Y, Z axis
+  aGMA.SetHalfAngMisAlig(0., 0., 0., 0., 0., 0.);  // half-MFT  rotated on Z, X, Y axis
+    
+  aGMA.SetDiskCartMisAlig(0., 0., 0., 0., 0., 0.); // Half-disks translated on X, Y, Z axis
+  aGMA.SetDiskAngMisAlig(0., 0., 0., 0., 0., 0.);  // Half-disks rotated on Z, X, Y axis
+    
+  aGMA.SetLadderCartMisAlig(0., 0., 0., 0., 0., 0.); // Ladders translated on X, Y, Z axis
+  aGMA.SetLadderAngMisAlig(0., 0., 0., 0., 0., 0.);   // Ladders rotated on Z, X, Y axis
+    
+  aGMA.SetSensorCartMisAlig(0., 0., 0., 0., 0., 0.); // Sensors translated on X, Y, Z axis
+  aGMA.SetSensorAngMisAlig(0., 0., 0., 0., 0., 0.);   // Sensors rotated on Z, X, Y axis
+
+  aGMA.MisAlign(true);  // Misalign the geometry
+
+  // lock the geometry, or unlock with "false"
+  gGeoManager->RefreshPhysicalNodes();
+
+  // store the misaligned geometry in a root file
+  gGeoManager->Export("o2sim_geometry_misaligned.root");
+        
 }
 
 //_____________________________________________________________________________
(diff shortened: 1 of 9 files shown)


shahor02:
@robincaron13 the errors reported by FullCI are real:
````
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/O2/6251/0/Detectors/ITSMFT/MFT/simulation/src/GeometryMisAligner.cxx:174:21: error: use '= default' to define a trivial destructor [modernize-use-equals-default]
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/O2/6251/0/Detectors/ITSMFT/MFT/simulation/src/GeometryMisAligner.cxx:190:9: error: statement should be inside braces [readability-braces-around-statements]
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/O2/6251/0/Detectors/ITSMFT/MFT/simulation/src/GeometryMisAligner.cxx:202:9: error: statement should be inside braces [readability-braces-around-statements]
````

File: Detectors/ITSMFT/MFT/simulation/src/Detector.cxx [hunk 1]
```diff
@@ -624,6 +625,7 @@ void Detector::addAlignableVolumes() const
   for (Int_t hf = 0; hf < nHalf; hf++) {
     addAlignableVolumesHalf(hf, path, lastUID);
   }
+  MisalignGeometry();
```
rpezzi: Hi @robincaron13 , notice that any simulation is calling MisalignGeometry() and saving `o2sim_geometry_misaligned.root`. This method should be called only in special circumstances, thus a configurable parameter is desirable.

Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
```diff
@@ -710,6 +712,37 @@ void Detector::addAlignableVolumesChip(Int_t hf, Int_t dk, Int_t lr, Int_t ms,
...
+  aGMA.SetHalfAngMisAlig(0., 0., 0., 0., 0., 0.);  // half-MFT  rotated on Z, X, Y axis
+
+  aGMA.SetDiskCartMisAlig(0., 0., 0., 0., 0., 0.); // Half-disks translated on X, Y, Z axis
+  aGMA.SetDiskAngMisAlig(0., 0., 0., 0., 0., 0.);  // Half-disks rotated on Z, X, Y axis
+
+  aGMA.SetLadderCartMisAlig(0., 0., 0., 0., 0., 0.); // Ladders translated on X, Y, Z axis
+  aGMA.SetLadderAngMisAlig(0., 0., 0., 0., 0., 0.);  // Ladders rotated on Z, X, Y axis
+
+  aGMA.SetSensorCartMisAlig(0., 0., 0., 0., 0., 0.); // Sensors translated on X, Y, Z axis
+  aGMA.SetSensorAngMisAlig(0., 0., 0., 0., 0., 0.);  // Sensors rotated on Z, X, Y axis
```
rpezzi: These parameters should be configurable, otherwise one needs to recompile O2 to produce any misalignment.

Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
```diff
@@ -710,6 +712,37 @@ void Detector::addAlignableVolumesChip(Int_t hf, Int_t dk, Int_t lr, Int_t ms,
...
+
+  aGMA.MisAlign(false); // Misalign the geometry ('option'=verbose)
+
+  // lock the geometry, or unlock with 'false'
+  //gGeoManager->RefreshPhysicalNodes();
+
+  // store the misaligned geometry in a root file
+  gGeoManager->Export("o2sim_geometry_misaligned.root");
```
rpezzi: if this file is relevant only for the MFT, its filename should refer to it. Also, this file may not contain the final geometry, since other detectors may not be ready at this point. The usual `o2sim_geometry.root` won't contain the complete detector including misalignment?

rpezzi (changes requested):
@robincaron13 , thanks for your contribution with detailed comments. 
Some things have to be clarified:
1. configurable options to enable the misaligner and  its configuration. 
2. the need of another geometry file.

Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
shahor02: The geometry misalignment will be handled centrally, I will open a PR today or tomorrow. The Detector should always produce an ideal geometry.

rpezzi: Hi @shahor02 , 
So we drop calling `MisalignGeometry()` here, but keep its implementation on `Detector.cxx`?

shahor02: The geometry should be not misaligned on the construction level. Instead, the "misaligner" should provide a vector of AlignParam objects, which are fed to `o2::base::GeometryManager::applyAlignment` method, which modifies ideal geometry. 
Note that I've updated many classes, so the interface changed, should push it after some more work: the PR https://github.com/AliceO2Group/AliceO2/pull/6287 will be amended by https://github.com/shahor02/AliceO2/commits/pr_algpar1.

shahor02:
please see https://github.com/AliceO2Group/AliceO2/pull/6287 , particularly the https://github.com/AliceO2Group/AliceO2/blob/b778332365c064825e78a6bc7ad0cecdfcbb0ee7/Detectors/Base/README.md
Essentially, the misaligner shall generate a `vector<AlignParam>`, with the output similar to that of https://github.com/AliceO2Group/AliceO2/blob/b778332365c064825e78a6bc7ad0cecdfcbb0ee7/Detectors/ITSMFT/ITS/macros/test/ITSMisaligner.C

shahor02:
Hi @robincaron13 
What is the context of this PR? I see you pushed changes after my https://github.com/AliceO2Group/AliceO2/pull/6287 and loaded some objects on the test ccdb, but your Detector still contains the Misalign method and the Misaligner seem to apply the alignment to the geom instead of just producing a vector of params.

Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
robincaron13: Thanks for your messages and the updated classes. Now this line has been droped.

Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
robincaron13: Thanks, sorry for the late reply, now I added the associated configurable parameters (using the MFTBaseParam.h).

Detectors/ITSMFT/MFT/simulation/src/Detector.cxx
robincaron13: I droped this line, since the misaligner should only produce an output file containing `vector<AlignParam>`.

javierecc:
Hi @shahor02, All,
I agree that the MisAligner should only produce a CCDB entry to be used later. Some changes are still needed in my opinion, we will come back.
I have however one remark about the Misaligner applying the alignments. This is indeed not the code to be used to apply the produced misalignments during a simulation, but we need to apply "temporarily" the misalignments to the geometry to produce reasonable misalignments. The reason is:
- AlignParam store parameters in global coordinates
- Generating misalignments in global coordinates is not practical because the same value will not have the same effect depending on the position of the element
- Thus, I would prefer to generate misalignments in local coordinates 
- AlignParam need the geometry when setting its parameters in local coordinates
- For hierarchical geometries as the MFT we need to first misalign the mother volume and propagate it to the geometry before misaligning the daughter volume so that the calculated global misalignment of the daughter volume makes sense

Does that make sense?
Cheers,
Javier

shahor02:
Hi @javierecc 

> This is indeed not the code to be used to apply the produced misalignments during a simulation, but we need to apply "temporarily" the misalignments to the geometry to produce reasonable misalignments. The reason is:

Sorry, I did not understand the point about the temporary misalignment, could you elaborate?

>     * AlignParam store parameters in global coordinates
> 
Yes, as it was done in the Aliroot, the meaning of parameters should be the same.

>     * Generating misalignments in global coordinates is not practical because the same value will not have the same effect depending on the position of the element
> 
>     * Thus, I would prefer to generate misalignments in local coordinates

You can extract them in local frame, the AlignParam constructor will take care of moving them to the global one.

>     * AlignParam need the geometry when setting its parameters in local coordinates

Does this pose a problem?
> 
>     * For hierarchical geometries as the MFT we need to first misalign the mother volume and propagate it to the geometry before misaligning the daughter volume so that the calculated global misalignment of the daughter volume makes sense
> 

This is taken care by the way the alignment is applied: AlignParam::applyToGeometry: the GeometryManager::applyAlignment
sorts the params in the level decreasing order and calls the `AlignParam::applyToGeometry` in this order.
In short, this is supposed to be an exact copy of what we had in the AliRoot. Is there a reason to change this?

shahor02:
PS: there is a format problem

javierecc:
Hi @shahor02
On 9 Jun 2021 at 16:27 +0200, Ruben Shahoyan ***@***.***>, wrote:
	Hi @javierecc
	This is indeed not the code to be used to apply the produced misalignments during a simulation, but we need to apply "temporarily" the misalignments to the geometry to produce reasonable misalignments. The reason is:
Sorry, I did not understand the point about the temporary misalignment, could you elaborate?
Delta^g = G Delta^l G^-1, with G = M L, where M is the transformation of the mother volume and L the transformation of the daughter volume inside the mother volume.

So if the transformation G is not the same when you produce Delta^g than when you apply it to your geometry you won’t get the intended result. And that is what would happen if you produce the Delta^g from Delta^l using and ideal geometry but apply it to a geometry that you are misaligning hierarchically as we are doing.
Note that I agree, AlignParam is correctly implemented (and it was already like that in AliRoot) and that the application of the misalignment parameters is correct also. I’m just saying that we should pay attention to it when generating misalignments from Delta_l. We need to first apply the misalignment to the mother volume before getting Delta^g from Delta^l for a daughter volume. But this should be done in the misalignment generation code.
		* AlignParam store parameters in global coordinates
Yes, as it was done in the Aliroot, the meaning of parameters should be the same.
Agree.
		* Generating misalignments in global coordinates is not practical because the same value will not have the same effect depending on the position of the element

* Thus, I would prefer to generate misalignments in local coordinates
You can extract them in local frame, the AlignParam constructor will take care of moving them to the global one.
Agree, but you need to have the right geometry for that
		* AlignParam need the geometry when setting its parameters in local coordinates
Does this pose a problem?
No, indeed.
		* For hierarchical geometries as the MFT we need to first misalign the mother volume and propagate it to the geometry before misaligning the daughter volume so that the calculated global misalignment of the daughter volume makes sense
This is taken care by the way the alignment is applied: AlignParam::applyToGeometry: the GeometryManager::applyAlignment
sorts the params in the level decreasing order and calls the AlignParam::applyToGeometry in this order.
Right, but for that the misalignment parameters need to have been produced properly following the same sorting and with the same misaligned transformations for the mother volumes.
	 In short, this is supposed to be an exact copy of what we had in the AliRoot. Is there a reason to change this?
Also agree. There is no reason to change. We were already doing what I describe in the misaligner code for muon.

In short, AlignParam has no problem. Our misalignment generation code needs to take care of generating the misalignments of the mother volumes and apply them before generating the misalignment of the daughter volumes. I said “temporarily", because once all misalignment parameters are generated and stored in the CCDB we forget about it and just let the framework take the CCDB and handle the misalignments.

Hope it is clearer and makes sense.
Cheers,
Javier

shahor02:
Hi @javierecc 

I know what is the difference between the global and local deltas, I was confused with your "temporary alignment". 
We always were doing the new alignment incrementally to existing one (as far as I understand, this is what you call "temporary") and extracted corrections were added on top of this initial alignment (obviously, this is not a simple addition).
The AlignParams created by any (mis)alignment code should be applicable to ideal geometry (which can be always loaded by setting the `Aligner.mDetectors` to "none").

So, all what you say is correct, but I don't understand your comment `this is indeed not the code to be used to apply the produced misalignments during a simulation`, does not it create full misalignment?

javierecc:
Hi @shahor02,

Good then we all agree and there was just some misunderstanding. 
The comment "this is indeed not the code to be used to apply the produced misalignments during a simulation" was related to the initial version in which, as a workaround, we were generating misalignments and applying them on the flight during creation of the geometry. That was a workaround because the reading of misalignments from CCDB was not yet in place. Now it is and thus this misaligner code should just be used to produce a CCDB entry.  
Robin made must of the needed changes, but some fixes are still needed. We will come back to this.

Cheers,
Javier

shahor02:
Hi @javierecc 
ok, now understood :)
Cheers,
 Ruben

rpezzi (dismissed):

rpezzi (approved):

rpezzi:
Hello, this PR should be good to go. CI failures are not related. Please merge.

PR #6252, opened 2021-05-26
Drop test for symbol names

ktf:
This seems to break on Big Sur, confusing XCode about a missing nm
and prompting for command line tool installation.

### DIFF
--- a/Detectors/MUON/MCH/Mapping/Impl3/CMakeLists.txt
+++ b/Detectors/MUON/MCH/Mapping/Impl3/CMakeLists.txt
@@ -39,12 +39,3 @@ set_target_properties(${targetName} PROPERTIES CXX_VISIBILITY_PRESET hidden)
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
         target_compile_options(${targetName} PRIVATE -fext-numeric-literals)
 endif()
-if(APPLE)
-        get_filename_component(
-                script ${CMAKE_CURRENT_LIST_DIR}/../../../check_nof_exported_symbols.sh
-                ABSOLUTE)
-        add_custom_command(
-                TARGET ${targetName} POST_BUILD
-                COMMAND ${script} $<TARGET_LINKER_FILE:${targetName}> 19
-                COMMENT "Checking number of exported symbols in the library")
-endif()


ktf:
@aphecetche something in the test triggers a command line tools reinstallation due to missing "nm".

aphecetche:
@ktf Hum, I doubt that script is the real origin of the issue. Has this been observed on your machine or on CI ones ? 

At some point a couple of weeks ago (last Xcode and/or OS upgrade, I don't remember clearly) I also got into trouble with some weird attempts to reinstall command line tools for missing tools like ar, ran, etc... I believe @dstocco also observed the same thing.  In the end "it solved itself" and I honestly did not look further (as I've already lost a couple of days of failing compiling at that point).

Anyway, even if we put a protection for missing nm (but a missing nm on dev machine is kind of strange), I would put it in the `check_nof_exported_symbols` directly instead of removing the test completely.

ktf:
I think @pzhristov had the same problem. I am sure it's probably some bug in XCode, but again, that's beyond our control.

What is not clear to me is what you are actually testing, and why only on mac.

aphecetche:
I'm just testing that I'm just exporting the least amount of symbols I've planned for for that library.
As to why only for Mac it's simply because I've never took the time to come up with the equivalent nm output "parsing" on other platforms; and one test on one platform was enough to catch instances where extra symbols were  inadvertently added (at the time this library was first developped quite a long time ago now)

dstocco:
Dear all,
I confirm that I had the issue with `nm` or `as`, or `ar`. The behaviour was rather random but basically, while compiling it complained that it could not find some command coming from command-line-tools and proposed to install them (except that they were already installed).
I've lost days. in the end I noticed that if I replaced `SYSTEM_VERSION_COMPAT: "1"` with `SYSTEM_VERSION_COMPAT: "0"` in `root.sh` it kind of worked...although I cannot compile AliRoot/AlliPhysics anymore. Frankly, I'm not even sure that this was really solving the issue or if, as Laurent says, at some points "it solved itself".

ktf:
Turns out the actual issue is due to:

```
export SYSTEM_VERSION_COMPAT=1
```

which confuses the script looking for uninstalled system tools.

PR #6253, opened 2021-05-26
Drop o2::byte

ktf:
Replace it with C++17 std::byte. This does not endorse the actual
idea behind std::byte usage, merely gets rid of an O2 specific type
and header for it.


ktf:
Even better then this would be to simply replace o2::byte with just char. When do we expect the extra type safety to be actually useful?

Barthelemy:
Concerning your second remark: what is the drawback of using std::byte ?

ktf:
@ironMann how would you like to proceed here? Shall we simply merge this and fix DataDistribution at the same time?

ktf:
@Barthelemy I missed you commented here. std::byte is a separate type from char, so there is casting involved everywhere this is not taken into account (e.g. when printing / formatting), which can catch some errors, but also might introduce a lot of boilerplate. As usual I guess it's a matter of tradeoffs.

ironMann:
@kft 
As far as I've seen, we mostly (only?) use byte*, and never use dereference of byte* directly. We still randomly cast byte* to different pointers and deref that (which is broken). From the standard, byte* has the same aliasing rules as char*, so there should be no surprises. 

Go ahead and merge if this is a simple rename.

ktf:
Ok, I did, could you tag DataDistribution with https://github.com/AliceO2Group/DataDistribution/pull/94 included?

PR #6254, opened 2021-05-27
Avoid tuple in analysis task

ktf:


ktf:
@aalkin I think this might reduce the amount of memory required to compile the analysis tasks. By applying a function directly on the decomposed structure, rather than going through a tuple, we should eliminate an optimisation step especially if we manage to have a trait which gets rid of the vector when the applied function returns void.

davidrohr (approved):
Reduces the memory slightly on my CC8 test, but the CC7 CI succeeded, so it seems to be helpful. And in any case, we can get rid of the tuple.

ktf:
Tested also in hyperloop and results seem compatible.

PR #6257, opened 2021-05-27
DPL: add --no-catch option

ktf:


davidrohr:
This is obsolete, since some of the code has changed anyway. Will open a new PR for this feature soon.

PR #6258, opened 2021-05-27
add tof file proxy for data w/o rdh

noferini:
add a file proxy reader to read TOF file acquired in slow control machines (no RDH) 
+ a fix for initOrbit when RDH is missing

### DIFF
--- a/Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
+++ b/Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
@@ -123,7 +123,9 @@ void CompressedDecodingTask::run(ProcessingContext& pc)
   //RS set the 1st orbit of the TF from the O2 header, relying on rdhHandler is not good (in fact, the RDH might be eliminated in the derived data)
   const auto* dh = o2::header::get<o2::header::DataHeader*>(pc.inputs().getByPos(0).header);
   mInitOrbit = dh->firstTForbit;
-  mDecoder.setFirstIR({0, mInitOrbit});
+  if (!mConetMode) {
+    mDecoder.setFirstIR({0, mInitOrbit});
+  }
 
   /** loop over inputs routes **/
   for (auto iit = pc.inputs().begin(), iend = pc.inputs().end(); iit != iend; ++iit) {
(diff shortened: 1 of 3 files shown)


preghenella (dismissed):

Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
```diff
@@ -162,6 +164,11 @@ void CompressedDecodingTask::endOfStream(EndOfStreamContext& ec)
 void CompressedDecodingTask::headerHandler(const CrateHeader_t* crateHeader, const CrateOrbit_t* crateOrbit)
 {
   if (mConetMode) {
+    if (mNCrateOpenTF == 0) {
```
shahor02, line 169: Is it safe to rely on the mNCrateOpenTF, and in general, on the rdhHandler? The delivery of every HBF is not guaranteed, so the 1st HBF RDH bearing the o2::trigger::TF might be missed so the mNCrateOpenTF will not be incremented. 
Also, here: https://github.com/AliceO2Group/AliceO2/blob/5b91ca11bd23384b0c742d6b2f8b3c5292270570/Detectors/TOF/workflow/src/CompressedDecodingTask.cxx#L343-L347
For the `mInitOrbit` you can switch to the `DataHeader::firstTForbit`

Detectors/TOF/workflow/src/file-proxy.cxx
```diff
@@ -0,0 +1,241 @@
...
+    return;
+  }
+
+  /** read data **/
+  int payload = 0;
+  if (mCONET)
+    payload = readCONET();
+  else
+    payload = readFLP();
+  if (payload == 0) {
```
shahor02: braces

shahor02 (commented):
Hi @noferini 
See a few comments below, not sure the current code is robust against dropped HBFs.

Detectors/TOF/workflow/src/file-proxy.cxx
noferini: done

Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
noferini: I kept mNCrateOpenTF counters but I removed the check on that for posting data (but the conet mode).
 mInitOrbit = RDHUtils::getHeartBeatOrbit(rdhr); removed as well

noferini:
o2 and macOS checks failing for reasons not related to this PR

preghenella (approved):

shahor02 (approved):

PR #6261, opened 2021-05-27
Inclusion of CosTheta variable in the helicity reference frame 

lucamicheletti93:
Inclusion of CosTheta variable in the helicity reference frame. The variable can be calculated for dimuons and electrons.

### DIFF
--- a/Analysis/PWGDQ/include/PWGDQCore/VarManager.h
+++ b/Analysis/PWGDQ/include/PWGDQCore/VarManager.h
@@ -646,8 +646,7 @@ void VarManager::FillPair(T const& t1, T const& t2, float* values, PairCandidate
   double cosTheta = 0;
   if (t1.sign() > 0) {
     cosTheta = zaxis.Dot(v1_CM);
-  }
-  else {
+  } else {
     cosTheta = zaxis.Dot(v2_CM);
   }
   values[kCosThetaHE] = cosTheta;
--- a/Analysis/PWGDQ/include/PWGDQCore/VarManager.h
+++ b/Analysis/PWGDQ/include/PWGDQCore/VarManager.h
@@ -20,6 +20,7 @@
 #include <TString.h>
 #include <Math/Vector4D.h>
 #include <TRandom.h>
+#include <TLorentzVector.h>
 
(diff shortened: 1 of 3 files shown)


Analysis/PWGDQ/include/PWGDQCore/VarManager.h
```diff
@@ -129,6 +130,7 @@ class VarManager : public TObject
     kMass,
     kCharge,
     kNBasicTrackVariables,
+    kCosThetaHE,
```
iarsene: Please add this in the group of pair variables, and it has to be before the "kNPairVariables"

Analysis/PWGDQ/include/PWGDQCore/VarManager.h
```diff
@@ -634,6 +636,29 @@ void VarManager::FillPair(T const& t1, T const& t2, float* values, PairCandidate
...
+  TLorentzVector pMuon1_CM = pMuon1_Lab;
+  TLorentzVector pMuon2_CM = pMuon2_Lab;
+  pMuon1_CM.Boost(beta);
+  pMuon2_CM.Boost(beta);
+  TVector3 zaxis;
+  zaxis = (pDimuon_Lab.Vect()).Unit();
+  double cosTheta;
+  if(t1.sign() > 0) cosTheta = zaxis.Dot((pMuon1_CM.Vect()).Unit());
+  else cosTheta = zaxis.Dot((pMuon2_CM.Vect()).Unit());
+  values[kCosThetaHE] = cosTheta;
```
iarsene: The 4-momentum of the two legs is actually already computed in this function using the Math::PtEtaPhiMVector. These are the v1 and v2 vectors just above in the function. So you should not redefine it. Those vectors have available the same methods as TLorentzVector and are more efficient. Please use those.

Try to use a different name than muon or electron if this is a general context. In this function, both dimuon and dielectron quantities are computed so maybe "lepton/dilepton" names can be used.

iarsene (dismissed):

iarsene (approved):

iarsene:
The o2 test fail is unrelated, merging

PR #6262, opened 2021-05-27
More fixes in doxygen:

ihrivnac:
Added doxygen keywords where missing, match the subpage references

### DIFF
--- a/Detectors/Align/Readme.md
+++ b/Detectors/Align/Readme.md
@@ -1,2 +1,6 @@
+<!-- doxy
+\page refDetectorsAlign Detectors Alignment
+/doxy -->
+
 #Align
 This is a top page for the detector alignment documentation.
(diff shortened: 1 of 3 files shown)


ihrivnac:
@vkucera ,
Is it expected that the space checker fails due to errors on the lines which are not modified in the PR? It's no problem to fix these, but I am just wondering if it is wanted behaviour. Thank you.

vkucera:
> @vkucera ,
> Is it expected that the space checker fails due to errors on the lines which are not modified in the PR? It's no problem to fix these, but I am just wondering if it is wanted behaviour. Thank you.

@ihrivnac , yes, it is expected. The test scans entire files. Actually it scans the file version of the merge commit (i.e. what it would look like after merging the PR), so it might even spot errors introduced in the `dev` after the PR branch diverged from it.
The main reason is that I didn't know how to isolate modified lines. :-) But eventually, I think this is the desired behaviour, as it cleans up the code progressively in a semi-adiabatic way. And since the instructions to fix the errors are simple and mentioned right in the log, I think it doesn't cause any inconvenience to the PR author. On top of that, the test runs already on push events in the fork, so the author can fix the errors before even creating a PR.

PR #6263, opened 2021-05-27
pythia8 generator flag will start empty pythia8 config, make sure you…

preghenella:
moving away from development mode where `o2-sim -g pythia8` automatically started event generation of inelastic pp collisions at 14 TeV.

`o2-sim -g pythia8` will from now on just start an empty pythia8 event generator setup.
if you will not provide the proper configuration, the initialisation of the event generator will fail with FATAL error.

Pythia8 configuration is passed via `--configKeyValues GeneratorPythia8.config=path/to/config/file`.

The old behaviour is kept for development purposes via `o2-sim -g pythia8pp`.


shahor02:
Thanks!

sawenzel:
This will likely have an effect on numerous tests done with this generator (sim_challenge.sh, alibi nightlies). Let's wait a bit until I have a proper list so that we can fix at the same time.

sawenzel (approved):

PR #6264, opened 2021-05-27
TRD fix: event time is relative to TF start

martenole:
Hi @shahor02 and @sawenzel 
I created a simulation with multiple TFs using the O2DPG tools. When running the TRD tracking for that simulation I noticed that matches are only found for the first TF. This was because the event time calculated from the TRD trigger records was based on the assumption that it is stored as an absolute time, but instead it is stored relative to the beginning of the TF.
I think other detectors use an absolute time. Is there a policy if one should use absolute or relative time?
Cheers,
Ole
ping @aberdnikova : this is the reason why you saw the TFs without any angular residuals

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/RecoInputContainer.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/RecoInputContainer.h
@@ -69,7 +69,11 @@ inline auto getRecoInputContainer(o2::framework::ProcessingContext& pc, o2::gpu:
   for (unsigned int iEv = 0; iEv < retVal->mNTriggerRecords; ++iEv) {
     const auto& trg = retVal->mTriggerRecords[iEv];
     retVal->trdTriggerIndices.push_back(trg.getFirstTracklet());
-    auto evTime = trg.getBCData().differenceInBC(inputTracks->startIR) * o2::constants::lhc::LHCBunchSpacingNS; // event time in ns
+    //auto evTime = trg.getBCData().differenceInBC(inputTracks->startIR) * o2::constants::lhc::LHCBunchSpacingNS; // event time in ns
+    // TODO at the moment the InteractionRecord contained in the TRD TriggerRecord stores the time relative to the beginning of the TF
+    // Other detectors are storing the absolute time, so that the event time needs to be shifted (as it is done in the commented line above)
+    // Should this be changed? Which format is expected from the real data?
+    auto evTime = trg.getBCData().toLong() * o2::constants::lhc::LHCBunchSpacingNS;                             // event time in ns
     retVal->trdTriggerTimes.push_back(evTime * 1e-3);                                                           // event time in us
   }
 


shahor02:
@martenole The clusters are stored in their natural units (ROF, TPC timebins, BC etc) referring either to abs bc,orbit since reset time or internal clock. Derived time stamps for tracks are stored wrt the TF start.

martenole:
Thanks!
@bazinski do you know if the trigger records which are created from the raw data stream will also store the event time relative to the TF or if it will be another unit?

bazinski:
@martenole relative to TF, because that is what sim does, but if we are going to change it now would be a good time.

martenole:
In general I think it is fine for the tracklets to have the time wrt TF start as it is now. We just need to be sure that in real data taking the trigger records are constructed in the same way as in the simulation. If I understand correctly this is the case so I don't see a problem.

shahor02:
> relative to TF, because that is what sim does, but if we are going to change it now would be a good time.

@bazinski  What do you mean? The TriggerRecord contains just `o2::InteractionRecord mBCData`, which counts from the orbit reset time. In the sim. digitization, the meaning of bc=0/orbit=0 has no meaning of the TF start, it is just some point in time when the orbit was reset. The start of TF is in fact the 1st orbit of the `HBFUtils::nHBFPerTF`-long interval containing the orbit `HBFUtils::orbitFirstSampled`

bazinski:
@shahor02 i went back and looked at the digitzation for trd and you are right its not stored relative to TF. @martenole i dont follow why you think its relative to timeframe or does it get changed somewhere after digitization https://github.com/AliceO2Group/AliceO2/blob/7a6c7f384bddc76e787b6efa32c5a0842c661cf1/Detectors/TRD/workflow/src/TRDDigitizerSpec.cxx#L97-L117 ? will fix raw data to be consistant.

martenole:
@bazinski this part of the TRD digitizer that you quote is done for each TF.  The collision context afaik contains the event times relative to the TF start so as a consequence also the time stored inside the TRD trigger records is relative to the TF start.

At the moment I can also clearly see that it is relative to the TF start, because with the changes from this PR the tracking also works for multiple TFs, while it does not work when the first orbit is subtracted for calculating the TRD trigger time.

bazinski:
@martenole cool that means raw is consistant. 
so we dont want to change it?

martenole:
For simulated raw data it should be consistent, yes. But in case of actual data taking I am not sure. The trigger records will be constructed on the EPNs, right? I assume that is the case since one needs to know the number of tracklets/digits for a given trigger and this is only known when the data from different FLPs is merged. The question is what information we get from the CRUs about the trigger time. Will that also be relative to the beginning of a TF?

shahor02:
Hi,
To be clear: there is no really TF defined on the digitization level, the sampling starts at `HBFUtils::orbitFirstSampled`, and the vector of InteractionTimeRecords generated starting from this point is what goes to the collision context. The collision context may span multiple TFs (whose boundaries are not defined during digitization), and its orbit/bc is non-decreasing.

During mc->raw conversion, the chopping to TFs starts from the `HBFUtils::orbitFirst` (where the SOX is written) and empty TFs of `HBFUtils::nHBFPerTF` orbits will be written (if any) until the 1st real TF containing the `HBFUtils::orbitFirstSampled` starts. 
In our usual simulations the `orbitFirst` and `orbitFirstSampled` are both 0, so the 1st TF seen will correspond to `orbit/bc={0,0}`, the next one to `{nHBFPerTF,0}` etc.

bazinski:
@martenole the flp builds what amounts to a trigger record in its raw headers (compressor version) and this then all gets merged on the epn for a complete timeframe and a TriggerRecord entry which is then passed along to the next device. @shahor02 @martenole I will review exactly the timing stuff a bit later, off the top of my head I am not going to guarantee it's consistent.

@martenole as for the info from the CRU, we get the bunch crossing in the cru header, we have the info from the RDH so bc and orbit, and we have the clock of the mcm in the half chamber headers. I think that is all the timing info.  So one gets the bunch crossing of the trigger and the rest is used for various forms of error checking both bunch crossing contestant, mcm clock increasing and consistent for event, etc.

martenole:
I am a bit confused here. I would like to get this merged, because otherwise we cannot work with larger simulations of multiple TFs. On the other hand we might need to change the implementation of the trigger records, I am not sure. Or are we save, because simulations start with orbitFirst and orbitFirstSampled equal to 0 such that the implementation in the TRD digitizer is OK?

shahor02:
@martenole not sure I understand what is the problem. The TriggerRecord contains the trigger orbit/bc, you can always convert this to number of bc's since the start of TF:
````
auto bc = triggerRec.getBCData().differenceInBC( recoContainer.startIR );
````
which then can be converted to \mus or TPC timebins etc. 
Do I misunderstand something?

martenole:
Hi Ruben,
`differenceInBC( recoContainer.startIR )` is what I used before, but this works only for the first TF. For all the later TFs I get negative values for the TRD time. That is why this fix is needed.

shahor02:
If TriggerRecord stores the bc/orbit wrt the TF start, this better to fix now. All InteractionRecords should contain raw bc/orbit (counts since "orbit reset" time)

martenole:
The reason for the observed issue was that I merged the output of the simulation of multiple TFs manually into single root files. When I tried to reconstruct the merged root files the reconstruction found multiple entries in the trees and added an offset of `iTimeFrame * nHBFPerTF` to each TF (=tree entry). But during the digitization each TF starts at HB/Orbit 0/0, so when I subtracted iTimeFrame * nHBFPerTF from the event time I got negative values

PR #6266, opened 2021-05-28
Add flag to check if a track has TOF

njacazio:
In agreement with the changes in https://github.com/alisw/AliPhysics/pull/17789 added a flag to the data model to check if a track has TOF


njacazio:
@jgrosseo this should handle what we discussed yesterday

jgrosseo (approved):

PR #6268, opened 2021-05-28
DPL: rationalise parent-child communication

ktf:
* We create the pipes separately from the forking, so that
  we can easily close all the unrelated pipes in the child.
* We first do all the forking and then we do the writing
  asynchronously for all children. It's important not to
  queue any work on the loop before the forking, because,
  that will be inherited by the child until execvp is called.
  For the same reason we create the pollers only after all
  the children have been forked.
  We should probably have a separate process (from the driver)
  which actually does all the forking.
* By doing the writing to the children asynchronously, we should
  speedup startup after a certain amount of children, given the bottleneck
  seems to be the writing of the configuration to stdin.

### DIFF
--- a/Framework/Core/src/runDataProcessing.cxx
+++ b/Framework/Core/src/runDataProcessing.cxx
@@ -641,32 +641,160 @@ void ws_connect_callback(uv_stream_t* server, int status)
   }
 }
 
-/// This will start a new device by forking and executing a
-/// new child
-void spawnDevice(std::string const& forwardedStdin,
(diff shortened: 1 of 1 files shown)


davidrohr (approved):
Thx a lot, this seems to work very well. Tested in full system test, particular for the digitizer (startup time on EPN goes from 1 minute to ~2 seconds) and for the full dpl workflow pipeline (20 seconds --> 2 seconds). And also fixes the test case of my Jira ticket.

PR #6269, opened 2021-05-28
reduce downscale noise

mwinn2:
@aphecetche : I made these two simple small changes to avoid crashes in the precluster and too many noise digits. One has to get back to it later, but at least, it should run with the preclusterer for the moment.


sawenzel:
This PR needs a conflict resolution / update.

aphecetche:
I believe those changes are no longer needed with the new digitizer. To be cross-checked with @mwinn2. Closing for now.

PR #6271, opened 2021-05-28
PWGHF: Switch MC on by default

vkucera:
Temporarily enable MC by default to make it possible to run MC on AliHyperloop.
@ginnocen


ginnocen (approved):

pzhristov:
Something is wrong with the checks - I do not see any obvious error in the logs. The compilation seems to be killed because of excessive memory consumption.

PR #6272, opened 2021-05-30
RawFileWriter can handle empty HBF fillers of >8KB

shahor02:
@lietava with this the empty HBF callback may provide a filler of arbitrary length (multiple of 16B), it will be automatically chopped to multiple RDH pages.


shahor02:
macos failure is unrelated, merging

PR #6274, opened 2021-05-31
Avoid inclusion of Vc/Vc in RandomRing

sawenzel:
Inclusion of Vc/Vc is argued to cause
gcc10 to take lot's of memory leading to crashes
in the CI.

In this particular case, I think we can avoid
the inclusion of Vc/Vc in headers and leave
it it to the source files to do so.
(But this might not always be possible);


sawenzel:
Merging. Let's see if we really get rid of the compiler crashes.

PR #6276, opened 2021-05-31
DPL DDS: use id rather than name for task id

ktf:
This will add support for the timePipelined devices.


ktf:
@AndreyLebedev this should fix the issue with the duplicate task name you are seeing.

PR #6277, opened 2021-05-31
TOF workflow to process the DCS configuration

chiarazampolli:
To test:

**o2-dcs-config-proxy --subscribe-to type=sub,method=connect,address=tcp://127.0.0.1:44772,rateLogging=1,transport=zeromq --acknowlege-to "" --severity debug | o2-calibration-tof-dcs-config-processor-workflow | o2-calibration-ccdb-populator-workflow --ccdb-path="http://localhost:8080"**

@shahor02 : It works with the dcs-config-proxy. 

workflow

clang-format

clang-format

### DIFF
--- a/Detectors/TOF/calibration/src/TOFFEElightReader.cxx
+++ b/Detectors/TOF/calibration/src/TOFFEElightReader.cxx
@@ -22,10 +22,10 @@ void TOFFEElightReader::loadFEElightConfig(const char* fileName)
   char* expandedFileName = gSystem->ExpandPathName(fileName);
   std::ifstream is;
   is.open(expandedFileName, std::ios::binary);
-  char* tempbuf = new char[sizeof(o2::tof::TOFFEElightConfig)];
-  is.read(tempbuf, sizeof(o2::tof::TOFFEElightConfig));
+  mFileLoadBuff.reset(new char[sizeof(o2::tof::TOFFEElightConfig)]);
+  is.read(mFileLoadBuff.get(), sizeof(o2::tof::TOFFEElightConfig));
   is.close();
-  mFEElightConfig = reinterpret_cast<TOFFEElightConfig*>(tempbuf);
+  mFEElightConfig = reinterpret_cast<const TOFFEElightConfig*>(mFileLoadBuff.get());
 }
 
 //_______________________________________________________________
@@ -34,9 +34,9 @@ void TOFFEElightReader::loadFEElightConfig(gsl::span<const char> configBuf)
   // load FEElight config from buffer
 
   if (configBuf.size() != sizeof(o2::tof::TOFFEElightConfig)) {
-    LOG(FATAL) << "Incoming message with TOFFEE configuration does not match expected size: " << configBuf.size() << " received, " << sizeof(*mFEElightConfig) << " expected";
+    LOG(INFO) << "Incoming message with TOFFEE configuration does not match expected size: " << configBuf.size() << " received, " << sizeof(*mFEElightConfig) << " expected";
   }
-  mFEElightConfig = const_cast<TOFFEElightConfig*>(reinterpret_cast<const TOFFEElightConfig*>(configBuf.data()));
+  mFEElightConfig = reinterpret_cast<const TOFFEElightConfig*>(configBuf.data());
 }
 
 //_______________________________________________________________
@@ -58,7 +58,7 @@ int TOFFEElightReader::parseFEElightConfig(bool verbose)
   mFEElightInfo.mRunType = runType;
 
   int nEnabled = 0, index;
-  TOFFEEchannelConfig* channelConfig = nullptr;
+  const TOFFEEchannelConfig* channelConfig = nullptr;
   for (int crateId = 0; crateId < Geo::kNCrate; crateId++) {
     for (int trmId = 0; trmId < Geo::kNTRM; trmId++) {
       for (int chainId = 0; chainId < Geo::kNChain; chainId++) {
@@ -90,7 +90,7 @@ int TOFFEElightReader::parseFEElightConfig(bool verbose)
     }
   }
 
-  TOFFEEtriggerConfig* triggerConfig = nullptr;
+  const TOFFEEtriggerConfig* triggerConfig = nullptr;
   for (Int_t iddl = 0; iddl < TOFFEElightConfig::NTRIGGERMAPS; iddl++) {
     triggerConfig = mFEElightConfig->getTriggerConfig(iddl);
     if (verbose) {
--- a/Detectors/TOF/calibration/src/TOFFEElightReader.cxx
+++ b/Detectors/TOF/calibration/src/TOFFEElightReader.cxx
@@ -32,6 +32,9 @@ void TOFFEElightReader::loadFEElightConfig(gsl::span<const char> configBuf)
   // load FEElight config from buffer
 
   //mFEElightConfig = reinterpret_cast<TOFFEElightConfig*>(*configBuf.data());
+  if (!configBuf.size() == sizeof(mFEElightConfig)) {
+    LOG(FATAL) << "Incoming message with TOFFEE configuration does not match expected size!";
+  }
   memcpy(&mFEElightConfig, configBuf.data(), sizeof(mFEElightConfig));
 }
 
--- a/Detectors/TOF/calibration/src/TOFFEElightReader.cxx
+++ b/Detectors/TOF/calibration/src/TOFFEElightReader.cxx
@@ -36,7 +36,7 @@ void TOFFEElightReader::loadFEElightConfig(gsl::span<const char> configBuf)
   if (configBuf.size() != sizeof(o2::tof::TOFFEElightConfig)) {
     LOG(FATAL) << "Incoming message with TOFFEE configuration does not match expected size: " << configBuf.size() << " received, " << sizeof(*mFEElightConfig) << " expected";
   }
-  mFEElightConfig = reinterpret_cast<TOFFEElightConfig*>(*configBuf.data());
+  mFEElightConfig = const_cast<TOFFEElightConfig*>(reinterpret_cast<const TOFFEElightConfig*>(configBuf.data()));
 }
 
 //_______________________________________________________________
@@ -52,9 +52,6 @@ int TOFFEElightReader::parseFEElightConfig(bool verbose)
   int version = mFEElightConfig->mVersion;
   int runNumber = mFEElightConfig->mRunNumber;
   int runType = mFEElightConfig->mRunType;
-  LOG(INFO) << "version = " << version;
-  LOG(INFO) << "run number = " << runNumber;
-  LOG(INFO) << "run type = " << runType;
 
   mFEElightInfo.mVersion = version;
   mFEElightInfo.mRunNumber = runNumber;
--- a/Detectors/TOF/calibration/src/TOFFEElightReader.cxx
+++ b/Detectors/TOF/calibration/src/TOFFEElightReader.cxx
@@ -22,8 +22,10 @@ void TOFFEElightReader::loadFEElightConfig(const char* fileName)
   char* expandedFileName = gSystem->ExpandPathName(fileName);
   std::ifstream is;
   is.open(expandedFileName, std::ios::binary);
-  is.read(reinterpret_cast<char*>(&mFEElightConfig), sizeof(mFEElightConfig));
+  char* tempbuf = new char[sizeof(o2::tof::TOFFEElightConfig)];
+  is.read(tempbuf, sizeof(o2::tof::TOFFEElightConfig));
   is.close();
+  mFEElightConfig = reinterpret_cast<TOFFEElightConfig*>(tempbuf);
 }
 
 //_______________________________________________________________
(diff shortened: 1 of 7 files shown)


chiarazampolli:
Dear @noferini , @preghenella , 

This PR is to process the TOF FEE from DCS. 

Chiara

Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h
```diff
@@ -0,0 +1,123 @@
...
+  {
+    auto configBuff = pc.inputs().get<gsl::span<char>>("inputConfig");
+    auto configFileName = pc.inputs().get<std::string>("inputConfigFileName");
+    auto timer = std::chrono::duration_cast<std::chrono::milliseconds>(HighResClock::now().time_since_epoch()).count();
+    ;
+    //std::string configFileName = "pippo";
+    LOG(INFO) << "got input file " << configFileName << " of size " << configBuff.size();
+    mTOFFEElightReader.loadFEElightConfig(configBuff);
+    mTOFFEElightReader.parseFEElightConfig(mVerbose);
+    //auto tfcounter = o2::header::get<o2::framework::DataProcessingHeader*>(pc.inputs().get("input").header)->startTime;
```
chiarazampolli: Hello @shahor02 ,
Do you know why this commented line does not work? Have we forgotten anything from the dcs-proxy?

Detectors/TOF/calibration/src/TOFFEElightReader.cxx
```diff
@@ -27,8 +27,17 @@ void TOFFEElightReader::loadFEElightConfig(const char* fileName)
 }
 
 //_______________________________________________________________
+void TOFFEElightReader::loadFEElightConfig(gsl::span<const char> configBuf)
+{
+  // load FEElight config from buffer
+
+  //mFEElightConfig = reinterpret_cast<TOFFEElightConfig*>(*configBuf.data());
+  memcpy(&mFEElightConfig, configBuf.data(), sizeof(mFEElightConfig));
```
chiarazampolli: Not sure memcpy is the best: any suggestion?

shahor02: yes it you want to have a copy (I guess you don't need it when receiving data from the DPL but do need when load from the file?). But I would anyway check that the `configBuf.size()==sizeof(mFEElightConfig)`.

Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  //---------------------------------------------------------
+
+  void run(o2::framework::ProcessingContext& pc) final
+  {
+    auto configBuff = pc.inputs().get<gsl::span<char>>("inputConfig");
+    auto configFileName = pc.inputs().get<std::string>("inputConfigFileName");
+    auto timer = std::chrono::duration_cast<std::chrono::milliseconds>(HighResClock::now().time_since_epoch()).count();
+    ;
```
shahor02: there is a stray `;`

Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h
shahor02: You don't have `"input"` input, the getTOFDCSConfigProcessorSpec defines only `"inputConfig"` and `"inputConfigFileName"`

You can use also 
````
o2::header::get<o2::framework::DataProcessingHeader*>( pc.inputs().getByPos(0).header )->startTime;
````
to be independent on particular names.

Comment (chiarazampolli) [reply Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h]: Correct, thanks! After it failed, I just started thinking that maybe one should use the current time, what do you think?

File: Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h (hunk 3 above)
Comment (chiarazampolli) [reply Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h]: Thanks

File: Detectors/TOF/calibration/src/TOFFEElightReader.cxx (hunk 2 above)
Comment (chiarazampolli) [reply Detectors/TOF/calibration/src/TOFFEElightReader.cxx]: I added the check. thanks.
I don't really want to have a copy, which is why I was not sure about this approach. I would have used a reinterpret_cast, but I did not manage to make it work correctly. 
What I want is to put in the `mFEElightConfig` variable the content of the buffer. I think that with reinterpret_cast it should work as commented out if `mFEElightConfig` was a pointer.

File: Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h (hunk 1 above)
Comment (shahor02) [reply Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h]: What should be the use of this startTime? The timestamp for CCDB?

File: Detectors/TOF/calibration/src/TOFFEElightReader.cxx (hunk 2 above)
Comment (shahor02) [reply Detectors/TOF/calibration/src/TOFFEElightReader.cxx]: the reinterpret_cast should work, but you don't need to dereference the `configBuf.data()`, i.e. should be
````
mFEElightConfig = reinterpret_cast<TOFFEElightConfig*>(configBuf.data());
````

But you also have an option of getting the `mFEElightConfig` from the file. If you want to avoid a copy of the DPL provided object, you should have an additional internal buffer where you will read in the data from the file.

**Review by noferini — DISMISSED** [review]

**Comment by shahor02** [comment]
Hi @chiarazampolli 
Could you fix the 
````
/System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SOURCES/O2/6277/0/Detectors/TOF/calibration/src/TOFFEElightReader.cxx:35:7: error: logical not is only applied to the left hand side of this comparison [-Werror,-Wlogical-not-parentheses]
  if (!configBuf.size() == sizeof(mFEElightConfig)) {
````
spotter by the CI.
Otherwise, should this be merged as it is or you want to address the issue of avoiding an extra copy of the payload?

File: Detectors/TOF/calibration/src/TOFFEElightReader.cxx (hunk 2 above)
Comment (chiarazampolli) [reply Detectors/TOF/calibration/src/TOFFEElightReader.cxx]: I will change `mFEElightConfig` to be a pointer, then I will use `reinterpret_cast` as in the commented line (see my comment above). This way, when going through DPL there will be no copy.

**Comment by chiarazampolli** [comment]
> Hi @chiarazampolli
> Could you fix the
> 
> ```
> /System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SOURCES/O2/6277/0/Detectors/TOF/calibration/src/TOFFEElightReader.cxx:35:7: error: logical not is only applied to the left hand side of this comparison [-Werror,-Wlogical-not-parentheses]
>   if (!configBuf.size() == sizeof(mFEElightConfig)) {
> ```
> 
> spotter by the CI.
> Otherwise, should this be merged as it is or you want to address the issue of avoiding an extra copy of the payload?

Hi Ruben, 
I will change the memcpy. I prefer not to copy.
Chiara

**Comment by chiarazampolli** [comment]
I moved the PR to WIP since I have some issues to solve. 

Chiara

**Comment by chiarazampolli** [comment]
Hi @shahor02 ,

Can you check if it is fine for you like this? Note that I needed to introduce also the `const_cast`.

I still have the doubt about the start time for the uploaded CCDB object, see previous comments. 

Chiara

File: Detectors/TOF/calibration/src/TOFFEElightReader.cxx [hunk 4]
```diff
@@ -22,20 +22,21 @@ void TOFFEElightReader::loadFEElightConfig(const char* fileName)
   char* expandedFileName = gSystem->ExpandPathName(fileName);
   std::ifstream is;
   is.open(expandedFileName, std::ios::binary);
-  is.read(reinterpret_cast<char*>(&mFEElightConfig), sizeof(mFEElightConfig));
+  char* tempbuf = new char[sizeof(o2::tof::TOFFEElightConfig)];
+  is.read(tempbuf, sizeof(o2::tof::TOFFEElightConfig));
   is.close();
+  mFEElightConfig = reinterpret_cast<TOFFEElightConfig*>(tempbuf);
```
shahor02: the tempbuf should be a class data member, which is reset on reading a new file and cleaned on destructing the object, e.g.
````
std::unique_ptr<char[]> mFileLoadBuff; // in the class declaration
...
mFileLoadBuff.reset(  new char[sizeof(o2::tof::TOFFEElightConfig)] );
is.read(mFileLoadBuff.get(), sizeof(o2::tof::TOFFEElightConfig));
is.close();
mFEElightConfig = reinterpret_cast<TOFFEElightConfig*>(mFileLoadBuff.get());
````

File: Detectors/TOF/calibration/src/TOFFEElightReader.cxx [hunk 5]
```diff
@@ -22,30 +22,53 @@ void TOFFEElightReader::loadFEElightConfig(const char* fileName)
...
+
+//_______________________________________________________________
+void TOFFEElightReader::loadFEElightConfig(gsl::span<const char> configBuf)
+{
+  // load FEElight config from buffer
+
+  if (configBuf.size() != sizeof(o2::tof::TOFFEElightConfig)) {
+    LOG(FATAL) << "Incoming message with TOFFEE configuration does not match expected size: " << configBuf.size() << " received, " << sizeof(*mFEElightConfig) << " expected";
+  }
+  mFEElightConfig = const_cast<TOFFEElightConfig*>(reinterpret_cast<const TOFFEElightConfig*>(configBuf.data()));
```
shahor02: you don't need const_cast, instead, better define `const TOFFEElightConfig* mFEElightConfig = nullptr`

Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  //---------------------------------------------------------
+
+  void endOfStream(o2::framework::EndOfStreamContext& ec) final
+  {
+  }
+
+ private:
+  void sendOutput(DataAllocator& output, TFType tf)
```
shahor02: did not understand which previous comment you were referring concerning the start time, but I don't see a problem with using now().  In opposite, I would not use `TFType` for something which is a real timestamp: the TFType as TF counter may change in future. Better use `long`

Detectors/TOF/calibration/src/TOFFEElightReader.cxx
chiarazampolli: I also thought about this possibility, but it was simpler to add the `const_cast` :-) 
Changed.

Detectors/TOF/calibration/testWorkflow/TOFDCSConfigProcessorSpec.h
chiarazampolli: Great, I agree, but wanted to be sure. 
`long` is now used.

chiarazampolli:
Hi @shahor02 ,

For the comment 

> the reinterpret_cast should work, but you don't need to dereference the configBuf.data()

on which I cannot reply since it is outdated: yes, no dereferencing, of course. 

And for:

> What should be the use of this startTime? The timestamp for CCDB?

yes, the timestamp is for CCDB.

Chiara

Detectors/TOF/calibration/src/TOFFEElightReader.cxx
chiarazampolli: Done. 
Would it there be a leak in the previous version?  
And, for my knowledge, you suggested to use a `unique_ptr` because in this way I don't have to manually delete anything, right when a new file comes, right? The default destructor would take care of deleting a standard pointer, if I am not wrong.

Detectors/TOF/calibration/src/TOFFEElightReader.cxx
```diff
@@ -22,30 +22,53 @@ void TOFFEElightReader::loadFEElightConfig(const char* fileName)
...
+  mFEElightConfig = reinterpret_cast<const TOFFEElightConfig*>(mFileLoadBuff.get());
+}
+
+//_______________________________________________________________
+void TOFFEElightReader::loadFEElightConfig(gsl::span<const char> configBuf)
+{
+  // load FEElight config from buffer
+
+  if (configBuf.size() != sizeof(o2::tof::TOFFEElightConfig)) {
+    LOG(INFO) << "Incoming message with TOFFEE configuration does not match expected size: " << configBuf.size() << " received, " << sizeof(*mFEElightConfig) << " expected";
```
chiarazampolli: Here the check triggers just a `LOG(INFO)` since the number of TRMs defined in TOF::Geo seems to be wrong, and this check currently fails. I am confirming with @noferini , then I will change it accordingly. 
@shahor02 : even if happy with everything else, do not approve till this is solved (otherwise we would have immediately a new PR). Thanks!

Detectors/TOF/calibration/src/TOFFEElightReader.cxx
shahor02: Yes, the previous version would lead to a leak on every reading and also because the destructor does not clean the the memory pointed by the pointer data member, it just destroys the pointer itself (which in your case was a local variable, anyway)

chiarazampolli:
Hello @noferini , 

Can you check the latest commit? This should fix the indexing of the TRM in the electronics compared to O2. The size of the file is now as expected. 

Chiara

chiarazampolli:
The error of the CI seems not related to this PR. @shahor02 , what do you think?
@noferini , can you check that the coding for the TRM numbering is fine for you and if so approve?

Thanks!

shahor02:
Yes, the failure is unrelated, squashing/merging

PR #6278, opened 2021-05-31
ctpdev: Digits to Raw data dev

lietava:
@shahor02 : please, have a look.

### DIFF
--- a/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
+++ b/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
@@ -48,9 +48,8 @@ struct CTPInput {
 struct CTPDescriptor {
   CTPDescriptor() = default;
   std::string name;
-  std::uint64_t inputsMask;
-  std::vector<std::string> inputNames;
-  std::vector<CTPInput> inputs;
+  std::vector<CTPInput*> inputs;
+  std::uint64_t getInputsMask() const;
   void printStream(std::ostream& strem) const;
   ClassDefNV(CTPDescriptor, 2)
 };
@@ -93,8 +92,11 @@ class CTPConfiguration
   void addCTPClass(const CTPClass& ctpclass);
   void printStream(std::ostream& stream) const;
   std::vector<CTPInput>& getCTPInputs() { return mInputs; }
+  std::vector<CTPClass>& getCTPClasses() { return mCTPClasses; }
   uint64_t getInputMask(const std::string& name);
-  bool isMaskInInputs(uint64_t& mask);
+  bool isMaskInInputs(const uint64_t& mask) const;
+  CTPInput* isInputInConfig(const std::string inpname);
+  uint64_t getDecrtiptorInputsMask(const std::string& name) const;
 
  private:
   std::string mName;
--- a/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
+++ b/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
@@ -41,7 +41,7 @@ struct CTPInput {
   std::string level;
   std::uint64_t inputMask;
   o2::detectors::DetID detector;
-  std::string getInputDetName() const {return detector.getName();}
+  std::string getInputDetName() const { return detector.getName(); }
   void printStream(std::ostream& strem) const;
   ClassDefNV(CTPInput, 2);
 };
@@ -91,9 +91,10 @@ class CTPConfiguration
   void addCTPCluster(const CTPCluster& cluster);
   void addCTPClass(const CTPClass& ctpclass);
   void printStream(std::ostream& stream) const;
-  std::vector<CTPInput>& getCTPInputs(){return mInputs;}
+  std::vector<CTPInput>& getCTPInputs() { return mInputs; }
   uint64_t getInputMask(const std::string& name);
   bool isMaskInInputs(uint64_t& mask);
+
  private:
   std::string mName;
   std::string mVersion;
--- a/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
+++ b/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
@@ -23,6 +23,9 @@ namespace o2
 {
 namespace ctp
 {
+/// Database constants
+const std::string CCDBPathCTPConfig = "CTP/Config";
+/// CTP Config items
 struct BCMask {
   BCMask() = default;
   std::string name;
@@ -33,18 +36,21 @@ struct BCMask {
 struct CTPInput {
   CTPInput() = default;
   std::string name;
-  o2::detectors::DetID::ID detID;
+  std::string level;
+  std::string detName;
   std::uint64_t inputMask;
+  o2::detectors::DetID::ID detID;
   void printStream(std::ostream& strem) const;
-  ClassDefNV(CTPInput, 1);
+  ClassDefNV(CTPInput, 2);
 };
 struct CTPDescriptor {
   CTPDescriptor() = default;
   std::string name;
   std::uint64_t inputsMask;
-  std::vector<std::string> inputsNames;
+  std::vector<std::string> inputNames;
+  std::vector<CTPInput> inputs;
   void printStream(std::ostream& strem) const;
-  ClassDefNV(CTPDescriptor, 1)
+  ClassDefNV(CTPDescriptor, 2)
 };
 /// The main part is Local Trigger Generator (LTG)
 struct CTPDetector {
@@ -56,17 +62,17 @@ struct CTPDetector {
 struct CTPCluster {
   CTPCluster() = default;
   std::string name;
-  uint32_t detectorsMask;
-  std::vector<std::string> detectorsNames;
+  o2::detectors::DetID::mask_t detectorsMask;
+  std::vector<o2::detectors::DetID> detectors;
   void printStream(std::ostream& strem) const;
-  ClassDefNV(CTPCluster, 1)
+  ClassDefNV(CTPCluster, 2)
 };
 struct CTPClass {
   CTPClass() = default;
   std::string name;
   std::uint64_t classMask;
-  CTPDescriptor* descriptor;
-  CTPCluster* cluster;
+  CTPDescriptor const* descriptor;
+  CTPCluster const* cluster;
   void printStream(std::ostream& strem) const;
   ClassDefNV(CTPClass, 1);
 };
@@ -74,6 +80,7 @@ class CTPConfiguration
 {
  public:
   CTPConfiguration() = default;
+  int loadConfiguration(std::string& ctpconfiguartion);
   void addBCMask(const BCMask& bcmask);
   void addCTPInput(const CTPInput& input);
   void addCTPDescriptor(const CTPDescriptor& descriptor);
@@ -81,16 +88,20 @@ class CTPConfiguration
   void addCTPCluster(const CTPCluster& cluster);
   void addCTPClass(const CTPClass& ctpclass);
   void printStream(std::ostream& stream) const;
-
+  std::vector<CTPInput>& getCTPInputs(){return mInputs;}
+  uint64_t getInputMask(const std::string& name);
+  bool isMaskInInputs(uint64_t& mask);
  private:
   std::string mName;
+  std::string mVersion;
   std::vector<BCMask> mBCMasks;
   std::vector<CTPInput> mInputs;
   std::vector<CTPDescriptor> mDescriptors;
   std::vector<CTPDetector> mDetectors;
   std::vector<CTPCluster> mClusters;
   std::vector<CTPClass> mCTPClasses;
-  ClassDefNV(CTPConfiguration, 1);
+  int processConfigurationLine(std::string& line, int& level);
+  ClassDefNV(CTPConfiguration, 2);
 };
 } // namespace ctp
 } // namespace o2
--- a/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
+++ b/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
@@ -58,7 +58,7 @@ struct CTPDescriptor {
 struct CTPDetector {
   CTPDetector() = default;
   o2::detectors::DetID::ID detID;
-  const char* getName()const { return o2::detectors::DetID::getName(detID); }
+  const char* getName() const { return o2::detectors::DetID::getName(detID); }
   uint32_t HBaccepted; /// Number of HB frames in TF to be accepted
   void printStream(std::ostream& strem) const;
 };
@@ -66,7 +66,7 @@ struct CTPCluster {
   CTPCluster() = default;
   std::string name;
   o2::detectors::DetID::mask_t maskCluster;
-  std::string getClusterDetNames() const { return o2::detectors::DetID::getNames(maskCluster,' '); }
+  std::string getClusterDetNames() const { return o2::detectors::DetID::getNames(maskCluster, ' '); }
   void printStream(std::ostream& strem) const;
   ClassDefNV(CTPCluster, 2)
 };
--- a/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
+++ b/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
@@ -58,14 +58,15 @@ struct CTPDescriptor {
 struct CTPDetector {
   CTPDetector() = default;
   o2::detectors::DetID::ID detID;
+  const char* getName()const { return o2::detectors::DetID::getName(detID); }
   uint32_t HBaccepted; /// Number of HB frames in TF to be accepted
   void printStream(std::ostream& strem) const;
 };
 struct CTPCluster {
   CTPCluster() = default;
   std::string name;
-  o2::detectors::DetID::mask_t getClusterDetMask() const;
-  std::vector<o2::detectors::DetID> detectors;
+  o2::detectors::DetID::mask_t maskCluster;
+  std::string getClusterDetNames() const { return o2::detectors::DetID::getNames(maskCluster,' '); }
   void printStream(std::ostream& strem) const;
   ClassDefNV(CTPCluster, 2)
 };
--- a/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
+++ b/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
@@ -25,6 +25,8 @@ namespace ctp
 {
 /// Database constants
 const std::string CCDBPathCTPConfig = "CTP/Config";
+///
+bool isDetector(o2::detectors::DetID& det);
 /// CTP Config items
 struct BCMask {
   BCMask() = default;
@@ -37,9 +39,9 @@ struct CTPInput {
   CTPInput() = default;
   std::string name;
   std::string level;
-  std::string detName;
   std::uint64_t inputMask;
-  o2::detectors::DetID::ID detID;
+  o2::detectors::DetID detector;
+  std::string getInputDetName() const {return detector.getName();}
   void printStream(std::ostream& strem) const;
   ClassDefNV(CTPInput, 2);
 };
@@ -62,7 +64,7 @@ struct CTPDetector {
 struct CTPCluster {
   CTPCluster() = default;
   std::string name;
-  o2::detectors::DetID::mask_t detectorsMask;
+  o2::detectors::DetID::mask_t getClusterDetMask() const;
   std::vector<o2::detectors::DetID> detectors;
   void printStream(std::ostream& strem) const;
   ClassDefNV(CTPCluster, 2)
@@ -80,7 +82,8 @@ class CTPConfiguration
 {
  public:
   CTPConfiguration() = default;
-  int loadConfiguration(std::string& ctpconfiguartion);
+  bool isDetector(const o2::detectors::DetID& det);
+  int loadConfiguration(const std::string& ctpconfiguartion);
   void addBCMask(const BCMask& bcmask);
   void addCTPInput(const CTPInput& input);
   void addCTPDescriptor(const CTPDescriptor& descriptor);
--- a/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
+++ b/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
@@ -19,6 +19,7 @@
 #include <string>
 #include <vector>
 #include <bitset>
+#include <map>
 namespace o2
 {
 namespace ctp
@@ -40,8 +41,8 @@ struct CTPInput {
   std::string name;
   std::string level;
   std::uint64_t inputMask;
-  o2::detectors::DetID detector;
-  std::string getInputDetName() const { return detector.getName(); }
+  o2::detectors::DetID::ID detID;
+  std::string getInputDetName() const { return o2::detectors::DetID::getName(detID); }
   void printStream(std::ostream& strem) const;
   ClassDefNV(CTPInput, 2);
 };
@@ -97,7 +98,7 @@ class CTPConfiguration
   bool isMaskInInputs(const uint64_t& mask) const;
   CTPInput* isInputInConfig(const std::string inpname);
   uint64_t getDecrtiptorInputsMask(const std::string& name) const;
-
+  std::map<o2::detectors::DetID::ID , std::vector<CTPInput> > getDet2InputMap();
  private:
   std::string mName;
   std::string mVersion;
--- a/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
+++ b/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
@@ -98,7 +98,8 @@ class CTPConfiguration
   bool isMaskInInputs(const uint64_t& mask) const;
   CTPInput* isInputInConfig(const std::string inpname);
   uint64_t getDecrtiptorInputsMask(const std::string& name) const;
-  std::map<o2::detectors::DetID::ID , std::vector<CTPInput> > getDet2InputMap();
+  std::map<o2::detectors::DetID::ID, std::vector<CTPInput>> getDet2InputMap();
+
  private:
   std::string mName;
   std::string mVersion;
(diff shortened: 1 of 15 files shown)


lietava:
oops, forgot cling. Doing now.

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
```diff
@@ -33,18 +36,21 @@ struct BCMask {
 struct CTPInput {
   CTPInput() = default;
   std::string name;
-  o2::detectors::DetID::ID detID;
+  std::string level;
+  std::string detName;
```
shahor02: note that the DetID has a method getName, i.e. instead of storing the name you can simply call
`DetID::getName(detID)``.
Alternatively, you can use DetID directly as a data member, instead of DetID::ID.

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
```diff
@@ -56,41 +62,46 @@ struct CTPDetector {
 struct CTPCluster {
   CTPCluster() = default;
   std::string name;
-  uint32_t detectorsMask;
-  std::vector<std::string> detectorsNames;
+  o2::detectors::DetID::mask_t detectorsMask;
+  std::vector<o2::detectors::DetID> detectors;
```
shahor02: why do you need both detectorsMask and `std::vector<o2::detectors::DetID> detectors` ? Don't they encode the same detectors list?

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
```diff
@@ -56,41 +62,46 @@ struct CTPDetector {
...
+  CTPDescriptor const* descriptor;
+  CTPCluster const* cluster;
   void printStream(std::ostream& strem) const;
   ClassDefNV(CTPClass, 1);
 };
 class CTPConfiguration
 {
  public:
   CTPConfiguration() = default;
+  int loadConfiguration(std::string& ctpconfiguartion);
```
shahor02: `const std::string& ctpconfiguartion`

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
```diff
@@ -33,18 +36,21 @@ struct BCMask {
 struct CTPInput {
   CTPInput() = default;
   std::string name;
-  o2::detectors::DetID::ID detID;
+  std::string level;
+  std::string detName;
   std::uint64_t inputMask;
```
shahor02, line 43: is this detectors input mask? If yes, you could probably use DetID::mask_t

DataFormats/Detectors/CTP/include/DataFormatsCTP/Digits.h
```diff
@@ -29,17 +29,21 @@ static constexpr std::uint32_t NumOfHBInTF = 256;
...
+/// CTP hits are inputs. Digits are just inputs collected from all detectors.
+/// digits->raw: CTP Config to be used
+/// raw->digits: CTP config to be used
 static constexpr std::pair<uint32_t, std::bitset<CTP_MAXL0PERDET>> CTP_INPUTMASK_FV0(0, 0x1f);
 static constexpr std::pair<uint32_t, std::bitset<CTP_MAXL0PERDET>> CTP_INPUTMASK_FT0(5, 0x1f);
 ///
 struct CTPDigit {
   o2::InteractionRecord intRecord;
   std::bitset<CTP_NINPUTS> CTPInputMask;
-  std::bitset<CTP_NCLASSES> CTPClassMask;
```
shahor02: for the record: the ROOT schema evolution does not support change of N in the `bitset<N>`.
Are these `CTP_NINPUTS` and `CTP_NCLASSES` constants which never change?

DataFormats/Detectors/CTP/src/Configuration.cxx
```diff
@@ -23,11 +25,13 @@ void BCMask::printStream(std::ostream& stream) const
 }
 void CTPInput::printStream(std::ostream& stream) const
 {
-  stream << "CTP Input:" << name << "Detector:" << detID << " Hardware mask:" << inputMask << std::endl;
+  stream << "CTP Input:" << name << " Detector:" << detName << " Level:" << level << " Hardware mask:0x" << std::hex << inputMask << std::dec << std::endl;
```
shahor02: see my comment about the DetID::getName

DataFormats/Detectors/CTP/src/Configuration.cxx
```diff
@@ -36,13 +40,186 @@ void CTPDetector::printStream(std::ostream& stream) const
...
 void CTPClass::printStream(std::ostream& stream) const
 {
-  stream << "CTP Class:" << name << " Hardware mask:" << classMask << std::endl;
+  stream << "CTP Class:" << name << " Hardware mask:" << classMask << " Descriptor:" << descriptor->name << " Cluster:"<< cluster->name << std::endl;
 }
 /// CTP configuration
+/// Assuming Run2 format + LTG
+int CTPConfiguration::loadConfiguration(std::string& ctpconfiguration)
+{
+  std::cout << "Loading CTP configuration." << std::endl;
```
shahor02: don't use `std::cout`, use `LOG(INFO)` instead.

DataFormats/Detectors/CTP/src/Configuration.cxx
```diff
@@ -36,13 +40,186 @@ void CTPDetector::printStream(std::ostream& stream) const
...
+  {
+    o2::utils::Str::trim(line);
+    if( (ret = processConfigurationLine(line,level)) !=0 )
+    {
+        return ret;
+    }
+  }
+  return ret;    
+}
+int CTPConfiguration::processConfigurationLine(std::string& line, int& level)
```
shahor02, line 92: `const std::string& line`

DataFormats/Detectors/CTP/src/Configuration.cxx
```diff
@@ -36,13 +40,186 @@ void CTPDetector::printStream(std::ostream& stream) const
...
+  }
+  if((first=line.find("CLASSES:")) != std::string::npos)
+  {
+    level = 7;
+    return 0;
+  }
+  /// Do parse levels
+  std::vector<std::string> tokens = o2::utils::Str::tokenize(line,' ');
+  size_t ntokens=tokens.size();
+  if(ntokens == 0) return 0;
```
shahor02: add `{}`

DataFormats/Detectors/CTP/src/Configuration.cxx
```diff
@@ -36,13 +40,186 @@ void CTPDetector::printStream(std::ostream& stream) const
...
+        }
+        CTPInput inp;
+        inp.name = tokens[0];
+        inp.detName = tokens[1];
+        inp.level = tokens[2];
+        o2::detectors::DetID det(inp.detName.c_str());
+        //std::cout << "id:" << det.getID() << " ndets:" << det.getNDetectors() << std::endl;
+        if(det.getID() >= det.getNDetectors())
+        {
+            std::cout << "INPUT line:" << line << " Detector does not exist: " << inp.detName << std::endl;
```
shahor02: better throw exception: `throw std::runtime_error(fmt::format("INPUT line {}: Detector {} does not exist", line, inp.detName);`

DataFormats/Detectors/CTP/src/Configuration.cxx
```diff
@@ -36,13 +40,186 @@ void CTPDetector::printStream(std::ostream& stream) const
...
+            exit(1);
+        }
+        try
+        {
+            inp.inputMask = std::stoull(tokens[3],nullptr,0);
+        }
+        catch(...)
+        {
+            std::cout << "Syntax error in INPUTS in mask:" << line << std::endl;
```
shahor02: `LOG(ERROR)`

DataFormats/Detectors/CTP/src/Configuration.cxx
```diff
@@ -36,13 +40,186 @@ void CTPDetector::printStream(std::ostream& stream) const
...
+        mClusters.push_back(cluster);
+        break;
+    }
+    case 7:
+    /// CLASSES: name mask descriptor cluster LM L0 L1
+    {
+        CTPClass cls;
+        if(ntokens != 4)
+        {
+            std::cout << "Syntax error in CLASSES:" << line << std::endl;
```
shahor02: same here, also below

Detectors/CTP/simulation/include/CTPSimulation/Digits2Raw.h
```diff
@@ -0,0 +1,78 @@
...
+  const int mNLinks=2;
+  const uint32_t CRULinkIDIntRec = 0;
+  const uint32_t NIntRecPayload = 48+12;
+  bool mZeroSuppressedIntRec = false;
+  const uint32_t CRULinkIDClassRec = 1;
+  const uint32_t NClassPayload = 64+12;
+  bool mZeroSuppressedClassRec = true;
+  //constexpr uint32_t CTPCRULinkIDMisc = 2;
+  const std::string HBFEmpty = "EMPTY_HBF       ";
+  ClassDefNV(Digits2Raw, 1);
```
shahor02: this will be a simple executable, why do you need root dictionary?

Detectors/CTP/simulation/src/Digits2Raw.cxx
```diff
@@ -0,0 +1,199 @@
...
+  if (outd.back() != '/') {
+    outd += '/';
+  }  
+  mLinkID = uint32_t(0);
+  mCruID = uint16_t(0);
+  mEndPointID = uint32_t(0);
+  for (int ilink = 0; ilink < mNLinks; ilink++) {
+    mFeeID = uint64_t(ilink);
+    std::string outFileLink = mOutputPerLink ? o2::utils::Str::concat_string(outd, "ctp_link", std::to_string(ilink), ".raw") : o2::utils::Str::concat_string(outd, "ctp.raw");
+    mWriter.registerLink(mFeeID, mCruID, mLinkID, mEndPointID, outFileLink);
```
shahor02: You should not have the same `mLinkID==0` on the same CRU. What is the point of this mLinkID at all if you know that you have 2 links?

Detectors/CTP/simulation/src/Digits2Raw.cxx
```diff
@@ -0,0 +1,199 @@
...
+      std::vector<std::bitset<NGBT>> digits;
+      for(uint32_t i = 0; i < o2::constants::lhc::LHCMaxBunches; i++) {
+          std::bitset<NGBT> dig = i;
+          digits.push_back(dig);
+      }
+      toAdd = digits2HBTPayload(digits,NIntRecPayload);
+    }
+  }
+  toAdd.resize(o2::raw::RDHUtils::GBTWord);
+  std::memcpy(toAdd.data(), HBFEmpty.c_str(), o2::raw::RDHUtils::GBTWord);
```
shahor02: this seems to be a copy/paste error from the example, remove it.

shahor02 (changes requested):

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
lietava: No, inputMask is position of CTP input in CTP hardware

DataFormats/Detectors/CTP/include/DataFormatsCTP/Digits.h
lietava: No, unless we change hardware.

Detectors/CTP/simulation/include/CTPSimulation/Digits2Raw.h
lietava: This is a class which would be used by simple executable. Do I need root dictionary ?

Detectors/CTP/simulation/src/Digits2Raw.cxx
lietava: Fixed: mLinkID -> ilink

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
lietava: added full DetID to CTPInput

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
lietava: ok, keeping only DEtID vector

lietava:
@shahor02 I tried to address your comments. Please, have a look.

Detectors/CTP/simulation/include/CTPSimulation/Digits2Raw.h
shahor02: no

lietava (commented):
created by mistake

Detectors/CTP/simulation/include/CTPSimulation/Digits2Raw.h
lietava: @shahor02  I removed Digits2Raw from LinkDef. Also there is no ClassImp in cxx. But when I remove also line ClassDefNV(Digits2Raw, 1); from header it does not compile,  so I left it. Is it ok ?

shahor02: I suspect this happens because you did not suppress https://github.com/AliceO2Group/AliceO2/blob/da03f266d029beb6119a6742a3abd18ee08aef34/Detectors/CTP/simulation/CMakeLists.txt#L20 after suppressing  `ClassDefNV(Digits2Raw, 1);`

lietava: After removing it still complains, please, see below.
Do I need to clean ?
R.
Detectors/CTP/simulation/G__O2CTPSimulation.cxx: In function 'ROOT::TGenericClassInfo* ROOT::GenerateInitInstanceLocal(const o2::ctp::Digits2Raw*)':
Detectors/CTP/simulation/G__O2CTPSimulation.cxx:91:65: error: 'Class_Version' is not a member of 'o2::ctp::Digits2Raw'
   91 |          instance("o2::ctp::Digits2Raw", ::o2::ctp::Digits2Raw::Class_Version(), "", 134,
      |                                                                 ^~~~~~~~~~~~~
Detectors/CTP/simulation/G__O2CTPSimulation.cxx:93:43: error: 'Dictionary' is not a member of 'o2::ctp::Digits2Raw'
   93 |                   &::o2::ctp::Digits2Raw::Dictionary, isa_proxy, 4,
      |                                           ^~~~~~~~~~

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
```diff
@@ -56,41 +64,48 @@ struct CTPDetector {
 struct CTPCluster {
   CTPCluster() = default;
   std::string name;
-  uint32_t detectorsMask;
-  std::vector<std::string> detectorsNames;
+  o2::detectors::DetID::mask_t getClusterDetMask() const;
+  std::vector<o2::detectors::DetID> detectors;
```
shahor02: in principle, instead of the std::vector<o2::detectors::DetID> you could simply use DetID::mask_t

DataFormats/Detectors/CTP/src/Configuration.cxx
shahor02: missed this?

Detectors/CTP/simulation/src/Digits2Raw.cxx
shahor02: On the git you still have ClassDefNV, could you push the code where you removed it and it still fails to compile?

lietava:
@shahor02 pushed

shahor02:
@lietava it should not produce such a message with this code, perhaps cleaning of the CTP/simulation dir. in the BUILD/... would help. But you don't need at all to add this class to the library, instead you should directly create an executable, like e.g.
https://github.com/AliceO2Group/AliceO2/blob/1299fa246ed5a13e31cc0625cfb7332c2b082e5a/Detectors/FIT/FT0/simulation/CMakeLists.txt#L34-L41

lietava:
@shahor02 I do not understand this comment:
https://github.com/AliceO2Group/AliceO2/pull/6278#discussion_r644047157

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
lietava: yes, fine. What is motivation ? We want to save memory ? Is it more elegant ?

lietava:
> @lietava it should not produce such a message with this code, perhaps cleaning of the CTP/simulation dir. in the BUILD/... would help. But you don't need at all to add this class to the library, instead you should directly create an executable, like e.g.
> https://github.com/AliceO2Group/AliceO2/blob/1299fa246ed5a13e31cc0625cfb7332c2b082e5a/Detectors/FIT/FT0/simulation/CMakeLists.txt#L34-L41

@shahor02  After rm -r BUILD and rebuild it is ok. I will also proceed in the way you suggested. For my education: What is motivation not having classes in root dictionary ?

DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h
shahor02: The size and serialization speed: if this goes to ccdb rather than to the data, then the size argument is irrelevant, but to send as a DPL message such object or to store it in the file you need to go via root serialization of every data member.
Anyway, this is a minor comment, if the class is for CCDB only.

shahor02:
@lietava the root dict. is needed only if the class should be persistent (including root serialization via DPL) or you want to use it from the root interpreter command line. Since this compiles to a simple executable, the dict. would just generate an unused code, no profit.

lietava:
@shahor02 I added digi2raw.cxx + minor changes you suggested. Please, have a look.

shahor02 (approved):
looks fine, I've just fixed a misleading option hint in the digi2raw

shahor02:
@lietava I've removed draft status to trigger CI tests. Do you have anything to add here?

lietava:
@shahor02 Thanks. I am adding nothing at the moment.
Please, proceed with review.

shahor02:
@lietava 
Could you correct this ?
````
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/O2/6278/0/DataFormats/Detectors/CTP/src/Configuration.cxx:85:24: error: statement should be inside braces [readability-braces-around-statements]
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/O2/6278/0/DataFormats/Detectors/CTP/src/Configuration.cxx:87:25: error: statement should be inside braces [readability-braces-around-statements]
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/O2/6278/0/DataFormats/Detectors/CTP/src/Configuration.cxx:117:20: error: statement should be inside braces [readability-braces-around-statements]
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/O2/6278/0/DataFormats/Detectors/CTP/src/Configuration.cxx:277:31: error: statement should be inside braces [readability-braces-around-statements]
````
Did you actually try to run digit2raw, does it produce readable file (can be checked by `o2-raw-file-check <rawfilename>`) ?

**Comment by lietava** [comment]
@shahor02 1 I will fix latest on Tuesday. I am off this long weekend.
2 i did not run digi2raw. Marek B can test it. I open issue for him in my rep.

File: Detectors/CTP/simulation/src/digi2raw.cxx [hunk 17]
```diff
@@ -0,0 +1,123 @@
...
+  }
+  // if needed, create output directory
+  if (!std::filesystem::exists(outDirName)) {
+    if (!std::filesystem::create_directories(outDirName)) {
+      LOG(FATAL) << "could not create output directory " << outDirName;
+    } else {
+      LOG(INFO) << "created output directory " << outDirName;
+    }
+  }
+  m2r.setOutDir(outDirName);
```
shahor02, line 117: I guess you should call `m2r.init()` here?

Detectors/CTP/simulation/src/Digits2Raw.cxx
```diff
@@ -0,0 +1,181 @@
...
+  }
+  std::vector<o2::ctp::CTPDigit> CTPDigits, *fCTPDigitsPtr = &CTPDigits;
+  if (digiTree->GetBranch("CTPDigits")) {
+    digiTree->SetBranchAddress("CTPDigits", &fCTPDigitsPtr);
+  } else {
+    LOG(FATAL) << "Branch CTPDigits is missing";
+    return;
+  }
+  // Get first orbit
+  digiTree->GetEntry(0);
```
shahor02: The simulation creates 1 single entry with vectors containing all data

Detectors/CTP/simulation/src/Digits2Raw.cxx
```diff
@@ -0,0 +1,181 @@
...
+  uint32_t orbit0;
+  if (CTPDigits.size() > 0) {
+    orbit0 = CTPDigits[0].intRecord.orbit;
+  } else {
+    LOG(FATAL) << "Branch CTPDigits there but no entries";
+    return;
+  }
+  // Add all CTPdigits for given orbit
+  for (int ient = 1; ient < digiTree->GetEntries(); ient++) {
```
shahor02: therefore you should start from 0 (upper bound `digiTree->GetEntries()` does not hurt, but you have just 1 entry).

Detectors/CTP/simulation/src/Digits2Raw.cxx
```diff
@@ -0,0 +1,181 @@
...
+        std::bitset<NGBT> gbtdig;
+        digit2GBTdigit(gbtdig, ctpdig);
+        hbf.push_back(gbtdig);
+      } else {
+        std::vector<char> buffer;
+        buffer = digits2HBTPayload(hbf, NIntRecPayload);
+        std::string PLTrailer;
+        std::memcpy(buffer.data() + buffer.size() - o2::raw::RDHUtils::GBTWord, PLTrailer.c_str(), o2::raw::RDHUtils::GBTWord);
+        orbit0 = ctpdig.intRecord.orbit;
+      }
```
shahor02: I don't see where do you call `writer.addData(...)`.
BTW, why do you watch orbit change? You just need to call addData for every orbit/bc where you have non-0 payload (and if you call it for 1 link, then this should be done for all. There is a writer-side protection against not doing this, but you better take care of this on your side).

shahor02 (changes requested):

Detectors/CTP/simulation/src/Digits2Raw.cxx
lietava: @shahor02  Yes, fixing it. 
Orbit: for non zero suppressed data I need to add (bcid, 0) for all bcs which are not in input record.
For the moment one link is zero suppressed and other is not.
 **Question** on emptyHBFMethod:  how can I do it link dependent ?

shahor02: > Orbit: for non zero suppressed data I need to add (bcid, 0) for all bcs which are not in input record.

OK, but it should be done in a single loop, w/o pre-reading the entry. 

> Question on emptyHBFMethod: how can I do it link dependent ?

You receive in the emptyHBFMethod the rdh for which the payload need to be generated, so you have the full info about the link.

Detectors/CTP/simulation/src/Digits2Raw.cxx
lietava: @shahor02 So can I assume that  I expect one entry, i.e. more than 1 entry is error ?

shahor02: Currently it is 1 entry, but better do not depend on this assumption, can't you loop over all entries?

lietava: yes, ok, just asking. I'll loop.

Detectors/CTP/simulation/src/Digits2Raw.cxx
lietava: @shahor02 what is PLTrailer ? Do I need it ?

lietava:
HI @shahor02 , I think I did all changes you requested. I added also payload on link 1 (Trigger Classes). Now it produces non empty file.
I am asking Marek to check what is there. Cheers, Roman.

lietava:
@shahor02 Going to move CTP classes calculation from Digits2Raw to Digitizer. 
Related question: we need to connect meaning of CTP inputs with detector inputs (eg V0, T0). Conceptually it can be done either by using input names or also code number - so called signature - we have in run2. Techically I guess I need to access detector config file in database.

shahor02:
@lietava Up to you: the digitizer is supposed to be responsible for the MC digits creation only, MC->raw: for the conversion which will not be done in the standard MC production (there we will use digits directly).
For the inputs: I am not sure if FIT detectors have these signatures, perhaps you should invent one, create a config object on your side and tell to FIT to use them?

lietava:
@shahor02:
1.) I moved detector to input/class translation to Digitizer meaning that Digits2Raw and Raw2Digits will not need CTP config. Difference between raw and digits is only different data packing in CTP case
2.) I will discus with detectors inputs conventions
3.) ROOT and bitset: does it mean we can never change CTP_NINPUTS and CTP_NCLASSES (see Digits.h) as only those are used in the class which is used by ROOT ? ( I mean there is also CTP_MAXTRIGINPPERDET).
4.) What next : merge or should I add Raw2Digit ?

shahor02:
> 3.) ROOT and bitset: does it mean we can never change CTP_NINPUTS and CTP_NCLASSES (see Digits.h) as only those are used in the class which is used by ROOT ? ( I mean there is also CTP_MAXTRIGINPPERDET).

Yes, ROOT schema evolution does not support template parameters change, and since bitset<N1> and bitset<N2> are different classes, if you change the constant, you cannot read old data.

For that reason in the detector mask (https://github.com/AliceO2Group/AliceO2/blob/aa1308a84e367aa334e2fb51c7f50cd036af77ef/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/DetID.h#L90) we are using fixed 32-slot mask although we have only 17 (or 20 with future upgrades) detectors. Note that we are using custom implementation of David, https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/Utils/GPUCommonBitSet.h which takes only 32 bits and cannot sore more (the std::bitset takes at least 64 bits)

> 4.) What next : merge or should I add Raw2Digit ?
We cannot use raw w/o decoder, but I would merge this once the tests are passed, to avoid divergences.

lietava:
@shahor02 ok, thanks, waiting for merge finished.  Give me examples on decoder, please.

lietava:
@shahor02 Do you still have request or is it previous ?

shahor02:
Merging, the mask problem, if needed, can be addressed in other PR. For the decoding, e.g., https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/CPV/workflow/src/RawToDigitConverterSpec.cxx, which is a part of https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/CPV/workflow/src/cpv-reco-workflow.cxx.

lietava:
@shahor02 : what is 'the mask problem' ?

shahor02:
The bitsets you were asking about. If you think the template parameters may change, better to reserve at once sufficient number of bits, even if now you don't use them.

PR #6283, opened 2021-05-31
DPL: cleanup AODReaderHelpers.cxx

ktf:
RDataFrame is not actually needed.


ktf:
@TimoWilken @MichaelLettrich I think there is an issue with the mac Builder.
@TimoWilken I think also the fullCI builder has troubles.

PR #6285, opened 2021-05-31
avoid ZMQ assert during shutdown

sawenzel:
This device had random ZMQ asserts during destruction.
The commit is providing a hotfix to stabilize the CI
until the real cause of the problem is found.


sawenzel:
better done by https://github.com/AliceO2Group/AliceO2/pull/6286

PR #6287, opened 2021-05-31
Fixes for alignment framework

shahor02:
* Modifications for the alignment application to the geometry
GeomManager::loadGeomerty(..) will trigger retrieval of the alignment objects for all detectors from CCDB
and their application to geometry (unless the applyMisalignment argument is set to false).
This process can be steered (CCDB to use, detectors list, timestamp) via
ConfigurableParam o2::base::Aligner, see O2/Detectors/Base/README.md

* Sample macros for dummy alignment objects uploading
-- macro/UploadDummyAlignment.C to populate CCDB with empty alignment containers for all detectors
-- Detectors/ITSMFT/ITS/macros/test/ITSMisaligner.C to upload ITS container with alignment objects
(with all parameters set to 0, so the geometry effectively will stay the same)

* Do not use FairMCApplication::MisalignGeometry
Instead, we will start with ideal geometry and misalign it with GeomManager on the fly

* Add standard alignment object path to NameConf (<DET>/Align)

* Misc fixes
-- Improve det.list to mask_t conversion
-- gpu::bitset prints bits from left to right (like std::bitser)


sawenzel (dismissed):

shahor02:
@peressounko : the fullCI (which fails for unrelated reasons), reports:
````
[7662:PHOSDigitizer]: [01:10:09][INFO] [PHOSDigitizer] No reading calibration from ccdb requested, set default
[7662:PHOSDigitizer]: [01:10:09][ERROR] impossible parameterization 0
[7662:PHOSDigitizer]: [01:10:09][ERROR] Available are:
````
Is this expected?

sawenzel:
@peressounko @shahor02 : I see the same ERROR messages in the GRID workflows.

iouribelikov (dismissed):

peressounko:
Dear @sawenzel, @shahor02 this message should not cause crush, but indeed reflects problem which I will fix in next PR

shahor02:
FullCI failure is unrelated, merging

PR #6289, opened 2021-06-01
DPL: avoid dependency on boost for trivial stuff

ktf:


ktf:
@TimoWilken this one was started Tue 12:22 CET and its 20:00 CET.

PR #6290, opened 2021-06-01
DPL: reduce template instanciation

ktf:


ktf:
Trivial and compiles on o2-dataflow and my own macos machine. Merging.

PR #6291, opened 2021-06-01
[MCH] use configurable parameters for MCH clustering

pillot:
This moves all the parameters of the original clustering that needs to be adjusted for run3 in a ConfigurableParam object.

For the moment, there are 2 ways to modify these parameters:

from command line with the workflow option --configKeyValues "key1=value1;key2=value2;..."
from a config file (JSON or INI format) with the processor option --config "file.json(ini)"

In the end I imagine they will be taken from the CCDB.

### DIFF
--- a/Detectors/MUON/MCH/Clustering/CMakeLists.txt
+++ b/Detectors/MUON/MCH/Clustering/CMakeLists.txt
@@ -12,4 +12,9 @@ o2_add_library(MCHClustering
                SOURCES src/ClusterOriginal.cxx
                        src/ClusterFinderOriginal.cxx
                        src/MathiesonOriginal.cxx
-               PUBLIC_LINK_LIBRARIES O2::MCHMappingInterface O2::MCHBase O2::MCHPreClustering O2::Framework)
+                       src/ClusterizerParam.cxx
+               PUBLIC_LINK_LIBRARIES O2::MCHMappingInterface O2::MCHBase O2::MCHPreClustering
+                                     O2::Framework O2::CommonUtils)
+
+o2_target_root_dictionary(MCHClustering
+                          HEADERS include/MCHClustering/ClusterizerParam.h)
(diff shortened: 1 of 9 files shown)


aphecetche (approved):

aphecetche:
@shahor02 I'm approving this one, like I've approved #6267, as it adds a feature we need and I believe we're not using ConfigurableParam more wrongly than in other places :smile:  but for the record I'd have a question / reservation. 
It's about the usage of the ConfigurableParam singleton in those classes (clustering and tracking). Basically if someone messes with the singleton (outside the clustering/tracking class itself) then the clustering got hit by those changes directly. Said otherwise the ConfigurableParam can't be made read-only. I was wondering if you'd have some advise concerning this or consider it as a non issue simply ?

sawenzel:
@aphecetche : The singleton should be already strictly read-only and the idea was to allow value-setting only during initialization (and to have provenance tracking).
I am aware of the shortcomings in the current system but I believe that we also offer a lot of good features: "Ease of declaring a parameter, automatic translation between text and C++ etc.". 

Eventually one could put some more effort into the system. To be constructive, it would be good to collect some wish-list of improvements (but also stating the features that are good and should be kept).

shahor02:
@aphecetche sorry for delay, just noticed the question was to me. Sandro has already replied, it has shortcomings, but the benefits outweigh. 
I don't think the singleton property in this context is so dangerous, unless your code has abuses it with run-time modification via `::setValue(..)`, which obviously should not be done. Otherwise, the params are updated from the command line (or ini file) at the start of the device, and then may be updated from CCDB during execution of new TF (if it is used as a CCDB object). In both cases it is updated for the whole process, one should simply not use this class if you may need different instances in the same process.
For me, the only real danger I see  is when you load the values via `updateFromFile` feature: the ini file written by the particular `ConfigurableParam` will contain all params visible in the scope of the writer. So, if in the code you have say 2 such classes, say `paramA` and `paramB`, and you want to update the `paramA` by reading the `ini` previously written by `paramA.writeINI()`, you risk to update also the `paramB` state.
So, I would recommend to use the feature added in https://github.com/AliceO2Group/AliceO2/pull/5854, which allows to read only limited range of main keys (param name).

aphecetche:
@sawenzel @shahor02 thanks for your answers. The thing is that I indeed don't have any constructive proposal for the moment. So let's put that discussion aside for now.
Regardless, the CI errors on this one seem unrelated to the PR itself. Can it be merged ?

PR #6292, opened 2021-06-01
[o2femtodream] First functional prototype of the filtering

a-mathis:
o Generic cut container class
o Generic cut class that allows for arbitrary specialization
o Track selection including systematic variations encoded bit-wise for customizable selection criteria
o Producer tasks writes minimal information to file

### DIFF
--- a/Analysis/Tasks/PWGCF/CMakeLists.txt
+++ b/Analysis/Tasks/PWGCF/CMakeLists.txt
@@ -8,6 +8,8 @@
 # granted to it by virtue of its status as an Intergovernmental Organization or
 # submit itself to any jurisdiction.
 
+add_subdirectory(FemtoDream)
+
 o2_add_library(PWGCFCore
                SOURCES  AnalysisConfigurableCuts.cxx
                PUBLIC_LINK_LIBRARIES O2::Framework O2::AnalysisDataModel)
--- a/Analysis/Tasks/PWGCF/FemtoDream/CMakeLists.txt
+++ b/Analysis/Tasks/PWGCF/FemtoDream/CMakeLists.txt
@@ -1,0 +1,14 @@
+# Copyright CERN and copyright holders of ALICE O2. This software is distributed
+# under the terms of the GNU General Public License v3 (GPL Version 3), copied
+# verbatim in the file "COPYING".
+#
+# See http://alice-o2.web.cern.ch/license for full licensing information.
+#
+# In applying this license CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization or
+# submit itself to any jurisdiction.
+
+o2_add_dpl_workflow(femtodream-producer
+          SOURCES femtoDreamProducerTask.cxx
+          PUBLIC_LINK_LIBRARIES O2::Framework O2::AnalysisCore O2::AnalysisDataModel
+          COMPONENT_NAME Analysis)
--- a/Analysis/Tasks/PWGCF/FemtoDream/femtoDreamProducerTask.cxx
+++ b/Analysis/Tasks/PWGCF/FemtoDream/femtoDreamProducerTask.cxx
@@ -1,0 +1,158 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
+// distributed under the terms of the GNU General Public License v3 (GPL
+// Version 3), copied verbatim in the file "COPYING".
+//
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file femtoDreamProducerTask.cxx
+/// \brief Tasks that produces the track tables used for the pairing
+/// \author Andi Mathis, TU München, andreas.mathis@ph.tum.de
+
+#include "include/FemtoDream/FemtoDreamCollisionSelection.h"
+#include "include/FemtoDream/FemtoDreamTrackSelection.h"
+#include "include/FemtoDream/FemtoDerived.h"
+#include "Framework/AnalysisDataModel.h"
+#include "Framework/AnalysisTask.h"
+#include "Framework/runDataProcessing.h"
+#include "Framework/HistogramRegistry.h"
+#include "Framework/ASoAHelpers.h"
+#include "AnalysisDataModel/TrackSelectionTables.h"
+#include "AnalysisDataModel/PID/PIDResponse.h"
+#include "AnalysisDataModel/EventSelection.h"
+#include "AnalysisDataModel/Multiplicity.h"
+#include "ReconstructionDataFormats/Track.h"
+
+using namespace o2;
+using namespace o2::analysis::femtoDream;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+namespace o2::aod
+{
+using FilteredFullCollision = soa::Filtered<soa::Join<aod::Collisions,
+                                                      aod::EvSels,
+                                                      aod::Mults>>::iterator;
+using FilteredFullTracks = soa::Filtered<soa::Join<aod::FullTracks,
+                                                   aod::TracksExtended,
+                                                   aod::pidTPCEl, aod::pidTPCMu, aod::pidTPCPi,
+                                                   aod::pidTPCKa, aod::pidTPCPr, aod::pidTPCDe,
+                                                   aod::pidTOFEl, aod::pidTOFMu, aod::pidTOFPi,
+                                                   aod::pidTOFKa, aod::pidTOFPr, aod::pidTOFDe>>;
+} // namespace o2::aod
+
+struct femtoDreamProducerTask {
+
+  Produces<aod::FemtoDreamCollisions> outputCollision;
+  Produces<aod::FemtoDreamParticles> outputTracks;
+  Produces<aod::FemtoDreamDebugParticles> outputDebugTracks;
+
+  Configurable<bool> ConfDebugOutput{"ConfDebugOutput", true, "Debug output"};
+
+  /// Event cuts
+  FemtoDreamCollisionSelection colCuts;
+  Configurable<float> ConfEvtZvtx{"ConfEvtZvtx", 10.f, "Evt sel: Max. z-Vertex (cm)"};
+  Configurable<bool> ConfEvtTriggerCheck{"ConfEvtTriggerCheck", false, "Evt sel: check for trigger"};
+  Configurable<int> ConfEvtTriggerSel{"ConfEvtTriggerSel", kINT7, "Evt sel: trigger"};
+  Configurable<bool> ConfEvtOfflineCheck{"ConfEvtOfflineCheck", false, "Evt sel: check for offline selection"};
+
+  Filter colFilter = nabs(aod::collision::posZ) < ConfEvtZvtx;
+
+  FemtoDreamTrackSelection trackCuts;
+  Configurable<std::vector<float>> ConfTrkCharge{"ConfTrkCharge", std::vector<float>{-1, 1}, "Trk sel: Charge"};
+  Configurable<std::vector<float>> ConfTrkPtmin{"ConfTrkPtmin", std::vector<float>{0.4f, 0.6f, 0.5f}, "Trk sel: Min. pT (GeV/c)"};
+  Configurable<std::vector<float>> ConfTrkPtmax{"ConfTrkPtmax", std::vector<float>{4.05f, 999.f}, "Trk sel: Max. pT (GeV/c)"};
+  Configurable<std::vector<float>> ConfTrkEta{"ConfTrkEta", std::vector<float>{0.8f, 0.7f, 0.9f}, "Trk sel: Max. eta"};
+  Configurable<std::vector<float>> ConfTrkTPCnclsMin{"ConfTrkTPCnclsMin", std::vector<float>{80.f, 70.f, 60.f}, "Trk sel: Min. nCls TPC"};
+  Configurable<std::vector<float>> ConfTrkTPCfCls{"ConfTrkTPCfCls", std::vector<float>{0.7f, 0.83f, 0.9f}, "Trk sel: Min. ratio crossed rows/findable"};
+  Configurable<std::vector<float>> ConfTrkTPCsCls{"ConfTrkTPCsCls", std::vector<float>{0.1f, 160.f}, "Trk sel: Max. TPC shared cluster"};
+  Configurable<std::vector<float>> ConfTrkDCAxyMax{"ConfTrkDCAxyMax", std::vector<float>{0.1f, 3.5f}, "Trk sel: Max. DCA_xy (cm)"}; /// here we need an open cut to do the DCA fits later on!
+  Configurable<std::vector<float>> ConfTrkDCAzMax{"ConfTrkDCAzMax", std::vector<float>{0.2f, 3.5f}, "Trk sel: Max. DCA_z (cm)"};
+  /// \todo maybe we need to remove the PID from the general cut container and have a separate one, these are lots and lots of bits we need
+  Configurable<std::vector<float>> ConfTrkPIDnSigmaMax{"ConfTrkPIDnSigmaMax", std::vector<float>{3.5f, 3.f, 2.5f}, "Trk sel: Max. PID nSigma"};
+  Configurable<std::vector<int>> ConfTrkTPIDspecies{"ConfTrkTPIDspecies", std::vector<int>{o2::track::PID::Electron, o2::track::PID::Pion, o2::track::PID::Kaon, o2::track::PID::Proton, o2::track::PID::Deuteron}, "Trk sel: Particles species for PID"};
+
+  // for now this selection does not work yet, however will very soon
+  // \todo once this is the case, set the limits less strict!
+  MutableConfigurable<float> TrackMinSelPtMin{"TrackMinSelPtMin", 0.4f, "(automatic) Minimal pT selection for tracks"};
+  MutableConfigurable<float> TrackMinSelPtMax{"TrackMinSelPtMax", 10.f, "(automatic) Maximal pT selection for tracks"};
+  MutableConfigurable<float> TrackMinSelEtaMax{"TrackMinSelEtaMax", 1.f, "(automatic) Maximal eta selection for tracks"};
+
+  Filter trackFilter = (aod::track::pt > TrackMinSelPtMin.value) &&
+                       (aod::track::pt < TrackMinSelPtMax.value) &&
+                       (nabs(aod::track::eta) < TrackMinSelEtaMax.value);
+
+  HistogramRegistry qaRegistry{"QAHistos", {}, OutputObjHandlingPolicy::QAObject};
+
+  void init(InitContext&)
+  {
+    colCuts.setCuts(ConfEvtZvtx, ConfEvtTriggerCheck, ConfEvtTriggerSel, ConfEvtOfflineCheck);
+    colCuts.init(&qaRegistry);
+
+    trackCuts.setSelection(ConfTrkCharge, femtoDreamTrackSelection::kSign, femtoDreamSelection::kEqual);
+    trackCuts.setSelection(ConfTrkPtmin, femtoDreamTrackSelection::kpTMin, femtoDreamSelection::kLowerLimit);
+    trackCuts.setSelection(ConfTrkPtmax, femtoDreamTrackSelection::kpTMax, femtoDreamSelection::kUpperLimit);
+    trackCuts.setSelection(ConfTrkEta, femtoDreamTrackSelection::kEtaMax, femtoDreamSelection::kAbsUpperLimit);
+    trackCuts.setSelection(ConfTrkTPCnclsMin, femtoDreamTrackSelection::kTPCnClsMin, femtoDreamSelection::kLowerLimit);
+    trackCuts.setSelection(ConfTrkTPCfCls, femtoDreamTrackSelection::kTPCfClsMin, femtoDreamSelection::kLowerLimit);
+    trackCuts.setSelection(ConfTrkTPCsCls, femtoDreamTrackSelection::kTPCsClsMax, femtoDreamSelection::kUpperLimit);
+    trackCuts.setSelection(ConfTrkDCAxyMax, femtoDreamTrackSelection::kDCAxyMax, femtoDreamSelection::kAbsUpperLimit);
+    trackCuts.setSelection(ConfTrkDCAzMax, femtoDreamTrackSelection::kDCAzMax, femtoDreamSelection::kAbsUpperLimit);
+    trackCuts.setSelection(ConfTrkPIDnSigmaMax, femtoDreamTrackSelection::kPIDnSigmaMax, femtoDreamSelection::kAbsUpperLimit);
+    trackCuts.setPIDSpecies(ConfTrkTPIDspecies);
+    trackCuts.init(&qaRegistry);
+
+    if (trackCuts.getNSelections(femtoDreamTrackSelection::kpTMin) > 0) {
+      TrackMinSelPtMin.value = trackCuts.getMinimalSelection(femtoDreamTrackSelection::kpTMin, femtoDreamSelection::kLowerLimit);
+    }
+    if (trackCuts.getNSelections(femtoDreamTrackSelection::kpTMax) > 0) {
+      TrackMinSelPtMax.value = trackCuts.getMinimalSelection(femtoDreamTrackSelection::kpTMax, femtoDreamSelection::kUpperLimit);
+    }
+    if (trackCuts.getNSelections(femtoDreamTrackSelection::kEtaMax) > 0) {
+      TrackMinSelEtaMax.value = trackCuts.getMinimalSelection(femtoDreamTrackSelection::kEtaMax, femtoDreamSelection::kAbsUpperLimit);
+    }
+  }
+
+  void process(aod::FilteredFullCollision const& col,
+               aod::FilteredFullTracks const& tracks)
+  {
+
+    if (!colCuts.isSelected(col)) {
+      return;
+    }
+    const auto vtxZ = col.posZ();
+    const auto mult = col.multV0M();
+    const auto spher = colCuts.computeSphericity(col, tracks);
+    colCuts.fillQA(col);
+    outputCollision(vtxZ, mult, spher);
+
+    for (auto& track : tracks) {
+      if (!trackCuts.isSelectedMinimal(track)) {
+        continue;
+      }
+      trackCuts.fillQA(track);
+      auto cutContainer = trackCuts.getCutContainer(track);
+      if (cutContainer > 0) {
+        trackCuts.fillCutQA(track, cutContainer);
+        outputTracks(outputCollision.lastIndex(), track.pt(), track.eta(), track.phi(), cutContainer, track.dcaXY());
+        if (ConfDebugOutput) {
+          outputDebugTracks(outputCollision.lastIndex(),
+                            track.sign(), track.tpcNClsFound(),
+                            track.tpcNClsFindable(),
+                            track.tpcNClsCrossedRows(), track.tpcNClsShared(), track.dcaXY(), track.dcaZ(),
+                            track.tpcNSigmaEl(), track.tpcNSigmaPi(), track.tpcNSigmaKa(), track.tpcNSigmaPr(), track.tpcNSigmaDe(),
+                            track.tofNSigmaEl(), track.tofNSigmaPi(), track.tofNSigmaKa(), track.tofNSigmaPr(), track.tofNSigmaDe());
+        }
+      }
+    }
+  }
+};
+
+WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
+{
+  WorkflowSpec workflow{adaptAnalysisTask<femtoDreamProducerTask>(cfgc)};
+  return workflow;
+}
--- a/Analysis/Tasks/PWGCF/FemtoDream/include/FemtoDream/FemtoDerived.h
+++ b/Analysis/Tasks/PWGCF/FemtoDream/include/FemtoDream/FemtoDerived.h
@@ -1,0 +1,98 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
+// distributed under the terms of the GNU General Public License v3 (GPL
+// Version 3), copied verbatim in the file "COPYING".
+//
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef ANALYSIS_TASKS_PWGCF_FEMTODREAM_FEMTODERIVED_H_
+#define ANALYSIS_TASKS_PWGCF_FEMTODREAM_FEMTODERIVED_H_
+
+#include "Framework/ASoA.h"
+#include "AnalysisDataModel/Multiplicity.h"
+#include "Framework/AnalysisDataModel.h"
+#include "AnalysisDataModel/TrackSelectionTables.h"
+#include "AnalysisDataModel/PID/PIDResponse.h"
+
+namespace o2::aod
+{
+
+namespace femtodreamcollision
+{
+DECLARE_SOA_COLUMN(MultV0M, multV0M, float);
+DECLARE_SOA_COLUMN(Sphericity, sphericity, float);
+} // namespace femtodreamcollision
+DECLARE_SOA_TABLE(FemtoDreamCollisions, "AOD", "FEMTODREAMCOLS",
+                  o2::soa::Index<>,
+                  o2::aod::collision::PosZ,
+                  femtodreamcollision::MultV0M,
+                  femtodreamcollision::Sphericity);
+using FemtoDreamCollision = FemtoDreamCollisions::iterator;
+
+namespace femtodreamparticle
+{
+DECLARE_SOA_INDEX_COLUMN(FemtoDreamCollision, femtoDreamCollision);
+DECLARE_SOA_COLUMN(Pt, pt, float);
+DECLARE_SOA_COLUMN(Eta, eta, float);
+DECLARE_SOA_COLUMN(Phi, phi, float);
+DECLARE_SOA_COLUMN(Cut, cut, uint64_t);
+DECLARE_SOA_COLUMN(TempFitVar, tempFitVar, float);
+// debug variables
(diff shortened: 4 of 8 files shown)


a-mathis:
This is part 1 of the o2femtodream framework, and contains all relevant files to filter the tracks to custom objects
In particular, we developed a class that enables arbitrary selections, that can further be applied for v0/cascade/... filtering
@sartozza

a-mathis:
The failing test is unrelated to this commit
FAILED: Framework/Core/CMakeFiles/O2test-framework-GroupSlicer.dir/test/test_GroupSlicer.cxx.o 
/mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/GCC-Toolchain/v10.2.0-alice2-1/bin/c++ -DBOOST_ALL_NO_LIB -DBOOST_CONTAINER_DYN_LINK -DBOOST_ERROR_CODE_HEADER_ONLY -DBOOST_FILESYSTEM_DYN_LINK -DBOOST_IOSTREAMS_DYN_LINK -DBOOST_PROGRAM_OPTIONS_DYN_LINK -DBOOST_REGEX_DYN_LINK -DBOOST_SERIALIZATION_DYN_LINK -DBOOST_UNIT_TEST_FRAMEWORK_DYN_LINK -DDPL_ENABLE_BACKTRACE -DENABLE_UPGRADES -DFAIRMQ_HAS_STD_FILESYSTEM=1 -DFMT_LOCALE -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/Framework/Core/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/Common/Utils/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/DataFormats/common/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/Common/Constants/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/GPU/Common -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/Framework/Logger/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/DataFormats/Headers/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/DataFormats/MemoryResources/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/Common/MathUtils/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/Framework/Foundation/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/Framework/Foundation/3rdparty/include -I/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/Utilities/PCG/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/Common-O2/v1.6.0-5/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/boost/v1.75.0-10/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/Configuration/v2.6.1-17/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/libInfoLogger/v2.1.0-2/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/Monitoring/v3.8.5-1/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/curl/7.70.0-29/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/FairMQ/v1.4.38-1/include/fairmq -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/FairMQ/v1.4.38-1/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/FairLogger/v1.9.1-5/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/fmt/7.1.0-17/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/ROOT/v6-20-08-alice1-78/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/FairRoot/v18.4.1-115/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/arrow/v1.0.0-68/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/ms_gsl/3.1.0-3/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/Vc/1.4.1-62/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/RapidJSON/v1.1.0-alice2-9/include -isystem /mnt/mesos/sandbox/sandbox/o2/sw/slc7_x86-64/libuv/v1.40.0-5/include -fPIC -O2 -std=c++17 -Werror -Wno-error=deprecated-declarations -O2 -g -std=gnu++17 -MD -MT Framework/Core/CMakeFiles/O2test-framework-GroupSlicer.dir/test/test_GroupSlicer.cxx.o -MF Framework/Core/CMakeFiles/O2test-framework-GroupSlicer.dir/test/test_GroupSlicer.cxx.o.d -o Framework/Core/CMakeFiles/O2test-framework-GroupSlicer.dir/test/test_GroupSlicer.cxx.o -c /mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6292/0/Framework/Core/test/test_GroupSlicer.cxx

Analysis/Tasks/PWGCF/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@
 # granted to it by virtue of its status as an Intergovernmental Organization or
 # submit itself to any jurisdiction.
 
+add_subdirectory(o2femtodream)
```
jgrosseo: let's remove "o2" here

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,111 @@
...
+namespace femtodreamparticle
+{
+DECLARE_SOA_INDEX_COLUMN(FemtoDreamCollision, femtoDreamCollision);
+DECLARE_SOA_COLUMN(pT, pt, float);
+DECLARE_SOA_COLUMN(Eta, eta, float);
+DECLARE_SOA_COLUMN(Phi, phi, float);
+DECLARE_SOA_COLUMN(Cut, cut, unsigned long int);
+DECLARE_SOA_COLUMN(TempFitVar, tempFitVar, float);
+// debug variables
+DECLARE_SOA_COLUMN(Charge, charge, int);
```
iarsene: This should be called Sign to keep in line with the framework data model. The column is computed anyway based on track.sign(). But maybe @jgrosseo could share his opinion here.

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,111 @@
...
+DECLARE_SOA_COLUMN(PIDTPCEl, pidTPCEl, float);
+DECLARE_SOA_COLUMN(PIDTPCPi, pidTPCPi, float);
+DECLARE_SOA_COLUMN(PIDTPCKa, pidTPCKa, float);
+DECLARE_SOA_COLUMN(PIDTPCPr, pidTPCPr, float);
+DECLARE_SOA_COLUMN(PIDTPCDe, pidTPCDe, float);
+DECLARE_SOA_COLUMN(PIDTOFEl, pidTOFEl, float);
+DECLARE_SOA_COLUMN(PIDTOFPi, pidTOFPi, float);
+DECLARE_SOA_COLUMN(PIDTOFKa, pidTOFKa, float);
+DECLARE_SOA_COLUMN(PIDTOFPr, pidTOFPr, float);
+DECLARE_SOA_COLUMN(PIDTOFDe, pidTOFDe, float);
```
iarsene: This is just a suggestion. I think its advantageous to reuse the columns already defined in the framework data model or in the additional producers (e.g. PID task), whenever reasonably possible. Its both because of code duplication, but also having similar member methods lets you use tracks from different data models to be used in templated functions.

a-mathis: very good point, I adjusted accordingly whenever possible!

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
a-mathis: good point, done!

a-mathis:
@iarsene @jgrosseo I updated the PR according to Ionut's suggestions
Thanks a lot!

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
```diff
@@ -0,0 +1,138 @@
...
+  Configurable<std::vector<float>> ConfTrkEta{"ConfTrkEta", std::vector<float>{0.8f, 0.7f, 0.9f}, "Trk sel: Max. eta"};
+  Configurable<std::vector<float>> ConfTrkTPCnclsMin{"ConfTrkTPCnclsMin", std::vector<float>{80.f, 70.f, 90.f}, "Trk sel: Min. nCls TPC"};
+  Configurable<std::vector<float>> ConfTrkTPCfCls{"ConfTrkTPCfCls", std::vector<float>{0.7f, 0.83f, 0.9f}, "Trk sel: Min. ratio crossed rows/findable"};
+  Configurable<std::vector<float>> ConfTrkTPCsCls{"ConfTrkTPCsCls", std::vector<float>{0.1f, 160.f}, "Trk sel: Max. TPC shared cluster"};
+  Configurable<std::vector<float>> ConfTrkDCAxyMax{"ConfTrkDCAxyMax", std::vector<float>{0.1f, 999.f}, "Trk sel: Max. DCA_xy (cm)"}; /// here we need an open cut to do the DCA fits later on!
+  Configurable<std::vector<float>> ConfTrkDCAzMax{"ConfTrkDCAzMax", std::vector<float>{0.2f, 999.f}, "Trk sel: Max. DCA_z (cm)"};    /// open cut needed for v0s etc. also there we need a lower limit bit
+  /// \todo maybe we need to remove the PID from the general cut container and have a separate one, these are lots and lots of bits we need
```
jgrosseo: Would it be an idea that not all pid bits are supported in one sample? E.g. a maximum of three, but configurable?

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
```diff
@@ -0,0 +1,138 @@
...
+  Configurable<std::vector<float>> ConfTrkPIDnSigmaMax{"ConfTrkPIDnSigmaMax", std::vector<float>{3.5f, 3.f, 2.5f}, "Trk sel: Max. PID nSigma"};
+  Configurable<std::vector<int>> ConfTrkTPIDspecies{"ConfTrkTPIDspecies", std::vector<int>{o2::track::PID::Electron, o2::track::PID::Pion, o2::track::PID::Kaon, o2::track::PID::Proton, o2::track::PID::Deuteron}, "Trk sel: Particles species for PID"};
+
+  /// todo with configurables - how to make sure we use the smallest/largest value?
```
jgrosseo: you want the smallest and largest from the std::vector array above?
I think you can configure an additional configurable here and overwrite it in the `init` function. I.e. loop over the array and put the smallest.

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
```diff
@@ -0,0 +1,138 @@
...
+    trackCuts.setSelection(ConfTrkTPCfCls, femtoDreamTrackSelection::kTPCfClsMin, femtoDreamSelection::kLowerLimit);
+    trackCuts.setSelection(ConfTrkTPCsCls, femtoDreamTrackSelection::kTPCsClsMax, femtoDreamSelection::kUpperLimit);
+    trackCuts.setSelection(ConfTrkDCAxyMax, femtoDreamTrackSelection::kDCAxyMax, femtoDreamSelection::kAbsUpperLimit);
+    trackCuts.setSelection(ConfTrkDCAzMax, femtoDreamTrackSelection::kDCAzMax, femtoDreamSelection::kAbsUpperLimit);
+    trackCuts.setSelection(ConfTrkPIDnSigmaMax, femtoDreamTrackSelection::kPIDnSigmaMax, femtoDreamSelection::kAbsUpperLimit);
+    trackCuts.setPIDSpecies(ConfTrkTPIDspecies);
+    trackCuts.init(&qaRegistry);
+  }
+
+  void process(o2::aod::FilteredFullCollision const& col,
```
jgrosseo: you can drop o2 because you use the namespace above

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,84 @@
...
+                  o2::soa::Index<>,
+                  o2::aod::collision::PosZ,
+                  femtodreamcollision::MultV0M,
+                  femtodreamcollision::Sphericity);
+using FemtoDreamCollision = FemtoDreamCollisions::iterator;
+
+namespace femtodreamparticle
+{
+DECLARE_SOA_INDEX_COLUMN(FemtoDreamCollision, femtoDreamCollision);
+DECLARE_SOA_COLUMN(pT, pt, float);
```
jgrosseo: Can you put Pt even if it looks funny?

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,84 @@
...
+DECLARE_SOA_COLUMN(Eta, eta, float);
+DECLARE_SOA_COLUMN(Phi, phi, float);
+DECLARE_SOA_COLUMN(Cut, cut, unsigned long int);
+DECLARE_SOA_COLUMN(TempFitVar, tempFitVar, float);
+// debug variables
+DECLARE_SOA_COLUMN(Sign, sign, int);
```
jgrosseo: int8 should be enough here

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,84 @@
...
+                  femtodreamcollision::Sphericity);
+using FemtoDreamCollision = FemtoDreamCollisions::iterator;
+
+namespace femtodreamparticle
+{
+DECLARE_SOA_INDEX_COLUMN(FemtoDreamCollision, femtoDreamCollision);
+DECLARE_SOA_COLUMN(pT, pt, float);
+DECLARE_SOA_COLUMN(Eta, eta, float);
+DECLARE_SOA_COLUMN(Phi, phi, float);
+DECLARE_SOA_COLUMN(Cut, cut, unsigned long int);
```
jgrosseo: please use uint64

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,84 @@
...
+DECLARE_SOA_COLUMN(TempFitVar, tempFitVar, float);
+// debug variables
+DECLARE_SOA_COLUMN(Sign, sign, int);
+DECLARE_SOA_COLUMN(TPCNClsFound, tpcNClsFound, int);
+DECLARE_SOA_COLUMN(TPCCrossedRowsOverFindableCls, tpcCrossedRowsOverFindableCls, float);
```
jgrosseo: better store findable as well (int8) only and compute TPCCrossedRowsOverFindableCls in a dynamic column. You save 3 byte per track

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,84 @@
...
+{
+DECLARE_SOA_INDEX_COLUMN(FemtoDreamCollision, femtoDreamCollision);
+DECLARE_SOA_COLUMN(pT, pt, float);
+DECLARE_SOA_COLUMN(Eta, eta, float);
+DECLARE_SOA_COLUMN(Phi, phi, float);
+DECLARE_SOA_COLUMN(Cut, cut, unsigned long int);
+DECLARE_SOA_COLUMN(TempFitVar, tempFitVar, float);
+// debug variables
+DECLARE_SOA_COLUMN(Sign, sign, int);
+DECLARE_SOA_COLUMN(TPCNClsFound, tpcNClsFound, int);
```
jgrosseo: uint8

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,84 @@
...
+DECLARE_SOA_COLUMN(TPCCrossedRowsOverFindableCls, tpcCrossedRowsOverFindableCls, float);
+DECLARE_SOA_COLUMN(TPCNClsCrossedRows, tpcNClsCrossedRows, int);
```
jgrosseo: uint8

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,84 @@
...
+
+} // namespace femtodreamparticle
+DECLARE_SOA_TABLE(FemtoDreamParticles, "AOD", "FemtoDreamParts",
```
jgrosseo: Third argument in capitals (same below)

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,84 @@
...
+                  soa::Index<>,
+                  femtodreamparticle::FemtoDreamCollisionId,
+                  femtodreamparticle::Sign,
+                  femtodreamparticle::TPCNClsFound,
+                  femtodreamparticle::TPCCrossedRowsOverFindableCls,
+                  femtodreamparticle::TPCNClsCrossedRows,
+                  track::TPCNClsShared,
+                  track::DcaXY,
+                  track::DcaZ,
+                  pidtpc::TPCNSigmaEl,
```
jgrosseo: This are a float each. Could you work with the 8 bit precision which we use here:
https://github.com/jgrosseo/AliceO2/blob/dev/Analysis/DataModel/include/AnalysisDataModel/PID/PIDResponse.h#L220

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamCollisionSelection.h
```diff
@@ -0,0 +1,121 @@
...
+  triggerAliases mTrigger = kINT7; ///< Trigger to check for
+  bool mCheckOffline = false;      ///< Check for offline criteria (might change)
+
+  HistogramRegistry* mHistogramRegistry = nullptr; ///< For QA output
+
+  ClassDefNV(FemtoDreamCollisionSelection, 1);
+};
+
+template <typename T>
+inline void FemtoDreamCollisionSelection::fillQA(T const& col)
```
jgrosseo: @eulisse Given the discussion this morning at the WP14, do you propose to inline this, or rather not?

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamCollisionSelection.h
```diff
@@ -0,0 +1,121 @@
...
+    return false;
+  }
+
+  return true;
+}
+
+template <typename T1, typename T2>
+inline float FemtoDreamCollisionSelection::computeSphericity(T1 const& col, T2 const& tracks)
+{
+  return 2.f;
```
jgrosseo: I like this :D

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamObjectSelection.h
```diff
@@ -0,0 +1,156 @@
...
+        break;
+    }
+    for (const auto sel : sels) {
+      mSelections.push_back(sel);
+    }
+  }
+
+  T1 getMinimalSelection(T2 selVar, femtoDreamSelection::SelectionType selType)
+  {
+    T1 minimalSel;
```
jgrosseo: These will be executed per object? We may need to profile this? I wonder if instead one could produce a partition expression which does the selection instead of doing it by hand. @aalkin @saganatt what do you think?

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamTrackSelection.h
```diff
@@ -0,0 +1,480 @@
...
+      break;
+  }
+}
+
+template <typename T>
+inline auto FemtoDreamTrackSelection::getNsigmaTOF(T const& track, o2::track::PID pid)
+{
+  /// skip tracks without TOF signal
+  /// \todo not sure what the error flags mean...
+  if (track.tofSignal() <= 0.f || std::abs(track.tofSignal() - 99998) < 0.01 || std::abs(track.tofSignal() - 99999) < 0.01) {
```
jgrosseo: why these comparisons with subtraction of 99998 and 99999?

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamTrackSelection.h
```diff
@@ -0,0 +1,480 @@
...
+        auto pidTPCVal = pidTPC.at(i);
+        auto pidTOFVal = pidTOF.at(i);
+        sel.checkSelectionSetBit(pidTPCVal, output, counter);
+        auto pidComb = std::sqrt(pidTPCVal * pidTPCVal + pidTOFVal * pidTOFVal);
+        sel.checkSelectionSetBit(pidComb, output, counter);
+      }
+
+    } else {
+      /// for the rest it's all the same
+      switch (selVariable) {
```
jgrosseo: Same here. I think all these cases will slow down processing a lot... We can profile this.

jgrosseo (commented):
Thanks for the extensive work. I have posted some suggestions and questions inline with the code.

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamTrackSelection.h
a-mathis: these appear to be some error flags from TOF PID. Maybe an isValid() function would be helpful here?

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
a-mathis: so something like this? I thought these filters were set up as one of the first things, I remember we had quite a lengthy discussion on this and for the same reason we also abandoned some of the rather nice features we had made up

Configurable<float> smallestSomething{ ... };
Filter filter = track::Something > smallestSomething;

init() {
  std::sort( .. vector .. );
  smallestSomething = vector.at(0);
..
}

aalkin: It does not need to be a configurable itself, it can simply be a data member that you sen in `init()` based on configurable vectors.

Analysis/Tasks/PWGCF/CMakeLists.txt
a-mathis: why? There is AliFemtoDream and o2femtodream. I like the nomenclature and want to keep it

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
a-mathis: but what would be the benefit of this? 
In the end we don't want to have several data sets, and imagine you want to look at Sigma0 -> Lambda + Gamma, for which you need electrons (conversions), protons and pions (Lambda) and unfortunately the electrons are in your other data set. Then you add another data set?
Same goes for Xi's (p, pi, pi), Omegas (p, pi, K), ...

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamObjectSelection.h
a-mathis: hmm, this is executed once per variable and run, so I don't think see why this should be a performance bottleneck

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
a-mathis: After all these columns here are just for debugging, i.e. we certainly won't write out these for a large fraction of the data. We optimise anyways?

aalkin: If the columns are "optional" it would make sense to have them in a separate table, joinable to the main one. And using types with the smallest required size is indeed important due to memory restrictions when running in the Grid.

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamTrackSelection.h
jgrosseo: Isn't the hasTOF() enough?
https://github.com/AliceO2Group/AliceO2/blob/dev/Framework/Core/include/Framework/AnalysisDataModel.h#L213

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamObjectSelection.h
jgrosseo: That is then fine, but you also go through the switch for each track elsewhere, don't you?

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
jgrosseo: Good point @aalkin , even better
Yes, @a-mathis , just the init() declaration is differently (you get also the InitContext, check corelations.cxx)

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
jgrosseo: How many species you think you put typically in each dataset?

Analysis/Tasks/PWGCF/CMakeLists.txt
jgrosseo: No class or library in O2 has an extra o2 in it. Please remove it. It will just be FemtoDream which sounds nice, doesn't it? :)

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
a-mathis: @jgrosseo ah yes of course, this was just a sketch.   void init(InitContext&) will do, right?

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
a-mathis: well, the five that we have here will be most relevant for us

a-mathis: in any case, I'm considering to have an extra field for the PID identifiers (could be 0 for v0 etc. and this should compress well, right?) Or we store the children ID of v0, cascade etc. in there

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamTrackSelection.h
a-mathis: in the case of 99998 and 99999 the TOF signal is clearly larger than 0, so I don't think this will be enough.

Analysis/Tasks/PWGCF/CMakeLists.txt
a-mathis: ok, but this is how we internally decided to call our framework, also to make the distinction to the existing FemtoDream in AliPhysics clear.

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
a-mathis: That was exactly the idea. In any case, we're now using the smaller data types

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamTrackSelection.h
a-mathis: yes if you have an idea how to improve this that would be very good, since this is one of the core pieces of the generic cut selection we set up (and that in principle could be used by anyone interested)

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
a-mathis: @aalkin @jgrosseo so now I tried running this, and it appears that only the values used to initialise that variable are used for the filter. So this appears not to be working, even though the minimum is properly found and set. Is there something conceptually wrong in my implementation?

aalkin: @a-mathis indeed, I missed the fact the you want to use it for the filter. For that it needs to be defined as a `Configurable`. The reasons is that filters are instantiated long before the configuration is available, and anything that is not a configurable will use the value it has at that point, while configurables will be updated later.

aalkin: Correction: right now `init()` is run after the filters are updated, so even with `Configurable` I will have to change that order to make your use case work. You can test this already by swapping two blocks in `AnalysisTask.h`, the one in lines 771-773 should be before the one in lines 755-769.

Analysis/Tasks/PWGCF/CMakeLists.txt
jgrosseo: You can still of course call it like that internally, but the directory should not include o2 (neither the class names, but I think they do not). Please change the directory name so that we can move ahead.

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDerived.h
```diff
@@ -0,0 +1,88 @@
...
+                  pidtpc_tiny::TPCNSigmaStoreEl,
+                  pidtpc_tiny::TPCNSigmaStorePi,
+                  pidtpc_tiny::TPCNSigmaStoreKa,
+                  pidtpc_tiny::TPCNSigmaStorePr,
+                  pidtpc_tiny::TPCNSigmaStoreDe,
+                  pidtof_tiny::TOFNSigmaStoreEl,
+                  pidtof_tiny::TOFNSigmaStorePi,
+                  pidtof_tiny::TOFNSigmaStoreKa,
+                  pidtof_tiny::TOFNSigmaStorePr,
+                  pidtof_tiny::TOFNSigmaStoreDe);
```
jgrosseo: I think you need the dynamic columns as well which expand the TPCNSigmaStoreEl to a usable value. I think you can simply reuse the definition, no copy paste of the code needed.

Analysis/Tasks/PWGCF/CMakeLists.txt
a-mathis: I don't understand this but whatever.

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
a-mathis: @aalkin Thanks a lot for the clarification. Is this something that is planned to be done soon? Then I will for now increase the initialisation of the variable and once the AnalysisTask.h is changed we'll adopt accordingly, ok?

jgrosseo: @aalkin will do a PR for this very soon, I assume. So you can already consider it done for your PR. Do you still need to adopt something?

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamCollisionSelection.h
ktf: The file in question is gone. I do see a few places still where there is complex code which should probably put out of line, though.

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
a-mathis: @jgrosseo @aalkin but actually how do I overwrite the value of a configurable? This doesn't compile
Thanks!

Configurable<float> TrackMinSelPtMin{"TrackMinSelPtMin", 0.4f, "(automatic) Minimal pT selection for tracks"};
void init(InitContext&) {
      TrackMinSelPtMin = ... something   
}

Analysis/Tasks/PWGCF/o2femtodream/include/o2femtodream/FemtoDreamCollisionSelection.h
a-mathis: OK, then I remove any inlining for now.

Analysis/Tasks/PWGCF/o2femtodream/femtoDreamProducerTask.cxx
aalkin: You also need to define it as a `MutableConfigurable`

jgrosseo (approved):

PR #6293, opened 2021-06-01
DPL: hide InputSpan from common headers

ktf:
InputSpan usage of std::function causes an avalanche of template
instantiantions, resulting (at least) in a rather large slowdown
of Framework/Core compilation. With this patch we go from

 Compilation (452 times):
   Parsing (frontend):         1686.8 s
   Codegen & opts (backend):   1120.7 s

to

 Compilation (452 times):
   Parsing (frontend):         1441.4 s
   Codegen & opts (backend):    842.8 s

### DIFF
--- a/Detectors/Raw/include/DetectorsRaw/SimpleSTF.h
+++ b/Detectors/Raw/include/DetectorsRaw/SimpleSTF.h
@@ -19,6 +19,7 @@
 #include <gsl/span>
 #include "Framework/InputRoute.h"
 #include "Framework/InputRecord.h"
+#include "Framework/InputSpan.h"
 
 namespace o2
 {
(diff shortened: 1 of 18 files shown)


ktf:
This needs https://github.com/AliceO2Group/QualityControl/pull/718. @Barthelemy could you do a new release with it?

ktf:
@TimoWilken @MichaelLettrich looks like one of the macs does not have one of the system packages. Could you check?

MichaelLettrich:
@ktf @TimoWilken : The application that cannot be found is part of the GTK (Gnome Toolkit) which is not available on macOS. It looks like some OS check fails or is missing and assumes  the script/application it's being run on Linux.

ktf:
I guess we installed it by hand before? I've added it to the system deps. Could you redeploy?

PR #6294, opened 2021-06-01
MCH: avoid using HB packets to compute digits time

aferrero2707:
This PR introduces a computation of the MCH Digits time is based on the orbits information in each Time Frame.

The SAMPA BC counter at the beginning of a given TF is estimated by counting the total number of orbits elapsed since the beginning of the run.

The computation assumes that the SAMPA BC counters are simultaneously reset via a dedicated trigger signal some time before the beginning of the run.

Two details in the code will need to be reviewed/corrected in the future:
* there is currently no way of reliably obtaining the number of orbits per time frame from the O2 framework, so for the moment this parameter can be specified from the command line and defaults to 128
* the Digit time computation requires a calibration constant representing the value of the SAMPA BC counters at the beginning of the first orbit in the run. This constant is currently implemented as a static data member the DataDecoder class, but it probably deserves a better place. It is also not yet clear whether this value will change over time, in which case a CDDB entry will be needed.

### DIFF
--- a/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
+++ b/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
@@ -155,8 +155,8 @@ class DataDecoder
   RawDigitVector mDigits;                               ///< vector of decoded digits
   std::unordered_set<OrbitInfo, OrbitInfoHash> mOrbits; ///< list of orbits in the processed buffer
 
-  std::optional<uint32_t> mFirstOrbitInRun{0}; ///< first orbit in the processed run
-  SampaTimeFrameStart mSampaTimeFrameStart;    ///< SAMPA bunch-crossing counter at the beiginning of the TF
+  std::optional<uint32_t> mFirstOrbitInRun; ///< first orbit in the processed run
+  SampaTimeFrameStart mSampaTimeFrameStart; ///< SAMPA bunch-crossing counter at the beiginning of the TF
 
   uint32_t mSampaTimeOffset{339986}; ///< SAMPA BC counter value at the beginning of the first orbit in the run
 
--- a/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
+++ b/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
@@ -95,10 +95,8 @@ class DataDecoder
     SampaTimeFrameStart(uint32_t orbit, uint32_t bunchCrossing) : mOrbit(orbit), mBunchCrossing(bunchCrossing) {}
 
     uint32_t mOrbit{0};
-    int32_t mBunchCrossing{0};
+    uint32_t mBunchCrossing{0};
   };
-
-  using SampaTimeFrameStarts = std::unordered_map<uint32_t, std::optional<SampaTimeFrameStart>>;
 
   struct RawDigit {
     o2::mch::Digit digit;
@@ -118,7 +116,7 @@ class DataDecoder
 
   using RawDigitVector = std::vector<RawDigit>;
 
-  DataDecoder(SampaChannelHandler channelHandler, RdhHandler rdhHandler, std::string mapCRUfile, std::string mapFECfile, bool ds2manu, bool verbose);
+  DataDecoder(SampaChannelHandler channelHandler, RdhHandler rdhHandler, uint32_t sampaBcOffset, std::string mapCRUfile, std::string mapFECfile, bool ds2manu, bool verbose);
 
   void reset();
   void decodeBuffer(gsl::span<const std::byte> buf);
@@ -126,9 +124,9 @@ class DataDecoder
   void setFirstOrbitInRun(uint32_t orbit) { mFirstOrbitInRun = orbit; }
   std::optional<uint32_t> getFirstOrbitInRun() { return mFirstOrbitInRun; }
   void setFirstOrbitInTF(uint32_t orbit);
-  std::optional<uint32_t> getFirstOrbitInTF() { return mFirstOrbitInTF; }
 
-  static uint32_t getSampaBcOffset() { return sSampaTimeOffset; }
+  void setSampaBcOffset(uint32_t offset) { mSampaTimeOffset = offset; }
+  uint32_t getSampaBcOffset() const { return mSampaTimeOffset; }
 
   static int32_t digitsTimeDiff(uint32_t orbit1, uint32_t bc1, uint32_t orbit2, uint32_t bc2);
   static void computeDigitsTime_(RawDigitVector& digits, SampaTimeFrameStart& sampaTimeFrameStart, bool debug);
@@ -158,10 +156,9 @@ class DataDecoder
   std::unordered_set<OrbitInfo, OrbitInfoHash> mOrbits; ///< list of orbits in the processed buffer
 
   std::optional<uint32_t> mFirstOrbitInRun{0}; ///< first orbit in the processed run
-  std::optional<uint32_t> mFirstOrbitInTF{0};  ///< first orbit in the processed time frame
   SampaTimeFrameStart mSampaTimeFrameStart;    ///< SAMPA bunch-crossing counter at the beiginning of the TF
 
-  static constexpr uint32_t sSampaTimeOffset = 339986; ///< SAMPA BC counter value at the beginning of the first orbit in the run
+  uint32_t mSampaTimeOffset{339986}; ///< SAMPA BC counter value at the beginning of the first orbit in the run
 
   SampaChannelHandler mChannelHandler;                  ///< optional user function to be called for each decoded SAMPA hit
   std::function<void(o2::header::RDHAny*)> mRdhHandler; ///< optional user function to be called for each RDH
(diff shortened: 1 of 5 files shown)


Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
```diff
@@ -147,19 +154,21 @@ class DataDecoder
...
 
-  RawDigitVector mDigits; ///< vector of decoded digits
-
+  RawDigitVector mDigits;                               ///< vector of decoded digits
   std::unordered_set<OrbitInfo, OrbitInfoHash> mOrbits; ///< list of orbits in the processed buffer
-  SampaTimeFrameStarts mSampaTimeFrameStarts;           ///< time stamps of the TimeFrames in the processed buffer
+
+  std::optional<uint32_t> mFirstOrbitInRun{0}; ///< first orbit in the processed run
+  std::optional<uint32_t> mFirstOrbitInTF{0};  ///< first orbit in the processed time frame
+  SampaTimeFrameStart mSampaTimeFrameStart;    ///< SAMPA bunch-crossing counter at the beiginning of the TF
```
pillot: you can remove "using SampaTimeFrameStarts..." at line 101, as it is not needed anymore.

Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
```diff
@@ -147,19 +154,21 @@ class DataDecoder
...
+
+  static constexpr uint32_t sSampaTimeOffset = 339986; ///< SAMPA BC counter value at the beginning of the first orbit in the run
```
pillot: You say this value might change over time. Can it change between the various test runs that you are currently performing or during the incoming global commissioning? If yes it should probably be settable from the command line, until we have a better solution (CCDB).

Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
```diff
@@ -132,6 +132,24 @@ static bool isValidDeID(int deId)
...
+    LOG(ERROR) << "[setFirstOrbitInTF] first orbit in run not set!";
+    return;
+  }
+
+  // the SAMPA BC value at the beginning of the TF is computed by counting the number of orbits
+  // since the beginning of the run, and then multiplying this number by the number of BC in one orbit.
+  // We then take the first 20 bits of the result to emulate the internal SAMPA BC counter.
+  mFirstOrbitInTF = orbit;
+  uint32_t nOrbitsInRun = mFirstOrbitInTF.value() - mFirstOrbitInRun.value();
+  uint32_t bc = (nOrbitsInRun * BCINORBIT + DataDecoder::sSampaTimeOffset) & TWENTYBITSATONE;
```
pillot: is uint32_t enough to hold an absolute number of BC? Or maybe it's fine to have a rollover at 32 bits since it is further restricted to 20 bits afterward anyway?

Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
```diff
@@ -363,28 +382,18 @@ void DataDecoder::computeDigitsTime_(RawDigitVector& digits, SampaTimeFrameStart
...
-      uint32_t bc = tfStart->mBunchCrossing;
-      uint32_t orbit = tfStart->mOrbit;
-      tfTime = DataDecoder::digitsTimeDiff(orbit, bc, info.orbit, info.getBXTime());
-      if (debug) {
-        std::cout << "\n[computeDigitsTime_] hit " << info.orbit << "," << info.getBXTime()
-                  << "    tfTime(1) " << orbit << "," << bc << "    diff " << tfTime << std::endl;
-      }
-      setDigitTime(d, tfTime);
-      info.tfTime = tfTime;
+    SampaTimeFrameStart& tfStart = sampaTimeFrameStart;
```
pillot: this is a detail but why not using directly sampaTimeFrameStart instead of a copy?

Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
```diff
@@ -91,6 +91,7 @@ class DataDecoder
   };
 
   struct SampaTimeFrameStart {
+    SampaTimeFrameStart() = default;
```
pillot, line 94: I missing that previously but shouldn't mBunchCrossing (at line 97) be a uint32_t instead of a int32_t (just for consistency as it probably does not matters since it is a 20 bits max int anyway).

Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
```diff
@@ -147,19 +154,21 @@ class DataDecoder
...
   o2::mch::raw::PageDecoder mDecoder; ///< CRU page decoder
 
-  RawDigitVector mDigits; ///< vector of decoded digits
-
+  RawDigitVector mDigits;                               ///< vector of decoded digits
   std::unordered_set<OrbitInfo, OrbitInfoHash> mOrbits; ///< list of orbits in the processed buffer
-  SampaTimeFrameStarts mSampaTimeFrameStarts;           ///< time stamps of the TimeFrames in the processed buffer
+
+  std::optional<uint32_t> mFirstOrbitInRun{0}; ///< first orbit in the processed run
+  std::optional<uint32_t> mFirstOrbitInTF{0};  ///< first orbit in the processed time frame
```
pillot: this data member mFirstOrbitInTF is not really needed since it is just copied into mSampaTimeFrameStart.mOrbit, which is the variable used later on. Beside, the std::optional is not really used since the validity of mSampaTimeFrameStart.mOrbit (or mFirstOrbitInTF) is never checked.

pillot (commented):
The new logic looks fine to me, I only have a few comments on the implementation below:

Detectors/MUON/MCH/Workflow/src/DataDecoderSpec.cxx
```diff
@@ -72,6 +75,7 @@ class DataDecoderTask
     mDebug = ic.options().get<bool>("debug");
     mCheckROFs = ic.options().get<bool>("check-rofs");
     mDummyROFs = ic.options().get<bool>("dummy-rofs");
+    mNHBPerTF = ic.options().get<int>("nhb-per-tf");
```
aphecetche: Can't you simply get this information from the HBFUtils ConfigurableParam (https://github.com/AliceO2Group/AliceO2/blob/d26801ce514a23b839e30e50f1a67aef6b46a38c/Detectors/Raw/include/DetectorsRaw/HBFUtils.h#L128) and then use the command line syntax `--configKeyValues "HBFUtils.nHBFPerTF=128` to set it ? (like what's done in the [digitization](https://github.com/AliceO2Group/AliceO2/blob/dev/doc/DetectorSimulation.md) for instance)  

Actually same remark/question for the first orbit ?

Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
aferrero2707: Ok.

Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
aferrero2707: Yes, good suggestion. The value should be constant across the test data we are currently collecting, but I still have to verify is that is really true.

Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
aferrero2707: Good point. I have modified the code in order to make sure we do not exceed the number of bits, at least not before we mask the lowest 20 bits.

Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
aferrero2707: Yes, you are right.

Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
aferrero2707: I agree, removed.

Detectors/MUON/MCH/Workflow/src/DataDecoderSpec.cxx
aferrero2707: Yes, good point, I was actually unsure of how to set this from the command line...

However, for the first orbit in the run we cannot rely on the command line, since it can change from one run to another in a random way.

Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
aferrero2707: Yes, indeed.

aferrero2707:
The last commit introduces the possibility of setting the number of orbits per time frame via the standard configuration option `--configKeyValues "HBFUtils.nHBFPerTF=256"` or via configuration file.

aphecetche (approved):

aferrero2707:
@ktf we are having again CI failures that are un-related to this PR... how do you suggest to proceed?

ktf:
I merged it taking your word for "the issues are unrelated". I think @sawenzel is aware of the issues on with o2Sim and working on fixing them.

PR #6295, opened 2021-06-01
Update ALICE 3 beam pipe default length

mconcas:
As per request the definition of BP length is updated. @qgp


mconcas:
Closing in favour of a broader pr

PR #6296, opened 2021-06-01
Updates for dimuon and Electron-Muon analysis

vfeuilla:
TableMakerMuon and TableMakerMuon_pp are simplified versions of TableMaker without configurables to be used on hyperloop in dimuon analysis in PbPb and pp respectively
Update of the dileptonMuMu.cxx to produce dimuon histograms
Moifying the range of some histograms for dimuon analysis
Adding basic cuts on muons in the CutsLibrary.h

Adding a loop on the TableReader to create Electron-Muon pairs


Analysis/Tasks/PWGDQ/dileptonMuMu.cxx
```diff
@@ -122,17 +145,29 @@ struct DQEventSelection {
...
+
+    fHistMan->FillHistClass("EventMixing_BeforeCuts", fValues);
+    if (fEventMixingCut->IsSelected(fValues)) {
+      fHistMan->FillHistClass("EventMixing_AfterCuts", fValues);
+      eventMixingSel(1);
+    } else {
+      eventMixingSel(0);
+    }
+    int hh = getHash(fValues[VarManager::kCentVZERO]);
+    hash(hh);
```
iarsene, line 157: Its not clear why you need to produce the additional EventMixingCuts table. This task is producing already a table with an event selection decision and normally i expect that event mixing is done with the same events as those selected for the same event analysis. Also, if the event does not fall into any event categories, the mixing has will be -1. Is this not sufficient ?

Analysis/Tasks/PWGDQ/dileptonMuMu.cxx
```diff
@@ -144,72 +179,51 @@ struct DQMuonTrackSelection {
...
-    kineMuonCut310.AddCut(VarManager::kEta, -4.0, -2.5);
-    kineMuonCut310.AddCut(VarManager::kMuonChi2, 0.0, 1e6);
-    // The value of sigma_PDCA depends on the postion w.r.t the absorber. For 17.6 < RAbs < 26.5cm, sigma_PDCA = 99 cmxGeV/c. For 26.5 < RAbs < 89.5 cm, sigma_PDCA = 54
-    // temporarily not applied
-    //    kineMuonCut310.AddCut(VarManager::kMuonRAtAbsorberEnd, 26.5, 89.5);
-    //    kineMuonCut310.AddCut(VarManager::kMuonPDca, 0.0, 324); // Cut is pDCA < 6*sigma_PDCA
-    fMuonCut310->AddCut(&kineMuonCut310);
+    fTrackCut = new AnalysisCompositeCut(true);
+    AnalysisCut kineMuonCut;
+    kineMuonCut.AddCut(VarManager::kPt, fConfigMuonPtLow, 100.0);
```
iarsene: This cut can be actually applied as a Filter, unless you want to check also tracks which do not pass this cut.

Analysis/Tasks/PWGDQ/dileptonMuMu.cxx
```diff
@@ -219,160 +233,169 @@ struct DileptonMuMu {
...
+  HistogramManager* fHistMan;
+  //NOTE: one could define also a dilepton cut, but for now basic selections can be supported using Partition
+
+  float* fValues;
+  // NOTE: The track filter produced by the barrel track selection contain a number of electron cut decisions and one last cut for hadrons used in the
+  //           dilepton - hadron task downstream. So the bit mask is required to select pairs just based on the electron cuts
+  uint8_t fTwoTrackFilterMask = 0;
+  std::vector<TString> fCutNames;
+
+  Filter filterEventSelected = aod::reducedevent::isEventSelected == 1;
```
iarsene, line 314: Here you use a different event selection compared to the mixing task

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -423,8 +423,33 @@ struct DQEventMixing {
           }
         } // end for (muon2)
       }   // end for (muon1)
-    }     // end for (event combinations)
-  }       // end process()
+
+      for (auto& track1 : tracks1) {
+        for (auto& muon2 : muons2) {
+          twoTrackFilter = track1.isBarrelSelected() & muon2.isMuonSelected();
```
iarsene: Here you have to be careful, since the electrons and muons are selected with different cuts, and possibly with a different number of cuts. At the moment at least, there is one single cut on the muons and possibly multiple ones on the electrons. So we must agree on a policy on how to combine these cuts. If we stay with just a single cut on the muon, then the bitwise AND should be changed here to a normal logical AND. Then, later when you fill electron-muon pair histograms, if you want to fill histograms separately for each electron cut, then you have to adapt the code. You should do it as in the electron-electron loop, where you must take into account the electron cuts fulfilled.

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -423,8 +423,33 @@ struct DQEventMixing {
...
+
+          if (!twoTrackFilter) { // the tracks must have at least one filter bit in common to continue
+            continue;
+          }
+          VarManager::FillPair(track1, muon2, fValues, VarManager::kElectronMuon);
+
+          for (int i = 0; i < fCutNames.size(); ++i) {
+            if (twoTrackFilter & (uint8_t(1) << i)) {
```
iarsene, line 437: If you have just one cut on the muon, and multiple ones on the electrons, than you need to make sure that twoTrackFilter is the electron filter map.

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -423,8 +423,33 @@ struct DQEventMixing {
...
+              if (track1.sign() * muon2.sign() < 0) {
+                fHistMan->FillHistClass(Form("PairsEleMuMEPM", fCutNames[i].Data()), fValues);
```
iarsene: here you need to add the electron cut name into the name of the electron-muon pair hist class

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -423,8 +423,33 @@ struct DQEventMixing {
...
+              } else {
+                if (track1.sign() > 0) {
+                  fHistMan->FillHistClass(Form("PairsEleMuMEPP", fCutNames[i].Data()), fValues);
```
iarsene: same as above

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -423,8 +423,33 @@ struct DQEventMixing {
...
+                } else {
+                  fHistMan->FillHistClass(Form("PairsEleMuMEMM", fCutNames[i].Data()), fValues);
```
iarsene: same as above

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -316,6 +315,7 @@ struct DQEventMixing {
       }
     }
     histNames += "PairsMuonMEPM;PairsMuonMEPP;PairsMuonMEMM;";
+    histNames += "PairsEleMuMEPM;PairsEleMuMEPP;PairsEleMuMEMM;";
```
iarsene: see also the comments i made lower in the code. Since multiple electron cuts are possible, then, when you combine with muons (currently selected with just one cut) you should have one Pairs histogram directory for each electron cut. You can define these histNames just above, in the loop over electron cuts.

iarsene (commented):
Besides the comments i placed through the code, i think you do not need the tableMakerMuon_pp.cxx since this completely overlaps with the tableMaker.cxx with the default option "--isPbPb false".

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -465,6 +490,7 @@ struct DQTableReader {
       }
     }
     histNames += "PairsMuonSEPM;PairsMuonSEPP;PairsMuonSEMM;";
+    histNames += "PairsEleMuSEPM;PairsEleMuSEPP;PairsEleMuSEMM;";
```
iarsene: here you have to add one pairs hist class for each electron cut, since multiple cuts are defined. You could just add this a few lines above in the electron cuts loop

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -531,6 +557,28 @@ struct DQTableReader {
         }
       }
     } // end loop over muon track pairs
+
+    for (auto& [trackBarrel, trackMuon] : combinations(tracks, muons)) {
+      twoTrackFilter = trackBarrel.isBarrelSelected() & trackMuon.isMuonSelected();
```
iarsene: here you ave just one cut for the muon and multiple ones for the electrons. So i have the same comment as in the event mixing task. The bitwise AND has to be changed to a normal AND.

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -531,6 +557,28 @@ struct DQTableReader {
...
+      if (!twoTrackFilter) { // the muons must have at least one filter bit in common to continue
+        continue;
+      }
+      // NOTE: The dimuons in this task ae pushed in the same table as the dielectrons.
+      //       In order to discriminate them, the dileptonFilterMap uses the first 8 bits for dielectrons and the last 8 for dimuons.
+      // TBD:  Other implementations may be possible, for example add a column to the dilepton table to specify the pair type (dielectron, dimuon, electron-muon, etc.)
+      dileptonFilterMap = uint16_t(twoTrackFilter) << 8;
```
iarsene: Make sure that the twoTrackFilter is basically the electron filter bit map

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -531,6 +557,28 @@ struct DQTableReader {
...
+      VarManager::FillPair(trackBarrel, trackMuon, fValues, VarManager::kElectronMuon);
+      dileptonList(event, fValues[VarManager::kMass], fValues[VarManager::kPt], fValues[VarManager::kEta], fValues[VarManager::kPhi], trackBarrel.sign() + trackMuon.sign(), dileptonFilterMap);
+      if (trackBarrel.sign() * trackMuon.sign() < 0) {
+        fHistMan->FillHistClass("PairsEleMuSEPM", fValues);
+      } else {
+        if (trackBarrel.sign() > 0) {
+          fHistMan->FillHistClass("PairsEleMuSEPP", fValues);
+        } else {
+          fHistMan->FillHistClass("PairsEleMuSEMM", fValues);
```
iarsene: Here you must have a loop over the electron cuts and fill the histogram classes accordingly. See how its done for the electron-electron pairs

Analysis/Tasks/PWGDQ/dileptonMuMu.cxx
vfeuilla: I want a different trigger for the event mixing wrt the same event. In dimuon analysis, the typical trigger (UnlikeDimuonLowPt) selects muon that have a non negligible probability to come from a J/Psi, therefore it induces a bias in the event mixing. Which is why we use a different trigger

Analysis/Tasks/PWGDQ/dileptonMuMu.cxx
vfeuilla: Good suggestion!

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -515,11 +541,11 @@ struct DQTableReader {
...
         continue;
       }
-      // NOTE: The dimuons in this task ae pushed in the same table as the dielectrons.
-      //       In order to discriminate them, the dileptonFilterMap uses the first 8 bits for dielectrons and the last 8 for dimuons.
+      // NOTE: The dimuons and electron-muon pairs in this task are pushed in the same table as the dielectrons.
+      //       In order to discriminate them, the dileptonFilterMap uses the first 8 bits for dielectrons, the next 8 for dimuons and the rest for electron-muon
       // TBD:  Other implementations may be possible, for example add a column to the dilepton table to specify the pair type (dielectron, dimuon, electron-muon, etc.)
-      dileptonFilterMap = uint16_t(twoTrackFilter) << 8;
-      VarManager::FillPair(muon1, muon2, fValues);
+      dileptonFilterMap = uint32_t(twoTrackFilter) << 8;
```
iarsene, line 547: The filter map is an uint16_t in the data model, not an uint32_t, so you have to change also in the data model

Analysis/Tasks/PWGDQ/tableReader.cxx
```diff
@@ -531,6 +557,31 @@ struct DQTableReader {
         }
       }
     } // end loop over muon track pairs
+
+    for (auto& [trackBarrel, trackMuon] : combinations(tracks, muons)) {
+      twoTrackFilter = trackBarrel.isBarrelSelected() & trackMuon.isMuonSelected() & fTwoTrackFilterMask;
```
iarsene: The selection filter bits for electrons are obviously different from those on muons. In particular, at the moment the muons selection task is applying just a single cut, while for the electrons multiple ones are possible. So when you do a bitwise AND you only request that the muon cut is fulfilled together with the first electron cut. All the rest of electron cuts are rejected.
There are many ways to combine electron and muon cuts, so one has to think of a strategy if you will want at some point to combine different electron and muon selections. Within the current model of this "example workflow", as i suggested in the previous comment you should combine the muon single cut with all the requested electron cuts. For this, the twoTrackFilter should be computed like this:
twoTrackFilter = (trackBarrel.isBarrelSelected() & fTwoTrackFilterMask) && trackMuon.isMuonSelected();

Analysis/Tasks/PWGDQ/tableReader.cxx
vfeuilla: Indeed, thanks for catching that.

Analysis/Tasks/PWGDQ/tableReader.cxx
vfeuilla: Oh, I get it! Sorry, I misunderstood what you were saying before.

iarsene (approved):

iarsene:
The macOS test fail is unrelated, merging

PR #6297, opened 2021-06-01
Fix in SVertices attachment to RecoContainer

shahor02:
@nburmaso thanks for spotting.


shahor02:
all CI's fail for unrelated reasons, merging.

PR #6298, opened 2021-06-01
Fix in requesting FT0 data for PVertex validation

shahor02:


shahor02:
failures unrelated, merging.

PR #6299, opened 2021-06-02
o2-sim: Improved signal handling

sawenzel:
This should improve the shutdown phase
of the distributed simulation system upon errors:

*) now terminate using process groups in
   order to prevent leaking child processes
*) improved signal broadcasting in the Geant transport
   processes (within the process group)

### DIFF
--- a/run/O2SimDeviceRunner.cxx
+++ b/run/O2SimDeviceRunner.cxx
@@ -35,20 +35,19 @@ namespace bpo = boost::program_options;
 std::vector<int> gChildProcesses; // global vector of child pids
 int gMasterProcess = -1;
 
-// custom signal handler to ensure that we
-// distribute signals to all detached forks
-void sighandler(int signal)
+void sigaction_handler(int signal, siginfo_t* signal_info, void*)
 {
-  if (signal == SIGTERM || signal == SIGINT) {
-    auto pid = getpid();
-    if (pid == gMasterProcess) {
-      // the master
-      for (auto child : gChildProcesses) {
-        kill(child, signal);
-      }
-    }
-    _exit(0);
+  auto pid = getpid();
+  LOG(INFO) << pid << " caught signal from source " << signal_info->si_pid;
+  auto groupid = getpgrp();
+  if (pid == gMasterProcess) {
+    killpg(pid, signal); // master kills whole process group
+  } else {
+    // forward to master
+    kill(groupid, signal);
   }
+  _exit(1); // we treat signal interruption as an error
+            // because only ordinary termination is good in the context of the distributed system
 }
 
 void addCustomOptions(bpo::options_description& options)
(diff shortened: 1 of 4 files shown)


run/O2SimDeviceRunner.cxx
```diff
@@ -35,20 +35,19 @@ namespace bpo = boost::program_options;
...
+  auto pid = getpid();
+  LOG(INFO) << pid << " caught signal from source " << signal_info->si_pid;
+  auto groupid = getpgrp();
+  if (pid == gMasterProcess) {
+    killpg(pid, signal); // master kills whole process group
+  } else {
+    // forward to master
+    kill(groupid, signal);
   }
+  _exit(1); // we treat signal interruption as an error
```
davidrohr, line 49: I am not 100% sure, but are we sure the only signal we will receive is KILL and TERM? Perhaps we should have a filter for the singal in here?

sawenzel: Why? The filter is specified further below in the main function.

davidrohr: ok, sorry, I missed that. Then it is OK. Perhaps we should also forward `SIGQUIT` and `SIGKILL`? Or probably SIGKILL cannot be handled?

sawenzel: SIGKILL can't be handled.

davidrohr: ok, I would put that in a comment to avoid confusion, and perhaps I'd add SIGQUIT just in case it is used at some time to avoid misleading behavior.

sawenzel:
@MichaelLettrich @TimoWilken : MacOS has a serious problem.

TimoWilken:
@sawenzel thanks, looking into it.

sawenzel:
@ktf: There are still compiler crashes despite the Vc improvement. I think the reason is something else and we rather need to rethink our way to build / the hardware on which we build:

```/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6299/0/Analysis/Tasks/PWGHF/taskLc.cxx
c++: fatal error: Killed signal terminated program cc1plus```

**Comment by davidrohr** [comment]
@ironMann : The FullCI fails in datadistribution probably due to the o2::byte / std::byte change:
```
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/DataDistribution/v0.10.16/v0.10.16/src/common/SubTimeFrameFileReader.cxx: In member function 'o2::header::Stack o2::DataDistribution::SubTimeFrameFileReader::getHeaderStack(std::size_t&)':
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/DataDistribution/v0.10.16/v0.10.16/src/common/SubTimeFrameFileReader.cxx:121:30: error: cannot convert 'unsigned char*' to 'std::byte*' in initialization
```
Could you have a look? (This ran yesterday evening)

ktf:
I am on it, but I still have one error when compiling it on my box.

ktf:
@MichaelLettrich I think you need to deploy one more mac with the updated system deps.

MichaelLettrich:
@ktf  just finished deploying on all macs- will take a while till the effects will show due to rebuilds.

PR #6300, opened 2021-06-02
Add back cerrno

ktf:


ktf:
Currently broken on mac.

PR #6301, opened 2021-06-02
DPL Analysis: fix AnalysisTask test after renaming data model table

aalkin:


ktf:
Currently broken. Merging.

PR #6302, opened 2021-06-02
DPL Analysis: avoid instantiating HistFactory functions for each TU

mario-krueger:
- move histogram creation code to HistogramSpec.cxx
- instantiate histogram creation templates only once

### DIFF
--- a/Analysis/Core/src/CorrelationContainer.cxx
+++ b/Analysis/Core/src/CorrelationContainer.cxx
@@ -26,7 +26,7 @@
 #include "TCanvas.h"
 #include "TF1.h"
 #include "THn.h"
-#include "Framework/HistogramRegistry.h"
+#include "Framework/HistogramSpec.h"
 
 using namespace o2::framework;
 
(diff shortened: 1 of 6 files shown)


mario-krueger:
This is only a first cleanup step to improve compile time and memory related to the HistogramRegistry.
More changes will be needed that directly affect the registry code, but I'd prefer to do this incrementally.

ktf:
Thank you for this. You can check how long a given file takes to compile by using:

```bash
scripts/profile-compilation ./Analysis/Tasks/PWGHF/taskXic.cxx
```

There is also a script which is called `scripts/profile-compilation-merge` which allows you to combine more then one compilation unit. You can look at the results with <https://speedscope.app>.

That said when I try to do a full compilation with clang and `-ftime-trace`, using https://github.com/aras-p/ClangBuildAnalyzer to do a full report, I still see roughly the same compilation time.

mario-krueger:
Thanks for the hints! That will be very helpful.
You are right, I also just realised these changes will not have much effect on the compilation since the templated function (createHist<>) which I extern'ed here was instanciated only in the HistogramRegistry.cxx itself and in the CorrelationContainer, but never in the user tasks. However, I still think these are useful changes if in the future people want to create histograms in the same manner as Jan Fiete does it in the CorrelationContainer.
The next modifications to the registry will probably have a larger impact.

ktf:
Yes, agreed.

ktf:
Errors unrelated. macOS tested on my own box.

PR #6303, opened 2021-06-02
Materials inside the MFT patch panel

frmanso:
Addition of copper plates inside the patch panel to simulate cables, pipes and electronic cards materials. The readout cables (defined in the HalfCone class) are a bit longer to improve the connection with the materials places inside the patch panel.


bovulpes (approved):

rpezzi:
This should be good to go. Anyone to merge? Thanks.

PR #6304, opened 2021-06-02
ITS tracking will produce output with reconstructed IRFrames (IR brackets)

shahor02:
* IRFrame: class to hold a bracket of InteractionRecords

* ITS tracking will produce output with reconstructed IRFrames (IR brackets)
The `vector<IRFrame>` will contain the 1st and the last IRs of the ITS strobe which was reconstructed (in the sync. workflow conditioned by mult. selection) and produced at least 1 track.

* Attach IRFrame to RecoContainer and InputHelper
````
Usage:
dataRequest->requestIRFramesITS(); // in the spec which needs it
o2::globaltracking::InputHelper::addInputSpecsIRFramesITS(configcontext, specs); // in the workflow building

alternatively, one can add the reader directly:
#include "ITSWorkflow/IRFrameReaderSpec.h"
 ...
specs.emplace_back(o2::its::getIRFrameReaderSpec());
recoContainer.getIRFramesITS() to access the span of IRFrame objects.
````

@martenole @davidrohr here is it. I wanted to attach it to the TRDtracker but I see that at the moment it does not use the InputHelper. Are you planning to?

### DIFF
--- a/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
+++ b/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
@@ -92,6 +92,7 @@ class VtxTrackRef;
 class V0;
 class Cascade;
 class TrackCosmics;
+class IRFrame;
 } // namespace o2::dataformats
 
 namespace o2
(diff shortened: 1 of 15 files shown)


shahor02:
Hi @sawenzel , the simulation (pythia8hi, after successful completion of the QED) failed in the FullCO with the error which I've never seen before:
````
[INFO] Waiting for configuration answer 
[ERROR] No configuration received within 4000ms
[ERROR] Could not initialize simulation
````
Any idea what could be the reason?

Merging since tested locally.

PR #6305, opened 2021-06-02
Time-dependent TPC dEdx calibration

badarots:
A first attempt to write the necessary code to track the MIP position variation over time.

### DIFF
--- a/Detectors/TPC/calibration/CMakeLists.txt
+++ b/Detectors/TPC/calibration/CMakeLists.txt
@@ -32,9 +32,13 @@ o2_add_library(TPCCalibration
                        src/IDCFourierTransform.cxx
                        src/RobustAverage.cxx
                        src/IDCCCDBHelper.cxx
-               PUBLIC_LINK_LIBRARIES O2::DataFormatsTPC O2::TPCBase O2::DetectorsCalibration
+                       src/dEdxHistos.cxx
+                       src/CalibdEdx.cxx
+               PUBLIC_LINK_LIBRARIES O2::DataFormatsTPC O2::TPCBase
                                      O2::TPCReconstruction ROOT::Minuit
-                                     Microsoft.GSL::GSL)
+                                     Microsoft.GSL::GSL
+                                     O2::DetectorsCalibration
+                                     O2::CCDB)
 
 o2_target_root_dictionary(TPCCalibration
                           HEADERS include/TPCCalibration/CalibRawBase.h
@@ -58,7 +62,9 @@ o2_target_root_dictionary(TPCCalibration
                                   include/TPCCalibration/IDCContainer.h
                                   include/TPCCalibration/RobustAverage.h
                                   include/TPCCalibration/IDCFourierTransform.h
-                                  include/TPCCalibration/IDCCCDBHelper.h)
+                                  include/TPCCalibration/IDCCCDBHelper.h
+                                  include/TPCCalibration/dEdxHistos.h
+                                  include/TPCCalibration/CalibdEdx.h)
 
 o2_add_test_root_macro(macro/comparePedestalsAndNoise.C
                        PUBLIC_LINK_LIBRARIES O2::TPCBase
--- a/Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
+++ b/Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
@@ -1,0 +1,116 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file CalibdEdx.h
+/// \brief This file provides the time dependent dE/dx calibrator, based on the MIP position.
+/// \author Thiago Badaró <thiago.saramela@usp.br>
+
+#ifndef ALICEO2_TPC_CALIBDEDX_H_
+#define ALICEO2_TPC_CALIBDEDX_H_
+
+#include <array>
+#include <cstdint>
+#include <string_view>
+
+// o2 includes
+#include "DataFormatsTPC/TrackTPC.h"
+#include "DataFormatsTPC/TrackCuts.h"
+#include "CCDB/CcdbObjectInfo.h"
+#include "DetectorsCalibration/TimeSlotCalibration.h"
+#include "DetectorsCalibration/TimeSlot.h"
+#include "TPCCalibration/dEdxHistos.h"
+#include "TPCCalibration/FastHisto.h"
+#include "CommonUtils/TreeStreamRedirector.h"
+
+namespace o2::tpc
+{
+
+/// Container used to store the dE/dx calibration output
+struct CalibMIP {
+  double ASide;
+  double CSide;
+};
+
+/// dE/dx calibrator class
+class CalibdEdx final : public o2::calibration::TimeSlotCalibration<o2::tpc::TrackTPC, o2::tpc::dEdxHistos>
+{
+  using TFType = o2::calibration::TFType;
+  using Slot = o2::calibration::TimeSlot<dEdxHistos>;
+  using CcdbObjectInfoVector = std::vector<o2::ccdb::CcdbObjectInfo>;
+  using MIPVector = std::vector<CalibMIP>;
+
+ public:
+  /// Contructor that enables track cuts
+  CalibdEdx(int nBins = 200, int minEntries = 100, float minP = 0.4, float maxP = 0.6, int minClusters = 60)
+    : mNBins{nBins}, mMinEntries{minEntries}, mCuts{minP, maxP, static_cast<float>(minClusters)}
+  {
+  }
+
+  /// Destructor
+  ~CalibdEdx() final = default;
+
+  /// \return if there are enough data to compute the calibration
+  bool hasEnoughData(const Slot& slot) const final
+  {
+    return slot.getContainer()->getASideEntries() >= mMinEntries && slot.getContainer()->getCSideEntries() >= mMinEntries;
+  }
+
+  /// Empty the output vectors
+  void initOutput() final;
+
+  /// Process time slot data and compute its calibration
+  void finalizeSlot(Slot& slot) final;
+
+  /// Creates new time slot
+  Slot& emplaceNewSlot(bool front, TFType tstart, TFType tend) final;
+
+  void setApplyCuts(bool apply) { mApplyCuts = apply; }
+  bool getApplyCuts() { return mApplyCuts; }
+  void setCuts(const TrackCuts& cuts) { mCuts = cuts; }
+
+  /// \return the computed calibrations
+  const MIPVector& getMIPVector() const { return mMIPVector; }
+
+  /// \return CCDB output informations
+  const CcdbObjectInfoVector& getInfoVector() const { return mInfoVector; }
+
+  /// Non const version
+  /// \return CCDB output informations
+  CcdbObjectInfoVector& getInfoVector() { return mInfoVector; }
+
+  /// Enable debug output to file of the time slots calibrations outputs and dE/dx histograms
+  void enableDebugOutput(std::string_view fileName);
+
+  /// Disable debug output to file. Also writes and closes stored time slots.
+  void disableDebugOutput();
+
+  /// \return if debug output is enabled
+  bool hasDebugOutput() const { return static_cast<bool>(mDebugOutputStreamer); }
+
+  /// Write debug output to file
+  void finalizeDebugOutput() const;
+
+ private:
+  int mNBins{};          ///< Number of bins in each time slot histogram
+  int mMinEntries{};     ///< Minimum amount of tracks in each time slot, to get enough statics
+  bool mApplyCuts{true}; ///< Flag to enable tracks cuts
+  TrackCuts mCuts;       ///< Cut object
+
+  CcdbObjectInfoVector mInfoVector; ///< vector of CCDB Infos, each element is filled with the CCDB description of the accompanying MIP positions
+  MIPVector mMIPVector;             ///< vector of MIP positions, each element is filled in "process" when we finalize one slot (multiple can be finalized during the same "process", which is why we have a vector. Each element is to be considered the output of the device, and will go to the CCDB
+
+  std::unique_ptr<o2::utils::TreeStreamRedirector> mDebugOutputStreamer; ///< Debug output streamer
+
+  ClassDefOverride(CalibdEdx, 1);
+};
+
+} // namespace o2::tpc
+#endif
--- a/Detectors/TPC/calibration/include/TPCCalibration/FastHisto.h
+++ b/Detectors/TPC/calibration/include/TPCCalibration/FastHisto.h
@@ -22,6 +22,7 @@
 #include "Framework/Logger.h"
 #include <sstream>
 #include <vector>
+#include <iomanip>
 
 namespace o2
 {
--- a/Detectors/TPC/calibration/include/TPCCalibration/dEdxHistos.h
+++ b/Detectors/TPC/calibration/include/TPCCalibration/dEdxHistos.h
@@ -1,0 +1,85 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file dEdxHistos.h
+/// \brief This file provides the container used for time based dE/dx calibration.
+/// \author Thiago Badaró <thiago.saramela@usp.br>
+
+#ifndef ALICEO2_TPC_DEDXHISTOS_H_
+#define ALICEO2_TPC_DEDXHISTOS_H_
+
+#include <array>
+#include <gsl/span>
+#include <string_view>
+
+// o2 includes
+#include "TPCCalibration/FastHisto.h"
+#include "DataFormatsTPC/TrackCuts.h"
+
+namespace o2::tpc
+{
+
+// forward declaration
+class TrackTPC;
+
+/// Class that creates dE/dx histograms from a sequence of tracks objects
+class dEdxHistos
+{
+  using Hist = FastHisto<float>;
+
+ public:
+  /// Default constructor
+  dEdxHistos() = default;
+
+  /// Constructor that enable tracks cuts
+  dEdxHistos(unsigned int nBins, const TrackCuts& cuts);
+
+  /// Constructor that enable tracks cuts, and creates a TrackCuts internally
+  dEdxHistos(unsigned int nBins, float minP = 0.4, float maxP = 0.6, int minClusters = 60)
+    : dEdxHistos(nBins, {minP, maxP, static_cast<float>(minClusters)}) {}
+
+  /// Fill histograms using tracks data
+  void fill(const gsl::span<const TrackTPC> tracks);
+
+  /// Add counts from other container
+  void merge(const dEdxHistos* other);
+
+  /// Print the number of entries in each histogram
+  void print() const;
+
+  void setApplyCuts(bool apply) { mApplyCuts = apply; }
+  bool getApplyCuts() { return mApplyCuts; }
+  void setCuts(const TrackCuts& cuts) { mCuts = cuts; }
+
+  /// \return number of entries in the A side histogram
+  double getASideEntries() const { return mEntries[0]; }
+
+  /// \return number of entries in the C side histogram
+  double getCSideEntries() const { return mEntries[1]; }
+
+  /// Get the underlying histograms
+  const std::array<Hist, 2>& getHists() const { return mHist; }
+
+  /// Save the histograms to a file
+  void dumpToFile(std::string_view fileName) const;
+
+ private:
+  bool mApplyCuts{true}; ///< Whether or not to apply tracks cuts
+  TrackCuts mCuts;       ///< Cut class
+
+  std::array<float, 2> mEntries{0, 0}; ///< Number of entries in each histogram
+  std::array<Hist, 2> mHist;           ///< MIP position histograms, for TPC's A and C sides
+
+  ClassDefNV(dEdxHistos, 1);
+};
+
+} // namespace o2::tpc
+#endif
--- a/Detectors/TPC/calibration/src/CalibdEdx.cxx
+++ b/Detectors/TPC/calibration/src/CalibdEdx.cxx
@@ -1,0 +1,101 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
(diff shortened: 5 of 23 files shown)


Detectors/TPC/calibration/CMakeLists.txt
```diff
@@ -20,9 +20,14 @@ o2_add_library(TPCCalibration
                        src/DigitDump.cxx
                        src/DigitDumpParam.cxx
                        src/CalibPadGainTracks.cxx
+                       src/CalibdEdx.cxx
                PUBLIC_LINK_LIBRARIES O2::DataFormatsTPC O2::TPCBase
                                      O2::TPCReconstruction ROOT::Minuit
-                                     Microsoft.GSL::GSL)
+                                     Microsoft.GSL::GSL
+                                     O2::DetectorsCalibration
```
badarots, line 40: I have to check which library I really need to add here, I think a can remove some of those, like O2::DetectorsDCS.

badarots: Removed

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -0,0 +1,108 @@
...
+ private:
+  int mMinClusters;
+  double mMinP, mMaxP;
+  std::array<float, 2> mHistEntries{0, 0};
+  std::array<Hist, 2> mHist;
+
+  ClassDefNV(CalibdEdx, 1);
+};
+
+class CalibratordEdx final : public o2::calibration::TimeSlotCalibration<o2::tpc::TrackTPC, o2::tpc::CalibdEdx>
```
wiechula: I prefer to put the `Calibrator` in a separate file.

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -0,0 +1,108 @@
...
+
+  double getASideEntries() const { return mHistEntries[0]; }
+  double getCSideEntries() const { return mHistEntries[1]; }
+
+  void dumpToFile(const std::string& file_name) const;
+  const std::array<Hist, 2>& getHist() const { return mHist; }
+
+ private:
+  int mMinClusters;
+  double mMinP, mMaxP;
```
wiechula: Default initialize, separate lines and comments:
```suggestion
  int mMinClusters{}; ///< bla
  double mMinP{}; ///< bla
  double mMaxP{}; ///< bla
```

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -0,0 +1,108 @@
...
+  const MIPVector& getMIPVector() const { return mMIPVector; }
+  const CcdbObjectInfoVector& getInfoVector() const { return mInfoVector; }
+  CcdbObjectInfoVector& getInfoVector() { return mInfoVector; }
+  void dumpToFile(const std::string& file_name) const;
+
+ private:
+  int mMinEntries;
+  int mMinClusters;
+  double mMinP, mMaxP;
+  int mNBins;
```
wiechula: see above, deafault initialize, doxygen comments, one variable per line.

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -0,0 +1,108 @@
...
+  CcdbObjectInfoVector mInfoVector; // vector of CCDB Infos , each element is filled with the CCDB description of the accompanying LHCPhase
+  MIPVector mMIPVector;             // vector of LhcPhase, each element is filled in "process" when we finalize one slot (multiple can be finalized during the same "process", which is why we have a vector. Each element is to be considered the output of the device, and will go to the CCDB
```
wiechula: doxygen comments
```suggestion
  CcdbObjectInfoVector mInfoVector; ///< vector of CCDB Infos , each element is filled with the CCDB description of the accompanying LHCPhase
  MIPVector mMIPVector;             ///< vector of LhcPhase, each element is filled in "process" when we finalize one slot (multiple can be finalized during the same "process", which is why we have a vector. Each element is to be considered the output of the device, and will go to the CCDB
```

Detectors/TPC/calibration/include/TPCCalibration/FastHisto.h
```diff
@@ -21,6 +21,7 @@
 #include "Framework/Logger.h"
 #include <sstream>
 #include <vector>
+#include <iomanip>
```
wiechula, line 25: Why is this needed?

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -0,0 +1,108 @@
...
+  double getCSideEntries() const { return mHistEntries[1]; }
+
+  void dumpToFile(const std::string& file_name) const;
+  const std::array<Hist, 2>& getHist() const { return mHist; }
+
+ private:
+  int mMinClusters;
+  double mMinP, mMaxP;
+  std::array<float, 2> mHistEntries{0, 0};
+  std::array<Hist, 2> mHist;
```
wiechula: doxygen comments

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -0,0 +1,108 @@
...
+
+#ifndef ALICEO2_TPC_CALIBDEDX_H_
+#define ALICEO2_TPC_CALIBDEDX_H_
+
+#include <string>
+#include <array>
+
+// o2 includes
+#include "DataFormatsTPC/TrackTPC.h"
+#include "DataFormatsTPC/dEdxInfo.h"
```
wiechula: is this needed here?

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -0,0 +1,132 @@
...
+ public:
+  void init(framework::InitContext& ic) final
+  {
+    int slot_length = ic.options().get<int>("tf-per-slot");
+    int max_delay = ic.options().get<int>("max-delay");
+    int min_entries = std::max(50, ic.options().get<int>("min-entries"));
+    double min_p = std::min(0.49, ic.options().get<double>("min-momentum"));
+    double max_p = std::max(0.51, ic.options().get<double>("max-momentum"));
+    int min_clusters = std::max(15, ic.options().get<int>("min-clusters"));
+    int nbins = std::max(50, ic.options().get<int>("nbins"));
```
wiechula:
```suggestion
    const int slot_length = ic.options().get<int>("tf-per-slot");
    const int max_delay = ic.options().get<int>("max-delay");
    const int min_entries = std::max(50, ic.options().get<int>("min-entries"));
    const double min_p = std::min(0.49, ic.options().get<double>("min-momentum"));
    const double max_p = std::max(0.51, ic.options().get<double>("max-momentum"));
    const int min_clusters = std::max(15, ic.options().get<int>("min-clusters"));
    const int nbins = std::max(50, ic.options().get<int>("nbins"));
```

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -0,0 +1,132 @@
...
+
+    mCalibrator = std::make_unique<tpc::CalibratordEdx>(min_entries, min_p, max_p, min_clusters, nbins);
+    mCalibrator->setSlotLength(slot_length);
+    mCalibrator->setMaxSlotsDelay(max_delay);
+  }
+
+  void run(ProcessingContext& pc) final
+  {
+    auto tfcounter = o2::header::get<DataProcessingHeader*>(pc.inputs().get("tracks").header)->startTime;
```
wiechula:
```suggestion
   const auto tfcounter = o2::header::get<DataProcessingHeader*>(pc.inputs().get("tracks").header)->startTime;
```

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  void endOfStream(EndOfStreamContext& eos) final
+  {
+    LOG(INFO) << "Finalizing calibration";
+    // FIXME: not sure about this
+    constexpr uint64_t INFINITE_TF = 0xffffffffffffffff;
+    mCalibrator->checkSlotsToFinalize(INFINITE_TF);
+    sendOutput(eos.outputs());
+    mCalibrator->dumpToFile("mip_position.root");
```
wiechula: Please add an option that this is only dumped if activated in the workflow, like here https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/TPC/workflow/include/TPCWorkflow/TPCCalibPedestalSpec.h#L180

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -0,0 +1,108 @@
...
+#include "CCDB/CcdbObjectInfo.h"
+#include "DetectorsCalibration/TimeSlotCalibration.h"
+#include "DetectorsCalibration/TimeSlot.h"
+#include "TPCCalibration/FastHisto.h"
+
+namespace o2
+{
+namespace tpc
+{
```
wiechula: please change this everywhere
```suggestion
namespace o2::tpc
{
```

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -0,0 +1,132 @@
...
+    },
+    outputs,
+    adaptFromTask<device>(),
+    Options{
+      {"tf-per-slot", VariantType::Int, 100, {"number of TFs per calibration time slot"}},
+      {"max-delay", VariantType::Int, 3, {"number of slots in past to consider"}},
+      {"min-entries", VariantType::Int, 100, {"minimum number of entries to fit single time slot"}},
+      {"min-momentum", VariantType::Double, 0.4, {"minimum momentum cut"}},
+      {"max-momentum", VariantType::Double, 0.6, {"maximum momentum cut"}},
+      {"min-clusters", VariantType::Int, 60, {"minimum number of clusters in a track"}},
```
wiechula: at some point we should put this into a track cut class. For the moment it is fine.

wiechula (changes requested):

wiechula (dismissed):
I think for the moment this is fine. Please mark ready for review. We'll continue then with further developments in another PR.

Detectors/TPC/calibration/include/TPCCalibration/dEdxHistos.h
```diff
@@ -0,0 +1,82 @@
...
+class dEdxHistos
+{
+  using Hist = FastHisto<float>;
+
+ public:
+  /// Default constructor
+  dEdxHistos() = default;
+
+  /// Constructor that enable tracks cuts
+  dEdxHistos(TrackCuts cuts, unsigned int nBins);
```
wiechula:
```suggestion
  dEdxHistos(const TrackCuts& cuts, unsigned int nBins);
```
below you move, which will 'return' an empty input object. Not necessarily what one would expect I think.

Detectors/TPC/calibration/include/TPCCalibration/dEdxHistos.h
```diff
@@ -0,0 +1,82 @@
...
+  double getASideEntries() const { return mEntries[0]; }
+
+  /// \return number of entries in the C side histogram
+  double getCSideEntries() const { return mEntries[1]; }
+
+  /// Get the underlying hitograms
+  const std::array<Hist, 2>& getHist() const { return mHist; }
+
+  /// Save the histograms to a file
+  void dumpToFile(const std::string& fileName) const;
```
wiechula:
```suggestion
  void dumpToFile(std::string_view fileName) const;
```

Detectors/TPC/calibration/include/TPCCalibration/dEdxHistos.h
```diff
@@ -0,0 +1,82 @@
...
+  dEdxHistos() = default;
+
+  /// Constructor that enable tracks cuts
+  dEdxHistos(TrackCuts cuts, unsigned int nBins);
+
+  /// Constructor that enable tracks cuts, and creates a TrackCuts internaly
+  dEdxHistos(double minP, double maxP, int minClusters, unsigned int nBins);
+
+  /// Constructor that disables any tracks cuts
+  dEdxHistos(unsigned int nBins);
```
wiechula: Perhaps it is better to set default Track cuts and steer if they should be applied by the flag `mApplyCuts` with setters and getter.

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -12,73 +12,101 @@
...
+  mShouldDump = true;
+  mDumpFile = std::make_unique<TFile>(fileName.c_str(), "recreate");
+  mDumpTree = std::make_unique<TTree>("mipPosition", "mip position information over time");
+
+  mMIPBranch = std::make_unique<CalibMIP>();
+  mHistosBranch = std::make_unique<dEdxHistos>();
+
+  mDumpTree->Branch("timeFrame", &mTimeFrameBranch, "timeFrame/l");
+  mDumpTree->Branch("calibMIP", mMIPBranch.get());
+  mDumpTree->Branch("dEdxHistos", mHistosBranch.get());
```
wiechula: It might be simpler to use the `TreeStream` functionality. Then you don't need to keep branches for data memebers that should be stored in the tree. See e.g. how it is done here:
https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/TPC/calibration/include/TPCCalibration/CalibLaserTracks.h#L120
https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/TPC/calibration/src/CalibLaserTracks.cxx#L82

Detectors/TPC/calibration/src/dEdxHistos.cxx
```diff
@@ -0,0 +1,86 @@
...
+#include "DataFormatsTPC/TrackCuts.h"
+#include "Framework/Logger.h"
+#include "TPCCalibration/FastHisto.h"
+
+//root includes
+#include "TFile.h"
+
+using namespace o2::tpc;
+
+dEdxHistos::dEdxHistos(TrackCuts cuts, unsigned int nBins)
```
wiechula: see comment above

Detectors/TPC/calibration/src/dEdxHistos.cxx
```diff
@@ -0,0 +1,86 @@
...
+  : dEdxHistos({}, nBins)
+{
+  mApplyCuts = false;
+}
+
+void dEdxHistos::fill(const gsl::span<const TrackTPC> tracks)
+{
+  for (const auto& track : tracks) {
+    const auto p = track.getP();
+    const int cluster_count = track.getNClusters();
```
wiechula: Naming convention camel case!

Detectors/TPC/workflow/src/MIPTrackFilterSpec.cxx
```diff
@@ -0,0 +1,96 @@
...
+  mCuts.setPMax(maxP);
+  mCuts.setNClusMin(minClusters);
+}
+
+void MIPTrackFilterDevice::run(ProcessingContext& pc)
+{
+  const auto tracks = pc.inputs().get<gsl::span<TrackTPC>>("tracks");
+
+  std::copy_if(tracks.begin(), tracks.end(), std::back_inserter(mMIPTracks),
+               [this](const auto& track) { return this->mCuts.goodTrack(track); });
```
wiechula, line 67: cool, I like this one!

wiechula (changes requested):
Thanks for the updates! It would be good if you could squash everything, since it is hard to read the diffs now after classes were renamed. You could keep 2 commits only:
1. moving of TrackCuts and all related changes
2. all MIP calibration stuff

Afterwards I'll go through carefully once more and have a look.

For the track cuts it might be nicer to have this configurable `ParameterTrackCuts` using a `ConfigurableParam` as e.g. here: https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/TPC/base/include/TPCBase/ParameterDetector.h and add function like `setCuts(TrackCuts& cuts)`. Then one can configure the cuts easily from the command line. We can discuss details.

Detectors/TPC/calibration/src/CalibdEdx.cxx
badarots: This is nice! I really struggled to get this data-keeping thing right, and it is so ugly. I wasn't aware of another way to do it and I am really glad you point this out. I will changes this right away!

Detectors/TPC/calibration/include/TPCCalibration/dEdxHistos.h
badarots: You are right, this added constructor just add complexity.

Detectors/TPC/calibration/src/dEdxHistos.cxx
badarots: It is stronger than me!

badarots:
I reworked the commit contents as you asked and applied all requested changes, but the `ConfigurableParam` stuff.

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -0,0 +1,113 @@
...
+  void initOutput() final;
+
+  /// Process time slot data and compute its calibration
+  void finalizeSlot(Slot& slot) final;
+
+  /// Creates new time slot
+  Slot& emplaceNewSlot(bool front, TFType tstart, TFType tend) final;
+
+  void setApplyCuts(bool apply) { mApplyCuts = apply; }
+  bool getApplyCuts() { return mApplyCuts; }
```
wiechula, line 76:
```suggestion
  bool getApplyCuts() const { return mApplyCuts; }
```

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -0,0 +1,104 @@
...
+  auto& slot = front ? cont.emplace_front(tstart, tend) : cont.emplace_back(tstart, tend);
+
+  auto container = std::make_unique<dEdxHistos>(mNBins, mCuts);
+  container->setApplyCuts(mApplyCuts);
+
+  slot.setContainer(std::move(container));
+  return slot;
+}
+
+void CalibdEdx::enableDumpToFile(std::string_view fileName)
```
wiechula: perhaps `enableDebugStream` or `enableDebugOutput`.

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -0,0 +1,104 @@
...
+{
+  mDumpStream = std::make_unique<o2::utils::TreeStreamRedirector>(fileName.data(), "recreate");
+}
+
+void CalibdEdx::disableDumpToFile()
```
wiechula: Same as above.

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -0,0 +1,104 @@
...
+{
+  // This will call the TreeStream destructor and write any stored data.
+  mDumpStream.reset();
+}
+
+void CalibdEdx::dumpToFile() const
```
wiechula: Also here the name is a bit misleading, since the dumping to file using the stream functionality is done in the place where it is used. Usually, dumpToFile should really put contents to a file. But this is not what is done here. So one might call it `finalizeDebugOutput` or so.

Detectors/TPC/calibration/include/TPCCalibration/dEdxHistos.h
```diff
@@ -0,0 +1,85 @@
...
+  void setCuts(const TrackCuts& cuts) { mCuts = cuts; }
+
+  /// \return number of entries in the A side histogram
+  double getASideEntries() const { return mEntries[0]; }
+
+  /// \return number of entries in the C side histogram
+  double getCSideEntries() const { return mEntries[1]; }
+
+  /// Get the underlying histograms
+  const std::array<Hist, 2>& getHist() const { return mHist; }
```
wiechula: perhaps
```suggestion
  const std::array<Hist, 2>& getHists() const { return mHist; }
```
since it is not only one.

wiechula (changes requested):
Thanks a lot @badarots! Only minor things left. Please amend them to the latest commit.

badarots:
Hi @wiechula, I applied the requested changes! As you can see I changed the `CalibMIP` from a `std::array` to a custom struct, and added it to the root dictionary. Now the debug output to file is working correctly, but this needs to be changed in the future.

wiechula (dismissed):
@badarots , thanks for all the changes.

badarots:
What now @wiechula? Should I squash the last two commits and proceed to the merge of this PR or do you prefer that I finish the `ParameterTrackCuts` first?

wiechula:
@badarots , merging can only be done by the code owners. Yes, please squash the last two commits. I'll then approve once more and after all test have finished successfully I'll ask to have it merged.

wiechula:
And make it ready for review, of course.

wiechula (approved):

shahor02 (changes requested):
@badarots the o2 CI failure is genuine:
````
/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6305/0/Detectors/TPC/calibration/src/dEdxHistos.cxx: In member function 'void o2::tpc::dEdxHistos::merge(const o2::tpc::dEdxHistos*)':
/mnt/mesos/sandbox/sandbox/o2/sw/SOURCES/O2/6305/0/Detectors/TPC/calibration/src/dEdxHistos.cxx:57:24: error: 'prev' was not declared in this scope; did you mean 'std::prev'?
   57 |     assert(binCount == prev->getHist()[i].getNBins());
      |                        ^~~~
````

File: Detectors/TPC/calibration/src/dEdxHistos.cxx [hunk 26]
```diff
@@ -0,0 +1,77 @@
...
+      }
+    }
+  }
+}
+
+void dEdxHistos::merge(const dEdxHistos* other)
+{
+  for (size_t i = 0; i < mHist.size(); i++) {
+    const auto binCount = mHist[i].getNBins();
+    assert(binCount == prev->getHist()[i].getNBins());
```
wiechula: I guess this should be
```suggestion
    assert(binCount == other->getHist()[i].getNBins());
```
@badarots , can you please check.

wiechula (changes requested):

badarots:
As discussed with @wiechula today, we agreed to just remove that assertion line.

badarots:
So there is an error in the `ensure-executable-naming-convention` test
```
ERROR: executable name o2-tpc-calib-dEdx does not follow naming convention : should only contains lowercase letters, numbers, or dashes
```

Should I just change the executable name to `o2-tpc-calib-dedx` (lowercase "de")?

Detectors/TPC/workflow/CMakeLists.txt
```diff
@@ -103,6 +105,11 @@ o2_add_executable(idc-to-vector
                   SOURCES src/tpc-idc-to-vector.cxx
                   PUBLIC_LINK_LIBRARIES O2::TPCWorkflow)
 
+o2_add_executable(calib-dEdx
```
shahor02: yes, just change this to `calib-dedx`

wiechula:
@shahor02 , @davidrohr , I think the macOS failure is unrelated.

davidrohr:
Probably, I have seen such errors often on the MacCI recently.
@TimoWilken @MichaelLettrich : Could you check why the MacCI fails here?

shahor02:
I thought this problem of compiled macro's test was fixed already... Anyway, the macOS log also contains legitimate warnings:
````
/System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SOURCES/O2/6305/0/Detectors/TPC/qc/macro/runPID.C:55:23: warning: comparison of integers of different signs: 'int' and 'size_t' (aka 'unsigned long') [-Wsign-compare]
    for (int k = 0; k < nTracks; k++) {
````
````
/System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SOURCES/O2/6305/0/Detectors/TPC/qc/macro/runTracks.C:59:23: warning: comparison of integers of different signs: 'int' and 'size_t' (aka 'unsigned long') [-Wsign-compare]
    for (int k = 0; k < nTracks; k++) {
                    ~ ^ ~~~~~~~
````

**Comment by wiechula** [comment]
Indeed, but these lines were not touched by this commit.
@badarots , can you fix it nevertheless in both places:
```c++
for (size_t k = 0; k < nTracks; k++) {
```
should do.

badarots:
I changed `int` -> `size_t` as requested. I'm not sure if I was supposed to amend my last commit or create a new one for this little change. So I amended it.

wiechula (approved):

wiechula:
@shahor02 , can this be merged? I think the macOS failure is again unrelated. cs8 fails with `No space left on device`.

shahor02 (approved):

PR #6306, opened 2021-06-02
Add unassigned tracks as the last entry of the Vtx->Track refs

shahor02:
The indices of unassigned tracks will be added in the end of the vector<VtxTrackIndex>
sent by the VertexTrackMatcher (PVTX_TRMTC DataDescription) and will be referenced
by an extra (last) element of the vector<VtxTrackRef> (sent as PVTX_TRMTCREFS DataDescription).

Therefore, for the TF with N vertices found the size of the vector<VtxTrackRef> will be N+1.

A getter VtxTrackRef::getVtxID() is added and it will return -1 for this special entry collecting
unassigned tracks.


shahor02:
FullCI failure unrelated (timeout?), merging.

PR #6312, opened 2021-06-03
DPL: allow workflows to have dangling inputs until fully merged

ktf:


ktf:
@davidrohr updated to handle also the single workflow with wildcards on the subSpec. We probably need something extra for the Condition case.

davidrohr:
@ktf : Thx, I checked this, and now it seems to work correctly for my cases. However, I have 2 points:
- When an input is really missing at the end, there is now a list of inputs that are missing. But what I don't see is the list of available inputs, which we had before. And also, at some point we added an indication what is the processor name that is missing the input. I don't see this anymore. Could we add this again? It has been quite helpful for debugging.
- Now, when the workflow fails to start, it quickly opens the debuggui and then closes it again. I think before this didn't happen in that case. No big deal, but I am wondering why this is the case now.

davidrohr:
@ktf : And two more points. If it fails because of missing inputs, it does return 0, which should not be the case.
Then, the digitizer workflow fails with
```
[ERROR] Some inputs are not available and therefore the topology cannot be executed
-  { DPL/ENUM/4064465409}
```

ktf:
Ok, I will look into those.

ktf:
Not stale but still not working.

PR #6313, opened 2021-06-03
Data Sampling: remove the possibility to use raw FairMQ channel

knopers8:
This was needed only in early implementation to feed Data Dump with data, which has been recently removed as well.


ktf:
@Barthelemy could you +1 here?

Barthelemy (approved):

PR #6319, opened 2021-06-03
Generate header providing build-time ENABLE_UPGRADES status

shahor02:
If the O2 was built with ENABLE_UPGRADES, using installed headers depending on the latter in the compiled macros
leads to the conflicts since the ACLIC is not aware of ENABLE_UPGRADES.
This PR will generate and install a header which defines ENABLE_UPGRADES if it was used at build-time.


shahor02:
@sawenzel did you mean something like this?

sawenzel:
Yes, exactly !

sawenzel (dismissed):

PR #6322, opened 2021-06-04
FITRaw: common digit2raw procedure is applied(implemented only for FV0)

afurs:


afurs:
It seems that the problem doesn't relate to this PR

afurs:
Issue doesn't relate to this code.

shahor02 (approved):
failures unrelated, local test ok, squashing/merging

PR #6323, opened 2021-06-04
DPL Analysis: avoid instantiating HistogramRegistry functions per histogram name

mario-krueger:
- allow to construct HistName objects from compile-time hashed names
- make HistName a nested class in HistogramRegistry since it has no meaning in global scope
- remove compile-time calculation of index in ConstStr<> since it is very registry specific
- don't inline all the functions

### DIFF
--- a/Framework/Core/include/Framework/HistogramRegistry.h
+++ b/Framework/Core/include/Framework/HistogramRegistry.h
@@ -33,7 +33,6 @@ class TList;
 
 namespace o2::framework
 {
-
 //**************************************************************************************************
 /**
  * Static helper class to fill existing root histograms of any type.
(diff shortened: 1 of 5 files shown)


Framework/Core/include/Framework/HistogramRegistry.h
```diff
@@ -179,28 +178,40 @@ struct HistFiller {
...
 {
+  //**************************************************************************************************
+  /**
+   * HistogramName object providing the associated hash and a first guess for the index in the registry.
+   */
+  //**************************************************************************************************
+  struct HistName {
+    // ctor for histogram names that are already hashed at compile time via HIST("myHistName")
+    template <char... chars>
+    constexpr HistName(const ConstStr<chars...>& hashedHistName)
```
ktf: I would still make the constructor out-of-line.

PR #6331, opened 2021-06-06
Add MID and RICH in TrackSelectorPID

vkucera:
Updated from `dev`


CCDB/src/CcdbApi.cxx
```diff
@@ -468,13 +469,22 @@ void* CcdbApi::extractFromLocalFile(std::string const& filename, std::type_info
 
 bool CcdbApi::checkAlienToken() const
 {
+#ifdef __APPLE__
```
costing, line 472: Grid production jobs see `JALIEN_TOKEN_CERT` set in the environment. So at least that `return true` path has to go first. 

More importantly, with this patch users will not be able to write objects to the production CCDB instance. It effectively forces all Mac users to go to a Linux box to push a new object in. I don't think this is a good idea.

costing (changes requested):

CCDB/src/CcdbApi.cxx
vkucera: @costing I don't think your comment is related to this PR.
This list of commits is just a result of a rebase from `upstream/dev`, so if you see something fishy related to CCDB, it is already in `dev`.

PR #6336, opened 2021-06-07
o2-sim: Stability fix

sawenzel:
Trying to address failures in the CI / lxplus-CVFMS, by:

* cleanup up sockets later
* giving more time to initialize (probably reading from CVMFS
  can slow down the init-process)
* set alignment to none by default as Geant4 crashes on the Mac when this
  is switched on


sawenzel:
@Barthelemy : I have some hope that this fixes O2-2365. Let's see how it goes.

Barthelemy:
@sawenzel thank you

PR #6337, opened 2021-06-07
DPL: guarantee the output-proxy device is always

ktf:
…kflow

### DIFF
--- a/Framework/Core/src/runDataProcessing.cxx
+++ b/Framework/Core/src/runDataProcessing.cxx
@@ -2262,6 +2262,11 @@ int doMain(int argc, char** argv, o2::framework::WorkflowSpec const& workflow,
   // check if DataProcessorSpec at i depends on j
   auto checkDependencies = [& workflow = physicalWorkflow](int i, int j) {
     DataProcessorSpec const& a = workflow[i];
+    // The output proxy is always last so that we
+    // can do the forwarding.
+    if (a.name == "dpl-output-proxy" && i != j) {
+      return true;
+    }
     DataProcessorSpec const& b = workflow[j];
     for (size_t ii = 0; ii < a.inputs.size(); ++ii) {
       for (size_t oi = 0; oi < b.outputs.size(); ++oi) {


ktf:
@matthiasrichter this should allow you to get rid of the memcpy in the OutputProxy without risking to interfere with the forwarding mechanism.

Framework/Core/src/runDataProcessing.cxx
```diff
@@ -2248,8 +2248,10 @@ int doMain(int argc, char** argv, o2::framework::WorkflowSpec const& workflow,
 
   // We sort dataprocessors and Inputs / outputs by name, so that the edges are
   // always in the same order.
+  // We put the output proxy as last device, to make sure we do not need to copy
+  // messages we are sending somewhere else.
   std::stable_sort(physicalWorkflow.begin(), physicalWorkflow.end(), [](DataProcessorSpec const& a, DataProcessorSpec const& b) {
-    return a.name < b.name;
+    return std::make_pair((int)(a.name == "output-proxy"), a.name) < std::make_pair((int)(b.name == "output-proxy"), b.name);
```
knopers8: I think it is called "dpl-output-proxy" by default, unless you plan to change that as well:
https://github.com/AliceO2Group/AliceO2/blob/707fb83a6cfc47d386d5c363c4206865559af5a3/Framework/Utils/src/dpl-output-proxy.cxx#L25

ktf: Nice catch. Thanks.

ktf:
Actually it needs to be done at the level of the topological sort. This seems to work correctly.

matthiasrichter:
Very good, so this makes sure that the output proxy is always the last processor in the DPL workflow. So it is enured that the messages can be forwarded without copying.

matthiasrichter (dismissed):

Framework/Core/src/runDataProcessing.cxx
```diff
@@ -2262,6 +2262,11 @@ int doMain(int argc, char** argv, o2::framework::WorkflowSpec const& workflow,
   // check if DataProcessorSpec at i depends on j
   auto checkDependencies = [& workflow = physicalWorkflow](int i, int j) {
     DataProcessorSpec const& a = workflow[i];
+    // The output proxy is always last so that we
+    // can do the forwarding.
+    if (a.name == "dpl-output-proxy") {
```
matthiasrichter: The only question would be if we have another condition to check. If we change the name, which is defined some other place, this might break the logic

ktf: I was thinking we could introduce some sort of "sink dataprocessor" label to attach to the `DataProcessorSpec`. However that would have the problem that the devices carrying that label in a topology must not share inputs.

That said, I think the output proxy is the only one which needs to behave that way, so I would say that for now the hardcoded check is "sustainable".

ktf:
PR updated with a proposal on how to workaround / fix the issue. I do not like the fact it needs to be handled internally by DPL, but maybe in the long term we could introduce a mechanism similar to the inputs "expiration" for the outputs.

matthiasrichter:
I'm testing right now a solution which handles the forward routes from the specific processor, so we could skip the special handling in DataProcessingDevice.

ktf:
Yes, that would be probably better, I agree. I just could not manage to get it to work in the past because the exposed API involves the buffers, and not the FairMQ messages (for a reason).

ktf:
Ok, I've reverted back to my original PR, because I like your mechanism better. We still need to guarantee output-proxy is last though, because otherwise that will trigger an implicit copy.

PR #6349, opened 2021-06-08
TPC: enable sampling of raw data for QC task

tklemenz:
This adds a workaround to get the raw data to the QC task after it passed data sampling.


wiechula (approved):

tklemenz:
The error seems to be unrelated. Can this be merged?

PR #6351, opened 2021-06-08
DPL: syntactic sugar for WorkflowOptions

ktf:
Rather than using the "customise" method, it is now possible do
define WorkflowOptions merely by defining a:

```
struct WorkflowOptions {
  Configurable<T> myConfigurable{name, defaultValue, help};
  ...
};
```

the system will automatically populate the vector of global
options by inspecting the contents of a default constructed WorkflowOptions
object.


ktf:
@aalkin @jgrosseo this will allow us to define workflow options as members of an optionally defined struct. Passing such a struct via the ConfigContext is however a bit more work and I need to think how we want to do that. The options are therefore for now available via the `ctx.options().get<int>("name")` API only.

PR #6353, opened 2021-06-09
Avoid data copy in the DPL output proxy

matthiasrichter:
This is a demonstrator to use forward routes of the DataProcessingDevice,
which allows to remove handling of messages in the processing callback
There is no need to do processing of the messages except keeping a copy
of the last DataProcessingHeader which is needed to forward EOS.

The remaining issue to be discussed is how to define these external forward
routes. ServiceRegistry is providing access to DeviceSpec, but this is a
const object.

### DIFF
--- a/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
+++ b/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
@@ -94,7 +94,8 @@ void sendOnChannel(FairMQDevice& device, FairMQParts& messages, std::string cons
       break;
     }
   }
-  if (timeout > 0 && timeout <= maxTimeout) {
+  // FIXME: we need a better logic for avoiding message spam
+  if (timeout > 1 && timeout <= maxTimeout) {
     LOG(WARNING) << "dispatching on channel " << channel << " was delayed by " << timeout << " ms";
   }
   // TODO: feeling this is a bit awkward, but the interface of FairMQParts does not provide a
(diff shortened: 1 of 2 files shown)


matthiasrichter:
@ktf, regardless the API we want to introduce, I'm using this prototype now to benchmark the proxies.

ktf:
Very smart! Indeed we could introduce some "ForwardPolicy" which would elegantly solve this (provided the output-proxy is always guaranteed to be last).

matthiasrichter:
This `ForwardPolicy` would be available at configuration time and can be used in the topological sort to make sure that processors using the policy are at the end. Rather then checking for processor names which might change.

ktf:
Having the policy would still not enough, because you do not know a-priori what the policy actually does. However we can have a "DependencyPolicy" which could allow specifying if there is some implicit priority between a device and the others. Let me give it a stab.

ktf:
I've made the dependency resolution for the topological sort customizable (see #6357). Unfortunately the policies still need to be hardcoded in the Framework itself, because the topological sort is something which can span multiple workflows.

matthiasrichter:
Still to be done in this PR: forwarding can only work, if the transports of the channels match. FairMQ does not handle this. In case of the transport not matching, we need to make the copy.

ktf:
I think FairMQ actually does that internally? If that is the only remaining issue, could you mark this PR as ready?

matthiasrichter:
No, FairMQ does not handle this transparently, it's crashing. But we can proceed and fix this in a follow up.

EDIT: I was wrong. It is handled, however it looks like the memory is not released so at some point no more shared memory can be allocated, depending on the parameters in the test, this happens sooner or later. I will collect a bit more statistics and discuss with the FairMQ developers.

ktf:
Issue unrelated. Merging.

PR #6354, opened 2021-06-09
DPL: extend call_if_defined to have a fallback if not defined

ktf:


ktf:
@aalkin unlike we discussed... Unfortunately we cannot leave the second lambda as a template argument because the TUndefined cannot be resolved due to the `auto`.

PR #6367, opened 2021-06-09
Due to G4 the alignment must be applied in the MCApplication::MisalignGeometry

shahor02:
* Move alignment application to MCApplication::MisalignGeometry()
since for the G4 VMC this should be done before the G4Root processing.
* Re-enable applying misalignment by default.

### DIFF
--- a/macro/o2sim.C
+++ b/macro/o2sim.C
@@ -108,6 +108,23 @@ FairRunSim* o2sim_init(bool asservice)
   TStopwatch timer;
   timer.Start();
 
+  o2::detectors::DetID::mask_t detMask{};
+  {
+    auto& modulelist = o2::conf::SimConfig::Instance().getActiveDetectors();
+    for (const auto& md : modulelist) {
+      int id = o2::detectors::DetID::nameToID(md.c_str());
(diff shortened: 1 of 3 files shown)


shahor02:
MacOS failure is not related (test failed to start). @sawenzel note that since the G4 problem is fixed, I am enabling application of alignment by default. Is this ok for you?

macro/o2sim.C
```diff
@@ -108,6 +108,23 @@ FairRunSim* o2sim_init(bool asservice)
   TStopwatch timer;
   timer.Start();
 
+  o2::detectors::DetID::mask_t detMask{};
+  {
+    auto& modulelist = o2::conf::SimConfig::Instance().getActiveDetectors();
```
sawenzel, line 113: just to be sure I understand correctly: Why do you put this code into braces ? ... To be somewhat isolated?  ... This looks like a function body but is not.

sawenzel (approved):

macro/o2sim.C
shahor02: just to isolate it, the braces are not really necessary.

PR #6370, opened 2021-06-10
Analysis/Tasksk/PWGDQ: update v0selector for std::map

dsekihat:
update v0selector for std::map.

### DIFF
--- a/Analysis/Tasks/PWGDQ/v0selector.cxx
+++ b/Analysis/Tasks/PWGDQ/v0selector.cxx
@@ -36,6 +36,7 @@
 
 #include <Math/Vector4D.h>
 #include <array>
+#include <map>
 #include "Framework/ASoAHelpers.h"
 #include "AnalysisDataModel/PID/PIDResponse.h"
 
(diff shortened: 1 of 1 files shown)


Analysis/Tasks/PWGDQ/v0selector.cxx
```diff
@@ -355,7 +293,7 @@ struct v0selector {
     fitter.setMaxChi2(1e9);
     fitter.setUseAbsDCA(true); // use d_UseAbsDCA once we want to use the weighted DCA
 
-    //printf("before entering V0 loop\n");
+    std::map<int, uint8_t> pidmap;
```
ktf, line 296: I would suggest to check if it's not better to have a vector of uint8_t. Each map element above has an overhead of ~24 bytes, so unless pidmap is very sparse, it's not gonna be a win memory wise (and lookup wise it's going to be O(1) vs O(log(N)).

dsekihat: Hi @ktf , thank you for your comments. But, sorry, I am not a memory expert and did not get your points.
What I need is upto 5 bits. So, uint8_t is minimum requirement, right?
So, could you tell me what  you suggest exactly, please?

ktf: I mean using a `std::vector<uint8_t> pidmap` which is sized for the maximum value of pidmap.size() so that you can use it for the lookup. There will be empty spots in the vector, but I guess it might still be a win.

dsekihat: Oh, I see. Thank you for your explanations. I will test it later.

dsekihat: By the way, how can I get maximum value of global track index?
The element for vector is global track index.
This the reason why std::map is easy and convenient in this task.

dsekihat: Hi again,
I use std::map because, I want to compare global track indices and pos/neg track indices from V0s and cascades.
That's why std::map is convenient and easy way for this task.
What do you think?

PR #6372, opened 2021-06-10
DPL: specify runNumber to the monitoring interface

ktf:


ktf:
@awegrzyn

awegrzyn:
Is it guaranteed to run when "run number" changes to new one?

ktf:
not for now. I need to add a callback on the DPL level for that.

davidrohr (approved):
works, thx

PR #6373, opened 2021-06-10
Adding benchmark workflow for the DPL proxies

matthiasrichter:
DPL output and input proxies are running in the same workflow, but connected
through an out-of-band channel. Note that the channel is not known in DPL, so
the connection is not displayed in the Gui either.

producer -> output proxy -> out-of-band-channel -> input proxy -> checker

```
Configurable parameters:
  --nChannels arg (=1)                  number of output channels of the producer
  --msgSize arg (=1024)                 message size in kB
  --runningTime arg (=30)               time to run the workflow
  --rate arg                            rate for a data source device (Hz)
```

### DIFF
--- a/Framework/Core/CMakeLists.txt
+++ b/Framework/Core/CMakeLists.txt
@@ -258,6 +258,7 @@ foreach(b
         HistogramRegistry
         TableToTree
         TreeToTable
+        ExternalFairMQDeviceProxies
         )
   o2_add_executable(benchmark-${b}
                     SOURCES test/benchmark_${b}.cxx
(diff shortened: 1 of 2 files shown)


matthiasrichter:
@ktf, Is there a way to get more information what clang-format CI is complaining about? More than
```
PR not clean
Error: Process completed with exit code 1.
```

I can not find any formatting issue with clang-format locally.

ktf:
I am not sure about the error. @TimoWilken might want to comment.

Very nice. Two comments:

* apparently `make<std::vector>` is very slow, you might want to use `make<char>`.
* how can I increase the number of splitSubParts?

matthiasrichter:
I have changed it to `make<char>`. That means we should at some point look into the performance of `make<vector<T>>` after putting so much effort into that.

As for the split parts, we do not support creation of split parts in DPL, do we? The number of messages can be controlled by parameter `nChannels`, but that's messages with unique identifier by choosing subspecs, that's not split parts like coming from STFSender.

I can implement this later via sending directly on `FairMQChannel`.instead of using `DataAllocator`

matthiasrichter:
merging, the fullCI failure is unrelated.

knopers8:
> As for the split parts, we do not support creation of split parts in DPL, do we?

Wasn't it possible by doing `make`/`snapshot`/`adopt` many times with the same output and within the same `process` invocation?

PR #6374, opened 2021-06-10
Add 2D histograms for pt jpsi vs pt daughter

zhangbiao-phy:


Analysis/Tasks/PWGHF/taskJpsi.cxx
```diff
@@ -188,6 +191,16 @@ struct TaskJpsiMC {
...
         registry.fill(HIST("hPtGen"), particle.pt());
         registry.fill(HIST("hEtaGen"), particle.eta());
+        registry.fill(HIST("hYGen"), RecoDecay::Y(array{particle.px(), particle.py(), particle.pz()}, RecoDecay::getMassPDG(particle.pdgCode())), particle.pt());
+
+        float ptProngs[2];
+        int counter = 0;
+        for (int iD = particle.daughter0(); iD <= particle.daughter1(); ++iD) {
+          ptProngs[counter] = particlesMC.iteratorAt(iD).pt();
+          counter++;
+        }
```
vkucera: What's the point of doing this?

zhangbiao-phy: here I reference "taskX.cxx" to assign the pt of daughters

vkucera: * It is not in the `dev` version, which means it has not been reviewed.
* I don't see why you want to use an array for storing two `float` values which you use just once.
* Why don't you fill them directly?

zhangbiao-phy: Ok, I can simply this

vkucera (approved):

PR #6375, opened 2021-06-10
DPL: Add --no-catchall option to disable catching of exceptions such …

ktf:
…that they are passed to e.g. gdb

### DIFF
--- a/Framework/Core/include/Framework/runDataProcessing.h
+++ b/Framework/Core/include/Framework/runDataProcessing.h
@@ -148,78 +148,95 @@ std::vector<T> injectCustomizations()
   policies.insert(std::end(policies), std::begin(policies), std::end(policies));
 }
 
+int mainNoCatch(int argc, char** argv)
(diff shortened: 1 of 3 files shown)


davidrohr:
thx, will try that later.

davidrohr:
ok, option parsing works.
But apparently I did not manage to apply it to all cases. At least some exceptions are still caught.
Will need one more update, working on it.

davidrohr:
@ktf : OK, actually, the option parsing doesn't work. It always returns false for me, even if I run with `--no-catchall`. Could you check again? (In the meantime I have fixed 2 typos of mine in your branch...)

ktf:
it seems also to break the analysis  stuff....

davidrohr:
How can it possibly break the analysis stuff? This just moved some code into functions.

ktf:
not sure yet, but reverting this systematically makes it work again.

PR #6382, opened 2021-06-10
[MRRTF-117] Various Encoder fixes.

aphecetche:
Those changes are intended to fix a number of issues in the MCH
digits-to-raw process. The changes are relatively numerous as some
refactoring was done.

The whole process sim->digits->raw->digits can be tested with the
`Detectors/MUON/MCH/Raw/test/codec-closure-test.sh` script. Note that a)
this script will be migrated to the full-system (extra) test somehow b)
it is _not_ currently passing because the digitizer has some issues
(that will be fixed in a forthcoming fix)


aphecetche:
@shahor02 this is the "meat" of the series of PRs (last but one probably) concerning the MCH Digits->raw->digits cycle. The intent of this one is to fix the issues you raised a long time ago about the MCH Digits to raw program (plus some more that I uncovered while working on it, like the correct treatment of the versioning of our UL format)

shahor02:
Hi @aphecetche 

Given the amount of changes, please let me know if you expect me to check something particular :)
If the raw encoding / decoding is operational now, perhaps you can add it to the FST with CTF extraction, this would be a good test of the full chain.

BTW, fullCI complains on malformed license in some files

aphecetche:
@shahor02 I realize the PR is quite massive, despite my attempt to split it in chunks, sorry about that. No, nothing in particular to be checked I would say. Concerning the FST, it's planned but first I need some fixes in the digitizer  (#6385).
For the FullCI thanks for noticing, indeed I forgot some copyrights. Will add them.

shahor02 (approved):
FST failure unrelated, merging

PR #6383, opened 2021-06-10
DPL: optimize handling of splitParts

ktf:


ktf:
@matthiasrichter @ironMann @knopers8  I think this is what is needed to handle the performance issues when too many splitParts are present. The idea is that DD guarantees that split parts all have the same origin/description/subspecification and they must travel together.

Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -825,8 +830,10 @@ void DataProcessingDevice::handleData(DataProcessorContext& context, FairMQParts
           auto headerIndex = 2 * pi;
           auto payloadIndex = 2 * pi + 1;
           assert(payloadIndex < parts.Size());
+          auto dh = o2::header::get<DataHeader*>(parts.At(headerIndex)->GetData());
           auto relayed = relayer.relay(std::move(parts.At(headerIndex)),
-                                       std::move(parts.At(payloadIndex)));
+                                       &parts.At(payloadIndex), dh->splitPayloadParts - 1);
```
matthiasrichter: `splitPayloadParts` can be zero, at least its easy to be fail-safe on that.

I suggest that numPayloads in the function below should denote all payloads not just the remainder if there are split parts. So we pass at least 1.
```suggestion
                                       &parts.At(payloadIndex), dh->splitPayloadParts > 0 ? dh->splitPayloadParts  : 1);
```

Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -825,8 +830,10 @@ void DataProcessingDevice::handleData(DataProcessorContext& context, FairMQParts
...
+          pi += dh->splitPayloadParts - 1;
```
matthiasrichter:
```suggestion
          pi += dh->splitPayloadParts > 0 ? dh->splitPayloadParts - 1 : 0;
```

Framework/Core/src/DataRelayer.cxx
```diff
@@ -196,6 +196,14 @@ void sendVariableContextMetrics(VariableContext& context, TimesliceSlot slot,
 DataRelayer::RelayChoice
   DataRelayer::relay(std::unique_ptr<FairMQMessage>&& header,
                      std::unique_ptr<FairMQMessage>&& payload)
+{
+  return relay(std::move(header), &payload, 1);
```
matthiasrichter, line 200: With the logic implemented below this should be zero as it describes the number of additional payloads. It does not harm since there is a division by 2, which make it to zero then. I would leave it like this and change the implementation below so that numPayloads denotes all payloads.

Framework/Core/src/DataRelayer.cxx
```diff
@@ -274,9 +283,13 @@ DataRelayer::RelayChoice
     cachedStateMetrics[cacheIdx] = CacheEntryStatus::PENDING;
     // TODO: make sure that multiple parts can only be added within the same call of
     // DataRelayer::relay
-    PartRef entry{std::move(header), std::move(payload)};
+    PartRef entry{std::move(header), std::move(payloads[0])};
     parts.emplace_back(std::move(entry));
-    assert(header.get() == nullptr && payload.get() == nullptr);
+    auto rest = payloads + 1;
+    for (size_t pi = 0; pi < numPayloads / 2; ++pi) {
```
matthiasrichter: `numPayloads` does not specify the number of messages but message pairs, so no division by 2. Also I would for clarity pass not splitParts - 1, but the total number of payloads, so the upper bound is one less
```suggestion
    for (size_t pi = 0; pi < numPayloads - 1; ++pi) {
```

Framework/Core/src/DataRelayer.cxx
```diff
@@ -274,9 +283,13 @@ DataRelayer::RelayChoice
     cachedStateMetrics[cacheIdx] = CacheEntryStatus::PENDING;
     // TODO: make sure that multiple parts can only be added within the same call of
     // DataRelayer::relay
-    PartRef entry{std::move(header), std::move(payload)};
+    PartRef entry{std::move(header), std::move(payloads[0])};
     parts.emplace_back(std::move(entry));
-    assert(header.get() == nullptr && payload.get() == nullptr);
+    auto rest = payloads + 1;
```
matthiasrichter: Simply would add a comment here
```suggestion
    // rest is now pointing to the header message of the next message pair
    auto rest = payloads + 1;
```

ktf:
Now with benchmark and test. Apparently there is an issue when using `CMAKE_BUILD_TYPE=Release` investigating.

ktf:
Good news is that in Debug mode we seem to pay a 2x slowdown for 100x more splitparts.

ktf:
No, actually the test and the benchmark are fine. In optimised mode it's more like 30% slowdown for 100x splitParts. Now investigating the crashes when running in DataProcessingDevice as opposed to the synthetic test.

matthiasrichter (commented):
I would say there is a flaw in the logic for looping over the additional parts.

Also, since `splitPayloadParts` minus one is passed, `numPayloads` should be renamed to indicate that its not all payloads but the remainder.

I would prefer to leave the name, pass `splitPayloadParts`, and use `numPayloads - 1` as upper bound for the loop.

Framework/Core/src/DataProcessingDevice.cxx
ktf: @ironMann what is the value of splitPayloadParts in case of a single (header, payload) pair? 0 or 1?

Framework/Core/src/DataRelayer.cxx
ktf: I fixed the logic to have `(restOfPartsSize - 1) / 2`. Also `numPayloads` was renamed to `restOfPartsSize` which is the total size of all the messages part of the multipart, minus the first header which is passed separately.

Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -825,8 +830,10 @@ void DataProcessingDevice::handleData(DataProcessorContext& context, FairMQParts
           auto headerIndex = 2 * pi;
           auto payloadIndex = 2 * pi + 1;
           assert(payloadIndex < parts.Size());
+          auto dh = o2::header::get<DataHeader*>(parts.At(headerIndex)->GetData());
           auto relayed = relayer.relay(std::move(parts.At(headerIndex)),
-                                       std::move(parts.At(payloadIndex)));
+                                       &parts.At(payloadIndex), dh->splitPayloadParts * 2 - 1);
```
ktf: @matthiasrichter I think I was lacking a `*2` here...

Framework/Core/src/DataRelayer.cxx
matthiasrichter: perfect

Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -807,7 +807,15 @@ void DataProcessingDevice::handleData(DataProcessorContext& context, FairMQParts
...
-      results[hi] = InputType::Data;
+      // We can set the type for the next splitPayloadParts
+      // because we are guaranteed they are all the same.
+      // If splitPayloadParts = 0, we assume that means there is only one (header, payload)
+      // pair.
+      size_t finalSplitPayloadIndex = hi + (dh->splitPayloadParts > 0 ? dh->splitPayloadParts : 1);
+      for (; hi < finalSplitPayloadIndex; ++hi) {
+        results[hi] = InputType::Data;
+      }
+      hi = finalSplitPayloadIndex - 1;
```
ktf, line 818: For the record, the lacking of `-1` here was the actual issue.

Framework/Core/src/DataProcessingDevice.cxx
matthiasrichter: yes, now the logic is consistent

PR #6384, opened 2021-06-10
Add bit counting to GPUCommonBitSet

shahor02:


davidrohr (changes requested):
Coult you use `GPUCommonMath::Popcount` for this, which should issue dedicated CPU/GPU instructions for that (just a minor optimization).
And actually, `GPUCommonMath::Popcount` has a fallback if no popc instruction is available.
Could you replace that fallback with your optimized code from here?

davidrohr (approved):

davidrohr:
error unrelated

PR #6385, opened 2021-06-11
MCHSimulation: Rewrite digitizer to make it simpler and faster.

aphecetche:
Also improve a bit the noise generation (e.g. avoid to add zero charge digits)

@mwinn2 @sawenzel I believe this version has all the features of the previous one, but I only really tested it in the context of my digits->raw->digits closure test exercise so far. Is there a definitive stress test for digitizers that I should pass first ?


sawenzel:
@aphecetche : There is `prodtests/check_embedding_pileup.sh` which is verifying if embedding / pileup is handled correctly (via MC labels).

sawenzel (dismissed):
looks ok to me

aphecetche:
@sawenzel Using `./check_embedding_pileup.sh` I got : 

```
MCH 1446 10875 pileup 1 embedding 1
```

which I guess means all is fine. 

For the record, I'm noticed a few things on my system (Ubuntu 20.04 in a VM) : 
- I often got o2-sim and o2-sim-hit-merger-runner stuck processes.
- I get plenty of PHOS related warnings : 
```
Warning in <TInterpreter::ReadRootmapFile>: class  o2::phos::CalibParams found in libO2PHOSCalib.so  is already in libO2DataFormatsPHOS.so
Warning in <TInterpreter::ReadRootmapFile>: class  o2::phos::Pedestals found in libO2PHOSCalib.so  is already in libO2DataFormatsPHOS.so
Warning in <TInterpreter::ReadRootmapFile>: class  o2::phos::TriggerMap found in libO2PHOSCalib.so  is already in libO2DataFormatsPHOS.so
```
- and also plenty (basically in all digilogXXX files) of ITS and MFT alignment related error messages (@shahor02) of the kind : 
```
617737:MCHDigitizer]: [19:40:16][ERROR] Volume path ITS/ITSULayer6/ITSUStave3/ITSUHalfStave0/ITSUModule3/ITSUChip6 is not valid
[617737:MCHDigitizer]: [19:40:16][ERROR] Error applying alignment object for volumeITS/ITSULayer6/ITSUStave3/ITSUHalfStave0/ITSUModule3/ITSU
Chip6
```

shahor02:
@aphecetche 
For the `Warning in <TInterpreter::ReadRootmapFile>: class  o2::phos::CalibParams found in libO2PHOSCalib.so  is already in libO2DataFormatsPHOS.so`: some time ago PHOS has moved these classes from Detectors to DataFormats. Apparently in your install directory you still have rootmaps from the old version.

For the `[ERROR] Volume path ITS/ITSULayer6/ITSUStave3/ITSUHalfStave0...`: when loading geometry it appears when you apply an alignment to detector which is not in the geometry, I guess you have simulated MCH only.
I can change it to warning, but then, if something is really wrong with the alignment, we will not notice this.
You can mute the error by passing `--configKeyValues "align-geom.mDetectors=none;"` (to disable applying alignment) or 
`--configKeyValues "align-geom.mDetectors=MCH,MID;"` (putting only detectors you have simulated or want to misalign (in any case, at the moment ccdb has only non-empty ITS and MFT alignment files).

We could avoid aligning detectors which are not in the GRP detector mask (at the moment I do this in o2-sim) but this is conceptually wrong: alignment is relevant not only for sensors but also for detectors as passive material. Instead, we should change the way the geometry is generated: selecting only particular modules in the o2-sim should desactivate the rest as active detectors, but still complete geometry should be built.

aphecetche:
@shahor02 assuming the fullCI failure is, as usual, not related to this PR, I would say this PR is ready to be merged (your point about the need to always get the full geometry is a very good one, btw, but to be followed-up in some Jira ticket probably ?)

shahor02:
Yes, fullCI regularly fails either due to the out-of-memory compiler or timeout in generation, merging.
Will create a ticket for the geometry.

davidrohr:
GIven this was the last PR mentioned in the JIRA ticket, does it mean the MCH raw data creation is finally working?

aphecetche:
@davidrohr yes, that should be the case. Note that I'm now working on integrating MCH into the FST (e.g. need to introduce a mch reco workflow to "wrap up" the piping command we currently use for our tests).

davidrohr:
ah, nice, so it seems I don't need to do anything :)
Thx. Ping me when you are ready, then I'd like to test it with a full 128 orbit TF.

PR #6386, opened 2021-06-11
MCH: do not throw exception when linkID equals 15

aferrero2707:
In the UserLogic data, the linkID=15 condition is used to identify special control and diagnostics words that are generated internally and do not contain data coming from the front-end electronics.

The exact format of those control words is not yet finalized, hence this PR does not introduce any specific decoding yet.
Instead, it simply avoids throwing an exception when the linkID=15 condition is found.

### DIFF
--- a/Detectors/MUON/MCH/Raw/Decoder/src/UserLogicEndpointDecoder.h
+++ b/Detectors/MUON/MCH/Raw/Decoder/src/UserLogicEndpointDecoder.h
@@ -46,11 +46,11 @@ class UserLogicEndpointDecoder : public PayloadDecoder<UserLogicEndpointDecoder<
                            std::function<std::optional<uint16_t>(FeeLinkId id)> fee2SolarMapper,
                            DecodedDataHandlers decodedDataHandlers);
 
-  /** @name Main interface 
+  /** @name Main interface
     */
   ///@{
 
-  /** @brief Append the equivalent n 64-bits words 
+  /** @brief Append the equivalent n 64-bits words
     * bytes size (=n) must be a multiple of 8
     *
     * @return the number of bytes used in the bytes span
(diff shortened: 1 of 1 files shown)


Detectors/MUON/MCH/Raw/Decoder/src/UserLogicEndpointDecoder.h
```diff
@@ -119,6 +119,15 @@ size_t UserLogicEndpointDecoder<CHARGESUM, VERSION>::append(Payload buffer)
 
     int gbt = ulword.linkID;
 
+    // The User Logic uses the condition gbt=15 to identify special control and diagnostics words
+    // that are generated internally and do not contain data coming from the front-end electronics.
+    if (gbt == 15) {
+      // TODO: the exact format of the UL control words is still being defined and tested,
+      // hence proper decoding will be implemented once the format is finalized.
+      // For the moment we simply avoid throwing an exception when linkID is equal to 15
+      continue;
```
aphecetche, line 128: I don't understand how that's going to prevent the exception below ? Did you mean `return 0;` instead of `continue` maybe ?

Anyway, I would put that == 15 test within the <0 || > 11 test itself to avoid a double if for regular gbt values (i.e. between 0 and 11) (assuming here an extra if has potential performance implications, but that should be measured of course)

aferrero2707: Unless I am really overlooking something, the `continue` statement prevents the code in lines 131 and below to be executed, the same way as we skip words that are zero or equal to `0xFEEDDEEDFEEDDEED`.

Also, the standard decoding (lines 141 and below) should not be applied to those words, since they do not contain front-end data. Therefore an `if (gbt == 15)` statement is needed anyhow.

One possibility to prevent the double `if` for each 64-bit word could be to replace the `if (gbt < 0 || gbt > 11)` statement on line 131 by `else if (gbt < 0 || gbt > 11)`. From the code flow point of view, that would result in something like
```
If gbt == 15 then 
    handle the special words
otherwise if invalid
    throw an exception or call the error handler
```

What do you think?

aphecetche: My bad. Forgot that 1st comment, was looking too fast so I've missed the external loop altogether ;-). So the continue actually makes perfect sense. Sorry about that.

For the double if, yes an else would do it.

aphecetche (approved):

aferrero2707:
@ktf the fullCI fails without any error log. Can we merge this anyhow? Thanks!

ktf:
@TimoWilken could you look into this?

PR #6387, opened 2021-06-11
Fix segfault in primary server device upon termination signal

sawenzel:
Apparently the Run function continues after a signal was
received which lead to an interrupt in the ZMQ Receive function.

Now detecting this situation and initiating clean shutdown ourselfs.

Fixes https://alice.its.cern.ch/jira/browse/O2-2367


sawenzel:
@davidrohr

davidrohr:
thx

PR #6390, opened 2021-06-11
TRD limit raw digits from MC like real data

bazinski:
- cli option to trap2raw to limit rate of digits in the raw stream
- default is 1000, 1 in 1000
- option "digitrate"

This then causes the mc to raw data to have the same 1 in 1000 events with digits, reducing the datarate significantly for full system tests.


bazinski:
This is on top of PR5912

davidrohr:
OK, tested this, but only on a small dataset:
before
```
[66784:trd-datareader]: [22:45:57][INFO] Digits found : 447967
[66784:trd-datareader]: [22:45:57][INFO] Digits returned : 447967
[66784:trd-datareader]: [22:45:57][INFO] Tracklets found : 48971
[66784:trd-datareader]: [22:45:57][INFO] Tracklets returned : 48971
```

Afterwards:
```
[67727:trd-datareader]: [22:47:45][INFO] Digits found : 13292
[67727:trd-datareader]: [22:47:45][INFO] Digits returned : 13292
[67727:trd-datareader]: [22:47:45][INFO] Tracklets found : 48971
[67727:trd-datareader]: [22:47:45][INFO] Tracklets returned : 48971
```

I think the reason that the digit count is not reduced by 1000 (besides cenrality fluctuations) is simply that I don't have enough events in my data set. So this seems to work, thanks. Also doesn't break any of the other parts of the FST.

davidrohr:
closing since superseded by #6400

PR #6391, opened 2021-06-11
Make sure CMAKE_MODULE_PATH is not reset. Provide second search using arrow_DIR

pzhristov:
This PR should enable the use of Arrow 3.0.0.


pzhristov:
Depends on https://github.com/alisw/alidist/pull/3088

pzhristov:
The error is not related to the change, merging.

PR #6393, opened 2021-06-11
TRD return digits not compresseddigits

bazinski:
the reader was returning compressed digits.
removed compressed digits, the compression is nominal due to the infrequent digit events.


bazinski:
needs to be rebased after PR5912

davidrohr:
For reference, I have tested the TRD entropy encoding with this fix, and it seems to work.

davidrohr:
closing since superseded by #6400

PR #6394, opened 2021-06-11
MFT: rename macro for dictionary creation; fix dictionary function call; move ClusterPattern init before loop

bovulpes:
- small fix in the tracker (variable "seed" is always true) 
- rename the macro CheckTopologies.C to CreateDictionaries.C
- fix boolean value argument in the function getClusterCoordinates from the dictionary
- move initialization of ClusterPattern before the loop (do it only once)

### DIFF
--- a/Detectors/ITSMFT/MFT/tracking/src/IOUtils.cxx
+++ b/Detectors/ITSMFT/MFT/tracking/src/IOUtils.cxx
@@ -55,11 +55,11 @@ int ioutils::loadROFrameData(const o2::itsmft::ROFRecord& rof, ROframe& event, g
         locXYZ = dict.getClusterCoordinates(c);
       } else {
         o2::itsmft::ClusterPattern patt(pattIt);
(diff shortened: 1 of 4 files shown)


shahor02 (approved):

Detectors/ITSMFT/MFT/tracking/src/IOUtils.cxx
```diff
@@ -54,12 +55,11 @@ int ioutils::loadROFrameData(const o2::itsmft::ROFRecord& rof, ROframe& event, g
       if (!dict.isGroup(pattID)) {
         locXYZ = dict.getClusterCoordinates(c);
       } else {
-        o2::itsmft::ClusterPattern patt(pattIt);
-        locXYZ = dict.getClusterCoordinates(c, patt, false);
+        locXYZ = dict.getClusterCoordinates(c, patt);
```
shahor02, line 58: this leads to segfault: the pattern must be set for each cluster individually

PR #6395, opened 2021-06-11
DPL: dump per arrow table metrics

ktf:


ktf:
@jgrosseo this should allow us to see per table memory usage in hyperloop.

ktf:
@Barthelemy @knopers8 multinode_test still fails.

Barthelemy:
@ktf @knopers8 I don't see it failing ?

PR #6396, opened 2021-06-11
ZDC New format for RecEvent and calibration objects for TDC and energy

cortesep:


shahor02:
@cortesep please fix errors reported by macOS CI

cortesep:
After modification the macOS still fails but no reason is listed

shahor02:
@cortesep MacOS failed to start, this happens and unrelated to PR.
But the FullCI fails with:
````
/mnt/mesos/sandbox/sandbox/o2-fullci/sw/SOURCES/O2/6396/0/Detectors/ZDC/reconstruction/src/DigiReco.cxx:42:19: error: statement should be inside braces [readability-braces-around-statements]
````

**Comment by cortesep** [comment]
Failure of fullCl seems not to be due to code issues. What do you think?

File: DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEventAux.h [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+  uint32_t triggers = 0;                                 /// pattern of channels with autotrigger bit
+  uint32_t flags;                                        /// reconstruction flags
+  std::map<uint8_t, float> ezdc;                         /// signal in ZDCs
+  int16_t tdcVal[NTDCChannels][MaxTDCValues];            /// TdcChannels
+  int16_t tdcAmp[NTDCChannels][MaxTDCValues];            /// TdcAmplitudes
+  int ntdc[NTDCChannels] = {0};                          /// Number of hits in TDC
+  std::array<bool, NTDCChannels> pattern;                /// Pattern of TDC
+  uint16_t fired[NTDCChannels] = {0};                    /// Position at which the trigger algorithm is fired
+  float inter[NTDCChannels][NTimeBinsPerBC * TSN] = {0}; /// Interpolated samples
+  uint32_t ref[NChannels] = {O2_ZDC_REF_INIT};           /// Cache of references
```
shahor02: Did not we agree to avoid this kind of initialization? It is error prone and not faster than doing an initialization in the c-tor.

cortesep: I must have forgot to note it down. Now the initialization is corrected.

shahor02:
@cortesep Do you have anything to add to this PR? For me it can be merged, the next step will be to add the reconstruction to the full system test.

cortesep:
Thanks Ruben. For me you can merge. It would be very useful to make a test of the reconstruction with the full system. It would be also useful to introduce the background from electromagnetic dissociation. I made a rough estimate and the reconstruction time could be around 1 hour * core / second_of_data_taking

I will try to add the protection in case of "deadbeef" next week in a pause from exams..

shahor02:
Hi @cortesep 
ok, will merge after one test. How do you generate EM dissociation?
1h / core for 1 s of data is quite slow. The is the time we spend to do tracking in the ITS, which has 10^9 hits/s. Should be optimized...

shahor02:
@cortesep ok, my tests passed, merging. The verbosity seems to be too high, will need to reduce it.

cortesep:
> Hi @cortesep
> ok, will merge after one test. How do you generate EM dissociation?
> 1h / core for 1 s of data is quite slow. The is the time we spend to do tracking in the ITS, which has 10^9 hits/s. Should be optimized...

EMD needs to be simulated starting with aliroot kine tree. I will ask to Chiara O. 
Concerning the hit rate, if you consider 5 MHz rate * 6 channels for neutron calorimeters * 24 samples we reach numbers comparable with ITS... :-)

shahor02:
Hi @cortesep 

For the generator, perhaps you should think of native O2 solultions, most of generators can be relatively easily adapted to O2.
Concerning your comparison of ITS and ZDC rates: not sure it is valid since the ITS does track finding with these hits, the combinatorics of considered segments is a few order of magnitude higher.
Anyway, let's speak after some benchmarks.

At the moment I see other potential problem: in the enf of `o2-zdc-digi2raw  --file-per-link -o raw/ZDC` output I see:
````
[WARN] RawFileWriter forced 225 dummy addData calls in 15 IRs for links which did not receive data
````

This means that there were 15  IRs where you have called `writer.addData` for at least one link but not for all.
This is in principle wrong: if detector "triggers", all its links should receive addData, even if the payload is empty. The RawFileWriter has a protection for this, adding `addData` calls w/o payload when it was missed, but it cannot know if just writing an RDH w/o any payload is enough for processing empty orbit.
Could you make sure that `addData` is called for every link?

cortesep:
I have to discuss with Chiara who ported already some generators for the ZDC. 
I'm inserting a call to the function with empty payload for the links that have no data.

cortesep:
Hi @shahor02 ,

I fixed the addData by inserting dummy calls but I still get the warning message but just once per link. Is it expected? Could it be the final stage of the processing where the Digits2Raw class is not called anymore?

[WARN] RawFileWriter forced 16 dummy addData calls in 4237 IRs for links which did not receive data

shahor02:
Hi @cortesep 
no, the protection on the writer side is triggered if one link does not get addData for the IR for which at least 1 other link got it.
Could you put some temporary logs, before you addData calls and check where do yo jump over the link?

shahor02:
PS: if you run with high verbosity, e.g. `o2-zdc-digi2raw -o raw/ZDC -v 12 |  tee xx.log`
then `grep Complete xx.log`, you should see the link-id and IR where RawWriter had to call addData.

cortesep:
Hi, I corrected the remaining bug. Now it seems to work. I opened a pull request (still with WIP for the moment). Do you want to proceed with this or you prefer to wait for the fix on the deadbeef?

shahor02:
Hi @cortesep , 
let's not mix different things, remove the WIP from https://github.com/AliceO2Group/AliceO2/pull/6515 and implement the missing TF protection in different PR.

PR #6402, opened 2021-06-12
TRD remove unused array causing segfaults for large datasets

martenole:
@davidrohr thanks for investigating


davidrohr:
trivial + tested locally. Also fixes the problem in #6400.
Merging

PR #6413, opened 2021-06-13
Ctpdev: RawToDigit

lietava:
Hi @shahor02 , I started to work on DigitToRaw. 
0I tried to follow the CPV example you suggested in previous PR. I added files in CTP workfloa and now trying to modify them. 
I have couple of questions:
1.) as we discussed CTP has no MC truth so I can ignore all related staff, right ?
2.)  main question:
void RawToDigitConverterSpec::run(framework::ProcessingContext& ctx) calls o2::xxx::RawReaderMemory rawreader(o2::framework::DataRefUtils::as<const char>(rawData));
which data like in DMA memory : I guess data created by o2::ctp::Digits2Raw are in that format ?
Should I create also RawReadrMemory like in CPV ?
Cheers, R.


shahor02:
Hi @lietava 

> 1.) as we discussed CTP has no MC truth so I can ignore all related staff, right ?

Right, I don't see what can be the MC truth for the CTP. One can think of the presence of hadronic interaction (the alternative being generating trigger on QED or some late particle/afterpulse), but this can be checked directly with the info we have in the AOD.

> 2.) main question:
> void RawToDigitConverterSpec::run(framework::ProcessingContext& ctx) calls o2::xxx::RawReaderMemory rawreader(o2::framework::DataRefUtils::as(rawData));
> which data like in DMA memory : I guess data created by o2::ctp::Digits2Raw are in that format ?

Yes, o2-raw-file-reader injects into the the DPL the the raw data for each link: RDH followed by the payload you have provided when creating raw data.

> Should I create also RawReadrMemory like in CPV ?

There is no any convention on naming the routines, but in one way or another you should navigate over the inputs of `CTP/RAWDATA` and convert raw data to digit. You can also check how I do this and cache the data in ITS/TPC decoder:
https://github.com/AliceO2Group/AliceO2/blob/3ad766bacf1513a9cf9dea485a9c02b95d8621a8/Detectors/ITSMFT/common/workflow/src/STFDecoderSpec.cxx#L110
https://github.com/AliceO2Group/AliceO2/blob/3ad766bacf1513a9cf9dea485a9c02b95d8621a8/Detectors/ITSMFT/common/reconstruction/src/RawPixelDecoder.cxx#L125

Cheers,
 Ruben

lietava:
> Hi @lietava
> 
> > 1.) as we discussed CTP has no MC truth so I can ignore all related staff, right ?
> 
> Right, I don't see what can be the MC truth for the CTP. One can think of the presence of hadronic interaction (the alternative being generating trigger on QED or some late particle/afterpulse), but this can be checked directly with the info we have in the AOD.
> 
> > 2.) main question:
> > void RawToDigitConverterSpec::run(framework::ProcessingContext& ctx) calls o2::xxx::RawReaderMemory rawreader(o2::framework::DataRefUtils::as(rawData));
> > which data like in DMA memory : I guess data created by o2::ctp::Digits2Raw are in that format ?
> 
> Yes, o2-raw-file-reader injects into the the DPL the the raw data for each link: RDH followed by the payload you have provided when creating raw data.
> 
> > Should I create also RawReadrMemory like in CPV ?
> 
> There is no any convention on naming the routines, but in one way or another you should navigate over the inputs of `CTP/RAWDATA` and convert raw data to digit. You can also check how I do this and cache the data in ITS/TPC decoder:
> https://github.com/AliceO2Group/AliceO2/blob/3ad766bacf1513a9cf9dea485a9c02b95d8621a8/Detectors/ITSMFT/common/workflow/src/STFDecoderSpec.cxx#L110
> 
> 
> https://github.com/AliceO2Group/AliceO2/blob/3ad766bacf1513a9cf9dea485a9c02b95d8621a8/Detectors/ITSMFT/common/reconstruction/src/RawPixelDecoder.cxx#L125
> 
> Cheers,
> Ruben

HI @shahor02 , sorry, I am still a bit lost - conceptually.
1.) input : I have a raw data file on disk : in your examples there is no file input. Do I get raw input  from workflow or I can do both  and workflow recognise it ?
2.) Decoding: Then when I get raw data : my case is probably simple. Is there a general method which gives me TF (RDH + Payload) ?
In that case everything is quite simple. Why I need to buffer anything ?
Cheers, Roman.

shahor02:
Hi @lietava 

> 1.) input : I have a raw data file on disk : in your examples there is no file input. Do I get raw input from workflow or I can do both and workflow recognise it ?

The simulated raw data can be injected into the DPL workflow by the `o2-raw-file-reader`, see https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/Raw#raw-data-file-reader-workflow
For the example of the complex workflow starting with this reader see e.g. https://github.com/AliceO2Group/AliceO2/blob/dev/prodtests/full-system-test/dpl-workflow.sh, which is started from the full system test https://github.com/AliceO2Group/AliceO2/blob/dev/prodtests/full_system_test.sh

> 2.) Decoding: Then when I get raw data : my case is probably simple. Is there a general method which gives me TF (RDH + Payload) ?
> In that case everything is quite simple. Why I need to buffer anything ?

The examples I've referred to do exactly this: you extract from the input the messages containing your RDH + payload. E.g.
````
void CTPDecoder::run(ProcessingContext& pc) 
{
  std::vector<InputSpec> filter{InputSpec{"filter", ConcreteDataTypeMatcher{"CTP", "RAWDATA"}, Lifetime::Timeframe}};
  
  o2::framework::DPLRawParser parser(pc.inputs(), filter);
  for (auto it = parser.begin(); it != parser.end(); ++it) {
     auto rdh = it.get_if<o2::header::RAWDataHeader>();
     gsl::span<const uint8_t> payload(it.data(), it.size());
    // process RDH and payload
  }
````
allows you to iterate over all raw pages (a single link message may have multiple pages).
Alternatively, you can iterate over the input messages using
````
auto walker = o2::framework::InputRecordWalker(pc.inputs(), filter);
for (auto it = walker.begin(); it != walker.end(); ++it) {
  auto rawspan = o2::framework::DataRefUtils::as<const char>(it);
  /// here you get the gsl::span<const char> and should process as you would do with data in DMA,
  /// parsing the RDHs and navigating to next one yourself.
}
````

**Comment by lietava** [comment]
@shahor02 Thanks for example above. Please, check if all looks reasonable (dpl stuff) and how to it.

File: Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx [hunk 1]
```diff
@@ -0,0 +1,157 @@
...
+  using Lifetime = o2::framework::Lifetime;
+  //mOutputHWErrors.clear();
+  std::vector<InputSpec> filter{InputSpec{"filter", ConcreteDataTypeMatcher{"CTP", "RAWDATA"}, Lifetime::Timeframe}};
+  o2::framework::DPLRawParser parser(ctx.inputs(), filter);
+  uint32_t payloadCTP;
+  for (auto it = parser.begin(); it != parser.end(); ++it) {
+    auto rdh = it.get_if<o2::header::RAWDataHeader>();
+    auto triggerOrbit = o2::raw::RDHUtils::getTriggerOrbit(rdh);
+    auto linkCRU = o2::raw::RDHUtils::getLinkID(rdh); // 0 = IR, 1 = TCR
+    if (linkCRU == o2::ctp::CRULinkIDIntRec) {
```
shahor02, line 61: The common wisdom was that the FEEIDs should be used to identify different links rather than  linkID's (which, apparently, may change). Not sure if this is relevant from the CTP.

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
```diff
@@ -0,0 +1,157 @@
...
+      } else if (wordCount > 9) {
+        wordCount++;
+      } else if (wordCount == 9) {
+        wordCount++;
+        std::vector<gbtword80_t> diglets;
+        makeGBTWordInverse(diglets, gbtWord, remnant, size_gbt, payloadCTP);
+        // save digit in buffer recs
+        for (auto diglet : diglets) {
+          gbtword80_t pld = (diglet & pldmask);
+          if (pld.count() == 0)
```
shahor02: missing {} will be signaled by CI

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
```diff
@@ -0,0 +1,157 @@
...
+    uint32_t size_gbt = 0;
+    int wordCount = 0;
+    for (auto payloadWord : payload) {
+      if (wordCount == 15) {
+        wordCount = 0;
+      } else if (wordCount > 9) {
+        wordCount++;
+      } else if (wordCount == 9) {
+        wordCount++;
+        std::vector<gbtword80_t> diglets;
```
shahor02: Instead of having dynamic object creation in a tight loop it would be better to define it outside of all loops and call `clear` here.

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
```diff
@@ -0,0 +1,157 @@
...
+                                          o2::framework::Options{}};
+}
+// Inverse of Digits2Raw::makeGBTWord
+void RawToDigitConverterSpec::makeGBTWordInverse(std::vector<gbtword80_t> diglets, gbtword80_t& GBTWord, gbtword80_t& remnant, uint32_t& size_gbt, uint32_t Npld) const
+{
+  gbtword80_t diglet = remnant;
+  uint32_t i = 0;
+  while (i < (NGBT - Npld)) {
+    std::bitset<NGBT> masksize = 0;
+    for (uint32_t j = 0; j < (Npld - size_gbt); j++)
```
shahor02: {}

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
```diff
@@ -0,0 +1,157 @@
...
+  //outputs.emplace_back("CPV", "RAWHWERRORS", 0, o2::framework::Lifetime::Timeframe);
+
+  return o2::framework::DataProcessorSpec{"CTPRawToDigitConverterSpec",
+                                          inputs, // o2::framework::select("A:CTP/RAWDATA"),
+                                          outputs,
+                                          o2::framework::adaptFromTask<o2::ctp::reco_workflow::RawToDigitConverterSpec>(),
+                                          o2::framework::Options{}};
+}
+// Inverse of Digits2Raw::makeGBTWord
+void RawToDigitConverterSpec::makeGBTWordInverse(std::vector<gbtword80_t> diglets, gbtword80_t& GBTWord, gbtword80_t& remnant, uint32_t& size_gbt, uint32_t Npld) const
```
shahor02: I guess there should be `std::vector<gbtword80_t>& diglets` here.

Detectors/CTP/workflow/src/ReaderSpec.cxx
```diff
@@ -0,0 +1,133 @@
...
+// distributed under the terms of the GNU General Public License v3 (GPL
+// Version 3), copied verbatim in the file "COPYING".
+//
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include "CTPWorkflow/ReaderSpec.h"
```
shahor02: From the filename is it not clear that this is a digit reader, better name it explicitly as DigitReaderSpec.
Also, note that we started to move gradually all readers and writers to separate IO directories, in order to avoid circular dependencies in global workflows.
Perhaps you could place this class directly in the CTP/WorkflowIO .

Detectors/CTP/workflow/src/ReaderSpec.cxx
```diff
@@ -0,0 +1,133 @@
...
+      constexpr auto persistency = Lifetime::Timeframe;
+      o2::header::DataHeader::SubSpecificationType subSpec = 0;
+      if (propagateMC) {
+        //processAttributes->reader = std::make_shared<RootTreeReader>(treename.c_str(), // tree name
+        //filename.c_str(), // input file name
+        //nofEvents,        // number of entries to publish
+        //publishingMode,
+        //Output{"CTP", "DIGITS", subSpec, persistency},
+        //"CTPDigit", // name of data branch
+        //Output{"CTP", "DIGITSMCTR", subSpec, persistency},"CPVDigitMCTruth");
```
shahor02: please remove commented code

Detectors/CTP/workflow/src/RecoWorkflow.cxx
```diff
@@ -0,0 +1,100 @@
...
+#include "SimulationDataFormat/MCTruthContainer.h"
+
+using namespace o2::dataformats;
+
+namespace o2
+{
+
+template <typename T>
+using BranchDefinition = framework::MakeRootTreeWriterSpec::BranchDefinition<T>;
+
```
shahor02: why is this here?

Detectors/CTP/workflow/src/WriterSpec.cxx
```diff
@@ -0,0 +1,61 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
```
shahor02: Also this one better to rename to DigitWriterSpec and to have it in the WorkflowIO

Detectors/CTP/workflow/src/ctp-reco-workflow.cxx
```diff
@@ -0,0 +1,63 @@
...
+// or submit itself to any jurisdiction.
+
+/// @file   ctp-reco-workflow.cxx
+/// @author RL from CPV example
+/// @brief  Basic DPL workflow for CTP reconstruction starting from digits
+#include "Framework/WorkflowSpec.h"
+#include "Framework/ConfigParamSpec.h"
+#include "CTPWorkflow/RecoWorkflow.h"
+#include "Algorithm/RangeTokenizer.h"
+#include "CommonUtils/ConfigurableParam.h"
```
shahor02, line 19: In order to create a correct DPL `DataHeadr.firstTForbit` in the root file digits driven mode , the workflows should have the lines like https://github.com/AliceO2Group/AliceO2/blob/59df0714e1705fb6e3d55b8f0bda869f2a2b866f/Detectors/ITSMFT/ITS/workflow/src/its-reco-workflow.cxx#L15 
https://github.com/AliceO2Group/AliceO2/blob/59df0714e1705fb6e3d55b8f0bda869f2a2b866f/Detectors/ITSMFT/ITS/workflow/src/its-reco-workflow.cxx#L40
https://github.com/AliceO2Group/AliceO2/blob/59df0714e1705fb6e3d55b8f0bda869f2a2b866f/Detectors/ITSMFT/ITS/workflow/src/its-reco-workflow.cxx#L72-L73

So, add here `#include "DetectorsRaw/HBFUtilsInitializer.h"`

Detectors/CTP/workflow/src/ctp-reco-workflow.cxx
```diff
@@ -0,0 +1,63 @@
...
+// including Framework/runDataProcessing
+void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
+{
+  std::vector<o2::framework::ConfigParamSpec> options{
+    {"input-type", o2::framework::VariantType::String, "digits", {"hits, digits, raw, clusters"}},
+    {"output-type", o2::framework::VariantType::String, "clusters", {"digits, clusters"}},
+    {"disable-mc", o2::framework::VariantType::Bool, false, {"disable sending of MC information"}},
+    {"disable-root-input", o2::framework::VariantType::Bool, false, {"disable root-files input reader"}},
+    {"disable-root-output", o2::framework::VariantType::Bool, false, {"disable root-files output writer"}},
+    {"configKeyValues", o2::framework::VariantType::String, "", {"Semicolon separated key=value strings ..."}}};
```
shahor02: add here `o2::raw::HBFUtilsInitializer::addConfigOption(options);`

Detectors/CTP/workflow/src/ctp-reco-workflow.cxx
```diff
@@ -0,0 +1,63 @@
...
+{
+  //
+  // Update the (declared) parameters if changed from the command line
+  o2::conf::ConfigurableParam::updateFromString(cfgc.options().get<std::string>("configKeyValues"));
+
+  return o2::ctp::reco_workflow::getWorkflow(cfgc.options().get<bool>("disable-root-input"),
+                                             cfgc.options().get<bool>("disable-root-output"),
+                                             !cfgc.options().get<bool>("disable-mc"),       //
+                                             cfgc.options().get<std::string>("input-type"), //
+                                             cfgc.options().get<std::string>("output-type") //
```
shahor02: change this to 
````
auto wf = o2::ctp::reco_workflow::getWorkflow( ... );

// configure dpl timer to inject correct firstTFOrbit: start from the 1st orbit of TF containing 1st sampled orbit
o2::raw::HBFUtilsInitializer <span class="pl-smi">hbfIni</span>(configcontext, wf);
return std::move(wf);
````

**Review by shahor02 — CHANGES_REQUESTED** [review]
Hi @lietava 

The dpl data fetching part looks ok, see few comments below.

File: Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx (hunk 1 above)
Comment (lietava) [reply Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx:61]: I see the point. But that would only work if our firmware would add FEEID to RDH, right ?

File: Detectors/CTP/workflow/src/RecoWorkflow.cxx (hunk 8 above)
Comment (lietava) [reply Detectors/CTP/workflow/src/RecoWorkflow.cxx]: remnant of copy and paste. removed.

File: Detectors/CTP/workflow/src/WriterSpec.cxx (hunk 9 above)
Comment (lietava) [reply Detectors/CTP/workflow/src/WriterSpec.cxx]: I rename. I do not understand second part of sentence: " to have it in the WorkflowIO" ?

File: Detectors/CTP/workflow/src/ctp-reco-workflow.cxx (hunk 10 above)
Comment (lietava) [reply Detectors/CTP/workflow/src/ctp-reco-workflow.cxx:19]: I added include and also equivalents of lines 40 and 73 ?

Comment (lietava) [reply Detectors/CTP/workflow/src/ctp-reco-workflow.cxx:19]: ok, your next two comments resolving this.

**Comment by lietava** [comment]
@shahor02 1.) using fee instead of link - I'll do it later. 2.) I am not sure I fully understand your comment on (Digit)WriterSpec - please see above. All other comments implemented, I think/ Cheers, R.

File: Detectors/CTP/workflow/src/WriterSpec.cxx (hunk 9 above)
Comment (shahor02) [reply Detectors/CTP/workflow/src/WriterSpec.cxx]: I mean put reader and writer workflows in a separate library, depending only on the data formats. E.g. this can be WorkflowIO directory creating a library of the same name. Or, see how TRD did this, creating `io` directory in the `O2/Detectors/TRD/workflow`

File: Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx (hunk 1 above)
Comment (shahor02) [reply Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx:61]: Afaik, the FEEID is supposed to be unique and fixed in all cases.

**Comment by lietava** [comment]
Hi @shahor02 , I created workflowIO on the same level as workflow. I was not sure where to put CTPDigitWriterSpec - kept it in workflow.
Anything else apart links (later) and missing STF (can do in this PR or next) ?
Cheers, Roman.

File: Detectors/CTP/workflowIO/src/ReaderSpec.cxx [hunk 13]
```diff
@@ -0,0 +1,133 @@
...
+using namespace o2::framework;
+
+namespace o2
+{
+
+namespace ctp
+{
+
+template <typename T>
+using BranchDefinition = MakeRootTreeWriterSpec::BranchDefinition<T>;
```
shahor02: you still have this stray lines.

Detectors/CTP/workflowIO/src/ReaderSpec.cxx
```diff
@@ -0,0 +1,133 @@
...
+// or submit itself to any jurisdiction.
+
+#include "CTPWorkflowIO/ReaderSpec.h"
+#include "Framework/ConfigParamRegistry.h"
+#include "Framework/ControlService.h"
+#include "Headers/DataHeader.h"
+#include "DPLUtils/RootTreeReader.h"
+#include "DPLUtils/MakeRootTreeWriterSpec.h"
+#include "Framework/DataSpecUtils.h"
+#include "DetectorsCommonDataFormats/NameConf.h"
```
shahor02: I thought you will rename this to DigitReaderSpec, to be symmetric with Writer.

shahor02 (commented):
Hi @lietava 

> I was not sure where to put CTPDigitWriterSpec - kept it in workflow.

Just noticed: why do you need both CTPDigitWriterSpec and DigitWriterSpec? They are doing exactly the same, except that in the CTPDigitWriterSpec you already have an option of indicating that it is called from the digitization or decoding. 
I would eliminate the DigitWriterSpec and move the CTPDigitWriterSpec to workflowIO.

Apart from that, the code seems to be complete, but did you try to decode simulated raw data and make sure that the decoded digits exactly match to simulated ones?

Once this is validated and the PR is merged I will implement the CTF creation and add the CTP to the FST.

lietava:
Hi @shahor02 , I did all modifications you suggested above. 
Also I tried to implement no lost TF trick - please, check.
If ok - I can test rawtodigit -  so I run rawtodigit and then you have some tool to check  ?
Cheers, R.

Detectors/CTP/workflow/include/CTPWorkflow/RawToDigitConverterSpec.h
```diff
@@ -0,0 +1,67 @@
...
+/// \class RawToDigitConverterSpec
+/// \brief Coverter task for Raw data to CTP digits
+/// \author Roman Lietava from CPV example
+///
+class RawToDigitConverterSpec : public framework::Task
+{
+ public:
+  /// \brief Constructor
+  /// \param propagateMC If true the MCTruthContainer is propagated to the output
+  RawToDigitConverterSpec() : framework::Task(){};
```
shahor02: I think CI will require to change this to `RawToDigitConverterSpec() = default`

Detectors/CTP/workflow/include/CTPWorkflow/RawToDigitConverterSpec.h
```diff
@@ -0,0 +1,67 @@
...
+  void makeGBTWordInverse(std::vector<gbtword80_t>& diglets, gbtword80_t& GBTWord, gbtword80_t& remnant, uint32_t& size_gbt, uint32_t Npld) const;
+
+ protected:
+ private:
+  std::vector<CTPDigit> mOutputDigits;
+};
+
+/// \brief Creating DataProcessorSpec for the CTP
+///
+o2::framework::DataProcessorSpec getRawToDigitConverterSpec();
```
shahor02: This should be removed

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
```diff
@@ -0,0 +1,195 @@
...
+  // if we see requested data type input with 0xDEADBEEF subspec and 0 payload this means that the "delayed message"
+  // mechanism created it in absence of real data from upstream. Processor should send empty output to not block the workflow
+  {
+    std::vector<InputSpec> dummy{InputSpec{"dummy", o2::framework::ConcreteDataMatcher{"CTP", "RAWDATA", 0xDEADBEEF}}};
+    for (const auto& ref : o2::framework::InputRecordWalker(inputs, dummy)) {
+      const auto dh = o2::framework::DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
+      if (dh->payloadSize == 0) {
+        LOGP(WARNING, "Found input [{}/{}/{:#x}] TF#{} 1st_orbit:{} Payload {} : assuming no payload for all links in this TF",
+             dh->dataOrigin.str, dh->dataDescription.str, dh->subSpecification, dh->tfCounter, dh->firstTForbit, dh->payloadSize);
+        return;
```
shahor02, line 51: See this comment: https://github.com/AliceO2Group/AliceO2/pull/6457#discussion_r653807498
You should push an empty output in this case.

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
```diff
@@ -0,0 +1,195 @@
...
+  std::vector<o2::framework::OutputSpec> outputs;
+  outputs.emplace_back("CTP", "DIGITS", 0, o2::framework::Lifetime::Timeframe);
+  //outputs.emplace_back("CPV", "RAWHWERRORS", 0, o2::framework::Lifetime::Timeframe);
+
+  return o2::framework::DataProcessorSpec{"CTPRawToDigitConverterSpec",
+                                          inputs, // o2::framework::select("A:CTP/RAWDATA"),
+                                          outputs,
+                                          o2::framework::adaptFromTask<o2::ctp::reco_workflow::RawToDigitConverterSpec>(),
+                                          o2::framework::Options{}};
+}
```
shahor02: not needed

shahor02 (commented):
Hi @lietava 
Looks fine but see some comments below. For the validation: I don't have tools for that, one should simply create raw data from simulated digits, then run your new decoding, write digits to other file and make sure that the content of both digit files are the same.

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
lietava: added output, pls, check

lietava:
HI @shahor02 , I did fixes you suggested, please,  check. For validation I involve Marek. Merging before or after validation - up to you.

lietava:
@shahor02 . I tried quick check and got:
[O2/latest-ctpdev-o2] ~/o2tests $> o2-ctp-reco-workflow -b
[ERROR] invalid workflow in o2-ctp-reco-workflow: No matching output found for <matcher query: (and origin:CTP (and startTime:$0  description:RAWDATA ))> as requested by data processor "CTP-RawStreamDecoder". Candidates:
-CTP/DIGITS/0

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
```diff
@@ -0,0 +1,179 @@
...
+          }
+        }
+        gbtWord = 0;
+      } else {
+        gbtWord |= payloadWord >> (wordCount * 8);
+        wordCount++;
+      }
+    }
+  }
+  mOutputDigits.clear();
```
shahor02: you already have `mOutputDigits.clear();` in the beginning.

shahor02 (commented):
Hi @lietava ,

As I wrote before, you should use `o2-raw-file-reader-workflow` to inject the raw data into the DPL : https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/Raw#raw-data-file-reader-workflow.
This means that having generated raw data, you should do e.g.
````
o2-raw-file-reader-workflow --input-conf raw/CTP/CTPraw.cfg | o2-ctp-reco-workflow 
````

lietava:
> o2-raw-file-reader-workflow --input-conf raw/CTP/CTPraw.cfg | o2-ctp-reco-workflow

lietava:
> > o2-raw-file-reader-workflow --input-conf raw/CTP/CTPraw.cfg | o2-ctp-reco-workflow

@shahor02  I did enter enviroment ...

[O2/latest-ctpdev-o2] ~/o2tests $> o2-raw-file-reader-workflow --input-conf CTPraw.cfg | o2-ctp-reco-workflow 
Error 65542: GLX: GLX version 1.3 is required
o2-ctp-reco-workflow: /home/rl/alice/sw/SOURCES/GLFW/3.3.2/3.3.2/src/input.c:863: glfwSetMouseButtonCallback: Assertion `window != ((void *)0)' failed.
Aborted

lietava:
> > > o2-raw-file-reader-workflow --input-conf raw/CTP/CTPraw.cfg | o2-ctp-reco-workflow
> 
> @shahor02 I did enter enviroment ...
> 
> [O2/latest-ctpdev-o2] ~/o2tests $> o2-raw-file-reader-workflow --input-conf CTPraw.cfg | o2-ctp-reco-workflow
> Error 65542: GLX: GLX version 1.3 is required
> o2-ctp-reco-workflow: /home/rl/alice/sw/SOURCES/GLFW/3.3.2/3.3.2/src/input.c:863: glfwSetMouseButtonCallback: Assertion `window != ((void *)0)' failed.
> Aborted

@shahor02 Sorry, it was just gui. Using -b  makes it run.

lietava:
@shahor02 . Thanks. It produces some digits now. As I said I suggest Marek with me starts to play with it.
Shall we merge to dev before ?
R.

shahor02:
If it does not break the stability of the code, we can merge it, though is there a problem to validate it before?
In any case, before merging you should remove the Draft tag from the PR and it should pass the CI.

lietava:
@shahor02 ok, we validate first.

lietava:
@shahor02 I was trying today to rebase dev to ctpdev  but it fails as there is renamed file I guess.
The validation may take weeks as I am not doing it now with full capacity and Marek needs some learning.
So my question is if it is not worth to merge at this stage just to reduce divergence. But if it is ok to stay like this I am fine with it. Decision up to you. Cheers, R.

shahor02:
Hi @lietava 
I would not merge the code which is in draft mode. So, remove the draft tag so that CI could start.

For the rebasing: yes, it complains on the conflict between the CTPDigitWriterSpec modified in the dev and removed in your branch. You should do (after updating your dev):
````
git rebase dev
# will complain on Detectors/CTP/workflow/src/CTPDigitWriterSpec.cxx, Detectors/CTP/workflow/include/CTPWorkflow/CTPDigitWriterSpec.h

git rm Detectors/CTP/workflow/src/CTPDigitWriterSpec.cxx Detectors/CTP/workflow/include/CTPWorkflow/CTPDigitWriterSpec.h
git rebase --continue
````

shahor02:
@lietava could you rebase and resolve the conflicts? The tests will not start w/o that.

lietava:
Conflict resolved.

lietava:
Looking in log build/AliceO2/O2/o2/macOS : are these erros because of CTP ?

shahor02:
@lietava no, macOS is broken in all PRs.

lietava:
I fixed license issues and now I do not see any CTP related issues in two failed logs. Please, let me know I am wrong.

shahor02:
right, they are not related. What about validation?

lietava:
@shahor02 waiting for Marek. I am busy with other things but may reorder. Is it urgent ? R.

shahor02:
no, just to understand if I shall merge this or not. Since this PR does not break anything, I could merge it.

lietava:
@shahor02 , please, merge. FYI: I did (de)coder checks outside O2 so bug maybe somewhere in creating HBF. Then I also like to involve Marek - other purpose of exercise. In parallel I start to add counters infrastructure. Cheers, R.

PR #6417, opened 2021-06-14
[OCTRL-496] AliECS dump: allow to preserve raw FMQ channels

knopers8:
This will allow us to use raw FairMQ channels (such as in proxies) with originally specified addresses in AliECS worfklow templates. This is required for workflows which run partially with ODC and partially with AliECS. The first does not have automatic port/machine assignment, thus we have to set them manually.


knopers8:
@ktf could you please have a look?

PR #6419, opened 2021-06-14
DPL: Allow to add Labels as command arguments

knopers8:
With this we can pass `DataProcessorLabels` to `DataProcessor` at runtime.

This is will help with adding AliECS-related labels (such as in https://github.com/AliceO2Group/AliceO2/pull/6417 ) at workflow template generation time. By that, we will be able enforce different template output for DPL proxies which connect AliECS- and ODC-controlled parts of O2 (e.g. calibration), and different for proxies which send data within AliECS-controlled environment (e.g. output proxy to STFS).

### DIFF
--- a/Framework/Core/src/runDataProcessing.cxx
+++ b/Framework/Core/src/runDataProcessing.cxx
@@ -2081,6 +2081,71 @@ void overridePipeline(ConfigContext& ctx, WorkflowSpec& workflow)
   }
 }
 
+void overrideLabels(ConfigContext& ctx, WorkflowSpec& workflow)
+{
+  struct LabelsSpec {
+    std::string_view matcher;
+    std::vector<std::string> labels;
+  };
+  std::vector<LabelsSpec> specs;
+
+  auto labelsString = ctx.options().get<std::string>("labels");
+  if (labelsString.empty()) {
+    return;
+  }
+  std::string_view sv{labelsString};
+
+  size_t specStart = 0;
+  size_t specEnd = 0;
+  constexpr char specDelim = ',';
+  constexpr char labelDelim = ':';
+  do {
+    specEnd = sv.find(specDelim, specStart);
+    auto token = sv.substr(specStart, specEnd == std::string_view::npos ? std::string_view::npos : specEnd - specStart);
+    if (token.empty()) {
+      throw std::runtime_error("bad labels definition. Syntax <processor>:<label>[:<label>][,<processor>:<label>[:<label>]");
+    }
+
+    size_t labelDelimPos = token.find(labelDelim);
+    if (labelDelimPos == 0 || labelDelimPos == std::string_view::npos) {
+      throw std::runtime_error("bad labels definition. Syntax <processor>:<label>[:<label>][,<processor>:<label>[:<label>]");
+    }
+    LabelsSpec spec{token.substr(0, labelDelimPos)};
+
+    size_t labelEnd = labelDelimPos + 1;
+    do {
+      size_t labelStart = labelDelimPos + 1;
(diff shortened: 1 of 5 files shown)


Framework/Core/test/test_OverrideLabels.cxx
```diff
@@ -0,0 +1,93 @@
...
+
+#define BOOST_TEST_MODULE Test Framework OverrideLabels
+#define BOOST_TEST_MAIN
+#define BOOST_TEST_DYN_LINK
+
+// We prevent runDataProcessing from starting a workflow
+#define main anything_else_than_main
+#include "Framework/runDataProcessing.h"
+#undef main
+o2::framework::WorkflowSpec defineDataProcessing(o2::framework::ConfigContext const&) { return {}; }
```
knopers8, line 20: @ktf I am not very happy about this, but I don't see any other workaround here (I need to test `overrideLabels` in this header). Would you have any other idea for that?

knopers8:
@ktf could you please have a look?

Framework/Core/test/test_OverrideLabels.cxx
ktf: We could move most of the header in runDataProcessing.h to runDataProcessing.inc and keep in runDataProcessing.h only the main. I will do it in a separate PR.

knopers8: thank you

ktf: For the moment let's go ahead with this, though. Since there is a separate PR which would have conflicts.

Framework/Core/src/runDataProcessing.cxx
```diff
@@ -1951,6 +1951,71 @@ void overridePipeline(ConfigContext& ctx, WorkflowSpec& workflow)
   }
 }
 
+void overrideLabels(ConfigContext& ctx, WorkflowSpec& workflow)
+{
+  struct LabelsSpec {
+    std::string_view matcher;
+    std::vector<std::string_view> labels;
```
ktf: Why a vector of `string_views` and not of `strings`?

knopers8: I hoped for less string copies during `vector` (re)allocation, both in `LabelsSpec.labels` and `std::vector<LabelsSpec>`.

ktf: sure, but string_view does not have any ownership on the data, where is it actually stored?

knopers8: Here is the original string:
```
auto labelsString = ctx.options().get<std::string>("labels")
```
It lives long enough before the labels are actually copied into `DataProcessorSpec::labels`.

Though I must admit I also wanted to exercise `string_view` a bit, so this may look like excessive optimization.

ktf: I would avoid it, yes. std::string is probably fine.

knopers8: sure, i'll change it.

ktf (dismissed):

ktf:
FIne by me if it passes tests.

sawenzel:
@knopers8 : There is a merge conflict here. Can you please fix it?

knopers8:
Thanks for the heads up, I actually thought this one was merged already.

ktf:
@knopers8 can you have a look on why it fails? i am away this week.

knopers8:
I will. I forgot to have a look the last time...

knopers8:
@ktf finally green.

PR #6421, opened 2021-06-14
[MID] DPL workflow to send MID tracks read from a root file

pillot:


Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
```diff
@@ -0,0 +1,131 @@
...
+
+ private:
+  /// accumulate the data
+  bool accumulate(std::string_view name, char* data)
+  {
+    if (name == "MIDTrackROF") {
+
+      auto rofs = reinterpret_cast<std::vector<ROFRecord>*>(data);
+
+      // accumulated the ROFs, shifting the track indexing accordingly
```
aphecetche: accumulated or accumulate (present tense) ?

Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
```diff
@@ -0,0 +1,131 @@
...
+                                                 Output{"MID", "TRACKS", 0, Lifetime::Timeframe}, "MIDTrack"},
+                                               RootTreeReader::BranchDefinition<std::vector<ROFRecord>>{
+                                                 Output{"MID", "TRACKSROF", 0, Lifetime::Timeframe}, "MIDTrackROF"},
+                                               &mAccumulator);
+  }
+
+  /// process the next entry
+  void run(ProcessingContext& pc)
+  {
+    if (mReader->next()) {
```
dstocco, line 55: To avoid useless if...else, can't you just use:
while(mReader->next())

Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
```diff
@@ -0,0 +1,131 @@
...
+class TrackSamplerTask
+{
+ public:
+  /// prepare the reader
+  void init(InitContext& ic)
+  {
+    auto inputFileName = ic.options().get<std::string>("infile");
+    mMinNumberOfROFsPerTF = ic.options().get<int>("repack-rofs");
+
+    mReader = std::make_shared<RootTreeReader>("midreco", inputFileName.c_str(), -1,
```
aphecetche: just curious : why must it be a shared pointer ?

aphecetche (dismissed):
Looks good to me

Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
```diff
@@ -0,0 +1,131 @@
...
+
+  /// process the next entry
+  void run(ProcessingContext& pc)
+  {
+    if (mReader->next()) {
+      (*mReader)(pc);
+      if (mROFs.size() >= mMinNumberOfROFsPerTF) {
+        publish(pc.outputs());
+      }
+    } else {
```
dstocco, line 60: Here you just close the while

Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
```diff
@@ -0,0 +1,131 @@
...
+      if (mROFs.size() > 0) {
+        publish(pc.outputs());
+      }
+      pc.services().get<ControlService>().endOfStream();
+    }
+  }
```
dstocco, line 66: and the last parenthesis is not needed anymore...

Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
pillot: accumutate indeed, this is typo.

Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
pillot: I followed examples I found here and there, but it's true that the way I am defining the algorithm spec, with a task instead of using lambda, I do not need to pass the reader to the lambda so I could probably use a unique_ptr instead.

Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
pillot: The test is performed for every entry in the while loop anyway. Beside, I don't want to send everything in a single pair of messages, so I would need to loop only over the required number of entries to accumulate enough ROF and send the 2 messages, then let the DPL call the run() function again for the next TF. And add a test when there is nothing left to read to stop the DPL loop.

Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
pillot: This is needed to stop the DPL loop

Detectors/MUON/MID/Workflow/src/tracks-reader-workflow.cxx
dstocco: ok, I see

aphecetche (approved):

PR #6423, opened 2021-06-14
generalize TOF matcher

noferini:


noferini:
TOF matcher generalized.

o2-tof-matcher-global -b
o2-tof-matcher-tpc -b 
still working but they will be removed in future


o2-tof-matcher-workflow -b
the new dpl:
- general writer still missing (to be implemented)
- TPC, ITC-TPC tracks allowed (easy to extend to TPC-TRD and ITS-TPC-TRD)
- high purity selection cuts implemented (MC info not managed), flag to be added

WIP

DataFormats/Reconstruction/include/ReconstructionDataFormats/MatchInfoTOF.h
```diff
@@ -58,8 +67,9 @@ class MatchInfoTOF
   evGIdx mEvIdxTrack;                ///< EvIdx for track (first: ev index; second: track global index)
   float mZatTOF = 0.0;               ///< Z position at  TOF
   float mDeltaT = 0.0;               ///< tTOF - TPC (microsec)
+  TrackType mTrackType;              ///< track type (TPC, ITSTPC, TPCTRD, ITSTPCTRD)
 
-  ClassDefNV(MatchInfoTOF, 2);
+  ClassDefNV(MatchInfoTOF, 3);
```
shahor02: not sure this is needed: the matches to different types should stay in different outputs, see https://docs.google.com/presentation/d/10llBSPh_cKY4CYxvwg2x4bD7GAWzlKRKGA6GrW2vNA8/edit#slide=id.p
flagged by corresponding DataDescriptions (or SubSpecs if we switch to them).

Detectors/GlobalTracking/include/GlobalTracking/MatchTOF.h
```diff
@@ -243,110 +80,20 @@ class MatchTOF
...
-  std::vector<o2::MCCompLabel> mOutTOFLabels;                ///< TOF label of matched tracks
-
-  int mNumOfTracks;                     // number of tracks to be matched
-  std::vector<int> mMatchedTracksIndex; // vector of indexes of the tracks to be matched
-  int mNumOfClusters;                   // number of clusters to be matched
-  int* mMatchedClustersIndex = nullptr; //[mNumOfClusters]
-
   std::string mTracksBranchName = "TPCITS";                       ///< name of branch containing input matched tracks
   std::string mTPCTracksBranchName = "TPCTracks";                 ///< name of branch containing actual TPC tracks
   std::string mTPCMCTruthBranchName = "MatchMCTruth";             ///< name of branch containing TPC labels
```
shahor02: Do we really need this? I think TOF is the only detector which still maintains the no-DPL version of the reconstruction. There will be some effort for maintenance but what is the profit?

Detectors/GlobalTracking/include/GlobalTracking/MatchTOFBase.h
```diff
@@ -0,0 +1,305 @@
...
+
+  bool mIsITSTPCused = false;
+  bool mIsTPCused = false;
+  bool mIsTPCTRDused = false;
+  bool mIsITSTPCTRDused = false;
+  bool mSetHighPurity = false;
+
+  ///>>>------ these are input arrays which should not be modified by the matching code
+  //           since this info is provided by external device
+  gsl::span<const o2::dataformats::TrackTPCITS> mITSTPCTracksArrayInp;  ///< input tracks
```
shahor02: also, by giving up non-DPL version you can also get rid of all these spans, since they will be available from the RecoContainer.

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -65,27 +69,33 @@ void MatchTOF::run()
...
+    mMatchedTracks[i].clear();
+    mTracksWork[i].clear();
+    mOutTOFLabels[i].clear();
+  }
+
   if (mIsworkflowON) {
     LOG(DEBUG) << "Number of entries in track tree = " << mCurrTracksTreeEntry;
 
-    if (mIsITSused) {
+    if (mIsITSTPCused) {
```
shahor02, line 91: We will need to cover the combinations of `{TPC and TPC-TRD}` seeds in "strict" mode and `{TPC, ITS-TPC, ITS-TPC-TRD, TPC,TRD}` for "relaxed" mode. And this should account for the redundancy: i.e. if TPC track is already used in some global track, it should not be considered anymore.
All this is naturally done via RecoContainer, which also takes care of different inputs and will absorb most of future changes. I would strongly suggest to switch to it. If you want, we can do this together.

Detectors/GlobalTrackingWorkflow/src/TOFMatcherSpec.cxx
```diff
@@ -104,6 +104,16 @@ void TOFMatcherSpec::run(ProcessingContext& pc)
...
 
+  LOG(INFO) << "isTrackSourceLoaded: TPC -> " << recoData.isTrackSourceLoaded(o2::dataformats::GlobalTrackID::Source::TPC);
+  LOG(INFO) << "isTrackSourceLoaded: ITSTPC -> " << recoData.isTrackSourceLoaded(o2::dataformats::GlobalTrackID::Source::ITSTPC);
+  LOG(INFO) << "isTrackSourceLoaded: TPCTRD -> " << recoData.isTrackSourceLoaded(o2::dataformats::GlobalTrackID::Source::TPCTRD);
+  LOG(INFO) << "isTrackSourceLoaded: ITSTPCTRD -> " << recoData.isTrackSourceLoaded(o2::dataformats::GlobalTrackID::Source::ITSTPCTRD);
+
+  bool isTPCused = recoData.isTrackSourceLoaded(o2::dataformats::GlobalTrackID::Source::TPC);
+  bool isITSTPCused = recoData.isTrackSourceLoaded(o2::dataformats::GlobalTrackID::Source::ITSTPC);
+  bool isTPCTRDused = recoData.isTrackSourceLoaded(o2::dataformats::GlobalTrackID::Source::TPCTRD);
+  bool isITSTPCTRDused = recoData.isTrackSourceLoaded(o2::dataformats::GlobalTrackID::Source::ITSTPCTRD);
```
shahor02, line 109: Also, with the RecoContainer you will have the same uniform interface for input types as other matches, e.g.:
https://github.com/AliceO2Group/AliceO2/blob/bd7be7b02e6ccc8c5545e64021a9af46cdb7bf35/Detectors/GlobalTrackingWorkflow/src/tpcits-match-workflow.cxx#L36

DataFormats/Reconstruction/include/ReconstructionDataFormats/MatchInfoTOF.h
noferini: I added
  TrackType mTrackType;              ///< track type (TPC, ITSTPC, TPCTRD, ITSTPCTRD)
as data member and then I increased the versioning.
I removed that member from streamer (//!) and restored previous version.
Is it ok?

Detectors/GlobalTracking/include/GlobalTracking/MatchTOF.h
noferini: I agree. I kept it to be consistent with the past.
If you also agree to remove it I am happy

Detectors/GlobalTrackingWorkflow/src/TOFMatcherSpec.cxx
noferini: About these last three comments.
I expected such comments anf this is the reason I didn't yet implemented the new writer and some features.
I am available to discuss how to move to RecoContainer directly in the MatchTOF class

DataFormats/Reconstruction/include/ReconstructionDataFormats/MatchInfoTOF.h
shahor02: But why do you need them? If this is just to keep the matches of different type in the single container during the reco: they will need to be anyway redistributed to different containers in end of the TF reco, so a copy will be needed.
So, at reco time you can simply use a class derived from MatchInfoTOF by adding this data member.

Detectors/GlobalTracking/include/GlobalTracking/MatchTOF.h
shahor02: Yes, would be good to get rid of this legacy code.

noferini: done

DataFormats/Reconstruction/include/ReconstructionDataFormats/MatchInfoTOF.h
noferini: Ok. I derived a class for that.

noferini:
I completed the job.

Old matching workflow removed.
Now tof matching is done only in 
o2-tof-matcher-workflow
fully compatible with general framework

and it accept for the moment TPC and ITS-TPC tracks.

    //    mMatcher.makeConstrainedTPCTracks(tracksTPCTOF);  // why is failing???? to be investigated
disabled in TOFMatcherSpec 

it doesn't work properly (to be checked but in a next PR)

extra features (e.g. enabling strong selection cuts) to be added 


This is the typical output
[264708:tof-matcher]: [13:53:46][INFO] Shifting Z for 20 matched TPC tracks according to TOF time info
[264709:TOFCalibWriter]: [13:53:46][INFO] RECEIVED MATCHED SIZE 43
                         43 matched tracks used for calibration written in a single file (stream)
[264710:TOFMatchedWriter]: [13:53:46][INFO] RECEIVED MATCHED SIZE 23
[264711:TOFMatchedWriter_TPC]: [13:53:46][INFO] RECEIVED MATCHED SIZE 20
                           23 matched with ITS-TPC + 20 matched with TPConly
[264710:TOFMatchedWriter]: [13:53:46][INFO] TOF LABELS GOT 23 LABELS 
[264708:tof-matcher]: [13:53:46][INFO] Correctly handshaken websocket connection.
[264708:tof-matcher]: [13:53:46][INFO] TOF matching total timing: Cpu: 4.600e-01 Real: 4.674e-01 s in 1 slots
[264711:TOFMatchedWriter_TPC]: [13:53:46][INFO] TOF LABELS GOT 20 LABELS 
                     written in two different files (streams)

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -40,416 +40,347 @@
...
+
+      if (mMCTruthON) {
+        mTracksLblWork[it].clear();
+        mTracksLblWork[it].reserve(mNumOfTracks[it]);
+      }
+      for (int sec = o2::constants::math::NSectors; sec--;) {
+        mTracksSectIndexCache[it][sec].clear();
+        mTracksSectIndexCache[it][sec].reserve(100 + 1.2 * mNumOfTracks[it] / o2::constants::math::NSectors);
+      }
+    }
```
shahor02, line 175: you could actually reserve less slots if you would count the numbers of different track types fetched by the `creator`, e.g. you cannot have  `mRecoCont->getTPCTracks().size()` TPC matches, since part of TPC tracks will come via ITS-TPC or TPC-TRD etc.

BTW, what prevents processing the TPC-TRD and ITS-TPC-TRD tracks at once?

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -40,416 +40,347 @@
...
-    }
+  // create working copy of track param
+  mTracksWork[trkType::ITSTPC].emplace_back(std::make_pair(_tr.getParamOut(), _tr.getTimeMUS()));
+  mLTinfos[trkType::ITSTPC].emplace_back(_tr.getLTIntegralOut());
+  auto& trc = mTracksWork[trkType::ITSTPC].back().first; // with this we take the TPCITS track propagated to the vertex
+  auto& intLT = mLTinfos[trkType::ITSTPC].back();        // we get the integrated length from TPC-ITC outward propagation
 
-  } else {
-    LOG(INFO) << "Output tree is not attached, matched tracks will not be stored";
+  if (trc.getX() < o2::constants::geom::XTPCOuterRef - 1.) { // tpc-its track outward propagation did not reach outer ref.radius, skip this track
```
shahor02, line 253: Why are you doing this check after adding the track? You could apply this trivial cut directly in the creator and simply not call the addXXXSeed at all if the track does not reach the TPC outer ref.

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -40,416 +40,347 @@
...
+  timeInfo.setTimeStampError((_tr.getDeltaTBwd() + 5) * mTPCTBinMUS + extraErr);
+  mSideTPC.push_back(_tr.hasASideClustersOnly() ? 1 : (_tr.hasCSideClustersOnly() ? -1 : 0));
+  mExtraTPCFwdTime.push_back((_tr.getDeltaTFwd() + 5) * mTPCTBinMUS + extraErr);
+
+  o2::track::TrackLTIntegral intLT0; //mTPCTracksWork.back().getLTIntegralOut(); // we get the integrated length from TPC-ITC outward propagation
+  mTracksWork[trkType::TPC].emplace_back(std::make_pair(_tr.getOuterParam(), timeInfo));
+  auto& trc = mTracksWork[trkType::TPC].back().first;
+  auto& intLT = mLTinfos[trkType::TPC].emplace_back(intLT0);
+
+  if (_tr.getNClusters() < nclustersMin) {
```
shahor02, line 312: same comment: all these checks can be done in the `creator`, avoiding unnecessary initializations.

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -40,416 +40,347 @@
...
+  trc.getXYZGlo(globalPos);
 
-  mCurrTracksTreeEntry = -1;
-  mCurrTOFClustersTreeEntry = -1;
+  mTracksSectIndexCache[trkType::TPC][o2::math_utils::angle2Sector(TMath::ATan2(globalPos[1], globalPos[0]))].push_back(it);
+  //delete trc; // Check: is this needed?
 }
-
 //______________________________________________
 bool MatchTOF::prepareTracks()
```
shahor02, line 346: This method is not used anymore, right? Why do you keep it?

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -523,39 +454,32 @@ bool MatchTOF::prepareTracks()
 bool MatchTOF::prepareTPCTracks()
```
shahor02, line 454: This one also seems to be unused (overridden by `prepareTPCData`

shahor02 (commented):
Thanks @noferini , please see some comments below.

What is the problem with `makeConstrainedTPCTracks` refit? W/o it the match cannot be used in the downstream workflows.

I see that you have eliminated old workflows like `tof-matcher-tpc.cxx` and `tof-matcher-global.cxx`, substituting by unique `tof-reco-workflow`.  I fully support this, old workflows currently used in different scripts: 

https://github.com/AliceO2Group/AliceO2/blob/269060e5b9a3bfdb32780e10ee6a1321edcf72c1/prodtests/full-system-test/dpl-workflow.sh#L133
https://github.com/AliceO2Group/AliceO2/blob/269060e5b9a3bfdb32780e10ee6a1321edcf72c1/prodtests/full-system-test/dpl-workflow.sh#L127

https://github.com/AliceO2Group/AliceO2/blob/269060e5b9a3bfdb32780e10ee6a1321edcf72c1/prodtests/sim_challenge.sh#L161
https://github.com/AliceO2Group/AliceO2/blob/269060e5b9a3bfdb32780e10ee6a1321edcf72c1/prodtests/sim_challenge.sh#L166

https://github.com/AliceO2Group/O2DPG/blob/b7a46515d5bbe54c3d9cd90dd4b063f01cf7e775/MC/bin/o2dpg_sim_workflow.py#L483
https://github.com/AliceO2Group/O2DPG/blob/b7a46515d5bbe54c3d9cd90dd4b063f01cf7e775/MC/bin/o2dpg_sim_workflow.py#L487

We should fix them also in the same PR to avoid breaking O2 (Note that the last 2 are in the https://github.com/AliceO2Group/O2DPG package, will require a separate PR). 
Also, would be good to fix the README.

PS: it may take a few iterations until this PR is merged, would be good to use a dedicated branch for such large changes rather than `dev`.

Detectors/GlobalTracking/src/MatchTOF.cxx
noferini: I forgot to remove it. I'm going to do that

Detectors/GlobalTracking/src/MatchTOF.cxx
noferini: I forgot to remove it. I'm going to do that

noferini:
PR reopend in 6455
I'm closing this PR since started from dev

PR #6434, opened 2021-06-15
Deprecated Muons data table removed from data model. Temporarily PWGUD/upcForward.cxx task commented out from CMakeLists until the Muons table dependence will be replaced with the new tables

iarsene:


jgrosseo:
Could you rebase this?

Why do we need to keep the muoncluster table? I think the idea was to drop it as well, wasn't it?

iarsene:
failing tests show unrelated compiler warnings, merging

PR #6438, opened 2021-06-16
jobutils: Allow debugging in case of problems

sawenzel:
Do not kill processes when debug mode is set
(so that a debugger can attach)


sawenzel:
trivial and tested locally

PR #6439, opened 2021-06-16
Move o2_dpg_wokflow_runner.py from O2DPG

benedikt-voelkel:
* move since it is generic to handle the execution of workflows

* rename o2_dpg_workflow_runner.py -> o2-workflow-runner.py


sawenzel:
As discussed privately, I'd like to do this with a bit more preparation/discussion.

PR #6442, opened 2021-06-16
DPL Analysis: add static for loop

mario-krueger:


mario-krueger:
Since we decided to do the hashing of histogram names at compile time, it is currently not very convenient to call the `fill` or `get` functions of the histogram registry multiple times in a similar way according to some pattern.
For instance we cannot do things like
```cpp
static constexpr std::string_view particles[] = {"pion", "kaon", "proton"};
for (int particleID = 0; particleID < 3; ++particleID){
  //registry.fill(HIST(Form("%s_pt", particles[particleID])), pt);
  registry.fill(HIST(particles[particleID]) + HIST("_pt"), pt);
}
``` 
Instead one has to 'expand' the loop and write all the lines down separately (as done for example here https://github.com/AliceO2Group/AliceO2/blob/dev/Analysis/Tasks/PWGLF/spectraTPC.cxx#L163).
Using a 'static' for loop for these cases would improve readability of the analysis code and make it easier to maintain.
What do you think?

ktf:
Have you checked if fmt::format does not allow you to do the same thing as Form, just constexpr?

mario-krueger:
> Have you checked if fmt::format does not allow you to do the same thing as Form, just constexpr?

Good point. I will check. Thanks.

Edit:
To me it looks like it is not possible to generate a formatted `static constexpr string_view` with `fmt::format` at compile time.

ktf:
You could also look into homogeneous_apply_ref, which has the advantage of not being recursive.

mario-krueger:
> You could also look into homogeneous_apply_ref, which has the advantage of not being recursive.

Could you point me to a good description of what this is?

ktf:
```c++
#include "Framework/StructToTuple.h"
o2::framework::homogeneous_apply_refs([&ctx](auto&& x) { return ...; }, anObj);
```

mario-krueger:
The CI errors appear to be unrelated to my changes.

ktf:
@TimoWilken looks like the cs8 container is lacking OpenGL.

mario-krueger:
FYI @ekryshen this new feature could help you here: 
https://github.com/AliceO2Group/AliceO2/blob/dev/Analysis/Tasks/eventSelectionQa.cxx#L159
Cheers, Mario

PR #6431, opened 2021-06-15
[ITS-tracking] Use MCTruthContainer to manage MC labels

mconcas:
This is a frist try to use MCTruthContainer directly. Open for discussion.

TODO:
- [x] Generalize fake tagging algorithm to cope with multiple labels
- [x] Extend MClabels comparison logic wherever it is used
- [x] Test consistency and debug (high pt cutoff in efficiency?)

These are 1000 pp events with minimum track length = 7 clusters.
Current state (AliceO2/dev):
pp_eff_dev

With this PR:
pp_eff_mcTruthContainer

Difference is subtle but non zero.

PR #6443, opened 2021-06-16
ITS: set track patterns for clusters and MC

mconcas:


mconcas:
@shahor02 did you mean something like this? I used the UserField as it was completely available.

I also took the chance to set the cluster pattern in the tracker itself.
This is likely to be overwritten if we use the executable:https://github.com/AliceO2Group/AliceO2/blob/fc78990b99c6108d16e91d9fd68288507a8e6113/Detectors/ITSMFT/ITS/workflow/src/TrackerSpec.cxx#L210
but now it also works with the steering macro. 

We could consider to remove it from the trackerSpec

shahor02 (commented):
Hi @mconcas 
thanks, yes, but:
1) I thought you will add this feature on top of the https://github.com/AliceO2Group/AliceO2/pull/6431, so that the fakes pattern would profit from all labels in the MCTruthContainer
2) I would keep the user field for transient info, in the TrackITS we have currently 
https://github.com/AliceO2Group/AliceO2/blob/fc78990b99c6108d16e91d9fd68288507a8e6113/DataFormats/Detectors/ITSMFT/ITS/include/DataFormatsITS/TrackITS.h#L99
The in-memory size of the TrackITS will not increase from extending it to 32 bytes and using upper 16 for the fakes pattern (with corresponding getters). 

Also, I agree that once the patterns are set in the tracker itself, we can suppress this part in the spec (but then the same should be implemented in the @iouribelikov's tracker).

mconcas:
> 1. I thought you will add this feature on top of the #6431, so that the fakes pattern would profit from all labels in the MCTruthContainer

This is a no-issue to me, I was just being orthogonal to that just in case we decided to merge this first (as this does not affect the I/O of the tracker and results easier to integrate on the multi roframe stuff). I can close this PR and update #6431.
> 2. I would keep the user field for transient info, in the TrackITS we have currently
>    https://github.com/AliceO2Group/AliceO2/blob/fc78990b99c6108d16e91d9fd68288507a8e6113/DataFormats/Detectors/ITSMFT/ITS/include/DataFormatsITS/TrackITS.h#L99
>    The in-memory size of the TrackITS will not increase from extending it to 32 bytes and using upper 16 for the fakes pattern (with corresponding getters).

True, I'll do that.

> Also, I agree that once the patterns are set in the tracker itself, we can suppress this part in the spec (but then the same should be implemented in the @iouribelikov's tracker).

Aren't anyways the two `*TrackerSpecs` separate?

shahor02:
> Aren't anyways the two `*TrackerSpecs` separate?

Right!

mconcas:
Ok, closing here as I am porting these changes to #6431.

PR #6444, opened 2021-06-16
MCH: digits merger implementation

aferrero2707:
This PR introduces the code required for the merging of SAMPA digits in continuous data taking mode, when the detector signals can be into multiple SAMPA acquisition windows that need to be combined in order to recover the full deposited charge.

The merging is performed when the first ADC sample of a given digit follows immediately the last ADC sample of an existing digit from the same readout channel.

To speed up the computations, the decoder keeps a table of references to the last stored digits for each readout channel, as well as the time stamps of the last ADC sample of those digits.  When a new digit is decoded, its time stamp is compared with the last existing digit for the same channel. If the time gap is exactly equal to one ADC clock cycle the new digit is merged into the existing one, otherwise the new digit is stored separately.

### DIFF
--- a/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
+++ b/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
@@ -159,6 +159,21 @@ class DataDecoder
   void dumpDigits();
   bool getPadMapping(const DsElecId& dsElecId, DualSampaChannelId channel, int& deId, int& dsIddet, int& padId);
   bool addDigit(const DsElecId& dsElecId, DualSampaChannelId channel, const o2::mch::raw::SampaCluster& sc);
+  int32_t getMergerChannelId(const DsElecId& dsElecId, DualSampaChannelId channel);
+  void updateMergerRecord(uint32_t mergerChannelId, uint32_t digitId);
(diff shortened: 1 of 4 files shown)


aferrero2707:
@TimoWilken  @ktf the MacOS CI job fails due to a lot of tests that fail, none apparently related to this PR... could you have a look?

Thanks!

aphecetche:
@aferrero2707 to fix the macOS build  I think you should rebase to dev

aferrero2707:
@aphecetche I rebased and force pushed today, but the macOs builds are still failing...

aphecetche:
@aferrero2707 I have no idea what's going on here. I really thought that problem was fixed.

ktf:
Could it be some leftover from old PRs?

aphecetche:
@ktf I don't understand how that could be. On that Mac ci machine I've looked for e.g. remnants of libs with main but found none : 
```
> pwd
/build/ci_checks/alice_ci_0/o2/sw/osx_x86-64/O2
> nm -aA latest*/lib/*.dylib | grep _main
```

There _is_ one location with old libs with the wrong _main symbol in, but it's under `/Users/alibuild/alice/` (that was used to debug the issue in the first place), so I guess this cannot (or should not, at least) interfere in any way, right ?

aphecetche (approved):

PR #6446, opened 2021-06-16
DCS config data format

chiarazampolli:
This is an example of data format to be used to fill the CCDB instance to store the calibration information to be used to configure detectors. Such object will be picked up by DCS, to be stored in Oracle and/or be used by the detectors (FEE or other). 

See https://alice.its.cern.ch/jira/browse/O2-2341

E.g. for testing, run (compile first) the following:

```
#include "DataFormatsDCS/DCSConfigObject.h"
#include <iostream>
#include <TFile.h>

int test()
{
  o2::dcs::DCSconfigObject_t config;
  o2::dcs::addConfigItem(&config, "pippo", "pluto");
  o2::dcs::addConfigItem(&config, "pippo_float", 12.f);

  o2::dcs::printDCSConfig(config);

  TFile fout("myDCSConfig.root", "recreate");
  fout.WriteObjectAny(&config, "std::vector<char>", "DCSConfig");
  fout.Close();

  TFile fin("myDCSConfig.root", "read");
  o2::dcs::DCSconfigObject_t* config_r = (o2::dcs::DCSconfigObject_t*)fin.Get("DCSConfig");
  std::cout << "\nReading back" << std::endl;
  o2::dcs::printDCSConfig(*config_r);
  fin.Close();
  
  return 0;
}
```


aphecetche:
@chiarazampolli thanks for the PR. Any reason why you are passing the dcsconfig by pointer in addConfigItem ? Why not just `addConfigItem(DCSconfigObject_t& configVector, std::string key, const T value)` ? Would eliminate one possible error (calling the method with nullptr)

chiarazampolli:
Hi @aphecetche ,

No reason, and now I don't remember if it was like this since initially I did it differently. Yours is a good suggestion, I will change accordingly, thanks!

Chiara

chiarazampolli:
Hello,

macos errors does not seem related to my PR, unless i overlooked something. CI log is not present. 

@shahor02 , do you think this PR can be merged?

Chiara

shahor02 (approved):

PR #6447, opened 2021-06-16
read number of channels from Geometry

AllaMaevskaya:
Temporary solution for O2-2380 "Mem corruption in MC-AOD production"

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -36,6 +36,7 @@
 #include "SimulationDataFormat/MCTrack.h"
 #include "SimulationDataFormat/MCTruthContainer.h"
 #include "SimulationDataFormat/MCEventLabel.h"
+#include "FT0Base/Geometry.h"
 #include "TMath.h"
 #include "MathUtils/Utils.h"
 #include <map>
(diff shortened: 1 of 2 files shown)


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -606,21 +607,24 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
     }
-    float aAmplitudesA[96];
-    float aAmplitudesC[112];
-    for (int i = 0; i < 96; i++) {
-      aAmplitudesA[i] = truncateFloatFraction(vAmplitudes[i], mT0Amplitude);
-    }
-    for (int i = 0; i < 112; i++) {
-      aAmplitudesC[i] = truncateFloatFraction(vAmplitudes[i + 96], mT0Amplitude);
+    float aAmplitudesA[nFT0ChannelsAside];
+    float aAmplitudesC[133];
```
shahor02, line 621: In this way you are introducing a dummy channels. Is not it better to introduce at once a proper mapping?
I thought something like 
````
int HW2SW(int id) const { return id < 200 ? id : (id>203 ? id - 4 : -1); }
int SW2HW(int id) const { return id<200 ? id : id + 4; }
````
should be enough.
Anyway, if it is more complex, I hope this solution is indeed temporary and will be overridden by a proper mapping.

jotwinow: Alla, I would go for the general solution.

AllaMaevskaya: General solution takes time

shahor02 (approved):
OK, merging this, but please, don't postpone correct solution.

PR #5765, opened 2021-03-22
DPL: improve handling of unbalanced queues

ktf:
In case a device is receiving data from multiple sources
we make sure we wait for the slow producers, rather than
dropping messages.

PR #6451, opened 2021-06-17
DPL: introduce Backpressure policy

ktf:
For now it behaves as before and cannot be customised.


ktf:
Baby steps towards a working #5765 integration.

PR #6452, opened 2021-06-17
Use expired messages mechanism for FDD

mbroz84:


shahor02 (commented):
Hi @mbroz84 
Thanks, the changes look ok, but it seems your branch was not up-to-date. Could you do `git pull` and push again? In general, opening the PR from dev branch is not a good idea...

AllaMaevskaya (dismissed):

afurs:
Hello @mbroz84 @shahor02 , 

I think there are no need in this implementation, because this mechanism is already applied in common FITWorkflow library(here https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/FIT/workflow/include/FITWorkflow/FITDataReaderDPLSpec.h) This new lib seems to be stable(no issues from FV0), so, it could be implemented also for FDD, in the same way as for FV0. What do you think about it?

shahor02:
@afurs @mbroz84 I saw that FITDataReaderDPL implements this protection but the FDD does not use this reader, so I wrote to @mbroz84. I would only support using common code by similar detectors.

afurs:
> @afurs @mbroz84 I saw that FITDataReaderDPL implements this protection but the FDD does not use this reader, so I wrote to @mbroz84. I would only support using common code by similar detectors.

Well, anyway I'm planning to fully implement common FIT libs(FITRaw:digit2raw and FITWorkflow) for remaining FIT sub-detectors :) I was waiting some time just to be sure that this libs are working correct, without any issues.

mbroz84:
> > @afurs @mbroz84 I saw that FITDataReaderDPL implements this protection but the FDD does not use this reader, so I wrote to @mbroz84. I would only support using common code by similar detectors.
> 
> Well, anyway I'm planning to fully implement common FIT libs(FITRaw:digit2raw and FITWorkflow) for remaining FIT sub-detectors :) I was waiting some time just to be sure that this libs are working correct, without any issues.

Ok so we don't need this PR and I will close it?

shahor02:
@mbroz84 The best would be if you could adopt the common reader class (which already supports this protection).

mbroz84:
> @mbroz84 The best would be if you could adopt the common reader class (which already supports this protection).

Ok, I will do that and open new PR. Closing here.

afurs:
@shahor02 if it shouldn't be done ASAP, I can do this in the next PR(I also planned it for FT0). 
@mbroz84 but anyway, if you decide to do it, here is example of adopting common FIT raw reader class for FV0: https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/FIT/FV0/workflow/src/fv0-flp-workflow.cxx

Just change namespaces and remove parts of code which related to extended TCM mode(like RawReaderFV0ext, because I haven't defined "extended TCM mode" for FDD yet)

shahor02:
It is not terribly urgent but if we can have it in a week or two it would be fine.

PR #6453, opened 2021-06-17
[MCH] fixed memory leak in pedestals decoder

aferrero2707:
The memory leak occurred when encountering decoding errors.
In this case the vector of error messages was continuously filled and never cleared.

### DIFF
--- a/Detectors/MUON/MCH/Calibration/src/pedestal-decoding-workflow.cxx
+++ b/Detectors/MUON/MCH/Calibration/src/pedestal-decoding-workflow.cxx
@@ -87,6 +87,29 @@ static bool isValidDeID(int deId)
   return false;
 }
 
+static void patchPage(gsl::span<const std::byte> rdhBuffer, bool verbose)
+{
+  static int sNrdhs = 0;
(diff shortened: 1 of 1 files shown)


Detectors/MUON/MCH/Calibration/src/pedestal-decoding-workflow.cxx
```diff
@@ -140,6 +163,7 @@ class PedestalsTask
   void reset()
   {
     mDigits.clear();
+    mErrors.clear();
```
aphecetche, line 166: I guess this is actual fix advertised in the PR title, right ? That one and the addition of the ERRORS output make perfect sense.

But what about the patchPage change : is it needed / related to this fix at all ?

aferrero2707: The patchPage change is not related to the memory leak, but introduces the same code we have in the standard decoder. I just wanted to avoid two independent PR with small changes... but we can split the two if you prefer.

aphecetche: As long as it is *actually* needed and not some leftover, that's fine with me then.

aphecetche (approved):

PR #6454, opened 2021-06-17
DPL: improve SimpleSink

ktf:


ktf:
Getting this out of the way while trying to fix the unbalanced queue issue which affects analysis / data taking.

ktf:
@Barthelemy the multinode test still fails. The error from DPL indicates that one of the devices terminated abruptly.

PR #6455, opened 2021-06-17
TOF matcher workflow re-arranged

noferini:
This is a new PR started from a non-dev branch.
Previous one was closed (PR6423).
Missing comments by Ruben to be addressed are reported below.

- What is the problem with makeConstrainedTPCTracks refit? W/o it the match cannot be used in the downstream workflows.
-  you could actually reserve less slots if you would count the numbers of different track types fetched by the creator, e.g. you cannot have mRecoCont->getTPCTracks().size() TPC matches, since part of TPC tracks will come via ITS-TPC or TPC-TRD etc.

BTW, what prevents processing the TPC-TRD and ITS-TPC-TRD tracks at once?
- Why are you doing this check after adding the track? You could apply this trivial cut directly in the creator and simply not call the addXXXSeed at all if the track does not reach the TPC outer ref. (2 times!)
-


noferini:
All comments fully addressed but:

- What is the problem with makeConstrainedTPCTracks refit? W/o it the match cannot be used in the downstream workflows.
-The error is
[171553:tof-matcher]: [12:18:48][INFO] Shifting Z for 20 matched TPC tracks according to TOF time info
[171553:tof-matcher]: terminate called without an active exception
[ERROR] websocket_server_callback: Error while reading from websocket


To be defined how to split outputs for constrained track once TPC-TRD and ITS-TPC-TRD will be included

noferini:
makeConstrainedTPCTracks problem fixes. Now it works (filling of one vector was missing)

noferini:
build/O2/o2 check expected to be fixed after
https://github.com/AliceO2Group/QualityControl/pull/735

Detectors/GlobalTrackingWorkflow/src/TOFMatcherSpec.cxx
```diff
@@ -135,34 +150,36 @@ void TOFMatcherSpec::endOfStream(EndOfStreamContext& ec)
...
+    dataRequest->requestClusters(GTrackID::getSourceMask(GTrackID::FT0), false);
   }
 
-  outputs.emplace_back(o2::header::gDataOriginTOF, "MATCHINFOS", 0, Lifetime::Timeframe);
+  std::vector<OutputSpec> outputs;
+
+  outputs.emplace_back(o2::header::gDataOriginTOF, "MATCHINFO_0", 0, Lifetime::Timeframe);
+  outputs.emplace_back(o2::header::gDataOriginTOF, "MATCHINFO_1", 0, Lifetime::Timeframe);
+  outputs.emplace_back(o2::header::gDataOriginTOF, "MATCHINFO_2", 0, Lifetime::Timeframe);
+  outputs.emplace_back(o2::header::gDataOriginTOF, "MATCHINFO_3", 0, Lifetime::Timeframe);
```
shahor02: At the moment ok, later may change this either to more meaningful names or use subspec.

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -130,7 +130,7 @@ WORKFLOW+="o2-trd-global-tracking $ARGS_ALL --disable-root-input --disable-root-
 
 # Workflows disabled in sync mode
 if [ $SYNCMODE == 0 ]; then
-  WORKFLOW+="o2-tof-matcher-tpc $ARGS_ALL --disable-root-input --disable-root-output $DISABLE_MC | "
+  WORKFLOW+="o2-tof-matcher-workflow $ARGS_ALL --disable-root-input --disable-root-output $DISABLE_MC | "
```
shahor02, line 133: Now the `o2-tof-reco-workflow` will produce only digits and clusters, while this one will  run with current defaults the  TPC and ITS-TPC mathing, right?

shahor02 (commented):
Thanks @noferini 
Since the FullCI fails due to the QC, did you run the FST locally to make sure it is not broken.
BTW, to not forget: once this is merged, the TOF reco should be updated also in the https://github.com/AliceO2Group/O2DPG/blob/b7a46515d5bbe54c3d9cd90dd4b063f01cf7e775/MC/bin/o2dpg_sim_workflow.py#L482-L487

prodtests/full-system-test/dpl-workflow.sh
noferini: right

noferini:
> Thanks @noferini
> Since the FullCI fails due to the QC, did you run the FST locally to make sure it is not broken.
> BTW, to not forget: once this is merged, the TOF reco should be updated also in the https://github.com/AliceO2Group/O2DPG/blob/b7a46515d5bbe54c3d9cd90dd4b063f01cf7e775/MC/bin/o2dpg_sim_workflow.py#L482-L487

Sorry, what is "FST"?
I compiled locally QC after removing the dependency  and then open PR
https://github.com/AliceO2Group/QualityControl/pull/735
I will fix O2DPG after the merging of this one.

shahor02:
Hi, FST is the prodtests/full_system_test.sh script

noferini:
I had problem in my laptop with FST with SHM which I didn't understand but I was able to run from scratch these commands (after a minor fix in tof-reco-workflow already ported)

o2-sim-serial -n 10 -g pythia8pp -e TGeant3 -m ITS TPC TRD TOF
o2-sim-digitizer-workflow -b --onlyDet ITS,TPC,TRD,TOF --shm-segment-size 4000000000
o2-tof-reco-workflow -b --output-type clusters --shm-segment-size 4000000000
o2-its-reco-workflow -b --shm-segment-size 4000000000
o2-tpc-reco-workflow -b --shm-segment-size 4000000000 --output-type clusters,tracks
o2-tpcits-match-workflow -b --shm-segment-size 4000000000
o2-tof-matcher-workflow -b --shm-segment-size 4000000000
o2-tof-matcher-workflow -b --high-purity --shm-segment-size 4000000000
[O2/latest AEGIS/latest] ~/prodtest $> 

I got proper outputs with tof-matcher-workflow (also with high-purity selected)

noferini:
vector management (resize, reserve, ...) optimized
QC is expected to compile now
minor fix for TGeo INFO (instead of warning) ported [@davidrohr ]

noferini:
I will re-open the PR since it seems to me that the last changes ported to QC are not included here for the tests.
Do you agree?

davidrohr:
I don't understand what you mean by reopen?

noferini:
> I don't understand what you mean by reopen?

I see that checks are running with the old QC even if my fix is merged.
I was wondering if the new QC master will be used in the check if I close and then open a new PR

davidrohr:
Hi Francesco, closing and reopening will not change what is used in the test. I think merging to QC master is not enough, we need a new QC tag in bump it in alidist, to test with your fix. @Barthelemy : could you take care?

Barthelemy:
@davidrohr QC v1.20 is out, the alidist bump is pending here: https://github.com/alisw/alidist/pull/3122

davidrohr:
thx
@noferini : Could you rebase this PR to dev?

noferini:
What is the plan for this PR?
I suspect that the conflict will continue to appear if we don't merge these changes

shahor02:
Hi @noferini 
I'll merge it as soon as it passes the FST (will test locally once the conflicts are resolved).

davidrohr:
Qc has been bumped, so if you rebase this it should hooefully pass the CI.

shahor02:
Hi @noferini 

Both the full_system_test.sh and sim_challenge are broken by this PR... I've opened a https://github.com/noferini/AliceO2/pull/28 in your repo which fixes them: apart from workflows options/reshuffling, I've also reverted the `MATCHINFO_0` and `MATCHINFO_1` to `MATCHINFOS_TPC` and `MATCHINFOS` resp.. The alternative would be to modify other workflows, but this should be anyway changed on TOF side:  (1) the TOF matcher should not always return all 4 `MATCHINFO_X` types and (2) there should be difference between the same (detector-wise) matchinfo in strict and relaxed modes.
I propose that you merge my fix, I merge this PR after the CI, then I will gradually fix the DataSpecs issue (most probably, using different subspecs).

noferini:
> Hi @noferini
> 
> Both the full_system_test.sh and sim_challenge are broken by this PR... I've opened a [noferini#28](https://github.com/noferini/AliceO2/pull/28) in your repo which fixes them: apart from workflows options/reshuffling, I've also reverted the `MATCHINFO_0` and `MATCHINFO_1` to `MATCHINFOS_TPC` and `MATCHINFOS` resp.. The alternative would be to modify other workflows, but this should be anyway changed on TOF side: (1) the TOF matcher should not always return all 4 `MATCHINFO_X` types and (2) there should be difference between the same (detector-wise) matchinfo in strict and relaxed modes.
> I propose that you merge my fix, I merge this PR after the CI, then I will gradually fix the DataSpecs issue (most probably, using different subspecs).

Merged in my own branch

shahor02:
Hi @noferini 
Could you please apply from O2 dir.:
````
git diff --diff-filter d --name-only $(git merge-base HEAD origin/dev) | xargs sed -i 's/\t/ /g'
git diff --diff-filter d --name-only $(git merge-base HEAD origin/dev) | xargs sed -i 's/[[:space:]]*$//'
````
The QC v1.20 was just merged, hopefully o2 CI will be ok at next iteration.

shahor02:
local FST passed, merging.

PR #6457, opened 2021-06-17
Add protection for dropped TFs

dstocco:


Detectors/MUON/MID/Workflow/src/RawDecoderSpec.cxx
```diff
@@ -53,7 +54,14 @@ class RawDecoderDeviceDPL
     auto tStart = std::chrono::high_resolution_clock::now();
 
     auto tAlgoStart = std::chrono::high_resolution_clock::now();
-    of::DPLRawParser parser(pc.inputs());
+
+    if (isDroppedTF(pc, header::gDataOriginMID)) {
+      return;
```
shahor02, line 63: @dstocco the whole point of this protection is that the device should send valid output in case the input allowing to start processing is not received. By doing nothing in case of lost TF you will block the downstream devices, which expect your input. Instead, your device should emit an "empty" output, with definition of "empty" suitable for your data types (I guess empty ROFRecord and channels vectors are OK).

shahor02 (changes requested):

Detectors/MUON/MID/Workflow/src/RawDecoderSpec.cxx
dstocco: Hi @shahor02 , ok, I see. I send valid (empty) messages it in the new commit. It can be squashed before merging.

shahor02 (approved):
@dstocco looks fine, merging

PR #6458, opened 2021-06-18
Raw: make the output directory creation more robust.

aphecetche:
Turns out that `std::filesystem::create_directories` on some platforms/compilers does not return true even when actually
creating the directory. So this PR changes a bit the logic used to force creation of output directory : call `std::filesystem::create_directories` but do not rely on the boolean return value. Make an explicit `std::filesystem::exists` call instead.

Discovered that while trying to get the FST working on macOS. The symptom was that the rawXXX.log showed a fatal due to the impossibility to create the raw/XXX directory. But in fact the directory was created just fine.

### DIFF
--- a/Detectors/Raw/src/RawFileWriter.cxx
+++ b/Detectors/Raw/src/RawFileWriter.cxx
@@ -740,10 +740,10 @@ void o2::raw::assertOutputDirectory(std::string_view outDirName)
     if (!std::filesystem::exists(outDirName)) {
       LOG(FATAL) << "could not create output directory " << outDirName;
     }
-  }
 #else
     if (!std::filesystem::create_directories(outDirName)) {
       LOG(FATAL) << "could not create output directory " << outDirName;
     }
 #endif
+  }
 }
--- a/Detectors/Raw/src/RawFileWriter.cxx
+++ b/Detectors/Raw/src/RawFileWriter.cxx
@@ -730,8 +730,20 @@ void RawFileWriter::DetLazinessCheck::completeLinks(RawFileWriter* wr, const IR&
 
 void o2::raw::assertOutputDirectory(std::string_view outDirName)
 {
-  std::filesystem::create_directories(outDirName);
   if (!std::filesystem::exists(outDirName)) {
-    LOG(FATAL) << "could not create output directory " << outDirName;
+#if defined(__clang__)
+    // clang `create_directories` implementation is misbehaving and can
+    // return false even if the directory is actually successfully created
+    // so we work around that "feature" by not checking the
+    // return value at all but using a second call to `exists`
+    std::filesystem::create_directories(outDirName);
+    if (!std::filesystem::exists(outDirName)) {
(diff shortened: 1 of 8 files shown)


Detectors/Raw/src/RawFileWriter.cxx
```diff
@@ -726,3 +727,11 @@ void RawFileWriter::DetLazinessCheck::completeLinks(RawFileWriter* wr, const IR&
   }
   clear();
 }
+
+void o2::raw::assertOutputDirectory(std::string_view outDirName)
+{
+  std::filesystem::create_directories(outDirName);
```
davidrohr: I think the logic here should be:
```
if (!exists) {
#ifdef MACOS
  create();
  if (!exists) error;
#else
  if (!create()) error;
#endif
}
```
with some comment that this is broken on mac OS currently and that the workaround should be reverted once it works. I like merging all the checks into this one function.

aphecetche: Well, yes, that's another option. But I believe the check should be done on the compiler, not the platform, as I think the issue is more a Clang vs GCC one (I understand we're only using Clang on Mac, but the intent of the check would be clearer that way)

davidrohr: Yes, sure, my point was just we should document why it is done like this, together with a comment that it should be cleaned up. Is there actually any reference that this is broken on Mac / clang? Clang is usually well-tested, so I would be surprised if `create_directories` would be misbehaving.

aphecetche: Would have to dig a reference if there is one, but can be seen in e.g. https://godbolt.org/z/qcE5fMq36

davidrohr: OK, then it is most likely a bug in clang?
Can you open a bug report? The clang developers are usually quite responsive.

davidrohr: And still, could we change it to have the exists check first, and then add a comment with a todo that this is a workaround due to a clang bug?

aphecetche: I've just added a commit (to be squashed on merge) in that respect : would that be ok for you ?

davidrohr (dismissed):
perfect, thx, will squash-merge once CI is through

davidrohr:
@aphecetche : This seems to be a genuine error:
```
/mnt/mesos/sandbox/sandbox/o2suite-dataflow/sw/SOURCES/O2/6458/0/Detectors/Raw/src/RawFileWriter.cxx: In function 'void o2::raw::assertOutputDirectory(std::string_view)':
/mnt/mesos/sandbox/sandbox/o2suite-dataflow/sw/SOURCES/O2/6458/0/Detectors/Raw/src/RawFileWriter.cxx:749:1: error: expected '}' at end of input
```

davidrohr (approved):

PR #6461, opened 2021-06-18
CcdbApi: Option to build/use local cache on the fly

sawenzel:
Offering a mode - useful for GRID - in which
we save downloaded objects into a given CCDB disc
store, thereby creating a local cache as we go.

This local store can be used to reduce the number
of network queries, as well as to know exactly which
objects have been requested by whom.

The cache can also be distributed with the software stack
in case someone doesn't have network access or credentials.

The usage is:

```
export ALICEO2_CCDB_LOCALCACHE=$pwd/.ccdb
process1
process2
```

where `process1` and `process2` would both build and use
the local cache located in the .ccdb folder (depending on the order).

Limitations:
- only to be used when all processes query for the same
  timestamp (typically the case for a single GRID job)

### DIFF
--- a/CCDB/src/CcdbApi.cxx
+++ b/CCDB/src/CcdbApi.cxx
@@ -35,6 +35,7 @@
 #include <boost/algorithm/string.hpp>
 #include <iostream>
 #include <mutex>
+#include <boost/interprocess/sync/named_semaphore.hpp>
 
 namespace o2
 {
@@ -228,12 +229,16 @@ string CcdbApi::getFullUrlForStorage(CURL* curl, const string& path, const strin
   return fullUrl;
 }
 
+std::string getSnapshotPath(std::string const& topdir, const string& path)
+{
+  return topdir + "/" + path + "/snapshot.root";
+}
+
 // todo make a single method of the one above and below
 string CcdbApi::getFullUrlForRetrieval(CURL* curl, const string& path, const map<string, string>& metadata, long timestamp) const
 {
   if (mInSnapshotMode) {
-    string snapshotPath = mSnapshotTopPath + "/" + path + "/snapshot.root";
-    return snapshotPath;
+    return getSnapshotPath(mSnapshotTopPath, path);
   }
 
   // Prepare timestamps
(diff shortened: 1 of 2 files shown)


costing:
That's a great development @sawenzel ! This caching strategy would indeed be efficient for a single MC job. 

As an extension to this functionality and to implement the full snapshot-in-a-file functionality I suggest to:
- store the objects in a `.ccdb/<path>/<unique identifier>.root` file name
- validate the structure at the beginning of the job, asking CCDB with the existing unique identifier. If the answer is not a `HTTP 304 (Not Modified)` then save instead the content in a `.ccdb/<path>/<new id>.root` and wipe the old file
- if CCDB is not available, simply use what you have locally
- have the code pick any (and there should only be one according to the above) file from `.ccdb/<path>/` to use
- log the file name in each job so that full traceability is insured 

With this a production precursor would download most if not all of the content once, we can package the directory and give it as input to all the other jobs. Any difference would be spotted at the beginning of the job and corrected for. With a minimal amount of network traffic overall.

costing (dismissed):

CCDB/src/CcdbApi.cxx
```diff
@@ -654,13 +666,65 @@ void* CcdbApi::retrieveFromTFile(std::type_info const& tinfo, std::string const&
...
+  if (cachedir) {
+    // protect this sensitive section by a multi-process named semaphore
+    bool use_sema = false;
+    boost::interprocess::named_semaphore* sem = nullptr;
+    std::hash<std::string> hasher;
+    const auto semhashedstring = "alice_ccdb" + std::to_string(hasher(std::string(cachedir) + path));
+    try {
+      sem = new boost::interprocess::named_semaphore(boost::interprocess::open_or_create_t{}, semhashedstring.c_str(), 1);
+    } catch (std::exception e) {
+      LOG(WARN) << "Exception occurred during GRP semaphore setup; Continuing without";
```
shahor02: is GRP a copy/paste typo?

shahor02 (dismissed):
Looks nice!

CCDB/src/CcdbApi.cxx
sawenzel: Yes indeed

PR #6462, opened 2021-06-18
[EMCAL-565]: Modified BoostHistogramUtils after testing.

hjbossi:


mfasDa (dismissed):

Common/Utils/include/CommonUtils/BoostHistogramUtils.h
```diff
@@ -270,41 +322,55 @@ fitResult<T, 4> fitGaus(Iterator first, Iterator last, AxisIterator axisfirst)
...
     rms2COG /= sumCOG;
+    /*
     result.setParameter<0>(xMax);
     result.setParameter<1>(meanCOG);
     result.setParameter<2>(TMath::Sqrt(TMath::Abs(meanCOG * meanCOG - rms2COG)));
     result.setChi2(-2);
+    */
+    result.at(0) = *xMax;
+    result.at(1) = meanCOG;
+    result.at(2) = TMath::Sqrt(TMath::Abs(meanCOG * meanCOG - rms2COG));
```
sawenzel, line 351: This looks like a template function. In this case it would really be better to use `std::` math functions since they are also templates. The ROOT ones are just `double` I think and there is no reason preferring ROOT math functions over the standard (for things such as exp, sin, cos,...).

sawenzel: Maybe this could be cleaned up in a separate PR

PR #6465, opened 2021-06-19
[O2-2132] CMake: get rid of custom test wrapper script

aphecetche:
Resuming work on this issue after a long hiatus...

=====

Get rid of test wrapper in favor of using native cmake/ctest facilities as much as possible.

DO NOT MERGE YET.

As agreed in the WP3 list thread, and quoting @davidrohr the idea would be to :

> make sure there are no random glitches afterwards.
> So I'd suggest we open a PR, and wait for the CI, which will check all tests.
> But then we don't merge it yet, but we force push an amended commit multiple times, and verify that the CI always succeeds.
> We merge when we have made sure that the CI success ratio is identical to before?
> In addition, if we want to keep the same feature as of now regarding the number of times a failed test is retried, then we should amend the o2.sh recipe to use ctest --repeat until-pass:3 (as just with this PR there is no retry at all).


aphecetche:
@davidrohr for the record, note that I will clean up this PR (to remove some changes unrelated to the ctest business), but other than that I guess the test wrapper is about to be gone ;-)

davidrohr:
@aphecetche : Could you rebase this to dev?

aphecetche:
@davidrohr done

aphecetche:
@davidrohr any idea what's the "aliDoctor error" for the fullCI ?

davidrohr:
@TimoWilken : I assume this is the error you fixed with the update to the container?

TimoWilken:
Not quite, this is separate issue I just fixed. I'll restart the check.

davidrohr:
@TimoWilken : Now the FullCI fails with `Error getting OpenCL Platform Count`.
I assume that is related to your changes to the container from yesterday?
Could be that the env variables are not set correctly.

davidrohr:
I can reproduce the fullCI issue locally. The new docker container that was created yesterday evening is broken.
@TimoWilken : Could you update the container again as I have written on Mattermost?

PR #6466, opened 2021-06-20
Move MFT track labels to a separate table + some cleanups

nburmaso:
* A separate table for MFT track labels was added
* Using local helper structure for track extra info instead of global variables
* Some cleanups and fixes for typos

### DIFF
--- a/Framework/Core/include/Framework/AnalysisDataModel.h
+++ b/Framework/Core/include/Framework/AnalysisDataModel.h
@@ -808,12 +808,12 @@ using McTrackLabel = McTrackLabels::iterator;
 namespace mcmfttracklabel
 {
 DECLARE_SOA_INDEX_COLUMN(McParticle, mcParticle); //! MC particle
-DECLARE_SOA_COLUMN(McMask, mcMask, uint16_t);     //! Bit mask to indicate detector mismatches (bit ON means mismatch). Bit 0-6: mismatch at ITS layer. Bit 7-9: # of TPC mismatches in the ranges 0, 1, 2-3, 4-7, 8-15, 16-31, 32-63, >64. Bit 10: TRD, bit 11: TOF, bit 15: indicates negative label
+DECLARE_SOA_COLUMN(McMask, mcMask, uint8_t);
 } // namespace mcmfttracklabel
 
-DECLARE_SOA_TABLE(McMftTrackLabels, "AOD", "MCMFTTRACKLABEL", //! Table joined to the mft track table containing the MC index
+DECLARE_SOA_TABLE(McMFTTrackLabels, "AOD", "MCMFTTRACKLABEL", //! Table joined to the mft track table containing the MC index
                   mcmfttracklabel::McParticleId, mcmfttracklabel::McMask);
-using McMftTrackLabel = McMftTrackLabels::iterator;
+using McMFTTrackLabel = McMFTTrackLabels::iterator;
 
 namespace mccalolabel
 {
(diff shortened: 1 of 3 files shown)


Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -805,6 +805,16 @@ DECLARE_SOA_TABLE(McTrackLabels, "AOD", "MCTRACKLABEL", //! Table joined to the
                   mctracklabel::McParticleId, mctracklabel::McMask);
 using McTrackLabel = McTrackLabels::iterator;
 
+namespace mcmfttracklabel
+{
+DECLARE_SOA_INDEX_COLUMN(McParticle, mcParticle); //! MC particle
+DECLARE_SOA_COLUMN(McMask, mcMask, uint16_t);     //! Bit mask to indicate detector mismatches (bit ON means mismatch). Bit 0-6: mismatch at ITS layer. Bit 7-9: # of TPC mismatches in the ranges 0, 1, 2-3, 4-7, 8-15, 16-31, 32-63, >64. Bit 10: TRD, bit 11: TOF, bit 15: indicates negative label
```
jgrosseo: The comment needs to be adjusted to the possible mismatches. I guess we can only have one per MFT layer? So probably uint8_t is enough? @rpezzi to confirm.

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -805,6 +805,16 @@ DECLARE_SOA_TABLE(McTrackLabels, "AOD", "MCTRACKLABEL", //! Table joined to the
...
+} // namespace mcmfttracklabel
+
+DECLARE_SOA_TABLE(McMftTrackLabels, "AOD", "MCMFTTRACKLABEL", //! Table joined to the mft track table containing the MC index
```
jgrosseo: Please use capital MFT here.

jgrosseo (changes requested):
This is fine with me apart from the two small suggestions.

Framework/Core/include/Framework/AnalysisDataModel.h
nburmaso: Hi @jgrosseo. Sure, I will make a 8-bit mask then. Although, I'm not sure what to put in the mask. Currently, I use label masks to indicate `isFake` and `isNoise` flags and also to show mismatches between ITS and TPC labels for ITS-TPC tracks.

jgrosseo: Ask the MFT experts what mismatches are possible. I'd assume one per layer.
I guess for ITS-TPC tracks it goes to the McMask of the MCTrackLabel, right? There the comment is correct, you can follow it.

rpezzi: Hi @nburmaso and @jgrosseo , at the moment we do not compute where the mismatch occurred. The MFT sets a simple true or fake flag according to defined criteria of 80% clusters from the same MC particle. At the moment we do not use such a mask. 
By the way, thanks for the PR.

nburmaso:
I have fixed the table naming and the data type of the mask. For the moment I set 7th bit to indicate `isFake` and 6th bit for `isNoise`. @rpezzi, is it ok to do this for a placeholder?

rpezzi:
> I have fixed the table naming and the data type of the mask. For the moment I set 7th bit to indicate `isFake` and 6th bit for `isNoise`. @rpezzi, is it ok to do this for a placeholder?

Thanks. Looks good.

jgrosseo:
You can add an enum for this to DataTypes.h

jgrosseo (approved):

PR #6472, opened 2021-06-20
CMake: appending to, instead of setting, {C,Fortran,CXX}_FLAGS

aphecetche:
This allows the developper to customize (e.g. to compile with sanitizers) the build w/o modifying this cmake file.

### DIFF
--- a/dependencies/O2CompileFlags.cmake
+++ b/dependencies/O2CompileFlags.cmake
@@ -10,6 +10,12 @@
 
 include_guard()
 
+# Define the C, CXX, Fortran (and possibly linker) flags
+# for the different build types we support :
+# Debug, Release, RelWithDebInfo
+# (FIXME: Coverage is left here but status unclear, to be reviewed ?)
+#
+
 set(CMAKE_CXX_FLAGS_COVERAGE "-g -O2 -fprofile-arcs -ftest-coverage")
 set(CMAKE_C_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_COVERAGE}")
 set(CMAKE_Fortran_FLAGS_COVERAGE "-g -O2 -fprofile-arcs -ftest-coverage")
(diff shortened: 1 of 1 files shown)


davidrohr:
I am not sure if this is the best thing to do.
If we append, other packages can put something in there globally which we might not want.
If you want to compile with custom flags, you can set the `CMAKE_BUILD_TYPE` to `CUSTOM_EXTERNAL`, in which case no flags should be set, and you can provide everything externally via the `CFLAGS` / `CXXFLAGS` env variable.

aphecetche:
@davidrohr I'm not so sure I understand what you mean by other packages ? With the aliBuild recipe we actually control what *_FLAGS  are used in the cmake invocation, so no harm's possible there I would say ?

davidrohr:
Well, everything that we have as a dependency has access to these flags and can change them in their CMake part, right?

That was the reason why originally in `dependenceis/CMakeLists.txt` we were loading first the dependencies and then setting the flags, to ensure our flags are used.
This order was reverted since it created some problems, when flags were not set for the dependencies.
Actually, perhaps we should change `dependenceis/CMakeLists.txt` to
```
include("${CMAKE_CURRENT_LIST_DIR}/O2CompileFlags.cmake")
include("${CMAKE_CURRENT_LIST_DIR}/O2Dependencies.cmake")
include("${CMAKE_CURRENT_LIST_DIR}/O2CompileFlags.cmake")
```
That would ensure the right flags for all dependencies, but also prevent that a dependency can change it.

aphecetche:
Well, the dependencies should be just that, dependencies, so already compiled and installed, and thus have no business with `CMAKE_<LANG>_FLAGS` in principle, or am I missing something trivial here ?

In addition, if I understand correctly the (somewhat counterintuitive) way CMAKE_<LANG>_FLAGS variables are used in CMake what really matters is their value _at the end_ of the directory scope (see Professional CMake : A Practical Guide §15.4.4). Thus : 

```
include("${CMAKE_CURRENT_LIST_DIR}/O2CompileFlags.cmake")
include("${CMAKE_CURRENT_LIST_DIR}/O2Dependencies.cmake")
include("${CMAKE_CURRENT_LIST_DIR}/O2CompileFlags.cmake")
```

should be the same as : 

```
include("${CMAKE_CURRENT_LIST_DIR}/O2Dependencies.cmake")
include("${CMAKE_CURRENT_LIST_DIR}/O2CompileFlags.cmake")
```

davidrohr:
@aphecetche : But that doesn't make sense to me. Since then actually the include of `O2CompileFlags.cmake` wouldn't do anything at all.

I think what had happened is that find_package(something) can run some CMake code, and that altered the flags at some point.

davidrohr:
What I also don't like: If you append, you cannot overwrite settings from the outside, e.g. if you wanted to compile with `-O1`, how would you do that?

aphecetche:
> @aphecetche : But that doesn't make sense to me. 

As I said it's quite counterintuitive. Had to try to it to convince myself... Basically the following `CMakeLists.txt` : 

```
cmake_minimum_required(VERSION 3.15)

project(play VERSION 1.2.3 LANGUAGES CXX)

set(CMAKE_CXX_FLAGS "-O3")

add_executable(insane)
target_sources(insane PRIVATE insane.cxx)

set(CMAKE_CXX_FLAGS "-g -O1 -fsanitize=address")
```

leads to the last `CMAKE_CXX_FLAGS` to be used : 

```
make VERBOSE=1
...
[ 50%] Building CXX object CMakeFiles/insane.dir/insane.cxx.o
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++   -g -O1 -fsanitize=address -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk -MD -MT CM
akeFiles/insane.dir/insane.cxx.o -MF CMakeFiles/insane.dir/insane.cxx.o.d -o CMakeFiles/insane.dir/insane.cxx.o -c /Users/laurent/github.com/aphecetche/cmake-play/insane.cxx
```

(see https://github.com/aphecetche/cmake-play/tree/cmake_lang_flags)

davidrohr:
ok, but isn't that what I said, that is why I want to have the flags after the dependencies.
The reason to have them also before the dependencies was that some dependencies might copy the flags for add_custom_commands etc... (This happens for the CUDA RTC compilation I am doing, and also for the AMD HIP compilation, since HIP is not a CMake language...)
Perhaps we should discuss this in the next CWG3 meeting.

aphecetche:
> What I also don't like: If you append, you cannot overwrite settings from the outside, e.g. if you wanted to compile with `-O1`, how would you do that?

Indeed, you are right, but with the current `set(CMAKE_CXX_FLAGS ...)` you can't overwrite them either, right ?

I guess the only way to allow full flexibility would be to simply let those variables alone (which is btw one of the advice in the Professional CMake book) and always set them from the outside : from the `o2.sh` recipe (default usage) or from command line `-DCMAKE_CXX_FLAGS:STRING="-O1..."` (expert usage for devs).

davidrohr:
Well again, if you want to customize the flags, the procedure we agreed on some years ago due to these problems is:

> If you want to compile with custom flags, you can set the `CMAKE_BUILD_TYPE` to `CUSTOM_EXTERNAL`, in which case no flags should be set, and you can provide everything externally via the `CFLAGS` / `CXXFLAGS` env variable.

You will have to set some things manually of course. I am using for instance:
```
  CFLAGS: "-fPIC -pipe -O3 -march=native -g -DNDEBUG -DFAIR_MIN_SEVERITY=info"
  CMAKE_BUILD_TYPE: "CUSTOM_EXTERNAL"
  CXXFLAGS: "-fPIC -pipe -O3 -march=native -g -DNDEBUG -DFAIR_MIN_SEVERITY=info -std=c++17"
```

aphecetche:
> Well again, if you want to customize the flags, the procedure we agreed on some years ago due to these problems is:
> 
> > If you want to compile with custom flags, you can set the `CMAKE_BUILD_TYPE` to `CUSTOM_EXTERNAL`, in which case no flags should be set, and you can provide everything externally via the `CFLAGS` / `CXXFLAGS` env variable.

ok, maybe that's simply my bad (I'm known to have a bad memory ;-) ), but was it agreed in the WP3 ?  is that actually documented somewhere ? (as far as I can tell the CUSTOM_EXTERNAL thing is not a CMake thing but a custom -pardon the pun- convention, right ?)

davidrohr:
Not sure it is documented at all. Actually I think that was done even before WP3 existed, we have the same in AliRoot / AliPhysics (since I was having exactly the problem you are describing some years ago). It was decided in the Computing Board back then.

davidrohr:
Note that you should use as build type `CUSTOM_EXTERNAL` and not something arbitrary, since there are build type checks.

aphecetche:
ok, so non-standard, non-documented and a bit confidential then :smile:
```
~/alice/brew$ ls
AliPhysics      AliRoot         O2              QualityControl  alidist         alo-o2          sw

~/alice/brew$ rg CUSTOM_EXTERNAL -t cmake
AliPhysics/CMakeLists.txt
47:set(CMAKE_ALLOWED_BUILD_TYPES DEBUG RELEASE RELWITHDEBINFO COVERAGE CUSTOM_EXTERNAL)

AliRoot/CMakeLists.txt
74:set(CMAKE_ALLOWED_BUILD_TYPES DEBUG RELEASE RELWITHDEBINFO COVERAGE CUSTOM_EXTERNAL)
~/alice/brew$
```

Not setting those CMAKE_<LANG>_FLAGS at all in our CMakeLists.txt would at least be "more standard" probably. And that would not even change the most common use case (RelWithDebInfo build = -g -O2 -NDEBUG), but would only affect Debug, right ? 
Anyway, can be discussed in WP3 indeed.

davidrohr:
was this PR discussed? I didn't have time to join WP3. How do we move ahead?

aphecetche:
@davidrohr no it was not discussed as far as I know (but did miss some meetings during summer). @ihrivnac  : maybe we can fit this discussion on the next meeting if time permit ?

PR #6473, opened 2021-06-20
MFT: Fix array size (spotted with address sanitizer)

aphecetche:
@frmanso , @rpezzi : as far as I can tell, and according to Address Sanitizer, the `ts` array should be of size 25, not 15. 
(see e.g. line 353)
To be cross-checked ?


frmanso:
Hello, I would propose to keep the size ts[15] but change ts[iline] by ts [iline - 6] in lines 353 and 357.

aphecetche:
@frmanso sure, fine with me, your call. Will you open a PR and I close this one then ?

frmanso:
As you want, I can manage it if you prefer.

aphecetche:
@frmanso yes, I'd prefer if you follow this up. I'm closing this one. Thanks !

PR #6476, opened 2021-06-20
Fix in stored vectors size, resize them and pad with 0s

shahor02:


MichaelLettrich:
@shahor02 Can you have a look at #6509 ? I think this should fix the memory sanitizer issues and improve the readability of the encoder code a bit.

DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
```diff
@@ -843,23 +843,34 @@ void EncodedBlocks<H, N, W>::encode(const S_IT srcBegin,     // iterator begin o
...
+      dataSize++;
+      vtmp.resize(dataSize * sizeof(stream_t), 0);
+    }
     // no dictionary needed
     expandStorage(dataSize);
     *meta = Metadata{messageLength, 0, sizeof(uint64_t), sizeof(stream_t), probabilityBits, opt, 0, 0, 0, dataSize, 0};
     //FIXME: no we don't need an intermediate vector.
     // provided iterator is not necessarily pointer, need to use intermediate vector!!!
-    std::vector<STYP> vtmp(srcBegin, srcEnd);
+    std::copy(srcBegin, srcEnd, std::back_inserter(vtmp));
```
MichaelLettrich, line 873: `std::back_inserter()` creates a `std::back_insertion_iterator` which will append at the end of the vector. If I read this correctly, you resize to size + padding in  line 866, which means you're ending up with a vector whith size + padding zeroes and the original payload n `vtmp`, so `storeData` will always store zeros.

MichaelLettrich (changes requested):
See comment.

shahor02:
superseded by https://github.com/AliceO2Group/AliceO2/pull/6509, closing

PR #6479, opened 2021-06-21
DPL: apparently memcpy(something, 0, 0) is UB

ktf:
Keeps address sanitizer happy.

### DIFF
--- a/Framework/Core/include/Framework/DataAllocator.h
+++ b/Framework/Core/include/Framework/DataAllocator.h
@@ -298,7 +298,9 @@ class DataAllocator
 
         if constexpr (std::is_pointer<typename T::value_type>::value == false) {
           // vector of elements
-          memcpy(payloadMessage->GetData(), object.data(), sizeInBytes);
+          if (object.data()) {
+            memcpy(payloadMessage->GetData(), object.data(), sizeInBytes);
+          }
         } else {
           // serialize vector of pointers to elements
           auto target = reinterpret_cast<unsigned char*>(payloadMessage->GetData());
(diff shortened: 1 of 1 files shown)


ktf:
@davidrohr

Framework/Core/include/Framework/DataAllocator.h
```diff
@@ -298,7 +298,9 @@ class DataAllocator
 
         if constexpr (std::is_pointer<typename T::value_type>::value == false) {
           // vector of elements
-          memcpy(payloadMessage->GetData(), object.data(), sizeInBytes);
+          if (object.data()) {
```
davidrohr: i would rather check sizeInBytes.
But actually this looks ridiculous, and I cannot really follow the discussions that claim it is UB.

ktf:
```suggestion
          if (object.data() && sizeInBytes) {
```

ktf: Checking sizeInBytes only might still trigger some static checker since what is UB is really having a nullptr...

davidrohr: I don't think so, sizeInBytes should be sufficient in principle. Since if that is non-zero, also the pointers passed in must be non-zero.

ktf: I think a static checker might not know that `sizeInBytes != 0 -> object.data() != nullptr`.

davidrohr: But then why should a static checker complain? I mean he would need to know that object.data() = nullptr without sizeInBytes == nullptr? And then, do we actually check our code with static checkers?

PR #6482, opened 2021-06-21
snprintf -> fmt::format to suppress warnings

shahor02:


shahor02:
FullCI failure is unrelated.

PR #6489, opened 2021-06-22
TPC: Add possibility to read StfBuilder files directly in raw reader

wiechula:


davidrohr (approved):
haven't tried it, but this is certainly a good idea

wiechula:
@davidrohr , seems that a lot of tests go to timeout in `macOS`. For `fullCI` there is no log. Any idea, or can it be merged?

davidrohr:
@shahor02 : Is this ok with you?

wiechula:
@shahor02 , are the changes fine for you? I think the failing tests are unrelated.

shahor02 (approved):
sorry, for delay, looks ok (after very superficially look).

PR #6493, opened 2021-06-22
Alice3: Set TRK thicknesses to matbud requirements

mconcas:
@preghenella @qgp FYI.


ddobrigk:
@mconcas, thanks a lot for this! We will need to rerun all simulations for strangeness tracking with these new thicknesses, too, so it would be good if we could merge this in the next few days (the simulations take many days and we still need to move to krypton-krypton)...

mconcas:
Will open a broader set of fixes to the geometry

PR #6495, opened 2021-06-22
Bug fix: use correct index over neighbour boards

dstocco:
A vector index in a loop was wrong. This caused an out-of-bound issue.


davidrohr (approved):
thx,
fix is trivial and currently broken, merging

PR #6496, opened 2021-06-22
[EMCAL-630, EMCAL-700] Fix handling of cell indices from module indices

mfasDa:
The function GetModuleIndexesFromCellIndexesInSModule
is returning the position of the module instead of
the position of the cell inside the module. Together
with the module ID the information is redundant and
useless, while what is actually needed is the position
of the cell within the module.

### DIFF
--- a/Detectors/EMCAL/base/include/EMCALBase/Geometry.h
+++ b/Detectors/EMCAL/base/include/EMCALBase/Geometry.h
@@ -392,12 +392,14 @@ class Geometry
   int SuperModuleNumberFromEtaPhi(Double_t eta, Double_t phi) const;
 
   /// \brief Get cell absolute ID number from location module (2 times 2 cells) of a super module
-  /// \param nSupMod super module number
-  /// \param nModule module number
-  /// \param nIphi index of cell in module in phi direction 0 or 1
-  /// \param nIeta index of cell in module in eta direction 0 or 1
+  /// \param supermoduleID super module number
+  /// \param moduleID module number
+  /// \param phiInModule index of cell in module in phi direction 0 or 1
+  /// \param etaInModule index of cell in module in eta direction 0 or 1
   /// \return cell absolute ID number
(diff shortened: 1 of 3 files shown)


mfasDa:
@sawenzel @davidrohr The PR errors seem to be unrelated. Would it be possible to merge the PR once the fullCI finished?

Thanks a lot!

shahor02:
@mfasDa fullCI complains on missing braces, please fix.

mfasDa:
@shahor02 for fullCI I don't get any build log, do you by chance still have them?

shahor02:
@mfasDa no, but you can do e.g. `git show HEAD~1 | grep 'if' | grep '(' | grep -v '{'`

mfasDa:
@shahor02 OK I hope I found the issue. In addition I changed a few cases where instructions depending on the supermodule type were handled by a change of if statements (1-1 porting from aliroot) by a switch statement.

mfasDa:
Now the latest failures seem to be unrelated

shahor02:
@mfasDa the failure in fullCI looks genuine: `reco_NOGPU.log:[14337:EMCALRawToCellConverterSpec]: [13:06:51][ERROR] Exception caught: Cell ID 17664 outside limits. `

mfasDa:
@shahor02 The issue with the fullCI is actually a different bug - there is some index shifting in the DCAL at online level with respect to offline mapping, and the back-shifting was missing. In particular the DCAL 1/3rd supermodules are shifted online in phi direction, leading to cell indices which are then outside the offline index range. In case of simulated raw data the same shift is applied as in raw data from the detector, in order to have the two consistent. It probably wasn't found because no particles were simulated in the DCAL 1/3rd supermodules so far, and we did the commissioning with the EMCAL only. I have pushed the fix for the raw-to-cell converter. The issue is unrelated to the change in the tower-module mapping because this change is not made use of in the raw-to-cell converter.

shahor02 (approved):

PR #6497, opened 2021-06-22
Do not use std::array in ConfigurableParam

shahor02:
Hi @cortesep 
The std containers are not handled by ConfigurableParams, I am changing bitset from std::array to C array.


shahor02:
trivial fix, merging
Will regenerate new RecoParam on the CCDB

PR #6500, opened 2021-06-22
TRD calibration for vDrift and ExB added

martenole:
@aberdnikova and @aschmah 
I took the latest changes from your branch and reformatted it a bit to adapt it a bit better to the other code in O2. The logic is still the same. I have added a few questions inline where I am not sure if we need to change something. Could you please have a look and let me know what you think?
Cheers,
Ole

### DIFF
--- a/Detectors/TRD/calibration/include/TRDCalibration/CalibratorVdExB.h
+++ b/Detectors/TRD/calibration/include/TRDCalibration/CalibratorVdExB.h
@@ -22,6 +22,7 @@
 #include "DataFormatsTRD/AngularResidHistos.h"
 
 #include "Rtypes.h"
+#include "TProfile.h"
 
 #include <array>
 #include <cstdlib>
@@ -31,11 +32,26 @@ namespace o2
 namespace trd
 {
 
+struct FitFunctor {
+  double operator()(const double* par) const;
+  double calculateDeltaAlphaSim(double vdFit, double laFit, double impactAng) const;
+  std::array<std::unique_ptr<TProfile>, constants::MAXCHAMBER> profiles; ///< profile histograms for each TRD chamber
+  double vdPreCorr;                                                      // TODO: these values should eventually be taken from CCDB
+  double laPreCorr;                                                      // TODO: these values should eventually be taken from CCDB
+  int currDet;                                                           ///< the current TRD chamber number
+  float lowerBoundAngleFit;
+  float upperBoundAngleFit;
+};
+
 class CalibratorVdExB final : public o2::calibration::TimeSlotCalibration<o2::trd::AngularResidHistos, o2::trd::AngularResidHistos>
 {
   using Slot = o2::calibration::TimeSlot<o2::trd::AngularResidHistos>;
 
  public:
+  enum ParamIndex : int {
+    LA,
+    VD
+  };
   CalibratorVdExB(size_t nMin = 40'000) : mMinEntries(nMin) {}
   ~CalibratorVdExB() final = default;
 
@@ -44,10 +60,12 @@ class CalibratorVdExB final : public o2::calibration::TimeSlotCalibration<o2::tr
   void finalizeSlot(Slot& slot) final;
   Slot& emplaceNewSlot(bool front, uint64_t tStart, uint64_t tEnd) final;
 
-  // TODO add calibration objects (vDrift and ExB values for each chamber) to this class and implement calibration in finalizeSlot()
+  void initProcessing();
 
  private:
+  bool mInitDone{false}; ///< flag to avoid creating the TProfiles multiple times
   size_t mMinEntries; ///< minimum total number of angular deviations (on average ~3 entries per bin for each TRD chamber)
+  FitFunctor mFitFunctor; ///< used for minimization procedure
   ClassDefOverride(CalibratorVdExB, 1);
 };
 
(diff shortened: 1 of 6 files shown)


Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -13,22 +13,160 @@
...
+
+void CalibratorVdExB::initProcessing()
+{
+  if (mInitDone) {
+    return;
+  }
+  mFitHelper.lowerBoundAngleFit = 80 * TMath::DegToRad();
+  mFitHelper.upperBoundAngleFit = 100 * TMath::DegToRad();
+  mFitHelper.vdPreCorr = 1.546;
+  mFitHelper.laPreCorr = -0.16133; // OS: what about magnetic field changes?
```
aberdnikova: this is a temporary fixed value (for simulated tracklets); in the future it will be replaced with vD and LA from CCDB from previous calibration

martenole: ok thanks then I will put a comment there that these values should eventually be taken from CCDB

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -13,22 +13,160 @@
...
 
+using namespace o2::trd::constants;
+
 namespace o2::trd
 {
 
+std::unique_ptr<TGraph> FitHelper::createTmpGraph(double vdFit, double laFit) const
+{
+  auto graph = std::make_unique<TGraph>();
+  double trdAnodePlane = .0335; // OS: what is this number? Is is actually needed?
```
martenole: what about this one? is that in units of m?

aberdnikova: this is m; Alex says we should keep this value like that for the moment, we are still discussing the precise value.

Detectors/TRD/calibration/include/TRDCalibration/CalibratorVdExB.h
```diff
@@ -30,11 +32,26 @@ namespace o2
 namespace trd
 {
 
+struct FitHelper {
```
tdietel: Is there a reason why this is a struct and not a class? 

It seems this essentially provides the function to be minimized, so FitFunctor might be a better name.

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -13,22 +13,160 @@
...
+#include <string>
+#include <map>
 #include <memory>
 
+using namespace o2::trd::constants;
+
 namespace o2::trd
 {
 
+std::unique_ptr<TGraph> FitHelper::createTmpGraph(double vdFit, double laFit) const
```
tdietel: What is the purpose of this TGraph? It looks like it is used to calculate analytic values of the reference function at certain points, and to then interpolate the points with splines. Wouldn't it be easier to just calculate the analytic values without the overhead of a TGraph and spline interpolation? Or am I missing something?

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -13,22 +13,160 @@
...
+    auto impactAngle = (profiles[currDet]->GetBinCenter(iBin) + 90) * TMath::DegToRad();
+    auto deltaAlpha = profiles[currDet]->GetBinContent(iBin) * TMath::DegToRad();
+    if (TMath::Abs(deltaAlpha) < 1e-7) {
+      continue;
+    }
+    if (impactAngle < lowerBoundAngleFit || impactAngle > upperBoundAngleFit) {
+      continue;
+    }
+    auto deltaAlphaSim = graphTmp->Eval(impactAngle);
+    chi2 += TMath::Power(deltaAlphaSim - deltaAlpha, 2);
```
tdietel: Strictly speaking, chi2 would require normalization with the uncertainties for each bin. But that's ok, if all uncertainties are the same, and if one does not check if the minimal chi2 is O(1).

Detectors/TRD/calibration/include/TRDCalibration/CalibratorVdExB.h
tdietel: The function calculates chi2 between a model and the data points, so it would be better to reflect this in the name to make the code more readable.

Detectors/TRD/calibration/include/TRDCalibration/CalibratorVdExB.h
```diff
@@ -30,11 +32,26 @@ namespace o2
...
+  double operator()(const double* par) const;
+  std::unique_ptr<TGraph> createTmpGraph(double vdFit, double laFit) const;
+  std::array<std::unique_ptr<TProfile>, constants::MAXCHAMBER> profiles; ///< profile histograms for each TRD chamber
```
tdietel, line 38: It seems like the FitHelper always processes one chamber at once, so why not have only one TProfile in here? Another question is why a TProfile is needed at all, it seems this is an extra copy of the data that does not add functionality.

tdietel (changes requested):

martenole:
Hi @tdietel 
I just uploaded the changes implemented by Nastia. As discussed we don't use a `TGraph` anymore but the `TProfile`s remain in so that they can eventually be used by QC.
Cheers,
Ole

tdietel (approved):

PR #6505, opened 2021-06-23
Correction of the numbering of ts array

frmanso:
The numbering of the ts array is now compatible with its size ts[15].


rpezzi (approved):

bovulpes (approved):

shahor02 (approved):
merging as (almost) trivial fix, checked locally

PR #6506, opened 2021-06-23
make ccdb-url a ZDC device option, rather than global of w.flow

shahor02:
@cortesep, @davidrohr , I am changing this since the global option of ZDC workflow conflicts with device option of CPV.
Will merge after the local test is passed.


shahor02:
@davidrohr for me with this PR + already merged https://github.com/AliceO2Group/AliceO2/pull/6504, the FST works. 
Could you also confirm, then will merge.

davidrohr (approved):
thx, works for me

PR #6509, opened 2021-06-23
[CTF] Fix out-of-bounds memory access in EncodedBlocks

MichaelLettrich:
Introduce padding bits to prevent out-of-bounds memory access in Encoded
blocks when writing in-compressible Symbols or data that is not passed
inside the entropy coder.
Additional cleanups.

### DIFF
--- a/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
+++ b/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
@@ -71,6 +71,23 @@ inline T* relocatePointer(const char* oldBase, char* newBase, const T* ptr)
   return (ptr != nullptr) ? reinterpret_cast<T*>(newBase + (reinterpret_cast<const char*>(ptr) - oldBase)) : nullptr;
 }
 
+template <typename source_T, typename dest_T, std::enable_if_t<(sizeof(dest_T) >= sizeof(source_T)), bool> = true>
+inline size_t calculateNDestTElements(size_t sourceElems) noexcept
+{
+  const size_t sizeOfSourceArray = sourceElems * sizeof(source_T);
+  return sizeOfSourceArray / sizeof(dest_T) + (sizeOfSourceArray % sizeof(dest_T) != 0);
+};
+
+template <typename source_T, typename dest_T, std::enable_if_t<(sizeof(dest_T) >= sizeof(source_T)), bool> = true>
+inline size_t calculatePaddedSize(size_t nElems) noexcept
+{
+  const size_t sizeOfSourceT = sizeof(source_T);
+  const size_t sizeOfDestT = sizeof(dest_T);
+
+  // this is equivalent to (sizeOfSourceT / sizeOfDestT) * std::ceil(sizeOfSourceArray/ sizeOfDestT)
+  return (sizeOfDestT / sizeOfSourceT) * calculateNDestTElements<source_T, dest_T>(nElems);
+};
+
 ///>>======================== Auxiliary classes =======================>>
 
 struct ANSHeader {
@@ -373,8 +390,8 @@ class EncodedBlocks
   size_t getFreeSize() const { return mRegistry.getFreeSize(); }
 
   /// expand the storage to new size in bytes
-  template <typename VB>
-  static auto expand(VB& buffer, size_t newsizeBytes);
+  template <typename buffer_T>
+  static auto expand(buffer_T& buffer, size_t newsizeBytes);
 
   /// copy itself to flat buffer created on the fly from the vector
   template <typename V>
@@ -394,15 +411,15 @@ class EncodedBlocks
   static void readFromTree(VD& vec, TTree& tree, const std::string& name, int ev = 0);
 
   /// encode vector src to bloc at provided slot
-  template <typename VE, typename VB>
-  inline void encode(const VE& src, int slot, uint8_t probabilityBits, Metadata::OptStore opt, VB* buffer = nullptr, const void* encoderExt = nullptr)
+  template <typename VE, typename buffer_T>
+  inline void encode(const VE& src, int slot, uint8_t symbolTablePrecision, Metadata::OptStore opt, buffer_T* buffer = nullptr, const void* encoderExt = nullptr)
   {
-    encode(std::begin(src), std::end(src), slot, probabilityBits, opt, buffer, encoderExt);
+    encode(std::begin(src), std::end(src), slot, symbolTablePrecision, opt, buffer, encoderExt);
   }
 
   /// encode vector src to bloc at provided slot
-  template <typename S_IT, typename VB>
-  void encode(const S_IT srcBegin, const S_IT srcEnd, int slot, uint8_t probabilityBits, Metadata::OptStore opt, VB* buffer = nullptr, const void* encoderExt = nullptr);
+  template <typename input_IT, typename buffer_T>
+  void encode(const input_IT srcBegin, const input_IT srcEnd, int slot, uint8_t symbolTablePrecision, Metadata::OptStore opt, buffer_T* buffer = nullptr, const void* encoderExt = nullptr);
 
   /// decode block at provided slot to destination vector (will be resized as needed)
   template <class container_T, class container_IT = typename container_T::iterator>
@@ -582,8 +599,8 @@ size_t EncodedBlocks<H, N, W>::estimateSizeFromMetadata() const
 ///_____________________________________________________________________________
 /// expand the storage to new size in bytes
 template <typename H, int N, typename W>
-template <typename VB>
-auto EncodedBlocks<H, N, W>::expand(VB& buffer, size_t newsizeBytes)
+template <typename buffer_T>
+auto EncodedBlocks<H, N, W>::expand(buffer_T& buffer, size_t newsizeBytes)
 {
 
   auto buftypesize = sizeof(typename std::remove_reference<decltype(buffer)>::type::value_type);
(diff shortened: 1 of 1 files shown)


MichaelLettrich:
@davidrohr this should fix the errors detected by the memory sanitizer. Can you have a look and see if it really is the case?

davidrohr:
I don't have the setup anymore to build with the sanitizer enabled. I could recheck, but can't we simply see from the code that what was fixed by Ruben is also fixed here? There was a lot of renaming, so without looking in detail it is not fully clear to me what is the actual fix.

DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
```diff
@@ -808,59 +815,79 @@ void EncodedBlocks<H, N, W>::encode(const S_IT srcBegin,     // iterator begin o
...
+        // introduce padding in case literals don't align;
+        const size_t nPaddingElements = (nSymbols * sizeof(input_t)) % sizeof(storageBuffer_t);
+        literals.insert(literals.end(), nPaddingElements, {});
+
+        const size_t literalStorageSize = (literals.size() * sizeof(input_t)) / sizeof(storageBuffer_t) + (sizeof(input_t) < sizeof(storageBuffer_t));
+        expandStorage(literalStorageSize);
+        thisBlock->storeLiterals(literalStorageSize, reinterpret_cast<const storageBuffer_t*>(literals.data()));
+      }
+      return nSymbols;
+    }();
```
MichaelLettrich: @davidrohr @shahor02  this introduces padding to the literal symbols

DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
```diff
@@ -808,59 +815,79 @@ void EncodedBlocks<H, N, W>::encode(const S_IT srcBegin,     // iterator begin o
...
+    const size_t nPaddingElements = messageSizeBytes % sizeof(storageBuffer_t);
+    //FIXME(milettri): we should be able to do without an intermediate vector;
     // provided iterator is not necessarily pointer, need to use intermediate vector!!!
-    std::vector<STYP> vtmp(srcBegin, srcEnd);
-    bl->storeData(meta->nDataWords, reinterpret_cast<const W*>(vtmp.data()));
+    std::vector<input_t> vtmp(srcBegin, srcEnd);
+    vtmp.insert(vtmp.end(), nPaddingElements, {});
+    thisBlock->storeData(thisMetadata->nDataWords, reinterpret_cast<const W*>(vtmp.data()));
+
+    *thisMetadata = Metadata{messageLength, 0, sizeof(ransState_t), sizeof(storageBuffer_t), symbolTablePrecision, opt, 0, 0, 0, dataSize, 0};
```
MichaelLettrich: @davidrohr @shahor02 this introduces padding to uncompressed messages.

MichaelLettrich:
@davidrohr  can you tell me how to run the sanitizer myself ? Did you use clang for that or did you use the GCC sanitizer?

shahor02:
Hi @MichaelLettrich 
I am on the phone now, will check bit later. But you don't need sanitizer to validated this, valgrind also shows the problem

davidrohr:
Then valgrind is actually the easier thing to do.

In order to run the sanitizer:
- Add build flags `-O0 -fsanitize=address`.
- Building the OpenCL part and the analysis will fail due to sanitizer errors. You can add `| true` to the CMake add_custom_command rules for those. This will basically not create those files, but they are not needed for the TPC entropy encoding.
- The sanitizer library will not correctly load for the forked processes for some reason, so you need to `LD_PRELOAD=[GCC_LIB_PATH]/libasan.so`
- Then there are some ROOT things causing sanitizer warnings, it should print out some env settings you have to apply.
- Afterwards just run the workflow normally.

DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
```diff
@@ -808,59 +815,79 @@ void EncodedBlocks<H, N, W>::encode(const S_IT srcBegin,     // iterator begin o
...
-                     encoder->getMinSymbol(), encoder->getMaxSymbol(), dictSize, dataSize, literalSize};
+    // store incompressible symbols if any
+    const size_t nLiteralSymbols = [&]() {
+      const size_t nSymbols = literals.size();
+      if (!literals.empty()) {
+        // introduce padding in case literals don't align;
+        const size_t nPaddingElements = (nSymbols * sizeof(input_t)) % sizeof(storageBuffer_t);
+        literals.insert(literals.end(), nPaddingElements, {});
+
+        const size_t literalStorageSize = (literals.size() * sizeof(input_t)) / sizeof(storageBuffer_t) + (sizeof(input_t) < sizeof(storageBuffer_t));
```
shahor02: Same comment as before: here for `sizeof(input_t) < sizeof(storageBuffer_t)` you are always adding 1, while it should be added only if `(literals.size() * sizeof(input_t)) % sizeof(storageBuffer_t) != 0`.
See how I did this in https://github.com/AliceO2Group/AliceO2/blob/1e054d67466173870d49c8726f955f90de4f4f97/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h#L848-L854

DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
```diff
@@ -808,59 +815,79 @@ void EncodedBlocks<H, N, W>::encode(const S_IT srcBegin,     // iterator begin o
...
+                             encoder->getMaxSymbol(),
+                             static_cast<int32_t>(frequencyTable.size()),
+                             dataSize,
+                             static_cast<int32_t>(nLiteralSymbols)};
 
   } else { // store original data w/o EEncoding
-    const size_t szb = messageLength * sizeof(STYP);
-    const int dataSize = szb / sizeof(stream_t) + (sizeof(STYP) < sizeof(stream_t));
+    const size_t messageSizeBytes = messageLength * sizeof(input_t);
+    const int dataSize = messageSizeBytes / sizeof(storageBuffer_t) + (sizeof(input_t) < sizeof(storageBuffer_t));
```
shahor02: same comment here: you add unneeded slot even if when there is no alignment problem.

DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
```diff
@@ -808,59 +815,79 @@ void EncodedBlocks<H, N, W>::encode(const S_IT srcBegin,     // iterator begin o
...
     expandStorage(dataSize);
-    *meta = Metadata{messageLength, 0, sizeof(uint64_t), sizeof(stream_t), probabilityBits, opt, 0, 0, 0, dataSize, 0};
-    //FIXME: no we don't need an intermediate vector.
+    const size_t nPaddingElements = messageSizeBytes % sizeof(storageBuffer_t);
+    //FIXME(milettri): we should be able to do without an intermediate vector;
     // provided iterator is not necessarily pointer, need to use intermediate vector!!!
-    std::vector<STYP> vtmp(srcBegin, srcEnd);
-    bl->storeData(meta->nDataWords, reinterpret_cast<const W*>(vtmp.data()));
+    std::vector<input_t> vtmp(srcBegin, srcEnd);
+    vtmp.insert(vtmp.end(), nPaddingElements, {});
```
shahor02: Since vtmp is a local vector with size given in constructor, `insert` will certainly lead to reallocation.
That's why to avoid it I calculate the padded size in the beginning, create a vector of the final size, then 
copy the data, the padding by 0s is there by construction, see https://github.com/AliceO2Group/AliceO2/blob/1e054d67466173870d49c8726f955f90de4f4f97/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h#L861-L869

shahor02 (commented):
Hi @MichaelLettrich 
I think the padding size calculation should be changed, see inline comments.

To validate with valgring, you can run it with 
````
... | o2-ctf-writer-workflow  --ctf-writer '--child-driver "valgrind "'  $ARGS_ALL  --onlyDet $CTF_DETECTORS 
````
W/o correct fix it will show `invalid reads`

File: DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h [hunk 6]
```diff
@@ -70,6 +70,23 @@ inline T* relocatePointer(const char* oldBase, char* newBase, const T* ptr)
   return (ptr != nullptr) ? reinterpret_cast<T*>(newBase + (reinterpret_cast<const char*>(ptr) - oldBase)) : nullptr;
 }
 
+template <typename source_T, typename dest_T, std::enable_if_t<(sizeof(dest_T) >= sizeof(source_T)), bool> = true>
+inline size_t calculateNDestTElements(size_t sourceElems)
+{
+  const size_t sizeOfSourceArray = sourceElems * sizeof(source_T);
+  return sourceElems / sizeof(dest_T) + (sourceElems % sizeof(dest_T) != 0);
```
shahor02: there should be `sizeOfSourceArray` here instead of `sourceElems`

DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
```diff
@@ -874,22 +891,20 @@ void EncodedBlocks<H, N, W>::encode(const input_IT srcBegin,      // iterator be
...
-    const int dataSize = messageSizeBytes / sizeof(storageBuffer_t) + (sizeof(input_t) < sizeof(storageBuffer_t));
-    // no dictionary needed
-    expandStorage(dataSize);
-    const size_t nPaddingElements = messageSizeBytes % sizeof(storageBuffer_t);
     //FIXME(milettri): we should be able to do without an intermediate vector;
     // provided iterator is not necessarily pointer, need to use intermediate vector!!!
     std::vector<input_t> vtmp(srcBegin, srcEnd);
+    // introduce padding in case literals don't align;
+    const size_t nPaddingElements = calculatePadding<input_t, storageBuffer_t>(messageLength);
     vtmp.insert(vtmp.end(), nPaddingElements, {});
```
shahor02: you still have 1 guaranteed re-allocation of vtmp here, why not booking it from very beginning with the right size?

shahor02 (changes requested):

DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
MichaelLettrich: Indeed. Stupid mistake thank you for spotting it!

DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h
MichaelLettrich: see comment below.

shahor02: which comment? This `//FIXME(milettri): we should be able to do without an intermediate vector`? 
Are you going to change the storeData interface?

MichaelLettrich:
@shahor02  thank you very much for the review especially spotting the stupid resizing mistake.

The vector sizes are interesting. What I have learned in my uni classes on algorithms and data structures was that a dynamic array like `std::vector` has both, a `size` and a `capacity`. `size` is the actual size of data it holds where as `capacity` refers to the size of the underlying allocated memory. To have low amortized cost of reallocation, insertion doubles the size of the container (see e.g. https://mortoray.com/2019/03/26/the-amortized-cost-of-vector-insert-is-3)
The standard itself doesn't state how the `capacity` is exactly affected by the constructor or  `resize`, `reserve` and `insert` apart from the obvious fact that `capacity >= size` after each operation.

Investigating things in compiler explorer yields surprisingly unsurprising results: https://godbolt.org/z/ccqqG5fP8
* `push_back` and `insert` behave as I would have expected from a dynamic array, doubling `capacity` upon resize.
* The constructor, `reserve` and `resize` behave as they intuitively would and really just allocate the exact amount of memory that is requested. I was expecting the same behavior as for `push_back` especially for the constructor. This btw seems to apply for GCC, clang, Intel and even MSVC. What this boils down to is that in case of the `literals` vector, which is created via `push_back`, `insert` will be most likely equivalent to `resize` whereas with the temporary vector it definitely will not. 

For the sake of readability I would have preferred the `insert` option as it seemed clearer to me that the new size will be larger than the old one.

This, at least to me was an interesting experiment and a really good to know fact about our most used container. I made the according changes to prevent resizing. Thank you !

shahor02:
@MichaelLettrich yes, constructor always books exact size requested, only dynamically expansion methods preallocate (by doubling the capacity) the memory if new size would exceed the capacity.

PR #6510, opened 2021-06-23
Updated g4config.in:

ihrivnac:
- Use new Geant4 command for Cerenkov process, as the old one is not working well in Geant4 10.7.
- This should solved the problem reported in JIRA (https://alice.its.cern.ch/jira/browse/O2-2390)


davidrohr:
thx for the fix, testing this locally in FST

davidrohr:
fixes the problem in the FST, thx.

PR #6511, opened 2021-06-23
TRD fix for empty frames

bazinski:
Fix trd raw data for empty frames.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/EventRecord.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/EventRecord.h
@@ -18,6 +18,7 @@
 #include "CommonDataFormat/RangeReference.h"
 #include "FairLogger.h"
 #include "DataFormatsTRD/Tracklet64.h"
+#include "Framework/ProcessingContext.h"
 
 namespace o2::trd
 {
@@ -93,6 +94,9 @@ class EventStorage
   void addTracklets(InteractionRecord& ir, std::vector<Tracklet64>& tracklets);
   void addTracklets(InteractionRecord& ir, std::vector<Tracklet64>::iterator& start, std::vector<Tracklet64>::iterator& end);
   void unpackDataForSending(std::vector<TriggerRecord>& triggers, std::vector<Tracklet64>& tracklets, std::vector<Digit>& digits);
+  void sendData(o2::framework::ProcessingContext& pc);
+  //this could replace by keeing a running total on addition TODO
+  void sumTrackletsDigitsTriggers(uint64_t& tracklets, uint64_t& digits, uint64_t& triggers);
   int sumTracklets();
   int sumDigits();
   std::vector<Tracklet64>& getTracklets(InteractionRecord& ir);
--- a/Detectors/TRD/reconstruction/src/DataReader.cxx
+++ b/Detectors/TRD/reconstruction/src/DataReader.cxx
@@ -35,6 +35,7 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
     {"trd-datareader-headerverbose", VariantType::Bool, false, {"Enable verbose header info"}},
     {"trd-datareader-dataverbose", VariantType::Bool, false, {"Enable verbose data info"}},
     {"trd-datareader-compresseddata", VariantType::Bool, false, {"The incoming data is compressed or not"}},
+    {"ignore-dist-stf", VariantType::Bool, false, {"do not subscribe to FLP/DISTSUBTIMEFRAME/0 message (no lost TF recovery)"}},
     {"trd-datareader-enablebyteswapdata", VariantType::Bool, false, {"byteswap the incoming data, raw data needs it and simulation does not."}}};
 
   o2::raw::HBFUtilsInitializer::addConfigOption(options);
@@ -62,7 +63,7 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   auto compresseddata = cfgc.options().get<bool>("trd-datareader-compresseddata");
   auto headerverbose = cfgc.options().get<bool>("trd-datareader-headerverbose");
   auto dataverbose = cfgc.options().get<bool>("trd-datareader-dataverbose");
-
+  auto askSTFDist = !cfgc.options().get<bool>("ignore-dist-stf");
   std::vector<OutputSpec> outputs;
   outputs.emplace_back("TRD", "TRACKLETS", 0, Lifetime::Timeframe);
   outputs.emplace_back("TRD", "DIGITS", 0, Lifetime::Timeframe);
@@ -90,9 +91,14 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   int idevice = 0;
   //  LOG(info) << "expected incoming data definition : " << inputspec;
   // this is probably never going to be used but would to nice to know hence here.
+  auto orig = o2::header::gDataOriginTRD;
+  std::vector<InputSpec> inputs{{"stf", ConcreteDataTypeMatcher{orig, "RAWDATA"}, Lifetime::Optional}};
+  if (askSTFDist) {
+    inputs.emplace_back("stdDist", "FLP", "DISTSUBTIMEFRAME", 0, Lifetime::Timeframe);
+  }
   workflow.emplace_back(DataProcessorSpec{
     std::string("trd-datareader"), // left as a string cast incase we append stuff to the string
-    select(std::string("x:TRD/" + inputspec).c_str()),
+    inputs,                        //select(std::string("x:TRD/" + inputspec).c_str()),
     outputs,
     algoSpec,
     Options{}});
--- a/Detectors/TRD/reconstruction/src/DataReader.cxx
+++ b/Detectors/TRD/reconstruction/src/DataReader.cxx
@@ -92,7 +92,11 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   //  LOG(info) << "expected incoming data definition : " << inputspec;
   // this is probably never going to be used but would to nice to know hence here.
   auto orig = o2::header::gDataOriginTRD;
-  std::vector<InputSpec> inputs{{"stf", ConcreteDataTypeMatcher{orig, "RAWDATA"}, Lifetime::Optional}};
+  auto inputs = o2::framework::select(inputspec.c_str());
+  for (auto& inp : inputs) {
+    // take care of case where our data is not in the time frame
+    inp.lifetime = Lifetime::Optional;
+  }
   if (askSTFDist) {
     inputs.emplace_back("stdDist", "FLP", "DISTSUBTIMEFRAME", 0, Lifetime::Timeframe);
   }
--- a/Detectors/TRD/reconstruction/src/DataReader.cxx
+++ b/Detectors/TRD/reconstruction/src/DataReader.cxx
@@ -76,23 +76,11 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
 
   WorkflowSpec workflow;
 
-  /*
-   * This is originally replicated from TOF
-     We define at run time the number of devices to be attached
-     to the workflow and the input matching string of the device.
-     This is is done with a configuration string like the following
-     one, where the input matching for each device is provide in
-     comma-separated strings. For instance
-  */
-
-  //  std::stringstream ssconfig(inputspec);
   std::string iconfig;
   std::string inputDescription;
   int idevice = 0;
-  //  LOG(info) << "expected incoming data definition : " << inputspec;
-  // this is probably never going to be used but would to nice to know hence here.
   auto orig = o2::header::gDataOriginTRD;
-  auto inputs = o2::framework::select(inputspec.c_str());
+  auto inputs = o2::framework::select(std::string("x:TRD/" + inputspec).c_str());
   for (auto& inp : inputs) {
     // take care of case where our data is not in the time frame
     inp.lifetime = Lifetime::Optional;
--- a/Detectors/TRD/reconstruction/src/DataReader.cxx
+++ b/Detectors/TRD/reconstruction/src/DataReader.cxx
@@ -29,7 +29,6 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
 {
 
   std::vector<o2::framework::ConfigParamSpec> options{
-    {"trd-datareader-inputspec", VariantType::String, "RAWDATA", {"TRD raw data spec"}},
     {"trd-datareader-output-desc", VariantType::String, "TRDTLT", {"Output specs description string"}},
     {"trd-datareader-verbose", VariantType::Bool, false, {"Enable verbose epn data reading"}},
     {"trd-datareader-headerverbose", VariantType::Bool, false, {"Enable verbose header info"}},
@@ -56,7 +55,6 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   //  o2::conf::ConfigurableParam::updateFromString(cfgc.options().get<std::string>("configKeyValues"));
   // o2::conf::ConfigurableParam::writeINI("o2trdrawreader-workflow_configuration.ini");
 
-  auto inputspec = cfgc.options().get<std::string>("trd-datareader-inputspec");
   //auto outputspec = cfgc.options().get<std::string>("trd-datareader-outputspec");
   auto verbose = cfgc.options().get<bool>("trd-datareader-verbose");
   auto byteswap = cfgc.options().get<bool>("trd-datareader-enablebyteswapdata");
@@ -69,7 +67,6 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   outputs.emplace_back("TRD", "DIGITS", 0, Lifetime::Timeframe);
   outputs.emplace_back("TRD", "TRKTRGRD", 0, Lifetime::Timeframe);
   //outputs.emplace_back("TRD", "FLPSTAT", 0, Lifetime::Timeframe);
-  LOG(info) << "input spec is:" << inputspec;
   LOG(info) << "enablebyteswap :" << byteswap;
   AlgorithmSpec algoSpec;
   algoSpec = AlgorithmSpec{adaptFromTask<o2::trd::DataReaderTask>(compresseddata, byteswap, verbose, headerverbose, dataverbose)};
@@ -80,7 +77,7 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   std::string inputDescription;
   int idevice = 0;
   auto orig = o2::header::gDataOriginTRD;
-  auto inputs = o2::framework::select(std::string("x:TRD/" + inputspec).c_str());
+  auto inputs = o2::framework::select(std::string("x:TRD/RAWDATA").c_str());
   for (auto& inp : inputs) {
     // take care of case where our data is not in the time frame
     inp.lifetime = Lifetime::Optional;
--- a/Detectors/TRD/reconstruction/src/DataReaderTask.cxx
+++ b/Detectors/TRD/reconstruction/src/DataReaderTask.cxx
@@ -43,12 +43,14 @@ void DataReaderTask::init(InitContext& ic)
   ic.services().get<CallbackService>().set(CallbackService::Id::Stop, finishFunction);
 }
 
-void DataReaderTask::sendData(ProcessingContext& pc)
+void DataReaderTask::sendData(ProcessingContext& pc, bool blankframe)
 {
   // mReader.getParsedObjects(mTracklets,mDigits,mTriggers);
-  mReader.getParsedObjects(mTracklets, mDigits, mTriggers);
+  if (!blankframe) {
+    mReader.getParsedObjects(mTracklets, mDigits, mTriggers);
+  }
 
-  LOG(info) << "Sending data onwards with " << mDigits.size() << " Digits and " << mTracklets.size() << " Tracklets and " << mTriggers.size() << " Triggers";
+  LOG(info) << "Sending data onwards with " << mDigits.size() << " Digits and " << mTracklets.size() << " Tracklets and " << mTriggers.size() << " Triggers and blankframe:" << blankframe;
   pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "DIGITS", 0, Lifetime::Timeframe}, mDigits);
   pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "TRACKLETS", 0, Lifetime::Timeframe}, mTracklets);
   pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "TRKTRGRD", 0, Lifetime::Timeframe}, mTriggers);
@@ -65,34 +67,31 @@ void DataReaderTask::run(ProcessingContext& pc)
   auto device = pc.services().get<o2::framework::RawDeviceService>().device();
   auto outputRoutes = pc.services().get<o2::framework::RawDeviceService>().spec().outputs;
   auto fairMQChannel = outputRoutes.at(0).channel;
-  int inputcount = 0;
-  std::vector<InputSpec> dummy{InputSpec{"filter", ConcreteDataTypeMatcher{"FLP", "DISTSUBTIMEFRAME"}, Lifetime::Timeframe}};
+  mDataSpec = o2::header::gDataDescriptionRawData;
+
+  std::vector<InputSpec> dummy{InputSpec{"dummy", ConcreteDataMatcher{"TRD", mDataSpec, 0xDEADBEEF}}};
   // if we see requested data type input with 0xDEADBEEF subspec and 0 payload this means that the "delayed message"
   //   // mechanism created it in absence of real data from upstream. Processor should send empty output to not block the workflow
 
   for (const auto& ref : InputRecordWalker(pc.inputs(), dummy)) {
     const auto dh = o2::framework::DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
-    if (dh->payloadSize == 16) {
+    if (dh->payloadSize == 16 || dh->payloadSize == 0) {
       LOGP(WARNING, "Found input [{}/{}/{:#x}] TF#{} 1st_orbit:{} Payload {} : assuming no payload for all links in this TF",
            dh->dataOrigin.str, dh->dataDescription.str, dh->subSpecification, dh->tfCounter, dh->firstTForbit, dh->payloadSize);
-      sendData(pc); //send the empty tf data.
+      sendData(pc, true); //send the empty tf data.
       return;
     }
     LOG(info) << " matched DEADBEEF";
   }
   //TODO combine the previous and subsequent loops.
-  int inputcounts = 0;
   /* loop over inputs routes */
   for (auto iit = pc.inputs().begin(), iend = pc.inputs().end(); iit != iend; ++iit) {
-    LOG(info) << " looping over inputs " << inputcounts;
-    inputcounts++;
     if (!iit.isValid()) {
       continue;
     }
     /* loop over input parts */
     int inputpartscount = 0;
     for (auto const& ref : iit) {
-      LOG(info) << " looping over parts " << inputpartscount;
       if (mVerbose) {
         const auto dh = DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
         LOGP(info, "Found input [{}/{}/{:#x}] TF#{} 1st_orbit:{} Payload {} : assuming no payload for all links in this TF",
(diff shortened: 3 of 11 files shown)


Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -62,28 +66,45 @@ void DataReaderTask::run(ProcessingContext& pc)
   auto outputRoutes = pc.services().get<o2::framework::RawDeviceService>().spec().outputs;
   auto fairMQChannel = outputRoutes.at(0).channel;
   int inputcount = 0;
+  std::vector<InputSpec> dummy{InputSpec{"filter", ConcreteDataTypeMatcher{"FLP", "DISTSUBTIMEFRAME"}, Lifetime::Timeframe}};
```
shahor02: you should check for the presence of not the `DISTSUBTIMEFRAME` (it is guaranteed to be there if you subscribed to it, and needed just to acknowledge the TF of no other input comes) but to your raw data with `0xdeadbeef` subspec: it will be created by the DPL when the actual input subscribed by wildcarded specification is absent.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -62,28 +66,45 @@ void DataReaderTask::run(ProcessingContext& pc)
...
+  // if we see requested data type input with 0xDEADBEEF subspec and 0 payload this means that the "delayed message"
+  //   // mechanism created it in absence of real data from upstream. Processor should send empty output to not block the workflow
+
+  for (const auto& ref : InputRecordWalker(pc.inputs(), dummy)) {
+    const auto dh = o2::framework::DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
+    if (dh->payloadSize == 16) {
```
shahor02: and its size will 0

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -62,28 +66,45 @@ void DataReaderTask::run(ProcessingContext& pc)
...
     /* loop over input parts */
+    int inputpartscount = 0;
     for (auto const& ref : iit) {
-
+      LOG(info) << " looping over parts " << inputpartscount;
+      if (mVerbose) {
+        const auto dh = DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
+        LOGP(info, "Found input [{}/{}/{:#x}] TF#{} 1st_orbit:{} Payload {} : assuming no payload for all links in this TF",
+             dh->dataOrigin.str, dh->dataDescription.str, dh->subSpecification, dh->tfCounter, dh->firstTForbit, dh->payloadSize);
+      }
```
shahor02: why is this here? Also,  please don't put `LOG(INFO)` in the loop.

Detectors/TRD/reconstruction/src/DataReader.cxx
```diff
@@ -90,9 +91,14 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   int idevice = 0;
   //  LOG(info) << "expected incoming data definition : " << inputspec;
   // this is probably never going to be used but would to nice to know hence here.
+  auto orig = o2::header::gDataOriginTRD;
```
shahor02: Does the reader supposed to handle other raw data than of RAWDATA type? You still have https://github.com/AliceO2Group/AliceO2/blob/cdb1d5b6df8c12a846e2c20a180e6a2049d809f1/Detectors/TRD/reconstruction/src/DataReader.cxx#L32 defined, but now it is ignored in favour of the hardcoded `RAWDATA`. If you will have only `RAWDATA`, then the `inputspec` should be removed. In case you need this flexibility in choosing the DataDescription, than you have to apply the trick which MCH implemented: https://github.com/AliceO2Group/AliceO2/blob/922cac2b59c7ea594baf17ef790972ed54dd6684/Detectors/MUON/MCH/Workflow/src/DataDecoderSpec.cxx#L291-L296 in https://github.com/AliceO2Group/AliceO2/pull/6512

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -62,28 +66,45 @@ void DataReaderTask::run(ProcessingContext& pc)
...
+  std::vector<InputSpec> dummy{InputSpec{"filter", ConcreteDataTypeMatcher{"FLP", "DISTSUBTIMEFRAME"}, Lifetime::Timeframe}};
+  // if we see requested data type input with 0xDEADBEEF subspec and 0 payload this means that the "delayed message"
+  //   // mechanism created it in absence of real data from upstream. Processor should send empty output to not block the workflow
+
+  for (const auto& ref : InputRecordWalker(pc.inputs(), dummy)) {
+    const auto dh = o2::framework::DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
+    if (dh->payloadSize == 16) {
+      LOGP(WARNING, "Found input [{}/{}/{:#x}] TF#{} 1st_orbit:{} Payload {} : assuming no payload for all links in this TF",
+           dh->dataOrigin.str, dh->dataDescription.str, dh->subSpecification, dh->tfCounter, dh->firstTForbit, dh->payloadSize);
+      sendData(pc); //send the empty tf data.
```
shahor02: I see that the `sendData` will call `mReader.getParsedObjects(mTracklets, mDigits, mTriggers);` also when it was triggered by the `0xdeadbeef`. Are you sure the reader will not use buffers of the previous TF and send some garbage (if not crash)?

shahor02 (changes requested):

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: oops, that was erroneously left in from when i was checking somethings.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: Its supposed to clear after being retrieved, but ja you are right, will change to be safer, so it can never be an issue.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: This is the thing i asked on skype last night. Will check again now when the compile is finished.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: if you are refering to the LOGP at line 98 its only for debugging via the command line option, there is a hell of a lot of this as mechanism to debug on the flp for when i get real data. They will mostly eventually come out.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: `[464712:trd-datareader]: [09:44:07][WARN] Found input [FLP/DISTSUBTIMEFRAMEFLP/0x0] TF#0 1st_orbit:0 Payload 16 : assuming no payload for all links in this TF`

So it definitely thinks its 16, I changed the loop to 16 or 0. Am I missing something?

shahor02: Yes, as I wrote above, `FLP/DISTSUBTIMEFRAMEFLP` is always there and its size is indeed 16B, but what you have to check is `TRD/RAWDATA/0xdeadbeef`, not the `DISTSUBTIMEFRAMEFLP`.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
shahor02: I was referring to `LOG(info) << " looping over parts " << inputpartscount;`

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: When I wrote that I got annoyed so just hardcoded it in(something was not working, cant remember what), not thinking that it needs compressed data as well. I would have bumped into that check this afternoon, so thanks for the catch.

Detectors/TRD/reconstruction/src/DataReader.cxx
bazinski: When I wrote that I got annoyed so just hardcoded it in(something was not working, cant remember what), not thinking that it needs compressed data as well. I would have bumped into that check this afternoon, so thanks for the catch.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: ah hell, never try work on "holiday" ok i get it, fixing now.

Detectors/TRD/reconstruction/src/DataReader.cxx
bazinski: done

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: removed

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: fixed

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: done

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -61,29 +67,43 @@ void DataReaderTask::run(ProcessingContext& pc)
...
-  int inputcount = 0;
+  mDataSpec = o2::header::gDataDescriptionRawData;
+
+  std::vector<InputSpec> dummy{InputSpec{"dummy", ConcreteDataMatcher{"TRD", mDataSpec, 0xDEADBEEF}}};
+  // if we see requested data type input with 0xDEADBEEF subspec and 0 payload this means that the "delayed message"
+  //   // mechanism created it in absence of real data from upstream. Processor should send empty output to not block the workflow
+
+  for (const auto& ref : InputRecordWalker(pc.inputs(), dummy)) {
+    const auto dh = o2::framework::DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
+    if (dh->payloadSize == 16 || dh->payloadSize == 0) {
```
shahor02: the size of dummy message must be strictly 0.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -61,29 +67,43 @@ void DataReaderTask::run(ProcessingContext& pc)
...
     }
     /* loop over input parts */
+    int inputpartscount = 0;
     for (auto const& ref : iit) {
-
+      if (mVerbose) {
+        const auto dh = DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
+        LOGP(info, "Found input [{}/{}/{:#x}] TF#{} 1st_orbit:{} Payload {} : assuming no payload for all links in this TF",
+             dh->dataOrigin.str, dh->dataDescription.str, dh->subSpecification, dh->tfCounter, dh->firstTForbit, dh->payloadSize);
+      }
```
shahor02: still here, please remove this

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -61,29 +67,43 @@ void DataReaderTask::run(ProcessingContext& pc)
   auto device = pc.services().get<o2::framework::RawDeviceService>().device();
   auto outputRoutes = pc.services().get<o2::framework::RawDeviceService>().spec().outputs;
   auto fairMQChannel = outputRoutes.at(0).channel;
-  int inputcount = 0;
+  mDataSpec = o2::header::gDataDescriptionRawData;
+
+  std::vector<InputSpec> dummy{InputSpec{"dummy", ConcreteDataMatcher{"TRD", mDataSpec, 0xDEADBEEF}}};
```
shahor02: not sure what is the point of using hardcoded `gDataDescriptionRawData` here and having the same flexible via `auto inputs = o2::framework::select(inputSpec.c_str())` in the creation of the device. If this "flexible" way is different from `gDataDescriptionRawData` (i.e. RAWDATA), the protection will not work.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -39,12 +43,14 @@ void DataReaderTask::init(InitContext& ic)
...
-  mReader.getParsedObjects(mTracklets, mDigits, mTriggers);
+  if (!blankframe) {
+    mReader.getParsedObjects(mTracklets, mDigits, mTriggers);
+  }
 
-  LOG(info) << "Sending data onwards with " << mDigits.size() << " Digits and " << mTracklets.size() << " Tracklets and " << mTriggers.size() << " Triggers";
+  LOG(info) << "Sending data onwards with " << mDigits.size() << " Digits and " << mTracklets.size() << " Tracklets and " << mTriggers.size() << " Triggers and blankframe:" << blankframe;
   pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "DIGITS", 0, Lifetime::Timeframe}, mDigits);
   pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "TRACKLETS", 0, Lifetime::Timeframe}, mTracklets);
   pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "TRKTRGRD", 0, Lifetime::Timeframe}, mTriggers);
```
shahor02: You never reset the mDigits, mTracklets, mTriggers, so, when `blankframe==true` is passed, you will snapshot vectors filled in previous TF.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -114,10 +134,9 @@ void DataReaderTask::run(ProcessingContext& pc)
         mTriggers = mCompressedReader.getIR();
```
shahor02: What is the reason of assigning mTracklets, mDigits, mTriggers from mCompressedReader, if in the `sendData` you reassign them again from `mReader`? 
Also, why are you doing so many copies? The `mReader` and `mCompressedReader` fill vectors, then you assign (copy) these vectors to mTracklets, mDigits, mTriggers members of this device just to create a snapshot from them. 
Why don't you do this directly from the vectors in the readers (using references)?

shahor02 (changes requested):

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: ok will take out the 16 ;-) i left it there to keep me entertained

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: but that is inside the mVerbose, its what I asked about the previous time, as there was a log info in the line previous to this

shahor02: Not sure which question you refer to, but here you are supposed to process a healthy data, which with `mVerbose` you would print `assuming no payload for all links ...`  ?

bazinski: bad paste, removing the text after the :

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: It was originally pointers something was messing up, cant remember what, gave up and made it references. Was planning to come back and make it quicker. Will do this now though.

shahor02: I would avoid having these data members at all, why not creating snapshots directly from the references provided by the mCompressedReader or mReader?

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: now changed to call getParsedObjectsandClear(...)

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: The stuff in mCompressedReader pulling the tracklets/digits/triggerrecords to a local variable was for checking purposes, and updated compressed reader is in a seperate branch to try and keep these updates logical seperate.

shahor02: > The stuff in mCompressedReader pulling the tracklets/digits/triggerrecords to a local variable was for checking purposes, and updated compressed reader is in a seperate branch to try and keep these updates logical seperate.

Sorry, did not understand what you mean, so, can we avoid extra copy?

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
shahor02: I don't see any change, did you forget to push?

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: yes implementing in this pr.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: yes sorry forgot to push.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
shahor02: What about this? If the input can be anything different from the `RAWDATA`, this protection will not work. Do you have such inputs?

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -39,12 +43,19 @@ void DataReaderTask::init(InitContext& ic)
...
-
-  LOG(info) << "Sending data onwards with " << mDigits.size() << " Digits and " << mTracklets.size() << " Tracklets and " << mTriggers.size() << " Triggers";
+  if (!blankframe) {
+    mReader.getParsedObjectsandClear(mTracklets, mDigits, mTriggers);
+  } else {
+    //ensure the objects we are sending back are indeed blank.
+    if (mTracklets.size() != 0)
+      LOG(warn) << "Sending data and tracklets should be zero for an empty frame but we have " << mTracklets.size() << " tracklets";
+    if (mTriggers.size() != 0)
+      LOG(warn) << "Sending data and trigger should be zero for an empty frame but we have " << mTriggers.size() << " tracklets";
```
shahor02: How they can be empty if you never clean them? 
Also, I did not understand, do you want to postpone getting rid of intermediate vectors to other PR?

bazinski: I got stuck with documents for cape town, finalising this now.

bazinski: There are more optimisations to come but various excessive copies are removed.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: Its now passed from the DataReader to the the DataReadTask.

bazinski:
@shahor02 if you are happy can we merge this once the ci is finished, although i am yet to get it all green, with failures in other places.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -61,63 +73,77 @@ void DataReaderTask::run(ProcessingContext& pc)
   auto device = pc.services().get<o2::framework::RawDeviceService>().device();
   auto outputRoutes = pc.services().get<o2::framework::RawDeviceService>().spec().outputs;
   auto fairMQChannel = outputRoutes.at(0).channel;
-  int inputcount = 0;
+  //auto datadesc = (mUserDataDescription == o2::header::gDataDescriptionInvalid) ? o2::header::gDataDescriptionRawData : mUserDataDescription;
+  auto datadesc = (mDataDesc == o2::header::gDataDescriptionInvalid.str) ? o2::header::gDataDescriptionRawData : mUserDataDescription;
```
shahor02: I don't see where do you set the `mUserDataDescription` (except initialization by gDataDescriptionInvalid) but what do you need it at all? Don't you have all needed info in the `mDataDesc` ? And I don't see how do you use at all this datadesc? You still hardcode the `RAWDATA`.
Could you explain in which cases the input will *not* be the `RAWDATA`, what it is will be and who will handle the `RAWDATA` in this case (I guess the StfBuilder always sends the `RAWDATA`?) ?

Concerning the treatment of `trd-datareader-inputspec` option: in the DataReader you are treating it is a global workflow option, in the DataReaderTask.cxx: as a local device option. This may pose a problem since in general these are different type of options. Please don't mix them, since it is in reality a workflow option (defined in DataReader.cxx) and you are parsing it there, you better simply pass its parsed value to DataReaderTask, as you do for other arguments in 
https://github.com/AliceO2Group/AliceO2/blob/3eb01802d28a3667adb1155b2825c4f9c8fd95d1/Detectors/TRD/reconstruction/src/DataReader.cxx#L74

Finally, we are placing the workflows and DPL divices classes in separate workflow directory, could eventually (in different PR) move DataReader and DataReaderTask to such dir.?

DataFormats/Detectors/TRD/include/DataFormatsTRD/EventRecord.h
```diff
@@ -93,6 +94,9 @@ class EventStorage
   void addTracklets(InteractionRecord& ir, std::vector<Tracklet64>& tracklets);
   void addTracklets(InteractionRecord& ir, std::vector<Tracklet64>::iterator& start, std::vector<Tracklet64>::iterator& end);
   void unpackDataForSending(std::vector<TriggerRecord>& triggers, std::vector<Tracklet64>& tracklets, std::vector<Digit>& digits);
+  void sendData(o2::framework::ProcessingContext& pc);
```
shahor02: Do I understand correctly that you fill the vectors twice, once in the unpackDataForSending and also in sendData?

DataFormats/Detectors/TRD/include/DataFormatsTRD/EventRecord.h
```diff
@@ -18,6 +18,7 @@
 #include "CommonDataFormat/RangeReference.h"
 #include "FairLogger.h"
 #include "DataFormatsTRD/Tracklet64.h"
+#include "Framework/ProcessingContext.h"
```
shahor02: It is not a good idea to expose framework headers to classes which generate root dictionary. 
This EventRecord is not really a data format which you store or exchange between the devices, it is just a helper class used in the decoding, right? I think it should go to the `O2/Detectors/TRD/reconstruction` and more important, don't generate dictionary for it (remove ClassDefNV, which is anyway wrong since you neither add this header to `o2_target_root_dictionary(DataFormatsTRD` in the DataFormats/Detectors/TRD/CMakeLists.txt nor add these classes to LinkDef.

bazinski: The classdef was there as i originally intended just to serialise it and send it. Will move it to reconstruction. Should I forward declare the processing context as well?

DataFormats/Detectors/TRD/include/DataFormatsTRD/EventRecord.h
bazinski: unpackDataForSending is just for debugging, the production version only uses sendData. It was easier to do it that way when i wanted to get access to all of vectors. I renamed it now to simpy unpackData

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: Original idea was to have RAWDATA and CRAWDATA for without and with the compressor. Will clean up now and only do RAWDATA, the parameter list kept failing for a string or char* or various other things i tried, so simply reverted to putting the option code you see in the init. Will simply change it all to RAWDATA and remove the inputspec option.

shahor02: OK, because if you would expect the CRAWDATA, the protection would need to be moved to the processor which converts RAWDATA to CRAWDATA. The point is to have a protection at the 1st instance where the TF might be missing.

bazinski: cool that will go in another pr

DataFormats/Detectors/TRD/include/DataFormatsTRD/EventRecord.h
shahor02: Yes, thanks, the less you expose framework headers, the better it is. You forgot to remove the ClassDefNV from EventRecord.h

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -37,18 +41,22 @@ void DataReaderTask::init(InitContext& ic)
...
-  pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "TRKTRGRD", 0, Lifetime::Timeframe}, mTriggers);
-  //    pc.outputs().snapshot(Output{o2::header::gDataOriginTRD,"STATS",0,Lifetime::Timerframe},mStats);
+  if (!blankframe) {
+    mReader.buildDPLOutputs(pc); //getParsedObjectsandClear(mTracklets, mDigits, mTriggers);
+  } else {
+    //ensure the objects we are sending back are indeed blank.
+    LOG(info) << "Sending data onwards with " << mDigits.size() << " Digits and " << mTracklets.size() << " Tracklets and " << mTriggers.size() << " Triggers and blankframe:" << blankframe;
+    pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "DIGITS", 0, Lifetime::Timeframe}, mDigits);
+    pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "TRACKLETS", 0, Lifetime::Timeframe}, mTracklets);
+    pc.outputs().snapshot(Output{o2::header::gDataOriginTRD, "TRKTRGRD", 0, Lifetime::Timeframe}, mTriggers);
```
shahor02: if I get it right, now the sending of the real data is done only via `mReader.buildDPLOutputs` while the mDigits, mTracklets and mTriggers are always empty and used on;y to push an empty output?
If so, then better to get rid of these data members and use here local vectors

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -61,63 +69,75 @@ void DataReaderTask::run(ProcessingContext& pc)
...
-        }
-        loopcounter++;
-        // mTracklets.insert(std::end(mTracklets), std::begin(mReader.getTracklets()), std::end(mReader.getTracklets()));
-        // mCompressedDigits.insert(std::end(mCompressedDigits), std::begin(mReader.getCompressedDigits()), std::end(mReader.getCompressedDigits()));
-        //mReader.clearall();
-        if (mVerbose) {
-          LOG(info) << "from parsing received: " << mTracklets.size() << " tracklets and " << mDigits.size() << " compressed digits";
-          LOG(info) << "relevant vectors to read : " << mReader.sumTrackletsFound() << " tracklets and " << mReader.sumDigitsFound() << " compressed digits";
+      if (std::string(headerIn->dataDescription.str) != std::string("DISTSUBTIMEFRAMEFLP")) {
+        if (!mCompressedData) { //we have raw data coming in from flp
```
shahor02, line 107: Now, that you have hardcoded RAWADAT, does this `mCompressedData` make sense? 
Which device is going to send compressed data?

bazinski: The idea was to make it full proof, one has to start the datareader task with a command line of option of  trd-datareader-compresseddata so the added safety of a different data type is not really required. You feel it should be done differently?

shahor02: Could you explain where the compressed data are coming from?

bazinski: datareader connects to crurawreader and compressedrawreader depending on the origin of its data.
So the CompressorTask uses the crurawreader and the DataReaderTask uses both depending data origin.
crurawreader-CompressorTask-send compressed to epn.

crurawreader/compressrawreader - DataReaderTask -- digits/tracklets/index--> ctf

my ascii art foo is weak.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
bazinski: ah good point. changing now.

DataFormats/Detectors/TRD/include/DataFormatsTRD/EventRecord.h
bazinski: my goodness, i am jumping around branches and took it out of the wrong one. fixing now.

bazinski:
@shahor02 could this be merged? 2 other prs need to be rebased/merged to this pr

PR #6514, opened 2021-06-24
swap HG/LG channels

peressounko:


Detectors/PHOS/reconstruction/src/AltroDecoder.cxx
```diff
@@ -65,8 +65,14 @@ void AltroDecoder::readChannels(const std::vector<uint32_t>& buffer, CaloRawFitt
       if (currentword != 0) {
         LOG(ERROR) << "Channel header mark not found, header word " << currentword;
         short fec = header.mHardwareAddress >> 7 & 0xf; //try to extract FEE number from header
-        if (fec < 0 || fec > 13) {
-          fec = kGeneralSRUErr;
+        short branch = header.mHardwareAddress >> 11 & 0x1;
+        if (branch < 0 || branch > 1) {
+          fec = kGeneralTRUErr + kGeneralSRUErr;
```
shahor02: Result of anything & 0x1 is either 0 or 1, so the condition of this `if` is always false. What is its point?

Detectors/PHOS/reconstruction/src/AltroDecoder.cxx
```diff
@@ -65,8 +65,14 @@ void AltroDecoder::readChannels(const std::vector<uint32_t>& buffer, CaloRawFitt
...
+        } else {
+          if (fec < 0 || fec > 13) {
+            fec = kGeneralSRUErr;
```
shahor02: also here, fec is limited to 0:15 ( = 0xf), so the <0 comparison is meaningless.

Detectors/PHOS/reconstruction/src/AltroDecoder.cxx
```diff
@@ -77,8 +83,14 @@ void AltroDecoder::readChannels(const std::vector<uint32_t>& buffer, CaloRawFitt
...
-      if (fec < 0 || fec > 13) {
-        fec = kGeneralSRUErr;
+      short branch = header.mHardwareAddress >> 11 & 0x1;
+      if (branch < 0 || branch > 1) {
+        fec = kGeneralTRUErr + kGeneralSRUErr;
+      } else {
+        if (fec < 0 || fec > 13) {
+          fec = kGeneralSRUErr;
+        }
+        fec += kGeneralTRUErr * branch;
```
shahor02: same here and below.

shahor02 (changes requested):

Detectors/PHOS/reconstruction/src/AltroDecoder.cxx
peressounko: fixed

Detectors/PHOS/reconstruction/src/AltroDecoder.cxx
peressounko: Rudiment. Fixed in new version

peressounko:
dear @shahor02 , could you please approve? test build/AliceO2/O2/o2/macOS fail for reasons not PHOS-related.

shahor02 (approved):

PR #6517, opened 2021-06-24
DPL: add code to enable multithreaded dispatch

ktf:
Dies due to the service registry not being correctly initialised for
multithreaded usage.


ktf:
Errors unrelated, merging.

PR #6524, opened 2021-06-25
DPL: add api to easily load and act on a plugin

ktf:


ktf:
Previous round of tests failed only on copyright. Merging.

PR #6528, opened 2021-06-25
Move to proper logger + protect file reads [O2-2423]

mpuccio:
Hi @sawenzel, @ktf 
this should fix O2-2423

Cheers,
Max

### DIFF
--- a/Analysis/EventFiltering/centralEventFilterProcessor.cxx
+++ b/Analysis/EventFiltering/centralEventFilterProcessor.cxx
@@ -36,7 +36,7 @@ Document readJsonFile(std::string& config)
 {
   Document d;
   FILE* fp = fopen(config.data(), "rb");
-  if (fp == NULL) {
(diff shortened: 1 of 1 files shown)


mpuccio:
@jgrosseo do you know if there is a way to restart the CI? Currently the o2 one shows that the compilation of o2 is fine (it tests successfully the macros in O2) but somehow it breaks in QC.

ktf:
@mpuccio the quickest way is probably to fix:

```
/sw/SOURCES/O2/6528-slc8_x86-64/0/Analysis/EventFiltering/centralEventFilterProcessor.cxx:39:13: error: use nullptr [modernize-use-nullptr]
```

and it should restart the PR checking.

mpuccio:
@kft thanks, I missed that!

mpuccio:
Test are back green, not sure if OSX test failure is related (to me it does not look like)

PR #6532, opened 2021-06-27
Kr-BoxClusterFinder can now take arbitrary amounts of timebins

HauenderPhilip:
Instead of a fixed 3D map, a std::deque is now used.
See comments in code.

Still to do: Re-implement functions for workflow.

### DIFF
--- a/Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
+++ b/Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
@@ -47,23 +47,29 @@
 /// At least mMinNumberOfNeighbours neighbours (= direct neighbours) has a signal.
 ///
 /// Implementation:
-/// The RAW data is "expanded" for each sector and stored in a big signal
-/// mMapOfAllDigits = 3d vector of all digits
+/// Old ansatz: Store every digit in a big 3D map
+/// Problem: Memory consumption too large if timebins > 10k
+/// New solution: Use rolling map aka "The set of timeslices"
 ///
-/// To make sure that one never has to check if one is inside the sector or not
-/// the arrays are larger than a sector. For the size in row-direction, a constant is used.
-/// For time and pad direction, a number is set (here is room for improvements).
+///      1 2 3 4 5 6 7
 ///
-/// ToDo: Find an elegant way to split the huge map into four (IROC, OROC1, OROC2 and OROC3) smaller maps. Unfortunately, this seems to interfere with the rest of the code.
+///   1  o o o x o o o
+///   2  o o o x o o o
+///      .............
+/// n-1  o o o x o o o
+///   n  o o o x o o o
 ///
-/// How to use:
-/// Load tpcdigits.root
-/// Loop over all events
+/// x-axis: timeslices
+/// y-axis: global pad number
+/// In this example, the fourth timeslice is the interesting one. Here, local maxima and clusters are found. After this slice has been processed, the first slice will be dropped and another slice will be added at the last position (seventh position).
+/// Afterwards, the algorithm looks for clusters in the middle time slice.
+///
+/// How to use (see macro: findKrBoxCluster.C):
+/// Create KrBoxClusterFinder object
+/// Load gainmap (if available)
+/// Make adjustments to clusterfinder (e.g. set min number of neighbours)
 /// Loop over all sectors
-/// Use KrBoxClusterFinder(sector) to create 3D map
-/// Use localMaxima() to find..well the local maxima
-/// localMaxima() yields a vector of a tuple with three int
-/// Loop over this vector and give the tuple to buildCluster()
+/// Use "loopOverSector" function for this
 /// Write to tree or something similar
 
 #ifndef ALICEO2_TPC_KrBoxClusterFinder_H_
@@ -80,6 +86,7 @@
 #include <tuple>
 #include <vector>
 #include <array>
+#include <deque>
 
 namespace o2
 {
@@ -96,6 +103,13 @@ namespace tpc
 
 class KrBoxClusterFinder
 {
+ private:
+  /// Maximum Map Dimensions
+  /// Here is room for improvements
+  static constexpr size_t MaxPads = 138; ///< Size of the map in pad-direction
+  static constexpr size_t MaxRows = 152; ///< Size of the map in row-direction
+  size_t mMaxTimes = 114048;             ///< Size of the map in time-direction
+
  public:
   /// Constructor:
   /// The constructor allocates a three dimensional array (Pad,Row,Time) which is
@@ -108,21 +122,12 @@ class KrBoxClusterFinder
   /// The function expects a CalDet file with a gain map (gain entry for each pad).
   void loadGainMapFromFile(const std::string_view calDetFileName, const std::string_view gainMapName = "GainMap");
 
-  /// Function used in macro to fill the map with all recorded digits
-  void fillAndCorrectMap(std::vector<o2::tpc::Digit>& eventSector, const int sector);
-
-  /// Function to fill single digi
-  void fillADCValue(int cru, int rowInSector, int padInRow, int timeBin, float adcValue);
-
   /// After the map is created, we look for local maxima with this function:
-  std::vector<std::tuple<int, int, int>> findLocalMaxima(bool directFilling = false);
+  std::vector<std::tuple<int, int, int>> findLocalMaxima(bool directFilling = true, const int timeOffset = 0);
 
   /// After finding the local maxima, we can now build clusters around it.
   /// It works according to the explanation at the beginning of the header file.
-  KrCluster buildCluster(int clusterCenterPad, int clusterCenterRow, int clusterCenterTime, bool directFilling = false);
-
-  /// reset the ADC map
-  void resetADCMap();
+  KrCluster buildCluster(int clusterCenterPad, int clusterCenterRow, int clusterCenterTime, bool directFilling = false, const int timeOffset = 0);
 
   /// reset cluster vector
   void resetClusters() { mClusters.clear(); }
@@ -145,6 +150,9 @@ class KrBoxClusterFinder
   /// Set Function for minimal charge required for maxCharge of a cluster
   void setMinQTreshold(int minQThreshold) { mQThresholdMax = minQThreshold; }
 
+  /// Set Function for minimal charge required for maxCharge of a cluster
+  void setMaxTimes(int maxTimes) { mMaxTimes = maxTimes; }
+
   /// Set Function for maximal cluster sizes in different ROCs
   void setMaxClusterSize(int maxClusterSizeRowIROC, int maxClusterSizeRowOROC1, int maxClusterSizeRowOROC2, int maxClusterSizeRowOROC3,
                          int maxClusterSizePadIROC, int maxClusterSizePadOROC1, int maxClusterSizePadOROC2, int maxClusterSizePadOROC3,
@@ -162,6 +170,11 @@ class KrBoxClusterFinder
 
     mMaxClusterSizeTime = maxClusterSizeTime;
   }
+
+  void fillADCValue(int cru, int rowInSector, int padInRow, int timeBin, float adcValue);
+  void resetADCMap();
+
+  void loopOverSector(std::vector<o2::tpc::Digit>& eventSector, const int sector);
 
  private:
   // These variables can be varied
@@ -187,18 +200,14 @@ class KrBoxClusterFinder
   int mSector = -1;                 ///< sector being processed in this instance
   std::unique_ptr<CalPad> mGainMap; ///< Gain map object
 
-  /// Maximum Map Dimensions
-  /// Here is room for improvements
-  static constexpr size_t MaxPads = 138;    ///< Size of the map in pad-direction
-  static constexpr size_t MaxRows = 152;    ///< Size of the map in row-direction
-  static constexpr size_t MaxTimes = 20000; ///< Size of the map in time-direction
-
   /// Values to define ROC boundaries
   static constexpr size_t MaxRowsIROC = 63;  ///< Amount of rows in IROC
   static constexpr size_t MaxRowsOROC1 = 34; ///< Amount of rows in OROC1
   static constexpr size_t MaxRowsOROC2 = 30; ///< Amount of rows in OROC2
   static constexpr size_t MaxRowsOROC3 = 25; ///< Amount of rows in OROC3
 
+  /// Vector of cluster objects
+  /// Used for returning results
   std::vector<o2::tpc::KrCluster> mClusters;
 
   /// Need an instance of Mapper to know position of pads
@@ -206,8 +215,38 @@ class KrBoxClusterFinder
 
   KrCluster mTempCluster; ///< Used to save the cluster data
 
-  /// Here the map is defined where all digits are temporarily stored
-  std::array<std::array<std::array<float, MaxPads>, MaxRows>, MaxTimes> mMapOfAllDigits{};
+  using TimeSliceSector = std::array<std::array<float, MaxPads>, MaxRows>;
+
+  /// A temporary timeslice that gets filled and can be added to the set of timeslices
+  // TimeSliceSector mTempTimeSlice;
+
+  /// Used to keep track of the digit that has to be processed
+  size_t mFirstDigit = 0;
+
+  /// The set of timeslices.
+  /// It consists of 2*mMaxClusterSizeTime + 1 timeslices.
+  /// You can imagine it like this:
+  /// \verbatim
+  ///
+  ///      1 2 3 4 5 6 7
+  ///
+  ///   1  o o o x o o o
+  ///   2  o o o x o o o
+  ///      .............
+  /// n-1  o o o x o o o
+  ///   n  o o o x o o o
+  ///
+  /// \endverbatim
+  ///
+  /// x-axis: Timeslice number
+  /// y-axis: Pad number
+  /// Time slice four is the interesting one. In there, local maxima are found and clusters are built from it. After it is processed, timeslice number 1 will be dropped and another timeslice will be put at the end of the set.
+  std::deque<TimeSliceSector> mSetOfTimeSlices{};
+
+  void createInitialMap(std::vector<o2::tpc::Digit>& eventSector);
+  void popFirstTimeSliceFromMap();
+  void fillADCValueInLastSlice(int cru, int rowInSector, int padInRow, float adcValue);
+  void addTimeSlice(std::vector<o2::tpc::Digit>& eventSector, const int timeSlice);
 
   /// For each ROC, the maximum cluster size has to be chosen
   void setMaxClusterSize(int row);
@@ -215,7 +254,7 @@ class KrBoxClusterFinder
   /// To update the temporary cluster, i.e. all digits are added here
   void updateTempCluster(float tempCharge, int tempPad, int tempRow, int tempTime);
   /// After all digits are assigned to the cluster, the mean and sigmas are calculated here
-  void updateTempClusterFinal();
+  void updateTempClusterFinal(const int timeOffset = 0);
 
   /// Returns sign of val (in a crazy way)
   int signnum(int val) { return (0 < val) - (val < 0); }
--- a/Detectors/TPC/reconstruction/macro/findKrBoxCluster.C
+++ b/Detectors/TPC/reconstruction/macro/findKrBoxCluster.C
@@ -40,8 +40,10 @@ void findKrBoxCluster(int lastTimeBin = 1000, int run = -1, int time = -1, std::
   TFile* fOut = new TFile(outputFile.c_str(), "RECREATE");
   TTree* tClusters = new TTree("Clusters", "Clusters");
 
+  // Create KrBoxClusterFinder object, memory is only allocated once
+  auto clFinder = std::make_unique<o2::tpc::KrBoxClusterFinder>();
+  auto& clusters = clFinder->getClusters();
   // Create a Branch for each sector:
-  std::vector<o2::tpc::KrCluster> clusters;
   tClusters->Branch("cls", &clusters);
   tClusters->Branch("run", &run);
   tClusters->Branch("time", &time);
@@ -52,18 +54,19 @@ void findKrBoxCluster(int lastTimeBin = 1000, int run = -1, int time = -1, std::
     tree->SetBranchAddress(Form("TPCDigit_%zu", iSec), &digitizedSignal[iSec]);
   }
 
-  // Create KrBoxClusterFinder object, memory is only allocated once
-  auto clFinder = std::make_unique<o2::tpc::KrBoxClusterFinder>();
   if (gainMapFile.size()) {
     clFinder->loadGainMapFromFile(gainMapFile);
   }
+
+  // clFinder->setMinNumberOfNeighbours(0);
+  // clFinder->setMinQTreshold(0);
+  clFinder->setMaxTimes(lastTimeBin);
 
   // Now everything can get processed
   // Loop over all events
   for (int iEvent = 0; iEvent < nEntries; ++iEvent) {
     std::cout << iEvent + 1 << "/" << nEntries << std::endl;
     tree->GetEntry(iEvent);
-    // Each event consists of sectors (atm only two)
 
     for (int i = 0; i < 36; i++) {
       auto sector = digitizedSignal[i];
@@ -71,27 +74,7 @@ void findKrBoxCluster(int lastTimeBin = 1000, int run = -1, int time = -1, std::
         continue;
       }
 
-      // Fill map and (if specified) correct with existing gain map
-      clFinder->fillAndCorrectMap(*sector, i);
-
-      // Find all local maxima in sector
-      std::vector<std::tuple<int, int, int>> localMaxima = clFinder->findLocalMaxima();
-
-      // Loop over cluster centers = local maxima
-      for (const std::tuple<int, int, int>& coords : localMaxima) {
-        int padMax = std::get<0>(coords);
-        int rowMax = std::get<1>(coords);
-        int timeMax = std::get<2>(coords);
-
-        if (timeMax >= lastTimeBin) {
-          continue;
-        }
-        // Build total cluster
-        o2::tpc::KrCluster tempCluster = clFinder->buildCluster(padMax, rowMax, timeMax);
-        tempCluster.sector = i;
-
-        clusters.emplace_back(tempCluster);
-      }
+      clFinder->loopOverSector(*sector, i);
     }
     // Fill Tree
     tClusters->Fill();
--- a/Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
+++ b/Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
@@ -38,66 +38,32 @@ void KrBoxClusterFinder::loadGainMapFromFile(const std::string_view calDetFileNa
   LOGP(info, "Loaded gain map object '{}' from file '{}'", calDetFileName, gainMapName);
 }
 
-void KrBoxClusterFinder::resetADCMap()
+void KrBoxClusterFinder::createInitialMap(std::vector<o2::tpc::Digit>& eventSector)
 {
-  // Reset whole map:
-  for (int iTime = 0; iTime < MaxTimes; iTime++) {
-    for (int iRow = 0; iRow < MaxRows; iRow++) {
-      std::fill(mMapOfAllDigits[iTime][iRow].begin(), mMapOfAllDigits[iTime][iRow].end(), 0.);
-      //for (int iPad = 0; iPad < MaxPads; iPad++) {
-      //mMapOfAllDigits[iTime][iRow][iPad] = 0;
-      //}
-    }
+  mSetOfTimeSlices.clear();
+
+  for (int iTimeSlice = 0; iTimeSlice <= 2 * mMaxClusterSizeTime; ++iTimeSlice) {
+    addTimeSlice(eventSector, iTimeSlice);
   }
 }
 
-// Fill the map with all digits.
-// You can pass a CalDet file to it so that the cluster finder can already correct for gain inhomogeneities
-// The CalDet File should contain the relative Gain of each Pad.
-void KrBoxClusterFinder::fillAndCorrectMap(std::vector<o2::tpc::Digit>& eventSector, const int sector)
+void KrBoxClusterFinder::popFirstTimeSliceFromMap()
 {
-  mSector = sector;
-  resetADCMap();
-
-  if (eventSector.size() == 0) {
-    // prevents a segementation fault if the envent contains no data
-    // segementation fault would occure later when trying to dereference the
-    // max/min pointer (which are nullptr if data is empty) empty events should
-    // be catched in the main function, hence, this "if" is no longer necessary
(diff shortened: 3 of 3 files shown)


Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +86,53 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
...
+}
+
+void KrBoxClusterFinder::resetADCMap() {
+  // Has to be reimplemented!
+  return;
+}
+
+void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, int timeBin, float adcValue) {
+  // Has to be reimplemented!
+  return;
```
HauenderPhilip: These functions are needed for the workflow but do not work anymore as they were implemented.

Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
```diff
@@ -206,16 +222,36 @@ class KrBoxClusterFinder
 
   KrCluster mTempCluster; ///< Used to save the cluster data
 
-  /// Here the map is defined where all digits are temporarily stored
-  std::array<std::array<std::array<float, MaxPads>, MaxRows>, MaxTimes> mMapOfAllDigits{};
+  /// A temporary timeslice that gets filled and can be added to the set of timeslices
+  std::array<std::array<float, MaxPads>, MaxRows> mTempTimeSlice;
```
wiechula:
```suggestion
  using TimeSliceSector = std::array<std::array<float, MaxPads>, MaxRows>;
  /// A temporary timeslice that gets filled and can be added to the set of timeslices
  TimeSliceSector mTempTimeSlice;
```

Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
```diff
@@ -206,16 +222,36 @@ class KrBoxClusterFinder
...
+  ///      .............
+  /// n-1  o o o x o o o
+  ///   n  o o o x o o o
+  ///
+  /// \endverbatim
+  /// 
+  /// x-axis: Timeslice number
+  /// y-axis: Pad number
+  /// Time slice four is the interesting one. In there, local maxima are found and clusters are built from it. After it is processed, timeslice number 1 will be dropped and another timeslice will be put at the end of the set.
+  std::deque<std::array<std::array<float, MaxPads>, MaxRows>> mSetOfTimeSlices{};
```
wiechula:
```suggestion
  std::deque<TimeSliceSector> mSetOfTimeSlices{};
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +86,53 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
     adcValue /= correctionFactor;
   }
 
-  mMapOfAllDigits[timeBin][rowInSector][corPad] = adcValue;
+  mTempTimeSlice[rowInSector][corPad] = adcValue;
+}
+
+void KrBoxClusterFinder::fillTempTimeSlice(std::vector<o2::tpc::Digit>& eventSector, const int timeSlice) {
```
wiechula: As it is implemented now, digits will be removed from the data. This means that devices after this one could not us the digits any longer. Also, this implies copy operations, most probably. Computing wise, it might be better to add a digits counter per sector in the header `std::array<size_t, 36> mFirstDigit;` and use:
```
for (mFirstDigit[mSector]; mFirstDigit[mSector] < eventSector.size(); ++mFirstDigit[mSector]) {
  const auto& digit = eventSector[mFirstDigit[mSector]];
...
```
`mFirstDigit` would then need to be reset for each TF, of course. Perhaps in `loopOverSector` `mFirstDigit[mSector] = 0;`. And `mSector` must be set properly before calling `loopOverSector`.
It it is clear that `loopOverSector` will only be called once per instance of `KrBoxClusterFinder`, then of course one could simply use `size_t mFirstDigit` and life would be simpler.

But we could also for the moment keep it like this.

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
wiechula: I think we will need to implement it in a different way. I already have an idea.

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +86,53 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
...
+  return;
+}
+
+const std::vector<KrCluster>& KrBoxClusterFinder::loopOverSector(std::vector<o2::tpc::Digit>& eventSector) {
+  mClusters.clear();
+  createInitialMap(eventSector);
+  for(int iTimeSlice = mMaxClusterSizeTime; iTimeSlice < MaxTimes - mMaxClusterSizeTime; ++iTimeSlice) {
+    findLocalMaxima(true, iTimeSlice);
+    popFirstTimeSliceFromMap();
+    fillTempTimeSlice(eventSector, iTimeSlice);
```
wiechula: Why is this not
```suggestion
    fillTempTimeSlice(eventSector, iTimeSlice + mMaxClusterSizeTime + 1);
```
? I thought the new slice should be added at the end, or? The `createInitialMap` goes up to `2*mMaxClusterSizeTime` including the last.

wiechula (changes requested):

wiechula:
Two more general comments.
1. While it is simple, of course to treat one sector as `max rows * max pads` this brings a size overhead of a factor 1.44. We should think about avoiding this. Should be rather simple to implement.
2. I like the present implementation, since it is very memory efficient. Only time bins which are strictly required for the cluster finding are kept in memory. We sitll might need to change this in future once the final zero suppression logic is in place. Just to keep in mind.

wiechula:
@HauenderPhilip , please fix the formatting and have a look at my questions / suggestions.

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
HauenderPhilip: You are right, yes!
This bug, however, had no big impact for the cluster finding, since it only affected the first few timebins. Nevertheless, I fixed it now. Thanks!

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
HauenderPhilip: I don't get the point, that this removes digits which can not be used afterwards. Is this not just a copy of digits?
The other points are clear and I implemented them. Thank you.

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
HauenderPhilip: Very cool, thanks. Do you want to explain your ideas?

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
wiechula: If we use the code in DPL we will receive the digits in the device. Since the function received the vector by reference, the vector will be empty at the end, since `loopOverSector` will loop over all time bins and `eventSector.erase(digit);` will remove the digit. So the `std::vector<o2::tpc::Digit>` passed from the DPL device to `loopOverSector` will then by empty in the DPL device.

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
wiechula: The DPL device currently fills the data digit by digit. We will now first need to buffer all digits and call `loopOverSector` instead of `fillADCValue`. I can quickly implement it once the changes are committed.

Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
```diff
@@ -163,6 +170,17 @@ class KrBoxClusterFinder
...
+  void fillADCValue(int cru, int rowInSector, int padInRow, int timeBin, float adcValue);
+  void resetADCMap();
+
+  void createInitialMap(std::vector<o2::tpc::Digit>& eventSector);
+  void addTimeSliceToMap();
+  void popFirstTimeSliceFromMap();
+  void fillADCValueInTempSlice(int cru, int rowInSector, int padInRow, float adcValue);
+  void fillTempTimeSlice(std::vector<o2::tpc::Digit>& eventSector, const int timeSlice);
+  void clearTempTimeSlice();
+  const std::vector<KrCluster>& loopOverSector(std::vector<o2::tpc::Digit>& eventSector);
```
wiechula:
```suggestion
  void loopOverSector(std::vector<o2::tpc::Digit>& eventSector);
```
I think it doesn't make sense to return mClusters, since there is already a getter for this.

Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
```diff
@@ -187,35 +205,55 @@ class KrBoxClusterFinder
...
   KrCluster mTempCluster; ///< Used to save the cluster data
 
-  /// Here the map is defined where all digits are temporarily stored
-  std::array<std::array<std::array<float, MaxPads>, MaxRows>, MaxTimes> mMapOfAllDigits{};
+  using TimeSliceSector = std::array<std::array<float, MaxPads>, MaxRows>;
+
+  /// A temporary timeslice that gets filled and can be added to the set of timeslices
+  TimeSliceSector mTempTimeSlice;
+
+  std::array<size_t, 36> mFirstDigit = {0};
```
wiechula:
```suggestion
  size_t mFirstDigit = {0};
```
Since the KrBoxClusterFinder can only handle one sector, it does not make sense to create an array for all sectors here.

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +90,59 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
...
+
+    // We don't need this digit anymore:
+    // digit = eventSector.erase(digit);
+  }
+  return;
+}
+
+const std::vector<KrCluster>& KrBoxClusterFinder::loopOverSector(std::vector<o2::tpc::Digit>& eventSector)
+{
+  mFirstDigit[mSector] = 0;
```
wiechula:
```suggestion
  mFirstDigit = 0;
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +90,59 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
...
 
-  mMapOfAllDigits[timeBin][rowInSector][corPad] = adcValue;
+  mTempTimeSlice[rowInSector][corPad] = adcValue;
+}
+
+void KrBoxClusterFinder::fillTempTimeSlice(std::vector<o2::tpc::Digit>& eventSector, const int timeSlice)
+{
+  // No third argument needed since iterator gets popped
+  for (mFirstDigit[mSector]; mFirstDigit[mSector] < eventSector.size(); ++mFirstDigit[mSector]) {
+    const auto& digit = eventSector[mFirstDigit[mSector]];
```
wiechula:
```suggestion
  for (mFirstDigit; mFirstDigit < eventSector.size(); ++mFirstDigit) {
    const auto& digit = eventSector[mFirstDigit];
```

Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
```diff
@@ -47,23 +46,29 @@
...
+/// How to use (see macro: findKrBoxCluster.C):
+/// Create KrBoxClusterFinder object
+/// Load gainmap (if available)
+/// Make adjustments to clusterfinder (e.g. set min number of neighbours)
 /// Loop over all sectors
-/// Use KrBoxClusterFinder(sector) to create 3D map
-/// Use localMaxima() to find..well the local maxima
-/// localMaxima() yields a vector of a tuple with three int
-/// Loop over this vector and give the tuple to buildCluster()
+/// Use "loopOverSector" function for this
```
wiechula, line 72: Please document somewhere that also `setSector` has to be called and that processing can only be done sector after sector. We could think about getting the sector from the Digits.

Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
```diff
@@ -96,6 +102,13 @@ namespace tpc
 
 class KrBoxClusterFinder
 {
+ private:
+  /// Maximum Map Dimensions
+  /// Here is room for improvements
+  static constexpr size_t MaxPads = 138; ///< Size of the map in pad-direction
+  static constexpr size_t MaxRows = 152; ///< Size of the map in row-direction
+  size_t MaxTimes = 114048;              ///< Size of the map in time-direction
```
wiechula:
```suggestion
  size_t mMaxTimes = 114048;              ///< Size of the map in time-direction
```

Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
```diff
@@ -145,6 +149,9 @@ class KrBoxClusterFinder
   /// Set Function for minimal charge required for maxCharge of a cluster
   void setMinQTreshold(int minQThreshold) { mQThresholdMax = minQThreshold; }
 
+  /// Set Function for minimal charge required for maxCharge of a cluster
+  void setMaxTimes(int maxTimes) { MaxTimes = maxTimes; }
```
wiechula:
```suggestion
  void setMaxTimes(int maxTimes) { mMaxTimes = maxTimes; }
```

Detectors/TPC/reconstruction/include/TPCReconstruction/KrBoxClusterFinder.h
```diff
@@ -163,6 +170,17 @@ class KrBoxClusterFinder
...
 
+  void fillADCValue(int cru, int rowInSector, int padInRow, int timeBin, float adcValue);
+  void resetADCMap();
+
+  void createInitialMap(std::vector<o2::tpc::Digit>& eventSector);
+  void addTimeSliceToMap();
+  void popFirstTimeSliceFromMap();
+  void fillADCValueInTempSlice(int cru, int rowInSector, int padInRow, float adcValue);
+  void fillTempTimeSlice(std::vector<o2::tpc::Digit>& eventSector, const int timeSlice);
+  void clearTempTimeSlice();
```
wiechula: Shouldn't they be private?

Detectors/TPC/reconstruction/macro/findKrBoxCluster.C
```diff
@@ -58,44 +58,28 @@ void findKrBoxCluster(int lastTimeBin = 1000, int run = -1, int time = -1, std::
     clFinder->loadGainMapFromFile(gainMapFile);
   }
```
wiechula, line 60: I propose to remove
```c++
auto clFinder = std::make_unique<o2::tpc::KrBoxClusterFinder>();
```
here, and the above, when the tree is created use replace
```c++
std::vector<o2::tpc::KrCluster> clusters;
```
by
```c++
auto clFinder = std::make_unique<o2::tpc::KrBoxClusterFinder>();
auto& clusters = clFinder->getClusters();
```
This will avoid the unnecessary copy of the clusters.
See also comment below.

Detectors/TPC/reconstruction/macro/findKrBoxCluster.C
```diff
@@ -58,44 +58,28 @@ void findKrBoxCluster(int lastTimeBin = 1000, int run = -1, int time = -1, std::
...
     tree->GetEntry(iEvent);
-    // Each event consists of sectors (atm only two)
 
     for (int i = 0; i < 36; i++) {
       auto sector = digitizedSignal[i];
       if (sector->size() == 0) {
         continue;
       }
+      clFinder->setSector(i);
+      clusters = clFinder->loopOverSector(*sector);
```
wiechula:
```suggestion
      clFinder->loopOverSector(*sector);
```
With my other comments.

Detectors/TPC/reconstruction/macro/findKrBoxCluster.C
```diff
@@ -58,44 +58,28 @@ void findKrBoxCluster(int lastTimeBin = 1000, int run = -1, int time = -1, std::
...
-        }
-        // Build total cluster
-        o2::tpc::KrCluster tempCluster = clFinder->buildCluster(padMax, rowMax, timeMax);
-        tempCluster.sector = i;
-
-        clusters.emplace_back(tempCluster);
-      }
+      // Fill Tree
+      tClusters->Fill();
+      clusters.clear();
```
wiechula:
```suggestion
```
The clean will not be needed, since `clusters` is now a refrence to `mClusters` and the clear will be called in `loopOverSector`.

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +90,59 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
     adcValue /= correctionFactor;
   }
 
-  mMapOfAllDigits[timeBin][rowInSector][corPad] = adcValue;
+  mTempTimeSlice[rowInSector][corPad] = adcValue;
+}
+
+void KrBoxClusterFinder::fillTempTimeSlice(std::vector<o2::tpc::Digit>& eventSector, const int timeSlice)
```
wiechula: Can't one simplify the filling of the slice and get rid of the `mTempTimeSlice`, remove
```
  void addTimeSliceToMap();
  void clearTempTimeSlice();
```

And call this function
```suggestion
void KrBoxClusterFinder::addTimeSlice(std::vector<o2::tpc::Digit>& eventSector, const int timeSlice)
```
together with the canges below.

This will avoid copying and unnecessary complexity.

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +90,59 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
...
+{
```
wiechula, line 85: Add
```c++
mSetOfTimeSlices.emplace_back();
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +90,59 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
...
+    mSector = cru / CRU::CRUperSector;
+
+    const int rowInSector = digit.getRow();
+    const int padInRow = digit.getPad();
+    float adcValue = digit.getChargeFloat();
+
+    fillADCValueInTempSlice(cru, rowInSector, padInRow, adcValue);
+
+    // We don't need this digit anymore:
+    // digit = eventSector.erase(digit);
```
wiechula:
```suggestion
    fillADCValueInLastSlice(cru, rowInSector, padInRow, adcValue);
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -38,66 +37,45 @@ void KrBoxClusterFinder::loadGainMapFromFile(const std::string_view calDetFileNa
...
 
-  // Every row starts at pad zero. But the number of pads in a row is not a constant.
-  // If we would just fill the map naively, we would put pads next to each other, which are not neighbours on the pad plane.
-  // Hence, we need to correct for this:
-  mSector = cru / CRU::CRUperSector;
-  const int pads = mMapperInstance.getNumberOfPadsInRowSector(rowInSector);
-  const int corPad = padInRow - (pads / 2) + (MaxPads / 2);
+void KrBoxClusterFinder::fillADCValueInTempSlice(int cru, int rowInSector, int padInRow, float adcValue)
+{
+  // Correct for pad offset:
```
wiechula, line 59: auto& timeSlice = mSetOfTimeSlices.back();

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -38,66 +37,45 @@ void KrBoxClusterFinder::loadGainMapFromFile(const std::string_view calDetFileNa
...
+  const int padsInRow = mMapperInstance.getNumberOfPadsInRowSector(rowInSector);
+  const int corPad = padInRow - (padsInRow / 2) + (MaxPads / 2);
 
+  // Get correction factor from gain map:
   const auto correctionFactorCalDet = mGainMap.get();
   if (!correctionFactorCalDet) {
-    mMapOfAllDigits[timeBin][rowInSector][corPad] = adcValue;
+    mTempTimeSlice[rowInSector][corPad] = adcValue;
```
wiechula:
```suggestion
    timeSlice[rowInSector][corPad] = adcValue;
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +90,59 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
     adcValue /= correctionFactor;
   }
 
-  mMapOfAllDigits[timeBin][rowInSector][corPad] = adcValue;
+  mTempTimeSlice[rowInSector][corPad] = adcValue;
```
wiechula:
```suggestion
  timeSlice[rowInSector][corPad] = adcValue;
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -38,66 +37,45 @@ void KrBoxClusterFinder::loadGainMapFromFile(const std::string_view calDetFileNa
...
   }
+}
 
-  // Every row starts at pad zero. But the number of pads in a row is not a constant.
-  // If we would just fill the map naively, we would put pads next to each other, which are not neighbours on the pad plane.
-  // Hence, we need to correct for this:
-  mSector = cru / CRU::CRUperSector;
-  const int pads = mMapperInstance.getNumberOfPadsInRowSector(rowInSector);
-  const int corPad = padInRow - (pads / 2) + (MaxPads / 2);
+void KrBoxClusterFinder::fillADCValueInTempSlice(int cru, int rowInSector, int padInRow, float adcValue)
```
wiechula:
```suggestion
void KrBoxClusterFinder::fillADCValueInLastSlice(int cru, int rowInSector, int padInRow, float adcValue)
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +90,59 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
...
+{
+  mFirstDigit[mSector] = 0;
+  mClusters.clear();
+  createInitialMap(eventSector);
+  for (int iTimeSlice = mMaxClusterSizeTime; iTimeSlice < MaxTimes - mMaxClusterSizeTime; ++iTimeSlice) {
+    findLocalMaxima(true, iTimeSlice);
+    popFirstTimeSliceFromMap();
+    fillTempTimeSlice(eventSector, iTimeSlice + mMaxClusterSizeTime + 1);
+    addTimeSliceToMap();
+    clearTempTimeSlice();
```
wiechula:
```suggestion
    addTimeSlice(eventSector, iTimeSlice + mMaxClusterSizeTime + 1);
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -38,66 +37,45 @@ void KrBoxClusterFinder::loadGainMapFromFile(const std::string_view calDetFileNa
...
-      //mMapOfAllDigits[iTime][iRow][iPad] = 0;
-      //}
-    }
+  mSetOfTimeSlices.clear();
+  clearTempTimeSlice();
+
+  for (int iTimeSlice = 0; iTimeSlice <= 2 * mMaxClusterSizeTime; ++iTimeSlice) {
+    fillTempTimeSlice(eventSector, iTimeSlice);
+    addTimeSliceToMap();
+    clearTempTimeSlice();
```
wiechula:
```suggestion
    addTimeSlice(eventSector, iTimeSlice);
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -38,66 +37,45 @@ void KrBoxClusterFinder::loadGainMapFromFile(const std::string_view calDetFileNa
...
-  // Reset whole map:
-  for (int iTime = 0; iTime < MaxTimes; iTime++) {
-    for (int iRow = 0; iRow < MaxRows; iRow++) {
-      std::fill(mMapOfAllDigits[iTime][iRow].begin(), mMapOfAllDigits[iTime][iRow].end(), 0.);
-      //for (int iPad = 0; iPad < MaxPads; iPad++) {
-      //mMapOfAllDigits[iTime][iRow][iPad] = 0;
-      //}
-    }
+  mSetOfTimeSlices.clear();
+  clearTempTimeSlice();
```
wiechula:
```suggestion
```

wiechula (changes requested):

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +77,56 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
...
+    const int cru = digit.getCRU();
+    mSector = cru / CRU::CRUperSector;
+
+    const int rowInSector = digit.getRow();
+    const int padInRow = digit.getPad();
+    float adcValue = digit.getChargeFloat();
+
+    fillADCValueInLastSlice(cru, rowInSector, padInRow, adcValue);
+  }
+  return;
```
wiechula:
```suggestion
```

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -197,128 +213,134 @@ void KrBoxClusterFinder::updateTempCluster(float tempCharge, int tempPad, int te
...
-      } else if (iRow == MaxRowsIROC + MaxRowsOROC1) {
-        mMaxClusterSizePad = mMaxClusterSizePadOROC2;
-        mMaxClusterSizeRow = mMaxClusterSizeRowOROC2;
-      } else if (iRow == MaxRowsIROC + MaxRowsOROC1 + MaxRowsOROC2) {
-        mMaxClusterSizePad = mMaxClusterSizePadOROC3;
-        mMaxClusterSizeRow = mMaxClusterSizeRowOROC3;
-      }
+  // loop over whole mSetOfTimeSlices the find clusters
+  // for (int iTime = 0; iTime < mMaxTimes; iTime++) { //mSetOfTimeSlices.size()
+  int iTime = mMaxClusterSizeTime;
```
wiechula:
```suggestion
  const int iTime = mMaxClusterSizeTime;
```

wiechula (changes requested):
Hi @HauenderPhilip , if I see correctly, you didn't update the `findKrBoxCluster.C` macro, yet. Please do so. Otherwise only minor comments.

HauenderPhilip:
> Hi @HauenderPhilip , if I see correctly, you didn't update the `findKrBoxCluster.C` macro, yet. Please do so. Otherwise only minor comments.

Hey @wiechula,
thanks a lot for your comments. Concerning the `findKrBoxCluster.C` macro: I changed it such that it now uses the `getClusters()` function. So no more unecessary copy operations have to be done. Or what am I missing?

wiechula:
@HauenderPhilip , in the present version of the file https://github.com/AliceO2Group/AliceO2/pull/6532/files#diff-92dc58619d30fc296ddfe8ec8526f65fbc68f64746ffde2735f95403c7fd7d5a I still don't see that my suggestions, mainly
```c++
auto clFinder = std::make_unique<o2::tpc::KrBoxClusterFinder>();
auto& clusters = clFinder->getClusters();
tClusters->Branch("cls", &clusters);
tClusters->Branch("run", &run);
tClusters->Branch("time", &time);
```
were implemented.

There is still a copy, now done differently using
```c++
clusters = clFinder->getClusters();
```
This line should be removed.

HauenderPhilip:
Ok, thanks for the explanation!
I see how it improves the code and I fixed it.

Detectors/TPC/reconstruction/macro/findKrBoxCluster.C
```diff
@@ -52,50 +54,31 @@ void findKrBoxCluster(int lastTimeBin = 1000, int run = -1, int time = -1, std::
...
-        tempCluster.sector = i;
-
-        clusters.emplace_back(tempCluster);
-      }
+      // Fill Tree
+      tClusters->Fill();
     }
-    // Fill Tree
-    tClusters->Fill();
-    clusters.clear();
```
wiechula: I think there is one more conceptual problem. If you clear the clusters in `loopOverSector` one entry in the tree will be done per sector. So one looses the one-to-one relation to the initial tree. I think it would be better to remove the clear from the `loopOverSector`, and call it manually as before. So revert this change.

Detectors/TPC/reconstruction/src/KrBoxClusterFinder.cxx
```diff
@@ -112,7 +77,55 @@ void KrBoxClusterFinder::fillADCValue(int cru, int rowInSector, int padInRow, in
...
+    fillADCValueInLastSlice(cru, rowInSector, padInRow, adcValue);
+  }
+}
+
+void KrBoxClusterFinder::loopOverSector(std::vector<o2::tpc::Digit>& eventSector, const int sector)
+{
+  mFirstDigit = 0;
+  mSector = sector;
+
+  mClusters.clear();
```
wiechula:
```suggestion
```

wiechula (changes requested):
@HauenderPhilip , sorry, one more thing ...

Detectors/TPC/reconstruction/macro/findKrBoxCluster.C
```diff
@@ -52,50 +54,32 @@ void findKrBoxCluster(int lastTimeBin = 1000, int run = -1, int time = -1, std::
...
-        }
-        // Build total cluster
-        o2::tpc::KrCluster tempCluster = clFinder->buildCluster(padMax, rowMax, timeMax);
-        tempCluster.sector = i;
-
-        clusters.emplace_back(tempCluster);
-      }
+      // Fill Tree
+      tClusters->Fill();
+      clusters.clear();
```
wiechula: Sorry, this has to go outside the sector loop.

wiechula (changes requested):

wiechula (dismissed):

wiechula (approved):

PR #6536, opened 2021-06-28
DPL: use provided size for shared memory limiting

ktf:


ktf:
@TimoWilken could you check what's up with the full CI?

ktf:
@TimoWilken there are 28 PRs for build/O2/o2-cs8 how is that actually possible? Do we have a single builder for it?

ktf:
Checked separately on hyperloop. Merging.

PR #6539, opened 2021-06-28
DPL Analysis: port bulk IO patch

ktf:
Adapted to compile with the new code layout. Not yet tested.

### DIFF
--- a/Framework/Core/src/TableTreeHelpers.cxx
+++ b/Framework/Core/src/TableTreeHelpers.cxx
@@ -14,6 +14,7 @@
 
 #include "arrow/type_traits.h"
 #include <arrow/util/key_value_metadata.h>
+#include <TBufferFile.h>
 
 namespace o2::framework
 {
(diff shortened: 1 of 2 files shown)


ktf:
@jgrosseo @pzhristov with this version, I have all the standard tests in hyperloop working fine, with quite a nice speedup and with reasonable results, as far as I can tell. If you could cross check with my tag (O2Physics::20210820-1) and after it passes all the tests here, I would suggest we merge this ASAP, so that we can spot early potential troubles elsewhere.

pzhristov:
@ktf Yes, I agree we should merge it ASAP.

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -443,32 +421,26 @@ TTree* TableToTree::process()
   }
 
 // is used in TreeToTable
-ColumnIterator::ColumnIterator(TTreeReader& reader, const char* colname)
+ColumnIterator::ColumnIterator(TTree* tree, const char* colname)
 {
+  mBranch = tree->GetBranch(colname);
```
pzhristov, line 426: Are we sure the pointers are not 0x0?

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -443,32 +421,26 @@ TTree* TableToTree::process()
...
-  // find branch
-  auto tree = reader.GetTree();
-  if (!tree) {
-    LOGP(FATAL, "Can not locate tree!");
-    return;
-  }
-
-  auto br = tree->GetBranch(colname);
-  if (!br) {
+  if (!mBranch) {
```
pzhristov, line 429: This check comes too late, we already use mBranch.

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -678,99 +606,81 @@ void ColumnIterator::reserve(size_t s)
...
+arrow::Status appendValues(Builder builder, TBuffer& buffer, int64_t size)
+{
+  return builder->AppendValues(reinterpret_cast<T const*>(buffer.GetCurrent()), size, nullptr);
+}
+
+size_t ColumnIterator::push()
 {
   arrow::Status stat;
 
+  TBufferFile buffer{TBuffer::EMode::kWrite, 64 * 1024};
```
pzhristov, line 619: What is 64 * 1024?

jgrosseo:
@ktf This is great progress. Can we wait with merging until the LHCC meetings are over. These are on Monday and Tuesday afternoon.

Framework/Core/src/TableTreeHelpers.cxx
ktf: A round number... :-) we can probably tweak this.

Framework/Core/src/TableTreeHelpers.cxx
ktf: I did not want to change the logic too much, but indeed it looks wrong at the moment.

PR #6540, opened 2021-06-28
Use RecoContainer for TPC interpolation workflow

martenole:
Hi @shahor02 
this is my try to use the RecoContainer for the TPC interpolation workflow. My next step is adding the TRD and also updating the TRD track type.
At the moment I use the `createTracksVariadic` method only to create a vector of GTrackIDs. Should I change this to rather create the seeding tracks instead?
Cheers,
Ole


shahor02:
> Should I change this to rather create the seeding tracks instead?

Why not?

martenole:
Hi @shahor02 
I have two more questions concerning this PR:
1. Inside `RecoContainer::getSingleDetectorRefs` the IDs for the individual contributing tracks are filled. For ITS and TPC these are the actual tracks and for TOF it is the matched cluster. I wonder if there is something I can put also for the TRD? Up to 6 tracklets can be attached. Storing all their indices inside a single index is I think not possible with the limited number of bits available. And we also don't create standalone TRD tracks.
2. I create the seeding tracks directly inside the `createTracksVariadic` method, but I think I should actually perform a refit to the ITS clusters with the most exact track time, is that correct?

shahor02:
Hi @martenole 

>     1. Inside `RecoContainer::getSingleDetectorRefs` the IDs for the individual contributing tracks are filled. For ITS and TPC these are the actual tracks and for TOF it is the matched cluster. I wonder if there is something I can put also for the TRD? Up to 6 tracklets can be attached. Storing all their indices inside a single index is I think not possible with the limited number of bits available. And we also don't create standalone TRD tracks.

Actually, the getSingleDetectorRefs name is somewhat misleading: it takes a GlobalTrackID and provides the references on the "more elementary" contributors on all levels, e.g. for the most complete GlobalTrackID::ITSTPCTRDTOF it will fill in the GlobalIDSet array the slots for the references on the
ITSTPCTRDTOF: index itself (i.e. input)
ITSTPCTRD: index of TrackTRD in the getITSTPCTRDTracks span
ITSTPC: index of MathTPCITS in the getTPCITSTracks span
TOF: index of the TOF cluster in the getTOFClusters span
TRD: see below
TPC: index of the TPC track in the getTPCTracks span
ITS: index of the ITS track in the getITSTracks span

Since the the array with TRD tracklet indices is contained in the TrackTRD, for the TRD slot I would simply fill the same index as filled in the ITSTPCTRD slot. Do you see any problem with that?

> 
>     2. I create the seeding tracks directly inside the `createTracksVariadic` method, but I think I should actually perform a refit to the ITS clusters with the most exact track time, is that correct?

Well, in general, the refit of the global track with TRD (or TOF) constraint should be done already in the corresponding matcher or in a dedicated refitter after all matchings done, but certainly before the TPCInterpolationDPL runs. But in the context of this device I don't understand why do you need the refit. The  `itsTrk->getParamOut()` does not depend on the timestamp assignment, on the opposite end, inward propagation through the TOF and TRD points also. The timestamp, which should come from the TRD trigger info is relevant only for defining the nominal TPC clusters position (to fix the drift distance). Or, have I misunderstood your question?

Do do you refer to the preparation of the seeds from MathTPCITS tracks? They are already refitted in

martenole:
Hi Ruben,
to 1.:
I think using the same index for the TRD contribution is fine. Since there is no more elementary contributor from the TRD side it should not lead to a problem. The index is only used for the array of ITS-TPC-TRD tracks.
to 2.:
For all seeds I start with the outer parameter of the corresponding ITS track, independent of whether they have an outer point or not. I thought that the ITS tracks after the ITS reconstruction are `const` or is the message of ITS tracks replaced by refitted ones after e.g. matching to TPC? The TPC cluster position is at the moment only fixed by the time estimate of the ITS-TPC match. I think this is fine for ITS-TPC matches, but if an outer point is available I should use the time estimate provided by TRD or TOF. Maybe I can use for the seeds a similar struct as in https://github.com/AliceO2Group/AliceO2/blob/bc294424d9073a8e77883c98c331903736aa0e94/Detectors/GlobalTracking/include/GlobalTracking/MatchCosmics.h#L75

shahor02:
@martenole OK, will add TRD tracks to getSingleDetectorRefs.
For (2): yes, all DPL inputs are const, therefore the refit of any track should be done in its producer device, or in an "after-burner" device which creates a clone of the input, but I don't see why this is relevant here. My point was that the ITS track params, including the `getParamOut()`, don't depend on the timestamp assumption, so I did not understand you original question.
For TPC clusters position constraint: I don't think you need a time bracket, you could simply convert the BC to TPC time-bin and fix the cluster position accordingly.

martenole:
@shahor02 ok thanks. Note that I already added TRD to `getSingleDetectorRefs` in this PR. My question to ITS tracks is then also clear. I just need to add the BC info.

shahor02:
> Note that I already added TRD to `getSingleDetectorRefs` in this PR.

Good, did not notice it. Perhaps one should extend the methods getTPCContributorGID and getITSContributorGID too.

martenole:
> Good, did not notice it. Perhaps one should extend the methods getTPCContributorGID and getITSContributorGID too.

Done. And I switched to using the time estimate that I get directly from the `createTracksVariadic` method. That should then take the BC time automatically for the TRD tracks if I understand correctly. For ITS-TPC-TRD tracks the track time estimate is always very close to the time of the corresponding ITS-TPC track. For ITS-TPC-TOF matches however I see significant differences between the TOF time and the ITS-TPC track time (up to 600 us!).  Is there maybe an issue with the time estimate for TOF matches?

shahor02:
Hi @martenole 
I've checked, indeed, there is a problem with ITSTPCTOF, even for those matches which have "correct" label.

@noferini , I've run `o2-tof-matcher-workflow --track-sources "TPC,ITS-TPC" ` with this patch:
````
diff --git a/Detectors/GlobalTracking/src/MatchTOF.cxx b/Detectors/GlobalTracking/src/MatchTOF.cxx
index c3fe595cec8..9c750e2f109 100644
--- a/Detectors/GlobalTracking/src/MatchTOF.cxx
+++ b/Detectors/GlobalTracking/src/MatchTOF.cxx
@@ -999,7 +999,6 @@ void MatchTOF::selectBestMatches()
     mMatchedTracksIndex[trkType][matchingPair.getTrackIndex()] = mMatchedTracks[trkType].size();                      // index of the MatchInfoTOF correspoding to this track
     mMatchedClustersIndex[matchingPair.getTOFClIndex()] = mMatchedTracksIndex[trkType][matchingPair.getTrackIndex()]; // index of the track that was matched to this cluster
     mMatchedTracks[trkType].push_back(matchingPair);                                                                  // array of MatchInfoTOF
-
     // get fit info
     double t0info = 0;
 
@@ -1029,6 +1028,14 @@ void MatchTOF::selectBestMatches()
       }
       mOutTOFLabels[trkType].emplace_back(labelsTOF[0].getTrackID(), labelsTOF[0].getEventID(), labelsTOF[0].getSourceID(), fake);
     }
+    if (trkType == o2::dataformats::MatchInfoTOFReco::ITSTPC) {
+      const auto& tofCL = mTOFClustersArrayInp[  matchingPair.getTOFClIndex() ];
+      const o2::dataformats::TrackTPCITS& trc = mRecoCont->getTPCITSTrack( matchingPair.getEvIdxTrack().getIndex() );
+      LOG(INFO) << "DBG: Times: TOFCl = " << tofCL.getTime()*1e-6 << " Times: MatchTPCITS: " << trc.getTimeMUS().getTimeStamp() << " -> diff = " <<  trc.getTimeMUS().getTimeStamp() - tofCL.getTime()*1e-6
+               << " MC: " << mOutTOFLabels[trkType].back();
+    }
+
+    
     i++;
   }
 }
````
and it produces:
````
[2740706:tof-matcher]: [18:44:29][INFO] DBG: Times: TOFCl = 1186.73 Times: MatchTPCITS: 1274.25 -> diff = 87.5239 MC: [0/62/+ 31733]
[2740706:tof-matcher]: [18:44:29][INFO] DBG: Times: TOFCl = 357.959 Times: MatchTPCITS: 740.551 -> diff = 382.592 MC: [0/16/+ 24850]
[2740706:tof-matcher]: [18:44:29][INFO] DBG: Times: TOFCl = 554.521 Times: MatchTPCITS: 1019.42 -> diff = 464.897 MC: [0/24/+ 50785]
[2740706:tof-matcher]: [18:44:29][INFO] DBG: Times: TOFCl = 232.855 Times: MatchTPCITS: 816.355 -> diff = 583.5 MC: [0/12/+123139]
[2740706:tof-matcher]: [18:44:29][INFO] DBG: Times: TOFCl = 29.4548 Times: MatchTPCITS: 839.105 -> diff = 809.651 MC: [0/2/+ 20133]
[2740706:tof-matcher]: [18:44:29][INFO] DBG: Times: TOFCl = 667.248 Times: MatchTPCITS: 1027.81 -> diff = 360.561 MC: [0/31/+ 59847]
[2740706:tof-matcher]: [18:44:29][INFO] DBG: Times: TOFCl = 740.555 Times: MatchTPCITS: 569.867 -> diff = -170.688 MC: [0/36/+ 30962]
[2740706:tof-matcher]: [18:44:29][INFO] DBG: Times: TOFCl = 667.25 Times: MatchTPCITS: 623.057 -> diff = -44.1927 MC: [0/31/+ 48922]
[2740706:tof-matcher]: [18:44:29][INFO] DBG: Times: TOFCl = 1093.16 Times: MatchTPCITS: 473.283 -> diff = -619.875 MC: [0/59/+ 55784]
````
The correction from the LT-integral is ignored, but it is <20ns anyway, while the differences between the ITS-TPC timestamp and TOF cluster->getTime() are huge. By construction they should be limited to TPC drift time for fake matches and be withing the ITS-TPC time resolution (<1\mus) for correct matches.
I suspect some index is set wrongly  when creating the output, could you please check?
Also, I see that the https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Reconstruction/include/ReconstructionDataFormats/MatchInfoTOF.h#L28-L29 still uses the "tree entry ID" in references, which is a rudiment of very old code. Could you please change this to simply e.g.
````
int mRefTOFCl;  // TOF cluster index
o2::dataformats::GlobalTrackID mRefTrack; // track global index
````
and also fix this part: https://github.com/AliceO2Group/AliceO2/blob/4989c1113ed8c7324126f0ca99ec1f8fe253dfe7/Detectors/GlobalTracking/src/MatchTOF.cxx#L627-L628

**Comment by noferini** [comment]
Hi,
I'm out of office in the next two days.
I will have a look as soon as possible (probably beginning next week).
In the meanwhile I would proceed with Ruben suggestions (I don't want to stop you)

**Comment by martenole** [comment]
Hi @noferini 
thanks! For me this is not blocking right now.
Cheers,
Ole

File: DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx [hunk 1]
```diff
@@ -659,6 +659,14 @@ RecoContainer::GlobalIDSet RecoContainer::getSingleDetectorRefs(GTrackID gidx) c
   GlobalIDSet table;
   auto src = gidx.getSource();
   table[src] = gidx;
+  if (src == GTrackID::ITSTPCTRD) {
+    const auto& parent0 = getITSTPCTRDTrack<o2::trd::TrackTRD>(gidx);
+    const auto& parent1 = getTPCITSTrack(parent0.getRefGlobalTrackId());
+    table[GTrackID::ITSTPC] = parent0.getRefGlobalTrackId();
+    table[GTrackID::ITS] = parent1.getRefITS();
```
shahor02: @martenole please note that now the ITSTPC track can return not only `GTrackID::ITS` ref but also `GTrackID::ITSAB`, which is just a reference on the range of clusters matched to TPC track. So, you should fill the table as
https://github.com/AliceO2Group/AliceO2/blob/e4e31d67a9389b74a00583956891f60e510704ed/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L683

Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/src/TPCInterpolationSpec.cxx
```diff
@@ -37,37 +45,80 @@ void TPCInterpolationDPL::init(InitContext& ic)
...
+  std::vector<float> trkTimes;
+  std::vector<GTrackID> gids;
+  bool processITSTPConly = mProcessITSTPConly; // so that the flag can be used inside the lambda
+  // the creator goes from most complete track (ITS-TPC-TRD-TOF) to least complete one (ITS-TPC)
+  auto creator = [&gidTables, &seeds, &trkTimes, &recoData, &processITSTPConly, &gids](auto& _tr, GTrackID _origID, float t0, float tErr) {
+    if constexpr (std::is_base_of_v<o2::track::TrackParCov, std::decay_t<decltype(_tr)>>) {
+      bool trackGood = true;
+      bool hasOuterPoint = false;
+      auto gidTable = recoData.getSingleDetectorRefs(_origID);
+      if (!gidTable[GTrackID::ITS].isIndexSet() || !gidTable[GTrackID::TPC].isIndexSet()) {
```
shahor02, line 69: This will make the creator to skip the AfterBurner tracks, since instead of ITS they have ITSAB reference, but this should be ok, I would not use the tracks with only a couple of ITS hits for the constraint.

shahor02 (changes requested):
@martenole there is a change to introduce due to the new type of tracks.

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
shahor02: @martenole I will push these RecoContainer modifications in my PR since need them also for other stuff

Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/src/TPCInterpolationSpec.cxx
martenole: I think it would be good to not exclude the ITSAB tracks here, since a few lines below I cut on the minimum number of ITS clusters anyways. So I have changed that now.

martenole:
I think it would be good to merge this (once the checks are through) and continue working on the outlier filtering after I have implemented the TRD refits

Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/src/TPCInterpolationSpec.cxx
shahor02: @martenole The ITSAB may have any number of clusters in 2-7 range (can be limited by options), so, rejection base on Nclusters only will not work. And since there is separate ITS track associated to ITSAB, the further code using ITS track outer param will not work. 
In any case, most probably we will not run the afterburner in the sync. mode, I would simply skip these tracks.

martenole: Ok I see, thanks. I will change this back again then.

martenole:
Hi @shahor02 would it be ok to merge this?

shahor02 (approved):

PR #6549, opened 2021-06-30
DPL: expire shm mem offers

ktf:


ktf:
I've tried this extensively on hyperloop and I am fasttracking it in order to get it inside today's build, to check if it solves the issues on the grid or at least provides better insight. 2 and 4 readers test behave correctly and clang format was used on it. I will revert it tomorrow in case it causes any harm.

PR #6550, opened 2021-06-30
ZDCRaw: do not include executable code in the library

ktf:


aphecetche:
@ktf it's actually more complicated than that : there's also messed up exe deps in ZDC/workflow/CMakeLists.txt. And a workflow included in a TPC library as well. I'm on it and will make a PR with the changes.

aphecetche:
Closing in favour of #6554

PR #6551, opened 2021-06-30
[WIP] DPL: match channel only for part 0, reuse later

knopers8:
@ktf This is what we wrote yesterday.
It allows me to get additional 5% improvement (35% to 40%) in Dispatcher. However, I think there is actually a way to break it (see the comment in the code), so I am not so sure we should merge it, given that the performance benefit is not too big.

### DIFF
--- a/Framework/Core/include/Framework/Output.h
+++ b/Framework/Core/include/Framework/Output.h
@@ -29,6 +29,8 @@ struct Output {
   header::DataOrigin origin;
   header::DataDescription description;
   header::DataHeader::SubSpecificationType subSpec = 0;
+  header::DataHeader::SplitPayloadIndexType splitIndex = 0;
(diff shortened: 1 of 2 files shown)


Framework/Core/include/Framework/Output.h
```diff
@@ -29,6 +29,8 @@ struct Output {
   header::DataOrigin origin;
   header::DataDescription description;
   header::DataHeader::SubSpecificationType subSpec = 0;
+  header::DataHeader::SplitPayloadIndexType splitIndex = 0;
+  header::DataHeader::SplitPayloadPartsType splitTotal = 1;
```
knopers8, line 33: i had to use 1 instead of 0, otherwise any workflow wouldn't start.

PR #6552, opened 2021-06-30
Fix the global angular position of staves in OB layers

mario6829:
The angular position of the staves in the Middle and Outer barrel layers was wrongly set to 0, determining a global rotation of the layers along the beam axis. This was fixed by setting a proper starting angle for each MB/OB layer based on blueprints.


mario6829:
Hi @iouribelikov @shahor02 this is the fix to the global OB layer rotation. Please let me now in case of problems. Thanks. Cheers.

iouribelikov (approved):
@mario6829 Seems to work perfectly fine. Thank you, Mario !

shahor02 (approved):

PR #6554, opened 2021-06-30
[O2-2419] No exe in libraries please

aphecetche:
Executables (e.g. those sources that include `runDataProcessing.h` and thus have a `main` function defined) should not be included in libraries.

This fix should solve the 
```
Undefined symbols for architecture x86_64:
  "_main", referenced from:
     implicit entry/start for main executable
ld: symbol(s) not found for architecture x86_64
clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
```
error in macro compilations on macOS.

Let's see how the tests go...

@cortesep note that following the removal of the `raw-parser` source from the `ZDCRaw` library I had to comment out the `zdc-write-reco` from `Detectors/ZDC/workflow/CMakeLists.txt`. You cannot turn a WorkflowSpec directly into an executable without e.g. defining `defineDataProcessing`. So, assuming the tests go green for this PR, you'd need to modify that `CMakeLists.txt` to recover your `zdc-write-reco` executable.

### DIFF
--- a/Detectors/TPC/workflow/CMakeLists.txt
+++ b/Detectors/TPC/workflow/CMakeLists.txt
@@ -20,7 +20,6 @@ o2_add_library(TPCWorkflow
                        src/ZSSpec.cxx
                        src/CalibProcessingHelper.cxx
                        src/ClusterSharingMapSpec.cxx
-                       src/FileReaderWorkflow.cxx
                        src/CalDetMergerPublisherSpec.cxx
                        src/KryptonClustererSpec.cxx
                        src/IDCToVectorSpec.cxx
--- a/Detectors/ZDC/raw/CMakeLists.txt
+++ b/Detectors/ZDC/raw/CMakeLists.txt
@@ -10,13 +10,19 @@
 # or submit itself to any jurisdiction.
 
 o2_add_library(ZDCRaw
-               SOURCES src/DumpRaw.cxx
-	               src/raw-parser.cxx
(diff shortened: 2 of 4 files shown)


davidrohr (dismissed):
That makes sense, although I didn't understand fully the problem with the ZDC writer, but OK, must be cleaned up. Why can't we clean it up within this PR? Does it require much work?

aphecetche:
@davidrohr maybe not much work no, but right now (until this evening) won't have much time, so wanted to confirm first the issue was solved.... (for the actual problem :  
```
o2_add_executable(write-reco
                   COMPONENT_NAME zdc
                   SOURCES src/ZDCRecoWriterDPLSpec.cxx
                   PUBLIC_LINK_LIBRARIES O2::ZDCWorkflow)
```
simply cannot work with the current [ZDCRecoWriterDPLSpec.cxx]( https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/ZDC/workflow/src/ZDCRecoWriterDPLSpec.cxx) which has no `main`whatsoever (previously was getting it from the library I suspect)

shahor02 (dismissed):

wiechula (dismissed):

shahor02:
@aphecetche indeed, the write-reco had no `WorkflowSpec` definition, just the Device spec. I've opened  https://github.com/aphecetche/AliceO2/pull/10 to your branch, could you merge it?

PR #6557, opened 2021-06-30
[O2-2424] DPL Analysis: Rework slicing algorithm for merged AODs

aalkin:
Solves O2-2424

### DIFF
--- a/Framework/Core/include/Framework/Kernels.h
+++ b/Framework/Core/include/Framework/Kernels.h
@@ -68,28 +68,20 @@ auto sliceByColumn(
   auto v = values.Value(0);
   auto vprev = v;
   auto vnext = v;
+  auto nzeros = 0;
   for (auto i = 0; i < size - 1; ++i) {
+    nzeros = 0;
     vprev = v;
     v = values.Value(i);
     count = counts.Value(i);
     if (v < 0) {
       vnext = values.Value(i + 1);
-      while (vnext - vprev > 1) {
-        makeSlice(offset, 0);
-        ++vprev;
-      }
-      makeSlice(offset, count);
-      offset += count;
-      continue;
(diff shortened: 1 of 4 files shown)


Framework/Core/include/Framework/Kernels.h
```diff
@@ -57,37 +58,52 @@ auto sliceByColumn(char const* key,
...
-    while (nextValue - value > 1) {
-      makeSlice(0);
-      ++value;
+  auto vprev = v;
+  auto vnext = v;
+  for (auto i = 0; i < size - 1; ++i) {
+    vprev = v;
+    v = values.Value(i);
+    count = counts.Value(i);
+    if (v >= 0) {
```
ktf, line 79: Frankly, I would rewrite things with less nesting:

```
for (...) {
if (v < 0) {
      vnext = values.Value(i + 1);
       while (vnext - vprev > 1) {
         makeSlice(offset, 0);
         ++vprev;
       }
       makeSlice(offset, count);
       offset += count;
       continue;
}
if (vprev < 0 || (v - vprev) == 1) {
         makeSlice(offset, count);
         offset += count;
         continue;
 }

while (v - vprev > 1) {
           makeSlice(offset, 0);
           ++vprev;
}
makeSlice(offset, count);
offset += count;
}
```

sawenzel:
Thanks @aalkin. This solves the problem indeed.

sawenzel (dismissed):

aalkin:
@ktf @sawenzel I've found even better solution - we do not, in fact, need to keep unassigned groups available for normal grouping operations. The slicing function will optionally fill and return those in a separate vector which may be needed later for different tasks. This also allows us to return to previous version of grouping that does not have any runtime checks.

sawenzel:
MacOS compilation checked locally.

PR #6558, opened 2021-06-30
DPL: export environment variables when dumping DDS config.

ktf:


davidrohr:
I think this does not respect the template keywords like `timeslice0`

ktf:
@davidrohr fixed

davidrohr (dismissed):
Works, thx
Only thing is that perhaps we should move the string replace stuff in a function to avoid the copy & paste

davidrohr:
@ktf : Could you try to fix the tests before your vacation. We need this in a tag to run with more than 1 GPU using AliECS.

ktf:
yes, sure, in a bit.

ktf:
Ok, there was actually an unrelated (to DDS) race condition in the case LD_PRELOAD=libSegFault.so already. Now fixed.

PR #6560, opened 2021-06-30
[QC-110] Performance improvements in Dispatcher

knopers8:
The Dispatcher::run() performs 45% faster with these changes for 1000 parts and 10% random sampling.
highlights:
 - match inputs, not parts
 - const ref instead of move ref in Dispatcher::send()
 - reuse the found output route when matching
 - compute deviceID once
 - decide once per input (not part) and policy
 - create DataSamplingHeader once per policy

These are all the fixes which don't touch the framework.


knopers8:
@Barthelemy

Barthelemy (approved):

PR #6564, opened 2021-07-01
some changes for cross check with dilepton group's setting

zhangbiao-phy:


vkucera:
@ginnocen Following our discussion about the *p* range, I don't think we want these changes.

zhangbiao-phy:
> @ginnocen Following our discussion about the _p_ range, I don't think we want these changes.


Ok, I just make a draft PR to let you and Gian know what I have changed in the local

ktf:
Any reason for this to still be a draft PR?

PR #6567, opened 2021-07-02
Use expression columns for MC particles

njacazio:


ktf (dismissed):

sawenzel:
@njacazio : Just for my understanding. Is this changing the data model or just the way we compute dynamic properties? (So it will not need changes in AOD conversion ?)

njacazio:
Ciao @sawenzel good call, the AOD conversion should stay unchanged as this affects only dynamic (i.e. derived) columns

aalkin:
@njacazio changing eta to an expression column would lead to division by zero for some of the particles, I've tried it :)

njacazio:
@aalkin is there a protection that we can add to avoid that? Is the division avoided with dynamic columns?

aalkin:
> @aalkin is there a protection that we can add to avoid that? Is the division avoided with dynamic columns?

With dynamic columns this is avoided simply because those are called only for some particles useful in analysis - with a reasonable values of pseudorapidity. Expression would run for every particles, including those very forward ones. Before we have conditional nodes in expressions or an ability to use arrow's null rows I'm afraid there is no workaround.

jgrosseo:
We do not trap FPE in O2 to my understanding. What is the problem here?

aalkin:
> We do not trap FPE in O2 to my understanding. What is the problem here?

This is outside O2, this is done by Gandiva library and it stops on FPE.

njacazio:
@aalkin I see. So, for now, we should leave Eta as a dynamic column right?

aalkin:
> @aalkin I see. So, for now, we should leave Eta as a dynamic column right?

Yes, unfortunately that is correct.

jgrosseo:
This is a severe limitation. When could we have conditional nodes in gandiva?

jgrosseo:
In the meanwhile we may think about adjusting slightly the momenta of the particles where we get fpe. These will be outside the physical region anyway, I suspect...

njacazio:
> > @aalkin I see. So, for now, we should leave Eta as a dynamic column right?
> 
> Yes, unfortunately that is correct.

And we cannot pre-cook variables in dynamic columns and then feed them to expression columns for the heavy lifting right?

aalkin:
> And we cannot pre-cook variables in dynamic columns and then feed them to expression columns for the heavy lifting right?

One could indeed create a non-dynamic column, fill it values using a normal task with safety checks and then append expression column that uses these new numbers.

njacazio:
With these two commits https://github.com/alisw/AliPhysics/pull/18164 and https://github.com/AliceO2Group/AliceO2/pull/6691 we should be able to avoid the FPE until we have the extension in the features

ktf:
Do we have any reason for not merging this? Yes, the PRs in AliPhysics are not yet merged.

njacazio:
Ciao @ktf I don't know, but to be working with AO2Ds we first need the fixes in AliPhysics and O2 workflows

jgrosseo (changes requested):
We have to reproduce the MCs first. Do not merge, yet.

njacazio:
Both PRs pending are now merged, we wait for the test on converted data and then we go on!

aalkin:
@njacazio #6753 just got merged, you can now try making expressions with `ifnode(condition_exp, then_exp, else_exp)`

njacazio:
@aalkin brilliant! I'm on it!! Thanks a lot

njacazio:
I tested that this works on LHC18g4 285011 181_20210626-1348 ready for review!

jgrosseo (commented):
As discussed, let's have hard coded values when it would trigger an FPE.

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -790,31 +790,57 @@ DECLARE_SOA_COLUMN(Vz, vz, float);
...
+                                     0.5f * nlog((nsqrt(aod::mcparticle::px * aod::mcparticle::px +
+                                                        aod::mcparticle::py * aod::mcparticle::py +
+                                                        aod::mcparticle::pz * aod::mcparticle::pz) +
+                                                  aod::mcparticle::pz) /
+                                                 (nsqrt(aod::mcparticle::px * aod::mcparticle::px +
+                                                        aod::mcparticle::py * aod::mcparticle::py +
+                                                        aod::mcparticle::pz * aod::mcparticle::pz) -
+                                                  aod::mcparticle::pz))));
+DECLARE_SOA_EXPRESSION_COLUMN(Pt, pt, float, //! Transverse momentum in GeV/c
+                              nsqrt(aod::mcparticle::px* aod::mcparticle::px +
```
jgrosseo, line 813: clang does not complain here? Please add a space :)

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -790,31 +790,57 @@ DECLARE_SOA_COLUMN(Vz, vz, float);
...
+                                    aod::mcparticle::py * aod::mcparticle::py));
+DECLARE_SOA_EXPRESSION_COLUMN(P, p, float, //! Total momentum in GeV/c
+                              nsqrt(aod::mcparticle::px* aod::mcparticle::px +
```
jgrosseo, line 816: Same here

Framework/Core/include/Framework/AnalysisDataModel.h
njacazio: No, it's the clang format that removes it, I tried already to add it ;)

jgrosseo (approved):

Framework/Core/include/Framework/AnalysisDataModel.h
aalkin: Clang-format does not recognize this as a multiplication, because it is in a macro argument.

PR #41, opened 2015-03-19
New functionality: EventSampler for HLT event header

matthiasrichter:
- synchronization of data publishers
- latency measurement
- consitency checks for aggregated timeframe parts on the EPN

PR #39, opened 2015-03-13
Update of DDS topologies and conditional build

matthiasrichter:
aliceHLTwrapper can now be built with and without DDS support

PR #6568, opened 2021-07-02
Use openMP to parallelize TOF calibration

chiarazampolli:
Done with @noferini .

------------------------------------------------------------------

Fix include of omp

Remove trailing whitespaces

again tabs and trailing whitespaces...

...again?...

...tabs...

Whitespaces

debugging

fix

Removing wrong call

moving functions to source file to use OpenMP

Does not work, but let's commit

move fitter to channel offset from cosmics to TLinearFitter for omp (#39)

calling FixParameter only once

move to local fitter for cosmics (#40)

removing obsolete code - but needs optimization

fix in memory allocation TOF calib (#41)

Some fixes and parallelization of the finalizeSlotWithTracks

Formatting changes

leftover in formatting

clang-format


Common/MathUtils/include/MathUtils/fit.h
```diff
@@ -264,6 +264,131 @@ Double_t fitGaus(const size_t nBins, const T* arr, const T xMin, const T xMax, s
...
+Double_t fitGaus(TLinearFitter& fitter, TMatrixD& mat, const size_t nBins, const T* arr, const T xMin, const T xMax, std::vector<T>& param)
+{
+  // fitGaus where we pass the TLinearFitter and TMatrixD; useful, e.g., when openMP
+  Double_t kTol = mat.GetTol();
+  fitter.StoreData(kFALSE);
+  fitter.ClearPoints();
+  TVectorD par(3);
+  TVectorD sigma(3);
+  TMatrixD A(3, 3);
+  TMatrixD b(3, 1);
```
shahor02: I would not use dynamic objects where fixed size arrays would work. The matrices A, b are always filled but used only for `npoints==3` for a trivial multiplication. I think these structures should be filled only if `npoints==3` and if matrices to be used, it is more appropriate to use root (templated fixed size) SMatrix.

Common/MathUtils/include/MathUtils/fit.h
```diff
@@ -264,6 +264,131 @@ Double_t fitGaus(const size_t nBins, const T* arr, const T xMin, const T xMax, s
...
+      par[0] = res(0, 0);
+      par[1] = res(0, 1);
+      par[2] = res(0, 2);
+      chi2 = -3.;
+    } else {
+      // use fitter for more than three points
+      fitter.Eval();
+      fitter.GetParameters(par);
+      fitter.GetCovarianceMatrix(mat);
+      chi2 = fitter.GetChisquare() / Double_t(npoints);
```
shahor02: why not using proper reduced chi2 definition, i.e. `chi2 = fitter.GetChisquare() / (npoints - 3);` ?

Common/MathUtils/include/MathUtils/fit.h
```diff
@@ -264,6 +264,131 @@ Double_t fitGaus(const size_t nBins, const T* arr, const T xMin, const T xMax, s
...
+  if (rms < kTol) {
+    return -4;
+  }
+
+  param[3] = entries;
+
+  Int_t npoints = 0;
+  for (Int_t ibin = 0; ibin < nBins; ibin++) {
+    Float_t entriesI = arr[ibin];
+    if (entriesI > 1) {
```
shahor02: why `>1` and not `>0`?

Common/MathUtils/include/MathUtils/fit.h
```diff
@@ -264,6 +264,131 @@ Double_t fitGaus(const size_t nBins, const T* arr, const T xMin, const T xMax, s
...
+      res.Mult(A, b);
+      par[0] = res(0, 0);
+      par[1] = res(0, 1);
+      par[2] = res(0, 2);
+      chi2 = -3.;
+    } else {
+      // use fitter for more than three points
+      fitter.Eval();
+      fitter.GetParameters(par);
+      fitter.GetCovarianceMatrix(mat);
```
shahor02: note that this cov.matrix makes no sense for the final params of the fit: it is calculated for the params of pol2, which you then transform to gaussian params.

Detectors/TOF/calibration/include/TOFCalibration/TOFChannelCalibrator.h
```diff
@@ -415,12 +211,54 @@ class TOFChannelCalibrator final : public o2::calibration::TimeSlotCalibration<T
...
+      if (icolumn < 1) {
+        if (kadd) {
+          mStripOffsetFunction.append("+");
+        }
+        mStripOffsetFunction.append(Form("(x > %d && x < %d)", irow + 96, irow + 96 + 1));
+      } else {
+        mStripOffsetFunction.append("-");
+        mStripOffsetFunction.append(Form("(x > %d && x < %d)", irow + 96, irow + 96 + 1));
+      }
+      mStripOffsetFunction.append(Form(") "));
```
shahor02, line 249: out of curiosity, what does this monster function do?

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
```diff
@@ -338,8 +345,312 @@ float TOFChannelData::integral(int ch) const
...
+    int offsetsector = sector * Geo::NSTRIPXSECTOR * Geo::NPADS;
+    for (int istrip = 0; istrip < Geo::NSTRIPXSECTOR; istrip++) {
+      LOG(INFO) << "Processing strip " << istrip;
+      int offsetstrip = istrip * Geo::NPADS + offsetsector;
+      int goodpoints = 0;
+
+      TLinearFitter localFitter(1, mStripOffsetFunction.c_str());
+
+      memset(&fracUnderPeak[0], 0, sizeof(fracUnderPeak));
+      memset(isChON, 0, 96);
```
shahor02: why not defining directly fracUnderPeak[Geo::NPADS]={0} and isChON[96]={0} ?

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
```diff
@@ -338,8 +345,312 @@ float TOFChannelData::integral(int ch) const
...
+      LOG(DEBUG) << "Strip = " << istrip << " fitted by thread = " << ithread << ", NDF" << goodpoints - localFitter.GetNumberFreeParameters();
+
+      if (goodpoints < localFitter.GetNumberFreeParameters()) {
+        LOG(DEBUG) << "Skipped";
+        continue;
+      }
+
+      LOG(DEBUG) << "N real params = " << nparams << ", fitter has " << localFitter.GetNumberFreeParameters() << " free parameters, " << localFitter.GetNumberTotalParameters() << " total parameters, " << localFitter.GetNpoints() << " points";
+      //LOG(INFO) << "Sector = " << sector << ", strip = " << istrip << " fitted by thread = " << ithread << ": ready to fit";
+      printf("Sector = %d, strip = %d, fitted by thread = %d: ready to fit\n", sector, istrip, ithread);
```
shahor02: avoid using printf, use LOG

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
```diff
@@ -338,8 +345,312 @@ float TOFChannelData::integral(int ch) const
...
+      LOG(INFO) << "channel = " << ich << " fitted by thread = " << ithread;
+      if (fitres >= 0) {
+        LOG(DEBUG) << "Channel " << ich << " :: Fit result " << fitres << " Mean = " << fitValues[1] << " Sigma = " << fitValues[2];
+      } else {
+        //        LOG(INFO) << "Channel " << ich << " :: Fit failed with result = " << fitres;
+        continue;
+      }
+
+      if (fitValues[2] < 0) {
+        fitValues[2] = -fitValues[2];
```
shahor02, line 608: how this can be?

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
```diff
@@ -338,8 +345,314 @@ float TOFChannelData::integral(int ch) const
...
+  std::map<std::string, std::string> md;
+  TimeSlewing& ts = mCalibTOFapi->getSlewParamObj(); // we take the current CCDB object, since we want to simply update the offset
+
+  int nbins = c->getNbins();
+  float range = c->getRange();
+  std::vector<int> entriesPerChannel = c->getEntriesPerChannel();
+
+#ifdef WITH_OPENMP
+  if (mNThreads < 1) {
+    mNThreads = std::min(omp_get_max_threads(), NMAXTHREADS);
```
davidrohr, line 373: how many sectors does TOF have? Is it really 20?
Since you parallelize over the sectors, I think it would make sense to set NMAXTHREADS to the number of tof sectors?

davidrohr (dismissed):
without checking in detail, looks ok, just one question below

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
chiarazampolli: They are 18. The 20 is the number that I got on my desktop. I can change it to 18, sure.

davidrohr: If you want to impose a limit, I would put nSectors, 20 doesn't make sense since with parallelization over the sectors you cannot run 20 threads.
But in principle you also do not need to impose a limit

chiarazampolli: Yes, you are right, only 18 threads will be created at max in any case. 20 was just the limit I had on my desktop, so what `omp_get_max_threads()` was returning. Then I thought that if you have something else running, this number will be less, so I used the minimum between the two. But if this is exactly what is done automatically, then I would remove this setting.

noferini (dismissed):

shahor02 (commented):
Hi,
I've added a lot of comments some 10 days age, but forgot to submit them, sorry.
Apart from the technical comments below, a general one: you are adding to general use math library a `fitGaus` function which in fact has nothing to do with Gaus: you are fitting the log of input data to user provided function which will be equivalent to a guassian fit *provided* the user supplied `TLinearFitter& fitter` happens to be pol2. 
What is the point of making a function with such specific requirements a general purpose function? I think it either should be defined a private method of your class, or better, implemented properly w/o user - supplied TLinearFitter: in the end, you are fitting a set of point to pol2, it takes a few lines to do this (one loop and 1 3x3 matrix inversion).

chiarazampolli:
Hi @shahor02 , 

Before going to the specific comments:

> Apart from the technical comments below, a general one: you are adding to general use math library a fitGaus function which in fact has nothing to do with Gaus: you are fitting the log of input data to user provided function which will be equivalent to a guassian fit provided the user supplied TLinearFitter& fitter happens to be pol2.
What is the point of making a function with such specific requirements a general purpose function? I think it either should be defined a private method of your class, or better, implemented properly w/o user - supplied TLinearFitter: in the end, you are fitting a set of point to pol2, it takes a few lines to do this (one loop and 1 3x3 matrix inversion).

Actually the `fitGaus` existed before, we only added another signature of the same. And it is used in several other classes. 

Chiara

Common/MathUtils/include/MathUtils/fit.h
chiarazampolli: Should I work on this, or should the author address the comment?

Common/MathUtils/include/MathUtils/fit.h
chiarazampolli: Same as above, I am not really the author of the function.

Common/MathUtils/include/MathUtils/fit.h
chiarazampolli: Same as above.

Common/MathUtils/include/MathUtils/fit.h
chiarazampolli: And same again.

Detectors/TOF/calibration/include/TOFCalibration/TOFChannelCalibrator.h
chiarazampolli: This function allows to fit the differences between the peaks for each channel. Using such function allows to get better results since we have 96 channels per strip to be calibrated, but we can use 144 points (all pairs between adjacent channels, adjacent in X or Z).

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
chiarazampolli: Yes, sorry. The problem was that with LOG and the thread we did not see the logs for some reason, so then we changed them to printf, but then forgot to change them again. 
Do you know if it is normal that the LOG did not show?

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
chiarazampolli: I actually do not remember :-) @noferini , do you recall if we had seen fitValues[2] (which is the sigma) negative?

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
chiarazampolli: I thought it did not work. Thanks for the suggestion!

chiarazampolli: I want to reset it for every strip, so I need `memset`, no?

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
chiarazampolli: So, now I checked the code. Here the logic is: if you don't set the number of threads from outside, then you want that it parallelize as much as it can. If I just put:
` mNThreads = omp_get_max_threads();`
I guess it will know that the max is 18 from the fact that we parallelize the loop over the 18 sectors, right?

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
shahor02: you could simply define them as local (in the strip loop) variable initialized by {0}, should be faster than calling memset.

shahor02:
Hi @chiarazampolli 

The original general purpose `MathUtils/fit.h` has a `fitGaus` method which indeed fits data to gauss and does not provide the error matrix.  
You are adding another function `fitGaus` which actually does no fit to gaus, (but fits the log of provided data to user-supplied TLinearFitter), extracts wrong covariance matrix and is certainly not optimal from the performance point of view. What's the point of doing this? If you need something quick and dirty, better move this function to your local code and suppress the wrong definition of the output cov.matrix. But, better to add a proper function for fast Gauss fit, usable with openMP.
If you want, I can provide the code.

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
chiarazampolli: done, I will commit all changes in one go, for simplicity.

chiarazampolli:
Hi @shahor02 , 

It is for sure my ignorance, but I am not sure I understand your comment: apart from passing the matrix, which we might even be able to avoid, why is the fitGaus that we added different from the other one? We only pass the TLinearFitter, since this is probably a static object that with threads gets messed up.

I would be very interested to see how to create the fitting function, if this is the way to go. Since next week we will be back to CERN, maybe we can find a slot offline to have a look together? Or if you have no time, you can point me to some documentation/examples. 

Thanks, 

Chiara

shahor02:
Hi @chiarazampolli 

The original version is still (in an inefficient way) fitting the logs of values to pol2 (defined in the function), which is equivalent to fitting the points to gaussian (provided the values are positive) and does not pretend to extract the errors.
With your modification this function 1) does not fit the gaussian if the provided TLinearFitter does not happen to be `pol2`, 
2) the matrix you pretend to extract is wrong by construction. This is not how the common-use utility should behave. 
On top of that, it the speed is so important that you use openMP, better to use optimize the function at 1st place.
I will write simpler code for log-normal fit in one of the next days.

chiarazampolli:
Hi @shahor02 . 
Thanks, now I understand better. I committed for now the other fixes. I would propose to keep this PR open till we have the more appropriate fit procedure. 
Cheers,
Chiara

shahor02:
Hi @chiarazampolli 
could you try to use the new function from https://github.com/AliceO2Group/AliceO2/pull/6659? With it you don't need to prebook bunch of linear fitters and you are free to extract or not a proper cov. matrix.
We can discuss Tue. afternoon, probably after 3:30 pm.

chiarazampolli:
Hi @shahor02 , 

I changed the fitting as we discussed, using your new fitGaus with the medMAD to reject outliers. 

Thanks for this development, it is very nice!

Chiara

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
```diff
@@ -338,8 +345,303 @@ float TOFChannelData::integral(int ch) const
...
+          }
+        } else {
+          localFitter.FixParameter(i, 0.);
+        }
+      }
+
+      LOG(DEBUG) << "Strip = " << istrip << " fitted by thread = " << ithread << ", goodpoints = " << goodpoints << ", number of free parameters = "
+                 << localFitter.GetNumberFreeParameters() << ",  NDF = " << goodpoints - localFitter.GetNumberFreeParameters();
+
+      if (goodpoints <= localFitter.GetNumberFreeParameters()) {
```
chiarazampolli, line 508: Ciao @noferini , 

Note that here I replaced: `if (goodpoints < localFitter.GetNumberFreeParameters())` with `if (goodpoints <= localFitter.GetNumberFreeParameters())` (note the "<=").

chiarazampolli:
Hello @TimoWilken ,

All tests are red, but no error is found... Any hint? Other PRs have issues, see https://github.com/AliceO2Group/AliceO2/pull/6772

Thanks,

Chiara

noferini:
+1

noferini (approved):

noferini:
Dear all,
we are close to deploy the calibration topology flp-epn-cal for TOF.
We need this PR as soon as possible to inject the changes and test the Workflow at P2 if there are no major comments.

davidrohr (approved):

shahor02 (approved):
forgot to merge this, sorry.

PR #6569, opened 2021-07-02
Do not consider type {DPL/EOS} as unknown in the raw proxy

matthiasrichter:
DPL output proxy automatically adds message of type {DPL/EOS} to forward
end-of-stream on all output channels. If another DPL workflow connects
via raw proxy, it handles the channel info headers and can ignore the
actual message.

Also adding an option to test_ExternalFairMQDeviceWorkflow which allows
to drop the subscriber part of the test and instead connect the raw proxy.


ktf (approved):

matthiasrichter:
all CI errors are unrelated, merging

PR #6572, opened 2021-07-02
AOD conversion: Correct indexing of MCparticle to MCCollision

sawenzel:
MCparticles referred to the original eventIDs
from the Geant transport, whereas they should refer
to the MCcollision AOD table entry.

Thanks to Nazar for discussions.


sawenzel:
@nburmaso : Could you check if ok ?

sawenzel:
MacOS compile tested locally.

PR #6573, opened 2021-07-02
[CPV] Check HBF orbit removed 

peressounko:
to fix https://alice.its.cern.ch/jira/browse/O2-2426

### DIFF
--- a/Detectors/CPV/reconstruction/src/RawDecoder.cxx
+++ b/Detectors/CPV/reconstruction/src/RawDecoder.cxx
@@ -63,14 +63,16 @@ RawErrorType_t RawDecoder::readChannels()
       }
       skipUntilNextHeader = false;
       currentBC = header.bc();
+      currentOrbit = header.orbit();
       wordCountFromLastHeader = 0;
       nDigitsAddedFromLastHeader = 0;
-      if (currentOrbit != header.orbit()) { //bad cpvheader
-        LOG(ERROR) << "RawDecoder::readChannels() : "
-                   << "currentOrbit != header.orbit()";
-        mErrors.emplace_back(5, 0, 0, 0, kCPVHEADER_INVALID); //5 is non-existing link with general errors
-        skipUntilNextHeader = true;
-      }
+      // // There is no requirement that event orbit be equal to HBF orbit.
+      // if (currentOrbit != header.orbit()) { //bad cpvheader
+      //   LOG(ERROR) << "RawDecoder::readChannels() : "
+      //              << "currentOrbit="<< currentOrbit<< "  != header.orbit() = "<< header.orbit();
+      //   mErrors.emplace_back(5, 0, 0, 0, kCPVHEADER_INVALID); //5 is non-existing link with general errors
+      //   skipUntilNextHeader = true;
+      // }
     } else {
       if (skipUntilNextHeader) {
         b += 16;
--- a/DataFormats/Detectors/CPV/include/DataFormatsCPV/RawFormats.h
+++ b/DataFormats/Detectors/CPV/include/DataFormatsCPV/RawFormats.h
@@ -86,7 +86,7 @@ class CpvHeader
     mBytes[1] = ((0x010 & 0xf00) >> 8)                                         //bits 11 - 0 trigger id (0x010 = physics trigger)
                 + 0b00100000 * isNoDataExpected + 0b0100000 * isDataContinued; //bit 13 (no data for this trigger) + bit 14 (payload continues from previous page)
     mBytes[2] = (orbitBC.bc & 0x00ff);                                         //bits 27 - 16 bunch crossing
-    mBytes[3] = (orbitBC.bc & 0x0f00) >> 8;                                    //bits 27 - 16 bunch crossing
+    mBytes[3] = (orbitBC.bc & 0xff00) >> 8;                                    //bits 27 - 16 bunch crossing
     mBytes[4] = (orbitBC.orbit & 0x000000ff);                                  //bits 63 - 32 orbit
     mBytes[5] = (orbitBC.orbit & 0x0000ff00) >> 8;                             //bits 63 - 32 orbit
     mBytes[6] = (orbitBC.orbit & 0x00ff0000) >> 16;                            //bits 63 - 32 orbit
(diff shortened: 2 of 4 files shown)


Detectors/CPV/reconstruction/src/RawDecoder.cxx
```diff
@@ -63,14 +63,16 @@ RawErrorType_t RawDecoder::readChannels()
...
       wordCountFromLastHeader = 0;
       nDigitsAddedFromLastHeader = 0;
-      if (currentOrbit != header.orbit()) { //bad cpvheader
-        LOG(ERROR) << "RawDecoder::readChannels() : "
-                   << "currentOrbit != header.orbit()";
-        mErrors.emplace_back(5, 0, 0, 0, kCPVHEADER_INVALID); //5 is non-existing link with general errors
-        skipUntilNextHeader = true;
-      }
+      // // There is no requirement that event orbit be equal to HBF orbit.
+      // if (currentOrbit != header.orbit()) { //bad cpvheader
```
shahor02, line 70: Hi @peressounko 
Why is it so? Isn't the payload after the RDH for given orbit supposed to correspond to triggers in this orbit?
I see that the RawFileWriter::addData is always called with the IR of the trigger, e.g.
https://github.com/AliceO2Group/AliceO2/blob/1bd90382249e1187060f52cc761c997025a9ceff/Detectors/CPV/simulation/src/RawWriter.cxx#L160
Is not your payload supposed to contain the same `trg.getBCData()`?

peressounko: Hi @shahor02 ,
What I see in RawFileWriter, first RDH header with some orbit is created.  Then payload is added with addData(...,IR,...) method where I provide IR. This IR is not used for header modification, but just for comparison that it is not too different from the initial RDH orbit. So I do not see where IR parameters corresponding to next payload are added to header.
Note that several events are normally written to one page, and here not only payload of the first event is checked, but all payloads in this page. And second and third should differ.

shahor02:
Hi @peressounko 

The writer adds a RDH page for the start of run for all kinds or detectors, since they all must acknowledge SOR trigger.
Then, for the CRU detectors in triggered mode it adds RDH only for the orbits which had addData call. For CRU detectors the RDH writes stores only HBF orbit, the specific trigger orbit/bc is stored in the detector payload. In case >1 addData is called for the same orbit, their payload will be stored in under the RDH created for the 1st addData call in this HBF.

Now, debugging the `o2-cpv-digi2raw`, I see that in some case RDH for the 1st orbit was not created, so the payload was stored under the RDH of previous HBF. This was a bug on the RawFileWriter side, I've fixed this in https://github.com/AliceO2Group/AliceO2/pull/6574.
But I see problem also in the CPV RawWriter: it calls `addData` with `preformatted=true` flag. It flags that the detector takes care itslef of the pagination of their payload: i.e. for every link it creates up to `8KB - sizeof(RDH)` payload pages (taking care of carry-over if HBF data does not fit to the single RDH page). In this case all what RawFileWriter does is to dump this payload prepending it by the RDH with IR provided in the addData call (only page counters are updated).
So far only TPC uses this `preformatted` mode.

I don't see corresponding preformatting in the CPV code: you are looping over triggers and for each of them calling addData with the IR of the trigger. So, if there were N triggers in the single HBF, you will ask RawWriter to create 5 RDH pages (usually <8KB).
Are you sure this is what your firmware does?
If not, the trivial remedy would be to set 
https://github.com/AliceO2Group/AliceO2/blob/1bd90382249e1187060f52cc761c997025a9ceff/Detectors/CPV/simulation/src/RawWriter.cxx#L131 to `false`, but I see this comment: https://github.com/AliceO2Group/AliceO2/blob/1bd90382249e1187060f52cc761c997025a9ceff/Detectors/CPV/simulation/src/RawWriter.cxx#L265
and not sure is come special carry-over action is needed or not.
Could you please verify what your firmware does? 
And I would revert the suppression of the `RDH.orbit == trigger.orbit` check, they must be consistent.

shahor02:
Hi @peressounko 
Any news about the  `preformatted=true` flag from the comment above (actually, my phrase `for the CRU detectors in triggered mode it adds RDH only for the orbits which had addData call` was wrong, the HBF is added always, even if it is empty)

peressounko:
Hi @shahor02, I am testing code - it looks like raw decoder now picks up digits from wrong event. May be @sevdokim will help when return.
     Dmitri

sevdokim:
Dear @shahor02 @peressounko 
I did RawWriter to write preformatted data, it's true. The reason is that cpv event is constructed from cpv header, cpv payload and cpv trailer. If we're splitting event somewhere, we need to add correct pair of cpv trailer and cpv header in the place of splitting; but carryOverMethod() is not able to do this so I prepare preformatted data and writing it. 
I just came back to work. Let me 1-2 days, I'll investigate the bug.
Best regards,
Sergey

shahor02:
Hi @sevdokim: why the carryOverMethod cannot handle the headers/trailers? It is intended for that.
You can also go on with "formatted" data, but then you should take care of collecting the triggers which should be on the same RDH page to a single payload, and then sending it via addData.

sevdokim:
@shahor02 one 8KB page can contain several cpv events (i.e. several BCs), every event has it's own header and trailer. But carryOverMethod() assumes same header & trailer for all the payload in the current page (as far as I understood). So just not to invent a bicycle I decided to use preformatted data option: I prepare cpv payload by processing digits event by event from earlier to later and flush it to page if orbit is finished or payload size is not enough to write more data (i.e. > = 8KB - RDHsize - 2-3 GBT words depending on actual case). Such a workflow is in accordance with cpv firmware algorithm of raw data creation.

shahor02:
Hi @sevdokim 
After checking your code again: it seems I was wrong, sorry, I did not notice that you are accumulating the data in the payload of the same page until it reaches max size. Also, as far as understand, when the page is close to 8KB and and you in the middle of the trigger, you do the splitting yourself, so this part should be fine also.
Then, there should be no problem, @peressounko, please just restore the orbit check and this PR can be merged.

Detectors/CPV/reconstruction/src/RawDecoder.cxx
```diff
@@ -63,14 +63,16 @@ RawErrorType_t RawDecoder::readChannels()
       }
       skipUntilNextHeader = false;
       currentBC = header.bc();
+      currentOrbit = header.orbit();
```
sevdokim, line 66: Hi @peressounko Here currentOrbit = mRawReader.getCurrentHBFOrbit() is orbit we read from RDH (see line 52). 
header.orbit() is orbit read from cpv header. In the lines 68-73 I try to compare orbit recorded in RDH with orbit recorded in cpv payload. The added line currentOrbit = header.orbit() will just void this comparison, and we will never know if cpv payload is self-consistent with its RDH.

peressounko: agree. If we revert check of orbit consistency in RDH and trigger header, this should be reverted too.

peressounko:
Closing PR: @sevdokim will revert changes and create his own

PR #6574, opened 2021-07-02
fix in RDH update for CRU det. in trig. mode

shahor02:


shahor02:
macOS failure is unrelated

PR #6576, opened 2021-07-04
TRD: Remove TRDBase/Digit.h

jolopezl:
Removes `TRDBase/Digit.h`. Now the TRD uses `DataFormatsTRD/Digit.h`. 

This pull request needs https://github.com/AliceO2Group/QualityControl/pull/758 merged first so that this change doesn't break QC.


tdietel (dismissed):
Looks like the QC PR was merged already, so we should be ready to go.

shahor02:
@tdietel @jolopezl you should request new QC tag, the failures happen because currently used v1.21.0 still expects `TRDBase/Digits.h`

PR #6577, opened 2021-07-05
DPL Analysis: add unsigned int types to Variant/Configurable

aalkin:
* 8,16,32 and 64bit unsigned ints are added to Variant
* Update serialization handlers to support these types in ConfigParamSpec and Configurable
* Note that uint8_t is considered a character when parsing command line arguments

### DIFF
--- a/Framework/Core/include/Framework/ConfigParamRegistry.h
+++ b/Framework/Core/include/Framework/ConfigParamRegistry.h
@@ -21,6 +21,24 @@
 #include <string>
 #include <cassert>
 
+namespace
+{
+template <typename T>
+constexpr auto isSimpleType()
+{
+  return std::is_same_v<T, int> ||
+         std::is_same_v<T, uint8_t> ||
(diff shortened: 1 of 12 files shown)


ktf:
Ok, but how do you manage the fact that `--foo 1` is an unsigned int rather than an int?

aalkin:
> Ok, but how do you manage the fact that `--foo 1` is an unsigned int rather than an int?

The type is determined by ConfigParamSpec (more precisely with boost option made from it), boost will use iostream to pick up the value with designated type. There may be some funny stuff happening for `--foo -1` for usigned foo, I will add a test...

ktf:
>  There may be some funny stuff happening for `--foo -1` for usigned foo, I will add a test...

Thanks.

aalkin:
I've checked explicitly, it behaves according to c++ rules for implicit type conversions (used by iostream), for example, for `uint16_t` foo, `--foo -1` would set it to `65535`. I do not believe we need an extra test for this.

PR #6579, opened 2021-07-05
Fixing typo on MB support shelf name

mario6829:
The MB support shelfs on side C were wrongly named as the ones on side A (probably a cut&paste error). The name is now correctly fixed (NB: there is no problem at all with their geometry, just the name)


mario6829:
Hi @shahor02 @iouribelikov this is a small PR to fix the name of shelfs located on side C whose name was erroneously the same of the ones on side A. Thanks. Cheers

sawenzel:
trivial typo fix. Merging.

PR #6581, opened 2021-07-05
TRD move tracklethalfchamber header 

bazinski:
Move tracklet half chamber header to front of tracklets.
Fix mc 2 raw to and the parsing.

### DIFF
--- a/Detectors/TRD/reconstruction/src/DigitsParser.cxx
+++ b/Detectors/TRD/reconstruction/src/DigitsParser.cxx
@@ -82,7 +82,7 @@ int DigitsParser::Parse(bool verbose)
     }
     LOG(info) << "digitdata to parse end";
     if (datacopy.size() > 1024) {
-      LOG(fatal) << "some very wrong with digit parsing >1024";
+      LOG(error) << "something likely very wrong with digit parsing >1024";
     }
   }
   int mcmdatacount = 0;
@@ -139,7 +139,7 @@ int DigitsParser::Parse(bool verbose)
       if (mState == StateDigitMCMData || mState == StateDigitEndMarker || mState == StateDigitHCHeader || mState == StateDigitMCMHeader) {
       } else {
 
-        LOG(fatal) << "Digit end marker found but state is not StateDigitMCMData(" << StateDigitMCMData << ") or StateDigit but rather " << mState;
+        LOG(error) << "Digit end marker found but state is not StateDigitMCMData(" << StateDigitMCMData << ") or StateDigit but rather " << mState;
       }
       //only thing that can remain is the padding.
       //now read padding words till end.
@@ -236,9 +236,9 @@ int DigitsParser::Parse(bool verbose)
         }
         // we dont care about the year flag, we are >2007 already.
       } else {
-        if (mState == StateDigitMCMHeader) {
-          LOG(warn) << " state is MCMHeader but we have just bypassed it as the bitmask is wrong :" << std::hex << *word;
-        }
+        //if (mState == StateDigitMCMHeader && *word!=o2::trd::constants::CRUPADDING32) {
+        //  LOG(warn) << " state is MCMHeader but we have just bypassed it as the bitmask is wrong :" << std::hex << *word;
+        //}
         if (*word == o2::trd::constants::CRUPADDING32) {
           if (mVerbose) {
             LOG(info) << "state padding and word : 0x" << std::hex << *word << "  state is:" << mState;
@@ -261,9 +261,9 @@ int DigitsParser::Parse(bool verbose)
               LOG(info) << " digit end marker state ...";
             }
           } else {
-            if (mState != StateDigitMCMData) {
-              LOG(warn) << "something is wrong we are in the statement for MCMdata, but the state is : " << mState << " and MCMData state is:" << StateDigitMCMData;
-            }
+            //if (mState != StateDigitMCMData) {
+            //  LOG(warn) << "something is wrong we are in the statement for MCMdata, but the state is : " << mState << " and MCMData state is:" << StateDigitMCMData;
+            //}
             if (mVerbose || mDataVerbose) {
               //LOG(info) << "mDigitMCMData with state=" << mState << " is at " << mBufferLocation << " had value 0x" << std::hex << *word << " mcmdatacount of : " << mcmdatacount << " adc#" << mcmadccount;
             }
--- a/Detectors/TRD/simulation/src/Trap2CRU.cxx
+++ b/Detectors/TRD/simulation/src/Trap2CRU.cxx
@@ -440,8 +440,6 @@ int Trap2CRU::buildDigitRawData(const int digitstartindex, const int digitendind
   if (digitswritten != digitendindex - digitstartindex) {
     LOG(error) << " something wrong the number of digitswritten does not correspond to the the loop count";
   }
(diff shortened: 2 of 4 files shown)


Detectors/TRD/reconstruction/src/DigitsParser.cxx
```diff
@@ -236,9 +236,9 @@ int DigitsParser::Parse(bool verbose)
         }
         // we dont care about the year flag, we are >2007 already.
       } else {
-        if (mState == StateDigitMCMHeader) {
-          LOG(warn) << " state is MCMHeader but we have just bypassed it as the bitmask is wrong :" << std::hex << *word;
-        }
+        //if (mState == StateDigitMCMHeader && *word!=o2::trd::constants::CRUPADDING32) {
```
tdietel, line 239: Can this be removed, or is this a temporary fix?

Detectors/TRD/reconstruction/src/DigitsParser.cxx
```diff
@@ -261,9 +261,9 @@ int DigitsParser::Parse(bool verbose)
               LOG(info) << " digit end marker state ...";
             }
           } else {
-            if (mState != StateDigitMCMData) {
-              LOG(warn) << "something is wrong we are in the statement for MCMdata, but the state is : " << mState << " and MCMData state is:" << StateDigitMCMData;
-            }
+            //if (mState != StateDigitMCMData) {
```
tdietel, line 264: Same here: can this be removed?

Detectors/TRD/simulation/src/Trap2CRU.cxx
```diff
@@ -668,28 +687,39 @@ void Trap2CRU::convertTrapData(o2::trd::TriggerRecord const& triggerrecord, cons
...
+          firstTrackletOfLink = false;
+        }
         int tracklets = buildTrackletRawData(mCurrentTracklet, linkid); //returns # of 32 bits, header plus trackletdata words that would have come from the mcm.
         mCurrentTracklet += tracklets;
         trackletcounter += tracklets;
         linkwordswritten += tracklets + 1;
         rawwords += tracklets + 1; //1 to include the header
       }
       if (mCurrentTracklet >= mTracklets.size()) {
+        LOG(info) << __func__ << " " << __LINE__;
```
tdietel: This looks like debug output.

Detectors/TRD/simulation/src/Trap2CRU.cxx
```diff
@@ -724,14 +758,17 @@ void Trap2CRU::convertTrapData(o2::trd::TriggerRecord const& triggerrecord, cons
           while (mDigits[mDigitsIndex[mCurrentDigit]].getMCM() == currentMCM &&
                  mDigits[mDigitsIndex[mCurrentDigit]].getROB() == currentROB &&
                  mDigits[mDigitsIndex[mCurrentDigit]].getDetector() == currentDetector) {
+            LOG(info) << __func__ << " " << __LINE__;
             LOG(debug) << " on index of : " << mDigitsIndex[mCurrentDigit] << " wuf channel=" << mDigits[mDigitsIndex[mCurrentDigit]].getChannel();
             mCurrentDigit++;
             digitcounter++;
             adccounter++;
             if (digitcounter > 22) {
+              LOG(info) << __func__ << " " << __LINE__;
```
tdietel: Actually, why is this 22? I would understand 10, 11, 30 or 31...

Detectors/TRD/simulation/src/Trap2CRU.cxx
```diff
@@ -668,28 +687,39 @@ void Trap2CRU::convertTrapData(o2::trd::TriggerRecord const& triggerrecord, cons
...
       //we have some data on this link
+      LOG(info) << __func__ << " " << __LINE__;
       int trackletcounter = 0;
+      bool firstTrackletOfLink = true;
       // if (mCurrentTracklet < endtrackletindex && mCurrentTracklet >= starttrackletindex && mCurrentDigit >= startdigitindex && mCurrentDigit < enddigitindex) {
       //if (mCurrentTracklet < endtrackletindex && mCurrentTracklet>=starttrackletindex mTracklets.size() || mCurrentDigit < mDigits.size()) {
       while (isTrackletOnLink(linkid, mCurrentTracklet) && mCurrentTracklet < endtrackletindex) { //}&& triggerrecord.getNumberOfTracklets()>0) {
+        LOG(info) << __func__ << " " << __LINE__;
         // still on an mcm on this link
+        if (firstTrackletOfLink) {
```
tdietel: This looks like the tracklet HCH is only written if there are tracklets in the event. The hardware will always send it.

tdietel (changes requested):
I think there is a bug: it looks like the HCH is written only if there are tracklets.

Otherwise there still seems to be some debug output left in the code.

Detectors/TRD/reconstruction/src/DigitsParser.cxx
bazinski: temp fix as there seems to be very rare occasions where the digits are created with out the digitmcmheader, in the monte carlo to raw, at least in the 100 event nominal example i am testing against.

Detectors/TRD/reconstruction/src/DigitsParser.cxx
bazinski: same reason as above

Detectors/TRD/simulation/src/Trap2CRU.cxx
bazinski: 21 digits per mcm and digitcounter has just been incremented before so 22.

Detectors/TRD/simulation/src/Trap2CRU.cxx
bazinski: oops removing

Detectors/TRD/simulation/src/Trap2CRU.cxx
tdietel: Ahh, digits as in ADCs. Which is in line with how we defined o2::trd::Digit, so that's ok.

Detectors/TRD/simulation/src/Trap2CRU.cxx
bazinski: 3 lines down from there is that question that you answer above. Fixing.

bazinski: done

bazinski:
@tdietel the digit hc header is only written in those instances where the digits themselves are written, the digit end marker is however written irrespective of digits being present or not. The tracklet halfchamber header and the tracklet end marker is written irrespective of whether there are tracklets in the event.
Additionally I see there is an option for the MCM to write a MCMHeader even when there is no data. This is not taken care of in the mc2raw however the parsing should be able to handle that.
I will go through the assembler again tomorrow as a double check.

tdietel:
The digit half chamber header is always there. The MCM that generates it does not know if the other MCMs have data.
But every normal MCM also sends a header, and there is a setting if this header is written when the MCM does not have any data in its 21 channels.

bazinski:
Ja I came to realise the same. For all of it the parsing is ok, its the generation of the raw data that is at issue. Hence also the ability to test all the combinations of options.

shahor02:
failures unrelated, merging

PR #6584, opened 2021-07-05
Implementation of the ITS Half Barrels

mario6829:
The Half Barrels, an intermediate level of volumes between the Layer and the Stave, has been implemented in the ITS geometry tree. Now each Layer is composed by two Half Barrels, each one containing half of the Staves previously placed directly into the Layer volume. This structure is more realistic with respect to the real world and allow for a better misalignment of all involved elements. Anyway it should be noted that the Staves and all their subvolumes do not change their global position in the world reference system: simply they are now daughters of the Half Barrel volume which in turn is a daughter of the Layer volume.
New methods in the GeometryTGeo class were introduced to handle this new volume level, while where needed the existing methods in the GeometryTGeo and Detector classes were adapted to take into account this new level in the volume tree. The construction of the Layer geometry in the V3Layer class was changed too to actually implement the Half Barrel volumes.
In this implementation only the Staves and their own services are part of the Half Barrels. Work is in progress to include in the new volumes also those parts of the supporting structures which belong to the Half Barrels in the physical ITS.

### DIFF
--- a/Detectors/ITSMFT/ITS/base/include/ITSBase/GeometryTGeo.h
+++ b/Detectors/ITSMFT/ITS/base/include/ITSBase/GeometryTGeo.h
@@ -108,7 +108,7 @@ class GeometryTGeo : public o2::itsmft::GeometryTGeo
   int getNumberOfModules(int lay) const { return mNumberOfModules[lay]; }
   int getNumberOfHalfStaves(int lay) const { return mNumberOfHalfStaves[lay]; }
   int getNumberOfStaves(int lay) const { return mNumberOfStaves[lay]; }
-  int getNumberOfHalfBarrels(int lay) const { return mNumberOfHalfBarrels[lay]; }
+  int getNumberOfHalfBarrels() const { return mNumberOfHalfBarrels; }
   int getNumberOfLayers() const { return mNumberOfLayers; }
   int getChipIndex(int lay, int detInLay) const { return getFirstChipIndex(lay) + detInLay; }
   /// This routine computes the chip index number from the layer, stave, and chip number in stave
@@ -313,7 +313,7 @@ class GeometryTGeo : public o2::itsmft::GeometryTGeo
 
   /// Determines the number of halfbarrels in the layer
   /// \param lay: layer number, starting from 0
-  int extractNumberOfHalfBarrels(int lay) const;
+  int extractNumberOfHalfBarrels() const;
 
   /// Determines the number of layers in the Geometry
   /// \param lay: layer number, starting from 0
@@ -351,7 +351,7 @@ class GeometryTGeo : public o2::itsmft::GeometryTGeo
   static constexpr int MAXLAYERS = 15; ///< max number of active layers
 
   Int_t mNumberOfLayers;                        ///< number of layers
-  std::vector<int> mNumberOfHalfBarrels;        ///< the number of halfbarrels/layer(layer)
+  Int_t mNumberOfHalfBarrels;                   ///< number of halfbarrels
   std::vector<int> mNumberOfStaves;             ///< number of staves/layer(layer)
   std::vector<int> mNumberOfHalfStaves;         ///< the number of substaves/stave(layer)
   std::vector<int> mNumberOfModules;            ///< number of modules/substave(layer)
--- a/Detectors/ITSMFT/ITS/base/include/ITSBase/GeometryTGeo.h
+++ b/Detectors/ITSMFT/ITS/base/include/ITSBase/GeometryTGeo.h
@@ -108,7 +108,7 @@ class GeometryTGeo : public o2::itsmft::GeometryTGeo
   int getNumberOfModules(int lay) const { return mNumberOfModules[lay]; }
   int getNumberOfHalfStaves(int lay) const { return mNumberOfHalfStaves[lay]; }
   int getNumberOfStaves(int lay) const { return mNumberOfStaves[lay]; }
-  int getNumberOfHalfBarrels() const { return mNumberOfHalfBarrels; }
+  int getNumberOfHalfBarrels(int lay) const { return mNumberOfHalfBarrels[lay]; }
   int getNumberOfLayers() const { return mNumberOfLayers; }
   int getChipIndex(int lay, int detInLay) const { return getFirstChipIndex(lay) + detInLay; }
   /// This routine computes the chip index number from the layer, stave, and chip number in stave
@@ -311,6 +311,10 @@ class GeometryTGeo : public o2::itsmft::GeometryTGeo
   /// \param lay: layer number from 0
   int extractNumberOfChipsPerModule(int lay, int& nrow) const;
 
+  /// Determines the number of halfbarrels in the layer
+  /// \param lay: layer number, starting from 0
+  int extractNumberOfHalfBarrels(int lay) const;
+
   /// Determines the number of layers in the Geometry
   /// \param lay: layer number, starting from 0
   int extractNumberOfStaves(int lay) const;
@@ -347,7 +351,7 @@ class GeometryTGeo : public o2::itsmft::GeometryTGeo
   static constexpr int MAXLAYERS = 15; ///< max number of active layers
 
   Int_t mNumberOfLayers;                        ///< number of layers
-  Int_t mNumberOfHalfBarrels;                   ///< the number of halfbarrels
+  std::vector<int> mNumberOfHalfBarrels;        ///< the number of halfbarrels/layer(layer)
   std::vector<int> mNumberOfStaves;             ///< number of staves/layer(layer)
   std::vector<int> mNumberOfHalfStaves;         ///< the number of substaves/stave(layer)
   std::vector<int> mNumberOfModules;            ///< number of modules/substave(layer)
--- a/Detectors/ITSMFT/ITS/base/include/ITSBase/GeometryTGeo.h
+++ b/Detectors/ITSMFT/ITS/base/include/ITSBase/GeometryTGeo.h
@@ -113,26 +113,29 @@ class GeometryTGeo : public o2::itsmft::GeometryTGeo
   int getChipIndex(int lay, int detInLay) const { return getFirstChipIndex(lay) + detInLay; }
   /// This routine computes the chip index number from the layer, stave, and chip number in stave
   /// \param int lay The layer number. Starting from 0.
+  /// \param int hba The halfbarrel number. Starting from 0
   /// \param int sta The stave number. Starting from 0
   /// \param int chipInStave The chip number in the stave. Starting from 0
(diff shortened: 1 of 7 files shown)


mario6829:
Hi @shahor02 @iouribelikov here it is the PR of the Half Barrels. Let me know in case of problems or if you need more info. Kind regards

shahor02:
Hi @davidrohr @TimoWilken 

The fullCI is amazingly green, but the log shows quite some warnings and errors:
````
Error while processing /O2/GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx.
/sw/SOURCES/O2/6584-slc8_x86-64/0/GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx:37:10: error: 'omp.h' file not found [clang-diagnostic-error]
#include <omp.h>
....
1 error generated.
Error while processing /sw/BUILD/de9f0d7f4b545fb32732038aff81f6daba397c25/O2/GPU/GPUTracking/Base/cuda/GPUReconstructionCUDArtc.src.S.
/sw/BUILD/de9f0d7f4b545fb32732038aff81f6daba397c25/O2/GPU/GPUTracking/Base/cuda/GPUReconstructionCUDArtc.src.S:1:2: error: expected identifier or '(' [clang-diagnostic-error]
        .global _curtc_GPUReconstructionCUDArtc_cu_src
        ^
Found compiler error(s).
....
End of search list.
3039 warnings and 9 errors generated when compiling for host.
Error while processing /O2/Detectors/ITSMFT/ITS/tracking/cuda/src/ClusterLinesGPU.cu.
error: cannot find libdevice for sm_20. Provide path to different CUDA installation via --cuda-path, or pass -nocudalib to build without linking with libdevice. [clang-diagnostic-error]
error: unknown argument: '-fPIC -O2 ' [clang-diagnostic-error]
error: unknown argument: '-forward-unknown-to-host-compiler' [clang-diagnostic-error]
error: unsupported option '--allow-unsupported-compiler' [clang-diagnostic-error]
error: unsupported option '--expt-relaxed-constexpr' [clang-diagnostic-error]
error: unsupported option '--extended-lambda' [clang-diagnostic-error]
error: unsupported option '--ftz=true' [clang-diagnostic-error]
error: unsupported option '--generate-code=arch=compute_86,code=[compute_86,sm_86]' [clang-diagnostic-error]
/sw/SOURCES/O2/6584-slc8_x86-64/0/GPU/Common/GPUCommonRtypes.h:32:12: error: 'Rtypes.h' file not found [clang-diagnostic-error]
  #include "Rtypes.h"
````
The FST itself finishes w/o errors, but I am not sure if I should trust it given all these errrors.

**Review by shahor02 — COMMENTED** [review]
Hi @mario6829 

Thanks a lot! Looks fine but I have 1 doubt: naively I thought the half-barrel is lower level alignable than the layer (i.e. if the half-barrel contains all half-layers). But I see that in your implementation each layer is made of half-barrels, and those of different layers are independent. 
Is this really supposed to be so or this is just an "effective" implementation?

Then, I am not sure a layer as alignable volume makes sense. Aren't top and bottom halves independent? If so, then I would suppress the https://github.com/AliceO2Group/AliceO2/blob/18b078c3c91235496e311a763e1100f29b99a4e3/Detectors/ITSMFT/ITS/simulation/src/Detector.cxx#L1049-L1051

Finally, I saw than during tests you've uploaded an ideal new alignment object to http://ccdb-test.cern.ch:8080/browse/ITS/Align. Since this creates a lot of error messages with current dev, I've masked it with empty object. If you play again with test ocdb, please clean it (before this PR is merged) by
````
root
.L $O2_ROOT/share/macro/UploadDummyAlignment.C+
UploadDummyAlignment("http://ccdb-test.cern.ch:8080",0,-1, o2::detectors::DetID::getMask("ITS"));
````

**Comment by mario6829** [comment]
Many thanks @shahor02 for reviewing the code!
For what concerns the layers: I implemented this way because I thought (maybe wrongly) that each single half barrel could be misaligned with respect to the others (in principle they are solidly linked to the support structures, but there may be a slight difference wrt the ideal position - this may be more relevant for MB/OB than IB). Indeed the two halves are independent, but I imagined that the layer itself could be misaligned as a whole (something like the whole ITS volume). In short: I preferred to leave open all possible degrees of freedom. But if you think this freedom is redundant, I'll remove the three lines you mention. In fact, I can even remove the layer as a volume if you think it is now useless (not in this PR, of course, I have to do it carefully - we can discuss in a WG meeting).
For the  alignment object in the CCDB: I really apologize! I didn't do it on purpose: I was trying to use a local file instead of the one on the db, I don't really know how I could have uploaded it... sorry for that.
Thanks again. Cheers

**Comment by davidrohr** [comment]
> Hi @davidrohr @TimoWilken
> 
> The fullCI is amazingly green, but the log shows quite some warnings and errors:
> 
> The FST itself finishes w/o errors, but I am not sure if I should trust it given all these errrors.

@shahor02 : That is OK, the warnings seem to be during the codechecker. The codechecker always has a lot of false warnings / errors during compilation. Important is the
```
```
at the end.

**Comment by iouribelikov** [comment]
> For what concerns the layers: I implemented this way because I thought (maybe wrongly) that each single half barrel could be misaligned with respect to the others (in principle they are solidly linked to the support structures, but there may be a slight difference wrt the ideal position - this may be more relevant for MB/OB than IB). Indeed the two halves are independent, but I imagined that the layer itself could be misaligned as a whole (something like the whole ITS volume). In short: I preferred to leave open all possible degrees of freedom. But if you think this freedom is redundant, I'll remove the three lines you mention. In fact, I can even remove the layer as a volume if you think it is now useless (not in this PR, of course, I have to do it carefully - we can discuss in a WG meeting).

@mario6829 @shahor02 As far as I understand, 
a) the ITS consists (mechanically) of 4 half-barrels:  two IB half-barrels and two OB half-barrels.
b) Within each of these 4 half-barrels, the corresponding half-layers could potentially be mis-aligned.
c) Then, the whole IB is mechanically independent of the  OB.  So the whole IB could be mis-aligned wrt the whole OB.
For me, this means that any of the layers (taken as whole) is not really "mis-alignable" wrt other layers.

**Comment by shahor02** [comment]
@mario6829 @iouribelikov OK, I would propose to suppress the whole layers as alignable DOF, because in this case I cannot apply a constraint on the movements of the half-layers of the same half-barrel (i.e. effectively consider e.g. MB half-barrel as an alignable).

**Comment by mario6829** [comment]
Ok @shahor02 I removed the alignement of Layer as you suggested. Then we can discuss how to properly re-arrange the whole volume tree. Thanks again. Cheers

File: Detectors/ITSMFT/ITS/simulation/src/Detector.cxx [hunk 1]
```diff
@@ -1037,38 +1037,64 @@ void Detector::addAlignableVolumesLayer(int lr, TString& parent, Int_t& lastUID)
   // Add alignable volumes for a Layer and its daughters
   //
   // Created:      06 Mar 2018  Mario Sitta First version (mainly ported from AliRoot)
+  // Updated:      06 Jul 2021  Mario Sitta Do not set Layer as alignable volume
   //
 
   TString wrpV =
     mWrapperLayerId[lr] != -1 ? Form("%s%d_1", GeometryTGeo::getITSWrapVolPattern(), mWrapperLayerId[lr]) : "";
   TString path = Form("%s/%s/%s%d_1", parent.Data(), wrpV.Data(), GeometryTGeo::getITSLayerPattern(), lr);
   TString sname = GeometryTGeo::composeSymNameLayer(lr);
```
shahor02, line 1047: This line should not be needed anymore

shahor02 (commented):
Thanks @mario6829 , should be fine now, with the half-layer being next to the full ITS coarse DOF I should be able to group them in global alignment (i.e. to constrain half-layers to move together).

Detectors/ITSMFT/ITS/base/include/ITSBase/GeometryTGeo.h
```diff
@@ -103,10 +103,12 @@ class GeometryTGeo : public o2::itsmft::GeometryTGeo
   int getNumberOfChipsPerModule(int lay) const { return mNumberOfChipsPerModule[lay]; }
   int getNumberOfChipsPerHalfStave(int lay) const { return mNumberOfChipsPerHalfStave[lay]; }
   int getNumberOfChipsPerStave(int lay) const { return mNumberOfChipsPerStave[lay]; }
+  int getNumberOfChipsPerHalfBarrel(int lay) const { return mNumberOfChipsPerHalfBarrel[lay]; }
   int getNumberOfChipsPerLayer(int lay) const { return mNumberOfChipsPerLayer[lay]; }
   int getNumberOfModules(int lay) const { return mNumberOfModules[lay]; }
   int getNumberOfHalfStaves(int lay) const { return mNumberOfHalfStaves[lay]; }
   int getNumberOfStaves(int lay) const { return mNumberOfStaves[lay]; }
+  int getNumberOfHalfBarrels(int lay) const { return mNumberOfHalfBarrels[lay]; }
```
iouribelikov: @mario6829 Do we need this function and the corresponding std::vector ?  The number of half-barrels is always 2, is not it ?

Detectors/ITSMFT/ITS/simulation/src/Detector.cxx
mario6829: Hi @shahor02 it was just a reminder that the layer was alignable and now it isn't anymore. But ok, I can remove it. Do you agree to do it in the next PR ? Otherwise I trigger again all the tests... :)
Many thanks. Cheers

Detectors/ITSMFT/ITS/base/include/ITSBase/GeometryTGeo.h
mario6829: Hi @iouribelikov I just wrote it like the other ones (pretty like the number of layers, which is actually fixed too but it has a member and a getter), but yes, I can change it into a constant (or in a single-valued member instead of a vector).  @shahor02 do you agree ?

shahor02:
Hi, I was not merging this PR since I understood @iouribelikov wants to discuss it at tomorrow meeting.
Fine for removing the line later

Detectors/ITSMFT/ITS/base/include/ITSBase/GeometryTGeo.h
shahor02: Sure

mario6829: Done: I changed it to a single-valued member, the same for all layers.

iouribelikov (approved):
@mario6829  For me, everything works fine now.  Thank you very much, Mario !

shahor02 (approved):

PR #6585, opened 2021-07-05
TRD tracking improvements

martenole:
- the TRD tracks in O2 are cleaned up: members which are transient and needed only during the tracking are moved to a helper struct and the TRD tracks thus become more lightweight
- a filter is added to the tracklet transformer which optionally masks TRD triggers for which no ITS data is available. This will speedup the synchronous reco for which the matching to the TRD will be done with ITS-TPC matched tracks only and thus we don't need to transform the tracklets which anyhow won't be matched


Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -144,13 +144,17 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
   // load ITS-TPC matched tracks
   for (int iTrk = 0; iTrk < mChainTracking->mIOPtrs.nTracksTPCITSO2; ++iTrk) {
     const auto& trkITSTPC = mChainTracking->mIOPtrs.tracksTPCITSO2[iTrk];
-    GPUTRDTrack trkLoad(trkITSTPC, mTPCVdrift);
+    GPUTRDTracker::HelperTrackAttributes trkAttribs;
+    trkAttribs.mTime = trkITSTPC.getTimeMUS().getTimeStamp();
+    trkAttribs.mTimeAddMax = trkITSTPC.getTimeMUS().getTimeStampError();
+    trkAttribs.mTimeSubMax = trkITSTPC.getTimeMUS().getTimeStampError();
```
shahor02: Note that for the ITS-TPC tracks the `getTimeStampError()` has a meaning of gaussian sigma, while for the TPC tracks `getDeltaTFwd(), getDeltaTBwd()` is uniform span. You should probably multiply (unless you do this later and I missed this) `getTimeStampError()` by n-sigma (in similar situations I take ~4 sigma, but this should be actually a settable parameter).

Detectors/TRD/workflow/src/TRDTrackingWorkflow.cxx
```diff
@@ -43,11 +44,13 @@ framework::WorkflowSpec getTRDTrackingWorkflow(bool disableRootInp, bool disable
       specs.emplace_back(o2::tpc::getTPCTrackReaderSpec(useMC));
     }
     specs.emplace_back(o2::trd::getTRDTrackletReaderSpec(useMC, false));
+    if (trigRecFilterActive) {
+      specs.emplace_back(o2::its::getIRFrameReaderSpec());
+    }
```
shahor02: why not using InputHelper?

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
```diff
@@ -27,45 +28,108 @@ namespace trd
...
 
-  for (const auto& tracklet : tracklets) {
-    calibratedTracklets.push_back(mTransformer.transformTracklet(tracklet));
+  std::vector<char> trigRecBitfield(trigRecs.size()); // flag TRD IR with ITS data (std::vector<bool> platform dependend)
+  int nTrackletsTransformed = 0;
+
+  if (mTrigRecFilterActive) {
+    const auto irFrames = inputData.getIRFramesITS();
+    for (const auto& irFrame : irFrames) {
+      int lastMatchedIdx = 0; // ITS IR are sorted in time and do not overlap
```
shahor02: should not this line be above the `for (const auto& irFrame : irFrames) {` in order to profit from the ordering?

GPU/GPUTracking/TRDTracking/GPUTRDInterfaces.h
```diff
@@ -255,13 +250,15 @@ class trackInterface<GPUTPCGMTrackParam> : public GPUTPCGMTrackParam
     for (int i = 0; i < 15; i++) {
       SetCov(i, param.getParamOut().getCov()[i]);
     }
+    /*
     mTime = param.getTimeMUS().getTimeStamp();
     mTimeAddMax = param.getTimeMUS().getTimeStampError();
     mTimeSubMax = param.getTimeMUS().getTimeStampError();
     float tmp = param.getTimeMUS().getTimeStampError() * 2.58f; // TPCvDrift = 2.58 cm/us fixed for now, should come from CCDB
     Cov()[2] += tmp * tmp;                                      // account for time uncertainty by increasing sigmaZ2
+    */
```
shahor02: can the obsolete code be removed, also below?

Detectors/TRD/workflow/src/TRDTrackingWorkflow.cxx
martenole: I tried that yesterday, but it is not straightforward for the TRD. The problem is that we have two tracklet types: Our tracklets which come from the FEE `Tracklet64` and the space points which are calculated from those tracklets with calibration applied called `CalibratedTracklet`. These come from the tracklet transformer.
The InputHelper tries to load everything for the TRD if TRD clusters are asked for. Basically I need another argument for ``InputHelper::addInputSpecs()` to distinguish between the two cases (if the calibrated tracklets should be loaded as well or not). Or I separate the tracklet transformer workflow from the tracking workflow. Then I can use the InputHelper for the tracking, but not for the tracklet transformer.

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
martenole: Absolutely, thanks for spotting!

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
```diff
@@ -27,45 +28,108 @@ namespace trd
...
+  std::shared_ptr<DataRequest> dataRequest = std::make_shared<DataRequest>();
+  if (trigRecFilterActive) {
+    dataRequest->requestIRFramesITS();
+  }
+  auto& inputs = dataRequest->inputs;
+  inputs.emplace_back("trdtracklets", "TRD", "TRACKLETS", 0, Lifetime::Timeframe);
+  inputs.emplace_back("trdtriggerrec", "TRD", "TRKTRGRD", 0, Lifetime::Timeframe);
+
+  std::vector<OutputSpec> outputs;
+  outputs.emplace_back("TRD", "CTRACKLETS", 0, Lifetime::Timeframe);
```
davidrohr, line 125: I would propose to set the subspec to 1 in case we apply the filtering, and then the tracker would request subspec 1 in case the filtering is active. That would ensure consistency.

martenole: There is also the `TRIGRECMASK` in the line below. This is always filled by the tracklet transformer and taken into account by the tracking. This is a bitmap flagging the TRD triggers which are ignored/used. I think that avoids inconsistencys.
On the other hand I want to warn the user in case the filtering is active, but matching to TPC-only tracks is requested (https://github.com/AliceO2Group/AliceO2/pull/6585/files#diff-70a51291b8fcbd8a6f7ee3ffcfb34e8295217d05bedb09a2d16a15b869c9a0e1R264). For that the tracker needs to know whether the filtering was active or not. So I think I will do as you suggest and force the user to set the option both for the tracklet transformer and for the tracking and use the subspec.

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
martenole: @davidrohr I would like to add this option to the TRD reco options in GPUSettingsList.h via:
`AddOptionRTC(nSigmaTerrITSTPC, float, 4.f, "", 0, "Number of sigmas for ITS-TPC track time error estimate")`
Then I should be able to set the parameter from command line via `--configKeyValues "GPU_rec_trd.nSigmaTerrITSTPC=4;"` right?
And how could I access the option in TRDGlobalTrackingSpec? Should I just create a GPUSettingsRec object and take the option from there?

davidrohr: The settings should be fetched automatically during
```
mRec->SetSettings(o2::base::Propagator::Instance()->getNominalBz(), &cfgRecoStep);
```
So they will be available for the TRDTracker.
In case you need to access them in the spec directly, you can use mRec->GetProcessingSettings()

Detectors/TRD/workflow/src/TRDTrackingWorkflow.cxx
shahor02: Not sure I understand the problem: you are anyway reading the Tracklet64, which could be done via InputHelper, and then you can couple this reading to transormer, like it is done for the TPC shared cluster creator: https://github.com/AliceO2Group/AliceO2/blob/ae21d9f42c0c2469d6e5595af955f6ba1dd4c15c/Detectors/GlobalTrackingWorkflow/helpers/src/InputHelper.cxx#L67
So, *if* you read the tracklets from the file, the InputHelper would also provide the CalibratedTracklet.
What is more tricky is having the "strict" and "relaxed" workflows in the same piped workflow, receiving the Tracklet64 from the upstream (e.g. from CTF-reader or raw-reader/DD):  you should either add to getTRDTrackingWorkflow a flag saying that the transformer should be attached (true in "strict" and false in "relaxed" instances) or to attach the transformer to the raw->tracklets and entropy decoders.

I guess when the Tracklet64 input is needed, in general, also the CalibratedTracklet will be needed? To me this is similar to TPC shared clusters container, 

````
{strict phase} [ read tracklets, ]
````

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
shahor02: In principle, the matching to TPC-only should be reserved to async mode, where there is no event filtering anyway.
But I did not understand how using subspec = 1 improves consistency (what problem does this solve?)

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
martenole: Thanks, I have added the option now

Detectors/TRD/workflow/src/TRDTrackingWorkflow.cxx
martenole: The tracklet transformer is always needed (for the TRD tracking, for the TRD calibration and also for the TPC interpolation for the refit). The InputHelper always loads both the Tracklet64 and the CalibratedTracklets. Since the latter are not available before the tracklet transformer has run I don't use the InputHelper for the tracklet transformer. But now I have decoupled the tracklet transformer and the global tracking so that I can use the InputHelper for the global tracking at least.
So TRD tracking for the sync. reco can now be run with:
`o2-trd-tracklet-transformer -b --filter-trigrec | o2-trd-global-tracking -b --tracking-sources ITS-TPC --run --configKeyValues "GPU_rec_trd.nSigmaTerrITSTPC=4;"`

martenole:
I don't understand the pragma-once-checker. Is it new? Looks to me like it fails due to
Analysis/DataModel/include/AnalysisDataModel/EMCALClusters.h
Analysis/DataModel/include/AnalysisDataModel/Jet.h
version/O2Version.h
So not related to this PR

Detectors/TRD/workflow/src/TRDTrackingWorkflow.cxx
shahor02: I had some junk copy/pasted in prev. message. OK, this will work also, but I thought you can calculate the calibTracklets on the fly, reading only Tracklet64 from the file.

martenole: Yes, I can also calculate calibTracklets on the fly. This is what happens with the above command. Read Tracklet64, calculate calibTracklets and pass them to the TRD tracker. 
To handle everything with the InputHelper would mean I need to distinguish in `DataRequest::requestTRDTracklets()` whether the calibTracklets are already there or not. And this I did not manage without breaking the logic in the RecoContainer.

martenole:
@sawenzel do you know why the pragma-once checker is failing here? I have rebased onto dev and I don't find any remaining pragma once statements in the repository now with the command `git grep -l '#pragma once' -- '*.h'`

martenole:
@davidrohr This is ready now _if_ the fullCI is happy, let's see

PR #6587, opened 2021-07-05
[MRRTF-122] Add MCH to FST

aphecetche:
Note that the last tracking part (extrap to vertex) is far from final
and will be updated in a forthcoming PR (in prep) where FwdTracks
will be added to AODs and extrapolation will be part of the AOD
producer.
Currently the vertex information that is used is completely fake (coming
from a vertex sampler that is used for debugging when using Run2 data).


shahor02:
Hi @aphecetche 
I see that your reco-workflow has no option to write to the root file (which we disable by default in raw production workflows) and I don't see separate writer workflow either. Will you extend the reco-workflow?

aphecetche:
@shahor02 the track writer is actually in another PR #6575 and might be integrated later on in the reco-workflow if need be, yes.

shahor02 (approved):

PR #6592, opened 2021-07-06
A simple github action to check pragma once rule

sawenzel:
We disallow `#pragma once` and check it with this github action.

### DIFF
--- a/.github/workflows/pragmaonce.yml
+++ b/.github/workflows/pragmaonce.yml
@@ -11,20 +11,8 @@ jobs:
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         persist-credentials: false
-        # We need the history of the dev branch all the way back to where the PR
-        # diverged. We're fetching everything here, as we don't know how many
-        # commits back that point is.
-        fetch-depth: 0
 
-    - name: Install prerequisites
-      env:
-        DEBIAN_FRONTEND: noninteractive
-      run: |
-        sudo apt update -y
-        sudo apt install -y git
     - name: Run pragma check
       id: pragma_check
-      env:
-        ALIBUILD_GITHUB_TOKEN: ${{secrets.ALIBUILD_GITHUB_TOKEN}}
       run: |
-        git grep "#pragma once" | grep "\.h"; if [ $? -eq "0" ]; then exit 1; fi; exit 0
+        git grep -l '#pragma once' -- '*.h' && exit 1
--- a/.github/workflows/pragmaonce.yml
+++ b/.github/workflows/pragmaonce.yml
@@ -1,0 +1,30 @@
+name: pragma-once checker
+
+on: [pull_request_target]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        persist-credentials: false
+        # We need the history of the dev branch all the way back to where the PR
+        # diverged. We're fetching everything here, as we don't know how many
+        # commits back that point is.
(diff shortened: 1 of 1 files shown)


sawenzel:
@TimoWilken : Could you please take a look if this looks reasonable? Eventually, we might make a combined action, encapsulating multiple of these simple checks.

.github/workflows/pragmaonce.yml
```diff
@@ -0,0 +1,30 @@
...
+        DEBIAN_FRONTEND: noninteractive
+      run: |
+        sudo apt update -y
+        sudo apt install -y git
+    - name: Run pragma check
+      id: pragma_check
+      env:
+        ALIBUILD_GITHUB_TOKEN: ${{secrets.ALIBUILD_GITHUB_TOKEN}}
+      run: |
+        git grep "#pragma once" | grep "\.h"; if [ $? -eq "0" ]; then exit 1; fi; exit 0
```
TimoWilken: I think this line could just be
```bash
git grep -l '#pragma once' -- '*.h' && exit 1
```

.github/workflows/pragmaonce.yml
```diff
@@ -0,0 +1,30 @@
...
+        # diverged. We're fetching everything here, as we don't know how many
+        # commits back that point is.
+        fetch-depth: 0
+
+    - name: Install prerequisites
+      env:
+        DEBIAN_FRONTEND: noninteractive
+      run: |
+        sudo apt update -y
+        sudo apt install -y git
```
TimoWilken: According to [the docs](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#tools), git is already installed in the container, so we don't need this step at all -- removing it should speed up the check.

.github/workflows/pragmaonce.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        persist-credentials: false
+        # We need the history of the dev branch all the way back to where the PR
+        # diverged. We're fetching everything here, as we don't know how many
+        # commits back that point is.
+        fetch-depth: 0
```
TimoWilken: We're only checking the latest commit, so we don't need to fetch everything. You should be able to remove the `fetch-depth` option for a speedup.

.github/workflows/pragmaonce.yml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Install prerequisites
+      env:
+        DEBIAN_FRONTEND: noninteractive
+      run: |
+        sudo apt update -y
+        sudo apt install -y git
+    - name: Run pragma check
+      id: pragma_check
+      env:
+        ALIBUILD_GITHUB_TOKEN: ${{secrets.ALIBUILD_GITHUB_TOKEN}}
```
TimoWilken: Unneeded, right?

TimoWilken (changes requested):
Looks good, but I think you can remove a bunch of stuff to make this faster.

TimoWilken (approved):

TimoWilken:
Thanks, merged.

vkucera:
@sawenzel @TimoWilken It seem that this action is reporting unrelated files.
See:
* https://github.com/AliceO2Group/AliceO2/pull/6615/checks?check_run_id=3024121465
* https://github.com/AliceO2Group/AliceO2/pull/6619/checks?check_run_id=3028476422

Also, there is no information in the log about what to do to fix the errors.

sawenzel:
@vkucera : This happens when the PR is not rebased on dev (the developer has an old state). Feel free to improve the action.

PR #6596, opened 2021-07-06
Add more information to TrackExtra table

nburmaso:


Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
```diff
@@ -40,6 +41,7 @@
 using namespace o2::framework;
 using GID = o2::dataformats::GlobalTrackID;
 using DataRequest = o2::globaltracking::DataRequest;
+using TPCClRefElem = uint32_t;
```
shahor02: it is already defined in the `TrackTPC.h`, better to no redefine it.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -671,7 +708,6 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
       auto& trackIndex = primVerGIs[ti];
```
shahor02: I think it is better to use `TrackExtraInfo extraInfoHolder;` local to the loop, then you don't need all these reinitializations

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -671,7 +708,6 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
       if (src == GIndex::Source::ITS && mFillTracksITS) {
```
shahor02: The types of different track will be growing, isn't it better to use a "track-sources" option like
https://github.com/AliceO2Group/AliceO2/blob/3d7bf78847055e5e16f41b8ddb88cc60cdd36a71/Detectors/GlobalTrackingWorkflow/src/primary-vertexing-workflow.cxx#L61
https://github.com/AliceO2Group/AliceO2/blob/3d7bf78847055e5e16f41b8ddb88cc60cdd36a71/Detectors/GlobalTrackingWorkflow/src/primary-vertexing-workflow.cxx#L74 ?
Then you can simply check
`if (GIndex::includesSource(src, sourcesMask)) {`

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -680,11 +716,11 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
       }
       if (src == GIndex::Source::TPC && mFillTracksTPC) {
```
shahor02: and then, instead of having separate block (with mostly repeating code) for every type of barrel track, you can have a single block
````
auto contributorsGID = recoData.getSingleDetectorRefs(src);
if (contributorsGID[GIndex::Source::ITS].isIndexSet()) {
 // process ITS part from contributorsGID[GIndex::ITS]
}
if (contributorsGID[GIndex::Source::TPC].isIndexSet()) {
 // process TPC part from contributorsGID[GIndex::TPC]
}
// etc.

````

File: Detectors/AOD/src/AODProducerWorkflowSpec.cxx [hunk 5]
```diff
@@ -718,23 +753,57 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+          extraInfoHolder.tpcInnerParam = tpcOrig.getP();
+          extraInfoHolder.tpcChi2NCl = tpcOrig.getNClusters() ? tpcOrig.getChi2() / tpcOrig.getNClusters() : 0;
+          extraInfoHolder.tpcSignal = tpcOrig.getdEdx().dEdxTotTPC;
+          extraInfoHolder.tpcNClsShared = countTPCSharedCl(tpcOrig, tpcClusRefs, tpcClusShMap, tpcClusAcc);
+        }
+        addToTracksTable(tracksCursor, tracksCovCursor, track, -1, src);
+        addToTracksExtraTable(tracksExtraCursor, extraInfoHolder);
+      }
+      if (src == GIndex::Source::TPCTRD && mFillTracksTPC) {
+        const auto& track = tracksTPCTRD[trackIndex.getIndex()];
```
shahor02: I don't see where do you save TPC Nclusters? BTW, with the current TPC tracking one can have >1 cluster on a given padrow. For the calculation of the tpcChi2NCl it is indeed better to use `tpcOrig.getNClusters()`, but as Nclusters stored in the AOD I would use actual number of contributing padrows, like it is calculated in the snippet I sent you before.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -718,23 +753,57 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
         extraInfoHolder.length = tofInt.getL();
```
shahor02: we discussed during the Monday meeting the way TOF info is stored, at the moment please store the `extraInfoHolder.tofExpMom` as in run2 AOD conversion:
https://github.com/alisw/AliPhysics/blob/820236aa6f0149a27d5204b64bb52620772df90f/RUN3/AliAnalysisTaskAO2Dconverter.cxx#L1657-L1665

Corresponding methods were implemented in https://github.com/AliceO2Group/AliceO2/blob/3d7bf78847055e5e16f41b8ddb88cc60cdd36a71/Analysis/DataModel/include/AnalysisDataModel/PID/PIDTOF.h#L58-L61 but I am not sure about the template parameters, hope @njacazio can clarify what should be used

shahor02 (commented):
Hi @nburmaso 
Please see comments below. BTW, on Monday we also discussed the s.vertices storage, we may eventually store not only prong indices but also corresponding `track.getX()` params. But at the moment, could you simply add the indices only, just to have the V0s stored?

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
nburmaso: Ok, I will clean it up

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Yes, it is a better way indeed

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Thank you for this advice, I didn't think about it :)

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: @shahor02, I thought we agreed on filling number of shared clusters for now, but maybe I misunderstood. Where should I put `found` and `crossed` for now then? The corresponding columns are dynamic https://github.com/AliceO2Group/AliceO2/blob/dc091dba2e910273d38d0a4dac48963b427f2adf/Framework/Core/include/Framework/AnalysisDataModel.h#L218-L221, so they can't be used in a file.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Thank you for the snippet. I will try adding this column

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: What I wrote before is that `tpcNClsFindable` is not available in run3 reco output with the same definition as in run2, so better to use a placeholder at the moment. For instance, you can use instead tpcNClsCrossedRows in order to define the differences currently present in the AOD.

nburmaso:
Hi @shahor02, @sawenzel. I have tried addressing the comments and updated the code. I have moved track table filling into a separate subroutine and also removed track flags, such as `--fill-tracks-its`, so information sources can now be passed via `--info-sources` flag as a comma-separated list. I still have some doubts about TPC clusters. Is this implementation https://github.com/AliceO2Group/AliceO2/blob/2a9f3f997dabec554c81ee3b0bb066f5e9758ef2/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L236-L240
is good for now as a placeholder?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -194,12 +196,86 @@ void AODProducerWorkflowDPL::addToMFTTracksTable(mftTracksCursorType& mftTracksC
...
+          if (contributorsGID[GIndex::Source::TPC].isIndexSet()) {
+            const auto& tpcOrig = data.getTPCTrack(contributorsGID[GIndex::TPC]);
+            extraInfoHolder.tpcInnerParam = tpcOrig.getP();
+            extraInfoHolder.tpcChi2NCl = tpcOrig.getNClusters() ? tpcOrig.getChi2() / tpcOrig.getNClusters() : 0;
+            extraInfoHolder.tpcSignal = tpcOrig.getdEdx().dEdxTotTPC;
+            uint8_t shared, found, crossed; // fixme: need to switch from these placeholders to something more reasonable
+            countTPCClusters(tpcOrig, tpcClusRefs, tpcClusShMap, tpcClusAcc, shared, found, crossed);
+            extraInfoHolder.tpcNClsFindable = tpcOrig.getNClusters();
+            extraInfoHolder.tpcNClsFindableMinusCrossedRows = tpcOrig.getNClusters() - crossed;
+            extraInfoHolder.tpcNClsShared = shared;
```
shahor02: There are 4 variables to fill, I would use:
````
extraInfoHolder.tpcNClsFindable = tpcOrig.getNClusters();
extraInfoHolder.tpcNClsFindableMinusFound = tpcOrig.getNClusters() - found;
extraInfoHolder.tpcNClsFindableMinusCrossedRows = tpcOrig.getNClusters() - crossed;
extraInfoHolder.tpcNClsShared = shared;
````
bearing in mind that such a definition of `tpcNClsFindable` is wrong and in some cases the `tpcOrig.getNClusters() - crossed` can be <0. Later we should find a proper substitute for the tpcNClsFindable.
As I wrote before, the `tpcOrig.getNClusters()` itself is not a good variable, since in the new reconstruction if the fitter considers that the 2 clusters on the same padrow may come in fact from the same split cluster, it will count (and fit) both of them, thus the `getNClusters()` may actually exceed the number of crossed rows.

**Comment by sawenzel** [comment]
Took away the "draft" label to trigger CI. So with this PR we have most of TPC and TOF data filled?

File: Detectors/AOD/src/AODProducerWorkflowSpec.cxx (hunk 7 above)
Comment (nburmaso) [reply Detectors/AOD/src/AODProducerWorkflowSpec.cxx]: Done, thank you

**Comment by nburmaso** [comment]
@sawenzel: Yes, it seems that we have most of the columns covered with this PR

File: Detectors/AOD/src/AODProducerWorkflowSpec.cxx [hunk 8]
```diff
@@ -194,12 +196,86 @@ void AODProducerWorkflowDPL::addToMFTTracksTable(mftTracksCursorType& mftTracksC
...
+            float mom = trackPar.getP();
+            float mass = o2::constants::physics::MassPionCharged; // default pid = pion
+            float expSig = mom > 0 ? std::sqrt(mom * mom + mass * mass) : 0.f;
+            extraInfoHolder.tofSignal = expSig;
+            float expMom = 0.f;
+            if (expSig > 0) {
+              float expBeta = (intLen / expSig / cSpeed);
+              expMom = mass * expBeta * cSpeed / std::sqrt(1.f - (expBeta * expBeta));
+            }
+            extraInfoHolder.tofExpMom = expMom;
```
shahor02: This does not look correct:
1) the `extraInfoHolder.tofSignal` should be the the measured time of flight rather than the energy. As in the current `o2::dataformats::MatchInfoTOF` there is no `getSignal()` method, the best if @noferini hints how it should be extracted (I think corresponding data member should be added).
The `extraInfoHolder.tofExpMom` should be calculated as `mass * expBeta / sqrt(1. - expBeta*expBeta) `, the `expBeta` will be correct provided the expSig is fixed. I've asked @njacazio to verify this by it seems he is offline.

shahor02: PS: details of TOF parameters are here: https://indico.cern.ch/event/1010843/contributions/4242241/attachments/2193866/3708654/jacazio_2021-02-22_O2PID.pdf

nburmaso: I see. I will be waiting for the response then

noferini: The TOF time is not propagated in the MatchingInfo but it can be recovever from the cluster (tofclusters.root) with position
tofMatch.getTOFClIndex() 
If you think it is better to propagate this infromation in the matching info I can add the data member and then the proper method. Do you need it?

noferini: By accepting this
https://github.com/nburmaso/AliceO2/pull/1
you could simply do
            extraInfoHolder.tofSignal = tofMatch.getSignal();

noferini: Please, note that the interaction time is not subtracted (you get a time inside the timeframe)

nburmaso: Hi @shahor02. I have fixed `tofExpMom` calculation, could you please check if it's ok?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -194,12 +196,88 @@ void AODProducerWorkflowDPL::addToMFTTracksTable(mftTracksCursorType& mftTracksC
...
+          }
+          if (contributorsGID[GIndex::Source::ITSTPCTOF].isIndexSet()) {
+            const auto& tofMatch = data.getTOFMatch(contributorsGID[GIndex::Source::ITSTPCTOF]);
+            extraInfoHolder.tofChi2 = tofMatch.getChi2();
+            const auto& tofInt = tofMatch.getLTIntegralOut();
+            float intLen = tofInt.getL();
+            extraInfoHolder.length = intLen;
+            float mass = o2::constants::physics::MassPionCharged; // default pid = pion
+            float expSig = tofMatch.getSignal();
+            extraInfoHolder.tofSignal = expSig;
```
shahor02: I see a problem here: in the Run2 -> Run3 AOD conversion we will a truncated float value, 
https://github.com/alisw/AliPhysics/blob/bb0e58318cd9d7fc90d659a724643b0f2b1c623e/RUN3/AliAnalysisTaskAO2Dconverter.cxx#L1650
to https://github.com/alisw/AliPhysics/blob/bb0e58318cd9d7fc90d659a724643b0f2b1c623e/RUN3/AliAnalysisTaskAO2Dconverter.h#L326.
This should be fine for Run2 AliESDtrack, where the time is defined wrt to the trigger, but, as @noferini explained, in Run3, with the time defined wrt the TF start, we should use double. And, we cannot redefine this time wrt some other reference (e.g. interaction time) since in general the track is not guaranteed to be related to any reference.
I would propose to change `float expSig` and `extraInfoHolder.tofSignal` to double and add it on
https://github.com/AliceO2Group/AliceO2/blob/af43e338c5ad94efa896f9afeb38231bdd98f085/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L175 directly, w/o truncation.
Then, the `AliAnalysisTaskAO2Dconverter` should be fixed in AliPhysics, and the lines
https://github.com/AliceO2Group/AliceO2/blob/2360d52d0484a1880a77e519cdfc6705d04394a3/Framework/Core/include/Framework/AnalysisDataModel.h#L209 and https://github.com/AliceO2Group/AliceO2/blob/2360d52d0484a1880a77e519cdfc6705d04394a3/Framework/Core/include/Framework/AnalysisDataModel.h#L214-L215 should be redefined with `double tofSignal`.
Pinging @pzhristov and @jgrosseo

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -194,12 +196,88 @@ void AODProducerWorkflowDPL::addToMFTTracksTable(mftTracksCursorType& mftTracksC
...
+            float expMom = 0.f;
+            if (expSig > 0) {
+              float tof = expSig - interactionTime;
```
shahor02, line 255: And I realize now that this will not work for unassigned tracks (there the `interactionTime` is passed as -1), i.e. the `tofExpMom` will be wrong. In fact, even for assigned tracks the real TOF PID evaluation should use not the nominal interaction time (as passed to fillTrackTablesPerCollision) but precise timeZero.
Now, I am wondering why at all we storing the `tofExpMom`? It can be always recalculated from precise time-of-flight and stored L-integral.

shahor02 (commented):
I think there is some general inconsistency in the TOF info stored, see comments below.
Pinging  to @pzhristov @jgrosseo @njacazio

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
noferini: I have a comment regarding propagation of tof time to aod: float or double?
Since tof time is filled in a method called with the collisionID as argument I would expect that at this stage you already know at least what is the BC candidate. In such a case you should be able to subtract a collision time with a precision depending on what you have. I think it doesn't make sense propagate to aod tof time referred to the the TF (even if I didn't check the logic of AOD producer).

shahor02: @noferini see my comment below: for unassigned tracks the collisionID and `interactionTime` are dummy parameters.
But even for assigned tracks the current `interactionTime` is not guaranteed to be even BC-precise (and, btw, it is rounded to ns): https://github.com/AliceO2Group/AliceO2/blob/af43e338c5ad94efa896f9afeb38231bdd98f085/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L762-L767

sawenzel:
It looks like some fundamental questions remain. Can we still merge this since it's in any case an improvement? And then TOF can take directly take in adjusting it's information?

shahor02:
@sawenzel fine for me

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
jgrosseo: I do not think this fully solves it.

For assigned tracks, we could store wrt to track.collision().bc()

But as you point out unassigned tracks, do not have a collision or bc. Storing with respect of TF start also does not work, because we do not store the TF start. As Ruben pointed out earlier, we can calculate TF start from any BC in a TF. However, we merge TFs into DFs, and therefore you have to select the *right* BC for the TF start calculation (not just any).

In principle, each unassigned track appears in the AmbiguousTrack table where it is assigned a BC slice or a collision array. One could define the time wrt the first BC in that table, but we have to look up the corresponding association track->ambiguoustrack while the index goes the other way round...

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
njacazio: Hi @shahor02 apologies I missed a bit the discussion because I was traveling. 
That said, I agree with you that if the TOF expected momentum is computed as in here i.e.:
```
              float expBeta = (intLen / tof / cSpeed);
              expMom = mass * expBeta / std::sqrt(1.f - expBeta * expBeta);
```
writing the expected momentum can be avoided entirely as it can be recomputed on the fly from the quantities in the table.
**Nonetheless** the TOF expected momentum **is to be computed with the expected time of pions** and not from the measured TOF.
This also simplifies a bit the issue with the `interactionTime` as it does not enter in the computation of the expected momentum.
The procedure would be as in https://github.com/alisw/AliPhysics/blob/bb0e58318cd9d7fc90d659a724643b0f2b1c623e/RUN3/AliAnalysisTaskAO2Dconverter.cxx#L1657-L1665 where we use the integrated times for pions to define the expected momentum.

```
      const AliPID::EParticleType tof_pid = AliPID::kPion;
      const float exp_time = TOFResponse.GetExpectedSignal(track, tof_pid);
      if (exp_time > 0) { // Check that the expected time is positive
        // Expected beta for such hypothesis
        const Float_t exp_beta = (track->GetIntegratedLength() / exp_time / cspeed);

        tracks.fTOFExpMom = AliMathBase::TruncateFloatFraction(
          AliPID::ParticleMass(tof_pid) * exp_beta * cspeed /
            TMath::Sqrt(1. - (exp_beta * exp_beta)),
          mTrack1Pt);
      }
```
With the present implementation in the RUN3 AOD Produce _I don't think that it will fully work as intended_ in the PID response.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -644,103 +752,10 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
-      }
-      if (src == GIndex::Source::MFT && mFillTracksMFT) {
-        const auto& track = tracksMFT[trackIndex.getIndex()];
-        isStoredMFT[trackIndex.getIndex()] = true;
-        addToMFTTracksTable(mftTracksCursor, track, -1);
-      }
-    }
-  }
+  // fixme: interaction time is undefined for unassigned tracks (?)
+  fillTrackTablesPerCollision(-1, -1, trackRefU, primVerGIs, recoData, tracksCursor, tracksCovCursor, tracksExtraCursor, mftTracksCursor);
```
shahor02: @nburmaso :  `trackRef` should be used here instead of `trackRefU`

nburmaso: Hi @shahor02. Thank you for notice. I have fixed the typo

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: Thanks @njacazio , indeed, the calculation should be (please cross-check)
````
float expSig = = tofInt.getTOF(o2::track::PID::Pion);
extraInfoHolder.tofSignal = tofMatch.getSignal();  // to be clarified what reference time we use for unassigned tracks
float expMom = 0.f;
if (expSig > 0) {
  float expBeta = (intLen / expSig / cSpeed);
  expMom = mass * expBeta / std::sqrt(1.f - expBeta * expBeta);
}
extraInfoHolder.tofExpMom = expMom;
````
If you agree, @nburmaso, could you change this.

@jgrosseo concerning your comment:
> In principle, each unassigned track appears in the AmbiguousTrack table where it is assigned a BC slice or a collision array.

Could you point on the code you refer to? I am little bit lost, since -1 is provided as a collision ID for unassigned tracks and I don't see where the track times are used at all.

njacazio: Hi @shahor02 yes, this would work better indeed. Thanks a lot!!

sawenzel:
Merging now as is (future improvements coming in other PRs). Tested it locally.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
jgrosseo: Hi @shahor02 
I have adjusted the ambiguous tracks table to what we have discussed in this PR:
https://github.com/AliceO2Group/AliceO2/pull/6629
Maybe it makes my point more clear.

PR #6597, opened 2021-07-06
C side modules positions and FV0 alignment

AllaMaevskaya:
- C side modules position according CAD drawings 
- FV0 zero misalignment

### DIFF
--- a/Detectors/FIT/FV0/base/include/FV0Base/Geometry.h
+++ b/Detectors/FIT/FV0/base/include/FV0Base/Geometry.h
@@ -17,13 +17,17 @@
 
 #ifndef ALICEO2_FV0_GEOMETRY_H_
 #define ALICEO2_FV0_GEOMETRY_H_
-
+#include "DetectorsBase/GeometryManager.h"
+#include "DetectorsCommonDataFormats/DetID.h"
+#include <Rtypes.h>
+#include <TGeoPhysicalNode.h>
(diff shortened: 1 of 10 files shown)


AllaMaevskaya:
It stacked already two days

sawenzel (approved):

Detectors/FIT/FV0/base/include/FV0Base/Geometry.h
```diff
@@ -120,6 +124,14 @@ class Geometry
   /// \return True if cellId belongs to ring 5.
   bool isRing5(UInt_t cellId);
 
+  static constexpr o2::detectors::DetID::ID getDetID() { return o2::detectors::DetID::FV0; }
+  TGeoPNEntry* getPNEntry(int index) const
+  {
+    /// Get a pointer to the TGeoPNEntry of a chip identified by 'index'
+    /// Returns NULL in case of invalid index, missing TGeoManager or invalid symbolic name
+    return o2::base::GeometryManager::getPNEntry(getDetID(), index);
+  }
```
shahor02, line 133: Looks fine, but why do you need this aliases, which are anyway never used?

AllaMaevskaya: May be in future FV0 will want to use it

shahor02 (approved):

PR #6598, opened 2021-07-06
TPC: Improve detection/removal of duplicate digits

wiechula:


davidrohr (approved):
works, and much faster than before

davidrohr:
full CI passed, errors unrelated, merging

PR #6600, opened 2021-07-06
DPL Analysis: correctly track consumed offers

ktf:


ktf:
Extensively tested in hyperloop. Merging to make sure we can try it on the grid later today.

Barthelemy:
it is creating an overwhelmingly large number of messages such as 
```
[84484:producer-0]: [09:29:06][INFO] Message DPL/ENUM is not of kind arrow, therefore we are not accounting its shared memory
[84484:producer-0]: [09:29:06][INFO] 0MB bytes being given back to reader, totaling 0MB
```

ktf:
fixed in #6606

PR #6601, opened 2021-07-07
[TPC] Update of the simulation

a-mathis:
o fine-tuning of the simulation parameters and workflow
o code cleanup and extended documentation
o inclusion of the pad response function


Detectors/TPC/simulation/include/TPCSimulation/SAMPAProcessing.h
```diff
@@ -210,33 +210,28 @@ inline T SAMPAProcessing::getGamma4(T time, T startTime, T ADC) const
...
 inline TimeBin SAMPAProcessing::getTimeBin(float zPos) const
 {
-  float timeBin = (mDetParam->TPClength - std::abs(zPos)) / (mGasParam->DriftV * mEleParam->ZbinWidth);
-  return static_cast<TimeBin>(timeBin);
+  return static_cast<TimeBin>((mDetParam->TPClength - std::abs(zPos)) / (mGasParam->DriftV * mEleParam->ZbinWidth));
 }
 
 inline float SAMPAProcessing::getZfromTimeBin(float timeBin, Side s) const
 {
   float zSign = (s == 0) ? 1 : -1;
```
wiechula:
```suggestion
  const float zSign = (s == 0) ? 1 : -1;
```

Detectors/TPC/simulation/include/TPCSimulation/SignalInduction.h
```diff
@@ -0,0 +1,111 @@
...
+{
+
+/// \class SignalInduction
+/// This class takes care of the signal induction on the pad plane.
+/// The actual Pad Response Function (PRF) is simulated with Garfield++/COMSOL and dumped to a file.
+/// This file is read by this class and buffered for each pad size individually (IROC / OROC1-2 / OROC3).
+
+struct PadResponse {
+  PadResponse() = default;
+  PadResponse(DigitPos digiPos, double padResp)
```
wiechula:
```suggestion
  PadResponse(const DigitPos& digiPos, double padResp)
```

Detectors/TPC/simulation/src/GEMAmplification.cxx
```diff
@@ -77,12 +77,13 @@ GEMAmplification::GEMAmplification()
...
   const float effStack = mGEMParam->EfficiencyStack;
-  const float sStack = gainStack / (kappaStack * (1.f - effStack));
+  // Correct for electron losses occurring when changing kappa and the efficiency
+  const float sStack = gainStack / (kappaStack * effStack);
   polya % kappaStack % sStack % sStack % (kappaStack - 1) % sStack;
   std::string name = polya.str();
   o2::math_utils::CachingTF1* polyaDistribution = nullptr;
   if (!cacheexists) {
-    polyaDistribution = new o2::math_utils::CachingTF1("polya", name.c_str(), 0, 50000);
+    polyaDistribution = new o2::math_utils::CachingTF1("polya", name.c_str(), 0, 25.f * gainStack);
```
wiechula, line 86: Where does the 25 come from?

Detectors/TPC/simulation/src/SignalInduction.cxx
```diff
@@ -0,0 +1,106 @@
...
+
+  return {shiftedDigitPos, computePadResponse(offsetX, offsetY, gemstack)};
+}
+
+void SignalInduction::getPadResponse(const GlobalPosition3D posEle, const DigitPos digitPos, std::vector<PadResponse>& signalArray, const PadResponseMode& mode) const
+{
+  signalArray.resize(0);
+  switch (mode) {
+    case PadResponseMode::FullPadResponse: {
+      std::vector<double> offset = {{-1, 0, 1}};
```
wiechula, line 94:
```suggestion
      const std::vector<double> offset = {{-1, 0, 1}};
```
This is independent of the pad / row size always +-1?

Detectors/TPC/base/include/TPCBase/ParameterGEM.h
```diff
@@ -54,10 +59,11 @@ struct ParameterGEM : public o2::conf::ConfigurableParamHelper<ParameterGEM> {
...
-  float TotalGainStack = 1644.f;                                 ///< Total gain of the stack for the EffectiveMode
-  float KappaStack = 1.2295f;                                    ///< Variable steering the energy resolution of the full stack for the EffectiveMode
-  float EfficiencyStack = 0.473805f;                             ///< Variable steering the single electron efficiency of the full stack for the EffectiveMode
+  float TotalGainStack = 2000.f;                                 ///< Total gain of the stack for the EffectiveMode
+  float KappaStack = 1.205f;                                     ///< Variable steering the energy resolution of the full stack for the EffectiveMode
+  float EfficiencyStack = 0.528f;                                ///< Variable steering the single electron efficiency of the full stack for the EffectiveMode
   AmplificationMode AmplMode = AmplificationMode::EffectiveMode; ///< Amplification mode [FullMode / EffectiveMode]
+  PadResponseMode PRFmode = PadResponseMode::FullPadResponse;    ///< Treatment of the pad response [FullPadResponse / ProjectivePadResponse]
```
wiechula: Did you check the difference 
* it takes in time to compute the full / projective response?
* in the effective shapes?

I think in the past we used projective only, so I wonder how different the results will be...

wiechula (changes requested):
Very nice! Only very minor things.

Detectors/TPC/simulation/src/GEMAmplification.cxx
a-mathis: Took a look to the distribution and this seems to be a reasonable choice to also have the long tail well represented

Detectors/TPC/simulation/src/SignalInduction.cxx
a-mathis: yes, essentially we take a look around the center of the avalanche, and since the PRF is anyways mostly projective +/-1 is more than sufficient. even for the IROC pads

Detectors/TPC/base/include/TPCBase/ParameterGEM.h
a-mathis: In general I think the projective PRF is more than enough and should be the default. Nevertheless I think it would be nice to have at least the option to test with PRF enabled. In particular for very short drifts it also makes an appreciable difference in the dE/dx and the charge distributions so it would be really good to have this

wiechula:
@a-mathis , please check why some tests are failing after the updates.

Detectors/TPC/simulation/src/SignalInduction.cxx
wiechula: can still be `const`

wiechula (approved):
Fine for me.

sawenzel:
This PR has real test problems. They can be fixed by adding `o2::tpc::PadResponseMode` to the dictionary generation. Thereafter this can be merged immediately.

sawenzel:
@a-mathis : I think this PR now needs rebase on dev (as part of this was merged in a different PR).

PR #6602, opened 2021-07-07
DPL Analysis: When bound to a filtered table, index column now returns iterator to the original

aalkin:
When following an index, bound to a filtered table, a correct iterator to the original table is now returned.


ktf (approved):

aalkin:
Needs to be updated for recently merged slice/array index.

PR #6604, opened 2021-07-07
o2-sim: Ensure correct order of events in output file

sawenzel:
So far, the output event sequence did not
necessarily match the order in whey they
were generated. This is now ensured to avoid
bookkeeping errors.

This comes at slight memory cost, as events
can't be flushed as soon as they arrive.

Minor other changes:
- fix pythia8 call in example


sawenzel:
MacOS compile checked locally.

PR #6605, opened 2021-07-07
MCH: Include CH6L in electronic mapping

aphecetche:


aferrero2707 (approved):
CH6L mapping tested with real data, all looks good.

PR #6612, opened 2021-07-08
PWGHF: Add task to perform optimisation of preselections 

fgrosa:


vkucera (commented):
Hi @fgrosa , on the first look, it seems fine to me. I have a general comment, though.

It is better to use the `<object><attribute>` (or `<general><specific>`) naming scheme, so that names of the same kind of objects start with same string and the different attributes follow.
This scheme makes names more readable, searchable and sortable.

For example, I think that
```c++
nCutsToTestCosp
nCutsToTestDecLen
nCutsToTestPtTrackMin

vecCutsCosp
vecCutsDecLen
vecCutsPtTrackMin
```
is more readable (and sortable) than
```c++
nCospCutsToTest
nDecLenCutsToTest
nMinTrackPtCutsToTest

cospCutsVec
decLenCutsVec
minTrackPtCutsVec
```
Btw, you already use that scheme for histogram names.
```c++
"hPromptVsPtD0ToPiK"
"hPromptVsPtJpsiToEE"
"hPromptVsPt2Prong"
```

fgrosa:
> Hi @fgrosa , on the first look, it seems fine to me. I have a general comment, though.
> 
> It is better to use the `<object><attribute>` (or `<general><specific>`) naming scheme, so that names of the same kind of objects start with same string and the different attributes follow.
> This scheme makes names more readable, searchable and sortable.
> 
> For example, I think that
> 
> ```c++
> nCutsToTestCosp
> nCutsToTestDecLen
> nCutsToTestPtTrackMin
> 
> vecCutsCosp
> vecCutsDecLen
> vecCutsPtTrackMin
> ```
> 
> is more readable (and sortable) than
> 
> ```c++
> nCospCutsToTest
> nDecLenCutsToTest
> nMinTrackPtCutsToTest
> 
> cospCutsVec
> decLenCutsVec
> minTrackPtCutsVec
> ```
> 
> Btw, you already use that scheme for histogram names.
> 
> ```c++
> "hPromptVsPtD0ToPiK"
> "hPromptVsPtJpsiToEE"
> "hPromptVsPt2Prong"
> ```

Hi @vkucera, thanks for your feedback! Indeed I did not pay too much attention to the naming scheme, I should have now changed all the variables according to your suggestion!

fgrosa:
This PR depends on https://github.com/AliceO2Group/AliceO2/pull/6783 (thanks @mpuccio!)

ginnocen (approved):

PR #6613, opened 2021-07-08
AOD: first steps to add forward tracks (mch standalone ones so far)

aphecetche:
It's mostly "plumbing" for now. 

The actual work of producing extrapolated-to-vertex mch tracks will have to be updated at some point, when we'll get real time windows associated with each track. For the moment we simply hard code a time window of 1.4 microsecond (around the ROF that's coming from the digits) for each track.

As a side note (not really related to this PR) there's quite a bit of code duplication in the `AODProducerWorkflowSpec.cxx` (for unassigned and assigned tracks) that should be addressed at some point IMO.


sawenzel:
This needs conflict resolution

aphecetche:
@sawenzel given the number of structural/logic changes in e.g. `AODProducerWorkflowSpec.cxx` that have been committed meanwhile might take some time.

aphecetche:
ok, tried to quickly fix the conflicts. There are a couple of open questions that will have to be resolved at some point, but I guess for the time being, this first step is "good enough" to be considered for merging.

sawenzel:
Nice. One question: Is the conversion automatically trying to access MCH tracks and will it crash if such a file is not present? I'd like to avoid breaking the O2DPG workflow: `https://github.com/AliceO2Group/O2DPG/blob/master/MC/bin/o2dpg_sim_workflow.py`

sawenzel:
@aphecetche : Sorry, but we need one more round of conflict resolution.

aphecetche:
@sawenzel I've re-fixed the conflicts. 
Concerning your question about the (non) crash if MCH tracks are not present, I'll check (but turns out that for the moment the sim-challenge is failing for me, on macOS or Ubuntu, even for dev version, trying to understand what's going on).

aphecetche:
bummer. yet another conflict. did not see that one coming. my bad. back to conflict resolution mode then...

aphecetche:
@sawenzel with #6838 I was able to run `sim_challenge.sh`. If I remove `mchtracks.root`and redo : `JOBUTILS_SKIPDONE=1 ../O2/prodtests/sim_challenge.sh` I do indeed get a crash. 

But I'd expect that to be expected in fact, right ? At least it seems there's no protection for any other kind of tracks ? (did not try all of them, but quickly tried to remove mfttracks.root  and got a crash also)

aphecetche:
@sawenzel hum, in addition seems in the various conflict resolution battles, the actual fwdtrack table filling was lost somehow (i.e. on my local tests the table is empty in the output AOD)... I'm investigating...

aphecetche:
ok, so either my initial implementation was wrong or something that I've not been able to identify was changed since the initial PR, but basically I was using : 

`data.getMCHTrack(trackIndex.getIndex())` instead of `data.getMCHTrack(trackIndex)` to get access to MCH tracks in `AODProducerWorkflowSpec.cxx`. Was getting back garbage and extrapolation to vertex failed for all MCH tracks, hence no fwd tracks.

@rpezzi note that the same getIndex mistake was done for MFT tracks also as far as I can tell. I've fixed it as well, please x-check if that's ok with you.

aphecetche:
one more conflict to be resolved...

rpezzi:
> @rpezzi note that the same getIndex mistake was done for MFT tracks also as far as I can tell. I've fixed it as well, please x-check if that's ok with you.

@aphecetche , thanks for spotting this. I am yet to get in par with these changes. I trust you, please proceed. I will soon get at AOD again to improve ambiguities on track-vertex association.

aphecetche:
@shahor02 @sawenzel as far as I can tell the two failing checks are not actually related to this PR.
Can it be reviewed/merged ? (would like to avoid yet another round of conflict resolution, and the longer we wait for the merge the more likely conflict appearance is ;-) )

pzhristov:
Yes, the issue with the tests is under investigation. I can squash and merge this PR.

PR #6614, opened 2021-07-08
[EMCAL-694] Summing all digits within one tower using SDigitizer

hahassan7:
The SDigitizer has been updated such that it can now sums all the different digits (with their labels) in one tower.
Also we moved the EMCALDigitizerSpec into the EMCAL workflow.


davidrohr:
@mfasDa : Can this be merged, and shall I retry the FST with this?

mfasDa (approved):
Looks fine to me

PR #6620, opened 2021-07-09
It's enough to use MCHeader file to create digi context

sawenzel:


sawenzel:
Trivial and compilation checked locally.

PR #6621, opened 2021-07-09
DPL: force processing of region callback without waiting for data

ktf:
Without this, it is possible that a late arriving event will not be processed
until some data actually arrives on one of the channels, unblocking the event
loop. This should obviate to the problem by triggering an aync libuv event
which will unblock the event loop without the need for data.


ktf:
@davidrohr not tested, but should improve the delivery of the callbacks.

davidrohr (approved):
I checked this and it doesn't break my tests.
But we need a release and RPMs to deploy on the EPNs and test with AliECS. Probably still needs some patches elsewhere.
Since it doesn't break anything, I propose to merge it.

PR #6624, opened 2021-07-09
trd fix sending whole messages

bazinski:
send whole messages instead of fragmented ones.
code clean up as well.

tested via 
`o2-raw-file-reader-workflow -b --session default --loop 1000 --delay 3 --input-conf raw/TRD/TRDraw.cfg --drop-tf "TRD,10" | o2-trd-datareader -b | o2-trd-tracklet-transformer --shm-segment-size 32000000000 -b`
all seems happy

### DIFF
--- a/Detectors/TRD/reconstruction/include/TRDReconstruction/DataReaderTask.h
+++ b/Detectors/TRD/reconstruction/include/TRDReconstruction/DataReaderTask.h
@@ -40,6 +40,7 @@ class DataReaderTask : public Task
   void init(InitContext& ic) final;
   void sendData(ProcessingContext& pc, bool blankframe = false);
   void run(ProcessingContext& pc) final;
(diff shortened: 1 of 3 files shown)


martenole:
Thanks! I see in the FST that the number of tracklets read from the raw files matches the number of simulated tracklets. Still, the number of attached tracklets is extremely low. I am not yet sure what causes this, but I will have a look at it today.

bazinski:
Hi @martenole you should be able to run this with `--fileper all` and circumvent the need for this merge. Applied to trap2raw of course. As for the number of attached tracklets I assume that is unrelated to this patch of even the rawreader, but to the incoming simulation?

martenole:
Yes, that issue is not related to this PR. Probably related to the track matching itself, I am looking into it.

tdietel (approved):

PR #6625, opened 2021-07-10
add DistStf as mandatory input for TOF decoder

noferini:


shahor02 (dismissed):

noferini:
failing checks doesn't appear to be related to this PR

shahor02:
Hi @noferini 
The fullCI has failed for unrelated reasons, but in the log I see all entropy-encoders writing data except the `tof-entropy-encoder`, also the `tof-matcher` did not produce any output, seems like the TOF data did not reach processors, but I don't see the "missing TF message" either. Something looks fishy, could you run the FST locally, please?

noferini:
TOF reco workflow work with option
--ignore-dist-stf
when skipping
("stdDist", "FLP", "DISTSUBTIMEFRAME", 0, Lifetime::Timeframe)
It seems that tfor some reason this input is not digested.

This is the command I ran

o2-raw-file-reader-workflow -b --input-conf TOFraw.cfg |o2-tof-compressor -b | o2-tof-reco-workflow -b --input-type raw --output-type clusters 

What is wrong?

Is DISTSUBTIMEFRAME stopped by compressor?

noferini:
> TOF reco workflow work with option
> --ignore-dist-stf
> when skipping
> ("stdDist", "FLP", "DISTSUBTIMEFRAME", 0, Lifetime::Timeframe)
> It seems that tfor some reason this input is not digested.
> 
> This is the command I ran
> 
> o2-raw-file-reader-workflow -b --input-conf TOFraw.cfg |o2-tof-compressor -b | o2-tof-reco-workflow -b --input-type raw --output-type clusters
> 
> What is wrong?
> 
> Is DISTSUBTIMEFRAME stopped by compressor?

Who is producing DISTSUBTIMEFRAME? RawFileReader?

noferini:
It seems to me that 
 --ignore-dist-stf option is mandatory if not running with DataDistribution.

shahor02:
@noferini The DISTSUBTIMEFRAME is produced both by the DataDistribution and by the RawFileReader.
I guess the problem is that here: https://github.com/AliceO2Group/AliceO2/blob/27792e4735309219871c0065b4b6c7a80bace63d/Detectors/TOF/workflow/src/CompressedDecodingTask.cxx#L171-L187
you are processing all inputs (including this DISTSUBTIMEFRAME) pretending these are compressed data. You should add a filter to select the TOF/CRAWDATA only.

noferini:
I 'm adding also Roberto since we are touching the decoder 
@preghenella

shahor02:
@noferini you don't need to touch the decoder itself, simply, itself of the https://github.com/AliceO2Group/AliceO2/blob/27792e4735309219871c0065b4b6c7a80bace63d/Detectors/TOF/workflow/src/CompressedDecodingTask.cxx#L171 loop you can loop over filtered data, e.g.

````
std::vector<InputSpec> sel{InputSpec{"filter", ConcreteDataTypeMatcher{"TOF", "CRAWDATA"}}};
 for (const auto& ref : InputRecordWalker(pc.inputs(), sel) {
 ...
````
Alternatively, you can simply reject in the original loop whatever is has not the header.DataDescription you need.

noferini:
Hi Ruben,
the problem is different.
I saw this message
[46916:tof-compressed-decoder]: [10:36:15][WARN] Found input [TOF/CRAWDATA/0xdeadbeef] TF#0 1st_orbit:0 Payload 0 : assuming no payload for all links in this TF
which means that it return before to go into the loop.

noferini:
Could it be due to the fact that the DISTSUBTIMEFRAME is much faster than CRAWDATA which comes from compressor?

noferini:
Filtering added. Now it works if I ask CRAWDATA as mandatory (not asked for the moment).
If CRAWDATA optional and DISTSUBTIMEFRAME requested, decoder invoked before the arrival of RAW DATA.
I don't know how to manage this (is there the possibility to set a longer timeout?)

shahor02:
Hi @noferini 
what is the exact workflow you are testing? The DISTSUBTIMEFRAME is sent by the STFBuilder (and TFBuilder on EPN), so it indeed should be available earlier than the output of the compressor. For the delays, I am not sure, @ktf should comment.
Formulating question for @ktf: the protection for missing STF in TOF is implemented on the compressed data decoder level, rather than raw data, and he sees that the "expired messages" mechanism triggers dummy CRAWDATA input creation even though the real is supposed to be there (but may arrive later than the FLP/DISTSUBTIMEFRAME since this latter comes directly from the stfbuilder, while the CRAWDATA comes from the device digesting the raw data coming from stfbuilder. 
Is it possible that this is a problem of delays?

noferini:
> Hi @noferini
> what is the exact workflow you are testing? The DISTSUBTIMEFRAME is sent by the STFBuilder (and TFBuilder on EPN), so it indeed should be available earlier than the output of the compressor. For the delays, I am not sure, @ktf should comment.
> Formulating question for @ktf: the protection for missing STF in TOF is implemented on the compressed data decoder level, rather than raw data, and he sees that the "expired messages" mechanism triggers dummy CRAWDATA input creation even though the real is supposed to be there (but may arrive later than the FLP/DISTSUBTIMEFRAME since this latter comes directly from the stfbuilder, while the CRAWDATA comes from the device digesting the raw data coming from stfbuilder.
> Is it possible that this is a problem of delays?

I simply run
o2-raw-file-reader-workflow -b --input-conf TOFraw.cfg |o2-tof-compressor -b | o2-tof-reco-workflow -b --input-type raw --output-type clusters

noferini:
workflow

ktf:
I think the problem is due to the fact you have two separate inputs which arrive unsynchronised. The "optional" mechanism will trigger on the first input which arrives. Can you add the second input also to the tof compressor?

noferini:
> I think the problem is due to the fact you have two separate inputs which arrive unsynchronised. The "optional" mechanism will trigger on the first input which arrives. Can you add the second input also to the tof compressor?

Added. Now it works.
@preghenella please check that changes in TOF compressor task are fine with you

noferini:
FYI
Also tested with
1) raw-reader/compressor(output-proxy
2) input-proxy/tof-reco
and it works

shahor02 (approved):

PR #6626, opened 2021-07-10
[O2-2461] MCHRaw: bc offset should be applied before the rollover

aphecetche:
@shahor02  should fix [O2-2461](https://alice.its.cern.ch/jira/browse/O2-2461)


shahor02 (approved):
Fixes broken test, merging

PR #6630, opened 2021-07-12
DPL Analysis: introduce self-index column type

aalkin:
* add `DECLARE_SELF_INDEX_COLUMN` for defining the index columns referring to their own table
* add binding hooks to analysis task so that self index is automatically set
* add a test

@jgrosseo this seems to work as expected, but I still need to test it with an actual analysis task.


sawenzel:
Nice. For my understanding, one application of this is in MCTracks who reference themselves for daughter + mothers?
If yes, one can test this with some MC tracks filtering/grouping analysis.

aalkin:
@sawenzel yes, this is the original intent. I'll check if I can improve the internals and then rewrite MC track table to use it.

aalkin:
@jgrosseo this now works with mc input files we have, but for some reason using old and new `isPhysicalPrimary` produces different results, could you also have a look? I used train 181 files to test.

@sawenzel this also works on nightlies. Unrelated to this PR, but I see that for nightlies in the output histograms for phi and eta difference between reco and gen levels are very wide, this indicates a problem with mccollision and/or mctracklabels in the input.

jgrosseo (changes requested):
Validating. Please do NOT merge, yet.

jgrosseo:
Included in https://github.com/AliceO2Group/AliceO2/pull/6688
Closing

PR #6640, opened 2021-07-14
Add V0s table to the converter

nburmaso:


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -265,6 +265,9 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
           }
           addToTracksTable(tracksCursor, tracksCovCursor, trackPar, collisionID, src);
           addToTracksExtraTable(tracksExtraCursor, extraInfoHolder);
+          // collecting table indices of barrel tracks for V0s table
+          mV0sIndices[trackIndex.getIndex()] = mTableTrID;
+          mTableTrID++;
```
shahor02, line 272: the `mTableTrID` is never reset, should be done together with resetting mV0sIndices at https://github.com/AliceO2Group/AliceO2/blob/e1ffc18da4c2cdd9ae0f8be17c13ace252f04b90/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L824

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
```diff
@@ -160,9 +160,13 @@ class AODProducerWorkflowDPL : public Task
   int64_t mTFNumber{-1};
   int mTruncate{1};
   int mRecoOnly{0};
-  bool mFillSVertices{false};
   TStopwatch mTimer;
 
+  // unordered map connects global indices and table indices of barrel tracks
+  // the map is used for V0s table filling
+  std::unordered_map<int, int> mV0sIndices;
```
shahor02: why are you using `int` as a key of the `mV0sIndices`. The V0 prong reference may be any barrel GlobalTrackID, I guess it should be used a key directly rather than using only its index part.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -793,10 +798,31 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
   }
 
+  // filling v0s table
+  for (auto& svertex : secVertices) {
+    auto trPosID = svertex.getProngID(0);
+    auto trNegID = svertex.getProngID(1);
+    int posTableIdx = -1;
+    int negTableIdx = -1;
+    LOG(DEBUG) << "V0s: " << trPosID.getIndex() << " " << trNegID.getIndex();
+    auto item = mV0sIndices.find(trPosID.getIndex());
```
shahor02: Same as above, the `trPosID.getIndex()` is not enough to define the prong track reference, e.g. the it could be ITS-TPC or ITS-TPC-TRD track with the same `getIndex()` but there are completely different track entries.
Use GlobalTrackID as the `mV0sIndices` key and, respectively, mV0sIndices.find(...) should be called with e.g. `trPosID` rather than `trPosID.getIndex()`

shahor02 (changes requested):
Hi @nburmaso 
Thanks, see a few comments below. In principle, also cascades can be added in the same PR, as a pair or V0 index and bachelor track index.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Thank you for notice. Fixed

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Thank you for advice @shahor02. I didn't notice that GlobalTrackIndex has a defined hash function, so I tried using `getIndex()`. I have switched to GIndex for keys

nburmaso:
@shahor02, I have also added cascades table

sawenzel:
Removed the draft label to trigger CI.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -810,19 +814,30 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+    if (item != mGIDToTableID.end()) {
       posTableIdx = item->second;
     }
-    item = mV0sIndices.find(trNegID);
-    if (item != mV0sIndices.end()) {
+    item = mGIDToTableID.find(trNegID);
+    if (item != mGIDToTableID.end()) {
       negTableIdx = item->second;
     }
     v0sCursor(0, posTableIdx, negTableIdx);
```
shahor02, line 829: I would produce an exception if posTableIdx or negTableIdx stay -1.

shahor02 (commented):
@nburmaso looks fine, just 1 comment below

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Hi @shahor02. Thank you, I have added a check for this https://github.com/AliceO2Group/AliceO2/blob/b4c18b8454c7777da3626bd967b7cb9f8ba514d3/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L817-L828

shahor02: could you do this also for the cascades?

shahor02 (approved):

sawenzel:
This broke the O2DPG workflow:
```
[WARN] MC-truth is not supported for FT0 recpoints currently
[ERROR] invalid workflow in o2-aod-producer-workflow: No matching output found for GLO/V0s/0 as requested by data processor "aod-producer-workflow". Candidates
:
-DPL/ENUM/105388270
-DPL/ENUM/1060785655
-
```

Could you please advice how `https://github.com/AliceO2Group/O2DPG/blob/master/MC/bin/o2dpg_sim_workflow.py` needs to be adjusted.
@nburmaso @shahor02

shahor02:
@sawenzel the script should run the SVertexer, I guess it should be (assuming the `needs` is recursive):
````
   SVFINDERtask = createTask(name='svfinder_'+str(tf), needs=[PVFINDERtask['name']], tf=tf, cwd=timeframeworkdir, lab=["RECO"], cpu=NWORKERS, mem='4000')
   SVFINDERtask['cmd'] = 'o2-seconday-vertexing-workflow ' + getDPL_global_options(nosmallrate=False)
   workflow['stages'].append(SVFINDERtask)
````

sawenzel:
Thank you!

sawenzel:
Added it but it doesn't fix the error message.

sawenzel:
https://github.com/AliceO2Group/AliceO2/pull/6655 solves it for me.

PR #6641, opened 2021-07-14
Export Framework headers required for O2Physics

TimoWilken:
In order to factor out O2Physics into its own repository, some extra O2 headers need to be exposed.

So far, these are only the headers required for `Analysis/Tutorials/src/histograms.cxx` -- copying other analysis code might require more headers to be exposed.


jgrosseo (approved):

jgrosseo:
Ci unrelated. Merging.

PR #6642, opened 2021-07-14
A utility to engineer collision structures in a timeframe

sawenzel:
Going beyond what was possible until now
via SimReader:

* make collision contexts of arbitrary timelength

  `o2-steer-colcontexttool -i bkg,50000 --tflength 10`
  (generated a context for 10ms timeframes and 50kHz interaction rate)

* allow to inject only into every N-th background event

  `o2-steer-colcontexttool -i bkg,50000 sgn,@0:e5`
  (injects signals into every 5-th background event)

* allow to inject respecting a minimal time distance between signal events

  `o2-steer-colcontexttool -i bkg,50000 sgn,@0:d10000`
  (injects signals into background but minimally 10000ns apart)

* allow to overlay arbitrary interactions (without matching of vertex)

  `o2-steer-colcontexttool -i bkg,50000 sgn,40000`
  (2 unrelated collisions, the first one at 50kHz the second one at 40kHz)

* allow to inject multiple signal events

  `o2-steer-colcontexttool -i bkg,50000 sgn1,@0:e1 sgn2,@0:e1`
  (injects 2 signals into every background event)

The tool can be upgraded to allow editing of existing contexts, etc. in the future.
One can also treat QED events in this new fashion without special treatment in digitizers.

To use the tool, simply generate a context with it before digitization
```
o2-sim -n ... -o bkg
o2-sim ... -o sgn
o2-steer-colcontexttool -i bkg,40000 sgn,@0:e2
o2-sim-digitizer-workflow --incontext collisioncontext.root --sims bkg
```

### DIFF
--- a/DataFormats/simulation/include/SimulationDataFormat/DigitizationContext.h
+++ b/DataFormats/simulation/include/SimulationDataFormat/DigitizationContext.h
@@ -127,7 +127,7 @@ class DigitizationContext
   std::vector<o2::InteractionTimeRecord> mEventRecordsWithQED;
   std::vector<std::vector<o2::steer::EventPart>> mEventPartsWithQED;
 
-  o2::BunchFilling mBCFilling; // patter of active BCs
+  o2::BunchFilling mBCFilling; // pattern of active BCs
 
   std::vector<std::string> mSimPrefixes;             // identifiers to the hit sim products; the key corresponds to the source ID of event record
   std::string mQEDSimPrefix;                         // prefix for QED production/contribution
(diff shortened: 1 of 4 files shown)


shahor02 (dismissed):
@sawenzel looks nice (though I did not analyze every line). What is missing, I think, is the possibility to impose an arbitrary bunch filling scheme, you can simply copy it from 
https://github.com/AliceO2Group/AliceO2/blob/4989c1113ed8c7324126f0ca99ec1f8fe253dfe7/Steer/DigitizerWorkflow/src/SimReaderSpec.cxx#L125-L128

PR #6644, opened 2021-07-15
DPL Analysis: add an exception on incorrect join

aalkin:
@jgrosseo this will show which tables cannot be joined

@ktf could you please take a look? I had to move `DECLARE_SOA_STORE()` to `ASoA.h` from `AnalysisDataModel.h` which is not great, but I do not see a workaround.

### DIFF
--- a/Framework/Core/include/Framework/AnalysisTask.h
+++ b/Framework/Core/include/Framework/AnalysisTask.h
@@ -31,6 +31,7 @@
 
 #include <arrow/compute/kernel.h>
 #include <arrow/table.h>
+#include <arrow/util/key_value_metadata.h>
 #include <gandiva/node.h>
 #include <type_traits>
 #include <utility>
--- a/Framework/Core/include/Framework/TableBuilder.h
+++ b/Framework/Core/include/Framework/TableBuilder.h
@@ -617,6 +617,8 @@ class TableBuilder
   }
 
  public:
+  void setLabel(const char* label);
+
   TableBuilder(arrow::MemoryPool* pool = arrow::default_memory_pool())
     : mHolders{nullptr},
       mMemoryPool{pool}
@@ -833,7 +835,7 @@ auto spawner(framework::pack<C...> columns, arrow::Table* atable, const char* na
   for (auto i = 0u; i < sizeof...(C); ++i) {
     arrays.push_back(std::make_shared<arrow::ChunkedArray>(chunks[i]));
   }
-
+  new_schema = new_schema->WithMetadata(std::make_shared<arrow::KeyValueMetadata>(std::vector{std::string{"label"}}, std::vector{std::string{name}}));
   return arrow::Table::Make(new_schema, arrays);
 }
 
--- a/Framework/Core/include/Framework/TableTreeHelpers.h
+++ b/Framework/Core/include/Framework/TableTreeHelpers.h
@@ -122,8 +122,12 @@ class TreeToTable
  private:
   std::shared_ptr<arrow::Table> mTable;
   std::vector<std::string> mColumnNames;
+  std::string mTableLabel;
 
  public:
+  // set table label to be added into schema metadata
+  void setLabel(const char* label);
+
   // add a column to be included in the arrow::table
   void addColumn(const char* colname);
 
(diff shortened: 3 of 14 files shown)


ktf:
Indeed, I think DECLARE_SOA_STORE should stay where it is... Why did you have to move it?

Framework/Core/src/ASoA.cxx
```diff
@@ -11,17 +11,20 @@
...
 
-std::shared_ptr<arrow::Table> ArrowHelpers::joinTables(std::vector<std::shared_ptr<arrow::Table>>&& tables)
+std::shared_ptr<arrow::Table> ArrowHelpers::joinTables(std::vector<std::shared_ptr<arrow::Table>>&& tables, std::vector<const char*> names)
 {
   if (tables.size() == 1) {
     return tables[0];
   }
   for (auto i = 0u; i < tables.size() - 1; ++i) {
-    assert(tables[i]->num_rows() == tables[i + 1]->num_rows());
+    if (tables[i]->num_rows() == tables[i + 1]->num_rows()) {
```
ktf: I would use the arrow::Schema metadata information to hold the name. Any reason for not doing that?

Framework/Core/src/ASoA.cxx
```diff
@@ -11,17 +11,20 @@
...
+      throw o2::framework::runtime_error_f("Tables %s and %s have different sizes and cannot be joined!", names[i], names[i + 1]);
```
jgrosseo: Can we also print the actual sizes of the two tables?

Framework/Core/src/ASoA.cxx
aalkin: I was not aware Schema had metadata, indeed that would be a much better solution.

Framework/Core/src/ASoA.cxx
aalkin: Added.

Framework/Core/include/Framework/AnalysisTask.h
```diff
@@ -31,6 +31,7 @@
 
 #include <arrow/compute/kernel.h>
 #include <arrow/table.h>
+#include <arrow/util/key_value_metadata.h>
```
ktf, line 34: Can we avoid exposing key_value_metadata.h by default?

Framework/Core/include/Framework/TableTreeHelpers.h
```diff
@@ -122,8 +122,12 @@ class TreeToTable
  private:
   std::shared_ptr<arrow::Table> mTable;
   std::vector<std::string> mColumnNames;
+  std::string mTableLabel;
```
ktf, line 125: Why an extra mTableLabel here? Don't you use the metadata of mTable?

Framework/Core/include/Framework/AnalysisTask.h
aalkin: I'm looking into it. Without this include compilation fails with undefined sizeof on a forward-declared KeyValueMetadata, but only with some workflows.

Framework/Core/include/Framework/TableBuilder.h
```diff
@@ -833,7 +835,7 @@ auto spawner(framework::pack<C...> columns, arrow::Table* atable, const char* na
   for (auto i = 0u; i < sizeof...(C); ++i) {
     arrays.push_back(std::make_shared<arrow::ChunkedArray>(chunks[i]));
   }
-
+  new_schema = new_schema->WithMetadata(std::make_shared<arrow::KeyValueMetadata>(std::vector{std::string{"label"}}, std::vector{std::string{name}}));
```
ktf, line 838: You can probably have a non templated helper function for this, which allows you not exposing the dependency on the key value stuff.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: Since this class creates the table, I need to supply it with label first.

Framework/Core/include/Framework/TableBuilder.h
aalkin: True, this can go in TableBuilder.cxx

ktf:
Given all the checks are green, I am merging this. However, please try to see if we can hide things a bit.

PR #6647, opened 2021-07-16
DPL: fix unassigned variable

anerokhi:


ktf:
Nicely spotted. What compilation flag did you use?

anerokhi:
AFAIR it is `-Wuninitialized` (or `-Wmaybe-uninitialized`) which is part of `-Wall` which is... not part of O2 build flags. A MLOC-sized C++ project doesn't use even `-Wall` flag, when `-Wall -Wextra -pedantic -Werror` should be used (maybe minus few defective GCC warnings)…

ktf:
I think the rationale at some point was that -Wall was using different flags on different platforms, which was deemed to be an issue by some and eventually turned down by WP3. The idea was to enable only the reasonably common flags explicitly.

anerokhi:
> I think the rationale at some point was that -Wall was using different flags on different platforms

Sounds like a great opportunity to fix them all!

ktf:
> Sounds like a great opportunity to fix them all!

I agree, but apparently considering all platforms as production was not an acceptable solution when this was discussed.

anerokhi:
> I agree, but apparently considering all platforms as production was not an acceptable solution when this was discussed.

Just consider other platforms as additional checkers))0)0

ktf:
I agree, but at the time the answer was usually "we should not waste time on (macOS|Ubuntu)". As I said in the framework meeting, I will raise the point again in WP3.

PR #6650, opened 2021-07-16
MathUtils: Fix a copy-paste bug

anerokhi:
The bug was introduced in [ae8838eec8c3148fef407725f69b416251c9ff8f](https://github.com/AliceO2Group/AliceO2/commit/ae8838eec8c3148fef407725f69b416251c9ff8f#diff-6e348dd817fa6913edb582f505ffeb8bc53df0146c56e93aa2a49c5346731c5bL285-R289)


sawenzel:
@anerokhi : Could you try to give more informative commit messages? To me it is not "obvious" that there is a bug here.
Please state what was wrong (how you noticed) and how a commit fixes it.

davidrohr:
Also I find the title misleading, how is it related to GPU?

anerokhi:
@davidrohr Indeed, I copied the prefix from the ae8838eec8c3148fef407725f69b416251c9ff8f commit, but the fix is not for GPU code

shahor02 (approved):
trivial fix, merging

PR #6651, opened 2021-07-16
Silence the most annoying unused variable

anerokhi:


ktf:
Why not simply dropping writer (or even the line, assuming `cursor<T>()` does not have side effects)?

anerokhi:
I don't know the details and I conservatively assumed that the result of `b.cursor<T>()` call should be alive till `b.finalize()` invocation.

ktf:
Ok, but then one should assume `[[used]]`, no?

anerokhi:
I'm not sure I understand what you mean

ktf:
I meant that if you do not know if something has side effect, the correct way of marking it is to assume it's used. That said, it's just academic.

PR #6652, opened 2021-07-16
No pessimizing moves

anerokhi:
Who taught you to do this?


sawenzel:
> Who taught you to do this?

I am sorry but this is not good style. Rather than blaming/attacking your colleagues, I would favour commit messages that 
explain the rational of your improvement (for instance how you found the bug, links to coding recommendations etc.).

PR #6654, opened 2021-07-16
TRD fix endpoint/side confusion, sort tracklets

bazinski:
- fix a mixup between endpoints and side.
- fix hcid in the process
- sort outgoing tracklets.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
@@ -67,6 +67,65 @@ struct HelperMethods {
     }
     return irob % 2;
   }
+
+  static int getStack(int det)
+  {
+    return det % (constants::NSTACK * constants::NLAYER) / constants::NLAYER;
+  };
+  static int getLayer(int det)
+  {
+    return det % constants::NLAYER;
+  };
+
+  static int getORIinSuperModule(int detector, int readoutboard)
+  {
+    //given a detector and readoutboard
+    int ori = -1;
+    int trdstack = HelperMethods::getStack(detector);
+    int trdlayer = HelperMethods::getLayer(detector);
+    int side = HelperMethods::getROBSide(readoutboard);
+    //TODO ccdb lookup of detector/stack/layer/side for link id.
+    bool aside = false;
+    if (trdstack == 0 || trdstack == 1) {
+      aside = true; //aside
+    } else {
+      if (trdstack != 2) {
+        aside = false; //cside
+      } else {
+        if (side == 0) {
+          aside = true; //stack
+        } else {
+          aside = false; //stack2, halfchamber 1
+        }
+      }
+    }
+    if (aside) {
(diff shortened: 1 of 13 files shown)


bazinski:
@martenole tracklets are now sorted, still not sure why the sort is needed, as the sort should be implied. The sort may or maynot stay when i sort out my confusion. with --trd-datareader-dataverbose you will get a dump at the end of all the tracklets, slow but thorough.

martenole:
@bazinski thanks, but I still don't see an improvement in the tracking. Actually, I tested this with the full system test and dumped the calibrated tracklets. Those are all at x=y=z=0 which cannot be correct.
Maybe the tracklet sorting is not needed in the end. There might be another more fundamental problem.

bazinski:
hcid, pos, slope, row and col are non zero, at least for the parts i just looked through. Will take a look at why x,y, and z end up zero.

bazinski:
I dont get x=y=z=0 will send pics on mattermost.

bazinski:
When running with --trd-datareader-dataverbose TRDTrackletTransformer gets called twice. I think its due to the speed of the datareader which in my example is 21000ms vs 100ms for with and without the option invoked respectively.
I will review the clearly overly verbose output in a seperate PR

bazinski:
This was all green, current error is not related. Could we get this approved and merged? @tdietel @martenole @jolopezl

martenole:
It does not make it worse for me, but still the issue is not fixed. When I convert a small simulation to raw and then read the tracklets from the raw files and compare them to the originally simulated tracklets I expect them to be the same. For offset/slope/pad row this seems to be true, but the half-chamber ID is not correct.
In the tracklets which I read from the raw files I see for example HCIDs in the range 804..816. This is sector 13, stack 2 (part of the PHOS hole) and there shouldn't be any tracklets.
trdtrackletsInPHOSHole

bazinski:
Incoming tracklets are sorted by linkid not hcid. Not equivalent. Will put mapping in TDP. Mapping should come from CCDB at some point, but for now algorithmic.
PhosHole
Outgoing tracklets are resorted by hcid.
Some other minor fixes to get all working.
mc to raw to parsing is consistent, whether its correct remains to be seen.
Some of the output was reduced as well.

bazinski:
a better drawn graph and showing the hole properly for phos.
There is a problem with the charge bits, but will postpone fixing for later. Q1 from sim looks wrong as does Q2, not related to this code, but Q1 looks binary and Q2 has a bump on the left, again not related to this code.
Canvas_1

bazinski:
@martenole could you verify the hcid linkid thing and we can get this merged?

What is the best way to have unit tests with outside data? pull in from eos? or simply run the fullsystem test (trd relevant parts) inside the unit tests?

martenole:
@bazinski thanks a lot this looks much better. The HCID for the simulated tracklets looks identical to the one from the raw files.

I also get a reasonable number of matched tracklets in the asynchronous reco. In the synchronous reco not, but this is because all the trigger records are filtered out. I need to check what is going wrong there. But since the asynchronous reco looks fine and that one uses the same TRD trigger records as the synch. reco it shouldn't have anything to do with the raw reader.

bazinski:
Hi @shahor02 is it ok to merge this? or should we wait for @tdietel ?

tdietel (approved):

bazinski:
Hi @davidrohr if you get a chance could you merge this?

PR #6655, opened 2021-07-16
Fix typo in RecoContainer concerning V0s

sawenzel:
The input spec should be all capitalized `V0S` to be consistent with output spec definitions.


shahor02 (approved):
Thanks!

sawenzel:
Just a string constant change. Tested locally. Merging.

PR #6658, opened 2021-07-16
Use `<fairmq/FwdDecls.h>` instead of downstream forward declarations

dennisklein:
This makes the forward declaration techniques more future-proof towards
renamed typenames in FairMQ, e.g. if forward declared types become a
type alias.

`<fairmq/FwdDecls.h>` is available since FairMQ `v1.4.39`.

---

* This PR supersedes #6657.
* FairMQ bump is proposed in alisw/alidist#3168.

### DIFF
--- a/Detectors/Base/include/DetectorsBase/Detector.h
+++ b/Detectors/Base/include/DetectorsBase/Detector.h
@@ -36,8 +36,7 @@
 #include <unistd.h>
 #include <cassert>
 
-class FairMQParts;
-class FairMQChannel;
+#include <fairmq/FwdDecls.h>
 
 namespace o2
 {
--- a/Framework/Core/include/Framework/ArrowContext.h
+++ b/Framework/Core/include/Framework/ArrowContext.h
@@ -18,7 +18,7 @@
 #include <string>
 #include <vector>
 
-class FairMQMessage;
+#include <fairmq/FwdDecls.h>
 
 namespace o2::framework
 {
--- a/Framework/Core/include/Framework/CallbackService.h
+++ b/Framework/Core/include/Framework/CallbackService.h
@@ -15,7 +15,7 @@
 #include "Framework/ServiceHandle.h"
 #include <tuple>
 
-class FairMQRegionInfo;
+#include <fairmq/FwdDecls.h>
 
 namespace o2
 {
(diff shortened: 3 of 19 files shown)


ktf:
What is the point of exposing all the FairMQ types everywhere?

dennisklein:
The plan is to rename all the remaining types that are not yet in a namespace:
* `::FairMQ<Name>` will become `::fair::mq::<Name>`
* There will be type aliases to the old type name. Unfortunately, this breaks downstream forward decls as we have seen yesterday. This is one way to enable a transition. 

If you have a better way of doing that, please tell us. If you are ok by being broken from one version bump to the next, then you can close this PR.

ktf:
Well, I was thinking about some finer grained subdivision like:

* FairMQMessage, FairMQParts
* Device, Channel, Transport
* MemoryResources
* Rest

but then again, maybe it's an overkill.

ktf:
Fixing a merge conflict and triggering a rebuild to pick up the new version of FairMQ.

ktf:
I am holding on this, because otherwise everyone will have to move to the new FairMQ. If you want this to be used, you should protect it with #if __has_include()...

dennisklein:
> I am holding on this, because otherwise everyone will have to move to the new FairMQ.

Sure. That was the whole point, enabling downstream to decide when to have to adapt and/or have to bump the minimum required FairMQ version.

> If you want this to be used, you should protect it with #if __has_include()...

Consider this PR just as a friendly hint/help. I do not insist for this specific change to be merged (since I do not know or care about which range of FairMQ versions you intend to support). If you prefer different variant, please go ahead!

davidrohr:
I think there are not so many fwd declaration, so it should be ok.
I would not put some #if has include code, it just makes the includes more complicated, at some point people will have to bump fairmq. Perhaps we can wait few more days, but then I would merge this.

ktf:
Yes, but I want to avoid recompiling everything two days before my holidays... ;-)

davidrohr:
Well, it is not so heavy. It will recompile FairMQ/DDS/FairROOT/O2.
Anyway, do you think it is better we do this throughout your vacation? Or do we wait for 2 more weeks until you return?

ktf:
Can we wait until tomorrow?

davidrohr:
Sure, merge it at your convenience. In case there are issues, I am also available at least until the end of the week to check. But I have already used the new FMQ in the FST and it seem good.

PR #6660, opened 2021-07-16
TPCReconstruction: Fix typo

anerokhi:


shahor02 (approved):
The fix looks correct to me, @davidrohr @wiechula is `float pad=0.0` correspond to the lower edge of the pad or its center?

PR #6665, opened 2021-07-18
Jet code: Prevent expensive copying

anerokhi:


Analysis/Tasks/PWGJE/jetsubstructure.cxx
```diff
@@ -79,11 +79,11 @@ struct JetSubstructure {
     jetConstituents.clear();
     jetReclustered.clear();
     if (b_DoConstSub) {
-      for (const auto constituent : constituentsSub) {
+      for (const auto& constituent : constituentsSub) {
```
ktf, line 82: Is this coming from some actual measurement or just some automatic tool detecting that a & is missing?

I am asking because in principle `constituent` should be a rather thin proxy object, so it should not matter if you copy or take a reference.

anerokhi: > Is this coming from some actual measurement or just some automatic tool detecting that a & is missing?

Clang told me that one should not copy here. Clang doesn't seem tell this for every copying range-for loop.

> I am asking because in principle constituent should be a rather thin proxy object

When I hover over `sizeof(constituent)` in VSCode it shows me `(unsigned long)336UL`.

anerokhi:
```
Analysis/Tasks/PWGJE/jetsubstructure.cxx:83:43: error: static assertion failed
   83 |         static_assert(sizeof(constituent) != 336);
      |                       ~~~~~~~~~~~~~~~~~~~~^~~~~~
```

ktf: Ok, then the optimization obviously makes sense. @aalkin we should try to see if we manage to slim it down by precomputing some of the offsets...

anerokhi: The option is `-Wrange-loop-analysis` (part of `-Wall` in Clang). For trivially-copyable types the threshold is above 64 bytes.

aalkin: @ktf iterator copying is a long-standing issue - it is not just the size, though, it is re-binding of iterators and dynamic columns, and we cannot really have less pointers (each already a 64bit entity) there.

jgrosseo: I am in favour of adding `-Wrange-loop-analysis` to our clang arguments...

jgrosseo (approved):

Analysis/Tasks/PWGJE/jetsubstructure.cxx
ktf: I fully agree. The issue is that GCC will not have it and people in the past were complaining that "it works on the production platform". We should probably raise the issue again.

anerokhi: Looks like GCC11 has it https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94695 https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94695#c4

ktf: Yes, but we cannot yet move to GCC 11 for the medium term due to issues in both ROOT and our GPU code, IIUC.

PR #6668, opened 2021-07-19
Fix invalid iterator use UB

anerokhi:
`mCache.pop_front()` invalidates iterators to the first element. Here, the current loop iterator value is always¹ iterator to the first element, so, in loop exit condition comparison against past-the-end iterator, the loop iterator will be invalid. Using invalid iterators in comparison may have undefined behavior.

---
1) Before invalidation


mslupeck (approved):
Thanks @anerokhi !

@arvindkhuntia - fyi and please check that it is corrected in the next version of the digitizer

PR #6671, opened 2021-07-19
Do not ignore O2_BUILTIN_(UN)LIKELY parameter

anerokhi:


ktf:
Obviously currently wrong. Merging. Clearly we never built on anything other than clang or GCC.

anerokhi:
I've tried to apply cppcheck to O2 codebase. Turned out it doesn't identify itself neither as `__GNUC__` nor as `__clang__`

PR #6676, opened 2021-07-19
DPL Analysis: fix and update table builder

aalkin:
* Removed unnecessary include in AnalysisTask.h
* Made sure that empty tables would still have label set in metadata


ktf:
The issue on macOS seems legit, although I am not sure it's related to this PR.

aalkin:
> The issue on macOS seems legit, although I am not sure it's related to this PR.

Seems to be caused by recently merged #6656

PR #6677, opened 2021-07-19
TRD raw reader can dump to file

martenole:
It is an option which can be enabled to directly compare the tracklets/digits read from raw files with the ones that come from the simulation. Can help to debug MC->raw conversion.
The trigger records will only be added to trdtracklets.root and not to trddigits.root file, because we have different data descriptions for the trigger records associated with the digits (`TRGRDIG`) and the tracklets (`TRKTRGRD`), although the data type is the same (`o2::trd::TriggerRecord`). This is needed, because the trigger records are both an input and an output for the TRAP simulation. And since the input cannot be changed we have to create a new message and that needs a new description.


bazinski (approved):

bazinski:
This removes the need for the option to dump it to the very long log files when doing --trd-datareader-dataverbose.

tdietel (approved):

bazinski:
@shahor02 can this be merged?

PR #6681, opened 2021-07-19
Add table for ALICE3 Outer Barrel TOF

njacazio:
@ddobrigk This is what we could do, what do you think?


njacazio:
Of course we would add the compressed Nsigma tables once we merge another PR that have some dependencies in the O2

PR #6683, opened 2021-07-19
Fix compilation issue on macOS

ktf:
Requires explicit cast, apparently.


ktf:
Currently broken. Merging.

PR #6689, opened 2021-07-20
Use array delete

anerokhi:


anerokhi:
(But better fix would be to switch to smth like vectors)

sawenzel:
Trivial. Merging. But please note that there more leaks here (and finally a move to std::vector indeed preferred).

PR #6691, opened 2021-07-20
AO2D: extend writing of tracks and particles

njacazio:
- Introduce hack to allow using expression columns for eta and y
- Fix TOF information writing
- Minor addition to analysis code
- Add dummy class for event time determination from TOF (to be used in the TOF PID Response)

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -393,6 +393,28 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
       if (item != toStore.end()) {
         daughterL = item->second;
       }
+      float pX = (float)mcParticles[particle].Px();
+      float pY = (float)mcParticles[particle].Py();
+      float pZ = (float)mcParticles[particle].Pz();
+      float energy = (float)mcParticles[particle].GetEnergy();
+      // HACK to avoid FPE in expression columns. Affect only particles in the Beam Pipe.
+      // TO BE REMOVED asap
+      {
+        const float limit = 1e-4;
+        const float mom = TMath::Sqrt(pX * pX + pY * pY + pZ * pZ);
+        const float eta = 0.5f * TMath::Log((mom + pZ) / (mom - pZ));
+        if (TMath::Abs(eta) > 0.9) {
+          if (TMath::Abs((mom - pZ) / pZ) <= limit) {
+            pX = truncateFloatFraction(TMath::Sqrt((1.f + limit) * (1.f + limit) - 1.f) * pZ * 0.5, mMcParticleMom);
+            pY = truncateFloatFraction(TMath::Sqrt((1.f + limit) * (1.f + limit) - 1.f) * pZ * 0.5, mMcParticleMom);
+          }
+          if (TMath::Abs(energy - pZ) < limit) {
+            energy = truncateFloatFraction(pZ + limit, mMcParticleMom);
+          }
+        }
+      }
+      // End of HACK
+
       mcParticlesCursor(0,
                         mccolid,
                         mcParticles[particle].GetPdgCode(),
@@ -403,10 +425,10 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
                         daughter0,
                         daughterL,
                         truncateFloatFraction(weight, mMcParticleW),
-                        truncateFloatFraction((float)mcParticles[particle].Px(), mMcParticleMom),
-                        truncateFloatFraction((float)mcParticles[particle].Py(), mMcParticleMom),
-                        truncateFloatFraction((float)mcParticles[particle].Pz(), mMcParticleMom),
-                        truncateFloatFraction((float)mcParticles[particle].GetEnergy(), mMcParticleMom),
+                        truncateFloatFraction(pX, mMcParticleMom),
+                        truncateFloatFraction(pY, mMcParticleMom),
+                        truncateFloatFraction(pZ, mMcParticleMom),
+                        truncateFloatFraction(energy, mMcParticleMom),
                         truncateFloatFraction((float)mcParticles[particle].Vx(), mMcParticlePos),
                         truncateFloatFraction((float)mcParticles[particle].Vy(), mMcParticlePos),
                         truncateFloatFraction((float)mcParticles[particle].Vz(), mMcParticlePos),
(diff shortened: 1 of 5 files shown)


njacazio:
@noferini we can start from here

jgrosseo:
For me this is fine, but @sawenzel or @nburmaso should also comment

sawenzel (dismissed):
Fine for me (although I can't judge correctness)

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -249,16 +249,14 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
...
-            if (expSig > 0 && interactionTime > 0) {
-              float tof = expSig - interactionTime;
-              float expBeta = (intLen / tof / cSpeed);
-              expMom = mass * expBeta / std::sqrt(1.f - expBeta * expBeta);
+            if (interactionTime > 0) {
+              extraInfoHolder.tofSignal = static_cast<float>(tofMatch.getSignal() - interactionTime);
+            }
+            const float mass = o2::constants::physics::MassPionCharged; // default pid = pion
+            if (tofInt.getTOF(o2::track::PID::Pion) > 0.f) {
+              const float expBeta = (intLen / tofInt.getTOF(o2::track::PID::Pion) / cSpeed);
```
shahor02: one division less with `const float expBeta = (intLen / (tofInt.getTOF(o2::track::PID::Pion) * cSpeed))`

njacazio:
Thanks @sawenzel let's see if @nburmaso is also happy and then we merge

nburmaso:
Hi @njacazio. Looks fine for me too

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -395,6 +393,28 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
...
+        const float mom = TMath::Sqrt(pX * pX + pY * pY + pZ * pZ);
+        const float eta = 0.5f * TMath::Log((mom + pZ) / (mom - pZ));
+        if (TMath::Abs(eta) > 0.9) {
+          if (TMath::Abs((mom - pZ) / pZ) <= limit) {
+            pX = truncateFloatFraction(TMath::Sqrt((1.f + limit) * (1.f + limit) - 1.f) * pZ * 0.5, mMcParticleMom);
+            pY = truncateFloatFraction(TMath::Sqrt((1.f + limit) * (1.f + limit) - 1.f) * pZ * 0.5, mMcParticleMom);
+          }
+          if (TMath::Abs(energy - pZ) < limit) {
+            energy = truncateFloatFraction(pZ + limit, mMcParticleMom);
+          }
```
shahor02: could you elaborate what produces FPE? In any case, if you want to approximate the 
````
pT^2 = ( (1+limit)^2 - 1) pZ^2
````
then one should have 
````
pX = truncateFloatFraction(TMath::Sqrt((1.f + limit) * (1.f + limit) - 1.f) * pZ  / sqrt(2), mMcParticleMom);
...
````

File: Detectors/AOD/src/AODProducerWorkflowSpec.cxx [hunk 3]
```diff
@@ -798,12 +818,12 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
                      truncateFloatFraction(vertex.getChi2(), mCollisionPositionCov),
                      vertex.getNContributors(),
-                     truncateFloatFraction(tsTimeStamp, mCollisionPosition),
+                     truncateFloatFraction(relInteractionTime, mCollisionPosition),
                      truncateFloatFraction(timeStamp.getTimeStampError() * 1E3, mCollisionPositionCov),
                      collisionTimeMask);
     auto& trackRef = primVer2TRefs[collisionID];
     // passing interaction time in [ps]
-    fillTrackTablesPerCollision(collisionID, tsTimeStamp * 1E3, trackRef, primVerGIs, recoData,
+    fillTrackTablesPerCollision(collisionID, interactionTime * 1E3, trackRef, primVerGIs, recoData,
```
shahor02, line 826: Was not the `interactionTime` already converted to ns at https://github.com/AliceO2Group/AliceO2/blob/0cd3f782fe75bf1550b47820df7684f3f9e34ff8/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L793

shahor02 (changes requested):

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Hi @shahor02. In `fillTrackTablesPerCollision()` interaction time is in `[ps]` since TOF signal is in `[ps]` too (if I'm not mistaken)

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
njacazio: Hi Ruben, indeed, I missed that thanks!

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: right, missed that, then forget this comment. But the one about the FPE protection seems to be valid one

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
njacazio: Here the FPE is observed when computing rapidity/pseudorapidity for particles that are in the beampipe. This hack is needed to bypass a limitation that will be solved in the future when we will be able to handle conditions in Gandiva for the computation of the expression columns.

shahor02: OK, but if you want to make `(mom - pZ) / pZ) = limit`, this means `pT^2 = ( (1+limit)^2 - 1) pZ^2` and with `pX=pY=pT/sqrt(2)` this gives `pX=pY=truncateFloatFraction(TMath::Sqrt((1.f + limit) * (1.f + limit) - 1.f) * pZ  / sqrt(2), mMcParticleMom);` (note the factor 1/sqrt(2) instead of 0.5)

njacazio: that's true, thanks for spotting this!

njacazio: nonetheless, the difference should be small i.e. `*0.7` vs `*0.5`

njacazio:
@shahor02 thanks for the useful comments! Everything should be taken care of, could we go on?

shahor02 (approved):

PR #6700, opened 2021-07-21
PWGHF: Implementing efficiency weighting for DDbar correlation framework

fcolamar:
Adding the possibility of applying 1/[eff(D0) eff(D0bar)] weights to correlation pairs and 2D mass plots, and 1/eff(D0) weights to trigger mass plots).
Efficiency values can be passed via dedicated configurabiles, and have the same binnings of the mass plots.
Additionally, there are some Sumw2() calls for few histograms and a fix to some array sizes.


vkucera:
Closing because of O2Physics migration

PR #6701, opened 2021-07-21
DPL: use C++20 aggregate initialisers for services

ktf:
Simplifies too much to still be negleted, at least in backend code.

### DIFF
--- a/Framework/Core/src/CommonMessageBackends.cxx
+++ b/Framework/Core/src/CommonMessageBackends.cxx
@@ -33,6 +33,10 @@
 #include <boost/program_options/variables_map.hpp>
 #include <csignal>
 
+// This is to allow C++20 aggregate initialisation
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
+
 namespace o2::framework
 {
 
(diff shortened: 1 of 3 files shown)


Framework/Core/src/CommonMessageBackends.cxx
```diff
@@ -33,6 +33,10 @@
 #include <boost/program_options/variables_map.hpp>
 #include <csignal>
 
+// This is to allow C++20 aggregate initialisation
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
+
```
knopers8, line 39: Just my curiosity - using c++20 features with c++17 flag produces only a pedantic warning, not an error?

anerokhi: `-Wpedantic` for such thing seems to be an artifact from pre-C++20 times when compilers warned about C99 feature in C++
Clang 11 (seems to be used here) uses `-Wc++20-designator` warning, GCC12 (not yet used) — `-Wc++20-extensions`.

And in general GCC and Clang allow newer language features in older mode. Like, structured bindings and explicit lambda captures in `-std=c++11`.

knopers8: Interesting, thanks!

PR #6702, opened 2021-07-21
DPL: use old value for maxMemory

ktf:


jgrosseo:
@ktf do you know why the new one did not work?

ktf:
That's what I am currently investigating.

PR #6703, opened 2021-07-21
Alice 3: embedding FT3 inner disks in BP vacuum

mconcas:
@rpezzi this workaround will append the FT3 inner disks to the embeddingvacuum voulme inside the BP if it is enabled.
I gave a quick look to the output with one PbPb event, hits on those disks seem to be restored.
Let me know if something is not working as expected.
photo_2021-07-21_14-42-31
photo_2021-07-21_14-42-27


marcovanleeuwen (approved):
Looks fine; this will also be tested with simulation.

rpezzi (approved):
Thanks, @mconcas.

PR #6705, opened 2021-07-21
Adding DPL parser tool for raw page sequences provided by DPL input.

matthiasrichter:
The DPLRawPageSequencer can be used to find sequences of consecutive
raw pages with a similar property, e.g. the FEE ID. The actual check
and callback to handle a sequence can be provided by lambda functions.

The raw pages within one buffer/message are expected to have the full
length except the last page which can be smaller. The fixed spacing
allows fast scanning by binary search.

Corresponding unit test is emulating variable-length sequences of
raw pages.

Before merging:
- [x] enhance class documentation
- [x] make the unit tests independent of``DetectorsRaw`

### DIFF
--- a/Framework/Utils/CMakeLists.txt
+++ b/Framework/Utils/CMakeLists.txt
@@ -76,6 +76,7 @@ foreach(t
         RootTreeWriter
         RawParser
         DPLRawParser
+        DPLRawPageSequencer
         )
   o2_add_test(${t}
               SOURCES test/test_${t}.cxx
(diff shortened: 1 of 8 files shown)


matthiasrichter:
@davidrohr I have encapsulated the optimized raw page parsing for the GPU reco workflow in this tool, and I'm currently testing it with the workflow. Using the binary search, there is a remarkable speedup for only this parsing. How much this can contribute to a total speedup depends on how the data usually looks like, how many pages we have per message etc. My assumption so far is that all pages within a message have the 8kB length, only the last one can be shorter. At least the pages must have a fixed length to allow to binary search. If the pages can be shorter but still of fixed length I can implement a configurable page length.

davidrohr:
Thx, the assumption is correct, only the last page for a given feeid can be less then 8kb

Kind Regards
David Rohr

On July 21, 2021 5:32:08 PM GMT+02:00, Matthias Richter ***@***.***> wrote:
***@***.*** I have encapsulated the optimized raw page parsing for the

davidrohr:
@matthiasrichter : will you also integrate this in the GPU workflow?

PR #6707, opened 2021-07-21
Rework filling process for MC track labels

nburmaso:
* `getTrackMCLabel()` method of RecoContainer is now used to obtain MC labels instead of full spans
* All sources are now used (in contrast to spans of only MFT, ITS and TPC labels), as well as contributing track labels for global tracks
* If label is not set, `-1` is stored into a table -- for analysis compatibility

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -277,43 +277,50 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
 }
 
 template <typename MCParticlesCursorType>
-void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader& mcReader, const MCParticlesCursorType& mcParticlesCursor,
-                                                  gsl::span<const o2::MCCompLabel>& mcTruthITS,
-                                                  gsl::span<const o2::MCCompLabel>& mcTruthMFT,
-                                                  gsl::span<const o2::MCCompLabel>& mcTruthTPC,
-                                                  TripletsMap_t& toStore,
+void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader& mcReader,
+                                                  const MCParticlesCursorType& mcParticlesCursor,
+                                                  gsl::span<const o2::dataformats::VtxTrackRef>& primVer2TRefs,
+                                                  gsl::span<const GIndex>& GIndices,
+                                                  o2::globaltracking::RecoContainer& data,
                                                   std::vector<std::pair<int, int>> const& mccolid_to_eventandsource)
 {
   // mark reconstructed MC particles to store them into the table
-  for (int i = 0; i < mcTruthITS.size(); i++) {
-    auto& mcTruth = mcTruthITS[i];
-    if (!mcTruth.isValid()) {
-      continue;
+  for (auto& trackRef : primVer2TRefs) {
+    for (int src = GIndex::NSources; src--;) {
+      int start = trackRef.getFirstEntryOfSource(src);
+      int end = start + trackRef.getEntriesOfSource(src);
+      for (int ti = start; ti < end; ti++) {
+        auto& trackIndex = GIndices[ti];
+        if (GIndex::includesSource(src, mInputSources)) {
+          auto mcTruth = data.getTrackMCLabel(trackIndex);
+          if (!mcTruth.isValid()) {
+            continue;
+          }
+          int source = mcTruth.getSourceID();
+          int event = mcTruth.getEventID();
+          int particle = mcTruth.getTrackID();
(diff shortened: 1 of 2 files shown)


nburmaso:
Hi @sawenzel. This should fix inconsistency in sizes between tracks and track labels.

sawenzel (approved):

sawenzel:
Thanks a lot !

sawenzel:
Compiled and tested locally. I'd like to have this in the nightlies. Merging.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -277,43 +277,50 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
...
+          if (!mcTruth.isValid()) {
+            continue;
+          }
+          int source = mcTruth.getSourceID();
+          int event = mcTruth.getEventID();
+          int particle = mcTruth.getTrackID();
+          mToStore[Triplet_t(source, event, particle)] = 1;
+          // treating contributors of global tracks
+          auto contributorsGID = data.getSingleDetectorRefs(trackIndex);
+          if (contributorsGID[GIndex::Source::ITS].isIndexSet() && contributorsGID[GIndex::Source::TPC].isIndexSet()) {
```
shahor02, line 305: Why `&&` here?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -874,104 +873,68 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+            mcMFTTrackLabelCursor(0,
+                                  labelHolder.labelID,
+                                  labelHolder.mftLabelMask);
+          } else {
+            if (mcTruth.isValid()) { // if not set, -1 will be stored
+              labelHolder.labelID = mToStore.at(Triplet_t(mcTruth.getSourceID(), mcTruth.getEventID(), mcTruth.getTrackID()));
+            }
+            // treating possible mismatches for global tracks
+            auto contributorsGID = recoData.getSingleDetectorRefs(trackIndex);
+            if (contributorsGID[GIndex::Source::ITS].isIndexSet() && contributorsGID[GIndex::Source::TPC].isIndexSet()) {
```
shahor02, line 907: Also here the reason of the `&&` is not clear. For the record: the MCCompLabel of standalone detector track (ITS, TPC, TRD tracklets collection etc) is set to refer to the dominant particle represented in the contributors (e.g. clusters) of the track and `setFake()` is set if the fraction of contributors w/o this dominant particle is above threshold (normally 10% for TPC, >0 for other detectors).
Then, for the compound tracks labels the rule is: it refers on the particle mentioned in the TPC track label and the isFake is set if the track/cluster in other detector matched to TPC has not this TPC label. I.e. the fakeness of individual tracks is ignored on this level.

So, if you have e.g. ITS-TPC-TOF match contributed by
ITStrack:  {mcIDA, non-fake}
TPCtrack: {mcIDB, fake}
TOCclus: {mcIDB}
then the label of ITS-TPC-TOF will be {mcIDB, non-fake}, it will refer to  track ITS-TPC with label {mcIDB, fake} which in turn will refer to ITS track  {mcIDA, non-fake} and TPCtrack  {mcIDB, non-fake}.

shahor02 (commented):
@nburmaso  sorry for reviewing after the merging, please see a few comments below

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Hi @shahor02. I added this additional check to represent possible ITS and TPC label mismatches for global tracks that have both ITS and TPC labels -- so I put `&&`. Do I understand correctly that such check is not needed, since `isFake()` can be used to indicate any mismatches?

nburmaso: Hi @shahor02. I added this additional check to represent possible ITS and TPC label mismatches in label masks for global tracks that have both ITS and TPC labels -- so I put &&.
https://github.com/nburmaso/AliceO2/blob/ab000dee139c59213711f14aaa15541942afd1d7/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L917-L920
Do I understand correctly that such check is not needed, since isFake() can be used to indicate any mismatches?

shahor02: Yes, for compound tracks (TPC always there) the label MC particle ID points to that of the TPC track label and isFake is set in case there is a mismatch between the MC particle IDs of the 2 contributors (as on schema below) of this global track
````
{ITS-TPC-TRD-TOF}
       ^
{ITS-TPC-TRD}, {TOF}
       ^
{ITS-TPC}, {TRD}
       ^
{ITS}, {TPC}
````

````
{ITS-TPC-TOF}
       ^
{ITS-TPC}, {TOF}
````

PR #6708, opened 2021-07-21
FIT: update from FEE, new field in raw data

afurs:
1) New field "OutputsAreBlocked" for TCM
2) Field "DataIsValid" bit position is now fixed


mslupeck (approved):
OK for me

PR #6709, opened 2021-07-21
DPL: introduce driverStart callback for services

ktf:


ktf:
@aalkin this should allow you to have access to ConfigurationInterface in the startup. I am not 100% sure of where we should put the callbacks. Maybe we need one for MATERIALSE_WORKFLOW and use that for Configuration.

ktf:
Now with all the files...

ktf:
@TimoWilken  I think there is some issue with o2-cs8: `build/O2/o2-cs8 Expected — Waiting for status to be reported`

ktf:
Ping @TimoWilken, still pending after 4 days...

PR #6711, opened 2021-07-22
[CPV] Fix RawWriter & proper CCDB handling in Digitizer

sevdokim:
A bug was found in RawWriter: it was repeatedly writing digits to raw file from different events in same orbit. Now it's fixed.
Also access to CCDB is added to Digitizer. Usage: 
`o2-sim-digitizer-workflow --configKeyValues 'CPVSimParams.mCCDBPath=ccdb-test.cern.ch:8080'`

### DIFF
--- a/Detectors/CPV/simulation/src/Digitizer.cxx
+++ b/Detectors/CPV/simulation/src/Digitizer.cxx
@@ -12,7 +12,8 @@
 #include "CPVSimulation/Digitizer.h"
 #include "SimulationDataFormat/MCCompLabel.h"
 #include "CPVBase/CPVSimParams.h"
-#include "CCDB/CcdbApi.h"
+#include "CCDB/CCDBTimeStampUtils.h"
+#include "CCDB/BasicCCDBManager.h"
 
 #include <TRandom.h>
 #include "FairLogger.h" // for LOG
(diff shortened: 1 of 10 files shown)


kharlov (dismissed):

Detectors/CPV/simulation/src/Digitizer.cxx
```diff
@@ -27,53 +28,48 @@ using namespace o2::cpv;
...
-      LOG(INFO) << "[CPVDigitizer] can not get calibration object from ccdb yet. Using default";
-      mCalibParams.reset(new CalibParams(1)); // test default calibration
-      //      o2::ccdb::CcdbApi ccdb;
-      //      std::map<std::string, std::string> metadata; // do we want to store any meta data?
-      //      ccdb.init("http://ccdb-test.cern.ch:8080");  // or http://localhost:8080 for a local installation
-      //      mCalibParams = ccdb.retrieveFromTFileAny<o2::cpv::CalibParams>("CPV/Calib", metadata, mEventTime);
-      //      if (!mCalibParams) {
-      //        LOG(FATAL) << "[CPVDigitizer] can not get calibration object from ccdb";
-      //      }
+  LOG(INFO) << "CPVDigitizer::init() : CCDB Url = " << o2::cpv::CPVSimParams::Instance().mCCDBPath.data();
```
sawenzel, line 31: I am wondering why each detector should have this path configurable (separately)? In principle there is already a global configurable `DigiParams` that can take this path.
I understand that several detectors are doing it like here, but eventually we should converge on a single source. (Not asking this necessarily now. I'll open a ticket to organize a campaign for all detectors.)

sevdokim: Dear @sawenzel, 

I am also wondering same question. Next PR I'll change it to DigiParams. Also if I do `cat o2sim_configuration.ini | grep -i ccdb` to some simulation config, I see 4 times ccdb path configured 4 times:
```
ccdb=http://ccdb-test.cern.ch:8080
mCCDB=http://ccdb-test.cern.ch:8080
mCCDBPath=localtest
mCCDBPath=ccdb-test.cern.ch:8080
```
 Just to inform you.

sevdokim:
Hello! It seems that PR's checks just stucked so I decided to add more changes to the PR and trigger checks again by new commit. Added are checks for valid geometry addresses. It appears that raw data which is coming from detector can contain non-existing addresses of channels. Now RawDecoder is checking channels for validity before creating digits. 
Also using of std::unique_ptr to handle calibration objects is removed from RawToDigitsConverterSpec.

kharlov (approved):

sevdokim:
Dear @shahor02 
It seems that some QC functional test has failed during build/O2/o2 check. I don't see a reason related to my changes. Can you help me to understand what happened?

shahor02:
@sevdokim this qc test timeout happens quite often, not related to your PR, merging

PR #6713, opened 2021-07-22
[EMCAL-701] Add protection for missing / empty timeframes

mfasDa:
Implementation according to AliceO2Group/AliceO2#5786


shahor02 (approved):
Thanks! Looks correct.

PR #6718, opened 2021-07-22
O2HitMerger: Fix memory leaks

sawenzel:
* Fix a few memory leaks by using unique_ptr;
* also change a few vector->at to operator[] for
  faster access

### DIFF
--- a/run/O2HitMerger.h
+++ b/run/O2HitMerger.h
@@ -445,7 +445,7 @@ class O2HitMerger : public FairMQDevice
   void reorderAndMergeMCTRacks(TTree& origin, TTree& target, const std::vector<int>& nprimaries, const std::vector<int>& nsubevents)
   {
     std::vector<MCTrack>* incomingdata = nullptr;
-    auto targetdata = new std::vector<MCTrack>;
+    auto targetdata = std::make_unique<std::vector<MCTrack>>();
     auto originbr = origin.GetBranch("MCTrack");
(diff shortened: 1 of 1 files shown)


sawenzel:
Thanks to @anerokhi for bringing this up.

run/O2HitMerger.h
```diff
@@ -530,16 +530,16 @@ class O2HitMerger : public FairMQDevice
...
     auto originbr = origin.GetBranch(brname.c_str());
     originbr->SetAddress(&incomingdata);
     const auto entries = origin.GetEntries();
 
     if (entries == 1) {
       // nothing to do in case there is only one entry
       originbr->GetEntry(0);
-      targetdata = incomingdata;
     } else {
+      targetdata = std::move(std::make_unique<T>());
```
anerokhi, line 542:
```suggestion
      targetdata = std::make_unique<T>();
```

run/O2HitMerger.h
```diff
@@ -530,16 +530,16 @@ class O2HitMerger : public FairMQDevice
     // This method is called by O2HitMerger::mergeAndFlushData(int)
     //
     T* incomingdata = nullptr;
-    auto targetdata = new T;
+    std::unique_ptr<T> targetdata(nullptr);
```
anerokhi, line 533:
```suggestion
    std::unique_ptr<T> targetdata;
```

PR #6719, opened 2021-07-22
Allow to set continueMode for GeneratorFromO2Kine

preghenella:


preghenella:
hi @sawenzel 
this should be trivial.
I added a setter which is needed for running this generator from ouside of the o2sim environment and at the same time be able to set/unset flags.

sawenzel (approved):

PR #6722, opened 2021-07-22
Optimizing GPUWorkflowSpec

matthiasrichter:
Using DPLRawPageSequencer utility which does the sequence scan based on
binary search instead of the current forward-match algorithm.

### DIFF
--- a/GPU/Workflow/src/GPUWorkflowSpec.cxx
+++ b/GPU/Workflow/src/GPUWorkflowSpec.cxx
@@ -38,6 +38,7 @@
 #include "TPCFastTransform.h"
 #include "TPCdEdxCalibrationSplines.h"
 #include "DPLUtils/DPLRawParser.h"
+#include "DPLUtils/DPLRawPageSequencer.h"
 #include "DetectorsBase/MatLayerCylSet.h"
 #include "DetectorsBase/Propagator.h"
 #include "DetectorsBase/GeometryManager.h"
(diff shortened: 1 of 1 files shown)


matthiasrichter:
@davidrohr I did a test on basis of the full system test. It seems that the tracking is not fully deterministic. There is a slight difference of +/- 10 tracks when I run the tracking multiple times with the dev branch version. Number of found clusters is the same. The same I can see with this PR.
```
run_new2.log:[2381455:tpc-tracker]: [21:35:37][INFO] Event has 8776 8kb TPC ZS pages, 22890646 digits
run_new2.log:[2381455:tpc-tracker]: [21:35:42][INFO] Event has 3365498 TPC Clusters, 0 TRD Tracklets
run_new.log:[2380866:tpc-tracker]: [21:35:06][INFO] Event has 8776 8kb TPC ZS pages, 22890646 digits
run_new.log:[2380866:tpc-tracker]: [21:35:11][INFO] Event has 3365498 TPC Clusters, 0 TRD Tracklets
run_ref2.log:[2384440:tpc-tracker]: [21:43:33][INFO] Event has 8776 8kb TPC ZS pages, 22890646 digits
run_ref2.log:[2384440:tpc-tracker]: [21:43:38][INFO] Event has 3365498 TPC Clusters, 0 TRD Tracklets
run_ref.log:[2383887:tpc-tracker]: [21:43:09][INFO] Event has 8776 8kb TPC ZS pages, 22890646 digits
run_ref.log:[2383887:tpc-tracker]: [21:43:14][INFO] Event has 3365498 TPC Clusters, 0 TRD Tracklets

run_new2.log:[2381455:tpc-tracker]: [21:35:50][INFO] found 31021 track(s)
run_new2.log:[2381458:tpc-track-writer]: [21:35:50][INFO] writing 31021 track(s)
run_new.log:[2380866:tpc-tracker]: [21:35:18][INFO] found 31030 track(s)
run_new.log:[2380869:tpc-track-writer]: [21:35:18][INFO] writing 31030 track(s)
run_ref2.log:[2384440:tpc-tracker]: [21:43:46][INFO] found 31025 track(s)
run_ref2.log:[2384443:tpc-track-writer]: [21:43:46][INFO] writing 31025 track(s)
run_ref.log:[2383887:tpc-tracker]: [21:43:22][INFO] found 31021 track(s)
run_ref.log:[2383890:tpc-track-writer]: [21:43:22][INFO] writing 31021 track(s)
```
where *ref* means reference dev branch, *new* is for this PR

Looking at some plots for parameters of the produced tracks, the result with the PR is identical within this uncertainty. Please confirm.

TPCTracks-time0_dev

TPCTracks-time0

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -410,51 +411,19 @@ DataProcessorSpec getGPURecoWorkflowSpec(gpuworkflow::CompletionPolicyData* poli
...
-            }
-            count++;
-          }
-          if (count) {
-            tpcZSmetaPointers[rawcru / 10][(rawcru % 10) * 2 + rawendpoint].emplace_back(ptr);
-            tpcZSmetaSizes[rawcru / 10][(rawcru % 10) * 2 + rawendpoint].emplace_back(count);
-          }
-        }
+        auto isSameRdh = [](const char* left, const char* right) -> bool {
+          return o2::raw::RDHUtils::getCRUID(left) == o2::raw::RDHUtils::getCRUID(right) && o2::raw::RDHUtils::getEndPointID(left) == o2::raw::RDHUtils::getEndPointID(right) && o2::raw::RDHUtils::getFEEID(left) == o2::raw::RDHUtils::getFEEID(right);
```
anerokhi: I'm sorry, there is no line width limit?

matthiasrichter: I think its fine, and would stick to what clang format detects to be fixed. But maybe its good enough to check only the FeeId, what do you think @davidrohr ?

matthiasrichter: Have changed this to only check the Fee Id, this also makes the line shorter

davidrohr:
Hi @matthiasrichter : yes, the tracking is not fully deterministic, the clusterizing is. If you have the same number of clusters, and slightly different numbers of tracks, that is ok. You can use `--configKeyValues `GPU_proc.ompKernels=0` to disable the multithreading, which will make also the tracking deterministic.

Thinking about the requirements we discussed last time for the 8kb size. Actually, the last page per feeid a per HBF can be != 8 kb, since the CRU must not send data that crosses a HBF in the same package, thus the last page of a HBF will usually not be full. Is this supported by your parser?

Finally, looking at the code, I think
```
rawcru = o2::raw::RDHUtils::getCRUID(*rdh);
rawendpoint = o2::raw::RDHUtils::getEndPointID(*rdh);
```
is actually not correct, it works for simulated data coincidentally, but there are 2 different numbering schemes, one for the physical CRUs used for the FLPs, and one for the logical TPC CRUs.
Instead, we should use
```
cruID = rdh_utils::getCRU(feeId);
endPoint = rdh_utils::getEndPoint(feeId);
```
from `Detectors/TPC/reconstruction/include/TPCReconstruction/RawReaderCRU.h`. Could you change that as well?

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -410,51 +411,19 @@ DataProcessorSpec getGPURecoWorkflowSpec(gpuworkflow::CompletionPolicyData* poli
...
-            const unsigned char* current = it.raw();
-            const RAWDataHeader* rdh = (const RAWDataHeader*)current;
-            if (current == nullptr || it.size() == 0 || (current - ptr) % TPCZSHDR::TPC_ZS_PAGE_SIZE || o2::raw::RDHUtils::getFEEID(*rdh) != lastFEE) {
-              if (count) {
-                tpcZSmetaPointers[rawcru / 10][(rawcru % 10) * 2 + rawendpoint].emplace_back(ptr);
-                tpcZSmetaSizes[rawcru / 10][(rawcru % 10) * 2 + rawendpoint].emplace_back(count);
-              }
-              count = 0;
-              lastFEE = o2::raw::RDHUtils::getFEEID(*rdh);
-              rawcru = o2::raw::RDHUtils::getCRUID(*rdh);
```
matthiasrichter, line 438: @davidrohr, you mean, that these lines have been outdated for a while already? How did it work then? I will change the updated code to use `rdh_utils::get*` as you suggested

davidrohr: We don't have real TPC data in the final ZS format, and in the MC data the rawcru in the RDH is set to the logical TPC CRU, so there it was working (more or less by chance).

davidrohr:
ok, thx. Just to confirm before merging:


> Thinking about the requirements we discussed last time for the 8kb size. Actually, the last page per feeid a per HBF can be != 8 kb, since the CRU must not send data that crosses a HBF in the same package, thus the last page of a HBF will usually not be full. Is this supported by your parser?

This will not pose a problem?

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -410,51 +411,19 @@ DataProcessorSpec getGPURecoWorkflowSpec(gpuworkflow::CompletionPolicyData* poli
...
-            }
-            count++;
-          }
-          if (count) {
-            tpcZSmetaPointers[rawcru / 10][(rawcru % 10) * 2 + rawendpoint].emplace_back(ptr);
-            tpcZSmetaSizes[rawcru / 10][(rawcru % 10) * 2 + rawendpoint].emplace_back(count);
-          }
-        }
+        auto isSameRdh = [](const char* left, const char* right) -> bool {
+          return o2::raw::RDHUtils::getFEEID(left) == o2::raw::RDHUtils::getFEEID(right);
```
matthiasrichter, line 415: should we replace `o2::raw::RDHUtils` also here? but it looks like there is no corresponding function in `rdh_utils` taking the raw pointer. Did not have the time to check more detailed.

GPU/Workflow/src/GPUWorkflowSpec.cxx
matthiasrichter: ok

GPU/Workflow/src/GPUWorkflowSpec.cxx
davidrohr: no, for the feeid it is ok. you are supposed to get the feeid via o2::raw::RDHUtils and then use rdh_utils to get the cru id from the feeid.
The confusing thing is that you can also get a CRU IT using RDHUtils from the RDH, which are not necessarily the same (so far in MC data they are the same, that is why it worked).

davidrohr:
Works for me, thx, failing CI has no log, seems an unrelated problem, merging

PR #6724, opened 2021-07-22
Refurbished TPC-ITS afterburner

shahor02:
Don't merge yet since a few workflows which have open PRs will need to be modified to avoid crashing.


shahor02:
Merging, the afterburner will be activated (with current MatchTPCITSParams) if `o2-tpcits-match-workflow` is run with option `--use-ft0` but should not be done in conjunction with AOD producer until the latter is adapted to process the output of the afterburner.

PR #6725, opened 2021-07-23
[EMCAL-703] Add optional settable subspecification for raw to digits output

mfasDa:
- Add option to define subspecification for the output
  specs of the raw to cell converter (default: 0) and
  propagate it into the class in order to send to the
  proper subspecification in the snapshot function
- Steer subspecification via the workflow option
  subspecification

Needed in order to run the workflow on the two FLPs
in parallel. In case of running offline or on the EPN the
default subspecification (0) should be used.

### DIFF
--- a/Detectors/EMCAL/workflow/include/EMCALWorkflow/RawToCellConverterSpec.h
+++ b/Detectors/EMCAL/workflow/include/EMCALWorkflow/RawToCellConverterSpec.h
@@ -15,6 +15,7 @@
 #include "Framework/Task.h"
 #include "DataFormatsEMCAL/Cell.h"
 #include "DataFormatsEMCAL/TriggerRecord.h"
+#include "Headers/DataHeader.h"
 #include "EMCALBase/Geometry.h"
 #include "EMCALBase/Mapper.h"
 #include "EMCALReconstruction/CaloRawFitter.h"
@@ -37,8 +38,8 @@ class RawToCellConverterSpec : public framework::Task
 {
  public:
   /// \brief Constructor
-  /// \param propagateMC If true the MCTruthContainer is propagated to the output
-  RawToCellConverterSpec() : framework::Task(){};
+  /// \param subspecification Output subspecification for parallel running on multiple nodes
+  RawToCellConverterSpec(int subspecification) : framework::Task(), mSubspecification(subspecification){};
 
   /// \brief Destructor
   ~RawToCellConverterSpec() override;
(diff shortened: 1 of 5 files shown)


mfasDa:
@shahor02 In case of running on the EPN we need a merger component which combines the cells from the two FLPs into a single cell container. This is in principle straight forward to write, however I have two questions:
- If we subscribe to the two cell container and trigger record containers with the different subspecifications, can we assume that we get in the run function the same entire timeframe from both subspecifications?
- For the numbering scheme I would propose to use the FLP ID (146 and 147) as subspecification, can the merger then create an output spec with the same name (i.e. EMC/CELLS and EMC/CELLSTRGR) as the raw to cell converter but with the (so far) unused subspecification 0?

Using the same output channel with subspecification 0 would make life easier for already existing components (QC, CTF coder, calibration)

shahor02:
>     * If we subscribe to the two cell container and trigger record containers with the different subspecifications, can we assume that we get in the run function the same entire timeframe from both subspecifications?

Yes, you will get 2 messages for each type instead of 1
> 
>     * For the numbering scheme I would propose to use the FLP ID (146 and 147) as subspecification, can the merger then create an output spec with the same name (i.e. EMC/CELLS and EMC/CELLSTRGR) as the raw to cell converter but with the (so far) unused subspecification 0?

Afaik, the FLP IDs may change, you better do this configurable (or, if you have some mapping, use as a subspec the FEEID of the 1st link etc.). In any case, on the merger side you don't need to specify the subspec explicitly, you can subscribe via `ConcreteDataTypeMatche{"EMC","CELLS"} and merge whatever comes in by iterating over the inputs.

But, in principle, you could avoid merger at all: the consumer of cells could simple iterate over inputs with different subspecs.
E.g. for the entropy-encoder, which is the principal consumer of cells, you would need to merge on the fly the trigger records only (since the encoding is done over the data sorted in time) and modify the iterators of CTFHelper to take 2 cell spans instead of 1.

> 
> 
> Using the same output channel with subspecification 0 would make life easier for already existing components (QC, CTF coder, calibration)

mfasDa:
I have made the subspecification settable (as a ConfigurableParam of the workflow). I was thinking that this could be steered on the FLP via the control workflow storing a unique ID in the consul on each of the two FLPs.

Concering the mergers: In particular for the QC we have cases where we need the full event, and at online level, i.e. for plots like number of cells / event. We therefore need the full event already online. This can be done via a merger device or simply a merger class that would run in each processor which needs merged cells (in my opinion not too efficient to duplicate work). Otherwise we would move the merging device from the synchronous to the asynchronous reconstruction, where in the later case the merging would be implemented within the AO2D producer workflow. In any case the work has to be done and then I would rather implement it in a dedicated DPL processor.

shahor02:
@mfasDa merging on every device might be expensive (if >1 device is foreseen) as well as the merging of cells output. 
From other side, merging of trigger records should be cheap, since this is a fixed record per trigger. What you can do is a compact class which does the on-the-fly merging of triggers, which each merged trigger having 2 range references of on the data of each FLP. Anyway, this is up to you.

mfasDa:
@shahor02 I think I understand where this is going. We still keep the cell container as separate objects but combine the trigger records in a way that they have multiple range references, one for each subtimeframe, in this case the component still sees the full event but needs to iterate over different (2) spans. That should work as well. We will keep it in consideration.

mfasDa:
@shahor02 Any objections preventing the PR from being merged? For the discussion how to propagate the subspecification via AliECS I started a topic in alice-talk https://alice-talk.web.cern.ch/t/flp-specific-subspecifications/1035

PR #6726, opened 2021-07-23
DPL RootTreeWriter: always amend branch name(s) if the callback is configured

matthiasrichter:
So far, branch names have only been amended for branch definitions handling more
than one branch. The more logical behavior is to amend whenever a callback is
configured.


matthiasrichter:
@sawenzel this should solve your problem with the tree names when running workflows for only a single TPC sector. You are right, the default behavior should be to always amend the branch name if the callback is configured, which is the case in the TPC reco workflow setup. I hope this does not have any other implication to other test/work routines of the TPC people ( @wiechula )

sawenzel (approved):
Thank you!

sawenzel (dismissed):

wiechula (approved):
My best guess is that it is fine.

PR #6729, opened 2021-07-23
TRD add Krypton cluster finder

martenole:
Based on the macro of @aschmah but fully revised. An executable `o2-trd-kr-clusterer` is added which takes as input TRD digits and trigger records and provides Krypton clusters.
Two concerns:
- The ROOT fit produces some log messages which I did not manage yet to turn off. This needs to be fixed before it can be used in production
- The TRDDigitReaderSpec provides the trigger records with the description `TRGRDIG` while the raw reader uses `TRKTRGRD`. A switch for both descriptions needs to be added depending on whether or not ROOT file input enabled
Adding this as draft PR for now.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/KrCluster.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/KrCluster.h
@@ -1,0 +1,92 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file KrCluster.h
+/// \brief A cluster formed from digits during TRD Krypton calibration
+
+#ifndef O2_TRD_KRCLUSTER_H
+#define O2_TRD_KRCLUSTER_H
+
+#include "Rtypes.h"
+#include "DataFormatsTRD/Constants.h"
+#include "DataFormatsTRD/HelperMethods.h"
+
+namespace o2
+{
+namespace trd
+{
+
+class KrCluster
+{
+ public:
+  KrCluster() = default;
+  KrCluster(const KrCluster&) = default;
+  ~KrCluster() = default;
+
+  void setGlobalPadID(int det, int row, int col);
+  void setAdcData(int adcSum, int adcRms, int adcMaxA, int adcMaxB, int adcEoT, int adcIntegral, int adcSumTrunc);
+  void setTimeData(int timeMaxA, int timeMaxB, int timeRms);
+  void setClusterSizeData(int rowSize, int colSize, int timeSize, int nAdcs);
+
+  // identify global pad number (row and column with the maximum ADC value contained in the cluster)
+  int getDetector() const { return mDet; }
+  int getSector() const { return HelperMethods::getSector(mDet); }
+  int getStack() const { return HelperMethods::getStack(mDet); }
+  int getLayer() const { return HelperMethods::getLayer(mDet); }
+  int getRow() const { return mRow; }
+  int getColumn() const { return mCol; }
+
+  // ADC related members
+  int getAdcSum() const { return mAdcSum; }
+  int getAdcRms() const { return mAdcRms; }
+  int getAdcMaxA() const { return mAdcMaxA; }
+  int getAdcMaxB() const { return mAdcMaxB; }
+  int getAdcSumEoverT() const { return mAdcSumEoverT; }
+  int getAdcIntegral() const { return mAdcIntegral; }
+  int getAdcSumTruncated() const { return mAdcSumTruncated; }
+
+  // cluster size
+  int getClSizeRow() const { return mDeltaRow; }
+  int getClSizeCol() const { return mDeltaCol; }
+  int getClSizeTime() const { return mDeltaTime; }
+  int getClSize() const { return mClusterSize; }
+
+  // time bin related members
+  int getTimeMaxA() const { return mTimeMaxA; }
+  int getTimeMaxB() const { return mTimeMaxB; }
+  int getTimeRms() const { return mTimeRms; }
+
+ private:
+  uint16_t mDet;             ///< chamber number [0..539]
+  uint16_t mAdcSum;          ///< sum of all ADCs contributing to this cluster (baseline subtracted)
+  uint16_t mAdcRms;          ///< RMS of the ADCs constributing to this cluster
+  uint16_t mAdcMaxA;         ///< sum of the ADCs of the first maximum
+  uint16_t mAdcMaxB;         ///< sum of the ADCs of the second maximum (the contribution from the first maximum is subtracted here)
+  uint16_t mAdcSumEoverT;    ///< same as mAdcSum, but ADCs below KrClusterFinder::mMinAdcClEoverT are ignored
+  uint16_t mAdcIntegral;     ///< integral of the Landau fit for all time bins
+  uint16_t mAdcSumTruncated; ///< same as mAdcSum, but only ADCs close to the RMS are counted
+  uint8_t mRow;              ///< pad row number contributing max ADC value contained in this cluster [0..15]
+  uint8_t mCol;              ///< pad column number contributing max ADC value contained in this cluster [0..143]
+  uint8_t mDeltaRow;         ///< cluster size in row direction
+  uint8_t mDeltaCol;         ///< cluster size in column direction
+  uint8_t mDeltaTime;        ///< cluster size in time bin direction
+  uint8_t mClusterSize;      ///< number of ADC values contributing to the cluster (ADC needs to be above KrClusterFinder::mMinAdcClContrib)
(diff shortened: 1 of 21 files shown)


bazinski:
@martenole given that i am busy fiddling with the datareader, I can change the outgoing triggers, to any name that is convenient. When reading simulated input there are 2 triggerrecords (digits file and tracklets file) but going out (simulated and raw) is a singular triggerrecord.

martenole:
Switch works fine, tested with `o2-raw-file-reader-workflow --input-conf raw/TRD/TRDraw.cfg | o2-trd-datareader | o2-trd-kr-clusterer -b --disable-root-input`.
To be checked whether it is better to write not one Kr cluster trigger record per TRD trigger, but only for a range of clusters with a minimum size. Or maybe one could simply write one trigger record per TF or similar.

martenole:
> @martenole given that i am busy fiddling with the datareader, I can change the outgoing triggers, to any name that is convenient. When reading simulated input there are 2 triggerrecords (digits file and tracklets file) but going out (simulated and raw) is a singular triggerrecord.

@bazinski thanks, but I think we need to keep to different descriptions for the digit and tracklet trigger records (although they have the same type), because of the TRAP sim. See my comment here: https://github.com/AliceO2Group/AliceO2/pull/6677#issue-692509491

martenole:
Hi @aschmah could you please have a look when you get a chance?
Especially at the `KrCluster` class and the data member names. I hope they are sufficiently descriptive. The size per cluster increases a bit from 20 to 25 bytes, but the access to data members is much simpler and the allowed ranges large enough so that no clusters are rejected.
The cluster finder itself is much faster and much less memory consuming than the original macro. For a 580MB digits file with 8k triggers I need about 30s computing time to find around 75k Kr clusters.

aschmah:
@martenole and I just discussed a few points. The code looks good to me. The following minor changes will be applied:
- line 307: // OS: shouldn't we instead check maxAdcA <= maxAdcB?  -> will be changed to "<="
- 365: maxTbB = 63; // OS: why 63?  -> is OK as it is. Larger than any maximum time bin we can have
- 320: if (!mFitter.FitFCN()) {...  -> the continue will go, instead the integral in line 330 will be set to 0 if the fit fails

Detectors/TRD/workflow/io/src/TRDDigitReaderSpec.cxx
```diff
@@ -87,7 +87,7 @@ void TRDDigitReaderSpec::run(ProcessingContext& pc)
   pc.services().get<ControlService>().readyToQuit(QuitRequest::Me);
 }
 
-DataProcessorSpec getTRDDigitReaderSpec(int channels, bool useMC)
+DataProcessorSpec getTRDDigitReaderSpec(bool useMC)
```
tdietel, line 84: I wonder what the original `channels` argument was for?

DataFormats/Detectors/TRD/include/DataFormatsTRD/KrCluster.h
```diff
@@ -0,0 +1,92 @@
...
+  uint16_t mAdcRms;          ///< RMS of the ADCs constributing to this cluster
+  uint16_t mAdcMaxA;         ///< sum of the ADCs of the first maximum
+  uint16_t mAdcMaxB;         ///< sum of the ADCs of the second maximum (the contribution from the first maximum is subtracted here)
+  uint16_t mAdcSumEoverT;    ///< same as mAdcSum, but ADCs below KrClusterFinder::mMinAdcClEoverT are ignored
+  uint16_t mAdcIntegral;     ///< integral of the Landau fit for all time bins
+  uint16_t mAdcSumTruncated; ///< same as mAdcSum, but only ADCs close to the RMS are counted
+  uint8_t mRow;              ///< pad row number [0..15]
+  uint8_t mCol;              ///< pad column number [0..143]
+  uint8_t mDeltaRow;         ///< cluster size in row direction
+  uint8_t mDeltaCol;         ///< cluster size in column direction
```
tdietel, line 79: From the comments; it's not clear if `mRow`, `mCol` refer to the 'first' row/column, or if they refer to some kind of 'central' value, e.g. weighted means.

DataFormats/Detectors/TRD/include/DataFormatsTRD/KrCluster.h
```diff
@@ -0,0 +1,92 @@
...
+  uint8_t mDeltaTime;        ///< cluster size in time bin direction
+  uint8_t mClusterSize;      ///< number of ADC values which contribute to the cluster (ADC needs to be above KrClusterFinder::mMinAdcClSize)
```
tdietel: I find the comment 'ADC needs to be above KrClusterFinder::mMinAdcClSize' confusing:
I would expect mMinAdcClSize to refer to the min number of ADCs to form a cluster, but here it sounds like it's the minimum ADC value required to add the value to the cluster.

Detectors/TRD/calibration/src/KrClusterFinder.cxx
```diff
@@ -0,0 +1,387 @@
...
+#include "TRDCalibration/KrClusterFinder.h"
+#include "Math/IntegratorOptions.h"
+#include <fairlogger/Logger.h>
+
+#include <limits>
+
+using namespace o2::trd;
+using namespace o2::trd::constants;
+
+double KrClusterFinder::LandauFunctor::operator()(const double* par) const
```
tdietel: It seems this functor calculates the chi2 w.r.t. the Landau distribution. Maybe add a Chi2 to the name, e.g. LandauChi2Functor.
Although I do not see the treatment of statistical uncertainties.

Detectors/TRD/workflow/io/src/TRDDigitReaderSpec.cxx
martenole: Maybe @jolopezl knows?

tdietel (dismissed):
Overall the code looks fine. The Kr clusterfinder is a rather hard read, with very few comments on the algorithm in the code. Is there a chance to improve this while the memory is still fresh.
If we want to get this into O2 quickly, we can merge this already.

DataFormats/Detectors/TRD/include/DataFormatsTRD/KrCluster.h
martenole: It is row and column number of the pad with the max ADC value. I will add that comment

DataFormats/Detectors/TRD/include/DataFormatsTRD/KrCluster.h
martenole: Yes, it is the minimum ADC value required to add the value to the cluster. Will change the name

Detectors/TRD/calibration/src/KrClusterFinder.cxx
martenole: Ok will do. Concerning treatment of statistical uncertainties I let @aschmah comment. I think in your macro you ignored them as well, right?

martenole:
Hi @tdietel 
thanks for your comments. I try to add some more elaborate comments to the cluster finder to improve its readability. After all the goal is to have this in O2 specifically to simplify the maintenance of the code..
Although I already added many more comments compared to the code which I got from Alex ;)

Detectors/TRD/calibration/src/KrClusterFinder.cxx
aschmah: The statistical error bars were taken into account in the fit. Not sure how it is in the new fit handling @martenole is using.

Detectors/TRD/workflow/io/src/TRDDigitReaderSpec.cxx
jolopezl: @tdietel, @martenole: "maybe" is just some initial design that has never followed

Detectors/TRD/calibration/src/KrClusterFinder.cxx
aschmah: Ok. Its not taken into account in the new functor. Its just (y[i] - f)^2. One just needs to add (y[i] - f)^2/sqrt(y[i])^2 = (y[i] - f)^2/y[i], just divide by y[i].

Detectors/TRD/workflow/io/src/TRDDigitReaderSpec.cxx
bazinski: I think that was originally me, and had something to do with pipe-lining the dpl, but was not done.

Detectors/TRD/calibration/src/KrClusterFinder.cxx
martenole: @aschmah are you sure? If I divide by y[i] then I get twice as many failed fits and lots of clusters with out-of-range values. What is ROOT doing per default if you call TH1::Fit like you did before? Is the error for each bin just sqrt of its bin content?

martenole: This is fixed, previously I did not set the fit range correctly. So the errors are accounted in the same way as previously in the TH1::Fit method.
Setting the fit ranges correctly also yields another factor 3 speedup :)

martenole:
Hi @shahor02 
could you please merge this when you get a chance?
Cheers,
Ole

davidrohr:
Hi @martenole : Do you want to partially squash the krypton CF commits, or shall I just squash-merge?
The CI has passed, so if you only do a partial squash and then ping me I can force-merge it.

martenole:
Hi @davidrohr
let me partially squash and then I'll ping you. Thanks!

martenole:
ping @davidrohr 
did nothing else but squash 3 of the 4 commits. Just took some time since I was working on a different branch

PR #6730, opened 2021-07-23
[rANS] Fix missing Encoder normalization initialization in Constructor

MichaelLettrich:
Fixes missing initialization of the symbol table precision that caused that all rANS encoders always initialized the dictionary normalized to the minimal number of bits. In some cases this caused wrong results in the decoder, as another normalization was used for the dictionary there.


shahor02 (dismissed):
Hi @MichaelLettrich 
How does this affect encoding/decoding with the dictionary generated with current dev?
We have some CTF-writing tests foreseen for Monday and the CTF data should be later analyzed.

MichaelLettrich:
@shahor02 This bug does not affect decoding of the `CTF`, since metadata in `EncodedBlocks` is consistent.

MichaelLettrich:
@shahor02 can we merge?

PR #6731, opened 2021-07-23
DPL: populate DataTakingContext from data

ktf:


ktf:
@shahor02 as discussed, this will populate the `DataTakingContext` service with the orbit reset time coming from `DataProcessingHeader::creation`.

PR #6732, opened 2021-07-23
TRD Make TrackletHCHeader optional

bazinski:
- make tracklet half chamber header optional
- defaulted to option 0 never there.
- options added to mc2raw to enable checking.

Options:
- 0 no hcheader
- 1 hcheader iff tracklet data
- 2 hcheader always there irrespective of tracklet data.

Option 1 is unlikely.

It will mostly give TrackletHCHeader sanity check warning if options get muddled. Not full proof and can cause a seg fault if options chosen badly. This is merely to enable data taking with unkown status of tracklet hc header, but suspect to be option 0.

### DIFF
--- a/DataFormats/Detectors/TRD/src/RawData.cxx
+++ b/DataFormats/Detectors/TRD/src/RawData.cxx
@@ -324,12 +324,22 @@ bool trackletHCHeaderSanityCheck(o2::trd::TrackletHCHeader& header)
 {
   bool goodheader = true;
   if (header.one != 1) {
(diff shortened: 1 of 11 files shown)


bazinski:
@tdietel can this get approved its needed for the Kr data next week to handle the tracklethalfchamber headers.

tdietel (approved):
Looks good. Let’s merge it.

bazinski:
Hi @shahor02 could this be merged?

bazinski:
ooops clicked the wrong button at the bottom

PR #6733, opened 2021-07-23
[FV0] Reconstruction skeleton

mslupeck:
Mainly forwards the digits to rec points + experimental collision time calculation
Some leftover commented code (copied from FT0) will be useful during the subsequent implementation of calibration.


AllaMaevskaya (dismissed):
May be CollisionTimeReco is not the best name for FV0?

DataFormats/Detectors/FIT/FV0/include/DataFormatsFV0/RecPoints.h
```diff
@@ -0,0 +1,104 @@
...
+
+class RecPoints
+{
+
+ public:
+  enum TimeTypeIndex : int { TimeFirst,
+                             TimeGlobalMean,
+                             TimeSelectedMean };
+
+  o2::dataformats::RangeReference<int, int> ref;
```
shahor02: If you define the class (rather than struct) then better keep all data members as private with corresponding getters/setters. Also, interleaving data members and methods is not a good practice.

@AllaMaevskaya since this is a copy paste from FT0 code, could you please fix it also there?

DataFormats/Detectors/FIT/FV0/include/DataFormatsFV0/RecPoints.h
```diff
@@ -0,0 +1,104 @@
...
+  o2::fv0::Triggers getTrigger() const { return mTriggers; }
+
+  o2::InteractionRecord getInteractionRecord() const { return mIntRecord; };
+
+  void SetMgrEventTime(Double_t time) { mTimeStamp = time; }
+
+  gsl::span<const ChannelDataFloat> getBunchChannelData(const gsl::span<const ChannelDataFloat> tfdata) const;
+
+ private:
+  std::array<Float_t, 3> mCollisionTime = {2 * o2::InteractionRecord::DummyTime,
```
shahor02: These are times in ps wrt the center of the BC indicated by the `mIntRecord`, right? In this case its range is limited to +- 25ns / 2. Is there a point to use float here (which would mean you have better than 1ps resolution) or simple short would suffice? 
@AllaMaevskaya , the same question for FT0

DataFormats/Detectors/FIT/FV0/include/DataFormatsFV0/RecPoints.h
```diff
@@ -0,0 +1,104 @@
...
+                                           2 * o2::InteractionRecord::DummyTime,
+                                           2 * o2::InteractionRecord::DummyTime};
+  Double_t mTimeStamp = 2 * o2::InteractionRecord::DummyTime; //event time from Fair for continuous
```
shahor02: This seems never used, can it be suppressed (@AllaMaevskaya also in FT0)

Detectors/FIT/FV0/workflow/include/FV0Workflow/FV0DataReaderDPLSpec.h
```diff
@@ -0,0 +1,101 @@
...
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// @file   FV0DataReaderDPLSpec.h
+
+#ifndef O2_FV0DATAREADERDPLSPEC_H
```
shahor02: To be sure: what is the role of this class? Currently the V0 decoding is done via https://github.com/AliceO2Group/AliceO2/blob/d4991a76d96b5460c02c43561030f133e6ad97f9/Detectors/FIT/FV0/workflow/src/fv0-flp-workflow.cxx#L79-L84, is this class supposed to substitute decoding via FIT-common FITDataReaderDPLSpec ?

DataFormats/Detectors/FIT/FV0/include/DataFormatsFV0/RecPoints.h
mslupeck: Agreed. This is a simple reshuffling of data members and doesn't involve modification of any other files. @AllaMaevskaya would you like me to do it also for FT0 or do you prefer to follow it yourself?

DataFormats/Detectors/FIT/FV0/include/DataFormatsFV0/RecPoints.h
mslupeck: I modified it for FV0 to use short instead of float - good point

DataFormats/Detectors/FIT/FV0/include/DataFormatsFV0/RecPoints.h
mslupeck: Done

Detectors/FIT/FV0/workflow/include/FV0Workflow/FV0DataReaderDPLSpec.h
mslupeck: It seems it's not needed. It's removed now

DataFormats/Detectors/FIT/FV0/include/DataFormatsFV0/RecPoints.h
AllaMaevskaya: I will do everything for FT0 - but at the end of week, OK?

shahor02:
Thanks @AllaMaevskaya.
@mslupeck shall I merge this partial PR once the tests are passed or you will finalize it?

mslupeck:
> @mslupeck shall I merge this partial PR once the tests are passed or you will finalize it?
Yes, please merge it as it is. I will have to play a little with calibration before it can be applied in reco.

shahor02:
OK, but can we already enable it in reconstruction, even if the results are not yet optimal?

mslupeck:
Currently, there's no FV0 calibration object in CCDB. I was planning to put some dummy there first and then test it in reconstruction, but it will take some time... I could uncomment the calibration-related lines in reco immediately, but I'd prefer to test it first.

mslupeck:
Hi @shahor02
Could you merge it? The tests are completed.
FYI - I already started working to include the first calibration to FV0, but still, I'd prefer to do it in a separate commit

shahor02 (approved):

PR #6735, opened 2021-07-23
Support for remote DebugGUI

luigifusco:
* Fix bit order in websocket header packets allowing connections from browsers (atm only Firefox supported)
* Use new DebugGUI API (multiple functions instead of single pollGUI)
* Changes to contexts:
  * Move GuiCallbackContext definition to its own header file
  * Add reference to GuiCallbackContext in DriverServerContext
  * Add support for string callbacks in GuiCallbackContext to be called with the json representation of a rendered frame

Callbacks are set by WSDPLHandlers which are connected to a browser (no set PID).
Callbacks are called every ~200ms in the main GUI loop with a json representation of the last rendered frame.
The callbacks send the data to the browser.

### DIFF
--- a/Framework/Core/include/Framework/DebugGUI.h
+++ b/Framework/Core/include/Framework/DebugGUI.h
@@ -36,9 +36,9 @@ struct DebugGUI {
   virtual void* initGUI(char const* windowTitle) = 0;
   virtual bool pollGUI(void* context, std::function<void(void)> guiCallback) = 0;
   virtual void getFrameJSON(void *data, std::ostream& json_data) = 0;
-  virtual bool pollGUI_gl_init(void* context) = 0;
-  virtual void* pollGUI_render(std::function<void(void)> guiCallback) = 0;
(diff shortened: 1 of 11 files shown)


ktf (dismissed):
Ok to run tests, not to merge.

ktf:
Please have a look at the errors.

luigifusco:
> Please have a look at the errors.

are they related to missing implementations in the DebugGUI?

peressounko:
Hi, I suspect that git pull --rebase was not done before commit. At least for PHOS this PR tries push again the latest changes. Please, do not commit them as is and pull latest PHOS code before.

Barthelemy:
@luigifusco I agree with @peressounko, there seems to be many commits that don't belong to you. You should probably rebase `dev` on your branch and then force push. Let us know if you need help.

ktf:
I guess this one is superseded by #6964?

ktf:
Yes, it is.

PR #6693, opened 2021-07-20
TPC: Update laser track drift velocity calibration

wiechula:
* add device to pre-filter laser track candiates
  (tpc-laser-track-filter)
* split device for
  - simple local processing (tpc-calib-laser-tracks)
  - distributed processing (tpc-laser-tracks-calibrator)
* new struct for calibration data storage
* drift velocity calibration separately for A-/C-Side
* code cleanup

PR #6736, opened 2021-07-23
TPC dEdx calibration values for each GEM stack

badarots:
In this pull request, I extended the TPC dE/dx calibration code to compute values for each GEM stack.
I also renamed some classes, again, sorry, to keep the namings more consistent. Now we have the following:

- `CalibdEdxHistos` is now the class the holds the dE/dx histograms. One for each GEM stack.
- `CalibdEdx` is the class the holds the calibration values extracted from the histograms.
- `CalibdEdxBase` is a templated base class that groups the common functionalities of the two above.
- `CalibratordEdx` is the time slot calibrator.

@wiechula let me know if you agree with my changes.

In `CalibdEdxBase` derived classes we have two arrays storing the Tot and Max dE/dx values for each GEM stack, 144 entries per array. The stack data are layout in the following way:

```
|         A Side          |         C Side           |
 IROC, OROC1, OROC2, OROC3, IROC, OROC1, OROC2, OROC3
```

And, for example, we can access the Tot data of the IROC 12 stack of A side with this call:
    
    CalibdEdxBase::getTotEntry(12, ReadoutType::IROC, TPCSide::A)

`CalibdEdxHistos` is reponsible to find out the GEM stacks crossed by a track. It do so finding the angle phi of the track using the following call in a "TrackTPC" object:

    TrackTPC::getOuterParam().getXYZGloAt(stack_center, 0.5, ok).Phi()

To get the calibration value from the histograms we construct a `CalibdEdx` object using as input a `CalibdEdxHistos`

    CalibdEdx calibOutput(CalibdEdxHistos)
    
During construction, the center of mass of every histogram is computed and stored in the same layout mentioned earlier.

This is a plot with if dEdx histograms and computed MIP position of every GEM stack (600 pp events).

dedx


Detectors/TPC/calibration/include/TPCCalibration/CalibdEdxBase.h
```diff
@@ -0,0 +1,92 @@
...
+    IROC,
+    OROC1,
+    OROC2,
+    OROC3,
+  };
+
+  enum class TPCSide {
+    A,
+    C
+  };
```
wiechula: This is duplication of information already available here
https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/TPC/include/DataFormatsTPC/Defs.h

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdxBase.h
```diff
@@ -0,0 +1,92 @@
...
+namespace o2::tpc
+{
+
+/// Define the commom properties of dEdx containers. Used by CalibdEdxHistos and CalibdEdx.
+template <typename Entry>
+class CalibdEdxBase
+{
+ public:
+  static const size_t stacksPerTurn = 18;
+  static const size_t totalStacks = 18 * 8;
```
wiechula: Please check information in 
https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/TPC/include/DataFormatsTPC/Defs.h
```suggestion
static const size_t totalStacks = SECTORSPERSIDE * SIDES * GEMSTACKSPERSECTOR;```
```
your `stacksPerTurn` is `SECTORSPERSIDE`

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdxBase.h
```diff
@@ -0,0 +1,92 @@
...
+    return mTotEntries[index];
+  }
+
+  Entry getMaxEntry(size_t stackNumber, ReadoutType type, TPCSide side) const
+  {
+    const size_t index = stackIndex(stackNumber, type, side);
+    return mTotEntries[index];
+  }
+
+ protected:
```
wiechula:
```suggestion
 private:
```

Detectors/TPC/calibration/src/CalibratordEdx.cxx
```diff
@@ -0,0 +1,110 @@
...
+{
+  // Here we initialize the vector of our output objects
+  mInfoVector.clear();
+  mMIPVector.clear();
+  return;
+}
+
+void CalibratordEdx::finalizeSlot(Slot& slot)
+{
+  LOG(INFO) << "Finalizing slot " << slot.getTFStart() << " <= TF <= " << slot.getTFEnd();
```
wiechula: Please change all logging message to `fmt::format` style using `LOGP(info|error|warning, "", ...)`
```suggestion
  LOGP(info, "Finalizing slot {} <= TF <= {}" , slot.getTFStart(), slot.getTFEnd());
```

wiechula (changes requested):
Some minor comments and a more conceptual one.
As layout I would propose somethink like
* CalibdEdxDataContainer -> Data storage for either histograms or floats per stack (now CalibdEdxBase)
* CaliddEdx -> standalone calibration class, something like
```c++
class CalibdEdx {

void fill();
void finalize(); 
...


private:
  CalibdEdxDataContainer<FastHisto<float>> mHistos;
  CalibdEdxDataContainer<float> mCalibrationData;
}

```
* CalibratordEdx -> Using CalibdEdx as calibration class
* As processors I would propose to implement two
  - one standalone CalibdEdxSpec
  - one time frame processor CalibratordEdxSpec

Please also have a look at https://github.com/AliceO2Group/AliceO2/pull/6693

badarots:
@wiechula I modified the class layout and changed the log type. The two CalibdEdxSpec and CalibratordEdxSpec are still missing but I will create them shortly.

We talked about adding a fill histogram function in the DataContainer class. I tried some ideas and didn't like the result.

One reason I didn't add the fill function is that I thought about start using a multidimensional histogram to keeping track of the GEM stacks, instead of keeping an array of 1D histograms. I can define a hist. with the following dimensions:

dEdx data (~100 bins)
Sectors (18 bins)
Type (4 bins)
Side (2 bins)
Charge type: Tot and Max (2 bins) -> maybe split into 2 histograms?

The nice thing about using the dimensions to separate the gem stacks data is that it is trivial to get rid/merge any of those dimensions using the project method of a THn or boost histogram. And it will be very easy to add more.

What do you think about this? Should I swap the array of FastHistos for a dense multidim. histogram?

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -11,91 +11,207 @@
...
+    bool ok = false;
+    const int sector = findTrackSector(track, stack, ok);
+
+    // Ignore stack if we are not able to find its sector
+    if (!ok) {
+      continue;
+    }
+
+    // Fill the correct readout type
+    switch (stack) {
```
wiechula: You could consider getting rid of the switch by adding before the `for` loop:
```
std::vector<float> dEdxTot{dEdx.dEdxTotIROC, dEdx.dEdxTotOROC1, dEdx.dEdxTotOROC2, dEdx.dEdxTotOROC3};
// same for Max
```
And here simply do
```
mTotHist(dEdxTot[stack], sector, side, stack);
// same for Max
```

Detectors/TPC/calibration/src/TPCCalibrationLinkDef.h
```diff
@@ -64,10 +64,11 @@
 #pragma link C++ class o2::tpc::IDCCCDBHelper<float> +;
 #pragma link C++ class o2::tpc::IDCCCDBHelper<short> +;
 #pragma link C++ class o2::tpc::IDCCCDBHelper<char> +;
-#pragma link C++ class o2::calibration::TimeSlotCalibration < o2::tpc::TrackTPC, o2::tpc::dEdxHistos> + ;
+#pragma link C++ class o2::tpc::CalibdEdxDataContainer<float> + ;
+#pragma link C++ class o2::tpc::CalibdEdxDataContainer<o2::tpc::FastHisto<float>> + ;
```
wiechula: Is the FastHisto option still used?

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -26,7 +26,7 @@
 #include "Framework/Task.h"
```
wiechula, line 21: I think this file should be called `CalibratordEdxSpec.cxx`, since it runs the `Calibrator`. We should have a separate workflow to directly run the `Calib` called `CalibdEdxSpec.cxx`

wiechula (changes requested):
@badarots , very nice! A few comments to be considered.

PR #6738, opened 2021-07-24
DPL Analysis: Configurable process switches

aalkin:
This introduces a way to dynamically configure which process functions a task will run.
* Added a `ProcessConfigurable` with a distinct kind to be used with hyperloop
* Added a macro to declare process configurables in a task
* Added workflow adjustment from supplied configuration (WARN: only JSON configuration is supported currently, commandline switches will not work)
* Updated multiple process tutorial example

### DIFF
--- a/Analysis/Tutorials/src/multiProcess.cxx
+++ b/Analysis/Tutorials/src/multiProcess.cxx
@@ -66,7 +66,7 @@ struct MultipleProcessExample {
 
   /// name, description, function pointer, default value
   /// note that it has to be declared after the function, so that the pointer is known
-  PROCESS_SWITCH(prec, "prec", "Process reco level", &MultipleProcessExample::processRec, true);
+  PROCESS_SWITCH(processRec, "Process reco level", MultipleProcessExample, true);
 
   void processGen(soa::Filtered<aod::McCollisions>::iterator const&, aod::McParticles const& mcParticles)
   {
@@ -76,7 +76,7 @@ struct MultipleProcessExample {
     }
   }
 
-  PROCESS_SWITCH(pgen, "pgen", "Process gen level", &MultipleProcessExample::processGen, false);
+  PROCESS_SWITCH(processGen, "Process gen level", MultipleProcessExample, false);
 
   void processResolution(soa::Filtered<soa::Join<aod::Collisions, aod::McCollisionLabels>>::iterator const& collision, soa::Join<aod::Tracks, aod::McTrackLabels> const& tracks, aod::McParticles const&, aod::McCollisions const&)
   {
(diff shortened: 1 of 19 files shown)


Analysis/Tutorials/src/multiProcess.cxx
```diff
@@ -63,48 +63,41 @@ struct MultipleProcessExample {
...
+  void processRec(soa::Filtered<aod::Collisions>::iterator const&, aod::Tracks const& tracks)
   {
     for (auto& track : tracks) {
       registry.fill(HIST("etaRec"), track.eta());
       registry.fill(HIST("phiRec"), track.phi());
     }
   }
 
-  void processGen(soa::Filtered<aod::McCollisions>::iterator const& mcCollision, aod::McParticles const& mcParticles)
+  PCONF(prec, "prec", "Process reco level", &MultipleProcessExample::processRec, true);
```
jgrosseo: Let's put a longer name `PROCESS_CONFIGURABLE` or smth better?

jgrosseo (commented):
Looks good to me! I'd change the name of the macro though.

Analysis/Tutorials/src/multiProcess.cxx
aalkin: Sure, this is was a proof-of-concept at first and I did not change it. It is still required to declare this configurable after the process function it refers to, as it needs the pointer.

aalkin:
Marking this ready for review.

Note that changes in tasks are made to preserve current functionality where the process selection is controlled by a a workflow option. Analyzers can later improve that by using the process switches directly.

ktf (dismissed):

jgrosseo (dismissed):

jgrosseo:
@aalkin 
I don't think this can be related to your PR, can it?
` 20/383 Test  #17: Framework/Utils/test/test_RootTreeWriterWorkflow.cxx ...........................Subprocess aborted***Exception:   1.05 sec
[INFO] Initialising O2 Data Processing Layer. Driver PID: 15525.
[INFO] O2 Data Processing Layer initialised. We brake for nobody.
[INFO] No device resulting from the workflow. Quitting.
Error in <TFile::TFile>: file /tmp/test_RootTreeWriter.root does not exist
Error in <TFile::TFile>: file /tmp/test_RootTreeWriter_alt.root does not exist
terminate called after throwing an instance of 'std::runtime_error'
  what():  missing file /tmp/test_RootTreeWriter.root
`

aalkin:
Not sure, I will check.

aalkin:
@jgrosseo seems to be genuine error, I'll look into it

jgrosseo (approved):

ktf:
This seems to correlate with failing train runs since the 31st. Is it being looked into?

aalkin:
> This seems to correlate with failing train runs since the 31st. Is it being looked into?

Yes, I missed to add an exclusion, testing the fix right now.

PR #6740, opened 2021-07-24
GPUTRDTracker: Use CAMath::Abs

anerokhi:
Clang on CentOS 7 warns that `int abs(int)` would be called for the argument of a floating point type


davidrohr (changes requested):
please use `CAMath::Abs(...)` for GPU-related code

PR #6742, opened 2021-07-26
PWGPP/qaEfficiency: Prevent expensive copying

anerokhi:


jgrosseo (approved):

anerokhi:
Unfortunately, can not be automatically checked now. `clang-tidy`'s `performance-unnecessary-value-param` check only captures explicit types, i.e.
```c++
    using TT = std::decay_t<decltype(*std::declval<o2::aod::McParticle>())>;
    auto rejectParticle = [&](TT p, ...
```
and the fact that https://github.com/AliceO2Group/AliceO2/blob/27cb4743c749edf5af7be6a5fcfec0b2aa5a1d18/Analysis/Core/include/AnalysisCore/MC.h#L64 is not a reference to `const` doesn't help `clang-tidy`. With this fixed and with concrete type specified, `clang-tidy` says:
```
Analysis/Tasks/PWGPP/qaEfficiency.cxx:215:34: warning: the parameter 'p' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
    auto rejectParticle = [&](TT p, auto h) {
                                 ^
                              const  &
```

PR #6745, opened 2021-07-26
Use refernce to const

anerokhi:


jgrosseo:
Thanks. This is included here: https://github.com/AliceO2Group/AliceO2/pull/6749

PR #6746, opened 2021-07-26
[ZDC] Event pedestal subtraction, decoding of reconstructed data, bug fix

cortesep:


DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEventFlat.h
```diff
@@ -0,0 +1,180 @@
...
+
+  inline NElem getNTDC()
+  {
+    return mNT;
+  }
+
+  inline NElem getNInfo()
+  {
+    return mNI;
+  }
```
shahor02: I guess these getters should be `const`

DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEvent.h
```diff
@@ -60,15 +61,68 @@ struct RecEvent {
...
+    printf("addInfo ch=%u code=%u info=%u 0x%04x\n", ch, code, info, info);
+#endif
     mInfo.emplace_back(info);
     mRecBC.back().addInfo();
   }
+
+  void addInfo(const std::array<bool, NChannels>& vec, const uint16_t code)
+  {
+    int cnt = 0;
+    for (uint8_t ich = 0; ich < NChannels; ich++) {
```
shahor02, line 92: you are already looping once over all channels, why don't you fill in the local `std::array<int, NChannels>` indices of cnt non-empty channels, and then in the further loops, scan only those.

DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEventFlat.h
```diff
@@ -0,0 +1,180 @@
...
+
+  float tdcV(uint8_t ich, uint64_t ipos)
+  {
+    if (ich < NTDCChannels) {
+      if (ipos < TDCVal[ich].size()) {
+        return FTDCVal * TDCVal[ich][ipos];
+      }
+    }
+    return -std::numeric_limits<float>::infinity();
```
shahor02, line 98: how are you later going to process the `infinity` values?

DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEventFlat.h
```diff
@@ -0,0 +1,180 @@
...
+
+  float EZDC(uint8_t ich)
+  {
+    std::map<uint8_t, float>::iterator it = ezdc.find(ich);
+    if (it != ezdc.end()) {
+      return it->second;
+    } else {
+      return -std::numeric_limits<float>::infinity();
+    }
+  }
```
shahor02: these are also const methods

DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEventFlat.h
```diff
@@ -0,0 +1,180 @@
...
+  float EZNC3() { return EZDC(IdZNC3); }
+  float EZNC4() { return EZDC(IdZNC4); }
+  float EZNCSum() { return EZDC(IdZNCSum); }
+
+  float EZPCC() { return EZDC(IdZPCC); }
+  float EZPC1() { return EZDC(IdZPC1); }
+  float EZPC2() { return EZDC(IdZPC2); }
+  float EZPC3() { return EZDC(IdZPC3); }
+  float EZPC4() { return EZDC(IdZPC4); }
+  float EZPCSum() { return EZDC(IdZPCSum); }
```
shahor02: these too.

DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEventFlat.h
cortesep: In this case it should be QC. Infinity should never happen, unless there is some memory corruption or a bug in the code.
If you use TreePlayer in root it also helps in producing a correct histogram by putting all missing entries in underflow

DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEventFlat.h
```diff
@@ -0,0 +1,180 @@
...
+
+namespace o2
+{
+namespace zdc
+{
+using FirstEntry = int;
+using NElem = int;
+
+struct RecEventFlat {
+  o2::InteractionRecord ir;
```
shahor02: The fullCI triggers here a warning `statement should be inside braces`, looks like a false positive.
@davidrohr @timo I don't know who sets up the O2codecheck, do you know?

davidrohr: That uses clang-tidy, if there is a bug, there is not much we can do except for waiting for a new clang version
Can you try to put `//NOLING: false positive in clang-tidy`
at the end of line 38, that should silence it. If it doesn't work, I would just force-merge it.

davidrohr: Argh, should be `//NOLINT: false positive in clang-tidy` !!

shahor02: thanks, let's see....

shahor02:
@cortesep This PR produces `reco_ASYNC.log:[22065:zdc-digi-reco]:  *** Break *** segmentation violation`

shahor02:
Hi @cortesep 
Could you also add a fix for this message:
````
[30507:zdc-digi-reco]: [ERROR] Unknown verbosity setting: '0', setting to default 'low'.
````
?

cortesep:
Fixed error with verbosity setting. I cannot reproduce the segmentation violation. Could you tell me more about how to reproduce the error?

cortesep:
Actually the segmentation violation was happening with a 20-30% probabiliy and seems to be related with an array of booleans that was not initialized. Should be fixed now.

cortesep:
Is it possible to merge also this pull request or are there any further improvements to be done?

shahor02 (approved):

PR #6747, opened 2021-07-26
DPL: avoid warning about uninitialised usage

ktf:
Still not clear to me wether the condition can actually be triggered.

It might explain phantom channels to the input proxy.

### DIFF
--- a/Framework/Core/src/DeviceSpecHelpers.cxx
+++ b/Framework/Core/src/DeviceSpecHelpers.cxx
@@ -487,6 +487,20 @@ void DeviceSpecHelpers::processInEdgeActions(std::vector<DeviceSpec>& devices,
     return deviceIt->deviceIndex;
   };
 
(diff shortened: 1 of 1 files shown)


ktf:
@TimoWilken macOS fails without logs.

ktf:
@TimoWilken I still regularly see PRs being sent back in the queue.

Framework/Core/src/DeviceSpecHelpers.cxx
```diff
@@ -708,14 +722,16 @@ void DeviceSpecHelpers::processInEdgeActions(std::vector<DeviceSpec>& devices,
   for (size_t edge : inEdgeIndex) {
     auto& action = actions[edge];
 
-    size_t consumerDevice;
+    size_t consumerDevice = -1;
```
anerokhi, line 725: Is this necessary with `else` branch added?

ktf: Not necessary, but allows us to spot actual issues if the code is changed.

PR #6750, opened 2021-07-26
DPL Analysis: sanitize writing of tables

ktf:
* Fix memory leaks
* Fix deletion of non-owned memory
* Fix memory corruption due to the fact wrong costructor is used
  for the an array of boolean.


ktf:
currently broken. Merging.

PR #6753, opened 2021-07-26
DPL Analysis: Introducing conditional expressions

aalkin:
* `ifnode(condition, then, else)` operation is added to expressions
* these can be nested
* all three arguments can be arbitrary valid expressions
* `condition` needs to have boolean result, `then` and `else` should return similar types (ideally the same - both floats, or both boolean, etc.)
* Added a test
* Added `conditionalExpressions.cxx` tutorial example (note that it uses bitwise operations in filter expression and thus will only work 'as is' with arrow > 3.0)

@jgrosseo this is fully functional

### DIFF
--- a/Analysis/Tutorials/CMakeLists.txt
+++ b/Analysis/Tutorials/CMakeLists.txt
@@ -228,3 +228,9 @@ o2_add_dpl_workflow(multiprocess-example
                   JOB_POOL analysis
                   PUBLIC_LINK_LIBRARIES O2::Framework O2::AnalysisCore O2::AnalysisDataModel
                   COMPONENT_NAME AnalysisTutorial)
+
+o2_add_dpl_workflow(conditional-expressions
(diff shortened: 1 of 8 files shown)


jgrosseo (dismissed):
Great!

jgrosseo:
Mac is not happy... `/System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SOURCES/O2/6753/0/Analysis/Tasks/PWGLF/spectraTPCPiKaPr.cxx:52:9: error: value of type 'o2::framework::expressions::Node' is not contextually convertible to 'bool'
    if (abs(nsigma) > cfgNSigmaCut) {`

aalkin:
There was an issue with operations between configurables and literal values. In the filter expressions one would may need to use `.node()` for configurables to ensure that it will be updated and not just used as default. The normal usage in `if ()` will be unaffected.

PR #6755, opened 2021-07-26
DPL: automatically load objecs from CCDB

ktf:


ktf:
@chiarazampolli with this I am able to correctly deserialise the object from CCDB. To complete this we just need to use the timestamp from the DataTakingContext (which probably should be renamed to account for usage in analysis).

ktf:
Skipping retesting of trivial warning fix. Merging.

PR #6759, opened 2021-07-27
DPL: fix handling of timepipelining in reader

ktf:
Up to now, a reader device was erroneously sending extra tables around if a timeline pipelined device (e.g. a spawner) was present. This causes deadlocks for certain topologies.


ktf:
@jgrosseo this fixes the issue with multiple `--spawners`, however unfortunately it is not the same problem responsible for the hanging HF workflows.

ktf:
In case of problems revert 6e93dcb1466355d3289ad726cd4d1cb35d721ce9 only. The other two commits are just "riders".

PR #6760, opened 2021-07-27
Next step towards the simulation challenge

victor-gonzalez:
Producing filtered tables for both
- reconstructed
- generated

Generating QA for the filter process for both
- reconstructed
- generated


victor-gonzalez:
There is a wrong use of `TParticlePDG` that seems to only be caught by the build on macOS
I will commit a fix

victor-gonzalez:
There is still one issue with the TracksExtended table remnant of the nightly tests
I will commit a fix

victor-gonzalez:
In my understanding, the error on macOS is basically saying no errors

jgrosseo (approved):
@victor-gonzalez Can you also start adding comments for all columns and tables (see data model for the format)

jgrosseo:
Mac CI empty. Merging.

PR #6761, opened 2021-07-27
[MFT] Standalone tracks with ambiguous vertices marked as orphans

rpezzi:
This PR sets MFT standalone tracks with ambiguous matched vertices as orphans to have them stored only once as unassigned MFT tracks.  At the moment track-vertex matching is performed by time compatibility.

On a sample run of 200 pp events I got the collision distribution for MFT standalone tracks at AOD level reducing from MFT 9141 tracks originally (~46 tracks/event - red line - current O2@dev), down to 1545 unassigned tracks (~8 tracks/event) - see figure. By reducing the interaction rate to 25000 Hz we observe about 20% of the tracks with unique track-vertex association.

image

It is a temporary solution to avoid explosion of AOD file size with redundant MFT tracks while a better MFT track to vertex matching is not available.

### DIFF
--- a/Detectors/Vertexing/src/VertexTrackMatcher.cxx
+++ b/Detectors/Vertexing/src/VertexTrackMatcher.cxx
@@ -118,16 +118,28 @@ void VertexTrackMatcher::process(const o2::globaltracking::RecoContainer& recoDa
       // track matches to vertex, register
       vtxList.push_back(vto.origID); // flag matching vertex
     }
-    if (vtxList.size()) {
-      nAssigned++;
-      for (auto v : vtxList) {
-        tmpMap[v].emplace_back(tro.origID).setAmbiguous();
+    if (tro.origID.getSource() == GIndex::Source::MFT) { // MFT tracks are treated differently: tracks with unresolved ambiguities are marked as orphans -> unassigned
+      if (vtxList.size() == 1) {
+        nAssigned++;
+        tmpMap[vtxList[0]].emplace_back(tro.origID).setAmbiguous();
+      } else {
+        orphans.emplace_back(tro.origID); // register as unassigned MFT track
       }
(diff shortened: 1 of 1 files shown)


jgrosseo:
This is fine with me. Given that it is outside of Analysis, someone from the muon project should approve.

shahor02:
Fine for me too as a temporary measure, but is it needed? I.e. are we now suffering from the multiply stored tracks?
The ambiguities will be always there, especially in pp and the proper solution would be to store them only once with proper reference on the BCs slice.

jgrosseo:
@shahor02 Are you saying that if we align ambiguousmfttracks with ambiguoustracks, i.e. add the BC slice, then this would not be needed?
https://github.com/jgrosseo/AliceO2/blob/dev/Framework/Core/include/Framework/AnalysisDataModel.h#L522

shahor02:
I think so: the ambiguous MFT tracks are fully equivalent to ambiguous ITS tracks (i.e. those which were not contributing to PVertex fit as matched only by time).

jgrosseo:
I agree.
@rpezzi Can you adjust ambiguousmfttracks then including the code which fills it? Thanks!

rpezzi:
Hi @jgrosseo ,  @shahor02 
> This is fine with me. Given that it is outside of Analysis, someone from the muon project should approve.

From the MFT/Muon part I approve.

> Fine for me too as a temporary measure, but is it needed? I.e. are we now suffering from the multiply stored tracks?

The impact of MFT tracks ambiguity is much higher since none is contributing to PV.

>  @rpezzi Can you adjust ambiguousmfttracks then including the code which fills it? Thanks!

For the intended purpose of realistic AOD file size, this PR does the job. Further developments to resolve ambiguities and handling ambiguous tracks will surely happen, after vacations :sunglasses:.

shahor02:
OK, merging until after-vacation

PR #6766, opened 2021-07-28
Add multithreading to ITS-TPC afterburner

shahor02:


Detectors/GlobalTracking/src/MatchTPCITS.cxx
```diff
@@ -1576,17 +1580,27 @@ void MatchTPCITS::runAfterBurner()
     return;
   }
   mTimer[SWABMatch].Start(false);
+  std::vector<ITSChipClustersRefs> itsChipClRefsBuff(mNThreads);
+#ifdef WITH_OPENMP
+  omp_set_num_threads(mNThreads);
+#pragma omp parallel for schedule(dynamic)
```
davidrohr: Could you instead use
`#pragma omp parallel for schedule(dynamic) num_threads(mNThreads)`

In that way, you will set the number of threads only for this loop, but not override it globally.

PR #6767, opened 2021-07-28
add parameters to eve workflow

jmyrcha:


EventVisualisation/Workflow/CMakeLists.txt
```diff
@@ -21,18 +21,31 @@ if(ALIGPU_BUILD_TYPE STREQUAL "O2"
...
+                O2::DetectorsRaw
+                O2::DetectorsVertexing
+                O2::EventVisualisationBase
+                O2::EventVisualisationDetectors
+                O2::FrameworkFoundation
+                O2::FT0Workflow
+                O2::GlobalTrackingWorkflowHelpers
+                O2::GlobalTrackingWorkflowReaders
+                O2::GPUO2Interface
+                O2::GPUWorkflowHelper
```
davidrohr: Why do you need these 2 GPU libraries?

EventVisualisation/Workflow/include/EveWorkflow/EveConfiguration.h
```diff
@@ -0,0 +1,95 @@
...
+#ifndef GPUCA_HAVE_O2HEADERS
+#define GPUCA_HAVE_O2HEADERS
+#endif
+#ifndef GPUCA_TPC_GEOMETRY_O2
+#define GPUCA_TPC_GEOMETRY_O2
+#endif
+#ifndef GPUCA_O2_INTERFACE
+#define GPUCA_O2_INTERFACE
+#endif
+
```
davidrohr: You should not set the defines if you are not GPU code.

EventVisualisation/Workflow/include/EveWorkflow/EveConfiguration.h
```diff
@@ -0,0 +1,95 @@
...
+#include <memory>
+#include <array>
+#include <vector>
+#include <functional>
+#include <gsl/gsl>
+#include <GPUO2Interface.h>
+#include "GPUSettings.h"
```
davidrohr: These 2 includes should not be needed.

EventVisualisation/Workflow/include/EveWorkflow/EveWorkflowHelper.h
```diff
@@ -0,0 +1,42 @@
...
+  using GID = o2::dataformats::GlobalTrackID;
+
+ public:
+  struct tmpDataContainer {
+    std::vector<o2::BaseCluster<float>> ITSClustersArray;
+    std::vector<int> tpcLinkITS, tpcLinkTRD, tpcLinkTOF;
+    std::vector<const o2::track::TrackParCov*> globalTracks;
+    std::vector<float> globalTrackTimes;
+  };
+  static std::shared_ptr<const tmpDataContainer> fillIOPtr(gpu::GPUTrackingInOutPointers& ioPtr, const o2::globaltracking::RecoContainer& recoCont, bool useMC, const CalibObjectsConst* calib = nullptr, GID::mask_t maskCl = GID::MASK_ALL, GID::mask_t maskTrk = GID::MASK_ALL, GID::mask_t maskMatch = GID::MASK_ALL);
```
davidrohr: I don't understand what you do with the `gpu::GPUTrackingInOutPointers& ioPtr` here. If I understand correctly, you have copied my fillIOPtr function into your WorkflowHelper. I have a few remarks here:
- So first, I don't understand why you need the copying into the ioPtr at all. For the GPU this is only done since the GPU doesn't support all the data types. For the event display, you could just use the data directly without copying to pointers to a special structure beforehand.
- If you use my structure, why do you create a copy of fillIOPtr function? I would avoid code duplication if possible.
- If you want to fill the ioPtr by yourself, I would propose you also create your own ioPtr object. the gpu::GPUTrackingInOutPointers is no fixed API and it can change when the GPU code is developed further, so I don't like the dependency to the event display.

EventVisualisation/Workflow/include/EveWorkflow/EveWorkflowHelper.h
```diff
@@ -0,0 +1,42 @@
...
+///
+/// \file   EveWorkflowHelper.h
+/// \author julian.myrcha@cern.ch
+
+#ifndef ALICE_O2_EVENTVISUALISATION_WORKFLOW_EVEWORKFLOWHELPER_H
+#define ALICE_O2_EVENTVISUALISATION_WORKFLOW_EVEWORKFLOWHELPER_H
+
+#include "ReconstructionDataFormats/GlobalTrackID.h"
+#include "DataFormatsGlobalTracking/RecoContainer.h"
+#include "GPUDataTypes.h"
```
davidrohr: I think you do not need the GPUDataTypes.
Well, currently I think you actually need it for the ioPtr structure, but perhaps we can get rid of that. See the comment below.

davidrohr (changes requested):
Hi @jmyrcha , this looks mostly OK. I have a couple of points inline where there is copy&pasted code from the GPU folder, which is not needed, and I think it should be removed.

But more importantly, I think 2 important points for the commissioning are missing:
- We need  a command line option to optionally limit the maximum number of tracks that can be written in a JSON file.
- We need a command line option to make skip the JSON export if the hostname of the node is not equal to a string provided via command line.

EventVisualisation/Workflow/CMakeLists.txt
jmyrcha: fixed

EventVisualisation/Workflow/include/EveWorkflow/EveWorkflowHelper.h
jmyrcha: I removed usage of ioPtr completely.

EventVisualisation/Workflow/include/EveWorkflow/EveWorkflowHelper.h
jmyrcha: done

jmyrcha:
I have implemented all points
(current pull request fails, but it looks like errors on the test environment - logs are empty). If there are any action on my side please let me know.

davidrohr (changes requested):
Hi @jmyrcha, thx looks better, checking the code again, I saw one bogus comment:
```
// Full configuration structure with all available settings of GPU...
```
And I am listed as author of several of the files you added, could you clean that up as well?

Finally, could you comment when we can have the 2 missing settings, to limit the number of tracks, and to check the hostname?

davidrohr:
@TimoWilken : Could you check why the CI is failing without a log?

jmyrcha:
Missing settings are already there:
eve-hostname - if set then workflows on computers which hostname do not match will do nothing
number-of_tracks - saves no more than given tracks
I will fix comments and author notes

davidrohr:
> Missing settings are already there:
> eve-hostname - if set then workflows on computers which hostname do not match will do nothing
> number-of_tracks - saves no more than given tracks

ok, sorry, I didn't recheck the code since you didn't comment on it. Looks good. Thx

davidrohr (dismissed):
looks good now

TimoWilken:
Sorry for the build failures this evening. I've reset the checks to pending as there weren't any previous statuses to fall back on.

davidrohr:
@jmyrcha : Could you please fix this CI warning:
```
/sw/SOURCES/O2/6767-slc8_x86-64/0/EventVisualisation/Workflow/src/O2DPLDisplay.cxx:38:10: error: inclusion of deprecated C++ header 'limits.h'; consider using 'climits' instead [modernize-deprecated-headers]
```

jmyrcha:
is it possible to run fullCl check locally?
If so - how?

davidrohr:
> is it possible to run fullCl check locally?
> If so - how?

- Pull the docker container alisw/slc8-gpu-builder
- In the container, export `ALIBUILD_O2_TESTS=1`
- Then use aliBuild to compile O2FullCI inside the container.

jmyrcha:
I have pull container, but then:
docker run -d alisw/slc8-gpu-builder
07175ddb9680d5e85d176aa7eab3ef4054be4709c0f0cf151488c36ec37ee98d
docker ps -a
CONTAINER ID   IMAGE                                        COMMAND                  CREATED          STATUS                      PORTS     NAMES
07175ddb9680   alisw/slc8-gpu-builder                       "/bin/sh"                19 seconds ago   Exited (0) 18 seconds ago             tender_hamilton
bf6d3bce64a1   mcr.microsoft.com/mssql/server:2019-latest   "/opt/mssql/bin/perm…"   55 minutes ago   Exited (0) 44 minutes ago             sqlserver
docker exec -it tender_hamilton "bash"
Error response from daemon: Container 07175ddb9680d5e85d176aa7eab3ef4054be4709c0f0cf151488c36ec37ee98d is not running
------------------------------------
how should I run the image? is there any user I should use/configure (passwords?) I guess I should clone my repo inside and build? aliBuild O2fullCl?

jmyrcha:
sorry for bold in comment - it was probably after copy from console

davidrohr:
Hi @jmyrcha : I am no docker expert, but I am usually doing `docker run -ti --shm-size 32G alisw/slc8-gpu-builder`.

PR #6768, opened 2021-07-28
[CPV] Fix CPV calibration objects

sevdokim:
Added protection against accessing non-existing addresses in o2::cpv::CalibParams,  o2::cpv::BadChannelMap and o2::cpv::Pedestals.

### DIFF
--- a/DataFormats/Detectors/CPV/include/DataFormatsCPV/BadChannelMap.h
+++ b/DataFormats/Detectors/CPV/include/DataFormatsCPV/BadChannelMap.h
@@ -94,20 +94,28 @@ class BadChannelMap
   /// Only bad or warm cells are added to the container. In case
   /// the mask type is GOOD_CELL, the entry is removed from the
   /// container if present before, otherwise the cell is ignored.
-  void addBadChannel(unsigned short channelID) { mBadCells.set(channelID); } //set bit to true
+  void addBadChannel(unsigned short channelID)
+  {
+    if (channelID < NCHANNELS)
+      mBadCells.set(channelID);
+  } //set bit to true
 
   /// \brief Mark channel as good
   /// \param channelID Absolute ID of the channel
   ///
   /// Setting channel as good.
-  void setChannelGood(unsigned short channelID) { mBadCells.set(channelID, false); }
+  void setChannelGood(unsigned short channelID)
+  {
+    if (channelID < NCHANNELS)
+      mBadCells.set(channelID, false);
+  }
 
   /// \brief Get the status of a certain cell
   /// \param channelID channel for which to obtain the channel status
   /// \return true if good channel
   ///
   /// Provide the mask status of a cell.
-  bool isChannelGood(unsigned short channelID) const { return !mBadCells.test(channelID); }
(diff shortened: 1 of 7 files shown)


kharlov (dismissed):

sevdokim:
Dear @shahor02, 

I added checks for valid addresses into calibration objects but probably this is too much checks (thanks @peressounko for pointing this out). Some checks are already present in the RawDecoder and RawWriter. Is it better to discard checks proposed in this PR?

shahor02:
Hi @sevdokim 
If you are sure non-existent channels are rejected upstream, then it is indeed better to not check them again. But where these channels some from? Are they produced by your firmware? 

In an case, the CI fails due to the missing {} after if.

sevdokim:
Hi @shahor02,

currently checks are actually done in Detectors/CPV/base/src/Geometry.cxx when calculating addresses from and to Hardware address. So I suppose that non-existent addresses shouldn't appear during reading or writing raw data. I think it's better to remove the checks from reading from calibration objects as this operation is quite often. However writing to calibration object is not happening so frequently so why not to let the check be in there, right?

shahor02:
Hi @sevdokim 
I want to understand where the non-existent channels may appear at all, Who produces them, your firmware?

sevdokim:
@shahor02, yes, sometimes non-existent channels were appearing due to firmware problems. I'm not sure that this is actually a problem of firmware because originally some FEE cards were designed to read 64 channels but we use only 48. Some times some of remaining 16 channels appears in raw data. Before my previous commit they weren't tracked and they become like non-existent channels in data flow which lead to many errors. Now I track and exclude such channels in RawDecoder, and in principle we will not see them anymore.

sevdokim:
Dear @shahor02,

I also want to undestand if I need @kharlov or @peressounko to approve the PR? 
It seems that their approval do nothing actually. Or not?

shahor02:
Thanks @sevdokim , then, if the FW cannot be fixed, I would introduce a check only in the beginning of the decoding, and would eliminate all others. Adding unnecessary protections might be even dangerous since will hide real bugs (apart from eventual overhead). If you suspect appearance of wrong channels beyond this FW problem, better to add an assert(...) (which will be compiled out in production) and compile w/o NDEBUG.

shahor02:
> I also want to undestand if I need @kharlov or @peressounko to approve the PR?
> It seems that their approval do nothing actually.

Technically, I can merge w/o approval, but it is better if a few detector experts crosscheck their code.

kharlov (approved):

shahor02:
Hi @sevdokim @kharlov 

Before merging: so what is the source and earliest point of rejection of invalid channels? If they come only from the real FW and can be rejected on the raw data decoding level (?), what is the reason to check them again in the calibration?

sevdokim:
Dear @shahor02, 

I just removed checks during the calibration. So the non-existent channels are checked now only during the decoding.
Checks for writing to calibration objects are present  because this operation is performed rarely and why not?

shahor02:
@sevdokim ok, merging, though I don't see the point of checking something with should not happen.

sevdokim:
@shahor02 All right! I'll reconsider this behaviour when will work on calibration procedure. For now I'm not sure if those checks are implemented to the CalibDevices.

Thank you for guiding!

PR #6769, opened 2021-07-28
DPL Analysis: simplify code in spawner / indexer

ktf:


ktf:
@aalkin I guess this is  actually what you wanted to do?

aalkin:
> @aalkin I guess this is actually what you wanted to do?

Indeed, I needed to extract origin and description from an `InputSpec`.

PR #6770, opened 2021-07-29
Add InputRecord::getFirstValid() and use it instead of getByPos(0)

shahor02:
Workflows which need to access DataHeader members common for all inputs (e.g. tfCounter, firstTForbit)
should use InputRecord::getFirstValid() instead of getByPos(0) used so far, since the particular input is not guaranteed
to be valid.


shahor02:
@TimoWilken the same here, is there a problem with fullCl?

TimoWilken:
@shahor02 fixed now, sorry. There was a fullCI status reported for this commit at 2021-07-29T09:55:08Z that was green, so I restored it to that.

shahor02:
@sawenzel The fullCI has failed on simulation (seg.fault, it seems on the exit from the merger).
Merging since before this test was passed (also locally).

PR #6772, opened 2021-07-29
[WIP][EMCAL-660, EMCAL-703] Add option to auto-configure subspecification

mfasDa:
In case of auto-configure mode the subspecification
is determined based on the first header in the
payload using the ID of the FLP determined via the
FEE ID. The output specs are defined with wildcard
subspecification according to the discussion in
https://alice-talk.web.cern.ch/t/flp-specific-subspecifications/1035/6.
In case of empty timeframe empty cell containers are
sent to the deadbeaf subspecification which must be
caught by receiver processors.

### DIFF
--- a/Detectors/EMCAL/base/src/Mapper.cxx
+++ b/Detectors/EMCAL/base/src/Mapper.cxx
@@ -131,6 +131,17 @@ int MappingHandler::getFEEForChannelInDDL(int ddl, int channelFEC, int branch)
   return fecID;
 }
 
+int MappingHandler::getFLPIndex(int ddl)
+{
+  if ((ddl >= 0 && ddl <= 23) || ddl == 44) {
+    return 146; // EMCAL Links (44 - STU) -> FLP 146
+  }
+  if ((ddl >= 24 && ddl <= 38) || ddl == 45) {
+    return 147; // DCAL Links (45 - STU) -> FLP 147
+  }
+  throw DDLInvalid(ddl);
+}
+
 std::ostream& o2::emcal::operator<<(std::ostream& stream, const Mapper::ChannelID& channel)
 {
   stream << "Row " << static_cast<int>(channel.mRow) << ", Column " << static_cast<int>(channel.mColumn) << ", type " << o2::emcal::channelTypeToString(channel.mChannelType);
--- a/Detectors/EMCAL/base/include/EMCALBase/Mapper.h
+++ b/Detectors/EMCAL/base/include/EMCALBase/Mapper.h
@@ -345,6 +345,13 @@ class MappingHandler
   /// \param branch Branch index (0 or 1) in DDL
   int getFEEForChannelInDDL(int dll, int channelFEC, int branch);
 
+  /// \brief Get index of the FLP based on the DDL ID
+  /// \param ddl Absolute DDL index
+  /// \return FLP index
+  ///
+  /// Current DDL - FLP indexing can be found in EMCAL-660
+  int getFLPIndex(int ddl);
+
  private:
   std::array<Mapper, 4> mMappings; ///< Mapping container
 
--- a/Detectors/EMCAL/workflow/include/EMCALWorkflow/RawToCellConverterSpec.h
+++ b/Detectors/EMCAL/workflow/include/EMCALWorkflow/RawToCellConverterSpec.h
@@ -37,9 +37,8 @@ namespace reco_workflow
 class RawToCellConverterSpec : public framework::Task
 {
  public:
-  /// \brief Constructor
-  /// \param subspecification Output subspecification for parallel running on multiple nodes
-  RawToCellConverterSpec(int subspecification) : framework::Task(), mSubspecification(subspecification){};
+  /// \brief Default constructor
+  RawToCellConverterSpec() = default;
 
   /// \brief Destructor
   ~RawToCellConverterSpec() override;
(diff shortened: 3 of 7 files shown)


mfasDa:
Hi @shahor02, 

I have implemented the setting of the subspecification according to our discussion in https://alice.its.cern.ch/jira/browse/EMCAL-703. Could you please have a look if it is fine like this?

Detectors/EMCAL/base/src/Mapper.cxx
```diff
@@ -131,6 +131,15 @@ int MappingHandler::getFEEForChannelInDDL(int ddl, int channelFEC, int branch)
   return fecID;
 }
 
+int MappingHandler::getFLPIndex(int ddl)
+{
+  if ((ddl >= 0 && ddl <= 23) || ddl == 44)
+    return 146; // EMCAL Links (44 - STU) -> FLP 146
+  if ((ddl >= 24 && ddl <= 38) || ddl == 45)
+    return 147; // DCAL Links (45 - STU) -> FLP 147
+  throw DDLInvalid(ddl);
```
shahor02: use {} after if, this will not pass the checker

shahor02 (dismissed):
@mfasDa looks correct, just the usual comment on missing braces

Detectors/EMCAL/base/src/Mapper.cxx
mfasDa: OK, brackets are fixed now

shahor02:
@TimoWilken the fullCL failed to start, could you please trigger it?

TimoWilken:
@shahor02 the fullCI failed previously, but subsequent failures wiped the log. The previous error was:

```
[ERROR] invalid workflow in o2-emcal-entropy-encoder-workflow: No matching output found for EMC/CELLS/0 as requested by data processor "emcal-entropy-encoder". Candidates:
[ERROR] -EMC/CTFDATA/0
[ERROR] -EMC/RAWDATA
[ERROR] -EMC/CELLS
[ERROR] -EMC/CELLSTRGR
[ERROR] -EMC/DECODERERR
[snip long list of other candidate names not starting with EMC]
```

shahor02:
Thanks @TimoWilken !
@mfasDa, the the input specs of all EMC cell consumers will need to be wildcarded, and, which is more tricky, the iterators in the CTFHelper will need to be adapted to multiple inputs.
Also, the cell writer will need to be fixed: since writing the root file does need to be terribly performant, I think the best would be to precede it with a device which merges the cells to single output with subspec 0... Given the amount of changes this PR will trigger, perhaps it is worth first write such a device and benchmark the overhead of producing a fully merged output. If it appears to be small, one could do the merging right after the TFbuilder and avoid all these complications. But, first of all, are you confident that your FLPs will be able to do the raw to cells conversion, which would indeed justify these complications?

mfasDa:
Hi @shahor02, as you recall we anyhow want to run a filter component rejecting cells which have no chance to end up in a cluster. As this has to iterate over all subevents it also can do the merging, and as such run in front of the CTF coder or cell writer. This class will need a few iterations but for the beginning it can run with a low-E filter (for current tests at P2 we will need to set E to 0 since we only process noise events) and therefore serve as merger.

shahor02:
@mfasDa well, the filtering of cells could be done only at the entropy compression stage, w/o actually creating a new cells output. 
But I am not sure that doing on the fly merging in a few places is significantly more economical that creating once a merged input.
Given that this PR will break many things, I would put in on hold and introduce a merging device (which optionally could also do filtering, but I guess the QC will need to see the complete data anyway?)

mfasDa:
@shahor02 First of all yes, the PR will need to be kept on hold and the checkpoint DPL on FLP needs to be postponed. More about that in dedicated tickets in JIRA (notably EMCAL-703 for the subspecification handling and EMCAL-615 for the filter). 

Concerning QC: The cell level QC needs to see the full event for particularly for occupancy monitoring, however the relevant changes are already implemented in https://github.com/AliceO2Group/QualityControl/pull/785 and merged, and should be straight forward to be ported also to other components. Only I am thinking about a generalised handler here iterating over all subverts instead of each processor having to implement it themselves and therefore copy/paste the code. The iteration over subevents is implemented with a variable amount of subevents which can be 1 in case the subevent contains input from all links and represents therefore the entire event (i.e. when raw->cell conversion offline or on the EPN).

A more tricky issue could be the clusterizer as in case the clusterizer gets multiple subevents as input the cell index container would need the cell subspecification as additional dimension. As we have dedicated FLPs for EMCAL and DCAL and they are distinct in clusterization it would be generally possible to run the clusterizer separate for the subspecification and push out clusters for each subspecification separately.

shahor02:
@mfasDa What I meant is create a separate DPL workflow with single merger device which would sit right after the TFBuilder, take multiple inputs and output merged one with subspec 0. Then all current devices would remain unchanged.
Just make sure that the FLPs don't send the output with subspec = 0 to not confuse it with merged one: if you mapping produces 0 subspec for one of FLPs, augment it by some constant so that it is different from 0.

mfasDa:
@shahor02 This is what we initially planned (referred to internally as EventBuilder - EMCAL-647). But in https://github.com/AliceO2Group/AliceO2/pull/6725 you suggested that we might be able to get rid of this component and merge on the fly, because this would be computing wise more efficient. In any case a merger component should be trivial to implement. We just need to figure out whether it should be separate processor (and eventually be implemented in a separate workflow) or whether it can be part of the filter (which however will at some point need to subscribe to the clusterizer - which raises issues as mentioned in my previous post).

mfasDa:
FLPs send to subspecification 146 for the EMCAL FLP and 147 for the DCAL FLP, so the FLP ID is used as subspecification based on the FEE-FLP mapping added to the mapper and logged in EMCAL-650. 0 is the default value which is supposed to be only used if we have a single instance of the raw to cell converter, either on the EPN (not affecting the pipelines) or offline.

shahor02:
> But in #6725 you suggested that we might be able to get rid of this component and merge on the fly,

Yes, I've suggested it, but now I see that this will trigger quite some changes in many places, which is probably not justified...

PR #6773, opened 2021-07-29
Improve GPU benchmark

mconcas:
TODO:
- [x] Replace single block kernels with kernels launches with 1 block
- [x] Change copyKernel to read on the same chunk but from different halves
- [x] Remove multiple streams approach and use only one also for concurrency (Max is 8)
- [x] Show result in bandwidth
- [x] Code refactoring to generalize test execution call
- [x] Add support for `int4`
- [x] Add support for targeted chunks


mconcas:
@davidrohr o2-cs8 seems stuck...

PR #6774, opened 2021-07-29
Script to update data model files

jgrosseo:


jgrosseo:
Does not touch code. Merging.

PR #6776, opened 2021-07-29
ITS tracker: Add smoothing class

mconcas:
@shahor02 : This PR is rebased on #6431 but it will likely be merged after.
Still have to do the tests, I'll keep you updated.


mconcas:
Hi @shahor02, I am starting again at looking into fake clusters.
The attached plot is produced with `CheckTrackerFakeClusters.C` (added in this PR).
fakeClusters

In the first row there are the distributions of fakes vs track length (no constraints on starting layer). Info on _fakiness_ is fetched via `TrackITS::isFakeOnLayer()` if `TrackITS::hasHitOnLayer()` is true.
Second row is same, but filtering on tracks with only one fake cluster.

Data is also here, it's 1K pp events: https://cernbox.cern.ch/index.php/s/qgPHswoFoSx2EB5
Over a total of 16K reconstructed tracjks, ~200 are `isFake()`.

Please let me know what further information I can provide.

trackEff

shahor02:
@mconcas thanks! Is this with or w/o fix of @mpuccio which eliminates ~10% pt-independent fake contamination in min.bias pbpb?
When checking for fakes, you don't check if there is a correct cluster available or not, right?

mconcas:
> @mconcas thanks! Is this with or w/o fix of @mpuccio which eliminates ~10% pt-independent fake contamination in min.bias pbpb?

This is w/o Max's fix. It's just using mCTruth containers, differently from what we have upstream atm.

> When checking for fakes, you don't check if there is a correct cluster available or not, right?

No, that was once done on a specific study on 7-clusters long tracks where we discovered they were not present. I can check for shorter tracks.

shahor02:
@max is going to merge his PR, for cleanness, I would prefer checking the fakes once it is merged

mconcas:
> @max is going to merge his PR, for cleanness, I would prefer checking the fakes once it is merged

Sounds reasonable

mconcas:
Hi @shahor02
I rebased locally on latest TF-based tracker.
Before force-pushing I just post the result of running on the same data now (using `run_trac_ca_its.C`):

__Efficiencies:__
- Total number of tracks: 20890
	- Total number of tracks not corresponding to particles: 88 (0.421254%)
	- Total number of fakes: 2196 (10.5122%)
	- Total number of good: 18606 (89.0665%)

Efficiencies

__Fake clusters:__
fakeClusters

@mpuccio: does this look reasonable? Nothing should've changed as the smoother code addition is non operational for the moment. 
 
If this is ok for the two of you I'll restart from here.

shahor02:
Hi @mconcas 

Qualitatively they look reasonable to me, the numbers are difficult to compare to previous set due to different pT coverage.
Do you have macro to produce these histos? Later I may play with them if find some time.
Anyway, please remove the `Drafts` tag to trigger the CI.

mconcas:
> Hi @mconcas
> 
> Qualitatively they look reasonable to me, the numbers are difficult to compare to previous set due to different pT coverage.
> Do you have macro to produce these histos? Later I may play with them if find some time.
> Anyway, please remove the `Drafts` tag to trigger the CI.

Yes, the macro is committed and it's called `CheckTrackerFakeClusters.C` I'll then proceed with the force pushing

mconcas:
@shahor02 : now check on fake clusters are improved and merged in `CheckTracksCA.C` macro. It is now possible to see for fake cluster distributions whether there is the correct one (blue) or not (red).

E.g. pp:
fakeClusters_pp

and PbPb:
fakeClusters_PbPb

If @mpuccio agrees with the edits to the macro I think this PR can go through, so that he can in parallel continue with his work while I can do more detailed studies on the fakes.

Detectors/ITSMFT/ITS/workflow/src/TrackerSpec.cxx
```diff
@@ -252,17 +252,14 @@ void TrackerDPL::run(ProcessingContext& pc)
         auto& trc{tracks[iTrk]};
         trc.setFirstClusterEntry(allClusIdx.size()); // before adding tracks, create final cluster indices
         int ncl = trc.getNumberOfClusters(), nclf = 0;
-        uint8_t patt = 0;
```
mconcas, line 255: @shahor02 please notice that I removed this part as the pattern is already set during the tracking step: https://github.com/AliceO2Group/AliceO2/blob/0dad46d7a140af2e44fbd97b496d58377483b63e/Detectors/ITSMFT/ITS/tracking/src/Tracker.cxx#L287

These line were actually resetting the part of the `mPattern` reserved for MC information.

mpuccio:
Unless there are other considerations, I would merge this PR as soon as tests are done, we will need this for the next development of the "track finding with holes".

shahor02:
@mconcas thanks! I am slightly confused by the histos: they show majority fake, even for 7 hit tracks. Why?

mconcas:
> @mconcas thanks! I am slightly confused by the histos: they show majority fake, even for 7 hit tracks. Why?

Hi, I am not sure I understood the question. The histos are stacked and both blue and red represent tracks with fake clusters. In red case there is not the corresponding correct cluster, in blue case yes.

shahor02:
@mconcas ah, ok, I did not realize only fakes contribute to both histos. Fine, thanks, will merge once CI passed

shahor02 (approved):

davidrohr:
mac os CI seems stuck since few days, merging

PR #6777, opened 2021-07-30
[ALICE3 - FT3] Enable beampipe fix for custom layout file

rpezzi:
@mconcas , this PR enables ALICE3 beampipe for the FT3 with custom geometry from external file.

### DIFF
--- a/Detectors/Upgrades/ALICE3/FT3/simulation/src/Detector.cxx
+++ b/Detectors/Upgrades/ALICE3/FT3/simulation/src/Detector.cxx
@@ -517,13 +517,24 @@ void Detector::createGeometry()
     }
   }
 
-  // Unsupported with beampipe
   if (mLayers.size() == 1) { // All layers registered at mLayers[0], used when building from file
     LOG(INFO) << "Creating FT3 layers:";
-    for (Int_t iLayer = 0; iLayer < mLayers[0].size(); iLayer++) {
-      mLayers[0][iLayer].createLayer(volFT3);
+    if (A3IPvac) {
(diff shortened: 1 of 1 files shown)


Detectors/Upgrades/ALICE3/FT3/simulation/src/Detector.cxx
```diff
@@ -517,13 +517,24 @@ void Detector::createGeometry()
...
   }
 
-  // Unsupported with beampipe
   if (mLayers.size() == 1) { // All layers registered at mLayers[0], used when building from file
     LOG(INFO) << "Creating FT3 layers:";
-    for (Int_t iLayer = 0; iLayer < mLayers[0].size(); iLayer++) {
-      mLayers[0][iLayer].createLayer(volFT3);
+    if (A3IPvac) {
+      for (Int_t iLayer = 0; iLayer < mLayers[0].size(); iLayer++) {
+        if (std::abs(mLayers[0][iLayer].getZ()) < 25) {
```
mconcas, line 524: Overall seems reasonable. I just wonder if you don't want an extra check on the disks' radii with cusom configuration to avoid overlaps. Or this is done somewhere else?

rpezzi: Nice feature, but I think does not worth the trouble. At this point the user should ensure the geometry is compatible with other elements in the simulation. In any case the already enabled pipe+telescope is also not failproof given that telescope  configurable parameters (configKeyValues) can also cross the limits of the pipe.

mconcas (approved):

PR #6779, opened 2021-07-30
make Cell HG/LG consistent with Mapping

peressounko:


shahor02 (approved):

peressounko:
Dear @shahor02, almost all tests passed - could you force merging?

shahor02:
ok, the fullCI died or log is wrong, but this should not be related to your PR. Merging as it is trivial.

PR #6783, opened 2021-07-30
Return ptr when adding histogram to registry

mpuccio:
Especially useful when setting labels for the axis

### DIFF
--- a/Framework/Core/include/Framework/HistogramRegistry.h
+++ b/Framework/Core/include/Framework/HistogramRegistry.h
@@ -90,9 +90,9 @@ class HistogramRegistry
   HistogramRegistry(char const* const name, std::vector<HistogramSpec> histSpecs = {}, OutputObjHandlingPolicy policy = OutputObjHandlingPolicy::AnalysisObject, bool sortHistos = true, bool createRegistryDir = false);
 
   // functions to add histograms to the registry
-  void add(const HistogramSpec& histSpec);
-  void add(char const* const name, char const* const title, const HistogramConfigSpec& histConfigSpec, bool callSumw2 = false);
-  void add(char const* const name, char const* const title, HistType histType, std::vector<AxisSpec> axes, bool callSumw2 = false);
+  HistPtr add(const HistogramSpec& histSpec);
+  HistPtr add(char const* const name, char const* const title, const HistogramConfigSpec& histConfigSpec, bool callSumw2 = false);
+  HistPtr add(char const* const name, char const* const title, HistType histType, std::vector<AxisSpec> axes, bool callSumw2 = false);
   void addClone(const std::string& source, const std::string& target);
 
   // function to query if name is already in use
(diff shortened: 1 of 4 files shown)


Framework/Core/include/Framework/HistogramRegistry.h
```diff
@@ -345,7 +345,7 @@ auto& HistogramRegistry::operator()(const HistName& histName)
 }
 
 template <typename T>
-void HistogramRegistry::insertClone(const HistName& histName, const std::shared_ptr<T>& originalHist)
+HistPtr HistogramRegistry::insertClone(const HistName& histName, const std::shared_ptr<T>& originalHist)
```
mario-krueger, line 348: This function is able also to clone whole directory structures. 
The pointer that will be returned here in this case would therefore be unpredictable (it will be the last histogram in the registry that matches the request).
My suggestion would be to keep this as a void function for now and later we can split it into a version for cloning single histograms that returns the corresponding pointer and a version that we call `addGroupClone` or similar which will then be a `void` function.

Framework/Core/src/HistogramRegistry.cxx
```diff
@@ -95,24 +96,25 @@ void HistogramRegistry::validateHistName(const char* name, const uint32_t hash)
...
-void HistogramRegistry::add(char const* const name, char const* const title, HistType histType, std::vector<AxisSpec> axes, bool callSumw2)
+HistPtr HistogramRegistry::add(char const* const name, char const* const title, HistType histType, std::vector<AxisSpec> axes, bool callSumw2)
 {
-  insert({name, title, {histType, axes}, callSumw2});
+  return insert({name, title, {histType, axes}, callSumw2});
 }
 
 // store a copy of an existing histogram (or group of histograms) under a different name
-void HistogramRegistry::addClone(const std::string& source, const std::string& target)
+HistPtr HistogramRegistry::addClone(const std::string& source, const std::string& target)
```
mario-krueger: This function is able also to clone whole directory structures. 
The pointer that will be returned here in this case would therefore be unpredictable (it will be the last histogram in the registry that matches the request).
My suggestion would be to keep this as a void function for now and later we can split it into a version for cloning single histograms that returns the corresponding pointer and a version that we call `addGroupClone` or similar which will then be a `void` function.

iarsene: Hi @mpuccio and @mario-krueger,
Hi think that Mario's suggestion is reasonable. Can you solve this so we can merge?

mario-krueger: we can also merge it as it is and I will then commit the follow-up changes

mpuccio: Yeah, sorry I got back from vacation today! I am fixing it right now :)

iarsene:
The macOS test failure is unrelated, merging

iarsene (approved):

PR #6784, opened 2021-07-30
set global reader default tracks/clusters request to none

shahor02:
@davidrohr requesting all tracks and clusters by default is confusing, e.g. if user wants only TPC tracks, it will also try to load all tracks. Setting to `none`. If needed, `--track-types all` can be requested.


shahor02:
trivial, merging

PR #6787, opened 2021-07-31
DPL Analysis: fix amending copy instead of an actual entry

aalkin:
@victor-gonzalez this should fix the issue you've encountered


aalkin:
@jgrosseo this fix is needed for process configurables to work properly for some complex workflows.

victor-gonzalez:
It is fixed now!!!

PR #6788, opened 2021-08-01
Modifications for extendend async workflow 

shahor02:
* Option `--strict-matching` (def: false) added to `o2-trd-global-tracking` and
`o2-tof-matcher-workflow` to impose the preliminary strict matching mode of the
extended workflow. If true, it will make sure that trd-matching gets as input
TPC tracks only while `o2-tof-matcher-workflow` TPC and/or TPC-TRD.
In this mode the outputs of these matcher will have `subspecs=1`, while
the `subspec=0` is reserved for the final matches.

* The TOF and TRD tracks/matches readers and writers modified to accept inputs
and provide outputs with this `subspec=1` when strict mode is selected.

* The `o2-tpcits-match-workflow` will automatically impose this strict mode on
TRD and TOF readers if TPC-TRD or TPC-TOF or TPC-TRD-TOF inputs are requested.

* DataRequest class for the RecoContainer got `setMatchingInputStrict()` method to
signal that TRD and TOF inputs has to be received with `subspec=1`.

* `InputHelper::addInputSpecs` got extra bool argument `subSpecStrict` to request
reading TRD and TOF inputs with `subspec=1` when needed.

* Set meaningful names for different TOF matching outputs.

* Added to `RecoContainer::getSingleDetectorRefs(..,)` filling of TRD-related slots.

* RecoContainer getter TOF matches got distinct names of input sources.

* GPUDataTypes and GPUWorkflowHelper will handle separately TOF mathches to
different input track sources.

* Add to TRDGlobalTrackReaderSpec option to load the material LUT, so far it was using TGeo.
----------------

* Temporary hack to avoid wrongly imposed refit time.
Will be removed once https://github.com/AliceO2Group/AliceO2/pull/6540#issuecomment-880060760 fixed

### DIFF
--- a/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
+++ b/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
@@ -20,6 +20,7 @@
 #include "ReconstructionDataFormats/GlobalTrackAccessor.h"
 #include "CommonDataFormat/RangeReference.h"
 #include "ReconstructionDataFormats/GlobalTrackID.h"
+#include "ReconstructionDataFormats/MatchingType.h"
 #include "CommonDataFormat/AbstractRefAccessor.h"
 #include "SimulationDataFormat/MCCompLabel.h"
 #include "SimulationDataFormat/MCTruthContainer.h"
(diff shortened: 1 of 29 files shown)


shahor02:
@davidrohr could you please check https://github.com/AliceO2Group/AliceO2/pull/6788/files#diff-798fc5ddba1d597d39697215f8cdbadf808da79945373ca698fe270eaf68392e and https://github.com/AliceO2Group/AliceO2/pull/6788/files#diff-232b4194900e1b0793e5688f43ea4c766641be6cd445f9577d77af3c9250bca3.

@noferini note that I've renamed your option `high-purity` to `strict-matching` (to have it uniform across this stage workflows)

davidrohr (approved):

PR #6791, opened 2021-08-02
DPL: Return the result code if noCatch is enabled

knopers8:
This should silence the warning at https://github.com/AliceO2Group/AliceO2/pull/6419 and bring back the expected (I suppose) behaviour.


ktf:
fullCI resolved by itself. Merging.

PR #6794, opened 2021-08-02
[EMCAL-630] Skip TRU and LEDMON cells in raw->cell conversion

mfasDa:
TRU and LEDMON need to be discarded in the
raw to cell conversion as they have different geometry
mapping than FEC cells and then raw fitting procedure
cannot be applied to them.

### DIFF
--- a/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
@@ -66,6 +66,8 @@ void RawToCellConverterSpec::init(framework::InitContext& ctx)
     mRawFitter = std::unique_ptr<CaloRawFitter>(new o2::emcal::CaloRawFitterGamma2);
   }
 
+  mPrintTrailer = ctx.options().get<bool>("printtrailer");
+
   mMaxErrorMessages = ctx.options().get<int>("maxmessage");
   LOG(INFO) << "Suppressing error messages after " << mMaxErrorMessages << " messages";
 
@@ -188,7 +190,11 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
         continue;
       }
 
-      LOG(DEBUG) << decoder.getRCUTrailer();
+      if (mPrintTrailer) {
+        // Can become very verbose, therefore must be switched on explicitly in addition
+        // to high debug level
+        LOG(DEBUG4) << decoder.getRCUTrailer();
+      }
       // Apply zero suppression only in case it was enabled
       mRawFitter->setIsZeroSuppressed(decoder.getRCUTrailer().hasZeroSuppression());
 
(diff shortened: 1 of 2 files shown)


mfasDa:
TRU decoding is currently be worked on. LEDMON cells are only used in the LED DA, here we need to discuss further what to do with them. For the ongoing tests we need the changes from this commit as several SMs have the LEDMON included.

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -123,6 +123,7 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
       auto found = cellBuffer.find(currentIR);
       if (found == cellBuffer.end()) {
         currentCellContainer = std::make_shared<std::vector<Cell>>();
+        std::cout << "Setup cell container with " << currentCellContainer->size() << " entries" << std::endl;
```
shahor02: please use LOG instead of the std::cout

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -244,6 +305,7 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
   for (auto [bc, cells] : cellBuffer) {
     mOutputTriggerRecords.emplace_back(bc, triggerBuffer[bc], mOutputCells.size(), cells->size());
     if (cells->size()) {
+      std::cout << "Event has " << cells->size() << " cells" << std::endl;
```
shahor02: idem

shahor02 (changes requested):
@mfasDa see cosmetic comments below. 
What is the reason to check for the invalid channels (where they come from)?

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
mfasDa: Ooops, sorry, these are leftover from debugging, will remove them again ...

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
mfasDa: See above

mfasDa:
@shahor02 The SRU payload contains 4 channel types

- High gain
- Low gain
- TRU (L0 trigger)
- LEDMON

The raw fitter can only work handle FEC cells (high- and low-gain). For L0 trigger a separate reconstruction is being worked on. The problem is however the LEDMON cells (dedicated sensors), which are only used in the LED DA. The mapping in the geometry only works for FEC cells, but we recently had them enabled and they lead to invalid cell IDs leading to a crash when processing. They need to be removed from the raw fitter. How we propagate then to the LED DA we will figure out later.

TRU cells we will reintroduce later once the porting is done, in a separate cell container (will be only used for the trigger QC).

mfasDa:
@shahor02 The couts are removed (second one changed to LOG(DEBUG). Could you please merge once the CI tests pass?

shahor02:
Merging although the fullCI crashed in the gpu-reconstruction. Could not reproduce the crash locally but it should be unrelated to this PR anyway.

PR #6797, opened 2021-08-02
[EMCAL-565]: Added conversion tools to convert ROOTTH1 and TH2 histog…

hjbossi:
…ras to corresponding BOOST histogrgrams.

### DIFF
--- a/Common/Utils/CMakeLists.txt
+++ b/Common/Utils/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+Common/Utils/CMakeLists.txt# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
 # See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
 # All rights not expressly granted are reserved.
 #
@@ -25,7 +25,6 @@ o2_target_root_dictionary(CommonUtils
                                   include/CommonUtils/TreeStreamRedirector.h
                                   include/CommonUtils/RootChain.h
                                   include/CommonUtils/BoostSerializer.h
-                                  include/CommonUtils/BoostHistogramUtils.h
                                   include/CommonUtils/ShmManager.h
                                   include/CommonUtils/RngHelper.h
                                   include/CommonUtils/StringUtils.h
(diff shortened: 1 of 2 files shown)


shahor02:
Hi @TimoWilken lot of tests (loading macros in interpreted mode) are failing with 
````
root.exe: /mnt/mesos/sandbox/sandbox/jenkins/workspace/DailyBuilds/DailyO2-slc7-py3/daily-tags.lohzINJ6pA/SOURCES/ROOT/v6-24-02/v6-24-02/interpreter/llvm/src/tools/clang/lib/CodeGen/ModuleBuilder.cpp:146: llvm::Module* clang::CodeGeneratorImpl::StartModule(llvm::StringRef, llvm::LLVMContext&, const clang::CodeGenOptions&): Assertion `OldBuilder->DeferredDeclsToEmit.empty() && "Should have emitted all decls deferred to emit."' failed.
````
Any idea why?

**Comment by TimoWilken** [comment]
I don't know, sorry!

**Review by shahor02 — CHANGES_REQUESTED** [review]
Hi @hjbossi 

It looks like the CI failures are indeed related to this PR: most probably boost headers you have added are polluting some namespace which makes the ROOT interpreter to fail. We already had such problem before. 
I noticed that you are adding the `BoostHistogramUtils.h` to https://github.com/AliceO2Group/AliceO2/blob/8058d0fc8908c57f5fb15cd5ca5bbab034322215/Common/Utils/CMakeLists.txt#L28 while the utility classes themselves have no ClassDef.
Could you remove it from the `o2_target_root_dictionary(CommonUtils` of CMakeLists ?

**Comment by hjbossi** [comment]
Hi @shahor02 - Thanks for looking into this! I have made the change you requested, which is included in the latest commit. We will see if this fixes the CI failures. 

Cheers, 

Hannah

File: Common/Utils/CMakeLists.txt [hunk 1]
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+Common/Utils/CMakeLists.txt# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
```
shahor02: Hi @hjbossi : there is a stray line here

hjbossi: Okay, this should be resolved in the most recent commit!

shahor02 (approved):

PR #6798, opened 2021-08-02
TPC: generalize cluster QC, speed up Kr cluster finder, unrelated fixes and cleanup

wiechula:
* move KrCluster to DataFormatsTPC
* modify cluster QC to accept ClusterNative and KrCluster as input


wiechula:
@davidrohr, can this be merged? I think the only failing test in macos is unrelated.
Note, I added two getters in ClusterNative, hope that is fine.

shahor02 (approved):

PR #6799, opened 2021-08-03
speed up and fix up trd raw data reading.

bazinski:
fix parsing of zero suppressed digits.
increase verbose output
fix bitcounting error causing slow readout
other speed ups, 7 hours per timeframe now down to .7ms per timeframe.
byteswap the adc bit pattern word
fix the  DigitMCMADCMask it was reversed.

attempt to jump the non tracklet data at the beginning of the tracklet parse block, partially successful.
erroneous additional digits are marked with EEE in info output

change datareader to write out triggers with the digits (temporary most likely)

This could well change depending on the testing tomorrow.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
@@ -307,10 +307,10 @@ struct DigitMCMADCMask {
   union {
     uint32_t word; //MCM ADC MASK header
     struct {
-      uint32_t j : 4; // unused always 0xc
+      uint32_t j : 4; // 0xc
       uint32_t adcmask : 21;
-      uint32_t c : 5; // unused always 0x1f
-      uint32_t n : 2; // unused always 0x3
+      uint32_t c : 5; // ~(number of bits set in adcmask)
+      uint32_t n : 2; // 0b01
     } __attribute__((__packed__));
   };
 };
@@ -440,6 +440,8 @@ std::ostream& operator<<(std::ostream& stream, const HalfCRUHeader& halfcru);
 bool trackletMCMHeaderSanityCheck(o2::trd::TrackletMCMHeader& header);
 bool trackletHCHeaderSanityCheck(o2::trd::TrackletHCHeader& header);
 bool digitMCMHeaderSanityCheck(o2::trd::DigitMCMHeader* header);
+bool digitMCMADCMaskSanityCheck(o2::trd::DigitMCMADCMask& mask, int numberofbitsset);
+bool digitMCMWordSanityCheck(o2::trd::DigitMCMData* word, int adcchannel);
 void printDigitMCMHeader(o2::trd::DigitMCMHeader& header);
 void printDigitHCHeader(o2::trd::DigitHCHeader& header);
 DigitMCMADCMask buildBlankADCMask();
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
@@ -307,10 +307,10 @@ struct DigitMCMADCMask {
   union {
     uint32_t word; //MCM ADC MASK header
     struct {
-      uint32_t n : 2; // unused always 0x3
-      uint32_t c : 5; // unused always 0x1f
-      uint32_t adcmask : 21;
       uint32_t j : 4; // unused always 0xc
+      uint32_t adcmask : 21;
+      uint32_t c : 5; // unused always 0x1f
+      uint32_t n : 2; // unused always 0x3
     } __attribute__((__packed__));
   };
 };
--- a/DataFormats/Detectors/TRD/src/RawData.cxx
+++ b/DataFormats/Detectors/TRD/src/RawData.cxx
@@ -451,6 +451,8 @@ int nextmcmadc(unsigned int& bp, int channel)
   while (!(bp & m)) {
     m = m << 1;
     position++;
+    if (position > 31)
+      break;
   }
   bp &= ~(1UL << (position));
   return position;
--- a/DataFormats/Detectors/TRD/src/RawData.cxx
+++ b/DataFormats/Detectors/TRD/src/RawData.cxx
@@ -124,7 +124,7 @@ uint32_t getQFromRaw(const o2::trd::TrackletMCMHeader* header, const o2::trd::Tr
       qa = header->pid2;
       break;
     default:
-      LOG(warn) << " unknown trackletindex to getQFromRaw : " << pidindex;
+      LOG(warn) << " unknown trackletindex of " << trackletindex << " to getQFromRaw : " << pidindex;
       break;
   }
   //qa is 6bits of Q2 and 2 bits of Q1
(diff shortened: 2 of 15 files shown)


DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
```diff
@@ -307,10 +307,10 @@ struct DigitMCMADCMask {
   union {
     uint32_t word; //MCM ADC MASK header
     struct {
-      uint32_t n : 2; // unused always 0x3
-      uint32_t c : 5; // unused always 0x1f
-      uint32_t adcmask : 21;
       uint32_t j : 4; // unused always 0xc
+      uint32_t adcmask : 21;
+      uint32_t c : 5; // unused always 0x1f
```
tdietel: Is the comment correct? I thought it's the inverted number of ADC channels with data.

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -440,14 +444,16 @@ void setNumberOfTrackletsInHeader(o2::trd::TrackletMCMHeader& header, int number
...
+    return 22;
+  int position = channel;
+  int m = 1 << channel;
+  while (!(bp & m)) {
+    m = m << 1;
+    position++;
   }
-  bp &= ~(1UL << (channel));
-  return channel; // zero based
+  bp &= ~(1UL << (position));
```
tdietel, line 509: Do you modify the bitpattern in the function? 
I don't see why this is necessary, if you tell the function what the current channel is.

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -440,14 +444,16 @@ void setNumberOfTrackletsInHeader(o2::trd::TrackletMCMHeader& header, int number
...
-  while ((bp & (1 << channel)) == 0) {
-    channel++;
-    if (channel == 21) {
-      break;
-    }
+  if (bp == 0)
+    return 22;
+  int position = channel;
+  int m = 1 << channel;
+  while (!(bp & m)) {
```
tdietel, line 502: Does this code do the same as:
```suggestion
  for (int pos=channel+1; pos<22; pos++) {
    if ( bp & (1<<pos) ) {
      return pos;
    }
  }
  return 22;
      
```

tdietel: I am not sure what your code does once it's at (or rather beyond) the last bit.

Detectors/TRD/reconstruction/src/DigitsParser.cxx
```diff
@@ -166,21 +179,57 @@ int DigitsParser::Parse(bool verbose)
           LOG(info) << "state mcmheader and word : 0x" << std::hex << *word;
           printDigitMCMHeader(*mDigitMCMHeader);
         }
-        if (mDigitHCHeader->major == 4) {
+        if (mDigitHCHeader->major == 0x21) {
```
tdietel: The 0x21 does not seem to go together with the 4.

Detectors/TRD/reconstruction/src/DigitsParser.cxx
```diff
@@ -270,80 +343,77 @@ int DigitsParser::Parse(bool verbose)
             }
             if (digitwordcount == constants::TIMEBINS / 3) {
```
tdietel: Does this actually work if the number of timebins is not divisible by 3?

Detectors/TRD/reconstruction/src/DigitsParser.cxx
```diff
@@ -270,80 +343,77 @@ int DigitsParser::Parse(bool verbose)
...
               //write out adc value to vector
               //zero digittimebinoffset
-              if (mDigitHCHeader->major == 4) {
+              if (mDigitHCHeader->major & 0x20) {
                 //zero suppressed, so channel must be extracted from next available bit in adcmask
-                if (mDataVerbose) {
-                  LOG(info) << "adcmask: 0x" << std::hex << mADCMask << " and channel : " << std::dec << mChannel;
-                }
                 mChannel = nextmcmadc(mADCMask, mChannel);
+                if (mChannel == 21) {
```
tdietel: I thought I saw a 22 up there...

tdietel (commented):
It's still work in progress. I made some comments, but I'm happy to merge if it helps with progress.

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
bazinski: correct as to my understanding at time of writing, but no not correct.

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: current channel is simply used as a position to jump to instead of each time shifting through the whole word.

Detectors/TRD/reconstruction/src/DigitsParser.cxx
bazinski: no the 4 is wrong, its the 6th bit that defines zero suppression 0x20. Table 15.8 in tdp, 6th bit is set for zero supression.
Did not matter when i was only looking at simulated data as generation and reading were self consistent.

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: the newer one works properly, verified. unit tests to come.
the bit is fliped so that each time is simply looks for the next right most bit with a 1 in it, so at the end one has no bits and has a condition for finishing.

bazinski: To be clearer line 447,448 handles the case of no bits valid, it is possible to start on the left of the left most bit.
adding a termination at position 32;

Detectors/TRD/reconstruction/src/DigitsParser.cxx
bazinski: it works in the sense that it wont write a digit, but carry on.

Detectors/TRD/reconstruction/src/DigitsParser.cxx
bazinski: 21 and 22, 22 gives logs the erroneous digit, and 21 should never get hit, there as a safety net should maybe change it to >21

Detectors/TRD/reconstruction/src/DigitsParser.cxx
bazinski: just above that there is a check for digitwordcount > timebins/3 with a fatal log message.

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
bazinski: now fixed.

bazinski:
should i leave the list of updates or collapse them all into 1? this can then be merged, stats will come shortly in a seperate pr @tdietel

shahor02:
@bazinski I can squash them at merging, shall I already do this?

tdietel:
Squashing looks like a good idea. Shall we go ahead and merge?

shahor02 (approved):
squashing/merging

PR #6802, opened 2021-08-03
Write CTFs to <partion_id>-<run_id> dir if partition is available

shahor02:
The o2-ctf-writer-workflow will try to fetch the `environment_id` (a.k.a. partition ID) set by the ECS. If it is defined, the CTFs will be stored in the `<output_dir>/<environment_id>-<runNumber>` directory, where `<output_dir>` is an existing directory provided by `--output-dir option` (default: cwd). 
The sub-directory `<environment_id>-<runNumber>`
will be created if needed.

If the `environment_id` property was not defined or an option `--ignore-partition-run-dir` is provided, the CTF will be written directly to the `<output_dir>`.


shahor02:
o2 CI silently stopped in the middle of the tests, but the modifications are relevant only for the FullCI which was passed, merging.

PR #6805, opened 2021-08-04
TRD: Introduce strict matching mode + MC labels for tracks

martenole:
Hi @shahor02 
a simple flag is set for the TRD track after the track following in case a competitor hypothesis was found close by. The chi2 separation cut needs to be tuned. The TRD tracking can be run with:
`o2-trd-tracklet-transformer -b | o2-trd-global-tracking -b --track-sources TPC --strict-matching --configKeyValues "GPU_rec_trd.chi2SeparationCut=2;GPU_proc.trdNCandidates=3"`
Would this already be sufficient for the strict matching mode?

I see that I have in 10pp events already one tracklet which is attached to two tracks. I am looking further into that and for that I will try to add also the MC labels now to the tracking. This will come in another PR.

Cheers,
Ole

### DIFF
--- a/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
+++ b/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
@@ -387,6 +387,8 @@ struct RecoContainer {
   }
   auto getITSTPCTRDTracksMCLabels() const { return getSpan<o2::MCCompLabel>(GTrackID::ITSTPCTRD, MCLABELS); }
   auto getITSTPCTRDTrackMCLabel(GTrackID id) const { return getObject<o2::MCCompLabel>(id, MCLABELS); }
+  auto getITSTPCTRDSATracksMCLabels() const { return getSpan<o2::MCCompLabel>(GTrackID::ITSTPCTRD, MCLABELSEXTRA); }
+  auto getITSTPCTRDSATrackMCLabel(GTrackID id) const { return getObject<o2::MCCompLabel>(id, MCLABELSEXTRA); }
 
   // TPC-TRD
   template <class U>
@@ -405,6 +407,8 @@ struct RecoContainer {
   }
   auto getTPCTRDTracksMCLabels() const { return getSpan<o2::MCCompLabel>(GTrackID::TPCTRD, MCLABELS); }
   auto getTPCTRDTrackMCLabel(GTrackID id) const { return getObject<o2::MCCompLabel>(id, MCLABELS); }
+  auto getTPCTRDSATracksMCLabels() const { return getSpan<o2::MCCompLabel>(GTrackID::TPCTRD, MCLABELSEXTRA); }
+  auto getTPCTRDSATrackMCLabel(GTrackID id) const { return getObject<o2::MCCompLabel>(id, MCLABELSEXTRA); }
   // TRD tracklets
   gsl::span<const o2::trd::Tracklet64> getTRDTracklets() const;
   gsl::span<const o2::trd::CalibratedTracklet> getTRDCalibratedTracklets() const;
--- a/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
+++ b/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
@@ -240,7 +240,7 @@ struct RecoContainer {
   void addITSClusters(o2::framework::ProcessingContext& pc, bool mc);
   void addTPCClusters(o2::framework::ProcessingContext& pc, bool mc, bool shmap);
   void addTOFClusters(o2::framework::ProcessingContext& pc, bool mc);
-  void addTRDTracklets(o2::framework::ProcessingContext& pc);
+  void addTRDTracklets(o2::framework::ProcessingContext& pc, bool mc);
 
   void addFT0RecPoints(o2::framework::ProcessingContext& pc, bool mc);
 
@@ -385,6 +385,8 @@ struct RecoContainer {
   {
     return getSpan<o2::trd::TrackTriggerRecord>(GTrackID::ITSTPCTRD, TRACKREFS);
   }
+  auto getITSTPCTRDTracksMCLabels() const { return getSpan<o2::MCCompLabel>(GTrackID::ITSTPCTRD, MCLABELS); }
+  auto getITSTPCTRDTrackMCLabel(GTrackID id) const { return getObject<o2::MCCompLabel>(id, MCLABELS); }
 
   // TPC-TRD
   template <class U>
@@ -401,6 +403,8 @@ struct RecoContainer {
   {
     return getSpan<o2::trd::TrackTriggerRecord>(GTrackID::TPCTRD, TRACKREFS);
   }
+  auto getTPCTRDTracksMCLabels() const { return getSpan<o2::MCCompLabel>(GTrackID::TPCTRD, MCLABELS); }
+  auto getTPCTRDTrackMCLabel(GTrackID id) const { return getObject<o2::MCCompLabel>(id, MCLABELS); }
   // TRD tracklets
   gsl::span<const o2::trd::Tracklet64> getTRDTracklets() const;
   gsl::span<const o2::trd::CalibratedTracklet> getTRDCalibratedTracklets() const;
--- a/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
+++ b/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
@@ -115,9 +115,9 @@ void DataRequest::requestITSTPCTRDTracks(bool mc)
   addInput({"trackITSTPCTRD", "TRD", "MATCHTRD_GLO", 0, Lifetime::Timeframe});
   addInput({"trigITSTPCTRD", "TRD", "TRKTRG_GLO", 0, Lifetime::Timeframe});
   if (mc) {
-    LOG(WARNING) << "TRD Tracks does not support MC truth, dummy label will be returned";
+    addInput({"trackITSTPCTRDMCTR", "GLO", "MTCHLBL_GLO", 0, Lifetime::Timeframe});
   }
-  requestMap["trackITSTPCTRD"] = false;
+  requestMap["trackITSTPCTRD"] = mc;
 }
 
 void DataRequest::requestTPCTRDTracks(bool mc)
@@ -126,9 +126,9 @@ void DataRequest::requestTPCTRDTracks(bool mc)
   addInput({"trackTPCTRD", "TRD", "MATCHTRD_TPC", ss, Lifetime::Timeframe});
   addInput({"trigTPCTRD", "TRD", "TRKTRG_TPC", ss, Lifetime::Timeframe});
   if (mc) {
-    LOG(WARNING) << "TRD Tracks does not support MC truth, dummy label will be returned";
+    addInput({"trackTPCTRDMCTR", "GLO", "MTCHLBL_TPC", 0, Lifetime::Timeframe});
   }
-  requestMap["trackTPCTRD"] = false;
+  requestMap["trackTPCTRD"] = mc;
 }
 
 void DataRequest::requestTOFMatches(bool mc)
@@ -354,7 +354,7 @@ void RecoContainer::collectData(ProcessingContext& pc, const DataRequest& reques
 
   req = reqMap.find("trackletTRD");
   if (req != reqMap.end()) {
-    addTRDTracklets(pc);
+    addTRDTracklets(pc, req->second);
   }
 
   req = reqMap.find("Cosmics");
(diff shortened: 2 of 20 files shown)


shahor02:
> `o2-trd-tracklet-transformer -b | o2-trd-global-tracking -b --track-sources TPC --strict-matching --configKeyValues "GPU_rec_trd.chi2SeparationCut=2;GPU_proc.trdNCandidates=3"`

Will check details bit later, but can't you propagate the `strict-matching` flag to the matcher and use GPU_rec_trd.chi2SeparationCut only when it is activated, so that we don't need to pass the options explicitly?

> I see that I have in 10pp events already one tracklet which is attached to two tracks. I am looking further into that and for that I will try to add also the MC labels now to the tracking. This will come in another PR.

This should be compared with fraction of fake attachments of the tracklets it it is used > 1 time.

martenole:
> Will check details bit later, but can't you propagate the `strict-matching` flag to the matcher and use GPU_rec_trd.chi2SeparationCut only when it is activated, so that we don't need to pass the options explicitly?

Actually that setting does not need to be passed explicitly when strict matching is requested. Only for parameter tuning this is needed. The matcher will always flag possible ambiguous tracks, independent of the matching mode, because that is only a very simple check and should not cost processing time. Therefore the matcher doesn't need the propagated strict matching flag.

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
```diff
@@ -115,9 +115,9 @@ void DataRequest::requestITSTPCTRDTracks(bool mc)
   addInput({"trackITSTPCTRD", "TRD", "MATCHTRD_GLO", 0, Lifetime::Timeframe});
   addInput({"trigITSTPCTRD", "TRD", "TRKTRG_GLO", 0, Lifetime::Timeframe});
   if (mc) {
-    LOG(WARNING) << "TRD Tracks does not support MC truth, dummy label will be returned";
+    addInput({"trackITSTPCTRDMCTR", "GLO", "MTCHLBL_GLO", 0, Lifetime::Timeframe});
```
shahor02: the DataSource should be TRD. In fact, since this data corresponds to ITS-TPC:TRD matches only, could you please rename all to MATCH_ITSTPC, TRKTRG_ITSTPC and MCLB_ITSTPC respectively (for uniformity, one can could also rename matches to TPC only according to the same schema). This should be done also in the DPL Specs.

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
```diff
@@ -126,9 +126,9 @@ void DataRequest::requestTPCTRDTracks(bool mc)
   addInput({"trackTPCTRD", "TRD", "MATCHTRD_TPC", ss, Lifetime::Timeframe});
   addInput({"trigTPCTRD", "TRD", "TRKTRG_TPC", ss, Lifetime::Timeframe});
   if (mc) {
-    LOG(WARNING) << "TRD Tracks does not support MC truth, dummy label will be returned";
+    addInput({"trackTPCTRDMCTR", "GLO", "MTCHLBL_TPC", 0, Lifetime::Timeframe});
```
shahor02: idem, DataSource does not match to other inputs.

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -111,6 +114,53 @@ void TRDGlobalTracking::updateTimeDependentParams()
   mTracker->SetTPCVdrift(mTPCVdrift);
 }
 
+void TRDGlobalTracking::fillMCTruthInfo(const TrackTRD& trk, const o2::MCCompLabel& lblSeed, std::vector<o2::MCCompLabel>& lblContainer, const o2::dataformats::MCTruthContainer<o2::MCCompLabel>* trkltLabels) const
```
shahor02: it is faster to pass by copy 8B `const o2::MCCompLabel lblSeed` than pass the reference of the same size then to dereference.

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -111,6 +114,53 @@ void TRDGlobalTracking::updateTimeDependentParams()
...
+    auto trkltIndex = trk.getTrackletIndex(iLy);
+    if (trkltIndex == -1) {
+      // no tracklet in this layer
+      continue;
+    }
+    const auto& lblsTrklt = trkltLabels->getLabels(trkltIndex);
+    for (const auto lblTrklt : lblsTrklt) {
+      ++labelCounter[lblTrklt];
+      if (labelCounter[lblTrklt] > maxOccurences) {
+        maxOccurences = labelCounter[lblTrklt];
```
shahor02: here you have 2 searches over the map, it is faster to do directly 
````
int occ = labelCounter[lblTrklt]++;
if (occ > maxOccurences) {
 maxOccurences = occ;
}
````

File: Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx [hunk 5]
```diff
@@ -228,14 +291,24 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
...
+      // skip tracks which have another hypothesis close to the best one or which do are above strict chi2 threshold
+      continue;
+    }
     nTrackletsAttached += trdTrack.getNtracklets();
     auto trackGID = trdTrack.getRefGlobalTrackId();
     if (trackGID.includesDet(GTrackID::Source::ITS)) {
       // this track is from an ITS-TPC seed
       tracksOutITSTPC.push_back(trdTrack);
+      if (mUseMC) {
+        fillMCTruthInfo(trdTrack, inputTracks.getTPCITSTrackMCLabel(trackGID), matchLabelsITSTPC, inputTracks.getTRDTrackletsMCLabels());
```
shahor02: getting `inputTracks.getTPCITSTrackMCLabel(trackGID)` in loop is more expensive (a span is created on the fly at every call) than fetching once `const auto lblSpan = inputTracks.getTPCITSTrackMCLabels()` then accessing in loop `lblSpan[trackGID]`.

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -111,6 +114,53 @@ void TRDGlobalTracking::updateTimeDependentParams()
...
+      if (labelCounter[lblTrklt] > maxOccurences) {
+        maxOccurences = labelCounter[lblTrklt];
+      }
+    }
+  }
+  o2::MCCompLabel mostFrequentLabel;
+  for (const auto& [lbl, count] : labelCounter) {
+    LOG(DEBUG) << "Label " << lbl << " occured " << count << " times.";
+    if (count == maxOccurences) {
+      if (lblSeed.compare(lbl) >= 0) {
```
shahor02: `MCCompLabel::compare` checks also for the isFake flag of each label, which you don't need, it is faster to do 
````
if (lblSeed == lbl) { 
````

File: Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx [hunk 7]
```diff
@@ -111,6 +114,53 @@ void TRDGlobalTracking::updateTimeDependentParams()
...
+        // most frequent label matches seed label
+        mostFrequentLabel = lbl;
+        mostFrequentLabel.setFakeFlag(maxOccurences != trk.getNtracklets());
```
shahor02, line 150: The convention (unfortunately, undocumented yet) for the matches label is that they have the MC track ID of the source track (effectively, that of the TPC, since the same is applied to ITS-TPC tracks) and isFake is set only if the mostFrequentLabel (even if by itself it would have a fake flag set)  does not match to that ID.
It is somewhat unfortunate that these labels have a meaning of both the track and match label. For the ITS and ITSAB (which is just a collection of the cluster references, we have their proper labels (with the fake flag set if mostFrequentLabel is not seen on every contributing layer) and there is a separate label for the TrackTPCITS. 
Since you don't have a separate TRD-info only track object, perhaps you could add simply a 2nd vector of labels which would bear the label you create now, while the label associated with the match should be `lblSeed.setFakeFlag(lblSeed == mostFrequentLabel)`

Detectors/TRD/workflow/src/TRDTrapSimulatorSpec.cxx
```diff
@@ -296,7 +298,7 @@ void TRDDPLTrapSimulatorTask::run(o2::framework::ProcessingContext& pc)
             for (const auto& newLabel : newLabels) {
               bool alreadyIn = false;
               for (const auto& currLabel : currentLabels) {
-                if (currLabel.compare(newLabel)) {
+                if (currLabel.compare(newLabel) == 1) {
```
shahor02: here too, you don't need to check the `isFake` flags, just use `currLabel == newLabel`

shahor02 (changes requested):

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
martenole: Thanks, this I did not take into account. But I need to do `int occ = ++labelCounter[lblTrklt]`, otherwise `occ` is set to `labelCounter[lblTrklt]` before the increment.

shahor02: right

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
martenole: Yes, I did read the comment in RecoContainer.h and wanted to do it that way. The random access was there for testing. Above in line 202 and following I already got the span of all labels, I just forgot to change to using `itstpcTrackLabels[trackGID]`. Thanks for spotting

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
martenole: Ok I created two label containers now:
1. `lblMatch` is simply the label of the seeding track and the fake flag is set if `lblSeed != mostFrequentLabel`.
2. `lblTrd` is the most frequent TRD label. Should I set the fake flag here only if the most frequent label does not appear in every layer where a tracklet is attached? I think this would make more sense than also requesting that the label matches the seed's label, since it is supposed to be the label of the TRD track.

And I am also wondering how to name the message with the TRD track labels. I need two different descriptions depending on the seed, so I thought `MCLB_ITSTPC_TRD` and `MCLB_TPC_TRD` would be ok. Or should I use `MCLB_TRD` with different subspecs?

martenole: Hm, `MCLB_TRD` with different subspecs is not good, since the subspec is used to indicate strict matching mode. So I will go with `MCLB_ITSTPC_TRD`

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
martenole: Done

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
martenole: Done (and made consistent with ITS-TPC match names)

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
martenole: Thanks, fixed

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
martenole: Done

Detectors/TRD/workflow/src/TRDTrapSimulatorSpec.cxx
martenole: Done

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
shahor02: Hi, yes, for the TRD label proper you should ignore the trackID of the seeding track, just the dominant contributor and its seFake(true) if it is not represented on every layer.
MCLB_ITSTPC_TRD and MCLB_TPC_TRD should be ok, indeed, better to not touch the subspec.

shahor02 (commented):
Hi @martenole 

Looks fine, but the extra MC labels reading has to be added to TRDTrackReader and also to RecoContainer (to MCLABELSEXTRA slot). Will you add this here or I merge this and you open a new PR?

martenole:
Hi @shahor02 
ok, I will add this here shortly.
Cheers,
Ole

DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
```diff
@@ -387,6 +387,8 @@ struct RecoContainer {
   }
   auto getITSTPCTRDTracksMCLabels() const { return getSpan<o2::MCCompLabel>(GTrackID::ITSTPCTRD, MCLABELS); }
   auto getITSTPCTRDTrackMCLabel(GTrackID id) const { return getObject<o2::MCCompLabel>(id, MCLABELS); }
+  auto getITSTPCTRDSATracksMCLabels() const { return getSpan<o2::MCCompLabel>(GTrackID::ITSTPCTRD, MCLABELSEXTRA); }
+  auto getITSTPCTRDSATraclMCLabel(GTrackID id) const { return getObject<o2::MCCompLabel>(id, MCLABELSEXTRA); }
```
shahor02: a small typo: getITSTPCTRDSATraclMCLabel -> getITSTPCTRDSATracklMCLabel

martenole: Thanks, I fixed that and squashed the commits for the separate TRD MC labels

shahor02 (approved):

PR #6809, opened 2021-08-05
[EMCAL-566]  Templated ChannelCalibrator and included time calibrator 

jokonig:
- EMCALChannelCalibrator is now templated to be used for bad channel calibration as well as time calibration (the .cxx file had to be removed as all template classes have to be defined in the .h file). The Channel calibrator has to be initialized with the Channel/Time Data class and the InitParams (simple struct with histogram ranges) for the Time/Channel Data
- Added seperate file for EMCALChannelData for better readability
- Added basic sceletton task for time calibration (EMCALTimeCalibData.h/cxx) similar to EMCALChannelData
- fixed merge confict/formatting in CMakeLists

### DIFF
--- a/Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelData.h
+++ b/Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelData.h
@@ -1,0 +1,108 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \class EMCALChannelData
+/// \brief  Perform the EMCAL bad channel calibration
+/// \author Hannah Bossi, Yale University
+/// \ingroup EMCALCalib
+/// \since Feb 11, 2021
+
+#ifndef CHANNEL_DATA_H_
+#define CHANNEL_DATA_H_
+
+#include "DetectorsCalibration/TimeSlotCalibration.h"
+#include "DetectorsCalibration/TimeSlot.h"
+#include "DataFormatsEMCAL/Cell.h"
+#include "EMCALBase/Geometry.h"
+#include "CCDB/CcdbObjectInfo.h"
+
+#include "Framework/Logger.h"
+#include "CommonUtils/MemFileHelper.h"
+#include "CCDB/CcdbApi.h"
+#include "DetectorsCalibration/Utils.h"
+#include <boost/histogram.hpp>
+#include <boost/histogram/ostream.hpp>
+#include <boost/format.hpp>
+
+#include <array>
+#include <boost/histogram.hpp>
+
+namespace o2
+{
+namespace emcal
+{
+struct ChannelCalibInitParams {
+  unsigned int nbins = 1000;
+  std::array<float, 2> range = {0, 0.35};
+};
+
+class EMCALChannelData
+{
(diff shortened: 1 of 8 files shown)


Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelData.h
```diff
@@ -0,0 +1,108 @@
...
+  //using Slot = o2::calibration::TimeSlot<o2::emcal::EMCALChannelData>;
+  using Cells = o2::emcal::Cell;
+  using boostHisto = boost::histogram::histogram<std::tuple<boost::histogram::axis::regular<double, boost::use_default, boost::use_default, boost::use_default>, boost::histogram::axis::integer<>>, boost::histogram::unlimited_storage<std::allocator<char>>>;
+
+ public:
+  // NCELLS includes DCal, treat as one calibration
+  o2::emcal::Geometry* mGeometry = o2::emcal::Geometry::GetInstanceFromRunNumber(300000);
+  int NCELLS = mGeometry->GetNCells();
+
+  EMCALChannelData(ChannelCalibInitParams hist) : mNBins(hist.nbins), mRange(1)
```
shahor02: why not `const ChannelCalibInitParams& hist` ?

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALTimeCalibData.h
```diff
@@ -0,0 +1,113 @@
...
+namespace o2
+{
+namespace emcal
+{
+
+// holds the time calibration parameters
+class TimeCalibParams
+{
+  std::array<float, 17776> arr;
+};
```
mfasDa: Can be using directives inside class:

``` c++
using TimeCalibParams = std::array<float, 17776> arr;
```

Actually the object is not needed because the process function should return directly o2::emcal::TimeCalibrationParams (Detectors/EMCAL/calib), so it can be removed entirely.

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALTimeCalibData.h
```diff
@@ -0,0 +1,113 @@
...
+  boostHisto& getHisto() { return mTimeHisto; }
+  const boostHisto& getHisto() const { return mTimeHisto; }
+
+  /// \brief Set number of events available for calibration
+  void setNEvents(int ne) { mEvents = ne; }
+
+  void PrintStream(std::ostream& stream) const;
+
+  /// \brief Actual function where calibration is done. Has to be called in has enough data when enough data is there
+  o2::emcal::TimeCalibParams process(); // how should the return type be be called? Or dont I have to return something and directly push to CCDB in process?
```
mfasDa: Should return o2::emcal::TimeCalibrationParams (Detectors/EMCAL/calib - library with EMCAL CCDB content)

In addition a dummy implementation for the function is missing, without there will be a linker error.

mfasDa (changes requested):
Hi @jokonig ,

There are a few changes in the time calib params needed (particularly in the process function). Please have a look!

Cheers

Markus

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALTimeCalibData.h
jokonig: deleted the TimeCalibParams and now use the dedicated TimeCalibrationParams

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALTimeCalibData.h
jokonig: added a dummy implementation and added the correct return type

shahor02 (changes requested):

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelData.h
jokonig: changed it as suggested.
Also did the same change to EMCALTimeCalibData.h

mfasDa (approved):
Looks good now!

mfasDa:
CI errors look unrelated

shahor02 (approved):

PR #6810, opened 2021-08-05
Fix in EncodedBlocks relocated pointers manipulations

shahor02:


shahor02:
macOS failure is unrelated (CI did not start), merging.

PR #6811, opened 2021-08-06
[QC-627] Mergers: add postDeserialization() callback and prioritize MergeInterface

knopers8:
This lets the children of MergeInterface to perform any cleanups after ROOT deserializes the object. The main use-case is setting the valid ownership, which cannot be propely set while serializing it for any reason.
We also prioritize storing an object as `MergeInterface*` over `TObject*` when a class inherits both (`MergeInterface::merge()` was already prioritized, this does not change).


Barthelemy (approved):

Barthelemy:
the macOS failure is not related, merging.

PR #6813, opened 2021-08-06
[QC-443] Data Sampling: ptree interfaces for everything

knopers8:
After we move to those in QC, it will be followed up with deleting methods which expect ConfigurationInterface.


Barthelemy (approved):

Barthelemy:
macos failure is unrelated, merging

PR #6814, opened 2021-08-06
getNPixels() method added to ITSMFT ClusterPattern

shahor02:
@akalweit


shahor02:
macOS failure is unrelated, merging.

PR #6815, opened 2021-08-06
WIP: DPL raw proxy optimization, avoid DataHeader parsing in special case

matthiasrichter:
We observe very poor performance of the raw proxy if the data consists
of a lot of message pairs, the overhead of parsing all data headers is
not negligible. Though the data model was not made for this use case,
an optimization can be applied to the proxy which is meaningful anyhow.
If there is only one output channel, the inspection of the data headers
to find the corresponding output channel can be skipped.

By specifying 'any' using (Note: 3x* for origin and 15x* for description)
```
--dataspec "ALL:***/**************"
```

More things need to be fixed:
- have to check if DPL creates the correct input and output routes
- in a test without a connected process, the internal-dpl-injected-dummy-sink
  has an error: "Could not match incoming data to any input route"
  the data relayer is probably not recognizing the 'any' data type
- we should allow to specify the 'any' wildcard by 'ALL:*/*' (ALL is arbitrary here)
- make sure that there is exactly one output channel if 'any' is specified
- there might be multiple output channels if time-pipelining is used, this
  needs to be correctly implemented.
- the handling of EOS needs to be optimized because every data set is parsed
  completely to find a potential SourceInfoHeader
- maybe we can detect whether or not to apply the optimization by only
  checking the number of output FairMQ channels, but if a downstream
  device does not exactly specify the input specs matching all data (which
  was injected into DPL without any check), the DataRelayer of the downstream
  device will report errors.

In summary it's probably better to optimize the parsing loop, at least the
EOS should be handled better to cut out one of the iterations, and we need
micro benchmarks. to know what to expect.

### DIFF
--- a/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
+++ b/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
@@ -222,6 +222,12 @@ InjectorFunction dplModelAdaptor(std::vector<OutputSpec> const& filterSpecs, boo
     std::string channelNameForSplitParts;
     static int64_t dplCounter = -1;
     dplCounter++;
+    if (filterSpecs.size() == 1 && DataSpecUtils::match(filterSpecs[0], OutputSpec{{header::gDataOriginAny, header::gDataDescriptionAny}})) {
+      // there should be only a single output channel
+      auto channelName = channelRetriever(filterSpecs[0], 0);
+      sendOnChannel(device, parts, channelName);
+      return;
+    }
     for (size_t msgidx = 0; msgidx < parts.Size() / 2; ++msgidx) {
       const auto dh = o2::header::get<DataHeader*>(parts.At(msgidx * 2)->GetData());
       if (!dh) {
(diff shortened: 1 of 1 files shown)


matthiasrichter:
@wiechula with this you should be able to check if it makes a difference. But probably this is not a possible solution because with the 'any' wildcard as the proxy output, DPL can only in very simple cases determine the necessary channels. I will find a better way, probably with a separate workflow option to force-skip the parsing.

Framework/Core/src/ExternalFairMQDeviceProxy.cxx
```diff
@@ -222,6 +222,12 @@ InjectorFunction dplModelAdaptor(std::vector<OutputSpec> const& filterSpecs, boo
     std::string channelNameForSplitParts;
     static int64_t dplCounter = -1;
     dplCounter++;
+    if (filterSpecs.size() == 1 && DataSpecUtils::match(filterSpecs[0], OutputSpec{{header::gDataOriginAny, header::gDataDescriptionAny}})) {
+      // there should be only a single output channel
+      auto channelName = channelRetriever(filterSpecs[0], 0);
```
ktf, line 227: I think another source of slowdown is the fact that we use the string based sending API. It would be much better to cache the FairMQChannel ptr and use that.

PR #6816, opened 2021-08-08
Part of modifications for the barrel alignment

shahor02:
Not functional yet...

* Multiple fixes to adapt from aliroot to O2
* ITS alignables definitions
* Steering workflow skeleton


shahor02:
macOS failure is unrelated, merging

PR #6821, opened 2021-08-09
Fix comments in trd rawdata.h

bazinski:
ordering is wrong in the last line of error flags in the comments portion of HalfCRUHeader.


shahor02:
trivial, comment change only, merging.

PR #6822, opened 2021-08-09
Set partition field of InfoLogger

shahor02:


shahor02:
macOS failure unrelated, merging.

PR #6826, opened 2021-08-09
[EMCAL-710] Error container for different error sources

mfasDa:
- Use error source and error code instead of multiple error codes
- Add page start error in case current pos. + header size is larger than the payload size
- Add FEE ID to of the current (last) FEE to the raw decoding error
- Update documentation for ErrorTypeFEE


mfasDa:
@shahor02 This is related to the crash on the EPN from this afternoon. Main goal is to add an error code for the case where the page start decoding is out of range. Furthermore I propagate also the FEE ID as requested by Martin to find the faulty card. This required some changes in the error container (ErrorTypeFEE) that is sent to the QC and will hopefully soon be implemented as QC component.

shahor02 (commented):
@mfasDa please fix clang-format errors.

mfasDa:
@shahor02 Done

shahor02 (approved):

PR #6829, opened 2021-08-10
TRD simulation cleanup

martenole:
My goal was actually to suppress the "Drift Velocity too small" warning in the TRD digitization, but I noticed that simply changing the severity to `DEBUG` might not be such a good idea, because then the underlying problem would be hidden. In case the local drift velocity for a pad is too low (about 1cm/us corresponding to a drift HV of less than 1.5kV) we don't have results from a Garfield simulation and simply set the drift velocity to the lowest value for which we have simulation results.
But since we usually always have some chambers with reduced drift HV we cannot trust the simulation for those chambers.
I wonder if it would be good to 
* keep track of the chambers for which vDrift is too low and output them to the log once at the end of one simulation
* add a very rough estimate for the drift time (which might still be better than just assuming 1cm/us)
In any case it would probably be good to find a way to suppress all those warning messages.

When looking at the simulation I noticed that the functions in DiffusionAndTimeStructEstimator were duplicates from the functions in CommonParam, so I removed them from CommonParam.  And the Garfield results I moved into a single header and did some more minor cleanup.


martenole:
With [e2227f1](https://github.com/AliceO2Group/AliceO2/pull/6829/commits/e2227f13f7fbec89846e5f62f83f8c66fadecd8f) I see for 10 PbPb events one log message:
`[51496:TRDDigitizer]: [22:55:32][INFO] List of TRD chambers for which at least once an unusual drift velocity was found: 5, 8, 26, 29, 30, 51, 113, 180, 181, 184, 215, 227, 228, 232, 233, 247, 265, 287, 476, 494, 498, 529`
Instead of 376 warnings similar to:
`[45082:TRDDigitizer]: [22:43:14][WARN] TRD: Drift Velocity too small 0.111536 < 1.032`.
The processing time stays around the same as without these changes.

Detectors/TRD/workflow/src/TRDDigitizerSpec.cxx
```diff
@@ -146,8 +146,9 @@ class TRDDPLDigitizerTask : public o2::base::BaseDPLDigitizer
     if (mctruth) {
       labelsAccum.mergeAtBack(labels);
     }
+    LOGF(INFO, "List of TRD chambers for which at least once an unusual drift velocity was found: %s", mDigitizer.dumpFlaggedChambers());
```
jolopezl: I think that `"List of TRD chambers with at least one drift velocity out-of-range: %s"` is more consistent with what the real error is.

Detectors/TRD/simulation/src/Digitizer.cxx
```diff
@@ -316,7 +311,11 @@ bool Digitizer::convertHits(const int det, const std::vector<Hit>& hits, SignalC
           zz = 0.5 - zz;
         }
         // Use drift time map (GARFIELD)
-        driftTime = mDriftEstimators[thread].TimeStruct(driftVelocity, 0.5 * AmWidth - 1.0 * locTd, zz) + hit.GetTime();
+        bool errFlag = false;
+        driftTime = mDriftEstimators[thread].timeStruct(driftVelocity, 0.5 * AmWidth - 1.0 * locTd, zz, &errFlag) + hit.GetTime();
+        if (errFlag) {
+          mFlagVdriftOutOfRange[det] = true;
+        }
```
jolopezl: What about dropping the `errFlag` variable and the if-check just by using `&mFlagVdriftOutOfRange[det]` instead of `errFlag`?

Detectors/TRD/simulation/src/Digitizer.cxx
```diff
@@ -74,6 +67,8 @@ void Digitizer::init()
     mDriftEstimators.emplace_back();
   }
 
+  std::fill(mFlagVdriftOutOfRange.begin(), mFlagVdriftOutOfRange.end(), false);
+
```
jolopezl: I think this isn't needed since `mFlagVdriftOutOfRange` is automatically initialized to its default value since it's a `std::array` of `bools`, see similar comment in the header.

Detectors/TRD/simulation/src/Digitizer.cxx
```diff
@@ -40,16 +40,9 @@ void Digitizer::init()
...
   mSimParam = SimParam::instance();          // Instance for simulation parameters
   mCommonParam = CommonParam::instance();    // Instance for common parameters
-  if (!mSimParam) {
-  }
-  if (!mCommonParam) {
-  } else {
-    if (!mCommonParam->cacheMagField()) {
-    }
-  }
+  mCommonParam->cacheMagField();
```
jolopezl, line 45: Thanks, I think what was there was really weird.

Detectors/TRD/simulation/include/TRDSimulation/Digitizer.h
```diff
@@ -108,6 +108,7 @@ class Digitizer
   std::vector<MCLabel> mMergedLabels;                                            // temporary label container
   std::array<SignalContainer, constants::MAXCHAMBER> mSignalsMapCollection;      // container for caching signals over a timeframe
   std::deque<std::array<SignalContainer, constants::MAXCHAMBER>> mPileupSignals; // container for piled up signals
+  std::array<bool, constants::MAXCHAMBER> mFlagVdriftOutOfRange;                 // flag chambers in which the drift velocity is out of range for which a parameterization exists
```
jolopezl: `std::array<bool, constants::MAXCHAMBER> mFlagVdriftOutOfRange{false};` to ensure the right initialization.

Detectors/TRD/base/test/testTRDDiffusionCoefficient.cxx
```diff
@@ -36,55 +36,16 @@ BOOST_AUTO_TEST_CASE(TRDDiffusionCoefficient_test1)
...
-/// \brief Test time structure
-BOOST_AUTO_TEST_CASE(TRDTimeStructure_test)
-{
-  auto commonParam = CommonParam::instance();
-  DiffusionAndTimeStructEstimator estimator;
-  BOOST_CHECK_CLOSE(estimator.TimeStruct(1.48, 1., 0.1), commonParam->timeStruct(1.48, 1., 0.1), 0.001);
-  BOOST_CHECK_CLOSE(estimator.TimeStruct(1.1, 1., 0.1), commonParam->timeStruct(1.1, 1., 0.1), 0.001);
-  BOOST_CHECK_CLOSE(estimator.TimeStruct(2, 1., 0.1), commonParam->timeStruct(2, 1., 0.1), 0.001);
-  BOOST_CHECK_CLOSE(estimator.TimeStruct(4, 1., 0.1), commonParam->timeStruct(4, 1., 0.1), 0.001);
-}
```
jolopezl, line 57: Back then, I liked having the two implementations set. However, since the second implementation evolved differently, changes can cause side effects, and these lines are like having a regression test of its functionality. I think it's ok to drop this test and the implementation from commonParam if one remembers that all of this comes from the same code in AliRoot.

Detectors/TRD/base/include/TRDBase/PadResponse.h
```diff
@@ -29,8 +29,8 @@ class PadResponse
  private:
   void samplePRF();                               // Initialized the PRF
   static constexpr int mPRFbin{500};              // Number of bins for the PRF
-  static constexpr float mPRFlo{-1.5};            // Lower boundary of the PRF
-  static constexpr float mPRFhi{1.5};             // Higher boundary of the PRF
+  const float mPRFlo{-1.5};                       // Lower boundary of the PRF
+  const float mPRFhi{1.5};                        // Higher boundary of the PRF
```
jolopezl, line 33: out of curiosity, what's the reason for this change?

DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
```diff
@@ -55,6 +55,10 @@ constexpr float PADGRANULARITYTRKLSLOPE = 1000.f;
 constexpr float GRANULARITYTRKLPOS = 1.f / PADGRANULARITYTRKLPOS;     // granularity of position in tracklet64 word in pad-widths
 constexpr float GRANULARITYTRKLSLOPE = 1.f / PADGRANULARITYTRKLSLOPE; // granularity of slope in tracklet64 word in pads/timebin
 
+// Simulation specific parameters
+constexpr int TIMEBINSGARFIELD = 38; // Number of bins in time direction used for garfield simulation
+constexpr int ZBINSGARFIELD = 11;    // Number of bins in z direction used for garfield simulation
+
```
jolopezl: I am happy with having all the constants in one place, particularly those used in several other units. Still, a few like these two are used only in a couple of places, for instance, in the definition of `Garfield` and in the drift estimators. So, any change in the constants will lead to recompiling and more classes that don't need to be recompiled. Therefore, I think `Garfield` can keep these constants for itself. Also, it's unlikely that these will change.

Detectors/TRD/base/include/TRDBase/Garfield.h
```diff
@@ -0,0 +1,348 @@
...
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef O2_TRD_GARFIELD_H
+#define O2_TRD_GARFIELD_H
+
+#include "DataFormatsTRD/Constants.h"
```
jolopezl: See my comment above in `o2::trd::Constants`.

jolopezl (commented):
Thanks, @martenole, this is very good. This is something that was definitively in the backlog frozen for a long time. I have a few comments you might want to consider. I bet there is a speedup in the simulation time; also, just making it much less verbose and dropping a message at the end is more pleasant and useful.

Detectors/TRD/workflow/src/TRDDigitizerSpec.cxx
martenole: Thanks, changed

Detectors/TRD/simulation/src/Digitizer.cxx
martenole: Good point

Detectors/TRD/simulation/src/Digitizer.cxx
martenole: Yes, changed

Detectors/TRD/base/include/TRDBase/PadResponse.h
martenole: I think there is no need for those variables to be static, so it is better to avoid it: https://rawgit.com/AliceO2Group/CodingGuidelines/master/coding_guidelines.html?showone=Global_Variables#Global_Variables

Detectors/TRD/base/test/testTRDDiffusionCoefficient.cxx
martenole: There is a comment in DiffusionAndTimeStructEstimator.h which should make that clear. I think there would be a few other things one can consider to clean up, but I did not want to go too deep now. For example we could consider merging CommonParam into SimParam, since it is used only for simulations. And in general one could probably use more the ConfigurableParam functionality than lots of singletons. But that is not so urgent at the moment I guess.

DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
martenole: I agree, that is changed now

Detectors/TRD/base/include/TRDBase/PadResponse.h
jolopezl: Thanks for the comments.

jolopezl (approved):
Looks great. Thanks for adding changes to the original pull request.

martenole:
@tdietel what do you think? Do you approve these changes?
If this is merged it should be squashed first

tdietel (approved):
Looks good to me. Just one comment: do we have the code to run the Garfield simulation somewhere? It seems this commit only contains the tabulated results?

PR #6830, opened 2021-08-10
[EMCAL-666] Include entropy decoding into the emcal reconstruction workflow

mfasDa:
- Add input type CTF and include entropy decoder
  spec for this input type
- Add cells and clusters as valid output types for
  input type CTF
- Add clusters as valid output type for input type raw

### DIFF
--- a/Detectors/EMCAL/workflow/include/EMCALWorkflow/RecoWorkflow.h
+++ b/Detectors/EMCAL/workflow/include/EMCALWorkflow/RecoWorkflow.h
@@ -28,10 +28,11 @@ namespace reco_workflow
 /// \enum InputType
 /// \brief Input types of the workflow
 /// \ingroup EMCALworkflow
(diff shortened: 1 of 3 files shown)


shahor02:
Hi @mfasDa 
Not sure what is the point of this PR? The `o2-ctf-reader` automatically deploys entropy decoders, either for all detectors which are in the GRP (skipping those mentioned with `--skipDet`) or for those which are requested via `--onlyDet`.
E.g. with
````
o2-ctf-reader-workflow --ctf-input <ctfs> --onlyDet EMC | o2-emcal-reco-workflow --input cells --output clusters ...`
````
your reco workflow will receive cells.

mfasDa:
Thanks @shahor02 , I thought we need to decoding inside the emcal reco workflow to get back the cells. Then this PR can be closed.

PR #6834, opened 2021-08-11
[EMCAL-647] Event builder for cell container from different subtimeframes

mfasDa:
Combining cells from multiple subevents in a single
cell / trigger record container according to cells
belonging to the same collsion. Output is sent to
the same output type as input cell / trigger record
container but with subspecification 0 for
compatibility with existing containers.

### DIFF
--- a/Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
@@ -1,0 +1,139 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction
+
+#include <algorithm>
+#include <iterator>
+#include <unordered_map>
+#include <gsl/span>
+
+#include "DataFormatsEMCAL/Cell.h"
+#include "DataFormatsEMCAL/TriggerRecord.h"
+#include "Framework/InputSpec.h"
+#include "Framework/Logger.h"
+#include "EMCALWorkflow/EventBuilderSpec.h"
+
+using namespace o2::emcal;
+
+void EventBuilderSpec::init(framework::InitContext& ctx)
+{
+}
+
+void EventBuilderSpec::run(framework::ProcessingContext& ctx)
+{
+  auto dataorigin = o2::header::gDataOriginEMC;
+  std::unordered_map<o2::header::DataHeader::SubSpecificationType, gsl::span<const TriggerRecord>> triggers;
+  std::unordered_map<o2::header::DataHeader::SubSpecificationType, gsl::span<const Cell>> cells;
+
+  std::vector<Cell> outputCells;
+  std::vector<TriggerRecord> outputTriggers;
+
+  auto posCells = ctx.inputs().getPos("inputcells"),
+       posTriggerRecords = ctx.inputs().getPos("inputtriggers");
+  auto numSlotsCells = ctx.inputs().getNofParts(posCells),
+       numSlotsTriggerRecords = ctx.inputs().getNofParts(posTriggerRecords);
+  for (decltype(numSlotsCells) islot = 0; islot < numSlotsCells; islot++) {
+    auto celldata = ctx.inputs().getByPos(posCells, islot);
+    auto subspecification = framework::DataRefUtils::getHeader<header::DataHeader*>(celldata)->subSpecification;
+    // Discard message if it is a deadbeaf message (empty timeframe)
+    if (subspecification == 0xDEADBEEF) {
+      continue;
+    }
+    cells[subspecification] = ctx.inputs().get<gsl::span<o2::emcal::Cell>>(celldata);
+  }
+  for (decltype(numSlotsTriggerRecords) islot = 0; islot < numSlotsTriggerRecords; islot++) {
+    auto trgrecorddata = ctx.inputs().getByPos(posTriggerRecords, islot);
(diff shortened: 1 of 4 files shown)


Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
```diff
@@ -0,0 +1,140 @@
...
+{
+  std::set<RangeCollection> events;
+
+  // Search interaction records from all subevents
+  std::set<o2::InteractionRecord> allInteractions;
+  for (auto& [subspecification, trgrec] : triggerrecords) {
+    for (auto rec : trgrec) {
+      auto eventIR = rec.getBCData();
+      if (allInteractions.find(eventIR) == allInteractions.end())
+        allInteractions.insert(eventIR);
```
shahor02: Why do you need `find`?  `insert` will anyway call it, like this you are doing it twice.

Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  // iterate over all subevents for all bunch crossings
+  for (auto collision : allInteractions) {
+    RangeCollection nextevent;
+    nextevent.mInteractionRecord = collision;
+    bool first = true,
+         hasSubevent = false;
+    for (auto [subspecification, records] : triggerrecords) {
+      auto found = std::find_if(records.begin(), records.end(), [&collision](const o2::emcal::TriggerRecord& rec) { return rec.getBCData() == collision; });
```
shahor02: Not sure if the overhead is significant, but instead of using find for every int.record, you could use the fact that the data in each of triggerrecords span is ordered in int.record, i.e. something like this:
````
std::vector<int> cursor(triggerrecords.size());
for (auto collision : allInteractions) {
  RangeCollection nextevent;
  nextevent.mInteractionRecord = collision;
  bool first = true, hasSubevent = false;
  int count = 0;
  for (auto [subspecification, records] : triggerrecords) {
     if (cursor[count] >= records.size()) {
       count++;
       continue;
     }
     while (cursor[count]<records.size() && records[cursor[count]].getBCData() < collision) {
       cursor[count]++;
     }
     auto slotID  = cursor[count++];
     if (slotID>=records.size()) { // not found !
       continue;
     }
     .... use data of the trigger  slotID
  }
````

Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
mfasDa: The goal of these lines was to avoid double counting.  The loop is doing a set merge of the sets of trigger records assuming the sets could differ (empty timeframe in either of the sets could be an option). In this case the interaction record should only be added in case not already present.

shahor02: std::set by definition contains only unique elements, if the trig.record is already there, insert will do nothing.

Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
mfasDa: I was doing this because I didn't want to imply time-ordering is guaranteed.

However indeed the Raw to cell converter should send the events in the timeframe time-ordered. It currently uses a map for the internal bookkeeping - which I understand is a sorted structure and if I iterate over them the entries should be ordered according to the key - however I was thinking of replacing it with an unordered_map for speed reason. In any case rebuilding the time ordering in the raw to cell converter would then have no impact on the event builder and this solution could be used.

Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
mfasDa: OK, then I will remove the check.

Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
mfasDa: @shahor02 I do the connection of the ranges now in a different way: When I iterate over trigger records from all subtimeframes I cache the index in the gsl span for the subspec range. The first find call is still kept, however as I am using a std::unordered_map for this the overhead should be small.

mfasDa:
With respect to the first implementation I reduced another overhead: The cells are now stored directly on the output container. The sorting according to tower ID is done in a range of the output vector handled by pointer arithmetics. The (unnecessary) copy from the event container to the output container is removed.

mfasDa:
@shahor02 I have one question: How can I test this? This needs to mimic a muti-node setup offline. Is there a way to produce a setup offline consisting of two  (FLP) senders and one (EPN) receiver? Is the Timeframe builder blocking? (i.e. when the senders are started not in sync - i.e. as 2 different terminals by hand - that it waits until both data packages are received and not sends while one sender is still processing?)

shahor02:
@mfasDa The only way to test this locally I see is
1) hack the RawToCellConverterSpec.cxx to accept an extra option with e.g. comma-separated feeIDs, to which it would subsrcribe AND which also would modify the name of the device, e.g. to EMCALRawToCellConverterSpec_<first_feeID_of_the_list>.
2) hack your RecoWorkflow to create e.g. 2 EMCALRawToCellConverterSpec instances by calling 
specs.emplace_back(o2::emcal::reco_workflow::getRawToCellConverterSpec(askDISTSTF, subspecification + extra list of FeeIDs (= subspecs) it will bind to.

Then, o2-raw-file-reader-workflow --input-conf raw/EMC/EMCraw.cfg | o2-emcal-reco-workflow ...
will create 2 converters whose inputs you can merge.

mfasDa:
Trying to reopen the PR but seems I can't - this is of relevance for porting the raw -> cell conversion to the FLP. @shahor02 was there anything missing that prevents merging the PR?

shahor02:
@mfasDa I also don't see reopen button, perhaps @ktf can do this. 
As for the merging, was it tested? I remember there was a problem of defining FLP-specific subspecs.

ktf:
Apparently I cannot reopen it either. Was the branch removed?

mfasDa:
> Apparently I cannot reopen it either. Was the branch removed?

No, it was not.  I have opened a new PR in the mean time  #7639 as I was unable to re-open that one.

mfasDa:
Seems in case the PR is closed by the stale bot I cannot open it myself any more. OK, for the future I will keep the PR open manually in case it need some more time to fix.

PR #6836, opened 2021-08-11
[MRRTF-138] fix improper usage of TrackReference DetectorID

aphecetche:
@dstocco @pillot

### DIFF
--- a/Detectors/MUON/MCH/Workflow/src/TrackMCLabelFinderSpec.cxx
+++ b/Detectors/MUON/MCH/Workflow/src/TrackMCLabelFinderSpec.cxx
@@ -23,6 +23,8 @@
 
 #include <gsl/span>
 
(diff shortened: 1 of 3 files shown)


Detectors/MUON/MCH/Workflow/src/TrackMCLabelFinderSpec.cxx
```diff
@@ -141,7 +143,7 @@ class TrackMCLabelFinderTask
         cluster.z = (cluster.z + trackRef.Z()) / 2.;
```
pillot, line 143: L139 should be:
if (trackRef.getUserId() == deId) {

The rest looks fine to me.

aphecetche: Indeed! Fixed.

sawenzel (approved):

PR #6837, opened 2021-08-11
Fix TOF matches index access in RecoContainer

shahor02:


shahor02:
Almost trivial fix, tested locally. Merging since current some workflows might be broken

PR #6842, opened 2021-08-13
Account for change in ITS cluster refs (PR6835)

shahor02:
@davidrohr this is incremental to https://github.com/AliceO2Group/AliceO2/pull/6835: now the ITS track cluster indices are counted wrt TF start rather than ROF start, no need to search for the ROF (but the ITS tracking of existing data has to be redone).


shahor02:
failure in fullCL is due to the MFT problem which was fixed afterwards, merging.

PR #6847, opened 2021-08-13
PWGHF: Modification of D0 task for step-by-step efficiency

strogolo:


jgrosseo (approved):

strogolo:
Hi @jgrosseo, @vkucera, @ginnocen, @iarsene
I see that the PR fails the build/AliceO2/O2/o2/macOS and in particular it fails the test_Framework_test_SlowConsumer (TimeOut). 
Nevertheless, I cannot find any explicit statement where the error is. Do you have any hint or suggestion to fix the problem?

Thanks,
Stefano

PR #6851, opened 2021-08-13
Default initialization of data members

dstocco:
Data members were not explicitly initialized

### DIFF
--- a/Detectors/MUON/MID/Clustering/include/MIDClustering/PreClustersDE.h
+++ b/Detectors/MUON/MID/Clustering/include/MIDClustering/PreClustersDE.h
@@ -53,8 +53,7 @@ class PreClustersDE
   BP& getPreClusterBP(int icolumn, int idx) { return mPreClustersBP[icolumn][idx]; }
 
   /// Gets pre-cluster in the BP (const version)
-  const BP&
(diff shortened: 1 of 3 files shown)


aphecetche (dismissed):

Detectors/MUON/MID/Clustering/include/MIDClustering/PreClustersDE.h
```diff
@@ -80,9 +79,9 @@ class PreClustersDE
   void reset();
 
  private:
-  int mDEId = 99;                                ///< Detection element ID
```
ktf: Frankly, I would prefer `= 99` and not {99}. Also, what is the point of adding {}? That is implied for containers because they have a default constructor.

dstocco: Well, in the coding convention the = sign is preferred but the brace initialisation is also possible. Anyways I changed back to assignment. Also, I removed the {}. I only keep the initialisation of the functional. 
The commits can be squashed.
Thanks.
Cheers,
Diego

dstocco:
Hi @ktf : can this be merged?

PR #6854, opened 2021-08-15
Extend o2-sim --field +-5U option to arbitrary <int_kGaus>U

shahor02:
and make sure the fastField param is invoked only when it is
consistent with full field settings.

@ddobrigk : with this you can invoke o2-sim with e.g. `--field 20U` option to obtain uniform 2T field.


ddobrigk:
Hi @shahor02, super thanks!! @sawenzel , I am afraid we need this quite urgently for Alice 3. Could you please give this extra attention? I do apologize for being pushy.

ddobrigk:
@sawenzel many many thanks!!

PR #6856, opened 2021-08-16
Fix HF workflow

ktf:


ktf:
@ginnocen @jgrosseo this seems to fix the issue with the HF workflow on my laptop, I am now testing also on hyperloop.

ktf:
Full CI works, also tested locally on macOS and on hyperloop. Merging.

PR #6860, opened 2021-08-17
DPL: reduce iterations in the slow consumer test

ktf:


ktf:
This will not harm, but the macOS builds are slower that's what needs to be fixed.

PR #6862, opened 2021-08-17
DPL: avoid one more race condition

ktf:


ktf:
@TimoWilken This is another example: it was bumped back in the queue for build/O2/o2-dataflow (now 6th, used to be 2nd).

ktf:
Merging this to have at least one build with the workaround, to try it on the grid. I will probably revert it later today to have a better fix.

PR #6863, opened 2021-08-17
Secondary vertexing for forward rapidities:

sadekr:
- Added FwdDCAFitter tool for fwd tracks secondary vertexing.
  - Helix Helper adapted for both Fwd and central tracks.
  - Some functions added on TrackFwd to accomodate changes.

### DIFF
--- a/DataFormats/Reconstruction/include/ReconstructionDataFormats/TrackFwd.h
+++ b/DataFormats/Reconstruction/include/ReconstructionDataFormats/TrackFwd.h
@@ -21,6 +21,8 @@
 #include <TMath.h>
 #include "Math/SMatrix.h"
 #include "MathUtils/Utils.h"
+#include "ReconstructionDataFormats/TrackUtils.h"
+#include "MathUtils/Primitive2D.h"
 
 namespace o2
 {
@@ -55,8 +57,23 @@ class TrackParFwd
   void setPhi(Double_t phi) { mParameters(2) = phi; }
   Double_t getPhi() const { return mParameters(2); }
 
+  Double_t getSnp() const
+  {
+    return o2::math_utils::sin(mParameters(2));
+  }
+
+  Double_t getCsp2() const
+  {
+    auto snp = o2::math_utils::sin(mParameters(2));
+    Double_t csp;
+    csp = std::sqrt((1. - snp) * (1. + snp));
+    return csp * csp;
+  }
+
   void setTanl(Double_t tanl) { mParameters(3) = tanl; }
   Double_t getTanl() const { return mParameters(3); }
+
+  Double_t getTgl() const { return mParameters(3); } // for the sake of helixhelper
 
   void setInvQPt(Double_t invqpt) { mParameters(4) = invqpt; }
   Double_t getInvQPt() const { return mParameters(4); } // return Inverse charged pt
@@ -69,6 +86,12 @@ class TrackParFwd
   Double_t getInverseMomentum() const { return 1.f / getP(); }
 
   Double_t getEta() const { return -TMath::Log(TMath::Tan((TMath::PiOver2() - TMath::ATan(getTanl())) / 2)); } // return total momentum
+
+  Double_t getCurvature(double b) const
+  {
+    auto invqpt = getInvQPt();
+    return o2::constants::math::B2C * b * invqpt;
+  }
 
   /// return the charge (assumed forward motion)
   Double_t getCharge() const { return TMath::Sign(1., mParameters(4)); }
@@ -96,6 +119,7 @@ class TrackParFwd
   void propagateParamToZlinear(double zEnd);
   void propagateParamToZquadratic(double zEnd, double zField);
   void propagateParamToZhelix(double zEnd, double zField);
+  void getCircleParams(float bz, o2::math_utils::CircleXY<float>& c, float& sna, float& csa) const;
 
  protected:
   Double_t mZ = 0.; ///< Z coordinate (cm)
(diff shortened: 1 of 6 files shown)


shahor02 (dismissed):
Conditionally approving just to trigger CI, don't merge

DataFormats/Reconstruction/include/ReconstructionDataFormats/TrackFwd.h
```diff
@@ -55,9 +57,24 @@ class TrackParFwd
   void setPhi(Double_t phi) { mParameters(2) = phi; }
   Double_t getPhi() const { return mParameters(2); }
 
+  Double_t getSnp() const
```
aphecetche, line 60: Hi Rita,

Thanks for the PR. I have some small comments. For the bulk of the PR (`FwdDCAFitterN.h`)  I'll let more knowledgable people comment on...  

Concerning `getSnp` and `getCsp2` I'd have a couple of remarks :

- given that they use data (phi) that are publicly available (getPhi) I don't see the point of making those member functions (I would see the point if they were internally cached or something, but that's not the case here)  

- why use such cryptic names ? Might be obvious to you, but honestly when I see `getSnp` I don't immediately understand `getSinPhi` ;-)  Would matter less in an implementation file, but here we are at the interface level.

DataFormats/Reconstruction/include/ReconstructionDataFormats/TrackFwd.h
```diff
@@ -96,6 +119,7 @@ class TrackParFwd
   void propagateParamToZlinear(double zEnd);
   void propagateParamToZquadratic(double zEnd, double zField);
   void propagateParamToZhelix(double zEnd, double zField);
+  void getCircleParams(float bz, o2::math_utils::CircleXY<float>& c, float& sna, float& csa) const;
```
aphecetche, line 122: A comment for this method would be welcome : what is this "circle" ? (and looking at the implementation I see two cases, maybe worth nothing them briefly in the comment if that's relevant) ?

Detectors/Vertexing/src/FwdDCAFitterN.cxx
```diff
@@ -0,0 +1,33 @@
...
+/// \author ruben.shahoyan@cern.ch, adapted from central barrel to fwd rapidities by Rita Sadek, rita.sadek@cern.ch
+
+#include "DetectorsVertexing/FwdDCAFitterN.h"
+
+namespace o2
+{
+namespace vertexing
+{
+
+void __test_instance__()
```
aphecetche, line 23: what is this one exactly ?

DataFormats/Reconstruction/include/ReconstructionDataFormats/TrackFwd.h
rpezzi: Hi Laurent, these methods and names were added to keep symmetry with [central barrel track classes](https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Reconstruction/include/ReconstructionDataFormats/TrackParametrization.h#L150-L160) and allow their use in template classes, such as HelixHelper. This is not the (yet) the case for `getSnp` and `getCsp2`. Rita could rename/remove these two in this PR, but this would only increase entropy by diverging from central barrel, unless this is also renamed in other classes in ReconstructionDataFormat.

aphecetche (approved):

Detectors/Vertexing/src/FwdDCAFitterN.cxx
shahor02: To trigger templates specialization.

Detectors/Vertexing/include/DetectorsVertexing/FwdDCAFitterN.h
```diff
@@ -0,0 +1,1266 @@
...
+
+  static_assert(sizeof...(args) == N, "incorrect number of input tracks");
+  assign(0, args...);
+  clear();
+
+  for (int i = 0; i < N; i++) {
+    mTrAux[i].set(*mOrigTrPtr[i], mBz);
+  }
+
+  if (!mCrossings.set(mTrAux[0], *mOrigTrPtr[0], mTrAux[1], *mOrigTrPtr[1])) { // even for N>2 it should be enough to test just 1 loop
```
shahor02, line 311: @sadekr If I understood correctly, you are seeding in the same way as for the barrel DCAFitter: taking the intersection of 2 circles in X,Y. Not should this is reliable: given a very small lever arm and large errors on helix parameters, such a seed at best will be far from the final convergence point (while the errors are fixed at the seeding point), at worst will not converge. I think for the seeding you should use the point of closest approach in Z in the straight line approximation

sadekr: Dear Ruben,
As you mentioned, we can not rely on information given from central barrel in forward rapidities. 
This is the reason to the developments of the different seed finder methods. 
As for the HelixHelper,  I only use it as an estimation. 
So it gives us information on the nCrossing (that of course in forward will be mostly 1) 
And also an estimation on X and Y, but that is only used for the conditions. For the actual secondary vertexing; 
the seed finder is used to find a Z (which we can then use to find X Y for each track), and for the time I use the dichotomy method (findZatXY_mid).

Detectors/Vertexing/include/DetectorsVertexing/FwdDCAFitterN.h
```diff
@@ -0,0 +1,1266 @@
...
+      continue;
+    }
+
+    mCrossIDCur = ic;
+    mCrossIDAlt = (mCrossings.nDCA == 2 && mAllowAltPreference) ? 1 - ic : -1; // works for max 2 crossings
+    mNIters[mCurHyp] = 0;
+    mTrPropDone[mCurHyp] = false;
+    mChi2[mCurHyp] = -1.;
+
+    findZatXY_mid(mCurHyp);
```
shahor02, line 339: Hi @sadekr OK, just noticed an independent seeding in `findZatXY_mid` (did not check details), but then why helical crossing and the associated rejections are used? The helical estimate with just a few cm level arm is simply not reliable, it is risky to take any decisions based on it.

shahor02 (approved):
Hi @sadekr 
Looks fine. One should also add the test which you already have, but this can be done in the next PR.

PR #6865, opened 2021-08-18
TRD event display feed workflow

jbarrella:
This is an O2 workflow for preparing data for use by the TRD event display created by @samperumal. The code can be found at https://github.com/samperumal/alice-trd-event-display. Sam also wrote an AliRoot analysis task that I have now ported to O2. 

The workflow requires input of tracks, tracklets, and digits for full visualization. Currently, it uses the `MATCH_ITSTPC` object for tracks. The outputs are one JavaScript `.js` file containing track and tracklet data and one `.json` file for each track for which digit data was found. The workflow is set to only process the first 5 events at the moment.

An example of data from a pp simulation can already be seen here: https://alicetrd.web.cern.ch/alicetrd/eventdisplay/. Not sure whether or not it makes the most sense to use this location for the server. We will also have to think about how to put the workflow output into the right location for the server to pick up automatically in the future.

There are a few missing pieces of event-level information that I am not sure how to retrieve. These are things like beam energy, beam type, and magnetic field. This needs to be added later.

Please let me know suggestions/changes.

### DIFF
--- a/Detectors/TRD/workflow/include/TRDWorkflow/TRDEventDisplayFeedSpec.h
+++ b/Detectors/TRD/workflow/include/TRDWorkflow/TRDEventDisplayFeedSpec.h
@@ -1,0 +1,60 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include <gsl/span>
+
+#include "TRDBase/Geometry.h"
+#include "TRDBase/TrackletTransformer.h"
+
+#include "Framework/Task.h"
+#include "Framework/DataProcessorSpec.h"
+
+#include "DataFormatsGlobalTracking/RecoContainer.h"
+#include "DataFormatsTRD/TrackTRD.h"
+#include "DataFormatsTRD/TriggerRecord.h"
+#include "DataFormatsTRD/Digit.h"
+
+#include "ITStracking/json.h"
+
+using json = nlohmann::json;
+
+namespace o2
+{
+namespace trd
+{
+
+class TRDEventDisplayFeedSpec : public o2::framework::Task
+{
+ public:
+  TRDEventDisplayFeedSpec(int nEventsMax) : mNeventsMax(nEventsMax){};
+  ~TRDEventDisplayFeedSpec() override = default;
+  void init(o2::framework::InitContext& ic) override;
+  void run(o2::framework::ProcessingContext& pc) override;
+  json getTracksJson(gsl::span<const TrackTRD> tracks, gsl::span<const Tracklet64> tracklets, gsl::span<const TrackTriggerRecord> trackTrigRecs, int iEvent);
+  json getTrackletsJson(gsl::span<const Tracklet64> tracklets, int iEvent);
+  void writeDigits(gsl::span<const Digit> digits, int iEvent);
+
+ private:
+  TrackletTransformer mTransformer;
+  o2::trd::Geometry* mGeo;
+
+  std::map<std::string, std::string> mTrackletMap;
+  std::bitset<constants::MAXCHAMBER> mUsedDetectors;
+  gsl::span<const TriggerRecord> mTrigRecs;
+
+  float mBz;
+  int mNeventsMax;
+};
+
+o2::framework::DataProcessorSpec getTRDEventDisplayFeedSpec(int nEventsMax);
+
+} // end namespace trd
+} // end namespace o2
--- a/Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
+++ b/Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
@@ -1,0 +1,283 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include "TMath.h"
+
+#include "TRDWorkflow/TRDEventDisplayFeedSpec.h"
+
+#include "DataFormatsTRD/TrackTriggerRecord.h"
+#include "DataFormatsTRD/Tracklet64.h"
+#include "DataFormatsTRD/CalibratedTracklet.h"
+#include "DataFormatsParameters/GRPObject.h"
+
+#include "DetectorsBase/Propagator.h"
+#include "Field/MagneticField.h"
+#include "DetectorsCommonDataFormats/DetID.h"
+
+using namespace o2::framework;
+using namespace o2::globaltracking;
+
+namespace o2
+{
+namespace trd
+{
+
+void TRDEventDisplayFeedSpec::init(o2::framework::InitContext& ic)
+{
+  LOG(info) << "Initializing event display feed...";
+}
+
+json TRDEventDisplayFeedSpec::getTracksJson(gsl::span<const TrackTRD> tracks, gsl::span<const Tracklet64> tracklets, gsl::span<const TrackTriggerRecord> trackTrigRecs, int iEvent)
+{
+  json trackArray = json::array();
+
+  const auto& trigRec = trackTrigRecs[iEvent];
+  for (int iTrack = 0; iTrack < trigRec.getNumberOfTracks(); ++iTrack) {
+    TrackTRD track = tracks[iTrack + trigRec.getFirstTrack()];
+    std::string trackId = Form("E%d_T%d", iEvent, iTrack);
+
+    float alpha = track.getAlpha();
+    if (alpha < 0) {
+      alpha += TMath::Pi() * 2;
+    }
+
+    int sector = TMath::Nint(18.0 * alpha / (2 * TMath::Pi()) - 0.5);
+    int stack;
+    for (int iLayer = 0; iLayer < 6; iLayer++) {
+      int trackletIdx = track.getTrackletIndex(iLayer);
+      if (trackletIdx != -1) {
+        int detector = tracklets[trackletIdx].getDetector();
+        stack = detector % 30 / 6;
+        break;
+      }
+    }
+
+    // Tangent of the track momentum dip angle
+    float tanLambda = track.getParam(3);
+    float lambdaDeg = TMath::ATan(tanLambda) * 180 / TMath::Pi();
+
+    // Record stacks with tracks for look-up by writeDigits() later
+    int layerZero = mGeo->getDetector(0, stack, sector);
+    mUsedDetectors.set(layerZero);
+
+    json trackJson = {
+      {"id", trackId},
+      {"stk", stack},
+      {"sec", sector},
+      {"typ", "Trd"},
+      {"i", {{"pT", track.getPt()}, {"alpha", track.getAlpha()}, {"lambda", lambdaDeg}, {"pid", (int)track.getPID().getID()}}},
+      {"path", json::array()},
+      {"tlids", json::array()}};
+
+    bool ok;
+    for (int x = 1; x <= 470; x += 10) {
+      auto xyz = track.getXYZGloAt(x, mBz, ok);
+      json point = {{"x", xyz.X()}, {"y", xyz.Y()}, {"z", xyz.Z()}};
+      trackJson["path"].push_back(point);
+    }
+
+    // Match TRD tracklets to track
+    for (int iLayer = 0; iLayer < 6; iLayer++) {
+      // trackletIdx gives absolute index of tracklet across all events
+      int trackletIdx = track.getTrackletIndex(iLayer);
+      if (trackletIdx != -1) {
+        int trackletCount = 0;
+        // count number of tracklets in all events prior to iEvent
+        for (int i = 0; i < iEvent; i++) {
+          trackletCount += mTrigRecs[i].getNumberOfTracklets();
+        }
+        // trackletId needs relative tracklet index within single event
+        std::string trackletId = Form("E%d_L%d", iEvent, trackletIdx - trackletCount);
+        // record {trackletId, trackId} pair in order to match tracks to tracklets in printTracklets()
+        mTrackletMap.insert(std::pair<std::string, std::string>(trackletId, trackId));
+        trackJson["tlids"].push_back(trackletId);
+      }
+    }
+    trackArray.push_back(trackJson);
+  }
+  return trackArray;
+}
+
+json TRDEventDisplayFeedSpec::getTrackletsJson(gsl::span<const Tracklet64> tracklets, int iEvent)
+{
+  json trackletArray = json::array();
+
+  const auto& trigRec = mTrigRecs[iEvent];
+  for (int iTracklet = 0; iTracklet < trigRec.getNumberOfTracklets(); ++iTracklet) {
+    Tracklet64 tracklet = tracklets[iTracklet + trigRec.getFirstTracklet()];
+    CalibratedTracklet cTracklet = mTransformer.transformTracklet(tracklet);
+
+    std::string trackletId = Form("E%d_L%d", iEvent, iTracklet);
+    // Find matched track if it exists
+    std::string trackId = (mTrackletMap.find(trackletId) != mTrackletMap.end() ? mTrackletMap.at(trackletId) : "null");
+
+    int detector = tracklet.getDetector();
+    int sector = mGeo->getSector(detector);
+    int stack = mGeo->getStack(detector);
+    int layer = mGeo->getLayer(detector);
+
+    // Start position of both raw and calibrated tracklet in event display
+    float localY = cTracklet.getY();
+    // Slope of raw tracklet (key dyDxAN in JSON)
+    float rawDyDx = tracklet.getUncalibratedDy() / mGeo->cdrHght();
+    // Slope of calibrated tracklet
+    float dyDx = cTracklet.getDy() / mGeo->cdrHght();
+
+    json trackletJson = {
+      {"id", trackletId},
+      {"stk", stack},
+      {"sec", sector},
+      {"lyr", layer},
+      {"row", tracklet.getPadRow()},
+      {"trk", trackId},
+      {"lY", localY},
+      {"dyDx", dyDx},
+      {"dyDxAN", rawDyDx}};
+
+    if (trackId == "null") {
+      trackletJson["trk"] = nullptr;
+    }
+
+    trackletArray.push_back(trackletJson);
+  }
+  return trackletArray;
+}
+
+void TRDEventDisplayFeedSpec::writeDigits(gsl::span<const Digit> digits, int iEvent)
+{
+  const auto& trigRec = mTrigRecs[iEvent];
+  for (int det = 0; det < constants::MAXCHAMBER; det += 6) {
+    if (mUsedDetectors[det]) {
+      int sector = mGeo->getSector(det);
+      int stack = mGeo->getStack(det);
+
+      json digitsJson = {
+        {"evid", iEvent},
+        {"lyrs", json::array()}};
+
+      for (int iLayer = 0; iLayer < 6; iLayer++) {
+        json layerJson = {
+          {"lyr", iLayer},
+          {"pads", json::array()}};
+
+        for (int iDigit = trigRec.getFirstDigit(); iDigit < trigRec.getFirstDigit() + trigRec.getNumberOfDigits(); ++iDigit) {
+          Digit digit = digits[iDigit];
+
+          int detector = digit.getDetector();
+
+          if (detector == det + iLayer) {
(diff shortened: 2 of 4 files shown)


Detectors/TRD/base/include/TRDBase/EventDisplayFeed.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef O2_TRD_EVENTDISPLAYFEED_H
+#define O2_TRD_EVENTDISPLAYFEED_H
+
+namespace o2
+{
+namespace trd
+{
+
+class EventDisplayFeed
```
martenole: It seems this class isn't used anywhere. Is it needed?

Detectors/TRD/workflow/include/TRDWorkflow/TRDEventDisplayFeedSpec.h
```diff
@@ -0,0 +1,58 @@
...
+namespace o2
+{
+namespace trd
+{
+
+class TRDEventDisplayFeedSpec : public o2::framework::Task
+{
+ public:
+  TRDEventDisplayFeedSpec(std::shared_ptr<o2::globaltracking::DataRequest> dataRequest) : mDataRequest(dataRequest){};
+  ~TRDEventDisplayFeedSpec() override = default;
```
martenole, line 38: Don't you need to close `summary` here?

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
```diff
@@ -0,0 +1,379 @@
...
+    if (alpha < 0)
+      alpha += TMath::Pi() * 2;
+
+    int sector = TMath::Nint(18.0 * alpha / (2 * TMath::Pi()) - 0.5);
+
+    // tangent of the track momentum dip angle
+    float tanLambda = track.getParam(3);
+    float lambdaDeg = TMath::ATan(tanLambda) * 180 / TMath::Pi();
+
+    int stack;
```
martenole: This is a bit vague. For example in your display I see for event 0, track 4 that it is found in stack 3, but actually the tracklets are attached in stack 2. Couldn't you simply assign the stack number from the first tracklet which is attached to the track?

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
```diff
@@ -0,0 +1,379 @@
...
+
+    *summary << "\t\t\t\t"
+             << "}," << std::endl;
+    *summary << "\t\t"
+             << "\"path\": [" << std::endl
+             << "\t\t\t";
+
+    // TODO: Magnetic field. Is this number available somewhere?
+    // -5 default value for o2-sim
+    int b = -5;
```
martenole: You can have a look e.g. in Detectors/TRD/workflow/src/TrackBasedCalibSpec.cxx:
```
o2::base::GeometryManager::loadGeometry();
o2::base::Propagator::initFieldFromGRP();
auto prop = o2::base::Propagator::Instance();
auto bz = prop->getNominalBz();
```

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
```diff
@@ -0,0 +1,379 @@
...
+      stack = 1;
+    else if (lambdaDeg > -8)
+      stack = 2;
+    else if (lambdaDeg > -30)
+      stack = 3;
+    else
+      stack = 4;
+
+    // TODO: Is there any use for this? (using trdmatches_itstpc.root)
+    bool isTrd = true;
```
martenole: I don't think so, since we don't have ESD tracks anymore

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
```diff
@@ -0,0 +1,379 @@
...
+             << "\"," << std::endl
+             << "\t\t\t"
+             << "}," << std::endl;
+
+    // Get only tracks, tracklets, digits for the current event
+    int start = trackTrigRec.getFirstTrack();
+    int end = start + trackTrigRec.getNumberOfTracks();
+    auto eventTracks = std::vector<TrackTRD>(tracks.begin() + start, tracks.begin() + end);
+
+    printTracks(eventTracks, iEvent);
```
martenole: I don't see why you need to copy the data so often. You could simply inside `printTracks(iEvent)` loop only over the tracks of the given event. The same also for tracklets and digits.

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
```diff
@@ -0,0 +1,379 @@
...
+    for (int det = 0; det < 540; det++)
+      mUsedDetectors[det] = false;
+
+    if (iEvent > 0)
+      *summary << "," << std::endl;
+    else
+      *summary << std::endl;
+
+    // TODO: Is this information available somewhere?
+    int beamEnergy = -1;
```
martenole: Should also be in the GRP

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
```diff
@@ -0,0 +1,379 @@
...
+  *summary << "\t"
+           << "];" << std::endl
+           << "}" << std::endl;
+}
+
+o2::framework::DataProcessorSpec getTRDEventDisplayFeedSpec()
+{
+  std::shared_ptr<DataRequest> dataRequest = std::make_shared<DataRequest>();
+
+  auto& inputs = dataRequest->inputs;
```
martenole: You can also use here the RecoContainer functionality and the InputHelper. Then you don't need to specify the input by hand. This is done for example in Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/src/TPCInterpolationSpec.cxx:
```
auto dataRequest = std::make_shared<DataRequest>();
dataRequest->requestTracks(o2::dataformats::GlobalTrackID::getSourcesMask("TRD"), false);
dataRequest->requestClusters(o2::dataformats::GlobalTrackID::getSourcesMask("TRD"), false);
```
In your run method you need to to:
```
  RecoContainer recoData;
  recoData.collectData(pc, *mDataRequest.get());
```
The TRD tracks are then available via `recoData.getITSTPCTRDTracks()`. Same also for tracklets and calibrated tracklets.

Detectors/TRD/workflow/src/TRDEventDisplayFeedWorkflow.cxx
```diff
@@ -0,0 +1,48 @@
...
+  o2::conf::ConfigurableParam::updateFromString(configcontext.options().get<std::string>("configKeyValues"));
+
+  WorkflowSpec spec;
+
+  bool useMC = true;
+  if (!configcontext.options().get<bool>("disable-root-input")) {
+    spec.emplace_back(o2::trd::getTRDGlobalTrackReaderSpec(useMC));
+    spec.emplace_back(o2::trd::getTRDTrackletReaderSpec(useMC, false));
+    spec.emplace_back(o2::trd::getTRDDigitReaderSpec(1, useMC));
+  }
```
martenole, line 44: Here you can use the InputHelper as for example in Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/src/tpc-interpolation-workflow.cxx:
```
  GID::mask_t sources = GID::getSourcesMask("TRD,TPC-TRD,ITS-TPC-TRD");
o2::globaltracking::InputHelper::addInputSpecs(configcontext, spec, sources, sources, sources, false);
```

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
martenole: I also see that for that track the digit information is not filled

martenole (commented):
Hi Jason, thanks a lot, this is very nice to have! A few comments to the code inline. Cheers, Ole

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
jbarrella: Hi @martenole. Thanks so much for all the comments. How do I receive digits in this way? Or do I just continue using my current way for those?

martenole: Hm right, for the digits this does not work. You could use the DataRequest for tracklets and tracks, but for the digits you still need to add the reader manually. Or you keep doing everything manually, but in that case you can remove the DataRequest entirely. So I guess it is up to you ;)

jbarrella: Thanks, Ole. I made an attempt to implement this but battled to get it working and it was beginning to get quite complicated. I think, in this case, doing it manually works out to be quite straightforward and fulfills the purpose so going to go with that :)

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
jbarrella: Thanks, can't remember why on Earth I did it this way...

Detectors/TRD/base/include/TRDBase/EventDisplayFeed.h
jbarrella: Yeah, I thought so too, but when I delete the file, O2 fails to compile because it looks for this file. Not sure why?

Detectors/TRD/workflow/include/TRDWorkflow/TRDEventDisplayFeedSpec.h
jbarrella: Should I close it at the end of `TRDEventDisplayFeedSpec::run()`?

martenole: Probably, since you open it at the beginning of that function. And I would also change it from `std::ofstream* summary` to `std::ofstream mSummary` (naming convention + avoid usage of raw `new` which is also not deleted anywhere)

Detectors/TRD/base/include/TRDBase/EventDisplayFeed.h
martenole: That cannot be, since it is not even added to a CMakeLists.txt file. Whether the file is present or not does not make a difference

Detectors/TRD/workflow/include/TRDWorkflow/TRDEventDisplayFeedSpec.h
```diff
@@ -0,0 +1,57 @@
...
+ private:
+  o2::trd::Geometry* mGeo;
+  TrackletTransformer mTransformer;
+  float mBz;
+
+  gsl::span<const TriggerRecord> mTrigRecs;
+  std::map<std::string, std::string> mTrackletMap;
+  std::ofstream* summary = new std::ofstream();
+
+  bool mUsedDetectors[540];
```
jolopezl: Prefer `std::array<bool, constants::MAXCHAMBER> mUsedDetectors{};`. You will need to include `DataFormatsTRD/Constants.h`.

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
```diff
@@ -0,0 +1,378 @@
...
+    const auto& trigRec = mTrigRecs[iEvent];
+    const auto& trackTrigRec = trackTrigRecs[iEvent];
+
+    if (trackTrigRec.getNumberOfTracks() == 0) {
+      continue;
+    }
+
+    for (int det = 0; det < 540; det++) {
+      mUsedDetectors[det] = false;
+    }
```
jolopezl: No need for this if you consider using the std::array with brace-initialization.

jolopezl (commented):
This is great. The event display by @samperumal is very cool, good-looking, and responsive. I added a minor code comment, but I have an important suggestion. Currently, you are using an `ofstream` pointer (that seems to open), which makes the implementation complicated, as seen in the code. It would be more helpful to rely on some JSON parser. Currently, the O2 codebase have dependencies in boost.json, rapidjson, and nholmann::json – all of them with pros and cons -. It would be very cool to have a cleaner implementation of the parsing with one of these libraries. See for example, https://github.com/nlohmann/json#examples where it shows how to push different types to the JSON object without much processing, also dumping the JSON object to a file is trivial. This is not mandatory since it requires some time to get it implemented, but it would be something nice to have.

Detectors/TRD/base/include/TRDBase/EventDisplayFeed.h
jbarrella: `2021-08-19@10:31:09:DEBUG:O2:O2:0: Detectors/TRD/base/G__O2TRDBase.cxx:62:10: fatal error: /root/alice/sw/SOURCES/O2/event-display-workflow/0/Detectors/TRD/base/include/TRDBase/EventDisplayFeed.h: No such file or directory
2021-08-19@10:31:09:DEBUG:O2:O2:0:    62 | #include "/root/alice/sw/SOURCES/O2/event-display-workflow/0/Detectors/TRD/base/include/TRDBase/EventDisplayFeed.h"
2021-08-19@10:31:09:DEBUG:O2:O2:0:       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2021-08-19@10:31:09:DEBUG:O2:O2:0: compilation terminated.`

I don't know how to explain it either...

The only possible connection I see is that the event display workflow is linked to `O2::TRDWorkflow` which is linked to `O2::TRDBase`. Maybe it expects this header file because of this, but I'm just guessing.

jolopezl:
On the same track: I think that https://github.com/samperumal/alice-trd-event-display should become a part of the AliceO2 codebase.

Detectors/TRD/base/include/TRDBase/EventDisplayFeed.h
martenole: But since EventDisplayFeed is not part of `O2::TRDBase` (have a look at Detectors/TRD/base/CMakeLists.txt where the library is defined) it is not needed at all.
I suspect that your build environment is somehow not cleaned up. You can try to remove everything from your O2 build directory and build again. Unfortunately this does recompile O2 completely of course :/
I bet that if you remove that file and push it to github O2 will compile here just fine ;)
But you need to fix the erros of the fullCI:
`/sw/SOURCES/O2/6865-slc8_x86-64/0/Detectors/TRD/workflow/include/TRDWorkflow/TRDEventDisplayFeedSpec.h:34:3: error: use '= default' to define a trivial default constructor [modernize-use-equals-default]`

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
jbarrella: I think I still need this since I have to reset it to false after each event

jolopezl: Still... :), if you change to std::array you can just use here: `mUsedDetectors = { false };` and drop the for-loop.

jbarrella:
> This is great. The event display by @samperumal is very cool, good-looking, and responsive. I added a minor code comment, but I have an important suggestion. Currently, you are using an `ofstream` pointer (that seems to open), which makes the implementation complicated, as seen in the code. It would be more helpful to rely on some JSON parser. Currently, the O2 codebase have dependencies in boost.json, rapidjson, and nholmann::json – all of them with pros and cons -. It would be very cool to have a cleaner implementation of the parsing with one of these libraries. See for example, https://github.com/nlohmann/json#examples where it shows how to push different types to the JSON object without much processing, also dumping the JSON object to a file is trivial. This is not mandatory since it requires some time to get it implemented, but it would be something nice to have.

Okay, excellent. I did not realize that we had JSON libraries available. This will make things a lot more readable and easy to work with. Let me take a look into that.

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
martenole: ..or you can go for a [std::bitest](https://en.cppreference.com/w/cpp/utility/bitset) :)

jolopezl: even cooler :)

jbarrella:
Thanks so much for the suggestions/improvements thus far @martenole, @jolopezl. I've made the move to `nlohmann::json` now which was a very significant improvement in general readability and efficiency. @jolopezl Merging the event display repo into AliceO2Group sounds like a good idea, but I'm not sure who would handle that. All other suggestions that were made here should also be implemented.

samperumal:
> 
> 
> On the same track: I think that https://github.com/samperumal/alice-trd-event-display should become a part of the AliceO2 codebase.

Hi all. It is very exciting news that the display is being used! I would be interested in merging the code into the main repo, if that makes sense. I'm also open to suggestions for improvements, if anyone has any.

martenole:
Hi Sam, it is a very nice tool indeed and it's good to have it back! :)
I have to say I don't see the point of merging it into O2 though. A big advantage of having it as a standalone application is that we can work on it without additional barriers (code checker, CI, etc). And also for using it one would not rely on O2.
But maybe I miss something?
Cheers, Ole
Edit: Maybe a fork for AliceO2Group would be an option?

Detectors/TRD/workflow/include/TRDWorkflow/TRDEventDisplayFeedSpec.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "Framework/Task.h"
+#include "Framework/DataProcessorSpec.h"
+
+#include "DataFormatsGlobalTracking/RecoContainer.h"
+#include "DataFormatsTRD/TrackTRD.h"
+#include "DataFormatsTRD/TriggerRecord.h"
+#include "DataFormatsTRD/Digit.h"
+
+#include "ITStracking/json.h"
```
jolopezl, line 25: Not urgent, but I wonder if it is worth moving this huge header to `Common/Utils/...` or somewhere else.

jolopezl (dismissed):
Looks great. Thanks for adding the JSON header, makes the code a lot better. I have a minor comment, but I don't think it's very urgent.

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
```diff
@@ -0,0 +1,290 @@
...
+
+  auto tracks = pc.inputs().get<gsl::span<TrackTRD>>("trdtracks");
+  auto tracklets = pc.inputs().get<gsl::span<Tracklet64>>("trdtracklets");
+  auto digits = pc.inputs().get<gsl::span<Digit>>("trddigits");
+
+  mTrigRecs = pc.inputs().get<gsl::span<TriggerRecord>>("trdtriggerrec");
+  auto trackTrigRecs = pc.inputs().get<gsl::span<TrackTriggerRecord>>("tracktriggerrec");
+
+  // Display only first 5 events
+  int nEvents = 5;
```
martenole: This should be compared to `mTrigRecs.size()` in order to avoid a segfault if nEvents is too large:
`int nEvents = std::min(mTrigRecs.size(), nEventsMax)`
One could also make `nEventsMax` configurable from command line.

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
```diff
@@ -0,0 +1,290 @@
...
+
+            json padJson = {
+              {"row", row},
+              {"col", col},
+              {"tbins", json::array()}};
+
+            // Event display draws 24 time bins
+            int nTimebins = 25;
+            int t = 0;
+            for (auto adc : digit.getADC()) {
```
martenole, line 186: Can the event display be changed to display 30 time bins as well? And since you do need a counter in this loop I would not use a range-based for loop, but an index one instead to improve the readability.

martenole (commented):
Hi Jason, I have two more small comments, but otherwise this should be good to go.
Cheers,
Ole

Detectors/TRD/workflow/include/TRDWorkflow/TRDEventDisplayFeedSpec.h
jbarrella: Should I move it there in this PR?

jolopezl: I don't think so. I don't know where it should really go.

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
jbarrella: Thanks Ole. I have changed the javascript code in the event display to show 30 timebins.

martenole (approved):
Looks good, thanks a lot!

Detectors/TRD/workflow/src/TRDEventDisplayFeedSpec.cxx
samperumal: Thanks for the PR @jbarrella . I've merged and extended so now a variable number of time-bins is accepted (but the hard-coded 20 timebins = 3cm drift remains).

samperumal:
> 
> 
> Hi Sam, it is a very nice tool indeed and it's good to have it back! :)
> I have to say I don't see the point of merging it into O2 though. A big advantage of having it as a standalone application is that we can work on it without additional barriers (code checker, CI, etc). And also for using it one would not rely on O2.
> But maybe I miss something?
> Cheers, Ole
> Edit: Maybe a fork for AliceO2Group would be an option?

That's a fair point. I'm happy to keep it separate for now, with a view to an AliceO2Group fork in the future if necessary.

jbarrella:
@davidrohr Hi, can this be merged or do I need to rebase again?

PR #6866, opened 2021-08-18
MCH: remove vertex handling from reco workflow

aphecetche:
Let the mch-reco-workflow last stage be the mch standalone tracks (i.e.
not extrapolated), as the vertex matching is done elsewhere (except for
Run2 tests in which case the removed connection to the vertex-sampler
can easily recovered with workflow piping)


aphecetche:
@davidrohr I'd assume the fullCI error is not related to this PR ?

aphecetche:

shahor02 (approved):

PR #6870, opened 2021-08-19
[SC] Adding possibility to use float for calculations

matthias-kleiner:
- adding function to calculate the drift path for an electron
- setting default values for omega tau
- adding simple debug function
- poisson solver: cleaning + using more than 1000 phi bins

### DIFF
--- a/Detectors/TPC/reconstruction/macro/createTPCSpaceChargeCorrection.C
+++ b/Detectors/TPC/reconstruction/macro/createTPCSpaceChargeCorrection.C
@@ -222,7 +222,7 @@ void debugInterpolation(utils::TreeStreamRedirector& pcstream,
           float gxyzf[3] = {gx, gy, gz};
           float pointCyl[3] = {gz, r, phi};
           double efield[3] = {0.0, 0.0, 0.0};
-          double charge = spaceCharge->getChargeCyl(pointCyl[0], pointCyl[1], pointCyl[2], side);
+          double charge = spaceCharge->getDensityCyl(pointCyl[0], pointCyl[1], pointCyl[2], side);
           double potential = spaceCharge->getPotentialCyl(pointCyl[0], pointCyl[1], pointCyl[2], side);
           spaceCharge->getElectricFieldsCyl(pointCyl[0], pointCyl[1], pointCyl[2], side, efield[0], efield[1], efield[2]);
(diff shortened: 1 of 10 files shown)


wiechula (dismissed):
@matthias-kleiner , there are quite some changes in the PoissonSolver. I guess you checked them against some benchmark (analytical solution)?

matthias-kleiner:
> @matthias-kleiner , there are quite some changes in the PoissonSolver. I guess you checked them against some benchmark (analytical solution)?

Hi, yes there are no differences when using doubles. When using floats there is a slight, but not significant, improvement for the resulting potential due to rounding of floats I think. But in general it looks like that the poisson solver doesnt converge when using floats instead of doubles and the calculation time for the distortions etc. is also not significantly different, which can lead sometimes to longer calculations times for floats. The difference between the resulting distortions are only a few micro meter between double and float, so it looks like at the end it doesnt really matter so much if one uses doubles or floats. But when using a really large grid the required memory can be a problem, where floats can help.

And there is always the test for the Poisson solver comparing the result of the potential with the analytical potential.

shahor02 (dismissed):
The o2 failure is unrelated. Look fine for me (with superficial check), shall I merge it?

matthias-kleiner:
> The o2 failure is unrelated. Look fine for me (with superficial check), shall I merge it?

It seems like there is now a new error during the O2/o2-cs8  build in generateTPCCorrectionNTuple.C: `error: definition with same mangled name` or is this unrelated? I dont actually understand why this error occurs. Could it have something to do with the changes of the includes I put into [Vector.h](https://github.com/AliceO2Group/AliceO2/blob/43567bc0a54547cb1f92de2ccf4b272c3444bf81/Detectors/TPC/spacecharge/include/TPCSpaceCharge/Vector.h#L20) or is this just some random error?

shahor02:
@matthias-kleiner I think the error is real. Can't you try to use a forward declaration for the Vc::Memory?

matthias-kleiner:
> @matthias-kleiner I think the error is real. Can't you try to use a forward declaration for the Vc::Memory?

I just tried it in a Docker installation and the error is indeed real. It seems to happen when one includes Vc partially and later includes the full Vc. Reordering the include order fixed the problem in the Docker environment. However this problem on my local installation also disappears when one compiles the macro a second time after it failed once.

wiechula (approved):

sgorbuno (approved):

matthias-kleiner:
Looks like all the tests are passing now. Can this PR then be merged?

PR #6874, opened 2021-08-20
Further adaption to inputCard.xml 

pbuehler:
Account for changed names of executables and namings in CMakeLists.txt files.


pbuehler:
delete

PR #6878, opened 2021-08-21
DPL: handle region notifications as required by GPU processing

ktf:
* Adds an option, --expected-region-callbacks <N>, to wait for
  enough callback before going to start.
* Deliver callbacks on the main thread, always.

### DIFF
--- a/Framework/Core/include/Framework/DataProcessingDevice.h
+++ b/Framework/Core/include/Framework/DataProcessingDevice.h
@@ -61,6 +61,7 @@ struct DeviceContext {
   DeviceState* state = nullptr;
   ComputingQuotaEvaluator* quotaEvaluator = nullptr;
   DataProcessingStats* stats = nullptr;
+  int expectedRegionCallbacks = 0;
 };
 
 struct DataProcessorContext {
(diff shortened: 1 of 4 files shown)


ktf:
@davidrohr

davidrohr:
@ktf : thx, will check on Monday. But this seems to break many framework ctests and also the full system test fails in the FullCI. Perhaps it is not yet fully working?

ktf:
I will check. I only tested the diamond topology.

ktf:
Ok, I am missing the default argument. If you pass the option explicitly it should work. Fix coming.

ktf:
I fixed it. Is the fullCI error actually related? The option is not really enabled.

davidrohr:
The FullCI error is no different compared to the one before, I think this one now is not related to this PR.
I have seen it a couple of times recent, it is some segfault in the MFT reconstruction, but yet I have not been able to reproduce it locally.

davidrohr (dismissed):
THx, both the callbacks coming from the main thread as well as the `--expected-region-callbacks` work as expected

ktf:
Clang format merged.

PR #6880, opened 2021-08-23
Avoid being screamed at every time I store an object.

Barthelemy:


sawenzel (dismissed):
Approving. ...Apart from the fact that DEBUG messages are excluded in production builds and it will probably be difficult to ever access this information.

Also not sure why this should happen at every object. Should only happen once during initialization.

Barthelemy:
You raise good points. Actually, I was checking the logs of some tests and saw many of these messages because it was reinitialized between each test. In this case, I can put it back to INFO. And just keep something in lower case :)

sawenzel (approved):

ktf:
Trivial. Merging.

PR #6884, opened 2021-08-23
DPL: stop looking for offers if we have enough

ktf:


ktf:
@Barthelemy @knopers8 build/O2/o2-cs8 fails in QC (batch_test). Is this a known issue?

ktf:
Tested on hyperloop. No regressions there.

Barthelemy:
@ktf Not known. When I try to access the logs it shows me that it passed though.

PR #6885, opened 2021-08-23
Global offsets and reco QC

AllaMaevskaya:


AllaMaevskaya:
What is wrong here? Could it be merged?

mslupeck (approved):

mslupeck:
HI @shahor02, could you have a look here and merge if possible?

shahor02:
Hi, before merging: what is the role of the recoQC? Is not it supposed to be in the QualityControl ?

AllaMaevskaya:
Hello
it should be in QC, but I did not find QC for reconstructed data after synch reconstruction. I will add it in QC later

PR #6889, opened 2021-08-23
DPL: correctly report consumed shared memory

ktf:


ktf:
Tested on hyperloop. No regressions.

PR #6892, opened 2021-08-24
DPL: fix race condition more likely with many drivers

ktf:


ktf:
Tested in hyperloop and with the previous, not clang formatted, version. Not really exercised in the CI (besides the newly introduced unit test which did work). fullCI was green as well.

PR #6896, opened 2021-08-24
[MCH] set bcWidth in digit ROFs

aferrero2707:
The width of the digit ROFs is set to a fixed width of 4 bunch crossings, corresponding to one SAMPA ADC cycle.

### DIFF
--- a/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/ROFRecord.h
+++ b/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/ROFRecord.h
@@ -37,6 +37,7 @@ class ROFRecord
  public:
   ROFRecord() = default;
   ROFRecord(const BCData& bc, int firstIdx, int nEntries) : mBCData(bc), mDataRef(firstIdx, nEntries) {}
+  ROFRecord(const BCData& bc, int firstIdx, int nEntries, int bcWidth) : mBCData(bc), mDataRef(firstIdx, nEntries), mBCWidth(bcWidth) {}
 
   /// get the interaction record
   const BCData& getBCData() const { return mBCData; }
@@ -56,8 +57,6 @@ class ROFRecord
 
   /// get the time span by this ROF, in BC unit
   int getBCWidth() const { return mBCWidth; }
-  /// set the time span by this ROF, in BC unit
-  void setBCWidth(int bcWidth);
 
   bool operator==(const ROFRecord& other) const
   {
--- a/DataFormats/Detectors/MUON/MCH/src/ROFRecord.cxx
+++ b/DataFormats/Detectors/MUON/MCH/src/ROFRecord.cxx
@@ -23,11 +23,5 @@ std::ostream& operator<<(std::ostream& os, const ROFRecord& rof)
                     rof.getBCWidth());
   return os;
 }
-void ROFRecord::setBCWidth(int bcWidth)
-{
-  if (bcWidth < 4) {
-    throw std::invalid_argument(fmt::format("bcWidth must be strictly greater than 4 bunch-crossings"));
-  }
-}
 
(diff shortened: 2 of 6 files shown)


aphecetche:
@aferrero2707 mch rofs records are actually constructed with a bcWidth of 4 by default (https://github.com/AliceO2Group/AliceO2/blob/f4e07de955fe58375cf1a69f51ceab151438be7a/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/ROFRecord.h#L83), so I assume your modification is simply to be more explicit, or do I miss a point here ?

aferrero2707:
@aphecetche yes, I prefer to have that explicitly set since it is not part of the constructor, just for the sake of clearness and documentation...

aferrero2707:
@aphecetche I have just realized that the BC width parameter was not actually updated in the `setBCWidth()`, now it is fixed.

aphecetche:
> @aphecetche yes, I prefer to have that explicitly set since it is not part of the constructor, just for the sake of clearness and documentation...

well, there's a point to be made about favouring documentation in include file instead of in implementation file, usually I'd expect the information to be readily available in the interface (include) file, not having to dig the implementation.

but ok, putting this point aside, if we go with your changes (btw, good catch about the setBcWith not doing anyway ;-) ), I would remove the `mBcWidth{4}` data member initialisation in the include file, otherwise we have the constant "4" in two places (and hence risk disconnect and/or confusion).

aferrero2707:
Why not to have the bcWidth parameter in the constructor? At this point it is just another parameter describing the group of digits. It takes a value of 4 for the ROFs coming out of the decoder, but it can take more or less arbitrary values in ROFs after the time clustering.

I am also realizing that the bcWidth could also be computed a posteriori, given that we know the time-ordered indexes of the first and last digits, and that the digits have their own time stamps in BC units. So the bcWidth information is somewhat redundant...

At the moment the bcWidth parameter seems to be only used in `DigitIOV4.cxx`.

aphecetche:
> Why not to have the bcWidth parameter in the constructor? At this point it is just another parameter describing the group of digits. It takes a value of 4 for the ROFs coming out of the decoder, but it can take more or less arbitrary values in ROFs after the time clustering.

Indeed. Seems to be a logical thing to do. But when I introduced the `bcWidth` I did not want to change the constructor signature, for a reason that escapes me now ... So let's go for it.
 
> I am also realizing that the bcWidth could also be computed a posteriori, given that we know the time-ordered indexes of the first and last digits, and that the digits have their own time stamps in BC units. So the bcWidth information is somewhat redundant...

It's redundant only for digits but ROFRecords are used for other data types as well.

> At the moment the bcWidth parameter seems to be only used in `DigitIOV4.cxx`.

It's also used in the AODProducer to assign a time window to MCH tracks.

aferrero2707:
@aphecetche ok then I will move the parameter in the constructor and let you know. Thanks!

aferrero2707:
@aphecetche there are several places where the call to the constructor should be changed when adding one more parameter.
We can either modify the code everywhere, or use a default value for the bcWidth in the constructor. What do you prefer?

aphecetche:
> @aphecetche there are several places where the call to the constructor should be changed when adding one more parameter.
> We can either modify the code everywhere, or use a default value for the bcWidth in the constructor. What do you prefer?

now I guess that may well be the reason why I did not want to change the default constructor at the time :wink: and opted for the `bcWidth{4}` initialisation, which achieves the same goal with minimal code change...

another option would be to _add_ a 2nd constructor with the bcWidth, for those cases where the bcWidth is non default.

aferrero2707:
@aphecetche I have added a second constructor with an extra `bcWidth` parameter. Can you have a look?

Thanks!

aphecetche (approved):

PR #6902, opened 2021-08-25
TRD fix digit reader in ED workflow

martenole:
@davidrohr sorry I missed that #6865 which was merged just before #6729 uses the TRD digit reader spec for which I changed the interface. Since one data member was obsolete. This breaks O2 build now :/


martenole:
I am compiling at the moment with these changes, but the build is not yet complete

davidrohr:
ok, no problem, both PRs had passed the CI, so it was not apparent that merging them would collide. Let's test this fix and if it works merge it right away.

davidrohr:
fixes a current problem, merging

PR #6909, opened 2021-08-26
[MCH] propagate tracks to MID

pillot:
This is the first step toward matching MCH and MID tracks. It propagates the MCH track parameters and covariances at the last attached cluster to the first MID chamber, where the matching will be performed, and store them in the TrackMCH output object.

### DIFF
--- a/Detectors/MUON/MCH/Tracking/src/TrackExtrap.cxx
+++ b/Detectors/MUON/MCH/Tracking/src/TrackExtrap.cxx
@@ -249,6 +249,32 @@ bool TrackExtrap::extrapToZCov(TrackParam* trackParam, double zEnd, bool updateP
 }
 
 //__________________________________________________________________________
+bool TrackExtrap::extrapToMID(TrackParam* trackParam)
+{
+  /// Extrapolate the track parameters and their covariances to the z position of the first MID chamber
+  /// Add to the track parameter covariances the effects of multiple Coulomb scattering in the muon filter
+
+  if (trackParam->getZ() == SMIDZ) {
+    return true; // nothing to be done
+  }
+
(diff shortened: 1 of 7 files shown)


Detectors/MUON/MCH/Tracking/src/TrackExtrap.cxx
```diff
@@ -248,6 +248,32 @@ bool TrackExtrap::extrapToZCov(TrackParam* trackParam, double zEnd, bool updateP
   return true;
 }
 
+//__________________________________________________________________________
+bool TrackExtrap::extrapToMID(TrackParam* trackParam)
```
aphecetche, line 252: why do you pass a pointer (which you anyway don't test for nullptr) instead of a (const) reference ?

pillot: just for consistency with all other extrapolation methods in this class (inherited from aliroot). A reference (not const since the parameters are modified) might be preferable indeed, but if I change that, I would prefer to change consistently all the methods (in a separate commit).

aphecetche: OK. Would be nice to have references instead indeed. When you say separate commit you mean an additional commit in this PR or another PR altogether ?

pillot: I was thinking another PR, as it requires changes in other places unrelated to this PR.

aphecetche: ok, then I will merge this one as soon as I've access to the green merge button...

aphecetche (approved):

PR #6911, opened 2021-08-26
Clear working containers when starting new TF

shahor02:
@rohr: some containers were not cleared, I am now checking if this PR fixes the crash


shahor02:
@davidrohr this seems to solve the seg.fault problem, could you check too?  Also the problem in increasing N.vertices is solved.
Merging since the fix is trivial and certainly needed.

davidrohr:
just tried with 100 time frames and seems ok now, thx!

PR #6912, opened 2021-08-27
Fix in CTF reader (for multi-TF files)

shahor02:
* CTFReader: fix in passing tree event entry in multi-CTF files
* ITSMFT CTFCoder: make exception message more informative

Temporarily compile EncodedBlocks w/o NDEBUG to enable asserts.


shahor02:
Failures are unrelated, merging since tested locally

PR #6914, opened 2021-08-27
DPL: add online average for metrics

ktf:
Refactor a bit the code while doing so.


ktf:
This will be useful to try to be more fair in offering shared memory to workers.

ktf:
Tested on hyperloop. 4 stars for the four test workflows.

PR #6915, opened 2021-08-27
DPL: do not allow for user printouts on --dds, --dump

ktf:


davidrohr (approved):

davidrohr:
thx, this looks good (without having tested it...)

ktf:
@davidrohr I actually managed to come up with this. We can still do the file option, but this should avoid having it mandatory and keep it unixy. Maybe we should simply redirect all the output to stderr, if --dds / --dump is specified.

@knopers8 FYI.

davidrohr:
I have just tried this with the full system test workflow and `--dds` option, and there are still `[INFO]` printouts, so apparently it is not working.

ktf:
Is this with a merged workflow? I guess what might happen is that since only the last dataprocessor has --dds, only the last one does not print extra stuff on the command line... Can you put the TRD workflow last, just to check?

davidrohr:
Yes, it is a merged workflow, but I cannot reorder it since then the outputs are used before they are defined.
I have added `--dds` to all executables, but that didn't change anything.

davidrohr (changes requested):
Revoking approval since it is not working (don't know why, but it doesn't make sense to merge as is).
As was discussed, I think it would anyway be better to provide an option to write the DDS output to a file.

PR #6917, opened 2021-08-27
Detect Alpide error: same pixels fired multiple times

shahor02:


shahor02:
Failures are unrelated, merging since tested locally

shahor02:
merged accidentally instead of other PR, will not revert it but open another PR to pass the  tests.

PR #6921, opened 2021-08-28
Add FDD and MID reco to sim_challenge

shahor02:
@sawenzel here is it, but I was not able to test it as o2-sim crashes in merger.


sawenzel:
The crash is due to ROOT 6.24. A patched ROOT version will be deployed next week and we should be back to normal in the simulation.

sawenzel (approved):

PR #6930, opened 2021-08-30
Send different decoded MID event types to different specs

dstocco:
During data taking MID receives software calibration triggers that are used to test noisy channels.
Few BCs after the reception of the calibration trigger, the MID returns a special event (Front-End-Test event) where all channels should anwer.
These events are used to test the dead channels.
These events are not reconstructed like the physics event, but are sent to separate workflows and used to generate masks.
Since they're not used in the standard reconstruction, they will be streamed on different specs to simplify operations.

### DIFF
--- a/DataFormats/Detectors/MUON/MID/include/DataFormatsMID/ROFRecord.h
+++ b/DataFormats/Detectors/MUON/MID/include/DataFormatsMID/ROFRecord.h
@@ -26,8 +26,8 @@ namespace mid
 
 enum class EventType {
   Standard = 0,
-  Noise = 1,
-  Dead = 2
+  Calib = 1,
+  FET = 2
 };
 
 /// ROFRecord class encodes the trigger interaction record of given ROF and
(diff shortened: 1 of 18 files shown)


dstocco:
Hi @aphecetche ,
the error is not on O2 but QualityControl. I've renamed a file which is used in QualityControl. Actually, I took care of removing the conflicting part from the QualityControl master branch before doing the O2 PR:
https://github.com/AliceO2Group/QualityControl/pull/817
But it seems that the CI does not compile the QualityControl master, but rather a tag...
I can assure you that this PR does not break the QualityControl master, but of course it will fail on old tags.

aphecetche:
@dstocco I guess the idea then would be to ask for a QC tag, right @Barthelemy  ?

dstocco:
Well, I see that the v1.28 was produced yesterday. So I guess it's just a matter of moving to the new tag in the CI...

shahor02:
@dstocco you can also put back the old file, and then remove it when the QC will have correct version.

dstocco:
@shahor02 : good point, I've just done it

aphecetche (approved):

aphecetche:

PR #6932, opened 2021-08-30
rename TF to DF

jgrosseo:


jgrosseo:
@ktf Do you expect any regressions somewhere from this?

ktf:
No, should be fine. I do not use those metrics for anything else.

PR #6936, opened 2021-08-31
TPC: add --file-for cru option

wiechula:


davidrohr (changes requested):
Thx, I was actually about to do the same when I saw this.
Unfortunately this one is not correct.
The fileFor = cru is different to fileFor = link. In the link case we want one raw file per endpoint, in the cru case there should be one file per CRU.
Also, the nameing in the CRU case should contain the FLP that will have that link.
I'll open a superseding PR soon.

wiechula:
Ok, sorry, then I misunderstood. In Rubens example also the endpoint was in the file name:
`ITS_<FLP_NAM>_cru<CRU_ID>_<EndPoint>.raw`
Then I let you implement it.

davidrohr:
Oh, good point, you are right, I didn't see that. Just thought that it would be per CRU. But I also added the FLP names.
OK, I'll update mine and keep the CRU name AND the endpoint.
Thx for catching!

PR #6939, opened 2021-08-31
[MRRTF-144] [O2-2486] mch-digits-to-raw: add per cru endpoint output

aphecetche:
and replaces `--file-per-link` option with `--file-for (all,cru,link)`
one.

### DIFF
--- a/Detectors/MUON/MCH/Raw/Encoder/Payload/PayloadPaginator.cxx
+++ b/Detectors/MUON/MCH/Raw/Encoder/Payload/PayloadPaginator.cxx
@@ -71,15 +71,39 @@ o2::header::RDHAny rdhFromLinkInfo(LinkInfo li)
   return rdh;
 }
 
+std::string flpName(LinkInfo li)
+{
+  static std::array<int, 33> cru2flp = {
+    148, 148, 148, //  0, 1, 2 // FIXME: to be x-checked for St12
+    149, 149, 149, //  3, 4, 5 // FIXME: to be x-checked for St12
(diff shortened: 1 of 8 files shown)


aphecetche:
@davidrohr @shahor02 here's the implementation of the per-cru-endpoint feature (and the removal of the --file-per-link in favour of --file-for) for `o2-mch-digits-to-raw`

Note that the actual cru->flp mapping is still to be x-checked (in particular for St12) by @aferrero2707  but should be fine for St345.

Detectors/MUON/MCH/Raw/Encoder/Payload/PayloadPaginator.cxx
```diff
@@ -71,15 +71,39 @@ o2::header::RDHAny rdhFromLinkInfo(LinkInfo li)
...
                    std::string outputBase,
                    const std::set<LinkInfo>& links,
-                   bool filePerLink)
+                   bool filePerLink,
+                   bool filePerCru)
 {
   std::string output = fmt::format("{:s}.raw", outputBase);
   for (auto li : links) {
     if (filePerLink) {
       output = fmt::format("{:s}_feeid{:d}.raw", outputBase, li.feeId);
```
shahor02: Just in case, in case of per-link output, could you simply extend the per-cru name by the link id, so that the flp and cru/ep are also present in the filename?

aphecetche: done

aphecetche:
@shahor02 I guess this one is ready to be merged

shahor02 (approved):

PR #6941, opened 2021-08-31
FIT LUT: common lookup table implemented for FV0 and FDD

afurs:


afurs:
@mslupeck , @shahor02  could you approve this PR? There are serious bug in FDD LUT(local channelID starts from 0), which I would like to fix(common FIT LUT fix it) before FLP suite new version release

shahor02 (approved):
ok, merging

PR #6942, opened 2021-09-01
Optional fall-back dir. for CTFwriter + free space check on primary one

shahor02:
Since in on the EPNs we may store the CTFs on the RAM disk of limited capacity, one can
indicate the fall-back storage via `--output-dir-alt <dir>` option. The writer will switch to it if
(i) `szCheck = max(min-file-size*1.1, max-file-size)` is positive and
(ii) estimated (accounting for eventual other CTFs files written concurrently) available space on
the primary storage is below the `szCheck`. The  available space is estimated as 
````
current physically available space
-
number still open CTF files from concurrent writers * szCheck
+
the current size of these files
````
The fall-back storage is not checked for the available space.

### DIFF
--- a/Detectors/CTF/workflow/src/CTFWriterSpec.cxx
+++ b/Detectors/CTF/workflow/src/CTFWriterSpec.cxx
@@ -47,6 +47,10 @@
 #include <TTree.h>
 #include <filesystem>
 #include <ctime>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <regex>
 
 using namespace o2::framework;
 
@@ -82,7 +86,7 @@ class CTFWriterSpec : public o2::framework::Task
 {
  public:
   CTFWriterSpec() = delete;
-  CTFWriterSpec(DetID::mask_t dm, uint64_t r = 0, bool doCTF = true, bool doDict = false, bool dictPerDet = false, size_t smn = 0, size_t szmx = 0);
+  CTFWriterSpec(DetID::mask_t dm, uint64_t r = 0, bool doCTF = true, bool doDict = false, bool dictPerDet = false);
   ~CTFWriterSpec() override = default;
   void init(o2::framework::InitContext& ic) final;
   void run(o2::framework::ProcessingContext& pc) final;
@@ -101,6 +105,9 @@ class CTFWriterSpec : public o2::framework::Task
   void closeTFTreeAndFile();
   void prepareTFTreeAndFile(const o2::header::DataHeader* dh);
   size_t estimateCTFSize(ProcessingContext& pc);
+  size_t getAvailableDiskSpace(const std::string& path, int level);
+  void createLockFile(const o2::header::DataHeader* dh, int level);
+  void removeLockFile();
 
   DetID::mask_t mDets; // detectors
   bool mWriteCTF = false;
@@ -111,6 +118,7 @@ class CTFWriterSpec : public o2::framework::Task
   uint64_t mRun = 0;
   size_t mMinSize = 0;     // if > 0, accumulate CTFs in the same tree until the total size exceeds this minimum
   size_t mMaxSize = 0;     // if > MinSize, and accumulated size will exceed this value, stop accumulation (even if mMinSize is not reached)
+  size_t mChkSize = 0;     // if > 0 and fallback storage provided, reserve this size per CTF file in production on primary storage
   size_t mAccCTFSize = 0;  // so far accumulated size (if any)
   size_t mCurrCTFSize = 0; // size of currently processed CTF
   size_t mNCTF = 0;        // total number of CTFs written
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Detectors/CTF/workflow/src/CTFWriterSpec.cxx [hunk 1]
```diff
@@ -141,6 +153,58 @@ class CTFWriterSpec : public o2::framework::Task
...
+      LOG(INFO) << "Multiple CTFs will be accumulated in the tree/file until its size exceeds " << mMinSize << " bytes";
+      if (mMaxSize > mMinSize) {
+        LOG(INFO) << "but does not exceed " << mMaxSize << " bytes";
+      }
+    }
+  }
+  mChkSize = std::max(size_t(mMinSize * 1.1), mMaxSize);
+  if (!std::filesystem::exists(mLockFileDir)) {
+    if (!std::filesystem::create_directories(mLockFileDir)) {
+      throw std::runtime_error(fmt::format("Failed to create {} directory", mLockFileDir));
```
davidrohr: this is a possible race condition. If 2 processes try to create the folder at the same time.
If the `create_directories` fails, I would propose to do another `exists` check before throwing an error.

Detectors/CTF/workflow/src/CTFWriterSpec.cxx
```diff
@@ -486,7 +525,63 @@ void CTFWriterSpec::closeDictionaryTreeAndFile(CTFHeader& header)
...
+{
+  // count number of CTF files in processing (written to storage at given level) from their lock files
+  std::regex pat{fmt::format("({}/ctfs{}-[[:alnum:]_]+\\.lock$)", mLockFileDir, level)};
+  int nLocked = 0;
+  size_t written = 0;
+  for (const auto& entry : std::filesystem::directory_iterator(mLockFileDir)) {
+    const auto& entryName = entry.path().native();
+    if (std::regex_search(entryName, pat) && (mLockFD < 0 || entryName != mLockFileName)) {
+      int fdt = open(entryName.c_str(), O_RDONLY);
+      if (lockf(fdt, F_TEST, 0) != 0) {
```
davidrohr: if you can lock a lockfile that you do not own, isn't that orphaned by definition? I.e. shouldn't you remove it? (The remove could trigger a race condition with the original owner removing it after unlocking, but that should not be a problem, we can just ignore the return value of the `remove`.

Detectors/CTF/workflow/src/CTFWriterSpec.cxx
```diff
@@ -486,7 +525,63 @@ void CTFWriterSpec::closeDictionaryTreeAndFile(CTFHeader& header)
...
+        if (nrd == sizeof(size_t)) {
+          written += sz;
+        }
+      }
+      close(fdt);
+    }
+  }
+  std::error_code ec;
+  const auto si = std::filesystem::space(path, ec);
+  int64_t avail = int64_t(si.available) - nLocked * mChkSize + written; // account already written part of unfinished files
```
davidrohr, line 606: can't this produce some double-counting of the file size while a file is written, i.e. if the file is written but the lock is not yet removed, the file size is obviously subtracted in `space(path, ec)` already, and then you subtract again the estimated size. But OK, I also don't see a better algorithm. Just wondering if a simple counting of the number of files would have been sufficient.

Detectors/CTF/workflow/src/CTFWriterSpec.cxx
```diff
@@ -486,7 +525,63 @@ void CTFWriterSpec::closeDictionaryTreeAndFile(CTFHeader& header)
...
+    mLockFD = -1;
+    std::filesystem::remove(mLockFileName);
+  }
+}
+
+//___________________________________________________________________
+size_t CTFWriterSpec::getAvailableDiskSpace(const std::string& path, int level)
+{
+  // count number of CTF files in processing (written to storage at given level) from their lock files
+  std::regex pat{fmt::format("({}/ctfs{}-[[:alnum:]_]+\\.lock$)", mLockFileDir, level)};
```
davidrohr: This will only work if the mLockFileDir of all processes is identical? But OK, probably we can assume that.

Detectors/CTF/workflow/src/CTFWriterSpec.cxx
shahor02: right, added another check after 10\ms wait

Detectors/CTF/workflow/src/CTFWriterSpec.cxx
shahor02: Yes, usually these would be remnants from crashed jobs. Their cleanup could create also some problems for checks in concurrent jobs. Anyway, added a cleanup of these files (provided they were not updated within last minute) with some extra protections.

Detectors/CTF/workflow/src/CTFWriterSpec.cxx
shahor02: In this marginal case indeed there would be a double counting and the writer may switch to real disk. Just counting number of files would not be reliable: one would need to count (1) finished but not yet evacuated files, (2) files currently in writing (with locks)  + wathever other files we may decide to have on the RAM disk or garbage (e.g. ctf*.part files from crashed jobs). And this will become even more complicated if there are directories on the RAM disk.

Detectors/CTF/workflow/src/CTFWriterSpec.cxx
shahor02: It is actually constant, `/tmp/ctf-writer-locks`, was indeed not following coding conventions.

davidrohr: ok, thx, one more thing here: I would propose we prepent all files/folders in /tmp with `o2_` to have some kind of namespacing (at least I have been doing this so far).

PR #6943, opened 2021-09-01
TRD tracking improvements

martenole:
Hi @shahor02 
the last commit [671a22e](https://github.com/AliceO2Group/AliceO2/commit/671a22e4746b0798d6f938ea2e6c39efb8f6b8e3) adds the inward and outward refits to the TPC-TRD tracks. For the moment I added the method to the DPL device, eventually it should go into `GPUTRDTracker`. For that I just need to check how to retrieve the relevant information (TPC cluster access/transformation) from inside that class.
But maybe the general functionality can already be checked? Or should I rather try to move the implementation directly into the TRD tracker?
Cheers,
Ole


Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -344,6 +365,102 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
...
+    std::array<float, 3> trkltCovUp;
+    mRecoParam.recalcTrkltCov(tilt, trk.getSnp(), pad->getRowSize(mTrackletsRaw[trkltId].getPadRow()), trkltCovUp);
+
+    chi2In += trk.getPredictedChi2(trkltPosUp, trkltCovUp);
+    if (!trk.update(trkltPosUp, trkltCovUp)) {
+      LOGF(INFO, "Failed to update track with space point in layer %i", iLy);
+      return false;
+    }
+  }
+  retVal = mTPCRefitter->RefitTrackAsTrackParCov(trk, mTPCTracksArray[recoCont->getSingleDetectorRefs(trk.getRefGlobalTrackId())[GTrackID::TPC]].getClusterRef(), timeTRD * mTPCTBinMUSInv, &chi2In, false, false); // inward refit
```
shahor02: Can you also propagate it to `o2::constants::geom::XTPCInnerRef` (w/o considering failure to propagate as refit failure).

shahor02 (commented):
Hi @martenole 
Thanks! Sorry, I did not understand what is the difference between the spec and the tracker  in extracting the TPC clusters access? In principle, it would be more logical to have the refit in the tracker itself.
Please see one comment below.

shahor02:
Could you remove the Draft tag to trigger CI?

davidrohr (dismissed):
I think it is OK in the DPL device for now.
Eventually I would move it into the GPU Refit class instead of the GPUTRDTracker class

martenole:
Hi @shahor02,

> Sorry, I did not understand what is the difference between the spec and the tracker in extracting the TPC clusters access?

the only difference is that I need to access it through the GPU IO pointers instead of through the RecoContainer. It should not be a problem to change it. But since I was not sure how to do it and wanted to have a first version ready fast I did it in the device for now.

Should I also store the chi2 for the inward/outward refits? The TRD track has one member for its chi2 which is currently the chi2 for the matching of the seed to the tracklets.

I will add also the refit for ITS-TPC-TRD and then mark the PR ready for review.

Cheers,
Ole

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
martenole: Yes, sorry I forgot about that

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -190,8 +202,27 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
...
+    auto mITSClusterROFRec = inputTracks.getITSClustersROFRecords();
+    const auto clusITS = inputTracks.getITSClusters();
+    if (mITSClusterROFRec.empty() || clusITS.empty()) {
+      LOG(FATAL) << "No ITS clusters";
+      return;
+    }
+    const auto patterns = inputTracks.getITSClustersPatterns();
+    auto pattIt = patterns.begin();
+    mITSClustersArray.reserve(clusITS.size());
+    //o2::its::ioutils::convertCompactClusters(clusITS, pattIt, mITSClustersArray, *mITSDict); // segfaults.. mITSDict is not set, is this the issue?
```
martenole: Hi @shahor02 do know what I am doing wrong here trying to load the ITS clusters?

shahor02: ITSDict must be a valid object, even if empty (i.e. created by default constructor). See how it is done here: https://github.com/AliceO2Group/AliceO2/blob/5897718f3ba61f306520709be3815d9e59f94080/Detectors/GlobalTrackingWorkflow/src/TPCITSMatchingSpec.cxx#L101-L107
(eventually it should come from the CCDB, the TPCITSMatchingSpec will update it in the `updateTimeDependentParams` so the MatchTPCITS::mITSDict pointer to TPCITSMatchingSpec::mITSDict will automatically provide the latest dict. In your case, since you keep the fitter in the spec, you can simply convert mITSDict from pointer to object, and read it like in the TPCITSMatchingSpec

davidrohr:
> the only difference is that I need to access it through the GPU IO pointers instead of through the RecoContainer. It should not be a problem to change it. But since I was not sure how to do it and wanted to have a first version ready fast I did it in the device for now.

HI @martenole : I think accessing the data through the GPUIOPtr is actually the way to go, since that will allow us to move it to the GPU more easily later on.

shahor02 (approved):

PR #6948, opened 2021-09-01
MCH: add time clustering of MCH digits

aferrero2707:
The time clustering algorithm uses a brute-force peak search algorithm to find the time regions where the digits clusterize around a given point in time.

### DIFF
--- a/Detectors/MUON/MCH/PreClustering/CMakeLists.txt
+++ b/Detectors/MUON/MCH/PreClustering/CMakeLists.txt
@@ -12,4 +12,4 @@
 o2_add_library(MCHPreClustering
         SOURCES src/PreClusterFinder.cxx
         src/PreClusterFinderMapping.cxx
-        PUBLIC_LINK_LIBRARIES O2::MCHMappingImpl3 O2::MCHBase O2::Framework)
+        PUBLIC_LINK_LIBRARIES O2::MCHMappingImpl4 O2::MCHBase O2::Framework)
--- a/Detectors/MUON/MCH/TimeClustering/CMakeLists.txt
+++ b/Detectors/MUON/MCH/TimeClustering/CMakeLists.txt
@@ -9,18 +9,16 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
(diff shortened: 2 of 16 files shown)


aferrero2707:
Here are the things to modify in this PR:
- remove the use of the first TF orbit in the time clustering
- minimize the changes in the PreClustering workflow
- simplify the optional input data specification in the time clustering workflow

aphecetche:
> * minimize the changes in the PreClustering workflow

actually note that the current changes break the `o2-mch-reco-workflow`, as the meaning of the `getPreClusterFinderSpec` parameters is changed by this PR (I guess that should have been caught by the fullCI anyway, if it went up to the actual FST test, but is currently failing for other reason).

aferrero2707:
@aphecetche the issue with the `reco-workflow` is even worse... the code compiles because one of the two string parameters of the `getPreClusterFinderSpec` function has a default value, hence in the `reco-workflow` we assign the data processor name to what is supposed to be the input description.

In fact, the input description should be provided from the command line in the `reco-workflow` case as well, unless the ROOT input is used, in which case the input description should be fixed to `DIGITROFS` (or whatever is defined in `getDigitSamplerSpec()`).

This looks quite over-complicated and error-prone for the moment... any idea?

aphecetche:
yep, the reco-workflow fails (because you've added a parameter in front of the existing one in getPreclusterFinderSpec :-( ). I've changed that in my (wip) dev branch to be able to pursue my tests : https://github.com/aphecetche/AliceO2/tree/mrrtf-129-rof-time-clustering. 
IMO the reco-workflow should be as simple as possible by default (so its number of command line option minimised if possible). 
Anyway, once I have my tests working I'll think of a clean(er) solution (if there's one)

aferrero2707:
@ktf @aphecetche the MacOS CI job is failing due to build errors on `O2HitMerger.h` (un-related to this PR).

Could you please have a look?

Thanks!

aphecetche (approved):

aphecetche:
@ktf @shahor02 as far as we can tell the CI errors are unrelated. Can this PR be merged (or the CI restarted if that's preferred ?). Thanks.

PR #6953, opened 2021-09-02
MCH: add first/last orbit option to rof-histogrammer

aphecetche:
in order to fix the boundaries of the output histogram, which is handy
when comparing simulations made with different time spread resolution
for instance (by default the firstIR and lastIR used in the histogram
will depend on the spread)


aphecetche:
@shahor02 @sawenzel can this one be merged please ? thanks.

shahor02 (approved):

PR #6955, opened 2021-09-02
DPL: speed up bulk reading

ktf:
Allocating and deallocating the buffer was actually a significant (10%)
fraction of the reading. This should fix it until we actually write
directly to the target message.


ktf:
Hold. Somehow this prevents stuff from terminating.

ktf:
Issue understood and unrelated. Better buffer size added. Merging.

PR #6957, opened 2021-09-03
macro to upload CTP orbit reset time object

shahor02:
@lietava this is the format for the CTP reset object: just plain `array<long,1>` so there is no need of O2 to create it on your side.

### DIFF
--- a/macro/CMakeLists.txt
+++ b/macro/CMakeLists.txt
@@ -55,6 +55,7 @@ install(FILES CheckDigits_mft.C
               run_trac_alice3.C
               CreateBCPattern.C
               UploadDummyAlignment.C
+       CreateCTPOrbitResetObject.C
         DESTINATION share/macro/)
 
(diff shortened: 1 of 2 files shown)


lietava:
@shahor02 thanks. So if I understand it well it is enough to run this macro as soon as possible after orbit reset at CTP. 
Or better tu run it with 
t = std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::system_clock::now()).time_since_epoch().count();
obtained at CTP reset.
Correct ?

shahor02:
@lietava well, this was just to show the format of the object and to create some entries for test. 
Yes, you should use the time from `now()` obtained at reset. Whether you use this macro or write it yourself from the python script, is up to you. Note that you should set the validity tmin equal to the time of reset and tmax = tmin + 0xffffffff (i.e. max lifetime of the orbit). The default values in the macro are just placeholder.

PR #6958, opened 2021-09-03
DPL GUI: speedup with many metrics

ktf:


ktf:
Compiles on macOS and fullCI. Not tested otherwise. Merging.

PR #6962, opened 2021-09-03
DPL Analysis: protect writer against empty messages

ktf:
Whether it is correct for the message to be empty still to be checked.


ktf:
Currently broken. Merging.

PR #6967, opened 2021-09-06
DataDistribution TF reader workflow + related

shahor02:
## DataDistribution TF reader workflow

`o2-raw-tf-reader-workflow` allows to inject into the DPL the `(s)TF` data produced by the DataDistribution, avoiding running DD sTFReader in a separate terminal and sending data via proxies.
The relevant command line options are:
```
--input-data arg
```
input data (obligatory): comma-separated list of input data files and/or files with list of data files and/or directories containing files

```
--onlyDet arg (=all)
```
list of dectors to select from available in the data. AT THE MOMENT NOT FUNCTIONAL

```
--max-tf arg (=-1)
```
max TF ID to process (<= 0 : infinite)

```
--loop arg (=0)
```
loop N times (-1 = infinite) over input files (but max-tf has priority if positive)

```
--delay arg (=0)
```
delay in seconds between consecutive TFs sending (provided TFs can be read fast enough)

```
--copy-cmd arg (=XrdSecPROTOCOL=sss,unix xrdcp -N root://eosaliceo2.cern.ch/?src ?dst)
```
copy command for remote files

```
--tf-file-regex arg (=.+\.tf$)
```
regex string to identify TF files: optional to filter data files (if the input contains directories, it will be used to avoid picking non-TF files)

```
--remote-regex arg (=^/eos/aliceo2/.+)
```
regex string to identify remote files (they will be copied to `/tmp/<random>/` directory, then removed after processing)

```
--max-cached-tf arg (=3)
```
max TFs to cache in memory: builds TFs asynchrously to sending, may speed up processing but needs more shmem.

```
--max-cached-files arg (=3)
```
max TF files queued (copied for remote source). For local files almost irrelevant, for remote ones asynchronously creates local copy.

```
--tf-reader-verbosity arg (=0)
```
verbosity level (1 or 2: check RDH, print DH/DPH for 1st or all slices, >2 print RDH)

```
--raw-channel-config arg
```
optional raw FMQ channel for non-DPL output, similar to `o2-raw-file-reader-workflow`.


shahor02:
MacOS errors unrelated (timeout for all macro tests), merging

PR #6924, opened 2021-08-30
Update AOD handling of ambiguous MFT and Fwd tracks 

rpezzi:
* Adds TrackTime and TrackTimeResolution for forward track table
* Ambiguous MFT and forward tracks assigned to BC Slice
* Add comments to Fwd tracks tables element definitions

The PR uses same column definitions for ambiguous as #6970 (and will result in merge conflicts) with it.

PR #6970, opened 2021-09-06
collision slice for ambiguous track table

jgrosseo:


jgrosseo:
@rpezzi I changed here the collision array to a slice as well (as suggested by @shahor02) and tried to use the same columns for the ambiguous track table as well as the mft track table. It seems to compile for me.

rpezzi:
Thanks @jgrosseo . Doesn't slices assume that compatible BCs or Collisions are ordered?  I may be missing something, but I believe this will not be the case, unless a range reference-style array of collisions is created for each ambiguous track table.

rpezzi:
image

An unique ordering of collisions won't address all track -types.

jgrosseo:
@shahor02 may comment more, but the assignment of both BCs and Collisions is done only according to the time of the track in the reconstruction (DCA is not used). And both, BCs and Collisions are stored time-ordered in the AOD. So therefore it can be a slice.

jgrosseo:
Nice plot. Do you plan this different for the muon reconstruction than for central barrel tracks?

I let @shahor02 comment...

shahor02:
Those tracks which are associated with vertex by the DCA (i.e. they are contributors, by construction they are also matching in time) are never ambiguous. The vertices are sorted in time.

jgrosseo:
> Those tracks which are associated with vertex by the DCA (i.e. they are contributors, by construction they are also matching in time) are never ambiguous. The vertices are sorted in time.

And if a secondary has two compatible vertices?

shahor02:
@jgrosseo there is no cut on DCA for attachment of "secondaries" to vertices, any track which is not contributing to vertex is a potential secondary and matched only in time (at least for now).

rpezzi:
> Those tracks which are associated with vertex by the DCA (i.e. they are contributors, by construction they are also matching in time) are never ambiguous. The vertices are sorted in time.

So, if tracking/position information solves any ambiguity, in general a slice is not enough. At forward, an array of collisions is needed to address ambiguity of tracks with MFT info. For MCH and MID+MCH tracks the slice should suffice.

jgrosseo:
So the MFT track extrapolation to the beam spot has sufficient resolution to select only a subset of the collisions which are within the beam spot?

shahor02:
To me it is opposite: the Z resolution of MFT tracks is much worse than that of the ITS....

rpezzi:
> So the MFT track extrapolation to the beam spot has sufficient resolution to select only a subset of the collisions which are within the beam spot?

Yes, the MFT should to be able to select the collision, at least on a portion of the data. As @shahor02 , the Z resolution of the MFT is not good, but we expect to separate if collisions are not too close in Z.

jgrosseo:
I am not against having a Collision array. That is what the current data model has and if now or in the future we will use the z position of ITS / MFT tracks to select the collisions, then this is the better choice. However, I think it is strange if we will do this for the MFT and consider it feasible there and we say at the same time that it will not be done for the ITS. What do you think?

shahor02:
The question is simple: suppose you have a track which in Z is 2 cm away from 1 vertex and 5 mm away from another one (both incompatible with any reasonable cut on nSigma DCA to assume it is a primary). Do we want to say that it is a secondary from 2nd vertex or consider that it may be secondary from any of them (V0 binding will consider both of them and if both give V0 converging to these vertices, both V0 candidates will be stored, one per vertex).

rpezzi:
> ... However, I think it is strange if we will do this for the MFT and consider it feasible there and we say at the same time that it will not be done for the ITS. What do you think?

If the MFT can resolve, surely the ITS can as well, and with better precision. 

> The question is simple: suppose you have a track which in Z is 2 cm away from 1 vertex and 5 mm away from another one (both incompatible with any reasonable cut on nSigma DCA to assume it is a primary). Do we want to say that it is a secondary from 2nd vertex or consider that it may be secondary from any of them (V0 binding will consider both of them and if both give V0 converging to these vertices, both V0 candidates will be stored, one per vertex).

In addition of these two compatible collisions 2 cm and 5 mm away from the track, there could be others incompatible further away that does not need to be re-evaluated at analysis-level. But honestly, I need to educate myself here. I learn as we go... I  see now that V0s for central barrel tracks are stored in AOD, while there is no equivalent at forward. Perhaps we should have them for forward and/or MFT tracks. 

What I realize is that the advent of v0 tables for forward and/or MFT tracks may reduce the criticality of the ambiguous track-tables. And maybe, just maybe, storing the slice of collisions instead of an array of collisions can be afforded. We will discuss this internally and get back here (and PR #6924). Thanks for the input.

Ping @aphecetche @pillot @urasantonio.

shahor02:
@rpezzi but how do define compatibility for the secondary w/o particular decay topology analysis? Morever, note that for the MFT an offset of ~1mm in transverse DCA vrt some vertex would mean cm-like offset in Z.

jgrosseo:
Just to add: The v0 tables only contain track references to existing tracks. They do not contain track information themselves.

rpezzi:
> @rpezzi but how do define compatibility for the secondary w/o particular decay topology analysis? Morever, note that for the MFT an offset of ~1mm in transverse DCA vrt some vertex would mean cm-like offset in Z.

This is exactly what we have to discuss internally. 

> Just to add: The v0 tables only contain track references to existing tracks. They do not contain track information themselves.

In addition to TrackIDs, v0 have a [collisionsId](https://github.com/AliceO2Group/AliceO2/blob/fc1d7cb14236d058890ab8d15ce87a1896faccd7/Framework/Core/include/Framework/AnalysisDataModel.h#L685) used in TransientV0s (indicated as "In-memory V0 table"). This is calculated on the fly once a particular decay topology is set? If this is to be calculated at analysis level, indeed  the MFT tracking information at AOD can be used to resolve ambiguities. The point of adding an array of collisionIDs in AOD is to reduce the computational overhead of re-testing all time-compatible collisions in every analysis.

I feel this became a long discussion for Github. Better move this conversation to conversation elsewhere. I will get back to you after tomorrow's MFT meeting.

shahor02:
@rpezzi ok to move this discussion elsewhere, just to answer your question:

> This is calculated on the fly once a particular decay topology is set? If this is to be calculated at analysis level, indeed the MFT tracking information at AOD can be used to resolve ambiguities. The point of adding an array of collisionIDs in AOD is to reduce the computational overhead of re-testing all time-compatible collisions in every analysis.

V0s produced in the SVertexer by construction are associated with 1 vertex to which they (or the cascade they belong to) converge. If the same V0 is compatible with 2 vertices (which would mean they are very close in Z) they will be created for each vertex separately.

jgrosseo:
I think this needs more discussion indeed. My conclusion is below (and I will change the PR accordingly):
* At present there is no code in central barrel tracking which makes use of the DCA information. It is assigned by track time.
* Similarly, such code does not exist for MFT
* Therefore, a bc slice has the same information content than a collision slice
* The selection of such tracks goes anyway through the BC at present: collision time --> compatible BCs --> selection on ambiguous table

My conclusion is that 
* For now we just add the bc slice to the ambiguous tables. 
* Once code exists which exploits the DCA information we add the needed collision table. There is no advantage adding it now.

jgrosseo:
@TimoWilken macOS is 21 hours behind?

TimoWilken:
> @TimoWilken macOS is 21 hours behind?

@jgrosseo The OSX O2 builder was slowed down recently. I changed the way it's run this morning and the number of parallel compilation jobs was increased today, so the builder should catch up soon. Sorry for the delay.

jgrosseo:
MacOS CI unrelated.

PR #6971, opened 2021-09-06
Use MCTruthContainer in the CA tracker

mpuccio:
Dear @shahor02,
with @mconcas we rebased #6431 onto the dev branch. Just to be clear (for GitHub at least), this closes #6431.
Cheers,
Matteo and Max

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/src/IOUtils.cxx
+++ b/Detectors/ITSMFT/ITS/tracking/src/IOUtils.cxx
@@ -78,56 +78,6 @@ void ioutils::convertCompactClusters(gsl::span<const itsmft::CompClusterExt> clu
   }
 }
 
-// std::vector<ROframe> ioutils::loadEventData(const std::string& fileName)
-// {
-//   std::vector<ROframe> events{};
-//   std::ifstream inputStream{};
-//   std::string line{}, unusedVariable{};
-//   int layerId{}, monteCarlo{};
-//   int clusterId{EventLabelsSeparator};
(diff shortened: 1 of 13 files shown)


mconcas (dismissed):

mconcas (dismissed):

Detectors/ITSMFT/ITS/tracking/src/IOUtils.cxx
```diff
@@ -78,55 +78,55 @@ void ioutils::convertCompactClusters(gsl::span<const itsmft::CompClusterExt> clu
   }
 }
 
-std::vector<ROframe> ioutils::loadEventData(const std::string& fileName)
```
shahor02, line 81: can this be removed?

shahor02 (dismissed):
Thanks, just a small comment below.

Detectors/ITSMFT/ITS/tracking/src/IOUtils.cxx
mpuccio: Good point, removed.

mpuccio:
uhm... osx is taking longer than expected (36h) and now also dataflow is spinning again... why? @TimoWilken

TimoWilken:
> uhm... osx is taking longer than expected (36h)

The OSX O2 builder was slowed down recently, I've changed the way it's run and the number of parallel compilation jobs was increased today, so the builder should catch up soon.

> and now also dataflow is spinning again... why? @TimoWilken

Earlier today, there was an issue with a few builders stepping on each other. Now fixed.

mpuccio:
osx failures look to me unrelated

PR #6977, opened 2021-09-06
MID workflow to test dead and noisy channels

dstocco:
The series of commits allow to generate a workflow that can be used to analyse the calibration triggers in order to compute the noisy and dead channels and produce the corresponding masks.
The commits are meant to be atomic: plese do not squash.


dstocco:
This seems to be a real error related to my code. I did not see it since I cannot reproduce it on MacOSX...but it seems I can reproduce it on a CentOS7 machine. I will have a deeper look.

dstocco:
A stupid mistake on the test indeed. Should be fixed now

dstocco:
The full CI test fails because it says that basically all of the classes have a missing or malformed copyright notice. I've actually copy pasted it from other classes. Is it a real error (i.e. the copyright notice changed) or a fake one?

shahor02:
@dstocco the error is real, you are using very obsolete license, it should be
```
// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
// All rights not expressly granted are reserved.
//
// This software is distributed under the terms of the GNU General Public
// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
//
// In applying this license CERN does not waive the privileges and immunities
// granted to it by virtue of its status as an Intergovernmental Organization
// or submit itself to any jurisdiction.
```

dstocco:
Hi @shahor02 : ok, thanks for the info. I've changed the licence.

dstocco:
@aphecetche : the latest errors seem to be all unrelated to the PR (although for two tests they complain about missing clang-format, so not sure up to where the tests went...)

Detectors/MUON/MID/Workflow/src/MaskHandlerSpec.cxx
```diff
@@ -0,0 +1,129 @@
...
+        LOG(INFO) << "Problematic digits found in " << name << " events. Corresponding masks:";
+        for (auto& mask : maskVec) {
+          LOG(INFO) << mask;
+        }
+      }
+      name = "FET";
+    }
+  }
+
+  // std::cout << "\nCorresponding boards masks:" << std::endl;
```
aphecetche: I guess this commented out part can go, right ?

aphecetche (dismissed):

aphecetche:
@dstocco indeed. Unfortunately I'm not amongst those who can bypass failed checks to merge PRs. So either you wait for the two failing tests to run again (and hopefully succeed) or you'd have to ping some power user (e.g. @shahor02) who can force merge.

Detectors/MUON/MID/Workflow/src/MaskHandlerSpec.cxx
dstocco: Well, this code allowed indeed to show the mask as a Local Board mask. This makes it easier to apply the mask at DCS level. I'm therefore uncommenting it ;)
But ok, this is secondary since, in any case, in the end we will have to put the mask in the CCDB...

dstocco:
The Mac error is unrelated. Anyone with force-merge power? ;)

PR #6978, opened 2021-09-06
DPL: propagate the DataHeader::runNumber information

ktf:


ktf:
@shahor02 apparently no one propagated the information about DataHeader::runNumber to DPL when it was added. This should fix it (not tested as I need to recompile the whole stack right now).

shahor02:
@ktf thanks! will test it

shahor02 (approved):
Now works fine, thanks!

ktf:
Tested locally on mac. Full CI works.

PR #6979, opened 2021-09-06
EMCAL DCS DP processing

mpoghos:


shahor02 (dismissed):
conditionally approving to trigger CI

Detectors/EMCAL/calib/src/EMCDCSProcessor.cxx
```diff
@@ -0,0 +1,344 @@
...
+  //		// tbd
+  //	}
+  //else
+
+  if (alias.find("EMC_DDL_LIST0") != std::string::npos) {
+    mFEECFG->setDDLlist0(val);
+	}
+  if (alias.find("EMC_DDL_LIST1") != std::string::npos) {
+    mFEECFG->setDDLlist1(val);
+	}
```
mfasDa: Misformated. Also lines below. Please run clang-format -I on that file.

mfasDa (changes requested):
Several files are not properly formatted - please run `clang-format -i ...` on the file. Also please change the commit message header to 

```
[EMCAL-556] EMCAL DCA DP processing
```

and specify in the message body which datapoints are implemented - I want to see the commit message in this format before approval.

The code has no documentation, even very few comments, which is not helpful but for the moment not a blocker from approval.

Otherwise the code looks good I would say.

mfasDa (dismissed):
Looks good now!

mfasDa (dismissed):

shahor02 (approved):
conditionally approving to restart CI

mfasDa:
@shahor02 Is the approval mandatory to start the CI? In this case, is the approval from me as codeowner sufficient?

shahor02:
Hi @mfasDa 
Approval is needed for the 1st 1st-timer commit. I guess the codeowner approval should work, you can try if you find another 1st-timer

mpoghos:
Hi @shahor02 and @mfasDa 
Do you understand why the build/AliceO2/O2/o2/macOS test filed? I think nothing related to my commit. is this a blocker for merging?

shahor02:
yes, it is unrelated

mfasDa:
@shahor02 Thanks for the reply! I already did the approval at the very first stage so it seems the code owner approval is not sufficient.

PR #6980, opened 2021-09-06
Fixes for ITSMFT cluster pattern and dictionary

shahor02:


shahor02:
macos seems to be dead, merging almost trivial fix.

PR #6981, opened 2021-09-06
Fix in recording CTF meta.nLiteralWords

shahor02:
@MichaelLettrich storing literals info in the matadata was still wrong: https://github.com/AliceO2Group/AliceO2/blob/81c6cf29e66e03a737b8840e04189a685d761706/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h#L506 was firing.
Fortunately, the data is decodable: overestimated `meta.nLiteralWords` was leading just to more than needed buffer allocation.


shahor02:
merging as current code produces wrong CTFs (still readable when compiled with NDEBUG)

PR #6982, opened 2021-09-07
MUON: dcs-ccdb: add option for dpConfName 

aphecetche:
(and fix default timestamp used for query)


aphecetche:
will be included in a forthcoming PR

PR #6983, opened 2021-09-07
Adapted inputCard.xml to account for the new directory structure

pbuehler:
considers table producers in O2Physics/Common
ignores table producers in O2Physics/PWG* for the moment


pbuehler:
@jgrosseo should work with the new directory structure BUT without PWG* - tables!

jgrosseo (approved):

PR #5826, opened 2021-03-30
Use JAliEn credentials in CCDB API

Atlantic777:
This PR adds JAliEn credentials manager to CCDB API and decorates curl handle with JAliEn credentials. Depends on: https://github.com/alisw/alidist/pull/2932

cc @costing @iraklic @yuw726

PR #6984, opened 2021-09-07
CCDBApi: protection against missing X509_CERT_DIR env var

aphecetche:
@Atlantic777 @costing @sawenzel 

Here's a small change for your consideration.

Assuming I understand correctly that the CCDB is not 100% tied to AliEn, I would expect it to work as it was before https://github.com/AliceO2Group/AliceO2/pull/5826 even if some env variables are not set. 

As far as I can tell with this minimal change the ccdb related tests of O2 (`ctest -L ccdb`) work even without the env var set, otherwise they crash (and any attempt to use the CcdbApi crash as well).

(tested on macOS, with AliceO2 built with Spack, but I'd say that's not relevant as much as added requirement on some variables to be present if they are not really, or not always).

### DIFF
--- a/CCDB/src/CcdbApi.cxx
+++ b/CCDB/src/CcdbApi.cxx
@@ -354,8 +354,10 @@ void CcdbApi::curlSetSSLOptions(CURL* curl_handle)
 
   TJAlienCredentialsObject cmo = mJAlienCredentials->get(cmk);
 
-  string CAPath = getenv("X509_CERT_DIR");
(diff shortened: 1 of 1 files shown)


aphecetche:
hum, actually @ktf proposed almost at the same time another version of this PR https://github.com/AliceO2Group/AliceO2/pull/6985 which does _not_ allow the X509_CERT_DIR to be unset. So, should it really be set or (as I've observed) can we live without it ?

ktf (approved):

sawenzel (approved):

costing (approved):

PR #6987, opened 2021-09-07
TPC: apply drift velocity calibration from ccdb, remove CE signal for laser track calibration

wiechula:


wiechula:
@shahor02 I think the errors are unrelated. Can this be merged?

PR #6988, opened 2021-09-07
o2-sim: hit merge optimizations

sawenzel:
The commit is speeding up the sim merging step by
not using TMemFiles as intermediate buffers,
in favour of dedicated plain merger pools keeping the
data in memory until streamed out.

In result, the processing time of the hit-merger
kernel is decreased by a factor 5.


shahor02:
Hi @sawenzel 

This PR makes the o2sim_HitsTRD.root to contain empty tree only, which in turn makes the FST to fail. 
I've checked that reverting it brings back TRD hits.
Shall we revert it until the problem is understood?

PR #6989, opened 2021-09-07
DPL: fix handling of empty tables

ktf:
* Actually send empty tables, so that there is no need for
  tricks on the receiving side.
* Actually write empty tables to disk.


ktf:
@knopers8 this actually highlights a quite severe issue with getByPos, which would return the wrong ref.spec if pipelining were involved. Could you cross check you do not depend on the bad behaviour?

jgrosseo:
This fixes the problem for me locally! :)

knopers8:
@ktf QC should be fine. I see this is used in EMCal code, but in their case there are no pipelines, so i suppose the behaviour should be preserved.

ktf:
I could not find any particular usage either. The only code affected by it would be the cleanup of the timers, which probably means right now there are chances that the timers are not cleaned up correctly. Anyways, merging. Let me know if you see something suspicious.

PR #6992, opened 2021-09-07
[Test] PWGHF: configure KFParticle package and add analysis task for 3-prongs reconstruction with KF

zhujh126:
The link of PR in O2: https://github.com/zhujh126/AliceO2/commit/4714a60d7a2296a8b7ca83089ece077d0998cd34
The link of PR in O2Physics: https://github.com/zhujh126/O2Physics/commit/08206ad894e9e58454f793e471d9d316596aba5e

### DIFF
--- a/dependencies/FindKFParticle.cmake
+++ b/dependencies/FindKFParticle.cmake
@@ -1,0 +1,32 @@
+# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+# See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+# All rights not expressly granted are reserved.
+#
+# This software is distributed under the terms of the GNU General Public
+# License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+#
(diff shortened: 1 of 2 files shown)


jgrosseo:
Why is this not added as a dependency in O2Physics? Will it also be needed in O2?

pzhristov:
KFParticle is required package in O2 and as such is build when one compiles O2Physics. For the moment we do not use any functionality in O2, so probably the dependency can be moved to alisw/o2physics.sh in a related PR.

zhujh126:
Thanks a lot for your comments. I tried to implement a similar way as other packages (like HepMC3). However, it does not work. Could you please tell me how to make KFParticle works? Now it is complied when O2 is installed, but it is not configured properly.

zhujh126:
I 

> KFParticle is required package in O2 and as such is build when one compiles O2Physics. For the moment we do not use any functionality in O2, so probably the dependency can be moved to alisw/o2physics.sh in a related PR.

I managed to add it in O2Physics. So we can close this PR.

PR #6993, opened 2021-09-07
Use proper method to create detector noise map filename 

shahor02:
Also, hoping to shed some light on [[O2-2332]](https://alice.its.cern.ch/jira/browse/O2-2329), added a few catch/(re)throw to `STFDecoder::init`


rpezzi (approved):

shahor02:
CI failures are unrelated, FST is currently broken since TRD hits generation fails, merging

PR #6998, opened 2021-09-08
allow digit writer in multi-thread

noferini:


noferini:
I think failing checks are not related to this PR

shahor02:
Hi @noferini 
Could you please clarify the purpose of this class? Why do you need to parallelize the writer? Note that the `pipeline` creates not the threads by completely independent processes, receiving TF with  TFid%NLanes = LaneID.
You mention that it should be used for the slewing calibration, but the input is digit. Where it is supposed to run?

noferini:
Hi Ruben,
I added this feature not for calibration and it is not supposed to be used in the normal workflow.

This writer feature is supposed to be used in commissioning phase in case we need to collect digits for some quick checks on epn for TOF standalone runs (debug mode).

It is an additional feature which doesn't affect the normal usage.

The normal tof workflow writes ctf (supported in multi-node epns) not digits (not supported in multi-node epns).

The only way to write temporary digits on epn is to use an aggregator/writer which needs multi-thread to deal with the high load. 

As I said I would use it only for debugging.

shahor02:
OK, but what is the reason to pipeline the writer? Is the writing bottleneck? Note that if you pipeline decoding, it does not mean that you need to pipeline also the writing.

noferini:
In the aggregator the writing is the unique running process. It is the bottleneck.

toplogy I tested:

10 epns (decoding/digitization) -> 1 aggregator (only digit writer)

I was also surprised that I need 8 writers to manage the full TOF load but this is what I found in test I did few days ago.

shahor02:
hm, if the writer is bottleneck, then it should be mostly a disk access problem + some contribution from the root serialization. The former will be almost not affected by the pipe-lining. If the real bottleneck is digits serialization to root tree, then it might be worth creating TOF-only CTFs which are more compact and easy to serialize.

Anyway, the PR itself is ok, merging.

shahor02 (approved):

noferini:
Hi Ruben,
thanks.
CTF creation is a working option but for quick checks is less efficient when I need to inspect few minutes of data (O(10k) TFs w.r.t O(10) digit outputs).

PR #6999, opened 2021-09-08
o2-sim: hitmerger scalability improvements (part 2)

sawenzel:
This should solve/reduce the slow-down problem observed
when running many events.


sawenzel:
@mpuccio : fyi

PR #7001, opened 2021-09-08
DPL GUI: update to new API

ktf:
The new API is present in v0.6.0 and provides support for creating remote GUIs. We can merge this separately as it is ultimately standalone.


ktf:
macOS tested on my own box. fullCI issue unrelated.

PR #7003, opened 2021-09-08
Fix printout

ktf:


ktf:
Currently broken on mac, works for build/O2/o2. Merging to get it into the nightly.

PR #7005, opened 2021-09-09
Fix index-shift error

sawenzel:


sawenzel:
I checked the fullCI test locally and it passes again. Will merge as a hotfix without waiting.

PR #7006, opened 2021-09-09
Fix reading of empty frames and cluster indices O2-2571

mpuccio:


mpuccio:
@shahor02, this will fix also the problem with the cluster indices

mpuccio:
I think that the CI is as green as it can get...

shahor02:
@mpuccio yes, macOS is hopeless. But did you solve the problem TGeo corrections?

PR #7007, opened 2021-09-09
Update DCS DP DeliveryType class by Float

shahor02:


shahor02:
trivial change, already tested locally. Merging as it is blocking DCS processing tests.

PR #7009, opened 2021-09-09
Noise calibration using clusters with pattern IDs

iouribelikov:


shahor02 (approved):
Hi @iouribelikov 
Thanks! Failure is unrelated, merging. I've realized that currently we cannot apply noise masks to data read from the CTF. I will need to add to ITS/MFT entropy decoder  options
1) to decompose decoded clusters and send digits
2) to decompose to digits, apply masks and reclusterize again before sending clusters out.

PR #7010, opened 2021-09-09
[MRRTF-142] MCH: add time resolution to simulated digits and noise on all ROFs

aphecetche:
The noise is added to potentially all ROFs (but within the limited range of first and
last hadronic interaction in the eventRecords) not only those with collisions.


aphecetche:
@TimoWilken this one failed on macOS for an unrelated error (in O2HitMergerRunner.cxx., which, I believe, has been fixed since then) : I assume it will  be reran at some point ?

aphecetche:
moving this PR to draft while I x-check a few things...

PR #7011, opened 2021-09-09
TPC: handle exceptions in calib_processing_helper::processRawData

shahor02:
RawParser used by processRawData throws on corrupted TPC raw data, eventually killing workflows
on all EPNs. This PR catches those exceptions, logs an error and skips the FMQ part, i.e.
1 HBF for single link with current packaging.
If an optional pointer on the counter is provided, it will be incremented for each throw.


wiechula (approved):

shahor02:
macOS still broken by O2HitMerger printf formatting, merging

PR #7020, opened 2021-09-10
Allow to have MID raw output grouped by cru

dstocco:
This should be compliant with https://alice.its.cern.ch/jira/browse/O2-2486

### DIFF
--- a/Detectors/MUON/MID/Workflow/src/RawWriterSpec.cxx
+++ b/Detectors/MUON/MID/Workflow/src/RawWriterSpec.cxx
@@ -44,7 +44,7 @@ class RawWriterDeviceDPL
   void init(o2::framework::InitContext& ic)
   {
     auto dirname = ic.options().get<std::string>("mid-raw-outdir");
-    auto fileFor = ic.options().get<std::string>("mid-file-for");
+    auto fileFor = ic.options().get<std::string>("file-for");
     if (!std::filesystem::exists(dirname)) {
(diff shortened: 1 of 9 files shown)


Detectors/MUON/MID/Workflow/src/RawWriterSpec.cxx
```diff
@@ -43,9 +43,8 @@ class RawWriterDeviceDPL
  public:
   void init(o2::framework::InitContext& ic)
   {
-    auto filename = ic.options().get<std::string>("mid-raw-outfile");
     auto dirname = ic.options().get<std::string>("mid-raw-outdir");
-    auto perLink = ic.options().get<bool>("mid-raw-perlink");
+    auto fileFor = ic.options().get<std::string>("mid-file-for");
```
aphecetche: isn't the option named `file-for` simply and not `mid-file-for` ?

dstocco: Indeed, I changed it to be consistent with other detectors but I forgot to replace it here. Actually at the beginning I added the prefix mid because the options with the same name where not treated correctly when piping different workflows. Not sure if this was solved

aphecetche (approved):

shahor02 (approved):
thanks! will merge once @TimoWilken repairs macOS CI

shahor02:
macOS is dead, merging.

PR #7021, opened 2021-09-10
DPL Analysis: allow std::vector column types backed by variable size list

aalkin:
* Update TableBuilder to handle writing vectors
* Update ColumnIterator to handle extracting vectors
* Add a test case to test_ASoA

@ktf this works, but I have doubts about using std::vector here (although very convenient)

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -29,6 +29,7 @@
 #include <cassert>
 #include <fmt/format.h>
 #include <typeinfo>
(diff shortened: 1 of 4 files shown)


aalkin:
C++20 `std::span` would be much better here...

aalkin:
Now list column getter will return a `gsl::span`, so no extra allocations will happen.

ktf:
As discussed we should probably use the filler with the iterator pairs, so that one does not need a std::vector on filling.

PR #7022, opened 2021-09-10
Bug fixes for MID decoder

dstocco:
The MID decoder was not correctly accounting for the possible delay between the BC and the MID local clock.
This delay is currently set to 0, but it will need to be tuned on real data.
Also, the event type was not correctly assigned if the calibration trigger occurred in coincidence with an orbit trigger.

### DIFF
--- a/Detectors/MUON/MID/Raw/src/ELinkDataShaper.cxx
+++ b/Detectors/MUON/MID/Raw/src/ELinkDataShaper.cxx
@@ -122,6 +122,7 @@ bool ELinkDataShaper::processTrigger(const ELinkDecoder& decoder, EventType& eve
   // From here we treat triggered events
   bool goOn = false;
   correctedClock = localClock;
+  eventType = EventType::Standard;
   if (decoder.getTriggerWord() & raw::sCALIBRATE) {
     // This is an answer to a calibration trigger
     eventType = processCalibrationTrigger(localClock);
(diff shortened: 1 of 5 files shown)


Detectors/MUON/MID/Raw/src/ELinkDataShaper.cxx
```diff
@@ -106,22 +129,24 @@ void ELinkDataShaper::processOrbitTrigger(uint16_t localClock, uint8_t triggerWo
...
 }
 
-bool ELinkDataShaper::processTrigger(const ELinkDecoder& decoder, EventType& eventType, uint16_t& correctedClock)
+bool ELinkDataShaper::processTrigger(const ELinkDecoder& decoder, EventType& eventType, InteractionRecord& ir)
 {
   /// Processes the trigger information
   /// Returns true if the event should be further processed,
   /// returns false otherwise.
   auto localClock = decoder.getCounter();
+  ir.bc = mIR.bc + localClock;
```
aphecetche: can't this addition results in a bc above `o2::constants::lhc::LHCMaxBunches` ?

dstocco: Not really because the mIR.bc is always zero and the local clock is reset at each orbit trigger. But it is true that, in the end, I do not use mIR.bc at all. So, if you prefer, I could just put: `ir.bc = localClock`

dstocco: @aphecetche : done and added a FIXME as a reminder.

aphecetche (approved):

dstocco:
Any chance to repaire the Mac CI? If not, can this be merged?

ktf:
Checked by hand on macOS. Merging.

PR #7023, opened 2021-09-10
Default MID FEEId config reflects current CERN config

dstocco:
The default FEEId config was an ideal one, but did not reflected the configuration at CERN.
With this commit the default can be used at is to correctly decode data

### DIFF
--- a/Detectors/MUON/MID/Raw/include/MIDRaw/FEEIdConfig.h
+++ b/Detectors/MUON/MID/Raw/include/MIDRaw/FEEIdConfig.h
@@ -58,6 +58,8 @@ class FEEIdConfig
 
  private:
   bool load(const char* filename);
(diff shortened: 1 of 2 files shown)


aphecetche:
@dstocco seems there’s a genuine failure of the raw test mid

dstocco:
Hi @aphecetche , indeed I've just realised that the test will fail until this PR: https://github.com/AliceO2Group/AliceO2/pull/7020 will be merged. After that it will work.

dstocco:
@aphecetche : as expected the error went away after the other PR was approved. The Mac CI is still dead, but I guess that one can safely review the PR now...

aphecetche (approved):

ktf:
Checked by hand on mac. Merging.

PR #7024, opened 2021-09-10
CTFReader can read remote files via FileFetcher

shahor02:
Provides the same local/remote file input/filtering interface as for the `o2-raw-tf-reader`.


shahor02:
@TimoWilken the macOS CI server is stuck and the only result it occasionally produces is a timeout for all macros test. Could you check please?

PR #7028, opened 2021-09-13
SpaceChage: Adding function to set the space-charge density from a CalDet object

matthias-kleiner:
- Adding conversion from CalDet (pad granularity) to TH3 (r*phi granularity)
- Catching division by 0
- updating function for calculating the electron drift path + creation of a gif animation of the drift path
- updating default value for omega tau
- Updating createSCHistosFromHits macro to allow for higher granularity histograms


wiechula (approved):

ktf:
Checked by hand on macOS. Merging.

PR #7031, opened 2021-09-13
TRD Tracklet Transformer CCDB interface

jbarrella:
This commit includes a change that makes the `TrackletTransformer` fetch the `CalVdriftExB` object from CCDB and use these values for calibration.

One question: line 50, I set the timestamp of the CCDB manager. I assume this value needs to correspond with whatever the latest timestamp is. If I do not set the timestamp, then it does not find anything. I am not sure how to get that information to keep the timestamp up to date automatically?

### DIFF
--- a/Detectors/TRD/base/src/TrackletTransformer.cxx
+++ b/Detectors/TRD/base/src/TrackletTransformer.cxx
@@ -30,19 +30,33 @@ TrackletTransformer::TrackletTransformer()
   // 5mm below cathode plane to reduce error propogation from tracklet fit and driftV
   mXDrift = mGeo->cdrHght() - 0.5;
   mXtb0 = -100;
-
-  // dummy values for testing. This will change in the future when values are pulled from CCDB
-  mt0Correction = -0.279;
-  mLorentzAngle = 0.14;
-  mDriftVRatio = 1.1;
 }
 
-void TrackletTransformer::loadPadPlane(int hcid)
+void TrackletTransformer::loadPadPlane(int detector)
 {
-  int detector = hcid / 2;
   int stack = mGeo->getStack(detector);
   int layer = mGeo->getLayer(detector);
   mPadPlane = mGeo->getPadPlane(layer, stack);
+}
+
+void TrackletTransformer::loadCalibrationParameters(int timestamp)
+{
+  LOG(info) << "loading calibration parameters with timestamp: " << timestamp;
+
+  auto& ccdbmgr = o2::ccdb::BasicCCDBManager::instance();
+  ccdbmgr.setTimestamp(timestamp);
+
+  mCalibration = ccdbmgr.get<o2::trd::CalVdriftExB>("TRD/Calib/CalVdriftExB");
+
+  // placeholder for when t0 is available
+  mT0 = ccdbmgr.get<std::vector<float>>("TRD/Calib/ChamberT0");
+
+  if (mCalibration == nullptr) {
+    LOG(error) << " failed to get vDrift and ExB parameters from ccdb";
+  }
+  if (mT0 == nullptr) {
+    LOG(warn) << " failed to get mT0 parameters from ccdb";
+  }
 }
 
 float TrackletTransformer::calculateY(int hcid, int column, int position)
@@ -75,12 +89,12 @@ float TrackletTransformer::calculateZ(int padrow)
   return rowPos - rowSize / 2. - middleRowPos;
 }
 
-float TrackletTransformer::calculateDy(int slope, double lorentzAngle, double driftVRatio)
+float TrackletTransformer::calculateDy(int detector, int slope)
 {
   double padWidth = mPadPlane->getWidthIPad();
 
-  // temporary dummy value in cm/microsecond
-  float vDrift = 1.5464f;
+  float vDrift = mCalibration->getVdrift(detector);
+  float exb = mCalibration->getExB(detector);
 
   int slopeSigned = 0;
   if (slope & (1 << (NBITSTRKLSLOPE - 1))) {
@@ -96,22 +110,23 @@ float TrackletTransformer::calculateDy(int slope, double lorentzAngle, double dr
   // NOTE: check what drift height is used in calibration code to ensure consistency
   // NOTE: check sign convention of Lorentz angle
   // NOTE: confirm the direction in which vDrift is measured/determined. Is it in x or in direction of drift?
-  double lorentzCorrection = TMath::Tan(lorentzAngle) * mXAnode;
+  double lorentzCorrection = TMath::Tan(exb) * mXAnode;
 
   // assuming angle in Bailhache, fig. 4.17 would be positive in our calibration code
   double calibratedDy = rawDy - lorentzCorrection;
 
-  return rawDy;
+  return calibratedDy;
 }
 
-float TrackletTransformer::calibrateX(double x, double t0Correction)
+float TrackletTransformer::calibrateX(double x)
 {
+  // Hard-coded value will be replaced once t0 calibration is available from CCDB
+  float t0Correction = -0.279;
   return x += t0Correction;
 }
 
-std::array<float, 3> TrackletTransformer::transformL2T(int hcid, std::array<double, 3> point)
+std::array<float, 3> TrackletTransformer::transformL2T(int detector, std::array<double, 3> point)
 {
-  int detector = hcid / 2;
   auto transformationMatrix = mGeo->getMatrixT2L(detector);
 
   ROOT::Math::Impl::Transform3D<double>::Point localPoint(point[0], point[1], point[2]);
@@ -122,6 +137,7 @@ std::array<float, 3> TrackletTransformer::transformL2T(int hcid, std::array<doub
 
 CalibratedTracklet TrackletTransformer::transformTracklet(Tracklet64 tracklet)
 {
+  uint64_t detector = tracklet.getDetector();
   uint64_t hcid = tracklet.getHCID();
   uint64_t padrow = tracklet.getPadRow();
   uint64_t column = tracklet.getColumn();
@@ -129,16 +145,18 @@ CalibratedTracklet TrackletTransformer::transformTracklet(Tracklet64 tracklet)
   uint64_t slope = tracklet.getSlope();
 
   // calculate raw local chamber space point
-  loadPadPlane(hcid);
+  loadPadPlane(detector);
   float x = getXDrift();
   float y = calculateY(hcid, column, position);
   float z = calculateZ(padrow);
 
-  float dy = calculateDy(slope, mLorentzAngle, mDriftVRatio);
-  float calibratedX = calibrateX(x, mt0Correction);
+  float dy = calculateDy(detector, slope);
+
+  float calibratedX = calibrateX(x);
+
   // NOTE: Correction to y position based on x calibration NOT YET implemented. Need t0.
 
-  std::array<float, 3> sectorSpacePoint = transformL2T(hcid, std::array<double, 3>{calibratedX, y, z});
+  std::array<float, 3> sectorSpacePoint = transformL2T(detector, std::array<double, 3>{calibratedX, y, z});
 
   LOG(debug) << "x: " << sectorSpacePoint[0] << " | "
              << "y: " << sectorSpacePoint[1] << " | "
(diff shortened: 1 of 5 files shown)


martenole:
Hi Jason,
I think just for testing the transformation using the calibration parameters form the CCDB this is ok. Although at the moment you query the CCDB only once when you create a TrackletTransformer object and afterwards they will never change.
Soon the framework will be able to provide the CCDB objects as a DPL message similar to the other inputs and you will be able to subscribe to the message. Then the framework will take care of providing the correct object for the current time (which will be deduced from the orbit counter relative to the start of run which in turn is translated into an absolute time which is assigned to the objects in the CCDB).
So, in the near future we will have to adapt to the new scheme, but for first tests now it should be ok. Can you add the timestamp as an option to the tracklet transformer workflow? Because then we can run it once using our fixed run number as timestamp to do the calibration and create a CCDB object. And then we could repeat the reconstruction using the timestamp of the newly created CCDB object to see how the performance changes.
Cheers,
Ole

jbarrella:
@martenole Thanks. I added the timestamp as an option now. Let me know if there is anything else.

bazinski:
@jbarrella what happened to "isHappy" ? :-1:

jbarrella:
> 
> 
> @jbarrella what happened to "isHappy" ? 👎

I realized that I could just check against the `mCalibration` variable itself. Unfortunately :(

jbarrella:
There is another bug that I need to fix. When using the default timestamp for `o2-sim` data, the calibration classes used in the CCDB are completely different. Will have to think about how to handle that without making the code too messy...

bazinski:
@jbarrella maybe just convert the data once off and move away from the old classes?

martenole:
You mean the CalVDriftExB object? I see that we assign the timestamp 265503 to that so I guess you need to use that one (see http://ccdb-test.cern.ch:8080/browse/TRD_test/CalVdriftExB)

martenole:
Actually it is not so good that we mix a timestamp with what we call run number.. We need to change that at some point. Per default the timestamp for a calibration workflow would be the time at which the processing was with a validity of one year.

bazinski:
Well i would suggest to move all to the new classes, maybe in a seperate pr though? I suspect the trapconfig will change as well once I get to parsing the version 0x47 events.

jbarrella:
> 
> 
> You mean the CalVDriftExB object? I see that we assign the timestamp 265503 to that so I guess you need to use that one (see http://ccdb-test.cern.ch:8080/browse/TRD_test/CalVdriftExB)

@martenole Yeah I understand that simulations are anchored to this 297595 run number/timestamp. And the objects are of the class `o2::trd::ChamberCalibrations`. What exactly is at timestamp 265503 in `TRD_test/CalVdriftExB`? Is it a conversion of the old data in `TRD_test/ChamberCalibrations`?

martenole:
These number are completely arbitrary.. I don't remember why I put 265503 for CalVdriftExB. I will change this in a future PR to the current time: `std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count()`. But for now in order to access it correctly you should just use 265503, because that is the only run aka timestamp for which we have a valid CCDB object.
For the simulations we use the OCDB from run 2 and copied the values from run  297595 into the CCDB, that is why we sticked to that number at the moment in run 3 simulations

jbarrella:
> 
> 
> These number are completely arbitrary.. I don't remember why I put 265503 for CalVdriftExB. I will change this in a future PR to the current time: `std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count()`. But for now in order to access it correctly you should just use 265503, because that is the only run aka timestamp for which we have a valid CCDB object.
> For the simulations we use the OCDB from run 2 and copied the values from run 297595 into the CCDB, that is why we sticked to that number at the moment in run 3 simulations

But if I just change the default `timestamp` to 265503, then we will not be able to calibrate any simulated data. Maybe we should do some sort of copy and conversion of the old 297595 one. I could figure out how to do that.

martenole:
> But if I just change the default `timestamp` to 265503, then we will not be able to calibrate any simulated data. Maybe we should do some sort of copy and conversion of the old 297595 one. I could figure out how to do that.

Why not? The tracklet transformer does not do any calibration it just applies whatever calibration objects it finds in the CCDB right?
Edit: maybe you can keep the default values for vDrift and so on and make it optional to take them from the CCDB?

jbarrella:
> > But if I just change the default `timestamp` to 265503, then we will not be able to calibrate any simulated data. Maybe we should do some sort of copy and conversion of the old 297595 one. I could figure out how to do that.
> 
> Why not? The tracklet transformer does not do any calibration it just applies whatever calibration objects it finds in the CCDB right?
> Edit: maybe you can keep the default values for vDrift and so on and make it optional to take them from the CCDB?

Yeah, but the simulations aren’t anchored to 265503 right? So it would apply an incorrect calibration.

martenole:
The 265503 is arbitrary, every object will get this number, see: https://github.com/AliceO2Group/AliceO2/blob/d8a32b7f8d338297f4d3358e4b17f2cc4ab3f833/Detectors/TRD/calibration/src/CalibratorVdExB.cxx#L166
We don't use the CCDB in the way it is intended yet. I am working on it and this will change, but for now you can just put 265503 and that will pull the 'correct' calibration object

martenole:
Hi Jason,
in #7101 I changed the Vd and ExB calibrator such that the CCDB object gets an actual timestamp with the validity of one year. I already did the calibration so there is an object in the CCDB: http://ccdb-test.cern.ch:8080/browse/TRD/Calib/CalVdriftExB
So you don't have to provide the run number anymore. I would just leave the default parameters in for now so that we can switch between either take the most recent CCDB object or the default values for each chamber.
Cheers,
Ole

jbarrella:
Okay just a small bug fix with regards to the CCDB path and then changed the default timestamp value to use `555555` which is a special object @martenole uploaded and contains default calibration parameters.

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
```diff
@@ -35,6 +35,13 @@ void TRDTrackletTransformerSpec::run(o2::framework::ProcessingContext& pc)
 {
   LOG(INFO) << "Running tracklet transformer";
 
+  TrackletTransformer transformer(mTimestamp);
```
martenole: Why do you create now a `TrackletTransformer` object each time run is called? I would keep it as a member of TRDTrackletTransformerSpec.

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
```diff
@@ -35,6 +35,13 @@ void TRDTrackletTransformerSpec::run(o2::framework::ProcessingContext& pc)
...
+
+  if (!transformer.hasCalibration()) {
+    // ccdb object was not found for specified timestamp
```
martenole, line 39: Would be good to throw an error message here

Detectors/TRD/base/src/TrackletTransformer.cxx
```diff
@@ -96,22 +112,23 @@ float TrackletTransformer::calculateDy(int slope, double lorentzAngle, double dr
...
 
   return rawDy;
 }
 
-float TrackletTransformer::calibrateX(double x, double t0Correction)
+float TrackletTransformer::calibrateX(double x)
 {
-  return x += t0Correction;
+  // Hard-coded value will be replaced once t0 calibration is available from CCDB
+  float t0 = -0.279;
```
martenole: Maybe call this t0Correction to not confuse it with the t0 of 4 time bins below?

Detectors/TRD/base/src/TrackletTransformer.cxx
```diff
@@ -96,22 +112,23 @@ float TrackletTransformer::calculateDy(int slope, double lorentzAngle, double dr
   // NOTE: check what drift height is used in calibration code to ensure consistency
   // NOTE: check sign convention of Lorentz angle
   // NOTE: confirm the direction in which vDrift is measured/determined. Is it in x or in direction of drift?
-  double lorentzCorrection = TMath::Tan(lorentzAngle) * mXAnode;
+  double lorentzCorrection = TMath::Tan(exb) * mXAnode;
 
   // assuming angle in Bailhache, fig. 4.17 would be positive in our calibration code
   double calibratedDy = rawDy - lorentzCorrection;
 
   return rawDy;
```
martenole: You could even return `calibratedDy` here. In the default calib object I put ExB to zero, so the correction does not have an an effect anyways.

martenole:
Hi @jbarrella 
looks good to me, just had some minor comments. Did you try to run the reconstruction with your changes to see if the fraction of TRD tracklets which is attached to global tracks remains the same? The instructions are in the README here: https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/TRD/workflow/README.md
The number of attached tracklets is written in the logs.
Cheers,
Ole

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
jbarrella: I wasn't sure how to actually implement that since the constructor takes an argument. I made an attempt. Let me know if it looks alright.

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
jbarrella: The error gets thrown at line 57 in `TrackletTransformer::loadCalibrationParameters()`.

jbarrella:
> 
> 
> Hi @jbarrella looks good to me, just had some minor comments. Did you try to run the reconstruction with your changes to see if the fraction of TRD tracklets which is attached to global tracks remains the same? The instructions are in the README here: https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/TRD/workflow/README.md The number of attached tracklets is written in the logs. Cheers, Ole

`[5386:trd-globaltracking[TPC_ITS-TPC]]: [11:36:25][INFO] The TRD tracker found 7 tracks from TPC seeds and 4 tracks from ITS-TPC seeds and attached in total 28 tracklets out of 346`

These appear to be the numbers that I am getting after my most recent commit. Is this in line with the norm?

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
martenole: Since your constructor argument has a default value this does not matter

jbarrella: Well `mTimestamp` is not within the scope of the TrackletTransformer. I think it is working how I implemented it though

martenole: I don't get why you need to know the time stamp at the time you create an `TrackletTransformer` object. I would not put the function which connects to the CCDB inside the constructor. Instead you can just call that function from the DPL device and then you would also get an error message if that fails in the place one would expect it ;)

jbarrella: True, but there are many cases where the TrackletTransformer is used outside of the workflow context. With this change, one would need to manually call the function to load the CCDB object.

martenole: In which cases for example? In any case one would need to load the CCDB object then, sure. But I would see this as a good thing, because that way you make sure that the user is aware that a calibration object is needed and a timestamp should be provided. A default time stamp would hide that

jbarrella: Okay, I went ahead and made this change.

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
```diff
@@ -35,6 +35,12 @@ void TRDTrackletTransformerSpec::run(o2::framework::ProcessingContext& pc)
 {
   LOG(INFO) << "Running tracklet transformer";
 
+  mTransformer.loadCalibrationParameters(mTimestamp);
```
martenole: Just one more point: Could you please move this line to TRDTrackletTransformerSpec::init()? At the moment the timestamp anyway never changes and otherwise a new object is fetched from the CCDB for every time frame which is a quite large overhead.

martenole (approved):

jbarrella:
@davidrohr Hello. Can this be merged?

martenole (changes requested):
Actually, probably this should not yet be merged, because on the EPNs we don't have access to the test CCDB and then this will fail. The default CCDB object needs to be uploaded to the production CCDB first

davidrohr:
What is the status for the upload of the default CCDB object? Is it there? If so, this could be merged now?

martenole:
@davidrohr  CCDB upload is done, this could be merged

Issue #347, opened 2017-04-26
config directory not installed to VMCWORKDIR


wiechula:
The 'config' directory containing the rootmanager.dat file should be installed to 
VMCWORKDIR 
otherwise FairRoot will not be able to use it.

sawenzel:
Issue fixed. The config folder is found under $VMCWORKDIR.

PR #7034, opened 2021-09-13
Cleaning up use of DataHeader

matthiasrichter:
- always use DataRefUtils to extract DataHeader properties
- cleanup usage of DataHeader and Stack within DPL processors
- removing obsolete headers

One thing to come is to use the orbit information from DPL TimeInfo and
replace code using the firstTFOrbit member of one of the DataHeaders.


ktf:
@sawenzel simulation seems to be timing out on slc8 in a number of PRs. E.g.:

```
346/392 Test #347: o2sim_G4 .............................................................................***Timeout 400.05 sec
```

matthiasrichter:
@Barthelemy @knopers8 can you please check the log for the `o2-dataflow-cs8` CI, one QC unit test is failing with timeout, but my conclusion from the full log is that it is not related to this PR but the FairMQ termination. All devices get the shutdown request but not all make the transition to IDLE.

matthiasrichter:
> @sawenzel simulation seems to be timing out on slc8 in a number of PRs. E.g.:
> 
> ```
> 346/392 Test #347: o2sim_G4 .............................................................................***Timeout 400.05 sec
> ```

Maybe the shm creation did not complete, here are the last messages from the log
316/392 Test #347: o2sim_G4 .............................................................................***Timeout 400.06 sec
[INFO] This is o2-sim version 1.2.0 (526585190)
[INFO] Built by --unavailable-- on OS:Linux-4.4.0-210-generic
[INFO] BINDING TO ADDRESS ipc:///tmp/o2sim-notifications-13249 type pub
[INFO] Running with 2 sim workers 
[INFO] CREATING SIM SHARED MEM SEGMENT FOR 2 WORKERS

Framework/Core/test/test_CCDBFetcher.cxx
```diff
@@ -31,11 +31,11 @@ WorkflowSpec defineDataProcessing(ConfigContext const&)
       AlgorithmSpec{
```
matthiasrichter, line 31: @ktf, looks like this test is not compiled anymore, should we remove it? There is almost identical code in  `Framework/TestWorkflows/src/test_CCDBFetcher.cxx`

Utilities/DataSampling/src/Dispatcher.cxx
```diff
@@ -87,7 +88,7 @@ void Dispatcher::run(ProcessingContext& ctx)
     if (firstPart.header == nullptr) {
       continue;
     }
-    const auto* firstInputHeader = header::get<header::DataHeader*>(firstPart.header);
+    const auto* firstInputHeader = DataRefUtils::getHeader<header::DataHeader*>(firstPart);
```
matthiasrichter, line 91: @knopers8 , please confirm that the changes are without effect to the functionality

Framework/Core/test/test_CCDBFetcher.cxx
ktf: yes

ktf:
Tested by hand on macOS. Merging.

PR #7037, opened 2021-09-14
DPL: use CTP information if available

ktf:


ktf:
Actual fetching of the CTP object still needs to happen. Now checking how difficult it is to do everything asynchronously, otherwise I will simply fall back to a blocking request.

PR #7039, opened 2021-09-15
DPL: get CTP information to retrieve the orbitResetTime

ktf:


ktf:
@chiarazampolli @shahor02 as privately discussed this seems to work in my test. 
The DataProcessingHeader::startTime must be populated correctly roughly the time when the data was taken, in nanoseconds and a valid CTP object must be available for that time. 

If those conditions are met, the value of the resetOrbit is then retrieved from CCDB and used to calculate the actual time precise time from the firstTFOrbit which is then used to query InputSpec with Lifetime::Condition.

If this PR passes tests I would suggest we merge it and then we finish off the remaining bits (most notably the per object overrides and fix the issue @wiechula reported with --ccdb-backend).

As discussed separately with @shahor02 and @ironMann, to playback RAW data we probably need to save somewhere in the CTF the rough time when it was taken, so that we can populate the DataProcessingHeader accordingly.

ktf:
Works in full CI and dataflow. Checked on my mac by hand. Merging.

PR #7043, opened 2021-09-15
Fix ITS chip mapping: swap module high/low cables

shahor02:
@davidrohr @mpuccio this will make already emulated ITS raw data tracking inefficient (will be ok if regenerated from existing digits): I had to swap some pairs of cables to match the decoder/encoder mapping to real detector.

### DIFF
--- a/Detectors/ITSMFT/common/reconstruction/include/ITSMFTReconstruction/RUInfo.h
+++ b/Detectors/ITSMFT/common/reconstruction/include/ITSMFTReconstruction/RUInfo.h
@@ -17,6 +17,7 @@
 
 #include <Rtypes.h>
 #include <cstdint>
+#include <string>
 
 namespace o2
 {
@@ -48,6 +49,7 @@ struct ChipOnRUInfo {
   std::uint8_t cableHWPos = DUMMY8;     // position of the bit of this cable in the activeLanes
   std::uint8_t chipOnCable = DUMMY8;    // chip within the cable (0 = master)
 
+  std::string asString() const;
   void print() const;
 };
 
@@ -61,6 +63,7 @@ struct ChipInfo {
   std::uint16_t ru = DUMMY16;     // RU sequential id
   std::uint16_t ruType = DUMMY16; // RU (or subBarrel) type
 
+  std::string asString() const;
   void print() const;
 };
 } // namespace itsmft
--- a/Detectors/ITSMFT/common/reconstruction/src/ChipMappingITS.cxx
+++ b/Detectors/ITSMFT/common/reconstruction/src/ChipMappingITS.cxx
@@ -39,32 +39,27 @@ constexpr std::array<uint8_t, ChipMappingITS::NSubB> ChipMappingITS::GBTHeaderFl
 constexpr std::uint8_t ChipMappingITS::ChipOBModSW2HW[14];
 constexpr std::uint8_t ChipMappingITS::ChipOBModHW2SW[15];
 
-//______________________________________________
 ChipMappingITS::ChipMappingITS()
 {
+  // FEE ID field in RDH is 16-bit wide
+  // Each RU has 10-bit DIPSWITCH, use bits 9:2 as an 8bit ID field to be mapped to FEE ID field in RDH
+
+  // | Lr |Stave Count| Bin. pref | Range of Binary Addresses
+  // |    |= RU Count | Stave addr|
+  // |----------------------------------------------------------
+  // | L0 |   12      |b 0000xxxx | 0000_0000 : 0000_1011 (0 – 11)
+  // | L1 |   16      |b 0001xxxx | 0001_0000 : 0001_1111 (0 – 15)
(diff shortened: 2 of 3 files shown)


iouribelikov:
@shahor02 Hi Ruben.  Does this PR mean that the clusterisation of real raw data needs to be re-done ?

shahor02:
> @shahor02 Hi Ruben. Does this PR mean that the clusterisation of real raw data needs to be re-done ?

Hi @iouribelikov yes, if the clusterization was done starting from raw TFs, it should be redone to get correct chips positions. 
Or, one can apply a hack by swapping the chipIDs of the module upper / lower 7 chips.
Since the data in the CTFs is already clusterized, there, the only way to get correct positions would be this trick with swapping chipIDs.

iouribelikov:
> > @shahor02 Hi Ruben. Does this PR mean that the clusterisation of real raw data needs to be re-done ?
> 
> Hi @iouribelikov yes, if the clusterization was done starting from raw TFs, it should be redone to get correct chips positions.
> Or, one can apply a hack by swapping the chipIDs of the module upper / lower 7 chips.
> Since the data in the CTFs is already clusterized, there, the only way to get correct positions would be this trick with swapping chipIDs.

@shahor02  The funny thing is that, even with the incorrect chip positions, I do not see anything wrong in the cosmic-track residuals...  So I wonder how it will look like once these changes are merged.
May be, we should wait a bit, until I get new residuals.  With my current capacity, this would take a few days of running.

shahor02:
What are the current residuals? Can't reprocess with this PR just a couple of TFs which where you see the tracks and compare results? The problem with delaying merging this PR is that the CTFs currently written contain a wrong mapping.

iouribelikov:
> What are the current residuals? Can't reprocess with this PR just a couple of TFs which where you see the tracks and compare results? The problem with delaying merging this PR is that the CTFs currently written contain a wrong mapping.

Here are the residuals that I seem to get on the layer6, for tracks initiated on layers3, 4 and 5.
(Two days of running clusterization, to get this number of tracks. Because of copying data.)
l6y

shahor02:
indeed look strangely good. But don't you record the TF filenames which contain these tracks?

iouribelikov:
> indeed look strangely good. But don't you record the TF filenames which contain these tracks?

Yes, this information can extracted from logs !    So we can re-reconstruct just these TFs. (which will be a lot faster)

iouribelikov:
@shahor02 Hi Ruben.  So I have selectively reconstructed a few TFs containing cosmics.  
The results are unexpected: The software chip IDs assigned to cosmic-track clusters do *not* change.
(Of course, the residuals do not change as well)

shahor02:
Hi @iouribelikov 

Thanks! I have to think, indeed this is strange, I have to think what's going on. You were running starting from the raw data, not CTF, right? 
In any case, I would merge this PR once the test is passed, since Felix confirmed that the HW addresses were swapped and the fix does not deteriorate the reconstruction.

iouribelikov:
> Hi @iouribelikov
> 
> Thanks! I have to think, indeed this is strange, I have to think what's going on. You were running starting from the raw data, not CTF, right?

Yes. This was a reconstruction from raw-data TFs.

 
> In any case, I would merge this PR once the test is passed, since Felix confirmed that the HW addresses were swapped and the fix does not deteriorate the reconstruction.

Could not it be that we still have (somewhere) another "swap" ?   that goes in a "compensating direction" ?

OK to merge it.    Thank you very much for this fix !

shahor02:
There could be still many errors, but this PR only reassign chips to other HW cable. Actually, you say that residuals did not change, it this true for the abs. tracks position?

iouribelikov:
> There could be still many errors, but this PR only reassign chips to other HW cable. Actually, you say that residuals did not change, it this true for the abs. tracks position?

Nothing changes.  The global positions of cosmic clusters remain the same.

shahor02:
@MichaelLettrich are macOS, o2 and o2-dataflow-cs8 builders dead?

shahor02:
@iouribelikov ok, I've sorted out what is going on: since the Alpide data encodes the complete chipID in module, as 0:6,8-14 (and not the chip position on cable), the global chip ID extracted from 2 cables serving the same module does not change if we swap these 2 cables: each one will anyway uniquely identify the chip within the module. That's why the result of decoding does not change. 
The effective result of this PR is that:
1) chips are now decoded in global chipID increasing order (actually, I've noticed the problem after seeing that the decoding is not ordered, as it was meant to be).
2) Mapping the MC raw data now is consistent with that of real ITS. The old MC raw data should still be ok: internally it has swapped cables but since what matters is the correspondence of the chipInModule to globalChipID (which is not affected by the cabling), the positions of the clusters don't change (the 2 cables in total ship exactly the same data, simply in a different order). 
Confusing but fact...

Some CI's are dead, merging.

iouribelikov:
@shahor02 Excellent news !

PR #7046, opened 2021-09-16
O2-2582: Correct order of hits when merging

sawenzel:
Fixing an index error, where
the merged hits chunk did not correspond
to the right sub-event index.

Fixes O2-2582


sawenzel:
Hotfixing the issue. A more optimized solution and test addition will follow ... but I'd like to provide this now to unblock other stuff.

sawenzel:
@mconcas

sawenzel:
full-system-test and all unit tests succeeded locally

PR #7047, opened 2021-09-16
parsing fixes, used for kr run

bazinski:
- fix 1 remaining parsing warning for mc2raw
- handles multiple end of digit markers correctly
- remove error and warning messages, see histograms and eventually qc for those messages.
- 2d histograms for parsing errors, rather illuminating
- fix corrupted halfchamber header where additional header words and data version are zero now.
- command line option to set DigitHCHeader.numberHCW --hcheaderwords
- command line option to set data version --hcheadermajor  only major version is used.
- command line option to expressly ignore halfchamber header (location, major version and addtional words blank or identical)
- histograms neatened up
- tracklets at the position of the stuck half chamber header are still there, bit corruption, so tracklets at hcid 512, row 8 are to be ignored, remove the spike
- if a digit mcmheader is corrupt digit data is dumped, there is however a possibilty to recover digits later on in the buffer of that rob, needs more investigation for producing corrupt digits.
- stats messages to qc are stripped out of this PR

This is the hash used for Kr runs.

I have not been through all the permutations of combinations there are too many now, the ones to process previous data work though.

### DIFF
--- a/DataFormats/Detectors/TRD/CMakeLists.txt
+++ b/DataFormats/Detectors/TRD/CMakeLists.txt
@@ -15,6 +15,7 @@ o2_add_library(DataFormatsTRD
                        src/AngularResidHistos.cxx
                        src/Tracklet64.cxx
                        src/RawData.cxx
+                       src/RawDataStats.cxx
                        src/CompressedDigit.cxx
                        src/CTF.cxx
                        src/Digit.cxx
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
@@ -74,6 +74,8 @@ constexpr int MAXDATAPERLINK32 = 13824;       // max number of 32 bit words per 
 constexpr int MAXDATAPERLINK256 = 1728;       // max number of linkwords per cru link. (256bit words)
 constexpr int MAXEVENTCOUNTERSEPERATION = 200; // how far appart can subsequent mcmheader event counters be before we flag for concern, used as a sanity check in rawreader.
 constexpr int MAXMCMCOUNT = 69120;             // at most mcm count maxchamber x nrobc1 nmcmrob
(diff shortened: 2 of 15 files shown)


bazinski:
@shahor02 @tdietel can this be merged, i dont know what the deal is with o2-cs8 and macOS (who uses mac??) but this has been stuck since thursday.

shahor02:
Some people use mac... And it is pickier on some code aspects than others. Anyway, if you have no chance to test it privately on mac, will merge it once @tdietel approves it.

bazinski:
@shahor02 the mac comment as a dig at @tdietel ;-)

tdietel (dismissed):
The changes look reasonable - although I only glanced over the. Let's go ahead with the merge.

bazinski:
@shahor02 @tdietel dont merge this, it crashes for 3 timeframes in 600k, investigating. Well something crashes, not sure where.

bazinski:
parsing error now fixed for the combination of cli options we are using for corrupted data and the instance where we receive the entire link data of 27 256 bit lines of 
0x0 0x0 0xeeeeeeee 0xeeeeeeee 0xeeeeeeee 0xeeeeeeee 0xeeeeeeee 0xeeeeeeee
The combination of cli options did not take into account skipping the half chamber header when half chamber header is to be ignored. Now fixed.

bazinski:
@shahor02 @tdietel can this be merged now, goes happily through the Kr data. All the problems converting the data are external to this code. Also works fine for the mc2raw.

tdietel:
Then we should merge it.

If this is the version we use to read the krypton data, we should also note it down somewhere. Maybe in a Jira ticket, or as a git tag, or both. What do you think?

tdietel (approved):

bazinski:
ja will put the git hash once merged in the jira ticket

bazinski:
@shahor02 can this be merged?

bazinski:
@shahor02 this one

PR #7049, opened 2021-09-16
DPL: forward message early when possible

ktf:
When coupled with FairMQ 1.4.41 this should allow for parallel
processing on data matching the same O2 Signature without sacrificing
the concept of being a message passing framework.

The optimization is disabled in the case an Optional field is present,
because we do not want to forward those before their potential by-products
due to processing.


ktf:
@davidrohr @wiechula this works as expected with FairMQ 1.4.41 and my synthetic tests. It would be good if you could test in your environment.

I need to test this for analysis as well, as it might interfere with the rate limiting.

davidrohr:
OK, thx, can test it with large FST on the EPN tonight, but will have to rebuild with the new FairMQ.

wiechula:
Thanks, I'll have a look!

davidrohr:
For reference, the FST is not running with the new FairMQ alisw/alidist#3312 thus testing of this PR is on hold.

davidrohr:
@ktf : Can you restart the CI for this? The FMQ part is merged. In parallel I'll try in the FST

ktf:
i actually already did. the issue is real and i am looking into it.

Ciao,
Giulio

ktf:
So the error seems real, but I get it ragardless of this PR. So I suspect it's actually a regression in FairMQ 1.4.41/42

ktf:
Maybe I did not actually rebuild...

ktf:
Ok, I confirm that rebuilding with the new FairMQ removes the issue with the test. I am trying to reset the tests...

davidrohr:
@ktf : Seems it fails again. And the FullCI failure seems like something is not working with the custom completion policy.

ktf:
Indeed, I was missing a `break` statement.

ktf:
I also suspect we cannot forward early with the custom completion policy, because otherwise we could end up forwarding many times...

davidrohr:
I don't understand that. Can't you keep track and make sure we only forward once?
Or let's say, you only forward-early at the point in time when the run function is called and the message is consumed?
The tpc workflow has a custom policy and it consumes raw data. So if we cannot forward-early there, it will be blocking.

ktf:
> I don't understand that. Can't you keep track and make sure we only forward once?

Yes, this is probably the best option.

> Or let's say, you only forward-early at the point in time when the run function is called and the message is consumed?

The problem is that we support "Process", besides "Cunsume" which means that the subsequent iteration the message will still be there and processed (or consumed) again.

> The tpc workflow has a custom policy and it consumes raw data. So if we cannot forward-early there, it will be blocking.

Ok.

ktf:
Ok, nevermind. I think it's actually just a matter of checking for CompletionOp::Consume, with the caveat that if one uses "Process" it will not be optimized.

davidrohr:
Ok, i think that should be ok for now. Process with a custom policy is anyhow a corner case.

wiechula:
I tested with my pedestal workflow on a few TFs with 4-12 workers. It is about 30-40% faster, depending a bit on the number of parallel workers! I think this looks good.

ktf:
@wiechula nice! can you tell me about the CPU consumption?

ktf:
Merging this, however there is an additional problem for the analysis. Given we now shallow copy messages, the accounting which is being done to regulate the incoming data will be wrong. I will workaround that in a separate PR.

wiechula:
@ktf , can you explain a bit more what you mean with CPU consumption or how to measure it?

ktf:
I mean CPU load during the processing. Given you are 40% faster, I assume that's due to more cores being used in parallel, so I was wondering how the load changed.

PR #7050, opened 2021-09-16
Fix order of MC track labels in table

nburmaso:


sawenzel (approved):

nburmaso:
Here's a comparison of MC-vs-Reco pT correlation before and after the fix.

| Before  | After |
| ------------- | ------------- |
| pt_mc_vs_reco_aod | pt_mc_vs_reco_aod_test |

sawenzel:
Needed for nightly testing. MacOS compile checked locally. Merging.

PR #7051, opened 2021-09-16
DPL: use Verbosity::Debug for DPL GUI metrics

ktf:


Framework/Core/src/CommonServices.cxx
```diff
@@ -84,7 +84,9 @@ o2::framework::ServiceSpec CommonServices::monitoringSpec()
       o2::monitoring::Monitoring* monitoring;
       if (useDPL) {
         monitoring = new Monitoring();
-        monitoring->addBackend(std::make_unique<DPLMonitoringBackend>(registry));
+        auto dplBackend = std::make_unique<DPLMonitoringBackend>(registry);
+        (dynamic_cast<o2::monitoring::Backend*>(dplBackend.get()))->setVerbosisty(o2::monitoring::Verbosity::Debug);
+        monitoring->addBackend(std::move(dplBackend));
```
ktf, line 89: @awegrzyn there is actually a typo (setVerbosisty) in the API.

awegrzyn: Yes, fixing, from v3.8.9 you will be able to use both.

ktf:
Full CI and dataflow on cc7 work. macOS tested on my box. Merging.

awegrzyn:
@ktf we're doing a test with @davidrohr atm, I can still see `data/relayer/xx` metrics, would it be possible to disable them as well?

ktf:
https://github.com/AliceO2Group/AliceO2/pull/7093 should remove the last ones.

PR #7052, opened 2021-09-16
[MCH MID] treat each entry of the input tree as a separate TF

pillot:
- use RootTreeReader to read the data from the input root file and send the messages, instead of doing it "by hand".

- In case the tree has several entries, send one set of messages per entry, thus treating each of them as a separate TF. Previously the entries where concatenated in the same messages, but without changing the references to the digits in the ROFs. Concatenation of several entries in the same messages is an option that could be introduced later if needed, with an appropriate reindexing.

### DIFF
--- a/Detectors/MUON/MCH/Workflow/CMakeLists.txt
+++ b/Detectors/MUON/MCH/Workflow/CMakeLists.txt
@@ -87,7 +87,7 @@ o2_add_executable(
         sim-digits-reader-workflow
         SOURCES src/DigitReaderSpec.cxx src/sim-digits-reader-workflow.cxx
         COMPONENT_NAME mch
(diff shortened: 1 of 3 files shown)


aphecetche (dismissed):

shahor02:
@pillot there are conflicts in this branch

aphecetche:
My bad, the conflict is due to https://github.com/AliceO2Group/AliceO2/pull/7081 .
Just (hopefully) resolved the conflict.

aphecetche (dismissed):

aphecetche:
@pillot I'm afraid you'll have to rebase and force push... My attempts to fix the conflicts using GitHub web interface apparently failed... Sorry about that.

pillot:
@aphecetche I fixed the conflict locally but now I see that everything is already green, so I guess I don't need to force push anything ?

aphecetche:
the checks are indeed ok, but the merge button is not accessible. This is what I see : 

Screenshot 2021-09-28 at 11 18 23

aphecetche (approved):

PR #7057, opened 2021-09-17
check addedd to solve decoding problem temporary needed (to be debugg…

noferini:
…ed with data)


noferini:
We observed an unexpected behavior in real data.
We added a check during raw decoding to trace the origin of errors due to mismatch of electronic indexes mapped on our reconstruction mapping (rare events but it should be zero)

shahor02 (approved):
The additions seem to be trivial. If this is urgent and you have tested it locally, let me know, I can merge w/o tests

noferini:
It is urgent but let's wait it passes at least the compilation in one system.
I couldn't run it yet since I compiling in a different branch for (PR 7053)

PR #7059, opened 2021-09-17
Rewrite of TreeToTable and TableToTree

aalkin:
Classes are rewritten to be more generic and to allow later extension of supported column/branch types. Table reader, TreeToTable, gets initial support for specifying memory pool for table allocation from outside. 

Relevant benchmarks and tests are updated.

### DIFF
--- a/Framework/Core/include/Framework/TableTreeHelpers.h
+++ b/Framework/Core/include/Framework/TableTreeHelpers.h
@@ -374,7 +374,7 @@ class TreeToTable
   {
   }
   void setLabel(const char* label);
-  void addColumns(TTree* tree, std::vector<const char*>&& names = {});
+  void addColumns(TTree* tree, std::vector<std::string>&& names = {});
   void read();
   auto finalize()
   {
--- a/Framework/Core/include/Framework/TableTreeHelpers.h
+++ b/Framework/Core/include/Framework/TableTreeHelpers.h
@@ -37,7 +37,7 @@ class ColumnToBranchBase
 {
  public:
   ColumnToBranchBase(arrow::ChunkedArray* column, arrow::Field* field, int size = 1);
-  virtual ~ColumnToBranchBase(){};
+  virtual ~ColumnToBranchBase() = default;
   void at(int64_t* pos)
   {
     mCurrentPos = pos;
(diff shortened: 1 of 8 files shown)


Framework/Core/include/Framework/TableTreeHelpers.h
```diff
@@ -22,122 +22,372 @@ namespace o2::framework
...
+template <>
+struct ROOTTypeString<uint16_t> {
+  static constexpr char const* str = "/s";
+};
+
+template <>
+struct ROOTTypeString<uint32_t> {
+  static constexpr char const* str = "/i";
+};
+
```
ktf, line 86: How about something like:

```
#include <iostream>


struct ConvertMe {
template <typename T> constexpr static char foo[3] = "**";
template <> constexpr static char foo<double>[3] = "do";
template <> constexpr static char foo<float>[3] = "fl";
};

int main() {
    std::cout << ConvertMe::foo<bool> << std::endl;
}
```
?

Framework/Core/include/Framework/TableTreeHelpers.h
```diff
@@ -22,122 +22,372 @@ namespace o2::framework
...
+  auto getCurrentArray() const
+  {
+    if (listSize > 1) {
+      return std::static_pointer_cast<o2::soa::arrow_array_for_t<std::remove_pointer_t<T>>>(std::static_pointer_cast<arrow::FixedSizeListArray>(mColumn->chunk(mCurrentChunk))->values());
+    } else {
+      return std::static_pointer_cast<o2::soa::arrow_array_for_t<std::remove_pointer_t<T>>>(mColumn->chunk(mCurrentChunk));
+    }
+  }
+
+  void nextChunk() override
```
ktf, line 177: I was actually thinking we cannot do that with the current implementation of the cursor, since it relies on the chunks in each column to have the same number of elements, no?

ktf (changes requested):

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: This is, in its essence, what we do in ColumnIterator for the RowView. Number of elements can still be the same, even if the elements are composed of several plain values.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: Implemented.

aalkin: I had to revert this, because GCC 10 fails to compile it...

PR #7062, opened 2021-09-17
Misc. fixes and optimization for global tracking / vertxing

shahor02:
@sawenzel this should fix https://alice.its.cern.ch/jira/browse/O2-2578


shahor02:
Also the build/O2/o2-dataflow-cs8  and build/O2/o2  CIs seem to be dead.
Merging as this is an important bug-fix

PR #7064, opened 2021-09-19
MCH: fix some (stupid) errors in DigitIOV4

aphecetche:


aphecetche:
@shahor02 one more green PR to be merged (this one is not relevant for data taking but only for internal Mch debugging)

shahor02 (approved):

PR #7065, opened 2021-09-19
Optionally read remote CTFs w/o copy (as root network file)

shahor02:
There is a possibility to read remote root files directly, w/o caching it locally. For that one should:
1) provide the full URL the remote files, e.g. if the files are supposed to be accessed by `xrootd` (the `XrdSecPROTOCOL` and `XrdSecSSSKT` env. variables should be set up in advance): `root://eosaliceo2.cern.ch//eos/aliceo2/ls2data/...root` 
(use `xrdfs root://eosaliceo2.cern.ch ls -u <path>` to list full URL).
2) provide proper regex to define remote files, e.g. for the example above: `--remote-regex "^root://.+/eos/aliceo2/.+"`.
3) pass an option `--copy-cmd no-copy`.

### DIFF
--- a/Common/Utils/include/CommonUtils/FileFetcher.h
+++ b/Common/Utils/include/CommonUtils/FileFetcher.h
@@ -99,6 +99,7 @@ class FileFetcher
   size_t mNRemote{0};
   size_t mMaxInQueue{5};
   bool mRunning = false;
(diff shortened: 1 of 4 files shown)


shahor02:
o2-dataflow-cs8  failure is unrelated, merging.

PR #7070, opened 2021-09-20
Fix access to track PID in AOD TOF storage

shahor02:


shahor02:
Merging as this fixes `o2-aod-producer-workflow` which breaks fullCI

PR #7073, opened 2021-09-21
Rewrite TableToTree to be more generic

aalkin:
@ktf first piece of the split #7059


Framework/Core/include/Framework/TableTreeHelpers.h
```diff
@@ -22,97 +22,205 @@ namespace o2::framework
...
-  float* mVariable_f = nullptr;
-  double* mVariable_d = nullptr;
-
-  // initialize a branch
-  bool initBranch(TTree* tree);
-
-  // initialize chunk ib
-  bool initDataBuffer(Int_t ib);
+ protected:
+  std::string mBranchName;
```
ktf: Please move this to a separate file.

Framework/Core/include/Framework/TableTreeHelpers.h
```diff
@@ -22,97 +22,205 @@ namespace o2::framework
...
+struct ROOTTypeString<bool> {
+  static constexpr char const* str = "/O";
+};
 
-  // a list of BranchIterator
-  std::vector<std::unique_ptr<BranchIterator>> mBranchIterators;
+template <>
+struct ROOTTypeString<uint8_t> {
+  static constexpr char const* str = "/b";
+};
```
ktf: Do you really need this in the .h? Can it be moved to the CXX?

Framework/Core/include/Framework/TableTreeHelpers.h
```diff
@@ -22,97 +22,205 @@ namespace o2::framework
...
+template <>
+struct ROOTTypeString<double> {
+  static constexpr char const* str = "/D";
+};
+
+template <typename T>
+constexpr auto ROOTTypeString_t = ROOTTypeString<T>::str;
+
+template <typename T>
+class ColumnToBranch : public ColumnToBranchBase
```
ktf: Can't you convert T to a string and remove the need for templating?

Framework/Core/include/Framework/TableTreeHelpers.h
```diff
@@ -22,97 +22,205 @@ namespace o2::framework
...
+};
+
+class GenericTreeToTable
+{
+ public:
+ private:
+  std::shared_ptr<arrow::Table> mTable;
+  std::vector<std::shared_ptr<arrow::Field>> mFields;
+  std::string mTableLabel;
 };
```
ktf: ??

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -85,310 +85,124 @@ class ColumnIterator
...
 
-BranchIterator::~BranchIterator()
+TableToTree::TableToTree(std::shared_ptr<arrow::Table> table, TFile* file, const char* treename)
 {
-  delete mBranchBuffer;
-
-  delete[] mVariable_o;
+  mTable = table.get();
+  mTree = static_cast<TTree*>(file->Get(treename));
+  if (mTree == nullptr) {
```
ktf:
```
if (mTree != nullptr) {
  return;
}
```

etc...

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -85,310 +85,124 @@ class ColumnIterator
...
-  // try to find branch in tree
-  mBranchPtr = tree->GetBranch(mBranchName.c_str());
-  if (mBranchPtr) {
-    return true;
+  if (mRows == 0) {
+    mRows = column->length();
+  } else {
+    if (mRows != column->length()) {
+      throw runtime_error_f("Adding incompatible column with size %d (num rows = %d)", column->length(), mRows);
+    }
```
ktf: `else if`

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -85,310 +85,124 @@ class ColumnIterator
...
-
-  bool togo = (mTreePtr->GetNbranches() > 0) && (mTable->num_rows() > 0);
-  while (togo) {
-    // fill the tree
-    mTreePtr->Fill();
-
-    // update the branches
-    for (auto& brit : mBranchIterators) {
-      togo &= brit->push();
+  int64_t mRow = 0;
```
ktf: do not prefix local variables with `m`.

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -85,310 +85,124 @@ class ColumnIterator
...
+  bool writable = (mTree->GetNbranches() > 0) && (mRows > 0);
+  if (writable) {
```
ktf:
```
if (!writable) {
   mTree->Write("", TObject::kOverwrite);
   return mTree;
}
```

etc.

Framework/Core/include/Framework/TableTreeHelpers.h
```diff
@@ -22,97 +22,205 @@ namespace o2::framework
...
-  // write table to tree
   TTree* process();
+  void addBranch(std::shared_ptr<arrow::ChunkedArray> column, std::shared_ptr<arrow::Field> field);
+  void addAllBranches();
+
+ private:
+  arrow::Table* mTable;
+  int64_t mRows = 0;
+  TTree* mTree = nullptr;
+  std::vector<std::unique_ptr<ColumnToBranchBase>> mColumnReaders;
```
ktf: Why do you need a unique_ptr? Are you moving them around?

ktf (changes requested):

ktf:
I wrote the first set of things I notices, I might have a few more comments.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: You mean move `CoumnToBranchBase`/`ColumnToBranch` classes?

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: It can be moved.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: I need the template to define the actual storing pointer, or at least cast it to correct type.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: A leftover from splitting, I'll remove it.

Framework/Core/src/TableTreeHelpers.cxx
aalkin: No, this is supposed to create a new tree in case file does not have it.

Framework/Core/src/TableTreeHelpers.cxx
aalkin: Leftover from this variable being a data member, I will rename it.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: I need pointers for polymorphism so this ensures ownership.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: Moved to .cxx

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: Moved.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: Removed.

Framework/Core/src/TableTreeHelpers.cxx
aalkin: Done.

Framework/Core/src/TableTreeHelpers.cxx
aalkin: Improved.

Framework/Core/src/TableTreeHelpers.cxx
aalkin: Fixed.

Framework/Core/src/TableTreeHelpers.cxx
aalkin: Sorry, misunderstood at first, fixed now.

ktf: Yes, that you can have afterwards, but if you already have a tree, there is nothing else that needs to be done.

Framework/Core/test/test_TreeToTable.cxx
```diff
@@ -96,15 +96,17 @@ BOOST_AUTO_TEST_CASE(TreeToTableConversion)
...
+  BOOST_REQUIRE(table->column(0)->type()->Equals(arrow::boolean()));
+  BOOST_REQUIRE(table->column(1)->type()->Equals(arrow::float32()));
+  BOOST_REQUIRE(table->column(2)->type()->Equals(arrow::float32()));
+  BOOST_REQUIRE(table->column(3)->type()->Equals(arrow::float32()));
+  BOOST_REQUIRE(table->column(4)->type()->Equals(arrow::float64()));
+  BOOST_REQUIRE(table->column(5)->type()->Equals(arrow::int32()));
+  BOOST_REQUIRE(table->column(6)->type()->Equals(arrow::fixed_size_list(arrow::float64(), nelem)));
+  BOOST_REQUIRE(table->column(7)->type()->Equals(arrow::fixed_size_list(arrow::boolean(), 5)));
+  BOOST_REQUIRE(table->column(8)->type()->Equals(arrow::fixed_size_list(arrow::float32(), 96)));
+  BOOST_REQUIRE(table->column(9)->type()->Equals(arrow::uint8()));
```
ktf, line 120: The old version has the advantage that if something breaks we see both sides of the equation. Please do not change it.

Framework/Core/test/test_TreeToTable.cxx
```diff
@@ -134,9 +136,9 @@ BOOST_AUTO_TEST_CASE(TreeToTableConversion)
   BOOST_REQUIRE_EQUAL(ntruein[1], ntrueout);
 
   // save table as tree
-  TFile* f2 = new TFile("table2tree.root", "RECREATE");
+  auto* f2 = TFile::Open("table2tree.root", "RECREATE");
   TableToTree ta2tr(table, f2, "mytree");
-  stat = ta2tr.addAllBranches();
+  ta2tr.addAllBranches();
```
ktf: This is not strictly needed, is it?

Framework/Core/include/Framework/TableTreeHelpers.h
ktf: That's actually the main issue I have. Why do you need polimorphism at all, can't you convert the type upfront and do everything at runtime, without the need for templates?

Framework/Core/test/test_TreeToTable.cxx
aalkin: The old version compares only type-id, but not the actual type object. I can just put in both versions then, redundancy in tests is not a problem.

Framework/Core/test/test_TreeToTable.cxx
aalkin: True. I'll revert it.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: With too many switches, which is exactly what I wanted to avoid. Adding other arrow types would be *much* easier with this.

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -83,312 +95,259 @@ class ColumnIterator
...
 {
-  // try to find branch in tree
-  mBranchPtr = tree->GetBranch(mBranchName.c_str());
-  if (mBranchPtr) {
-    return true;
+ public:
+  ColumnToBranch(TTree* tree, arrow::ChunkedArray* column, arrow::Field* field, int size = 1)
+    : ColumnToBranchBase(column, field, size)
+  {
+    if constexpr (std::is_pointer_v<T>) {
```
ktf: This can be expressed as well using field, no? Why do you need it to be constexpr?

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -83,312 +95,259 @@ class ColumnIterator
...
+      mLeaflist = mBranchName + "[" + std::to_string(listSize) + "]" + ROOTTypeString_t<std::remove_pointer_t<T>>;
+    } else {
+      mLeaflist = mBranchName + ROOTTypeString_t<T>;
```
ktf: Same for ROOTTypeString_t. Can't you derive that information from field?

ktf (changes requested):

Framework/Core/include/Framework/TableTreeHelpers.h
ktf: I am all for removing the switches, but since you do have the field I think most of the if constexpr which you can become normal if, taking advantage of the field, which you have in any case. Basically, rather than having something which converts from a basic type t to root type, what you need is arrow type to root type. I see already quite a few places where that would actually work.

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -83,312 +95,259 @@ class ColumnIterator
...
-    chunkToUse = std::dynamic_pointer_cast<arrow::FixedSizeListArray>(chunkToUse)->values();
+  void accessChunk(int at)
+  {
+    auto array = getCurrentArray();
+    if constexpr (std::is_same_v<bool, std::remove_pointer_t<T>>) {
+      for (auto i = 0; i < listSize; ++i) {
+        mCurrent[i] = (bool)array->Value((at - mFirstIndex) * listSize + i);
+      }
+    } else {
+      mCurrent = (std::remove_pointer_t<T>*)array->raw_values();
```
ktf: This is the only place where you really need T* with the current code, but even in that case you are basically using it just to get the size of a single element, which again you could do based on the field.

Framework/Core/include/Framework/TableTreeHelpers.h
aalkin: Field is still dynamic, one would still need a switch to use it. Which is done once when the instance of the ColumnToBranch is added.

ktf:
As we discussed privately, I think you can simply use the type erased API
to write a branch.

Ciao,
Giulio

aalkin:
There is a genuine crash in `internal-dpl-aod-writer`, I'm looking into it.

ktf:
@aalkin looks good now. I also checked it on hyperloop and I did not see any regression.

PR #7080, opened 2021-09-21
DDS to ODC

ktf:


ktf:
@davidrohr indeed @dennisklein gave me a heads up about this but I forgot when we switched to the new FairMQ.

davidrohr:
@ktf : Could you check what failed here in the ctest? Would be good to merge this soon.

dennisklein:
Looks like a false negative to me: https://github.com/AliceO2Group/AliceO2/blob/8f7f5ef3bc1c5c46b977799e6510b237aaa0150e/Framework/Core/test/test_FrameworkDataFlowToDDS.cxx#L118 - expected string needs to be adapted, too.

ktf:
Indeed, I must have accidentally hit u before saving.

PR #7081, opened 2021-09-21
MCH: consistent specnames for DataProcessorSpec getters

aphecetche:
@pillot @aferrero2707 The idea of this PR is to be a bit more consistent in our parameter list for the `getXXX` functions that return `DataProcessSpec` objects. So the proposal is, when defining a function that returns a `DataProcessorSpec`, to stick to the following pattern for its parameters :

```
DataProcessorSpec getXXX([bool useMC], const char* specName="mch-xxx", other parameters);
```

- first parameter, if relevant, should be a boolean indicating whether the processor has to deal with Monte Carlo data or not. For a processor that never has to deal with MC, leave that parameter out
- second parameter is the name by which that device will be referenced in all log files, so, in order to be easily recognizable, it *must* start with the prefix `mch-`
- the rest of the parameters (if any) is specific to each device

As an aside, @davidrohr, this should fix the "anonymous DataDecoder issue" in workflow logs you reported some time ago.

### DIFF
--- a/Detectors/MUON/MCH/Calibration/src/PedestalCalibSpec.h
+++ b/Detectors/MUON/MCH/Calibration/src/PedestalCalibSpec.h
@@ -158,12 +158,7 @@ class PedestalCalibDevice : public o2::framework::Task
 namespace framework
 {
 
-std::string getMCHPedestalCalibDeviceName()
-{
-  return "calib-mch-pedestal";
-}
-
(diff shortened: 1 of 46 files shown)


davidrohr:
Thx, then plase update the decoder name in full system test dpl-workflow.sh.

aphecetche:
@davidrohr `dpl-workflow.sh` change has been made and the PR is fully green (a situation I've not witnessed for a very long time 😄)

PR #7082, opened 2021-09-21
Fix for extreme case of cluster phi

mpuccio:


mpuccio:
@shahor02, this is the "proper" fix: in Sandro's event there was a cluster with a ridiculously small negative phi angle (~-1.e-8), that when summed with 2pi gave a border condition in our code that was not properly treated. The current dev just skips these clusters (leading to extremely small inefficiency), this PR fixes that.
(cc @mconcas)

shahor02 (approved):
Thanks @mpuccio ! Merging as the change is trivial.

PR #7084, opened 2021-09-21
CPV pedestal calibration based on TimeSlotCalibration

sevdokim:
Hello! Workflow based on TimeSlotCalibration for CPV pedestal calibration is created.

### DIFF
--- a/Detectors/CPV/base/include/CPVBase/CPVSimParams.h
+++ b/Detectors/CPV/base/include/CPVBase/CPVSimParams.h
@@ -63,11 +63,11 @@ struct CPVSimParams : public o2::conf::ConfigurableParamHelper<CPVSimParams> {
   bool mUnfoldClusters = false;      ///< Perform cluster unfolding?
 
   // Parameters used in pedestal calibration
-  uint16_t mPedClbToleratedGapWidth = 5;    ///< Tolerated gap between bins: if |bin1 - bin2| < width -> bin1 and bin2 belongs to same peak
-  float mPedClbToleratedChannelEfficiencyLow = 0.9; ///< Tolerated channel efficiency (lower limit)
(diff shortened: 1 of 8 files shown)


kharlov (dismissed):

kharlov (dismissed):

sevdokim:
Hello @davidrohr @shahor02 all,

today I was fighting with o2codechecker and finally found a way of running it locally. 
Why not to add several strings to https://github.com/AliceO2Group/CodingGuidelines#readme explaining how to run it? I think it could save some time for people like me and also CPU resources of testing facilities. Running the codechecker is pretty simple (it was enough to run `aliBuild build o2checkcode --defaults o2` in my case), but it really helps to speed up passing the checks. Sorry if I write something obvious. 

Cheers,
Sergey

davidrohr:
Hi, thx for the feedback. Feel free to add some documentation..my problem is that e.g. on my laptop it doesn't work due to different place of system headers. What i personally do is to run the command you wrote in the docker container. Thus, you see, it is not so simple :(

kharlov (approved):

sevdokim:
Hi @shahor02 
I'm a bit worry that build/O2/fullCI check has fell into "Expected - Waiting for status to be reported" for a while. Could it be that it is stuck somewhere? Other checks are passed successfully and I'm wandering which status is expected to be reported?
Thanks, cheers,
Sergey

shahor02:
Hi @sevdokim 
Giulio has checked and did not see any anomaly, one should wait a little bit more. Different checks are independent of each other.

sevdokim:
> Hi @sevdokim
> Giulio has checked and did not see any anomaly, one should wait a little bit more. Different checks are independent of each other.

Hi @shahor02 
Thanks for the info. Now I keep patience.

shahor02 (approved):

PR #7085, opened 2021-09-22
Alpide decoder protection againg wrong row + mute extra lanes error reporting

shahor02:


davidrohr (changes requested):
I think instead of setting the verbosity to 0 for all but the first instance, it would be better to use a static variable, that counts the errors, and then mute after 10 errors. In that way, Over the EPN farm we could get a spam at the beginning, but each process could only send 10 errors.
Conversely, if there are errors, it is guaranteed that we get at least some messages, i.e. when there are no error messages we know that the data is clean.

shahor02:
@davidrohr as I wrote in the email, already now each type of error is logged only once, we see so many errors simply because there are many chips (and error types). If we limit the total number of error messages sent by a process, we will simply lose some errors. The way I've implemented it here, we will see at least all repeating errors.

davidrohr:
ah, ok, didn't fully understand. Before there was already a check that every error can be printed only once. With the new patch, every error will be printed only once, but in addition also for the first instance. This means if an error happens only for instances > 1 we will not see it?

Then I would propose:
```
static int globalCount == 0;
if (thisErrorCount <= 1 && (globalCount++ <= 10 || instance == 0) printError(...);
```

shahor02:
> static int globalCount == 0;
> if (thisErrorCount <= 1 && (globalCount++ <= 10 || instance == 0) printError(...);

but these are different processes, not threads, each will have its own copy of the globalCount. One can do this probably via names semaphores, but I am afraid this will cause overheads (we already pay a price of checking lot of error conditions). Let me clarify 1st with ITS HW people why there are some many errors and if they need their reporting.

davidrohr:
But how many processes are there? there is one its raw decoder per EPN? So obviously we will get the message up to 250 times. I think this is unavoidable. Perhaps we should have a chat on this...

shahor02:
Currently, production/dpl-workflow.sh defines 3 ITS decoders per EPN in the global workflows.

davidrohr (approved):
ok, sorry, only now fully understood how this works. It is perfect as is...
sorry for the noise...

PR #7086, opened 2021-09-22
DPL: workaround ICE on XCode 13

ktf:
Apparently having a critical section in a lambda breaks the new clang.
Moving it to when the lambda is actually called keeps omp happy.

### DIFF
--- a/Detectors/TPC/workflow/src/ChunkedDigitPublisher.cxx
+++ b/Detectors/TPC/workflow/src/ChunkedDigitPublisher.cxx
@@ -178,7 +178,7 @@ void publishMergedTimeframes(std::vector<int> const& lanes, std::vector<int> con
 
         using AccumType = std::decay_t<decltype(makePublishBuffer<decltype(data)>(pc, sector, activeSectors))>;
         AccumType accum;
-#pragma omp critical
+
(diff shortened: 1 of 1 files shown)


ktf:
@sawenzel only happens with the new xcode.

Detectors/TPC/workflow/src/ChunkedDigitPublisher.cxx
```diff
@@ -199,6 +199,7 @@ void publishMergedTimeframes(std::vector<int> const& lanes, std::vector<int> con
     //data definitions
     using DigitsType = std::vector<o2::tpc::Digit>;
     using LabelType = o2::dataformats::MCTruthContainer<o2::MCCompLabel>;
+#pragma omp critical
```
sawenzel, line 202: It probably then needs to be applied to every function invocation, not just this one (and then it becomes ugly).
Also now the lock will be on a  much larger section of code than before. So, in the end, we will probably need to convert the lambda into a regular template function.

sawenzel:
I made https://github.com/AliceO2Group/AliceO2/pull/7105 to solve this (hopefully) without changing the runtime behaviour.

ktf:
thanks, I will try it in a bit.

Ciao,
Giulio

PR #7088, opened 2021-09-22
[WIP][EMCAL-534] Move error handling to CaloFitResults

mfasDa:
- Raw fitter classes do not throw exceptions any more
  but functions which might fail return std::optional
- CaloFitResults keeps track of the error status
- In case of a fit error the CaloFitResult is initialized
  only with the error status
- CaloFitResult throws a RawFitterError_t in case
  users try to access parameters of the fit, however
  this can be prevented if the user checks the fit
  status with isFitOK and ignores the fit results in case
  the fit results are not OK.

### DIFF
--- a/Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CaloFitResults.h
+++ b/Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CaloFitResults.h
@@ -37,8 +37,34 @@ namespace emcal
 /// procedures than A different meaning Fitting is applied
 class CaloFitResults
 {
-
  public:
+  /**
+   * \enum RawFitterError_t
+   * \brief Error codes for failures in raw fitter procedure
+   */
+  enum class RawFitterError_t {
+    SAMPLE_UNINITIALIZED, ///< Samples not initialized or length is 0
+    FIT_ERROR,            ///< Fit procedure failed
+    CHI2_ERROR,           ///< Chi2 cannot be determined (usually due to insufficient amount of samples)
+    BUNCH_NOT_OK,         ///< Bunch selection failed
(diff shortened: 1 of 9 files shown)


mfasDa:
WIP - don't yet merge!

mfasDa:
It was discussed in on the SHIFT-EPN-PDP channel that the raw fitter might be too slow because a lot of exceptions are thrown. This is indeed the case particularly for zero-suppressed data which cannot be fit because the max. ADC value is at a calo bunch edge, inconsistent with the pulse shape model represented by a gamma2 function. In the case of non-ZS data ADC samples are read out as 1 bunch for ALL channels in the EMCAL, and for all channels failing in the fit as described above an exception is raised. I want to emphasise that this situation is very much connected to the noise data we are taking at the moment. 

In order to circumvent this the error handling can be absorbed in the CaloFitResults object. The raw fitter and helper functions are exception-free, where all functions which might fail return std::optional instead. In case any of the std::optional fail the evaluate functions construct a CaloFitResults objects with only the error state. Users can check the error state with the isFitOK function and disregard the object in case of a fit error. If the users decide to ignore this and nevertheless try to access parameters of the fit results in case of a fit error, the exception is thrown nevertheless,  but this time by the getters. As the fit status is checked in the raw to cell converter the exception should never occur.

I am not sure this is faster, because we now return std::optional by the functions that might end up in error, which need to be handled, and std::optionals are sometimes tuples or arrays. In the exception handling we were just throwing integers, not full objects.

Adding @mpoghos to the discussion as well.

shahor02:
@mfasDa note that there are conflicts here to resolve

mfasDa:
Up

PR #7090, opened 2021-09-22
HistogramRegistry: provide add() function that directly returns hist pointer

mario-krueger:
fyi @njacazio 
with this we could do things like
```
auto myHist = histos.add<TH1>("myHist", "", kTH1F, {{10, 0.5, 10.5}}));
myHist->GetXaxis()->SetBinLabel(1, "abc"); 
```


jgrosseo:
Does not compile on Mac

jgrosseo (dismissed):

jgrosseo (approved):

PR #7091, opened 2021-09-22
[EMCAL-630] Fill cell container only for cells where the raw fit was …

mfasDa:
…successful.


mfasDa:
Is there a chance to get this in the next FLP suite?

shahor02 (approved):

PR #7092, opened 2021-09-22
skip TOF LTM during decoding

noferini:


noferini:
This should remove many errors coming from LTM which cannot be mapped in digits

noferini:
All checks passed. Waiting for review to be merged

Detectors/TOF/reconstruction/src/Decoder.cxx
```diff
@@ -182,8 +187,18 @@ void Decoder::readTRM(int icru, int icrate, uint32_t orbit, uint16_t bunchid)
   DigitInfo digitInfo;
 
   for (int i = 0; i < nhits; i++) {
+    if (icrate % 2 == 1 && itrm == 3 && mUnion[icru]->packedHit.tdcID / 3 > 0) { // Signal not coming from a TOF pad but -probably- from a TOF OR signal
+      // add operation on LTM (Trigger) if needed (not used)
+      continue;
+    }
```
shahor02: In this check there is nothing which depends on the loop index or local variables, why is it done inside the loop?

noferini: Hi Ruben,
thanks to point out this.
Actually mUnion[icru] is depending on the loop but I forgot to increment it before the "continue" statements (two times).
I was introducing a dangerous bug. Now, I fixed it

PR #7093, opened 2021-09-22
DPL: more verbosity reduction for GUI metrics

ktf:


davidrohr:
tested locally, merging

PR #7094, opened 2021-09-22
Improve checker for MID raw data consistency

dstocco:
This PR consists pf three commits:
- the first one temorarily remove some checks. In real life we do not expect calibration triggers and physics triggers happening at the same time. But in the tests we do, so we temporairly remove this check.
- the second one improves the treatment of busy local board
- the third one decouples the clear of the statistics information and of the inner object. This is needed for the Raw data QC since the statistics information needs to be cleared at every cycle.


shahor02:
is this ready to be merged?

aphecetche (approved):

Issue #294, opened 2017-04-03
Run unit tests during PR checking 


sawenzel:
As a developer I would like that the unit tests are run in the PR - CI.

ktf:
Agreed. alisw/alidist#668 should take care of running them. We should really clean the tests up, as discussed, before we enable the actual testing.

matthiasrichter:
I think the unit test are executed now, this issue can probably be closed

sawenzel:
Ok. It seems to be the case indeed. Closing.

PR #7099, opened 2021-09-23
TPC: multidimentional histograms for the CalibdEdx

badarots:
This is a revival of #6736.

- Uses multidimensional boost histograms to store dEdx values from tracks.
- The dimensions stored are: dEdx, sectors, stack type, charge type, Z, tan(lambda).
- It is possible to project any of those dimensions to reduce the histogram dimensionality.
- Fit correction based on the histogram data, the type of fit depends on the available statistics.
- Apply corrections to the clusters during reconstruction


DataFormats/Detectors/TPC/include/DataFormatsTPC/CalibdEdxCorrection.h
```diff
@@ -0,0 +1,60 @@
...
+{
+
+class CalibdEdxCorrection
+{
+ public:
+  using Params = std::array<float, 6>;
+  CalibdEdxCorrection() { clear(); }
+  CalibdEdxCorrection(std::string_view fileName) { loadFile(fileName); }
+
+  float correction(const StackID&, ChargeType, float z = 0, float tgl = 0) const;
```
wiechula:
```suggestion
  float getCorrection(const StackID&, ChargeType, float z = 0, float tgl = 0) const;
```

DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx
```diff
@@ -0,0 +1,70 @@
...
+
+// o2 includes
+#include "DataFormatsTPC/Defs.h"
+
+// root includes
+#include "TFile.h"
+
+using namespace o2::tpc;
+
+size_t CalibdEdxCorrection::stackIndex(const StackID& stack, ChargeType charge)
```
wiechula: Perhaps you can put the calculation directly in the return statement and put it directly in the header.

DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx
```diff
@@ -0,0 +1,70 @@
...
+  for (auto& x : mParams) {
+    std::fill(x.begin(), x.end(), 0.0);
+  }
+  std::fill(mChi2.begin(), mChi2.end(), 0.0);
+  mDims = std::numeric_limits<int>::min();
+}
+
+void CalibdEdxCorrection::saveFile(std::string_view fileName) const
+{
+  TFile file(fileName.data(), "recreate");
```
wiechula: I would prrefer
```suggestion
  std::unique_ptr<TFile> file(TFile::Open(fileName.data(), "recreate"));
```
And the same below.

DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx
```diff
@@ -0,0 +1,70 @@
...
+  file.WriteObject(this, "CalibdEdxCorrection");
+  file.Close();
+}
+
+void CalibdEdxCorrection::loadFile(std::string_view fileName)
+{
+  TFile file(fileName.data());
+  auto corr = file.Get<CalibdEdxCorrection>("CalibdEdxCorrection");
```
wiechula: Can't one use
```suggestion
 file.GetObject("CalibdEdxCorrection", *this);
```

wiechula (changes requested):
Very nice! A few quick comments. More later.

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -11,91 +11,283 @@
...
+    constexpr std::array<float, 4> xks{108.475f, 152.2f, 189.4f, 228.4f};
+
+    bool ok = cpTrack.propagateTo(xks[stack], mField);
+
+    // Ignore stack if we are not able to find its sector
+    if (!ok) {
+      continue;
+    }
+
+    float phi = cpTrack.getAlpha();
```
wiechula:
```suggestion
   const float phi = o2::math_utils::bringTo02PiGen(cpTrack.getAlpha());
```
And make the ones below const.

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -11,91 +11,283 @@
...
+}
+
+void CalibdEdx::fill(const gsl::span<const TrackTPC> tracks)
+{
+  for (const auto& track : tracks) {
+    fill(track);
+  }
+}
+
+void CalibdEdx::fill(const std::vector<TrackTPC>& tracks)
```
wiechula: Perhaps call the `gsl::span` version

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -11,91 +11,283 @@
...
+{
+  for (const auto& track : tracks) {
+    fill(track);
+  }
+}
 
-void CalibdEdx::initOutput()
+void CalibdEdx::merge(const CalibdEdx* other)
```
wiechula, line 100: Check for `nullptr` or make const ref instead of pointer.
For interace rasons we only need pointer in the calibrator.

wiechula (changes requested):
A few more remarks.

Detectors/TPC/calibration/src/CalibdEdx.cxx
badarots: Uhhh, I didnt know I could just write gsl::span(tracks) to convert to span, awesome!

Detectors/TPC/calibration/src/CalibdEdx.cxx
badarots: Not sure about the const ref. I thought the time slot container needs a merge method that receives a pointer.
https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/Calibration#timeslot

DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx
badarots: I tried the following:

```file.GetObject("CalibdEdxCorrection", this); // passing a pointer```

but didnt work and produced this error:
Candidate function [with T = o2::tpc::CalibdEdxCorrection] not viable: expects an lvalue for 2nd argument
Im not sure how I can get around this.

File: DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx (hunk 3 above)
Comment (badarots) [reply DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx]: Do I need to call file->Close() in the end or the destructor will do it for me?

File: DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx (hunk 4 above)
Comment (wiechula) [reply DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx]: Could be `&this`. I wiil check.

File: Detectors/TPC/calibration/src/CalibdEdx.cxx (hunk 7 above)
Comment (wiechula) [reply Detectors/TPC/calibration/src/CalibdEdx.cxx:100]: Ahh, sorry. You are right, we use it in the time slot. Sorry, I forgot about this.

File: DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx (hunk 3 above)
Comment (wiechula) [reply DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx]: I think the destuctor should do it. But I would need to check the code. In doubt just add the `Close` statement.

File: Detectors/TPC/calibration/src/CalibdEdx.cxx (hunk 7 above)
Comment (badarots) [reply Detectors/TPC/calibration/src/CalibdEdx.cxx:100]: No problem, and thank you for catching this bug, I added the nulltprt check!

**Comment by wiechula** [comment]
All in all very nice!
As next step we need to finalize the GPU part and make sure it compiles and works on nvidia and amd.

File: DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx (hunk 4 above)
Comment (wiechula) [reply DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx]: Ok, indeed this does not work. No sure why.

Comment (badarots) [reply DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx]: maybe it is a good idea to keep it as I did, and also add a check for nulltptr

```
  auto tmp = file->Get<CalibdEdxCorrection>("CalibdEdxCorrection");
  if (tmp != nullptr) {
    *this = *tmp;
  }
```

File: DataFormats/Detectors/TPC/include/DataFormatsTPC/CalibdEdxCorrection.h [hunk 8]
```diff
@@ -0,0 +1,96 @@
...
+
+ private:
+  GPUd() static size_t stackIndex(const StackID& stack, ChargeType charge)
+  {
+    return static_cast<size_t>(stack.index() + charge * SECTORSPERSIDE * SIDES * GEMSTACKSPERSECTOR);
+  }
+
+  std::array<Params, 288> mParams{};
+  std::array<float, 288> mChi2{};
+  int mDims{-1}; ///< Fit dimension
```
wiechula, line 90: Please add
```suggestion
  int mDims{-1}; ///< Fit dimension
  
  ClassDefNV(CalibdEdxCorrection, 1);
```

DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx
```diff
@@ -0,0 +1,49 @@
...
+void CalibdEdxCorrection::clear()
+{
+  for (auto& x : mParams) {
+    std::fill(x.begin(), x.end(), 0.0);
+  }
+  std::fill(mChi2.begin(), mChi2.end(), 0.0);
+  mDims = -1;
+}
+
+void CalibdEdxCorrection::saveFile(std::string_view fileName) const
```
wiechula: perhpas
```suggestion
void CalibdEdxCorrection::dumpToFile(std::string_view fileName) const
```
or
```suggestion
void CalibdEdxCorrection::writeToFile(std::string_view fileName) const
```

DataFormats/Detectors/TPC/src/CalibdEdxCorrection.cxx
```diff
@@ -0,0 +1,49 @@
...
+{
+  std::unique_ptr<TFile> file(TFile::Open(fileName.data(), "recreate"));
+  file->WriteObject(this, "CalibdEdxCorrection");
+}
+
+void CalibdEdxCorrection::loadFile(std::string_view fileName)
```
wiechula: Perhaps
```suggestion
void CalibdEdxCorrection::loadFromFile(std::string_view fileName)
```

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -11,91 +11,296 @@
...
-    mDebugOutputStreamer->Close();
+  const float lower = mHist.axis(0).begin()->lower();
+  const float upper = mHist.axis(0).end()->lower();
+
+  auto projectedHist = getHist(projected_axis);
+
+  const int nBins = mHist.axis(Axis::dEdx).size();
+  const int nHists = projectedHist.size() / nBins;
+
+  TH2F rootHist("", "", nHists, 0, nHists, nBins, lower, upper);
```
wiechula: Why no name and title? Can this be deduced from the axes names? Also the axis titles could be set I think.

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -38,86 +35,72 @@ class CalibdEdxDevice : public Task
...
+    const int minEntries1D = ic.options().get<int>("min-entries-1d");
+    const int minEntries2D = ic.options().get<int>("min-entries-2d");
+
+    const int dEdxBins = ic.options().get<int>("dedxbins");
+    const int zBins = ic.options().get<int>("zbins");
+    const int angularBins = ic.options().get<int>("angularbins");
+    const float mindEdx = ic.options().get<float>("min-dedx");
+    const float maxdEdx = ic.options().get<float>("max-dedx");
+
+    const bool dumpData = ic.options().get<bool>("file-dump");
```
wiechula:
```suggestion
   mDumpToFile  = ic.options().get<bool>("file-dump");
```

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -38,86 +35,72 @@ class CalibdEdxDevice : public Task
...
+    float field = ic.options().get<float>("field");
+
+    const auto inputGRP = o2::base::NameConf::getGRPFileName();
+    const auto grp = o2::parameters::GRPObject::loadFrom(inputGRP);
+    if (grp != nullptr) {
+      field = 5.00668f * grp->getL3Current() / 30000.;
+      LOGP(info, "Using GRP file to set the magnetic field to {} kG", field);
     }
```
wiechula: Like this, the grp would overwrite the value from the command line. Since the grp will always be required to not crash, the `field` command line parameter is useless. I guess the logic should be the other way around.
This should then be combined with some dummy value in the default parameter, like `-100`. And the checked
```c++
if (ic.options().get<float>("field") > -10) {
field = ic.options().get<float>("field");
}
```
or so.

Detectors/TPC/workflow/src/CalibratordEdxSpec.cxx
```diff
@@ -0,0 +1,169 @@
...
+
+    const bool dumpData = ic.options().get<bool>("file-dump");
+    float field = ic.options().get<float>("field");
+
+    const auto inputGRP = o2::base::NameConf::getGRPFileName();
+    const auto grp = o2::parameters::GRPObject::loadFrom(inputGRP);
+    if (grp != nullptr) {
+      field = 5.00668f * grp->getL3Current() / 30000.;
+      LOGP(info, "Using GRP file to set the magnetic field to {} kG", field);
+    }
```
wiechula: same as above.

Detectors/TPC/workflow/src/tpc-calibrator-dEdx.cxx
```diff
@@ -0,0 +1,21 @@
...
+
+#include "TPCWorkflow/CalibratordEdxSpec.h"
+#include "Framework/runDataProcessing.h"
+
+using namespace o2::framework;
+
+WorkflowSpec defineDataProcessing(ConfigContext const&)
+{
+  using namespace o2::tpc;
+  return WorkflowSpec{getCalibratordEdxSpec()};
```
wiechula: We should also have this `configKeyValues` stuff you put in `tpc-calib-dEdx.cxx` here, I think.

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -251,13 +253,22 @@ DataProcessorSpec getGPURecoWorkflowSpec(gpuworkflow::CompletionPolicyData* poli
...
       } else {
         processAttributes->dEdxSplines.reset(new TPCdEdxCalibrationSplines);
       }
       config.configCalib.dEdxSplines = processAttributes->dEdxSplines.get();
 
+      if (!confParam.dEdxCorrFile.empty()) {
+        LOGP(info, "Loading dEdx correction file: {}", confParam.dEdxCorrFile);
+        processAttributes->dEdxCorrection.reset(new o2::tpc::CalibdEdxCorrection(confParam.dEdxCorrFile));
+      } else {
+        processAttributes->dEdxCorrection.reset(new o2::tpc::CalibdEdxCorrection());
```
wiechula, line 268: We should check with David if we cannot just leave this as a `nullptr`

wiechula (changes requested):
@badarots , very nice! Just a few minor things.

wiechula:
@davidrohr , could you have a look at the modifications in the GPU classes and the classes where we use the `GPUd()` macros if this looks fine.
Note, that in the fillCluster we now pass the full calibration object. Is this fine? For the splines you suggested to always have valid pointers, is there a special reason? Should be do the same for the residual correction?

DataFormats/Detectors/TPC/include/DataFormatsTPC/CalibdEdxCorrection.h
```diff
@@ -0,0 +1,96 @@
...
+#endif
+
+ private:
+  GPUd() static size_t stackIndex(const StackID& stack, ChargeType charge)
+  {
+    return static_cast<size_t>(stack.index() + charge * SECTORSPERSIDE * SIDES * GEMSTACKSPERSECTOR);
+  }
+
+  std::array<Params, 288> mParams{};
+  std::array<float, 288> mChi2{};
```
davidrohr: Please don't use std::array for classes that shall run on the GPU.
Either just use C arrays, or you can use GPUCommonArray.h

DataFormats/Detectors/TPC/include/DataFormatsTPC/CalibdEdxCorrection.h
```diff
@@ -0,0 +1,96 @@
...
+// or submit itself to any jurisdiction.
+
+/// \file CalibdEdxCorrection.h
+/// \author Thiago Badaró <thiago.saramela@usp.br>
+
+#ifndef ALICEO2_TPC_CALIBDEDXCORRECTION_H_
+#define ALICEO2_TPC_CALIBDEDXCORRECTION_H_
+
+#include "GPUCommonDef.h"
+#ifndef GPUCA_ALIGPUCODE
```
davidrohr: For hiding statndard c++ headers, you can use `#ifndef GPUCA_GPUCODE_DEVICE`

GPU/Workflow/src/GPUWorkflowSpec.cxx
davidrohr: why would you leave this as nullptr. From what I have seen above, you actually check for the presense of the object and you would fail the dEdx computation if it is not present. I think having a default object here that doesn't apply any correction makes sense.
(The alternative would be to check inside the fillCluster function for the presense of the object and use it only if not nullptr).

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -532,7 +543,8 @@ DataProcessorSpec getGPURecoWorkflowSpec(gpuworkflow::CompletionPolicyData* poli
               if (verbosity) {
                 end = std::chrono::high_resolution_clock::now();
                 std::chrono::duration<double> elapsed_seconds = end - start;
-                LOG(INFO) << "Allocation time for " << name << " (" << size << " bytes)" << ": " << elapsed_seconds.count() << "s";
+                LOG(INFO) << "Allocation time for " << name << " (" << size << " bytes)"
+                          << ": " << elapsed_seconds.count() << "s";
```
davidrohr: Please don't break my lines :)

davidrohr (dismissed):
Haven't tested it, but looks all good except for the `std::array`. Let's see what the GPU CI says.

Detectors/TPC/calibration/src/CalibdEdx.cxx
badarots: Sure, I should set a name and title, but the axis will be tricky to set because I'm doing a 2D unrolling of the multi D hist and I also can apply projections to it.
I am thinking about removing it, to be honest. I keep this method so I can plot 2D hists easily, but since now I can dump the hist content to a tree maybe it is not useful anymore.

DataFormats/Detectors/TPC/include/DataFormatsTPC/CalibdEdxCorrection.h
wiechula: Ok, however I ran this code on the GPU, NVIDIA and AMD. And it seems to work. But we'll change it.

GPU/Workflow/src/GPUWorkflowSpec.cxx
wiechula: I think this was clang-format.

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
badarots: So, if a value is passed to the command line we don't try to read the grp file, right? Its very nice. I thought that we would get a nullptr if no grp file was found. I will change this.

Detectors/TPC/calibration/src/CalibdEdx.cxx
wiechula: I think you can keep it. But just give it some name, e.g. "hdEdxProjection" or so.

DataFormats/Detectors/TPC/include/DataFormatsTPC/CalibdEdxCorrection.h
davidrohr: it would fail with OpenCL

GPU/Workflow/src/GPUWorkflowSpec.cxx
badarots: haha, sorry. My auto format keeps breaking, and I keep forgetting to fix it.

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
badarots: Like this
```C++
    float field = ic.options().get<float>("field");

    if (field <= -100) {
      const auto inputGRP = o2::base::NameConf::getGRPFileName();
      const auto grp = o2::parameters::GRPObject::loadFrom(inputGRP);
      if (grp != nullptr) {
        field = 5.00668f * grp->getL3Current() / 30000.;
        LOGP(info, "Using GRP file to set the magnetic field to {} kG", field);
      }
    }
```

wiechula (dismissed):

badarots:
The error happened during tests:
Test #294: Detectors/TPC/reconstruction/test/testGPUCATracking.cxx

The message was:
[ERROR] Cannot run dE/dx without calibration splines or time correction object

It occurred because no calibration file was supplied.

wiechula (approved):

badarots:
Some complainings about compiler_commands.json. Not sure why.

```
++ perl -p -i -e 's|/sw/SOURCES/O2Physics/7099-slc7_x86-64/0||' compile_commands.json
++ ln -sf /sw/BUILD/75741165bb0853846a221410c7f34d9edcfdec4f/O2Physics/compile_commands.json /compile_commands.json
ln: failed to create symbolic link '/compile_commands.json': Permission denied
```

wiechula:
@davidrohr , not sure what the problem with the CIs is. The only thing I find is `ln: failed to create symbolic link '/compile_commands.json': Permission denied`, as written by @badarots . Do you have an idea?

davidrohr:
Issues should be fixed meanwhile, when the CI reprocesses the PR, it should become OK.

davidrohr (approved):
Looks good.
Could you provide me with some up to date calibration objection I can use to test it?

badarots:
> Looks good. Could you provide me with some up to date calibration objection I can use to test it?

Of course, here is one: https://cernbox.cern.ch/index.php/s/ylQVmOUDqKY1wk2

davidrohr:
thx, do you also have an up to date splines file? :)

badarots:
Here!! https://cernbox.cern.ch/index.php/s/s6ebTJUU94DZ1Rz

wiechula:
@davidrohr can this be merged?

PR #7101, opened 2021-09-23
[TRD] Adapt to 3-level CCDB naming scheme

martenole:
- TRD TRAP simulator can use different TRAP configs, was hardcoded before
- minor cleanup

### DIFF
--- a/Detectors/TRD/base/macros/OCDB2CCDB.C
+++ b/Detectors/TRD/base/macros/OCDB2CCDB.C
@@ -213,8 +213,9 @@ void UnpackGainTable(std::string& gainkey, CalOnlineGainTables* gtbl)
   }
 }
 
+// choose ccdbPath = "http://ccdb-test.cern.ch:8080" to write to the test CCDB at CERN
 //__________________________________________________________________________________________
-void OCDB2CCDB(Int_t run, const Char_t* storageURI = "alien://folder=/alice/data/2010/OCDB/")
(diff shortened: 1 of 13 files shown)


bazinski (dismissed):

martenole:
Hi @tdietel would it be OK for you to merge this?

Detectors/TRD/base/macros/OCDB2CCDB.C
```diff
@@ -214,7 +214,7 @@ void UnpackGainTable(std::string& gainkey, CalOnlineGainTables* gtbl)
 }
 
 //__________________________________________________________________________________________
-void OCDB2CCDB(Int_t run, const Char_t* storageURI = "alien://folder=/alice/data/2010/OCDB/")
+void OCDB2CCDB(Bool_t writeToLocalCCDB = kTRUE, Int_t run = 297595, const Char_t* storageURI = "local:///cvmfs/alice-ocdb.cern.ch/calibration/data/2018/OCDB")
```
tdietel: Wouldn't it be more flexible to pass the ccdbPath (which is actually a URI) instead of a boolean variable? The dafault could be the test CCDB.

tdietel (dismissed):
Looks ok. Can be merged, but consider my minor comment.

Detectors/TRD/base/macros/OCDB2CCDB.C
martenole: Sure, I changed this now to take the CCDB path directly as parameter

martenole:
Hi @shahor02 could this be merged when you get a chance? The dataflow-cs8 error is a timeout issue. Yesterday all checks were green.

PR #7102, opened 2021-09-23
Fix buffer size in get_timestamp

chiarazampolli:
to avoid memory corruption.
@mpoghos , FYI.
Kudos to @ktf .


shahor02 (approved):
merging since the fix is trivial.

PR #7103, opened 2021-09-23
DPL: flush more metrics to avoid race condition in analysis

ktf:


ktf:
Tested on hyperloop. Fixes the stuck workflows there, including nice speedup.

PR #7104, opened 2021-09-23
DPL: extend ccdbParamSpec to allow run dependent queries

ktf:


ktf:
@shahor02 @chiarazampolli this roughly implements what we discussed today. The GET will be extended with the `/runNumber=<number>` if `ccdbParamSpec()` gets passed a true as second argument. I did some preliminary tests and it seems to work as expected.

PR #7105, opened 2021-09-23
Move lambda to template helper to fix XCode13 compile problem

sawenzel:
omp critical section inside lambda makes XCode13 fail.
Therefore moving lambda to template helper to work around.


ktf (approved):
Works fine with my XCode 13 setup. Thanks.

PR #7107, opened 2021-09-23
MCH: implemented digit time computation based on HBPackets

aferrero2707:


Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
```diff
@@ -418,23 +462,30 @@ void DataDecoder::decodePage(gsl::span<const std::byte> page)
...
-           bunchCrossingCounter, mOrbit,
-           bunchCrossingCounter % 3564);
+      LOGP(info, "HeartBeat: {}-CHIP{} -> {}/{} offset {} -> bcTF {}",
+           s, chip, mFirstOrbitInTF, bunchCrossing, mSampaTimeOffset, bcTF);
+    }
+
+    if (!mTimeFrameStartRecords[chipId].check()) {
+      if (mErrorCount < MCH_DECODER_MAX_ERROR_COUNT) {
+        auto s = asString(dsElecId);
+        LOGP(warning, "Inconsistent HeartBeat packet received: {}-CHIP{} {}/{} (last {}/{})",
```
aphecetche, line 483: I guess this is fine for now, but cutting out messages might be dangerous in the long run. A better option (even if it'll cost some CPU) would be to store messages, counting their occurrences, and showing that store once in a while. (might require to think a bit in order not to generate too different messages for similar errors, otherwise the message store would blow up the memory...)

Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
```diff
@@ -492,75 +543,109 @@ void DataDecoder::decodePage(gsl::span<const std::byte> page)
...
+  // SAMPA BC at the beginning of the current TF
+  bcTF = tfStart.mBunchCrossing;
+
+  if (orbitHBP != orbitTF) {
+    // we correct the BC from the last received HB packet, if it was recorded from an older TF
+    bcTF += (orbitTF - orbitHBP) * bcInOrbit;
+    // only keep 20 bits
+    bcTF &= twentyBitsAtOne;
+
+    //std::cout << fmt::format(" -> {}/{}", orbitTF, bcTF) << std::endl;
```
aphecetche, line 609: commented out parts should be removed

Detectors/MUON/MCH/Raw/Decoder/src/ROFFinder.cxx
```diff
@@ -145,10 +145,12 @@ void ROFFinder::sortDigits()
   for (size_t i = 0; i < mInputDigits.size(); i++) {
     auto& digit = mInputDigits[i];
     if (!digit.timeValid()) {
+#ifdef ROFDEBUG
       LOG(ERROR) << "Digit with invalid time, DS " << digit.info.solar << "," << digit.info.ds << "," << digit.info.chip
```
aphecetche, line 149: note that FairLogger is deprecating the upper case version of the log levels, so from now on, we should use the lower case versions when we think about it

aphecetche (approved):

aphecetche:
@kft @davidrohr any idea what's wrong with the fullCI ?

PR #7109, opened 2021-09-23
ConfigParam - Ccdb-Api integration

sawenzel:
Update ConfigurableParams when they are fetched from
CCDB via CcdbApi. This guarantees the following:

a) ConfigParams are updated with the CCDB instance

b) However, members overriden by the user (kRT provenance) are not overriden
   by CCDB values (kRT takes higher priority)

### DIFF
--- a/CCDB/CMakeLists.txt
+++ b/CCDB/CMakeLists.txt
@@ -53,6 +53,13 @@ o2_add_test(CcdbApi
             PUBLIC_LINK_LIBRARIES O2::CCDB
             LABELS ccdb)
 
+o2_add_test(CcdbApiConfigParam
+		SOURCES test/testCcdbApi_ConfigParam.cxx
+		COMPONENT_NAME ccdb
+		PUBLIC_LINK_LIBRARIES O2::CCDB O2::DetectorsVertexing
+		LABELS ccdb)
+
+
 o2_add_test(CcdbApi-Alien
 		SOURCES test/testCcdbApi_alien.cxx
 		COMPONENT_NAME ccdb
--- a/CCDB/include/CCDB/CcdbApi.h
+++ b/CCDB/include/CCDB/CcdbApi.h
@@ -24,6 +24,8 @@
 #include <TObject.h>
 #include <TMessage.h>
 #include "CCDB/CcdbObjectInfo.h"
+#include <CommonUtils/ConfigurableParam.h>
+#include <type_traits>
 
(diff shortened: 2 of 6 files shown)


sawenzel:
Implements a feature requested by @shahor02

shahor02:
Hi @sawenzel 
Thanks! I had a chat with @ktf, in the framework CCDB service he does not use CcdbApi but fetches the filed directly with curl: https://github.com/AliceO2Group/AliceO2/blob/3206b5602f6f8a02d189a099172d177e22bc781b/Framework/Core/src/LifetimeHelpers.cxx#L259-L356 and extracts the object with https://github.com/AliceO2Group/AliceO2/blob/3206b5602f6f8a02d189a099172d177e22bc781b/Framework/Core/src/DataRefUtils.cxx#L36
The user accesses the object via DataRefUtils: 
https://github.com/AliceO2Group/AliceO2/blob/3206b5602f6f8a02d189a099172d177e22bc781b/Framework/Core/include/Framework/DataRefUtils.h#L157
I guess at this place a hook handling Configurable params should be added?

jgrosseo:
> I had a chat with @ktf, in the framework CCDB service he does not use CcdbApi but fetches the filed directly with curl

I think we cannot do it this way - at least for analysis. Unless you also re-implement the whole alien access with certificate. CCDB can also sit in AliEn...

shahor02:
> > I had a chat with @ktf, in the framework CCDB service he does not use CcdbApi but fetches the filed directly with curl
> 
> I think we cannot do it this way - at least for analysis. Unless you also re-implement the whole alien access with certificate. CCDB can also sit in AliEn...

Here @ktf should comment, I don't know what was the reason behind using the curl directly

ktf:
Doing my own curl was meant as a stop gap solution until I made up my mind on how:

* Allow the current CCDBApi to use curl_multi* API so that we can do parallel fetching of objects without extra threads.
* Allow the current CCDBApi to handle the case in which multiple timeframes with different conditions are in flight at the same time in different threads.

It is not meant to be a final solution, but so far what we have is how much I managed to integrate the CCDBApi in the DPL processing model.

shahor02 (approved):
>     * Allow the current CCDBApi to handle the case in which multiple timeframes with different conditions are in flight at the same time in different threads.

Hm, this would make usage of the ConfigParam problematic, I thought the DPL device can see just 1 ProcessingContext at a time.
Anyway, this deserves a separate discussion, merging this.
Thanks, @sawenzel !

PR #7114, opened 2021-09-24
Adding WF for calibration based on diagnostic words

chiarazampolli:
This is the workflow to propagate to the CCDB the diagnostic words from decoding. 

Example usage:

`o2-calibration-data-generator-diagnostic-workflow --max-timeframes 10 | o2-calibration-tof-diagnostic-workflow  | o2-calibration-ccdb-populator-workflow --ccdb-path="http://localhost:8080"`

### DIFF
--- a/DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
+++ b/DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
@@ -17,6 +17,7 @@
 
 #include <map>
 #include <TObject.h>
+#include <gsl/gsl>
 
 namespace o2
 {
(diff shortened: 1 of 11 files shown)


chiarazampolli:
Hi @noferini , 

The CCDB entries that the workflow adds are valid till infinity to always have an entry that is valid. I have a doubt about this which we should check, namely whether the time slots are closed for sure in increasing order. If this is the case, then it should be not an issue. 

Chiara

noferini:
> Hi @noferini ,
> 
> The CCDB entries that the workflow adds are valid till infinity to always have an entry that is valid. I have a doubt about this which we should check, namely whether the time slots are closed for sure in increasing order. If this is the case, then it should be not an issue.
> 
> Chiara

In principle it shouldn't be a problem.
The information will be used only in simulation.
Let see, in case we will limit the validity

shahor02 (approved):
This looks fine for me, @noferini shall I wait for your review?

noferini (approved):
It looks fine

PR #7117, opened 2021-09-24
Rewrite TreeToTable to be more generic

aalkin:
@ktf second part, I've found a way to avoid template here as well.


ktf:
Tested on hyperloop. Works fine with the "O2 Development" test with no regressions on the performance / memory usage.

PR #7118, opened 2021-09-24
[MFT] Use unique volume names

benedikt-voelkel:
* MFT half cones seem to be only volumes in geo tree with non-unique
  names

* fixes potential issues related to finding correct volume by name or
  geo path


rpezzi (approved):

shahor02 (approved):
fullCI failed because of my wrong commit (already fixed), merging as the change is trivial

PR #7119, opened 2021-09-24
CPV documentation

sevdokim:
Hello!

Some documentation is added to Detectors/CPV/README.md


sevdokim:
Dear @shahor02 , it seems that PR is trivial. I also ran "Build documentation" workflow in my fork, and build was successful. So maybe there is no need to run building tests?
https://github.com/sevdokim/AliceO2/runs/3703097289?check_suite_focus=true

shahor02 (approved):
indeed, merging as trivial.

PR #7121, opened 2021-09-24
Implement CTP CTF and add CTP to FST

shahor02:
Hi @lietava 

CTP decoding was producing more logs than all other detectors together, so I've changed a lot of INFOs to DEBUG in your code.
Also, had to touch your digits class.

The CTP digit2row still has this custom option `--file-per-link` instead of  `--file-for [all,link,cru]`, could you please fix this? The output naming is ok.

Then, your raw->digits decoder is for some reason named `o2-ctp-reco-workflow`, Could you please rename it to something more appropriate, e.g. `o2-ctp-raw-decoder`. More serious is that it declares a lot of options which it is not supposed to support, I guess it comes from some copy/paste.
```
  --input-type arg (=raw)               digits, raw
  --output-type arg (=digits)           digits, raw
  --disable-mc                          disable sending of MC information
  --disable-root-input                  disable root-files input reader
  --disable-root-output                 disable root-files output writer
```
Can you have input other than raw and output other than digits?
Raw decoders are not supposed to support MC, so you don't need this option at all.
And for the raw decoders there is no root input anyway.

### DIFF
--- a/DataFormats/Detectors/CTP/CMakeLists.txt
+++ b/DataFormats/Detectors/CTP/CMakeLists.txt
@@ -11,15 +11,16 @@
 
 o2_add_library(DataFormatsCTP
   SOURCES src/Digits.cxx
-  SOURCES src/Configuration.cxx
-  SOURCES src/Scalers.cxx
+          src/Configuration.cxx
+          src/Scalers.cxx
+          src/CTF.cxx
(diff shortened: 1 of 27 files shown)


lietava:
Hi @shahor02 , thanks. I'll fix it latest after apw

PR #7126, opened 2021-09-25
Fit dev update

afurs:
1. New option for defining CCDB and LUT path in raw2digit/digit2raw
2. Some pre-allocation options for raw2digit
3. New feature, only headers(no data) from PM


afurs:
In CheckPR: `error: failed to push some refs to 'https://github.com/alibuild/AliceO2.git'`

PR #7127, opened 2021-09-25
FIT: LUT url/path option, pre-allocation option, some update in RawEventData

afurs:
1. New option for defining CCDB and LUT path in raw2digit/digit2raw
2. Some pre-allocation options for raw2digit
3. New feature, only headers(no data) from PM

### DIFF
--- a/DataFormats/Detectors/FIT/FDD/include/DataFormatsFDD/LookUpTable.h
+++ b/DataFormats/Detectors/FIT/FDD/include/DataFormatsFDD/LookUpTable.h
@@ -215,7 +215,7 @@ class SingleLUT : public LUT
  public:
   static constexpr char sDetectorName[] = "FDD";
   static constexpr char sDefaultCCDBpath[] = "http://ccdb-test.cern.ch:8080/";
-  static constexpr char sDefaultLUTpath[] = "FV0/LookUpTable";
+  static constexpr char sDefaultLUTpath[] = "FDD/LookUpTable";
   inline static std::string sCurrentCCDBpath = sDefaultCCDBpath;
   inline static std::string sCurrentLUTpath = sDefaultLUTpath;
   //Before instance() call, setup url and path
--- a/DataFormats/Detectors/FIT/FDD/include/DataFormatsFDD/LookUpTable.h
+++ b/DataFormats/Detectors/FIT/FDD/include/DataFormatsFDD/LookUpTable.h
@@ -214,21 +214,16 @@ class SingleLUT : public LUT
 
  public:
   static constexpr char sDetectorName[] = "FDD";
+  static constexpr char sDefaultCCDBpath[] = "http://ccdb-test.cern.ch:8080/";
+  static constexpr char sDefaultLUTpath[] = "FV0/LookUpTable";
+  inline static std::string sCurrentCCDBpath = sDefaultCCDBpath;
+  inline static std::string sCurrentLUTpath = sDefaultLUTpath;
+  //Before instance() call, setup url and path
+  static void setCCDBurl(const std::string& url) { sCurrentCCDBpath = url; }
+  static void setLUTpath(const std::string& path) { sCurrentLUTpath = path; }
   static SingleLUT& Instance()
   {
-    static SingleLUT instanceLUT("http://ccdb-test.cern.ch:8080/", "FDD/LookUpTable");
-    return instanceLUT;
-  }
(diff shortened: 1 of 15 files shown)


AllaMaevskaya (commented):
sDefaultCCDBpath should be "http://o2-ccdb.internal/" and could be changed as argument

afurs:
> sDefaultCCDBpath should be "http://o2-ccdb.internal/" and could be changed as argument

Hello @AllaMaevskaya , yes I already added such option to raw2digit and digit2raw (`--ccdb-path` and `--lut-path`) . Adding this urls by default is not good idea, cause this url unable from outside of CERN. I will change default values later.

afurs:
And it seems that for FDD and FV0 one need to request/register LUT path within CCDB. For FT0 it is done(I mentioned path when we had discussion by email concerning CCDB, and it seems that path was registered)

afurs:
@shahor02 hello, could you take a look to this PR?

DataFormats/Detectors/FIT/FDD/include/DataFormatsFDD/LookUpTable.h
```diff
@@ -214,21 +214,16 @@ class SingleLUT : public LUT
 
  public:
   static constexpr char sDetectorName[] = "FDD";
+  static constexpr char sDefaultCCDBpath[] = "http://ccdb-test.cern.ch:8080/";
+  static constexpr char sDefaultLUTpath[] = "FDD/LookUpTable";
```
shahor02, line 218: We are switching to 3-level CCDB directories, perhaps you can directly change them to e.g. `FDD/Config/LookUpTable` ?

DataFormats/Detectors/FIT/FDD/include/DataFormatsFDD/LookUpTable.h
```diff
@@ -214,21 +214,16 @@ class SingleLUT : public LUT
...
+  inline static std::string sCurrentCCDBpath = sDefaultCCDBpath;
+  inline static std::string sCurrentLUTpath = sDefaultLUTpath;
```
shahor02, line 220: since your lookup table is constant within the run, I can understand why you make it singleton, hence static. 
But why do you need to have these paths as static? They are used only once at the Instance initialization, cannot you have theme simply as arguments of the Instance, defaulted to the values you need?

DataFormats/Detectors/FIT/FDD/include/DataFormatsFDD/LookUpTable.h
afurs: This is old CCDB directory, but target `FDD/Config/LookUpTable` is configurable via `--lut-path FDD/Config/LookUpTable --ccdb-path http://o2-ccdb.internal/` for raw2digit and digit2raw. I can upload new 3-level dir into `http://ccdb-test.cern.ch:8080/` (`http://o2-ccdb.internal/` unfortunately is unable for me right now, I'm discussing this situation with Chiara)

DataFormats/Detectors/FIT/FDD/include/DataFormatsFDD/LookUpTable.h
afurs: I decided to keep `Instance()` without any arguments( `Instance(const std::string="default value" ...)` will call copy operation any time when LUT is needed ) and configuration via static fields in https://github.com/AliceO2Group/AliceO2/blob/4de04de161bd24ba8683c285f37bb4bc2d4917fd/Detectors/FIT/workflow/include/FITWorkflow/FITDataReaderDPLSpec.h#L52 looks reasonable. Should I change this?

shahor02: Why? If you define it as `Instance(const std::string& str = "default value" ...)` or `Instance(const std::string& str =  sDefaultCCDBpath)` then no copy will be involved, it will simply pass the reference on const.
I would prefer to minimize the number of statics. 
Anyway, if this PR is urgent, I can merge it is as it is, the cleanup can be done in other PR

afurs: Oh, sure. Strange, I saw problems when I tried to use refs with defaults, thats why I decided to exclude default args... Could you merge this PR I will prepare cleanup in the next?

PR #7129, opened 2021-09-26
Replace --dds, rather than --odc

ktf:
The option is still named --dds.


ktf:
Currently broken and not checked in any case, since this is breaking production.

PR #7132, opened 2021-09-27
TPC: fixes und updates in laser drift velocity calibration

wiechula:


Detectors/TPC/workflow/src/tpc-krypton-clusterer.cxx
```diff
@@ -76,7 +77,8 @@ WorkflowSpec defineDataProcessing(ConfigContext const& config)
   const auto tpcSectors = o2::RangeTokenizer::tokenize<int>(config.options().get<std::string>("sectors"));
   const auto nSectors = (int)tpcSectors.size();
   const auto nLanes = std::min(config.options().get<int>("lanes"), nSectors);
-  const auto disableWriter = config.options().get<bool>("disable-writer");
+  const auto disableWriter = config.options().get<bool>("disableWriter");
+  const auto autoSaveAfter = config.options().get<int>("autoSaveAfter");
```
shahor02: note that the `--autosave` is generic option of the root tree writer (default -1), `-h full` on any workflow containing RootTreeWriterSpec should show it.
With your `--autoSaveAfter` you are basically redefining its default shown in the help (you could simply hardcode it), which then can be changed with `--autosave`.

wiechula:
@shahor02 , thanks for the info. I didn't know that. I'll remove the commit then and push again.

wiechula:
@shahor02 , I think the failing test is unrelated. Could you merge?

PR #7135, opened 2021-09-27
raw input optional in tof compressor

noferini:
Urgent: this is required to fix an issue in the tof chain because of recent changes

### DIFF
--- a/Detectors/TOF/compression/src/tof-compressor.cxx
+++ b/Detectors/TOF/compression/src/tof-compressor.cxx
@@ -100,6 +100,9 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   while (getline(ssconfig, iconfig, ',')) {
     std::vector<InputSpec> inputs = select(iconfig.c_str());
     if (!ignoreStf) {
+      for (auto& val : inputs) {
+        val.lifetime = Lifetime::Optional;
+      }
       inputs.emplace_back("stdDist", "FLP", "DISTSUBTIMEFRAME", 0, Lifetime::Timeframe);
     }
     workflow.emplace_back(DataProcessorSpec{


noferini:
@ktf @preghenella

shahor02:
@noferini not sure it is enough: now in case of missing TOF/RAWDATA on all links of the FLP, your compressor will get empty `TOF/RAWDATA/0xdeadbeaf` message. You have to check for this message in the same way as are doing in the decompressor, and make sure that either valid dummy CRAWDATA output is sent (then the 0xdeadbeaf will not be triggered in the decompressor) or that you send nothing (then the 0xdeadbeaf will be triggered in the decompressor). 
But your compressor should be able to digest such a dummy message. Perhaps you should try with raw-file-reader using `--drop-tf` option.

Now I remember why you made the CRAWDATA to be optional rather than the upstream RAWDATA: if the TF data on one of the FLPs is completely missing, your compressor on this FLP would receive TOF/RAWDATA/0xdeadbeaf dummy message. If both FLPs lose the data, they would both send the message with the same signature. Perhaps the best if you send nothing in this case from the compressor, triggering another 0xdeadbeaf on the decompressor side).

noferini:
> @noferini not sure it is enough: now in case of missing TOF/RAWDATA on all links of the FLP, your compressor will get empty `TOF/RAWDATA/0xdeadbeaf` message. You have to check for this message in the same way as are doing in the decompressor, and make sure that either valid dummy CRAWDATA output is sent (then the 0xdeadbeaf will not be triggered in the decompressor) or that you send nothing (then the 0xdeadbeaf will be triggered in the decompressor).
> But your compressor should be able to digest such a dummy message. Perhaps you should try with raw-file-reader using `--drop-tf` option.
> 
> Now I remember why you made the CRAWDATA to be optional rather than the upstream RAWDATA: if the TF data on one of the FLPs is completely missing, your compressor on this FLP would receive TOF/RAWDATA/0xdeadbeaf dummy message. If both FLPs lose the data, they would both send the message with the same signature. Perhaps the best if you send nothing in this case from the compressor, triggering another 0xdeadbeaf on the decompressor side).

Hi @shahor02 , the problem here is the following.
Compressor sends forward only what it received (RDH -> compressor -> RDH).
If no RDH reaches compressor there is no output.
I would not send anything in such a case. I would expect that a deadbeef will be recognized in the decoder (as for other detectors). Am I wrong?

shahor02:
> I would not send anything in such a case. I would expect that a deadbeef will be recognized in the decoder (as for other detectors).

On order to generate a `deadbeaf`, the decoder needs to receive something, that's why you subscribe it to  `DISTSUBTIMEFRAME`. Hopefully, you will receive it because other detectors send something (and if the TOF alone, then we don't mind if the TF processing is abandoned). But I think it is worth trying how the chain with this double "optional" behaves when TF is absent.
Actually, I realize that we may have problems also with input proxy, not sure how it behaves if one requested inputs are absent, need to test.

noferini:
> > I would not send anything in such a case. I would expect that a deadbeef will be recognized in the decoder (as for other detectors).
> 
> On order to generate a `deadbeaf`, the decoder needs to receive something, that's why you subscribe it to `DISTSUBTIMEFRAME`. Hopefully, you will receive it because other detectors send something (and if the TOF alone, then we don't mind if the TF processing is abandoned). But I think it is worth trying how the chain with this double "optional" behaves when TF is absent.
> Actually, I realize that we may have problems also with input proxy, not sure how it behaves if one requested inputs are absent, need to test.

Hi ruben,
I am a bit confused. 
If the compressor receives DISTSUBTIMEFRAME, DISTSUBTIMEFRAME will be also propagated forward to the decoder on epn independently of other detectors.
I have understood that the processing of deadbeef was required only in case a specific output is required in the next steps.
Since the CRAWDATA are optional in the next step (decoder) I don't understand why I need to digest deadbeef in tof compressor.

shahor02:
@noferini you do subscribe to DISTSUBTIMEFRAME in the compressor, right? In absence of the TF this will trigger on the compressor generation of TOF/RAWDATA/0xdeadbeef. Do you know how the compressor will behave?

noferini:
> @noferini you do subscribe to DISTSUBTIMEFRAME in the compressor, right? In absence of the TF this will trigger on the compressor generation of TOF/RAWDATA/0xdeadbeef. Do you know how the compressor will behave?

Hi Ruben,
it seems to me (@preghenella can you confirm?) that a 0 payload message is sent out.

This is the last call in run method of Detectors/TOF/compression/src/CompressorTask.cxx
 /** send message **/
  device->Send(partsOut, fairMQChannel);

So I would expect no difference with respect to what happens in case of payload > 0.

shahor02:
OK, btw, you can run with current O2 w/o waiting for this PR to pass the CI, just pass `--ignore-dist-stf` to `o2-tof-reco-workflow`, then it will not subscribe to DISTSUBTIMEFRAME and the premature 0xdeadbeef will not be triggered.

shahor02 (approved):

noferini:
@shahor02 
I did some checks by disabling processing input in compressors.
It works if the output 0-payload sending is disable otherwise it gives problem.
I will spoke with @preghenella to manage such a case.

shahor02:
Ok, you can always disable this deadbeef by option --ignore-dist-stf

PR #7136, opened 2021-09-27
[EMCAL-534] Add handling of non-fatal decoding errors

mfasDa:
Non-fatal errors only skip parts of the decoding
(i.e. channel) as part of the payload can be
decoded, but do not skip the entire page
- Add container for minor errors in AltroDecoder
- Add error states for missing bunch header,
  unexpected end of payload, and payload out-
  of-bounds
- Add category for minor errors in ErrorTypeFee
- Add handling of minor errors in raw-to-cell
  converter


mfasDa:
@shahor02 This is connected to the crashes we observed over the weekend while decoding on the EPN. We were able to connect them to channels sending faulty bunch headers. As they don't affect the full payload for the given links but only certain towers we treat them as minor errors that do not raise skipping the page (in contrast to errors i.e. in the RCU trailer leading to corrupted pages). Towers in this case will be skipped and errors will be raised on the infoBrowser + displayed in the QC.

shahor02 (dismissed):

mfasDa:
Making CI happy ...

shahor02 (approved):

PR #7139, opened 2021-09-27
fix for ITSTPC matching timers/counters

shahor02:


shahor02:
Failure is unrelated, merging

PR #7141, opened 2021-09-28
DPL: have resource accounting for analysis work with 200MB per reader

ktf:


ktf:
Tested on hyperloop. Merging.

PR #7142, opened 2021-09-28
flag also TOF noisy channel in diagnostic

noferini:


noferini:
Hi @chiarazampolli,
I added also noisy channels flag to diagnostic.
This will guarantee that we will discover them also in the sync phase in case we miss to switch off some of them in the slow control

DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
```diff
@@ -36,9 +36,11 @@ class Diagnostic
   int getFrequency(ULong64_t pattern);                                                    // Get frequency
   int getFrequencyROW() { return getFrequency(0); }                                       // Readout window frequency
   int getFrequencyEmptyCrate(int crate) { return getFrequency(getEmptyCrateKey(crate)); } // empty crate frequency
+  int fillNoisy(int channel, int frequency = 1) { return fill(getNoisyChannel(channel), frequency); }
   int fillROW() { return fill(0); }
   int fillEmptyCrate(int crate, int frequency = 1) { return fill(getEmptyCrateKey(crate), frequency); }
   ULong64_t getEmptyCrateKey(int crate);
+  ULong64_t getNoisyChannel(int channel);
```
shahor02: isn't this const method?

noferini: Yes, I will fix it (the same for getEmptyrateKey)
Even better -> static

chiarazampolli:
Thanks, it should be transparent for the diagnostic calibration, unless I overlooked something.

noferini:
> Thanks, it should be transparent for the diagnostic calibration, unless I overlooked something.

Yes, it is transparent

noferini:
all checks passed!

shahor02 (approved):

PR #7145, opened 2021-09-28
Correctly convert relative vertex times into global BC information

sawenzel:
Using the orbit information from digitization to convert
relative vertex times into global BC information.

Fixes https://alice.its.cern.ch/jira/browse/O2-2610


sawenzel:
Like to have in nightlies. Merging.

PR #7147, opened 2021-09-28
Avoid overallocations

ktf:


ktf:
Seems to improve both performance and memory usage. Tested on hyperloop and my mac. Merging to get it in the build.

PR #7148, opened 2021-09-28
Implementation of parsing of LHC-IF file

chiarazampolli:
Parser

Workflow and data type

space checker

clang-format

### DIFF
--- a/DataFormats/Parameters/CMakeLists.txt
+++ b/DataFormats/Parameters/CMakeLists.txt
@@ -10,13 +10,15 @@
 # or submit itself to any jurisdiction.
 
 o2_add_library(DataFormatsParameters
-               SOURCES src/GRPObject.cxx
+  SOURCES src/GRPObject.cxx
+  src/LHCIFData.cxx
                PUBLIC_LINK_LIBRARIES FairRoot::Base O2::CommonConstants
                                      O2::CommonTypes
                                      O2::DetectorsCommonDataFormats)
 
 o2_target_root_dictionary(DataFormatsParameters
-                          HEADERS include/DataFormatsParameters/GRPObject.h
+  HEADERS include/DataFormatsParameters/GRPObject.h
+  include/DataFormatsParameters/LHCIFData.h
                           LINKDEF src/ParametersDataLinkDef.h)
 # note we are explicitely giving the LINKDEF parameter as the LinkDef does not
 # follow the usual naming scheme [module]LinkDef.h
--- a/DataFormats/Parameters/CMakeLists.txt
+++ b/DataFormats/Parameters/CMakeLists.txt
@@ -10,16 +10,18 @@
 # or submit itself to any jurisdiction.
 
 o2_add_library(DataFormatsParameters
-  SOURCES src/GRPObject.cxx
-  src/LHCIFData.cxx
+               SOURCES src/GRPObject.cxx
+                       src/LHCIFData.cxx
                PUBLIC_LINK_LIBRARIES FairRoot::Base O2::CommonConstants
                                      O2::CommonTypes
                                      O2::DetectorsCommonDataFormats)
 
 o2_target_root_dictionary(DataFormatsParameters
-  HEADERS include/DataFormatsParameters/GRPObject.h
-  include/DataFormatsParameters/LHCIFData.h
+                          HEADERS include/DataFormatsParameters/GRPObject.h
+                                  include/DataFormatsParameters/LHCIFData.h
                           LINKDEF src/ParametersDataLinkDef.h)
+
+
 # note we are explicitely giving the LINKDEF parameter as the LinkDef does not
 # follow the usual naming scheme [module]LinkDef.h
 #
--- a/DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
+++ b/DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
@@ -30,11 +30,11 @@ class LHCIFData
   LHCIFData() = default;
   ~LHCIFData() = default;
 
-  std::pair<uint64_t, int32_t> getBeamEnergy() const { return mBeamEnergy; }
-  std::pair<uint64_t, int32_t> getFillNumber() const { return mFillNumber; }
-  std::pair<uint64_t, std::string> getInjectionScheme() const { return mInjectionScheme; }
-  std::pair<uint64_t, int32_t> getAtomicNumberB1() const { return mAtomicNumberB1; }
-  std::pair<uint64_t, int32_t> getAtomicNumberB2() const { return mAtomicNumberB2; }
+  std::pair<long, int32_t> getBeamEnergy() const { return mBeamEnergy; }
+  std::pair<long, int32_t> getFillNumber() const { return mFillNumber; }
+  std::pair<long, std::string> getInjectionScheme() const { return mInjectionScheme; }
+  std::pair<long, int32_t> getAtomicNumberB1() const { return mAtomicNumberB1; }
+  std::pair<long, int32_t> getAtomicNumberB2() const { return mAtomicNumberB2; }
 
   int32_t getBeamEnergyVal() const { return mBeamEnergy.second; }
   int32_t getFillNumberVal() const { return mFillNumber.second; }
@@ -42,30 +42,30 @@ class LHCIFData
   int32_t getAtomicNumberB1Val() const { return mAtomicNumberB1.second; }
   int32_t getAtomicNumberB2Val() const { return mAtomicNumberB2.second; }
 
-  uint64_t getBeamEnergyTime() const { return mBeamEnergy.first; }
-  uint64_t getFillNumberTime() const { return mFillNumber.first; }
-  uint64_t getInjectionSchemeTime() const { return mInjectionScheme.first; }
-  uint64_t getAtomicNumberB1Time() const { return mAtomicNumberB1.first; }
-  uint64_t getAtomicNumberB2Time() const { return mAtomicNumberB2.first; }
+  long getBeamEnergyTime() const { return mBeamEnergy.first; }
+  long getFillNumberTime() const { return mFillNumber.first; }
+  long getInjectionSchemeTime() const { return mInjectionScheme.first; }
+  long getAtomicNumberB1Time() const { return mAtomicNumberB1.first; }
+  long getAtomicNumberB2Time() const { return mAtomicNumberB2.first; }
 
-  void setBeamEnergy(std::pair<uint64_t, int32_t> p) { mBeamEnergy = p; }
-  void setFillNumber(std::pair<uint64_t, int32_t> p) { mFillNumber = p; }
-  void setInjectionScheme(std::pair<uint64_t, std::string> p) { mInjectionScheme = p; }
-  void setAtomicNumberB1(std::pair<uint64_t, int32_t> p) { mAtomicNumberB1 = p; }
-  void setAtomicNumberB2(std::pair<uint64_t, int32_t> p) { mAtomicNumberB2 = p; }
+  void setBeamEnergy(std::pair<long, int32_t> p) { mBeamEnergy = p; }
+  void setFillNumber(std::pair<long, int32_t> p) { mFillNumber = p; }
+  void setInjectionScheme(std::pair<long, std::string> p) { mInjectionScheme = p; }
+  void setAtomicNumberB1(std::pair<long, int32_t> p) { mAtomicNumberB1 = p; }
+  void setAtomicNumberB2(std::pair<long, int32_t> p) { mAtomicNumberB2 = p; }
 
-  void setBeamEnergy(uint64_t t, int32_t v) { mBeamEnergy = std::make_pair(t, v); }
-  void setFillNumber(uint64_t t, int32_t v) { mFillNumber = std::make_pair(t, v); }
-  void setInjectionScheme(uint64_t t, std::string v) { mInjectionScheme = std::make_pair(t, v); }
-  void setAtomicNumberB1(uint64_t t, int32_t v) { mAtomicNumberB1 = std::make_pair(t, v); }
-  void setAtomicNumberB2(uint64_t t, int32_t v) { mAtomicNumberB2 = std::make_pair(t, v); }
+  void setBeamEnergy(long t, int32_t v) { mBeamEnergy = std::make_pair(t, v); }
+  void setFillNumber(long t, int32_t v) { mFillNumber = std::make_pair(t, v); }
+  void setInjectionScheme(long t, std::string v) { mInjectionScheme = std::make_pair(t, v); }
+  void setAtomicNumberB1(long t, int32_t v) { mAtomicNumberB1 = std::make_pair(t, v); }
+  void setAtomicNumberB2(long t, int32_t v) { mAtomicNumberB2 = std::make_pair(t, v); }
 
  private:
-  std::pair<uint64_t, int32_t> mBeamEnergy;
-  std::pair<uint64_t, int32_t> mFillNumber;
-  std::pair<uint64_t, std::string> mInjectionScheme;
-  std::pair<uint64_t, int32_t> mAtomicNumberB1;
-  std::pair<uint64_t, int32_t> mAtomicNumberB2;
+  std::pair<long, int32_t> mBeamEnergy;
+  std::pair<long, int32_t> mFillNumber;
+  std::pair<long, std::string> mInjectionScheme;
+  std::pair<long, int32_t> mAtomicNumberB1;
+  std::pair<long, int32_t> mAtomicNumberB2;
 
   ClassDefNV(LHCIFData, 1);
 };
--- a/DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
+++ b/DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
@@ -32,38 +32,38 @@ class LHCIFData
 
   std::pair<float, int32_t> getBeamEnergy() const { return mBeamEnergy; }
   std::pair<float, std::string> getFillNumber() const { return mFillNumber; }
-  std::pair<float, std::string> getInjectionSchema() const { return mInjectionSchema; }
+  std::pair<float, std::string> getInjectionScheme() const { return mInjectionScheme; }
   std::pair<float, int32_t> getAtomicNumberB1() const { return mAtomicNumberB1; }
   std::pair<float, int32_t> getAtomicNumberB2() const { return mAtomicNumberB2; }
 
   int32_t getBeamEnergyVal() const { return mBeamEnergy.second; }
   std::string getFillNumberVal() const { return mFillNumber.second; }
-  std::string getInjectionSchemaVal() const { return mInjectionSchema.second; }
+  std::string getInjectionSchemeVal() const { return mInjectionScheme.second; }
   int32_t getAtomicNumberB1Val() const { return mAtomicNumberB1.second; }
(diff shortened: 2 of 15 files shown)


chiarazampolli:
Hello @shahor02 , 

This is the workflow to process the LHC IF file. There is one open point: the dcs-config-proxy expects that the name of the file starts with the three letters of the detector, but we don't have GRP yet. Should it be added? For the development I used "TOF", but it is clearly wrong. 

This was tested simulating the DCS server sending the file, as we did long ago when developing the dcs-config-proxy.

Chiara

shahor02:
> There is one open point: the dcs-config-proxy expects that the name of the file starts with the three letters of the detector, but we don't have GRP yet. Should it be added?

Hi @chiarazampolli Expanding DetID will create a mess in many places, we better patch the `dcs-config-proxy` in such a way, that when failing to recognize the detector from the 1st 3 letters, before throwing an error it will check a list of allowed extra names.

chiarazampolli:
Hi @shahor02 , 

Thanks for the comment and suggestion, I implemented GRP as an exception. 

Chiara

chiarazampolli:
Hello @shahor02 , @ktf , @TimoWilken 

The error in the CI is unrelated to this PR, I don't know why it appears. The same happened on this same PR the other day, but with a different error (then it changed because I added a commit).

Chiara

shahor02:
It is due to my bug which is already fixed

ktf:
I restarted the test.

chiarazampolli:
Thanks!

DataFormats/Parameters/CMakeLists.txt
```diff
@@ -10,13 +10,15 @@
 # or submit itself to any jurisdiction.
 
 o2_add_library(DataFormatsParameters
-               SOURCES src/GRPObject.cxx
+  SOURCES src/GRPObject.cxx
+  src/LHCIFData.cxx
```
shahor02: could the formatting be restored?

DataFormats/Parameters/CMakeLists.txt
```diff
@@ -10,13 +10,15 @@
...
                PUBLIC_LINK_LIBRARIES FairRoot::Base O2::CommonConstants
                                      O2::CommonTypes
                                      O2::DetectorsCommonDataFormats)
 
 o2_target_root_dictionary(DataFormatsParameters
-                          HEADERS include/DataFormatsParameters/GRPObject.h
+  HEADERS include/DataFormatsParameters/GRPObject.h
+  include/DataFormatsParameters/LHCIFData.h
```
shahor02: same

Detectors/DCS/testWorkflow/src/dcs-config-proxy.cxx
```diff
@@ -44,10 +46,10 @@ void sendAnswer(const std::string& what, const std::string& ack_chan, FairMQDevi
   }
 }
 
-auto getDetID(const std::string& filename)
+auto getDetID(const std::string& dIDStr)
 {
   // assume the filename start with detector name
-  return DetID::nameToID(filename.substr(0, 3).c_str(), DetID::First);
+  return DetID::nameToID(dIDStr.c_str(), DetID::First);
```
shahor02: I would keep the old definition of the method, so that it does what the name tells: gets the detID from the string.
Also, it can directly return DataOrigin, e.g.
```
auto getDetID(const std::string& dIDStr)
{
  auto did = DetID::nameToID(filename.substr(0, 3).c_str(), DetID::First);
  return did<0 ? gDataOriginInvalid : DetID::getDataOrigin(did);
}
```

Detectors/DCS/testWorkflow/src/dcs-config-proxy.cxx
```diff
@@ -65,15 +67,29 @@ InjectorFunction dcs2dpl(const std::string& acknowledge)
     std::string filename{static_cast<const char*>(parts.At(0)->GetData()), parts.At(0)->GetSize()};
     size_t filesize = parts.At(1)->GetSize();
     LOG(INFO) << "received file " << filename << " of size " << filesize;
-    int dID = getDetID(filename);
+    auto dIDStr = filename.substr(0, 3);
+    int dID = getDetID(dIDStr);
+    o2::header::DataOrigin dataOrigin;
```
shahor02: you can do w/o `found`, use directly (and moving `auto dIDStr = filename.substr(0, 3);` in the non-detector part:

```
auto dataOrigin = getDetID(filename);
if (dataOrigin == gDataOriginInvalid) {
  auto dIDStr = filename.substr(0, 3);
  for (auto& el : exceptionsDetID) {
     if (el.as<std::string>() == dIDStr) {
       dataOrigin = el;
       break;
     }
  }
}
if (dataOrigin == gDataOriginInvalid) {
  LOG(ERROR) << "unknown detector for " << filename;
  sendAnswer("error1: unrecognized filename", acknowledge, device);
  return;
}
```

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
```diff
@@ -0,0 +1,126 @@
...
+  LOG(DEBUG) << "posStart = " << posStart << ", posEnd = " << posEnd;
+  if (posEnd == std::string::npos) {
+    posEnd = mFileBuffStr.size();
+  }
+  std::string subStr = mFileBuffStr.substr(posStart, posEnd - posStart);
+  LOG(DEBUG) << "subStr = " << subStr;
+  boost::char_separator<char> sep("\t");
+  boost::tokenizer<boost::char_separator<char>> tokens(subStr, sep);
+  std::vector<std::string> tokensStr{begin(tokens), end(tokens)};
+  LOG(DEBUG) << "size of tokensStr = " << tokensStr.size();
```
shahor02: You can use https://github.com/AliceO2Group/AliceO2/blob/b1c1818be96481f59e94d3826aabbe38195cbad5/Common/Utils/include/CommonUtils/StringUtils.h#L105 
```
auto tokensStr = o2::utils::Str::tokenize(subStr, '\t');
```

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
```diff
@@ -0,0 +1,126 @@
...
+  if (tokensStr.size() < 5) {
+    LOG(FATAL) << "Number of tokens too small: " << tokensStr.size() << ", should be at 5 (alias, type, nelements, value(s), timestamp(s)";
+  }
+  boost::char_separator<char> sep_type(":");
+  boost::tokenizer<boost::char_separator<char>> tokens_type(tokensStr[1], sep_type);
+  std::vector<std::string> tokensStr_type{begin(tokens_type), end(tokens_type)};
+  LOG(DEBUG) << "size of tokensStr_type = " << tokensStr_type.size();
```
shahor02: idem:
```
auto tokensStr_type = o2::utils::Str::tokenize(tokensStr[1], ':');
```

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
```diff
@@ -0,0 +1,126 @@
...
+          LOG(INFO) << alias << ": value string = " << tokensStr[shift + iele];
+          vect.emplace_back(tokensStr[shift + iele]);
+        }
+      } else {
+        LOG(FATAL) << "templated function called with wrong type, should be string";
+      }
+    }
+
+    LOG(DEBUG) << "timestamp = " << std::stof(tokensStr[shift + nele]);
+    meas.emplace_back(std::make_pair(std::stof(tokensStr[shift + nele]), vect));
```
shahor02: with emplace you don't need to call constructor, it will call it itself:
```
meas.emplace_back(std::stof(tokensStr[shift + nele]), vect);
```

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
```diff
@@ -0,0 +1,126 @@
...
+  void readValue(const char* alias, std::string& type, int& nel, int& nmeas, std::vector<std::pair<float, std::vector<T>>>& meas);
+
+ private:
+  std::string mFileBuffStr; // buffer containing content of LHC IF file
+
+  ClassDefNV(LHCIFfileReader, 1);
+};
+
+template <typename T>
+void LHCIFfileReader::readValue(const char* alias, std::string& type, int& nele, int& nmeas, std::vector<std::pair<float, std::vector<T>>>& meas)
```
shahor02: better use `const std::string& alias`.
Why do you need `nmeas`, isn't `meas.second.size()` enough?

Detectors/GRP/calibration/src/LHCIFfileReader.cxx
```diff
@@ -0,0 +1,47 @@
...
+#include "GRPCalibration/LHCIFfileReader.h"
+#include "TSystem.h"
+#include <fstream>
+
+namespace o2
+{
+namespace grp
+{
+
+void LHCIFfileReader::loadLHCIFfile(const char* fileName)
```
shahor02: use `const std::string& fileName`

Detectors/GRP/calibration/src/LHCIFfileReader.cxx
```diff
@@ -0,0 +1,47 @@
...
+{
+  // load the LHC IF file into a string
+  char* expandedFileName = gSystem->ExpandPathName(fileName);
+  std::ifstream ifs(expandedFileName);
```
shahor02: why do you need `gSystem`?  If you wand to check file existence or expand its path, you can use
`o2::utils::Str::pathExists(filename);`
and 
`o2::utils::Str::getFullPath(filename);`

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
```diff
@@ -0,0 +1,126 @@
...
+        for (int iele = 0; iele < nele; ++iele) {
+          LOG(INFO) << alias << ": value string = " << tokensStr[shift + iele];
+          vect.emplace_back(tokensStr[shift + iele]);
+        }
+      } else {
+        LOG(FATAL) << "templated function called with wrong type, should be string";
+      }
+    }
+
+    LOG(DEBUG) << "timestamp = " << std::stof(tokensStr[shift + nele]);
```
shahor02, line 114: Why is the time represented by float?

DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
```diff
@@ -0,0 +1,74 @@
...
+{
+
+class LHCIFData
+{
+ public:
+  LHCIFData() = default;
+  ~LHCIFData() = default;
+
+  std::pair<float, int32_t> getBeamEnergy() const { return mBeamEnergy; }
+  std::pair<float, std::string> getFillNumber() const { return mFillNumber; }
```
shahor02: I would convert fill to integer

DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
```diff
@@ -0,0 +1,74 @@
...
+namespace parameters
+{
+
+class LHCIFData
+{
+ public:
+  LHCIFData() = default;
+  ~LHCIFData() = default;
+
+  std::pair<float, int32_t> getBeamEnergy() const { return mBeamEnergy; }
```
shahor02: why the time is float?

shahor02 (changes requested):

DataFormats/Parameters/CMakeLists.txt
chiarazampolli: This was changed by clang-format. I can change it back, but then it will probably not pass the test.

DataFormats/Parameters/CMakeLists.txt
chiarazampolli: same too

DataFormats/Parameters/CMakeLists.txt
chiarazampolli: Now it worked, I will try committing the fix.

DataFormats/Parameters/CMakeLists.txt
chiarazampolli: Also here clang-format did not complain now.

DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
chiarazampolli: This is provided like this from LHC_IF. I asked Antonio, and he said that this is what they get. Should I change it to uint64_t, as all our timestamps?

DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
chiarazampolli: Ok. I indeed find the types quite weird.

DataFormats/Parameters/CMakeLists.txt
shahor02: I think clang-format does not touch the CMakeFiles, it is probably tab/spaces substitution

DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
shahor02: yes, we better stick to the same types and units: long, microseconds (but the CCDB validity range in ms).

Detectors/DCS/testWorkflow/src/dcs-config-proxy.cxx
chiarazampolli: Done

Detectors/DCS/testWorkflow/src/dcs-config-proxy.cxx
chiarazampolli: Yes, I already changed it like this while implementing the previous comment

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
chiarazampolli: Ok for std::string instead of char*. 
As for nmeas, I store it just because it is in the LHC-IF file, in case we want to use it as a x-check with what we get (for now I don't do it).

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
chiarazampolli: Good to know, but you prefer that I change it, or it does not really matter?

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
chiarazampolli: will change depending on the answer to the previous comment.

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
chiarazampolli: True. changed

Detectors/GRP/calibration/src/LHCIFfileReader.cxx
chiarazampolli: Ok, I did not know.

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
shahor02: Yes, I would minimize the usage of the boost, also the StringUtil does trimming of whitespaces

DataFormats/Parameters/CMakeLists.txt
chiarazampolli: Yes, I had the doubt, it was probably that. I will see when I commit all modifications.

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
chiarazampolli: done

Detectors/GRP/calibration/include/GRPCalibration/LHCIFfileReader.h
chiarazampolli: It is like this that we get it from LHC-IF. I will change it.

Detectors/DCS/testWorkflow/src/dcs-config-proxy.cxx
chiarazampolli: Actually I changed it a bit differently. Once I commit, maybe you can check if you "like" it? The check on the excpetions is done in `getDetID`, which I now called `getDataOriginFromFilename`.

DataFormats/Parameters/include/DataFormatsParameters/LHCIFData.h
chiarazampolli: done

DataFormats/Parameters/CMakeLists.txt
chiarazampolli: all fine now

DataFormats/Parameters/src/ParametersDataLinkDef.h
```diff
@@ -24,5 +24,7 @@
 #pragma link off all functions;
 
 #pragma link C++ class o2::parameters::GRPObject + ;
+#pragma link C++ class o2::parameters::LHCIFData + ;
+#pragma link C++ class std::pair < float, std::string> + ;
```
shahor02: Forgot to change float here.

Detectors/DCS/testWorkflow/src/dcs-config-proxy.cxx
```diff
@@ -164,6 +177,10 @@ WorkflowSpec defineDataProcessing(ConfigContext const& config)
     dcsOutputs.emplace_back(DetID(id).getDataOrigin(), "DCS_CONFIG_FILE", 0, Lifetime::Timeframe);
```
shahor02, line 177: Seems we have a bug in all proxies: splitPayloadIndex must go from 0 to splitPayloadParts-1. And, for DCS to CCDB, where we send the multipart of 2, it should be 
```
    hdrN.splitPayloadParts = 2;
    hdrN.splitPayloadIndex = 0;
    hdrN.payloadSize = parts.At(0)->GetSize();
    hdrN.firstTForbit = 0; // this should be irrelevant for DCS


    hdrF.tfCounter = *timesliceId; // this also
    hdrF.payloadSerializationMethod = o2::header::gSerializationMethodNone;
    hdrF.splitPayloadParts = 2;
    hdrF.splitPayloadIndex = 1;
```
instead of 
https://github.com/AliceO2Group/AliceO2/blob/89e3fe068f17e371c1edc06ebcf688ad5cead9db/Detectors/DCS/testWorkflow/src/dcs-config-proxy.cxx#L100-L108

shahor02 (changes requested):
Hi @chiarazampolli 

Please see more comments below. 
For the time: we use everywhere `long`, better to stick to it rather than uint64_t, to avoid later warnings about signed/unsigned comparison.

Detectors/DCS/testWorkflow/src/dcs-config-proxy.cxx
chiarazampolli: Done

DataFormats/Parameters/src/ParametersDataLinkDef.h
chiarazampolli: Done

chiarazampolli:
`uint64_t` changed to `long`.

Chiara

shahor02 (approved):

PR #7149, opened 2021-09-29
extend TOF matching to TRD

noferini:


noferini:
@davidrohr @chiarazampolli

noferini:
Hi @martenole ,
I implemented TOF matching with TRD tracks.
I run with different configurations.

Here few stats when using different sources for tracks
Single cases
ITS-TPC 55
TPC 99
TPC-TRD 0
ITS-TPC-TRD 31

Combined ones
TPC,ITS-TPC 99
TPC,ITS-TPC,ITS,ITS-TPC-TRD 99
TPC,ITS-TPC,ITS,ITS-TPC-TRD,TPC-TRD 81

It seems that TPC-TRD tracks (TrackTRD) don't match anything
while ITS-TPC-TRD tracks (also TrackTRD) match

The code for both is exactly the same since they are both o2::trd::TrackTRD
From what we should expect 18 matching with TPC-TRD tracks in my simulation.

The command to check matching with TPC-TRD tracks is this one
o2-tof-matcher-workflow -b  --track-sources "TPC-TRD"

martenole:
Hi Francesco,
thanks for trying this. It looks as if either I am doing a mistake in the refit of TPC-TRD tracks or maybe it is because of the time estimate. I see that upon creation ITS-TPC-TRD tracks get the time estimate of the BC together with an error of 1us:
https://github.com/AliceO2Group/AliceO2/blob/8b51bd1dc9ab1de9cf2cc81ac09504dc6c7ac0ec/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L111
While the TPC-TRD tracks get an uncertainty of only 1 ns:
https://github.com/AliceO2Group/AliceO2/blob/8b51bd1dc9ab1de9cf2cc81ac09504dc6c7ac0ec/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L155
I think the 1 us error is a typo and it should also be 1 ns for ITS-TPC-TRD tracks, right @shahor02 ?
I will try to have a look also myself with the changes from your PR, but I am not sure if I will manage today.
Cheers,
Ole

noferini:
I guess that 1E-6 means 1 ps

shahor02:
@martenole  the time is in \mus (except for TPC tracks, where it is in timebins), so 1e-3 is indeed 1ns.

noferini:
Moving to 1E-3 I got 23 matching with TPC-TRD

This is the overall case

[52528:TOFCalibWriter]: [13:35:55][INFO] RECEIVED MATCHED SIZE 101

[52529:TOFMatchedWriter_ITSTPC]: [13:35:55][INFO] RECEIVED MATCHED SIZE 25
[52530:TOFMatchedWriter_ITSTPCTRD]: [13:35:55][INFO] RECEIVED MATCHED SIZE 30
[52531:TOFMatchedWriter_TPC]: [13:35:55][INFO] RECEIVED MATCHED SIZE 25
[52532:TOFMatchedWriter_TPCTRD]: [13:35:55][INFO] RECEIVED MATCHED SIZE 21

I will add this fix in a while

noferini:
Done
Thanks to everybody

noferini:
I think there is still a problem when running in strict mode related to INPUT matching.

I guess that
  o2::globaltracking::InputHelper::addInputSpecs(configcontext, specs, clustermask, nonemask, src, useMC, mcmaskcl, GID::getSourcesMask(GID::ALL), strict);

move di subspec 1 all input but the TRD reader complains with

[96075:tpctrd-track-reader]: [16:11:02][ERROR] Exception caught: Worker is not authorised to create message with origin(TRD) description(MATCH_TPC) subSpec(0)

noferini:
I found the problem. I will fix and update the PR in
  TRDTrackReader(bool useMC, Mode mode, bool subSpecStrict = false) : mUseMC(useMC), mMode(mode), mSubSpecStrict(subSpecStrict) {}

last initialization was missing

noferini:
Now validated both for
o2-tof-matcher-workflow -b
o2-tof-matcher-workflow -b --strict-matching

noferini:
FullCI check failing for reasons not related to this PR (if I am not wrong)

martenole (approved):
Looks good and thanks a lot for the fixes for the TRD track reader + track creator

noferini:
checks:
- fullCI seems not related to this PR
- macOS: failed at the beginning... has it to be re-run?

davidrohr:
@TimoWilken : could you restart the failed CIs`?

shahor02 (approved):

PR #7150, opened 2021-09-29
[EMCAL-606] Test all mappings in the test for the mapper

mfasDa:
- Add missing DDL mappings to the test
- Exclude ambiguous channels from the test of the
  inverse mapping
- Use boost::hash_combine to calculate hash value
  for inverse mapping

### DIFF
--- a/Detectors/EMCAL/base/include/EMCALBase/Mapper.h
+++ b/Detectors/EMCAL/base/include/EMCALBase/Mapper.h
@@ -19,6 +19,7 @@
 #include <unordered_map>
 #include <sstream>
 #include <string>
+#include <boost/container_hash/hash.hpp>
 
 #include "fmt/format.h"
 #include "RStringView.h"
@@ -64,10 +65,17 @@ class Mapper
     /// \return hash value for channel ID
     size_t operator()(const ChannelID& s) const
     {
+      std::size_t seed = 0;
+      boost::hash_combine(seed, s.mRow);
+      boost::hash_combine(seed, s.mColumn);
+      boost::hash_combine(seed, o2::emcal::channelTypeToInt(s.mChannelType));
+      return seed;
+      /*
       size_t h1 = std::hash<int>()(s.mRow);
       size_t h2 = std::hash<int>()(s.mColumn);
       size_t h3 = std::hash<int>()(o2::emcal::channelTypeToInt(s.mChannelType));
       return ((h1 ^ (h2 << 1)) >> 1) ^ (h3 << 1);
+      */
     }
   };
 
--- a/Detectors/EMCAL/base/test/testMapper.cxx
+++ b/Detectors/EMCAL/base/test/testMapper.cxx
@@ -12,6 +12,7 @@
 #define BOOST_TEST_MAIN
 #define BOOST_TEST_DYN_LINK
 #include <boost/test/unit_test.hpp>
+#include <fmt/format.h>
 #include <DataFormatsEMCAL/Constants.h>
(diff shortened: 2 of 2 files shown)


mfasDa:
The faulty channel was put there on purpose in order to test that the exception are caught correctly (with BOOST_CHECK_EXCEPTION). Both on my mac and under Ubuntu 18.04 the test works fine, no clue why it fails in the CI.

shahor02:
@mfasDa I am not particularly familiar with Boost tests, perhaps @aphecetche may suggest how to test the exceptions handling?

mfasDa:
@shahor02 This part of the test is not new and has worked in the past. Furthermore it works on both systems I use locally for development, including a mac system.

shahor02:
> @shahor02 This part of the test is not new and has worked in the past. Furthermore it works on both systems I use locally for development, including a mac system.

@mfasDa but if it breaks all CI's it will create problems for all other PRs, let's wait until boost test experts respond

aphecetche:
@mfasDa actually did try this PR on my Mac : Big Sur with Xcode Version 12.5.1 (12E507)(i.e. not the latest one) with a slightly different build system (Spack instead of aliBuild, but I don't expect this to be relevant here), and I think I can reproduce the failure : 

```
📦dev🚧...O2/spack-build-fuqytjn [pr/7150] $ export O2_ROOT=$(spack location -i o2-aliceo2)
📦dev🚧...O2/spack-build-fuqytjn [pr/7150] $ ninja && stage/tests/o2-test-emcal-Mapper
ninja: no work to do.
Running 1 test case...
Test mapping RCU0A.data
unknown location:0: fatal error: in "Mapper_test": o2::emcal::Mapper::ChannelNotFoundException: Channel with Row 0, Column 16, type LEDmon not found.
../Detectors/EMCAL/base/test/testMapper.cxx:70: last checkpoint

*** 1 failure is detected in the test module "Test EMCAL Base"

📦dev🚧~$ uname -a
Darwin  20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64
```

mfasDa:
@aphecetche Thanks for checking. The exception thrown on purpose, the goal is that BOOST_TEST checks whether the correct exception is thrown for an inexistent channel. For this BOOST_CHECK_EXCEPTION should catch the exception and check against the type. This has worked in the pasty but seems to not work in this test. If I disable the test what is its purpose then. In principle BOOST test should support exception tests.

aphecetche:
Sure, I completely agree on the intent, but turns out `BOOST_CHECK_EXCEPTION` is not the culprit here : the error actually comes from line 70 of that test, which is a regular `BOOST_CHECK_EQUAL`, so it seems to be a genuine error after all.

shahor02:
@mfasDa you test still fails on macos with
```
unknown location:0: [4;31;49mfatal error: in "Mapper_test": o2::emcal::Mapper::ChannelNotFoundException: Channel with Row 0, Column 17, type LEDmon not found.[0;39;49m
/System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SOURCES/O2/7150/0/Detectors/EMCAL/base/test/testMapper.cxx:70: [1;36;49mlast checkpoint[0;39;49m
```

mfasDa:
Hmm, there was indeed a bug going to the boost hasher for the unordered map that the default initialization was missing when setting up the combined hasher, which can indeed lead to errors in the inverse mapping. Fixed.

PR #7157, opened 2021-09-29
Avoid aliasing problems in the DataRelayer

ktf:


ktf:
@shahor02 @davidrohr can you check if this gets rid of the backpressure? if it does, it means there is some caching alias like issue which I need to understand. I am trying with your reproducer but I have (unrelated) troubles getting it to run on my laptop.

shahor02:
@ktf For the workflow w/o pipelines it does not make any difference, the warnings are there (and as far as I can see, your change matters only for the pipelines.
I've tested the same workflow with its-tracker pipelined, and now the processing simply gets stuck with such a giu (and the warnings are still there):

image

W/o this PR, although still with warnings, the processing of same 300 TFs completes.

PR #7158, opened 2021-09-29
EventVisualisation documentation

jmyrcha:


shahor02 (approved):
Documentation and config files only, merging

PR #7163, opened 2021-09-29
DPL Analysis: convince ROOT not to keep buffers around

ktf:


ktf:
Tested on hyperloop. Seems to give 10-15% gain PSS in certain workflows.

PR #7168, opened 2021-09-30
TPCQC: fix cluster task

tklemenz:
- fixed a bug where normalization to number of clusters did not properly work
 - added reset function to get a fresh set of CalPad members
 - renamed analyze function to normalize

### DIFF
--- a/Detectors/TPC/qc/src/Clusters.cxx
+++ b/Detectors/TPC/qc/src/Clusters.cxx
@@ -31,6 +31,11 @@ using namespace o2::tpc::qc;
 template <class T>
 bool Clusters::processCluster(const T& cluster, const o2::tpc::Sector sector, const int row)
 {
+  if (mIsNormalized) {
+    denormalize();
+    LOGP(warning, "calling denormalize() before filling");
+  }
+
   const int nROC = row < 63 ? int(sector) : int(sector) + 36;
   const int rocRow = row < 63 ? row : row - 63;
 
@@ -66,6 +71,11 @@ bool Clusters::processCluster(const T& cluster, const o2::tpc::Sector sector, co
 //______________________________________________________________________________
 void Clusters::fillADCValue(int cru, int rowInSector, int padInRow, int timeBin, float adcValue)
 {
+  if (mIsNormalized) {
+    denormalize();
+    LOGP(warning, "calling denormalize() before filling");
+  }
+
   const CRU cruID(cru);
   float val;
   val = mNClusters.getValue(cruID.sector(), rowInSector, padInRow);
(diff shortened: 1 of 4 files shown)


tklemenz:
@wiechula This works properly now. Any further remarks?

Detectors/TPC/qc/src/Clusters.cxx
```diff
@@ -88,6 +88,27 @@ void Clusters::analyse()
...
+
+//______________________________________________________________________________
+void Clusters::reset()
+{
+  mNClusters = CalPad("N_Clusters");
+  mQMax = CalPad("Q_Max");
+  mQTot = CalPad("Q_Tot");
+  mSigmaTime = CalPad("Sigma_Time");
+  mSigmaPad = CalPad("Sigma_Pad");
+  mTimeBin = CalPad("Time_Bin");
```
wiechula: I think you could simply do
```suggestion
  mNClusters *= 0.f;
  mQMax *= 0.f;
  mQTot *= 0.f;
  mSigmaTime *= 0.f;
  mSigmaPad *= 0.f;
  mTimeBin *= 0.f;
```
Or am I overlooking something?

wiechula (changes requested):

Detectors/TPC/qc/src/Clusters.cxx
tklemenz: Yes, this also works and is probably faster.

wiechula (changes requested):
Perhaps it would be good to add a Datamember
```
bool mIsNormalized{false};
```
And only allow normalization in case it is not already, only allow de-normalization in case it is normalized and denormalize in case of filling and it is normalized. What do you think?

tklemenz:
I can do that but I don't think it is needed. De-normalizing the data when it is not normalized happens before the first data is processed or after the reset. In this case the zeros are multiplied by zero. We can of course spare the computing time with the flag but for the data itself I would say it doesn't matter.
I'll implement it anyway since spared computing time is always good.

wiechula:
Yes, but it might safe trouble in future if the code will also be used elsewhere...

Detectors/TPC/qc/src/Clusters.cxx
```diff
@@ -79,13 +89,38 @@ void Clusters::fillADCValue(int cru, int rowInSector, int padInRow, int timeBin,
 }
 
 //______________________________________________________________________________
-void Clusters::analyse()
+void Clusters::normalize()
 {
   mQMax /= mNClusters;
```
wiechula, line 98:
```suggestion
  if (mIsNormalized) {
    return;
  }
  mQMax /= mNClusters;
```

Detectors/TPC/qc/src/Clusters.cxx
```diff
@@ -79,13 +89,38 @@ void Clusters::fillADCValue(int cru, int rowInSector, int padInRow, int timeBin,
...
   mSigmaPad /= mNClusters;
   mTimeBin /= mNClusters;
+
+  mIsNormalized = true;
+}
+
+//______________________________________________________________________________
+void Clusters::denormalize()
+{
+  mQMax *= mNClusters;
```
wiechula, line 114:
```suggestion
if (!mIsNormalized) {
    return;
  }
  mQMax *= mNClusters;
```

wiechula (changes requested):
Please add missing checks.

wiechula (dismissed):

wiechula (dismissed):

tklemenz:
I wasn't aware of the fact that CI tests in O2 also build QC. Little surprisingly the tests fail since there are two new functions and a renamed one (`analyse() --> normalize()`) being introduced with this PR. Therefore, QC fails since in this QC branch the not-renamed one is still being called. Do I really need to introduce a function with the old name calling the newly named one to make the CI tests work?

wiechula (approved):

Detectors/TPC/qc/src/Clusters.cxx
```diff
@@ -79,13 +89,46 @@ void Clusters::fillADCValue(int cru, int rowInSector, int padInRow, int timeBin,
...
+
+//______________________________________________________________________________
+void Clusters::reset()
+{
+  mNClusters *= 0;
+  mQMax *= 0;
+  mQTot *= 0;
+  mSigmaTime *= 0;
+  mSigmaPad *= 0;
+  mTimeBin *= 0;
```
shahor02, line 131: I don't know how often this method is used, but what is the reason of multiplying by 0 instead of assignment?

shahor02 (approved):
Hi @tklemenz merging as it is now, but please check 1 comment below, perhaps this should be fixed in the next iteration

Detectors/TPC/qc/src/Clusters.cxx
tklemenz: Hi @shahor02, these member variables are CalDet<float> objects. Assigning 0 would not work.

PR #7169, opened 2021-09-30
MCH: optimized DataDecoder::reset() function

aferrero2707:
The function had a slow initialization loop over a large vector. It is now replaced by a memset() over a smaller data structure. This provides a speed-up of more than x100.

### DIFF
--- a/Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
+++ b/Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
@@ -275,20 +275,40 @@ void dumpOrbits(const std::unordered_set<OrbitInfo, OrbitInfoHash>& mOrbits)
 
 //_________________________________________________________________________________________________
 
-int32_t DataDecoder::getMergerChannelId(const DsElecId& dsElecId, DualSampaChannelId channel)
+bool DataDecoder::getMergerChannelId(const DsElecId& dsElecId, DualSampaChannelId channel, uint32_t& chId, uint32_t& boardId)
 {
+  static constexpr uint32_t sChannelsInOneDs = 64;
+  static constexpr uint32_t sDsInOneSolar = 40;
+  static constexpr uint32_t sChannelsInOneSolar = sChannelsInOneDs * sDsInOneSolar;
   auto solarId = dsElecId.solarId();
   uint32_t dsId = static_cast<uint32_t>(dsElecId.elinkGroupId()) * 5 + dsElecId.elinkIndexInGroup();
   if (solarId > DataDecoder::sMaxSolarId || dsId >= 40 || channel >= 64) {
-    return -1;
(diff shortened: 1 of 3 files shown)


Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
```diff
@@ -715,6 +738,21 @@ void DataDecoder::init()
...
   initElec2DetMapper(mMapFECfile);
+
+  mTimeFrameStartRecords.resize(sReadoutChipsNum);
+  std::fill(mTimeFrameStartRecords.begin(), mTimeFrameStartRecords.end(), TimeFrameStartRecord());
+
+  mMergerRecords.resize(sReadoutChannelsNum);
+  mMergerRecordsReady.resize(sReadoutBoardsNum);
+
+  if (mMergerRecords.size() != DataDecoder::sReadoutChannelsNum) {
+    std::cout << "WRONG SIZE OF mMergerRecords" << std::endl;
```
aphecetche: could you use `LOGP(error,".......")` instead of bare `cout` so that the messages are accessible in the info logger as well ?

Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
```diff
@@ -300,13 +320,14 @@ bool DataDecoder::mergeDigits(uint32_t mergerChannelId, o2::mch::raw::SampaClust
     return false;
   }
 
-  auto& mergerCh = mMergerRecords[mergerChannelId];
-
   // if there is not previous digit for this channel then no merging is possible
-  if (mergerCh.digitId < 0) {
+  //flog << fmt::format("  mask {:X}  ready {:X}", mergerChannelBitmask, mMergerRecordsReady[mergerBoardId]) << std::endl;
```
aphecetche: remove commented out code

aphecetche:
@aferrero2707 great ! Just to be sure : the 100x speedup is for the `reset` method alone, not the full decoder, right ? :wink:

Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
aferrero2707: Those checks are actually not really needed, just a leftover of some debugging. I have removed  them.

Detectors/MUON/MCH/Raw/Decoder/src/DataDecoder.cxx
aferrero2707: Done.

aferrero2707:
Yes, just the reset() method... but it was taking around 1/3 of the whole decoding time, if I am not wrong, so the improvement is not too bad...

aphecetche (approved):

PR #7175, opened 2021-09-30
Warn instead of crashing fataly

sawenzel:


sawenzel:
@nburmaso : I see some crashes with this and I prefer to treat it non-fataly for the moment.

PR #7176, opened 2021-09-30
Changes + fixes in CTF writer

shahor02:
* Write the CTF meta-info file to the dedicated directory provided via `--meta-output-dir` option, skip writing it if `/dev/null` is supplied (default).
* if `--max-ctf-per-file <N>` provides values `>0`, close CTF file with multiple CTFs after writing N CTFs, even if the `--min-file-size` is not reached
* `--save-ctf-after <N>` allows to auto-save CTF file with multiple CTFs after every `N>0` CTFs.

* Reshuffle some options + fix in CTF tree saving


shahor02:
@ktf for the record: the workflows piped together are allowed to have global options with the same name and they even can take different values. But if the device of one workflow has a local option with the same name as the global option in *different* workflow, this leads to boost error about the ambiguous options. This seems illogical: if similar global options of 2 workflows don't clash, moreover the global option of one workflow should not clash with the local option of another one.

PR #7183, opened 2021-10-01
MID: add protection for data from unexpected elements

dstocco:
During data taking, we sometime get uncorrect data. When this happens, the decode might crash.
With this protection, an error message is issued but the decoder does not crash


dstocco:
@aphecetche : all green ;)

aphecetche (approved):

PR #7185, opened 2021-10-01
Timeframe rate limit

ktf:


jgrosseo:
That was fast :)

ktf:
Unfortunately it does not improve the memory usage, at least not so far. I am back to thinking that the current rate limiting is actually doing the correct thing already...

jgrosseo:
I think this fixes the use case of a slow task at the end. Did you try to the use case `CorrelationsFilteredOnTheFly`? I can set it up in `O2 Development` if you have the code on hyperloop.

ktf:
I am not sure actually. What I saw (but maybe I did not not look too well) is that the number of in-fly dataframes was already low. I will look better this week. Anyways, we need it in any case also for the online, so I will make this ready to merge sometimes today.

ktf:
@davidrohr this fixes the conflicts.

PR #7188, opened 2021-10-03
[MRRTF-152] update electronic mapping (St1 complete, St2 partial)

aphecetche:
This PR completes the electronic mapping for MCH, by adding Station 1 (CH1,CH2) (which mapping should be complete now), and St2 (CH3,CH4). Note that for St2 only the CRU mapping is ok : the FEC mapping still needs to be finalised (but with this PR the data decoder should no longer crash when encountering St2 data at least).

Also add feature to generate `cru.map` file to `elecmap.py` script and to
`o2-mch-elecmap-cli` program (`--generate-cru-map` option).

`o2-mch-elecmap-cli` also gains the option to read from custom mapping
cru.map file (`--cru-map`option).

Note that the `cru.map` and `fec.map` can be used by the MCH raw data decoder (`o2-mch-raw-to-digits-workflow`) using its `--fec-map` and `--cru-map` options (which replace the integrated mapping that comes with the MCH libraries).


aphecetche:
@shahor02 this is not the end of the electronic mapping story for MCH, but at least with this PR the TED shots data can be decoded w/o crash

shahor02:
@aphecetche shall I merge it already?

aphecetche:
@shahor02  yes please. Some updates might come as early as tomorrow, but I'd prefer to have a least this one usable asap.

shahor02 (approved):

PR #7193, opened 2021-10-04
vertex trigger threshold  the same as in  TCM

AllaMaevskaya:


mslupeck (approved):

AllaMaevskaya:
What is wrong here? Can it be merged?

PR #7195, opened 2021-10-04
Tracking mode "cosmics" for the CM tracker

iouribelikov:
This PR provides a set of loose cuts to allow for catching cosmic tracks from the ITS commissioning data.
It can be used like this:   
```shell
... |  o2-its-reco-workflow --tracking-mode cosmics ...
```

### DIFF
--- a/Detectors/ITSMFT/ITS/reconstruction/include/ITSReconstruction/CookedTracker.h
+++ b/Detectors/ITSMFT/ITS/reconstruction/include/ITSReconstruction/CookedTracker.h
@@ -75,6 +75,20 @@ class CookedTracker
     gRoadZ = par[12];
     gminNumberOfClusters = par[13];
   }
+  void setParametersCosmics()
+  {
+    // seed "windows" in z and phi: makeSeeds
+    gzWin = 84.; // length of the L3
+    gminPt = 10.;
+    // Maximal accepted impact parameters for the seeds
+    gmaxDCAxy = 19.4; // radius of the L3
+    gmaxDCAz = 42.;   // half-lenght of the L3
+    // Space point resolution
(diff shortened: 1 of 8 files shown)


shahor02 (dismissed):

shahor02:
@iouribelikov there are formatting errors: `/sw/SOURCES/O2/7195-slc8_x86-64/0/Detectors/ITSMFT/ITS/reconstruction/src/CookedTracker.cxx:392:34: error: statement should be inside braces `

Detectors/ITSMFT/ITS/reconstruction/include/ITSReconstruction/CookedTracker.h
```diff
@@ -133,6 +167,29 @@ class CookedTracker
...
+  // Space point resolution
+  static Float_t gSigma2;
+  // Max accepted chi2
+  static Float_t gmaxChi2PerCluster;
+  static Float_t gmaxChi2PerTrack;
+  // Tracking "road" from layer to layer
+  static Float_t gRoadY;
+  static Float_t gRoadZ;
+  // Minimal number of attached clusters
+  static Int_t gminNumberOfClusters;
```
shahor02, line 191: I think this better define as ConfigurableParam (in a separate PR)

shahor02 (approved):

Detectors/ITSMFT/ITS/reconstruction/include/ITSReconstruction/CookedTracker.h
iouribelikov: > I think this better define as ConfigurableParam (in a separate PR)

Good point !  This would make setting the parameters much more flexible.

PR #7198, opened 2021-10-04
[EMCAL-551] Align indices for HG/LG cells

mfasDa:
Adapt order of HG/LG cells to comply with
the indices used in the mapping (0 - low
gain, 1 - high gain)


mfasDa:
Even though we use getter/setter functions or converters for the channel type I think it is safer to align the enum index with the enum index used in the mapping file (0-low gain, 1-high gain) and not vice versa. This swaps the cell type for existing (CTF) data but there is nothing yet taken which is really of relevance.

shahor02 (approved):

mfasDa:
Errors unrelated - @shahor02 can this be merged?

shahor02 (approved):

PR #7199, opened 2021-10-04
Incremental table reading

ktf:


ktf:
Hold. This actually does not work with 2 readers. Fix in progress.

PR #7203, opened 2021-10-04
DPL Analysis: prevent adding extra empty slices when the last index is -1

aalkin:
* Check for the last non-negative value before padding the slices to full size
* Add a test case


jgrosseo (approved):

ktf:
Analysis only. Merging.

PR #7206, opened 2021-10-04
[MFT] Flip ALPIDE sensor for MFT

rpezzi:
The sensor was upside down.


rpezzi:
Converting to draft while running more tests.

rpezzi:
More data is needed to fully test this change. Unfortunately x-ray data obtained in March is in old format. For now, this PR improve things.

PR #7210, opened 2021-10-05
ClusterNativeHelper: fixing convert to tree method

matthias-kleiner:
the ClusterNative format was already adapted in an earlier [PR](https://github.com/AliceO2Group/AliceO2/commit/a740637a016f2f2f93fd73cc454dc8647bd70d06), but the runtime error not commented out.
Also the event and tree was not filled consistently.


shahor02 (approved):
t

PR #7214, opened 2021-10-05
Minor improvement in hw error handling

peressounko:


Detectors/PHOS/calib/src/PHOSHGLGRatioCalibDevice.cxx
```diff
@@ -142,13 +144,16 @@ void PHOSHGLGRatioCalibDevice::checkRatios()
   }
 
   LOG(INFO) << "Got current calibration from CCDB";
-
   //Compare to current
   int nChanged = 0;
   for (short i = o2::phos::Mapping::NCHANNELS; i > 1792; i--) {
```
shahor02, line 149: @peressounko just to be on the safe side: your loop corresponds to root TAxis convention, where the 1st bin is 1, not 0.
But do you use the same convention for your calib objects, i.e. do you count the channels from 1?

shahor02 (approved):

Detectors/PHOS/calib/src/PHOSHGLGRatioCalibDevice.cxx
peressounko: @shahor02, we have convention about absId described in Geometry.cxx file: absId starts from 1 and first half of module 1 up to channel 1792 included is empty. I try to follow this convention everywhere. In Calib objects to save space we shift absId and start internal numbering from 0 what corresponds to absId=1793.

shahor02:
ok, thanks for the explanations.

PR #7215, opened 2021-10-05
DPL Analysis: improve group slicer test with several unassigned groups

aalkin:
Adds several unassigned groups to the slicing test to imitate merged AOD file.


ktf:
@sawenzel there is an unrelated error from the simulation tests:

```
357/395 Test #354: o2sim_G3 .............................................................................***Failed  Required regular expression not found. Regex=[SIMULATION RETURNED SUCCESFULLY
```

is it known?

PR #7216, opened 2021-10-05
Switch from options to ClustererParam for ITS/MFT clust.dict. and noise

shahor02:
ITS and MFT DPL-device options cluster-dict-file, dict-file, noise-file ets were eliminated in favour of
configurable param ITSClustererParam.dictFilePath and ITSClustererParam.noiseFilePath (idem for MFT), i.e.
the option should be provided as
`--configKeyValues "ITSClustererParam.dictFilePath=XXX"` with XXX is either the directory where the default filename
is checked or explicit full path.

@jmyrcha I did some fixes in the dictionaries initialization and usage in the EveWorkflowHelper and O2DPLDisplay, including the setting of `mConfig->configCalib.itsPatternDict` (and same for MFT) in  https://github.com/shahor02/AliceO2/blob/7fa33850b62dbedd4f4e80f76c7553db757f6b5b/EventVisualisation/Workflow/src/O2DPLDisplay.cxx#L81-L99, but I guess this is an obsolete code, you can probably remove it later.

@davidrohr this will invalidate O2DataProcessing `dpl-workflow.sh`, once the test are passed I'll prepare a sync. PR for O2DataProcessing so that both can be merged simultaneously. But then the updated O2DataProcessing should not be used on epns, until the O2 with this PR propagates there (and at that moment all workflows involving ITS/MFT will need to be regenerated).

### DIFF
--- a/Detectors/CTF/README.md
+++ b/Detectors/CTF/README.md
@@ -115,14 +115,14 @@ For the ITS and MFT entropy decoding one can request either to decompose cluster
 or to apply the noise mask to decoded clusters (or decoded digits). If the masking (e.g. via option `--its-entropy-decoder " --mask-noise "`) is requested, user should provide to the entropy decoder the noise mask file (eventually will be loaded from CCDB) and cluster patterns decoding dictionary (if the clusters were encoded with patterns IDs).
 For example,
 ```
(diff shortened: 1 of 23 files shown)


davidrohr:
@shahor02 : I would merge this once it is green, and then merge the O2DataProcessing PR at the time we update the software on the EPN.

PR #7217, opened 2021-10-05
[EMCAL-630] Handling of overlap region of HG/LG digits

mfasDa:
- In case both HG and LG are present we keep
  the HG if the amp is < 950 ADC counts, otherwise
  the low gain
- In case only HG is present we keep the cell if the
  amp is < 950 ADC counts, otherwise we flag it as
  HG-out-of-range
- In case only LG is present we flag the cell as
  LGnoHG.
All flagged cells are filtered out (discarded). In this
case a error is logged on the infoBrowser and an
error object is added to the raw data errors for QC.

### DIFF
--- a/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
@@ -9,6 +9,7 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 #include <string>
+#include <iomanip>
 #include <iostream>
 #include <bitset>
 
@@ -21,6 +22,7 @@
 #include "Framework/Logger.h"
 #include "Framework/WorkflowSpec.h"
 #include "DataFormatsEMCAL/EMCALBlockHeader.h"
+#include "DataFormatsEMCAL/Constants.h"
 #include "DataFormatsEMCAL/TriggerRecord.h"
 #include "DataFormatsEMCAL/ErrorTypeFEE.h"
 #include "DetectorsRaw/RDHUtils.h"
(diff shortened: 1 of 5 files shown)


Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -47,6 +49,7 @@ RawToCellConverterSpec::~RawToCellConverterSpec()
 
 void RawToCellConverterSpec::init(framework::InitContext& ctx)
 {
+  std::cout << "Init called" << std::endl;
```
shahor02: Please use `LOG(info)`

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -87,6 +90,7 @@ void RawToCellConverterSpec::init(framework::InitContext& ctx)
 
 void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
 {
+  std::cout << "Called" << std::endl;
```
shahor02: idem, though this looks like debug message

shahor02 (changes requested):

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
mfasDa: Ooops, yes, same as above. Will fix immediately

shahor02 (dismissed):

mfasDa:
For the record: Both digitizers operate in an overlapping region which is from 100 MeV to 16 GeV, so in this range the the front-end sends always 2 bunches per tower. Below only the HG digitizer sends data. Up to 16 GeV we use the HG digitizer as it has the better energy resolution and drop the LG digits. At the end of the dynamic range the HG digitiser goes into overflow and we swap to the LG digitizer.

In certain rare cases it can happen that the HG digitizer sends data beyond the overflow, or that we have LG digits without HG digits. In both cases the energy cannot be obtained reliably and the cell must be considered as error.

The double cell information must be filtered as above. This was not an issue for the commissioning phase but becomes an issue for the pilot beam.

shahor02:
@mfasDa the error in fullCI is genuine: apparently, your PR introduced a difference between the size of the cells vector and the sum of N channels referred by each trigger. Can you check?

mfasDa:
Ooops, sorry! Fixed

shahor02 (approved):

PR #7219, opened 2021-10-05
FV0: fixes for pilot beams

sbysiak:
- fix in digit reader which reads only 1st entry from the tree
- remove silly fatal logging
- fix name


jotwinow (approved):

mslupeck (approved):
Thanks @sbysiak for fixing these

sbysiak:
@sawenzel @shahor02 there is only one error in one o2 build – timeout in o2sim_G3 – which is rather unrelated to the changes. 
Is it possible to repeat the test?

shahor02 (approved):

PR #7220, opened 2021-10-05
TPC static distortion map creation

martenole:
- Add conversion from TPC cluster residuals to local binned residuals
- Outlier filtering for Run 3 data
- Skip intermediate conversion from `float` to `short`

@shahor02 with this one can in principle create a static distortion map for a single time frame. One just needs to uncomment a few lines in the TPCInterpolationSpec.cxx
I have some ideas how the collected local residuals could be send to an aggregator (which is not yet contained in this PR), but I would like to discuss with you in person. I will try to pass by your office tomorrow. I am for instance not sure if I should use the mechanism described in [here](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/Calibration/README.md) for time-interval based calibrations.


martenole:
Hi @shahor02 
in the `o2-tpc-scdcalib-interpolation-workflow` I am now measuring the cluster residuals, doing the outlier rejection and then sending the unbinned residuals to the residual aggregator which does the binning and writes the residuals to a file. I am not sure, if the way I am writing the residuals to a file is correct. Maybe we can discuss this quickly tomorrow? Right now the timeslot-based calibration is foreseen to process only one slot at a time and collect the residuals for about 1 minute. For each timeslot a new file is written.
After this is finalized I will add another workflow with a reader for the binned residuals and which creates then the maps.
Cheers,
Ole

shahor02:
@martenole sure, can discuss tomorrow. The time-slot calibration should eventually process the multiple slots, depending on arriving TF time it should fill it to corresponding slot

Detectors/TPC/calibration/SpacePoints/include/SpacePoints/TrackResiduals.h
```diff
@@ -654,7 +648,7 @@ inline void TrackResiduals::getVoxelCoordinates(int isec, int ix, int ip, int iz
   x = getX(ix);
   p = getY2X(ix, ip);
   z = getZ2X(iz);
-  if (isec >= SECTORSPERSIDE) {
+  if (isec < SECTORSPERSIDE) {
```
shahor02: Why, did you change the convention of A:0-17, C:18-35?

Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
```diff
@@ -0,0 +1,205 @@
...
+  // the previous slot is merged to this one and afterwards
+  // the previous one will be deleted
+  for (int iSec = 0; iSec < SECTORSPERSIDE * SIDES; ++iSec) {
+    // merge statistics
+    const auto& statVecPrev = prev->stats[iSec];
+    auto& statVec = stats[iSec];
+    for (int iVox = 0; iVox < trackResiduals->getNVoxelsPerSector(); ++iVox) {
+      const auto& statPrev = statVecPrev[iVox];
+      auto& stat = statVec[iVox];
+      float norm = 1. / (statPrev.nEntries + stat.nEntries);
```
shahor02: one should probably have a protection against division by 0 here.

shahor02 (commented):
Hi @martenole 

sorry for delay in reviewing this PR, see just 2 small comments below.

Detectors/TPC/calibration/SpacePoints/include/SpacePoints/TrackResiduals.h
martenole: Thanks a lot for spotting, this should not have been included in the PR

Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
martenole: Done

shahor02:
@martenole could you resolve conflicts, clang-format and remove WIP?

martenole:
Sure I can do that. I am just not yet finished with my tests with a larger data set from the pilot beam. So more changes will be probably be needed. But since this is already quite a large PR it makes sense to merge this sooner.

shahor02:
@martenole so, shall I merge it?

martenole:
@shahor02 yes please, thanks

PR #7224, opened 2021-10-06
DPL Analysis: add exception when grouping with improperly sorted index

aalkin:
If non-monotonous order detected when grouping, an exception is thrown, stating the name of the table being grouped, name of the index and offending values.


ktf:
Can you add a test that we actually do throw on an unsorted grouping?

jgrosseo (approved):

PR #7227, opened 2021-10-06
DPL: better defaults for the parallel processing support

ktf:


ktf:
@davidrohr @ironMann this should trigger the shallow copies only when strictly needed and supported.

davidrohr:
thx, this will be enough for checking

davidrohr:
OK, so I did some tests with this:
- The early forwarding in the FST now works as it should, so it is definitely an improvement.
- There is no early forwarding on the EPN at the readout-proxy level, even when I remove your 2 checks for `RAWDATA` and for `Optional`. Not sure why, isn't the readout proxy calling the forward function?. Anyway, so no problem at that level.
- I have added 2 workflows in a merged workflow both reading TPC RAWDATA. The RAWDATA check correctly avoids the forwarding (actually already the Optional check does for the GPU tracking). I am wondering, isn't the raw data marked as Optional on the FLP? If so, the early forwarding would have been disabled there as well.
- When I remove both your checks, I get plenty of messages forwarded which takes some time as expected. Having tested locally, I cannot judge whether this is the slowdown that hits the FLP or not, but it is plausible.

My proposal would be that we add a command line option `--enable-early-forward [0/1/2]` in addition:
- 0 = no forwarding at all.
- 1 = forwarding with the checks as currently implemented in this PR.
- 2 = as 1 but the RAWDATA check is disabled.
For now 0 should be the default, that will hopefully fix it on the FLP. We can raise the default to 1 at some time after the pilot beam (should actually also work on the FLP).

ktf:
> There is no early forwarding on the EPN at the readout-proxy level, even when I remove your 2 checks for RAWDATA and for Optional. Not sure why, isn't the readout proxy calling the forward function?. Anyway, so no problem at that level.

Readout proxy is the first device in the chain, so it does not forward, it creates the messages, which go through a different chain.

> I have added 2 workflows in a merged workflow both reading TPC RAWDATA. The RAWDATA check correctly avoids the forwarding (actually already the Optional check does for the GPU tracking). I am wondering, isn't the raw data marked as Optional on the FLP? If so, the early forwarding would have been disabled there as well.

IIRC, it's only on the EPN that the mechanism is in place, synchronized by the "timeframe is there" message. @shahor02 might know better.

> My proposal would be that we add a command line option --enable-early-forward [0/1/2] in addition:

Doing that right now. I suggest we still merge this (since it's green).

davidrohr:
Actually, I would not merge it as is before the pilot beam, since it will reanable the early forwarding by default, just to avoid that we again break something by accident.
I would only merge once we have the command line option, which is off by default.

wiechula:
Just for my understanding. This would mean that for the TPC processing of RAWDATA (e.g. pedestal data taking on single EPN) we would not profit from the early forwarding. Unless the proposed command line option is implemented and we use parameter `2`, right?

davidrohr:
> Just for my understanding. This would mean that for the TPC processing of RAWDATA (e.g. pedestal data taking on single EPN) we would not profit from the early forwarding. Unless the proposed command line option is implemented and we use parameter `2`, right?

yes, exactly. And for some transition time, it would also be disabled completely, unless command line parameter `1` is used.

davidrohr (approved):
thx, tried all modes and they work correctly

PR #7228, opened 2021-10-06
Match itstpcqc

chiarazampolli:
To run:

`o2-global-track-cluster-reader --track-types "TPC,ITS-TPC" | o2-itstpc-match-qc -b`

For now it does not implement the cyclic saving of histograms that we have in QC. 
To be done: attach a QC task that publishes the output to the QCG.


Detectors/GlobalTracking/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ o2_add_library(GlobalTracking
           src/MatchCosmics.cxx
           src/MatchCosmicsParams.cxx
           src/MatchGlobalFwd.cxx
+	  src/MatchITSTPCQC.cxx
+	  src/ITSTPCMatchingQCParams.cxx
```
shahor02: formatting again slipped

Detectors/GlobalTracking/CMakeLists.txt
```diff
@@ -39,11 +41,14 @@ o2_add_library(GlobalTracking
...
     O2::MFTTracking
-    O2::MCHTracking)
+    O2::MCHTracking
+    O2::MathUtils
+    O2::ReconstructionDataFormats)
 o2_target_root_dictionary(
   GlobalTracking
   HEADERS include/GlobalTracking/MatchTPCITSParams.h include/GlobalTracking/MatchGlobalFwd.h
-          include/GlobalTracking/MatchTOF.h include/GlobalTracking/MatchCosmics.h include/GlobalTracking/MatchCosmicsParams.h)
+  include/GlobalTracking/MatchTOF.h include/GlobalTracking/MatchCosmics.h include/GlobalTracking/MatchCosmicsParams.h include/GlobalTracking/MatchITSTPCQC.h include/GlobalTracking/ITSTPCMatchingQCParams.h)
```
shahor02: here too

Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
```diff
@@ -0,0 +1,174 @@
...
+bool MatchITSTPCQC::init()
+{
+
+  mPtTPC = new TH1F("mPtTPC", "Pt distribution of TPC tracks; Pt; dNdPt", 100, 0.f, 20.f);
+  mFractionITSTPCmatch = new TH1F("mFractionITSTPCmatch", "Fraction of ITSTPC matched tracks vs Pt; Pt; Eff", 100, 0.f, 20.f);
+  mPt = new TH1F("mPt", "Pt distribution of matched tracks; Pt; dNdPt", 100, 0.f, 20.f);
+  mEta = new TH1F("mEta", "Eta distribution of matched tracks; Eta; dNdEta", 100, -1.2f, 1.2f);
+  mChi2Matching = new TH1F("mChi2Matching", "Chi2 of matching; chi2", 200, 0, 20);
+  mChi2Refit = new TH1F("mChi2Refit", "Chi2 of refit; chi2", 200, 0, 20);
+  mTimeResVsPt = new TH2F("mTimeResVsPt", "Time resolution vs Pt; Pt; time res", 100, 0.f, 20.f, 100, 0.f, 2.f);
```
shahor02, line 66: Should not histos boundaries be defined by the cuts?

Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
```diff
@@ -0,0 +1,174 @@
...
+  if ((mSrc[GID::Source::ITSTPC] == 0 || mSrc[GID::Source::TPC] == 0)) {
+    LOG(fatal) << "We cannot do ITSTPC QC, some sources are missing, check sources in " << mSrc;
+  }
+
+  mDataRequest = std::make_shared<o2::globaltracking::DataRequest>();
+  mDataRequest->requestTracks(mSrc, mUseMC);
+
+  o2::base::GeometryManager::loadGeometry();
+  o2::base::Propagator::initFieldFromGRP();
+  o2::base::Propagator::Instance()->getNominalBz();
```
shahor02: This will create a problem in real processing, where the geometry and grp files are not in current dir. 
Usually, we set the paths via ConfigurableParam, e.g. 
```
-configKeyValues "NameConf.mDirGRP=/home/epn/odc/files;NameConf.mDirGeom=/home/epn/odc/files;NameConf.mDirCollContext=/home/epn/odc/files;NameConf.mDirMatLUT=/home/epn/odc/files;keyval.input_dir=/home/epn/odc/files;keyval.output_dir=/dev/null;HBFUtils.nHBFPerTF=128"
```
Doing this in QC seems to be impossible. 
BTW, why do you need field and geometry? Eventually, we will do some progatations for DCA checks etc., then it will be needed, but at the moment the seem not to be used.

shahor02 (changes requested):
There are some genuine errors in the failed CIs

Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
chiarazampolli: Hi Ruben, 
Actually it is my bad, I forgot then to initialize the Bz, sorry. But this is why I need it, because Bz is needed in the track selection, for the DCA cut. 
Yes, in fact I wanted to ask you. Shouldn't we read it from the CCDB?

shahor02: in general yes, but at the moment we are reading it from files, and I guess for MC we will be doing this (at least non-anchored ones). We should talk to @Barthelemy to allow at least ConfigurableParams in the qc.

Detectors/GlobalTracking/CMakeLists.txt
chiarazampolli: I honestly don't understand what I did wrong, I just used the formatting that was already there. I am anyway trying to improve it, not sure it is the correct way (is there some documentation? an example?)

Detectors/GlobalTracking/CMakeLists.txt
chiarazampolli: Done

Detectors/GlobalTracking/CMakeLists.txt
shahor02: I suspect your editor uses tabs, which are then substituted by spaces, producing a shift

chiarazampolli:
The error on o2-dataflow-cs8 seems not related to my PR:
```
The following tests FAILED:
	 32 - multinode_test (Timeout)
Errors while running CTest
## End sw/BUILD/QualityControl-latest/log
```

Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
chiarazampolli: For now I am using the NameConf for the GRP and geom file names. In QC (see https://github.com/AliceO2Group/QualityControl/pull/878), I will pass them in the json file, as the other parameters.

Detectors/GlobalTracking/CMakeLists.txt
chiarazampolli: You were right, I could not see with emacs. Now it is fixed.

shahor02 (approved):
failures unrelated, merging

PR #7233, opened 2021-10-07
[EMCAL-630] Shift cell time by 600 ns

mfasDa:
Shift cell time by 600 ns in order to store
time value approximately around the nominal
delay.


mfasDa:
We expect our main signal with a delay of ~600 ns from the trigger. We will store for the beam the cells moved close to the trigger in order to make best use of the data format. The calibration is designed in run 3 to calibrate the time with respect to 0 and no longer with respect to 600 ns as in run2.

shahor02 (approved):

shahor02:
ci failures unrelated.

PR #7236, opened 2021-10-07
DPL: speedup forwarding in case of split payload parts

ktf:


ktf:
@davidrohr @shahor02 @chiarazampolli this should speedup the dispatcher.

Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -1205,45 +1205,60 @@ bool DataProcessingDevice::tryDispatchComputation(DataProcessorContext& context,
...
+        // We need to find the forward route only for the first
+        // part of a split payload. All the others will use the same.
+        if (fdh->splitPayloadIndex == 0) {
+          cachedForwardingChoice = -1;
+          for (size_t fi = 0; fi < spec->forwards.size(); fi++) {
+            auto& forward = spec->forwards[fi];
+            if (DataSpecUtils::match(forward.matcher, dh->dataOrigin, dh->dataDescription, dh->subSpecification) == false || (dph->startTime % forward.maxTimeslices) != forward.timeslice) {
+              continue;
+            }
+            cachedForwardingChoice = fi;
```
shahor02, line 1241: should not one break here?

davidrohr:
This works, and #7235 also works.
#7235 has 30% less CPU load in the dispatcher compared to this one.

Framework/Core/src/DataProcessingDevice.cxx
ktf:
```suggestion
            cachedForwardingChoice = fi;
            break;
```

ktf: yes.

davidrohr (approved):

ktf:
Tested by @davidrohr and myself. Seems quite an improvement, so we merge it to get it in tomorrow build.

PR #7239, opened 2021-10-07
Fix CTP digit reader

shahor02:


shahor02:
some CIs are broken for unrelated reasons

PR #7246, opened 2021-10-08
DPL Analysis: fix issue with empty tables

ktf:
The mock writer was not doing the correct thing to calculate
the size of the empty tables since we have our own custom writer
for them.

This also gets rid of the extra space reserved when writing the tables
since we now calculate the size of the buffer upfront to the byte.


ktf:
@jgrosseo, @aalkin and myself found the issue with the empty table. Could you run the test train to see if it also fixes the extra rows?

ktf:
This was tested in Hyperloop, since the unit test in O2 pass, that should be enough. Merging.

PR #7247, opened 2021-10-08
suppress usage of Slewing correction for pp data

AllaMaevskaya:


shahor02:
trivial, merging to have it the nightly.

PR #7248, opened 2021-10-08
Workflow to write to file the MID digits decoded from raw data

dstocco:
Writing decoded data to file is useful for debugging.


aphecetche (approved):

Detectors/MUON/MID/Workflow/src/decoded-digits-writer-workflow.cxx
```diff
@@ -0,0 +1,49 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
```
aphecetche: @dstocco that's the old copyright. must be changed to new one...

aphecetche (changes requested):

dstocco:
All failures seem unrelated (multinode_test)

PR #7249, opened 2021-10-08
DPL: minor fixes in AOD table writer

aalkin:
* Remove std::move that prevented copy elisions
* Fix potential null pointer dereference
* Use .empty()
* Fix incorrect description string handling
* Do not copy strings in a loop
* Do not use else after `continue;`

### DIFF
--- a/Framework/Core/src/CommonDataProcessors.cxx
+++ b/Framework/Core/src/CommonDataProcessors.cxx
@@ -289,7 +289,7 @@ DataProcessorSpec
     std::map<uint64_t, uint64_t> tfNumbers;
 
     // this functor is called once per time frame
-    return std::move([dod, tfNumbers](ProcessingContext& pc) mutable -> void {
+    return [dod, tfNumbers](ProcessingContext& pc) mutable -> void {
       LOGP(DEBUG, "======== getGlobalAODSink::processing ==========");
       LOGP(DEBUG, " processing data set with {} entries", pc.inputs().size());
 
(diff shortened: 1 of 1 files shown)


Framework/Core/src/CommonDataProcessors.cxx
```diff
@@ -312,7 +312,7 @@ DataProcessorSpec
       // loop over the DataRefs which are contained in pc.inputs()
       for (const auto& ref : pc.inputs()) {
         if (!ref.spec) {
-          LOGP(DEBUG, "The input \"{}\" is not valid and will be skipped!", ref.spec->binding);
+          LOGP(DEBUG, "Invalid input will be skipped!");
```
ktf, line 315: Why dropping the actual input being skipped? I would find it useful.

Framework/Core/src/CommonDataProcessors.cxx
```diff
@@ -350,7 +350,8 @@ DataProcessorSpec
         if (!table->Validate().ok()) {
           LOGP(WARNING, "The table \"{}\" is not valid and will not be saved!", tableName);
           continue;
-        } else if (table->schema()->fields().empty() == true) {
+        }
+        if (table->schema()->fields().empty() == true) {
```
ktf:
```suggestion
        if (table->schema()->fields().empty()) {
```

Framework/Core/src/CommonDataProcessors.cxx
aalkin: It is a null pointer in this context.

ktf (approved):

PR #7250, opened 2021-10-08
fix in tof decoder wrong rdh

noferini:


noferini:
@preghenella @shahor02 
I added few checks in TOF decoder of compressed data (one in particular for a memcp) to avoid crash when something is not set properly in rdh.
Let me know if you have any comment.
Please consider this as urgent (this should fix many troubles we have in FLP and EPN).

shahor02:
can you run locally FST? Then I'll merge it once one CI is passed

noferini:
I am recompiling because I moved to a different branch.
I will run the test as soon as I compiled (hopying to be faster than checks)

noferini:
Since now it is required also O2sim I cannot run the test (let's wait for the official checks)
Then I will update my alidist...

shahor02:
@noferini  I run FST with your PR, looks ok, will merge before midnight so that it appears in the nightly

noferini:
Hi @shahor02, thanks!

shahor02:
tested locally, merging to have in next nightly

PR #7254, opened 2021-10-08
DPL: automatically create DDS properties for unresolved zeromq channels

ktf:


davidrohr (dismissed):
seems to work, should be merged tonight so we have it in the nightly and can test it tomorrow

PR #7260, opened 2021-10-10
Fixes for TRD matcher

shahor02:


martenole (approved):
Thanks a lot this looks good to me. I don't know why I did not cache the results of `recoCont->getSingleDetectorRefs()` before.

PR #7261, opened 2021-10-10
Fixes for TRD decoder

shahor02:
* Loop only over TRD data, send output once per TF
* Option to limit number of errors and warnings logged (default 20, 20)
* Some cleanup (more needed)

### DIFF
--- a/Detectors/TRD/reconstruction/src/DataReaderTask.cxx
+++ b/Detectors/TRD/reconstruction/src/DataReaderTask.cxx
@@ -18,14 +18,11 @@
 
 #include "Framework/ControlService.h"
 #include "Framework/ConfigParamRegistry.h"
-#include "Framework/RawDeviceService.h"
 #include "Framework/DeviceSpec.h"
 #include "Framework/DataSpecUtils.h"
 #include "Framework/InputRecordWalker.h"
 
 #include "DataFormatsTRD/Constants.h"
-
-#include <fairmq/FairMQDevice.h>
 #include <TH3F.h>
 #include "TH2F.h"
 #include "TFile.h"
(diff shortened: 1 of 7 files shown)


bazinski (dismissed):

bazinski:
@shahor02 Seems I chose the perfect afternoon/evening to drive back from italy :-D

shahor02:
Hi @bazinski 
I'll merge this PR in order to have it tomorrow on epns, but please check nevertheless. 
I see lot of unused lines in the code, some major cleanup would be welcome.

Here:  https://github.com/AliceO2Group/AliceO2/blob/0f92e528ff052ca8127569566f859d220a6db7ac/Detectors/TRD/reconstruction/src/CruRawReader.cxx#L551-L554 and https://github.com/AliceO2Group/AliceO2/blob/0f92e528ff052ca8127569566f859d220a6db7ac/Detectors/TRD/reconstruction/src/CruRawReader.cxx#L594-L596
there was a clear bug, leading to `Rejected: 18446744073708029992 bytes` messages in the printout. I've moved ` linkstart = linkend;` after the distance calculation, the numbers are reasonable now but please crosscheck. 
BTW, I see that `CruRawReader::mTotalDigitWordsRead` reported as `mReader.getWordsRead()` is always 0, while corresponding rejected words size is not...

bazinski:
That is weird that was fixed about 6 weeks ago, the rejected going to a huge number.... will check what happened.
The physics trigger vs calibration trigger I will need merged urgently tomorrow once i have tested it. That is the thing I mentioned on mattermost this afternoon, physics trigger never has a half chamber header.

As for clean up, ja its on the todo list, but need qc urgently.

Detectors/TRD/reconstruction/src/DataReaderTask.cxx
```diff
@@ -253,79 +250,53 @@ void DataReaderTask::run(ProcessingContext& pc)
...
-  uint64_t inputcounter = 0;
-  auto device = pc.services().get<o2::framework::RawDeviceService>().device();
-  auto outputRoutes = pc.services().get<o2::framework::RawDeviceService>().spec().outputs;
-  auto fairMQChannel = outputRoutes.at(0).channel;
-  /* loop over inputs routes */
-  for (auto iit = pc.inputs().begin(), iend = pc.inputs().end(); iit != iend; ++iit) {
-    if (!iit.isValid()) {
-      continue;
+
+  std::vector<InputSpec> sel{InputSpec{"filter", ConcreteDataTypeMatcher{"TRD", "RAWDATA"}}};
```
bazinski, line 254: The buffer creation was actually going to move init, but this way is cool as well.

bazinski (approved):

bazinski:
@shahor02 as for the merge, cool with me.

shahor02:
tested locally on FST and real raw data, merging.

PR #7267, opened 2021-10-11
DPL: allow Lifetime::Enumeration to repeat values

ktf:
Is now possible to specify the "repetitions" metadata for an InputSpec which uses Lifetime::Enumeration.
When the associated input gets expired, it will create "repetitions" entries with the same startTime. 
This allows things like reading a file in "repetitions" steps, rather than at once.


ktf:
@TimoWilken I think something is wrong with the CS8 test as it did not get picked up the whole day. Merging this as all the rest work.

PR #7268, opened 2021-10-11
DPL: improve notification of backpressure

ktf:


ktf:
@davidrohr as discussed. This does not take into account the time, yet, but should already reduce quite drastically the number of messages.

Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -967,11 +965,21 @@ void DataProcessingDevice::handleData(DataProcessorContext& context, InputChanne
...
+                info.backpressureNotified = true;
+                info.normalOpsNotified = false;
+              }
               policy.backpressure(info);
               break;
             case DataRelayer::Dropped:
             case DataRelayer::Invalid:
             case DataRelayer::WillRelay:
+              if (info.normalOpsNotified == false && info.backpressureNotified == true) {
+                LOGP(WARN, "Back to normal on channel {}.", info.channel->GetName());
```
davidrohr: I would actually make this info severity

ktf: Sure. Both or just going back to normal?

davidrohr: just going back to normal, the other is a real warning :)

ktf:
```suggestion
                LOGP(info, "Back to normal on channel {}.", info.channel->GetName());
```

PR #7272, opened 2021-10-11
DPL Analysis: ensure correct slicing of boolean, fixed and variable list columns

aalkin:
This fixes long-standing issue with slicing special column types by handling the view offset correctly: boolean, fixed lists and recently introduced variable lists. Test cases are updated.

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -157,6 +157,11 @@ struct unwrap<std::vector<T>> {
   using type = T;
 };
 
+template <>
+struct unwrap<bool> {
+  using type = char;
+};
+
 template <typename T>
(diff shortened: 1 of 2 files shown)


Framework/Core/include/Framework/ASoA.h
```diff
@@ -290,17 +296,20 @@ class ColumnIterator : ChunkingPolicy
   arrow::ChunkedArray const* mColumn;
   mutable int mFirstIndex;
   mutable int mCurrentChunk;
+  mutable int mOffset;
```
ktf, line 299: Fine for now, but maybe we should have a "slicing policy" in the long term.

ktf (approved):

ktf:
Checked also on hyperloop.

PR #7275, opened 2021-10-11
DPL: do not add condition-backend option if already there

ktf:


ktf:
@wiechula @chiarazampolli this should fix the issue with the duplicate condition-backend. Still not sure why that happens in the first place, but as a workaround it should be good enough.

ktf:
@TimoWilken again, the cs8 builders seem not to be picking up jobs...

ktf:
fullCI should be enough to validate this. Merging.

PR #7276, opened 2021-10-11
FT0 digit write standalone workflow

shahor02:


AllaMaevskaya (approved):
Thank you, we really need it

PR #7281, opened 2021-10-12
DeltaROF default should be 0 for CA tracker

mpuccio:
This will fix the combinatorics issue with the emulated data + sync reco.

@shahor02


davidrohr:
I don't fully understand what does this do? Does it just disable the TF tracking? Shall we merge it?

mpuccio:
Yes, by default disable the check across different ROFs, I would merge it as we should only require this for late iterations of the tracker

PR #7283, opened 2021-10-12
DPL GUI: show again driver metrics

ktf:


ktf:
Currently broken. Tested locally on my mac. Previous optimisation was clearly wrong.

PR #7285, opened 2021-10-12
DPL GUI: avoid looping over data

ktf:
The maximum number of points to show is given by mod, not by size,
which includes the points which were dropped from the ring buffer.


ktf:
Not really tested in CI. Works on my box and compiles on cc7 and macOS. Merging.

PR #7286, opened 2021-10-12
DPL: introduce the --dds-workflow-suffix option

ktf:
Workaround for now. It might be merged with --workflow-suffix at some
point when we generalise it from being analysis specific.

Yet it might still make sense to have two levels (one to decouple things
at DPL level, one to decouple DDS level).

### DIFF
--- a/Framework/Core/src/DDSConfigHelpers.cxx
+++ b/Framework/Core/src/DDSConfigHelpers.cxx
@@ -13,23 +13,12 @@
 #include <map>
 #include <iostream>
 #include <cstring>
+#include <regex>
 #include <fmt/format.h>
 #include <libgen.h>
 
 namespace o2::framework
 {
-
-std::string replaceFirstOccurrence(
-  std::string s,
-  const std::string& toReplace,
-  const std::string& replaceWith)
-{
-  std::size_t pos = s.find(toReplace);
-  if (pos == std::string::npos) {
-    return s;
-  }
-  return s.replace(pos, toReplace.length(), replaceWith);
-}
 
 struct ChannelProperties {
   std::string_view key;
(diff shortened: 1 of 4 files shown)


ktf:
@davidrohr the option needs to be called differently, but this should do for now.

davidrohr:
thx, will try!

davidrohr:
Could this be done such, that it is sufficient to pass it to the last workflow in a pipe, i.e.
```
o2-... | o2-... | o2-dpl-run --dds --dds-workflow-suffix foo
```

And, this generates a --dump-workflow-suffix in the dds command lines, which comes from replacing --dds with --dump. That should be fixed.

ktf:
Ok, i fixed the issue with `--dump-workflow-prefix`, however I cannot reproduce the one with the merged workflows. I.e.:

```
stage/bin/o2-testworkflows-sync-reconstruction-dummy |  stage/bin/o2-testworkflows-diamond-workflow | stage/bin/o2-dpl-run --dds-workflow-suffix _foo --dds
```

seems to produce what I would have expected. Am I missing something?

davidrohr (approved):
OK, this fully works now. No idea what was my problem before, was probably sitting in front of the screen... sorry for the noise.

davidrohr:
@Barthelemy : This PR failed due to QC multinode ctest failed. That happened before we switched to QC 1.33, and it stopped with 1.33. Any idea why it shows up again now?

ktf:
getting this in the nightly.

Barthelemy:
@davidrohr no idea, calling @knopers8 as expert of the multi_node test ;)

knopers8:
Apparently I am too late see the logs. My first guess would be that it was caused by the move from push/pull to pub/sub channels between machines, which is, in short, less reliable to transfer all messages in difficult conditions (like on an overwhelmed CI machine).
If this is the case, then perhaps we could make this test retry 1-2 times if it fails, or move to push/pull only for tests on CI.

PR #7287, opened 2021-10-12
Splitting of GRP object 

chiarazampolli:
Object to fill information from ECS, see https://alice.its.cern.ch/jira/browse/OCTRL-564


DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
```diff
@@ -0,0 +1,125 @@
...
+
+  /// print itself
+  void print() const;
+
+  static GRPECSObject* loadFrom(const std::string& grpecsFileName = "", const std::string& grpecsName = "GRPECS");
+
+ private:
+  timePoint mTimeStart = 0; ///< DAQ_time_start entry from DAQ logbook
+
+  uint32_t mNHBFPerTF = 256; /// Number of HBFrames per TF
```
shahor02: I think we can set the default to 128

macro/CreateGRPECSObject.C
```diff
@@ -0,0 +1,60 @@
...
+  grpecs.setDetsTrigger(detsTrigger);
+  grpecs.setRun(run);
+  grpecs.setDataPeriod(dataPeriod);
+
+  CcdbApi api;
+  api.init(ccdbPath);
+  std::map<std::string, std::string> metadata;
+  metadata["responsible"] = "ECS";
+  long ts = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+  api.storeAsTFileAny(&grpecs, "GRP/GRP/ECS", metadata, ts, ts + 60 * 60 * 24 * 365); // making it 1-year valid to be sure we have something
```
shahor02: I think the start of validity should be equal to `start` time (but in `ms`), I would also provide an end of validity, with the convention that <0 is interpreted as start + e.g. 10h, to cover the fill.

macro/CreateGRPECSObject.C
```diff
@@ -0,0 +1,60 @@
...
+  grpecs.setDetsReadOut(detsReadout);
+  grpecs.setDetsContinuousReadOut(detsContinuousRO);
+  grpecs.setDetsTrigger(detsTrigger);
+  grpecs.setRun(run);
+  grpecs.setDataPeriod(dataPeriod);
+
+  CcdbApi api;
+  api.init(ccdbPath);
+  std::map<std::string, std::string> metadata;
+  metadata["responsible"] = "ECS";
```
shahor02, line 58: For the run-specific objects we agreed to pass metadata runNumber

shahor02 (commented):
Thanks @chiarazampolli ! 
Please see a few comments below

macro/CreateGRPECSObject.C
shahor02: Also, I would store it in e.g. `GLO/Config/GRPECS`, we anyway have plenty of other global objects, better collect them in one place.

DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
chiarazampolli: done

macro/CreateGRPECSObject.C
chiarazampolli: done

chiarazampolli:
Hi @shahor02 , 

I moved this PR to WIP, I would add here the split to the MagField GRP object and some changes in the LHCIF object, to avoid too many PRs. 

Chiara

chiarazampolli:
Hello @shahor02 . 
I implemented the filling of the Bunch Filling information in the LHCIF part of GRP. I need to wait for  https://github.com/AliceO2Group/AliceO2/pull/7367/files to be merged, then I can fix the (anti)clockwise) beam indices. 
Last to be implemented is the workflow for filling the B field information. Then I think we have everything - to be reviewed, of course.
Chiara

chiarazampolli:
Hi @shahor02 , 

In this PR I added the change for the path of the MeanVertex.
BTW, what if I remove the [WIP] from this PR, and we use another one for the processing of the B field Data Points?

Chiara

shahor02:
@chiarazampolli thanks! BTW, the MeanVertexObject (and calibrator) should be eventually changed to account for the diamond slopes.
OK for removing WIP and eventually merging.

shahor02 (approved):
failures are unrelated.

PR #7291, opened 2021-10-12
Stop reporting 0xdeadbeef TFs after such N(D=5) consecutive TFs seen

shahor02:
The threshould can be changed via e.g. `--configKeyValues "VerbosityConfig.maxWarnDeadBeef=10"`
Healthy TF will reset the counter, so the deadbeef TF will be reported again if reappear


davidrohr:
tested locally + should be in tomorrows nightly, + cs8 and macos already passed, merging

PR #7295, opened 2021-10-13
Update Compressor.cxx

preghenella:
ciao @noferini 
I think this should fix the issues observed in the TOF decoder with bad RDH.
it is a problem generated by the compressor, which in case of fatal errors was not properly storing the size of the data in the RDH payload.


noferini (approved):
I checked and the fix works properly.
We recovered 40% of the TOF signal in the bad link

shahor02:
trivial lines swap, fullCI passed, merging

PR #7299, opened 2021-10-13
MCH: latest updates for mapping (both electronic and detector ones)

aphecetche:
Besides the electronic mapping changes (due to changes in cabling at
Pt2), we also change one thing in the detector mapping, as we learnt
that Run3 version of station2 quadrants have one less front end card.

@aferrero2707 this should fix the mapping issues in the decoder.


Detectors/MUON/MCH/Raw/ElecMap/src/CH10R.cxx
```diff
@@ -1299,183 +885,87 @@ void fillElec2DetCH10R(std::map<uint32_t, uint32_t>& e2d)
...
   add(e2d, 1020, 1134, 475, 3, 0);
   add(e2d, 1020, 1133, 475, 3, 1);
-  add(e2d, 1020, 0, 475, 3, 2);
-  add(e2d, 1020, 0, 475, 3, 3);
-  add(e2d, 1020, 0, 475, 3, 4);
   add(e2d, 1020, 1129, 475, 5, 0);
   add(e2d, 1020, 1128, 475, 5, 1);
-  add(e2d, 1020, 0, 475, 5, 2);
-  add(e2d, 1020, 0, 475, 5, 3);
-  add(e2d, 1020, 0, 475, 5, 4);
```
aferrero2707, line 1478: For curiosity: why there are so many lines removed from CH10R?

aferrero2707:
Ok for me.

Detectors/MUON/MCH/Raw/ElecMap/src/CH10R.cxx
aphecetche: I realised that at some point (or maybe even always ? ) my python script was generating code for all 5 DS, even where only 1,2,3 or 4 DS are actually present. Hence the zeros as 3rd parameter in the `add` calls above. I finally fixed that in order not to add extra useless code.

aferrero2707: Ok, got it.

aphecetche:
seems I have genuine errors here. 
I'm on it...

aphecetche:
Tests are helpful to developers... as long as they are executed ;-)

Problem was that locally I was using : `ctest -L mch -j 16` to run the MCH tests. 
But that command misses two tests that are disabled for `BUILD_TYPE Debug` (for performance reason). 

To really get _all_ MCH tests one _must_ specify the build configuration to ctest `ctest -C RelWithDebInfo -L mch -j 16`

The actual test failure (in one of the two tests that were not ran) was indeed genuine as the expectations were not changed in that (cathode based) test to account for the St2 removed FEE for Run>=3 (while they were correctly changed for the segmentation based test)

aphecetche:
@ktf @TimoWilken is there some problem with the `build/O2/o2-cs8` check ? it seems it's not been progressing for a very long time (or am I just being impatient because that fix is needed for the pilot beam ;-) ? )

TimoWilken:
Hi @aphecetche, there was a problem with one of the build machines, so the jobs have been moved to another builder.

aphecetche:
@TimoWilken thanks !

davidrohr:

PR #7300, opened 2021-10-13
MCH: prevent decoder crashes due to unknown link id error

aphecetche:
@aferrero2707 here's a proposal to make the decoder a bit more robust and avoid crashing on some common errors. 

In addition I've put a basic message map in place to avoid having too many messages. To be refined later on (e.g. we should also define what kind of detail we need in the messages for the info logger, i.e. on production, versus what is more debug and oriented towards developers only)

### DIFF
--- a/Detectors/MUON/MCH/Raw/Decoder/src/ErrorCodes.cxx
+++ b/Detectors/MUON/MCH/Raw/Decoder/src/ErrorCodes.cxx
@@ -1,0 +1,73 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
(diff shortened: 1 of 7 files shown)


Detectors/MUON/MCH/Raw/Decoder/src/ErrorCodes.cxx
```diff
@@ -0,0 +1,73 @@
...
+    append("Parity", msg);
+  }
+  if (ec & ErrorHammingCorrectable) {
+    append("Hamming Correctable", msg);
+  }
+  if (ec & ErrorHammingUncorrectable) {
+    append("Hamming Uncorrectable", msg);
+  }
+  if (ec & ErrorBadClusterSize) {
+    append("Cluster Size", msg);
```
aferrero2707, line 45: I would write "Bad Cluster Size" to be consistent will all other "Bad Something" error messages.

aferrero2707 (approved):
Apart from the suggestion for the "Bad Cluster Size" error message, it is ok for me.

aphecetche:
@ktf @TimoWilken is there some problem with the `build/O2/o2` check ? it seems it's not been progressing for a very long time (or am I just being impatient because that fix is needed for the pilot beam ;-) ? )

TimoWilken:
Hi @aphecetche, there was a problem with one of the build machines, so the jobs have been moved to another builder.

aphecetche:
@TimoWilken thanks !

davidrohr:

PR #7301, opened 2021-10-13
fix in tof geo indexing initialization

noferini:
urgent to fix memory leak affecting also QC

PR #7303, opened 2021-10-13
Separate initialization flags, or it can cause mem leak

chiarazampolli:
if you don't call o2::tof::base::Geo::Init

@noferini , @njacazio , @shahor02 , @davidrohr


chiarazampolli:
It is exactly the same as: https://github.com/AliceO2Group/AliceO2/pull/7301 (apart from clng-format), so i close this one.

PR #7306, opened 2021-10-13
DPL Analysis: generalized grouping

aalkin:
* Adds "sorted" property to index columns
* Adds generic slicing using Filtered mechanism to define groups
* Adds a test

Note: analysis data model in O2Physics will also need to be updated to mark the index columns in Produced tables that are sorted by construction.

@jgrosseo @ktf 
The issue is that this will work for unsorted grouping only when grouped arguments are already `Filtered` in process method signature, otherwise we cannot call it. Any suggestions?

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -1574,7 +1574,7 @@ constexpr auto is_binding_compatible_v()
 #define DECLARE_SOA_ARRAY_INDEX_COLUMN(_Name_, _Getter_, _Size_) DECLARE_SOA_ARRAY_INDEX_COLUMN_FULL(_Name_, _Getter_, int32_t, _Size_, _Name_##s, "")
 
 ///NORMAL
-#define DECLARE_SOA_INDEX_COLUMN_FULL(_Name_, _Getter_, _Type_, _Table_, _Suffix_)                                                \
+#define DECLARE_SOA_INDEX_COLUMN_FULL_BASE(_Name_, _Getter_, _Type_, _Table_, _Suffix_, _Sorted_)                                 \
   struct _Name_##Id : o2::soa::Column<_Type_, _Name_##Id> {                                                                       \
     static_assert(std::is_integral_v<_Type_>, "Index type must be integral");                                                     \
     static_assert((*_Suffix_ == '\0') || (*_Suffix_ == '_'), "Suffix has to begin with _");                                       \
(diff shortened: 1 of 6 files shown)


Framework/Core/include/Framework/ASoA.h
```diff
@@ -1641,7 +1642,9 @@ constexpr auto is_binding_compatible_v()
   static const o2::framework::expressions::BindingNode _Getter_##Id { "fIndex" #_Table_ _Suffix_, typeid(_Name_##Id).hash_code(), \
                                                                       o2::framework::expressions::selectArrowType<_Type_>() }
 
-#define DECLARE_SOA_INDEX_COLUMN(_Name_, _Getter_) DECLARE_SOA_INDEX_COLUMN_FULL(_Name_, _Getter_, int32_t, _Name_##s, "")
+#define DECLARE_SOA_INDEX_COLUMN_FULL(_Name_, _Getter_, _Type_, _Table_, _Suffix_) DECLARE_SOA_INDEX_COLUMN_FULL_BASE(_Name_, _Getter_, _Type_, _Table_, _Suffix_, false)
```
jgrosseo, line 1645: Maybe we need also SORTED_FULL ?

aalkin: The sole reason for the usage of `FULL` declaration was to have multiple index columns referring to the same table. I do not think there are cases where these need to be sorted.

Framework/Core/include/Framework/Kernels.h
```diff
@@ -25,6 +25,40 @@
...
+{
+  groups->resize(fullSize);
+  auto column = input->GetColumnByName(key);
+  int64_t row = 0;
+  for (auto iChunk = 0; iChunk < column->num_chunks(); ++iChunk) {
+    auto chunk = static_cast<arrow::NumericArray<typename detail::ConversionTraits<T>::ArrowType>>(column->chunk(iChunk)->data());
+    for (auto iElement = 0; iElement < chunk.length(); ++iElement) {
+      auto v = chunk.Value(iElement);
+      if (v >= 0) {
+        if (v >= groups->size()) {
```
ktf, line 46: This should probably be moved outside the if, no?

aalkin: It needs to be checked only for positive values, and for each of the positive values.

aalkin:
Closing in favor of different development.

PR #7307, opened 2021-10-13
Write MID masks to file in the format expected from DCS

dstocco:
This PR consists of three commits:
- the first one add to the mapping the information on which local board has a direct input from the Front-End electronics
- the second one introduces a class to convert the MID digits into a mask suitable for the reaodut electronics
- the third commit uses the first two to write a file that can be used by WinCC to setup the mask
Notice that such a mask should go to the ccdb at some point, but this will be done when the MID DCS can handle the communication with ccdb.
Text files are used for the time being.

### DIFF
--- a/Detectors/MUON/MID/Filtering/exe/mask-maker.cxx
+++ b/Detectors/MUON/MID/Filtering/exe/mask-maker.cxx
@@ -29,12 +29,13 @@
 #include "MIDRaw/ColumnDataToLocalBoard.h"
 #include "MIDFiltering/ChannelMasksHandler.h"
 #include "MIDFiltering/ChannelScalers.h"
+#include "MIDFiltering/ColumnDataMaskToROMask.h"
 #include "MIDFiltering/FetToDead.h"
 #include "MIDFiltering/MaskMaker.h"
 
 namespace po = boost::program_options;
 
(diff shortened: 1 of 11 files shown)


Detectors/MUON/MID/Filtering/exe/mask-maker.cxx
```diff
@@ -47,15 +48,17 @@ bool processScalers(const o2::mid::ChannelScalers& scalers, unsigned long nEvent
...
-  colToBoard.process(masks);
-  for (auto& mapIt : colToBoard.getData()) {
-    for (auto& board : mapIt.second) {
-      std::cout << board << std::endl;
-    }
+  auto roMasks = colMasksToRO.convert(masks);
+  for (auto& board : roMasks) {
+    std::cout << board << std::endl;
   }
+  std::cout << "\nMask file produced: " << outFilename << std::endl;
```
aphecetche, line 60: well, you're getting a bit ahead of yourself here, aren't you ? what if the write method fails to actually create a file ? ;-)

aphecetche (approved):

Detectors/MUON/MID/Filtering/exe/mask-maker.cxx
dstocco: Well, I can move it after the writing...but since it's difficult to have all green I will do it only if one of the checks fails ;) This said, the filename is in the current working directory, so, unless the user has no write permission, the file will be written..

PR #7308, opened 2021-10-13
--append-det-to-period N(Def=1) will add dets.names to LHCPeriod if their n<=N

shahor02:
Default  use-case:  runs with single detector will be registered under <LHCPeriod>_<DET>, while "global" runs: under <LHCPeriod>.


shahor02:
almost trivial and tested locally, merging to have it in the nightly

PR #7310, opened 2021-10-14
DPL: driver errors should exit with 1

ktf:


ktf:
@davidrohr

davidrohr (approved):
without compiling and running an actual test, looks good and should fix the reported issue.
Thx

ktf:
@TimoWilken there is definitely some issue with build/O2/o2-dataflow-cs8.

PR #7313, opened 2021-10-14
DPL: prevent OutputProxy from using 100% CPU usage when idle

ktf:
This is because the output proxy channel polled as readable and
therefore it kept awakening the event loop.

### DIFF
--- a/Framework/Core/src/DataProcessingDevice.cxx
+++ b/Framework/Core/src/DataProcessingDevice.cxx
@@ -457,6 +457,12 @@ void DataProcessingDevice::initPollers()
         LOGP(debug, "{} is to send data. Not polling.", channelName);
         continue;
       }
+
+      if (channelName.rfind("from_") != 0) {
+        LOGP(info, "{} is not a DPL socket. Not polling.", channelName);
+        continue;
+      }
+
       // We assume there is always a ZeroMQ socket behind.
       int zmq_fd = 0;
       size_t zmq_fd_len = sizeof(zmq_fd);


ktf:
@knopers8 this should fix the output-proxy going to 100% CPU utilisation before actually starting to receive data.

davidrohr:
This PR still breaks some tests:
```
91/98 Test #105: test_Framework_test_ExternalFairMQDeviceWorkflow ......***Timeout  30.05 sec
--
93/98 Test #107: test_Framework_test_ProcessorOptions ..................***Timeout  60.07 sec
```
What is the status?

ktf:
The error is genuine. I need to have a look at it.

iravasen:
@ktf is there any update here? Sorry for asking, just to know since our ITSClusterTask for QC was involved in this. Thank you.

ktf:
No, sorry, I will try to raise it in priority.

iravasen:
@ktf Hi, any time estimate for this PR?

ktf:
I am doing some work for CCDB now which takes precedence. Nevertheless, I will try to have another look at it to see if there is anything quick to fix.

iravasen:
@ktf do you intend to reopen it or in the meantime this became obsolete?

ktf:
I will keep this open, but I think I will have a better solution in a bit.

ktf:
I rechecked what was the status on this because I thought some other development would have solved the issues with this PR and indeed it did. The remaining problem was the conflicts, which I fixed. On my box output proxy has 0% CPU usage when idle (tested with `o2-dpl-raw-proxy --dataspec "foo:TST/C1/0" | stage/bin/o2-dpl-output-proxy --dataspec "foo:TST/C1/0"`) and all the tests now pass.

davidrohr:
thx, ok, can check tonight if it helps in FST + QC

ktf:
Yes, that would be great. I see the macOS-arm tests are passing, so it indeed means that any unittest which was failing before is now actually ok, so I have good hopes...

ktf:
Did you manage to test so that we can call this done as well, given this is also considered a blocker?

davidrohr:
does not break the FST, and in a quick test I could not see the output proxy spinning at 100%.
Though I must admit that I did not write down a reproducer command where I saw this, so not 100% sure the problem I reported originally is fixed. But worst case I'll complain again.

PR #7318, opened 2021-10-15
DPL: add metric for processed timeslices in Hz

ktf:
Calculated over a ~5s interval.

### DIFF
--- a/Framework/Core/include/Framework/DataProcessingStats.h
+++ b/Framework/Core/include/Framework/DataProcessingStats.h
@@ -40,6 +40,10 @@ struct DataProcessingStats {
   std::atomic<uint64_t> lastSlowMetricSentTimestamp = 0; /// The timestamp of the last time we sent slow metrics
   std::atomic<uint64_t> lastMetricFlushedTimestamp = 0;  /// The timestamp of the last time we actually flushed metrics
   std::atomic<uint64_t> beginIterationTimestamp = 0;     /// The timestamp of when the current ConditionalRun was started
+
+  std::atomic<uint64_t> performedComputations = 0;             // The number of computations which have completed so far
+  std::atomic<uint64_t> lastReportedPerformedComputations = 0; // The number of computations which have completed until lastSlowMetricSentTimestamp
+
   InputLatency lastLatency = {0, 0};
 
   std::atomic<int> relayerState[MAX_RELAYER_STATES];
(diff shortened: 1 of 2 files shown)


ktf:
@davidrohr

davidrohr:
Thx, actually I would make it even 10 seconds average, or 20.
Just to understand what this does? It counds how often the doProcessing function is called?
And what happens with pipelined processes? Are their values summed up at the monitoring level?

ktf:
5 seconds is the current delay for sending out all of those kind of metrics. I can make it 10 but I need to check that does not create issues for the analysis (which uses the same mechanism to do the rate limiting).

ktf:
> Just to understand what this does? It counds how often the doProcessing function is called?

Yes, and it keeps track of the last count reported and when it was reported.

> And what happens with pipelined processes? Are their values summed up at the monitoring level?

Not yet. In the DebugGUI you will see two different metrics and the same on graphana, I guess. @awegrzyn might know better. In principle we could aggregate them in the DebugGUI, if you'd find that useful (maybe not immediately, though).

ktf:
(bad rebase fixed).

awegrzyn:
All kind of processing is possible after metrics are reported, I don't know what "pipelined processes" are.

davidrohr:
@awegrzyn : There are multiple instances of the same process running on the EPN to handle the load. Their ids are suffixed with `_t0`, `_t1`, ...
Ideally, the values reported by them should be summed up in Grafane.
And actually, looking in the plots at http://alio2-cr1-hv-aliecs.cern.ch:3000/d/GVh_uHlGz/dpl?orgId=1&refresh=5s&var-Hostname=epn002.internal
I see only one curve for `gpu-reconstruction` even though there are 4 instances running. So either they are already summed up, or it is showing only one of them?

ktf:
They are multiple instances of the same code, handling different timeframes. Do you keep track of the process name that is sending the metric?

awegrzyn:
okay, understood, would it be possible to tag each instance with `t0`, `t1`, .. suffix? This could be done by `.addTag(tags::Key::Id, 0)` or `addGlobalTag("instance", "t0")`

For DPL I have only `dataprocessor_id` so atm I think Grafana just shows all instances on same curve.

ktf:
Yes, sure. I can make dataprocessor_id the actual unique id and dataprocessor_name the generic name.

ktf:
I assume I can have more than one global tag?

awegrzyn:
> I assume I can have more than one global tag?

Yes of course.

ktf:
Ok, done.

davidrohr:
Thx, I assume there is no way to test it, except for deploying it on the EPN tomorrow?

ktf:
You can check in the DebugGUI:

image

davidrohr:

PR #7322, opened 2021-10-15
[EMCAL-728] Reduce verbosity of the error logging

mfasDa:
Errors related to raw channels (HW channels, DDL,
cell ID are cached in a std::unordered_map. In case
the same message has already been printed before
the message is suppressed in all further occurrences
and instead the number of occurrences is counted.

In addition option to switch off merging of high gain/
low gain cells, this is particular of relevance for MC
where only either of the two channel types are stored,
resulting in a large amount of raw data errors.

### DIFF
--- a/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
@@ -84,6 +84,8 @@ void RawToCellConverterSpec::init(framework::InitContext& ctx)
   mMaxErrorMessages = ctx.options().get<int>("maxmessage");
   LOG(INFO) << "Suppressing error messages after " << mMaxErrorMessages << " messages";
 
+  mMergeLGHG = ctx.options().get<bool>("mergeHGLG");
+
   mRawFitter->setAmpCut(mNoiseThreshold);
   mRawFitter->setL1Phase(0.);
 }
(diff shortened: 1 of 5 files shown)


mfasDa:
@davidrohr @shahor02 According to the e-mail discussion I have adjusted the logging of the Raw to cell converter for the error handling. I have tried to follow Ruben's suggestion to print the error message per link only once in case of multiple occurrences. I handle this internally with a std::unordered_map. We need the error message however either per HW channel, per tower or per DDL, depending on the type of the error. Also, if the channel raises different errors we print all of them and suppress errors in case they are the same. This is simply necessary for the on-call to know in order to mask faulty channels once they appear. Still it should catch the dominant amount of repetitions and lead to a sizable reduction of verbosity in the infoBrowser, because in case a channel is bad it will send repetitively faulty data leading to the same error message. You currently see this in the infoBrowser: The message flood comes from a few channels with always the same message and can be trivially suppressed.

mfasDa:
@davidrohr For MC I have added an option to switch off the merging of HG and LG cells, because in order to simplify the situation at simulation level and in order to not blow up the data volume for digits we currently decide at digits level to store either the low or the high gain digits, but not both. With the LG/HG merger - which is ultimately needed for data - this leads to a huge amount of error messages, because in case of data we expect the data to come from both digitizers and treat cases where only one of the two are present as error case. The HG/LG merger can be switched off via the option `--mergeHGLG` (boolean, default=true). Can the treatment for MC be somehow included in the scipts?

mfasDa:
I guess the current errors come from the network glitch at CERN. Does the CI need to be restarted once the situation at CERN resolves?

davidrohr:
checked locally, works for me, merging to have it in the nightly (all CIs failed for a different reason)

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -501,5 +510,6 @@ o2::framework::DataProcessorSpec o2::emcal::reco_workflow::getRawToCellConverter
                                           o2::framework::Options{
                                             {"fitmethod", o2::framework::VariantType::String, "gamma2", {"Fit method (standard or gamma2)"}},
                                             {"maxmessage", o2::framework::VariantType::Int, 100, {"Max. amout of error messages to be displayed"}},
-                                            {"printtrailer", o2::framework::VariantType::Bool, false, {"Print RCU trailer (for debugging)"}}}};
+                                            {"printtrailer", o2::framework::VariantType::Bool, false, {"Print RCU trailer (for debugging)"}},
+                                            {"mergeHGLG", o2::framework::VariantType::Bool, true, {"Merge HG and LG channels for same tower"}}}};
```
davidrohr, line 514: @mfasDa : unfortunately it doesn't work like this. As far as I know, one cannot set a boolean to false on the command line. You would need to invert the option and make the default false, then I can set it to true.

PR #7325, opened 2021-10-15
Fix for entropy encoder buffer overflow

shahor02:
This is a least invasive and tested locally fix, need to be deployed on EPNs since currently hight-rate EMCAL CTFs are mostly corrupted. More flexible but complex solution will be pushed later with full check.


shahor02:
merging to have it in the nightly, tested locally

PR #7331, opened 2021-10-16
Prepend MCH local config and debug options by mch- to avoid conflict

shahor02:
DPL workflow building fails if the local device option has the same name as the global option of (any) workflow.
This is a quick fix until this unnecessary conflict will be resolved

@aphecetche merging this since otherwise it is impossible to have a workflow including MCH reconstruction and any QC at the same time (needed for Sunday's TED shots)


aphecetche:
@shahor02 I understand this is just temporary, right ? Is there a JIRA ticket or something where a proper fix is being followed-up ?

shahor02:
@aphecetche theoretically it is temporary, but the problem exists since ever and we tacitly avoid using generic names like `config` or `debug` in particular workflows. As far as I know, there is no currently work ongoing to solve this, so no jira...
But this is indeed very annoying, should be resolved.

PR #7338, opened 2021-10-17
DPL GUI: make some progress towards remote gui integration

ktf:


ktf:
Compiles and not really active yet. Merging.

PR #7339, opened 2021-10-17
o2-eve refactoring

jmyrcha:
- screenshot
- handling save folder
- fixing numbering events
- adding thread safety to events
- full geometry paths in simple geometry specs


pzhristov:
The failure on the o2 builder is a timeout in one of the tests.
The failure on cs8 is not expected - it looks like an old alidist was used where the LLVM_ROOT is not propagated to the arrow build.

PR #7343, opened 2021-10-18
Add ZDC table to the AOD producer

nburmaso:
Hi @sawenzel, @shahor02. While the reader problem is not resolved yet (related to some issue in the reader, which I contacted Ruben about; see log attached), I'm putting the code I prepared for the ZDC table in this draft PR to have it ready. 

[test.log](https://github.com/AliceO2Group/AliceO2/files/7363185/test.log)


Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
```diff
@@ -211,6 +212,10 @@ class AODProducerWorkflowDPL : public Task
   std::unordered_map<GIndex, int> mGIDToTableID;
   int mTableTrID{0};
 
+  // zdc helper maps to avoid a number of "if" statements
+  map<string, float> mZDCEnergyMap;
+  map<int, float> mZDCTDCMap;
```
sawenzel: could you add what is being mapped to what in the comments?
Is `map` better than `unordered_map` ?

Detectors/AOD/CMakeLists.txt
```diff
@@ -47,15 +48,16 @@ o2_add_executable(
...
+        O2::FT0Reconstruction
+        O2::TPCFastTransformation
+        O2::TPCBase
+        O2::TPCReconstruction
+        O2::TOFBase
+        O2::TOFCalibration
+        O2::SimConfig
+        O2::DataFormatsFT0
+        O2::Steer
+        O2::ZDCBase
```
sawenzel, line 69: Is this global change necessary?

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
nburmaso: Hi @sawenzel. I have added some comments for these maps. Concerning performance of `map` and `unordered_map`, I think in this case there should be no real differences, since these maps will consist of a small number of key-value pairs. Here, keys correspond to ZDC sub-parts and TDC channels, so their number is small indeed.  Perhaps, performance-wise it would be better to simply use if-clauses, but there will be a lot of them :)

Detectors/AOD/CMakeLists.txt
nburmaso: Sorry, it is a typo. I just added `ZDCBase`, but accidentally indented some lines.

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
cortesep: Hello,
I'm from the ZDC team. To me it looks a bit strange to have a map that uses signal TDC ids and another that uses signal names. One should probably unify the approach and use signal names for both maps. What do you think?

nburmaso: Hi @cortesep. Thank you for your comment. I will try to clean it up a bit.

sawenzel (dismissed):

sawenzel:
@nburmaso : I thought this was merged already but apparently the auto-merge feature didn't go through. Unfortunately, we now have merge conflicts again. Could you please take a look?

nburmaso:
Hi @sawenzel. I have resolved the conflicts and cleaned up the code a bit.

sawenzel (approved):

PR #7344, opened 2021-10-18
Flag for background particles

jgrosseo:


jgrosseo:
@sawenzel What about using this flag 0x2 for now?

sawenzel (approved):

PR #7345, opened 2021-10-18
Revert "[EMCAL-728] Reduce verbosity of the error logging"

mfasDa:
- Revert "[EMCAL-728] Reduce verbosity of the error logging"
- Suppress gain-related errors after the max. amount of messages is reached
- Reset error counters after 10 minutes

### DIFF
--- a/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
@@ -84,6 +84,10 @@ void RawToCellConverterSpec::init(framework::InitContext& ctx)
   mMaxErrorMessages = ctx.options().get<int>("maxmessage");
   LOG(INFO) << "Suppressing error messages after " << mMaxErrorMessages << " messages";
 
+  mMergeLGHG = ctx.options().get<bool>("mergeHGLG");
+
+  LOG(INFO) << "Running gain merging mode: " << (mMergeLGHG ? "yes" : "no") << std::endl;
+
   mRawFitter->setAmpCut(mNoiseThreshold);
   mRawFitter->setL1Phase(0.);
 }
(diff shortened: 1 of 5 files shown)


Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -510,6 +501,5 @@ o2 ::framework::DataProcessorSpec o2::emcal::reco_workflow::getRawToCellConverte
                                           o2::framework::Options{
                                             {"fitmethod", o2::framework::VariantType::String, "gamma2", {"Fit method (standard or gamma2)"}},
                                             {"maxmessage", o2::framework::VariantType::Int, 100, {"Max. amout of error messages to be displayed"}},
-                                            {"printtrailer", o2::framework::VariantType::Bool, false, {"Print RCU trailer (for debugging)"}},
-                                            {"mergeHGLG", o2::framework::VariantType::Bool, true, {"Merge HG and LG channels for same tower"}}}};
```
davidrohr: could you please only revert the logging, not the mergeHGLG?

mfasDa: Of course, sorry!

mfasDa:
@davidrohr nomerge mode is re-established.

mfasDa:
CS8 error seems to be unrelated

davidrohr:
ok, will try this later and merge it for the next nightly

davidrohr (approved):
ok for now, we have to see whether 100 messages per processes is not too much...
this can still lead to severel 100k messages per hour.

davidrohr:
had to squash-merge, since rebase-merge didn't work cleanly.
But this should go into the nightly so EMCAL can use it tomorrow.

PR #7346, opened 2021-10-18
Assign MCCollID -1 for PVs with undefined labels [O2-2631]

benedikt-voelkel:
PVs made from tracks of mostly noise hits have labels with source and
event ID of -1. In that case assign MCCollID of -1 as well

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -1169,7 +1169,7 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   for (auto& label : primVerLabels) {
     auto it = std::find_if(mcColToEvSrc.begin(), mcColToEvSrc.end(),
                            [&label](const std::pair<int, int>& item) { return (item.first == label.getEventID() && item.second == label.getSourceID()); });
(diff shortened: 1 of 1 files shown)


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1169,7 +1169,7 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   for (auto& label : primVerLabels) {
     auto it = std::find_if(mcColToEvSrc.begin(), mcColToEvSrc.end(),
                            [&label](const std::pair<int, int>& item) { return (item.first == label.getEventID() && item.second == label.getSourceID()); });
-    int32_t mcCollisionID = it - mcColToEvSrc.begin();
+    int32_t mcCollisionID = (it != mcColToEvSrc.end()) ? it - mcColToEvSrc.begin() : -1;
```
sawenzel, line 1172: @aalkin : Would we need to put all such unassigned MC collision labels also to the front/beginning of the table ...  or is that not necessary here?

aalkin: No, the MCCollisionLabels should have exactly the same order, as Collisions, there it no need to alter it.

sawenzel:
Tested manually.

PR #7178, opened 2021-10-01
DPL: add metric for consumed timeframes

ktf:
In order for this to work correctly a few things are needed.

* Outputs which is are not created on a timeframe basis needs to be marked
  as Lifetime::QA
* For all the timeframe sinks, we produce automatically a message DPL/SUMMARY/<unique id for a given sink>
  which marks that (part of) a timeframe has reached that sink.
* All those DPL/SUMMARY outputs are garbage collected by the
  dummy-sink, since they are dangling. Whenever all of them arrives, we
  increase the counter of consumed timeframes.
* Notice that in order to avoid accumulating stuff on the dummy sink,
  we introduced a new CompletionOp::ConsumeExisting which is like process,
  but it will drop the payload (but not the header) of a given O2 message pair.

PR #7349, opened 2021-10-18
DPL: baby-steps towards reporting timeframes consumed

ktf:


davidrohr:
In contrast to #7178 : this PR seems to work. FST behaves as before.

ktf:
Ok, good. Indeed this one should be harmless and we can merge it once the rest of the tests pass. I will prepare now the one which has the real behavioural change.

PR #7351, opened 2021-10-18
Proposal to attach track statistics in MCHeader

sawenzel:
https://alice.its.cern.ch/jira/browse/O2-2646

### DIFF
--- a/run/O2HitMerger.h
+++ b/run/O2HitMerger.h
@@ -65,6 +65,7 @@
 #include <csignal>
 #include <mutex>
 #include <filesystem>
+#include <functional>
 
 #include "SimPublishChannelHelper.h"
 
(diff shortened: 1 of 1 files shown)


sawenzel:
@chiarazampolli : This is a proposal concerning https://alice.its.cern.ch/jira/browse/O2-2646.
Since the MCEventHeader class already allows the possibility to attach arbitrary information (via key - value store), my proposal would be not to add further specific API/data members to the class.

For the user it means to access information as:
```
eventheader->getInfo<int>("prims_eta_1.2");
```

rather than

```
eventheader->getNPrimsSmallerEta1_2();
```
(The second one needs schema/class evolution while the first one just needs some naming convention and we could attach many more metadata of similar nature). Would that be ok for you?

sawenzel:
@preghenella

chiarazampolli:
Ciao Sandro, 
I think it will work fine, yes, thanks!
So you will fill the values of the keys during simulation, I guess. Right?
Chiara

shahor02 (approved):

PR #7353, opened 2021-10-18
Guarantee order of PWG items in Data Model menu

pbuehler:


pbuehler:
@jgrosseo The order of the PWG items in the Data Model menu is guaranteed to follow the order the PWGs are listed in the inputCard.xml.

jgrosseo (approved):

PR #7354, opened 2021-10-18
[WIP][EMCAL-728] Error handler for cached error logging

mfasDa:
Log only the first occurrence of an error type,
and count number of subsequent errors of the
same type per channel. Using linear containers
in dimension DDL and hwaddress for direct access
to channel ID. Providing error stats at the end
of the run.

The raw error handler is included also in the
RawToCellConverter and used for all Channel / DDL
based errors. Only the rare errors corresponding
to an invalid cell ID are suppressed after their
first occurrence by a dedicated handler using
std::unordered_set.

### DIFF
--- a/Detectors/EMCAL/reconstruction/CMakeLists.txt
+++ b/Detectors/EMCAL/reconstruction/CMakeLists.txt
@@ -13,6 +13,7 @@ o2_add_library(EMCALReconstruction
                SOURCES src/RawReaderMemory.cxx
                        src/RawBuffer.cxx
                        src/RawHeaderStream.cxx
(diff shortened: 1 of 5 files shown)


mfasDa:
WIP - please don't merge before the pilot beam!

This is a new version of the error message handling more similar to the approach used in the ITSMFT code: Instead of a std::unordered_map a 3D set of array (Error type, DDL, hwaddress) is used to cache the errors, resulting in O(1) complexity. Only the rare cell ID error which cannot be covered with the fixed array sizes is handled by a separate std::unordered_set.

mfasDa:
Up

PR #7355, opened 2021-10-18
[EMCAL-507] Setting default files for objects handled by ROOTTreeReader

mfasDa:


mfasDa:
blocked by https://github.com/AliceO2Group/QualityControl/pull/917

shahor02 (changes requested):
@mfasDa the CI errors are genuine

mfasDa:
@shahor02 Yes, this is known. As stated above we had a dedicated qc workflow for EMCAL for initial development due to problems with workflow chaining in the early time, which is now legacy and no longer used as chaining works smoothly, but depends on the publisher interface, which is changed by this commit. The change that removes the dedicated EMCAL QC workflow in QualityControl is already merged but not yet tagged. @Barthelemy for reference. I expect new tags will only come after the pilot beam, not sure how urgent this change is.

mfasDa:
@shahor02 The commit in QualityControl is tagged now so this PR can be processed. I see the only test failing is the fullCI, which is unrelated to the commit. I think this can be merged now.

shahor02 (approved):

PR #7359, opened 2021-10-18
DPL: provide unique DDS id as --id when running with --dds

ktf:
This makes sure the XML is generated with the template macro
we then need to make sure FairMQ sees a unique id across the whole
DDS deployment.


ktf:
@AndreyLebedev @davidrohr as discussed...

davidrohr:
@ktf : I think this works only with standalone workflows but not with merged workflows.
`o2-dpl-raw-proxy --dds` does what it should but `o2-dpl-raw-proxy | o2-dpl-run --dds` does not work.

ktf:
Ok, should be fixed.

davidrohr:
This seems to work correctly now, at least looking at the XML file.
Still need to test the generated XML on the EPNs.

davidrohr (approved):
works on the EPN, should go in tomorrows nightly

PR #7363, opened 2021-10-19
DPL: add metric for shm memory usage

ktf:


ktf:
@davidrohr as discussed.

Framework/Core/include/Framework/DataProcessingStats.h
```diff
@@ -37,6 +37,7 @@ struct DataProcessingStats {
   std::atomic<int> totalProcessedSize = 0;
   std::atomic<int> totalSigusr1 = 0;
   std::atomic<int> consumedTimeframes = 0;
+  std::atomic<int> availableManagedShm = 0;
```
davidrohr: Is this in bytes or in GB? If it is bytes, the `int` will overflow

Framework/Core/src/CommonServices.cxx
```diff
@@ -472,6 +479,7 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
   monitoring.send(Metric{(stats.lastProcessedSize / (stats.lastLatency.maxLatency ? stats.lastLatency.maxLatency : 1) / 1000), "input_rate_mb_s"}
                     .addTag(Key::Subsystem, Value::DPL));
   monitoring.send(Metric{((float)performedComputationsSinceLastUpdate / (float)timeSinceLastUpdate) * 1000, "processing_rate_hz"}.addTag(Key::Subsystem, Value::DPL));
+  monitoring.send(Metric{(uint64_t)stats.availableManagedShm, "available_managed_shm"}.addTag(Key::Subsystem, Value::DPL));
```
davidrohr: could we appent `_[segment_id]` here, in the end the monitoring will receive both, and we need to distinguish.

davidrohr:
And then it actually doesn't work, I get an error like
```
[21:29:29][ERROR] Could not find management segment for shmid '17a7d816'
```
followed by an exception
I think the wrong shmid is computed:
```
[O2PDPSuite/latest] ~/alitest/tmp-small $> ls /dev/shm/
com.freerdp.xfreerdp.tsmf_000000000000923F  fmq_eca859e3_cv  fmq_eca859e3_m_0  fmq_eca859e3_mng  sem.fmq_eca859e3_mtx
```

davidrohr:
Ah, and one last point, can we enable it only for the readout-proxy, or enable it via command line option / env variable?
Just to reduce the metric load a bit.

Framework/Core/src/CommonServices.cxx
ktf: Shouldn't we distinguish based on the dataprocessor_id global tag?

ktf:
What did you use as session id?

Framework/Core/src/CommonServices.cxx
davidrohr: When we have 2 NUMA domains, with 2 DPL chains, don't we have each dataprocessor_id twice? Only this id we put in the XML will be unique.

davidrohr:
> What did you use as session id?

`--session default`. However, I think DDS somehow overwrite the session via the odc plugin. It is always dífferent, even though I always provide `--session default`. Can't we get the session that is used from FairMQ?

ktf:
If the property is set, sure. This might also affect memory cleanup...

davidrohr:
> If the property is set, sure. This might also affect memory cleanup...

OK, then we should try to get it from there. And perhaps we should put the whole complex in a `try` / `catch` to avoid that an exception kills us.

ktf:
Ok. I have updated this. I tested it and it still works correctly without DDS.

Framework/Core/src/CommonServices.cxx
```diff
@@ -438,6 +440,17 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
...
   }
+  // Derive the amount of shared memory used
+  auto& runningWorkflow = registry.get<RunningWorkflowInfo const>();
+  using namespace fair::mq::shmem;
+  auto& spec = registry.get<DeviceSpec const>();
+
+  // FIXME: Ugly, but we do it only every 5 seconds...
+  if (spec.name == "readout-proxy") {
+    auto device = registry.get<RawDeviceService>().device();
+    stats.availableManagedShm.store(Monitor::GetFreeMemory(SessionId{device->fConfig->GetProperty<std::string>("session")}, runningWorkflow.shmSegmentId));
```
davidrohr: Now it works perfectly. But can you put a try {...} catch {} around this. I want to make sure we don't create a build tonight that will crash on the EPN due to exception.

davidrohr:
tested on the EPN, merging

PR #7364, opened 2021-10-19
DPL: set run number in start rather than in init

ktf:


ktf:
Updated to fix the warning / error.

davidrohr (commented):
Why don't you do both? If it is already avaialble at init time, messages at that time will already be flagged correctly. Then you overwrite it at start.

vascobarroso:
@davidrohr the run number is only available at start, not before.

davidrohr:
ah, ok, that makes sense. I though it might be different between EPN and FLP, but it is by design...

PR #7366, opened 2021-10-19
[EMCAL-612] Add option to disable decoding errors in cell writer

mfasDa:
Needed for converting CTF data into cell trees.


davidrohr (approved):

mfasDa:
For the record: Decoding errors exist only online for sending to the QC where a dedicated task (currently in development) will provide several histograms with error statistics (error type per DDL link, finer granularity, ...). The errors are exported because the raw QC itself sees only ~ 1% of the statistics, while we want to monitor all of them. Thus they are obviously not exported to the CTF file, and when streaming CTF content to the cell writer the channel needs to be excluded, otherwise the workflow won't start due to the missing channel `EMC/DECODERERR/0`.

mfasDa:
CS8 CI error unrelated

shahor02 (approved):
@mfasDa the option does not hurt, so merging. But I don't understand its point. To extract cells from the CTF to the root tree you don't need to go via `emcal-reco-workflow`, it is enough to do
```
o2-ctf-reader ... | o2-emcal-cell-writer-workflow
```

PR #7372, opened 2021-10-20
DPL: use @ as default IPC path on linux

ktf:
This will make sure that the pipes do not exists as actual files on disk.

### DIFF
--- a/Framework/Core/src/WorkflowCustomizationHelpers.cxx
+++ b/Framework/Core/src/WorkflowCustomizationHelpers.cxx
@@ -18,12 +18,16 @@ namespace
 {
 std::string defaultIPCFolder()
 {
+#ifdef __linux__
+  return "@";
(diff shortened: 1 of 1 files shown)


ktf:
@ironMann can you check if this works as expected for you?

ktf:
@ironMann @davidrohr This seems to work now. Any objections to it?

davidrohr:
Makes sense to me if it works

ktf:
The only thing which could be a regression is if we have a name conflict between the channels which is disambiguated by running in different folders. With this it will not work anymore. Do we do anything like that?

ktf:
Of course we could also wait for after the pilot beam, just to be on the safe side.

davidrohr:
Well, i think in most places on the epn we already have @ in the name.

Will that work actually?

PR #7374, opened 2021-10-20
add to ctf-writer stop() with same functionality as EoS()

shahor02:
Includes https://github.com/AliceO2Group/AliceO2/pull/7373, should be merged after it.


ktf:
Actually this one would be better, because you have the proper renaming of stop() to finalise() in the CTFWriter. I am closing #7373.

PR #7377, opened 2021-10-20
[QC-667] Add detector to logging in merger

Barthelemy:


knopers8 (approved):

Barthelemy:
@ktf @TimoWilken Could you force merge it ? and what is going on with cs8 ?

TimoWilken:
The CS8 problem should be fixed soon (there's a PR in alidist, but I'm just waiting for some Jenkins builds to finish so that merging the fix won't cause huge delays for all the builders).

PR #7378, opened 2021-10-20
DPL Analysis: Generic grouping

aalkin:
@jgrosseo 
Discussing this with @ktf, we've found a minimal workaround: `SmallGroups<>` is defined as a simple alias to `Filtered` and thus the arguments of process function are always compatible when generic grouping is used.

Now back to marking columns sorted and not sorted, what is the proposed scheme? Assume that default macro always declares sorted index, and require `UNSORTED` to be explicit?


ktf:
This is fine per se, but when I run on hyperloop I got a stuck workflow with this so I am trying to understand what is going on.

ktf:
Ok, checked again and it works. Maybe I had something else also on top of the build.

PR #7379, opened 2021-10-20
TRD fix non zerosuppressed data and sending qc messages

bazinski:
- fixes detector number, and channel in non zero suppressed digit data.
- expands accepted formats to all but config events.
- qc stats, and its subsequent message.
- defaults tracklethalfchambers header to 2, it was 0 at @shahor02 request.
- histogram option is left in until the qc version of the histograms is properly qualified.
- white bars traced to 2 uses of the word "side", detector side and halfchamber side, now fixed.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
@@ -16,17 +16,21 @@
 #ifndef O2_TRD_RAWDATASTATS
 #define O2_TRD_RAWDATASTATS
 
+#include "TObject.h"
 #include <iostream>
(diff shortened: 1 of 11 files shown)


bazinski:
@shahor02 This fixes a bug in parsing non zero suppressed data, and a bracket in the wrong place causing the print out of digits and tracklets found at the end of a tf to now be correct. The rest is related to qc.

bazinski:
Those errors are llvm errors not related to the code :-(

bazinski:
Those failures are still llvm, after kicking ci.

davidrohr:
merging, now that code freeze for pilot beam is over, I understand this is better than the current situation, although it will still be refined later.

PR #7382, opened 2021-10-20
Add MID to createTracksVariadic + fixes

shahor02:


DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
```diff
@@ -326,25 +331,49 @@ void o2::globaltracking::RecoContainer::createTracksVariadic(T creator) const
...
   }
 
+  // MID standalone tracks
+  {
+    const auto& rofs = getMIDTracksROFRecords();
+    for (const auto& rof : rofs) {
+      auto bcWidth = 56;
+      auto rofMeanBC = rof.interactionRecord.differenceInBC(startIR) + bcWidth / 2;
+      float t0 = rofMeanBC * o2::constants::lhc::LHCBunchSpacingMUS;
+      float t0err = o2::constants::lhc::LHCBunchSpacingMUS * bcWidth / 2;
```
shahor02: @dstocco I've borrowed t0 and t0err definition from the MCH, most probably wrong. Could you please check?

@jmyrcha for your ED tests small error in time should not be a problem.

dstocco: Hi @shahor02 , differently from MCH the MID resolution allows to separate each BC. So we do not need an average. The BC in the ROF is the BC of the event.

shahor02: Hi @dstocco 
Shall I take 0.6ns (bc length) ad an error estimate?

dstocco: Not sure why one would need an error on this. But yes, you can take the bc length as error for the moment.
I'll have a deeper look at the code tomorrow to understand it better.

PR #7383, opened 2021-10-20
Use Lifetime::QA for sampled quantities

ktf:
This will make sure that sampled data will not be used to do counting
of in-fly timeframes.

### DIFF
--- a/Utilities/DataSampling/src/DataSamplingPolicy.cxx
+++ b/Utilities/DataSampling/src/DataSamplingPolicy.cxx
@@ -69,6 +69,7 @@ DataSamplingPolicy DataSamplingPolicy::fromConfiguration(const ptree& config)
     }
     for (const auto& outputAsInputSpec : outputsAsInputSpecs) {
       outputSpecs.emplace_back(DataSpecUtils::asOutputSpec(outputAsInputSpec));
+      outputSpecs.back().lifetime = Lifetime::QA;
     }
   } else { // otherwise default format will be used
     for (const auto& inputSpec : inputSpecs) {
@@ -78,12 +79,12 @@ DataSamplingPolicy DataSamplingPolicy::fromConfiguration(const ptree& config)
           createPolicyDataOrigin(),
           createPolicyDataDescription(name, outputId++),
           DataSpecUtils::getOptionalSubSpec(inputSpec).value(),
-          inputSpec.lifetime});
+          Lifetime::QA});
       } else {
         outputSpecs.emplace_back(OutputSpec{
           {inputSpec.binding},
           {createPolicyDataOrigin(), createPolicyDataDescription(name, outputId++)},
-          inputSpec.lifetime});
+          Lifetime::QA});
       }
     }
   }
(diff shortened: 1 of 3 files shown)


davidrohr:
Is it safe to just overwrite it? Shouldn't we request the developer to set it, and give a failure if not? Otherwise I think this can be confusing?

ktf:
@knopers8 @Barthelemy in order to correctly account for how many timeframes have been processed, we need to have all OutputSpec which are associated to sampled data marked as Lifetime::QA. While this will not change how the data is processed, it will make sure it will not be accounted to consider a timeframe "complete" (because only a fraction of the timeframes have it, by definition).

Do you see any problems with doing it?

ktf:
@davidrohr see the other message. It Lifetime::Timeframe and Lifetime::QA should be identical in terms of processing, QA simply means that the data is sampled or created once in a while (like in the case of Analysis histograms).

davidrohr:
Then why do we call it Lifetime::QA? Can we rename it to Lifetime::Sampled or something like that?

ktf:
That's what I called it historically at the dawn of DPL. I am happy to rename it, however I do not like "Sampled" much, since it needs to apply also for histograms, which fit "Accumulated" more than "Sampled". Maybe "Occasional"?

davidrohr:
In some sense I would call it Optional, but we have optional already for another purpose. What is actually exactly the difference?
How about sporadic / irregular / arbitrary / selective ?

Barthelemy:
Hi, 
@knopers8 is absent till next Monday and I think that he is the one who should give his green light. I propose to wait for him.

davidrohr:
Hi @Barthelemy : this is actually on the critical path for commissioning, so I would not wait till Monday. Worst case we can revert. But it seems this breaks the DataSampling ctest. That must obviously be fixed.

davidrohr:
There seems to be still this error message as debug: `[1101455:trd-entropy-encoder]: [09:33:31][ERROR] Consume:true`

ktf:
Are you sure you are on the right branch?

Barthelemy:
> Hi @Barthelemy : this is actually on the critical path for commissioning, so I would not wait till Monday. Worst case we can revert. But it seems this breaks the DataSampling ctest. That must obviously be fixed.

Ok, I think that I fixed some of the failing tests. Although, @knopers8 will have to review it when he is back. 

However for lines 112 and 113 I don't know how to fix it. If I put a lifetime, then the subspec is wrong, if I don't put it, it fails.

ktf:
I fixed the remaining tests.

davidrohr:
The test is still failing:
```
 58/401 Test  #67: test_DataSampling_test_DataSampling ..................................................***Failed    0.14 sec
```

Barthelemy:
ok, forgot this one.

davidrohr:
> ok, forgot this one.

and it is not the DDS test :)

Barthelemy:
hopefully it is ok now

ktf:
So it seems ok now. @Barthelemy Can we merge?

knopers8:
I think this should be fine, but I will immediately contact you if I see something not working anymore.

PR #7384, opened 2021-10-20
DPL: add fairmq out of band channels

ktf:


ktf:
Not yet tested, however it should compile and not affect anything, since it's enabled only when Lifetime::OutOfBand is used. Yet, it moves the configuration of expiring inputs from Init to InitTask, so that the channels exists and I can add a poller on them. For this reason it's probably a good idea to test and merge this separately.

davidrohr:
This PR standalone does not pose problems in the FST, so it can be merged.
@ktf : One possibly important point I want to mention is: In the FST with 2 NUMA domains we have 2 readout-proxies in the same session. If you create out-of band-channels for them, we have to make sure these are 2 individual channels, so they should have a random suffix. I think you already add such a random suffix for the normal in-band channels. I remember we had to do this when we set up the 2 NUMA domains in 1 session in the first place.

ktf:

PR #7388, opened 2021-10-21
DPL: do not dump metrics which are empty

ktf:


ktf:
@jgrosseo in order for the metrics in the hyperloop GUI to display correctly I need to make sure I do not include empty collections, so this drops it. Do you confirm that's the required behaviour?

jgrosseo:
We can also protect in the GUI. What do you find more logical for the json?

ktf:
> What do you find more logical for the json?

I would drop it in any case in the JSON.

The important thing is that the GUI does not depend on the metric being there, IMHO.

PR #7391, opened 2021-10-21
Protection from zero payload

peressounko:


davidrohr:
@peressounko : Thx, this fix works (I cherry-picked the commit with the fix). Could you please rebase to dev? Currently this PR contains changes that are already in dev and there is a merge collision.

shahor02 (commented):
Hi @peressounko 
thanks for the fix! I see you have a lot of `throw`s but not all exceptions are caught a few try/catch clauses you have are rethrowing. The correct behaviour should be to catch the exception, LOG an error and to skip the trigger or the whole TF (sending empty output in that case). There should be a clean way to abandon the processing, could you please rectify this part?

peressounko:
Failed test build/O2/o2-cs8  seems to be unrelated to PHOS

peressounko:
Hi @shahor02, I forward exceptions to RawToCellConverter where I classify them and store in HWerror stream which in turn is analyzed by QC. Looking into logs is not efficient  during data taking, so I preferred this way. More details will not help on-call expert to identify problem. Depending on exception I either abandon one RCU package or one channel payload. Skipping whole TF is overreacting I think. Best regards, Dmitri

shahor02:
Hi @peressounko 

I don't see where these exceptions
https://github.com/AliceO2Group/AliceO2/blob/02276bf560a9790cc8d1664de5495a09769ee963/Detectors/PHOS/reconstruction/src/RawReaderMemory.cxx#L141
https://github.com/AliceO2Group/AliceO2/blob/02276bf560a9790cc8d1664de5495a09769ee963/Detectors/PHOS/reconstruction/src/RawReaderMemory.cxx#L149 (though this function seems to be unused).
are caught.
`getRawHeader` can throw in https://github.com/AliceO2Group/AliceO2/blob/02276bf560a9790cc8d1664de5495a09769ee963/Detectors/PHOS/reconstruction/src/AltroDecoder.cxx#L30 and the call to decode is seems not to be in any `try` scope.

davidrohr:
Merging this since it fixes an actual problem affecting the full system test. Catching the exceptions if still missing can be done in a follow-up PR.

peressounko:
Thanks, last commit treats exceptions properly, so issue if fixed too.

PR #7392, opened 2021-10-21
Include PWG code in name of executable

pbuehler:
@jgrosseo Please have a look.


jgrosseo:
Is `dmname.lower()` `PWGCF` or only `CF`?

pbuehler:
> Is `dmname.lower()` `PWGCF` or only `CF`?

only cf.
Anyway, good point. There is an other change I need to make. Wait a second.

PR #7393, opened 2021-10-21
DPL: provide metric for rate in and out of each channel

ktf:


ktf:
@vascobarroso as requested.

ktf:
The rates will be called `channel_{}_rate_out_mb_s` / `channel_{}_rate_out_mb_s` where `{}` is the actual channel name.

vascobarroso:
@ktf cool thanks! What channel(s) should we use in the grafana dashboard ? Do we have well known channel names ?

awegrzyn:
Would it be possible to use `channel_rate_in_mb_s` name, and then `.addTag(tags::Key::ID, 0);` ? Instead of having different name for each channel

ktf:
@awegrzyn can ID be alphanumeric? I do not have a unique numeric ID per channel.

ktf:
Merging this to have something in tomorrow build. Not clear to me how one can distinguish between the different channels actually.

awegrzyn:
@ktf Is it possible to revert this? It's produces ~10 kHz of metrics, but it does not provide what we need.

ktf:
For the record: metrics were marked as Debug.

PR #7395, opened 2021-10-21
[WIP][EMCAL-681] Implement pileup simulation for pre-trigger bunches

hahassan7:
The current pileup simulation only includes collisions that happen up to 1.5 mus after the trigger, contributions from collisions before the trigger are not included. Still they can leave a signal in the readout window. The logic needs to be changed in the following way:

 -Digitization needs to be started again 600 mus before the end of the busy time.
 -Trigger time still needs to be set to the event that actually triggered the collision.
 -Collisions with a time > trigger time + 900 mus will not leave a signal in the readout window and can therefore be discarded.

In addition to some changes in the DigitsWriteoutBuffer.


mfasDa:
Still needed but first needs to be fully debugged locally.

PR #7396, opened 2021-10-21
o2-eve: adding maximum allowed memory option

jmyrcha:
in case of any potential memory leaks, o2-eve will exit after exceeding memory limit option
(so it can be run again)
not setting a limit (-m limit) causes not checking


jmyrcha:
build/O2/o2-cs8 failed because :libgandiva.so.500.0.0: undefined reference to `llvm::createNewGVNPass()
(which not my fault)

shahor02 (approved):

PR #7398, opened 2021-10-22
[MFT] Workaround for MFT tracking with misaligned detector

rpezzi:
This PR adds option to force zero B field tracking for the MFT and set this as default for convenience in pilot beam assessment and event display. At the moment, B=0 case uses a dummy MFT tracker: Kalman filter is disabled, track position and direction from first and last track-clusters. This is considered a valid approach while alignment corrections are not applied to the detector geometry.
To restore normal tracking configuration use the following configKeyValues:

`o2-mft-reco-workflow --configKeyValues "MFTTracking.forceZeroField=false;MFTTracking.LTFclsRCut=0.0100;"`


rpezzi:
Errors here also are not related to the code. The track finding configuration here will help get tracks for first assessment of the data and event display while detector is not aligned. @shahor02 @bovulpes , OK to merge?

shahor02 (approved):

PR #7399, opened 2021-10-22
Embedding: combine associated signal and background MC collisions

nburmaso:
* Corresponding signal and background events are combined into "composite" collisions
* MC particles from a background event in these collisions have second bit = 1 in `flags` (see https://github.com/AliceO2Group/AliceO2/blob/dev/Framework/Core/include/Framework/DataTypes.h#L68)
* Note that there may be several signal events inside a collision (see e.g. comments here https://github.com/AliceO2Group/O2DPG/blob/055b8759c5f58822b5006602a69df76bfca88a1b/MC/bin/o2dpg_sim_workflow.py#L15-L17)
* A primary vertex is pointed to a "composite" MC collision that contains its corresponding MC event. That means that, in principle, several vertices may point to one MC collision.

@sawenzel, @iarsene: please take a look and see if this implementation is ok for you

### DIFF
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -330,7 +330,7 @@ class AODProducerWorkflowDPL : public Task
                             gsl::span<const o2::dataformats::VtxTrackRef>& primVer2TRefs,
                             gsl::span<const GIndex>& GIndices,
                             o2::globaltracking::RecoContainer& data,
-                            std::vector<std::pair<int, int>> const& mcColToEvSrc);
+                            std::map<std::pair<int, int>, int> const& mcColToEvSrc);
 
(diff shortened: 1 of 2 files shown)


iarsene:
Thanks @nburmaso, i will run a few events locally to test this and let you know, but it might take until monday.
Ionut

iarsene:
Hi @nburmaso,
I run locally a simulation using the setup with one prompt jpsi injected over a pythia event (https://github.com/AliceO2Group/O2DPG/blob/master/MC/run/PWGDQ/runPromptJpsi_midy_pp.sh). I simulated 10 time-frames each with 10 events.
I can see that the pp and the jpsi signal are now merged in a single event. Also, the indices seem to work properly now, because i manage to find all of the injected jpsi assigned to the proper MC event. There are still a few things i need to understand, but i need some more events, so if its fine with everyone, i suggest to merge this PR and run it in the nightlies.
Ionut

sawenzel:
Merging after positive feedback from @iarsene.

PR #7400, opened 2021-10-22
DPL: introduce Lifetime::Sporadic as alias to Lifetime::QA

ktf:


ktf:
Adds a new name for Lifetime::QA so that we can get rid of it everywhere and then remove it.

PR #7403, opened 2021-10-22
fix in TOF diagnostic words

noferini:


noferini:
@chiarazampolli 
This is the PR to fix a mismatch in TRM mapping

DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
```diff
@@ -41,11 +41,25 @@ class Diagnostic
...
   void merge(const Diagnostic* prev);
+  unsigned long size() const { return mVector.size(); }
+  ULong64_t getPattern(int i) const
+  {
+    auto iter = mVector.begin();
+    for (; i-- > 0;) {
+      iter++;
+    }
+    return iter->first;
+  }
```
shahor02, line 58: this method is not used at the moment but it is very expensive and unsafe.

shahor02 (approved):
@noferini merging, but please see comment below, if you really need an "indexed" access to the map, you better first build a vector<key>, then use it as `map[vector[i]]`. 
But I don't see a point of such access since the order of elements in the map changes upon insertion.

DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
noferini: Hi Ruben,
thanks.
This method is not used. I added it just to temporary check the entry in the CCDB are consisten with expectation (I didn't want to make map public).
I will add a method to decode the words (needed in simulation).
I didn't implement it yet since I am still thinking what is the best way (I'll do it soon)

PR #7411, opened 2021-10-23
DPL: miscellaneous fixes propaedeutic for rate limiting

ktf:


ktf:
These are definitely safe to have. Checking them separately to reduce moving parts.

ktf:

PR #7414, opened 2021-10-25
Misaligned sync reco parameters

mpuccio:
- First version of misaligned selections
- Update selections and introduce new mode


mpuccio:
Hi @shahor02, sorry I had it since last Wednesday but I wanted to finalise the "proper" version first... I am not sure how relevant it is with the code freeze.

shahor02 (approved):
Hi @mpuccio 
do you have by chance also `async_misaligned`?

mpuccio:
Hi @shahor02 , not yet! But consider that the efficiency is already saturating to the ~60% reconstructed vertices we have. For the async phase I would also move to the new version of the tracker with nsigma cuts that I will discuss today in the WP12 meeting.

PR #7421, opened 2021-10-25
Adding some plots + use labels to avoid duplicated tracks

chiarazampolli:
Use MC info for ITSTPC matching QC

Using lables to remove duplicates

clang-format

### DIFF
--- a/Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
+++ b/Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
@@ -16,6 +16,7 @@
 #include "ReconstructionDataFormats/TrackParametrization.h"
 #include "DetectorsBase/Propagator.h"
 #include "DetectorsBase/GeometryManager.h"
+#include <algorithm>
 
 using namespace o2::globaltracking;
 
@@ -33,6 +34,9 @@ void MatchITSTPCQC::deleteHistograms()
   delete mPtTPC;
   delete mFractionITSTPCmatch;
   delete mPt;
+  delete mPhiTPC;
+  delete mFractionITSTPCmatchPhi;
+  delete mPhi;
   delete mEta;
   delete mChi2Matching;
   delete mChi2Refit;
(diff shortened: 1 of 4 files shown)


chiarazampolli:
@shahor02 : with the last commit, one can have plots of efficiency with only physical primaries. Since, though, the functionality is in O2Physics, for now the selection does not do anything. I can of course port here (for the time being) the algorithm.

Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
```diff
@@ -56,16 +73,33 @@ void MatchITSTPCQC::reset()
...
+  // ...till here
+  mEta = new TH1F("mEta", "Eta distribution of matched tracks; Eta; dNdEta", 100, -2.f, 2.f);
   mChi2Matching = new TH1F("mChi2Matching", "Chi2 of matching; chi2", 200, 0, 20);
   mChi2Refit = new TH1F("mChi2Refit", "Chi2 of refit; chi2", 200, 0, 20);
-  mTimeResVsPt = new TH2F("mTimeResVsPt", "Time resolution vs Pt; Pt; time res", 100, 0.f, 20.f, 100, 0.f, 2.f);
+  mTimeResVsPt = new TH2F("mTimeResVsPt", "Time resolution vs Pt; Pt [GeV/c]; time res [us]", 100, 0.f, 20.f, 100, 0.f, 2.f);
 
   mPtTPC->Sumw2();
   mPt->Sumw2();
+  mPhiTPC->Sumw2();
```
mario-krueger, line 97: ciao Chiara,
to my understanding one should only call sumw2 if one intends to fill a histogram with weights (and even then it would be called automatically at the first fill).
This effectively triples the size in memory, since a weights structure (TArrayD) of nBins*sizeof(double) is added to the histogram (https://github.com/root-project/root/blob/master/hist/hist/src/TH1.cxx#L8879). To get the errorbars for a plot or further processing (scale, divide, etc) one could just call sumw2 offline which then calculates sqrt(n) on the already filled histogram.

chiarazampolli: Ciao @mario-krueger , 
Yes, i know that sumw2 is quite demanding, and i was not sure it is needed for a TEfficiency, but I did not have the time to check. The division for the TEfficiency is not done offline, but in the QC framework, this is why I thought one should add it there. I can check if the errors are correctly calculated in the TEfficiency without sumw2 (unless you know already). 
Thanks for the comment!!
Chiara

mario-krueger: Ah, I see. Actually, I dont know about TEfficiency (never tried it).
I was just triggered by all the Sumw2() calls because in many AliPhysics tasks this function was called unnecessarily (also by myself) and since I understood that this is not such a great idea, I keep bugging people about it :)

chiarazampolli:
@shahor02 : errors are unrelated to the PR. When you have time, you could review this PR.

Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
chiarazampolli: We also had issues in Run 2 QA about it, so I know... :-) 
I really appreciated that you took time to check! 
And as I said, I will cross-check too.

chiarazampolli:
BTW: I noted that with the new checks for the physical primary, the task became a bit slower (it was not really a measurement actually), so if anything is spotted from the code, any comment is very welcome! I will also try to run with some profiler.

shahor02 (approved):

PR #7424, opened 2021-10-25
CPV: fix in cluster to ctf packing

sevdokim:
Hello! Small but essential for success reading clusters from ctf fix added.

### DIFF
--- a/DataFormats/Detectors/CPV/include/DataFormatsCPV/Cluster.h
+++ b/DataFormats/Detectors/CPV/include/DataFormatsCPV/Cluster.h
@@ -13,7 +13,7 @@
 #define ALICEO2_CPV_CLUSTER_H_
 
 #include "DataFormatsCPV/Digit.h"
-#include <math.h>
+#include <cmath>
 
 namespace o2
 {
--- a/DataFormats/Detectors/CPV/include/DataFormatsCPV/Cluster.h
+++ b/DataFormats/Detectors/CPV/include/DataFormatsCPV/Cluster.h
@@ -13,6 +13,7 @@
 #define ALICEO2_CPV_CLUSTER_H_
 
 #include "DataFormatsCPV/Digit.h"
+#include <math.h>
 
 namespace o2
 {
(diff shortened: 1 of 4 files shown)


sevdokim:
Wth this commit ctf dictionary must be regenerated I suppose.

kharlov (dismissed):

shahor02 (commented):
@sevdokim yes, such changes make the dictionary less effective (at least for the data it was created with). I would not change the dictionary 1 day before the beam. Is this change essential for the the pilot beam?
Note that with modified code you also make unreadable the existing CTF data (or at least make its result different from what it would be before).

sevdokim:
Hi @shahor02 
the problem is that now ctf module number decoded from ctf wrongly and cluster energy range is absolutely not enough.
This is how typical energy spectrum is looking like:
image
And this is how it looks like now after compress-decompress ctf:
image
All the info for energies > 255 is lost. This is not good for us. So mainly energy spectrum is important for us, and we want to have small energies with maximum available info (we will decide if small-energy clusters are real signal or some fake) and do not care much about high-energy clusters (those are most likely real signals). With proposed formula spectrum decoded spectrum is looking like this:
image
Max info is kept for small energies and with increase of energy digitization step is also increasing which is absolutely acceptable for us. Of cause there is no need to change dictionary one day before pilot beam but after it I would do so.

Also now the module number is decoded wrongly: M4 decoded as M0 which does not exist. It also to be fixed.

sevdokim:
By the way, how to run tests without aliBuild?

shahor02:
@sevdokim the compiled tests are in the `..BUILD/O2-latest/O2/stage/tests` directory.
I don't know when this PR will appear on the epn's and if it can be used for sync processing. In any case, I would prefer to not change the dictionary now, the old one should still work with new getters (obviously, the decoding of old data will also change).
Can you point me on some CPV run with raw data stored, I'll check if the dictionary is reliable?

sevdokim:
Hi @shahor02 ,
surely there is no need to change ctfdict now. This is not critical for us. You are right, getters will work. 
A good mc run is stored at /eos/aliceo2/raw/2021/OCT/504798/

shahor02:
@davidrohr the fullCI stopped w/o any obvious reason, any idea why?

shahor02:
@sevdokim I've tried `/eos/aliceo2/raw/2021/OCT/504798/` with new code and old dictionary, it is good enough.

PR #7428, opened 2021-10-26
DPL Analysis: save histograms in the order they were added

mario-krueger:
- by default store histograms in the order they were added to the registry
- check if histogram names are valid
- make histogram sorting case insensitive

### DIFF
--- a/Framework/Core/include/Framework/HistogramRegistry.h
+++ b/Framework/Core/include/Framework/HistogramRegistry.h
@@ -87,7 +87,7 @@ class HistogramRegistry
   };
 
  public:
-  HistogramRegistry(char const* const name = "histograms", std::vector<HistogramSpec> histSpecs = {}, OutputObjHandlingPolicy policy = OutputObjHandlingPolicy::AnalysisObject, bool sortHistos = true, bool createRegistryDir = false);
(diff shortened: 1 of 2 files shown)


ktf:
This is fine for me. I assume all these features where requested at some point, correct?

mario-krueger:
Well, I intended to add this feature since the beginning but never did.
The way it was before, the histograms (if not sorted) were just put to file in a random order according to their index in the registry. This is why back then I put the standard mode to 'sorted' to avoid having the file contents scattered arbitrarily.
I suppose usually the analyser wants to keep the order in which he added the histograms.
We could of course also have it sorted alphabetically by default in case this is what most people want.
@jgrosseo do you have an opinion on this?

mario-krueger:
By checking the naming patterns, I want to avoid that people put histogram names like
`/asdf///aa` or `myHist/` or `aaa/bbb.sdfer@` or `my___asdfHist`
Where at least a `/` at the beginning or a double `//` would most likely break something.

mario-krueger:
regarding the sorting, I think it is more intuitive if things are sorted like {`abc`, `Abc`, `def`} instead of {`abc`, `def`, `Abc`}

jgrosseo:
In what use cases is the order relevant?

mario-krueger:
For developing new analysis code, I find being able to quickly find again a certain histogram in the output file very useful.

jgrosseo (approved):

jgrosseo:
OK, I have no objections...

PR #7435, opened 2021-10-27
fix in tof sim for large TF offset

noferini:


noferini:
@chiarazampolli 
I implemented a fix. I still need to check it. I will move from draft to reviewed once checked

sawenzel (approved):

sawenzel:
It works! Thanks.

chiarazampolli:
Thanks @noferini !

sawenzel:
fullCI tested locally. Merging since important fix.

chiarazampolli:
One more comments: with this we should see 3 * N_orbits_per_TF * N_TF_in_job per job. So with a 128 orbits TF and jobs with 25 TF, we should have 6400 RO windows. @sawenzel , did you see something consistent in your local test (it clearly depends on N_orbits_per_TF and N_TF_in_job)?

PR #7436, opened 2021-10-27
DPL: make TimingInfo a service

ktf:


ktf:
Tested on hyperloop as well. Merging.

PR #7446, opened 2021-10-28
noise suppression enabled for tof calib with cosmics (diagostic now p…

noferini:
…ropagated also when reading from digits/ctf)


noferini:
@chiarazampolli 
Diagnostic ferquency added also when reading from CTF and digits to allow filling of ccdb also in async phase for diagnostic and to apply noise filter for calibration (only for clusters)

noferini:
checked with 123 TF

o2-ctf-reader-workflow --shm-segment-size 5000000000 -b --ctf-input o2_ctf_run00505207_orbit0004708979_tf0000035744.root --onlyDet TOF \
| o2-tof-reco-workflow -b --input-type none --output-type clusters --disable-root-output --calib-cluster --cluster-time-window 10000 --cosmics --disable-mc \
| o2-tof-cluster-calib-workflow -b

noferini:
all checked passed

PR #7448, opened 2021-10-28
DPL: mark rate metrics as debug

ktf:


ktf:
@awegrzyn as discussed. @davidrohr do you need any of these metrics for the EPNs?

davidrohr:
These are per-channel data rates? I think we do not need them.

PR #7449, opened 2021-10-28
Few improvements for the ITS CA async processing

mpuccio:
- Protect from spurious clusters while reading CTFs
Fixes some occasional crashes and reports in case of "strange" clusters are detected

- Optionally use a diamond instead of PV from finder
Helps greatly to process the runs from yesterday

- Add misalignment to cosmic processing
Makes cosmics reconstruction up to 100% more efficient than before (even some interactions with the detector material are reconstructed)

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/src/Tracker.cxx
+++ b/Detectors/ITSMFT/ITS/tracking/src/Tracker.cxx
@@ -656,6 +656,10 @@ void Tracker::getGlobalConfiguration()
     params.NSigmaCut *= tc.nSigmaCut > 0 ? tc.nSigmaCut : 1.f;
     params.CellDeltaTanLambdaSigma *= tc.deltaTanLres > 0 ? tc.deltaTanLres : 1.f;
     params.TrackletMaxDeltaPhi *= tc.phiCut > 0 ? tc.phiCut : 1.f;
+    for (int iD{0}; iD < 3; ++iD) {
+      params.Diamond[iD] = tc.diamondPos[iD];
+    }
+    params.UseDiamond = tc.useDiamond;
   }
 }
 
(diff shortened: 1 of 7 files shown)


Detectors/ITSMFT/ITS/tracking/src/Tracker.cxx
```diff
@@ -625,9 +625,12 @@ track::TrackParCov Tracker::buildTrackSeed(const Cluster& cluster1, const Cluste
...
-                   20.f; // FIXME: MS contribution to the cov[14] (*20 added)
+  float cy = 1.e15f;
+  if (std::abs(getBz()) > o2::constants::math::Almost0) {
+    cy = (math_utils::computeCurvature(x1, y1, x2, y2 + resolution, x3, y3) - crv) /
+         (resolution * getBz() * o2::constants::math::B2C) *
+         20.f; // FIXME: MS contribution to the cov[14] (*20 added)
+  }
   const float s2 = resolution;
 
   return track::TrackParCov(tf3.xTrackingFrame, tf3.alphaTrackingFrame,
```
shahor02, line 636: In case of B=0 one should assign not an infinite momentum but the most probable pT, in order to have at least mat. corrections correct at least in average. @davidrohr we discussed this long ago, I assume you already have it, do you have some constants defined? Should be something like 0.6 GeV but better to have common convention.

mpuccio: OK, I assume you are referring to the line below (638), I would still blow up the uncertainty in case of zero B field. I will use `1. / o2::track::kMostProbablePt` (1./0.6). I will ruin the charm of all green CI for this fix... :P

PR #7450, opened 2021-10-28
DPL Analysis: fix incorrect use of offset

aalkin:


ktf:
Checked on hyperloop as well.

PR #7453, opened 2021-10-29
Improve sending channel related metrics

awegrzyn:
@ktf This is how I would have done it to avoid sending with high rate
Not sure if this is actually needed now.

Referencing https://github.com/AliceO2Group/AliceO2/pull/7448

### DIFF
--- a/Framework/Core/src/CommonServices.cxx
+++ b/Framework/Core/src/CommonServices.cxx
@@ -527,22 +527,21 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
   stats.channelBytesIn.resize(device->fChannels.size());
   stats.channelBytesOut.resize(device->fChannels.size());
   size_t ci = 0;
+  auto inMetric = Metric{"channel_in_rate", o2::monitoring::Verbosity::Debug}.addTag(Key::Subsystem, Value::DPL);
+  auto outMetric = Metric{"channel_out_rate", o2::monitoring::Verbosity::Debug}.addTag(Key::Subsystem, Value::DPL);
   for (auto& channel : device->fChannels) {
     auto newBytesOut = channel.second[0].GetBytesTx();
     auto newBytesIn = channel.second[0].GetBytesRx();
-    monitoring.send(Metric{(float)(newBytesOut - stats.channelBytesOut[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f),
-                           fmt::format("channel_{}_rate_in_mb_s", channel.first),
-                           o2::monitoring::Verbosity::Debug}
-                      .addTag(Key::Subsystem, Value::DPL));
-    monitoring.send(Metric{(float)(newBytesIn - stats.channelBytesIn[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f),
-                           fmt::format("channel_{}_rate_out_mb_s", channel.first),
-                           o2::monitoring::Verbosity::Debug}
-                      .addTag(Key::Subsystem, Value::DPL));
+
+    inMetric.addValue((float)(newBytesOut - stats.channelBytesOut[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f), channel.first);
+    outMetric.addValue((float)(newBytesIn - stats.channelBytesIn[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f), channel.first);
+
     stats.channelBytesOut[ci] = newBytesOut;
     stats.channelBytesIn[ci] = newBytesIn;
     ci++;
   }
-
+  monitoring.send(std::move(inMetric));
+  monitoring.send(std::move(outMetric));
   stats.lastVerySlowMetricSentTimestamp.store(stats.beginIterationTimestamp.load());
 };
 
--- a/Framework/Core/src/CommonServices.cxx
+++ b/Framework/Core/src/CommonServices.cxx
@@ -533,8 +533,8 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
     auto newBytesOut = channel.second[0].GetBytesTx();
     auto newBytesIn = channel.second[0].GetBytesRx();
 
-    inMetric.addValue((float)(newBytesOut - stats.channelBytesOut[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f), channel.first);
-    outMetric.addValue((float)(newBytesIn - stats.channelBytesIn[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f), channel.first);
+    outMetric.addValue((float)(newBytesOut - stats.channelBytesOut[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f), channel.first);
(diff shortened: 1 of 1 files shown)


awegrzyn:
btw. it looks to me that "out" metrics are being send under "in" name, am I correct?

ktf:
Swapping input and output might explain why we had so many 0s... That said, of course we can do globally, but then it will only be meaningful for the input and output devices. If that's what is needed, I am happy with this PR being merged.

Framework/Core/src/CommonServices.cxx
```diff
@@ -527,22 +527,21 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
   stats.channelBytesIn.resize(device->fChannels.size());
   stats.channelBytesOut.resize(device->fChannels.size());
   size_t ci = 0;
+  auto inMetric = Metric{"channel_in_rate", o2::monitoring::Verbosity::Debug}.addTag(Key::Subsystem, Value::DPL);
```
ktf, line 530: Maybe we can also drop the debug if this is merged?

Framework/Core/src/CommonServices.cxx
```diff
@@ -527,22 +527,21 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
...
-                           fmt::format("channel_{}_rate_in_mb_s", channel.first),
-                           o2::monitoring::Verbosity::Debug}
-                      .addTag(Key::Subsystem, Value::DPL));
-    monitoring.send(Metric{(float)(newBytesIn - stats.channelBytesIn[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f),
-                           fmt::format("channel_{}_rate_out_mb_s", channel.first),
-                           o2::monitoring::Verbosity::Debug}
-                      .addTag(Key::Subsystem, Value::DPL));
+
+    inMetric.addValue((float)(newBytesOut - stats.channelBytesOut[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f), channel.first);
+    outMetric.addValue((float)(newBytesIn - stats.channelBytesIn[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f), channel.first);
```
ktf: still swapped...

Framework/Core/src/CommonServices.cxx
awegrzyn: We could, but as we need just values from `readout` and `output` proxies would it be possible to send just from these 2 devices? If not I could suppress irrelevant metrics on Telegraf side.

ktf: You could check if the name of all the channels start with `from_`. If they do, then it's not a device on the edge of the DPL topology.

ktf: We could merge this as is, then I can do the change, if you want.

ktf (dismissed):

ktf (approved):

Framework/Core/src/CommonServices.cxx
awegrzyn: Okay, it's an improvement, but looking at stats from P2 most values are coming from QC channels, prefixes `from_MERGER...` and `from_Dispatcher...` so would be nice to somehow drop that as well.

ktf: Well, in principle we can simply check the device id to contain "proxy" in the name. Would that be enough?

awegrzyn: okay, sounds good to me.

davidrohr:
@ktf @awegrzyn : Do I understand correctly that this is for per-channel input rates in MB/s?

Could we have the same for the total input rate of a device?
We have `input_rate_mb_s` but I think that is reporting something else, at least the reported rates are much higher than the actual input rates. Perhaps it is not taking into accound the idle time of the device?
Total input rate in MB/s per device is certainly something I would like to have.
In particular also for the "Input Rate" panel in the DPL grafana, and for the CTF data rate in the full dataflow monitoring dashboard

ktf:
@davidrohr with @awegrzyn change in this PR, they become the **per-device** input and output rates as reported by FairMQ. Indeed `input_rate_mb_s` is actually something different. I.e. the rate we have been processing input messages BEFORE they get processed by the reported device. Most likely the latter is not particularly useful.

davidrohr:
OK, perhaps if `input_rate_mb_s` is not useful, I would drop it. In particular since it is misleading.

For the per-device input and output rates. If they are already reported by FairMQ, why don't we just use the FairMQ values?

Still, do I understand that this PR creates these metrics per channel? Could we also have them globally, i.e. total input/output per device?

awegrzyn:
> @awegrzyn change in this PR, they become the **per-device** input and output rates

They are not per device, it is still per channel but sent as single metric (opposed to 1 metric per channel). They will need to be summed afterwards to produce per device value.

`addValue` does not mean `+=` but add another metric value to given metric (single metric can have multiple values).
I think what David proposed makes sense, why not summing all values in the loop?

davidrohr:
ok, thx, so what do you prefer? Do you want to sum them up in grafana? Or should we instead send a single per-device metric? Could be send in addition / alternatively? That could reduce the grafana load / size in influxdb.

vascobarroso:
@davidrohr, small off-topic: will the improvements under discussion in this PR also fix the CTF writer metrics ? The value we see at P2 is usually much higher than expected.

davidrohr:
Yes, exactly, that will fix it. We will need to switch to the other metric then in the `full dataflow monitoring dashboard`
The only question I still have is whether it would not be better to already send an aggregate metric instead of per-channel, to reduce the grafana/influx load. But both will work of course.

ktf:
@awegrzyn I do not understand how you distinguish between the various channels then...

awegrzyn:
So there's no way to distinguish `in` and `out` channels?

davidrohr:
Sure, you need one aggregate metric for in, and one aggregate metric for out.
But e.g. for the CTF writer, you would not need to sum up the individual channels for the inputs of all detectors.

ktf:
To be clear:

* Each fairmq channel is bidirectional, so each one will have a metric for `in` and a metric for `out`.
* However DPL uses those channels in a single direction, because the data processing graph is really a DAG. For each channel, only one of the two metrics will have a value different than 0.

So, the only way to distinguish between a in and out channel is to check the name. If it starts with `from_<device-name>` it's an output channel. This is for DPL channels. For the non-DPL channels, e.g. a proxy downstream channel, we need to know what a given device is used for.

awegrzyn:
I'd be in fav of sending already summed value, lemme know what do you need from my side?

davidrohr:
Then I would propose we leave the current metrics in (potentially mark them as debug, I think they already are after @ktf's) other commit, and add 2 new metrics for total_in_mb_s and total_out_mb_s?
@ktf : Could you do that?

ktf (commented):
How about #7486 instead of this?

davidrohr:
> How about #7486 instead of this?

I don't fully understand the instead? Why not in addition. If I understand this PR correctly, it still sends the metrics for all channels, but in a smarter way. So I would merge this, and then #7486 on top?

ktf:
Well, actually who needs the per channel metrics? If no one, we can simply drop them.

davidrohr:
> Well, actually who needs the per channel metrics? If no one, we can simply drop them.

I don't really need them. If you need them neither, we can drop them. Then we can also get rid of the 2 different update intervals for the metrics I think.

davidrohr:
not needed anymore

PR #7457, opened 2021-10-29
Automatically push workflow changes to alibuild/AliceO2

TimoWilken:
This would solve the occasional breakage of the clang-format check, which can't push branches to its fork when GitHub workflows differ between its old dev and PRs based on the current dev.

It would probably be better to just give the clang-format check the workflow permission, if possible.


ktf:
Why not giving the clang-format check the workflow permissions? Can it actually execute any user code?

ktf (commented):
If possible, I would use the other option you suggest.

TimoWilken:
> Why not giving the clang-format check the workflow permissions? Can it actually execute any user code?

Done. I assume it is the "Pull request bot" token, not the "CI Token" one, as the former was used more recently.

PR #7461, opened 2021-10-29
TFReader sleeps for 10s at startup, waiting for other devices to start

shahor02:
@ktf this indeed works, so added temporarily until more clever solution is found

### DIFF
--- a/Detectors/Raw/TFReaderDD/src/TFReaderSpec.cxx
+++ b/Detectors/Raw/TFReaderDD/src/TFReaderSpec.cxx
@@ -80,6 +80,10 @@ void TFReaderSpec::init(o2f::InitContext& ic)
 //___________________________________________________________
 void TFReaderSpec::run(o2f::ProcessingContext& ctx)
 {
+  if (!mTFCounter) { // RS FIXME: this is a temporary hack to avoid late-starting devices to lose the input
+    LOG(WARNING) << "This is a hack, sleeping 10 s at startup";
+    usleep(10000000);
+  }
   if (!mDevice) {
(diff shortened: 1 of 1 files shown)


davidrohr:
Can you explain why we need this? Can this be related to the problem I see with raw proxy and StfBuilder: when the StfBuilder starts too early, the chain fails.

shahor02:
The tf-reader sends out raw data fast while the DISTSUBTIMEFRAME is forwarded later.
 If the device expecting this conditional RAWDATA and DISTSUBTIMEFRAME starts later than the  tf-reader sends the RAWDATA, it receives the DISTSUBTIMEFRAME only, triggering a `deadbeef`. 
At least this is a working hypothesis for why I was seeing lot of `deadbeefs` when using tr-reader and adding this delay seems to help.
Perhaps @ktf can give more intelligent explanation

davidrohr:
ok, so it is different to my problem. For me it just crashes...

ktf:
I think it actually receives both, however due to the late start, it misses the notification on the channel of the RAWDATA. Then, when the DISTSUBTIMEFRAME arrives, it triggers the receive on that channel and queries the other channel only after, resulting in the deadbeef message.

shahor02:
@ktf thanks, do you think we can have better solution? I am not sure how mush safe is this 10s sleep.

ktf:
Yes, of course. I will try waking up the event look on each connections /
disconnection event. If what I described it's what is happening, that
should be enough. Long term we should probably also sort the order of
polled channels so that those that have an optional input get checked
first. I will try to have something tonight.

PR #7468, opened 2021-10-29
A few more selections

chiarazampolli:
Hello @sawenzel , 
After the QC meeting, I would propose to add a few selections. 
What do you think?
Pinging also @fprino.


sawenzel:
fine with me.

sawenzel (approved):

PR #7479, opened 2021-10-31
Fix TRD tracking error messages and option to write TPC tracking outp…

martenole:
…ut to ROOT file

### DIFF
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -97,9 +97,8 @@ TOF_CONFIG=
 TOF_INPUT=raw
 TOF_OUTPUT=clusters
 ITS_CONFIG_KEY=
-TRD_CONFIG=
+TRD_FILTER=
 TRD_CONFIG_KEY=
-TRD_TRANSFORMER_CONFIG=
 CPV_INPUT=raw
 EVE_CONFIG=" --jsons-folder $EDJSONS_DIR"
 MIDDEC_CONFIG=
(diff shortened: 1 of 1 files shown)


shahor02:
@martenole can you add a dummy commit to retrigger fullCI? 
BTW, I see in the logs 
```
[109124:TRDTRACKLETTRANSFORMER]: [08:21:39][ERROR] Requested resource does not exist: http://ccdb-test.cern.ch:8080/TRD/Calib/ChamberT0/555555/
```

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -205,6 +209,14 @@ done
...
 has_detector_reco FT0 && PVERTEX_CONFIG+=" --validate-with-ft0"
 
+# In case matching of TPC-TRD tracks is requested, force off the TRD filtering
+TRD_SOURCES_ARRAY=$(echo $TRD_SOURCES | tr "," "\n")
+for src in $TRD_SOURCES_ARRAY; do
+  if [[ $src == "TPC-TRD" ]]; then
+    TRD_FILTER=""
+  fi
+done
+
```
davidrohr, line 219: This 1-liner should do the same :)
```
[[ $TRD_SOURCES =~ (^|,)"TPC-TRD"(,|$) ]] && TRD_FILTER=""
```
For my understanding, why do you want to disable the filtering in case we match to TPC? As far as I understand @shahor02 , this is mostly for constraining the TPC tracks before the TPC-ITS matching. So if we do not do TPC-ITS matching for that collision, I don't know if we need the TRD matching anyway?

shahor02: indeed, I don't see any reason to run TPC-TRD matching in sync stage, unless for QC

PR #7483, opened 2021-11-01
Provide MC generator status code as part of MCTrack

sawenzel:
Status codes provided from MC generators (as part
of TParticle) are now propagated and stored in MCTrack
(for primary particles).

This should allow to fill the necessary information
in AOD tables.

Some future memory optimization can be done and is targeted
(avoiding extra data member for this).

### DIFF
--- a/DataFormats/simulation/include/SimulationDataFormat/MCTrack.h
+++ b/DataFormats/simulation/include/SimulationDataFormat/MCTrack.h
@@ -187,6 +187,9 @@ class MCTrackT
   /// get the production process (id) of this track
   int getProcess() const { return ((PropEncoding)mProp).process; }
 
+  /// get generator status code
+  int getStatusCode() const { return mStatusCode; }
+
   void setToBeDone(bool f)
   {
     auto prop = ((PropEncoding)mProp);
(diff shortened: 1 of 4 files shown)


DataFormats/simulation/include/SimulationDataFormat/MCTrack.h
```diff
@@ -339,6 +348,8 @@ inline MCTrackT<T>::MCTrackT(const TParticle& part)
     setToBeDone(true); // if inhibited, it had to be done: restore flag
     setInhibited(true);
   }
+  // set MC generator status code only for primaries
+  mStatusCode = part.TestBit(ParticleStatus::kPrimary) ? part.GetStatusCode() : -1;
```
sawenzel, line 352: @amorsch : Is `-1` ok as `not set` or should we use 0.

sawenzel:
@nburmaso : Using this, you can fill the AOD by calling `track.getStatusCode()`.

DataFormats/simulation/include/SimulationDataFormat/MCTrack.h
amorsch: @sawenzel 
The process type for secondaries can be obtained from 
part.GetUniqueID();

sawenzel: I know. But the process is already encoded in MCTrack elsewhere (for now). I would leave the logic of mixing status codes and the process to the filling part in the AOD creator. It is better that users call `getStatusCode()` and `getProcess()`, so that we can change the data model internally without causing any disruption.

amorsch: ok then -1 is ok.

PR #7484, opened 2021-11-01
DPL: provide callback on timeslice creation

ktf:


ktf:
@shahor02 this is probably all we need. You should be able to implement your own logic in the callback and simply override the various quantities in the DataHeader to match the state at data taking time.

ktf:
Error unrelated.

PR #7486, opened 2021-11-01
DPL: report rate metrics on a per device level

ktf:


awegrzyn (dismissed):

Framework/Core/src/CommonServices.cxx
```diff
@@ -527,22 +527,34 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
   stats.channelBytesIn.resize(device->fChannels.size());
   stats.channelBytesOut.resize(device->fChannels.size());
   size_t ci = 0;
+  auto lastTotalBytesIn = stats.totalBytesIn.load();
+  auto lastTotalBytesOut = stats.totalBytesOut.load();
```
davidrohr: shouldnt't these 2 be `.exchange(0)` instead of `.load()`?
Actually I don't fully understand why these are atomic variables in the first place, but in any case if we use atomics we should use proper atomic semantics. 

At second sight, also below we don't do proper atomic updates ...
I think it should rather be something like this (pseudo-code for only the in-part):
```
uint64_t sum = 0;
for (auto& channel : device->fChannels) sum += channel.second[0].GetBytesRx();
uint64_t lastBytes = stats.totalBytesIn.exchange(sum);
monitoring.send(sum - lastBytes);
```

Framework/Core/src/CommonServices.cxx
```diff
@@ -527,22 +527,34 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
...
                            o2::monitoring::Verbosity::Debug}
                       .addTag(Key::Subsystem, Value::DPL));
+    stats.totalBytesIn += newBytesIn;
+    stats.totalBytesOut += newBytesOut;
     stats.channelBytesOut[ci] = newBytesOut;
     stats.channelBytesIn[ci] = newBytesIn;
     ci++;
   }
 
+  monitoring.send(Metric{(float)(stats.totalBytesOut - lastTotalBytesOut) / 1000000.f / (timeSinceLastLongUpdate / 1000.f), "total_rate_out_mb_s"}
```
davidrohr: If I understand correctly the latency for aggregating `timeSinceLastLongUpdate` is minimum 30 seconds. I'd actually prefer a shorter granularity for the total in/out metric, perhaps 10 seconds?
I don't care so much for the per-channel metrics.

davidrohr (changes requested):
I would propose to combine this with #7453, and I have some minor comments below.

Framework/Core/src/CommonServices.cxx
ktf: They are atomic because otherwise we used to get a thread-sanitiser warning when running DPL multithreaded albeit I think it cannot actually be called in a multithreaded fashion, so I was just doing something quick. I will update it.

Framework/Core/src/CommonServices.cxx
ktf: ok, I will move it to be on the 5 seconds interval.

Framework/Core/src/CommonServices.cxx
```diff
@@ -517,32 +517,30 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
...
 
-  stats.channelBytesIn.resize(device->fChannels.size());
-  stats.channelBytesOut.resize(device->fChannels.size());
-  size_t ci = 0;
+  auto lastTotalBytesIn = 0;
+  auto lastTotalBytesOut = 0;
+  stats.totalBytesIn.exchange(lastTotalBytesIn);
+  stats.totalBytesOut.exchange(lastTotalBytesOut);
+  auto totalBytesIn = 0;
+  auto totalBytesOut = 0;
```
davidrohr: The auto variables here are not safe I think. They will become integers, but we want uint64_t.

davidrohr (changes requested):

Framework/Core/src/CommonServices.cxx
```diff
@@ -517,32 +517,30 @@ auto sendRelayerMetrics(ServiceRegistry& registry, DataProcessingStats& stats) -
...
-                      .addTag(Key::Subsystem, Value::DPL));
-    monitoring.send(Metric{(float)(newBytesIn - stats.channelBytesIn[ci]) / 1000000.f / (timeSinceLastLongUpdate / 1000.f),
-                           fmt::format("channel_{}_rate_out_mb_s", channel.first),
-                           o2::monitoring::Verbosity::Debug}
-                      .addTag(Key::Subsystem, Value::DPL));
-    stats.channelBytesOut[ci] = newBytesOut;
-    stats.channelBytesIn[ci] = newBytesIn;
-    ci++;
+    totalBytesIn += channel.second[0].GetBytesTx();
+    totalBytesOut += channel.second[0].GetBytesRx();
```
davidrohr: mixed up

PR #7487, opened 2021-11-01
TRD remove unused placeholder

martenole:
@jbarrella I would remove this for now, as it is not used anyway and otherwise we get errors from the CCDB when it tries to access the object. As a short term solution I have uploaded a dummy object with the lifetime of a couple days

### DIFF
--- a/Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
+++ b/Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
@@ -70,7 +70,6 @@ class TrackletTransformer
   float mXtb0;
 
   o2::trd::CalVdriftExB* mCalibration;
-  std::vector<float>* mT0;
 };
 
 } // namespace trd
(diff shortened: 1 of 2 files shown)


jbarrella:
Okay yeah, I am not sure when t0 calibration is going to happen... might be a while still so makes sense to remove this

bazinski (dismissed):

martenole:
Not sure why `CCDB/test/testCcdbApi.cxx` failed. Trying again the CI without any changes..

tdietel (approved):
I agree, it makes sense to remove this for now.

martenole:
Hi @davidrohr or @shahor02 
could this be merged please, since it is a trivial change and the o2-cs8 build only had a timeout in one of the tests, otherwise everything is passed.
This would also fix https://github.com/AliceO2Group/AliceO2/pull/7479#issuecomment-958796187

PR #7488, opened 2021-11-01
MC particles table: fill status codes and fix typo in flags

nburmaso:
Following PR #7483

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -658,8 +658,11 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
       }
       int statusCode = 0;
       uint8_t flags = 0;
-      if (!mcParticles[particle].isTransported()) {
+      if (!mcParticles[particle].isPrimary()) {
         flags |= 1 << 0; // mark as transported
+        statusCode = mcParticles[particle].getProcess();
+      } else {
+        statusCode = mcParticles[particle].getStatusCode();
       }
       if (source == 0) {
         flags |= 1 << 1; // mark as particle from background event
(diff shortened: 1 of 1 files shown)


sawenzel (dismissed):

sawenzel:
@amorsch : This is what we discussed. @jgrosseo : I think it would make sense to offer a (dynamic columns) API to allow extracting physics processes for secondaries (via the status code storage). If the particle is in fact a primary ... the physics code should be set to TMCProcess::kPrimary

jgrosseo:
I am confused now: Is the "from transport" bit now used for isPrimary?

Sure, we can add a dynamic column. If you give me the exact mapping, I'll add it.

sawenzel:
@jgrosseo : I think there is a "not" operator in front. So not primary is from transport.

sawenzel:
See also https://github.com/AliceO2Group/AliceO2/pull/7492 in relation to this.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -658,8 +658,11 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
       }
       int statusCode = 0;
       uint8_t flags = 0;
-      if (!mcParticles[particle].isTransported()) {
+      if (!mcParticles[particle].isPrimary()) {
         flags |= 1 << 0; // mark as transported
```
sawenzel: the comment is misleading. I would say `// mark as produced by transport`

jgrosseo: Yes, and also use o2::aod::mcparticle::enums::ProducedByTransport instead of `1 << 0`

nburmaso: Thank you for comments. I have cleaned up the code

PR #7489, opened 2021-11-01
DPL: use libuv to retrieve time

ktf:
Minimizes number of calls to `clock_gettime`.


ktf:
@TimoWilken could you check what is going on with the fullCI?

TimoWilken:
What was the problem? It's green now and I can't find any previous errors in the builder logs.

ktf:
It took over 4h to complete, apparently. Maybe we should check why.

PR #7110, opened 2021-09-23
Extending DPL input handling and message caching related to [O2 1325] 

matthiasrichter:
This is a prototype which also will allow with a few more changes to activate support for an updated O2 message model where split payloads can be transferred in a sequence with one preceding data header.

Has to be finally activated in `DataProcessingDevice` together with a workflow test.

For using this in the workflows together with Data Distribution, also the raw proxy needs to be extended, and for that a better encapsulation of the message model is in development based on the outcome of the prototype test.

PR #7493, opened 2021-11-02
DPL: Extending initial parsing of input messages

matthiasrichter:
This is a part of  #7110 in preparation for support of split payloads with pruned headers.


matthiasrichter:
@davidrohr @ktf @shahor02  I don't actually think, the fullCI failure is related to the PR, have you seen this `bad_alloc` in the its tracker before?

shahor02:
pinging to @mpoghos concerning the `bad alloc`

ktf:
I've never seen it myself.

shahor02:
used wrong name, now pinging to @mpuccio

PR #7495, opened 2021-11-02
Improve GPU memory benchmark

mconcas:
@davidrohr to keep track of the status:

- [x] sequential kernels should use that approach to split the access of the blocks among the chunks, concurrent kernels should stay as they are.
- [x] concurrent total throughput measurement should be based on a measurement of the host time.
- [x] benchmark should check that the chunk size does not go out of bounds of the allocated memory and do not intersect each others.
- [x] benchmarks to accept fractionary boundaries and sizes for chunks

### DIFF
--- a/GPU/GPUbenchmark/cuda/Kernels.cu
+++ b/GPU/GPUbenchmark/cuda/Kernels.cu
@@ -445,7 +445,6 @@ float GPUbenchmark<chunk_t>::runDistributed(void (*kernel)(chunk_t**, size_t*, T
   }
 
   // Setup
-
   chunk_t** block_ptr;
   size_t* block_size;
   GPUCHECK(cudaMalloc(reinterpret_cast<void**>(&block_ptr), 60 * sizeof(chunk_t*)));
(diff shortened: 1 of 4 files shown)


davidrohr:
thx, looks good so far!

GPU/GPUbenchmark/cuda/Kernels.cu
```diff
@@ -142,17 +142,16 @@ __global__ void rand_read_k(
...
   chunk_t sink{0};
-  auto* ptr = blockPtrs[blockIdx.x];
-  auto n = blockSizes[blockIdx.x];
+  chunk_t* ptr = block_ptr[blockIdx.x];
+  size_t n = block_size[blockIdx.x];
   for (size_t i = threadIdx.x; i < n; i += blockDim.x) {
     sink += ptr[i];
   }
-  blockPtrs[blockIdx.x][threadIdx.x] = sink;
+  block_ptr[blockIdx.x][threadIdx.x] = sink;
```
davidrohr: you can just use ptr here instead of block_ptr[blockIdx.x] :)

davidrohr:
thx, this looks good!
Will try to merge for tomorrow's nightly, then we can easily test it on the EPN tomorrow.

mconcas:
> thx, this looks good! Will try to merge for tomorrow's nightly, then we can easily test it on the EPN tomorrow.

thanks a lot!

davidrohr:
do you want to partially squash, or shall I squash the whole PR?

mconcas:
You can squash, np

PR #7497, opened 2021-11-02
Use mc flag to be able to run with data

chiarazampolli:
Done with @martenole . 
@sawenzel , @nburmaso , this is needed to create AO2D with the pilot beam data. It uses the MC flag to avoid processing MC information if it is not present.

### DIFF
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -191,13 +191,15 @@ typedef boost::unordered_map<Triplet_t, int, TripletHash, TripletEqualTo> Triple
 class AODProducerWorkflowDPL : public Task
 {
  public:
-  AODProducerWorkflowDPL(GID::mask_t src, std::shared_ptr<DataRequest> dataRequest) : mInputSources(src), mDataRequest(dataRequest) {}
+  AODProducerWorkflowDPL(GID::mask_t src, std::shared_ptr<DataRequest> dataRequest, bool useMC = true) : mInputSources(src), mDataRequest(dataRequest), mUseMC(useMC) {}
   ~AODProducerWorkflowDPL() override = default;
   void init(InitContext& ic) final;
   void run(ProcessingContext& pc) final;
   void endOfStream(framework::EndOfStreamContext& ec) final;
 
  private:
+  bool mUseMC = true;
+
   const float cSpeed = 0.029979246f; // speed of light in TOF units
 
   GID::mask_t mInputSources;
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -1067,40 +1067,42 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   // correct ordering when iterating over container elements
   std::map<std::pair<int, int>, int> mcColToEvSrc;
 
-  // TODO: figure out collision weight
-  float mcColWeight = 1.;
-  // filling mcCollision table
-  int nMCCollisions = mcContext->getNCollisions();
-  for (int iCol = 0; iCol < nMCCollisions; iCol++) {
-    auto time = mcRecords[iCol].getTimeNS();
-    auto globalBC = mcRecords[iCol].toLong();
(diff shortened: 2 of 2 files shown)


Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
```diff
@@ -191,13 +191,15 @@ typedef boost::unordered_map<Triplet_t, int, TripletHash, TripletEqualTo> Triple
...
  public:
-  AODProducerWorkflowDPL(GID::mask_t src, std::shared_ptr<DataRequest> dataRequest) : mInputSources(src), mDataRequest(dataRequest) {}
+  AODProducerWorkflowDPL(GID::mask_t src, std::shared_ptr<DataRequest> dataRequest, bool useMC = true) : mInputSources(src), mDataRequest(dataRequest), mUseMC(useMC) {}
   ~AODProducerWorkflowDPL() override = default;
   void init(InitContext& ic) final;
   void run(ProcessingContext& pc) final;
   void endOfStream(framework::EndOfStreamContext& ec) final;
 
  private:
+  bool mUseMC = true;
```
sawenzel, line 201: I guess we should then directly offer an option for the executable to set this from outside.

martenole: The `aod-producer-workflow` already offers this option, but is was so far only used by the readers: https://github.com/AliceO2Group/AliceO2/blob/36498389afcfff3925b78729dfb6ad990172d38d/Detectors/AOD/src/aod-producer-workflow.cxx#L30

chiarazampolli: As Ole said, the flag is currently an option from the workflow, but we pass it now explicitly to the AOD producer to be able do disable some functionalities (those related to the MC information).

shahor02:
@chiarazampolli could you also change the following:

1) substitute  https://github.com/AliceO2Group/AliceO2/blob/6a462f7b6c8cda85edf4205f40a2e9a2f3e3f511/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L977-L986
by 
```
  std::unique_ptr<o2::steer::MCKinematicsReader> mcReader;
  if (mUseMC) {
    mcReader = std::make_unique<o2::steer::MCKinematicsReader>("collisioncontext.root");
    LOG(DEBUG) << "FOUND " << mcReader->getDigitizationContext()->getEventRecords().size()
	       << " records" << mcReader->getDigitizationContext()->getEventParts().size() << " parts";
  }
  
  std::map<uint64_t, int> bcsMap;
  collectBCs(fddRecPoints, ft0RecPoints, fv0RecPoints, primVertices, mUseMC ? mcReader->getDigitizationContext().getEventRecords() : std::vector<o2::InteractionTimeRecord>{}, bcsMap);
```
2) change https://github.com/AliceO2Group/AliceO2/blob/6a462f7b6c8cda85edf4205f40a2e9a2f3e3f511/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L1074 to 
```
int nMCCollisions = mcReader->getDigitizationContext()->getNCollisions();
const auto& mcRecords = mcReader->getDigitizationContext()->getEventRecords();
const auto& mcParts = mcReader->getDigitizationContext()->getEventParts();
```
3) change in https://github.com/AliceO2Group/AliceO2/blob/6a462f7b6c8cda85edf4205f40a2e9a2f3e3f511/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L1093 to `mcReader->` and in https://github.com/AliceO2Group/AliceO2/blob/6a462f7b6c8cda85edf4205f40a2e9a2f3e3f511/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L1297 to `(*mcReader)`

`mcReader` to (*mcReader)

chiarazampolli:
Hi @shahor02 . 

The changes you asked for are in (but I committed before compiling :) ).

Chiara

sawenzel:
Is this tested/exercised as part of the full-system test? ... Ah I see #7500. So question answered.

davidrohr:
> Is this tested/exercised as part of the full-system test? ... Ah I see #7500. So question answered.

as of today the AOD creation is included in the full system test.

chiarazampolli:
Hi,
It was also tested with reconstructed data from pilot beam.
Chiara

PR #7498, opened 2021-11-02
Move dataRequest from init

chiarazampolli:


Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
```diff
@@ -123,15 +114,33 @@ bool MatchITSTPCQC::init()
 
 //__________________________________________________________
 
-void MatchITSTPCQC::run(o2::framework::ProcessingContext& ctx)
+void MatchITSTPCQC::initDataRequest()
 {
 
+  // initialize data request, if it was not already done
```
shahor02, line 120: Why this method is needed at all? I don't see where are you calling it. 
You have there `mDataRequest->requestTracks(mSrc, mUseMC);` but the https://github.com/AliceO2Group/AliceO2/blob/78a872a72192359f280eb0b9e2aa1424ad866fa3/Detectors/GlobalTrackingWorkflow/qc/src/ITSTPCMatchingQCSpec.cxx#L92 already contains it...

chiarazampolli: Hi, 

QC needs it, see https://github.com/AliceO2Group/QualityControl/pull/938

Chiara

PR #7499, opened 2021-11-02
Injection of custom DataHeader TF counters 

shahor02:


shahor02:
@ktf I've now the vector of DataHeader counters to be set via callback, need just a way to add callback function from the WorkflowSpec definition rather than from device init function.

shahor02:
Don't merge yet, even if passes all CIs

shahor02:
@ktf the way with workflow + device options did not work, it would require setting the correct file name for the global option of all workflows, because every workflow tries to set the callback with the global options from its own ConfigContext.
In the end, I got rid of the global options at all, including setting the linearly increasing `firstTForbit` via `orbit-offset-enumeration` and `orbit-multiplier-enumeration`. Instead, now I am passing equivalent `offset + increment*tfCounter` via the same callback also in case of MC.

PR #7500, opened 2021-11-02
Enable AOD production in the FST

shahor02:


pzhristov (dismissed):

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -447,6 +447,11 @@ workflow_has_parameter CALIB && has_detector_calib TPC && has_detectors TPC ITS
 [ -z "$ED_CLUSTERS" ] && ED_CLUSTERS=$TRACK_SOURCES
 workflow_has_parameter EVENT_DISPLAY && [ $NUMAID == 0 ] && add_W o2-eve-display "--display-tracks $ED_TRACKS --display-clusters $ED_CLUSTERS $EVE_CONFIG $DISABLE_MC" "$ITSMFT_FILES"
 
+# ---------------------------------------------------------------------------------------------------------------------
+# AOD
+[ -z "$AOD_INPUT" ] && AOD_INPUT=$TRACK_SOURCES
+has_detector_matching AOD && add_W o2-aod-producer-workflow "--info-sources $AOD_INPUT --disable-root-input --aod-writer-keep dangling --aod-writer-resfile "AO2D" --aod-writer-resmode UPDATE $DISABLE_MC"
```
davidrohr: can we change this to `workflow_has_parameter` since AOD is not a detector?

davidrohr (changes requested):

prodtests/full-system-test/dpl-workflow.sh
shahor02: 1st I did it via `workflow_has_parameter`, but the problem is that the parameters are set in advance rather than built in the dpl-workflow, e.g. how would you enable it by default in the FST and at the same time allow its disabling?
As for the semantics, PRIMVTX and SECVTX are also not detectors...

davidrohr: AliceO2Group/AliceO2#7508 would disable it in the FST
Agreed semantically for VTXs, but I didn't want to add yet another list for vertexers.

shahor02:
Don't merge until https://github.com/AliceO2Group/AliceO2/pull/7508 is merged

PR #7501, opened 2021-11-03
Add metadata to AOD

nburmaso:
@jgrosseo: This update partially covers [O2-2656](https://alice.its.cern.ch/jira/browse/O2-2656). TMap does not include `ProducerROOTVersion`, as I did not find a way to fetch it from the framework in format like `heads/v5-34-00-patches`.

@sawenzel: Metadata is written to AOD before any tables, so that TMap could be written separately outside of DF. Tables should be written by aod-writer in `update` mode, so that metadata is not purged.

This PR will be updated following changes in #7497 related to MC/RAW input data type.

I'm not sure if I write `TMap` object correctly, so please take a look

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -52,10 +52,13 @@
 #include "SimulationDataFormat/MCEventLabel.h"
 #include "SimulationDataFormat/MCTrack.h"
 #include "SimulationDataFormat/MCTruthContainer.h"
+#include "O2Version.h"
 #include "TMath.h"
 #include "MathUtils/Utils.h"
 #include "Math/SMatrix.h"
-#include <TMatrixD.h>
+#include "TMatrixD.h"
+#include "TString.h"
+#include "TObjString.h"
 #include <map>
 #include <unordered_map>
 #include <vector>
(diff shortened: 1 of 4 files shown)


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -903,6 +896,45 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
   // Needed by MCH track extrapolation
   o2::base::GeometryManager::loadGeometry();
 
+  // writing metadata if it's not yet in AOD file
+  // note: `--aod-writer-resmode "UPDATE"` have to be used,
+  //       so that metadata is not overwritten
+  mResFile += ".root";
+  auto* fResFile = TFile::Open(mResFile, "UPDATE");
```
jgrosseo, line 903: Shouldn't this be written by using some O2 framework functions instead of doing it manually?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1363,10 +1395,11 @@ DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src, bool useMC)
     "aod-producer-workflow",
     dataRequest->inputs,
     outputs,
-    AlgorithmSpec{adaptFromTask<AODProducerWorkflowDPL>(src, dataRequest)},
+    AlgorithmSpec{adaptFromTask<AODProducerWorkflowDPL>(src, dataRequest, resFile)},
     Options{
       ConfigParamSpec{"aod-timeframe-id", VariantType::Int64, -1L, {"Set timeframe number"}},
       ConfigParamSpec{"enable-truncation", VariantType::Int, 1, {"Truncation parameter: 1 -- on, != 1 -- off"}},
+      ConfigParamSpec{"prod-tags", VariantType::String, "LHC21Axx,pass1,LHC15o,pass1", {"Comma separated list of production tags: `LPMProductionTag,AnchorPassName,AnchorProduction,RecoPassName`"}},
```
jgrosseo, line 1402: I propose one configSpec for each of the four strings

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: It would be nice, but I'm not aware of a method that would allow writing an object outside of DF folder. Also, the code in this PR got mixed with some older code, so I will make another PR, where we can continue discussion :)

jgrosseo: @ktf could you comment?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: I agree, this would be better. Done in #7503

PR #7502, opened 2021-11-03
fix in tof latency window

noferini:


noferini:
@chiarazampolli 
This fixes the TOF-track time misalignement which affects the matching

noferini:
It would be nice to have it in the nightly by tomorrow (epn cvmfs)

PR #7503, opened 2021-11-03
Add metadata to AOD

nburmaso:
@jgrosseo, @sawenzel: This is a fixed PR, following #7501

### DIFF
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -208,10 +208,10 @@ class AODProducerWorkflowDPL : public Task
   int mTruncate{1};
   int mRecoOnly{0};
   TString mResFile{"AO2D"};
-  TString mLPMProdTag{"LHC21Axx"};
-  TString mAnchorPass{"pass1"};
-  TString mAnchorProd{"LHC15o"};
-  TString mRecoPass{"pass1"};
+  TString mLPMProdTag{""};
+  TString mAnchorPass{""};
+  TString mAnchorProd{""};
+  TString mRecoPass{""};
   TStopwatch mTimer;
 
   // unordered map connects global indices and table indices of barrel tracks
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -34,6 +34,7 @@
 #include "ReconstructionDataFormats/VtxTrackIndex.h"
 #include "SimulationDataFormat/MCCompLabel.h"
 #include "Steer/MCKinematicsReader.h"
+#include "TMap.h"
 #include "TStopwatch.h"
 
 #include <boost/functional/hash.hpp>
@@ -191,7 +192,7 @@ typedef boost::unordered_map<Triplet_t, int, TripletHash, TripletEqualTo> Triple
 class AODProducerWorkflowDPL : public Task
 {
  public:
-  AODProducerWorkflowDPL(GID::mask_t src, std::shared_ptr<DataRequest> dataRequest) : mInputSources(src), mDataRequest(dataRequest) {}
+  AODProducerWorkflowDPL(GID::mask_t src, std::shared_ptr<DataRequest> dataRequest, std::string resFile) : mInputSources(src), mDataRequest(dataRequest), mResFile(resFile) {}
   ~AODProducerWorkflowDPL() override = default;
   void init(InitContext& ic) final;
   void run(ProcessingContext& pc) final;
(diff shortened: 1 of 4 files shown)


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -901,6 +908,34 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
...
+    if (!fResFile->FindObjectAny("metaData")) {
+      // populating metadata map
+      TString dataType = mUseMC ? "MC" : "RAW";
+      mMetaData.Add(new TObjString("DataType"), new TObjString(dataType));
+      mMetaData.Add(new TObjString("Run"), new TObjString("3"));
+      TString converterVersion = "o2 ";
+      converterVersion += o2::fullVersion();
+      converterVersion += " ; root ";
+      converterVersion += ROOT_RELEASE;
+      mMetaData.Add(new TObjString("Run3ConverterVersion"), new TObjString(converterVersion));
```
jgrosseo: You can rename this to "O2Version" and just put `o2::fullVersion` there, then let's make an additional one "ROOTVersion" for `ROOT_RELEASE`

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
```diff
@@ -206,6 +207,11 @@ class AODProducerWorkflowDPL : public Task
   int64_t mTFNumber{-1};
   int mTruncate{1};
   int mRecoOnly{0};
+  TString mResFile{"AO2D"};
+  TString mLPMProdTag{"LHC21Axx"};
+  TString mAnchorPass{"pass1"};
+  TString mAnchorProd{"LHC15o"};
+  TString mRecoPass{"pass1"};
```
jgrosseo: Please put empty defaults. We should not have meaningful defaults (like it was for the run number), this causes trouble often. ;)

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -901,6 +908,34 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
...
+      mMetaData.Add(new TObjString("RecoPassName"), new TObjString(mRecoPass));
+      mMetaData.Add(new TObjString("AnchorProduction"), new TObjString(mAnchorProd));
+      mMetaData.Add(new TObjString("AnchorPassName"), new TObjString(mAnchorPass));
+      mMetaData.Add(new TObjString("LPMProductionTag"), new TObjString(mLPMProdTag));
+      LOGF(info, "Metadata: writing into %s", mResFile);
+      fResFile->WriteObject(&mMetaData, "metaData");
+    } else {
+      LOGF(info, "Metadata: target file not found or metadata is already written");
```
jgrosseo: I think this is not the case of file not found. It is just the if  (!fResFile->FindObjectAny("metaData"))
Are there valid cases when this is allowed? Otherwise it should be `fatal`
I think it should also be `fatal` if the output file is not found

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1366,10 +1401,14 @@ DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src, bool useMC)
...
     outputs,
-    AlgorithmSpec{adaptFromTask<AODProducerWorkflowDPL>(src, dataRequest, useMC)},
+    AlgorithmSpec{adaptFromTask<AODProducerWorkflowDPL>(src, dataRequest, resFile, useMC)},
     Options{
       ConfigParamSpec{"aod-timeframe-id", VariantType::Int64, -1L, {"Set timeframe number"}},
       ConfigParamSpec{"enable-truncation", VariantType::Int, 1, {"Truncation parameter: 1 -- on, != 1 -- off"}},
+      ConfigParamSpec{"lpmp-prod-tag", VariantType::String, "LHC21Axx", {"LPMProductionTag"}},
+      ConfigParamSpec{"anchor-pass", VariantType::String, "pass1", {"AnchorPassName"}},
+      ConfigParamSpec{"anchor-prod", VariantType::String, "LHC15o", {"AnchorProduction"}},
+      ConfigParamSpec{"reco-pass", VariantType::String, "LHC15o", {"RecoPassName"}},
```
jgrosseo: Empty defaults also here

nburmaso:
Hi @jgrosseo. Thank you for taking a look at the code. I have fixed it according to your comments

sawenzel:
@nburmaso : Can we remove the draft status to trigger the CI?

nburmaso:
@sawenzel, yes sure. Also, notice that TMap is written separately from tables, and some O2DPG scripts may need tuning because of it (e.g. due to files merging)

sawenzel:
@nburmaso : Could you suggest necessary changes in O2DPG yourself? It's not clear to me what needs to be done.

nburmaso:
@sawenzel: I'm not sure myself if changes are even needed, it's only my guess. Sorry if it was too bothersome :)

jgrosseo:
The merger already deals with the meta data correctly.

sawenzel (approved):

PR #7507, opened 2021-11-03
Remove TrackCov from FullTracks

jgrosseo:
To avoid that it is used by habit.

Needs https://github.com/AliceO2Group/O2Physics/pull/246 for CI


jgrosseo:
FullCI OK and changes minimal. Merging.

PR #7510, opened 2021-11-03
Adapt workflow for async processing of pilot beam data

chiarazampolli:
With @martenole

### DIFF
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -6,6 +6,10 @@ command -v zsh > /dev/null 2>&1 && MYDIR=$(dirname $(zsh -c 'echo ${0:A}' "$0"))
 test -z ${MYDIR+x} && MYDIR="$(dirname $(readlink -f $0))"
 source $MYDIR/setenv.sh
 
+if [ -z $FILEWORKDIRRUN ]; then FILEWORKDIRRUN=$FILEWORKDIR; fi              # directory where to find the run-related files (grp, collision context)
+
+if [ ! -z $ISPILOTBEAMDATA ]; then BEAMTYPE=pp; fi  # pilot beam was pp
+
 # ---------------------------------------------------------------------------------------------------------------------
 #Some additional settings used in this workflow
 if [[ -z $OPTIMIZED_PARALLEL_ASYNC ]]; then OPTIMIZED_PARALLEL_ASYNC=0; fi     # Enable tuned process multiplicities for async processing on the EPN
@@ -128,6 +132,11 @@ fi
 
 has_processing_step ENTROPY_ENCODER && has_detector_ctf TPC && GPU_OUTPUT+=",compressed-clusters-ctf"
 
+if [ -z $DISABLE_ROOT_OUTPUT ]; then
+  # enable only if root output is written, because it slows down the processing
+  GPU_OUTPUT+=",send-clusters-per-sector"
+fi
+
 has_detector_flp_processing CPV && CPV_INPUT=digits
 ! has_detector_flp_processing TOF && TOF_CONFIG+=" --ignore-dist-stf"
 
@@ -212,7 +221,11 @@ get_N() # USAGE: get_N [processor-name] [DETECTOR_NAME] [RAW|CTF|REST] [optional
   local NAME_DET="MULTIPLICITY_FACTOR_DETECTOR_$2"
   local NAME_PROC="MULTIPLICITY_FACTOR_PROCESS_${1//-/_}"
   local NAME_DEFAULT="N_$4"
-  echo $1:${!NAME_PROC:-$((${!NAME_FACTOR} * ${!NAME_DET:-1} * ${!NAME_DEFAULT:-1}))}
+  if [ -z $DISABLEPIPELINE ]; then
+    echo $1:${!NAME_PROC:-$((${!NAME_FACTOR} * ${!NAME_DET:-1} * ${!NAME_DEFAULT:-1}))}
+  else
+    echo $1:1
+  fi
 }
 
 math_max()
@@ -360,7 +373,7 @@ if [[ $CTFINPUT == 0 ]]; then
   has_detector TRD && add_W o2-trd-datareader "$TRD_DECODER_OPTIONS --pipeline $(get_N trd-datareader TRD RAW TRDRAWDEC)" "" 0
   has_detector ZDC && add_W o2-zdc-raw2digits "$DISABLE_ROOT_OUTPUT --pipeline $(get_N zdc-datareader-dpl ZDC RAW)"
   has_detector HMP && add_W o2-hmpid-raw-to-digits-stream-workflow "--pipeline $(get_N HMP-RawStreamDecoder HMP RAW)"
-  has_detector CTP && add_W o2-ctp-reco-workflow "--pipeline $(get_N CTP-RawStreamDecoder CTP RAW)"
+  has_detector CTP && add_W o2-ctp-reco-workflow "$DISABLE_ROOT_OUTPUT --pipeline $(get_N CTP-RawStreamDecoder CTP RAW)"
   has_detector PHS && ! has_detector_flp_processing PHS && add_W o2-phos-reco-workflow "--input-type raw --output-type cells --disable-root-input $DISABLE_ROOT_OUTPUT --pipeline $(get_N PHOSRawToCellConverterSpec PHS REST) $DISABLE_MC"
   has_detector CPV && add_W o2-cpv-reco-workflow "--input-type $CPV_INPUT --output-type clusters --disable-root-input $DISABLE_ROOT_OUTPUT --pipeline $(get_N CPVRawToDigitConverterSpec CPV REST),$(get_N CPVClusterizerSpec CPV REST) $DISABLE_MC"
   has_detector EMC && ! has_detector_flp_processing EMC && add_W o2-emcal-reco-workflow "--input-type raw --output-type cells $EMCRAW2C_CONFIG $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N EMCALRawToCellConverterSpec EMC REST EMCREC)"
@@ -368,7 +381,15 @@ fi
 
 # ---------------------------------------------------------------------------------------------------------------------
 # Common reconstruction workflows
-(has_detector_reco TPC || has_detector_ctf TPC) && WORKFLOW+="o2-gpu-reco-workflow ${ARGS_ALL//-severity $SEVERITY/-severity $SEVERITY_TPC} --input-type=$GPU_INPUT $DISABLE_MC --output-type $GPU_OUTPUT --pipeline gpu-reconstruction:${N_TPCTRK:-1} $GPU_CONFIG $ARGS_EXTRA_PROCESS_o2_gpu_reco_workflow --configKeyValues \"$ARGS_ALL_CONFIG;GPU_global.deviceType=$GPUTYPE;GPU_proc.debugLevel=0;$GPU_CONFIG_KEY;$GPU_EXTRA_CONFIG;$CONFIG_EXTRA_PROCESS_o2_gpu_reco_workflow\" | "
+PIPELINEGPU=1
+if [ -z $DISABLEPIPELINE ]; then
+  PIPELINEGPU=${N_TPCTRK:-1}
+fi
+if [ ! -z $ISPILOTBEAMDATA ]; then
+  ITSTPC_EXTRA_CONFIG="tpcitsMatch.maxVDriftUncertainty=0.2;tpcitsMatch.safeMarginTimeCorrErr=10.;tpcitsMatch.cutMatchingChi2=1000;tpcitsMatch.crudeAbsDiffCut[0]=5;tpcitsMatch.crudeAbsDiffCut[1]=5;tpcitsMatch.crudeAbsDiffCut[2]=0.3;tpcitsMatch.crudeAbsDiffCut[3]=0.3;tpcitsMatch.crudeAbsDiffCut[4]=10;tpcitsMatch.crudeNSigma2Cut[0]=200;tpcitsMatch.crudeNSigma2Cut[1]=200;tpcitsMatch.crudeNSigma2Cut[2]=200;tpcitsMatch.crudeNSigma2Cut[3]=200;tpcitsMatch.crudeNSigma2Cut[4]=900"
+fi
+echo extra config = $ITSTPC_EXTRA_CONFIG, ispilot = $ISPILOTBEAMDATA
+(has_detector_reco TPC || has_detector_ctf TPC) && WORKFLOW+="o2-gpu-reco-workflow ${ARGS_ALL//-severity $SEVERITY/-severity $SEVERITY_TPC} --input-type=$GPU_INPUT $DISABLE_MC --output-type $GPU_OUTPUT --pipeline gpu-reconstruction:$PIPELINEGPU $GPU_CONFIG $ARGS_EXTRA_PROCESS_o2_gpu_reco_workflow --configKeyValues \"$ARGS_ALL_CONFIG;GPU_global.deviceType=$GPUTYPE;GPU_proc.debugLevel=0;$GPU_CONFIG_KEY;$GPU_EXTRA_CONFIG;$CONFIG_EXTRA_PROCESS_o2_gpu_reco_workflow\" | "
 (has_detector_reco TOF || has_detector_ctf TOF) && add_W o2-tof-reco-workflow "$TOF_CONFIG --input-type $TOF_INPUT --output-type $TOF_OUTPUT --disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N tof-compressed-decoder TOF RAW),$(get_N TOFClusterer TOF REST)"
 has_detector_reco ITS && add_W o2-its-reco-workflow "--trackerCA $ITS_CONFIG $DISABLE_MC --clusters-from-upstream $DISABLE_ROOT_OUTPUT --pipeline $(get_N its-tracker ITS REST ITSTRK)" "$ITS_CONFIG_KEY;$ITSMFT_FILES"
 has_detectors_reco ITS TPC && has_detector_matching ITSTPC && add_W o2-tpcits-match-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N itstpc-track-matcher MATCH REST TPCITS)" "$ITSTPC_EXTRA_CONFIG;$ITSMFT_FILES"
@@ -376,6 +397,9 @@ has_detector_reco FT0 && add_W o2-ft0-reco-workflow "--disable-root-input $DISAB
 has_detector_reco TRD && add_W o2-trd-tracklet-transformer "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC $TRD_TRANSFORMER_CONFIG --pipeline $(get_N TRDTRACKLETTRANSFORMER TRD REST TRDTRK)"
 has_detectors_reco TRD TPC ITS && [[ ! -z "$TRD_SOURCES" ]] && add_W o2-trd-global-tracking "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC $TRD_CONFIG --track-sources $TRD_SOURCES" "$TRD_CONFIG_KEY;$ITSMFT_FILES"
 has_detectors_reco TOF TRD TPC ITS && [[ ! -z "$TOF_SOURCES" ]] && add_W o2-tof-matcher-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC --track-sources $TOF_SOURCES --pipeline $(get_N tof-matcher TOF REST TOFMATCH)" "$ITSMFT_FILES"
+if [ -z "$DISABLE_ROOT_OUTPUT" ]; then
+  has_detectors TPC && add_W o2-tpc-reco-workflow "--input-type pass-through --output-type clusters,tracks,send-clusters-per-sector $DISABLE_MC"
+fi
 
 # ---------------------------------------------------------------------------------------------------------------------
 # Reconstruction workflows normally active only in async mode in async mode ($LIST_OF_ASYNC_RECO_STEPS), but can be forced via $WORKFLOW_EXTRA_PROCESSING_STEPS
@@ -389,6 +413,7 @@ has_detectors_reco MFT MCH && has_detector_matching MFTMCH && add_W o2-globalfwd
 
 if [[ $BEAMTYPE != "cosmic" ]]; then
   has_detectors_reco ITS && has_detector_matching PRIMVTX && add_W o2-primary-vertexing-workflow "$DISABLE_MC --disable-root-input $DISABLE_ROOT_OUTPUT $PVERTEX_CONFIG --pipeline $(get_N primary-vertexing MATCH REST)" "$PVERTEX_EXTRA_CONFIG"
+
   has_detectors_reco ITS && has_detector_matching SECVTX && add_W o2-secondary-vertexing-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT --vertexing-sources $TRACK_SOURCES --pipeline $(get_N secondary-vertexing MATCH REST)" "$SVERTEX_EXTRA_CONFIG"
 fi
 
@@ -448,7 +473,7 @@ workflow_has_parameter AOD && add_W o2-aod-producer-workflow "--info-sources $AO
 
 # ---------------------------------------------------------------------------------------------------------------------
 # Quality Control
-workflow_has_parameter QC && { source $MYDIR/qc-workflow.sh; [[ $? != 0 ]] && exit 1; }
+workflow_has_parameter QC && { source $O2DATAPROCESSING_ROOT/production/qc-workflow.sh; [[ $? != 0 ]] && exit 1; }
 
 # ---------------------------------------------------------------------------------------------------------------------
 # DPL run binary
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -8,8 +8,6 @@ source $MYDIR/setenv.sh
 
 if [ -z $FILEWORKDIRRUN ]; then FILEWORKDIRRUN=$FILEWORKDIR; fi              # directory where to find the run-related files (grp, collision context)
 
-if [ ! -z $ISPILOTBEAMDATA ]; then BEAMTYPE=pp; fi  # pilot beam was pp
-
 # ---------------------------------------------------------------------------------------------------------------------
 #Some additional settings used in this workflow
 if [[ -z $OPTIMIZED_PARALLEL_ASYNC ]]; then OPTIMIZED_PARALLEL_ASYNC=0; fi     # Enable tuned process multiplicities for async processing on the EPN
@@ -81,7 +79,7 @@ fi
 [[ $NUMAGPUIDS != 0 ]] && ARGS_ALL+=" --child-driver 'numactl --membind $NUMAID --cpunodebind $NUMAID'"
 [[ ! -z $TIMEFRAME_RATE_LIMIT ]] && [[ $TIMEFRAME_RATE_LIMIT != 0 ]] && ARGS_ALL+=" --timeframes-rate-limit $TIMEFRAME_RATE_LIMIT --timeframes-rate-limit-ipcid $NUMAID"
 
-ARGS_ALL_CONFIG="NameConf.mDirGRP=$FILEWORKDIR;NameConf.mDirGeom=$FILEWORKDIR;NameConf.mDirCollContext=$FILEWORKDIR;NameConf.mDirMatLUT=$FILEWORKDIR;keyval.input_dir=$FILEWORKDIR;keyval.output_dir=/dev/null;$ALL_EXTRA_CONFIG"
+ARGS_ALL_CONFIG="NameConf.mDirGeom=$FILEWORKDIR;NameConf.mDirMatLUT=$FILEWORKDIR;NameConf.mDirCollContext=$FILEWORKDIRRUN;NameConf.mDirGRP=$FILEWORKDIRRUN;keyval.input_dir=$FILEWORKDIR;keyval.output_dir=/dev/null;$ALL_EXTRA_CONFIG"
 
 # ---------------------------------------------------------------------------------------------------------------------
 # Set some individual workflow arguments depending on configuration
@@ -132,7 +130,7 @@ fi
 
 has_processing_step ENTROPY_ENCODER && has_detector_ctf TPC && GPU_OUTPUT+=",compressed-clusters-ctf"
 
-if [ -z $DISABLE_ROOT_OUTPUT ]; then
+if [[ -z $DISABLE_ROOT_OUTPUT ]]; then
   # enable only if root output is written, because it slows down the processing
   GPU_OUTPUT+=",send-clusters-per-sector"
 fi
@@ -203,6 +201,7 @@ for det in `echo $LIST_OF_DETECTORS | sed "s/,/ /g"`; do
     has_detector_reco $det && add_comma_separated TRACK_SOURCES "$det"
   fi
 done
+[[ -z $VERTEXING_SOURCES ]] && VERTEXING_SOURCES="$TRACK_SOURCES"
 PVERTEX_CONFIG="--vertexing-sources $TRACK_SOURCES --vertex-track-matching-sources $TRACK_SOURCES"
 
 # this option requires well calibrated timing beween different detectors, at the moment suppress it
@@ -221,11 +220,7 @@ get_N() # USAGE: get_N [processor-name] [DETECTOR_NAME] [RAW|CTF|REST] [optional
   local NAME_DET="MULTIPLICITY_FACTOR_DETECTOR_$2"
   local NAME_PROC="MULTIPLICITY_FACTOR_PROCESS_${1//-/_}"
   local NAME_DEFAULT="N_$4"
-  if [ -z $DISABLEPIPELINE ]; then
-    echo $1:${!NAME_PROC:-$((${!NAME_FACTOR} * ${!NAME_DET:-1} * ${!NAME_DEFAULT:-1}))}
-  else
-    echo $1:1
-  fi
+  echo $1:${!NAME_PROC:-$((${!NAME_FACTOR} * ${!NAME_DET:-1} * ${!NAME_DEFAULT:-1}))}
 }
 
 math_max()
@@ -373,7 +368,7 @@ if [[ $CTFINPUT == 0 ]]; then
   has_detector TRD && add_W o2-trd-datareader "$TRD_DECODER_OPTIONS --pipeline $(get_N trd-datareader TRD RAW TRDRAWDEC)" "" 0
   has_detector ZDC && add_W o2-zdc-raw2digits "$DISABLE_ROOT_OUTPUT --pipeline $(get_N zdc-datareader-dpl ZDC RAW)"
   has_detector HMP && add_W o2-hmpid-raw-to-digits-stream-workflow "--pipeline $(get_N HMP-RawStreamDecoder HMP RAW)"
-  has_detector CTP && add_W o2-ctp-reco-workflow "$DISABLE_ROOT_OUTPUT --pipeline $(get_N CTP-RawStreamDecoder CTP RAW)"
+  has_detector CTP && add_W o2-ctp-reco-workflow "--pipeline $(get_N CTP-RawStreamDecoder CTP RAW)"
   has_detector PHS && ! has_detector_flp_processing PHS && add_W o2-phos-reco-workflow "--input-type raw --output-type cells --disable-root-input $DISABLE_ROOT_OUTPUT --pipeline $(get_N PHOSRawToCellConverterSpec PHS REST) $DISABLE_MC"
   has_detector CPV && add_W o2-cpv-reco-workflow "--input-type $CPV_INPUT --output-type clusters --disable-root-input $DISABLE_ROOT_OUTPUT --pipeline $(get_N CPVRawToDigitConverterSpec CPV REST),$(get_N CPVClusterizerSpec CPV REST) $DISABLE_MC"
   has_detector EMC && ! has_detector_flp_processing EMC && add_W o2-emcal-reco-workflow "--input-type raw --output-type cells $EMCRAW2C_CONFIG $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N EMCALRawToCellConverterSpec EMC REST EMCREC)"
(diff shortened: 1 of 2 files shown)


prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -6,6 +6,8 @@ command -v zsh > /dev/null 2>&1 && MYDIR=$(dirname $(zsh -c 'echo ${0:A}' "$0"))
 test -z ${MYDIR+x} && MYDIR="$(dirname $(readlink -f $0))"
 source $MYDIR/setenv.sh
 
+if [ -z $FILEWORKDIRRUN ]; then FILEWORKDIRRUN=$FILEWORKDIR; fi              # directory where to find the run-related files (grp, collision context)
```
shahor02, line 9: Why do we need 2 directories, what is the difference between these two?

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -125,17 +127,27 @@ else
...
+      [ -z ${ITS_CONFIG+x} ] && ITS_CONFIG=" --tracking-mode sync_misaligned"
+    else
+      ITS_CONFIG_KEY+="fastMultConfig.cutMultClusLow=1;fastMultConfig.cutMultClusHigh=2000;fastMultConfig.cutMultVtxHigh=500;ITSVertexerParam.phiCut=0.5;ITSVertexerParam.clusterContributorsCut=3;ITSVertexerParam.tanLambdaCut=0.2"
+      [ -z ${ITS_CONFIG+x} ] && ITS_CONFIG=" --tracking-mode async"
+    fi
   elif [ $BEAMTYPE == "cosmic" ]; then
     [ -z ${ITS_CONFIG+x} ] && ITS_CONFIG=" --tracking-mode cosmics"
   else
-    [ -z ${ITS_CONFIG+x} ] && ITS_CONFIG=" --tracking-mode async"
+    [ -z ${ITS_CONFIG+x} ] && ITS_CONFIG=" --tracking-mode sync_misaligned"
```
shahor02: all these parameters can be set via env. vars, as it was done during the pilot beam, why do we need to hardcode them?

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -125,17 +127,27 @@ else
...
   fi
 fi
 
 has_processing_step ENTROPY_ENCODER && has_detector_ctf TPC && GPU_OUTPUT+=",compressed-clusters-ctf"
 
+if [ -z $DISABLE_ROOT_OUTPUT ]; then
+  # enable only if root output is written, because it slows down the processing
+  GPU_OUTPUT+=",send-clusters-per-sector"
```
shahor02, line 135: actually, I would set `send-clusters-per-sector` by default

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -202,7 +214,13 @@ for det in `echo $LIST_OF_DETECTORS | sed "s/,/ /g"`; do
...
   fi
 done
-PVERTEX_CONFIG="--vertexing-sources $TRACK_SOURCES --vertex-track-matching-sources $TRACK_SOURCES"
+
+if [ -z $ISPILOTDATA ]; then
+  PVERTEX_CONFIG="--vertexing-sources $TRACK_SOURCES --vertex-track-matching-sources $TRACK_SOURCES"
+else
+  PVERTEX_CONFIG="--vertexing-sources ITS --vertex-track-matching-sources ITS --hbfutils-config none"
+  PVERTEX_EXTRA_CONFIG="pvertexer.nSigmaTimeCut=100;pvertexer.dbscanMaxDist2=30;pvertexer.dcaTolerance=3.;pvertexer.pullIniCut=100;pvertexer.addZSigma2=0.1;pvertexer.tukey=20.;pvertexer.addZSigma2Debris=0.01;pvertexer.addTimeSigma2Debris=1.;pvertexer.maxChi2Mean=30;"
+fi
```
shahor02: this too can be set via env.vars.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: This is because the GRP and collision context are run dependent, while the geometry, dictionaries... are not. E.g. on the gird, for the processing of the pilot beam data, we can copy only 1x the files that are in common.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: We adapted to the script, also so to reduce to the minimum the env variables that we pass, which is error prone.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Ok for me. But what about the "slow-down" hinted at by the comment?

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: BTW, the syntax mean that if they are already there, what is defined is taken before the hardcoded values, so it is already what yo suggest.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Would setting it only if the env variables are not already set be ok for you? I would keep it in the script, to have less env variables when running.

prodtests/full-system-test/dpl-workflow.sh
shahor02: Aren't we going to upload external files per job as we were doing in run2? OK, if we read from predefined directories, this change makes sense.

prodtests/full-system-test/dpl-workflow.sh
shahor02: But you can define these extra params in a separate script which calls dpl-workflow. It is much cleaner than patching general script by custom settings. Also, I am sure the custom parameters for the prim.vertexin and TPC/ITS matching we will use in the async pass will not be the same quick and dirty settings used in the sync. reco.

prodtests/full-system-test/dpl-workflow.sh
shahor02: For pp this slowdown will be insignificant (and we anyway not going to write routinely root trees), in general, for PbPb after some number of events per TF the trees written w/o this option become unreadable.

prodtests/full-system-test/dpl-workflow.sh
shahor02: see above.

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -125,17 +129,27 @@ else
   if [ $BEAMTYPE == "PbPb" ]; then
     [ -z ${ITS_CONFIG+x} ] && ITS_CONFIG=" --tracking-mode async"
   elif [ $BEAMTYPE == "pp" ]; then
-    ITS_CONFIG_KEY+="fastMultConfig.cutMultClusLow=1;fastMultConfig.cutMultClusHigh=2000;fastMultConfig.cutMultVtxHigh=500;ITSVertexerParam.phiCut=0.5;ITSVertexerParam.clusterContributorsCut=3;ITSVertexerParam.tanLambdaCut=0.2"
-    [ -z ${ITS_CONFIG+x} ] && ITS_CONFIG=" --tracking-mode async"
+    if [ -z $ISPILOTBEAMDATA ]; then # configuration of ITS tracking for pilot beam data
+      ITS_CONFIG_KEY+="ITSVertexerParam.phiCut=0.5;ITSVertexerParam.clusterContributorsCut=3;ITSVertexerParam.tanLambdaCut=0.2;ITSCATrackerParam.useDiamond=true;ITSCATrackerParam.diamondPos[2]=35.;ITSCATrackerParam.pvRes=10."
```
shahor02: BTW, probably @mpuccio should comment, but this shifted IP Z `ITSCATrackerParam.diamondPos[2]=35` was valid only in 1st 2 days, I don't see the point of reconstructing these data anyway.

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -125,17 +129,27 @@ else
...
+      [ -z ${ITS_CONFIG+x} ] && ITS_CONFIG=" --tracking-mode sync_misaligned"
+    else
+      ITS_CONFIG_KEY+="fastMultConfig.cutMultClusLow=1;fastMultConfig.cutMultClusHigh=2000;fastMultConfig.cutMultVtxHigh=500;ITSVertexerParam.phiCut=0.5;ITSVertexerParam.clusterContributorsCut=3;ITSVertexerParam.tanLambdaCut=0.2"
```
shahor02: actually, this line should have not been there, we don't need any selection in the async mode, could you simply suppress it?

shahor02: could you please set `fastMultConfig.cutMultClusLow=-1`? With the current noise, the estimated multiplicity cuts all events.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Sorry, I said it wrong, what I meant is when we are not on the grid, one does not need to copy the common files multiple times. On the grid they will be copied on the worker node anyway - I would prefer to avoid reading remotely.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Yes, an extra script for env variables is what I am planning to have anyway - we will need it e.g. to define which detectors are in the workflow and other things. But if we can get a default of the settings, isn't it practical to have them directly in the script? 
I am aware that settings will change, so we should always set the parameters only if they are not set from outside. I will check and fix where it is not done.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Ok, then I will change it.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: As I mentioned above, I would give priority to what comes from outside, but also have some defaults in case nothing is provided. 
A sort of side comment: from past experience having settings only on the grid is not always easy to trace afterwards.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: All the settings need to be vetted by the experts, sure.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Sure.

chiarazampolli: Yes, I will.

prodtests/full-system-test/dpl-workflow.sh
davidrohr: I think for now it is ok, eventually we should anyway read most objects from the CCDB, then we don't need this any more.
Instead of creating copies locally, one could also just link the files locally and then work with a single `FILEWORKDIR`. That might avoid confusion. E.g. the matlut.root file is optional. If one puts it to the wrong folder, it will not be used.

prodtests/full-system-test/dpl-workflow.sh
shahor02: Well, the dpl-workflow is a general purpose script, I don't see what we gain by hardcoding the custom pilot beam option in it instead of defining the same env. var in the external steering script. But we certainly increase its complexity by putting there custom settings. Also, I am not sure we will use the same settings for different periods of this data...

prodtests/full-system-test/dpl-workflow.sh
davidrohr: I am not sure if we want such special setting for few runs in this script. They can also be set externally via
```
$CONFIG_EXTRA_PROCESS_o2_its_reco_workflow="ITSVertexerParam.phiCut=0.5;ITSVertexerParam.clusterContributorsCut=3;ITSVertexerParam.tanLambdaCut=0.2;ITSCATrackerParam.useDiamond=true;ITSCATrackerParam.diamondPos[2]=35.;ITSCATrackerParam.pvRes=10."
```

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -6,6 +6,10 @@ command -v zsh > /dev/null 2>&1 && MYDIR=$(dirname $(zsh -c 'echo ${0:A}' "$0"))
 test -z ${MYDIR+x} && MYDIR="$(dirname $(readlink -f $0))"
 source $MYDIR/setenv.sh
 
+if [ -z $FILEWORKDIRRUN ]; then FILEWORKDIRRUN=$FILEWORKDIR; fi              # directory where to find the run-related files (grp, collision context)
+
+if [ ! -z $ISPILOTBEAMDATA ]; then BEAMTYPE=pp; fi  # pilot beam was pp
```
davidrohr: I think the `BEAMTYPE` should come externally, and not be changed here.
BTW. which this check also `ISPILOTBEAMDATA=0` will lead to `BEAMTYPE=pp`

prodtests/full-system-test/dpl-workflow.sh
davidrohr: Please do not set this by default. It slows down the processing, on the sending and the reading side, and it doubles (temporarily) the SHM that is used, since the data is produced in a linear buffer during clusterization, and then it is copied to per-sector buffers for the ROOT output. And actually, also all consumers will become slower, since they will read the per-sector buffers and then copy it back to a linear buffer.
E.g. it would probably break the full system test due to the slow down.
I think it should only be enabled when the ROOT_OUTPUT is enabled.

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -395,6 +434,7 @@ has_detectors_reco MFT MCH && has_detector_matching MFTMCH && add_W o2-globalfwd
 
 if [ $BEAMTYPE != "cosmic" ]; then
   has_detectors_reco ITS && has_detector_matching PRIMVTX && add_W o2-primary-vertexing-workflow "$DISABLE_MC --disable-root-input $DISABLE_ROOT_OUTPUT $PVERTEX_CONFIG --pipeline $(get_N primary-vertexing MATCH REST)" "$PVERTEX_EXTRA_CONFIG"
+
```
davidrohr: could you remove the unnecessary new line?

prodtests/full-system-test/dpl-workflow.sh
shahor02: Ah, this is my fault, I thought it is relevant for writing only. But then, for large PbPb datasets we will not be able to store clusters in the root file, since they exceed the 1GB limit.

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -366,22 +394,33 @@ if [ $CTFINPUT == 0 ]; then
...
 # Common reconstruction workflows
-(has_detector_reco TPC || has_detector_ctf TPC) && WORKFLOW+="o2-gpu-reco-workflow ${ARGS_ALL//-severity $SEVERITY/-severity $SEVERITY_TPC} --input-type=$GPU_INPUT $DISABLE_MC --output-type $GPU_OUTPUT --pipeline gpu-reconstruction:${N_TPCTRK:-1} $GPU_CONFIG $ARGS_EXTRA_PROCESS_o2_gpu_reco_workflow --configKeyValues \"$ARGS_ALL_CONFIG;GPU_global.deviceType=$GPUTYPE;GPU_proc.debugLevel=0;$GPU_CONFIG_KEY;$GPU_EXTRA_CONFIG;$CONFIG_EXTRA_PROCESS_o2_gpu_reco_workflow\" | "
+PIPELINEGPU=1
+if [ -z $DISABLEPIPELINE ]; then
+  PIPELINEGPU=${N_TPCTRK:-1}
+fi
+if [ ! -z $ISPILOTBEAMDATA ]; then
+  ITSTPC_EXTRA_CONFIG="tpcitsMatch.maxVDriftUncertainty=0.2;tpcitsMatch.safeMarginTimeCorrErr=10.;tpcitsMatch.cutMatchingChi2=1000;tpcitsMatch.crudeAbsDiffCut[0]=5;tpcitsMatch.crudeAbsDiffCut[1]=5;tpcitsMatch.crudeAbsDiffCut[2]=0.3;tpcitsMatch.crudeAbsDiffCut[3]=0.3;tpcitsMatch.crudeAbsDiffCut[4]=10;tpcitsMatch.crudeNSigma2Cut[0]=200;tpcitsMatch.crudeNSigma2Cut[1]=200;tpcitsMatch.crudeNSigma2Cut[2]=200;tpcitsMatch.crudeNSigma2Cut[3]=200;tpcitsMatch.crudeNSigma2Cut[4]=900"
```
davidrohr: Also here, I don't really like setting this based on $ISPOLOTBEAMDATA.
And by the way $ITSTPC_EXTRA_CONFIG is deprecated and will be removed.
I'd prefer to set `CONFIG_EXTRA_PROCESS_o2_tpcits_match_workflow="..."` from extern

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -366,22 +394,33 @@ if [ $CTFINPUT == 0 ]; then
...
+fi
+echo extra config = $ITSTPC_EXTRA_CONFIG, ispilot = $ISPILOTBEAMDATA
```
davidrohr: There should not be printouts during workflow generation (except for with `WORKFLOWMODE=print`. This will break the XML generation.
In case of error messages, they must go to stderr (`echo foo 1>&2`)

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -366,22 +394,33 @@ if [ $CTFINPUT == 0 ]; then
...
 has_detector_reco FT0 && add_W o2-ft0-reco-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N ft0-reconstructor FT0 REST)"
 has_detector_reco TRD && add_W o2-trd-tracklet-transformer "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC $TRD_TRANSFORMER_CONFIG --pipeline $(get_N TRDTRACKLETTRANSFORMER TRD REST TRDTRK)"
 has_detectors_reco TRD TPC ITS && [ ! -z "$TRD_SOURCES" ] && add_W o2-trd-global-tracking "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC $TRD_CONFIG --track-sources $TRD_SOURCES" "$TRD_CONFIG_KEY;$ITSMFT_FILES"
 has_detectors_reco TOF TRD TPC ITS && [ ! -z "$TOF_SOURCES" ] && add_W o2-tof-matcher-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC --track-sources $TOF_SOURCES --pipeline $(get_N tof-matcher TOF REST TOFMATCH)" "$ITSMFT_FILES"
+if [ -z "$DISABLE_ROOT_OUTPUT" ]; then
+  has_detectors TPC && add_W o2-tpc-reco-workflow "--input-type pass-through --output-type clusters,tracks,send-clusters-per-sector $DISABLE_MC"
```
davidrohr: could you use the more compact syntax?
```
has_detectors TPC && [ -z "$DISABLE_ROOT_OUTPUT" ] && add_W o2-tpc-reco-workflow ...
```

prodtests/full-system-test/dpl-workflow.sh
martenole: This change is also in https://github.com/AliceO2Group/AliceO2/pull/7479 which doesn't pass the CI unfortunately. Let's see what can be merged first :)

prodtests/full-system-test/run-workflow-on-inputlist.sh
```diff
@@ -17,10 +17,15 @@ if [[ $2 != "LOCAL" && ! -f $2 ]]; then
   exit 1
 fi
 
+if [[ $1 == "CTF" ]]; then
+  export EXTINPUT=0
+fi
+
```
davidrohr: I am a bit confused by this. `EXTINPUT` is only set to 1 in case `if [[ $1 == "DD" ]]; then` below.
So what is this needed for? It would only make sense in case the user would set EXTINPUT=1 on the command line when invoking this script. That would actually indeed break it, but then there are 10 other env variables that could break it if the user sets them. So I am not sure this single check here has much benefit.
What we should perhaps do is define a list of env variables that are set by this script ( e.g. `EXTINPUT`, `CTFINPUT`, `RAWTFINPUT`, `INPUT_FILE_LIST`) and that should not be set by the user, and then check if any of them are set at startup, and if so abort with an error.

prodtests/full-system-test/dpl-workflow.sh
martenole: In order to allow that `PVERTEX_CONFIG` can be set from the outside we need to change the way it is defined in the dpl-workflow.sh script, so we should at least change this to something like
```
if [ -z $PVERTEX_CONFIG ]; then
  PVERTEC_CONFIG=...
fi
```

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -202,7 +216,13 @@ for det in `echo $LIST_OF_DETECTORS | sed "s/,/ /g"`; do
     has_detector_reco $det && add_comma_separated TRACK_SOURCES "$det"
   fi
 done
-PVERTEX_CONFIG="--vertexing-sources $TRACK_SOURCES --vertex-track-matching-sources $TRACK_SOURCES"
+
+if [ -z $ISPILOTBEAMDATA ]; then
+  PVERTEX_CONFIG="--vertexing-sources $TRACK_SOURCES --vertex-track-matching-sources $TRACK_SOURCES"
+else
+  PVERTEX_CONFIG="--vertexing-sources ITS --vertex-track-matching-sources ITS --hbfutils-config none"
+  PVERTEX_EXTRA_CONFIG="pvertexer.nSigmaTimeCut=100;pvertexer.dbscanMaxDist2=30;pvertexer.dcaTolerance=3.;pvertexer.pullIniCut=100;pvertexer.addZSigma2=0.1;pvertexer.tukey=20.;pvertexer.addZSigma2Debris=0.01;pvertexer.addTimeSigma2Debris=1.;pvertexer.maxChi2Mean=30;"
```
davidrohr: Please use `CONFIG_EXTRA_PROCESS_o2_primary_vertexing_workflow="..."`, the other is deprecated and will be removed.

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -202,7 +216,13 @@ for det in `echo $LIST_OF_DETECTORS | sed "s/,/ /g"`; do
     has_detector_reco $det && add_comma_separated TRACK_SOURCES "$det"
   fi
 done
-PVERTEX_CONFIG="--vertexing-sources $TRACK_SOURCES --vertex-track-matching-sources $TRACK_SOURCES"
+
+if [ -z $ISPILOTBEAMDATA ]; then
+  PVERTEX_CONFIG="--vertexing-sources $TRACK_SOURCES --vertex-track-matching-sources $TRACK_SOURCES"
+else
+  PVERTEX_CONFIG="--vertexing-sources ITS --vertex-track-matching-sources ITS --hbfutils-config none"
```
davidrohr: I think it would be more flexible if we changed it in this way:
```
[[ -z "VERTEXING_SOURCES" ]] && VERTEXING_SOURCES="$TRACK_SOURCES"
PVERTEX_CONFIG="--vertexing-sources $VERTEXING_SOURCES --vertex-track-matching-sources $VERTEXING_SOURCES"
```
Then one could set externally `VERTEXING_SOURCES=ITS` for pilot beam data.

prodtests/full-system-test/run-workflow-on-inputlist.sh
martenole: We just had the problem that even when we ran with `CTF` mode then we got the error message `EXTINPUT and CTFINPUT are incompatible`, because when running on the EPNs EXTINPUT is set to 1 by default (if it is not set before) in setenv.sh

davidrohr (changes requested):
Hi @chiarazampolli @martenole : Thx for the PR and sorry for having so many comments, and sorry again for not being available this morning...

So except for some of the minor semantic and deprecation comments I put below, I have 2 general comments:
- I don't like to have special settings for PILOTBEAMDATA in several places in the workflow. I would prefer to set them externally. One idea that comes to my mind is: we could have a script with settings for a " period" that is sourced before processing by `run-workflow-on-inputlist.sh`. In there we could collect all the special settings applied for that period. That would leave the general script shorter and cleaner, and one would see all the modifications for that period in one place.
- I also don't like the DISABLEPIPELINE option. Actually, all pipelines should be 1 by default, and they will only be != 1 if some pipelined mode is requested, e.g.
  - `EPNPIPELINES!=0` which comes when `EPNMODE=1` is set.
  - `OPTIMIZED_PARALLEL_ASYNC != 0`
Now, I have seen that some pipelines are actually != 1 also by default. That is unintended, and I will fix that. I think then we don't need any additional DISABLEPIPELINES any more.

prodtests/full-system-test/run-workflow-on-inputlist.sh
davidrohr: OK, now I am confused. Did you perhaps set EPNMODE=1, which sets some defaults for the synchronous processing, like to use 4 GPUs per NUMA domain and it also enables the EPNPIPELINES?
That would explain why EXTINPUT was set for you, and that would also explain some of the !=1 pipeline settings you had. If that was the case, sorry for the confusion :(
Perhaps we should rename that in EPNSYNCMODE=1...

martenole: Actually yes, we naively set `EPNMODE=1`, because we were running on the EPN... We will change this then

prodtests/full-system-test/dpl-workflow.sh
davidrohr: Sorry, commented below before I saw @shahor02 comment here. I agree I would not hardcode stuff in the general script. I think this should come via env vars. To make it clean, I would propose to collect all env vars you set in an extra script. If you want to track which extra env vars you set in the logs (to avoid confusion when running something on the grid later), that extra script could have some logic to print the extra env vars it is setting. Then you would always have it logged.
@martenole : see my comment below about adding `VERTEXING_SOURCES`

prodtests/full-system-test/run-workflow-on-inputlist.sh
davidrohr: OK, I have renamed it in #7516 and updated the READMEs to make it more clear.
In order to use it, you will also need the new O2DataProcessing with new setenv, I have bumped it in alidist.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Yes, one could also just link the run-dependent files locally, and keep the rest in the common folder - this is how I do it actually, e.g. (file names are fake):
```
commonFolder/grp_505548.root
commonFolder/grp_505673.root
commonFolder/geometry.root
cd myFolder
ln -s commonFolder/grp_505673.root grp.root
```
then I set:
``` 
FILEWORKDIR=commonFolder
FILEWORKDIRRUN=myFolder
```

chiarazampolli:
Hello @shahor02 , @davidrohr , @martenole , 
Thanks for the comments, all very useful!!
Since we got many comments, I propose to not reply to them one by one (unless the answer does not require change), but just re-work the changes, trying to remove everything that is specific to the pilot beam data.
I will move this PR to WIP, unless you prefer that I close it, and reopen a new one with a different proposal (for now I move it to WIP).
Chiara

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: I would anyway always print everything. Experience showed that it is very useful. 
@martenole : yes, your comment is in line to what I wrote earlier to Ruben ("Would setting it only if the env variables are not already set be ok for you?"), but let's change approach.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Yes, we don't care about the value, provided it is there.
But ok, we will refactorize with an external script.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Sure, will do.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: This was a leftover not meant to stay, sorry.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Sure.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Done

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Left as it is.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Done.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Not done as discussed here.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Done.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Done.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Full line removed for syncmode = 0.

chiarazampolli: Sorry, I am confused, you wrote above to remove this line, because selections are not done in async mode, but here you say to use `fastMultConfig.cutMultClusLow=-1`...

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Done.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Done

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Done

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Done

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Done

prodtests/full-system-test/run-workflow-on-inputlist.sh
```diff
@@ -68,7 +73,8 @@ start_process $MYDIR/dpl-workflow.sh
 
 if [[ "0$4" != "00" ]]; then
   sleep 1
-  tail -f ${LOG_PREFIX}*.log &
+  tail -n 1000000 -f ${LOG_PREFIX}*.log &
+  ln -sf ${LOG_PREFIX}*.log latest.log
```
davidrohr: The `ln` with `*` will fail in DD mode, since in that case there are 2 log files

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -378,18 +387,19 @@ fi
...
 has_detector MID && has_processing_step MID_RECO && add_W o2-mid-reco-workflow "$DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N MIDClusterizer MID REST),$(get_N MIDTracker MID REST)"
 has_detector MCH && has_processing_step MCH_RECO && add_W o2-mch-reco-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N mch-track-finder MCH REST MCHTRK),$(get_N mch-cluster-finder MCH REST),$(get_N mch-cluster-transformer MCH REST)"
 has_detector MFT && has_processing_step MFT_RECO && add_W o2-mft-reco-workflow "--clusters-from-upstream $DISABLE_MC $DISABLE_ROOT_OUTPUT --pipeline $(get_N mft-tracker MFT REST MFTTRK)" "$ITSMFT_FILES"
-has_detector FDD && has_processing_step FDD_RECO && add_W o2-fdd-reco-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC"
-has_detector FV0 && has_processing_step FV0_RECO && add_W o2-fv0-reco-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC"
+has_detector FDD && has_processing_step FDD_RECO && add_W o2-fdd-reco-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC --hbfutils-config none"
```
davidrohr: I assume this means the workflow doesn't read the HBFUtils settings from the digitization ini file.
@shahor02 : Could you remind me what we actually still read from that ini file? the NHBFPERTF is now in the GRP.
Do we need to read something from the ini file for MC data?

prodtests/full-system-test/run-workflow-on-inputlist.sh
chiarazampolli: I see. I would like to link to a file that always points to the latest log.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Actually they try to access that file, but since I don't have it, I used this option. It is not meant to be the final way, though.

davidrohr: You can just run a short simulation, or a `NEvents=5 NEventsQED=100 $O2_ROOT/prodtests/full_system_test.sh` to create missing files.

prodtests/full-system-test/run-workflow-on-inputlist.sh
davidrohr: Could you try `ln -sf ${LOG_PREFIX}${NUM_PROCS}.log latest.log` ?

prodtests/full-system-test/dpl-workflow.sh
shahor02: The ini file from the digitization is needed for the digit readers to correctly set up HBFutils as it was used for MC and to disable this `--hbfutils-config none` should be used. 
If you rebase your branch to @davidrohr's fix for FDD, there will be no need for this option anymore.

prodtests/full-system-test/run-workflow-on-inputlist.sh
chiarazampolli: Done and working

chiarazampolli:
Hi, 
Could this be merged if there are no further comments?
Chiara

prodtests/full-system-test/run-workflow-on-inputlist.sh
```diff
@@ -17,10 +17,15 @@ if [[ $2 != "LOCAL" && ! -f $2 ]]; then
   exit 1
 fi
 
+if [[ $1 == "CTF" ]]; then
```
davidrohr: I don't understand this setting. EXTINPUT should not be set from the outside anyway. So this should not be needed.

prodtests/full-system-test/run-workflow-on-inputlist.sh
```diff
@@ -64,11 +69,19 @@ else
...
 fi
-start_process $MYDIR/dpl-workflow.sh
+
+if [ -f dpl-workflow.sh ]; then
+    echo "Use dpl-workflow.sh macro passed as input"
+else
+    echo "Use dpl-workflow.sh macro from ${MYDIR}"
+    cp $O2_ROOT/prodtests/full-system-test/dpl-workflow.sh .
+fi
+start_process ./dpl-workflow.sh
```
davidrohr: As discussed privately, I would prefer to have an explicit setting via an env variable to set a directory for the dpl-workflow.sh script, instead of taking one from the local directory.

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -448,7 +457,15 @@ workflow_has_parameter AOD && add_W o2-aod-producer-workflow "--info-sources $AO
 
 # ---------------------------------------------------------------------------------------------------------------------
 # Quality Control
-workflow_has_parameter QC && { source $MYDIR/qc-workflow.sh; [[ $? != 0 ]] && exit 1; }
+if [[ ! -z $QC_JSON_FROM_OUTSIDE ]]; then
+    if [[ -z $QC_CONFIG_PARAM ]]; then
+        QC_CONFIG_PARAM="--local --host $QC_HOST"
+    fi
+    echo "qc json from outside $QC_JSON_FROM_OUTSIDE"
```
shahor02: There should be no echo from this script

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -448,7 +457,15 @@ workflow_has_parameter AOD && add_W o2-aod-producer-workflow "--info-sources $AO
...
+    workflow_has_parameter QC && add_W o2-qc "--config json://$QC_JSON_FROM_OUTSIDE $QC_CONFIG_PARAM" "" 0
+else
+    workflow_has_parameter QC && { source $O2DATAPROCESSING_ROOT/production/qc-workflow.sh; [[ $? != 0 ]] && exit 1; }
+fi
```
davidrohr: There should be no echo, or it should go to stderr.
If you want to provide multiple parameters, why don't you just provide a full string providing multiple parameters to `o2-qc`? That would be simpler and single-line.

davidrohr (commented):
In principle ok, but I have few comments:

prodtests/full-system-test/run-workflow-on-inputlist.sh
davidrohr: Actually, Now I remember I think we discussed this with Ole before: At the beginning of this script, we should actually check for multiple variables, which are set by this script, that they are not set from the outside. I.e.
```
for i in EXTINPUT CTFINPUT RAWTFINPUT; do
  [[ ! -z ${!i} ]] && { echo "$i must not be set!"; exit 1; }
done
```

prodtests/full-system-test/dpl-workflow.sh
davidrohr: Thinking a bit about it, I think it makes actually sense to have multiple variables. But then I would align it between the JSON_FROM_OUTSIDE case here, and what is done in qc-workflow.sh.
How about we use
```
add_W o2-qc "--config json://$QC_JSON_FROM_OUTSIDE ${QC_CONFIG_PARAM:---local --host ${QC_HOST:-localhost}}" "" 0
```
here (also a bit shorter :). Sorry, but I try to keep the script short), and then
```
"--config json://$MERGED_JSON_FILENAME ${QC_CONFIG_PARAM:---local --host ${QC_HOST:-localhost}}"
```
in O2DataProcessing/production/qc-workflow.sh

davidrohr: Actually, I just added the suggested simplification and also the QC_JSON_FROM_OUTSIDE handling to qc-workflow.sh:
AliceO2Group/O2DataProcessing#125. If that makes sense for you, I'd merge it today and bump it in alidist, so it will go in tomorrow's tag.

davidrohr: OK, I did some more fixes in the O2DataProcessing part, so I just went ahead and merged the PR. QC_JSON_FROM_OUTSIDE is now supported there in any case.

prodtests/full-system-test/run-workflow-on-inputlist.sh
davidrohr: OK, after thinking about this again, this is actually dangerous. You are creating a local copy of the script, and afterwards you will be using the local copy if present. I.e., if you run once, then you update O2, and then you run again in the same folder, it will use the old version of dpl-workflow.sh. I think we should not do this. This will lead to confusion.

#7606 is a proposal how we could add this feature, does this work for you? Btw, `start_process` will anyway print `Started $@ with PID $PID` with the full path, so I didn't add an additional echo.

shahor02: Btw, concerning multiple running in the same folder: the AOD production is requested in update mode and fails if AO2D.root is found in the folder. Either update should be suppressed or the file should removed before starting the job.

davidrohr:
@chiarazampolli : Testing your PR, I saw also that the vertexing_sources are not applied everywhere. I have opened chiarazampolli/AliceO2#47 which would apply all the changes I have proposed to your PR + fixes the vertexing.

prodtests/full-system-test/run-workflow-on-inputlist.sh
martenole: I think this is a leftover from when we were still using the EPN mode, so it can be safely removed. The check at the beginning would be helpful

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Thanks David, perfect! I merged also your PR to my branch here, so now it should be fine, I guess.

chiarazampolli:
> @chiarazampolli : Testing your PR, I saw also that the vertexing_sources are not applied everywhere. I have opened [chiarazampolli#47](https://github.com/chiarazampolli/AliceO2/pull/47) which would apply all the changes I have proposed to your PR + fixes the vertexing.

Yes, sorry, again victims of the rebase...

PR #7512, opened 2021-11-03
MCUtils: functionality to query ALICE properties on MCtracks/particles

sawenzel:


sawenzel:
@chiarazampolli : Could you see if this is ok for the use case you have in mind? In particular the interface?

chiarazampolli:
Hi, 
If I am not wrong, you just moved the class from O2Physics to O2, right?
Chiara

sawenzel:
@chiarazampolli : No, the code is similar but not a copy. In O2Physics, it operates and navigates directly within AOD tables and the interface is different. It's not straightforward to make a real common code. I am buying into a slight code duplication if it unblocks you.

DataFormats/simulation/src/MCUtils.cxx
```diff
@@ -0,0 +1,165 @@
...
+//
+// Created by Sandro Wenzel on 03.11.21.
+//
+
+#include <SimulationDataFormat/MCUtils.h>
+#include "TPDGCode.h"
+
+using namespace o2::mcutils;
+
+bool isStable(int pdg)
```
jgrosseo: We could expose this function in the header and reuse it in O2Physics "at least". What do you think?

jgrosseo:
@amorsch You mentioned that some of the "hacks" here are not needed anymore for Run 3 if I am not mistaken. Which are those?

DataFormats/simulation/src/MCUtils.cxx
sawenzel: Yes, absolutely.

chiarazampolli:
Hello @sawenzel . 
To me it looks like you kept the same functionalities, so it seems fine. 
Thanks!
Chiara

amorsch:
> @amorsch You mentioned that some of the "hacks" here are not needed anymore for Run 3 if I am not mistaken. Which are those?
@jgrosseo 
In principle a particle from the generator which is in the list of stable particles is a physical primary. 
Since we decay HF during generation this should be also ok for HF decay particles. 
Hence the special cases for pi0, K0, Xi and HF can go out. A problem we have with Pythia8 is that sometimes direct gammas are put several times on the stack. So one checks that the gamma has no child among the generated particles. This avoids using the status code. So this part has to stay. 
What about calling IsPhysicalPrimary only once and storing the result in a flag?

sawenzel:
I put `isStable` in the header.

Storing isPhysicalPrimary as a status bit in the AOD sounds appealing. It has the advantage of less code-duplication and less compute during analysis (assuming that the logic is more or less fixed and constant). 

@amorsch : I left the code as is. May I ask you to make a future PR according to your understanding? (But I suppose in any case that the current version works ok as a start).

jgrosseo:
> (assuming that the logic is more or less fixed and constant).

@amorsch Are we sure about this for Run 3 (it seemed to not have been the case for Run 2)?

PR #7513, opened 2021-11-03
DPL: introduce CallbacksPolicy

ktf:
Useful to customise which callbacks are set up by default depending on the device type.


ktf:
@TimoWilken looks like something in wrong with build/O2/o2 and build/O2/o2-dataflow. Could you have a look?

TimoWilken:
The failures were due to errors cloning git repos. I restarted the slc7 builders to add one for the newly available machine, so they all had to fetch stuff in `sw/MIRROR` again. I've restarted the checks.

ktf:
macos still has troubles. @TimoWilken @MichaelLettrich

MichaelLettrich:
@ktf restarted the builder. should be back now.

PR #7514, opened 2021-11-03
limit ITS fast mult.estimate from below by 0

shahor02:
trivial, merging


shahor02:
trivial, failures unrelated, merging

PR #7515, opened 2021-11-03
DPL Analysis: fix incorrect use of offset for filtered iteratorAt()

aalkin:


ktf:
Didn't we already fix something like this? Would it be possible to have a test workflow which exercises this in the test workflows we have in hyperloop?

aalkin:
Previous fix was for unfiltered iterator, this one is for filtered. As far as I know, it is not used in any workflows, but I will create a test example for this and the other indexing features.

ktf:
The issue in the o2-cs8 seems to come from O2Physics. Could you have a look?

ktf (commented):
Please check the error is not related to this.

aalkin:
@ktf this is probably due to https://github.com/AliceO2Group/AliceO2/pull/7507 that removed tracks covariance from the default join

ktf:
Yes, you are right. o2-cs8 does not use the O2Physics master. Should be ok now that I bumped alidist.

PR #7520, opened 2021-11-04
Mat.LUT object must be saved as ccdb_object to be reachable from CCDB

shahor02:


Detectors/Base/src/MatLayerCylSet.cxx
```diff
@@ -157,37 +157,32 @@ void MatLayerCylSet::dumpToTree(const std::string outName) const
...
 
   TFile outf(outFName.data(), "recreate");
   if (outf.IsZombie()) {
     return;
   }
-  if (name.empty()) {
-    name = "matBud";
-  }
-  outf.WriteObjectAny(this, Class(), name.data());
+  outf.WriteObjectAny(this, Class(), "ccdb_object");
```
sawenzel, line 168: I was under the impression that this would be automatically done if you uploaded this using CcdbApi (or the command-line tools)? (I guess you use alternative paths to upload?)

sawenzel (approved):

Detectors/Base/src/MatLayerCylSet.cxx
shahor02: Yes, but I use `curl -L -F "blob=@file.root" http://alice-ccdb.cern.ch/<URL>/<from>/<to> --cert ~/.globus/usercert.pem --key ~/.globus/userkey.pem -k` which preserves the name of the object in the file.

PR #7111, opened 2021-09-24
[WIP][EMCAL-687] AODProducerWorkflowSpec: implemented calo cells

fjonasALICE:
- implemented treatment of calo cells and trigger records for CTF -> AO2D
- since CTF data from all detectors is needed, code has not been tested yet
**open questions:**
- used BCId stored in table needs to be double checked. Unclear if absolute or relative BCId is needed
- trigger record for now filled with dummy values until implementation of offline triggers is done
- MC information is currently ignored until proper implementation on previous stages of reco chain

PR #7521, opened 2021-11-04
[EMCAL-687] AODProducerWorkflowSpec: implemented CALO cells

fjonasALICE:
sorry for reopening. I changed branches since i had opened the original pull request: https://github.com/AliceO2Group/AliceO2/pull/7111
and somehow my plan of forcing the changes from my new branch to the existing PR did not fully work and it automatically closed the original request instead ... sorry for the noise.

The changes suggested in the original PR are implemented and conflicts with Rubens implementation fixed.

### DIFF
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -207,7 +207,6 @@ class AODProducerWorkflowDPL : public Task
   GID::mask_t mInputSources;
   int64_t mTFNumber{-1};
   int mRunNumber{-1};
-  int mFillCaloCells{1};
   int mTruncate{1};
   int mRecoOnly{0};
   TString mResFile{"AO2D"};
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -24,6 +24,7 @@
 #include "DataFormatsMCH/TrackMCH.h"
 #include "DataFormatsTPC/TrackTPC.h"
 #include "DataFormatsTRD/TrackTRD.h"
+#include "DataFormatsEMCAL/EventHandler.h"
 #include "Framework/AnalysisDataModel.h"
 #include "Framework/AnalysisHelpers.h"
 #include "Framework/DataProcessorSpec.h"
@@ -206,6 +207,7 @@ class AODProducerWorkflowDPL : public Task
   GID::mask_t mInputSources;
   int64_t mTFNumber{-1};
   int mRunNumber{-1};
+  int mFillCaloCells{1};
   int mTruncate{1};
   int mRecoOnly{0};
   TString mResFile{"AO2D"};
@@ -300,6 +302,7 @@ class AODProducerWorkflowDPL : public Task
                   gsl::span<const o2::ft0::RecPoints>& ft0RecPoints,
                   gsl::span<const o2::fv0::RecPoints>& fv0RecPoints,
                   gsl::span<const o2::dataformats::PrimaryVertex>& primVertices,
+                  gsl::span<const o2::emcal::TriggerRecord>& caloEMCCellsTRGR,
                   const std::vector<o2::InteractionTimeRecord>& mcRecords,
                   std::map<uint64_t, int>& bcsMap);
 
@@ -337,6 +340,7 @@ class AODProducerWorkflowDPL : public Task
                                    const dataformats::PrimaryVertex& vertex);
 
   template <typename MCParticlesCursorType>
+
   void fillMCParticlesTable(o2::steer::MCKinematicsReader& mcReader,
                             const MCParticlesCursorType& mcParticlesCursor,
                             gsl::span<const o2::dataformats::VtxTrackRef>& primVer2TRefs,
@@ -361,6 +365,12 @@ class AODProducerWorkflowDPL : public Task
 
   // helper for trd pattern
   uint8_t getTRDPattern(const o2::trd::TrackTRD& track);
+
+  o2::emcal::EventHandler<o2::emcal::Cell>* mCaloEventHandler = nullptr; ///< Pointer to the event builder for emcal cells
+
+  template <typename TCaloCells, typename TCaloTriggerRecord, typename TCaloCursor, typename TCaloTRGTableCursor>
+  void fillCaloTable(const TCaloCells& calocells, const TCaloTriggerRecord& caloCellTRGR, const TCaloCursor& caloCellCursor,
+                     const TCaloTRGTableCursor& caloCellTRGTableCursor, std::map<uint64_t, int>& bcsMap);
 };
 
 /// create a processor spec
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -987,7 +987,6 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
   // Needed by MCH track extrapolation
   o2::base::GeometryManager::loadGeometry();
 
-
   // writing metadata if it's not yet in AOD file
   // note: `--aod-writer-resmode "UPDATE"` has to be used,
   //       so that metadata is not overwritten
@@ -1015,7 +1014,6 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
     LOGF(fatal, "Metadata: target file %s already has metadata", mResFile);
   }
   fResFile->Close();
-
 
   mTimer.Reset();
 }
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -930,7 +930,6 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
   mAnchorProd = ic.options().get<string>("anchor-prod");
   mRecoPass = ic.options().get<string>("reco-pass");
   mTFNumber = ic.options().get<int64_t>("aod-timeframe-id");
-  mFillCaloCells = ic.options().get<int>("fill-calo-cells");
   mRecoOnly = ic.options().get<int>("reco-mctracks-only");
   mTruncate = ic.options().get<int>("enable-truncation");
   mRunNumber = ic.options().get<int>("run-number");
@@ -942,7 +941,6 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
     LOG(INFO) << "The Run number will be obtained from DPL headers";
   }
 
-  LOG(INFO) << "calo filling flag is set to: " << mFillCaloCells;
 
   // create EventHandler used for calo cells
   mCaloEventHandler = new o2::emcal::EventHandler<o2::emcal::Cell>();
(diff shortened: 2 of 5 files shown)


fjonasALICE:
sorry for the reopen, see decription.  @nburmaso: I implemented your changes. Tagging also @sawenzel so he gets an update about the status of the PR

mfasDa (dismissed):

DataFormats/common/include/CommonDataFormat/InteractionRecord.h
```diff
@@ -117,6 +117,13 @@ struct InteractionRecord {
     return (int64_t(orbit) * o2::constants::lhc::LHCMaxBunches) + bc;
   }
 
+  // get global BC ID from bc and orbit
+  // this ID is needed to calculate BC rel. to BCID of vertex
+  uint64_t getGlobalBC() const
+  {
+    return bc + orbit * o2::constants::lhc::LHCMaxBunches;
+  }
```
sawenzel: this function seems introduced but not used in the code anywhere. Do you need it?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -854,17 +860,85 @@ uint8_t AODProducerWorkflowDPL::getTRDPattern(const o2::trd::TrackTRD& track)
...
+
+    // check with Markus if globalBC ID is needed or globalBC - minGlBC
+    // in case of collision vertex what is used is
+    // uint64_t globalBC = std::round(tsTimeStamp / o2::constants::lhc::LHCBunchSpacingNS);
+    auto item = bcsMap.find(globalBC);
+    int bcID = -1;
+    if (item != bcsMap.end()) {
+      bcID = item->second;
+    } else {
+      throw CollisionIDNotFoundException(globalBC, bcsMap.size());
```
sawenzel: as far as I know, we don't use exceptions in that code. It would be better to not divert here and simply use LOG(FATAL) or LOG(ERROR) ... which is much easier to treat.

DataFormats/common/include/CommonDataFormat/InteractionRecord.h
sawenzel: note that there is helper function ```uint64_t relativeTime_to_GlobalBC(double relativeTimeStampInNS)
``` in the AOD converter itself which may be interesting!?

File: Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h [hunk 3]
```diff
@@ -351,6 +355,48 @@ class AODProducerWorkflowDPL : public Task
 
   // helper for trd pattern
   uint8_t getTRDPattern(const o2::trd::TrackTRD& track);
+
+  o2::emcal::EventHandler<o2::emcal::Cell>* mEventHandler = nullptr; ///< Pointer to the event builder for emcal cells
```
sawenzel: could you call this `mCaloEventHandler` or similar? Just `mEventHandler` looks a bit too generic.

DataFormats/common/include/CommonDataFormat/InteractionRecord.h
fjonasALICE: the function was not needed  and an artifact from previous implementations. I removed it

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
fjonasALICE: changed to your suggestion, following usage from other detectors

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
fjonasALICE: renamed to mCaloEventHandler

DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
```diff
@@ -221,6 +223,8 @@ struct DataRequest {
   void requestPrimaryVerterticesTMP(bool mc);
   void requestSecondaryVertertices(bool mc);
 
+  void requestCaloCells(bool mc);
```
shahor02: EMCAL cells and triggers were already implemented, see https://github.com/AliceO2Group/AliceO2/blob/7274c228d3335892a51987bb4dc420fa2e9bdfaa/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h#L217
https://github.com/AliceO2Group/AliceO2/blob/7274c228d3335892a51987bb4dc420fa2e9bdfaa/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h#L351
https://github.com/AliceO2Group/AliceO2/blob/7274c228d3335892a51987bb4dc420fa2e9bdfaa/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h#L603-L604

shahor02 (changes requested):

DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
fjonasALICE: I am aware. This change should not have been pushed .. working on it. Thanks for spotting

fjonasALICE: I removed the changes. Sorry again for the mess ... Note to self: never change branches when you still have an open pull request and then force push from the wrong branch. Now everything should be fine. Please let me know if you still spot something

shahor02 (dismissed):

mfasDa (dismissed):

sawenzel (dismissed):

mfasDa:
@shahor02 Seems the name of the trigger record container is not correct in the reco container:
```
[ERROR] invalid workflow in o2-aod-producer-workflow: No matching output found for EMC/CELLTRIGREC/0 as requested by data processor "aod-producer-workflow". Candidates:
```
Needs to be 
```
EMC/CELLSTRGR/0
```

fjonasALICE:
@shahor02 the full CI test fails. This is due to a wrong name in the RecoContainer in the function requestEMCALCells implemented a month ago. CELLTRIGREC should be CELLSTRGR How should I proceed? I guess it never failed before because requestEMCALCells was never called. Do you want me to change this or do we merge this request and then you fix it?

Edit: markus was quicker

shahor02:
Hi, will fix, just need to finish one job to switch the branch.

shahor02:
Hi, fixed in (already merged) https://github.com/AliceO2Group/AliceO2/pull/7561, thanks for spotting

fjonasALICE:
rebased to have @shahor02 changes included so that test won't fail

sawenzel:
@ktf @TimoWilken : Is there any reason this is not compiling for the last 16h?

sawenzel (changes requested):
I tested this branch in a O2DPG MC workflow. However, it fails with the message 
```[ERROR] invalid workflow in o2-aod-producer-workflow: No matching output found for EMC/CELLS/0 as requested by data processor "aod-producer-workflow"``` regardless if I set `--fill-calo-cells` or not.

So, I assume this needs a bit more testing. So, as a first step, could you please add CALO treatment to `https://github.com/AliceO2Group/AliceO2/blob/dev/prodtests/sim_challenge.sh` as a first step. Once this works, we can go on with O2DPG integration.

As a side note, please rethink your command line flag `--fill-calo-cells` as this breaks the symmetry with other detectors. The better way would be to treat it as any other source "ITS,TPC,...".

See also `https://github.com/AliceO2Group/O2DPG/blob/master/MC/bin/o2dpg_sim_workflow.py` around line 751.

File: Detectors/AOD/src/AODProducerWorkflowSpec.cxx [hunk 5]
```diff
@@ -1377,6 +1468,7 @@ DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src, bool useMC, std::s
   if (src[GID::TPC]) {
     dataRequest->requestClusters(GIndex::getSourcesMask("TPC"), false); // TOF clusters are requested with TOF tracks
   }
+  dataRequest->requestEMCALCells(useMC);
```
mfasDa: Looks the request is globally enabled, meaning even if the option to fill EMCAL cells is not enabled the request is performed. I guess this leads to the observation that the fill-calo-cell option has no effect on it. This needs to be handled in a different way, probably as well via `GIndex::getSourcesMask`. @shahor02 Can we add a key for EMCAL in the GIndex?

mfasDa (changes requested):
Enabling / disabling EMCAL must be handled in a different way, because the current way enables the data request regardless whether EMCAL is used or not.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: @mfasDa what do you mean? All detectors are there: https://github.com/AliceO2Group/AliceO2/blob/e54e15af180e30c233fe85a278cd077f42bed5e9/DataFormats/Reconstruction/include/ReconstructionDataFormats/GlobalTrackID.h#L47

sawenzel: this also looks wrong (depending on flag `useMC`?)

mfasDa: @shahor02 I see, then the line simply changes to

``` c++
  if (src[GID::EMC]) {
    dataRequest->requestEMCALCells(useMC);
  }
```

right? + of course the same for filling the calo table ...

fjonasALICE: ok I will implement the changes and remove --fill-calo-cells. @shahor02 you are right that all detectors are there, but i think I also need to add it to GID::mask_t allowedSrc in aod-producer-workflow.cxx

shahor02: @fjonasALICE sure, can you add this in this PR?

fjonasALICE: sure, i will push the changes once my local build finishes

fjonasALICE: @shahor02 all changes shoud be included now

sawenzel:
@fjonasALICE : Can you still please add the CALO treatment in the sim_challenge.sh (as mentioned in my comment above)?

fjonasALICE:
> @fjonasALICE : Can you still please add the CALO treatment in the sim_challenge.sh (as mentioned in my comment above)?

@sawenzel : Could you clarify what changes are still needed? I looked at sim_challenge.sh and the o2-emcal-reco-workflow is already included. In the o2-aod-producer-workflow i added EMC as an allowed source and removed the treatment via a flag, as you suggested. Could you point me to what I am missing?

sawenzel:
@fjonasALICE : Then it is fine for me, if you have tested that sim_challenge.sh runs through and the AOD contains CALO output!?

sawenzel:
@fjonasALICE : Can you please still resolve the merge conflict? Not sure where it comes from now.

shahor02:
@fjonasALICE @sawenzel it was conflicting with my yesterday switch to disable secondary vertices processing, I've resolved it.

fjonasALICE:
@shahor02 i think we simultaneously fixed it now ... thanks!
@sawenzel I will perform a full test then and report back

fjonasALICE:
@sawenzel sim_challenge.sh failed. There was a problem in the calo EventHandler. Should be fixed by [7584](https://github.com/AliceO2Group/AliceO2/pull/7584) from @mfasDa

sawenzel (dismissed):

sawenzel:
The code compiles and runs ok. Let's get this code in to avoid further merge conflicts with other PRs and get over the "first committer" status which keeps blocking the CI.

However, I think that this needs more testing and further adjustments. A simple run of `sim_challenge.sh` goes through, produces the AOD trees ... but, as far as I can see, they are empty.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1312,6 +1394,15 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
   }
+  for (int src = GIndex::NSources; src--;) {
+    if (GIndex::includesSource(src, mInputSources)) {
+      if (src == GIndex::Source::EMC) {
+        // fill EMC cells to tables
+        // TODO handle MC info
+        fillCaloTable(caloEMCCells, caloEMCCellsTRGR, caloCellsCursor, caloCellsTRGTableCursor, bcsMap);
+      }
+    }
+  }
```
shahor02: there is no reason for such loop, fixing to 
```
if (mInputSources[GIndex::EMC]) {
  fillCaloTable(caloEMCCells, caloEMCCellsTRGR, caloCellsCursor, caloCellsTRGTableCursor, bcsMap);
}
```

sawenzel (approved):

mfasDa (approved):

PR #7524, opened 2021-11-04
Move mem-factor option from MCH decoder to encoder

shahor02:


shahor02:
by mistake the option was added to the entropy decoder instead of the encoder, trivial fix, merging

PR #7526, opened 2021-11-04
AODProducer fixes: runNumber and ITSAB tracklets

shahor02:
@jgrosseo @nburmaso I've added fetching the runNumber from the DPL, as for the TF number: I assume you refer on the timestamp in `ms` calculated in https://github.com/shahor02/AliceO2/blob/e74a284012f34235db3b12e9242917612f4f5405/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L132 ?

The way it is done now will hardly work, it is very difficult to have `ms`-precise start of run, moreover, to record its orbit.

The exact time-stamp should be extracted from the http://ccdb-test.cern.ch:8080/browse/CTP/Calib/OrbitReset?report=true and DataHeader::firstTForbit, but this should be done using DPL methods to find the rough time (to query this CTP object). 
Is the timestamp necessary now?

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -299,12 +299,10 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
         } else {
           auto contributorsGID = data.getSingleDetectorRefs(trackIndex);
           const auto& trackPar = data.getTrackParam(trackIndex);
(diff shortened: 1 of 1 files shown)


jgrosseo:
The TF number is used as output folder in the AOD file. It simply has to be unique. It does not matter if it is not precise (or a bit off), but simply be unique.

shahor02:
OK, I'added assignment of the `tfNumber` from the `dh->tfCounter` which is a unique counter *within the run*.

jgrosseo:
This is good for now. We have to extend it in the future to make an absolute unique number, but this can wait a bit...

jgrosseo:
I have added a comment on this

jgrosseo (approved):

PR #7527, opened 2021-11-04
ITS/MFT: pass origin/datadescription asked in CL to RawPixelDecoder

shahor02:


shahor02:
fullCI did not start, tested locally, merging.

PR #7533, opened 2021-11-05
DPL: using linear message store for the DataRelayer cache

matthiasrichter:
Part of #7110

### DIFF
--- a/Framework/Core/include/Framework/MessageSet.h
+++ b/Framework/Core/include/Framework/MessageSet.h
@@ -23,22 +23,28 @@ namespace framework
 
 /// A set of associated inflight messages.
 struct MessageSet {
-  std::vector<PartRef> parts;
+  struct Index {
+    Index(size_t p, size_t s) : position(p), size(s) {}
+    size_t position = 0;
+    size_t size = 0;
+  };
+  std::vector<FairMQMessagePtr> messages;
+  std::vector<Index> index;
 
   MessageSet()
-    : parts()
+    : messages(), index()
   {
   }
 
   template <typename F>
-  MessageSet(F&& getter, size_t size)
-    : parts()
+  MessageSet(F getter, size_t size)
+    : messages(), index()
   {
     add(std::forward<F>(getter), size);
   }
 
   MessageSet(MessageSet&& other)
-    : parts(std::move(other.parts))
+    : messages(std::move(other.messages)), index(std::move(other.index))
   {
     other.clear();
   }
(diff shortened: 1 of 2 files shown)


Framework/Core/include/Framework/MessageSet.h
```diff
@@ -23,22 +23,28 @@ namespace framework
...
 /// A set of associated inflight messages.
 struct MessageSet {
-  std::vector<PartRef> parts;
+  struct Index {
+    Index(size_t p, size_t s) : position(p), size(s) {}
+    size_t position = 0;
+    size_t size = 0;
+  };
+  std::vector<FairMQMessagePtr> messages;
+  std::vector<Index> index;
```
ktf, line 32: Can you remind why we need the index at all?

matthiasrichter: this will contain the length of the message train, so we have the position in the linear vector and the number of messages for a particular part (header plus n payloads)

matthiasrichter:
@davidrohr is this something which happens from time to time?
```
## sw/BUILD/O2-latest/log
--
 98/379 Test  #25: GPU/TPCFastTransformation/macro/fastTransformQA.C_compiled ...........................***Failed   16.25 sec
--
104/379 Test  #24: GPU/TPCFastTransformation/macro/SplineDemo.C_compiled ................................***Failed   16.31 sec
--
134/379 Test  #26: GPU/TPCFastTransformation/macro/generateTPCCorrectionNTuple.C_compiled ...............***Failed   28.92 sec
```

davidrohr:
I have seen that before, but I don't know what to make out of it. Apparently it fails here:

```
Info in <TUnixSystem::ACLiC>: creating shared library /sw/BUILD/6e016cf8f69d5775714d089afcba9e67a9975265/O2/compiled_macros//sw/SOURCES/O2/7533-slc8_x86-64/0/GPU/TPCFastTransformation/macro/SplineDemo_C.so
/sw/slc8_x86-64/GCC-Toolchain/v10.2.0-alice2-4/bin/../lib/gcc/x86_64-unknown-linux-gnu/10.2.0/../../../../x86_64-unknown-linux-gnu/bin/ld: cannot open output file /sw/BUILD/6e016cf8f69d5775714d089afcba9e67a9975265/O2/compiled_macros//sw/SOURCES/O2/7533-slc8_x86-64/0/GPU/TPCFastTransformation/macro/SplineDemo_C.so: No such file or directory
collect2: error: ld returned 1 exit status
```

that seems unrelated to the GPU code. ROOT cannot open the output file when creating the lib for the compiled macro. Now idea why that happens, and why it happens only for the GPU macros...

PR #7534, opened 2021-11-05
DPL: allow CallbacksPolicy to have access to command line options

ktf:


ktf:
@shahor02 this still requires some adjustments to the tests, but it should be all we need.

ktf:
@shahor02 @davidrohr this is now compiling also the tests. Testing on hyperloop to make sure it does not break stuff.

Detectors/ITSMFT/ITS/workflow/src/its-reco-workflow.cxx
```diff
@@ -14,15 +14,33 @@
...
 
 using namespace o2::framework;
 
 // ------------------------------------------------------------------
 
+void customize(std::vector<o2::framework::CallbacksPolicy>& policies)
+{
+  policies.push_back(CallbacksPolicy{
+    [](DeviceSpec const& spec, ConfigContext const& context) -> bool {
+      auto filename = context.options().get<std::string>("filename");
```
ktf:
```suggestion
      // auto filename = context.options().get<std::string>("filename");
```

Detectors/ITSMFT/ITS/workflow/src/its-reco-workflow.cxx
```diff
@@ -14,15 +14,33 @@
...
 // ------------------------------------------------------------------
 
+void customize(std::vector<o2::framework::CallbacksPolicy>& policies)
+{
+  policies.push_back(CallbacksPolicy{
+    [](DeviceSpec const& spec, ConfigContext const& context) -> bool {
+      // auto filename = context.options().get<std::string>("filename");
+      return false;
+    },
+    [](CallbackService& service, InitContext& context) {
```
shahor02, line 37: Hi @ktf 
seems to work as expected except 1 problem: the filename for the tf-info vector comes via workflow global options, i.e. `ConfigContext`, as you correctly have it in the 1st lambda. Can the second lambda also have `ConfigContext` supplied (`InitContext` does not hurt but useless in this particular case).

ktf: I thought InitContext actually had also the workflow option. I will try to fix that.

shahor02: let me check, did not know it is supposed to pass the global options.

shahor02: @ktf I've checked, global options are not accessible via `InitContext.options()`

ktf: It's supposed to pass all the options which apply to a given DataProcessorSpec. I do not remember if workflow options are considered "global".

shahor02: well, in the `o2-global-track-cluster-reader --track-types ITS --cluster-types ITS --hbfutils-config o2_tfidinfo.root `
the particular reader devices don't see the option `--hbfutils-config`.

shahor02: @ktf Looks like there is more serious problem: the `customize(std::vector<o2::framework::CallbacksPolicy>& policies)` defined for particular workflow, checks *all* devices of the global merged workflow. So, to make it work currently I need to pass the option (filename with TF header infos) to all these workflows. I.e. since both `o2-global-track-cluster-reader` and `o2-its-reco-workflow` have this `customize`, I need to call the workflow as 
```
o2-global-track-cluster-reader --track-types ITS --cluster-types ITS --hbfutils-config o2_tfidinfo.root ... | \
o2-its-reco-workflow --clusters-from-upstream  --hbfutils-config o2_tfidinfo.root ...
```
otherwise the `o2-its-reco-workflow` checks the devices of the `o2-global-track-cluster-reader` against its own `hbfutils-config` option (by default pointing to `o2simdigitizerworkflow_configuration.ini).

PR #7538, opened 2021-11-05
Use extra ITS error on the ioutils::convertCompactClusters level

shahor02:


shahor02:
Hi @mpuccio 
The per-layer misalignemnt parameters (in fact, in Y and Z separately, for the sake of being general) must be introduced not only for ITS tracking but in all ITS refit (done at different  matching steps).  Therefore I've changed the TrackingConfigParam.h to https://github.com/AliceO2Group/AliceO2/blob/0a89233d198f487369c072e2bd491278b560b4a4/Detectors/ITSMFT/ITS/tracking/include/ITStracking/TrackingConfigParam.h#L43-L44 and apply it directly on the cluster conversion level in https://github.com/AliceO2Group/AliceO2/blob/0a89233d198f487369c072e2bd491278b560b4a4/Detectors/ITSMFT/ITS/tracking/src/IOUtils.cxx#L87-L91.
Note that the meaning of the parameter is an additional *squared* error.

I did not want to touch your code too much, so changed only https://github.com/AliceO2Group/AliceO2/blob/0a89233d198f487369c072e2bd491278b560b4a4/Detectors/ITSMFT/ITS/tracking/src/Tracker.cxx#L653 assuming that your 'params.LayerMisalignment' is indeed the error and not an error^2.
But I think you have a bug in https://github.com/AliceO2Group/AliceO2/blob/0a89233d198f487369c072e2bd491278b560b4a4/Detectors/ITSMFT/ITS/tracking/src/Tracker.cxx#L374-L375:
the `cov[0]` and `cov[2]` are already squared errors, `hypot` will square them once more...
Instead, one would need to simply add misalignments squared errors to cov.

Because I (almost) did not touch your code, with this PR the ITS tracking would need to be run with TrackingConfigParam containing 0s (since you apply the misalignment errors in the tracker) while the ITS-TPC and TRD matchers (which refit ITS tracks) will use them with 
```
ITSEXTRAERR="ITSCATrackerParam.misalignmentY2[0]=9e-4;ITSCATrackerParam.misalignmentZ2[0]=9e-4;ITSCATrackerParam.misalignmentY2[1]=9e-4;ITSCATrackerParam.misalignmentZ2[1]=9e-4;ITSCATrackerParam.misalignmentY2[2]=9e-4;ITSCATrackerParam.misalignmentZ2[2]=9e-4;ITSCATrackerParam.misalignmentY2[3]=1e-2;ITSCATrackerParam.misalignmentZ2[3]=1e-2;ITSCATrackerParam.misalignmentY2[4]=1e-2;ITSCATrackerParam.misalignmentZ2[4]=1e-2;ITSCATrackerParam.misalignmentY2[5]=1e-2;ITSCATrackerParam.misalignmentZ2[5]=1e-2;ITSCATrackerParam.misalignmentY2[6]=1e-2;ITSCATrackerParam.misalignmentZ2[6]=1e-2;"
```
But for the uniformity and to avoid different settings, I would suggest to do the same in the ITS tracker: leave adding extra errors to ioutils. I see that in https://github.com/AliceO2Group/AliceO2/blob/0a89233d198f487369c072e2bd491278b560b4a4/Detectors/ITSMFT/ITS/tracking/src/TrackerTraitsCPU.cxx#L259 you are using the misalignment params, but then the `resolution` is used only for the logging

mpuccio:
Indeed that is a bug! I am fine with the other changes. I will then avoid using the misalignment again in `resolution`, but I will need to change a bit the code as I use it to compute the sigma_z (TrackerTraitsCPU.cxx#L86) and during trackleting I don't have access to the uncertainties of the single clusters. Thanks for these modifications!

Detectors/Vertexing/include/DetectorsVertexing/PVertexer.h
```diff
@@ -37,7 +37,7 @@
 
 //TODO: MeanVertex and parameters input from CCDB
 
-//#define _PV_DEBUG_TREE_ // if enabled, produce dbscan and vertex comparison dump
+#define _PV_DEBUG_TREE_ // if enabled, produce dbscan and vertex comparison dump
```
mpuccio: @shahor02  Just be careful here, I think this is a leftover

shahor02: Thanks,indeed, need to comment this out

shahor02: Why don't you use directly the misalignment from the configurable param to compute the sigmas, applying some default value if it is empty?

shahor02:
all checks were passed, merging after a trivial shortening of the parameter name from `misalignmentY/Z2` to `sysErrY/Z2`.

PR #7540, opened 2021-11-05
Tag vertices as messageable

shahor02:


shahor02:
Hi @davidrohr 

I guess the fullCI fails because the `#include <type_traits>` is guarded in the GPU headers? should I also guard `is_messageable` declarations? By which guard? I am always confused by the multitude of various GPU guards.

davidrohr:
I can only check tonight in detaol. The guards are described in GPUCommonDef.h. And in the meantine there shoupd also be a GPUCommonTypeTraits, but I think for you case guarding will be enough.

Kind Regards
David Rohr

davidrohr (approved):

PR #7541, opened 2021-11-06
DPL Analysis: fix Build<> to respect index table declaration

aalkin:
`Builds<T>` template will now use index table metadata to chose building algorithm, previously building exclusive table would choose sparse index builder path unless `BuildsExclusive<T>` was used, regardless if index table `T` was declared as exclusive or not.

### DIFF
--- a/Framework/Core/include/Framework/AnalysisHelpers.h
+++ b/Framework/Core/include/Framework/AnalysisHelpers.h
@@ -348,7 +348,7 @@ struct IndexSparse {
 };
 
 /// This helper struct allows you to declare index tables to be created in a task
-template <typename T, typename IP = IndexSparse>
+template <typename T, typename IP = std::conditional_t<aod::MetadataTrait<T>::metadata::exclusive, IndexExclusive, IndexSparse>>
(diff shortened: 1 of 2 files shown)


Framework/Core/include/Framework/AnalysisHelpers.h
```diff
@@ -348,7 +348,7 @@ struct IndexSparse {
 };
 
 /// This helper struct allows you to declare index tables to be created in a task
-template <typename T, typename IP = IndexSparse>
+template <typename T, typename IP = std::conditional_t<aod::MetadataTrait<T>::metadata::exclusive, IndexExclusive, IndexSparse>>
```
ktf: Why do you need the IP in the template signature? Can't you simply use it below?

ktf (dismissed):

Framework/Core/include/Framework/AnalysisHelpers.h
ktf: Thanks. Looks like there is some other adjustments to be done following that.

ktf (approved):

PR #7544, opened 2021-11-07
TPC VDrift calibration from ITS/TPC tracks tgl comparison

shahor02:


shahor02:
failures unrelated, merging.

PR #7546, opened 2021-11-08
add bit for physical primary

jgrosseo:


jgrosseo:
@sawenzel @ekryshen @nburmaso as discussed this morning

PR #7547, opened 2021-11-08
[EMCAL-534] Add option to switch off pedestal subtraction for non-ZS …

mfasDa:
…events

Pedestal subtraction steered via the ZS status in the RCU
trailer. In case the data is taken in no-ZS mode the
pedestal subtraction is enabled. A dedicated setter is
added to bypass the determination via the RCU trailer
in case the pedestals should not be evaluated on
request.

In case of ZS Data the pedestal subtraction is always
disabled as it is performed online.

### DIFF
--- a/Detectors/EMCAL/reconstruction/src/CaloRawFitter.cxx
+++ b/Detectors/EMCAL/reconstruction/src/CaloRawFitter.cxx
@@ -63,7 +63,7 @@ int CaloRawFitter::getErrorNumber(CaloRawFitter::RawFitterError_t fiterror)
   return -1;
 }
 
-//Default constructor
+// Default constructor
 CaloRawFitter::CaloRawFitter(const char* name, const char* nameshort) : mMinTimeIndex(-1),
                                                                         mMaxTimeIndex(-1),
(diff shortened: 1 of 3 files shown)


mfasDa:
Needed under certain condition when we need to run with more than 15 ALTRO samples. Reason was that in the pilot beam our signal was shifted by additional 400 ns (so peak at 1 mus), leading to exceeding the 1.5 mus readout window for large signals. When enlarging the window and running in non-ZS mode we had some cases in which we were unable to evaluate the pedestals due to insufficient amount of samples. In such cases the pedestal subtraction must be switched off manually. This might be needed again at the beginning of the commissioning next year for trigger alignment.

In the normal scenario we run with ZS on and therefore pedestals are evaluated and subtracted in the ALTRO chip and therefore not handled in the reconstruction. This is steered by the ZS mode propagated via the RCU trailer. Also in MC we also use ZS-mode.

shahor02 (approved):

shahor02:
Failure is unrelated, merging.

@mpuccio now the ITS tracker even produced `bad memalloc`

PR #7558, opened 2021-11-08
DPL: Storing actual message size in the DataRef

matthiasrichter:
When multiple payloads per header will be supported, the payload
size in the DataHeader can only store one value and we need the
actual payload message size. DataRefUtils::getPayloadSize should
always be used.

Most of the user code has already been updated in #7034.

Part of #7110

### DIFF
--- a/Framework/Core/include/Framework/DataRefUtils.h
+++ b/Framework/Core/include/Framework/DataRefUtils.h
@@ -168,7 +168,14 @@ struct DataRefUtils {
     if (!header) {
       return 0;
     }
-    return header->payloadSize;
+    // in case of an O2 message with multiple payloads, the size of the message stored
+    // in DataRef is returned,
+    // as a prototype solution we are using splitPayloadIndex == splitPayloadParts to
+    // indicate that there are splitPayloadParts payloads following the header
+    if (header->splitPayloadParts > 1 && header->splitPayloadIndex == header->splitPayloadParts) {
+      return ref.payloadSize;
+    }
+    return header->payloadSize < ref.payloadSize || ref.payloadSize == 0 ? header->payloadSize : ref.payloadSize;
   }
 
   template <typename T>
--- a/Framework/Core/include/Framework/InputRecord.h
+++ b/Framework/Core/include/Framework/InputRecord.h
@@ -234,9 +234,7 @@ class InputRecord
       // the buffer to be deleted when it goes out of scope. The string is built
       // from the data and its lengh, null-termination is not necessary.
       // return std::string object
-      auto header = DataRefUtils::getHeader<header::DataHeader*>(ref);
-      assert(header);
-      return std::string(ref.payload, header->payloadSize);
+      return std::string(ref.payload, DataRefUtils::getPayloadSize(ref));
 
       // implementation (c)
     } else if constexpr (std::is_same<T, char const*>::value) {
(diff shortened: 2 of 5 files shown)


matthiasrichter:
@ktf the PR concerning the message consumption is merged, right? So there should not be a new conflict.

Framework/Core/include/Framework/DataRefUtils.h
```diff
@@ -168,7 +168,12 @@ struct DataRefUtils {
     if (!header) {
       return 0;
     }
-    return header->payloadSize;
+    // in case of an O2 message with multiple payloads, the size of the message stored
+    // in DataRef is returned
+    if (header->splitPayloadParts > 1 && header->splitPayloadIndex == header->splitPayloadParts) {
```
ktf, line 175: I am not sure I understand this if. Why would `header->splitPayloadIndex == header->splitPayloadParts` matter?

Framework/Core/include/Framework/InputRecord.h
```diff
@@ -285,25 +279,27 @@ class InputRecord
         throw runtime_error("Inconsistent serialization method for extracting span");
       }
       using ValueT = typename T::value_type;
-      if (header->payloadSize % sizeof(ValueT)) {
+      auto payloadSize = DataRefUtils::getPayloadSize(ref);
+      if (payloadSize % sizeof(ValueT)) {
         throw runtime_error(("Inconsistent type and payload size at " + std::string(ref.spec->binding) + "(" + DataSpecUtils::describe(*ref.spec) + ")" +
```
ktf, line 284: I would just use fmt::format, actually...

Framework/Core/src/DataRelayer.cxx
```diff
@@ -512,9 +512,10 @@ void DataRelayer::getReadyToProcess(std::vector<DataRelayer::RecordAction>& comp
         auto payload = partial[idx].payload(part).get();
         return DataRef{nullptr,
                        reinterpret_cast<const char*>(header->GetData()),
-                       reinterpret_cast<const char*>(payload ? payload->GetData() : nullptr)};
+                       reinterpret_cast<char const*>(payload ? payload->GetData() : nullptr),
+                       payload ? payload->GetSize() : 0};
       }
-      return DataRef{nullptr, nullptr, nullptr};
+      return DataRef{nullptr, nullptr, nullptr, 0};
```
ktf: I think this can now be `return {};` which will keep clang-code-checker happy (there is a modernize check which fails, when enabled).

ktf (changes requested):

Framework/Core/include/Framework/DataRefUtils.h
matthiasrichter: we have agreed on with @ironMann that this will be the indication that `splitPayloadParts` payloads will follow. This is a wip solution, think we will later introduce a dedicated header version without member layout change. I will make the comment more descriptive.

Framework/Core/include/Framework/InputRecord.h
matthiasrichter: did not change in the runtime error message here, its just after the change, but we can correct all that later

Framework/Core/src/DataRelayer.cxx
matthiasrichter: ok, fixed

matthiasrichter:
All requested changes addressed.

PR #7559, opened 2021-11-09
[WIP] DPL: make sure Dispatchers outputs are using Pub/Sub

ktf:


ktf:
@Barthelemy @knopers8 this should be roughly what is needed. Possible additional things which might be done:

* Tweak the output queue size so that messages are dropped earlier (defaults to 10000 elements, iirc, which is obviously not a good idea).
* Have a command line option to enable the feature.

ktf:
@dennisklein @rbx I now realise that pub/sub is not supported for shared memory, at least according to (https://github.com/FairRootGroup/FairMQ/blob/master/docs/Transport.md#2-transport-interface) is there any hope to get that supported?

rbx:
With the shared ownership we are one step towards being able to do pub/sub. But there are still number of other technical points that would have to be addressed in a possible pub/sub implementation - lack of delivery guarantee, dynamically changing number of peers, dealing with references to already-destroyed messages, possibly more. Short answer: it will require time to think about and implement.

Could you now do something like?:
```
for (o in outputs) {
   FairMQMessagePtr msgCopy(NewMessageFor(o));
   msgCopy->Copy(originalMsg);
   o->Send(msgCopy);
}
```
This would be the equivalent of pub/sub with delivery guarantee, and the drawback of needing to know all the peers by name.

ktf:
I think all we need is actually dropping messages if the receiver is not ready. Communication is in any case one to one.

rbx:
Then you could call `Send()` with a timeout of `0`:
https://github.com/FairRootGroup/FairMQ/blob/master/fairmq/Device.h#L88-L89

ktf:
Right... Would it be possible to have the default timeout configured via channel-config? Otherwise, I will have to track out-of-band the extra option...

rbx:
Yes, I can add that.

ktf:
Thanks. Notice that it should be the default timeout. I still want the ability to override it on a per-Send-call basis.

PR #7560, opened 2021-11-09
DPL: Adjusting the InputSpan backend

matthiasrichter:
InputRecord is providing all payloads as header payload pairs, even
if there are multiple payloads per header. The InputSpan backend is
using an additional index to allow the mapping.

### DIFF
--- a/Framework/Core/src/DataProcessingDevice.cxx
+++ b/Framework/Core/src/DataProcessingDevice.cxx
@@ -1163,18 +1163,26 @@ bool DataProcessingDevice::tryDispatchComputation(DataProcessorContext& context,
       currentSetOfInputs = relayer->consumeExistingInputsForTimeslice(slot);
     }
     auto getter = [&currentSetOfInputs](size_t i, size_t partindex) -> DataRef {
-      if (currentSetOfInputs[i].size() > partindex) {
-        auto header = currentSetOfInputs[i].header(partindex).get();
-        auto payload = currentSetOfInputs[i].payload(partindex).get();
-        return DataRef{nullptr,
-                       static_cast<char const*>(header->GetData()),
-                       static_cast<char const*>(payload ? payload->GetData() : nullptr),
-                       payload ? payload->GetSize() : 0};
+      if (currentSetOfInputs[i].getNumberOfPairs() > partindex) {
+        const char* headerptr = nullptr;
+        const char* payloadptr = nullptr;
+        size_t payloadSize = 0;
+        // - each input can have multiple parts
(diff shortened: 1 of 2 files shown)


Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -1155,26 +1155,54 @@ bool DataProcessingDevice::tryDispatchComputation(DataProcessorContext& context,
   };
 
   //
+  // the number of parts per input depends on the number of added PartRef object
+  // and the number of payloads in these objects. ProcessingContext provdes data
+  // through the input span where every payload is provided through a DataRef
+  // object. The part index within the input needs to be translated into a 2 dim
+  // index to access the payload message i from PartRef object j and we generate
+  // a consecutive map of these tuples
+  std::vector<std::vector<std::pair<size_t, size_t>>> partRefMaps;
```
matthiasrichter: By isolating the individual parts of #7110, its now verified that the unit test failure of QualityControl/testCheckWorkflow is triggered by introducing this extra variable and using it in the lambdas. Since this variable is defined at the functions' scope like `currentSetOfInputs` and it is captured by reference we can rule out a simple out-of-scope volatile object.

This needs to be further investigate to exclude other errors. Here we have solved the problem by adding the ref index to the MessageSet. Adding this commit solves the unit test failure.

matthiasrichter: found the problem. This index can go out of sync with the vector of message sets in the loop for processing the list of ready-actions. In this loop, the message set can now be cleared by early forwarding. The index has now been incorporated into `MessageSet`.

matthiasrichter:
@TimoWilken what is wrong with the CI? Are the logs lost so that the status can not be evaluated?
```
Finished building on alibuild06.cern.ch at 2021-11-22-04:59:03
Built commit 2618d2ef1e03d4aa969be28daf4779e8edcb77cb from https://github.com/AliceO2Group/AliceO2/pull/7560
To log into the build machine, use:
    aurora task ssh -l root build/mesosci/prod/ci_alisw_slc7-builder_latest/6
No logs found. Please check the aurora log.
See http://alisw.github.io/infrastructure-pr-testing for more instructions.
```

TimoWilken:
Hi @matthiasrichter, unfortunately, CI is having some trouble reaching GitHub at the moment. This doesn't appear in the logs at the moment. The checks have been retried, and are all green now.

matthiasrichter:
Hi @TimoWilken, great, thanks

PR #7562, opened 2021-11-09
[MID] improve MID tracking

pillot:
- fix chi2 and NDF calculation and chi2 cut
- fix track selection in the algorithm keeping only the best track candidates (keep track with more cluster, then best chi2 in case of equality)
- really consider all combinations (attaching a maximum of clusters) in the algorithm keeping all track candidates
- remove unnecessary calculations + other improvements to speed up the code
- change the devices to use the tracking algorithm keeping all track candidates


dstocco:
@aphecetche : approved from my side

aphecetche (approved):

PR #7564, opened 2021-11-09
TFInfo recording + disable option requiring perfect calibration + fixes

shahor02:


shahor02:
tested locally, merging

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -196,7 +196,8 @@ for det in `echo $LIST_OF_DETECTORS | sed "s/,/ /g"`; do
 done
 PVERTEX_CONFIG="--vertexing-sources $TRACK_SOURCES --vertex-track-matching-sources $TRACK_SOURCES"
 
-has_detector_reco FT0 && PVERTEX_CONFIG+=" --validate-with-ft0"
```
davidrohr, line 198: Do we want to disable this also for the FST with simulated data?
We could use `[[ $IS_SIMULATED_DATA == 1 ]] && `?

shahor02: I don't think this is needed: it is used for the cleanup of fake vertices and has a relatively small effect.

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -381,14 +386,14 @@ has_detector FV0 && has_processing_step FV0_RECO && add_W o2-fv0-reco-workflow "
 has_detector ZDC && has_processing_step ZDC_RECO && add_W o2-zdc-digits-reco "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC"
 has_detectors_reco MFT MCH && has_detector_matching MFTMCH && add_W o2-globalfwd-matcher-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N globalfwd-track-matcher MATCH REST)"
 
-if [ $BEAMTYPE != "cosmic" ]; then
+if [[[ $BEAMTYPE != "cosmic" ]]; then
```
davidrohr, line 389: Are you sure you tested it locally? This should have failed? (just fixed in #7567)

shahor02: I run only sync part on already existing data and it worked w/o a problem. But now I see that in complains (still running well):
```
WORKFLOWMODE=print prodtests/full-system-test/dpl-workflow.sh 
prodtests/full-system-test/dpl-workflow.sh: line 389: [[[: command not found
#Workflow command:

o2-raw-file-reader-workflow --session default --severity info --shm-segment-id 0 --shm-segment-size 8589934592  --early-forward-policy noraw --monitoring-backend no-op:// --fairmq-rate-logging 0 --detect-tf0 --delay 100 --loop 1 --max-tf 0 --input-conf /home/shahoian/alice/O2/rawAll.cfg  --configKeyValues "NameConf.mDirGRP=/home/shahoian/alice/O2;NameConf.mDirGeom=/home/shahoian/alice/O2;NameConf.mDirCollContext=/home/shahoian/alice/O2;NameConf.mDirMatLUT=/home/shahoian/alice/O2;keyval.input_dir=/home/shahoian/alice/O2;keyval.output_dir=/dev/null;;HBFUtils.nHBFPerTF=128;" | \
...
o2-dpl-run --session default --severity info --shm-segment-id 0 --shm-segment-size 8589934592  --early-forward-policy noraw --monitoring-backend no-op:// --fairmq-rate-logging 0 
```

But the async reco got some problem with cascades filtering in AOD, will check

chiarazampolli: There is a "[" too much on line 389.

shahor02: this is what @davidrohr fixed.

PR #7566, opened 2021-11-10
ctf-reader sends dummy output if requested detector is not in CTF

shahor02:


peressounko (dismissed):

davidrohr (approved):
I think with that additional option it makes fully sense and is safe.

PR #7568, opened 2021-11-10
Fix TF 1st orbit usag + flag to disable s-vertices processing

shahor02:


shahor02:
Hi @TimoWilken , the clang-format  fails due to the docker, could you please check?

TimoWilken:
Hi @shahor02, that seems to have been a transient issue on GitHub's side. A re-run fixed it.

PR #7570, opened 2021-11-10
Fix needed for TOF calibration to avoid truncation double->float

noferini:
@chiarazampolli 
This is needed to avoid do manage info calib (t - txp) with large values (t up to 11000000000 ps in timeframe) which are then truncated to float.

This happens if FT0 is not available.

Now (t - texp) convert to time in orbit instead of timeframe before to be converted to calibration info.

Other optimization in...


noferini:
FullCI failing because of
task timeout reached .. killing all processes
I don't know if it is related to the recent problem with the dpl getting stuck.
It doesn't seems related to this PR.
Can we merge it (I need it to collect calibrations on EPNs)

shahor02 (approved):

PR #7572, opened 2021-11-10
DPL: revert completion policy for dummy sink

ktf:


ktf:
Looks like this still is not enough... :-/

PR #7575, opened 2021-11-10
Few bug fixes for CI and cosmics

mpuccio:
- SigmaTanL for cosmics set to sync_misaligned << This should mitigate the issue in the technical runs @davidrohr
- Change default for cluster studies << Minor
- Compute deltaPhi cut starting from minpt << This makes it easier to tune the cut
- Don't use TGeo as default for material correction << This change makes the tracker 4 times faster wrt LUT and it is the new default when matbut.root is not available

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/src/TimeFrame.cxx
+++ b/Detectors/ITSMFT/ITS/tracking/src/TimeFrame.cxx
@@ -39,6 +39,11 @@ float MSangle(float mass, float p, float xX0)
   return 0.0136f * std::sqrt(xX0) * (1.f + 0.038f * std::log(xX0)) / (beta * p);
 }
 
+float Sq(float v)
+{
+  return v * v;
+}
+
 } // namespace
 
 namespace o2
@@ -188,7 +193,6 @@ void TimeFrame::initialise(const int iteration, const MemoryParameters& memParam
     mTrackletsLookupTable.resize(trkParam.CellsPerRoad());
     mIndexTables.clear();
     mIndexTableUtils.setTrackingParameters(trkParam);
-    mMSangles.resize(trkParam.NLayers);
     mPositionResolution.resize(trkParam.NLayers);
     mBogusClusters.resize(trkParam.NLayers, 0);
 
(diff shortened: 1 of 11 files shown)


shahor02 (approved):
failures are unrelated, merging

Detectors/ITSMFT/ITS/tracking/src/TimeFrame.cxx
```diff
@@ -275,6 +278,27 @@ void TimeFrame::initialise(const int iteration, const MemoryParameters& memParam
...
+
+    if (iLayer < mClusters.size() - 1) {
+      const float& r1 = trkParam.LayerRadii[iLayer];
+      const float& r2 = trkParam.LayerRadii[iLayer + 1];
+      const float res1 = std::hypot(trkParam.PVres, mPositionResolution[iLayer]);
+      const float res2 = std::hypot(trkParam.PVres, mPositionResolution[iLayer + 1]);
+      const float cosTheta1half = std::sqrt(1.f - Sq(0.5f * r1 * oneOverR));
+      const float cosTheta2half = std::sqrt(1.f - Sq(0.5f * r2 * oneOverR));
+      float x = r2 * cosTheta1half - r1 * cosTheta2half;
+      float delta = std::sqrt(1. / (1.f - 0.25f * Sq(x * oneOverR)) * (Sq(0.25f * r1 * r2 * Sq(oneOverR) / cosTheta2half + cosTheta1half) * Sq(res1) + Sq(0.25f * r1 * r2 * Sq(oneOverR) / cosTheta1half + cosTheta2half) * Sq(res2)));
```
sawenzel, line 297: better to put `1.f` everywhere (`1.f/(...`)! Otherwise, the division will be done as double.

mpuccio: Indeed! On some machine I even get the warning to fix this. In the next PR it will be done

davidrohr:
@mpuccio : I don't fully understand, does this now cut the combinatorics that led to the quasi endless loops? Or does it just reduce it? Can't we just have a limit that it aborts after attemping a certain unreasonable amount of track candidates?

What about the full system test CI? In there we have Pb-Pb events. Do we meanwhile know why it times out there? Is it fixed with this PR? OK, I mean we can just wait tomorrow and then see.

mpuccio:
The full system test CI now will take 1 second, the async cuts were too loose. I still have to had the protection for bogus number of roads, but I need to understand what is an unreasonable number of track candidates with the final cuts before.

PR #7578, opened 2021-11-11
DPL: inputs need to be consumed also on discard

ktf:


ktf:
Currently broken. Merging.

shahor02:
@ktf I confirm that it solves the problem.

ktf:
@chiarazampolli do you want a daily build with this in to attempt production on the grid?

chiarazampolli:
This will be in tonight’s tag, right? Or you are proposing to tag earlier? I will be online from 22, if a new tag is there, i can try it.

PR #7580, opened 2021-11-11
Fix completion policy

ktf:


ktf:
@davidrohr this fixes the completion policy as we discussed yesterday.

Framework/Core/src/CompletionPolicyHelpers.cxx
```diff
@@ -112,27 +113,50 @@ CompletionPolicy CompletionPolicyHelpers::consumeWhenAll(const char* name, Compl
...
+          }
+        }
+        if (input.payload != nullptr) {
+          withPayload++;
+        }
+        current++;
       }
-      if (input.payload != nullptr) {
-        withPayload++;
+      // * In case we have all inputs but the sporadic ones: Consume, since we do know if the sporadic ones
```
davidrohr: should be "since we do not know"

Framework/Core/src/CompletionPolicyHelpers.cxx
```diff
@@ -112,27 +113,50 @@ CompletionPolicy CompletionPolicyHelpers::consumeWhenAll(const char* name, Compl
...
+      // * In case we have only sporadic inputs: Consume, since we do not know if we already Consumed
+      // the non sporadic ones above.
+      // * In case we do not have payloads: Wait
+      // * In all other cases we consume what is there, but we wait for the non sporadic ones to be complete
+      //   (i.e. we wait for present + maxSporadic).
+      if (present + maxSporadic == inputs.size()) {
+        return CompletionPolicy::CompletionOp::Consume;
+      } else if (present - sporadic == 0) {
+        return CompletionPolicy::CompletionOp::Consume;
+      } else if (withPayload == 0) {
```
davidrohr: I don't understand this part.
- `if (present - sporadic == 0) {` covers the case that we have only sporadic inputs, so I agree we have to consume here.
- Which case does `if (present + maxSporadic == inputs.size()) {` cover? I don't think it covers the case "In case we have all inputs but the sporadic ones". Since `present` can contain the sporadic ones.I think it should be `present - sporadic + maxSporadic == inputs.size()`, or do I miss something?

Framework/Core/src/CompletionPolicyHelpers.cxx
ktf:
```suggestion
      // * In case we have all inputs but the sporadic ones: Consume, since we do not know if the sporadic ones
```

Framework/Core/src/CompletionPolicyHelpers.cxx
ktf:
```suggestion
      if (present - sporadic + maxSporadic == inputs.size()) {
        return CompletionPolicy::CompletionOp::Consume;
      } else if (present - sporadic == 0) {
        return CompletionPolicy::CompletionOp::Consume;
      } else if (withPayload == 0) {
```

ktf:
Of course you are right about the `present - sporadic + maxSporadic`. Fixed.

davidrohr:
@ktf : some framework tests are failing,. Looks like real errors to me.

ktf:
Indeed they were real. Fixed.

ktf:
Works fine on hyperloop.

davidrohr:
Rate throttling still not working, but seems not to break the full system test, so let's merge it to have it in. Can follow up in a separate PR.

ktf:
rate throttling is indeed a separate PR, which I think we can now merge:

ktf:
Nevermind... I will check why it's not working.

davidrohr:
I meant rate throttling for sync reco. I have commented in #7605 why it is not working.

PR #7584, opened 2021-11-11
[EMCAL-645] Fix reading event IR from cell trigger records

mfasDa:
- Fix incorrect use of trigger record container for
  cells, which was incorrectly take from the
  trigger records from the cell index container
- Use std::optional for internal handling which
  trigger record is available


mfasDa:
Blocks #7521

shahor02 (approved):

PR #7592, opened 2021-11-12
DPL: introduce sending policy

ktf:
Useful to customise behavior based on certain output types,
e.g. to make sure QC sampled data can be dropped in case of
backpressure.

### DIFF
--- a/Framework/Core/CMakeLists.txt
+++ b/Framework/Core/CMakeLists.txt
@@ -36,6 +36,7 @@ o2_add_library(Framework
                        src/ControlService.cxx
                        src/ControlServiceHelpers.cxx
                        src/DispatchPolicy.cxx
+                       src/DataSender.cxx
                        src/ProcessingPoliciesHelpers.cxx
                        src/ConfigParamStore.cxx
                        src/ConfigParamsHelper.cxx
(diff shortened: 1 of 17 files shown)


ktf:
@Barthelemy @knopers8  this should allow you to customise the timeout for sending the data, based either on a device by device basis, or even at the level of the single message. Have a look at:

https://github.com/AliceO2Group/AliceO2/pull/7592/files#diff-7f4b4bbc3aff8ff891aa123dfbcc02cf21485ec49cbf9388b902f07ba048f450R47-R56

for an example. We can discuss off-line, if you have any doubts.

knopers8 (approved):

davidrohr:
Do I understand correctly that this would be a timeout to drop the message if the receiver is not yet ready and the timeout expires.
I.e. we could configure that if QC doesn't accept a message in 5 seconds, we do not forward it to QC (we would still send it to other receivers). If that is the case, this makes totally sense.

ktf:
@davidrohr dropping the message would be possible so this would be ok for sampled data. 
However skipping one non-responding device would not be so easy, because it would need an actual channel to it (it would also be difficult to decide wether a given device can be just skipped).
On the other hand this would also allow having an (optional) way to dump all messages to disk or things like that.

davidrohr:
ok, so this would basically be after the dispatcher and in front of QC devices with sampling, or perhaps actually also between all QC devices with sampling. And it would guarantee that if one of them is stuck, the messages are dropped?

ktf:
Correct.

Ciao,
Giulio

PR #7593, opened 2021-11-12
Skip dedicated macro config for MCReplay simulation

benedikt-voelkel:


sawenzel (approved):

davidrohr:
Does this fix
```
G__MCReplayCore dictionary payload:29:10: fatal error: 'MCStepLogger/StepInfo.h' file not found
#include "MCStepLogger/StepInfo.h"
```
in `Detectors/gconfig/mcreplayConfig.C` since then I can stop investigating :)

davidrohr (approved):
indeed, it does. Thx!

PR #7603, opened 2021-11-13
Changes to inject TF creation time + set default remote access to alien

shahor02:
* CTF/TF readers: set default remote file access to alien instead of EOS
since by default the access to CTF or rawTF data will be done via alien, the defaults of `--remote-regex` and `--copy-cmd` of `o2-ctf-reader-workflow` and `o2-raw-tf-reader-workflow` are adapted to alien.

* Ad hoc correction of pilot beam run CTF times
CTFs created during the pilot beam test have no TFs creation time stored. This implements an ad-hoc correction for the global runs of the pilot beam.

* Possibility to set TF creation time in raw-file-reader workflow
By the default the DataProcessingHeader of each message will have its creation time set to `now()`. This can be changed by passing an option `--start-time <t>`: in this case the creation time will be defined as `t + firstTForbir*orbit_duration` in milliseconds.

### DIFF
--- a/DataFormats/Detectors/Common/src/CTFHeader.cxx
+++ b/DataFormats/Detectors/Common/src/CTFHeader.cxx
@@ -18,7 +18,7 @@ using DetID = o2::detectors::DetID;
 /// describe itsel as a string
 std::string CTFHeader::describe() const
 {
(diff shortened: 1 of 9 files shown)


shahor02:
dataflow-cs8 CI is stuck, merging

PR #7605, opened 2021-11-14
DPL: if all the inputs are sporadic, then also the outputs will be

ktf:
This is done to avoid having to change the lifetime of
all the devices which run on sampled and unsampled data alike.

### DIFF
--- a/Framework/Core/src/WorkflowHelpers.cxx
+++ b/Framework/Core/src/WorkflowHelpers.cxx
@@ -577,8 +577,45 @@ void WorkflowHelpers::injectServiceDevices(WorkflowSpec& workflow, ConfigContext
   extraSpecs.clear();
 }
 
-void WorkflowHelpers::adjustServiceDevices(WorkflowSpec& workflow, ConfigContext const& ctx)
+void WorkflowHelpers::adjustTopology(WorkflowSpec& workflow, ConfigContext const& ctx)
 {
+  for (auto& spec : workflow) {
+    auto& inputs = spec.inputs;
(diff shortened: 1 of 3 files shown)


davidrohr:
Does this work recursively? I.e. if device A gets sporadic outputs since all its inputs are sporadic, and device B receives messages from device A, the inputs of device B will only B all sporadic after the outputs of device A were adjusted. In that case also the outputs of device B must be adjusted. Does this PR work that way?

ktf:
I am actually not completely sure. At the point the adjustTopology is invoked, I think the devices are topologically sorted, so it should work as you want, but I would need to check.

ktf:
Works fine on hyperloop.

ktf:
Works fine on hyperloop.

davidrohr (changes requested):
@ktf : This seems to work for the sporadic inputs, but it still does not work for QC. The problem seems to be that QC gets an QC/TIMER-***/0 input from the internal dpl clock.
I think the check `allSporadic` has to ignore that, and then it also has to check that there is at least one real sporadic input.

Currently, the QC tasks are still sending summaries to the internal DPL clock (see attached plot)
qc

ktf:
I think this implements the changes you asked for although I am not 100% sure of the logic.

davidrohr (approved):
Passes the FST on the EPN if rate throttling is not enabled, i.e. doesn't break anything.
The incorrect summary messages are gone, but the test with rate throttling still gets stuck. I think we can merge this and keep serching for the next problem.

PR #7612, opened 2021-11-15
TPC: track and cluster filter, Kr file writer sync. reco

wiechula:
* Add a track and cluster filter for quick visualization and development purposes. The custers are stored together with track
* For synchronous data taking with the Kr-cluster finder a dedicated writer is implemented which stores also the required meta data file for automatic eos export and alien registration. This is based on the CTFWriter.


wiechula:
@davidrohr , can this be merged?

PR #7614, opened 2021-11-15
DPL GUI: add table with possible outputs in device inspector.

ktf:


ktf:
Merged by accident, however it should be fine (and it's not really tested).

PR #7615, opened 2021-11-15
TRD possibility to flag noisy MCMs

martenole:
Hi @tdietel 
this would add a simple CCDB object with a flag for each MCM whether it is noisy or not. A MCM is considered noisy if it sends 10 times more data then the average MCM does. For calculating the average 10% of MCMs with the largest amount of tracklets are excluded.
For the pilot beam I see 77 MCMs in total are flagged which corresponds to 0.12 % of all MCMs.
Do you think it would make sense to ignore data from noisy MCMs in reconstruction? I had posted the tracklet spectra with and without noise suppression on mattermost.

### DIFF
--- a/DataFormats/Detectors/TRD/CMakeLists.txt
+++ b/DataFormats/Detectors/TRD/CMakeLists.txt
@@ -20,6 +20,7 @@ o2_add_library(DataFormatsTRD
                        src/CTF.cxx
                        src/Digit.cxx
                        src/KrCluster.cxx
+                       src/NoiseCalibration.cxx
                PUBLIC_LINK_LIBRARIES O2::CommonDataFormat O2::SimulationDataFormat)
 
 o2_target_root_dictionary(DataFormatsTRD
@@ -36,6 +37,7 @@ o2_target_root_dictionary(DataFormatsTRD
                        include/DataFormatsTRD/Digit.h
                        include/DataFormatsTRD/KrCluster.h
                        include/DataFormatsTRD/KrClusterTriggerRecord.h
+                       include/DataFormatsTRD/NoiseCalibration.h
                        include/DataFormatsTRD/CTF.h
                        include/DataFormatsTRD/CalVdriftExB.h
                        include/DataFormatsTRD/SignalArray.h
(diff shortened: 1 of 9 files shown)


bazinski (dismissed):

tdietel (commented):
Hello Ole. 

Thanks for looking into this.

The concern I have is that this looks like a short term solution, and might not be in line with what we want in the long term. In the long term, I would expect that noise would be determined per pad, and not (only) per MCM. And once we parse the config events, we would also have a status flag per MCM - although I'm not sure if this would be handled in the same way as the MCM noise issue here.

What do you think? It might be that we drop this item from the CCDB at some point, but actually I think the name would not conflict with what we would do in the future, so there would not be an issue with this.

Cheers,
Tom.

martenole:
Hi Tom,

indeed, eventually the noise should be determined per pad. I attach here the tracklet position distribution for a noisy MCM. Some tracklets have reasonable values while the vast majority of tracklets comes from an individual pad. So in the end we would want to exclude only that pad. Will it be possible to turn off the noisy pads completely in the MCMs such that they don't contribute tracklets anymore?

In that case this calibration would not be needed. But it might be a good tool to double-check the TRAP performance and I think it would be useful for the analysis of pilot beam data. So maybe it can be added now and at some point be removed again when we have the final solution for the noise?

positionNoisyMCM

tdietel:
Hello Ole,

I completely agree that the analysis of the pilot beam data should be done rather quickly, so in these cases the simple rejection of all noisy MCMs is a reasonable thing to do.

Your plot is  a very clear indication that we are dealing with a single noisy pad.

In the long run, we should mask all these noisy pads, and this can be done in the FEE. If the whole MCM is questionable, we can also exclude that at the FEE level. The key is that we keep track of these patches in the CCDB, and this is the configuration events that we need to parse.

Cheers,
Tom.

tdietel (approved):
I approve this implementation as a temporary measure to quickly analyze the pilot beam data.

martenole:
That sounds very good to me. Thanks!

martenole:
All checks have passed. I also tried the noise calibration for simulated data sample and there 0 MCMs were flagged as expected. The object is already in the test ccdb and is ready to be used

PR #7617, opened 2021-11-15
CcdbApi: no need to follow in retrieveHeaders

sawenzel:
While retrieving header meta-data, we should not / don't need to follow re-directions.
This query may otherwise fail when the object is on Alien/etc.

### DIFF
--- a/CCDB/src/CcdbApi.cxx
+++ b/CCDB/src/CcdbApi.cxx
@@ -1216,7 +1216,7 @@ std::map<std::string, std::string> CcdbApi::retrieveHeaders(std::string const& p
     curl_easy_setopt(curl, CURLOPT_URL, fullUrl.c_str());
     /* get us the resource without a body! */
     curl_easy_setopt(curl, CURLOPT_NOBODY, 1L);
-    curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
+    curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 0L);
     curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, header_map_callback<>);
     curl_easy_setopt(curl, CURLOPT_HEADERDATA, &headers);
 


costing (approved):

ktf:
@Barthelemy @sawenzel QC testTriggers test seems to not work with this consistently (and only for this PR, AFAICT). Can you have a look?

sawenzel:
I confirm that the test breaks with this PR. Not sure what's going on.

I made the PR because I observed errors when querying from the official CCDB which has ALIEN redirection:

Before:
```
o2::ccdb::CcdbApi api
api.init("alice-ccdb.cern.ch")
api.retrieveHeaders("MFT/Calib/ClusterDictionary",std::map<string,string>())
**curl_easy_perform() failed: Unsupported protocol**
(std::map<std::string, std::string>) { "Access-Control-Allow-Headers" => "range", "Access-Control-Allow-Methods" => "HEAD,GET", "Access-Control-Allow-Origin" => "*", "Access-Control-Expose-Headers" => "content-range,content-length,accept-ranges", "Cache-Control" => "no-cache", "Content-Location" => "alien:///alice/data/CCDB/MFT/Calib/ClusterDictionary/06/28404/3e275460-4229-11ec-8357-200114580202", "Created" => "1636550544550", "Date" => "Tue, 16 Nov 2021 09:12:16 GMT", "ETag" => ""3e275460-4229-11ec-8357-200114580202"", "Last-Modified" => "Wed, 10 Nov 2021 13:22:24 GMT", "Location" => "alien:///alice/data/CCDB/MFT/Calib/ClusterDictionary/06/28404/3e275460-4229-11ec-8357-200114580202", "Transfer-Encoding" => "chunked", "UploadedBy" => "shahoian", "Valid-From" => "1635771600000", "Valid-Until" => "1640955600000", "partName" => "blob" }
```

After:
```
o2::ccdb::CcdbApi api
api.init("alice-ccdb.cern.ch")
api.retrieveHeaders("MFT/Calib/ClusterDictionary",std::map<string,string>())
(std::map<std::string, std::string>) { "Access-Control-Allow-Headers" => "range", "Access-Control-Allow-Methods" => "HEAD,GET", "Access-Control-Allow-Origin" => "*", "Access-Control-Expose-Headers" => "content-range,content-length,accept-ranges", "Cache-Control" => "no-cache", "Content-Location" => "alien:///alice/data/CCDB/MFT/Calib/ClusterDictionary/06/28404/3e275460-4229-11ec-8357-200114580202", "Created" => "1636550544550", "Date" => "Tue, 16 Nov 2021 09:10:20 GMT", "ETag" => ""3e275460-4229-11ec-8357-200114580202"", "Last-Modified" => "Wed, 10 Nov 2021 13:22:24 GMT", "Location" => "alien:///alice/data/CCDB/MFT/Calib/ClusterDictionary/06/28404/3e275460-4229-11ec-8357-200114580202", "Transfer-Encoding" => "chunked", "UploadedBy" => "shahoian", "Valid-From" => "1635771600000", "Valid-Until" => "1640955600000", "partName" => "blob" }
```

The answer seems to be the same in both cases but maybe something changes when we use meta-data filter.

We could also just disable the error message, as it doesn't seem to be relevant here (because we don't expect a payload response)?

PR #7623, opened 2021-11-16
Analysis task: sketching new structure

jgrosseo:


jgrosseo:
Would work with e.g. such a task: https://github.com/jgrosseo/O2Physics/blob/analysistask/PWGCF/Tasks/correlations.cxx

Framework/Core/test/test_AnalysisTask.cxx
```diff
@@ -122,10 +122,12 @@ struct ITask {
 };
 
 struct JTask {
-  Configurable<o2::test::SimplePODClass> cfg{"someConfigurable", {}, "Some Configurable Object"};
+  struct Config {
+    Configurable<o2::test::SimplePODClass> cfg{"someConfigurable", {}, "Some Configurable Object"};
+  } config;
```
ktf, line 127: Why not simply:
```
Config config {
  Configurable<o2::test::SimplePODClass> cfg{"someConfigurable", {}, "Some Configurable Object"};
};

PR #7624, opened 2021-11-16
DPL: avoid including Stack.h everywhere

ktf:


ktf:
For the record this brings the clean slate cumulative time to compile the framework down to 13 minutes, from 16...

ktf:
@TimoWilken the fullCI has been processing for over 4 hours.

PR #7627, opened 2021-11-16
TRD fix tracklet64 position and slope XOR with 0x80

bazinski:
Fix the position and slope extraction to include the XOR with 0x80


davidrohr:
@bazinski @ktf : Why do we have merge requests that come from AliceO2Group? Isn't the proposed behavior that everyone uses his own fork for PRs? This is not limited to sean, I have also seen it from others. This increases the number of branches in the AliceO2Group fork, and thus also the git clone time / size. Perhaps we should also clean up stale branches in here.

ktf:
This happens when someone edits the file inline and has write access to the repository.

bazinski:
@davidrohr Ah sorry I was editing on github website (not usual for me) thought i was on my own, my bad.

ktf:
Indeed we should cleanup branches

ktf:
@TimoWilken there is a bunch of rc/branches which are not really needed, after the tag is there. Can you clean them up?

PR #7629, opened 2021-11-16
TRD add XOR to slope and pos extraction of Tracklet64

bazinski:
- add the xor with 0x80 to the slope and position extraction from the raw values.
- fix padsize related to step size in the front electronics

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
@@ -51,7 +51,7 @@ constexpr int NCOLMCM = 18;     // the number of pads per MCM
 
 constexpr int NBITSTRKLPOS = 11;                   // number of bits for position in tracklet64 word
 constexpr int NBITSTRKLSLOPE = 8;                  // number of bits for slope in tracklet64 word
-constexpr float PADGRANULARITYTRKLPOS = 80.f;
+constexpr float PADGRANULARITYTRKLPOS = 40.f;      // TODO this should come out of the trapconfig.
 constexpr float PADGRANULARITYTRKLSLOPE = 1000.f;
 constexpr float GRANULARITYTRKLPOS = 1.f / PADGRANULARITYTRKLPOS;     // granularity of position in tracklet64 word in pad-widths
 constexpr float GRANULARITYTRKLSLOPE = 1.f / PADGRANULARITYTRKLSLOPE; // granularity of slope in tracklet64 word in pads/timebin
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
@@ -85,8 +85,8 @@ class Tracklet64
   GPUd() uint64_t getHCID() const { return ((mtrackletWord & hcidmask) >> hcidbs); };       // no units 0..1077
   GPUd() uint64_t getPadRow() const { return ((mtrackletWord & padrowmask) >> padrowbs); }; // pad row number [0..15]
   GPUd() uint64_t getColumn() const { return ((mtrackletWord & colmask) >> colbs); };       // column refers to MCM position in column direction on readout board [0..3]
-  GPUd() uint64_t getPosition() const { return ((mtrackletWord & posmask) >> posbs); };     // in units of 1/80 pads, 11 bit granularity [-12.8..12.8] relative to MCM center
-  GPUd() uint64_t getSlope() const { return ((mtrackletWord & slopemask) >> slopebs); };    // in units of 1/1000 pads/timebin, 8 bit granularity [-0.128 to 0.128]
+  GPUd() uint64_t getPosition() const { return ((mtrackletWord & posmask) >> posbs) ^ 0x80; };  // in units of 1/80 pads, 11 bit granularity [-12.8..12.8] relative to MCM center
+  GPUd() uint64_t getSlope() const { return ((mtrackletWord & slopemask) >> slopebs) ^ 0x80; }; // in units of 1/1000 pads/timebin, 8 bit granularity [-0.128 to 0.128]
   GPUd() uint64_t getPID() const { return ((mtrackletWord & PIDmask)); };                   // no unit, all 3 charge windows combined
   GPUd() uint64_t getQ0() const { return ((mtrackletWord & Q0mask) >> Q0bs); };             // no unit
   GPUd() uint64_t getQ1() const { return ((mtrackletWord & Q1mask) >> Q1bs); };             // no unit
--- a/Detectors/CTF/test/test_ctf_io_trd.cxx
+++ b/Detectors/CTF/test/test_ctf_io_trd.cxx
@@ -114,6 +114,12 @@ BOOST_AUTO_TEST_CASE(CTFTest)
   sw.Stop();
   LOG(info) << "Decompressed in " << sw.CpuTime() << " s";
 
+  //fix for XOR of 0x80 of pos and slope, the get methods have the XOR inside.
+  for (auto& tracklet : trackletsD) {
+    tracklet.setPosition(tracklet.getPosition());
+    tracklet.setSlope(tracklet.getSlope());
+  }
+
   BOOST_CHECK(triggersD.size() == triggers.size());
   BOOST_CHECK(trackletsD.size() == tracklets.size());
   BOOST_CHECK(digitsD.size() == digitsD.size());
--- a/Detectors/TRD/simulation/src/TrapSimulator.cxx
+++ b/Detectors/TRD/simulation/src/TrapSimulator.cxx
@@ -2028,11 +2028,13 @@ void TrapSimulator::fitTracklet()
         }
 
         slope = slope & 0xff; // 8 bit
+        slope = slope ^ 0x80; // TODO temporary hack to compensate for XOR in Tracklet64::getSlope() as per data coming off the trap chips
 
         if (position > 0x3ff || position < -0x400) {
           LOG(warning) << "Overflow in position with position of " << position << " in hex 0x" << std::hex << position;
         }
         position = position & 0x7ff; // 11 bits
+        position = position ^ 0x80;  //TODO temporary hack to compensate for XOR in Tracklet64::getPosition() as per data coming off the trap chips
 
         // assemble and store the tracklet word
         TrackletMCMData trackletword;


tdietel (dismissed):
Looks good.

Do we need a patch for the TRAP simulator for consistency?

bazinski:
> Looks good.
> 
> Do we need a patch for the TRAP simulator for consistency?

well I think the trapsim fix will be bigger, so for now just want to get the raw data correct?

DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
```diff
@@ -50,7 +50,7 @@ constexpr int NCOLMCM = 18;     // the number of pads per MCM
 
 constexpr int NBITSTRKLPOS = 11;                   // number of bits for position in tracklet64 word
 constexpr int NBITSTRKLSLOPE = 8;                  // number of bits for slope in tracklet64 word
-constexpr float PADGRANULARITYTRKLPOS = 80.f;
+constexpr float PADGRANULARITYTRKLPOS = 160.f;
```
martenole: Shouldn't this be 40?

bazinski: mmm multitiplied it by 2 in the qc to get a nice plot.  but i think you might have a point.

martenole:
> well I think the trapsim fix will be bigger, so for now just want to get the raw data correct?

But we should not break our simulation as well..

bazinski:
> > well I think the trapsim fix will be bigger, so for now just want to get the raw data correct?
> 
> But we should not break our simulation as well..

you mean the change to the pad size will break the simulation?

martenole:
The TRAP sim takes the pad granularity also from the constants, so that is fine. The XOR part is what is not included in the simulation. It would not break it in the sense of crashing, but it would have an impact on the performance (have not tested it though)

DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
martenole: did you multiply `PADGRANULARITYTRKLPOS` or `GRANULARITYTRKLPOS` which is the inverse of the former?

bazinski: ah you are indeed right. I meant to change GRANULATIRYTRKLPOS

bazinski: changed to 40

tdietel: Can you add a TODO comment that this should actually come from the TRAP config?

bazinski: done

tdietel (dismissed):

martenole:
Hi Sean,
could you please add this patch for the TRAP simulation?
```
diff --git a/Detectors/TRD/simulation/src/TrapSimulator.cxx b/Detectors/TRD/simulation/src/TrapSimulator.cxx
index e8ffd3bd57cf..f3ae60a73071 100644
--- a/Detectors/TRD/simulation/src/TrapSimulator.cxx
+++ b/Detectors/TRD/simulation/src/TrapSimulator.cxx
@@ -2033,6 +2033,7 @@ void TrapSimulator::fitTracklet()
           LOG(warning) << "Overflow in position with position of " << position << " in hex 0x" << std::hex << position;
         }
         position = position & 0x7ff; // 11 bits
+        position = position ^ 0x80; // TODO temporary hack to compensate for XOR in Tracklet64::getPosition()
 
         // assemble and store the tracklet word
         TrackletMCMData trackletword;

```
Otherwise the tracklet position in simulations is distorted and I checked that this costs about 50% of efficiency in the tracking. Plot below shows the tracklet distribution without the patch for simulations.

positionSimNew

Of course the change to a step of 1/40 pad widths means we use less of the available 11 bits for the position (in this plot the patch is applied)
positionNew

martenole:
Actually, the same should of course also be done for the deflection

bazinski:
> Actually, the same should of course also be done for the deflection

will check the simulations in a bit, just want to finish these slides and the pics associated with them.

martenole:
I wonder if changing the getters of Tracklet64 leads to issues with the CTF encoding/decoding. These CI issues seem to be real?
```
## sw/BUILD/O2-latest/log
406/406 Test #115: Detectors/CTF/test/test_ctf_io_trd.cxx ...............................................***Failed    9.77 sec
99% tests passed, 1 tests failed out of 406
100% tests passed, 0 tests failed out of 93
```
They were also there in the previous CI checks

bazinski:
> I wonder if changing the getters of Tracklet64 leads to issues with the CTF encoding/decoding. These CI issues seem to be real?
> 
> ```
> ## sw/BUILD/O2-latest/log
> 406/406 Test #115: Detectors/CTF/test/test_ctf_io_trd.cxx ...............................................***Failed    9.77 sec
> 99% tests passed, 1 tests failed out of 406
> 100% tests passed, 0 tests failed out of 93
> ```
> 
> They were also there in the previous CI checks

oh when one is completely convinced a pr is going to be trivial .... and then it comes back and bites you ... :-(

bazinski:
> > I wonder if changing the getters of Tracklet64 leads to issues with the CTF encoding/decoding. These CI issues seem to be real?
> > ```
> > ## sw/BUILD/O2-latest/log
> > 406/406 Test #115: Detectors/CTF/test/test_ctf_io_trd.cxx ...............................................***Failed    9.77 sec
> > 99% tests passed, 1 tests failed out of 406
> > 100% tests passed, 0 tests failed out of 93
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > They were also there in the previous CI checks
> 
> oh when one is completely convinced a pr is going to be trivial .... and then it comes back and bites you ... :-(

ah @martenole we need to xor the incoming position and slope in the constructor. The test is generating its own position and slope from a random distribution. This is becoming messy. Maybe there is a better way of doing this.

bazinski:
the last commit will get rebased away, just testing on an epn.

tdietel:
I am thinking about what we should actually store in the Tracklet64? If we want to change this, now would be a good time.

It might be easier to store the XORed and signed position and deflection, i.e. to do the bit manipulations when the tracklet is created. However, I just looked at the code, and everything is stored in a uint64_t anyways, so it might not make a big difference.

The other question is where we store the granularity - we kind of postponed it for now and said it would come out of the CCDB. Shall we have a short chat when Sean feels better?

bazinski:
> I am thinking about what we should actually store in the Tracklet64? If we want to change this, now would be a good time.
> 
> It might be easier to store the XORed and signed position and deflection, i.e. to do the bit manipulations when the tracklet is created. However, I just looked at the code, and everything is stored in a uint64_t anyways, so it might not make a big difference.
> 
> The other question is where we store the granularity - we kind of postponed it for now and said it would come out of the CCDB. Shall we have a short chat when Sean feels better?

On the question of storage, I would prefer if the trapsim accurately reflects the electronics/firmware i.e. it does the xoring itself, and then send pos and slope into tracklet64 already xored.
The problem with this pr is the ctf test, will sort out when screens dont seem so bright.
As for the granularity, yes a chat would be a good idea.

bazinski:
@shahor02 @martenole not a pretty fix to the test, but it gets the final boost tests pass as it should be.
I cant think of an elegant way of doing it, if you are happy with this cool, else will figure out a less cludgy way.

DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
```diff
@@ -85,8 +85,8 @@ class Tracklet64
   GPUd() uint64_t getHCID() const { return ((mtrackletWord & hcidmask) >> hcidbs); };       // no units 0..1077
   GPUd() uint64_t getPadRow() const { return ((mtrackletWord & padrowmask) >> padrowbs); }; // pad row number [0..15]
   GPUd() uint64_t getColumn() const { return ((mtrackletWord & colmask) >> colbs); };       // column refers to MCM position in column direction on readout board [0..3]
-  GPUd() uint64_t getPosition() const { return ((mtrackletWord & posmask) >> posbs); };     // in units of 1/80 pads, 11 bit granularity [-12.8..12.8] relative to MCM center
-  GPUd() uint64_t getSlope() const { return ((mtrackletWord & slopemask) >> slopebs); };    // in units of 1/1000 pads/timebin, 8 bit granularity [-0.128 to 0.128]
+  GPUd() uint64_t getPosition() const { return ((mtrackletWord & posmask) >> posbs) ^ 0x80; };  // in units of 1/80 pads, 11 bit granularity [-12.8..12.8] relative to MCM center
+  GPUd() uint64_t getSlope() const { return ((mtrackletWord & slopemask) >> slopebs) ^ 0x80; }; // in units of 1/1000 pads/timebin, 8 bit granularity [-0.128 to 0.128]
```
shahor02, line 89: what is the point of these `^0x80`? You are just flipping bit 8-th bit

Detectors/CTF/test/test_ctf_io_trd.cxx
```diff
@@ -52,7 +52,7 @@ BOOST_AUTO_TEST_CASE(CTFTest)
 
       for (int i = nTrk; i--;) {
         tracklets.emplace_back(formatTrk, hcid, gRandom->Integer(0x1 << 4), gRandom->Integer(0x1 << 2),
-                               gRandom->Integer(0x1 << 11), gRandom->Integer(0x1 << 8), gRandom->Integer(0x1 << 24));
+                               gRandom->Integer(0x1 << 11) ^ 0x80, gRandom->Integer(0x1 << 8) ^ 0x80, gRandom->Integer(0x1 << 24));
```
shahor02: here too

bazinski: Its done in the electronics, so the pos and slope stored in the tracklet64 class has the bit flipped.
The get method then reflips it so that the values used are the correct slope and position.
The final lines of the test  test::boost.....per_element the raw value not via the getslope and getposition method as far as i can see, hence the cludge. Correct me if i am wrong.

shahor02: ok, so this is just to emulate the weirdness of the firmware...

shahor02 (dismissed):

Detectors/CTF/test/test_ctf_io_trd.cxx
bazinski: oui

tdietel: In principle, we would have to keep the data fields as not bitflipped in Tracklet64, and do the conversion when decoding the raw data to tracklets (and digits).
Some background on the weirdness: the flipped bits are necessary to avoid misinterpretation of a tracklet word as a tracklet data end marker.

shahor02: > Some background on the weirdness: the flipped bits are necessary to avoid misinterpretation of a tracklet word as a tracklet data end marker.

Does this mean that the flipped bit is always there?

bazinski: the bit flipped data is stored inside the tracklet64 class, hence the conversion in the get methods. 
This leaves the current ctf alone.

martenole: Would it make sense to remove the change in this line, since it is not relevant (the actual fix is in the lines below)? Or do you think it makes the test more readable? Then we can leave it of course. 
That would also kick the CI again and then we can hopefully get this merged soon

bazinski: Sure it does seem rather random, to flip a bit, to someone not familiar reading the code. Or even us in 3 years time.

martenole (dismissed):

martenole:
Not sure if I understand the failed check correctly. Maybe you need to switch to lowercase fairlogger API now in all files which are touched by your commits? I see that in Detectors/CTF/test/test_ctf_io_trd.cxx there are still uppercase messages

bazinski:
ugh just when you think things are simple ... case changed.

martenole (dismissed):

martenole:
Hi @bazinski there seem to be some conflicts now :/ Could you please fix them?

bazinski:
This is like the never ending pr ...

davidrohr:
@martenole  @bazinski : Can this be merged now?

martenole:
Yes

PR #7630, opened 2021-11-16
[EMCAL-566] Fix Gauss fit in boost histogram utils

jokonig:
- debugged gaus fit function for boost histograms. Gaussian is
transformed into pol2 using logarithm. Pol2 is then fitted using
TLinearFitter. Procedure was validated with direct comparison to ROOT.

- Added function for 2d boost histogram projection to 1d boost
histogram. Therefore 2d histo is reduced and then the under/overflow bin
has to be set to 0 as the under/overflow is taken into account in the
projection when using boost histograms but not for root histograms.
Procedure was validated with direct comparison to ROOT.

- Adjusted bin numbers for boost histos in functions
boostHistoFromRoot_2D and boostHistoFromRoot_1D

### DIFF
--- a/Common/Utils/include/CommonUtils/BoostHistogramUtils.h
+++ b/Common/Utils/include/CommonUtils/BoostHistogramUtils.h
@@ -215,21 +215,27 @@ std::string createErrorMessage(FitGausError_t errorcode)
 /// \param first begin iterator of the histogram
 /// \param last end iterator of the histogram
 /// \param axisFirst axis iterator over the bin centers
+/// \param ignoreUnderOerflowBin switch to disable taking under and overflow bin into fit
 /// \return result
 ///      result is of the type fitResult, which contains 4 parameters (0-Constant, 1-Mean, 2-Sigma,  3-Sum)
 ///
 /// ** Temp Note: For now we forgo the templated struct in favor of a std::vector in order to
 /// have this compile while we are working out the details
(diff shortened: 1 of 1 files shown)


jokonig:
Please do not approve this PR yet. 
The PR 7631 has to be merged before and then this PR has to be adjusted

shahor02 (commented):
Note that there is already a robust and faster pseudo-gaussian (log-normal) fit method:
https://github.com/AliceO2Group/AliceO2/blob/e0fdbc8739d33d00579b696c04de6c32250f0f6c/Common/MathUtils/include/MathUtils/fit.h#L131-L132

mfasDa:
@shahor02 This version was taken from the one you mentioned, the difference is in the interface to use iterators instead to simplify the usage with boost histograms. It was implemented by Hannah some time ago and already merged - this commit are fixes found in tests by Joshua comparing the results of this fit to ROOTs fitter - see EMCAL status report from the S&C week yesterday. Can you please merge the PR nevertheless?

shahor02:
@mfasDa will merge, but take into account that the version with the TLinearFitter is unstable and sensitive to outliers, the one I refer to is more robust.

mfasDa:
@shahor02 Thanks for your comment! We will have a look. This is basically in the kernel of the function which should be easy to combine. Eventually one can also unify the kernel and access it via the two interfaces.

PR #7633, opened 2021-11-16
TRD additional configurable params

martenole:
- deflection cut for tracklets disabled by default, needed until proper calibration available
- some more comments in the code

With the changes in https://github.com/AliceO2Group/AliceO2/pull/7629 added as well I find a reasonable number of tracklets which is attached to ITS-TPC and to TPC tracks (more than 10% of the tracklets are attached when I ignore noisy MCMs).
An important change to get a proper matching was to disable the cut on the deflection of the tracklets. After a proper vD and ExB calibration this should be re-enabled and the cut tuned.


tdietel (approved):

bazinski (approved):
There is a comment somewhere where the instruction to LOG must be lower case, the fairmq change was postponed though. i.e. DEBUG to debug

PR #7634, opened 2021-11-16
[EMCAL-735] Switch of trailer merging in RawReaderMemory for STU pages

mfasDa:
Use  FEE ID range (settable) in which the trailer merging is
enabled, and for all other FEE IDs it is disabled, meaning
pages are added as if there was no trailer at the end of the
page. Default range is defined in the constructor to be within
0 and 40 in order to avoid conflicts in the raw QC also using
the RawReaderMemory as page parser.


mfasDa:
This is for the preparation of the decoding of STU pages. STU pages do not contain RCU trailers, consequently RCU trailer merging needs to be disabled for these links. For the moment no trailer merging for STU pages is applied.

shahor02 (approved):

PR #7635, opened 2021-11-16
Class to handle the MID readout card configuration and response

dstocco:
This PR introduces new classes for better handling of the MID card cnfiguration.
The electronics can be configured with masks and with parameters stating if the card should send the output when the X AND Y strips are fired or the X OR Y are fired.
The aim of this PR is to be able to load and write such configurations, as well as to better mimic the electronic behaviour in the simulation/reconstruction code.

### DIFF
--- a/Detectors/MUON/MID/Raw/include/MIDRaw/GBTMapper.h
+++ b/Detectors/MUON/MID/Raw/include/MIDRaw/GBTMapper.h
@@ -1,0 +1,42 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file   MIDRaw/GBTMapper.h
+/// \brief  MID GBT mapping
+/// \author Diego Stocco <Diego.Stocco at cern.ch>
+/// \date   16 November 2021
+#ifndef O2_MID_GBTMAPPER_H
+#define O2_MID_GBTMAPPER_H
+
+#include "DataFormatsMID/ROBoard.h"
+#include "MIDRaw/CrateParameters.h"
+
+namespace o2
+{
+namespace mid
+{
+/// Functions to relate the unique local board ID with the corresponding GBT ID
+namespace gbtmapper
+{
+/// Gets the GBT ID from the unique loc ID
+/// \param uniqueLocId Unique local board ID
+/// \return GBT ID
+inline uint16_t getGBTIdFromUniqueLocId(uint8_t uniqueLocId) { return crateparams::makeGBTUniqueId(raw::getCrateId(uniqueLocId), crateparams::getGBTIdFromBoardInCrate(raw::getLocId(uniqueLocId))); }
+/// Check if the board is in the GBT
+/// \param uniqueLocId Unique local board ID
+/// \param gbtUniqueId Unique GBT ID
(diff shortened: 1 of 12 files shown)


Detectors/MUON/MID/Raw/include/MIDRaw/GBTMapper.h
```diff
@@ -0,0 +1,42 @@
...
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file   MIDRaw/GBTMapper.h
+/// \brief  MID GBT mapping
+/// \author Diego Stocco <Diego.Stocco at cern.ch>
+/// \date   16 December 2021
```
aphecetche: time travelling ? ;-)

Detectors/MUON/MID/Raw/include/MIDRaw/ROBoardConfigHandler.h
```diff
@@ -0,0 +1,71 @@
...
+
+  /// Updates the mask values
+  void updateMasks(const std::vector<ROBoard>& masks);
+
+  /// Writes the configuration to file
+  void write(const char* filename) const;
+
+ private:
+  /// Loads the board  from a configuration file
+  /// The file is in the form:
```
aphecetche, line 56: would be nice to have the description of this file (and its format) also in some `README.md` and not only within the code

Detectors/MUON/MID/Raw/src/CrateMapper.cxx
```diff
@@ -155,35 +160,40 @@ uint8_t CrateMapper::deLocalBoardToRO(uint8_t deId, uint8_t columnId, uint8_t li
...
+      if (gbtUniqueId != 0xFFFF) {
+        if (!gbtmapper::isBoardInGBT(locId, gbtUniqueId)) {
+          continue;
+        }
+      }
+      roBoardIds.emplace_back(locId);
+    }
   }
-  return item->second;
+  std::sort(roBoardIds.begin(), roBoardIds.end());
```
aphecetche, line 195: if the sorting is actually of some importance, I would mention it in the description of the function in the include file as well

Detectors/MUON/MID/Raw/src/ROBoardConfigHandler.cxx
```diff
@@ -0,0 +1,181 @@
...
+      if ((cfgIt->second.statusWord & crateconfig::sXorY) == 0) {
+        cfgIt->second.masksNBP = mask.patternsNBP;
+      }
+    }
+  }
+}
+
+std::vector<ROBoardConfig> makeDefaultROBoardConfig(uint16_t gbtUniqueId)
+{
+  // FIXME: in the current configuration, when an Y strip covers several local boards
```
aphecetche, line 145: is there a JIRA ticket corresponding to this (rather important) FIXME ?

Detectors/MUON/MID/Raw/src/CrateMapper.cxx
dstocco: Well, it is not fundamental, it is just convenient when one needs to write data or print them in a human-friendly way.
I mentioned it in the description

Detectors/MUON/MID/Raw/src/ROBoardConfigHandler.cxx
dstocco: Not sure to which JIRA project this belongs. It is just an internal decision. Also, since the Y signal is physically not sent to some boards, we will have to leave with almost half of the detector in X OR Y. So, for consistency and to avoid biases we might even decide to keep X OR Y by default for all boards.

aphecetche: > Not sure to which JIRA project this belongs

Good point. As it's an MID-wide thing, one option would be to use the run coordination project (R3C ; MID component) ?

aphecetche (approved):

PR #7639, opened 2021-11-17
[EMCAL-647] Event builder for cell container from different subtimefr…

mfasDa:
…ames

Combining cells from multiple subevents in a single
cell / trigger record container according to cells
belonging to the same collsion. Output is sent to
the same output type as input cell / trigger record
container but with subspecification 0 for
compatibility with existing containers.

### DIFF
--- a/Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
@@ -1,0 +1,139 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction
+
+#include <algorithm>
+#include <iterator>
+#include <unordered_map>
+#include <gsl/span>
+
+#include "DataFormatsEMCAL/Cell.h"
+#include "DataFormatsEMCAL/TriggerRecord.h"
+#include "Framework/InputSpec.h"
+#include "Framework/Logger.h"
+#include "EMCALWorkflow/EventBuilderSpec.h"
+
+using namespace o2::emcal;
+
+void EventBuilderSpec::init(framework::InitContext& ctx)
+{
+}
+
+void EventBuilderSpec::run(framework::ProcessingContext& ctx)
+{
+  auto dataorigin = o2::header::gDataOriginEMC;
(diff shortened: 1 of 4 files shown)


mfasDa:
From #6834
> @mfasDa I also don't see reopen button, perhaps @ktf can do this. As for the merging, was it tested? I remember there was a problem of defining FLP-specific subspecs.

Not yet, as we are missing a test setup and for the pilot we had other priorities. 

For what concerns the FLP-specific subspec I have an implementation here: https://github.com/mfasDa/AliceO2/commit/c0b1101a75f0d44413b30653d1d98206032f4196 (#6772 - also stale-closed - reopened in #7642). It is not yet in a PR, it still needs to be tested on a dedicated test FLP.

Detectors/EMCAL/workflow/src/EventBuilderSpec.cxx
```diff
@@ -0,0 +1,139 @@
...
+
+  auto posCells = ctx.inputs().getPos("inputcells"),
+       posTriggerRecords = ctx.inputs().getPos("inputtriggers");
+  auto numSlotsCells = ctx.inputs().getNofParts(posCells),
+       numSlotsTriggerRecords = ctx.inputs().getNofParts(posTriggerRecords);
+  for (decltype(numSlotsCells) islot = 0; islot < numSlotsCells; islot++) {
+    auto celldata = ctx.inputs().getByPos(posCells, islot);
+    auto subspecification = framework::DataRefUtils::getHeader<header::DataHeader*>(celldata)->subSpecification;
+    // Discard message if it is a deadbeaf message (empty timeframe)
+    if (subspecification == 0xDEADBEEF) {
```
shahor02, line 46: see comment https://github.com/AliceO2Group/AliceO2/pull/7642/files#r752583346 
Why not sending with the standard subspec and skip if the payload is absent (which will happen if the raw data had 0xdeadbeef)?

shahor02 (commented):
With so many map operations you may have overheads (I think T0 had these in an almost similar situation). It could be more efficient to 1st to record in the vector<int> the trigger_bc.toLong() - TF_first_IR.toLong() of every trigger, associating with every filled slot vectors of corresponding cells and in triggers, then process them.

ktf:
This will be closed soon. @mfasDa is it still being worked on?

shahor02:
Hi @ktf ,
This PR is waiting for the feature we were discussing: possibility to define wildcarded subspecs which does not match to one specific subspec, to avoid short circuits

mfasDa:
Hi @ktf ,

Yes, this is needed, for a triggered detector we have to merge triggered events within a timeframe from the two subtimeframes manually once we move the reconstruction from the EPN to the FLP. However as @shahor02 explained there are some questions in the handling of input and output channels, i.e. related to the subspecification, and waiting for some features, and therefore the PR is on hold. It would be good if the pull request could be kept open, it was already closed once and reopened manually. In AliRoot we had the feature to make the PR a draft if it was not yet ready for merging, but I don't see the feature in O2, would be nice to add it here as well. The same applies to PR #7642 which is the corresponding on the sender side (FLP) which has to be adapted and merged in sync with this PR on the EPN (receiver side).

PR #7642, opened 2021-11-17
[EMCAL-660, EMCAL-703] Add option to auto-configure subspecification

mfasDa:
In case of auto-configure mode the subspecification
is determined based on the first header in the
payload using the ID of the FLP determined via the
FEE ID. The output specs are defined with wildcard
subspecification according to the discussion in
https://alice-talk.web.cern.ch/t/flp-specific-subspecifications/1035/6.
In case of empty timeframe empty cell containers are
sent to the deadbeaf subspecification which must be
caught by receiver processors.

### DIFF
--- a/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
@@ -67,6 +67,12 @@ void RawToCellConverterSpec::init(framework::InitContext& ctx)
     LOG(error) << "Failed to initialize mapper";
   }
 
+  if (ctx.options().get<bool>("autosubspec")) {
+    mAutoDefineSubspec = true;
+  } else {
+    mSubspecification = ctx.options().get<int>("subspec");
+  }
+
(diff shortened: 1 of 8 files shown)


mfasDa:
Reopening stale PR #6772

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -115,7 +121,7 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
   mOutputDecoderErrors.clear();
 
   if (isLostTimeframe(ctx)) {
-    sendData(ctx, mOutputCells, mOutputTriggerRecords, mOutputDecoderErrors);
+    sendData(ctx, mOutputCells, mOutputTriggerRecords, mOutputDecoderErrors, 0xdeadbeaf);
```
shahor02, line 123: Why do you need to send with 0xdeadbeaf subspec? Since the cell consumers (except the new event-builder) expect subspec 0, with this you will be obliged to always run the event-builder, even if the full processing is done on the EPN and event building is not needed.

PR #7643, opened 2021-11-17
Make raw-reader.sh parse on macOS

ktf:


davidrohr:
I don't fully understand why this should be needed? Is NUMAGPUIDS not set on MacOS? It should be set by setenv.sh. Perhaps `$MYDIR` is not obtained correctly on MacOS? Then we could fix that instead.

ktf:
yes, indeed the issue is MYDIR not being detected correctly (`readlink -f` does not exist). That said, IMHO, it's in any case good practice to follow posix conventions for `if`s:

```
if [ "X$VAR" = X1 ]; then
...
fi
```

it would also a good idea to run with `bash -e`.

davidrohr:
> if [ "X$VAR" = X1 ]; then

I would prefer to avoid those if we know the variable is set, since I find it less readable.

> it would also a good idea to run with `bash -e`.

Makes sense to me

For MacOS: I think it was Laurent who added this code in `dpl-workflow.sh`:
```
command -v zsh > /dev/null 2>&1 && MYDIR=$(dirname $(zsh -c 'echo ${0:A}' "$0"))
test -z ${MYDIR+x} && MYDIR="$(dirname $(readlink -f $0))"
```
Isn't there a simply way to get the path of the current script?
Could we use `greadlink -f` on Mac? It comes with `brew install coreutils`. Is that commonly available?

ktf:
> > if [ "X$VAR" = X1 ]; then
> I would prefer to avoid those if we know the variable is set, since I find it less readable.

Maybe you can get away with just "$VAR"?

> > it would also a good idea to run with `bash -e`.
> 
> Makes sense to me

ok, I will try changing it.

> For MacOS: I think it was Laurent who added this code in `dpl-workflow.sh`:
> Isn't there a simply way to get the path of the current script? Could we use `greadlink -f` on Mac? It comes with `brew install coreutils`. Is that commonly available?

Is greadlink commonly available on linux? By now macOS comes with `realpath`, maybe we could use that.

davidrohr:
> > > if [ "X$VAR" = X1 ]; then
> > > I would prefer to avoid those if we know the variable is set, since I find it less readable.
> 
> Maybe you can get away with just "$VAR"?
> 

For integer variables I don't really see the point. The scripts all anyway require bash, since they use the `[[` and `$((` syntax which is from bash. And bash doesn't require quotes around variables, so I would leave them out.

> > For MacOS: I think it was Laurent who added this code in `dpl-workflow.sh`:
> > Isn't there a simply way to get the path of the current script? Could we use `greadlink -f` on Mac? It comes with `brew install coreutils`. Is that commonly available?
> 
> Is greadlink commonly available on linux? By now macOS comes with `realpath`, maybe we could use that.

No, greadlink is the gnu readlink on MacOS, on linux readlink is what greadlink is on mac but there is no greadlink on Linux... Quite annoying. OK, then we can perhaps also use realpath. Since we have this logic in several places, perhaps you could come up with a nice bash one-liner that works for both mac and Linux :)

ktf:
> For integer variables I don't really see the point. The scripts all anyway require bash, since they use the `[[` and `$((` syntax which is from bash. And bash doesn't require quotes around variables, so I would leave them out.

Apart from the fact that requiring bash might not work in the long term on macOS (they recently moved to zsh by default), they are using `[`, not `[[`, hence the problem. If you prefer to use `[[` we could also do that, I thought using `[` was deliberate to support non-bash shells.

davidrohr:
ok, actually raw-reader uses `[` apparently, bust most other full system test scripts use `[[`.
And basically, I have little understanding if MacOS decides to drop bash support. Probably even then you could still get it from homebrew.

ktf:
Ok, I will change things to use `[[` and realpath.
Maybe they will not drop it, but for sure they will never move to bash 4, due to GPLv3. I guess indeed that we can solve via homebrew.

ktf:
Will reopen with the agreed fix.

PR #7644, opened 2021-11-17
[EMCAL-732] implementation of EMCAL standalone AOD producer workflow

fjonasALICE:


Detectors/EMCAL/workflow/include/EMCALWorkflow/StandaloneAODProducerSpec.h
```diff
@@ -0,0 +1,60 @@
...
+ private:
+  int64_t mTFNumber = -1;                                                // Timeframe ID
+  int mRunNumber = -1;                                                   // Run number
+  uint32_t mCaloAmp = 0xFFFFFF00;                                        // 15 bits
+  uint32_t mCaloTime = 0xFFFFFF00;                                       // 15 bits
+  o2::emcal::EventHandler<o2::emcal::Cell>* mCaloEventHandler = nullptr; ///< Pointer to the event builder for emcal cells
+  TStopwatch mTimer;
+
+  static const char* BININGCELLS;
+  static const char* BINDINGCELLSTRG;
```
mfasDa, line 51: No longer needed.

mfasDa (changes requested):
Only a small change from my side requested

Detectors/EMCAL/workflow/src/StandaloneAODProducerSpec.cxx
```diff
@@ -0,0 +1,173 @@
...
+
+    LOG(INFO) << "Found " << cellsInEvent.size() << " cells in event";
+
+    auto bcID = interactionRecord.toLong();
+    bcMap.emplace_back(bcID);
+    for (auto& cell : cellsInEvent) {
+
+      // fill table
+      caloCellsCursor(0,
+                      bcID,
```
mfasDa: This is not going to work. We need the index in the BC table, not the global BCID. Since the BCs are sorted in the trigger record table it is sufficient to add counter and fill the BC table for the same event right after the cell and calo table with the index the same counter value.

Detectors/EMCAL/workflow/src/StandaloneAODProducerSpec.cxx
```diff
@@ -0,0 +1,173 @@
...
+                      cell.getTower(),
+                      o2::math_utils::detail::truncateFloatFraction(cell.getAmplitude(), mCaloAmp),
+                      o2::math_utils::detail::truncateFloatFraction(cell.getTimeStamp(), mCaloTime),
+                      cell.getType(),
+                      1); // hard coded for emcal (-1 would be undefined, 0 phos)
+    }                     // end of cell loop
+
+    // filled only with BCID, rest dummy for no2
+    caloCellsTRGTableCursor(0,
+                            bcID,
```
mfasDa: Needs same BC counter as mentioned before.

Detectors/EMCAL/workflow/src/StandaloneAODProducerSpec.cxx
```diff
@@ -0,0 +1,173 @@
...
+                            0,  // fastOrAbsId (dummy value)
+                            0., // lnAmplitude (dummy value)
+                            0,  // triggerBits (dummy value)
+                            1); // caloType (dummy value)
+
+    // fill collision cursor
+    collisionsCursor(0,
+                  bcID,
```
mfasDa: Once more

Detectors/EMCAL/workflow/src/StandaloneAODProducerSpec.cxx
```diff
@@ -0,0 +1,173 @@
...
+  // std::cout << "Finished cell loop" << std::endl;
+
+  std::sort(bcMap.begin(), bcMap.end(), std::less<int64_t>());
+  uint64_t triggerMask = 1;
+  for (auto& bcID : bcMap) {
+    bcCursor(0,
+             runNumber,
+             bcID,
+             triggerMask);
+  }
```
mfasDa: This should go directly in the event loop. Instead of filling 0 in the index column we use the counter value used in the other tables.

mfasDa (changes requested):
The BC mapping does not work, when testing I get negative values in the index column and consequently a crash once I try to access the BC. This must be changed using a counter for the BC ID that is used for all 4 tables. The BC table in this case can be filled inside the event loop.

Detectors/EMCAL/workflow/src/StandaloneAODProducerSpec.cxx
```diff
@@ -0,0 +1,173 @@
...
+  std::vector<int64_t> bcMap;
+  // loop over events
+  for (int iev = 0; iev < mCaloEventHandler->getNumberOfEvents(); iev++) {
+    o2::emcal::EventData inputEvent = mCaloEventHandler->buildEvent(iev);
+    auto cellsInEvent = inputEvent.mCells;                  // get cells belonging to current event
+    auto interactionRecord = inputEvent.mInteractionRecord; // get interaction records belonging to current event
+
+    LOG(INFO) << "Found " << cellsInEvent.size() << " cells in event";
+
+    auto bcID = interactionRecord.toLong();
```
jgrosseo, line 90: This is wrong. You have to fill into the tables the tree entry of the BC entry in the BC tree. You get this with `bcCursor.lastIndex()` after filling the respective BC entry.
So you need a loop over BCs, you cannot fill first one, then the other (unless you know which ones you fill and you do *not* reorder as you do below)

mfasDa: `bcCursor` is a lambda function, `lastIndex()` doesn't exist. Can't we just take a running index (i.e. iev) that we increment per event/bc in the loop?

mfasDa (dismissed):
Tested and working, should be fine now. @jgrosseo Could you please merge once the CI finishes?

shahor02:
sorry I've missed something, but what is the point of the standalone producer? Isn't it better to change the usual one is such a way that it supports EMCal only?

mfasDa:
@shahor02 I was actually checking with you in a private Mattermost conversation whether there is any chance to have the AODProducer supporting also stand-alone runs. Currently the global producer will not work if there is not at least FIT and a primary vertex reconstructed, which we don't have in the stand-alone run. Definitely we want AODs from the stand-alone runs we took in the pilot beam, we would like to test the analysis chain using the analysis framework, we want to repeat the pi0 analysis using the run3 analysis framework as validation of the analysis task. As you see the conversion from CTF to AOD for triggered detectors in the stand alone case is super simple, because we just need to port the cell container and link with the BC table, where the BC we provide ourselves. Adding this option to the global AODProducer will need quite sizeable modification. I am not generally against adding a stand alone mode to the global AOD producer, I also prefer common code instead of many code duplications, and maybe other detectors have similar interests as well. But for such a simple purpose I am wondering whether a stand-alone workflow is the more efficient way to go.

shahor02:
I've just rechecked our mm conversation, there I was also suggested to make the standard aodproducer to work with caloriemeters only. it should not be a big deal, I think one should simply foresee the possibility to run w/o primary vertex and use triggered detectors triggers to create globalBC maps.
This standalone code is simple, but if the something is changing in the AOD, you will need to change it also here...
Anyway, one can have this for the moment, but I would suggest eventually switching to the standard one. Pinging to @nburmaso .

mfasDa:
@shahor02 Thanks for your comment! If the option would be foreseen in the standard AOD producer that would of course be better. Still it does not hurt to merge this one, because then we have something in the mean time to work with and commission the EMCAL analysis framework. Once the global producer is ready to support stand-alone runs of triggered detectors we will move to that and delete this one again.

Detectors/EMCAL/workflow/src/StandaloneAODProducerSpec.cxx
```diff
@@ -0,0 +1,171 @@
...
+void StandaloneAODProducerSpec::run(ProcessingContext& pc)
+{
+  auto cput = mTimer.CpuTime();
+  mTimer.Start(false);
+
+  const auto* dh = o2::header::get<o2::header::DataHeader*>(pc.inputs().getFirstValid(true).header);
+  uint64_t tfNumber;
+  if (mTFNumber == -1L) {
+    // TODO has to be made globally unique (by using absolute time of TF). For now is unique within the run
+    tfNumber = dh->tfCounter; // getTFNumber(startIR, runNumber);
```
shahor02: Note that this part was changed in the standard AOD producer: https://github.com/AliceO2Group/AliceO2/blob/c394a041e0f17a0fd6971b8373597d2e407a2e13/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L1101

I'll fix this and merge

shahor02 (approved):
all tests were passed, merging after applying a trivial fix

Detectors/EMCAL/workflow/src/StandaloneAODProducerSpec.cxx
mfasDa: Thanks a lot for fixing!

mfasDa:
@shahor02 Thanks a lot for merging! We promise we will migrate to the standard AOD producer once it supports stand alone runs and remove this one again. We will however not mess around with changing other tables there, this should be done by the experts, too much damage could be done for other detectors if something goes wrong in the implementation.

PR #7647, opened 2021-11-17
Method to refit vertex with partial set of its tracks

shahor02:


shahor02:
@mfaggin : example of refitting the vertex with some tracks suppressed is in the attached `vtxRefit.C`.
This is just a simple exercise with the vertex made of ITS tracks only, so I could use the tracks from the ITS reco file.
I assume that in general, you will be refitting the vertices from the AOD. In this case you will need to prepare a vector of TrackParCov tracks (or any class derived from it) selecting the tracks contributing to vertex (needs https://github.com/AliceO2Group/AliceO2/pull/7654 + implementing filling this info) and pass it to the vertexer as in this example.
Note that only position is refitted, the time info in ignored.

[tstVtxRefit.tar.gz](https://github.com/AliceO2Group/AliceO2/files/7564534/tstVtxRefit.tar.gz)

PR #7648, opened 2021-11-18
DPL: avoid confusing stacktrace when workflow is empty

ktf:


ktf:
Tested locally. Not really tested in CI. Full CI and macOS compile and run. Merging.

PR #7660, opened 2021-11-19
DPL: speedup compilation

ktf:


ktf:
@Barthelemy I fixed the issues  here in today build  of QualityControl. Any objections switching to it (i.e. updating alidist/master with it)?

PR #7662, opened 2021-11-19
fix for ccdb in TOF workflow

noferini:
Two fixes:

- port of the production ccdb fixed (otherwise pointing to test ccdb)

- remove request of ccdb condition input when asking for ccdb (provided by the api)
@chiarazampolli with previous version --use-ccdb is not working because of that (please not use it)

@davidrohr it is not clear to me if I need to specify the ccdb or inherit it from global conf (ccdb retrived with CcdbManager::instance() in our api and then setting url)

### DIFF
--- a/Detectors/GlobalTrackingWorkflow/tofworkflow/src/tof-reco-workflow.cxx
+++ b/Detectors/GlobalTrackingWorkflow/tofworkflow/src/tof-reco-workflow.cxx
@@ -52,6 +52,7 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
     {"tof-sectors", o2::framework::VariantType::String, "0-17", {"TOF sector range, e.g. 5-7,8,9 ,TBI"}},
     {"tof-lanes", o2::framework::VariantType::Int, 1, {"number of parallel lanes up to the matcher, TBI"}},
     {"use-ccdb", o2::framework::VariantType::Bool, false, {"enable access to ccdb tof calibration objects"}},
+    {"ccdb-url", o2::framework::VariantType::String, "http://ccdb-test.cern.ch:8080", {"CCDB Url"}},
     {"input-desc", o2::framework::VariantType::String, "CRAWDATA", {"Input specs description string"}},
     {"disable-root-input", o2::framework::VariantType::Bool, false, {"disable root-files input readers"}},
     {"disable-root-output", o2::framework::VariantType::Bool, false, {"disable root-files output writers"}},
@@ -137,6 +138,7 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   auto isCalibFromCluster = cfgc.options().get<bool>("calib-cluster");
   auto isCosmics = cfgc.options().get<bool>("cosmics");
   auto ignoreDistStf = cfgc.options().get<bool>("ignore-dist-stf");
+  auto ccdb_url = cfgc.options().get<std::string>("ccdb-url");
 
   LOG(INFO) << "TOF RECO WORKFLOW configuration";
   LOG(INFO) << "TOF input = " << cfgc.options().get<std::string>("input-type");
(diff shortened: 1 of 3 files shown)


noferini:
@davidrohr @chiarazampolli 
If you all agree I would remove
calibapi.setURL("http://alice-ccdb.cern.ch");
in the spec and using what provided by the common framework (default is test ccdb)
That call modifies the common instance and potentially affecting also others

noferini:
In the meanwhile I added the flag
--ccdb-url
to change the url

chiarazampolli:
Hi, 

I don't think the errors are related to the PR, even if one (o2) is from a CCDB test. 

Chiara

shahor02:
Hi @noferini 

> @davidrohr @chiarazampolli If you all agree I would remove calibapi.setURL("http://alice-ccdb.cern.ch"); in the spec and using what provided by the common framework (default is test ccdb) That call modifies the common instance and potentially affecting also others

Which others do you mean? It is set in the run() and affects only the BasicManager instance of this process.

The failures are unrelated but the fullCI failed to start. Did you run reconstruction locally, to make sure that the change of the ccdb server default does not pose problems?

chiarazampolli:
Hi, 

I think that Francesco refers to the others using CcdbManager. For TOF, that is what is used:
https://github.com/AliceO2Group/AliceO2/blob/ef59be8ddb35a36fede5025a73f009e8d8034380/Detectors/TOF/calibration/include/TOFCalibration/CalibTOFapi.h#L48
which has the CcdbManager inside.

If I check the logs of the async reco, https://alimonitor.cern.ch/users/download.jsp?view=true&path=/alice/cern.ch/user/a/alidaq/PilotBeam/grid/finalInputs/updates/forCentralProd/output_2021119_100TF_TPCcalib/505673/001/log_20211119-163116_1.log for example, I see:
```
[INFO]  ccdbpath http://alice-ccdb.cern.ch
```
```
[830:ft0-recpoint-writer]: [INFO]  ccdbpath http://alice-ccdb.cern.ch
[830:ft0-recpoint-writer]: [INFO] WorkflowSpec getRecoWorkflow useMC 0 CCDB  http://alice-ccdb.cern.ch
```
```
[837:MIDTracker]: [16:31:56][INFO] Alignment from http://ccdb-test.cern.ch:8080 for timestamp 1637335907858: , empty object for [ITS,TPC,TRD,TOF,PHS,CPV,EMC,HMP,MFT,MCH,MID,ZDC,FT0,FV0,FDD]
```
(and many more for the alignment, of course)
```
[825:fv0-reconstructor]: Info in <TJAlienFile::Open>: Accessing file /alice/data/CCDB/FV0/Calibration/ChannelTimeOffset/11/42337/1ae55a20-2900-11ec-ab62-2a010e0a09fb in SE <ALICE::CERN::EOS>
[824:ft0-reconstructor]: Info in <TJAlienFile::Open>: Accessing file /alice/data/CCDB/FT0/Calibration/ChannelTimeOffset/02/39742/c6304da0-393f-11ec-b299-200114580202 in SE <ALICE::CERN::EOS>
```
So, apart from the fact that alignment should be taken from production CCDB (no?), it seems to use the CcdbManager too:
https://github.com/AliceO2Group/AliceO2/blob/8feb5833b484c757386edc39ebf4b58e6783e97f/Detectors/Base/src/Aligner.cxx#L43
FT0 and FV0 use as well the CcdbManager (https://github.com/AliceO2Group/AliceO2/blob/ef59be8ddb35a36fede5025a73f009e8d8034380/Detectors/FIT/FT0/workflow/src/ReconstructionSpec.cxx#L40, https://github.com/AliceO2Group/AliceO2/blob/ef59be8ddb35a36fede5025a73f009e8d8034380/Detectors/FIT/FV0/workflow/src/ReconstructionSpec.cxx#L40)
I don't know if there is a lock or something, but if all alignment, FT0 and FV0 use it, and they seem to set different paths, it seems that the CcdbManager is not shared between processes, so whatever TOF does, will not interfere. Right?

But in general, we need to clean up how CCDB is accessed...

noferini:
I tested reconstruction (ITS+TPC+TOF) on EPN but I don't think other dpl were using ccdb.

chiarazampolli:
Ciao,

Yes, as I wrote, everyone accessing alignment + FT0 and FV0 do. But there is no interference among them, so I guess it is fine for TOF too.

Chiara

noferini:
In any case currently we can work with both (test and production) since I loaded the same calibration object in both

chiarazampolli:
I would put the production CCDB.

shahor02:
> 

these are separated processes, their statics, singletons etc cannot interfere.

chiarazampolli:
Hi, 
Yes, exactly. 
Since we answered Francesco's doubts, can this PR be merged, or are there still pending points? The errors seem unrelated.
Chiara

PR #7663, opened 2021-11-19
AOD-tracks: propagate PVContributor flag

sawenzel:


jgrosseo:
I am sorry, I made a mistake in the other PR. This should of course be checked on the `Flags` (as says the name of the enum) and not on the `Type`. I have moved the dynamic column in the right place here:
https://github.com/AliceO2Group/AliceO2/pull/7665

sawenzel:
Ok. Then the code becomes even easier and requires less changes.

jgrosseo (approved):

PR #7670, opened 2021-11-19
DPL GUI: send up to 10 updates per second for the data relayer GUI

ktf:


ktf:
@davidrohr

davidrohr:
Error unrelated, merging

PR #7672, opened 2021-11-19
TPC: Add simple raw data filter for krypton clusters

wiechula:


wiechula:
@davidrohr , @shahor02 , the error looks unrelated. Can this be merged?

Detectors/TPC/workflow/src/FileWriterSpec.cxx
```diff
@@ -47,13 +47,40 @@ using o2::dataformats::FileMetaData;
...
+};
+
+const std::unordered_map<BranchType, std::string> BranchName{
+  {BranchType::Krypton, "TPCBoxCluster"},
+  {BranchType::Digits, "TPCDigit"},
+};
+
+const std::unordered_map<BranchType, std::string> TreeName{
+  {BranchType::Krypton, "Clusters"},
+  {BranchType::Digits, "o2sim"},
```
shahor02, line 67: @wiechula there are extra `,` at the end of each block. The compiler does not complain but I think it is better to remove them.

wiechula: @shahor02 as far as I know it is allowed since C99. I would like to keep it, sine it looks more symmetric and makes adding new elements a bit easier.
I think it is also used in other places in o2.

shahor02: ok, thought looks a bit awkward...

PR #7674, opened 2021-11-20
Subevent merger implemented

peressounko:
Impement EventBuilder which takes 2 messages from PHOS FLPs and combines sub-events with same BC

### DIFF
--- a/Detectors/PHOS/base/include/PHOSBase/Mapping.h
+++ b/Detectors/PHOS/base/include/PHOSBase/Mapping.h
@@ -53,7 +53,7 @@ class Mapping
 
   ~Mapping() = default;
 
-  //Getters for unique instance of Mapping
+  // Getters for unique instance of Mapping
   static Mapping* Instance();
   static Mapping* Instance(std::basic_string_view<char> path);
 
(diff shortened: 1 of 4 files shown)


peressounko:
Dear Ruben @shahor02, 
I implemented event builder as we discussed with Markus @mfasDa and mostly copied his code. But when I try to run as 
```o2-raw-file-reader-workflow --input-conf rawreader.cfg | o2-phos-reco-workflow --input-type raw --output-type cells --disable-root-output --disable-mc --flpId 1| o2-phos-event-builder-workflow ``` it fails:  PHOSEventBuilder depends on itself
It looks like wildcards accept output as well. Do I miss something? @mfasDa did you manage to run your code?

shahor02:
Hi @peressounko @mfasDa 

Yes, it will not work, also for EMCAL: device cannot have input matching to its own output. We discussed this once with Giulio, he suggested to have a version of ConcreteDataTypeMatcher which allows to exclude specific subspec, e.g. 0. 
Pinging to @ktf to ask if he can add such a feature.

ktf:
Let's follow up it here: https://alice.its.cern.ch/jira/browse/O2-2699. In
principle it should be already possible using DataDescriptorMatcher,
however I guess I lack at least an example, if not an helper.

Ciao,
Giulio

PR #7676, opened 2021-11-20
change --ccdb-url to --ccdb-url-tof, no LOG(info) from workflow definition

shahor02:
The same option cannot be local and global at the same time, even in different devices.

Merging as currently the FST is brocken


chiarazampolli:
Hi Ruben, 

The conflict was between the tof-reco-workflow and the e.g. FIT ones? Or which ones?

Chiara

shahor02:
@chiarazampolli CPV and ZDC have it as a device-level option.

chiarazampolli:
Ok, thanks for clarifying!

Chiara

noferini:
Thanks for checking this

PR #7678, opened 2021-11-21
Restore --hbfutils-config as global and add --tf-info-file local option to readers

shahor02:
To make sure that if HBFUtils should be updated from ini file, this is done for every device.


shahor02:
failures unrelated, FST checked locally.

PR #7610, opened 2021-11-14
add fill scheme info in TOF event time checker

noferini:

PR #7679, opened 2021-11-21
fixes in TOF calibration workflow

noferini:


noferini:
WIP

noferini:
@shahor02 -> this is now ready for the review (many fixes in TOF calibration workflow after running several tests on Beam Test data). Two comments:
- only ITS-TPC tracks are used now (TPC only tracks have wrong track length to be fixed before to use in calibration)
- added calibInfo clear when changing TF (this was missing, bug, and leading to a calibInfo size 50 times larger with CTF containing 100 TFs + some problems in calibration)

@njacazio -> once this PR will be merger you can open yours in O2Physics (adding EventTimeMaker moved from TOFReconstruction to TOFBase)

DataFormats/Detectors/TOF/src/CalibTimeSlewingParamTOF.cxx
```diff
@@ -168,3 +189,14 @@ CalibTimeSlewingParamTOF& CalibTimeSlewingParamTOF::operator+=(const CalibTimeSl
...
 }
+//______________________________________________
+CalibTimeSlewingParamTOF::CalibTimeSlewingParamTOF(const CalibTimeSlewingParamTOF& source)
+{
+  bind();
+  for (int i = 0; i < NSECTORS; i++) {
+    *(mTimeSlewing[i]) = *(source.mTimeSlewing[i]);
+    (*(mChannelStart[i])) = (*(source.mChannelStart[i]));
+    *(mFractionUnderPeak[i]) = *(source.mFractionUnderPeak[i]);
+    *(mSigmaPeak[i]) = *(source.mSigmaPeak[i]);
```
shahor02, line 201: don't you need to copy also mGlobalOffset?

shahor02 (commented):
Hi @noferini 

see one comment bellow. BTW, what about this question: https://github.com/AliceO2Group/AliceO2/pull/7610#pullrequestreview-808291976 ?

DataFormats/Detectors/TOF/src/CalibTimeSlewingParamTOF.cxx
noferini: You are right. I probably missed it when copy and paste (and workflow didn't complain also in execution).

About #7610
I implemented an autodetection for fill scheme in case it is not provided.
It will be automatically skipped once the fill scheme will be filled with the information of collision context (at the moment I didn't know how to retrieve that info from there).

noferini: fixed

shahor02 (approved):

PR #7680, opened 2021-11-22
CcdbApi - Add handling multiple addresses

mucharafal:
Add hosts pool.

To use hosts pool, string parameter with address should has hosts addresses separated with commas or semicolons.

It is initial implementation, it must be discussed and improved (but preferably in next PRs).

Methods, which behaviour should be discussed:
- delete - for now it iterates over all addresses and delete object everywhere
- truncate - like above
- list - take first response. As I discussed with Costin, it should rather ask all hosts and merge response. Is it correct, or current semantics is fine?

For others methods looping over addresses in case error (currently `404`, change to above `400`?) seems reasonable for me. But please, review it deeply, I am not C++ expert ;).

If you prefer, I can split changes into smaller PRs - just let me know.

### DIFF
--- a/CCDB/test/testCcdbApiMultipleUrls.cxx
+++ b/CCDB/test/testCcdbApiMultipleUrls.cxx
@@ -1,0 +1,98 @@
+// Copyright 2019-2021 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#define BOOST_TEST_MODULE CCDB
+#define BOOST_TEST_MAIN
+#define BOOST_TEST_DYN_LINK
+
+#include "CCDB/CcdbApi.h"
+#include "CCDB/CCDBTimeStampUtils.h"
+#include <boost/test/unit_test.hpp>
+#include <cstdio>
+#include <TH1F.h>
+
+using namespace std;
+using namespace o2::ccdb;
+namespace utf = boost::unit_test;
+namespace tt = boost::test_tools;
+
+static string ccdbUrl;
+static string basePath;
+bool hostReachable = false;
+
+/**
+ * Global fixture, ie general setup and teardown
+ */
+struct Fixture {
+  Fixture()
+  {
+    CcdbApi api;
+    ccdbUrl = "https://localhost:22,https://localhost:8080,http://ccdb-test.cern.ch:8080";
+    api.init(ccdbUrl);
+    cout << "ccdb url: " << ccdbUrl << endl;
+    hostReachable = api.isHostReachable();
+    cout << "Is host reachable ? --> " << hostReachable << endl;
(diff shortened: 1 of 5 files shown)


costing:
Hi @mucharafal ,

Thank you for the PR. I think this addresses the point of falling back between endpoints/specific storages.

Yes, everything above `400` should be considered an error and the fallback should apply.

As for the listing (`/browse/` or `/latest/`), this is a bit trickier since ideally it would reflect what the fallback would do. Maybe it's as simple as iterating the endpoints in reverse order? And for `/latest/` overriding the existing objects in the same path, while for `/browse/` ... append to the list? This can be left for a future PR though.

Cheers,

.costin

costing (dismissed):

ktf (changes requested):
It looks like the code checker is not happy?

```
/sw/SOURCES/O2/7680-slc8_x86-64/0/CCDB/src/CcdbApi.cxx:41:10: error: inclusion of deprecated C++ header 'stdio.h'; consider using 'cstdio' instead [modernize-deprecated-headers]
/sw/SOURCES/O2/7680-slc8_x86-64/0/CCDB/src/CcdbApi.cxx:486:42: error: statement should be inside braces [readability-braces-around-statements]
/sw/SOURCES/O2/7680-slc8_x86-64/0/CCDB/src/CcdbApi.cxx:488:15: error: statement should be inside braces [readability-braces-around-statements]
/sw/SOURCES/O2/7680-slc8_x86-64/0/CCDB/src/CcdbApi.cxx:1297:19: error: use nullptr [modernize-use-nullptr]
/sw/SOURCES/O2/7680-slc8_x86-64/0/CCDB/src/CcdbApi.cxx:1299:20: error: use nullptr [modernize-use-nullptr]
/sw/SOURCES/O2/7680-slc8_x86-64/0//CCDB/test/testCcdbApiMultipleUrls.cxx:1:1: error: missing or malformed copyright notice
```

could you please fix them?

CCDB/test/testCcdbApiMultipleUrls.cxx
```diff
@@ -0,0 +1,126 @@
...
+///
+/// \file   testCcdbApiMultipleUrls.cxx
+/// \author Rafał Mucha
+///
+
+#define BOOST_TEST_MODULE CCDB
+#define BOOST_TEST_MAIN
+#define BOOST_TEST_DYN_LINK
+
+#include "CCDB/CcdbApi.h"
```
Barthelemy, line 16: That's a lot of includes. Are they all necessary ?

Barthelemy (dismissed):

CCDB/test/testCcdbApiMultipleUrls.cxx
mucharafal: No, you are right. I have left necessary

costing:
@ktf could this be merged?

ktf (dismissed):

ktf:
I think it required an  approval before the tests could start, because @mucharafal is a first time contributor. Let's wait for the tests and then we can merge.

mucharafal:
> I think it required an approval before the tests could start, because @mucharafal is a first time contributor. Let's wait for the tests and then we can merge.

@ktf When the test will be started? What is the algorithm?

ktf:
Some of them already started. It should update things shortly.

sawenzel (approved):
triggering build

sawenzel:
Merging. I would still be good to put (doxygen) documentation on the new functions introduced as well as to enhance the main markdown README with the features.

PR #7681, opened 2021-11-22
DPL: make dispatcher lossy

ktf:
This makes the QC dispatcher drop data if the receiving side is not
ready to process them, avoiding building backpressure. Notice
that this also depends on the length of the message queues in FairMQ,
which by default is 4 (which could be already too much) and which
can be user customized.

### DIFF
--- a/Framework/Core/src/SendingPolicy.cxx
+++ b/Framework/Core/src/SendingPolicy.cxx
@@ -24,7 +24,7 @@ std::vector<SendingPolicy> SendingPolicy::createDefaultPolicies()
   return {SendingPolicy{
             .name = "dispatcher",
             .matcher = [](DeviceSpec const& spec, ConfigContext const& ctx) { return spec.name == "Dispatcher" || DeviceSpecHelpers::hasLabel(spec, "Dispatcher"); },
-            .send = [](FairMQDevice& device, FairMQParts& parts, std::string const& channel) { device.Send(parts, channel, 0, 0); }},
+            .send = [](FairMQDevice& device, FairMQParts& parts, std::string const& channel) { device.Send(parts, channel, 0, -1); }},
           SendingPolicy{
             .name = "default",
             .matcher = [](DeviceSpec const& spec, ConfigContext const& ctx) { return true; },
--- a/Framework/Core/src/SendingPolicy.cxx
+++ b/Framework/Core/src/SendingPolicy.cxx
@@ -24,7 +24,7 @@ std::vector<SendingPolicy> SendingPolicy::createDefaultPolicies()
   return {SendingPolicy{
             .name = "dispatcher",
             .matcher = [](DeviceSpec const& spec, ConfigContext const& ctx) { return spec.name == "Dispatcher" || DeviceSpecHelpers::hasLabel(spec, "Dispatcher"); },
-            .send = [](FairMQDevice& device, FairMQParts& parts, std::string const& channel) { device.Send(parts, channel, 0, 0); }},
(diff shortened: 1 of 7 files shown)


ktf:
@knopers8 @Barthelemy, in principle this is all it would be needed now...

Framework/Core/src/SendingPolicy.cxx
```diff
@@ -20,8 +21,12 @@ namespace o2::framework
 std::vector<SendingPolicy> SendingPolicy::createDefaultPolicies()
 {
   return {SendingPolicy{
-    .name = "default",
-    .matcher = [](DeviceSpec const& spec, ConfigContext const& ctx) { return true; },
-    .send = [](FairMQDevice& device, FairMQParts& parts, std::string const& channel) { device.Send(parts, channel, 0); }}};
+            .name = "dispatcher",
+            .matcher = [](DeviceSpec const& spec, ConfigContext const& ctx) { return spec.name == "Dispatcher"; },
```
knopers8: Instead of this, could you please check if a label "Dispatcher" exists? In the future we might have to add some hashes in the name to allow for multiple Dispatchers.

knopers8 (changes requested):
Thanks, this should be indeed all that we need. Could you consider applying my one comment?

Framework/Core/src/SendingPolicy.cxx
ktf: I guess something like this?

ktf:
@knopers8 @Barthelemy the batch_test seems to be failing consistently. Could it be misbehaving because of the new lossy setup?

Framework/Core/src/SendingPolicy.cxx
knopers8: Yes, thank you

knopers8:
Yes, it appears that the new mechanism is quite lossy indeed. You can see:
```
+ entries=157
+ '[' 157 -ne 200 ']'
```
Now that we allow for losses, this line should be changed, but 25% loss on a CI machine is actually quite worrying. What is the fmq buffer size currently?

ktf:
I think the default queue is something like 4 elements, to avoid being able to explode memory. We could either make it longer via a ChannelConfigurationPolicy or try with 1ms timeout.

knopers8:
Indeed that helps (198/200 sampled received), however I think that in this case the mechanism will not actually allow to unblock dataflow. If message rate is higher than 1000/s, then it will eventually fill the queues.

Is it possible to set queue sizes per device? Maybe we could risk increasing it e.g. to 16 and keeping 0ms timeout as a hopefully good balance.

ktf:
Yes, agreed that this only works if the rate is < 1000Hz.

You can customise the size of the queue for a given channel using the ChannelConfigurationPolicy, IIRC. There is something like sndBufSize which you can override...

ktf:
Ok, I am changing this to have a separate policy for the dispatcher, but behaving as before. Enabling it will be a matter of turning a -1 in a 0.

ktf:
@TimoWilken I see there is now an aliDoctor error every now and then. What is that about?

knopers8:
> Ok, I am changing this to have a separate policy for the dispatcher, but behaving as before. Enabling it will be a matter of turning a -1 in a 0.

Sounds good, thanks

TimoWilken:
> @TimoWilken I see there is now an aliDoctor error every now and then. What is that about?

@ktf I think aliDoctor is timing out on hosts where multiple CI builders are running, due to the disk activity from other builders blocking the cleanup steps for the many Docker containers created by aliDoctor. https://github.com/alisw/alibuild/pull/728 should fix this.

PR #7683, opened 2021-11-22
Removing interaction time for track times

njacazio:


njacazio:
@noferini as we discussed yesterday I open the PR to subtract the interaction time to the trackTime

njacazio:
Btw, probably the `getMass2Z` returns a float maybe I could avoid using double for the mass

noferini (approved):
the one failing check not related to this PR

PR #7685, opened 2021-11-22
MCH: add first version of digit filtering

aphecetche:
For the moment we only filter out the digits that end up with an ADC=0
(to protect the clustering). This is done by using the
`--min-adc-value` option (with default value=1).

Note that the addition of the digit filtering forces us to add a few
`--input-digits-xxx-data-description` options to other devices to be
able to select the digit (and digit rofs) data descriptions.

We also add a `--sanity-check` debug option to ensure some coherence
between digits and digit rofs, primarily to cross-check that no
out-of-bound indexing occur.

### DIFF
--- a/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/Digit.h
+++ b/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/Digit.h
@@ -36,6 +36,7 @@ class Digit
   ~Digit() = default;
 
   bool operator==(const Digit&) const;
+  bool operator<(const Digit&) const;
 
   // time in bunch crossing units, relative to the beginning of the TimeFrame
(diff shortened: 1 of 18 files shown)


aphecetche:
I assume the o2 failing test is unrelated to this PR, right ?

shahor02 (approved):
The failed CI is unrelated and traditional. As a general comment: why don't you filter out 0-ADC digits at the decoding level (I understand that this PR goes beyond that, allowing arbitrary filter).

aphecetche:
@shahor02 the option of filtering those digits at the decoding level was indeed considered but rejected as anyway we'll need sooner or later a more complex filtering (which needs e.g. access to some info from CCDB) and that was considered too far from the decoding responsibility

PR #7687, opened 2021-11-23
DPL: introduce Lifetime::Dangling

ktf:


ktf:
Actual usage to prune unneeded messages to the dummy sink coming.

davidrohr:
That could be a genuine error: `64/98 Test  #78: test_Framework_test_WorkflowHelpers ...................***Failed    0.10 sec`

davidrohr:
Just had a quick test of this locally. What I see in the FST is that all processers now send something to the internal-dpl-injected-dummy-sink. So the logic to identify the sinks seems wrong.
Also, many processors die with
```
[28276:ft0-reconstructor]: [11:45:53][ERROR] requested channel has not been configured? check channel names/configuration.
[28276:ft0-reconstructor]: [11:45:53][ERROR] channel: from_ft0-entropy-encoder_to_ft0-reconstructor, index: 0
[28276:ft0-reconstructor]: [11:45:53][ERROR] out of range: _Map_base::at
[28276:ft0-reconstructor]: [11:45:53][ERROR] out of range: _Map_base::at
[28276:ft0-reconstructor]: [11:45:53][ERROR] incorrect/incomplete channel configuration?
[28276:ft0-reconstructor]: [11:45:53][ERROR] error while setting up workflow in o2-ft0-reco-workflow: _Map_base::at
```

ktf:
I realised I actually need also the rest of the feature and that I cannot do things incrementally. I am on it.

ktf:
Ok, I fixed the test and I think also the logic change wrt what was done before. Now verifying something more complex.

ktf:
For the record, the automatic refactoring bot seems to work nicely.

ktf:
Works on hyperloop.

davidrohr:
Doesn't look that good in the full system test with QC. Still some processors are crashing at startup, and the dummy sink still receives a message from basically every processor.
dplgui

ktf:
Ok, i understand the problem. We want to mark things as dangling only at the end of the workflow merging, not on a per workflow basis... I am thinking how to address that...

davidrohr (dismissed):
seems ok in the full system test.

PR #7714, opened 2021-11-23
WP3: apply clang-tidy --checks=modernize-return-braced-init-list --fix

aphecetche:
as agreed in WP3 meeting, here are the fixes for the `modernize-return-braced-init-list` warnings from `clang-tidy`. This PR is to see if nothing breaks.

### DIFF
--- a/Detectors/MUON/MCH/Raw/ElecMap/src/FeeLinkId.cxx
+++ b/Detectors/MUON/MCH/Raw/ElecMap/src/FeeLinkId.cxx
@@ -31,7 +31,7 @@ FeeLinkId decodeFeeLinkId(uint32_t x)
 {
   uint16_t feeId = static_cast<uint16_t>((x & 0xFFFF0000) >> 16);
   uint8_t linkId = static_cast<uint8_t>(x & 0xF);
-  return FeeLinkId(feeId, linkId);
+  return {feeId, linkId};
 }
 
 std::ostream& operator<<(std::ostream& os, const FeeLinkId& id)
--- a/Common/Utils/src/ConfigurableParamHelper.cxx
+++ b/Common/Utils/src/ConfigurableParamHelper.cxx
@@ -167,13 +167,13 @@ std::string asString(TDataMember const& dm, char* pointer)
         for (int i = 0; i < constantlist->GetEntries(); ++i) {
           const auto e = (TEnumConstant*)(constantlist->At(i));
           if (val == std::to_string((int)e->GetValue())) {
-            return std::string(e->GetName());
+            return {e->GetName()};
           }
         }
       }
     }
 
-    return std::string(val);
+    return {val};
   }
 
   // if data member is a std::string just return
--- a/DataFormats/Detectors/TPC/src/ClusterNativeHelper.cxx
+++ b/DataFormats/Detectors/TPC/src/ClusterNativeHelper.cxx
@@ -66,7 +66,7 @@ std::unique_ptr<ClusterNativeAccess> ClusterNativeHelper::createClusterNativeInd
   for (int i = 0; i < clusters.size(); i++) {
     if (retVal->nClusters[clusters[i].sector][clusters[i].globalPadRow]) {
       LOG(error) << "Received two containers for the same sector / row";
-      return std::unique_ptr<ClusterNativeAccess>();
+      return {};
     }
     retVal->nClusters[clusters[i].sector][clusters[i].globalPadRow] = clusters[i].clusters.size();
     retVal->nClustersTotal += clusters[i].clusters.size();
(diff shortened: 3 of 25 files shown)


davidrohr:
FYI: You have to revert this line to reenable the check in the codechecker: https://github.com/alisw/alidist/pull/3506/files#diff-35397bd99ec0ce432c089b4962d848df6faaf20441ceaa66d44e344b7f61bc0dL56

ktf:
Isn't this change actually dangerous? Take:

```C++
struct Foo {
    int i;
};

struct Bar {
    float f = 0;
    int i = 0;
};

Foo foo() {
    return {1};
}

int foobar() {
    return foo().i;
}
```

if I change foo() to return Bar, it will still compile (because int casts to float automatically) but most likely it will not do what one wants. On the other hand, the version where one has:

```
Foo foo() {
  return Foo{1};
}
```

will complain that Foo cannot be converted to Bar. In this particular case, it's fine because the function is trivial enough. However, for large enough methods with multiple return statements, it suddenly becomes possible to miss a substitution and still to be able to compile.

run-clang-tidy
```diff
@@ -0,0 +1,347 @@
+#!/usr/bin/env python
+#
+#===- run-clang-tidy.py - Parallel clang-tidy runner --------*- python -*--===#
```
sawenzel: this file should probably not be committed

run_clang_tidy.log
```diff
@@ -0,0 +1,2832 @@
+Enabled checks:
+    modernize-return-braced-init-list
```
sawenzel: this file should probably not be committed

aphecetche: @sawenzel indeed. will remove that one and the one above when updating the PR if / when we decide to go with (and_ when the fair logger related PRs are done, as they created some conflict)

Detectors/MUON/MCH/Raw/ElecMap/src/FeeLinkId.cxx
```diff
@@ -31,7 +31,7 @@ FeeLinkId decodeFeeLinkId(uint32_t x)
 {
   uint16_t feeId = static_cast<uint16_t>((x & 0xFFFF0000) >> 16);
   uint16_t linkId = static_cast<uint8_t>(x & 0xF);
-  return FeeLinkId(feeId, linkId);
+  return {feeId, linkId};
```
ihrivnac, line 34: This code produces a compiler error on Mac builder:

```
/System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SOURCES/O2/7714/0/Detectors/MUON/MCH/Raw/ElecMap/src/FeeLinkId.cxx:34:18: error: non-constant-expression cannot be narrowed from type 'uint16_t' (aka 'unsigned short') to 'uint8_t' (aka 'unsigned char') in initializer list [-Wc++11-narrowing]
  return {feeId, linkId};
                 ^~~~~~
```
The` linkId` is declared as `uint16_t` while it takes value of  `uint8_t`, what is also expected in the class constructor.

ktf:
q.e.d. let’s just forget about that option…

Ciao,
Giulio

ihrivnac:
> q.e.d. let’s just forget about that option… Ciao, Giulio
> […](#)
> 

The compiler error happens due to implicit narrowing at `FeeLinkId(feeId, linkId)`,  which passes silently with the current code, so it speaks rather for accepting the option, and fixing the code where needed; in this case

`uint8_t linkId = static_cast<uint8_t>(x & 0xF);
`

Or using `auto` would also work here.

ktf:
The problem I have is that the suggested fix is to remove the type, not to add auto or explicit, possibly poisoning our code of hidden bugs which might take long to understand. All this to end up with code which will look almost exactly like the original one.

ihrivnac:
> 

The example above shows just the opposite: it reveals a hidden bug in the code.
Returning via the class construction may induce an implicit conversion, while when the type is omitted, it is guaranteed that the exact type will be constructed from the initialization list.

ktf:
This has now conflicts, could you regenerate it?

ktf (commented):
This has now conflicts, could you regenerate it?

ihrivnac:
@aphecetche , 
There is another narrowing arror (MacOS build):

```
/System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SOURCES/O2/7714/0/Framework/GUISupport/src/FrameworkGUIDebugger.cxx:269:21: error: non-constant-expression cannot be narrowed from type 'size_t' (aka 'unsigned long') to 'double' in initializer list [-Wc++11-narrowing]
            return {histoData->time[pos], ((int*)(histoData->points))[pos]};
                    ^~~~~~~~~~~~~~~~~~~~

```
Can you commit a fix by changing
`            return {histoData->time[pos], ((int*)(histoData->points))[pos]};
`
with
`            return {static_cast<double>(histoData->time[pos]), ((int*)(histoData->points))[pos]};
`

And similarly for the other errors of this type, if they happen after this fix?

aphecetche:
@ihrivnac will do as soon as I get a chance

aphecetche:
@ihrivnac with the relevant modifications this one is now green.

ihrivnac:
@aphecetche , thank you. I will add the link to the changes in this PR in the discussion item in the WP3 online notes for the discussion tomorrow.

PR #7716, opened 2021-11-23
AOD: include all primaries put by generator

sawenzel:


sawenzel:
@amorsch @nburmaso : This is a simple adjustment to have all primaries from generators (pythia8) included in the AOD. (sorry for eventual overlap but it's just a small change)

nburmaso:
Thank you @sawenzel

PR #7718, opened 2021-11-23
Add trigger word to EMC and FV0 CTF

shahor02:


Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFHelper.h
```diff
@@ -129,6 +129,15 @@ class CTFHelper
     value_type operator*() const { return mData[mIndex].getNumberOfObjects(); }
   };
 
+  //_______________________________________________
+  // Trigger word
+  class Iter_trigger : public _Iter<Iter_trigger, TriggerRecord, uint16_t>
+  {
+   public:
+    using _Iter<Iter_trigger, TriggerRecord, uint16_t>::_Iter;
+    value_type operator*() const { return uint16_t(mData[mIndex].getTriggerBits() & 0xffff); }
```
shahor02, line 138: @mfasDa @mpoghos I've allocated 16 bits max for the trigger bits. Here I simply take the lowest 16 bits of your 32bit word, this is just a placeholder. Instead, you should build the bit pattern (as short as possible), setting the lowest bits of the returned word. If later you decide to add new bits (still<16 in total) you should add them incrementally w/o touching already defined bits (for backward compatibility).

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
```diff
@@ -145,7 +156,8 @@ void CTFCoder::decode(const CTF::base& ec, VTRG& trigVec, VCELL& cellVec)
       cellVec.emplace_back(cell);
       cellCount++;
     }
-    trigVec.emplace_back(ir, firstEntry, entries[itrig]);
+    uint32_t trigBits = trigger[itrig];
```
shahor02, line 159: @mfasDa @mpoghos here you should implement the inverse conversion (in case the content of the `TriggerRecord::mTriggerBits` has different layout wrt the 16 bit word stored in the CTF.

Detectors/FIT/FV0/reconstruction/src/CTFCoder.cxx
```diff
@@ -66,6 +67,7 @@ void CTFCoder::compress(CompressedDigits& cd, const gsl::span<const BCData>& dig
     const auto chanels = digit.getBunchChannelData(channelVec); // we assume the channels are sorted
 
     // fill trigger info
+    cd.trigger[idig] = digit.getTriggers().triggerSignals;
```
shahor02, line 70: @afurs @mslupeck: this implements storage of the trigger word in the CTF. Please let me know asap if there are other changes in FIT CTFs.

afurs: Hello @shahor02, actually there are also last modification which should be applied, @mslupeck what about PM bits for FV0's ChannelData class (like https://github.com/AliceO2Group/AliceO2/blob/630cf86bde29fae35eed3fae9f6b42cf08c473f9/DataFormats/Detectors/FIT/FT0/include/DataFormatsFT0/ChannelData.h#L34) ? I guess that maybe it will be better to add this field in CTF, but without encoding/decoding (temporary), I'm not sure that we will be able to read FV0's ChannelData objects from raw TFs after adding new field.

shahor02: Hi @afurs 
What is this field? There should be no temporary additions to the CTF, if its format needs to be changed, we should finalize and freeze it asap. In general, we cannot modify the existing fields of the CTF, we can only add new ones and this is also not for free.
During the pilot beam you were mentioning that the FT0 may need to store extra trigger info rather than deduce it from channels. Did you come to some conclusion?

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFHelper.h
mfasDa: OK, we can simply add a function getTriggerBitsCoded() to o2::emcal::TriggerRecord to encode the trigger bits we are interested in.

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
mfasDa: OK, here we can then do the inverse coding adding a setCodedTriggerBits() to o2::emcal::TriggerRecord.

Detectors/FIT/FV0/reconstruction/src/CTFCoder.cxx
mslupeck: @afurs I think we really need to freeze the structure of the digits and CTF ASAP. Let's raise this subject during the meeting today and ask if we expect to see any more changes to the electronics flags. I will then implement all the needed changes into digits and this CTF modification can follow.

As for the access to the PB CTFs after this change, didn't we already get most of what we could out of the PB CTF data? To read them we will always be able to use a bit older version of O2. This access problem will be gone when the first collisions of Run3 begin (in about 3 months), because we will collect equivalent data but with all the fixes in place

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFHelper.h
shahor02: Yes, will you send me the code to add it to this PR or do this in a separate one? Also, an inverse method should be provided.

Detectors/FIT/FV0/reconstruction/src/CTFCoder.cxx
shahor02: @mslupeck you will not need old O2 version to access existing CTFs after this change, so far it is backward compatible. Simply, the trigger bits "decoded" from existing CTFs will be empty. The same is true if you add new members. You should not modify existing ones, this will make it backward incompatible and will create a problem not only for you but also for other detectors which read CTF containing FV0.

mslupeck: OK, one problem less to worry about then. Then, the only question is if this PR should go as it is or wait for the update in FV0 digits (if any). This should become clear today afternoon.

shahor02: OK, I will wait for your decision, just try to make it final :)

shahor02: Hi @mslupeck 
Any news ?

mslupeck: Yes. I think we finally understood what is the current status. So this PR is OK for FV0, but we have a related question about FT0. The `eventFlags` field is not needed for FT0: https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/FIT/FT0/include/DataFormatsFT0/Digit.h#L50
It has never been used for any meaningful data and never will be used. As you know, we would like to keep as similar data formats within FIT as possible so it would be nice if this field could be removed (FV0 and FDD don't have it). But I understand that removing it now would cause problems with backward compatibility. Would you (@shahor02) recommend keeping it (even if always empty) or removing it eventually? Perhaps we could minimize the impact if we waited with removing it until late March - at that time we'd no longer need access to this year's PB data, because a newer, more relevant data would become available with the start of Run 3. What do you advise?

shahor02: Hi @mslupeck 

The `FT0::Triggers::eventFlags` was actually not stored and it indeed neither set nor used anywhere. So I've simply suppressed it. Not that the CTF is encoding the https://github.com/AliceO2Group/AliceO2/blob/7f4d8ef8a92062e26887c933b403280d1eaf41ca/DataFormats/Detectors/FIT/FT0/include/DataFormatsFT0/Digit.h#L121, though it was aliased as eventFlags. And actually, it was not set back to digit at encoding. 
I've fixed this and also made the naming more appropriate. 
So far, there is no problem with backward compatibility.
Please confirm that https://github.com/AliceO2Group/AliceO2/pull/7718/commits/5bb2b3ef11a7a52233c1ebf7e97e819b9cc3470f is what you want.

shahor02:
@mslupeck @afurs any comment?

Detectors/FIT/FV0/reconstruction/src/CTFCoder.cxx
mslupeck: Hi @shahor02 
From what I can see you've just renamed the eventFlags to eventStatus. I don't see the mechanism by which you suppressed it by just renaming it...
From the point of view of the datastream coming from the detector, we don't need either eventFlags or eventStatus so it could be removed if such a modification wouldn't break backward compatibility. If it would break the compatibility, then your proposal is OK for now.

shahor02: @mslupeck You wrote above that you want to suppress https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/FIT/FT0/include/DataFormatsFT0/Digit.h#L50 
As I wrote, this data member is never used and actually not stored in the CTF. Under the name of `eventFlags` FT0 CTF was storing different data member: https://github.com/AliceO2Group/AliceO2/blob/25d3ee193255a795ee4d45a5e69a1be2b1948b8c/DataFormats/Detectors/FIT/FT0/include/DataFormatsFT0/Digit.h#L121, so I've renamed the CTF block identifier to make it more explicit. 
Did you want to suppress this data member or both `Triggers::triggersignals` and `Digit::mEventStatus`?

AllaMaevskaya: Please don' suppress this event flag word. We will use it

shahor02: @AllaMaevskaya could you specify what do you call `even flag words`: `Triggers::triggersignals` or `Digit::mEventStatus`?
The 1st is suppressed by this PR, the 2nd: is stored in the CTF

AllaMaevskaya: Triggers::triggersignals should stay, we use it 
Digit::mEventStatus may be can go

shahor02: Sorry, I meant the `Triggers::eventFlags` and / or `Digit::mEventStatus`

AllaMaevskaya: OK, this information could be useful in analysis, but we don't propagate it in AOD any case.

shahor02: Please let me know what should I do with these data members

AllaMaevskaya: Remove them if you want

shahor02:
Does this mean that you are not going to use `Triggers::eventFlags` (which is already suppressed in this PR) and  `Digit::mEventStatus`? If so, I will suppress also the second. The corresponding slot in the CTF will need to stay for the backward compatibility but it will be always empty.

AllaMaevskaya:
Yes, it means that Triggers::eventFlags will not be used. 
Digit::mEventStatus better to keep if  corresponding slot in the CTF will need to stay.

shahor02:
OK, then I am merging this PR, if changes will be needed, please let me know asap.

mfasDa:
@shahor02 I have prepared a PR in your branch with the EMCAL-related encoding. I guess it is not merged in, so I open this as a new PR.

shahor02:
@mfasDa why don't you simply rebase to dev?

mfasDa:
@shahor02  See #7774 there is a remapping of the trigger bits in the compression from 32 -> 16 bit

PR #7724, opened 2021-11-24
TRD fix to get tracklets and digits out of sm16 and sm17

bazinski:
- another boolean cli option to raw data --fixsm1617
- sm16 and sm17 are missing a bit in the trackletHCHeader, sadly the very bit that uniquely identifies them.
- this also requires --tracklethcheader option of 2 i.e. tracklethcheader is always there.
- tracklets and digits for the last 2 sm then appear 
test on desktop due to no power at p2, so testing not finished, but will be by the time this gets through ci.


martenole:
There is also (almost) no data for sector 6. Is it another issue there?

bazinski:
sector 6 has digits but no tracklets, nothing in the raw stream that i could find. Will do a more randomised search when the epn are back.

davidrohr:
@martenole @bazinski : Can we merge this as well?

bazinski:
@davidrohr yes thanks

PR #7725, opened 2021-11-24
Add trigger masks from CTP

nburmaso:


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1396,17 +1399,32 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
-  uint64_t triggerMask = 1;
+  uint64_t triggerMask;
   for (auto& item : bcsMap) {
     uint64_t bc = item.first;
+    auto bcClassPair = bcToClassMask.find(bc);
+    if (bcClassPair != bcToClassMask.end()) {
+      triggerMask = bcClassPair->second;
+    } else {
+      triggerMask = 0;
+    }
```
shahor02, line 1427: Don't we need to account CTP IRs in the collectBCs ?

nburmaso: Yes, you are right. Thank you, @shahor02, I will fix it

sawenzel (approved):

PR #7726, opened 2021-11-24
DPL Analysis: introduce filter selection caching

aalkin:
* Renamed `FilteredPolicy` to `FilteredBase` to better describe its role.
* `Filtered<>` now uses span to handle the selection, backed either by internal cache or external buffer. 
* Framework-managed `Filtered` and `Partition` now have their selections and gandiva filters cached by the framework per data frame, improving performance in cases where they are used in several `process()` functions
* Derived `Filtered` objects (sum or intersection) switch to internal cache automatically
* Combinations with filters will use internal caches
* Fixed bug in Combinations where temporary `Filtered` objects were not owned

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -2072,6 +2072,13 @@ class FilteredBase : public T
     return result;
   }
 
+  auto select(framework::expressions::Filter const& f) const
+  {
+    auto t = o2::soa::select(*this, f);
+    copyIndexBindings(t);
+    return t;
+  }
+
  protected:
   auto slice(uint64_t start, uint64_t end)
   {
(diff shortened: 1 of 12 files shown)


ktf (changes requested):
Seems to break the tests.

aalkin:
Now it passes tests, but breaks tasks in O2Physics that use combinations, investigating.

aalkin:
@ktf the latest change fixes the remaining issues.

ktf:
Seems to fail systematically on hyperloop https://alimonitor.cern.ch/train-workdir/tests/0001/00016096/stdout.log

aalkin:
@ktf should be working now: apparently when I removed `const` in various places, I've also removed `&` which lead to selection cache not being updated. Seeing as everything else was already working, could you test this on hyperloop without waiting for CI?

ktf:
Tested on hyperloop. Merging.

PR #7727, opened 2021-11-24
Get raw-reader.sh to work on macOS

ktf:


ktf:
Is this acceptable? BTW, how could you find `rawAll.cfg` before?

davidrohr:
I just tried, `realpath` works totally fine, thx! I'll also apply that change to the other scripts, to make them work with MacOS.

didn't you want to change `#!/bin/bash` to `#!/bin/bash -e` as well?

For rawAll.cfg, it should be taken from the current directory.
I would leave it like that, since also the other scripts that can read rawData expect it in the local directory, and I would prefer to have the handling homogeneous.

If we want to make it configurable, I would add a `$RAW_DATA_PATH` to `setenv.sh` that defaults to `$(pwd)`, and then use that everywhere.

ktf:
I did not test the `-e` change yet, and since this allows me to have a working reproducer, i did not bother.

Ok to have the rawAll.cfg in the current folder.

davidrohr:
@ktf : Please check #7728 : It adds a `RAWINPUTDIR=foo` env variable (taken from O2DataProcessing) to be able to set the path to the input data. That is now used in all scripts. I also applied your MacOS fix to the other scripts.

ktf:
Sounds good to me.

PR #7729, opened 2021-11-24
DPL: Extending ExternalFairMQDeviceProxy benchmark

matthiasrichter:
- handle exception from  shm allocation

Depending on the configuration, there might not be enough shared memory,
handling this exception gracefully.

- more proxy bypassing options

Option '--bypass-proxies' none, all(a), output(o)

We can now bypass both proxies to get the metrics for the pure producer
to checker traffic, or bypass output proxy only to benchmark with one
proxy in between. It is not possible to have only the output proxy.

### DIFF
--- a/Framework/Core/test/benchmark_ExternalFairMQDeviceProxies.cxx
+++ b/Framework/Core/test/benchmark_ExternalFairMQDeviceProxies.cxx
@@ -27,6 +27,18 @@ using namespace o2::framework;
 #include "Headers/DataHeader.h"
 #include "fairmq/FairMQDevice.h"
 #include <chrono>
+#include <sstream>
+
+namespace benchmark_config
+{
+enum struct ProxyBypass {
+  None,
+  All,
+  Output,
+};
+}
+std::istream& operator>>(std::istream& in, enum benchmark_config::ProxyBypass& val);
+std::ostream& operator<<(std::ostream& out, const enum benchmark_config::ProxyBypass& val);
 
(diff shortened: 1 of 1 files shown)


ktf:
This looks good. Could you please merge the two automatically opened PRs you have for it and then merge this one?

ktf (dismissed):

matthiasrichter:
nice tool with the automatic PRs, but naturally doesn't work with two conflicting changes, in this case I had to do it manually.

And can "squash and commit" be selected by default?

ktf:
If you merged one, then the second one would have been updated at some point.

I will check if it's possible to make "squash and merge" default.

ktf:
Apparently it's not possible...

matthiasrichter:
@ktf I would add the reference to the text of the PR not only the title, and the project needs to be in the reference I think. So that GitHub is adding cross links to the original PR. This will make the suggestions more visible. I have noticed the feature only now when you mentioned it.

Something like "This is a follow-up to AliceO2Group/AliceO2#PR ." I'm not exactly sure about the syntax

I reopened, think you have closed accidentally.

ktf:
I think I already updated the script in a way which is close to your suggestion. See:

https://github.com/AliceO2Group/AliceO2/commit/dc0dd953c8a683c9e1665217a551f83b737c2e98

you might have to rebase in order to see the new message...

matthiasrichter:
very good!

PR #7730, opened 2021-11-24
DPL GUI: drastically improve ability to debug large topologies

ktf:
This marks in purple incoming and outgoing connections to the selected
node. Notice that the tip of the outgoing arrow is still yellow,
due to the fact it's considered to be part of the target node. Good
enough improvement for now.


ktf:
Obligatory screenshot:
image

PR #7737, opened 2021-11-25
add TOF diagnostic from ccdb in digitisation (empty crates, TRM errors)

noferini:
Changes in simulation (@chiarazampolli ) to read from ccdb
- offset calibration (to decalibrate)
- diagnostic words -> switch off crates, add TRM errors words

tested with pythia 1000 events@50 kHz  (1 TF) using diagnostic of a run with only 1 crate -> digit QC as expected (matching data QC)

### DIFF
--- a/DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
+++ b/DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
@@ -33,10 +33,10 @@ class Diagnostic
   Diagnostic() = default;
   int fill(ULong64_t pattern);
   int fill(ULong64_t pattern, int frequency);
-  int getFrequency(ULong64_t pattern);                                                    // Get frequency
-  int getFrequencyROW() { return getFrequency(0); }                                       // Readout window frequency
-  int getFrequencyEmptyCrate(int crate) { return getFrequency(getEmptyCrateKey(crate)); } // empty crate frequency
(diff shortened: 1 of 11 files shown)


noferini:
I forgot to say that for the moment to enable ccdb in MC I followed the same philosophy of reconstruction by adding a flag
--use-ccdb-tof

noferini:
all checks have passed

DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
```diff
@@ -49,7 +49,8 @@ class Diagnostic
   void fill(const Diagnostic& diag);                       // for calibration
   void fill(const gsl::span<const o2::tof::Diagnostic>){}; // for calibration
   void merge(const Diagnostic* prev);
-  void getNoisyMap(Bool_t* output); // set true in output channel array
+  void getNoisyMap(Bool_t* output, int noisyThr = 1); // set true in output channel array
+  void getNoisyLevelMap(Char_t* output);              // set true in output channel array
```
shahor02: Aren't these const methods?

noferini: Yes, they are. Moved to const (including also other non static getters).

shahor02 (approved):

PR #7739, opened 2021-11-25
AOD: fixe memory leak + allow overwriting metadata

shahor02:


Framework/Core/src/CommonDataProcessors.cxx
```diff
@@ -354,7 +354,7 @@ DataProcessorSpec
         auto s = pc.inputs().get<TableConsumer>(ref.spec->binding);
         auto table = s->asArrowTable();
         if (!table->Validate().ok()) {
-          LOGP(WARNING, "The table \"{}\" is not valid and will not be saved!", tableName);
+          LOGP(warnING, "The table \"{}\" is not valid and will not be saved!", tableName);
```
sawenzel: is that a typo? `LOGP(warn)` ?

ktf:
```suggestion
          LOGP(warning, "The table \"{}\" is not valid and will not be saved!", tableName);
```

Framework/Core/src/CommonDataProcessors.cxx
```diff
@@ -383,7 +383,7 @@ DataProcessorSpec
           } else {
             ta2tr.addAllBranches();
           }
-          ta2tr.process();
+          delete ta2tr.process();
```
sawenzel: What is the problem here? `ta2tr` does not seem to be a pointer. Is this deleting an object returned by `process()` which is not **owned** by `ta2tr`? This looks a bit awkward.

jgrosseo:
Hi Ruben,
Why we need the case of overwriting the metadata?

Framework/Core/src/CommonDataProcessors.cxx
shahor02: It is actually a feature of the upper to lower case converter: it checks only for `|LOGP[(]WARN|LOGP(warn|g;`
@davidrohr can this be fixed?

shahor02:
@jgrosseo because currently, it produces FATAL when metadata is there

Framework/Core/src/CommonDataProcessors.cxx
shahor02: TableToTree returns the pointer to the tree after the writing. It has only a default constructor, so although it has a pointer data member, it is not deleted. The alternative would be to define it as unique_ptr, but then its potential consumer should be aware of this.

Framework/Core/src/CommonDataProcessors.cxx
shahor02: @ktf I've learned that this case converter is yours, can you add `WARNING`?

jgrosseo:
> @jgrosseo because currently, it produces FATAL when metadata is there

Yes, that is not good of course. I would simply change that to WARNING then. The overwrite does not harm, but does not help either...

shahor02:
@jgrosseo do you want simply skip creating the metada if it is already there?

shahor02:
@sawenzel @ktf I've changed the raw mTree pointer in the TableToTree to shared one and removed the deletion of the returned pointer.

ktf:
@shahor02 i would actually check if we cannot simply avoid returning the tree.

jgrosseo:
> @jgrosseo do you want simply skip creating the metada if it is already there?

Yes, exactly. It will be identical (unless we have a quite bad setup...)

shahor02:
@ktf check with whom? At the moment the returned pointer is never used, but I understood the returning was done on purpose in case somebody will need this pointer.

shahor02:
@jgrosseo ok, I've changed the code to preserve the existing one

ktf:
@shahor02 what i meant is that the pointer might have been used, so before changing it to anything else, we should have checked. If it's not used, I would be in favour of simply deleting the object inside the function and not return anything.

shahor02:
Hi @ktf, currently the pointer is not used in any real code, but is used in the test: https://github.com/AliceO2Group/AliceO2/blob/b82fa607e5278a265da5d04c9c25231873795b56/Framework/Core/test/test_TreeToTable.cxx#L150-L154
Let me know if you want to preserve it.
In any case, to have this fix in the nightly, I would merge this PR once the CIs are green, then, if decided, can avoid returning a pointer and implement a destructor.

ktf:
yeah... I guess the correct thing is to return a unique ptr, then. We can merge this and think about it later.

shahor02:
> yeah... I guess the correct thing is to return a unique ptr, then. We can merge this and think about it later.

For the record: at the moment it returns shared ptr, which is destroyed since never assigned.

ktf:
Yes, sure, that would work as well. I guess the unique_ptr is maybe better / safer, but we can always change it later.

PR #7742, opened 2021-11-25
DPL Analysis: ability to read and write variable list arrays

aalkin:


ktf:
As discussed offline, we should probably move everything to use GetEntriesSerialised and do the byteswap ourselves as part of the copy. It should boost a (large?) bit the performance and simplify the VLAs to a simple corner case.

aalkin:
@ktf the error in `o2-dataflow-cs8` is very confusing, can you have a look?

ktf:
It now returns a shared_ptr, due to a change made yesterday to fix a leak.

aalkin:
> It now returns a shared_ptr, due to a change made yesterday to fix a leak.

Ah, it came in after I forked the dev for this PR. Rebased and fixed.

aalkin:
@ktf it seems macos ARM is not happy about byteswap...

ktf:
Weird, in principle M1 is little endian. Can you add a unit test for the byteswap function?

ktf:
Also the function itself we should probably move to `Framework/Endian.h`

aalkin:
It seems I've fixed the byteswap, but shared pointer in `TableToTree` does not prevent the crashes:
```
o2-test-framework-TreeToTable(56676,0x10a00ee00) malloc: *** error for object 0x7face369f268: pointer being freed was not allocated
o2-test-framework-TreeToTable(56676,0x10a00ee00) malloc: *** set a breakpoint in malloc_error_break to debug
```
This happens very rarely, on closing the file that has the same tree as an instance of the `TableToTree`, because, as I understand, the file owns it and does not use a shared pointer.

ktf:
Adding also @shahor02.

Yes, that's what I suspected as the reason why the file was "leaked". I am surprised however that it does not crash consistently. I will have to look on how to properly close the tree without closing the branch.

PR #7751, opened 2021-11-26
DPL GUI: fix crash on macOS

ktf:


ktf:
Calling pollGUIPostRender alone obviously can create some serious troubles if it needs to finalise something which was started in pollGUIPreRender. @luigifusco do you remember the rationale for it?

luigifusco:
PostRender is a bad naming choice.
It does the actual rendering from the perspective of opengl (takes the imgui data and performs the opengl calls).
Render generates the imgui data and handles stuff like timing.

So it should stay outside the if.

ktf:
Mmm... Are you sure that's correct? If `glfwWindowShouldClose(window)` returns true, the code in pollGUIPreRender() is not executed, but the pollGUIPostRender is still executed, with all it could derive from it. I suspect this is actually the reason for the crash on exit...

luigifusco:
If `glfwWindowShouldClose(window)` is called and returns true the `gui_callback` function returns as well.
The only problematic flow I see is: glfw wants to close but less than 15ms have passed, so `gui_callback` never realizes it and tries to draw anyway with `pollGUIPostRender` using an already generated frame.

luigifusco:
`remoteGuiCallback` could be a source of problems in that sense as well if the crash is related to having remote guis connected.

ktf:
merging as it is currently broken. we can nevertheless continue the discussion

PR #7752, opened 2021-11-26
DPL: baby steps going to Lifetime::Dangling

ktf:
* do not try to cache the AOD information
* keep dangling information as a vector of bools


ktf:
Mostly mechanic transformation. Works also on hyperloop. That said, since  a couple of days we seem to have performance issues there.

PR #7754, opened 2021-11-26
TPC dEdx: adding container class for calibrations

matthias-kleiner:
Provide a container class that contains all necessary calibration objects for the dE/dx instead of providing them separately.
Adding a polynomial correction (5D) for the track topology correction as an alternative to the spline (3D) method.
One can select the topology correction method (splines or polynomials) by providing the specific input file for the corresponding method.

TPCdEdxCalibrationSplines:
- renaming and moving TPCdEdxCalibrationSplines class to be consistent with the new polynomial correction class.
- updating splines from 2D to 3D
- renaming file name for loading the topology correction for consistency


wiechula (dismissed):
For me this looks fine! Perhaps just one minor comment for the future. The pol4 with 5D correction is now hard coded. But you could think about templating the pol order and dimensionality. We discussed to perhaps use pol3 and 6D including the zero suppression threshold.

wiechula:
I also checked the GPU impact. The pol4 5D is a bit faster than the spline3D:
```
0.no-dedx : 21.16 +- 4.46
1.with-dedx : 30.78 +- 4.94
2.dedx-spline3D : 33.69 +- 5.16
3.dedx-pol4_5D : 31.21 +- 4.85
```

wiechula (dismissed):

sgorbuno (dismissed):

wiechula (approved):

wiechula:

PR #7755, opened 2021-11-26
MRRTF-145: only exes (not libs) should bake the mapping implementation

aphecetche:
@pillot, as we discussed, this is an attempt to make the muon libraries independent from the actual mapping implementation (as they were intended to be in the first place but weren't really). 

The actual implementation should now only be chosen when building an executable. The current use case is to allow to have run2-compatible executables (using Impl3 of the mapping = run2 version) at the same time as Run3 executables (using Impl4 = current run3 version with e.g. one DS less).

Note that so far I've only tested it on my Mac (ctest and reco-related workflows). 

I'm not making it a draft (which it kind of is still for the moment) to see what the CI says on other platforms...

### DIFF
--- a/Detectors/MUON/MCH/Mapping/Interface/CMakeLists.txt
+++ b/Detectors/MUON/MCH/Mapping/Interface/CMakeLists.txt
@@ -9,4 +9,7 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-o2_add_header_only_library(MCHMappingInterface)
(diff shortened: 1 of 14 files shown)


aphecetche:
As far as I can tell, the modifications should now be ok for linux and Mac. 
I guess the CI failure is not related to this one.

aphecetche:
@shahor02 @sawenzel this one is green and ready for merging. Can it been reviewed and/or merged ? Thanks.

shahor02 (approved):

PR #7756, opened 2021-11-27
enable optical surface properties for G4

AllaMaevskaya:


sawenzel (approved):

sawenzel:
It appears to fix https://alice.its.cern.ch/jira/browse/O2-2703. So fine with me.

PR #7761, opened 2021-11-29
[EMCAL-507] Remove custom end-of-stream handling via EMCALBlockHeader

mfasDa:
Terminate condition checking the custom EMCALBlockHeader
added as additional payload by the publishers no longer needed.
- Remove publishing of EMCALBlockHeader
- Remove checkReady terminate condition in
  ROOTTreeWriterSpecs
- Remove includes of headers in various Spec

In addition increase verbosity level from LOG of the number
of cells at the end of run of the RawToCellConverter from debug
to info.


mfasDa:
@matthiasrichter Do you see any problem with this?

PR #7762, opened 2021-11-29
Adding protection for straight MFT tracks, to prevent divison by zero

sarahherrmann:


ktf:
@jgrosseo and @aalkin should probably comment, but shouldn't we just avoid the division in that case and multiply by the provided 1/Pt? Where is this used?

ktf (commented):
@jgrosseo and @aalkin should probably comment, but shouldn't we just avoid the division in that case and multiply by the provided 1/Pt? Where is this used?

jgrosseo:
This is in principle fine. Of course alternatively we could consider to fill 1/pt with a very large number for straight tracks instead of 0. What do you think about this?
Tagging also @rpezzi

sarahherrmann:
Hi ! Anton and Rafael are actually already aware of this proposal of modification (they were the ones that suggested me to do that for MFT tracks), and I don't know for the provided 1/Pt question sorry

rpezzi:
Hi, I am OK with the current implementation.

jgrosseo:
OK, thanks.

Last question: did you try to use `Almost0` instead of the hard-coded small float value? Is `Almost0` too small? Do we need a second constant?

sarahherrmann:
I didn't try `Almost0`, I didn't know it existed. I will try it

rpezzi:
Indeed `Almost0` is more appropriate. It is safe for a floating point division by `Almost0`:
```
1.f/o2::constants::math::Almost0
(float) 8.50709e+37f
```

jgrosseo (approved):

PR #7764, opened 2021-11-29
CTF dictionaries are ready to be fetched from CCDB

shahor02:
+ macro to create plain vector-based per-detector dictionaries from the global tree-based CTF dictionary produced by the `o2-ctf-writer-workflow`.

The dictionaries used for the commissioning are uploaded to http://alice-ccdb.cern.ch/<DET>/Calib/CTFDictionary with validity from 16/07/21 to 12/05/22.

@MichaelLettrich note there were some changes in the createCoders methods.

### DIFF
--- a/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/CTFDictHeader.h
+++ b/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/CTFDictHeader.h
@@ -18,6 +18,7 @@
 
 #include <Rtypes.h>
 #include <string>
+#include "DetectorsCommonDataFormats/DetID.h"
 
 namespace o2
 {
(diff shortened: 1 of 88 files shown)


MichaelLettrich (commented):
Changes look fine for me. I will rebase once you merged your changes and add my adaptations.
However I would strongly recommend to factor out the `getFreq` and `getProbBits` lambdas as functions from the individual CTF Coders, as this level of copy-paste code duplication is widely considered a code smell. If you share my opinion but lack the time to do it I can take care of it in a separate PR.

shahor02:
@MichaelLettrich the `getFreq` and `getProbBits` don't need to be lambdas, now these are members of the EncodedBlocks. Also added more simplifications.

MichaelLettrich (dismissed):
Thanks for taking my comments into consideration. I think this makes the CTF Coders more maintainable.

PR #7766, opened 2021-11-30
fix in TOF latency

noferini:
@chiarazampolli 
this is to fix the 3.5 mus misalignment of TOF (140 BC)


noferini:
This is a trivial change.
Do we need all checks passed?

shahor02 (approved):
trivial fix, merging

PR #7768, opened 2021-11-30
DPL: cleanup struct to tuple

ktf:


ktf:
@jgrosseo @aalkin as discussed. Having support for 99 elements rather than 40 does not seem to have any noticeable slowdown.

ktf:
Of course this does not mean that actually parsing 99 is not slower than parsing 40. Merely that if the concern is allowing for more configurables, we can do that without a global impact.

ktf:
Tested also on hyperloop, works.

jgrosseo:
@ktf Thanks! Have you tried how long it takes to compile something with 99 elements?

jgrosseo:
Any thoughts about making this also consuming structs?

PR #7769, opened 2021-11-30
TRD remove straggling error output and add option for trigger fix

bazinski:
- option --fixforoldtrigger ignore the 2 step trigger for physics/calibration nodigit/digit data.
- remove a straggling text output for non sequential ROB, printing digit mcm header.

### DIFF
--- a/Detectors/TRD/reconstruction/src/CruRawReader.cxx
+++ b/Detectors/TRD/reconstruction/src/CruRawReader.cxx
@@ -542,7 +542,7 @@ int CruRawReader::processHalfCRU(int cruhbfstartoffset)
       ** DIGITS NOW ***
       *****************/
       // Check if we have a calibration trigger ergo we do actually have digits data. check if we are now at the end of the data due to bugs, i.e. if trackletparsing read padding words.
-      if (linkstart != linkend && mCurrentHalfCRUHeader.EventType == o2::trd::constants::ETYPECALIBRATIONTRIGGER) { // calibration trigger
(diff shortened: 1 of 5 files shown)


martenole (dismissed):

davidrohr:
can someone approve, then I would merge?

bazinski:
@tdietel @martenole could one of you approve this so i can get this off my list ;-) thanks

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -542,7 +542,7 @@ int CruRawReader::processHalfCRU(int cruhbfstartoffset)
       ** DIGITS NOW ***
       *****************/
       // Check if we have a calibration trigger ergo we do actually have digits data. check if we are now at the end of the data due to bugs, i.e. if trackletparsing read padding words.
-      if (linkstart != linkend && mCurrentHalfCRUHeader.EventType == o2::trd::constants::ETYPECALIBRATIONTRIGGER) { // calibration trigger
+      if (linkstart != linkend && (mCurrentHalfCRUHeader.EventType == o2::trd::constants::ETYPECALIBRATIONTRIGGER || mOptions[TRDIgnore2StageTrigger])) { // calibration trigger
```
martenole, line 545: So without the 2-stage trigger we always had digits in the raw data? Did you test it with both old and new raw data format?

bazinski: Yes, but only hcheader and trailers, but it gets caught by the linkstart != linkend if there are no digits.

martenole: sounds good

martenole (approved):

PR #7770, opened 2021-11-30
DPL: make Dispatcher lossy

ktf:
* Allow any device to be configured with send timeout via `--fairmq-send-timeout` option.
* Return immediately if Dispatcher is not able to send and queue messages in the FairMQ queue, which is dimensioned according to `--fairmq-send-buffer-size`. By default 4 when the DPL driver is used. If `--fairmq-send-timeout` is provided, the custom timeout is used. Notice that in order for this to be working as expected, we had to swap the bind / connect method for the source / sink devices.


ktf:
@Barthelemy @knopers8 this should make it less lossy...

knopers8:
Weirdly it is not any better:
```
+ entries=153
+ '[' 153 -ne 200 ']'
```
```
+ entries=139
+ '[' 139 -ne 200 ']'
```
With message queues of size we should see 100% delivery rate in this test, so either the corresponding channels are not set to the correct size or `Send()` with timeout value does not do what we would expect...

ktf:
@rbx @dennisklein to the rescue. I checked in lldb and the rcvBufSize and sndBufSize are correctly set to 256 messages for the outgoing channels. Do we need to explicitly  set also the SndKernelSize (or whatever it is called)?

rbx:
`rcvBufSize` and `sndBufSize` are the sizes of the zeromq queue, measured in number of messages. `rcvKernelSize` and `sndKernelSize` are the sizes of the underlying transport buffers (for TCP/IPC will be given to SO_SNDBUF/SO_RCVBUF), measured in bytes.

The latter, if unset, should take system setting.

ktf:
Thanks, then i do not understand. Why would it drop messages if the queue is longer than the number of messages?

rbx:
Also note that the `*kernelSize` is tricky for shared memory transport, since only meta info of each message is transferred via zeromq, which is 40 bytes, no matter the actual msg size in shmem. So if you need a low limit here, you'll have to consider this.

rbx:
> Thanks, then i do not understand. Why would it drop messages if the queue is longer than the number of messages?

Should not.
Are both endpoints connected?

ktf:
i would say so. if we change the timeout to -1, every message is received. if we change the timeout to 1, 99% of the messages are received. the problem seems to be timeout 0.

rbx:
If a socket starts sending before any peers connect to it, then the behavior will depend on whether the socket binds or connects. If it connects, msg will be queued as expected. If it binds, it will not be able to queue. This is because the zmq queues exist per peer. When connecting it is clear that there is one queue, but the binding side can have multiple peers, each with its own queue. So when nothing is connected, there isn't actually any queue yet.

rbx:
Perhaps the sending happens so early that the bind/connect is not fully complete yet. The bind/async of zmq happens asynchronously, and it does not block until it actually succeeds...

rbx:
Perhaps we need to set this option: http://api.zeromq.org/master:zmq-setsockopt#toc25
Although timeout of 0 requests non-blocking transfer, so it may return in this case still, need to test.

ktf:
Indeed the sources are binding right  now... Maybe I should swap connect / bind... For now let's do it on the Dispatcher only. https://github.com/AliceO2Group/AliceO2/pull/7770/files#diff-3ab9ab3068fa42fdd0c56c60179107507b9676f5fd6b461026e1c58cf1477325R30-R48 should be enough.

rbx:
Can you add a sleep before starting the transfer, to make sure what we are seeing here. This should give zmq enough time to complete the connection. if everything completes as you expect with the added sleep then it is indeed the effect i suspect.

ktf:
Sure. However, if it works by swapping connect and bind, doesn't that mean the issue is indeed using bind on the push side?

rbx:
> Sure. However, if it works by swapping connect and bind, doesn't that mean the issue is indeed using bind on the push side?

True, that should confirm it too. I wrote that before I read your previous comment..

ktf:
Swapping seems to have done the trick, at least for o2-cs8. @knopers8, in case this works, what other tests should  we do before deployment?

ktf:
@rbx will I have the same problem  if I use PAIR rather than push/pull?

rbx:
> @rbx will I have the same problem if I use PAIR rather than push/pull?

Just tested it - it is the same behavior also with PAIR. A bit unexpected, since there he should know that there can be only one peer. Perhaps there is more that depends on this.

ktf:
Ok, thanks for checking. Indeed I find it bizarre... I guess I will move to use connect on the sources everywhere. Do I understand correctly that if -1 is passed, it will still wait? Or will it drop messages also in that case?

knopers8:
@ktf I will try to play with it myself today. 
Then, as for now, the DPL-generated AliECS workflow templates do not include queues' sizes, so perhaps we should change this first, otherwise we will use some default value and I do not know what it is.
Then we could merge and test on FLPs on the next opportunity.

BTW, how come that your modification does not cause the workflow to bind on both sides of Dispatcher's input channel and connect on both sides of the output channel?

rbx:
> Ok, thanks for checking. Indeed I find it bizarre... I guess I will move to use connect on the sources everywhere. Do I understand correctly that if -1 is passed, it will still wait? Or will it drop messages also in that case?

With timeout of -1 it will wait.

rbx:
> Or will it drop messages also in that case?

Actually it doesn't drop, it just tells you that it cannot send with timeout of 0. You can then drop or not drop.

ktf:
@knopers8 the queue is 256 element long right now for the dispatcher (and only for the dispatcher) and can be modified by using `--fairmq-recv-buffer-size` and `fairmq-send-buffer-size`. I am not sure I get the other point: what happens is that the policy allows you to modify independently the method and takes care of doing the correct thing on both ends. If you notice there is two policies one for channels associated to an outgoing connection and one for channels associated to an incoming one.

@rbx. I am not sure I understand what you mean by: "Actually it doesn't drop, it just tells you that it cannot send with timeout of 0. You can then drop or not drop." Currently what I do is that the source device binds, and then starts sending messages with timeout -1. Will it wait for a peer to connect, if no one is connected?

rbx:
> Will it wait for a peer to connect, if no one is connected?

With a timeout of -1 it will wait.

What I mean is that even with a timeout of 0, it doesn't actually drop anything. It returns the call if it cannot send immediately and leaves the message intact. You can then retry with the same message or drop it.

knopers8:
@ktf Regarding the queue sizes, I meant that we will not have influence on those in AliECS-controlled environments if we do not add buffer sizes in workflow templates, since AliECS takes care of configuring the channels. I will take care of modifying this.

ktf:
@knopers8 of course. I forgot the channels are actually created by aliecs...

@rbx: yes, sure. It's just that DPL deletes the messages it cannot deliver. I could also change that but I would prefer it is handled on the FairMQ side, frankly. I will experiment a bit.

ktf:
Ok, with the default queue length (4) we get 190 out of 200, I guess we can increase it a bit and call it done. @knopers8 would you agree? 16?

knopers8:
Yes, thank you. I will also make the check more tolerant to dropped messages, say 150 out of 200 for the start.

ktf:
@knopers8 @Barthelemy what do we want to  do with this? It seems that in the test the only way to get it to not lose any message is to use 256 for the queue. Shall we leave it like that?

knopers8:
Since the latest QC bump (https://github.com/alisw/alidist/commit/1890565fcadf3a8556dd00fa33d32a99932018d3) we accept >150/200 entries in the test. However, I noticed before the bump (around Thu/Fr) that even with queues of size 256 we did not get all messages, so something still looks suspicious. I would propose to wait with merging.

PR #7771, opened 2021-11-30
DPL Analysis: Rework AOD I/O

aalkin:
* Use `GetEntriesSerialized` and build arrow buffer directly
* Add support for reading/writing variable size arrays
* Fix double free on ROOT file close
* Fix pointer to buffer not being moved

### DIFF
--- a/Framework/Core/include/Framework/TableTreeHelpers.h
+++ b/Framework/Core/include/Framework/TableTreeHelpers.h
@@ -20,6 +20,10 @@
 // =============================================================================
 namespace o2::framework
 {
+namespace TableTreeHelpers
+{
+static constexpr char const* sizeBranchsuffix = "_size";
+}
 
 // -----------------------------------------------------------------------------
 // TableToTree allows to save the contents of a given arrow::Table into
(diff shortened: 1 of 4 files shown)


Framework/Core/include/Framework/TableTreeHelpers.h
```diff
@@ -20,6 +20,10 @@
 // =============================================================================
 namespace o2::framework
 {
+namespace TableTreeHelpers
+{
+static constexpr char const* sizeBranchsuffix = "_size";
+}
```
ktf: This should be moved to a private header and probably renamed to sizeBranchSuffix (or just renamed in place...)

Framework/Foundation/include/Framework/Endian.h
```diff
@@ -15,10 +15,16 @@
 // Lookup file for __BYTE_ORDER
 #ifdef __APPLE__
 #include <machine/endian.h>
+#define swap16_ ntohs
+#define swap32_ ntohl
+#define swap64_ ntohll
```
ktf, line 20: Why do you actually need the aliases? Can't you simply stick to the POSIX nomenclature?

aalkin: No, I need uniform definitions here, both apple and gcc have aliases to builtin functions for swapping, only apple ones are mapped to `ntoh/hton`.

ktf: all you need  is to define ntohll on linux, no? ntohl and ntohs are POSIX.1-2001 and are available on both mac and linux.

aalkin: No, these functions are somewhere in network headers on linux, while the builtins are gcc's own, which is simpler to use.

ktf: I disagree, the header in question is quite minimal and in any case it would end up in a couple of CXX files. Anyways, we can fix it later.

ktf:
Tested on hyperloop. Merging.

PR #7773, opened 2021-11-30
sim_challenge.sh: Add basic analysis testing

sawenzel:


sawenzel:
@davidrohr : This needs some adjustment in the full-system-ci recipe which I'll do separately.

sawenzel:
Should be activated after https://github.com/alisw/alidist/pull/3619

PR #7774, opened 2021-11-30
[EMCAL-729] Add compression of trigger bits

mfasDa:
Reduction from 32 bit to 16 bit filtering the most
relevant triggers for EMCAL. For the moment
3 triggers are propagated
- TF trigger
- Physics trigger
- Calib trigger
In case other triggers will be included in the
compressed output, they need to be appended.

### DIFF
--- a/Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
+++ b/Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
@@ -141,6 +141,7 @@ void CTFCoder::decode(const CTF::base& ec, VTRG& trigVec, VCELL& cellVec)
   o2::InteractionRecord ir(header.firstBC, header.firstOrbit);
 
   Cell cell;
+  TriggerRecord trg;
   for (uint32_t itrig = 0; itrig < header.nTriggers; itrig++) {
     // restore TrigRecord
     if (orbitInc[itrig]) {  // non-0 increment => new orbit
@@ -156,8 +157,10 @@ void CTFCoder::decode(const CTF::base& ec, VTRG& trigVec, VCELL& cellVec)
       cellVec.emplace_back(cell);
       cellCount++;
     }
-    uint32_t trigBits = trigger[itrig];
-    trigVec.emplace_back(ir, trigBits, firstEntry, entries[itrig]);
+    trg.setBCData(ir);
+    trg.setDataRange(firstEntry, entries[itrig]);
+    trg.setTriggerBitsCompressed(trigger[itrig]);
+    trigVec.emplace_back(trg);
   }
   assert(cellCount == header.nCells);
 }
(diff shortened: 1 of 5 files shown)


shahor02 (changes requested):
@mfasDa could you also fix the CTF unit test?

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
```diff
@@ -156,8 +156,8 @@ void CTFCoder::decode(const CTF::base& ec, VTRG& trigVec, VCELL& cellVec)
       cellVec.emplace_back(cell);
       cellCount++;
     }
-    uint32_t trigBits = trigger[itrig];
-    trigVec.emplace_back(ir, trigBits, firstEntry, entries[itrig]);
+    auto currenttrigger = trigVec.emplace_back(ir, firstEntry, entries[itrig]);
+    currenttrigger.setTriggerBitsCompressed(trigger[itrig]);
```
mfasDa: Reference for some reason seems to not work. The trigger bits stay unmodified. Will change it in the same way as done for the cells.

mfasDa:
@shahor02 OK, in fact besides missing adaption of the reference they were showed indeed a bug in the new decoding of compressed trigger bits. One question: Why do you prefer BOOST_CHECK with respect to BOOST_CHECK_EQUAL for the check for equal? If nothing speaks against I will change to BOOST_CHECK_EQUAL as this not only tells me that the test failed but why it failed.

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
shahor02: you should use `auto& currenttrigger = trigVec.emplace_back(ir, firstEntry, entries[itrig])`

mfasDa:
In addition: Based on the discussion in the EMCAL meeting we drop the timeframe trigger from encoding.

shahor02:
> If nothing speaks against I will change to BOOST_CHECK_EQUAL as this not only tells me that the test failed but why it failed.

you can change it if you prefer.

shahor02:
> In addition: Based on the discussion in the EMCAL meeting we drop the timeframe trigger from encoding.

could you clarify what do you mean? If this is about the TF trigger bit of the RDH, you anyway don't have it, right?

mfasDa:
> > In addition: Based on the discussion in the EMCAL meeting we drop the timeframe trigger from encoding.
> 
> could you clarify what do you mean? If this is about the TF trigger bit of the RDH, you anyway don't have it, right?

Exactly, and that's why we drop it from being encoded (with respect to the initial version)

mfasDa:
@shahor02 should be ready for merge now?

shahor02 (approved):

PR #7775, opened 2021-11-30
[WIP] [MRRtF-150] improved digits time computation

aferrero2707:
The new code improves the selection of the HeartBeat packets used for the computation of the SAMPA digit time.


aphecetche:
finalized in #7996

PR #7777, opened 2021-12-01
DPL Analysis: add natan2() node for expressions

aalkin:
@jgrosseo as `atan2()` function is already there in Gandiva, we only need a new functional node in expressions to be able to use it.


jgrosseo (approved):

aalkin:
@ktf could you please merge this?

PR #7778, opened 2021-12-01
Configure CCDB server via NameConf::mCCDBServer and NameConf::getCCDBServer()

shahor02:


AllaMaevskaya (dismissed):

noferini (approved):
For TOF it is fine,
since I have an open PR affecting calibration and ccdb I will take care to align the TOF code also there once this is in dev

martenole (approved):
Hi @shahor02 
from TRD side looks good. I was wondering if it would make sense to change the default URL for the CCDB manager as well? If I do
```
root [0] auto &mgr = o2::ccdb::BasicCCDBManager::instance()
[INFO] Is alien token present?: 0
(o2::ccdb::BasicCCDBManager &) @0x7f0c8ccb3028
root [1] mgr.getURL()
(const std::string &) "http://ccdb-test.cern.ch:8080"
root [2] 
```
I still get the test CCDB. And for instance the tracklet transformer is not setting the URL explicitly: https://github.com/AliceO2Group/AliceO2/blob/e076a97458688349c03c613f160782aa37fd5135/Detectors/TRD/base/src/TrackletTransformer.cxx#L46-L49
I would change this now, but I don't know if this is the case also in other places of O2..

davidrohr (approved):
thx a lot for taking care!
Full CI is failing, I am just trying it locally.

davidrohr:
> Hi @shahor02
> from TRD side looks good. I was wondering if it would make sense to change the default URL for the CCDB manager as well? If I do

Absolutely, I would do that. This configKeyValues should set all CCDB defaults.

shahor02:
> Full CI is failing, I am just trying it locally.
This is because some ccdb objects, particularly, that of ZDC, exist only on the test ccdb server.

shahor02:
> > Hi @shahor02
> > from TRD side looks good. I was wondering if it would make sense to change the default URL for the CCDB manager as well? If I do
> 
> Absolutely, I would do that. This configKeyValues should set all CCDB defaults.

This would mean making CCDB dependent on the DataFormats. So far, we were trying to keep CCDB clean, the only O2 dependency is on O2::CommonUtils (which contains the ConfigurableParam class). What we can do is to define a special configurable CCDBParam and move the NameConf::mCCDBServer there.

davidrohr:
I don't see a large problem if the CCDB depended on data formats.
Although, perhaps I would move the nameConf from DataFormats/Detectors/Common to DataFormats/common.

Having another configKeyValues class for the CCDB would also be fine for me.
Don't have a strong preference.

shahor02:
@cortesep I've copied the last objects in ZDC/Config/Sim and ZDC/Config/Module from http://ccdb-test.cern.ch:8080 to http://alice-ccdb.cern.ch. If you need to update them, please do this directly on the production server.

shahor02:
> Although, perhaps I would move the nameConf from DataFormats/Detectors/Common to DataFormats/common.

NameConf depends on DetID which is in the DataFormats/Detectors/Common and which in turn depends on the  DataFormats/common. We cannot move NameConf there w/o major refactoring...

davidrohr:
I am still getting errors, also for ZDC (ran 1 minute ago), these objects are missing:
```
Requested resource does not exist: http://alice-ccdb.cern.ch//ZDC/Config/Module/0/
Requested resource does not exist: http://alice-ccdb.cern.ch//CTP/Config/1638438988548/
Requested resource does not exist: http://alice-ccdb.cern.ch//CPV/Calib/Gains/1638438995255/
Requested resource does not exist: http://alice-ccdb.cern.ch/download/dcbe56d0-3aa1-11ec-9661-0aa202a21b9a
Requested resource does not exist: http://alice-ccdb.cern.ch/download/61a99a90-3aa6-11ec-9661-0aa202a21b9a
```
I don't fully understand what is the last one? It is requested by CPV.

davidrohr:
> > Although, perhaps I would move the nameConf from DataFormats/Detectors/Common to DataFormats/common.
> 
> NameConf depends on DetID which is in the DataFormats/Detectors/Common and which in turn depends on the DataFormats/common. We cannot move NameConf there w/o major refactoring...

Well, it is only used only in some helper functions to get the detector name. Let me see whether I can find a clever way to move it to `DataFormats/common` without much refactoring.

davidrohr:
Here is an attempt to move NameConf: https://github.com/davidrohr/AliceO2/commit/8a59520ab8c191cee3d432d62e28cf3f318f771e
I had to move it to CommonUtils, for StringUtils and ConfigurableParam.
Not fully working yet, will fix the remaining issues later.

davidrohr:
I have opened #7798, which includes this one.
I wanted to open the PR against `shahor02:pr_fit_ccdb` but that doesn't work since `pr_fit_ccdb` is not on `dev`, and meanwhile there were commits in dev that needed adaptation after I have moved the NameConf class to CommonUtils.

mfasDa (approved):

PR #7779, opened 2021-12-01
Allow user to define output tables list

nburmaso:
This is an attempt to segregate output tables by allowing a user to choose what to store in AOD.
Summary:
*Names and descriptions of tables needed for output are defined in `static constexpr` arrays
*A comma-separated list of table names can be passed with `--fill-tables` flag
*The list is parsed in `parseTablesList()` and an array of corresponding `bool` flags is set respectively
*Due to some dependencies between tables, sanity checks are made in the `checkTablesDeps()` and in some places in the `process()`
*Tables are bound to output at the end of `process()` to avoid writing empty (and not requested) tables

This implementation is somewhat clumsy, so comments and suggestions are very appreciated


nburmaso:
Hi @sawenzel, @jgrosseo, @mfasDa. These changes address O2-2690 and O2-2704. Could you please take a look at the code and see what should be changed/improved?

jgrosseo:
@aalkin @ktf 
There is lots of duplication of names in this PR as pointed out by @nburmaso 
My understanding is that in the AOD reader/writer we derive the tree name from the table name in the data model. Couldn't it be done the same here. What do you think?

aalkin:
In the reader we do determine the tree name by adding O2 to the requested table name, but these are not checked against any list. In the rest of the framework we also do not keep any "database" of tables, since we use types to identify them. I do not see how we can avoid this duplication in a case where we need to have a list to check against.

jgrosseo:
Thanks Anton.
In this PR we have three lists: enum, string list, table name. So from what I understand from you, we could reduce it to one list at least.
Even better if we can get the table name `o2bc` from `o2::aod::BCs` (@aalkin could you advise how?) then we could check without any list if `o2bc` is included in `tablesList`.
What do you think?

aalkin:
For the predefined types we have metadata template, one can retrieve the label with, for example `aod::MetadataTrait<aod::BCs>::metadata::tableLabel()`. But this is when starting from the type, as far as I understand here we start from input, so there is no workaround, there needs to be a list.

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
```diff
@@ -199,6 +201,93 @@ class AODProducerWorkflowDPL : public Task
...
+    O2mcparticle,
+    O2mctracklabel,
+    O2mfttrack,
+    O2track,
+    O2trackcov,
+    O2trackextra,
+    O2v0,
+    O2zdc,
+    O2caloCell,
+    O2caloCellTRGR,
```
mfasDa, line 230: Should be renamed to `O2caloTrigger`.

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
```diff
@@ -199,6 +201,93 @@ class AODProducerWorkflowDPL : public Task
...
+    "O2mcparticle",
+    "O2mctracklabel",
+    "O2mfttrack",
+    "O2track",
+    "O2trackcov",
+    "O2trackextra",
+    "O2v0",
+    "O2zdc",
+    "O2caloCell",
+    "O2caloCellTRGR"};
```
mfasDa, line 258: Should be renamed to `O2caloTrigger`.

mfasDa (changes requested):
I think this could work for supporting only a subset of detectors present, however there might be some details I overlooked.

One comment from my side is that the enum values for the calorimeter trigger information has the tag "cellTRGR", which is confusing and should be renamed. What we understand as "TRGR" is a kind of header object referred to as TriggerRecord we propagate through the reconstruction stage in order to mark ranges in various containers (cells, triggers, clusters) belonging to the same BC, and add some information from the RDH. It is adapted from the ROFRecord from ITS, however as EMCAL is not continuous but triggered has been renamed as it contains header information belonging rather to a trigger than a time range. It has itself nothing to do with the trigger information we provide but rather the triggers we receive. It is used for iterating over the bunch crossings but is itself not propagated as the AOD is already using the BC for linking columns. Therefore the value should be renamed. Also "cell" for the trigger table is not appropriate as the module in trigger terminology has a different name in order to avoid confusion with the cell itself.

PR #7780, opened 2021-12-01
DPL Analysis: event mixing: allow for creating empty combinations

saganatt:
First prerequisite for event mixing - allowing empty combinations, so they can be defined outside `process()`.
Simplified some template parameters.

### DIFF
--- a/Framework/Core/include/Framework/ASoAHelpers.h
+++ b/Framework/Core/include/Framework/ASoAHelpers.h
@@ -787,7 +787,7 @@ struct CombinationsBlockUpperSameIndexPolicy : public CombinationsBlockSameIndex
       setRanges();
     }
   }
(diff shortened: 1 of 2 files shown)


aalkin:
You probably also want a version with consuming `setTables()`, so you can give temporary objects to the combinations constructor, which may be useful in some contexts.

saganatt:
By consuming, do you mean with `Ts&&... tables`?

saganatt:
I don't see for now the application of setTables() besides AnalysisTask, where I get just table references.
We have now constructors for temporaries, don't we? When we do not create empty combinations first.

aalkin:
If we want to have ad-hoc filtering, we'll need to create temporary filtered objects for combinations, even in the analysis task. Same goes for combing filtered: while framework-managed filters have their selections kept by the task, using intersection or union would create a filtered object with selection kept in its internal cache.

saganatt:
Ok, I added these functions, can you look at them @aalkin ?
I am building locally with sanitizer, I will see if there will be anything else to correct.

saganatt:
Update: sanitizer complains about TableBuilder, but it seems it is OK with the combinations themselves.

jgrosseo (approved):

PR #7781, opened 2021-12-01
DPL Analysis: event mixing: Move GroupSlicer to a separate file, add operator+ and +=

saganatt:
Moved GroupSlicer to avoid circular dependencies later in event mixing code.


Framework/Core/include/Framework/AnalysisTask.h
```diff
@@ -21,6 +21,7 @@
 #include "Framework/Expressions.h"
 #include "Framework/ExpressionHelpers.h"
 #include "Framework/EndOfStreamContext.h"
+#include "Framework/GroupSlicer.h"
```
ktf, line 24: Do you really need to include it by default? Can't we include it only when needed?

Framework/Core/include/Framework/GroupSlicer.h
```diff
@@ -0,0 +1,362 @@
...
+      GroupSlicerIterator copy = *this;
+      copy.position += inc;
+      copy.mGroupingElement = copy.mGroupingElement + inc;
+      return copy;
+    }
+
+    GroupSlicerIterator operator+=(int64_t inc)
+    {
+      position += inc;
+      mGroupingElement = mGroupingElement + inc;
```
aalkin: Shouldn't it be `mGroupingElement += inc;`?

Framework/Core/include/Framework/AnalysisTask.h
saganatt: It is used in this file, for grouping for analysis task.

jgrosseo (approved):

jgrosseo:
Needs forced sync merge with O2Physics after local test (ongoing)

PR #7782, opened 2021-12-01
Use TOF Utils to subtract BC time to tofSignal

njacazio:
@noferini please have a look

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -53,6 +53,7 @@
 #include "SimulationDataFormat/MCTrack.h"
 #include "SimulationDataFormat/MCTruthContainer.h"
 #include "GPUTPCGMMergedTrackHit.h"
+#include "TOFBase/Utils.h"
 #include "O2Version.h"
 #include "TMath.h"
 #include "MathUtils/Utils.h"
@@ -301,6 +302,7 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
           // normal tracks table
           auto contributorsGID = data.getSingleDetectorRefs(trackIndex);
           const auto& trackPar = data.getTrackParam(trackIndex);
+          extraInfoHolder.flags |= trackPar.getPID() << 28;
           if (contributorsGID[GIndex::Source::ITS].isIndexSet()) {
             int nClusters = itsTracks[contributorsGID[GIndex::ITS].getIndex()].getNClusters();
             float chi2 = itsTracks[contributorsGID[GIndex::ITS].getIndex()].getChi2();
@@ -326,10 +328,9 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
             extraInfoHolder.tofChi2 = tofMatch.getChi2();
             const auto& tofInt = tofMatch.getLTIntegralOut();
             const float intLen = tofInt.getL();
+            double timeOrbit = static_cast<int>(tofMatch.getSignal() / o2::constants::lhc::LHCOrbitNS * 1E-3) * o2::constants::lhc::LHCOrbitNS * 1E3;
+            const float tofSignal = static_cast<float>(o2::tof::Utils::subtractInteractionBC(tofMatch.getSignal() - timeOrbit));
             extraInfoHolder.length = intLen;
-            if (interactionTime > 0) {
-              extraInfoHolder.tofSignal = static_cast<float>(tofMatch.getSignal() - interactionTime * 1E3);
-            }
             const float mass = o2::constants::physics::MassPionCharged; // default pid = pion
             if (tofInt.getTOF(o2::track::PID::Pion) > 0.f) {
               const float expBeta = (intLen / (tofInt.getTOF(o2::track::PID::Pion) * cSpeed));
@@ -339,8 +340,8 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
             const double massZ = o2::track::PID::getMass2Z(trackPar.getPID());
             const double energy = sqrt((massZ * massZ) + (extraInfoHolder.tofExpMom * extraInfoHolder.tofExpMom));
             const double exp = extraInfoHolder.length * energy / (cSpeed * extraInfoHolder.tofExpMom);
-            extraInfoHolder.trackTime = static_cast<float>((tofMatch.getSignal() - exp) * 1e-3 - interactionTime); // tof time in \mus, FIXME: account for time of flight to R TOF
-            extraInfoHolder.trackTimeRes = 200e-3;                                                                 // FIXME: calculate actual resolution (if possible?)
+            extraInfoHolder.trackTime = static_cast<float>((tofSignal - exp) * 1e-3); // tof time in ns, taken from the definition of Ruben scaled by 1000 to convert from mus to ns to match the collisionTime
+            extraInfoHolder.trackTimeRes = 200e-3;                                    // FIXME: calculate actual resolution (if possible?)
           }
           if (src == GIndex::Source::TPCTRD || src == GIndex::Source::ITSTPCTRD) {
             const auto& trdOrig = data.getTrack<o2::trd::TrackTRD>(src, contributorsGID[src].getIndex());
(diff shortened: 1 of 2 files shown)


noferini (dismissed):
Ok, from my side.
@chiarazampolli: we couldn't test with aod-producer (we don't know how to do)

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -339,8 +340,9 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
             const double massZ = o2::track::PID::getMass2Z(trackPar.getPID());
             const double energy = sqrt((massZ * massZ) + (extraInfoHolder.tofExpMom * extraInfoHolder.tofExpMom));
             const double exp = extraInfoHolder.length * energy / (cSpeed * extraInfoHolder.tofExpMom);
-            extraInfoHolder.trackTime = static_cast<float>((tofMatch.getSignal() - exp) * 1e-3 - interactionTime); // tof time in \mus, FIXME: account for time of flight to R TOF
-            extraInfoHolder.trackTimeRes = 200e-3;                                                                 // FIXME: calculate actual resolution (if possible?)
+            // extraInfoHolder.trackTime = static_cast<float>((tofMatch.getSignal() - exp) * 1e-3 - interactionTime); // tof time in \mus, FIXME: account for time of flight to R TOF
```
jgrosseo: Why has the unit changed by 3 orders of magnitude and the comment stays the same: ` \mus` ?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -301,6 +302,7 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
           // normal tracks table
           auto contributorsGID = data.getSingleDetectorRefs(trackIndex);
           const auto& trackPar = data.getTrackParam(trackIndex);
+          extraInfoHolder.flags |= trackPar.getPID() << 28;
```
jgrosseo, line 305: Good catch! :)

njacazio: It was indeed affecting the TOF signal

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
njacazio: This is a fix, the time was stored in `ns` before and the comment was inconsistent. Now we store both trackTime and trackTimeRes with the same units

noferini: in the comment you still have mus. Could you write "ns"? I guess this was the comment of @jgrosseo

noferini: Sorry, you move to mus everything (I was mislead by the part commented)

noferini (dismissed):
I re-generated new AO2D with the last fixes and now everything looks fine.

Let remember to fix in O2Physics this

 --- a/Common/Core/PID/PIDTOF.h 
 +++ b/Common/Core/PID/PIDTOF.h
 @@ -227,7 +227,7 @@ class TOFSignal
   /// Computes the expected time of a track, given its TOF expected time
   /// \param trackTime trackTime in ns
   /// \param response response to use to compute the expected time for the propagation from the vertex to TOF
// -  static float ComputeTOFSignal(const float& trackTime, const float& expTime) { return trackTime * 1e+3 + expTime; }
// +  static float ComputeTOFSignal(const float& trackTime, const float& expTime) { return trackTime * 1e+6 + expTime; }

njacazio:
Thanks @noferini I forgot this, I fixed the comments and am ready to merge!

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -339,8 +340,8 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
             const double massZ = o2::track::PID::getMass2Z(trackPar.getPID());
             const double energy = sqrt((massZ * massZ) + (extraInfoHolder.tofExpMom * extraInfoHolder.tofExpMom));
             const double exp = extraInfoHolder.length * energy / (cSpeed * extraInfoHolder.tofExpMom);
-            extraInfoHolder.trackTime = static_cast<float>((tofMatch.getSignal() - exp) * 1e-3 - interactionTime); // tof time in \mus, FIXME: account for time of flight to R TOF
-            extraInfoHolder.trackTimeRes = 200e-3;                                                                 // FIXME: calculate actual resolution (if possible?)
+            extraInfoHolder.trackTime = static_cast<float>((tofSignal - exp) * 1e-3); // tof time in ns, taken from the definition of Ruben scaled by 1000 to convert from mus to ns to match the collisionTime
+            extraInfoHolder.trackTimeRes = 200e-3;                                    // FIXME: calculate actual resolution (if possible?)
```
jgrosseo, line 344: @shahor02 I am in favour of merging this. Unrelated: Don't we have an estimate of the track type already somewhere?

jgrosseo (approved):

njacazio:
@shahor02 if you have no objections I would merge this

shahor02 (approved):

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: @jgrosseo sorry, just saw the question. If you are asking whether we do reco-time PID: no, for that the TPC would need to provide a dedx -> PID algorithm.

jgrosseo: @shahor02 I wrote track type but mean track time resolution (sorry for this; and that's why I made the comment here). 
So do you already have a track time resolution (even course based on the type of track)?

shahor02: Not for ITS-TPC-TOF tracks: they are somewhat special since their time is given by the TOF signal stored in the match info (while the kinematic comes from ITS-TPC). In the https://github.com/AliceO2Group/AliceO2/blob/ce3e7457c3e9794d25d47ef156417b29f8a138a7/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L169-L170 a 10 ns error is assigned (which is surely an overkill but is much more precise that the ITS-TPC time errors), but AOD producer does not use this method, since it relias on the vertex->tracks attachments tables.

jgrosseo: I think we should propagate these values into the AOD (even if you mark them as "overkill" for now). All this info should be available in the AOD converter, so it just requires adjusting the code, right?

shahor02: Yes, but I would do this in the opposite way: if TOF could provide a method to get an error for given {TOFCluster, track, LIntegral}, which could be used both in the RecoContainerCreateVariadic::create(...) and for the tracks directly accessed in the RecoContainer (like in the AODProducer, which would store it).

jgrosseo: * OK, who we should ping for this?
* In addition, how do we fill `trackTimeRes` for non TOF tracks?

noferini: What kind of TOF error do you mean? The time resolution for PID?

shahor02: @jgrosseo for tracks with TRD the error corresponds to the nominal triggered BC duration (i.e. a few 100 ps), for ITS-TPC tracks: uncertainty in matching point Z / vdrift.

@noferini : essentially, the uncertainty on the expected time.

noferini: For uncertainties on the expected times we need a tuned parameterization consistent with data and MC.
In run-2 it was encoded in a base class + OADB. I guess we will do the same for Run-3 (OADB -> CCDB).
If I'm not wrong @njacazio started to work on that

jgrosseo: TrackTime contains here the track time estimate based on TOF (absolute wrt BC). We do *not* talk about the estimated time with a momentum and particle hypothesis. Do we talk about the same thing?

jgrosseo: > @jgrosseo for tracks with TRD the error corresponds to the nominal triggered BC duration (i.e. a few 100 ps), for ITS-TPC tracks: uncertainty in matching point Z / vdrift.

Can we add this to the AOD producer already?

noferini: I see. A deltat (t_tof - t_track) is avaliable with the matching info.
For the resolution we depend on the track time precision which is available with the track.
Is this what you need?

shahor02: @jgrosseo 

> We do not talk about the estimated time with a momentum and particle hypothesis. Do we talk about the same thing?

Does this matter if we speak about the uncertainty? 

> Can we add this to the AOD producer already?

Do you mean to fill `extraInfoHolder.trackTime` and `extraInfoHolder.trackTimeRes` for all tracks types?

jgrosseo: > Do you mean to fill extraInfoHolder.trackTime and extraInfoHolder.trackTimeRes for all tracks types?

Yes!

PR #7783, opened 2021-12-01
[Global Fwd Tracking] Improve configuration of MFTMCH matching

rpezzi:
Allows external definition of MFTMCH pair matching score.


shahor02 (approved):
@rpezzi looks fine, but eventually the functions should be loaded not from the disk file but from the CCDB.

shahor02 (approved):

PR #7785, opened 2021-12-01
[EMCAL-645] Make cell index range in event handler const

mfasDa:


mfasDa:
@shahor02  the CI error unrelated. Also ready for merge. Needed to subscribe directly to cell index container from input spec, which is of type gsl::span<const int>.

shahor02 (approved):

PR #7787, opened 2021-12-01
Suppress unneeded clusters request by AOD producer

shahor02:


chiarazampolli:
Thanks, Ruben!

shahor02:
Hi @chiarazampolli 

Unfortunately, I was wrong: forgot that the TPC clusters are needed to count the shared clusters in the AOD producer.
So, we need them...

Cheers,
 Ruben

chiarazampolli:
Ok, thank you!

Chiara

PR #7793, opened 2021-12-02
Add TRD tracking QC (for expert analysis)

martenole:
This adds the possibility to add detailed QC to the global TRD tracking. A QC module should come in the future, which would also rely on the `Tracking` class (and then produce histograms instead of a large tree). This for now is more for debugging purposes and parameter tuning of the matching cuts etc. It would probably not be enabled in production workflows.

### DIFF
--- a/Detectors/TRD/qc/include/TRDQC/Tracking.h
+++ b/Detectors/TRD/qc/include/TRDQC/Tracking.h
@@ -1,0 +1,117 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file Tracking.h
+/// \brief Check the performance of the TRD in global tracking
+/// \author Ole Schmidt
+
+#ifndef O2_TRD_TRACKINGQC_H
+#define O2_TRD_TRACKINGQC_H
+
+#include "DataFormatsTRD/TrackTRD.h"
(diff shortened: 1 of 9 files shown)


Detectors/TRD/qc/include/TRDQC/Tracking.h
```diff
@@ -0,0 +1,113 @@
...
+  int type; // 0 TPC-TRD track; 1 ITS-TPC-TRD track
+  int nTracklets;
+  float chi2;
+  float reducedChi2;
+  float pt;
+  std::array<float, constants::NLAYER> findable{};
+  std::array<float, constants::NLAYER> trackX{};
+  std::array<float, constants::NLAYER> trackY{};
+  std::array<float, constants::NLAYER> trackZ{};
+  std::array<float, constants::NLAYER> trackSnp{};
```
tdietel: Can you add a brief comment what Snp is? And maybe what coordinate system is used for the helix: global / tracking=sector / chamber?

Detectors/TRD/qc/include/TRDQC/Tracking.h
```diff
@@ -0,0 +1,113 @@
...
+  std::array<float, constants::NLAYER> trackTgl{};
+  std::array<float, constants::NLAYER> trackQpt{};
+  std::array<float, constants::NLAYER> trackletYraw{};
+  std::array<float, constants::NLAYER> trackletZraw{};
+  std::array<float, constants::NLAYER> trackletY{};
+  std::array<float, constants::NLAYER> trackletZ{};
+  std::array<float, constants::NLAYER> trackletDy{};
+  std::array<float, constants::NLAYER> trackletSlope{};
```
tdietel: Can you add a comment what is slope, and what is dy?

tdietel (dismissed):
Looks good.

I want to look into residuals at some point as well. It would be good to add a few comments as indicated to make it easier to use.

martenole:
Hi @tdietel 
thanks for the review. I have added comments to all members of the QC struct. Could you please approve again? Then I could merge this as soon as the CI is passed.
Cheers,
Ole

tdietel (approved):
Ahh, snp is sin(phi).
Thanks for the comments, approved.

PR #7796, opened 2021-12-02
DPL Analysis: make sure all supported types can be used in vector columns

aalkin:


jgrosseo (approved):

Framework/Core/include/Framework/ArrowTypes.h
```diff
@@ -77,21 +77,26 @@ struct arrow_array_for<double[N]> {
...
+  template <>                                   \
+  struct arrow_array_for<std::vector<_type_>> { \
+    using type = arrow::ListArray;              \
+    using value_type = _type_;                  \
+  };
+
+ARROW_VECTOR_FOR(uint8_t);
+ARROW_VECTOR_FOR(uint16_t);
+ARROW_VECTOR_FOR(uint32_t);
+ARROW_VECTOR_FOR(uint64_t);
```
ktf, line 91: I think this might need to use the ROOT type. Otherwise  we end up with the usual problem that uint64_t is different types on different platform...

aalkin: That is handled by TableBuilder, using stuff in `RootTableBuilderHelpers.h`. It will do remapping when needed.

PR #7797, opened 2021-12-02
DPL Analysis: cleanups in TreeToTable code

aalkin:


ktf:
Please fix the conflict.

Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -228,13 +228,11 @@ std::pair<std::shared_ptr<arrow::ChunkedArray>, std::shared_ptr<arrow::Field>> B
     std::unique_ptr<TBufferFile> offsetBuffer;
 
     uint32_t offset = 0;
-    uint32_t lastOffset = offset;
+    uint32_t lastOffset;
```
ktf, line 231: Why not initialising?

ktf:
Tested on hyperloop. I am not sure the idea of leaving some variable uninitialised is actually a good idea, but we can fix it separately.

Framework/Core/src/TableTreeHelpers.cxx
aalkin: It is redundant, as it is always reset in a loop, before it is used, and if loop is not executed it is not used.

ktf: It's  redundant until we change the code and then we an hard to spot random memory access. I would always initialise  things.

PR #7801, opened 2021-12-03
print branch type in exception

jgrosseo:


Framework/Core/src/TableTreeHelpers.cxx
```diff
@@ -61,7 +61,7 @@ auto arrowTypeFromROOT(EDataType type, int size)
     case EDataType::kDouble_t:
       return typeGenerator(arrow::float64(), size);
     default:
-      throw runtime_error("Unsupported branch type");
+      throw std::runtime_error(fmt::format("Unsupported branch type: {}", static_cast<int>(type)));
```
ktf, line 64: Please use `runtime_error_f("Unsupported branch type: %d", type)`.

PR #7805, opened 2021-12-03
TRD calibration classes cleanup

martenole:
Hi Sean, Tom, Jorge,
This is quite a large cleanup of the TRD calibration classes (mostly CCDB/OCDB related). There were quite some objects which were copied over from Run 2 OCDB, but which were not used anywhere in the code. I think it would make sense to remove those and only add them when they are also taken into account by either reconstruction or simulation.
I want to move the remaining parts to `DataFormatsTRD` and thus removed the dependencies from geometry etc were it was not needed. But that I would do in a follow up PR. I already put new objects in the test-ccdb and tested that the simulation still works with these changes.
Please let me know what you think. If you would prefer to keep some of the stuff in then let me know.
Cheers,
Ole


Detectors/TRD/base/include/TRDBase/PadCalibrations.h
```diff
@@ -57,33 +51,15 @@ PadCalibrations<T>::PadCalibrations()
...
-template <class T>
-void PadCalibrations<T>::reset(int roc, int col, int row, std::vector<T>& data)
-{
-  //reset the readoutchamber
-  //primarily here for setting the values incoming from run2 ocdb, but might find other use cases.
-  // you need to send it the roc as it is used to calculate internal parameters.
-  mreadOutChamber[roc].reset(roc, col, row, data); // it *should* not actually matter as this *should* be set correctly via init.
-}
+// instantiate templates always needed
+template class PadCalibrations<float>; // for LocalT0, LocalVDrift, LocalGainFactor, PadNoise
```
tdietel, line 61: I don't think we will have local, i.e. pad-by-pad, T0 or local Vdrift calibrations.

martenole: Probably not, but the template instantiation will be needed anyhow for the pad-wise gain factor (`LocalGainFactor`)

bazinski:
These were always just a semi hack to pull in the stuff from run2 to get the simulations running. Its good they getting redone/deleted.

tdietel (dismissed):

martenole:
@davidrohr I kicked the CI again. Let's just see if this goes through (I believe it should...) and afterwards merge it

Detectors/TRD/base/include/TRDBase/PadCalibrations.h
martenole: @tdietel I checked and at the moment the values we are using for LocalT0 and LocalVDrift are 1 and 0 resp., so they don't have any effect. So I would indeed remove them now completely. Do you agree?

PR #7806, opened 2021-12-03
Changing indices of multipart-messages or it won't work

chiarazampolli:


chiarazampolli:
@noferini , @shahor02

PR #7813, opened 2021-12-05
Event Display is now capable of displaying tracks from AOD files

pnwkw:
The current version of Event Display renders track JSON files produced by a custom workflow running on EPNs as part of the "online" functionality.

I have created a workflow which can create these JSON files locally from AODs.
This workflow is executed by the Event Display itself on-the-fly when the "offline" mode is selected and an AOD file is provided.

This is not a final solution to the problem, however due to some open questions regarding the DPL Framework (i.e. "random access" support to collisions/tracks), I consider it to be a good attempt for marrying the ED and the DPL together in their current state of development.

To view an AOD file in the Event Display, execute it in the following way:

```
o2-eve --filename AO2D.root
```

Screenshot from 2021-12-06 14-43-07


pzhristov:
@pnwkw Would you put some short description of the current state and the planned development?
@aalkin Would you also review these changes?

EventVisualisation/Workflow/src/AO2DConverter.cxx
```diff
@@ -0,0 +1,63 @@
...
+  LOG(info) << "------------------------    AO2DConverter::init version " << mWorkflowVersion << "    ------------------------------------";
+  mData.init();
+}
+
+void AO2DConverter::process(o2::aod::Collisions const& collisions, EveWorkflowHelper::AODFullTracks const& tracks)
+{
+  for (auto const& c : collisions) {
+    auto const tracksCol = tracks.sliceBy(aod::track::collisionId, c.globalIndex());
+
+    EveWorkflowHelper helper;
```
aalkin: Should it be loop-local? Can you initialize it once outside the loop?

EventVisualisation/Workflow/src/AO2DConverter.cxx
```diff
@@ -0,0 +1,63 @@
...
+      helper.drawAOD(track, c.collisionTime());
+    }
+
+    helper.save(jsonPath, collisions.size(), GlobalTrackID::MASK_ALL, GlobalTrackID::MASK_NONE, mWorkflowVersion);
+  }
+}
+
+WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
+{
+  LOG(info) << "------------------------    defineDataProcessing " << AO2DConverter::mWorkflowVersion << "    ------------------------------------";
```
aalkin: Please avoid output within this function, it will break combining workflows by piping through stdin.

EventVisualisation/Workflow/src/AO2DConverter.cxx
pnwkw: This requires modifying the `EveWorkflowHelper` to support clearing its current list of tracks. I'll do it.

EventVisualisation/Workflow/src/AO2DConverter.cxx
pnwkw: Should I remove it from `O2DPLDisplay.cxx` as well?

EventVisualisation/View/src/Initializer.cxx
```diff
@@ -56,11 +57,17 @@ void Initializer::setup()
...
-  //}
+  if (options->online()) {
+    eventManager.setDataSource(new DataSourceOnline(options->dataFolder()));
+  } else {
+    eventManager.setDataSource(new DataSourceOffline(options->AODConverterPath(), options->dataFolder(), options->fileName(), options->hideDplGUI()));
+  }
+
+  // if (Options::Instance()->json()) {
+  //   eventManager.getDataSource()->registerDetectorX(new DataReaderJSON(nullptr), EVisualisationGroup::JSON);
+  // }
```
ktf: Please remove commented code.

ktf:
It looks nice, thanks. Apart from a few cosmetic  fixes, given it's really fully separate from DPL itself, I guess we can simply go ahead and merge it and do bug fixes afterwards.

ktf:
Could you add a picture as well?

EventVisualisation/View/src/Initializer.cxx
```diff
@@ -56,11 +57,13 @@ void Initializer::setup()
   auto& eventManager = EventManager::getInstance();
   eventManager.setCdbPath(ocdbStorage);
```
shahor02, line 58: The storage `settings.GetValue("OCDB.default.path", "local://$ALICE_ROOT/OCDB");` seems to be Aliroot legacy.
Are you using env file for anything else that this path? If yes, then please set the default to `o2::base::NameConf::getCCDBServer()`, if not, then you can used directly `o2::base::NameConf::getCCDBServer()`.

EventVisualisation/Workflow/src/DetectorData.cxx
```diff
@@ -0,0 +1,79 @@
...
+using namespace o2::tpc;
+using namespace o2::trd;
+
+void DetectorData::init()
+{
+  const auto grp = o2::parameters::GRPObject::loadFrom();
+  o2::base::GeometryManager::loadGeometry();
+  o2::base::Propagator::initFieldFromGRP();
+  mConfig.reset(new EveConfiguration);
+  mConfig->configGRP.solenoidBz = 5.00668f * grp->getL3Current() / 30000.;
```
shahor02: you can use here `o2::base::Propagator::Instancer()->getNominalField()`.

EventVisualisation/View/src/Initializer.cxx
pnwkw: The env file is used currently for some o2Eve settings (fonts, colors), but I think @jmyrcha planned to refactor it out of the code at some point. I have put `o2::base::NameConf::getCCDBServer()` as the default for now.

shahor02 (approved):

PR #7815, opened 2021-12-05
DPL Analysis: use internal selection cache for partitions

aalkin:
* Use internal selection cache for partitions
* Use `std::make_unique` to prevent memory leak
* Remove redundant index binding (we never bind indices to partitions)

### DIFF
--- a/Framework/Core/include/Framework/AnalysisHelpers.h
+++ b/Framework/Core/include/Framework/AnalysisHelpers.h
@@ -492,30 +492,25 @@ struct Service {
 };
 
 template <typename T>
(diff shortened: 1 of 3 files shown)


aalkin:
@ktf could you please test this on hyperloop and merge if it is successful? We need this for tomorrow tutorials, since currently partitions are broken.

ktf:
Sorry the build was failing for unrelated reasons. I am checking it right now.

ktf:
Tested on hyperloop. Merging. Do you need a daily build?

PR #7816, opened 2021-12-06
Fix logic to print TGrid connect error msg

sawenzel:


davidrohr:
trivial and currently broken, merging

PR #7817, opened 2021-12-06
jobutils improvements

sawenzel:
version 2 fixing some performance issues:
* instead of actively monitoring logs every N-s, we now
  use an asynchronous background task using the "tail -f | grep"
  pattern

Committed as a separate script for the moment in order
to test this in parallel to existing solution. A complete
switch-over will be made once tests are successful.


sawenzel:
untested script not used in production; will be tested locally before being made operational.

PR #7819, opened 2021-12-06
TRD merge CommonParam into SimParam and more

martenole:
- singleton implementation not required here and thus removed
- shifted code relevant only to simulations from TRDBase into TRDSimulation
- added missing ClassDefs
- hoping to achieve better readability and maintainability of the code


Detectors/TRD/simulation/include/TRDSimulation/SimParam.h
```diff
@@ -111,50 +127,9 @@ class SimParam
...
 
- private:
-  /// This is a singleton, constructor is private!
-  SimParam();
-  ~SimParam() = default;
-
-  /// Fill the arrays mTRDsmp and mCTsmp for the given gas mixture
-  void sampleTRF();
-
   ClassDefNV(SimParam, 1); // The TRD simulation parameters
```
sawenzel: I guess you'll need to increment the version as well

martenole: Yes that makes sense. In principle though this is only important if we have stored the SimParam in a file somewhere and would like to open it again, right? Because otherwise ROOT complains about a different class layout? I don't think we do have it in a file somewhere, but I will change it nevertheless to be sure.

martenole:
Hi @tdietel and @jolopezl 
do you agree with the changes here? Would it be okay for you if we merge this?
Performance is unaffected, the digitization produces the same results.
Cheers,
Ole

jolopezl (approved):
Hi @martenole, I find this change very nice and welcome. The only thing that caught my attention while looking at the code changes is that there are two classes with similar names: `TRDSimParams` and `SimParam`. So I would propose the following renaming: `TRDSimParams` to `SimConfigParams`, and `SimParam` to `SimParams`. But this shouldn't stop getting these changes merged.

tdietel (approved):
Looks ok.
But we really need to think about documenting what settings are where...

martenole:
Thanks a lot for your comments, then I'll merge this now as it is. The `TRDSimParams` are what we can change from the command line and the `SimParam` class holds some parameters which can currently only be changed by recompiling the code. In that sense they are rather constants. I mean there are setters in that class, but they are not used anywhere else in O2 (at least the ones I checked).
If we decide to make them configurable at some point those can be moved into `TRDSimParams` or the digitizer can get a command line option for them. Of course, if we conclude that the TRD will not be operated with Ar at some point in the future one could also make a lot of things just constants. :)

I was also thinking about the documentation. We do have the AliRoot software writeup in `Detectors/TRD/doc/` but this is to a large extend out of date. I think it would be good, although also quite some work, to transfer some of the key information into README files in O2, for example for simulation and reconstruction. And then maybe remove the `doc` folder from O2. The AliRoot documentation is still available in AliRoot.

PR #7820, opened 2021-12-06
AOD: also keep particles marked as keepPhysics

sawenzel:
This should fix the K0 -> pipi branching ratio analysis. Now also copying all particles
which were originally marked as "keepPhysics" in the MC stack treatment (this seems to be needed
in addition to isPhysicalPrimary()).


sawenzel:
@aimeric-landou @amorsch

PR #7822, opened 2021-12-07
TOF full access to ccdb calibration and anchoring for MC

noferini:


noferini:
added @chiarazampolli 

open the PR to produce and use all TOF calibration ccdb objects (both in reco and sim).

For simulation I still passing a timestamp via command line (it will be replace as soon as there will be provided centrally)

DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
```diff
@@ -60,15 +60,19 @@ class Diagnostic
...
     return iter->first;
   }
-  int getSlot(ULong64_t pattern) const;
-  int getCrate(ULong64_t pattern) const;
-  int getChannel(ULong64_t pattern) const;
-  int getNoisyLevel(ULong64_t pattern) const;
+  static int getSlot(ULong64_t pattern);
+  static int getCrate(ULong64_t pattern);
+  static int getChannel(ULong64_t pattern);
+  static int getNoisyLevel(ULong64_t pattern);
```
shahor02: better to inline these simple methods.

DataFormats/Detectors/TOF/include/DataFormatsTOF/TOFFEElightInfo.h
```diff
@@ -0,0 +1,72 @@
...
+  std::array<uint64_t, NTRIGGERMAPS> mTriggerMask; // trigger mask, can it be uint32_t?
+  TOFFEElightInfo()
+  {
+    mVersion = -1;
+    mRunNumber = -1;
+    mRunType = -1;
+    mChannelEnabled.fill(false);
+    mMatchingWindow.fill(0);
+    mLatencyWindow.fill(0);
+    mTriggerMask.fill(0);
```
shahor02: you don't need to call fill(0) in the default constructor, better define them with default initialer, i.e.
```
std::array<bool, NCHANNELS> mChannelEnabled{};
...
```
Also, since you already defined e.g.   `int mVersion = -1;   // version`, you don't need to do the same assignment in the constructor.

Detectors/GlobalTrackingWorkflow/src/TOFMatcherSpec.cxx
```diff
@@ -102,6 +102,8 @@ void TOFMatcherSpec::run(ProcessingContext& pc)
   RecoContainer recoData;
   recoData.collectData(pc, *mDataRequest.get());
 
+  auto creationTime = DataRefUtils::getHeader<DataProcessingHeader*>(pc.inputs().getFirstValid(true))->creation;
```
shahor02, line 105: for the record: this is not guaranteed to be the exact time stamp of the TF, only approximate (the difference may reach seconds).

Detectors/TOF/base/src/CalibTOFapi.cxx
```diff
@@ -41,13 +44,42 @@ CalibTOFapi::CalibTOFapi(const std::string url)
 
 //______________________________________________________________________
 
+void CalibTOFapi::readActiveMap()
+{
+  // getting the active TOF map
+  memset(mIsOffCh, false, Geo::NCHANNELS);
+
+  auto& mgr = CcdbManager::instance();
+  long timems = long(mTimeStamp) * 1000;
```
shahor02, line 53: What are there units of the mTimeStamp? The query should be done in ms, so I guess the mTimeStamp is in seconds.
But then in the next method you use mTimeStamp directly to query the ccdb. One of these methods is wrong.

Detectors/TOF/base/src/CalibTOFapi.cxx
```diff
@@ -41,13 +44,42 @@ CalibTOFapi::CalibTOFapi(const std::string url)
...
 {
 
   // getting the LHCphase calibration
 
   auto& mgr = CcdbManager::instance();
-  mLHCphase = mgr.getForTimeStamp<LhcPhase>("TOF/Calib/LHCphase", mTimeStamp);
+  long timems = long(mTimeStamp) * 1000;
+  mLHCphase = mgr.getForTimeStamp<LhcPhase>("TOF/Calib/LHCphase", timems);
+  if (mLHCphase) {
+    LOG(info) << "read LHCphase for TOF " << mLHCphase->getLHCphase(mTimeStamp);
```
shahor02, line 79: here you query both with `mTimeStamp*1000` and with `mTimeStamp`. Why don't you simply set  the timestamp to the CCDBManager once and query with simple `get<...>`? Also, take into account that all these custom queries will become obsolete once we switch to DPL ccdb service.

Detectors/TOF/workflow/src/TOFClusterizerSpec.cxx
```diff
@@ -146,6 +170,28 @@ class TOFDPLClustererTask
     }
 
     for (int i = 0; i < row.size(); i++) {
```
shahor02: this will complain about comparison of signed and unsigned, use `unsigned i = 0...`

Detectors/GlobalTrackingWorkflow/src/TOFMatcherSpec.cxx
noferini: This I know (we are not so sensitive in our calibration).
However, this will be replaced... waiting for a standard way working both in sync and async phase.

Detectors/TOF/base/src/CalibTOFapi.cxx
noferini: I guess you referred to
' -  corr += mLHCphase->getLHCphase(int(mTimeStamp / 1000)); // timestamp that we use in LHCPhase is in seconds, but for CCDB we need it in ms'
'+  corr += mLHCphase->getLHCphase(mTimeStamp); // timestamp that we use in LHCPhase is in seconds, but for CCDB we need it in ms'

since LHCphase get it in second (fixed)

Detectors/TOF/base/src/CalibTOFapi.cxx
noferini: One call is to ccdb
The second one is to a LHCphase object which has an internal time structure in seconds

The first one with timems will be moved to ccdb service
The second one will remain here (I will discuss with @chiarazampolli if we still need this redundant structure)

DataFormats/Detectors/TOF/include/DataFormatsTOF/TOFFEElightInfo.h
noferini: mv to default constructor.

Detectors/TOF/base/src/CalibTOFapi.cxx
noferini: mTimeStamp -> seconds

Detectors/TOF/workflow/src/TOFClusterizerSpec.cxx
noferini: mvd to unsigned

Detectors/TOF/base/src/CalibTOFapi.cxx
shahor02: ok, did not notice that the 2 nd call is not a ccdb query.

DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h
noferini: done

Detectors/TOF/base/src/CalibTOFapi.cxx
noferini: Yes, we planned in the past to have an internal structure here.
However, we receently moved to use calibration timeslot to fill ccdb so this is probably an obsolete feature (but we will decide to keep one ccdb object per run with the LHCphase trending, this should be discussed).

shahor02 (approved):

PR #7824, opened 2021-12-07
print branch type in exception

jgrosseo:


jgrosseo:
@ktf I think you put auto merge but didn't approve...

PR #7825, opened 2021-12-07
o2-ccdb-downloadfile: Support for ALIEN resources

sawenzel:
o2-ccdb-downloadfile now works when the resource is on ALIEN.
This wasn't previously possible.

At the same time, this commit enables the use of the
local CCDB caching mechanism in conjunction with resources
on ALIEN.


sawenzel:
Private comments on an earlier version #7613 have been addressed. This is also an important fix for the CCDB_LOCALCACHE mechanism which is broken since 3 days ago.

PR #7826, opened 2021-12-07
DPL: support VariantType::Dict in merged workflows

ktf:
While it's still not serialised correctly, this should
allow a placeholder dict to be provide and then filled from
e.g. AliECS


ktf:
@Barthelemy

Barthelemy:
ok, I will test it

Barthelemy:
The crash is gone and it seems that I get the tuple. I did not check the content but that's already good.

PR #7827, opened 2021-12-07
DPL Analysis: Event mixing interface

saganatt:
Hi, this actually introduces the new event mixing. I will refresh the tutorial in O2Physics once this PR will get merged.

In the meantime, you can see the new interface [here](https://github.com/saganatt/O2Physics/blob/event-mixing/Tutorials/src/eventMixing.cxx#L130). It is possible to define it both outside and inside `process()`.

### DIFF
--- a/Framework/Core/include/Framework/AnalysisManagers.h
+++ b/Framework/Core/include/Framework/AnalysisManagers.h
@@ -12,6 +12,7 @@
 #ifndef FRAMEWORK_ANALYSISMANAGERS_H
 #define FRAMEWORK_ANALYSISMANAGERS_H
 #include "Framework/AnalysisHelpers.h"
+#include "Framework/GroupedCombinations.h"
 #include "Framework/Kernels.h"
 #include "Framework/ASoA.h"
 #include "Framework/ProcessingContext.h"
@@ -28,6 +29,29 @@
 
 namespace o2::framework
 {
+
+template <typename ANY>
+struct GroupedCombinationManager {
+  template <typename TG, typename... T2s>
+  static void setGroupedCombination(ANY&, TG&, T2s&...)
+  {
+  }
+};
+
+template <typename T1, typename GroupingPolicy, typename H, typename G, typename... Us, typename... As>
+struct GroupedCombinationManager<GroupedCombinationsGenerator<T1, GroupingPolicy, H, G, pack<Us...>, As...>> {
+  template <typename TH, typename TG, typename... T2s>
+  static void setGroupedCombination(GroupedCombinationsGenerator<T1, GroupingPolicy, H, G, pack<Us...>, As...>& comb, TH& hashes, TG& grouping, std::tuple<T2s...>& associated)
+  {
+    static_assert(sizeof...(T2s) > 0, "There must be associated tables in process() for a correct pair");
+    static_assert(!soa::is_soa_iterator_t<std::decay_t<H>>::value, "Only full tables can be in process(), no grouping");
+    if constexpr (std::conjunction_v<std::is_same<G, TG>, std::is_same<H, TH>>) {
+      // Take respective unique associated tables for grouping
+      auto associatedTuple = std::tuple<Us...>(std::get<Us>(associated)...);
+      comb.setTables(hashes, grouping, associatedTuple);
+    }
+  }
+};
 
 template <typename ANY>
 struct PartitionManager {
--- a/Framework/Core/include/Framework/GroupedCombinations.h
+++ b/Framework/Core/include/Framework/GroupedCombinations.h
@@ -1,0 +1,300 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef FRAMEWORK_PAIR_H
+#define FRAMEWORK_PAIR_H
+
+#include "Framework/ASoAHelpers.h"
+#include "Framework/GroupSlicer.h"
+#include "Framework/StringHelpers.h"
+
+#include <iostream>
+
+namespace o2::framework
+{
+
+// Define the type of a tuple interleaved from 2 tuples
+// Base struct without specialization
+template <typename T1, typename T2>
+struct InterleavedTupleType {
+  using interleaved_pack_t = typename interleave_pack<T1>::template with<T2>::type;
+  template <typename... Is>
+  struct TupleType {
+    using type = std::false_type;
+  };
+  template <std::size_t... Is>
+  struct TupleType<std::index_sequence<Is...>> {
+    using type = std::tuple<pack_element_t<Is, interleaved_pack_t>...>;
+  };
+  using type = typename TupleType<std::make_index_sequence<2>>::type;
+};
+
+// Proper specialization with tuples
+template <typename... T1s, typename... T2s>
+struct InterleavedTupleType<std::tuple<T1s...>, std::tuple<T2s...>> {
+  using interleaved_pack_t = typename interleave_pack<T1s...>::template with<T2s...>::type;
+  template <typename... Is>
+  struct TupleType {
+    using type = std::false_type;
+  };
+  template <std::size_t... Is>
+  struct TupleType<std::index_sequence<Is...>> {
+    using type = std::tuple<pack_element_t<Is, interleaved_pack_t>...>;
+  };
+  using type = typename TupleType<std::make_index_sequence<sizeof...(T1s) + sizeof...(T2s)>>::type;
+};
+
+// Create an instance of a tuple interleaved from given tuples
+template <typename... T1s, typename... T2s, std::size_t... Is>
+auto interleaveTuplesImpl(std::tuple<T1s...>& t1, std::tuple<T2s...>& t2, std::index_sequence<Is...>)
+{
+  return std::tuple_cat(std::make_tuple(std::get<Is>(t1), std::get<Is>(t2))...);
+}
+
+template <typename... T1s, typename... T2s>
+auto interleaveTuples(std::tuple<T1s...>& t1, std::tuple<T2s...>& t2)
+{
+  return interleaveTuplesImpl(t1, t2, std::index_sequence_for<T1s...>());
+}
+
+// Functions to create a tuple from N runs of a function that returns a value
+template <std::size_t I, typename R, typename C, typename... Args>
+R execFunctionWithDummyIndex(R (C::*f)(Args...), C& obj, Args... args)
+{
+  return (obj.*f)(args...);
+}
+
+template <typename R, typename C, typename... Args, std::size_t... Is>
+auto functionToTupleImpl(R (C::*f)(Args...), C& obj, Args... args, std::index_sequence<Is...>)
+{
+  return std::make_tuple((execFunctionWithDummyIndex<Is>(f, obj, args...))...);
+}
+
+template <std::size_t N, typename R, typename C, typename... Args>
+auto functionToTuple(R (C::*f)(Args...), C& obj, Args... args)
+{
+  return functionToTupleImpl(f, obj, args..., std::make_index_sequence<N>());
+}
+
+template <typename T1, typename GroupingPolicy, typename H, typename G, typename... Ts>
+struct GroupedCombinationsGenerator {
+};
+
+template <typename T1, typename GroupingPolicy, typename H, typename G, typename... Us, typename... As>
+struct GroupedCombinationsGenerator<T1, GroupingPolicy, H, G, pack<Us...>, As...> {
+  using joinIterator = typename soa::Join<H, G>::table_t::iterator;
+  using GroupedIteratorType = typename InterleavedTupleType<typename o2::soa::NTupleType<joinIterator, sizeof...(As)>::type, typename std::tuple<As...>>::type;
+
+  struct GroupedIterator : public std::iterator<std::forward_iterator_tag, GroupedIteratorType>, public GroupingPolicy {
+   public:
+    using reference = GroupedIteratorType&;
+    using value_type = GroupedIteratorType;
+    using pointer = GroupedIteratorType*;
+    using iterator_category = std::forward_iterator_tag;
+
+    GroupedIterator(const GroupingPolicy& groupingPolicy) : GroupingPolicy(groupingPolicy) {}
+    GroupedIterator(const GroupingPolicy& groupingPolicy, const H& hashes, const G& grouping, const std::shared_ptr<GroupSlicer<G, Us...>>&& slicer_ptr) : GroupingPolicy(groupingPolicy), mSlicer{std::move(slicer_ptr)}, mGrouping{std::make_shared<G>(std::vector{grouping.asArrowTable()})}
+    {
+      GroupingPolicy::setTables(join(hashes, grouping), join(hashes, grouping));
+      if (!this->mIsEnd) {
+        setCurrentGroupedCombination();
+      }
+    }
+
+    GroupedIterator(GroupedIterator const&) = default;
+    GroupedIterator& operator=(GroupedIterator const&) = default;
+    ~GroupedIterator() = default;
+
+    void setTables(const H& hashes, const G& grouping, std::shared_ptr<GroupSlicer<G, Us...>> slicer_ptr)
+    {
+      mGrouping = std::make_shared<G>(std::vector{grouping.asArrowTable()});
+      mSlicer = slicer_ptr;
+      setMultipleGroupingTables<sizeof...(As)>(join(hashes, grouping));
+      if (!this->mIsEnd) {
+        setCurrentGroupedCombination();
+      }
+    }
+
+    template <std::size_t N, typename T, typename... Args>
+    void setMultipleGroupingTables(const T& param, const Args&... args)
+    {
+      if constexpr (N == 1) {
+        GroupingPolicy::setTables(param, args...);
+      } else {
+        setMultipleGroupingTables<N - 1>(param, param, args...);
+      }
+    }
+
+    void moveToEnd()
+    {
+      GroupingPolicy::moveToEnd();
+    }
+
+    // prefix increment
+    GroupedIterator& operator++()
+    {
+      if (!this->mIsEnd) {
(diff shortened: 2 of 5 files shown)


saganatt:
Note: I left the MixedEventPartitionedTask, as I haven't tested partitions and filters thoroughly yet, after the recent filter changes. I will do it soon.

Note 2: Hash task will be replaced by a dedicated struct / lambda, this will come in another PR.

jgrosseo:
@saganatt it does not pass the CI in O2Physics. This is unexpected, right?

Framework/Core/include/Framework/GroupedCombinations.h
```diff
@@ -0,0 +1,300 @@
...
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef FRAMEWORK_PAIR_H
+#define FRAMEWORK_PAIR_H
```
ktf: Please  follow the convention:

```
O2_FRAMEWORK_GROUPEDCOMBINATIONS_H_
```

Framework/Core/include/Framework/GroupedCombinations.h
```diff
@@ -0,0 +1,300 @@
...
+    using type = std::tuple<pack_element_t<Is, interleaved_pack_t>...>;
+  };
+  using type = typename TupleType<std::make_index_sequence<sizeof...(T1s) + sizeof...(T2s)>>::type;
+};
+
+// Create an instance of a tuple interleaved from given tuples
+template <typename... T1s, typename... T2s, std::size_t... Is>
+auto interleaveTuplesImpl(std::tuple<T1s...>& t1, std::tuple<T2s...>& t2, std::index_sequence<Is...>)
+{
+  return std::tuple_cat(std::make_tuple(std::get<Is>(t1), std::get<Is>(t2))...);
```
ktf, line 26: Do you really need to use std::tuple? Can't you use our o2::framework::Pack and homogeneous_apply_ref?

Framework/Core/include/Framework/GroupedCombinations.h
```diff
@@ -0,0 +1,300 @@
...
+
+template <typename... T1s, typename... T2s>
+auto interleaveTuples(std::tuple<T1s...>& t1, std::tuple<T2s...>& t2)
+{
+  return interleaveTuplesImpl(t1, t2, std::index_sequence_for<T1s...>());
+}
+
+// Functions to create a tuple from N runs of a function that returns a value
+template <std::size_t I, typename R, typename C, typename... Args>
+R execFunctionWithDummyIndex(R (C::*f)(Args...), C& obj, Args... args)
```
ktf, line 37: There is already functions which do something like this elsewhere. Check Framework/FunctionalHelpers.h.

Framework/Core/include/Framework/GroupedCombinations.h
```diff
@@ -0,0 +1,300 @@
...
+        o2::soa::for_<k>([&](auto i) {
+          std::get<i.value>(associatedTables).bindExternalIndices(mGrouping.get());
+        });
+
+        mCurrentGrouped.emplace(interleaveTuples(currentGrouping, associatedTables));
+      }
+    }
+
+    std::shared_ptr<GroupSlicer<G, Us...>> mSlicer = nullptr;
+    std::shared_ptr<G> mGrouping;
```
ktf, line 204: Why a shared ptr?

Framework/Foundation/include/Framework/Pack.h
```diff
@@ -90,6 +90,14 @@ constexpr auto concatenate_pack(P1 p1, P2 p2, Ps... ps)
...
 using concatenated_pack_t = decltype(concatenate_pack(Ps{}...));
 
+template <typename... Args1>
+struct interleave_pack {
+  template <typename... Args2>
+  struct with {
+    using type = concatenated_pack_t<pack<Args1, Args2>...>;
+  };
+};
+
```
ktf: Doesn't:

```
template <typename... Args1, typename... Args2>
constexpr auto interleave_pack(pack<Args1...>, pack<Args2...>)
{
  return concatenated_pack_t<pack<Args1, Args2>...>{};
}
```

work?

Framework/Core/include/Framework/GroupedCombinations.h
```diff
@@ -0,0 +1,300 @@
...
+// or submit itself to any jurisdiction.
+
+#ifndef FRAMEWORK_PAIR_H
+#define FRAMEWORK_PAIR_H
+
+#include "Framework/ASoAHelpers.h"
+#include "Framework/GroupSlicer.h"
+#include "Framework/StringHelpers.h"
+
+#include <iostream>
```
ktf: Do not include iostream in an header.

ktf (changes requested):
Thank you for the contribution! We need however a few changes before this can be merged. Could you have a look at the inline comments? It would also be good if you could split up the PR in smaller changes, so that things which are unambiguous can be merged immediately while less trivial changes have less noise around them.

Framework/Core/include/Framework/GroupedCombinations.h
saganatt: As far as I understand, a pack does not hold any data. But I create this interleaved tuple to output the collisions and tracks to the user. Like in usual combinations.

Framework/Core/include/Framework/GroupedCombinations.h
saganatt: I am not sure this does what we want. Or, at least I don't see how this enables repeated calls of a function?

Framework/Core/include/Framework/GroupedCombinations.h
saganatt: With unique_ptr I cannot use the default copy constructor in l. 112, used e.g. in l. 244 - 259

Framework/Foundation/include/Framework/Pack.h
saganatt: Indeed it worked, thanks!

saganatt:
Thanks @ktf , I corrected according to your suggestions, please see my comments.

I am not sure about splitting - I can push Pack.h separately, but the rest depends on GroupedCombinations as a whole. So it is still the matter of completing grouped combinations.

Framework/Foundation/include/Framework/Pack.h
```diff
@@ -90,6 +90,15 @@ constexpr auto concatenate_pack(P1 p1, P2 p2, Ps... ps)
...
 
+template <typename... Args1, typename... Args2>
+constexpr auto interleave_pack(pack<Args1...>, pack<Args2...>)
+{
+  return concatenated_pack_t<pack<Args1, Args2>...>{};
+}
+
+template <typename... Ps>
+using interleaved_pack_t = decltype(interleave_pack(Ps{}...));
+
```
ktf: Thanks. It would be good if you moved the Pack.h changes in a separate PR and add a couple of tests to show the newly introduced  features.

Framework/Foundation/include/Framework/Pack.h
```diff
@@ -248,6 +257,25 @@ constexpr auto concatenate_pack_unique(P1 p1, P2 p2, Ps... ps)
...
+template <typename P, typename PT = pack<>>
+struct unique_types_impl {
+  using type = PT;
+};
+template <typename PT, typename T, typename... Ts>
+struct unique_types_impl<pack<T, Ts...>, PT> : unique_types_impl<pack<Ts...>, add_type<T, PT>> {
+};
+template <typename P>
+using unique_types = typename unique_types_impl<P>::type;
+
```
ktf: You should try to follow the conventions which we use in the rest of the file. In particular:

* `unique_types` /  `add_type` should probably be `uniq_pack_t`
* `add_type` should probably be renamed to`set_insert_pack` or something like that.

I also wonder if the whole logic would not be better expressed by a constexpr function, like done elsewhere.

saganatt: Ok, indeed it is possible

Framework/Foundation/include/Framework/Pack.h
saganatt: OK, the pack PR is [here](https://github.com/AliceO2Group/AliceO2/pull/7847)

Framework/Core/include/Framework/GroupedCombinations.h
```diff
@@ -0,0 +1,296 @@
...
+  template <std::size_t... Is>
+  struct TupleType<std::index_sequence<Is...>> {
+    using type = std::tuple<pack_element_t<Is, interleaved_pack_t<pack<T1s...>, pack<T2s...>>>...>;
+  };
+  using type = typename TupleType<std::make_index_sequence<sizeof...(T1s) + sizeof...(T2s)>>::type;
+};
+
+// Create an instance of a tuple interleaved from given tuples
+template <typename... T1s, typename... T2s, std::size_t... Is>
+auto interleaveTuplesImpl(std::tuple<T1s...>& t1, std::tuple<T2s...>& t2, std::index_sequence<Is...>)
```
ktf: Why not simply introducing a `pack_to_tuple(pack<...>)` helper, with the associated `pack_to_tuple_t<pack<...>>`? Then expressing the interleaved tuple should be just a matter of using those helpers, no? I think this part would become greatly simplified. Could you give it a try?

Framework/Core/include/Framework/GroupedCombinations.h
ktf: Ok, let's keep this pending for now. If you address the `pack_to_tuple` comment maybe I get a clearer idea of  what you want to do here.

Framework/Core/include/Framework/GroupedCombinations.h
ktf: Same here. Please, address the comment about `pack_to_tuple` and the we discuss again.

Framework/Foundation/include/Framework/Pack.h
ktf: Same here. Please, address the comment about `pack_to_tuple` and the we discuss again.

jgrosseo:
I see the other PR was merged. @saganatt can you then rebase this. @ktf what else needs fixing here?

ktf:
@jgrosseo see the other comment, I think the whole InterleavedTupleType can be simplified. While I appreciate this should go in asap, I really would like to try to reduce it to the minimum the usage of std::tuple, because in the past it was correlated with (even slower) compilation times.

saganatt:
@ktf, I am trying to re-implement what you suggest, I will push it today/early tomorrow (sorry for the weekend time!).

In general: I could replace tuple with a vector, for example. But not with a pack, I need something which holds the actual data. And I thought tuple is the recommended choice? Like for the usual combinations?

ktf:
Sure, you can keep the tuple, but to calculate an interleaved tuple you can simply interleave the associated pack and then transform the pack into a tuple declaration. If you can use an std::array (not a vector, which would allocate extra memory on the heap) it would be even better.

saganatt:
I simplified the code and put pack-related functions again in Pack.h.
I sticked to tuples, though - in each case there might be different types mixed, so other containers cannot be used.

Framework/Foundation/include/Framework/Pack.h
```diff
@@ -272,6 +272,29 @@ constexpr auto unique_pack(pack<T, Ts...>, PT p2)
...
+template <typename P, typename Is>
+struct pack_to_tuple;
+
+template <typename P, std::size_t... Is>
+struct pack_to_tuple<P, std::index_sequence<Is...>> {
+  using type = std::tuple<pack_element_t<Is, P>...>;
+};
+
+template <typename P>
+using pack_to_tuple_t = typename pack_to_tuple<P, std::make_index_sequence<pack_size(P{})>>::type;
```
ktf: Wouldn't:

```C++
template <typename...ARGS>
inline constexpr std::tuple<ARGS...> pack_to_tuple(pack<ARGS...>) { return std::tuple<ARGS...>{};}

template <typename...ARGS>
using pack_to_tuple_t = decltype(pack_to_tuple(pack<ARGS...>{}));
```

be more convenient? Or alternatively something like (not 100% sure though):

```C++
template <typename T>
struct pack_to_tuple_t{};

template <typename... ARGS>
struct pack_to_tuple_t<pack<ARGS...>> {
   using type = std::tuple<ARGS...>;
};
```

Sorry for being pedantic, but Pack.h is quite in the critical path for compile time performance.

saganatt:
Sorry, I forgot to add the tests. All checks passed before, so should be still OK. Do you want a separate PR for pack changes?

@ktf are we happy now with interleaveTuples() and functionToTuple() in GroupedCombinations.h? Let me know what needs fixing.

And a perhaps trivial question to @jgrosseo : I quietly skip collision pairs (tuples) if any of the collisions does not contain any tracks. Is there a need to output to the user all collision pairs, even those with some track tables empty?

Framework/Core/include/Framework/AnalysisManagers.h
```diff
@@ -29,6 +30,29 @@
...
+};
+
+template <typename T1, typename GroupingPolicy, typename H, typename G, typename... Us, typename... As>
+struct GroupedCombinationManager<GroupedCombinationsGenerator<T1, GroupingPolicy, H, G, pack<Us...>, As...>> {
+  template <typename TH, typename TG, typename... T2s>
+  static void setGroupedCombination(GroupedCombinationsGenerator<T1, GroupingPolicy, H, G, pack<Us...>, As...>& comb, TH& hashes, TG& grouping, std::tuple<T2s...>& associated)
+  {
+    static_assert(sizeof...(T2s) > 0, "There must be associated tables in process() for a correct pair");
+    static_assert(!soa::is_soa_iterator_t<std::decay_t<H>>::value, "Only full tables can be in process(), no grouping");
+    if constexpr (std::conjunction_v<std::is_same<G, TG>, std::is_same<H, TH>>) {
```
ktf, line 48: why not &&?

Framework/Foundation/include/Framework/Pack.h
```diff
@@ -272,6 +272,28 @@ constexpr auto unique_pack(pack<T, Ts...>, PT p2)
...
+template <typename... Ts>
+inline constexpr std::tuple<Ts...> pack_to_tuple(pack<Ts...>)
+{
+  return std::tuple<Ts...>{};
+}
+
+template <typename P>
+using pack_to_tuple_t = decltype(pack_to_tuple(P{}));
+
+template <typename T, std::size_t N, typename... REST>
```
ktf: Why do you need REST at all? Isn't T the  only type involved?

Framework/Foundation/include/Framework/Pack.h
```diff
@@ -272,6 +272,28 @@ constexpr auto unique_pack(pack<T, Ts...>, PT p2)
...
+constexpr auto repeat_type_pack(pack<REST...>)
```
ktf: Why do you need to pass the pack?

Framework/Foundation/include/Framework/Pack.h
ktf: Ok, nevermind, I understand the logic now.

Framework/Core/include/Framework/AnalysisManagers.h
saganatt: For passing the tables to the manager? But then the original tables (collisions, tracks) accessible inside process() would be invalidated (ownership loss), no? Even if there is no use case for using these tables in an event mixing task, this would be surely confusing to users. Similarly, in PartitionManager, we pass tables just by &.

Framework/Foundation/include/Framework/Pack.h
saganatt: To repeat T. REST is nothing, then T, then T, T and so on, till REST becomes T repeated N times.

Framework/Foundation/include/Framework/Pack.h
saganatt: right, perhaps I can get rid of this, let me try.

ktf:
I was actually thinking:

```C++
template <typename T,int... Is>
inline auto sequence_to_pack(std::integer_sequence<int, Is...>){
    return pack<decltype((Is,T{}))...>{};
};

template <typename T, std::size_t N>
using repeated_type_pack_t = decltype(sequence_to_pack<T>(std::make_index_sequence<N>()));
```

might be a bit more orthogonal.

saganatt:
Now I don't fully understand how `decltype((Is,T{}))` returns only T, but indeed it is simpler.

ktf:
Given:
```
(expr1, expr2)
```

the comma operator evaluates all expression in order and then returns the value of the last expression.

saganatt:
ah, right, thank you!

Framework/Core/include/Framework/GroupedCombinations.h
```diff
@@ -0,0 +1,267 @@
...
+}
+
+template <typename T1, typename GroupingPolicy, typename H, typename G, typename... Ts>
+struct GroupedCombinationsGenerator {
+};
+
+template <typename T1, typename GroupingPolicy, typename H, typename G, typename... Us, typename... As>
+struct GroupedCombinationsGenerator<T1, GroupingPolicy, H, G, pack<Us...>, As...> {
+  using joinIterator = typename soa::Join<H, G>::table_t::iterator;
+  using GroupedIteratorType = pack_to_tuple_t<interleaved_pack_t<repeated_type_pack_t<joinIterator, sizeof...(As)>, pack<As...>>>;
```
ktf, line 61: `repeated_type_pack_t<joinIterator, sizeof...(As)>` is probably better written as:

```
pack<(decltype(As, joinIterator{}))...>
```

so that you avoid the repacking...

saganatt: yes, I can avoid repeated_type_pack_t.
I wonder also about the interleaved_pack_t now, but apparently it is too tricky to get two variadics expanding at once, and not just the second one....

ktf:
That said, let's avoid other changes for  now and get the tests to finish. We can improve later. It's already quite an improvement.

Framework/Core/include/Framework/GroupedCombinations.h
ktf: Yes, I wondered about the same, but again, let's merge this and then we can continue.

Framework/Core/include/Framework/AnalysisManagers.h
ktf: No, I meant why not using `std::is_same_v<G, TG> && std::is_same_v<H, TH>`. We can fix it later.

ktf:
Tested on hyperloop, merging. It would be good to go through Framework/GroupedCombinations.h and see if some further simplifications cannot be made.

PR #7828, opened 2021-12-07
Use Root units for geometry and material construction

sawenzel:
materials had wrong radiation length because TGeo
was using G4 units internally (probably introduced
without us knowing). Now switching explicitely back to Root units.


sawenzel:
@amorsch @shahor02

PR #7835, opened 2021-12-08
o2-sim: Introduce persistent aligned geometry

sawenzel:
We stream out an aligned version of the geometry file,
for direct use in digitization and other steps in the MC chain.
This prevents a lot of calls to CCDB and uncessary dynamic realignment.

Digitization is moved to using the new aligned geometry file.
We can think of doing same for other steps.

A couple of minor changes accompany this commit:
- use correct timestamp when aligning in o2-sim

### DIFF
--- a/Detectors/Base/include/DetectorsBase/GeometryManager.h
+++ b/Detectors/Base/include/DetectorsBase/GeometryManager.h
@@ -49,7 +49,9 @@ class GeometryManager : public TObject
 {
  public:
   ///< load geometry from file
-  static void loadGeometry(std::string_view geomFilePath = "", bool applyMisalignment = true);
+  ///< When applyMisalignedment == false --> read from unaligned file
+  ///< When preferAlignedFile == true and applyMisalignment == true : Prefer reading from existing aligned file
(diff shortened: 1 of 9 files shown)


Detectors/Base/include/DetectorsBase/GeometryManager.h
```diff
@@ -49,7 +49,9 @@ class GeometryManager : public TObject
 {
  public:
   ///< load geometry from file
-  static void loadGeometry(std::string_view geomFilePath = "", bool applyMisalignment = true);
+  ///< When applyMisalignedment == false --> read from unaligned file
+  ///< When preferAlignedFile == true and applyMisalignment == true : Prefer reading from existing aligned file
+  static void loadGeometry(std::string_view geomFilePath = "", bool applyMisalignment = true, bool preferAlignedFile = true);
```
shahor02: setting `preferAlignedFile = true` by default will trigger loading aligned geometry everywhere, can use set it to false and require it explicitly in the digitization?

sawenzel: Good catch. Yes, of course.

PR #7838, opened 2021-12-08
TPC: change track params in dEdx calibration

badarots:
Correct for Tgl and Snp, insted of Z and Tgl

Other  changes:
* Save run number as CCDB metadata
* ajustable Tgl limits
* record number of tracks of each correction fit
* Method to convert a boost hist into a THn

### DIFF
--- a/Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
+++ b/Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
@@ -28,9 +28,7 @@
 
 // boost includes
 #include <boost/histogram.hpp>
-
-// forward declarations
-class TH2F;
+#include "THn.h"
 
 namespace o2::tpc
 {
@@ -44,9 +42,9 @@ class CalibdEdx
  public:
   enum Axis {
     dEdx,
-    Z,
     Tgl,
-    // Snp,
+    Snp,
+    // Z,
     Sector,
     Stack,
     Charge,
@@ -63,9 +61,9 @@ class CalibdEdx
   // and the HistAxis enum
   using AxesType = std::tuple<
     FloatAxis, // dEdx
-    FloatAxis, // Z
     FloatAxis, // Tgl
-    // FloatAxis, // Snp
+    FloatAxis, // Snp
+    // FloatAxis, // Z
     IntAxis, // sector
     IntAxis, // stackType
     IntAxis  // Charge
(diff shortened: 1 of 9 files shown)


Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -63,18 +61,18 @@ class CalibdEdx
...
+    FloatAxis, // Snp
+    // FloatAxis, // Z
     IntAxis, // sector
     IntAxis, // stackType
     IntAxis  // Charge
     >;
 
   using Hist = boost::histogram::histogram<AxesType>;
-  using FitCuts = std::array<int, 3>;
+  using FitCuts = std::array<float, 3>;
```
wiechula: Why did you change this to float?

badarots: For stupid reasons, never mind. Its back an int.

Detectors/TPC/calibration/include/TPCCalibration/CalibdEdx.h
```diff
@@ -130,15 +118,17 @@ class CalibdEdx
   /// Save the histograms to a TTree.
   void writeTTree(std::string_view fileName) const;
 
- private:
-  constexpr static float mipScale = 1.0 / 50.0; ///< Target value for MIP dE/dx
+  constexpr static float mipScale = 1.0 / 50.0; ///< Inverse of target MIP dE/dx value
```
wiechula, line 121:
```suggestion
  constexpr static float MipScale = 1.0 / 50.0; ///< Inverse of target MIP dE/dx value
```
see https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html?showone=Variable_Names#Variable_Names

Detectors/TPC/calibration/src/CalibdEdx.cxx
```diff
@@ -150,9 +167,9 @@ void fitHist(const Hist& hist, CalibdEdxCorrection& corr, TLinearFitter& fitter,
 
         // constexpr float dEdxResolution = 0.05;
         // const double error = dEdx * dEdxResolution / sqrt(counts);
-        const double error = 1. / sqrt(counts);
+        const double error = 1. / counts;
```
wiechula, line 170: I'm still puzzled about what the correct error is ...

wiechula (changes requested):

PR #7839, opened 2021-12-08
DPL: actually populate Dict with associated ptree

ktf:


ktf:
Trivial. Checked by both @Barthelemy and @myself. Merging.

PR #7843, opened 2021-12-08
Do not cleanup cached CCDB objects of some special types

shahor02:
The BasicCCDBManager cache is based on shared_ptr, which assumes that the object is never
deleted via its raw pointer. We cannot control this for some global ROOT pointers, e.g.
gGeoManager, which is deleted on exit internally by ROOT (thus creating a race condition with
destruction of the same pointer via shared_ptr.
Such objects will be cached to raw pointers in the BasicCCDBManager but will be never deleted by it.

### DIFF
--- a/CCDB/include/CCDB/BasicCCDBManager.h
+++ b/CCDB/include/CCDB/BasicCCDBManager.h
@@ -25,6 +25,8 @@
 
 // #include <FairLogger.h>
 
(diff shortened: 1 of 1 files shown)


shahor02:
@victor-gonzalez @jgrosseo Once this is merged, you can probably revert the https://github.com/AliceO2Group/O2Physics/pull/402.

jgrosseo (approved):

PR #7844, opened 2021-12-08
[rANS] Reegineer Frequency Tables and Renorming

MichaelLettrich:
* Implements a new rANS renorming algorithm that marks very low frequency symbols as incompressible during renorming with the effect of increasing the renorming accuracy and thus compression effectiveness. As a positive side-effect, a more effective renorming reduces the size of the inverse symbol lookup table required for decoding which automatically leads to a drastic reduction in its size, especially for large alphabets.
* Reorganize and Rewrite the rANS Frequency table to tightly integrate the concept of incompressible symbols and reduce computational and memory overhead during the construction of symbol tables.
* Adapt and simplify O2 Code that uses librANS. 
* Extensive unittests

 Currently the legacy renorming functions are kept in place for backward compatibility. A scheme for using both new and old renorming algorithms will follow after discussing them more in-depth with other stakeholders.


davidrohr:
Having a brief look, the changes look fine.
The MacOS issue looks genuine.
It passes the fullCI, so should be fine at first glance, but we should do some more testing for this.
- I think we should run with some old dictionary on some real CTF data we have recorded and compare it to before.
- And then we should run the full system test sync reco with and without this PR on the same dataset, and see whether we get the same result in the async reco?
Who will take care of this? I can do if needed, but also not too excited about volunteering :)

MichaelLettrich:
Indeed the macOS error looks legit and it's a very interesting finding. Since I am sorting the values to be corrected when renorming. In this case both values are 2 before correction, so their ordering with sort can be arbitrary. My suspicion is that `libc++` and `stdlibc++` implementations of `std::sort` produce both valid results but with different ordering. Choosing `std::stable_sort` should give the same results on each platform. 

As for the testing, I did indeed run the full system test with each commit and also reused the same dictionary across commits without errors. I agree however that the tests you proposed should be performed. I can do them but would need some assistance on how to perform checks on real data.

shahor02:
Hi @MichaelLettrich 

Here are the tests I would suggest to do:

1) make sure old CTFs (e.g. from the pilot beam) are readable:
e.g. you can
alien_cp alien:///alice/data/2021/OCT/505669/raw/0630/o2_ctf_run00505669_orbit0000328291_tf0000000001.root ./
fetch a few necessary files from https://cernbox.cern.ch/index.php/s/m2wSU3ieTCRqXc3, unpack to the same dir. where the ctf file is,
then run (e.g. for 100 TFs)
```
o2-ctf-reader-workflow --allow-missing-detectors --ctf-input o2_ctf_run00505669_orbit0000328291_tf0000000001.root --max-tf 100 --delay 0.1 | o2-ft0-digits-writer-workflow --disable-mc  | o2-its-cluster-writer-workflow --disable-mc --shm-segment-size 32000000000 --run | tee ctfread.log
```
Apart from the flawless decoding it is worth to check the result of the decoding for a few channels, so I've added the writers for ft0 and ITS. You can make sure that the histograms
```
root o2_ft0digits.root
o2sim->Draw("FT0DIGITSCH.ChainQTC");
o2sim->Draw("FT0DIGITSCH.CFDTime");
```
and
 ```
root o2clus_its.root
o2sim->Draw("ITSClusterPatt");
o2sim->Draw("ITSClusterComp.mData");
```
are exactly the same with the current dev and your PR.

2) make sure you can write CTF (and read back freshly written data) with new external dictionary:
run the FST locally, save somewhere produced AO2D.root
Then `rm reco_NOGPU.log_done reco_ASYNC.log_done`, and run

CREATECTFDICT=0 $O2_ROOT/prodtests/full_system_test.sh

You can compare a few columns in the new and saved AO2D.root, e.g.
root AO2D.root
DF_7032750922744725504->cd(); // note that this directory name will be random.
O2ft0->Draw("fAmplitudeC"); // must be exactly the same
O2track->Draw("fSnp"); // must be almost the same, since the TPC reco is not deterministic, there might be some differences.

3) the same for the old dictionary: like in (2), but after `rm reco_NOGPU.log_done reco_ASYNC.log_done` copy to the working directory the ctf_dictionary.root from the tarball I've provided for (1).

Let me know in case of difficulties or if something is not clear.

**Comment by MichaelLettrich** [comment]
@shahor02  thank you very much for your support and the clear and easy to follow instructions. I can confirm that the data for t0`is identical now. 

This is data from the TPC plot you asked me to create. @shahor02  @davidrohr  please let me know if it looks similar enough.
Screenshot from 2021-12-16 12-53-20
`

**Comment by MichaelLettrich** [comment]
We discovered an issue with uninitialized values in ft0 digits which resulted in writing of random values in the ctf (and thus could not be reproduced.) I'm opening another PR for ft0.

**Comment by shahor02** [comment]
Hi @MichaelLettrich 

Thanks for checking,  what it what on `snp` plots? The only non-deterministic tracking step is TPC, but even in this case a difference in the number of reconstructed track over the same data is <<1, while on your plots I see 20 and 9% difference of 2nd and 3d wrt the 1st.

**Comment by MichaelLettrich** [comment]
@shahor02 sorry, first is with this PR, second is dev, third is another run of dev.

**Comment by shahor02** [comment]
Hi @MichaelLettrich 

Here what I see in your log files (for @davidrohr:  dev, newRenorming and newRenorming with old external dictionary.  the "0" means we create the dictionary, "1" means we reuse the dictionary created in step "0")

```
						TPC			TPC-ITS/TPC	PV   NAODTracks
reco_ASYNC.log.dev0		5193690 TPC Clusters 42718 track(s)	9465/25139	12   218335
reco_ASYNC.log.dev1 		5194521 TPC Clusters 42740 track(s)	9465/25141	11   199026
reco_ASYNC.log.externalDict 	5196724 TPC Clusters 42743 track(s)	9469/25132	10   182028
reco_ASYNC.log.renorm0		5193805 TPC Clusters 42732 track(s)	9468/25137	10   181971
reco_ASYNC.log.renorm1		5194055 TPC Clusters 42725 track(s)	9467/25136	10   181921
```
TPC-ITS/TPC gives number of matches / number of TPC tracks preselected for matching (reaching outermost ITS layer), PV: number of primary vertices found, NAODTracks: number of barrel tracks stored in the AOD.

Since the TPC tracking is non-deterministic, for the same input data one can get sometimes stray vertices (with a few tracks only) which are time-wise compatible with tracks belonging to other vertices. Since the AOD at the moment stores separately all tracks compatible with given vertex, there is a lot of redundancy in the AOD, which scales not only with the number of the input tracks but also with the number of vertices (even if the vertex was fitther with 2 tracks it can have a few K tracks time-wise compatible). This explains the huge difference in the number of AOD tracks. 
What matters is that the reconstruction produces consistent results in all cases, so for me it is fine to merge this PR (fullCI has failed for unrelated reasons, but it was green before).
The only thing I would ask @davidrohr to confirm is that the variation in the number of TPC tracks and clusters is within the expectation).

MichaelLettrich:
@shahor02 actually the `fullCI` and `o2-cs8` for some reason crash in the `ctf-coder` unit test.  I will investigate. refrain from merging until CI is fully green.

davidrohr:
@shahor02 : Sorry, saw your question only now. The variation of tracks is expected. I am a bit surprised by the variation of clusters. That should not happen running from raw data, but obviously when you use the track model compression, which discards hits of low momentum tracks, this can affect the number of clusters. Could you try to run with `--configKeyValues "GPU_proc.ompKernels=0"`. That should make it deterministic, while reducing the speed. But for such a test it should be helpful.

shahor02:
Hi @davidrohr 
thanks, I think @MichaelLettrich run also the sync rec in each test, so the starting CTF was not the same. @MichaelLettrich could you try with the deterministic option which @davidrohr indicated?

shahor02:
@MichaelLettrich the CI failures are genuine

MichaelLettrich:
@shahor02 I identified the breaking change in an extra commit. There was however no problems on my machine or with the macs with it, so it seems like maybe memory or undefined behavior. I still don't see how it's incorrect or undefined. Do you see anything?

shahor02:
Hi @MichaelLettrich ,
now it is green, not sure if your last commit fixed it. I did not save the log of the failed test, but given that it had lot of dumps from the rANS, I guess one of your `asserts` was triggered.

MichaelLettrich:
@shahor02 : I dumped the data that enters the encoder and exits the decoder via a new logging mechanism in the rANS coder, which allows me to compare all payloads in an automated manner. I can confirm that with and without external dictionaries encoder input always matches decoder output without exception within the execution of a full system test which to me means that the entropy coder is working correctly.

When I execute the full system test a second time, and compare the inputs and outputs to those of the first execution , the combined `QTotQMaxA` array yields a different array length.  @davidrohr . It seems I did not put `--configKeyValues "GPU_proc.ompKernels=0"` into the right spot ( I put it here: https://github.com/AliceO2Group/AliceO2/blob/e69b9ad9069273619bd95a12832ad5a635546149/prodtests/full-system-test/dpl-workflow.sh#L396. Is there any other place where I need to put it?)

davidrohr:
> . Is there any other place where I need to put it?)

No, that looks correct. Let me check...
In the console output, does the gpu-reconstruction find the same number of tpc clusters and tpc tracks in the 2 iterations?

MichaelLettrich:
@davidrohr the error was on my side. Now everything matches: We can get the same results with dev and this PR. From my side we can merge after my final, cleaned up version passes the CI.

davidrohr:
ok, thx. read your post only now.
But apparently something is broken with dev. Have been trying to run a large full system test since 8 hours now, and it is always crashing, but independent from your PR. Something else got broke apparently. But since it is independent of this one, and as you say you confirmed now this is OK, I am fine with merging it once the CI is green.

davidrohr:
@MichaelLettrich : just to confirm before I merge, this is now the final cleaned up version?

MichaelLettrich:
@davidrohr yes. this can be merged.

PR #7845, opened 2021-12-09
GPU benchmark: Add --raw output option, fix float chunk size TP

mconcas:
@davidrohr relevant fix is e.g. Kernels.cu:#L798.
Added a `--raw` option: minimal metrics/parameters. Can be extended at occurrence.
I know the code it's becoming a bit 🍝 ... ;)


davidrohr:
@mconcas  thx, just to be 100% sure: fractional chunk sizes are supported, just the throughput was computed incorrectly?

mconcas:
> @mconcas  thx, just to be 100% sure: fractional chunk sizes are supported, just the throughput was computed incorrectly?

Yes, indeed the elapsed time was growing proportionally with the size. The chunk size at numerator was casted to size_t, hence being rounded down.

PR #7846, opened 2021-12-09
Fix T0 filling

jgrosseo:


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1353,9 +1353,9 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   // vector of FT0 amplitudes
   int nFT0Channels = o2::ft0::Geometry::Nsensors;
   int nFT0ChannelsAside = o2::ft0::Geometry::NCellsA * 4;
-  std::vector<float> vAmplitudes(nFT0Channels, 0.);
   // filling FT0 table
   for (auto& ft0RecPoint : ft0RecPoints) {
+    std::vector<float> vAmplitudes(nFT0Channels, 0.);
```
shahor02, line 1358: this means an allocation of a new vector for every trigger, I would simply get rid of it.

jgrosseo: Yes, see mattermost. I already confirmed that.
https://mattermost.web.cern.ch/alice/pl/1pwtw45b1trmiygnp3tupu6qkr

AllaMaevskaya: set amplitude vector in zero before each event #7848, 
I think FV0 and FDD need similar fix

PR #7852, opened 2021-12-09
DPL: do not initialise InfoLogger when not requested

ktf:
Unless --infologger-mode or O2_INFOLOGGER_MODE is set to
a valid backend, we do not instanciate InfoLoggerClient.

### DIFF
--- a/Framework/Core/src/CommonServices.cxx
+++ b/Framework/Core/src/CommonServices.cxx
@@ -262,6 +262,9 @@ auto createInfoLoggerSinkHelper(InfoLogger* logger, InfoLoggerContext* ctx)
   };
 };
 
+struct MissingService {
+};
+
 o2::framework::ServiceSpec CommonServices::infologgerSpec()
 {
   return ServiceSpec{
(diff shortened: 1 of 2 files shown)


ktf:
@sy-c it looks like it's enough to instantiate InfoLoggerClient to have periodic reconnection attempts to some backend. Is that  expected?

@davidrohr this should explain and correct the short-lived thread you see.

@chiarazampolli I see also a (small) reduction in the number of open files with this.

ktf:
@awegrzyn looks like Monitoring is using an out-of-band InfoLogger. Would it be possible to disable it or to provide my own hook?

sy-c:
> @sy-c it looks like it's enough to instantiate InfoLoggerClient to have periodic reconnection attempts to some backend. Is that expected?

Indeed, that would happen when O2_INFOLOGGER_MODE is infoLoggerD (default), it will try to reconnect every 5s. If this is not desirable, the output mode can be instead set to file, stdout, or none.

awegrzyn:
> @awegrzyn looks like Monitoring is using an out-of-band InfoLogger. Would it be possible to disable it or to provide my own hook?

It is if CMake successfully detects it, otherwise it falls back to internal logger that just prints the logs to stdout.

What do you try to achieve? Maybe you could just use internal logger?
If you want to mute all the messages then maybe there's global option in InfoLogger?

ktf:
@sy-c maybe we should make the default stdout or none?
@awegrzyn we always compile with InfoLogger enabled, because there are of course cases in which we want to use it, of course, e.g. the sync reconstruction. Maybe I can simply override the O2_INFOLOGGER_MODE to be none by default, before  I create the Monitoring instance. I will try with that.

sy-c:
I think the default behavior is correct, creating an infoLogger instance should push messages to infoLogger system. The env variable is exactly meant to change this default. e.g. I set it automatically to stdout in o2-readout-exe when running interactively from the console, before creating the log handle.

ktf:
Yes, the issue is what  to do when the creation of  the infologger is done with a singleton like in monitoring. As I said, I will try to override O2_INFOLOGGER_MODE in case it's empty and hopefully that's actually done early enough.

PR #7853, opened 2021-12-09
DPL: close (almost) all file descriptors before forking

ktf:


ktf:
@chiarazampolli I think this is it.

Framework/Core/src/runDataProcessing.cxx
```diff
@@ -726,20 +726,29 @@ void spawnDevice(DeviceRef ref,
...
+      }
+      if (childFds[ref.index].childstderr[1] == i) {
+        continue;
+      }
+      if (i == STDERR_FILENO) {
+        continue;
+      }
+      auto res = close(i);
+      if (res == 0) {
+        std::cerr << "pid: " << getpid() << "successfully closed " << i << std::endl;
```
davidrohr: do we want this printout, or is it just a test?

ktf: This was just a test. I will drop it.

ktf:

PR #7856, opened 2021-12-10
DPL: use fatal to report uncaught exceptions

ktf:
Otherwise the machinery will not harvest the errors correctly.


ktf:
@TimoWilken I think o2-dataflow is  having  troubles "Started Fri 13:37 CET @ mesosci/ci_alisw_slc7-builder_latest/4".

TimoWilken:
Hm, that builder spent a lot of time rebuilding almost everything (including Clang and ROOT). I'll try to find out why that is.

PR #7864, opened 2021-12-11
DPL: double the number of processes which can be created by the driver

ktf:
This makes sure that the extra file descriptors which are needed to
setup the parent / child communication channel created as late as
possible, and closed as soon as the fork happens.

Before this change, since we were creating the file descriptors in
one go, the extra endpoints which get closed after the fork were needlessly
using resources, leading to early exhaustion.

### DIFF
--- a/Framework/Core/src/runDataProcessing.cxx
+++ b/Framework/Core/src/runDataProcessing.cxx
@@ -613,13 +613,6 @@ struct DeviceStdioContext {
   int childstdout[2];
 };
 
-void prepareStdio(std::vector<DeviceStdioContext>& deviceStdio)
-{
-  for (auto& context : deviceStdio) {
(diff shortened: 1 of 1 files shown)


ktf:
@davidrohr this should allow the driver to spawn up to ~110 devices with ulimit -n  256.

davidrohr (approved):
Thx, this is a significant improvement.
Now I can run the workflow (still without QC) with `ulimit -n 256`
(ping @chiarazampolli)

ktf:
We can try to improve a few more things, however if we want to have a stable connection to the driver to report logs / metrics, we clearly cannot go beyond less than 1 fd per child.

davidrohr:
ok, but 1 socket per child sounds more than reasonable, let's see if this helps on the grid

PR #7866, opened 2021-12-13
[O2-2712] DPL Analysis: add an exception on invalid index access

aalkin:
This adds an exception when invalid index is accessed, to prevent returning an incorrect table iterator.


aalkin:
Solves O2-2712.

ktf (approved):

ktf:
I am not sure the `if` and  the inlined exception is a good idea, but I do  not have a better suggestion for now.

ktf:
Maybe optional...

PR #7867, opened 2021-12-13
DPL Analysis: update advanced indices

aalkin:
* Slice index is now considered valid only when both entries are >0
* Invalid slice index getter would return an empty slice of the target table
* Array index is now backed by variable size array
* Updated test


ktf:
Tested on hyperloop. Merging.

aalkin:
@ktf it is not merged yet, are there any issues?

ktf:
it had automerge enabled but one of the tests did not work.

PR #7868, opened 2021-12-13
DPL: fix quoting environment

ktf:
Since we use putenv, there is no need to provide a quoted string.


ktf:
Discovered while running with `-fsanitize=address` on hyperloop.

PR #7869, opened 2021-12-13
[MCH] add protections

pillot:
- add out-of-bound protection for internal DE and pad arrays (this is just extra protection of the code against bad memory access, since the validity of digits' pad and DE IDs should be guaranteed at this point)

- add protection against multiple digits on the same pad. I am not sure to which extend this can/should be addressed in e.g. the time clustering. For the moment I just keep the first digit and skip the others when that happens.


aphecetche (approved):
Noting (for the future) that we should at some point review our usages of both the error log and the exception.

As we already discussed that's quite inefficient on the long run to just log on error, and we should find a more persistent way of e.g. aggregating error). And for the throws we should also ensure that we are able to recover from them (idea here being of not crashing the reco for one single bad ROF for instance).

PR #7870, opened 2021-12-13
CCDBManager: Fatal when CCDB return nullptr

sawenzel:
Access to CCDB via BasicCCDBManager now
fatals by default whenever the return value is nullptr.
This is introduced upon user request.

One can still decide to treat this situation in
user code by setting `fatalWhenNull = false`.

For now, the lower level API in CcdbApi remains
untouched.

### DIFF
--- a/CCDB/include/CCDB/BasicCCDBManager.h
+++ b/CCDB/include/CCDB/BasicCCDBManager.h
@@ -23,8 +23,6 @@
 #include <unordered_map>
 #include <memory>
 
-// #include <FairLogger.h>
-
 class TGeoManager; // we need to forward-declare those classes which should not be cleaned up
 
 namespace o2::ccdb
@@ -90,7 +88,6 @@ class CCDBManagerInstance
   template <typename T>
   T* get(std::string const& path)
   {
-    // TODO: add some error info/handling when failing
     return getForTimeStamp<T>(path, mTimestamp);
   }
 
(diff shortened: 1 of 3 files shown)


shahor02 (commented):
Hi @sawenzel 
I am wondering if we should not add a data member switch to the CCDBManager, and then the decision to produce a fatal will be taken as && of this global switch and the extra argument you added.

sawenzel:
It is possible but not sure if it's a good approach as this thing is a singleton. We might get influenced by someone else's switch. In the current approach, we could decide based on types and paths. But I can also live with the global switch.

jgrosseo:
From the discussion yesterday, I had actually thought it is a switch in the manager. I find this better than an extra argument (additional default arguments quickly go out of hand because they get more and more)

jgrosseo:
It is a singleton per process, right?

shahor02:
@sawenzel this singleton anyway has some settable properties + cache and it is either configured by a single used for his process.

sawenzel:
I'll modify this afternoon. Currently in a training.

sawenzel:
@shahor02 @jgrosseo : Reworked according to your suggestion. Beware that for objects returning just meta-data (such as `RunInformation` ?) you will now have to toggle the switch in the manager. Or we introduce a new function just returning the meta-data answer.

jgrosseo:
> @shahor02 @jgrosseo : Reworked according to your suggestion. Beware that for objects returning just meta-data (such as `RunInformation` ?) you will now have to toggle the switch in the manager. Or we introduce a new function just returning the meta-data answer.

Thanks for reworking this!

We shouldn't merge yet then (it will break things in O2Physics). Yes, we need a new function...

sawenzel:
@jgrosseo : I checked the situation in O2Physics. It appears that RunInformation is directly fetched using the ccdbApi.getHeaders function. So I think it is ok to merge this.

jgrosseo:
I see. OK - agreed!

jgrosseo (approved):

ktf:
macOS x86 seems to have unrelated hw troubles. Merging this as it passes tests on macOS ARM.

PR #7871, opened 2021-12-13
TPC: Restructure pad-wise calibration, implement missing workflows

wiechula:
* Generalize and rename pedestal workflow
      - to be able to process pulser and CE data
      - use digits as input
* Inernally use unordered_map of CalPad in pedestal and pulser calib
* Write Pedestal and Noise output to one ccdb entry
* CalDetMergerPublisherSpec can take unordered_map of CalPads
* MonitorWorkflowSpec can use digits as input


wiechula:

PR #7876, opened 2021-12-14
TPC: Make FileWriterSpec buffer data until all sectors received

wiechula:
Using multiple decoders it happens that data from different time frames
are received intermixed. To avoid wrong writing a wrong order to the
file, the data is buffered and only written once data from all sectors
are received.

### DIFF
--- a/Detectors/TPC/workflow/include/TPCWorkflow/FileWriterSpec.h
+++ b/Detectors/TPC/workflow/include/TPCWorkflow/FileWriterSpec.h
@@ -19,16 +19,34 @@
 #include "Framework/DataProcessorSpec.h"
 #include <string>
 
-namespace o2
+namespace o2::tpc
 {
-namespace tpc
-{
+
+enum class BranchType {
+  Krypton,
+  Digits,
+};
+
+const std::unordered_map<std::string, BranchType> BranchTypeMap{
+  {"krypton", BranchType::Krypton},
+  {"digits", BranchType::Digits},
+};
+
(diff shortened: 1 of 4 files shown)


davidrohr:
I assume this is a TPC internal writer for krypton data? For normal TPC digits / clusters, the general logic should actually make sure to receive clusters for all sectors for the same TF at a given time.

wiechula:
Yes, this is a writer for special krypton raw data. How is it done for the general data flow, that the data of the same TF arrive? I didn't find this.

wiechula:
@shahor02 , @davidrohr , can this be merged?

shahor02:
> How is it done for the general data flow, that the data of the same TF arrive? I didn't find this.

Does not  TPCSectorCompletionPolicy (which you use also in tpc-krypton-clusterer) provide this ?

shahor02 (approved):

wiechula:
> > How is it done for the general data flow, that the data of the same TF arrive? I didn't find this.
> 
> Does not TPCSectorCompletionPolicy (which you use also in tpc-krypton-clusterer) provide this ?

It seems not. Perhaps I use it in the wrong way...

davidrohr:
I think TPCSectorCompletionPoicy should do this, if you use it in the right way. Just loading the policy will not work, you have to configure it for which inputs it is active.
See Line 759 in GPUWorkflowSpec.cxx:
```
auto createInputSpecs = [&tpcsectors, &specconfig, policyData]() {
```
and check how policyData is used.

PR #7877, opened 2021-12-15
use natan2, index for ambigoustracks

jgrosseo:


jgrosseo:
@aalkin here the (bad) solution...

aalkin:
It does not break anything, and those two templates did make the resolution ambiguous, so it will do for now. I'll fix the issue with templates for binary functions in a subsequent PR.

PR #7881, opened 2021-12-15
DPL: get rid of mutex for RegionInfo events

ktf:


ktf:
@davidrohr if you have a bit of time could you verify this with the GPU setup? It works fine in my local tests.

davidrohr:
@ktf : I am not fully sure what I need to verify. I can try whether it works with the GPU and the registration at all by just running with dev+this PR on the EPNs. Or should I check whether there are race conditions now? The latter would be quite tricky, since it would mean I would have to set up a test bench to trigger the race conditions.

I think instead we should specify the requirements, and if you want to make sure they are fulfilled, we could introduce unit tests for the framework side? That would be better than to test with the full system.

Just to clarify again, the requirements are:
- There must be only one Region Callback called at a time.
- All region callbacks that have arrived before the data arrives must have been executed (and must have returned) before the first processing function is called.
Is this still fulfilled after this PR?

ktf:
The race condition will still be there whenever there is a sudden death, although it will be less likely.
I agree we should have a proper unit test, it's just not clear to me what the GPU code / DataDistribution does to trigger the extra RegionInfo registrations. In all my test I see a single callback being reported.

In principle the new code should do what you require. The only limit is that if more than 2048 events are reported and the callbacks are not invoked, it will start losing events.

davidrohr:
I don't understand your point with the race condition. This mutex can be limited to the scope of one process, i.e. within a process, multiple threads must be mutexed.
Multiple processes can do in parallel whatever they want.

If you run the start-tmux.sh script in the prodtests/full-system-test (on a dataset created by the fst and converted to tf file) you will get 5 region callbacks, 3 from dpl regihns for the 3 managed segments wuth different segment-id, and 2 from dd for its 2 unmanaged segments.

Kind Regards
David Rohr

ktf:
So, there are two mutexes involved. One which is used to make sure callbacks are not run during processing, which this PR gets  rid of, and one which is in shared memory and that FairMQ uses internally to synchronise RegionInfo notification to all processes. This PR gets rid of the former, reducing a possible race condition which could happen on the latter in the event of a sudden death of  a process.

davidrohr:
OK, but I do not understand. If this PR gets rid of the former one, is it still guaranteed that callbacks are not executed during processing? That would mean the mutex wasn't needed in the first place?

I have no easy way to check whether callbacks are executed during the processing. For this I would need to hack together some test case, where I create new regions during runtime, while I know that some long-running processing is running.

I can only check if this PR works in general with GPU processing. Will do so tonight.

davidrohr (approved):
OK in the sense that GPU processing in the full system test still works with this PR. Note that I have not checked for race conditions during region callbacks, in my tests the region callbacks are arriving way before the first data, so it is automatically serialized.

ktf:
Yes, it's still guaranteed. The mutex is there to prevent a multithreading issue when moving the region info from the callback to the DPL queue. The queue itself it's still flushed between one processing invocation and the other.

PR #7882, opened 2021-12-15
DPL: force exit on error after 15 seconds

ktf:
Apparently the kill -9 was done only after a ctrl-c, not on a
generic error resulting in a workflow getting stuck.


ktf:
This should get any stuck workflow to terminate.

jgrosseo:
👍

PR #7885, opened 2021-12-16
CCDBPopulator adds runNumber (if any) tag to metadata

shahor02:
Since the tag might be queried, define its name via common constexpr.

@wiechula to make sure the same tag name is used I've also modified its setting in the TPC part.


shahor02:
Hi @TimoWilken 
The macOS queue grows with time instead of shrinking, is this normal?

wiechula (approved):
@shahor02 , sure fine with me. Thanks for the changes!

PR #7886, opened 2021-12-16
[EMCAL-566] Add executable for performing time and bad cell calibration

jokonig:
- executable to run calibration, both time and bad channel, in offline
mode with local root histograms. Goal is to validate and debug the
calibration using run2 data.

- modified root to boost histogram conversion function: Specified the
return type, corrected bin numbers for boost histograms

- Added OpenMP option to cmake list. Goal is to parallelize the
time-calibration as many independent fits are performed

- Added implementation of time calibration in EMCALCalibExtractor

### DIFF
--- a/Common/Utils/include/CommonUtils/BoostHistogramUtils.h
+++ b/Common/Utils/include/CommonUtils/BoostHistogramUtils.h
@@ -24,6 +24,7 @@
 #include <TH2.h>
 #include <TFile.h>
 
+#include "Framework/Logger.h"
 #include "Rtypes.h"
 #include "TLinearFitter.h"
 #include "TVectorD.h"
@@ -42,6 +43,9 @@
 #include <boost/histogram/axis.hpp>
 #include <boost/histogram/make_histogram.hpp>
 #include <boost/histogram/accumulators/mean.hpp>
+
+using boostHisto2d = boost::histogram::histogram<std::tuple<boost::histogram::axis::regular<double, boost::use_default, boost::use_default, boost::use_default>, boost::histogram::axis::regular<double, boost::use_default, boost::use_default, boost::use_default>>, boost::histogram::unlimited_storage<std::allocator<char>>>;
+using boostHisto1d = boost::histogram::histogram<std::tuple<boost::histogram::axis::regular<double, boost::use_default, boost::use_default, boost::use_default>>, boost::histogram::unlimited_storage<std::allocator<char>>>;
 
 namespace o2
 {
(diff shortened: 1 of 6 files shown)


Detectors/EMCAL/calibration/CMakeLists.txt
```diff
@@ -10,8 +10,9 @@
 # or submit itself to any jurisdiction.
 
 o2_add_library(EMCALCalibration
-               SOURCES  src/EMCALChannelData.cxx 
-                        src/EMCALTimeCalibData.cxx 
+                TARGETVARNAME targetName
```
mfasDa: Was probably unintended

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALCalibExtractor.h
```diff
@@ -88,16 +95,46 @@ class EMCALCalibExtractor
...
   {
+
+    auto histReduced = boost::histogram::algorithm::reduce(hist, boost::histogram::algorithm::shrink(minTime, maxTime), boost::histogram::algorithm::shrink(0, NCELLS));
+
     o2::emcal::TimeCalibrationParams TCP;
-    TCP.addTimeCalibParam(1234, 600, 0);
+
+#ifdef WITH_OPENMP
+    if (mNThreads < 1) {
+      mNThreads = std::min(omp_get_max_threads(), NCELLS);
```
mfasDa, line 107: 17644 threads is probably out of order for local machines :) Needs to be checked with @shahor02 and @chiarazampolli where to put the maximum here, I guess omp threads will need to be shared among components.

Detectors/EMCAL/calibration/run/runCalibOffline.cxx
```diff
@@ -0,0 +1,160 @@
...
+{
+
+  bpo::variables_map vm;
+  bpo::options_description opt_general("");
+  bpo::options_description opt_hidden("");
+  bpo::options_description opt_all;
+  bpo::positional_options_description opt_pos;
+
+  std::string CalibInputPath = "";
+  std::string ccdbServerPath = "http://ccdb-test.cern.ch:8080";
```
mfasDa: We might want to use 
``` c++
o2::base::NameConf::getCCDBServer();
```
introduced in #7778 as the default server path.

Detectors/EMCAL/calibration/run/runCalibOffline.cxx
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  // instance of the calib extractor
+  o2::emcal::EMCALCalibExtractor CalibExtractor;
+
+  // get boost histo from root input histogram
+  auto hCalibInputHist = o2::utils::boostHistoFromRoot_2D(hCalibInputHist_ROOT);
+
+  // instance of CalibDB
+  o2::emcal::CalibDB calibdb(ccdbServerPath);
```
mfasDa, line 150: @sawenzel @shahor02 Currently the o2::emcal::CalibDB is still using o2::ccdb::CcdbApi as underlying CCDB API. In principle it would be nice to migrate to o2::ccdb::BasicCcdbManager for the CCDB access, but the BasicCcdbManager does not provide write functions, which is i.e. used in this case. Do you see a problem adding a function storing objects which forwards to StoreAsTFileAny to the BasicCcdbManager? The goal of the o2::emcal::CalibDB is to provide a specific interface for EMCAL CDB access so users don't need to mess around with paths.

mfasDa (changes requested):
I think something entered in the CMakeLists by accident that needs to be fixed now. For what concerns the OMP thread allocation I guess this can be handled in the next iterations.

Detectors/EMCAL/calibration/run/runCalibOffline.cxx
shahor02: @mfasDa the custom access to the CCDB from the user code in the workflows should be anyway outphased once the DPL CCDB service becomes functional.

Detectors/EMCAL/calibration/CMakeLists.txt
jokonig: The "TARGETVARNAME targetName" is needed for OpenMP. Based on O2/Detectors/TOF/calibration/CMakeLists.txt

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALCalibExtractor.h
jokonig: Yes NCells is unrealistic. But like this it will use all available resources when calling the omp_get_max_threads(), right?

Common/Utils/include/CommonUtils/BoostHistogramUtils.h
```diff
@@ -215,24 +219,24 @@ std::string createErrorMessage(o2::utils::FitGausError_t errorcode);
...
     first++;
     last--;
   }
 
-  TLinearFitter fitter(3, "pol2");
-  TMatrixD mat(3, 3);
-  Double_t kTol = mat.GetTol();
+  static TLinearFitter fitter(3, "pol2");
+  static TMatrixD mat(3, 3);
+  static Double_t kTol = mat.GetTol();
```
mfasDa, line 239: I suspect this is to circumvent the problem with the multiple definition of TFormula when running with OMP threads > 1. This could become tricky in a multi-threaded environment. As far as I understand the variables should be allocated per thread (thread-local storage) based on our build settings. However would be good if we could find a way to avoid making members static.

shahor02:
@jokonig could you please apply changes suggested by the fullCI (braces)?

mfasDa (approved):

mfasDa:
@shahor02 Should now be ready for merge.

shahor02 (approved):

PR #7887, opened 2021-12-16
DPL: demangle stacktraces, if possible.

ktf:


ktf:
@sawenzel I think this is new:

```
[23:13:44][INFO] Setting up CAVE to host ZDC
Info in <TGeoManager::SetTopVolume>: Top volume is cave. Master volume is cave
Info in <TGeoNavigator::BuildCache>: --- Maximum geometry depth set to 100
Error in <TGeoCompositeShape::MakeNode>: Expression has no boolean operation
Error in <TGeoCompositeShape::ctor>: Composite halfFixationToPipe: cannot parse expression: pipeSupportTubeCarbon
[23:13:44][INFO] Loaded TOF geometry
Info in <TGeoManager::CheckGeometry>: Fixing runtime shapes...
[23:13:45][INFO] GeometryBuilder::buildGeometry volume name = MFT
```

ktf:
This works only for exceptions which are thrown and caught by DPL, it cannot do anything for libSegFault, though...

PR #7890, opened 2021-12-17
DPL Analysis: groundwork for caching the slicing information of a table

aalkin:
@jgrosseo @ktf This adds a cache for a `Table` object that holds the results of arrow's `value_counts` function, so that individual slices can be requested without rerunning it (e.g. in a loop). The caveat is that this will not work on `const` process function arguments. I'll improve that later with the same approach that was used to cache the filters. For now, this already makes slicing much faster for non-const tables (in my task has it has saved about 1/3 of runtime), like the ones coming from partitions or temporaries created in-process.

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -2122,7 +2122,7 @@ class FilteredBase : public T
   {
     uint64_t offset = 0;
     std::shared_ptr<arrow::Table> result = nullptr;
-    auto status = ((table_t)this)->getSliceFor(value, node.name.c_str(), result, offset);
+    auto status = ((table_t*)this)->getSliceFor(value, node.name.c_str(), result, offset);
     if (status.ok()) {
       auto start = offset;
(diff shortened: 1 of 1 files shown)


Framework/Core/include/Framework/ASoA.h
```diff
@@ -2066,6 +2118,29 @@ class FilteredBase : public T
...
 
+  auto sliceByCached(framework::expressions::BindingNode const& node, int value)
+  {
+    uint64_t offset = 0;
+    std::shared_ptr<arrow::Table> result = nullptr;
+    auto status = (table_t)this->getSliceFor(value, node.name.c_str(), result, offset);
+    if (status.ok()) {
+      auto start = offset;
+      auto end = start + result->num_rows();
+      auto start_iterator = std::lower_bound(mSelectedRows.begin(), mSelectedRows.end(), start);
```
ktf, line 2129: Isn't this `std::equal_range`?

ktf:
Tested on hyperlooop. Merging.

PR #7891, opened 2021-12-17
Propaedeutic for CCDB - DPL integration

ktf:


Framework/Core/src/CCDBHelpers.cxx
```diff
@@ -0,0 +1,36 @@
...
+          continue;
+        }
+        LOGP(info, "The following route is a condition {}", route.matcher);
+        for (auto& metadata : route.matcher.metadata) {
+          LOGP(info, "- {}", metadata.name);
+        }
+      }
+      
+      return adaptStateless([](DataAllocator& ctx) {
+                              }); });
```
ktf, line 33: @shahor02 here is where we should plug the "blob only" CCDB manager.

ktf:
Should not change anything for the moment. Merging so that other developments do not have this in the way.

PR #7892, opened 2021-12-17
IDCs: Adding the possibility to group pads at the sector edges

matthias-kleiner:


Detectors/TPC/calibration/include/TPCCalibration/IDCCCDBHelper.h
```diff
@@ -47,15 +47,12 @@ class IDCDelta;
 
 /// \tparam DataT the data type for the IDCDelta which are stored in the CCDB (short, char, float)
 template <typename DataT = short>
-class IDCCCDBHelper
+class IDCCCDBHelper : public o2::ccdb::BasicCCDBManager
```
shahor02: It is not a good idea to derive non-sigleton class from the singleton BasicCCDBManager. For that purpose better use directly underlying CCDBManagerInstance. 
More general, where are you supposed to use this helper. Once the CCDB DPL support will be finalized, it will be disadvised to use local CCDB managers in the standard workflows, the objects will be provided to the DPL device as usual DPL inputs to which device needs to be subscribed.

sawenzel: I agree. Independently, why inheriting? Did you consider composition/usage over inheritance?

matthias-kleiner: Hi Ruben and Sandro, thanks for the comment and the further information. I have currently only used this class for debugging, i.e. loading the IDCs from the CCDB and then drawing or accessing them. I think parts of this functionality will be used later for QA purposes or when applying the space-charge fluctuation correction. 
I think I will then just remove the CCDB stuff from this class and implement a method to set the IDC members manually?

shahor02: @matthias-kleiner yes, better to set the ccdb objects from outside, to be able to use both the DPL service and CCDBManager depending on circumstances. You can add 2 methods 
```
void fetchCCDBObjects();
void fetchCCDBObjects(o2::framework::ProcessingContext& pc);
```
the 1st to fetch what you need via BasicCCDBManager::instance, and the 2nd to get them via 
`pc.inputs().get<>(...)`

wiechula (approved):
Fine from my side.

shahor02 (approved):

PR #7895, opened 2021-12-20
TRD: fix for Vd and ExB calibration and a bit more tracking QC

martenole:
We use `o2::trd::AngularResidHistos` as container for our vDrift and ExB calibration. Since we only collect sums of histogram bin entries and the number of bins these are trivially merged and we don't need to have a vector or span of these objects. During the processing this should be fine now as it is. When the histograms are written to a root file there is still one entry in the tree per time frame and this makes the root file very large for longer productions. But I don't see a way at the moment to avoid this, because the logic of the time-slot-based calibration requires one entry per TF for the processing.
When processing 50k TFs from the pilot beam I get a 1.2 GB large file which could in principle be reduced to less than 1 MB. But as said during the processing on the aggregator node it will accumulate the data and not keep so many objects in the memory in parallel so for now this might be fine.
Maybe I could add a custom writer and add an option to the calibrator (the one which would run on the aggregator node) such that in case it is processing from a ROOT file then it requires only 1 TF per slot.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/AngularResidHistos.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/AngularResidHistos.h
@@ -31,12 +31,14 @@ class AngularResidHistos
   AngularResidHistos() = default;
   AngularResidHistos(const AngularResidHistos&) = default;
   ~AngularResidHistos() = default;
+  void reset();
   bool addEntry(float deltaAlpha, float impactAngle, int chamberId);
   float getHistogramEntry(int index) const { return mHistogramEntries[index]; }
   int getBinCount(int index) const { return mNEntriesPerBin[index]; }
   size_t getNEntries() const { return mNEntriesTotal; }
 
-  void fill(const gsl::span<const AngularResidHistos> input);
+  void fill(const AngularResidHistos& input);
+  void fill(const gsl::span<const AngularResidHistos> input); // dummy!
   void merge(const AngularResidHistos* prev);
   void print();
 
--- a/DataFormats/Detectors/TRD/src/AngularResidHistos.cxx
+++ b/DataFormats/Detectors/TRD/src/AngularResidHistos.cxx
@@ -19,6 +19,13 @@
 using namespace o2::trd;
 using namespace o2::trd::constants;
 
+void AngularResidHistos::reset()
+{
+  mHistogramEntries.fill(0);
+  mNEntriesPerBin.fill(0);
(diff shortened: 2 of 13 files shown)


shahor02:
> When the histograms are written to a root file there is still one entry in the tree per time frame and this makes the root file very large for longer productions. But I don't see a way at the moment to avoid this, because the logic of the time-slot-based calibration requires one entry per TF for the processing.

Do you mean that the output gets large when the reco output gets saved to the tree for the further processing starting from the root reader? For the test / validation 1.2 GB file is not too large and in the production you don't need this, right?
Anyway, strictly speaking, for the time-stot-based calibration you don't need to ship data per TF, e.g. if your slot is 1000 TFs long, you can ship data merged in any divisor of 1000 TF: provided the merged block fits to the slot, you just need to take care that the message headers correspond to the 1st TF of the slot

martenole:
> Do you mean that the output gets large when the reco output gets saved to the tree for the further processing starting from the root reader? For the test / validation 1.2 GB file is not too large and in the production you don't need this, right? 

Yes that's right

> you just need to take care that the message headers correspond to the 1st TF of the slot

But how can I do that? I tried in https://github.com/AliceO2Group/AliceO2/pull/7895/commits/a778f81c0bc4450e2d633228288c36c70f66fa67 to send the output only for every 200 TFs and thought I would prepare the header only for the first TF of that block. But when I run the calibration afterwards from ROOT file it treats every message of 200 TFs as a single TF.
I.e. in Detectors/TRD/workflow/include/TRDWorkflow/VdAndExBCalibSpec.h `auto tfcounter = o2::header::get<o2::framework::DataProcessingHeader*>(pc.inputs().get("input").header)->startTime` still counts up continuously from 1 to 4

Edit: forgot to ping @shahor02

shahor02:
@martenole :  with the current state of framework / time-slot calibration, where the header->startTime is contiguous counter (incremented by 1 for every TF/tree-entry when the data comes from the root file) you indeed cannot do this directly. Instead of sending the output once every 200 TFs you would need to send an empty output in the remaining 199 TFs to have dummy entries. But the time-slots will be eventually based on the real time intervals rather than counters.

martenole:
Ok thanks then I will for now leave it as it is and wait for the changes in the framework. For my tests I can simply reduce the number of TFs required for a slot, taking into account that a counter increment by one indicates 200 TFs. That way the debugging output size is reduced a lot.
If I send empty dummy output it still increases the size and does not really help me.

bazinski (approved):

PR #7896, opened 2021-12-21
Reducing memory consumption of the digitization in AliceO2/prodtests/sim_challenge.sh

zovarga:
Added features:
- Digitization step split into 3 parts (TPC, TRD, REST) to reduce memory consumption.
- Usage() now prints a line saying the code is deprecated.


sawenzel (approved):
Approving to trigger test (newcomer)

PR #7898, opened 2021-12-21
TRD TRAP simulation refactoring

martenole:
Hi @tdietel and @bazinski ,
this is the announced PR for the TRAP simulation adaption to Venelin's C++ model of the TRAP. When using the correct LUT for the y-position correction the results to his model are identical. Inside the `mMCMT` array I store the same 32 bit tracklet words as in the actual TRAP including the header from the first CPU.
We need to think about how we want to handle the slope. In the current TRAP it has the wrong sign which I correct for simply by returning the slope with the correct sign by `Tracklet64::getSlopeBinSigned()`. We could also change the TRAP, but then we need to handle the pilot beam data differently then the data we take in Run 3.
There are also some more questions in the code, for example which ADC channels should be checked for the charge calculation of the tracklets. Currently it is the leftmost channel + 3 channels to the right, but one could also choose the channels symmetric around those ones which are taken into account for the fit.
Also some values are still hard-coded now which should go into the TRAP config. For example the fit window and the cluster threshold for tracklet candidates. Someone would need to take a look at the tracklet PID to judge this, which anyhow needs to be done for the gain calibration.
As mentioned in the mail the granularity for the slope is 1/1024 pad, but it is realized by a scaling factor of 1/128 pad and another division by 8 by a different bit shift. I implemented it in the same way here now.
What I am not sure about is the double tracklet rejection. It seems it is active always, but before we still had a flag in the FeeParam for that. Maybe that is not needed anymore?
So, still some work to do for the tracklets. Maybe we can chat about this at the beginning of next year?
Cheers,
Ole


DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
```diff
@@ -51,10 +51,12 @@ constexpr int NCOLMCM = 18;     // the number of pads per MCM
 
 constexpr int NBITSTRKLPOS = 11;                   // number of bits for position in tracklet64 word
 constexpr int NBITSTRKLSLOPE = 8;                  // number of bits for slope in tracklet64 word
-constexpr float PADGRANULARITYTRKLPOS = 40.f;      // TODO this should come out of the trapconfig.
-constexpr float PADGRANULARITYTRKLSLOPE = 1000.f;
+constexpr int ADDBITSHIFTSLOPE = 1 << 3;           // in the TRAP the slope is shifted by 3 additional bits compared to the position
```
tdietel, line 56: I guess the original TODO comment still applies: this should come out of the trapconfig.

DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
```diff
@@ -81,16 +81,16 @@ class Tracklet64
...
-  GPUd() uint64_t getPadRow() const { return ((mtrackletWord & padrowmask) >> padrowbs); }; // pad row number [0..15]
-  GPUd() uint64_t getColumn() const { return ((mtrackletWord & colmask) >> colbs); };       // column refers to MCM position in column direction on readout board [0..3]
-  GPUd() uint64_t getPosition() const { return ((mtrackletWord & posmask) >> posbs) ^ 0x80; };  // in units of 1/80 pads, 11 bit granularity [-12.8..12.8] relative to MCM center
-  GPUd() uint64_t getSlope() const { return ((mtrackletWord & slopemask) >> slopebs) ^ 0x80; }; // in units of 1/1000 pads/timebin, 8 bit granularity [-0.128 to 0.128]
-  GPUd() uint64_t getPID() const { return ((mtrackletWord & PIDmask)); };                   // no unit, all 3 charge windows combined
-  GPUd() uint64_t getQ0() const { return ((mtrackletWord & Q0mask) >> Q0bs); };             // no unit
-  GPUd() uint64_t getQ1() const { return ((mtrackletWord & Q1mask) >> Q1bs); };             // no unit
-  GPUd() uint64_t getQ2() const { return ((mtrackletWord & Q2mask) >> Q2bs); };             // no unit
+  GPUd() int getFormat() const { return ((mtrackletWord & formatmask) >> formatbs); };     // no units 0..15
```
tdietel, line 90: Just a comment: shouldn't this use an explicit size, e.g. int32_t?

tdietel (commented):
Looks good, I just have a few minor comments. 
I did not go through the calculations of the slope in detail, but I assume this has been checked with data.
Let's see if the others have comments.

DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
martenole: Hm, I don't see a necessity for using an explicit size here. We can be sure that `int` holds at least 2 bytes which is sufficient for all methods except for `getPID()` which requires 4 bytes. But 4 bytes is apparently the de-facto standard for an `int` so I guess this is fine: https://stackoverflow.com/questions/589575/what-does-the-c-standard-state-the-size-of-int-long-type-to-be

DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
martenole: Ok, I will put the TODO comment back in place. Implementing this will require also changes in the TrackletTransformer which needs to know the granularities for slope and position. Or I could also change this now, should I better do that?

martenole:
Thanks for the comments @tdietel. What do we do about the sign of the slope and the additional XOR for slope and position of the tracklets? We could leave it as it is now, but I think it would be a cleaner solution of we would store the slope and the position inside the `Tracklet64` in a way such that no additional `^0x80` or `*-1` is needed. The raw reader would need to do this when the `Tracklet64`s are created.
The disadvantage is of course that we would need to treat the pilot beam data differently from the data we take after the change.

Concerning the checks with data I did test the reconstruction with the corrected granularity and slope sign which seems fine, but I would still like to run the TRAP simulation for some data we took during the pilot beam. For that I would need some raw TF data where we have TRD digits, tracklets and ideally also some global tracks. Not sure if I will find some TFs which have all of that, but I will start to compare digits and tracklets from raw data with simulated tracklets from those digits.

DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
tdietel: Good point. It's not stored anywhere, so the size really does not matter.

bazinski:
The raw reader can do what ever we want, and can trivially add a new command line parameter, particularly given the current rewrite. The question is what exactly do we want tracklet class to store internally.... 
with regards the XOR, I would say keep the tracklet64 class contents identical to the output of the trap. It also means no hoop jumping for the pilot beam data. I suppose this assumes/implies the pilot beam data is going to be used at some significant level past march? In anycase I am easy either way.

bazinski:
@tdietel @martenole on second thoughts, off the top of my head, we should probably build a version into the spare bits of the 64 to allow for the other options of pid as well, this then could also take care of the xor and the -1 then as well. Will think about that some more though.

tdietel:
> The raw reader can do what ever we want, and can trivially add a new command line parameter, particularly given the current rewrite. The question is what exactly do we want tracklet class to store internally.... with regards the XOR, I would say keep the tracklet64 class contents identical to the output of the trap. It also means no hoop jumping for the pilot beam data. I suppose this assumes/implies the pilot beam data is going to be used at some significant level past march? In anycase I am easy either way.

I am leaning towards Sean's view. I think it makes sense to leave the tracklet64 class as close as possible to the implementation in the TRAP.

tdietel:
> @tdietel @martenole on second thoughts, off the top of my head, we should probably build a version into the spare bits of the 64 to allow for the other options of pid as well, this then could also take care of the xor and the -1 then as well. Will think about that some more though.

That would need some thought. In principle, we have a lot of freedom in the calculation of the tracklets, and there could be many different settings. Change of fit ranges and charge integration windows are the most obvious. And it could become tricky to synchronize all the changes between MCM, raw data, tracklet64 and trap simulator.

I think the central piece will be the trapconfig that we need to somehow get into the CCDB.

martenole:
I like the idea of using the spare bits for either a versioning of the tracklets or as individual flags for XOR, -1 and so on. The idea of the Tracklet64 class was to have a format which has all the information we need easily accessible without needing additional information from other places. So e.g. take the MCM position into account and so on. The XOR is only needed for the readout, but later on it is an unnecessary complication. That is why I would like to drop it. And the -1 we could ask Venelin to fix. This code snippet shows what I don't like about the current 'fix':
```
root [0] o2::trd::Tracklet64 trklt
(o2::trd::Tracklet64 &) @0x7fbc6cde5018
root [1] trklt.getPosition()
(unsigned long) 128
root [2] trklt.setPosition(trklt.getPosition())
root [3] trklt.getPosition()
(unsigned long) 0
```
The Tracklet64 is by definition already different from the actual TRAP output.

bazinski:
> The Tracklet64 is by definition already different from the actual TRAP output.

But the output has a xor of 0x80 on the position and slope?

tdietel (approved):
Let's go ahead with the merge. Looks like all comments have been addressed, and the results look good.

PR #7900, opened 2021-12-22
File blobs fetching for DPL CCDB service

shahor02:
@ktf I've modified CcdbApi and CCDBManagerInstance to fetch and catch the CDB files to memory, e.g.
```
o2::ccdb::CCDBManagerInstance mm("http://alice-ccdb.cern.ch");
mm.setTimestamp(1654257300000);
auto b = mm.getBlob("HMP/Align");
```
to receive the blob, from which the object can be extracted using a static method

```
auto al = o2::ccdb::CcdbApi::extractFromMemoryBlob<std::vector<o2::detectors::AlignParam>>(*b);
```

At the moment the blob is simple `std::vector<char>`, if you tell me how  CCDBManagerInstance should access the raw device,  I will convert it to FairMQ message. From your comment https://github.com/AliceO2Group/AliceO2/pull/7891/files/4eb8448babc24209523f4ba22f4ab3bf6e8b91bd#r771505147 I did not understand how exactly the blob manager should be attached. 
Also, I see two potential problems:
1) to create a FMQ message I have to extract the channel from the raw device and usually the channel depends not only the message specs but also on the TF counter in case of pipelines, e.g. https://github.com/AliceO2Group/AliceO2/blob/b4c482bb2d536fbf64e8c085545c101421e5ee96/Detectors/Raw/src/RawFileReaderWorkflow.cxx#L171
So, in case of pipelines the cached message will be valid for the 1st processed lane. Do you have a way to hack the channel before sending it?

2) We foresee a possibility to use different CCDB sources for different conditions, while the  CCDBManagerInstance is reading just 1 source (one can surely reset the server in the given instance of the manager, but this will lead to cleaning the cache and extra latency for the new connection to server). Shall I create a container class for multiple  CCDBManagerInstance instances, each one for a particular source + the mapping between some URLs and non-default servers?

### DIFF
--- a/CCDB/include/CCDB/BasicCCDBManager.h
+++ b/CCDB/include/CCDB/BasicCCDBManager.h
@@ -45,11 +45,23 @@ class CCDBManagerInstance
     void* noCleanupPtr = nullptr; // if assigned instead of objPtr, no cleanup will be done on exit (for global objects cleaned up by the root, e.g. gGeoManager)
     std::string uuid;
     long startvalidity = 0;
-    long endvalidity = 0;
+    long endvalidity = -1;
+    bool isBlob = false; // is this raw blob or parsed object?
     bool isValid(long ts) { return ts < endvalidity && ts > startvalidity; }
+    void clear()
+    {
+      objPtr.reset();
+      uuid = "";
+      startvalidity = 0;
+      endvalidity = -1;
+      isBlob = false;
+    }
   };
 
  public:
+  using BLOB = std::vector<char>;
+  using MD = std::map<std::string, std::string>;
+
   CCDBManagerInstance(std::string const& path) : mCCDBAccessor{}
   {
     mCCDBAccessor.init(path);
@@ -77,7 +89,7 @@ class CCDBManagerInstance
 
   /// retrieve an object of type T from CCDB as stored under path, timestamp and metaData
   template <typename T>
-  T* getSpecific(std::string const& path, long timestamp = -1, std::map<std::string, std::string> metaData = std::map<std::string, std::string>())
+  T* getSpecific(std::string const& path, long timestamp = -1, MD metaData = MD())
   {
     // TODO: add some error info/handling when failing
     mMetaData = metaData;
@@ -90,6 +102,11 @@ class CCDBManagerInstance
   {
     return getForTimeStamp<T>(path, mTimestamp);
   }
+
+  /// aliases for BLOB retrieval
+  BLOB* getBlobForTimeStamp(std::string const& path, long timestamp) { return getForTimeStamp<BLOB>(path, timestamp); }
+  BLOB* getSpecificBlob(std::string const& path, long timestamp = -1, MD metaData = MD()) { return getSpecific<BLOB>(path, timestamp, metaData); }
+  BLOB* getBlob(std::string const& path) { return get<BLOB>(path); }
 
   bool isHostReachable() const { return mCCDBAccessor.isHostReachable(); }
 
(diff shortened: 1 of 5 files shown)


CCDB/include/CCDB/BasicCCDBManager.h
```diff
@@ -91,6 +103,11 @@ class CCDBManagerInstance
     return getForTimeStamp<T>(path, mTimestamp);
   }
 
+  /// aliases for BLOB retrieval
+  BLOB* getBlobForTimeStamp(std::string const& path, long timestamp) { return getForTimeStamp<BLOB>(path, timestamp); }
+  BLOB* getSpecificBlob(std::string const& path, long timestamp = -1, MD metaData = MD()) { return getSpecific<BLOB>(path, timestamp, metaData); }
+  BLOB* getBlob(std::string const& path) { return get<BLOB>(path); }
```
shahor02, line 109: @ktf these are the methods which can be used to extract/cache the blob.

ktf:
Regarding 1. we can probably use `DataAllocator::getMemoryResource` or introduce a new specialization `make<FairMQMessage>` I will try to do it ASAP. Regarding 2. I would prefer we simply keep a list of rules to decide which url to use given the metadata of the InputSpec.

shahor02:
@ktf: for (2), I meant that if we want to be able to fetch objects from different servers, we should either instantiate a  CCDBManagerInstance per server (+ some mapping from inputSpecs on particular manager instance) or to reinitialize the underlying CcdbApi with other server string each time we read from different server, but this may have significant overheads (connection, credentials check)

costing:
The credentials are attached to each connection, irrespective of the server address. So there shouldn't be any overhead from having different CcdbApi calls with different server address(es). What matters is only if the request is made or not and there the in-memory caching should help.

Since #7680 introduced support for multiple servers, I suggest an approach like:
- a default address, globally configured
- specific storages as needed, that take the form of a different base URL. Where they exist, the specific storages are prepended to the above string (comma separated)

With this the list of rules is kept to only the overrides, and non-existing objects in the specific storage are transparently handed over to the default location.

shahor02:
Thanks @costing ,
The caching is there by default and is not affected by the server change. For the specific storage: i assume if the object from there is requested one should not allow a fall back to default server since the object will likely differ from what is needed.

costing:
In that case it can override, instead of being prepended to, the default URL, so that it becomes the only address that's checked (and fail if not found). 

But I imagine situations where a simple configuration (i.e. all `HMP/.*` objects should be first checked in a specific storage, but only a few are actually there) would be beneficial. 

So what do you think about a convention/flag on the specific storage configuration, that chooses between the two behaviors? An in-order matching of the regexes and either concatenating (`+`) or overriding the address (without it) would cover a lot of different use cases:
```
.*=https://alice-ccdb.cern.ch/
HMP/.*=+https://alice-ccdb.cern.ch/Users/g/grigoras/
HMP/Calib/.*=+https://alice-ccdb.cern.ch/Users/o/otheruser/
TPC/Align/.*=http://ccdb-test.cern.ch:8080/,http://other.server/
```

shahor02:
Hi @costing 
I think such simple rules with concatenation  or overriding would be useful. Though, the i am not sure we will have a wildcarded request like HMP/.*: current CCDB DPL service assumes that each CCDB object (path) is requested by the device specs as a usual DPL they input.

BTW, @ktf , would it be possible to automatically add the condition type requests to device options, with the CCDB path as an option name and the server as a default option? This would save lot of boilerplate code

ktf:
@shahor02 yes, I think it's a good idea, like for the timers. I would use `--<data-origin>-<data-description>`as an option name, though. I will try to add it.

PR #7907, opened 2022-01-03
get timestamp from hbfutil in tof digitisation

noferini:


Steer/DigitizerWorkflow/src/SimpleDigitizerWorkflow.cxx
```diff
@@ -516,6 +516,11 @@ WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
     auto useCCDB = configcontext.options().get<bool>("use-ccdb-tof");
     auto ccdb_url_tof = o2::base::NameConf::getCCDBServer();
     auto timestamp = configcontext.options().get<int>("timestamp-tof");
+
+    if (timestamp < 0) {
+      timestamp = o2::raw::HBFUtils::Instance().startTime / 1000;
+    }
```
shahor02, line 522: For the record: the HBFUtils.startTime at the moment is 0, unless it is set explicitly via command line option. Is it ok for what you want to do with the timestamp?

noferini: timestamp is a parameter of the workflow (default = -1)
In this way I can overwrite the HBFUtils in a local test in case I need it

noferini: overwrite=means use my own timestamp instead of the one provided by utils

shahor02: @noferini ok, but the fullCI is crashing after failing to fetch a TOF CCDB object, I guess you need to upload a default object with infinite coverage.

noferini: That's strange (default entry is supposed to be present). I will have a look

shahor02: Hi @noferini ,
Indeed, the problem appears to be different:
````
89783:ZDCDigitizer]: [18:41:05][WARN] CCDB: Did not find an alien token; Cannot serve objects located on alien:/
````
leads to
````
[ERROR] SEVERE: Device TOFDigitizer (89779) had at least one message above severity fatal: Got nullptr from CCDB for path TOF/Calib/ChannelCalib and timestamp 1641234700000
````
@davidrohr, @TimoWilken: after disabling the http fetching, we need to have an alien token initialized on the build servers...

**Comment by shahor02** [comment]
@davidrohr @TimoWilken 
FullCI again fails with 
```
[79293:TOFDigitizer]: [11:53:15][WARN] CCDB: Did not find an alien token; Cannot serve objects located on alien://
[79293:TOFDigitizer]: [11:53:15][ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch/download/d103d68c-52a8-11ec-bf8f-0aa204261b9a
[79293:TOFDigitizer]: [11:53:15][FATAL] Got nullptr from CCDB for path TOF/Calib/ChannelCalib and timestamp 1641642269000
```
Could you clarify if alien-token is always supposed to be initialized on the build server? We should have it.

noferini:
I close this PR since in the meanwhile many things changed

PR #7909, opened 2022-01-03
[EMCAL-637, O2-2683] Fix extruding DCAL supermodules

mfasDa:
The size of the DCAL supermodules did not take into
account the projective layout and therefore not the
tiling of the inner DCAL modules, leading to extruding
at the inner edge in eta at the inner R. To compensate the
tilting the size of the supermodules had to be increased
by 7 cm, shifting the origin back in PHOS direction by 3.5
cm in order to fix the upper edge of the supermodule. The
relative tower z-position had to be adapted to the origin shift
in order ot keep the towers at the same position.


mfasDa:
FullCI errors unrelated

PR #7912, opened 2022-01-05
DPL Analysis: fall back to per-entry reading of VLA branches with >616 entries

aalkin:
Temporary workaround for VLA reading bug (https://alice.its.cern.ch/jira/browse/O2-2732)


jgrosseo (approved):

ktf:
What's magic about 616?

aalkin:
This is a question for ROOT experts, I just found that the error only appears when there are more than 616 entries.

PR #7913, opened 2022-01-05
Add Origin table

jgrosseo:


jgrosseo:
Unrelated. Merging.

jgrosseo:
@ekryshen @nburmaso @sawenzel 
This table should be filled in the AOD converter with just 1 entry which is the DF id which is used for the folder as well. 
This then later allows to track which DFs have been merged together.

PR #7914, opened 2022-01-06
DPL raw proxy: Supporting message sequences/pruned headers

matthiasrichter:
This is following up to #7110 and O2 1325

More tests are ongoing right now and there is an issue with the message forwarding in the output proxy.

### DIFF
--- a/Framework/Core/include/Framework/ExternalFairMQDeviceProxy.h
+++ b/Framework/Core/include/Framework/ExternalFairMQDeviceProxy.h
@@ -56,12 +56,29 @@ InjectorFunction incrementalConverter(OutputSpec const& spec, uint64_t startTime
 /// multipart ensemble.
 InjectorFunction o2DataModelAdaptor(OutputSpec const& spec, uint64_t startTime, uint64_t step);
 
+/// @struct DPLModelAdapterConfig
+/// Configuration object for dplModelAdaptor
+struct DPLModelAdapterConfig {
+  /// throw runtime error if an input message is not matched by filter rules
+  bool throwOnUnmatchedInputs = true;
+  /// do all kinds of consistency checks
+  bool paranoid = false;
+  /// blindly forward on one channel
+  bool blindForward = false;
+};
+
 /// This is to be used when the input data is already formatted like DPL
 /// expects it, i.e. with the DataProcessingHeader in the header stack
 /// The list of specs is used as a filter list, all incoming data matching an entry
 /// in the list will be send through the corresponding channel
 InjectorFunction dplModelAdaptor(std::vector<OutputSpec> const& specs = {{header::gDataOriginAny, header::gDataDescriptionAny}},
-                                 bool throwOnUnmatchedInputs = true);
+                                 DPLModelAdapterConfig config = DPLModelAdapterConfig{});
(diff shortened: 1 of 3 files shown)


Framework/Core/include/Framework/ExternalFairMQDeviceProxy.h
```diff
@@ -56,12 +56,29 @@ InjectorFunction incrementalConverter(OutputSpec const& spec, uint64_t startTime
...
+/// @struct DPLModelAdapterConfig
+/// Configuration object for dplModelAdaptor
+struct DPLModelAdapterConfig {
+  /// throw runtime error if an input message is not matched by filter rules
+  bool throwOnUnmatchedInputs = true;
+  /// do all kinds of consistency checks
+  bool paranoid = false;
+  /// blindly forward on one channel
+  bool blindForward = false;
+};
```
ktf, line 68: Maybe this should be an enum?

Framework/Core/src/ExternalFairMQDeviceProxy.cxx
```diff
@@ -241,49 +241,63 @@ InjectorFunction dplModelAdaptor(std::vector<OutputSpec> const& filterSpecs, boo
...
           }
-          outputs[channelName].AddPart(std::move(parts.At(msgidx * 2)));
-          outputs[channelName].AddPart(std::move(parts.At(msgidx * 2 + 1)));
-          LOG(debug) << "associating part with index " << msgidx << " to channel " << channelName << " (" << outputs[channelName].Size() << ")";
-          indexDone = true;
           break;
         }
       }
-      if (indexDone == false && !DataSpecUtils::match(query, "DPL", "EOS", 0)) {
+      if (!channelName.empty()) {
```
ktf, line 258: I would invert the logic and have a `continue` on `channelName.empty() == true`.

Framework/Core/include/Framework/ExternalFairMQDeviceProxy.h
matthiasrichter: it's not just definitions of options but we can have multiple active options at the same time. That would mean to bind the enum definitions to bits and then probing for the bits. But true, I wanted to work on a constructor with a flexible argument list to set the members, got a bit out of scope.

Framework/Core/src/ExternalFairMQDeviceProxy.cxx
matthiasrichter: ah, yeah, can revise the logic. Actually I did not want to continue in the loop because the unmatched specs are logged. But I will combine this, there is room for cleanup in the logic.

matthiasrichter:
I think we simply go on with that now and I'm solving the minor cleanups in subsequent commits.

Framework/Core/include/Framework/ExternalFairMQDeviceProxy.h
matthiasrichter: I'm going to implement to paranoid checks and forward optimization in a coming commit, and then I will also add the constructor.

ktf:
Yes, sure. Thank you for the contribution.

PR #7916, opened 2022-01-06
DPL Analysis: inspect Configurables in tasks recursively

ktf:
This will inspect any member of a task recursively, looking for
"Configurables", providing the ability to nest them inside
other structures. This effectively allows making any number of
configurables as long as:
* there is no more than 100 elements in the toplevel struct or in the nested ones.
* the configurables are named uniquely (at least for the moment).

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -896,7 +896,7 @@ using is_soa_iterator_t = typename framework::is_base_of_template<RowViewCore, T
 template <typename T>
 constexpr bool is_soa_iterator_v()
 {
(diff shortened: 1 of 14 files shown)


ktf:
@jgrosseo @aalkin I think this is all we need, actually.

jgrosseo:
Very cool!

jgrosseo:
It somehow broke `aodDataModelGraph` ...

ktf:
yes, but the fix should be easy. i will try to do it today.

ktf:
Should be fixed.

ktf:
Does not break hyperloop.

PR #7917, opened 2022-01-07
bug fix in TOF diagnostics

noferini:


noferini:
This introduces a fix for a bug to avoid to process trm diagnostic words not related to readout channels.

The current bugs produces a mismatch in channel mapping for crate=0 slot<3 (readout slots >=3) (rare events but not for one of the beam test runs).
Needed for async pass2 (@chiarazampolli)

shahor02 (approved):
MacOS CI did not start, merging.

PR #7922, opened 2022-01-07
Fix typo in CovMat calculation for V0s

fmazzasc:
Hi @shahor02 , is it correct?


shahor02 (approved):
@fmazzasc sure, thanks for spotting the typo.

shahor02:
merging as the fix is trivial

PR #7927, opened 2022-01-11
Revert "DPL raw proxy: Supporting message sequences/pruned headers"

davidrohr:
Reverts AliceO2Group/AliceO2#7914

This breaks the processing on the EPN. The raw-proxy segfaults with this PR. If I revert this PR the segfault is gone.

I can reproduce this in the full system test starting from DataDistribution.
It is not seen in the FullCI, since the FullCI does not use DataDistribution, but it starts from the raw reader.

Will merge when CI through, since currently broken.

In order to reproduce:
You can use my DD FST dataset on the EPNs:
```
cd /home/drohr/alitest/tmp-fst10
TFDELAY=5 NTIMEFRAMES=8 $O2_ROOT/prodtests/full-system-test/start_tmux.sh dd
```

Or alternatively generating a DD FST dataset is describer here, for reference: https://alice.its.cern.ch/jira/browse/O2-1492
Then please use the same `start_tmux.sh dd` command to reproduce.

### DIFF
--- a/Framework/Core/include/Framework/ExternalFairMQDeviceProxy.h
+++ b/Framework/Core/include/Framework/ExternalFairMQDeviceProxy.h
@@ -56,29 +56,12 @@ InjectorFunction incrementalConverter(OutputSpec const& spec, uint64_t startTime
 /// multipart ensemble.
 InjectorFunction o2DataModelAdaptor(OutputSpec const& spec, uint64_t startTime, uint64_t step);
 
-/// @struct DPLModelAdapterConfig
-/// Configuration object for dplModelAdaptor
-struct DPLModelAdapterConfig {
-  /// throw runtime error if an input message is not matched by filter rules
-  bool throwOnUnmatchedInputs = true;
-  /// do all kinds of consistency checks
-  bool paranoid = false;
-  /// blindly forward on one channel
-  bool blindForward = false;
-};
-
 /// This is to be used when the input data is already formatted like DPL
 /// expects it, i.e. with the DataProcessingHeader in the header stack
 /// The list of specs is used as a filter list, all incoming data matching an entry
 /// in the list will be send through the corresponding channel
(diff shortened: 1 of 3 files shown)


ktf:
@sawenzel there seems to be a genuine (unrelated to this PR) issue in simulation?

```
/sw/SOURCES/O2/7927-slc8_x86-64/0/Detectors/MUON/MID/Simulation/test/testSimulation.cxx(464): [1;31;49merror: in "o2_mid_simulation/MID_SimClusters/_5": check sortedPC.size() == 1 has failed [0 != 1]
Failure occurred in a following context:
    deId = 20; [0;39;49m
/sw/SOURCES/O2/7927-slc8_x86-64/0/Detectors/MUON/MID/Simulation/test/testSimulation.cxx(466): [1;31;49merror: in "o2_mid_simulation/MID_SimClusters/_5": (
Failure occurred in a following context:
    deId = 20; [0;39;49m
```

ktf:
@matthiasrichter are you ok with the revert or you know what is wrong and can do an hotfix?

davidrohr:
> @matthiasrichter are you ok with the revert or you know what is wrong and can do an hotfix?

ok, but I would like to have a working build for the EPNs tomorrow.
So we can wait a bit but if it is not fixed by tonight, I'll revert it.

ktf:
Yes, agreed. Since it's 3 commits it is always a pain when we revert so if we can avoid it, it would be better. It also breaks QC (@knopers8 @Barthelemy) although it happens sporadically there.

matthiasrichter:
@davidrohr I'm on it, but just merge to get the build up and running.

matthiasrichter (approved):

ktf:
@TimoWilken the build here is ongoing since 11:00 (now 17).

ktf:
Verified by David for full CI. Merging since it's broken in any case.

TimoWilken:
About the lagging CI: there was an issue on some of the hosts with installing PyGithub (as they don't have a C compiler on the host). Using PyGithub 1.54.1 (instead of the current 1.55) worked as a temporary workaround.

PR #7928, opened 2022-01-11
DPL: add ability to cache adopted messages

ktf:


ktf:
@shahor02 
This is what we need to cache adopted blobs at DPL level and send them by cacheId. The receiver is guaranteed that as long as the DataRef.payload pointer is unmodified, the object which was sent is always the same and there is no need to e.g. deserialise it multiple times. The mechanics to use it in my opinion should be something like:

```cpp
o2::pmr::vector<char>&& outputBuffer = ctx.outputs().makeVector<char>(*outputPtr);
// fill the output buffer with the stuff fetched from CCDB.
cacheID = ctx.outputs().adoptContainer(output, std::move(outputBuffer));
```

while you can reuse a cacheID via:

```cpp
ctx.outputs().adoptContainer(output, cacheID);
```

on the sender side. 

If you agree to modify the CCDBManagerInstance so that it takes a o2::pmr::vector to where it needs to write, I can do the changes and provide a concrete example. Alternatively I can try to refactor CCDBManagerInstance::getForTimeStamp and we do not even need the manager, just the CCDB API. Notice that this also means  there is no need to cache the blobs on the CCDBManagerInstance side. We just need to keep a map of the cacheId associated to a given unique identifier for the CCDB object.

shahor02:
@ktf sorry for delay, was taken by other stuff/meetings... I see you have already modified the CCDB part, will comment in https://github.com/AliceO2Group/AliceO2/pull/7933. 
But I am not sure how the cache in your implementation should be used: in the CCDBManager, the cache stores the object unique ID provided by the query: https://github.com/AliceO2Group/AliceO2/blob/45973e4711cd9794f5b293706fc6c8213ba9aacf/CCDB/include/CCDB/BasicCCDBManager.h#L235.
When checking if the cached object is good to use for another timestamp, this uuid is passed to CURL query, which refrains from reading the new entry if the latter has the same uuid as the cached one.
I think you cannot avoid something similar here: you will need one `map<std::string ccdb_url, std::string uuid>` which should be queried to check if the object was already cached. Then, another `map<std::string uuid,  std::unique_ptr<FairMQMessage>>` should cache the message itself. 

If you don't want to put all this into the framework, then one can keep your https://github.com/AliceO2Group/AliceO2/blob/012c0fa3bcfd919d5a6dcde91e751863de311e19/Framework/Core/include/Framework/MessageContext.h#L582 as it, and move url->uuid->yourCacheID mapping to the central CCDB fetcher task, something like:

```
std::unordered_map<std::string, std::string> mapURL2UUID;
std::unordered_map<std::string, int64_t> mapURL2DPLCache;

// process of fetching the object with particular URL
const auto url2uuid = mapURL2UUID.find(url);
std::string uuid;
if (url2uuid != mapURL2UUID.end()) {
  uuid = url2uuid->second;
}
 o2::pmr::vector<char>&& outputBuffer = ctx.outputs().makeVector<char>(*outputPtr);
 ccdbAPI.loadFileToMemory(outputBuffer, url, metadata, timestamp, headers, uuid, createdNotAfter, createdNotBefore);
 if (mHeaders.count("Error"))  {
   throw ...
 }
 if (!uuid.empty) { // object was already cached
   if (outputBuffer.size()) { // but should be overridden by fresh object
     mapURL2UUID[URL] = mHeaders["ETag"]; // update uuid
     mapURL2DPLCache[URL] = ctx.outputs().adoptContainer(output, std::move(outputBuffer));
  } else { // cached object is fine
     ctx.outputs().adoptContainer(output, mapURL2DPLCache[URL] );
     // the outputBuffer was not used, can we destroy it?
  }
} else { // 1st reading of this URL, fill the cache
   mapURL2UUID[URL] = mHeaders["ETag"]; // update uuid
   mapURL2DPLCache[URL] = ctx.outputs().adoptContainer(output, std::move(outputBuffer));  
}
 ```
Note that we will need to destroy the outputBuffer created in advance if the object was already cached

**Comment by ktf** [comment]
Yes, of course, there are two ids. One which is the ETag from the server, the other one which is the cacheId which I use to identify the blob associated to a give etag, without having to know the ETag itself in the consumers (basically I can simply rely on the fact that the resource will never move after it has been created). I was indeed thinking to keep some of the logic in the fetcher itself, similar to what you propose. I think the caveat is that you cannot shallow copy a FairMQMemoryResource, but you need a full FairMQMessage, hence the need for handling things in MessageContext. Anyways, it's still in flux, I think I will have a more concrete proposal by later today.

**Comment by ktf** [comment]
Actually, I think I had in mind exactly what you have there. You are just lacking a couple of `adoptContainer(... true)` where the caching on  the DPL side needs to happen, and an adoptFromCache where the uuid  is available.

**Comment by ktf** [comment]
Merging to simplify integration. Feature is unused by default and all tests pass.

**Comment by shahor02** [comment]
Indeed there were some typos, below is the fixed code.  Also, in case of overriding of cached blob one should `pruneFromCache` with the previous cacheID. But if this code stays in the fetched, one should have a way to access the MessageContext.
Perhaps, the easiest is to add to the  `adoptContainer` and extra argument `uint64_t oldCacheID` like in the commented code below, then `adoptContainer` would do
`if (oldCacheID)  context.pruneFromCache(oldCacheID)`

```
std::unordered_map<std::string, std::string> mapURL2UUID;
std::unordered_map<std::string, int64_t> mapURL2DPLCache;

// process of fetching the object with particular URL
const auto url2uuid = mapURL2UUID.find(url);
std::string uuid;
if (url2uuid != mapURL2UUID.end()) {
  uuid = url2uuid->second;
}
 o2::pmr::vector<char>&& outputBuffer = ctx.outputs().makeVector<char>(*outputPtr);
 ccdbAPI.loadFileToMemory(outputBuffer, url, metadata, timestamp, headers, uuid, createdNotAfter, createdNotBefore);
 if (mHeaders.count("Error"))  {
   throw ...
 }
 if (!uuid.empty) { // object was already cached
   if (outputBuffer.size()) { // but should be overridden by fresh object
 
    // somewhere here pruneFromCache should be called

     mapURL2UUID[URL] = mHeaders["ETag"]; // update uuid
     mapURL2DPLCache[URL] = ctx.outputs().adoptContainer(output, std::move(outputBuffer), true);
    // one could modify the    adoptContainer to take optional old cacheID to clean:
    // mapURL2DPLCache[URL] = ctx.outputs().adoptContainer(output, std::move(outputBuffer), true, mapURL2DPLCache[URL]);
  } else { // cached object is fine
     ctx.outputs().adoptFromCache(output, mapURL2DPLCache[URL] );
     // the outputBuffer was not used, can we destroy it?
  }
} else { // 1st reading of this URL, fill the cache
   mapURL2UUID[URL] = mHeaders["ETag"]; // update uuid
   mapURL2DPLCache[URL] = ctx.outputs().adoptContainer(output, std::move(outputBuffer), true);  
}
```

**Comment by ktf** [comment]
We could have the oldCacheID API you suggest, but wouldn't it make sense to avoid always pruning old messages? After all, there are cases in which we might be jumping back and forward between two.

We probably  also need something to tell the consumers they can now get rid of the deserialised object associated to a given cacheId.

**Comment by shahor02** [comment]
But then, if you have multiple objects cached for a given url, how do you know which one to pass as ETAG in the CcdbAPI query?

**Comment by ktf** [comment]
if-none-match should take multiple etags, iirc. now being boosted. i will
write more in a bit.

Ciao,
Giulio

**Comment by ktf** [comment]
So indeed the standard  seems to allow for multiple etags to be specified:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match

that would allow us to have multiple objects in fly by changing mapURL2UUID to a multi_map and formatting the equal_range as a list of hashes.

However I tried by hand and unless I am doing something stupid it seems that CCDB always replies with 200, rather than a 304 and setting the ETag field. I.e:

```bash
curl -I -H 'If-None-Match: "04dc0fef-554e-11ec-8438-808d13fc250c"' http://ccdb-test.cern.ch:8080/TOF/Calib/Diagnostic/10
HTTP/1.1 304
Access-Control-Allow-Origin: *
Access-Control-Allow-Headers: range
Access-Control-Expose-Headers: content-range,content-length,accept-ranges
Access-Control-Allow-Methods: HEAD,GET
Cache-Control: no-cache
Date: Wed, 12 Jan 2022 15:57:34 GMT
```

however:

```bash
curl -I -H 'If-None-Match: "04dc0fef-554e-11ec-8438-808d13fc250c", "04dc0fef-554e-11ec-8438-808d13fc250d"' http://ccdb-test.cern.ch:8080/TOF/Calib/Diagnostic/10
HTTP/1.1 200
Access-Control-Allow-Origin: *
Access-Control-Allow-Headers: range
Access-Control-Expose-Headers: content-range,content-length,accept-ranges
Access-Control-Allow-Methods: HEAD,GET
Cache-Control: no-cache
Content-Location: /download/04dc0fef-554e-11ec-8438-808d13fc250c
Date: Wed, 12 Jan 2022 15:56:59 GMT
Valid-Until: 999999
Valid-From: 0
Created: 1638655411898
ETag: "04dc0fef-554e-11ec-8438-808d13fc250c"
Last-Modified: Sat, 04 Dec 2021 22:03:31 GMT
ObjectType: o2::tof::Diagnostic
partName: send
Content-Disposition: inline;filename="o2::tof::Diagnostic_1638655411863.root"
Accept-Ranges: bytes
Content-MD5: 1b946af7ee74e6e2e52f83dfaff5ad3e
Content-Type: application/octet-stream
Content-Length: 2048
```

@costing Is that on your side? Could that be fixed? Alternatively one could simply check that one of the provided etags matches the etag in the reply headers (without then fetching the rest of the body).

costing:
I haven't implemented multiple etag values parsing. But how would that work out? The server would return a HTTP 304 "Not modified" response, how would you know which of the two objects to use?

shahor02:
In this approach, the query to the server would need return, probably in the headers, which one of ETAGs (if any) matched to the object valid for the queried time. 
But, @ktf, within the approach of the central fetcher, with queries triggered by the incoming TF, what  would be the use case of multi-object cache? Isn't the query done only once per TF, in the order of arriving TF timestamps?

ktf:
@costing indeed, I would simply add the matching ETag to the 304 response (I  actually think you are supposed to do it in any case, as I understand the standard).

@shahor02 the usecase I had in mind is:

```
GET /foo/bar/1 -> 200 + Etag: abcdef
GET /foo/bar/2 Etag: abcdef -> 304
GET /foo/bar/4 Etag: abcdef -> 200 + Etag: fedcba
GET /foo/bar/3 Etag: fedcba -> 200 + Etag: abcdef
GET /foo/bar/5 Etag: abcdef -> 200 + Etag: fedcba
```

which would become:
```
GET /foo/bar/1 -> 200 + Etag: abcdef
GET /foo/bar/2 Etag: abcdef -> 304 + Etag: abcdef 
GET /foo/bar/4 Etag: abcdef -> 200 + Etag: fedcba
GET /foo/bar/3 Etag: abcdef, fedcba -> 304 + Etag: abcdef
GET /foo/bar/5 Etag: abcdef, fedcba -> 304 + Etag: fedcba
```

i.e. it would save to full fetches compared to now.

costing:
```
curl -I -H 'If-None-Match: "04dc0fef-554e-11ec-8438-808d13fc250c", "04dc0fef-554e-11ec-8438-808d13fc250d"' http://ccdb-test.cern.ch:8080/TOF/Calib/Diagnostic/10
```

This call should work now as you describe above. Please give it a try and if everything is ok I'll propagate it to the production instance too.

ktf:
To me it looks fine, thank you for the prompt development. Let's see what @shahor02 says, though, before we merge it to production.

shahor02:
@ktf, I understood the way you want to query with multiple etags, my question was in which situation we will have queries jumping back and forth in time, e.g.
```
GET /foo/bar/4 Etag: abcdef -> 200 + Etag: fedcba
GET /foo/bar/3 Etag: fedcba -> 200 + Etag: abcdef
GET /foo/bar/5 Etag: abcdef -> 200 + Etag: fedcba
```
I assume the fetcher will sit somewhere at the beginning of the workflow and will see incrementing TFs.
In any case, we should have a way to limit the number of blobs cached for the same URL, perhaps by using e.g.
```
std::unordered_map<std::string, std::deque<std::string>> mapURL2UUID;
```
and suppressing the `front` when updating the cache if the size exceeds some thresholds.

ktf:
Is that (processing timeframes in order) always the case also for merged files or multiple files? In analysis we read timeframes from 2 files in parallel for example, that would trigger the worst case scenario, no?

shahor02:
For the multi-TF files (CTF, rawTF) the TFs are monotonously increasing, with multiple files, one can surely have jumps in any direction. Anyway, if it does not cost too much, it is good to have multi-blob cache, provided we have a mechanism to constrain its size.

ktf:
Yes, agreed. I am implementing first the caching as you suggest and then we can evolve it from there. I agree it needs to have a mechanism to handle the cache size.

PR #7933, opened 2022-01-11
Adapt CCDBApi to work with MemoryResources

ktf:


ktf:
WIP do not merge.

shahor02:
Looks fine, but if you are not going to use the CCDBManagerInstance, then better either leave it unchanged (if somebody will want to get the BLOB in my old definition) or to revert it completely to pre-https://github.com/AliceO2Group/AliceO2/pull/7900 state. In your version you leave the cache wrapper but destroy the cached blob, so it will crash at the next query.

ktf:
This seems to work with my prototype CCDB fetcher, which indeed does not need a CCDB manager at this point, so I kept the changes there to the bare minimum. Notice I had to introduce one copy to avoid duplicating the loadFileToMemory to be used with `o2::vector`.

PR #7934, opened 2022-01-11
[EMCAL-750] Add file per C-RORC, fix filenames

mfasDa:
- Add option to encode each C-RORC card to a separate file
- Encode source ID in the filename based on the granularity
  selected.


mfasDa:
@davidrohr We use the source ID as C-RORC ID. In the file name you need the tag "cru" and not "crorc" (since we are a C-RORC detector).

PR #7935, opened 2022-01-12
Add option to consider single half barrel for PV seeding

mconcas:
@iouribelikov: this should do what we discussed yesterday at WP2. I tested it on some monte carlo data, seems to be working.
I cannot carry on studies on the pilot beam data at the moment, hope it can help you anyways.

To enable the switch:

```bash
o2-its-reco-workflow --trackerCA --configKeyValues="ITSVertexerParam.ignoreHalfBarrel=0"
```
One can ignore the halfbarrel IDs 0 and 1 provided by https://github.com/AliceO2Group/AliceO2/blob/5c0bb3432b94c54877509b6ac6ce7f143647aa50/Detectors/ITSMFT/ITS/base/src/GeometryTGeo.cxx#L145
Default value is 2 that does the usual thing.
If `LOG(debug)` is too strong I can also remove the message entirely.

PS. this change affects the PV finder only. Tracker uses another data loader. Soon the two approaches are going to be unfied.

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/src/IOUtils.cxx
+++ b/Detectors/ITSMFT/ITS/tracking/src/IOUtils.cxx
@@ -95,7 +95,7 @@ void ioutils::convertCompactClusters(gsl::span<const itsmft::CompClusterExt> clu
 
 void ioutils::loadEventData(ROframe& event, gsl::span<const itsmft::CompClusterExt> clusters,
                             gsl::span<const unsigned char>::iterator& pattIt, const itsmft::TopologyDictionary& dict,
-                            const dataformats::MCTruthContainer<MCCompLabel>* clsLabels)
+                            const dataformats::MCTruthContainer<MCCompLabel>* clsLabels, HalfBarrel ignBarrel)
 {
   if (clusters.empty()) {
     std::cerr << "Missing clusters." << std::endl;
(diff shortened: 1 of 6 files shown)


Detectors/ITSMFT/ITS/tracking/src/IOUtils.cxx
```diff
@@ -107,8 +107,12 @@ void ioutils::loadEventData(ROframe& event, gsl::span<const itsmft::CompClusterE
   int clusterId{0};
 
   for (auto& c : clusters) {
-    int layer = geom->getLayer(c.getSensorID());
+    if (geom->getHalfBarrel(c.getSensorID()) == ignBarrel) { // Top barrel Id: 0, Bot: 1, never ignored otherwise
```
shahor02, line 110: `getHalfBarrel` does some looping and calculation, better first check if the selection is requested at all:
```
if (ignBarrel<2 && geom->getHalfBarrel(c.getSensorID())) {
```

Detectors/ITSMFT/ITS/tracking/src/IOUtils.cxx
```diff
@@ -157,8 +161,12 @@ int ioutils::loadROFrameData(const o2::itsmft::ROFRecord& rof, ROframe& event, g
   auto first = rof.getFirstEntry();
   auto clusters_in_frame = rof.getROFData(clusters);
   for (auto& c : clusters_in_frame) {
-    int layer = geom->getLayer(c.getSensorID());
+    if (geom->getHalfBarrel(c.getSensorID()) == ignBarrel) { // Top barrel Id: 0, Bot: 1, never ignored otherwise
```
shahor02, line 164: same here

shahor02 (changes requested):

mconcas:
Closing for the moment as these probably will be superseded by a more generic approach to disable clusters from specific chips.

PR #7939, opened 2022-01-12
Threshold scan workflow

ezradlesser:
Implemented threshold scan workflow for ITS:
`o2-its-threshold-cali-workflow`


ezradlesser:
Need to remove old files -- will open new pull request

PR #7940, opened 2022-01-12
ITS threshold calibration workflow

ezradlesser:
Created new workflow for ITS threshold calibration:
`o2-its-threshold-calib-workflow`

It can be run (for example) in the following way (VCASN scan):
```
o2-raw-tf-reader-workflow --input-data alien:///alice/data/2021/DEC/506615/raw/2140/o2_rawtf_run00506615_tf00000001_epn033.tf --remote-regex "^alien:///alice/data/.+" --copy-cmd "alien.py cp ?src file:?dst" --max-tf 5000 --onlyDet ITS | \
o2-itsmft-stf-decoder-workflow --no-clusters --no-cluster-patterns --enable-calib-data --digits | \
o2-its-threshold-calib-workflow -b --fittype derivative  --output-dir "./" --meta-output-dir "./" | \
o2-calibration-ccdb-populator-workflow --ccdb-path="http://ccdb-test.cern.ch:8080" -b
```


Detectors/ITSMFT/ITS/workflow/CMakeLists.txt
```diff
@@ -21,14 +21,19 @@ o2_add_library(ITSWorkflow
                        src/IRFrameReaderSpec.cxx
                        src/TrackReaderSpec.cxx
                        src/VertexReaderSpec.cxx
+                       src/CalibrationWorkflow.cxx
+                       src/CalibratorSpec.cxx
```
shahor02: The names are too generic, can this be changed to e.g. ThresholdCalibration...?

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
```diff
@@ -0,0 +1,206 @@
...
+  std::map<short int, std::vector<threshold_obj>> thresholds;
+};
+
+template <class Mapping>
+class ITSCalibrator : public Task
+{
+ public:
+  // using Mapping=ChipMappingITS;
+  ITSCalibrator();
+  ~ITSCalibrator(); // override = default;
```
shahor02: `override = default;` should not be commented.
The naming conventions are mostly not respected: data members in camel-case starting with `m`, constants w/o `m` and capitalized: https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
```diff
@@ -0,0 +1,206 @@
...
+  // The number of injections per data value
+  const short int nInj = 50;
+
+  // Number of charges in a threshold scan (from 0 - 50 inclusive)
+  const short int nCharge = 51;
+  // Number of points in a VCASN tuning (from 30 - 70 inclusive)
+  const short int nVCASN = 51;
+  // Number of in a ITHR tuning (from 30 to 100 inclusive)
+  const short int nITHR = 71;
+  // Refernce to one of the above values; updated during runtime
```
shahor02: are these really constants or parameters of the run? According to comments, it is the second.

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
```diff
@@ -0,0 +1,206 @@
...
+  // Hash tables to store the hit and threshold information per pixel
+  std::map<short int, short int> currentRow;
+  std::map<short int, std::vector<std::vector<short int>>> pixelHits;
+  // std::map< short int, TH2F* > thresholds;
+  //  Unordered vector for saving info to the output
+  std::map<short int, std::vector<threshold_obj>> thresholds;
+  // std::unordered_map<unsigned int, int> mHitPixelID_Hash[7][48][2][14][14]; //layer, stave, substave, hic, chip
+
+  // Tree to save threshold info in full threshold scan case
+  TTree* threshold_tree = new TTree("ITS_calib_tree", "ITS_calib_tree");
```
shahor02: Please move this to the constructor or init. Also, avoid memory-based trees, i.e. the Tree must be created after opening a TFile in recreate or update mode.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
```diff
@@ -0,0 +1,1083 @@
...
+    this->threshold_tree->Branch("noise", &(this->noise), "noise/b");
+  this->threshold_tree->Branch("success", &(this->success), "success/O");
+
+  // Get metafile directory from input
+  try {
+    this->metafile_dir = ic.options().get<std::string>("meta-output-dir");
+  } catch (std::exception const& e) {
+    LOG(warning) << "Unable to access meta-output-dir, reason:\n"
+                 << e.what()
+                 << "\n*** Setting metafile output directory to /dev/null";
```
shahor02: I don't understand these lines, here you are not trying to access the directory, contrary to what the message says. I think this try/catch  should envelop the next line,

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
```diff
@@ -0,0 +1,1083 @@
...
+  // Get ROOT output directory from input
+  try {
+    this->output_dir = o2::utils::Str::rectifyDirectory(
+      ic.options().get<std::string>("output-dir"));
+  } catch (std::exception const& e) {
+    LOG(warning) << "Unable to access output-dir, reason:\n"
+                 << e.what()
+                 << "\n*** Setting ROOT output directory to ./";
+  }
+  this->output_dir = "./";
```
shahor02: this line makes meaningless the lines before.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
```diff
@@ -0,0 +1,1083 @@
...
+
+  return;
+}
+
+//////////////////////////////////////////////////////////////////////////////
+// Initialize arrays of chip columns/rows for ROOT histograms
+template <class Mapping>
+void ITSCalibrator<Mapping>::get_row_col_arr(const short int& chipID,
```
shahor02: I don't see where this method is used, but is dynamic allocation really needed? Usually we avoid bare `new` calls.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
```diff
@@ -0,0 +1,1083 @@
...
+  if (flip) { // ITHR case. Lower is at large x[i], Upper is at small x[i]
+
+    for (int i = 0; i < NPoints; i++) {
+      if (data[i] == 0) {
+        Upper = i;
+        break;
+      }
+    }
+
+    if (Upper == -1)
```
shahor02: This will not pass CI: the code checker requires to have {} even for 1-liners, i.e.
```
if (Upper == -1) {
  return false;
} 
```

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
```diff
@@ -0,0 +1,1083 @@
...
+  fitfcn->Write(0, TObject::kOverwrite);
+
+  // Close file and clean up memory
+  tf->Close();
+  delete tf;
+  exit(1);
+  */
+
+  delete g;
+  delete fitfcn;
```
shahor02: given the number of channels, why do you create and delete the function for every fit, instead of reusing the same function?

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
```diff
@@ -0,0 +1,206 @@
...
+} PIXEL;
+
+class threshold_map
+{
+ public:
+  threshold_map(const std::map<short int, std::vector<threshold_obj>>& t) : thresholds(t){};
+  std::map<short int, std::vector<threshold_obj>> thresholds;
+};
+
+template <class Mapping>
```
shahor02: Why do you need the Mapping as a template parameter? Are you planning to use anything else than ChipMappingITS?

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
```diff
@@ -0,0 +1,1083 @@
...
+    // Check if file already exists
+    struct stat buffer;
+    if (recreate && stat(filename.c_str(), &buffer) == 0) {
+      LOG(warning) << "File " << filename << " already exists, recreating";
+    }
+
+    // Initialize ROOT output file
+    // to prevent premature external usage, use temporary name
+    const char* option = recreate ? "RECREATE" : "UPDATE";
+    TFile* tf = TFile::Open(filename.c_str(), option);
```
shahor02: as I wrote above: please open the output file before the creation of the TTree

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
```diff
@@ -0,0 +1,1083 @@
...
+// Set the run_type for this run
+// Initialize the memory needed for this specific type of run
+template <class Mapping>
+void ITSCalibrator<Mapping>::set_run_type(const short int& runtype)
+{
+
+  // Save run type info for future evaluation
+  this->run_type = runtype;
+
+  if (runtype == 42) {
```
shahor02: Perhaps it is better to use some enums instead of hardcoded identifiers?

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
```diff
@@ -0,0 +1,1083 @@
...
+    this->update_LHC_period(pc);
+  }
+
+  // Calibration vector
+  const auto calibs = pc.inputs().get<gsl::span<o2::itsmft::GBTCalibData>>("calib");
+  const auto digits = pc.inputs().get<gsl::span<o2::itsmft::Digit>>("digits");
+  const auto ROFs = pc.inputs().get<gsl::span<o2::itsmft::ROFRecord>>("digitsROF");
+  const auto tfcounter = o2::header::get<o2::framework::DataProcessingHeader*>(
+                           pc.inputs().get("digitsROF").header)
+                           ->startTime;
```
shahor02: Use 
```
const auto tfcounter = DataRefUtils::getHeader<DataProcessingHeader*>(pc.inputs().getFirstValid(true))->startTime;
```
instead of this line. Though, I don't see where do you use this counter.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
```diff
@@ -0,0 +1,1083 @@
...
+  // Below is CCDB stuff
+  long tstart, tend;
+  tstart = o2::ccdb::getCurrentTimestamp();
+  constexpr long SECONDSPERYEAR = 365 * 24 * 60 * 60;
+  tend = o2::ccdb::getFutureTimestamp(SECONDSPERYEAR);
+
+  auto class_name = o2::utils::MemFileHelper::getClassName(tuning);
+
+  // Create metadata for database object
+  std::string* ft;
```
shahor02: why pointer rather than object?

shahor02 (dismissed):
Conditionally approving to trigger the CI. Don't merge yet.

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
ezradlesser: These values are constants. Which one is actually used is determined by the type of run, but these values themselves are global constants. We will modify their names to give the naming convention you point out above.

Detectors/ITSMFT/ITS/workflow/CMakeLists.txt
ezradlesser: Yes, we can change the names of these files.

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
ezradlesser: The code does not compile using `override = default;` -- note that we are defining our own deconstructor in order to clean up dynamic memory. 

Thank you for the pointer, we will rename everything according to this sheet.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
ezradlesser: Thanks, we will rephrase this log warning to be more clear and fix the following line.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
ezradlesser: Thanks -- we will add an enum in the header file to define all of these numbers.

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
shahor02: Right, then you would need the `override` (or `final`), otherwise the code checker may complain.

ezradlesser: Feel free to suggest an edited line and I will try it here -- but again, the code does not compile with `override = default` and I don't recall any complaints from the code checker (previous to the last push).

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
ezradlesser: Thanks -- this will be moved elsewhere in the next commit (and initialized with nullptr). We will improve the output so that a ROOT file is opened before each new TTree.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
ezradlesser: Thanks -- we will add this.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
ezradlesser: A pointer is better in this case so that the object is not initialized twice -- though in theory either would work. I don't have a strong preference one way or the other.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
ezradlesser: Thanks, we have switched to this line and commented it out. We were previously using this for print statements while debugging, so we will keep it (but commented out) for convenience while things are still in development.

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
shahor02: The previous comment was about `override` only, w/o `default`, which is surely inappropriate if you define your method. Iiirc, it is the fullCI codechecker which complains about overridden virtual methods w/o override flag, did this code ever passed the fullCI?

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
shahor02: `std::string s;` declaration does not involve any heap memory allocation (in opposite to `new std::string()` which does 2 heap allocations: one to create the string object and another one to set the string value if it is longer than the fixed size internal buffer, but this is STL implementation dependent).

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
ezradlesser: Okay, I will try with just `override` -- thanks!
What is fullCI? I can only see the checkers at the bottom of this PR. 
(edit) ah sorry I see it now -- I don't recall this one in particular.

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
ezradlesser: `std::string s` calls the default constructor for `std::string` and creates a new (empty) `std::string` which is then stored on the stack. `s = std::string("stuff")` calls a copy constructor for `std::string` and creates a second `std::string` object, which replaces the original (empty) `std::string`. The deconstructor should then called for the original (empty) `std::string`.

To my understanding whether or not the object is stored on the heap or stack is not so important, but using a pointer the object is at least only created once.

ezradlesser: If really one is worried about efficiency here the best is to use `char*` but I think really this will have no perceivable effect at all on the code...

shahor02: this is just in our coding guidelines: avoid bare `new` operators if possible.

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
shahor02: it is there, called `build/O2/fullCI `

Detectors/ITSMFT/ITS/workflow/src/CalibratorSpec.cxx
ezradlesser: OK, understood -- thanks. I will change to char array.

shahor02: ok, it becomes somewhat academic, but... what is expensive the heap memory allocation, creation of the stack object is very cheap. The 'std::string x;' creates just a fixed size stack object, and `x = "something"`; will call assignment operator with just 1 memory allocation.
With the `new std::string("something")`, you have 2 allocations.
That said, both cases will be likely optimized out to predefined const strings with needed one picked depending on the if condition. What matters is that the code with local objects is safer than that will explicit new / deletes.

shahor02:
@ezradlesser could you apply clang-format, the CIs will not start while the formatting has problems.

ezradlesser:
> @ezradlesser could you apply clang-format, the CIs will not start while the formatting has problems.

Yes of course, sorry, I am still updating things -- I had to build O2 from scratch again after having pulled from upstream, so I am doing testing now and will let you know when ready for another review.

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/CalibratorSpec.h
ezradlesser: Thanks for pointing this out, at the moment we foresee no use so the mapping template has now been removed.

ezradlesser:
Hi @shahor02 -- I believe that everything has been updated now.

shahor02 (dismissed):
approving to trigger CIs

ezradlesser:
Hi @shahor02 -- I just pushed again to fix a compilation issue in the macOS build checker.

shahor02 (dismissed):

ezradlesser:
Hi again @shahor02 -- after reviewing the build/O2/fullCI I believe that I have now corrected the few remaining points, and so I have pushed an update once more. (There was also some compilation warning that does not appear to be correlated with our changes.) Thank you!

shahor02 (dismissed):

ezradlesser:
I am a bit confused by the current copyright failure of `build/O2/fullCI` -- I checked the existing copyright statement at the top of the two mentioned files here on GitHub, and made a diff to the suggested text in the `fullLog.txt`, and the two texts appear to be identical. Is there something else which could trigger this error?

shahor02:
indeed, I don't see anything wrong in the license header, please force-push some dummy commit to restart the CI,

ezradlesser:
Looks like those two files were in DOS format (using `\r` instead of `\n`) for some reason -- this must have caused the error. It should be resolved now.

shahor02 (approved):

shahor02:
@ezradlesser all checks are passed, shall I squash/merge or you have to add something?

ezradlesser:
> @ezradlesser all checks are passed, shall I squash/merge or you have to add something?

Yes, please merge, thank you!

PR #7941, opened 2022-01-13
Ccdb fetcher

ktf:
Ongoing:

- [x] Understand why my changes to the CCDBApi seem to crash some other tests.
- [x] Add the fetching of the CTP and associated calculation of the timestamp based on OrbitResetTime and firstTFOrbit
- [x] Do not deserialise the objects every timeframe
- [x] Add a callback when it actually needs to happen.
- [x] Use DISTSUBTIMEFRAME / TF/TFNUM where appropriate to get the time for the data
- [x] Add ConfigParamSpec for createdNotAfter and createdNotBefore
- [x] Support for metadata on query in the InputSpec
- [x] Allow configuring the CCDB path / backend on a per input basis.
- [ ] Support for headers as result message.
- [ ] Better strategy to cleanup the cache and eventually support prefetching of CCDB objects / multiple in-fly objects for the same path.

### DIFF
--- a/Framework/Core/include/Framework/Task.h
+++ b/Framework/Core/include/Framework/Task.h
@@ -38,6 +38,24 @@ class has_endOfStream
   enum { value = sizeof(test<T>(nullptr)) == sizeof(char) };
 };
 
+/// Check if the class task has EndOfStream
+template <typename T>
+class has_finaliseCCDB
+{
+  typedef char one;
+  struct two {
+    char x[2];
+  };
+
+  template <typename C>
+  static one test(decltype(&C::finaliseCCDB));
+  template <typename C>
+  static two test(...);
+
+ public:
+  enum { value = sizeof(test<T>(nullptr)) == sizeof(char) };
+};
+
 /// Check if the class task has Stop
 template <typename T>
 class has_stop
@@ -79,6 +97,13 @@ class Task
   /// This is invoked whenever we have an EndOfStream event
   virtual void endOfStream(EndOfStreamContext& context) {}
 
+  /// This is invoked whenever a new CCDB object associated to
+  /// a given ConcreteDataMatcher is deserialised
+  virtual void finaliseCCDB(ConcreteDataMatcher& matcher, void* obj)
+  {
+    LOGP(error, "CCDB deserialization invoked");
+  }
+
   /// This is invoked on stop
   virtual void stop() {}
 };
@@ -94,6 +119,12 @@ AlgorithmSpec adaptFromTask(Args&&... args)
       auto& callbacks = ic.services().get<CallbackService>();
       callbacks.set(CallbackService::Id::EndOfStream, [task](EndOfStreamContext& eosContext) {
         task->endOfStream(eosContext);
+      });
+    }
+    if constexpr (has_finaliseCCDB<T>::value) {
+      auto& callbacks = ic.services().get<CallbackService>();
+      callbacks.set(CallbackService::Id::CCDBDeserialised, [task](ConcreteDataMatcher& matcher, void* obj) {
+        task->finaliseCCDB(matcher, obj);
       });
     }
     if constexpr (has_stop<T>::value) {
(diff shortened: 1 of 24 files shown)


ktf:
First iteration, do not merge.

Framework/Core/src/CCDBHelpers.cxx
```diff
@@ -12,24 +12,110 @@
...
+              }
+            }
+            const auto url2uuid = helper->mapURL2UUID.find(path);
+            if (url2uuid != helper->mapURL2UUID.end()) {
+              etag = url2uuid->second;
+            }
+
+            LOGP(info, "{} {} {}\n", timingInfo.timeslice, timingInfo.tfCounter, timingInfo.creation);
+            helper->api.loadFileToMemory(v, path, metadata, timingInfo.timeslice, &headers, etag, createdNotAfter, createdNotBefore);
+            if ((headers.count("Error") != 0) || !v.size()) {
```
shahor02: I think `v.size()==0` should be treated as an error only if the ETAG was empty (cache was not filled yet), with ETAG for the valid object the `v` must be empty.

ktf: agreed.

ktf:
With the latest commit commit the example in `Framework/TestWorkflows/src/test_CCDBFetcher.cxx` works as before when running with:

```
stage/bin/o2-testworkflows-test-ccdb-fetcher --start-value-enumeration `date +%s` --condition-backend http://alice-ccdb.cern.ch:8080
```

but using the new centralised logic and honouring the etag.

Framework/Core/src/CCDBHelpers.cxx
ktf: done.

shahor02:
>> Understand why my changes to the CCDBApi seem to crash some other tests.

I think this is because of exposing boost headers to root cling. Perhaps we can avoid this by templating the vector to fill?

ktf:
I was thinking about that, but then we would need to expose everyone a  bunch of ROOT headers. I am now trying something  slightly different with `ifndef __CLING__`.

ktf:
New version fetches the OrbitResetTime from the CTP (plus a bit of cleanup).

ktf:
Update now does not deserialise CCDB objects if the associated blob was not changed.

ktf:
Now with the callback for deserialised objects.

Framework/Core/include/Framework/Task.h
```diff
@@ -38,6 +38,24 @@ class has_endOfStream
   enum { value = sizeof(test<T>(nullptr)) == sizeof(char) };
 };
 
+/// Check if the class task has EndOfStream
+template <typename T>
+class has_finalseCCDB
```
shahor02: typo in "finaliseCCDB"?

ktf: yes

ktf:
Fixed some tests and update to include DISTSUBTIMEFRAME / TF/TFNUM subscription, if present in the topology.

ktf:
Metadata support added.

ktf:
I now have the remapping in place. There is probably something still to adapt for the multihost case, but it should work for simple, same server remappings. The syntax is:

```
--condition-remap <prefix1>=<target1>,[<target2>;<prefix2>=<target3>[,...]]
```

where prefix can be either `https://` / `http://` or `/`. The targets are the paths which are specified by the `ccdbParamSpec()`. For the time being the must fully match.

If this works I would merge this PR and address the remaining issues in separate developments.

shahor02:
> For the time being they must fully match.

If you refer to URLs with the wildcard, I would not do that as this could be error prone.

ktf:
@davidrohr @shahor02 I think this could be merged as is. It is a step forward the current implementation and if the Lifetime::Condition is not used, there should not be any side effects at all. What do you think? The FLP suite will use in any case the tag of yesterday, no?

shahor02:
Hi @ktf 
thanks, fine with me to merge this. I'll try to play with it after finishing another branch I am working with.

shahor02:
@ktf were not you going to merge this PR? Is there any reason to not do this?

davidrohr:
I discussed with @ktf to check this in FST with DD before merging, this test is currently running. Will comment here soon.

davidrohr (approved):
Don't see any problem in FST, fine for me to merge this.

ktf:
Ok, thanks. As discussed, let's take it from here for the remaining bugs / features.

PR #7945, opened 2022-01-13
TPC: Fix krypton workflow

wiechula:
* Fix completion policy
* Fix setting of branch address
* Implement sector selection


wiechula:
@davidrohr @shahor02 can this be merged?

PR #7946, opened 2022-01-13
prototype schema evolution for V0s

jgrosseo:


jgrosseo:
Needs https://github.com/AliceO2Group/O2Physics/pull/445

jgrosseo:
@shahor02 @ktf @aalkin Here a prototype of the schema evolution. I applied it to V0s now as this is more pressing than for MCs. This still needs some testing but outlines the idea.
With @ktf I discussed that instead of putting the version string explicitly into the table name, we can use the subspec possibly (and make that an SOA macro argument).
Also we run into issues with long table names with the explicit listing...

ktf:
I gave it a bit of thought and indeed I think using (part of?) the subSpec might be what we want. We would probably need to add a wildcard for  the subspec in a few InputSpec we generate from the table descriptions, but apart from that it should be fine. It would also allow to have multiple version read at the same time and e.g. do comparison between them.

jgrosseo:
Synced force merge with O2Physics

PR #7948, opened 2022-01-14
[MCH MID] event finder algorithm and workflow

pillot:
Uses MID as a trigger, with a parametrizable window, to gather the MCH digits potentially produced by the same interaction.

### DIFF
--- a/Detectors/MUON/MCH/Triggering/src/EventFinder.cxx
+++ b/Detectors/MUON/MCH/Triggering/src/EventFinder.cxx
@@ -1,0 +1,120 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file EventFinder.cxx
+/// \brief Implementation of a class to group MCH digits based on MID information
+///
+/// \author Philippe Pillot, Subatech
+
+#include "MCHTriggering/EventFinder.h"
+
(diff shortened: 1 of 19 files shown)


Detectors/MUON/MCH/Workflow/CMakeLists.txt
```diff
@@ -216,15 +222,15 @@ o2_add_executable(
             src/ClusterTransformerSpec.cxx
             src/DigitFilteringSpec.cxx
             src/DigitReaderSpec.cxx
-            src/TrackAtVertexSpec.cxx
+            src/EventFinderSpec.cxx
```
aphecetche, line 225: why is the time clustering device dependent on the event finder ?

Detectors/MUON/MCH/Triggering/src/EventFinder.cxx
```diff
@@ -0,0 +1,120 @@
...
+#include "MCHTriggering/EventFinderParam.h"
+
+namespace o2
+{
+namespace mch
+{
+
+//_________________________________________________________________________________________________
+/// run the event finder algorithm
+void EventFinder::run(const gsl::span<const mch::ROFRecord>& mchROFs,
```
aphecetche, line 33: (note: this is nick picking on an otherwise very fine and well prepared PR ;-) ).
I would have split this run method in its actual 4 working parts to ease reading. 
(plus I suspect the digit merging part we'll have to externalise at some point to reuse it elsewhere...)
Again, not a call for action for this PR, just a comment.

Detectors/MUON/MCH/Workflow/CMakeLists.txt
pillot: sorry, my previous message didn't show up... it's not the time clustering, it's the reco-workflow ;-)

aphecetche: indeed. got confused by the way GitHub diffs are presented. sorry about that.

aphecetche:
@TimoWilken the Mac builds seem to have some issues...

pillot:
@shahor02 the failing build on macOS does not seem related to this PR. Could you merge it please ?

PR #7955, opened 2022-01-16
Fix cluster->pixel expansion in ITS noise calibration

shahor02:


shahor02:
Fixes https://alice.its.cern.ch/jira/browse/O2-2745

PR #7957, opened 2022-01-17
fix the way ambigous tracks are written in AODs for muons

mguilbau:
This is an attempt to fix the way ambiguous muon tracks are written in the AOD. In the current version, each ambiguous track is written for all compatible ROFs. This create incompatibility in track table and MC label length at a later stage.

In this fix, the loop over ROF is removed as it should not be necessary. Compatible ROFs for a given track can be find back using track time and track time error at later stages.

I have left the vertex to 0 0 0 for the moment. This could be change with a better value later (average BS x, y, z position?)

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -290,17 +290,10 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
           // FwdTracks tracks are treated separately since they are stored in a different table
           const auto& track = data.getMCHTrack(trackIndex);
           if (collisionID < 0) {
-            InteractionRecord meanIR;
-            auto rofsMCH = data.getMCHTracksROFRecords();
-            for (const auto& rof : rofsMCH) {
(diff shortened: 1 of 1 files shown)


mguilbau:
@aphecetche you may want to look at this

mguilbau:
@ktf @shahor02 This PR did not pass the fullCI test due to copyright fail on a file that is not related to this PR (/sw/SOURCES/O2/7957-slc8_x86-64/0//Utilities/ShmManager/src/ShmManager.cxx). Could you take a look please?

Thanks a lot

shahor02:
@mguilbau I've fixed the copyright and merging this. Note that the way the ambiguous tracks are stored is still not a proper one, I am now implementing it for barrel tracks, then it should be extended to the forward tracks.

mguilbau:
Hi Ruben, 

Thanks a lot. Yeah I know. This change is meant to avoid the Analysis workflow to breakdown because of table length inconsistency between data and MC.
We will extend to Fwd tracks when you are done with the barrel

Cheers

— 
Maxime Guilbaud
Assistant Professor / Enseignant-Chercheur
SUBATECH - IMT Atlantique
Tél :  +33 (0)2 51 85 80 23
e-mail : ***@***.***
—

PR #7958, opened 2022-01-17
Move cut on min N TPC Clusters from 40 to 60

chiarazampolli:
Following async-QC meeting.
@shahor02 , @stheckel


shahor02 (approved):
trivial, merging

PR #7960, opened 2022-01-17
Add ShmManager utility

rbx:
- Starting code for the shm utility needed for R3C-646.
@davidrohr This can be already using in this form. Example config:
```
./shm-manager --shmid 1 --segments 1,1000 2,3000 --regions 1,2000 2,5000
[INFO] Created segment 1 of size 1000, starting at 0x7f3277c8e000. Locking...
[INFO] Done.
[INFO] Zeroing...
[INFO] Done.
[INFO] Created segment 2 of size 3000, starting at 0x7f3277c8d000. Locking...
[INFO] Done.
[INFO] Zeroing...
[INFO] Done.
[INFO] Created unamanged region 1 of size 2000, starting at 0x7f3277c8c000. Locking...
[INFO] Done.
[INFO] Zeroing...
[INFO] Done.
[INFO] Created unamanged region 2 of size 5000, starting at 0x7f3277c8a000. Locking...
[INFO] Done.
[INFO] Zeroing...
[INFO] Done.
^C[INFO] stopping.
[INFO] Cleaning up for shared memory id '00000001'...
[INFO] Found 2 unmanaged regions...
[INFO] Found RegionInfo with path: '', flags: 0, fDestroyed: 0.
[INFO] Successfully removed 'fmq_00000001_rg_2'.
[INFO] Found RegionInfo with path: '', flags: 0, fDestroyed: 0.
[INFO] Successfully removed 'fmq_00000001_rg_1'.
[INFO] Found 2 managed segments...
[INFO] Successfully removed 'fmq_00000001_m_2'.
[INFO] Successfully removed 'fmq_00000001_m_1'.
[INFO] Successfully removed 'fmq_00000001_mng'.
```

### DIFF
--- a/Utilities/ShmManager/CMakeLists.txt
+++ b/Utilities/ShmManager/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+# Copyright 2019-2022 CERN and copyright holders of ALICE O2.
 # See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
 # All rights not expressly granted are reserved.
 #
@@ -9,11 +9,7 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-add_subdirectory(DataSampling)
-add_subdirectory(DataCompression)
-add_subdirectory(Mergers)
-add_subdirectory(PCG)
-add_subdirectory(rANS)
-add_subdirectory(Tools)
-add_subdirectory(EPNMonitoring)
-add_subdirectory(ShmManager)
+o2_add_executable(shm-manager
+                  COMPONENT_NAME epn
+                  SOURCES src/ShmManager.cxx
+                  PUBLIC_LINK_LIBRARIES FairMQ::FairMQ)
--- a/Utilities/ShmManager/src/ShmManager.cxx
+++ b/Utilities/ShmManager/src/ShmManager.cxx
@@ -1,0 +1,161 @@
+// Copyright 2019-2022 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include <fairmq/shmem/Common.h>
+#include <fairmq/shmem/UnmanagedRegion.h>
+#include <fairmq/shmem/Segment.h>
+#include <fairmq/shmem/Monitor.h>
+
+#include <fairmq/tools/Unique.h>
+
+#include <fairlogger/Logger.h>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/program_options.hpp>
+
+#include <csignal>
+
+#include <chrono>
+#include <map>
+#include <string>
+#include <thread>
+
+using namespace std;
+using namespace boost::program_options;
+
+namespace
+{
+volatile sig_atomic_t gStopping = 0;
+}
+
+void signalHandler(int /* signal */)
+{
+  gStopping = 1;
+}
+
+struct ShmManager {
+  ShmManager(uint64_t _shmId, const vector<string>& _segments, const vector<string>& _regions)
+    : shmId(fair::mq::shmem::makeShmIdStr(_shmId))
+  {
+    AddSegments(_segments);
+    AddRegions(_regions);
+  }
+
+  void AddSegments(const vector<string>& _segments)
+  {
+    for (const auto& s : _segments) {
+      vector<string> conf;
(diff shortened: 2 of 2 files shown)


rbx:
Fixed formatting.

davidrohr:
Thx a lot, from the example it looks nice. Will test with the managed region in the full system test.
@ironMann : When can we expect support in DD for this?

rbx:
To remind you what is needed on DD/DPL side (as I wrote in the [ticket](https://alice.its.cern.ch/jira/browse/R3C-646)):

DD:
```cpp
fair::mq::RegionConfig regionCfg;
regionCfg.id = 1; // the id corresponding to the region IDs set by the tool, to locate the open region instead of creating it
regionCfg.removeOnDestruction = false; // do not remove this region when it destructs.
fRegion = fair::mq::UnmanagedRegionPtr(NewUnmanagedRegionFor("channel", 0, regionSize, yourAckCallback, regionCfg));
```
In principle the code above can be the same even if workflow+DD run without this tool. The ID would then be used to create (not open) the region and `removeOnDestruction` will skip only the destruction in the destructor of `UnmanagedRegion`, but the final `Monitor::Cleanup()` will still clean everything unconditionally.

DPL & DD:
When this tool is active, both have to run with:
- `--shm-monitor false` (should already be the case, at least for DPL) to not launch shmmonitor.
- `--shm-no-cleanup true` to skip the `Monitor::Cleanup()` step when final device in a session leaves. Or whatever the equivalent is on DPL side, to actually skip shmem cleanup.
- `--shmid <id-corresponding to the shmid in the tool>` to actually find everything.

davidrohr:
ok, thx. How does shmid work in detail? By default it is deduced from the sessionid, which normally comes from DDS?
But if I set --shmid explicitly, it overrides what comes from the sessionid?

rbx:
> How does shmid work in detail? By default it is deduced from the sessionid, which normally comes from DDS? But if I set --shmid explicitly, it overrides what comes from the sessionid?

That's right, it overrides it for shm objects only. Devices still need session id for other purposes.

ktf:
Can you also fix the copyright statement?

rbx:
> Can you also fix the copyright statement?

Done.

ktf:
Copyright is actually correct now, it's the test that was sticking to 2020. Fixed.

davidrohr:
Hi @rbx : I have one question for the usage:
at some point you had added a feature to "reinitialize" the shm segment.
We need a way to do that in between.
If we run one workflow, and it fails somehow, it might leave messages in the segment, correct?
So before we start the next workflow, we should get rid of all messages in there.
Is that already foreseen? Perhaps I just missed it?

rbx:
This line does the "soft reset": https://github.com/AliceO2Group/AliceO2/pull/7960/files#diff-4bd4b5fb9352adb5460d9cdff3697bbf396c0e04db4cfefde633f193316f0d48R100.
It is not called anywhere, but you can extend the tool and call that as a reaction to some signal/message/input. You could even make a separate executable and call it from there, as long as all the ids stay the same it should be fine to do so. Or the same executable with a boolean switch - `ShmManager --shmid <id> --soft-reset`, which would only call the soft reset.

I would only call soft reset when you have more or less error-free shutdown - which can still leave buffers in shm if there were orphaned messages in the IPC queues.

But if you get a real crash, all kinds of things could be wrong, including in management parts of shared memory, then I would only do a full reset: https://github.com/AliceO2Group/AliceO2/pull/7960/files#diff-4bd4b5fb9352adb5460d9cdff3697bbf396c0e04db4cfefde633f193316f0d48R103-R108. Full reset cannot be in a separate executable in this case, then "the state" in this tool is no longer correct. For full reset you'll need either some signal, or just restart the tool with the same parameters.

davidrohr:
OK, I tested the tool with a DPL workflow (just a single workflow, so no cleanup needed in between) and it seems to work as expected, very nice. workflow startup is much faster as expected.
I couldn't test with DataDistribution, since it lacks the feature still.
Giulio already merged it, so nothing left to do.

We need one more additional feature:
We must have per-segment-id control over the NUMA policy, to bind to particular NUMA domain or to interlaved memory.
I'll comment everything on the jira ticket.

Thx for the tool. Good to have finally some progress here!

ktf:
This breaks the EPN builds:

https://alijenkins.cern.ch/job/DailyBuilds/job/DailyO2Dataflow/395/console

because the flp-suite branch we are using does not yet pick-up the latest FairMQ. @vascobarroso @awegrzyn do you plan to do so for the next version?

@rbx ~can we replace fairmq/shmem/Common.h by something equivalent which is backward compatible?~ Actually the file seems to be there since ever. Maybe it needs to be installed?

davidrohr:
I'll add FairMQ to the overrides for the EPN builds temporarily, FMQ will be bumped on the FLPs on Monday, and in the meantime we don't need to update software on the EPNs anyway.

awegrzyn:
@ktf Yes, FLP Suite branches are always based on master.

ktf:
@davidrohr ok

rbx:
> @rbx ~can we replace fairmq/shmem/Common.h by something equivalent which is backward compatible?~ Actually the file seems to be there since ever. Maybe it needs to be installed?

This file was not installed before because it is an implementation detail. Although it is still a detail now, some of the shmem headers are now public, to provide shmem-specific APIs needed by this tool, including this header.

PR #7961, opened 2022-01-18
Improving sensitive hit creation for FT0 in Detector.cxx

zovarga:
- The Detector::ProcessHits(FairVolume* v) function had too many string comparisons which made it slow. The comparison was modified to be based on integer IDs instead.
- The calculation of some quantities (etot, enDep, parentID) were moved inside the relevant if() statement block (i.e. when the particle is a photon), so they are only calculated when necessary.

### DIFF
--- a/Detectors/FIT/FT0/simulation/include/FT0Simulation/Detector.h
+++ b/Detectors/FIT/FT0/simulation/include/FT0Simulation/Detector.h
@@ -247,6 +247,11 @@ class Detector : public o2::base::DetImpl<Detector>
 
   /// Container for data points
   std::vector<o2::ft0::HitType>* mHits = nullptr;
+  
+  // Define volume IDs
+  int regVolID = -1; 
+  int topVolID = -1; 
+  int mtoVolID = -1;
 
   /// Define the sensitive volumes of the geometry
   void defineSensitiveVolumes();
--- a/Detectors/FIT/FT0/simulation/include/FT0Simulation/Detector.h
+++ b/Detectors/FIT/FT0/simulation/include/FT0Simulation/Detector.h
@@ -249,9 +249,9 @@ class Detector : public o2::base::DetImpl<Detector>
   std::vector<o2::ft0::HitType>* mHits = nullptr;
 
   // Define volume IDs
-  int regVolID = -1;
-  int topVolID = -1;
-  int mtoVolID = -1;
+  int mREGVolID = -1;
+  int mTOPVolID = -1;
+  int mMTOVolID = -1;
 
   /// Define the sensitive volumes of the geometry
   void defineSensitiveVolumes();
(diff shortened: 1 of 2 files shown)


Detectors/FIT/FT0/simulation/src/Detector.cxx
```diff
@@ -870,7 +870,11 @@ void Detector::defineFrameTransformations()
 Bool_t Detector::ProcessHits(FairVolume* v)
 {
 
-  TString volname = fMC->CurrentVolName();
+  Int_t copy;
+  Int_t volID = fMC->CurrentVolID(copy);
+  Int_t volREG = fMC->VolId("0REG");
+  Int_t volTOP = fMC->VolId("0TOP");
+  Int_t volMTO = fMC->VolId("0MTO");
```
sawenzel: these should be cached as data members inside the Detector class to avoid calling `VolId` again and again.

sawenzel (changes requested):
Looks good. Please still address the comment concerning caching volume ids as data members.

AllaMaevskaya (dismissed):

zovarga:
Thank you for the review, I implemented the requested changes.

Detectors/FIT/FT0/simulation/include/FT0Simulation/Detector.h
```diff
@@ -248,6 +248,11 @@ class Detector : public o2::base::DetImpl<Detector>
   /// Container for data points
   std::vector<o2::ft0::HitType>* mHits = nullptr;
 
+  // Define volume IDs
+  int regVolID = -1;
+  int topVolID = -1;
+  int mtoVolID = -1;
```
sawenzel: please note that class member variables need to follow the `mFooBar` notation.

sawenzel (approved):

sawenzel:
@ktf @TimoWilken : There is a weird alibuild error here:
```
++ alibuild-generate-module --bin --lib
/System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SPECS/osx_arm64/TBB/v2021.5.0-local1/TBB.sh: line 13: alibuild-generate-module: command not found
```
I have seen this locally as well this morning.

TimoWilken:
Looks like TBB should depend on alibuild-recipe-tools!

sawenzel:
@TimoWilken : Right. But why did all the tests in alidist pass? It appears to be a bit platform-dependent at least.

TimoWilken:
Maybe the build order changed, and something else that depends on alibuild-recipe-tools was built before TBB. In that case, even though TBB doesn't explicitly depend on a-r-t, it would have it available. In this case, TBB was built before anything that pulls in a-r-t, so a-r-t wasn't available.

(Edited to add: this may well be platform-dependent, as using packages from the system would change the build graph and therefore the build order.)

PR #7962, opened 2022-01-18
Fixes assessment workflows

rpezzi:
Also 
- Adds 5ms delays on the MFT track and cluster readers to avoid backpressure.
- Release memory when processing MC Kinematics

### DIFF
--- a/Detectors/GlobalTracking/src/MatchGlobalFwdAssessment.cxx
+++ b/Detectors/GlobalTracking/src/MatchGlobalFwdAssessment.cxx
@@ -250,8 +250,9 @@ void GloFwdAssessment::processGeneratedTracks()
         auto evh = mcReader.getMCEventHeader(src, event);
         addMCParticletoHistos(&mcParticle, kGen, evh);
       } // mcTracks
-    }   // events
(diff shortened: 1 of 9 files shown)


shahor02 (commented):
> Adds 5ms delays on the MFT track and cluster readers to avoid backpressure

Perhaps this can be activated with an option and run w/o delay by default? Otherwise, it can add unwanted delays when the reader is used from the globalTrackCluster reader.

Does the backpressure really pose a problem?

rpezzi:
Hi @shahor02 , the delay resolves the non critical issue of flooding the logger with back-pressure messages in some cases. I agree that it is better to have this as a configurable option. I am removing this for now since this PR solves a critical issue. 

The cluster reader is defined on the ITSMFT common library, and MFT track reader on the MFT lib. Do you have any preference on how to have this implement this in a system-wide way?

shahor02:
Hi @rpezzi 
There is no need for the system-wide delay, the backpressure issue will be handled by the rate (of the number of TFs in flight) limiting which is in preparation.

Cheers,
 Ruben

PR #7963, opened 2022-01-18
MFT cluster noise calibration with dictionary patterns

mcoquet642:
Adding possibility to use cluster dictionary for MFT noise calibration.


shahor02 (approved):
This implements fixes discussed before, merging

PR #7968, opened 2022-01-19
Embedding of independent signal and background

pbuehler:
In AODProducerWorkflowSpec.cxx it was assumed, that when signals are embedded into background, then the signals belong to the same collision as a background event. Embedding however also allows for independent signal and background vertices. This change enables this feature in AODProducerWorkflowSpec.cxx.


sawenzel (approved):
makes total sense. Thanks.

PR #7970, opened 2022-01-19
DPL Analysis: improve spawner logic for handling extensions

aalkin:
* Use `InputSpec` metadata feature to track inputs for spawnable table extensions
* Spawner now also supports extended tables over Joins
* Improved exceptions' phrasing
* Consolidated input lists and input metadata management

@ktf @jgrosseo this seems to work already, but I want to reduce code duplication and test the topology update logic, that has changed slightly.

### DIFF
--- a/Framework/Core/include/Framework/InputSpec.h
+++ b/Framework/Core/include/Framework/InputSpec.h
@@ -62,9 +62,6 @@ struct InputSpec {
             enum Lifetime lifetime_ = Lifetime::Timeframe,
             std::vector<ConfigParamSpec> const& metadata_ = {});
 
-  /// Create an InputSpec from metadata string
-  static InputSpec fromString(std::string s);
-
   /// A mnemonic name for the input spec.
   std::string binding;
 
@@ -81,8 +78,5 @@ struct InputSpec {
   friend std::ostream& operator<<(std::ostream& stream, InputSpec const& arg);
   bool operator==(InputSpec const& that) const;
 };
-
-void updateInputList(std::vector<InputSpec>& list, InputSpec&& input);
-
 } // namespace o2
 #endif
--- a/Framework/Core/include/Framework/InputSpec.h
+++ b/Framework/Core/include/Framework/InputSpec.h
@@ -78,5 +78,6 @@ struct InputSpec {
   friend std::ostream& operator<<(std::ostream& stream, InputSpec const& arg);
   bool operator==(InputSpec const& that) const;
 };
+
 } // namespace o2
 #endif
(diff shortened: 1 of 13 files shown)


ktf:
Please do not do changes which are not  strictly necessary, I am changing the same code and I would prefer not having to merge conflicts.

aalkin:
@ktf probably it would be easier to change after this one, it removes quite a lot of repetitions

Framework/Core/src/InputSpec.cxx
```diff
@@ -89,10 +88,43 @@ InputSpec::InputSpec(std::string binding_,
 {
 }
 
+InputSpec InputSpec::fromString(std::string s)
```
ktf: Did you check some of the utilities we already have in `Framework/Core/include/Framework/DataDescriptorQueryBuilder.h`?
If you really need something else, could you move this code there in any case?

Framework/Core/include/Framework/InputSpec.h
```diff
@@ -79,5 +82,7 @@ struct InputSpec {
   bool operator==(InputSpec const& that) const;
 };
 
+void updateInputList(std::vector<InputSpec>& list, InputSpec&& input);
```
ktf: Also this one should be moved to the utilities.

ktf:
Maybe, but things like `size() > 0` -> `!empty()` or `0` -> `0u` are both not strictly needed and questionably better from the readability point of view.

Framework/Core/src/InputSpec.cxx
aalkin: For the source table tracking I have a special format, as we also want to track if the particular input is enabled by a process switch. While I can use that for parsing the substring, it is easier and faster to just apply regex. My code is very specific for InputSpec metadata, though, DataDescriptorQueryBuilder has a different purpose.

Framework/Core/include/Framework/InputSpec.h
aalkin: In which header should I put it then?

Framework/Core/src/InputSpec.cxx
ktf: I would use either DataDescriptorQueryBuilder or DataSpecUtils. How to parse something to produce an InputSpec is not something which is part of the "essence" of being an `InputSpec`, so it should not be in `InputSpec.h`.

Framework/Core/include/Framework/InputSpec.h
ktf: See above, either DataDescriptorQueryBuilder or DataSpecUtils.

ktf:
Perfect. Thank you for your changes!

ktf (approved):

PR #7974, opened 2022-01-20
DPL: distinguish between exit and stop callbacks

ktf:


ktf:
@matthiasrichter @davidrohr @awegrzyn this should fix the current issue we saw with start/stop. The "Exit" callback was erroneously hooked to the PostRun() transition. This introduces a new callback ("stop") to be called on PostRun and moves "exit" to when DPL exits for real. It happened for monitoring just because monitoring is one of the services which we want to survive FairMQ state machine transitions (because it might be interesting to send metrics regardless of which state you are in).

awegrzyn:
Could we have daily tag once this will have merged? So we include it in today's release

ktf:
Yes, sure.

vascobarroso:
Any ETA for this ? Can we expect to have a tag with this in the coming hours or is it too ambitious ?

ktf:
The build is now started at:

https://alijenkins.cern.ch/job/DailyBuilds/job/AdHocO2Physics/86/

it should be there in a couple of hours.

ktf:
For the record, the build is there.

PR #7977, opened 2022-01-20
Introduction of a second charge-collection table in ALPIDE response

AlexBigO:
With this PR we introduced a second charge-collection table in ALPIDE response. 
A possibility to choose either of the tables or both from the command line is implemented as well.

For the moment, the second table is just a clone of the previously existing table for Vbb=-3V.
But it will eventually be replaced with a true table for Vbb=0V.


shahor02 (commented):
Hi @AlexBigO 

Thanks for the development. Not sure it makes sense to upload 800+ dummy files cloned from Vbb0 just to override them later by real Vbb3. Since the Vbb3 anyway cannot be used now, I would propose to leave the code part of your PR (making sure that if Vbb3 is selected it prints a proper message and produce fatal) + moving current data files to the proper Vbb0 directory.

AlexBigO:
> Hi @AlexBigO
> 
> Thanks for the development. Not sure it makes sense to upload 800+ dummy files cloned from Vbb0 just to override them later by real Vbb3. Since the Vbb3 anyway cannot be used now, I would propose to leave the code part of your PR (making sure that if Vbb3 is selected it prints a proper message and produce fatal) + moving current data files to the proper Vbb0 directory.

Hi @shahor02 

It is not yet clear when the true table for Vbb=0 will become available. The current idea is to try to tweak the existing table for Vbb=-3V, working freely on an independent clone of the table.
Does this create problems with managing the repository? If yes, we can remove the clone, of course. 
But eventually, we would have to commit the second table anyway.

shahor02:
Hi @AlexBigO 
I let to @ktf to comment if committing 800+ new files which are anyway cannot be used until they are completely overwritten (not just a small diff applied) will cause extra stress to the git repository. Personally, I would not do this.

ktf:
I agree it would be best to avoid it. How much bigger does the repository become? Do you have any other way this could be done?

shahor02:
> I agree it would be best to avoid it. How much bigger does the repository become? Do you have any other way this could be done?

6.4MB made of 817 files, duplicated from the data of the currently used map as a placeholder for the alternative map, which will be produced later.

shahor02:
@AlexBigO thanks for changing! Could you apply clang-format, then I'll approve to start CI tests.

shahor02 (dismissed):
approving to trigger CI tests.

iouribelikov (approved):

shahor02 (approved):

iouribelikov:
@everybody.   I do not understand... This PR was already "in green" during the weekend.
Now, 5 checks failed :(  
And I do not see anything calling for corrections in the logs.

Cannot we simply merge this PR ?

shahor02:
some builders are failing to start. Merging.

ktf:
For the record, Gitlab at CERN is having troubles, so this might result in failed builds with no logs.

PR #7979, opened 2022-01-20
Implementing multivariate polynomial class

matthias-kleiner:
multivariate polynomial class:
- The formula for the multivariate polynomials is unpacked at compile time and equally fast as the hardcoded formula when dimensions and degree is specified as template parameters
- The runtime version of the polynomials are significantly slower (dimensions and degree not known during compile time)

dE/dx:
- Using of template polynomial class for dE/dx correction instead of hardcoded formula for polynomials
- implementing loading of zero supression threshold map for dE/dx correction


DataFormats/Detectors/TPC/include/DataFormatsTPC/CalibdEdxTrackTopologyPol.h
```diff
@@ -12,33 +12,57 @@
...
-  {
-    clear();
-  }
-  CalibdEdxTrackTopologyPol(std::string_view fileName) { loadFromFile(fileName); }
-#else
-  CalibdEdxTrackTopologyPol() CON_DEFAULT;
+  /// constructor constructs an object Initialized from file
+  /// \param fileName name of the input file containing the object
+  /// \parma name name of the object
+  CalibdEdxTrackTopologyPol(std::string_view fileName, const char* name = "CalibdEdxTrackTopologyPol") { loadFromFile(fileName.data(), name); };
```
wiechula: Why once `string_view` and once `const char*`?

DataFormats/Detectors/TPC/include/DataFormatsTPC/CalibdEdxTrackTopologyPol.h
```diff
@@ -49,71 +73,114 @@ class CalibdEdxTrackTopologyPol
...
   /// \param relPad absolute relative pad position of the track
   /// \param relTime relative time position of the track
-  GPUd() float getCorrection(const int region, const ChargeType charge, const float tanTheta, const float sinPhi, const float z, const float relPad, const float relTime) const
+  GPUd() float getCorrection(const int region, const ChargeType charge, const float tanTheta, const float sinPhi, const float z, const float relPad, const float relTime, const float threshold = 0) const
   {
-    const auto& param = mParams[regionIndex(region, charge)];
-    const float x[FXDim]{tanTheta, sinPhi, z, relPad, relTime};
-    const float corr = evalPol4_5D(x, param);
+    const float x[]{tanTheta, sinPhi, z, relPad, relTime, threshold};
+    const float corr = charge == ChargeType::Tot ? mCalibPolsqTot[region].eval(x) : mCalibPolsqMax[region].eval(x);
```
wiechula: Perhaps
```suggestion
    const float corr = (charge == ChargeType::Tot) ? mCalibPolsqTot[region].eval(x) : mCalibPolsqMax[region].eval(x);
```
for readability.

wiechula (dismissed):
Just very minor things. Very nice! Especially the MultivariatePolynomial. Did you make some timing tests compared to the explicit formula?

matthias-kleiner:
> Just very minor things. Very nice! Especially the MultivariatePolynomial. Did you make some timing tests compared to the explicit formula?

Hi @wiechula, 
Thanks for the comments, I have changed it accordingly.

When performing the interpolation on CPU tested with googlebenchmark the compile time version is around 20% slower than the hardcoded formula and the run time version is around 17 times slower than the hardcoded formula.
However when measuring the time for the tracking (GPUTPCGMMergerTrackFit) on an NVIDIA GPU the difference seems negligible for the compile time version. [See this pdf (the y axis and y axis errors are just arbitrary to better distinguish the values.)](https://github.com/AliceO2Group/AliceO2/files/7914755/performance_tracking.pdf)

You can also look at the output from [godbolt](https://godbolt.org/z/7cK8e399d) (left panel compile time evaluation and right panel hard coded formula) to see the differences.

wiechula (dismissed):
Very nice! Thanks a lot for the additional information.

sgorbuno (dismissed):

wiechula (dismissed):

wiechula:
@shahor02 @davidrohr , I guess the error is unrelated. Can this be merged?

matthias-kleiner:
Hello @davidrohr @sgorbuno,

now an error appeared during the slc8-GPU test (https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/AliceO2/7979/a4302bcf068fcc27beee25c8402df65390b2aed1/build_O2_fullCI/pretty.html#errors) and I am not quite sure why it happens: `/sw/SOURCES/O2/7979-slc8_x86-64/0/GPU/TPCFastTransformation/MultivariatePolynomial.h:92:74: error: cannot initialize a parameter of type 'const __private float *' with an lvalue of type '__generic o2::gpu::MultivariatePolynomial::DataTParams *const __generic' (aka '__generic float *const __generic')
  GPUd() float eval(const float x[/*Dim*/]) const { return this->evalPol(mParams, x); }`

When I look to similar functions running on GPU e.g.: https://github.com/AliceO2Group/AliceO2/blob/4fc3847aa49643c67f06cb1447b4a26606c3dfa7/GPU/TPCFastTransformation/SplineSpec.h#L272 I see that there is sometimes a `GPUgeneric()` statement. 
Do you know if this will fix the problem above and when exactly do I need it? 

By looking at the definition https://github.com/AliceO2Group/AliceO2/blob/03608ff899d444d52571dbed14a0106ae4616562/GPU/Common/GPUCommonDefAPI.h#L47 it is not really clear to me when to use it and when it is not needed.

davidrohr:
It is needed in some cases when you are passing references/pointers to non-local variables to a function, which automatically deduces a template parameter from the type of that variable. The whole story is super complicated and related to OpenCL memory address space qualifiers.

basically, in
```
templace <class T> void foo(T& bla) {...}

int x;
boid bar()
{
  int y;
  foo(x);
  foo(y);
}
```
`foo(x)` yields `T = int&` while `foo(y)` yields `T = __global__ int&`, which are incompatible.
particularly, the latter will not work in case you have
```
templace <class T> void foo(T& bla) {
T z;
}
```
because it cannot instantiate `__global__ int z;` as local variable. This is prevented by adding `GPUgeneric()` before, which avoids that `__global__` becomes part of `T`.

matthias-kleiner:
Thanks for the nice explanation. I hope it will fix the error.

davidrohr:
thx, now it compiles. Have you tested on the EPN whether there are performance implications with the GPU?

matthias-kleiner:
> thx, now it compiles. Have you tested on the EPN whether there are performance implications with the GPU?

I havent done a performance test on EPNs. @wiechula Could you run such a test on EPNs?

davidrohr:
Please note, we have switched EPN access to using the slurm scheduler. Some instructions can be found here: https://indico.cern.ch/event/1111506/contributions/4675857/attachments/2370168/4047932/2022-01-01%20EPN%20access%20via%20Scheduler.pdf

I have added Jens to the users that may submit to the `prod` partition, so that he can use the GPUs.

davidrohr:
Any update on the performance test of the new class on the EPN?

wiechula:
@davidrohr , here the comparison
Black: no dE/dx calculation
Red: dE/dx without any calibrations applied
Orange: dE/dx with Kr map and the new topological correction of this PR
The overhead of applying these calibrations during the dE/dx is ~4.5%, the overhead of dE/dx with calibrations in the tracking is ~30%.
[gpuStats.pdf](https://github.com/AliceO2Group/AliceO2/files/8039133/gpuStats.pdf)

matthias-kleiner:
> @davidrohr , here the comparison Black: no dE/dx calculation Red: dE/dx without any calibrations applied Orange: dE/dx with Kr map and the new topological correction of this PR The overhead of applying these calibrations during the dE/dx is ~4.5%, the overhead of dE/dx with calibrations in the tracking is ~30%. [gpuStats.pdf](https://github.com/AliceO2Group/AliceO2/files/8039133/gpuStats.pdf)

Thanks @wiechula. Is it then ok to merge this @davidrohr ?

PR #7981, opened 2022-01-21
Fix data races in simulation merging using TBB

sawenzel:
std::unordered_map is not thread-safe, when one
thread is inserting and another reading.

Fix this by using dedicated and efficient
thread-safe hashmaps from TBB.
This is preferred over manual access management with
mutexes.


davidrohr (dismissed):
Tested with system TBB, compiles correctly

PR #7984, opened 2022-01-21
TPC: Decouple BetheBloch from Detector class

sawenzel:
Provide function in standalone header. Clients (such as O2Physics)
should not have to include the whole TPC detector to have access to
this function.


wiechula (commented):
@sawenzel , in general this is a good idea, we are working on moving this in parallel. If you need it now I would propose to move it to dataformatstpc, since it will be needed not only in simulation.

sawenzel:
Moved to DataFormatsTPC as suggested.

PR #7987, opened 2022-01-22
[O2-2749] Fix MCH mapping library implementations link options on Ubuntu

aphecetche:
To be accurate the change affects all non Apple platforms, but it's only on Ubuntu that it was not working properly without that fix.


aphecetche:
@shahor02 this one should solve the [problem you had](https://alice.its.cern.ch/jira/browse/O2-2749) with MCH digitizer. 

Actually, _all_ programs using the MCH mapping were broken on Ubuntu since #7755 (which I did test on CS8 and Mac, but  indeed not on Ubuntu. And as the CI does not run on Ubuntu either it went unnoticed until now... Sorry about that).

shahor02 (approved):
Thanks!

PR #7989, opened 2022-01-23
Multiple changes in AODProducerWorkflowSpec

shahor02:
### Multiple changes in AODProducerWorkflowSpec
    
*   Store ambiguous tracks only once, filling their BCslice corresponding to the time bracket as used in track-vertex matching.
*  Store BC slice for tracks w/o vertices
*  The time of ambiguous and orphan tracks is stored in ns wrt the globalBC corresponding to   bcSlice[0].
*  Add time info for MFT tracks
* Store in flags (at the moment for the barrel tracks only, since Fwd and MFT tracks have no  flags) the bit for the track being orphan, as well as the track time error being gaussian or  half-length of the range.
* Store in V0s and Cascades the collisonID.
* Add separate truncation data member for tracks time error.
* To guarantee required precision (set to 16ps) of the TOF tracks time the max distance to eventual reference globalBC is automatically estimated and if needed, dummy globalBCs are added to BCs table.

*   Add MFT track time and resolution to DataModel

*   Add flag for error being half-interval (non-gaussian)

*   Fix decomposition of TRD tracks indices

@jgrosseo The MFT tracks had no time info so I've added it to the model: https://github.com/shahor02/AliceO2/blob/384212ca1185de45b1885b50ab764677bb584890/Framework/Core/include/Framework/AnalysisDataModel.h#L463 but for some reason calling the cursor with time and error added does not work: if I uncoment https://github.com/shahor02/AliceO2/blob/384212ca1185de45b1885b50ab764677bb584890/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L277-L279 compilation fails with the complaint the too many arguments supplied. Any idea what could be the reason?

I see that for the Fwd and MFT tracks the precision truncation is not applied, was this done on purpose.

Currently, the flags exist only for barrel tracks (in Extra...), shall we add them also for the Fwd and MFT tracks?

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -24,11 +24,13 @@
 #include "DataFormatsZDC/ZDCEnergy.h"
 #include "DataFormatsZDC/ZDCTDCData.h"
 #include "CommonUtils/NameConf.h"
+#include "MathUtils/Utils.h"
 #include "DetectorsBase/GeometryManager.h"
 #include "CCDB/BasicCCDBManager.h"
 #include "CommonConstants/PhysicsConstants.h"
 #include "CommonDataFormat/InteractionRecord.h"
 #include "DataFormatsTRD/TrackTRD.h"
+#include "DataFormatsTRD/TrackTriggerRecord.h"
 #include "DataFormatsGlobalTracking/RecoContainer.h"
 #include "Framework/AnalysisDataModel.h"
 #include "Framework/ConfigParamRegistry.h"
@@ -44,6 +46,8 @@
 #include "ReconstructionDataFormats/Cascade.h"
 #include "MCHTracking/TrackExtrap.h"
 #include "MCHTracking/TrackParam.h"
+#include "ITSMFTBase/DPLAlpideParam.h"
+#include "DetectorsVertexing/PVertexerParams.h"
 #include "ReconstructionDataFormats/GlobalTrackID.h"
 #include "ReconstructionDataFormats/Track.h"
 #include "ReconstructionDataFormats/TrackTPCITS.h"
@@ -57,6 +61,7 @@
 #include "SimulationDataFormat/MCTruthContainer.h"
 #include "SimulationDataFormat/MCUtils.h"
 #include "ZDCBase/Constants.h"
+#include "TPCBase/ParameterElectronics.h"
 #include "GPUTPCGMMergedTrackHit.h"
 #include "TOFBase/Utils.h"
 #include "O2Version.h"
@@ -82,7 +87,7 @@ using SMatrix55Sym = ROOT::Math::SMatrix<double, 5, 5, ROOT::Math::MatRepSym<dou
 namespace o2::aodproducer
 {
 
-void AODProducerWorkflowDPL::collectBCs(o2::globaltracking::RecoContainer& data,
+void AODProducerWorkflowDPL::collectBCs(const o2::globaltracking::RecoContainer& data,
                                         const std::vector<o2::InteractionTimeRecord>& mcRecords,
                                         std::map<uint64_t, int>& bcsMap)
 {
@@ -93,6 +98,8 @@ void AODProducerWorkflowDPL::collectBCs(o2::globaltracking::RecoContainer& data,
   const auto& caloEMCCellsTRGR = data.getEMCALTriggers();
   const auto& ctpDigits = data.getCTPDigits();
   const auto& zdcBCRecData = data.getZDCBCRecData();
+
+  bcsMap[mStartIR.toLong()] = 1; // store the start of TF
 
   // collecting non-empty BCs and enumerating them
   for (auto& rec : mcRecords) {
@@ -233,15 +240,31 @@ void AODProducerWorkflowDPL::addToTracksExtraTable(TracksExtraCursorType& tracks
                     truncateFloatFraction(extraInfoHolder.tofExpMom, mTrack1Pt),
                     truncateFloatFraction(extraInfoHolder.trackEtaEMCAL, mTrackPosEMCAL),
                     truncateFloatFraction(extraInfoHolder.trackPhiEMCAL, mTrackPosEMCAL),
-                    truncateFloatFraction(extraInfoHolder.trackTime, mTrackSignal),
-                    truncateFloatFraction(extraInfoHolder.trackTimeRes, mTrackSignal));
+                    truncateFloatFraction(extraInfoHolder.trackTime, mTrackTime),
+                    truncateFloatFraction(extraInfoHolder.trackTimeRes, mTrackTimeError));
 }
 
-template <typename mftTracksCursorType>
-void AODProducerWorkflowDPL::addToMFTTracksTable(mftTracksCursorType& mftTracksCursor,
-                                                 const o2::mft::TrackMFT& track, int collisionID)
+template <typename mftTracksCursorType, typename AmbigMFTTracksCursorType>
+void AODProducerWorkflowDPL::addToMFTTracksTable(mftTracksCursorType& mftTracksCursor, AmbigMFTTracksCursorType& ambigMFTTracksCursor,
+                                                 GIndex trackID, const o2::globaltracking::RecoContainer& data, int collisionID,
+                                                 std::uint64_t collisionBC, const std::map<uint64_t, int>& bcsMap)
 {
   // mft tracks
+  int bcSlice[2] = {-1, -1};
+  const auto& track = data.getMFTTrack(trackID);
+  const auto& rof = data.getMFTTracksROFRecords()[mMFTROFs[trackID.getIndex()]];
+  float trackTime = rof.getBCData().differenceInBC(mStartIR) * o2::constants::lhc::LHCBunchSpacingNS + mMFTROFrameHalfLengthNS;
+  float trackTimeRes = mMFTROFrameHalfLengthNS;
+  bool needBCSlice = collisionID < 0 || trackID.isAmbiguous();
+  std::uint64_t bcOfTimeRef;
+  if (needBCSlice) {
+    double error = mTimeMarginTrackTime + trackTimeRes;
+    bcOfTimeRef = fillBCSlice(bcSlice, trackTime - error, trackTime + error, bcsMap);
+  } else {
+    bcOfTimeRef = collisionBC - mStartIR.toLong(); // by default (unambiguous) track time is wrt collision BC
+  }
+  trackTime -= bcOfTimeRef * o2::constants::lhc::LHCBunchSpacingNS;
+
   mftTracksCursor(0,
                   collisionID,
                   track.getX(),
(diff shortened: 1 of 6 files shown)


shahor02:
Removing WIP to run tests, do not merge even if green.

ktf:
The error is genuine.

jgrosseo:
"@jgrosseo The MFT tracks had no time info so I've added it to the model: https://github.com/shahor02/AliceO2/blob/384212ca1185de45b1885b50ab764677bb584890/Framework/Core/include/Framework/AnalysisDataModel.h#L463 but for some reason calling the cursor with time and error added does not work: if I uncoment https://github.com/shahor02/AliceO2/blob/384212ca1185de45b1885b50ab764677bb584890/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L277-L279 compilation fails with the complaint the too many arguments supplied. Any idea what could be the reason?"

-> Tables are redefined in the AODProducerWorkflowSpec.h (not sure if that could be done better). You need to add the columns also there.

jgrosseo:
"I see that for the Fwd and MFT tracks the precision truncation is not applied, was this done on purpose." 
--> No, we should do it.

jgrosseo:
- Thanks for adding the TrackTime to MFT. Do we need a resolution actually given that it is only one detector contributing?
- A FwdTrack (which has already TrackTime) references an MFTTrack and therefore such a track has two TrackTimes defined. It is not a problem, I mention it for completeness.
- I agree that we should add Flags to MFT and FwdTrack. For MFT it can be done like this, for FwdTrack it invalidates the converted data, so we need to apply there already now the schema evolution...

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -827,6 +831,57 @@ void AODProducerWorkflowDPL::fillMCTrackLabelsTable(const MCTrackLabelCursorType
...
+      LOG(warn) << "Could not find a positive track index for prong ID " << trPosID;
+    }
+    item = mGIDToTableID.find(trNegID);
+    if (item != mGIDToTableID.end()) {
+      negTableIdx = item->second;
+    } else {
+      LOG(warn) << "Could not find a negative track index for prong ID " << trNegID;
+    }
+    if (posTableIdx != -1 and negTableIdx != -1) {
+      v0Cursor(0, v0.getVertexID(), posTableIdx, negTableIdx);
```
jgrosseo: `v0.getVertexID()` is the entry number in the collision table?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -980,6 +1035,8 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
     mTrackCovDiag = 0xFFFFFFFF;
     mTrackCovOffDiag = 0xFFFFFFFF;
     mTrackSignal = 0xFFFFFFFF;
+    mTrackTime = 0xFFFFFFFF;
+    mTrackTimeError = 0xFFFFFFFF;
```
jgrosseo, line 1051: Why do we have different defaults here than above? (`-999`)

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: you are right, for multi-TF processing this will create problems, will use mapping from vertexID to collisionID. But them I am not sure if the old code is correct: the reconstructed track indices are mapped to stored AOD index as https://github.com/AliceO2Group/AliceO2/blob/30327663640c088193254dd45d3e3a6f92040a1f/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L364-L365 (used to indicate V0's prong's indices). 
But they are reset after processing each TF:   https://github.com/AliceO2Group/AliceO2/blob/30327663640c088193254dd45d3e3a6f92040a1f/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L1475-L1476
It is OK to reset the map, since the reco indices are specific for each TF, but resetting the `mTableTrID` will break the correspondence to AOD track index (which is accumulated from one TF to another), right?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: sorry, did not understand the question: the lines you refer to are compression masks (when no compression is requested, what they have to do with -999?)

rpezzi:
> Thanks for adding the TrackTime to MFT. Do we need a resolution actually given that it is only one detector contributing?

MFT track time resolution depends on the readout configuration defined at the start of each run. There should be an optimal way to store this and propagate to AOD. Notice that there are also corner cases at the interface of ROFs where tracks are reconstructed only after merging clusters in two consecutive read out frames - this case has not yet been addressed in the MFT reconstruction but could be handled by a flag.

We have not foreseen MFT flags yet, but it may come handy in other cases as well. What should be a good place for flagging MFT tracks? MFTTrackExtra or add it to the main MFTTrack table?

Thanks a lot.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
jgrosseo: Sorry, forget about it.

shahor02:
>> >> "I see that for the Fwd and MFT tracks the precision truncation is not applied, was this done on purpose."

>> --> No, we should do it.

OK, will add this

>> Do we need a resolution actually given that it is only one detector contributing?

ITS and MFT tracks time uncertainties depend on used strobe lengths which are the parameters of the run. Currently, they are constant within the run (so will be compressed out) but this may change when the tracking will properly account the tracks on the edge of the strobe (with clusters found in 2 strobes the time resolution will be on the level of ~100 ns).

>>  A FwdTrack (which has already TrackTime) references an MFTTrack and therefore such a track has two TrackTimes defined. It is not a problem, I mention it for completeness.

1st of all, the time of FwdTrack is more precise and different from the corresponding MFT track. And only a tiny fraction of MFT tracks have the associated FwdTrack. But in general, the way the MFT and Fwd tracks are related to each other breaks the "minimum redundancy" approach used for the barrel: e.g. ITS tracks participating in the global tracks are not counted as independent tracks (not stored in the AOD, instead, some of their features are stored in the Extra of the global track). 
In principle, the should have been done for the Fwd and MFT track on the level of RecoContainer selection: flag also MFT tracks as used in https://github.com/AliceO2Group/AliceO2/blob/30327663640c088193254dd45d3e3a6f92040a1f/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L284
Don't know if this was not done on purpose (given the tiny amount of the ForwardTracks). Pinging to @rpezzy and @aphecetche. 

>>   I agree that we should add Flags to MFT and FwdTrack. For MFT it can be done like this, for FwdTrack it invalidates the converted data, so we need to apply there already now the schema evolution...

Will you add it? I am not feeling confident and modifications in the data model.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1541,6 +1580,299 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+      }
+      double massZ = o2::track::PID::getMass2Z(data.getTrackParam(trackIndex).getPID());
+      double energy = sqrt((massZ * massZ) + (tofExpMom * tofExpMom));
+      double exp = intLen * energy / (cSpeed * tofExpMom);
+      auto tofSignal = (tofMatch.getSignal() - exp) * 1e-3; // time in ns wrt TF start
+      auto bc = relativeTime_to_GlobalBC(tofSignal);
+
+      auto it = bcsMap.lower_bound(bc);
+      if (it == bcsMap.end() || it->first > bc + maxGapBC) {
+        bcsMap.emplace_hint(it, bc, 1);
```
jgrosseo, line 1840: So you now add the dummy one closest to the tof time, right?
An alternative would be to use a bc number which actually had a valid bunch crossing. So all BCs would sit where we were sure true bunch crossings have taken place (even if it did not trigger any data taking).

Framework/Core/include/Framework/DataTypes.h
```diff
@@ -35,7 +35,8 @@ enum TrackTypeEnum : uint8_t {
 };
 enum TrackFlags {
   TrackTimeResIsRange = 0x1, // Gaussian or range
-  PVContributor = 0x2        // This track has contributed to the collision vertex fit
+  PVContributor = 0x2,       // This track has contributed to the collision vertex fit
+  OrphanTrack = 0x4          // Track has no association with any collision vertexd
```
jgrosseo, line 38: As you are on it, could you remove `ITSStandaloneTrack`  in line 32. I think it is not needed anymore.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: This will not work for the cosmics. Also, for the beam with a few BCs only it is not guaranteed that you can find a BC with the `maxGapBC`. In any case, this fake BC is merely a reference.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
jgrosseo: Yes, it does not look sane.
Better in any case would be `mGIDToTableID.emplace(trackIndex, tracksCursor.lastIndex()); `

Framework/Core/include/Framework/DataTypes.h
shahor02: ok, done

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
jgrosseo: @aalkin @ktf Is `lastIndex` not defined for the cursor? We use it in O2Physics task on something defined by `Produces`. How can one have the last insert id here?

jgrosseo:
> > I agree that we should add Flags to MFT and FwdTrack. For MFT it can be done like this, for FwdTrack it invalidates the 
> > converted data, so we need to apply there already now the schema evolution...

> Will you add it? I am not feeling confident and modifications in the data model.

Yes, I put it on my list.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
jgrosseo: I thought 1 BC per turn gives us already 4.5 ps precision and here you aim at 16ps?
Yes, cosmics needs to be treated differently.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: ok, I'll simply avoid resetting the counters like mTableTrID

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: Wth the maxGap=orbit and for the non-truncated float the precision in ps is indeed 
```
88924.e3/(1<<24)
(double) 5.3002834
```
but with default truncation of 8 lowest bits it is not enough:
```
88924.e3/(1<<16)
(double) 1356.8726
```
So, I am calculating the maxGap depending on the requested precision and adding fake BC to respect it.

rpezzi:
> In principle, the should have been done for the Fwd and MFT track on the level of RecoContainer selection: flag also MFT tracks as used in
> 
> https://github.com/AliceO2Group/AliceO2/blob/30327663640c088193254dd45d3e3a6f92040a1f/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L284
> Don't know if this was not done on purpose (given the tiny amount of the ForwardTracks). 
Hi Ruben, this was not done on purpose. We do not intend to exclude used MFT or MCH tracks used in global forward tracks.

shahor02:
Hi @rpezzi 

>> Hi Ruben, this was not done on purpose. We do not intend to exclude used MFT or MCH tracks used in global forward tracks.

OK, so actually it was done on purpose: when filling MFT-MCH tracks you don't flag its constituent MFT and MCH as used, so they are provided to the data requestor on top of the global tracks.

shahor02:
@jgrosseo Shall I merge it or it will affect the O2Physics? I still did not understand if recovering multiple vertices from the BCSlice is already supported.

jgrosseo:
I'd merge it. Not everything is supported but this should not cause regressions and is an important step to the final data model.

jgrosseo (approved):

PR #7992, opened 2022-01-23
Condition analysis

ktf:


ktf:
This is AliceO2Group/AliceO2#7941, plus the bit to support analysis.

ktf:
Tested on hyperloop. No regressions, merging.

PR #7998, opened 2022-01-24
DPL Analysis: add exception for unsorted unassigned groups

aalkin:
If unassigned groups are in incorrect order (or split), fast grouping is invalid due to incorrect offsets, so the execution should stop.

### DIFF
--- a/Framework/Core/include/Framework/Kernels.h
+++ b/Framework/Core/include/Framework/Kernels.h
@@ -81,22 +81,29 @@ auto sliceByColumn(
   for (auto i = 0; i < column->num_chunks(); ++i) {
     T prev = 0;
     T cur = 0;
-    T nprev = -1;
+    T lastNeg = -1;
+    T lastPos = 0;
+
     auto array = static_cast<arrow::NumericArray<typename detail::ConversionTraits<T>::ArrowType>>(column->chunk(i)->data());
     for (auto e = 0; e < array.length(); ++e) {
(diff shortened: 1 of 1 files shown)


ktf:
The change itself is fine, but can we remove the templating of sliceByColumn?

aalkin:
Yes, but it needs some changes in other places, I have it on my todo list.

ktf (dismissed):

jgrosseo (dismissed):
This is fine but does not catch broken groups, right?
So 1 1 1 2 2 2 -1 -1 2 2 will not trigger any assert right?

What keeping last_positive, last_negative and prev
- if cur == prev do nothing, else:
- if cur >= 0 && cur <= last_positive then assert
- if cur < 0 && cur >= last_negative then assert
- else continue and set last_positive / last_negative

aalkin:
@jgrosseo something like this?

jgrosseo (commented):
Why it does not assert on startup without changing lastPos to start from -1 and lastNeg from 0?

jgrosseo:
I think if we start with -1, it should give an assert right away...

aalkin:
You are right, the first iteration needs a special care.

aalkin:
Setting all initial values to 0 should prevent any of the exceptional conditions triggering when parsing the initial `0... -1...` or `-1... 0...` segment.

jgrosseo (approved):

PR #8000, opened 2022-01-24
Add DCS Config -> CCDB codes in Detectors/ITSMFT/MFT/condition

syano0822:
Add DCS config -> CCDB function in MFT


rpezzi (dismissed):

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigInfo.h
```diff
@@ -0,0 +1,79 @@
...
+    add = mAdd;
+  }
+  void getType(int& type)
+  {
+    type = mType;
+  }
+  void getVersion(std::string& version)
+  {
+    version = mVersion;
+  }
```
shahor02: These methods should be declared as `const`. But why do you pass the target argument instead of using traditional getters, e.g. `const std::string& geVersion() const { return Version; }` ?

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
```diff
@@ -0,0 +1,135 @@
...
+namespace o2
+{
+namespace mft
+{
+
+class DCSConfigUtils
+{
+
+ public:
+  void init(DCSConfigInfo info)
```
shahor02: `void init(const DCSConfigInfo& info)` ?

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
```diff
@@ -0,0 +1,137 @@
...
+  {
+    return mVersion;
+  }
+  const std::string& getName() const
+  {
+    if (mType == 0) {
+      return mMapAddNameRU.find(mAdd)->second;
+    } else if (mType == 1) {
+      return mMapAddNameALPIDE.find(mAdd)->second;
+    }
```
rpezzi, line 69: @syano0822 , macOS tests are still failing due to missing return.

rpezzi (approved):

shahor02 (approved):
CI failures are unrelated

PR #8002, opened 2022-01-25
DPL Analysis: add template to declare index binding equivalence

aalkin:
@jgrosseo this should solve the index problem with versioning, for example

`DECLARE_EQUIVALENT_FOR_INDEX(McParticles, McParticles_001)`


jgrosseo:
Very nice! Is something still missing as it is in draft?

aalkin:
@jgrosseo it can only be tested together with your versioning modification, it would be easier if you integrate this change there.

jgrosseo:
OK, I'll do that.

aalkin:
@jgrosseo this change should fix the "cannot be used in constexpr" problem, and the declaration itself can be placed outside `o2::aod` namespace, adding `o2::aod::` to the table types, of course.

PR #8003, opened 2022-01-25
GRP DCS DP processing

chiarazampolli:
Eg. to run:
```
o2-calibration-grp-dcs-dps-sim-workflow --max-timeframes 100 --delta-fraction 0.5 --max-cycles-no-full-map 3 | o2-calibration-grp-dcs-dps-workflow  | o2-calibration-ccdb-populator-workflow --ccdb-path http://localhost:8080
```

### DIFF
--- a/Detectors/Calibration/include/DetectorsCalibration/Utils.h
+++ b/Detectors/Calibration/include/DetectorsCalibration/Utils.h
@@ -31,19 +31,18 @@ namespace o2
 namespace calibration
 {
 
-using TFType = uint64_t;
-
 struct Utils {
   static constexpr o2::header::DataOrigin gDataOriginCDBPayload{"CLP"}; // generic DataOrigin for calibrations payload
   static constexpr o2::header::DataOrigin gDataOriginCDBWrapper{"CLW"}; // generic DataOrigin for calibrations wrapper
+  static constexpr long INFINITE_TIME = 99999999999999;
   template <typename T>
-  static void prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const std::string& path, TFType tf,
-                                    const std::map<std::string, std::string>& md);
+  static void prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const std::string& path,
+                                    const std::map<std::string, std::string>& md, long start, long end = -1);
 };
 
 template <typename T>
-void Utils::prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const std::string& path, TFType tf,
-                                  const std::map<std::string, std::string>& md)
+void Utils::prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const std::string& path,
+                                  const std::map<std::string, std::string>& md, long start, long end)
 {
 
   // prepare all info to be sent to CCDB for object obj
@@ -52,8 +51,8 @@ void Utils::prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const 
   info.setPath(path);
   info.setObjectType(clName);
   info.setFileName(flName);
-  info.setStartValidityTimestamp(tf);
-  info.setEndValidityTimestamp(99999999999999);
+  info.setStartValidityTimestamp(start);
+  info.setEndValidityTimestamp(end);
   info.setMetaData(md);
 }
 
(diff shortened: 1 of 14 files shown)


Detectors/GRP/workflows/src/GRPDCSDPsSpec.cxx
```diff
@@ -0,0 +1,247 @@
...
+    mProcessor->useVerboseMode();
+  }
+  mProcessor->init(vect);
+  mTimer = HighResClock::now();
+}
+//__________________________________________________________________
+
+void GRPDCSDPsDataProcessor::run(o2::framework::ProcessingContext& pc)
+{
+  auto tfid = o2::header::get<o2::framework::DataProcessingHeader*>(pc.inputs().get("input").header)->startTime;
```
shahor02: `startTime` is a dummy counter, it cannot be used as an alias for the time. Moreover, when it comes from the dcs-proxy, where the TF has no usual meaning, it even cannot be converted to the abs. time.
If you want, I can change the dcs2dpl to send 'now()' in the `DataProcessingHeader::creation`, then you can use 
```
auto creationTime = DataRefUtils::getHeader<DataProcessingHeader*>(pc.inputs().getFirstValid(true))->creation;
```
directly for the time. Alternatively, you can simply use `now()` here. But the most precise would be extracting the timestamp from the DP itself.

Detectors/Calibration/include/DetectorsCalibration/Utils.h
```diff
@@ -22,17 +22,41 @@
...
 struct Utils {
   static constexpr o2::header::DataOrigin gDataOriginCDBPayload{"CLP"}; // generic DataOrigin for calibrations payload
   static constexpr o2::header::DataOrigin gDataOriginCDBWrapper{"CLW"}; // generic DataOrigin for calibrations wrapper
+  template <typename T>
+  static void prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const std::string& path, TFType tf,
+                                    const std::map<std::string, std::string>& md);
 };
 
+template <typename T>
+void Utils::prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const std::string& path, TFType tf,
```
shahor02: The TF is not a time, it cannot be used to populate the CCDB. Since these are generic methods, why don't you simply add explicit arguments for start/end time
```
  static void prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const std::string& path, 
                                    const std::map<std::string, std::string>& md, std::uint64_t tStart, std::unit64_t tEnd);
```

Detectors/GRP/workflows/src/GRPDCSDPsSpec.cxx
chiarazampolli: Hi Ruben, 
This was the plan, I started with `startTime` more as a placeholder (see also our other mail exchange). I planned to add a comment to the PR, but I forgot :-) 
I would use for each object probably the oldest time of all datapoints, or, as you say, "now". 
This is something that should be changed in several places, which is also why I kind of postponed it.
In general, it would be better to change the dcs2dpl, also for the configurations.

Detectors/Calibration/include/DetectorsCalibration/Utils.h
chiarazampolli: Yes, I can change that too.

Detectors/GRP/workflows/src/GRPDCSDPsSpec.cxx
chiarazampolli: I decided to not use the timestamps of the measurements because they might be out of sync, and they could give the wrong message. 
I am using now "creation", and if this is -1, then I use "now". 
But I think we should set it in the proxy, same for the configurations.

shahor02: Then simply use 
```
auto creationTime = DataRefUtils::getHeader<DataProcessingHeader*>(pc.inputs().getFirstValid(true))->creation;
```
I will add its filling to the proxy today.

chiarazampolli: OK, I now have:

```
auto startValidity = o2::header::get<o2::framework::DataProcessingHeader*>(pc.inputs().get("input").header)->creation;
```

but I can change it to the one above (I guess they are equivalent, no?).

chiarazampolli: done.

chiarazampolli:
Hi @TimoWilken , @ktf , 
macos is stuck, I think the same was reported somewhere else in other PRs, right?
@shahor02 , were the comments above the only ones you had?
Chiara

shahor02 (approved):
The macOS is stuck since a few days, merging.

PR #8004, opened 2022-01-25
DPL: allow GUI to transition states

ktf:


ktf:
@matthiasrichter @rbx @dennisklein I have done this PR to allow mimicking the `RUN->STOP->RUN` transition which we do in online and which results in DPL not seeing the data.
The transition itself seems to work correctly, and I do see the expected messages:

```
[STATE] RUNNING ---> READY
[26403:A]: [13:06:23][INFO] Now in READY
[26403:A]: [13:06:23][INFO] Changing to RUN
[26403:A]: [13:06:23][STATE] READY ---> RUNNING
[26403:A]: [13:06:23][INFO] Now in RUNNING
[26403:A]: [13:06:23][INFO] DEVICE: Running...
[26403:A]: [13:06:23][INFO] invoked at start
```

and then the Run method of the device is called correctly. However, for some reason I get that:

```
NewStatePending() == true
```

and the second time the Run() method is invoked it exits immediately. @dennisklein @rbx am I doing something wrong or is  this a race condition? Shouldn't "NewStatePending" be false right before entering a given state?

dennisklein:
Looking at the changes in this PR, it looks to me as if your `stateWatcher` could be calling `ChangeState` on entering `RUNNING` state, no? But I cannot match your log snippet to the code in here - did you trim it? And I do not understand, what `registry.get<DeviceState>()` is returning on each invocation.

dennisklein:
see https://github.com/FairRootGroup/FairMQ/blob/ebcbe2dde6224b89ebc1df085401847b4014b826/fairmq/StateMachine.cxx#L202-L203, this means `stateWatcher` is called before `Run()` is called.

ktf:
No, when entering the Running state the vector with the state transitions I keep in DeviceState (which for what we are concerned here is just a vector<std::string>) is actually empty, since I fill it with 2 elements every time you press the button and a pop_back() each time I trigger a transition.

`registry.get<DeviceState>()` is simply returning a pointer to an instance of DeviceState, which is an DPL internal structure.

dennisklein:
just for reference, we found that the issue is caused by using the builtin static controller of FairMQ which queues state transitions as soon as the RUNNING state is left the first time.

ktf:
@matthiasrichter this + https://github.com/FairRootGroup/FairMQ/pull/415 should allow you to select a device a "Restart" it by clicking  on the associated button in the DeviceInspector.

ktf:
@TimoWilken @MichaelLettrich it looks like one of the mac machines has troubles again with autotools. Is this always the same one? Can we get rid of it?

ktf:
Hold until new FairMQ adopted.

PR #8006, opened 2022-01-25
[EMCAL-751] Fixing the problem with pre-trigger collisions

hahassan7:
The EMC digitizer was crashing because of pre-trigger collisions. The problem is now fixed.


mfasDa (approved):
Looks OK to me

mfasDa:
CI failure unrelated

PR #8007, opened 2022-01-25
[WIP] Tesbench for CCDB fetcher callback 

shahor02:
Hi @ktf,

the 1st 3 commits are necessary for the fetcher to work properly, the last 3 others are temporary patches for extra verbosity and testbench: overall it works fine except the callback: I see that it is recognized since 
https://github.com/shahor02/AliceO2/blob/a8fc9aaa63b504937d2823dc6eb74ece8450d8a5/Framework/Core/include/Framework/Task.h#L126 is printed and an attempt to call it is made since https://github.com/AliceO2Group/AliceO2/blob/a8fc9aaa63b504937d2823dc6eb74ece8450d8a5/Framework/Core/include/Framework/CallbackService.h#L116 logs the call. But then the https://github.com/shahor02/AliceO2/blob/a8fc9aaa63b504937d2823dc6eb74ece8450d8a5/Detectors/ITSMFT/common/workflow/src/EntropyDecoderSpec.cxx#L147 is never called. 

The problem can be reproduced by unpacking https://cernbox.cern.ch/index.php/s/EVFNwBQHH2S2GWM and running 
```
o2-ctf-reader-workflow --max-tf 80 --onlyDet MFT  --ctf-input o2_ctf_run00507092_orbit0000000252_tf0000000001.root --condition-backend http://ccdb-test.cern.ch:8080  --run | tee xx.log
```
It will process 80 TFs, fetching specially tailored `Test/GRP/Data` and `Test/MFT/AlpideParam` objects (there are 7 updates over the timespan of these 80 TFs.

### DIFF
--- a/Detectors/ITSMFT/common/workflow/include/ITSMFTWorkflow/EntropyDecoderSpec.h
+++ b/Detectors/ITSMFT/common/workflow/include/ITSMFTWorkflow/EntropyDecoderSpec.h
@@ -15,6 +15,7 @@
 #ifndef O2_ITSMFT_ENTROPYDECODER_SPEC
 #define O2_ITSMFT_ENTROPYDECODER_SPEC
 
+#include "Framework/CCDBParamSpec.h"
 #include "Framework/DataProcessorSpec.h"
 #include "Framework/Task.h"
 #include "Headers/DataHeader.h"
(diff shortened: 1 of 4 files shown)


Framework/Core/src/CCDBHelpers.cxx
```diff
@@ -74,7 +74,7 @@ CCDBHelpers::ParserResult CCDBHelpers::parseRemappings(char const* str)
     switch (state) {
       case IN_BEGIN: {
         if (*str == 0) {
-          return {remappings, "Empty string provided"};
```
shahor02: @ktf this was triggering an exception which supposed to print details https://github.com/AliceO2Group/AliceO2/blob/592ed27ed7deedf5931bed170f5ac612720eec22/Framework/Core/src/CCDBHelpers.cxx#L145 but instead was producing 
```
[1526580:internal-dpl-ccdb-backend]: [14:16:32][FATAL] Unhandled o2::framework::runtime_error reached the top of main of o2-ctf-reader-workflow, device shutting down.
[1526580:internal-dpl-ccdb-backend]:  Reason:
[1526580:internal-dpl-ccdb-backend]:  Backtrace follow: 
```

ktf: Yes, of course an empty string for the remappings is not an error.

ktf:
Your changes look fine to me. Maybe we can have the first 4 commits in a different PR and merge them right away?

ktf:
I will check what is wrong with the callback this evening.

shahor02:
Ok, will open separated or for the stable part

PR #8010, opened 2022-01-26
Fixes for CCDB Fetcher

shahor02:


shahor02:
Failure is unrelated, merging

PR #8014, opened 2022-01-26
Using single quotes around command line arguments in DDS configuration

matthiasrichter:
This is related to #7972. If the problem reported there is related to using double quotes, we can switch to single quotes. But this might be a mistake in the updated unit test, e.g https://github.com/AliceO2Group/AliceO2/pull/7972/files#diff-04403506b63a13d27536d4c1fc04e50ca05f63428f2a1d526dbfc526ca2578deL123

Not sure if we want to switch to single quote, we drop this PR if not. DDS configurations are probably using double quotes on purpose.


davidrohr:
I have changed my ctest in #7972 to double quotes, the DDS ctest works now. I think there is no reason to generally change to single quotes.

matthiasrichter:
Closing, as this is not an option.

PR #8015, opened 2022-01-26
fix typo in cascades sorting/merging

shahor02:


shahor02:
trivial fix, tested locally, merging since the current dev is broken

PR #8016, opened 2022-01-26
[EMCAL-566] Added function for fast projection of boost histograms

jokonig:
- New function for faster projection needed as boost build-in functions
"project and reduce" are very slow (takes 20 min for approx. 10000
projections). New function copies the content of the desired bins in the 2d boost histogram into a
new 1d boost histogram and returns the 1d histogram.
- Improved offline calibration macro for time calibration: Added option
for fit range and possibility to store calibration factors in local root
file
- Added faster projection to calib extractor


mfasDa (approved):
Looks good to me!

mfasDa:
@sawenzel @shahor02 I think this can be merged. The macO2 CI errors are unrelated ...

shahor02 (approved):

PR #8018, opened 2022-01-26
CPV: Remove unnecessary CCDB server check

sawenzel:
Remove check if CCDB server is reachable, for the following
reasons:

- BasicCCDBManager fatals anyway when it can't deliver an object

- The error handling can be done depending on if we get an object or not

- The isHostReachable() check prevents use with local CCDB snapshots

- in removing this line, we are able to run the O2DPG MC workflow
  against a CVMFS CCDB snapshot (verified without network connection)

### DIFF
--- a/Detectors/CPV/simulation/src/Digitizer.cxx
+++ b/Detectors/CPV/simulation/src/Digitizer.cxx
@@ -36,11 +36,6 @@ void Digitizer::init()
     LOG(info) << "[CPVDigitizer] No reading calibration from ccdb requested, set default";
   } else {
     auto& ccdbMgr = o2::ccdb::BasicCCDBManager::instance();
-    bool isCcdbReachable = ccdbMgr.isHostReachable(); //if host is not reachable we can use only dummy calibration
-    if (!isCcdbReachable) {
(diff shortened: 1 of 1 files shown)


sevdokim:
Dear @sawenzel 
Digitizer is a part of CPVDigitizerSpec.cxx which is DPL processor. According to @shahor02 announcement, it is not allowed to access ccdb manager in the DPL workflows anymore so not only these part of code to be removed but the whole code of digitizer must be free of ccdb manager. I was about to do so but your commit confused me a little bit. Please confirm that my understanding is correct or not. Thanks!

sawenzel:
@sevdokim : Your understanding is correct: Finally, there need to be bigger refactorings to be done. This commit is not related to this but a small step that I need urgently for something else (being able to use a CCDB snapshot from CVMFS).
For you, nothing should change.

shahor02:
Hi @sevdokim 

For the record: my comment that the processors should not create CCDB Api instances was referring to the moment we fully deploy the DPL CCDB service.

sevdokim:
@sawenzel ok, thanks for the clarification!

PR #8005, opened 2022-01-25
DPL Analysis: disentangle slicing kernels

aalkin:
* Remove unnecessary templates
* Move code to .cxx
* Consolidate usage of arrow value_counts kernel
* Remove unnecessary includes

@ktf

PR #8020, opened 2022-01-26
Revert "DPL Analysis: add exception for unsorted unassigned groups"

jgrosseo:
Reverts AliceO2Group/AliceO2#7998


jgrosseo:
In case we have no other fix before the tag tonight.
Let me know @aalkin

Pinging also @benedikt-voelkel

aalkin:
@jgrosseo I'm in favor of merging this one to unblock analysis. The proper fix will come with #8005, but it also moves the code to another file.

PR #8021, opened 2022-01-26
Parallelization of ITS threshold calibration workflow + new features

iravasen:
Changes applied:
- parallelization of thr extraction and sum of thr for avg calculation
- modification of avg and rms calculation: sum values as soon as they are calculated so that we don't take into memory all thresholds
- modification of tree format to allow parallelization: for each row we have now C arrays.
- class name changed from ITSCalibrator to ITSThresholdCalibrator (as suggested by Ruben)
- Removed commented lines of code (as suggested by Ruben)
- added option nthreads to chose the number of threads (def is 1) 
- added avg/rms calculation also for thr scan so that we push to CCDB also avg thr&noise and rms for each chip (as for the tuning scans)

### DIFF
--- a/Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
+++ b/Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
@@ -111,11 +111,11 @@ void ITSThresholdCalibrator::init(InitContext& ic)
   this->mVerboseOutput = ic.options().get<bool>("verbose");
 
   //Get number of threads
-  try {
-    this->mNThreads = ic.options().get<int>("nthreads");
-  } catch (std::exception const& e) {
-    LOG(warning) << "Input parameter nthreads not found"
-                 << "\n*** Setting nthreads equal to 1";
+  this->mNThreads = ic.options().get<int>("nthreads");
+
+  //Check fit type vs nthreads (fit option is not thread safe!)
+  if (mFitType == FIT && mNThreads > 1) {
+    throw std::runtime_error("Multiple threads are requested with fit method which is not thread safe");
   }
 
   return;
@@ -315,7 +315,7 @@ bool ITSThresholdCalibrator::findThresholdDerivative(
   }
 
   int deriv_size = upper - lower;
-  float* deriv = new float[deriv_size];
+  float deriv[deriv_size];
   float xfx = 0, fx = 0;
 
   // Fill array with derivatives
@@ -337,7 +337,6 @@ bool ITSThresholdCalibrator::findThresholdDerivative(
   stddev /= fx;
   noise = std::sqrt(stddev);
 
-  delete[] deriv;
   return (noise < 15);
 }
 
@@ -362,7 +361,7 @@ bool ITSThresholdCalibrator::findThresholdHitcounting(
   // if (numberOfHits < N_INJ) { return false; }
   if (!is50) {
     if (this->mVerboseOutput) {
-      LOG(warning) << "Too few hits, skipping this pixel";
+      LOG(warning) << "Calculation unsuccessful: too few hits. Skipping this pixel";
     }
     return false;
   }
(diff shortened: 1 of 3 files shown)


Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -106,12 +110,20 @@ void ITSCalibrator::init(InitContext& ic)
 
   this->mVerboseOutput = ic.options().get<bool>("verbose");
 
+  //Get number of threads
+  try {
+    this->mNThreads = ic.options().get<int>("nthreads");
+  } catch (std::exception const& e) {
+    LOG(warning) << "Input parameter nthreads not found"
+                 << "\n*** Setting nthreads equal to 1";
+  }
```
shahor02: you have default nthreads=1 in the option definition, why do you need this try/catch?

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -288,7 +299,7 @@ bool ITSCalibrator::findThresholdFit(
 // data is the number of trigger counts per charge injected;
 // x is the array of charge injected values;
 // NPoints is the length of both arrays.
-bool ITSCalibrator::findThresholdDerivative(
+bool ITSThresholdCalibrator::findThresholdDerivative(
   const char* data, const short int* x, const short int& NPoints,
```
shahor02, line 303: You can save some cpu time by using a stack array instead of `new[]/delete[]` at https://github.com/AliceO2Group/AliceO2/blob/c8432bd40a03211b20849c9c6bba61ff8ecc8045/Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx#L318, i.e. simply `float derivfloat[deriv_size];`

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -240,7 +251,7 @@ bool ITSCalibrator::findThreshold(
 // x is the array of charge injected values;
 // NPoints is the length of both arrays.
 // thresh, noise, chi2 pointers are updated with results from the fit
-bool ITSCalibrator::findThresholdFit(
+bool ITSThresholdCalibrator::findThresholdFit(
```
shahor02, line 254: Note that this method is not thread-safe due to the minuit used in the fit.  To avoid a race condition which is very difficult to debug, I would suggest to put somewhere in the initialization check
```
if (mFitType==FIT && mNThreads>1) {
  throw std::runtime_error("Multuple threads are requested with fit method");
}
```

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -393,56 +406,60 @@ void ITSCalibrator::extractThresholdRow(const short int& chipID, const short int
...
         LOG(warning) << "Start-finding unsuccessful for chipID " << chipID
-                     << " row " << row << " column " << (col_i + 1) << '\n';
+                     << " row " << row << " column " << (col_i) << '\n';
       }
       continue;
     } catch (int* i) {
       if (this->mVerboseOutput) {
         LOG(warning) << "Start-finding unsuccessful for chipID " << chipID
-                     << " row " << row << " column " << (col_i + 1) << '\n';
+                     << " row " << row << " column " << (col_i) << '\n';
```
shahor02: Your findThreshold never throws an exception, what is the purpose of these catches. Also, you have 2 catches with the same action but you never use the catch argument. Usually, if one wants to catch `any` exception, one does 
`catch(...) { /* your processing code */ }  `

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
iravasen: yes actually not needed. Removed.

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
iravasen: good catch, done.

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
iravasen: done

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
iravasen: you are right. Actually now that I check in detail, we print already a warning in case success = false in each of the fit methods. So I removed try try-catch block that actually was doing nothing (as you said) and in case success = false we keep the info into the c-arrays and hence in the root tree filled in ```saveThreshold```

iravasen:
the error for the macOS build doesn't seem to be related to the modifications done in this PR.

shahor02 (approved):

PR #8024, opened 2022-01-27
DPL Analysis: add an error message on unbound index access

aalkin:


jgrosseo (dismissed):

Framework/Core/include/Framework/ASoA.h
```diff
@@ -1664,7 +1668,9 @@ constexpr auto is_binding_compatible_v()
     template <typename T>                                                                                                         \
     auto _Getter_##_as() const                                                                                                    \
     {                                                                                                                             \
-      assert(mBinding != nullptr);                                                                                                \
+      if (O2_BUILTIN_UNLIKELY(mBinding == nullptr)) {                                                                             \
+        throw o2::framework::runtime_error("Index to " #_Table_ " is not bound! Did you subscribe to the table?");                \
+      }                                                                                                                           \
       if (O2_BUILTIN_UNLIKELY(!has_##_Getter_())) {                                                                               \
```
ktf: We should probably hide all this is a separate, non templated, function.

ktf (dismissed):

ktf:
Looks like there is a conflict, actually...

Framework/Core/include/Framework/ASoA.h
aalkin: @ktf something like this?

Framework/Core/src/ASoA.cxx
```diff
@@ -110,4 +110,11 @@ arrow::ChunkedArray* getIndexFromLabel(arrow::Table* table, const char* label)
   return table->column(index[0]).get();
 }
 
+void checkBinding(void const* binding, const char* tableName)
+{
+  if (O2_BUILTIN_UNLIKELY(binding == nullptr)) {
```
ktf: The if statement (and only the `if`) should stay inlined, otherwise you will pay the cost of a function call.

aalkin: So only the `throw` should be extracted in a function? Would it have a benefit?

aalkin: Should I use `o2::framework::throw_error()` instead?

ktf: You benefit the fact that the string / stack unrolling is not embedded in every single instance of that function.

ktf (approved):

PR #8027, opened 2022-01-27
Ccdb callback

ktf:


ktf:
@shahor02 this now invokes the callback correctly.
@matthiasrichter apparently the callback registry needs some extra guidance to understand that `void*&` and `void*`. Would it be possible to at least have static_assert when it cannot find the callback?

ktf:
Relevant tests pass. Tested locally.

PR #8036, opened 2022-01-28
DCS proxies use now() in ms to fill DPH.creation time

shahor02:
@chiarazampolli


chiarazampolli:
Thanks @shahor02 ! And sorry for not having added it to my other PR.

PR #8037, opened 2022-01-28
Add SIGUSR1 signal handler to ShmManager

rbx:
When `SIGUSR1` signal is received ShmManager will perform a soft segment reset (without recreating).
Take special care not to reset content when messages are still in-flight or sockets are alive (that may contain messages). Otherwise these messages will point to buffers that no longer there.

For R3C-646.


davidrohr (approved):
seems to work so far, I have some more comments, which I will put on the JIRA ticket. merging, since this part is working.

PR #8038, opened 2022-01-28
Revert "Improving sensitive hit creation for FT0 in Detector.cxx"

sawenzel:
Reverts AliceO2Group/AliceO2#7961

Broke FT0 hit creation with Geant4.


mpuccio:
Thanks a lot for the prompt reaction! I confirm that with the PR reverted we have again the hits in FT0 with G4.
Just to give context: we noticed this because the event selection task on AO2D rejects all the events with the current dev and all the nightlies from the 21st of January.
Thanks again!

PR #8041, opened 2022-01-30
Introducing McParticle version 001

jgrosseo:
Default version still McParticle version 000


jgrosseo:
This includes https://github.com/AliceO2Group/AliceO2/pull/8002

jgrosseo:
Was green before clang fixes. Merging.

PR #8043, opened 2022-01-31
MCH: implement handling of SAMPA reset at each TF

aferrero2707:
The new code introduces a new method to compute the time of the MCH digits, which assumes a reset of the internal SAMPA bunch-crossing counters at the beginning of each Time Frame.
This new method does not rely on the decoding of Heart-Beat data packets in order to extract the reference SAMPA time at the TF start, hence it is more reliable in case of high data rates.

The choice of the method is driven by a new "--time-reco-mode" command-line option of the MCH decoder. Possible values are:
"hbpackets": the SAMPA time reference is computed from the time stamp of the Heart-Beat packets (old method)
"bcreset": a reset of the SAMPA bc counters is assumed at the beginning of each time frame (new method)

### DIFF
--- a/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
+++ b/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/DataDecoder.h
@@ -58,6 +58,11 @@ class DataDecoder
   static constexpr int32_t tfTimeMax{0x7FFFFFFF};
   static constexpr int32_t tfTimeInvalid{-tfTimeMax};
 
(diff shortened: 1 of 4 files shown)


aphecetche (approved):

aferrero2707:
There is one failing CI job, with errors like
```
[ERROR] Exception caught: Couldn't get TTree "DF_1/O2timestamps" from "/sw/BUILD/baefc713d5a57d5c5be0ad239a540f35f954008a/O2-full-system-test/sim-challenge/AO2D.root"
```
that are completely unrelated with this PR...

PR #8045, opened 2022-01-31
switch to disable fall-back to TGeo if MatLUT is missing

shahor02:
@jgrosseo @victor-gonzalez 
With this PR, `o2::base::Propagator::Instance()->setTGeoFallBackAllowed(false);` will throw an exception if the matLUT is requested but not attached


victor-gonzalez:
Thanks for doing this @shahor02 
In the case of the `tracksextended` task we will keep the default behavior (fall-back to `TGeo`) because we will anyway get an exception when trying to `o2::base::MatLayerCylSet::rectifyPtrFromFile()` passing a null pointer when `LUT` is not gotten from `CCDB`

PR #8046, opened 2022-01-31
Update analyses in sim_challange.sh

benedikt-voelkel:


benedikt-voelkel:
This PR should help us to spot issues at CI-level which might lead to failing analyses (at least on sim data). The same analyses run during the alibi nightlies are tested.

sawenzel:
only affecting full-CI (not compiled code or other tests), merging

PR #8048, opened 2022-01-31
McParticle: moving default to 001

jgrosseo:


jgrosseo:
@nburmaso I have removed the duplication of the table definition. Do I overlook something or is it not needed actually?
https://github.com/AliceO2Group/AliceO2/pull/8048/files#diff-d43bf1e210cd8966890634920e751cd6801d9125fe5c48e4268115b9dc211115L154

jgrosseo:
This PR depends on https://github.com/AliceO2Group/O2Physics/pull/488

nburmaso:
> @nburmaso I have removed the duplication of the table definition. Do I overlook something or is it not needed actually? https://github.com/AliceO2Group/AliceO2/pull/8048/files#diff-d43bf1e210cd8966890634920e751cd6801d9125fe5c48e4268115b9dc211115L154

Hi @jgrosseo. I implemented a separate table for MC particles so that I had a table cursor without dynamic columns in it. At least it was needed at the moment when I first added MC particles to the converter. I guess `MCParticlesTable` can be removed now if it works fine with the `StoredMcParticles_001`

sawenzel:
AliceO2Group/O2Physics#488 breaks the analysis testing which is done inside the full system test. This is unfortunately still the case with the PR here.

I see in the fullCI log:
```
++ grep 'Return status' sim-challenge.log
Return status of Efficiency: 1
Return status of EventTrackQA: 1
```
where the first seems to be a simple transition to the new table on the O2Physics side. The second one is a segfault.

jgrosseo:
> [AliceO2Group/O2Physics#488](https://github.com/AliceO2Group/O2Physics/pull/488) breaks the analysis testing which is done inside the full system test. This is unfortunately still the case with the PR here.
> 
> I see in the fullCI log:
> 
> ```
> ++ grep 'Return status' sim-challenge.log
> Return status of Efficiency: 1
> Return status of EventTrackQA: 1
> ```
> 
> where the first seems to be a simple transition to the new table on the O2Physics side. The second one is a segfault.

Do you know what needs to be fixed?

sawenzel (approved):
Ok let's merge it. FullCI is back to normal in a local test.

sawenzel:
Remaining issues followed up here: https://alice.its.cern.ch/jira/browse/O2-2766

jgrosseo:
Thanks a lot!

jgrosseo:
I have asked @ktf for a daily tag. Then we can merge https://github.com/AliceO2Group/O2Physics/pull/491 and all tests should be back in business.

PR #8049, opened 2022-01-31
DPL: improved Conditions support

ktf:


ktf:
@chiarazampolli @shahor02 using `ccdbParamSpec(std::string const& path, bool runDependent)` with `runDependent = true` should allow you to fetch runDependent objects using the run number provided in the DataTakingContext.

PR #8051, opened 2022-01-31
[EMCAL-751] Temporal fix for the digitizer

hahassan7:
Temporal fix for the digitizer by disabling the "DigitsRighOutBuffer" and adding normal "std::unordered_map" instead.


sawenzel:
What issue is this fixing? Is it still related to https://alice.its.cern.ch/jira/browse/EMCAL-751 ? (If yes, did https://github.com/AliceO2Group/AliceO2/pull/8006 not resolve the issue completely?)

hahassan7:
Hi @sawenzel,
That old pull request was a temporal fix not complete. In the old one, the pre-trigger collisions are dealt with properly. But in this pull request, all the digits are being dealt with properly and results makes more sense.
By the way, this is also temporal fix (since we are planning to update the DigitsWriteoutBuffer).

mfasDa (approved):

PR #8052, opened 2022-02-01
[MFT] Fix digi2raw output segmentation

rpezzi:
This PR adapts the MFT MC to raw converter to produce one file per CRU endpoint. `o2-mft-digi2raw --file-for cru` produces the following raw files:

```
MFT_alio2-cr1-flp182_cru567_0.raw
MFT_alio2-cr1-flp182_cru567_1.raw
MFT_alio2-cr1-flp182_cru570_0.raw
MFT_alio2-cr1-flp182_cru570_1.raw
MFT_alio2-cr1-flp183_cru548_0.raw
MFT_alio2-cr1-flp183_cru548_1.raw
MFT_alio2-cr1-flp183_cru554_0.raw
MFT_alio2-cr1-flp183_cru554_1.raw
MFT_alio2-cr1-flp184_cru543_0.raw
MFT_alio2-cr1-flp184_cru543_1.raw
MFT_alio2-cr1-flp184_cru569_0.raw
MFT_alio2-cr1-flp184_cru569_1.raw
MFT_alio2-cr1-flp185_cru541_0.raw
MFT_alio2-cr1-flp185_cru541_1.raw
MFT_alio2-cr1-flp185_cru552_0.raw
MFT_alio2-cr1-flp185_cru552_1.raw
MFT_alio2-cr1-flp186_cru542_0.raw
MFT_alio2-cr1-flp186_cru542_1.raw
MFT_alio2-cr1-flp186_cru547_0.raw
MFT_alio2-cr1-flp186_cru547_1.raw
```

### DIFF
--- a/Detectors/ITSMFT/MFT/simulation/src/digi2raw.cxx
+++ b/Detectors/ITSMFT/MFT/simulation/src/digi2raw.cxx
@@ -186,111 +186,143 @@ void digi2raw(std::string_view inpName, std::string_view outDir, std::string_vie
   swTot.Print();
 }
 
+struct MFTRUMapping {
+  std::string flp{};
+  int feeID;
+  int cruHWID = 0;
+  int idInCRU = 0;
+  int endpoint = 0;
+};
+
+// FLP_Name, FEEId, CRU_HW, idInCRU, endpoint
+const MFTRUMapping mftHWMap[o2::itsmft::ChipMappingMFT::getNRUs()] = {
+  {"alio2-cr1-flp182", 0, 570, 0, 0},
+  {"alio2-cr1-flp182", 1, 570, 1, 0},
+  {"alio2-cr1-flp182", 2, 570, 2, 0},
+  {"alio2-cr1-flp182", 3, 570, 3, 0},
+  {"alio2-cr1-flp186", 64, 542, 4, 0},
+  {"alio2-cr1-flp186", 65, 542, 5, 0},
+  {"alio2-cr1-flp186", 66, 542, 6, 0},
+  {"alio2-cr1-flp186", 67, 542, 7, 0},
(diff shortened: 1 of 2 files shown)


shahor02:
Hi @sawenzel, @davidrohr ,
 
If I interpret the fullCI log correctly, it has failed due to the analysis test returning 1 instead of 1, which seems to be something recently added:
```
++ SIM_CHALLENGE_ANATESTING=ON
++ /sw/slc8_x86-64/O2/8052-slc8_x86-64-local1/prodtests/sim_challenge.sh
++ grep 'Return status' sim-challenge.log
++ grep -v ': 0'
Return status of MCHistograms: 1
++ false
```

benedikt-voelkel:
Hej @shahor02 
The analysis testing was not working properly actually and we fixed that for the first in https://github.com/AliceO2Group/AliceO2/pull/8046. After that, the introduction of the versioned MCParticle table caused a failure of the analysis tests in the fullCI. @jgrosseo fixed everything related to that and it was tested again (with 2 other analyses failing for related reasons).
In https://github.com/AliceO2Group/AliceO2/pull/8054 we see that all analyses pass now. So also here it should be green when tested again I reckon.
I will propagate this info tomorrow to one or two appropriate mattermost channels.

For clarification: The expected return value is `0` in which case no output is printed.

davidrohr:
I don't really understand how this change in digi2raw can break the sim challenge.

davidrohr:
OK, thx, that explains it, will wait for the CI to rerun and then merge.

shahor02:
Hi @benedikt-voelkel 
thanks for the info (I had typo in my comment, should have been `failed due to the analysis test returning 1 instead of 0`). I see that the test finished in the morning, let's see if at next test it gets green.

davidrohr (approved):
tested, and output looks good.

PR #8053, opened 2022-02-01
[QC-725] Allow to store TRFCollections as CSV files

knopers8:
Which is needed for the integration with RCT


Barthelemy:
Do I understand correctly that the interface with RCT is via CSV files ? just asking, I am fine with it. Perhaps it could be explained in the doc  somewhere.

Barthelemy (approved):

knopers8:
> Do I understand correctly that the interface with RCT is via CSV files ? just asking, I am fine with it. Perhaps it could be explained in the doc somewhere.

Yes, I will add some doc in the next iteration.

PR #8054, opened 2022-02-01
Re-introduce analyses

benedikt-voelkel:


benedikt-voelkel:
After fixes made in https://github.com/AliceO2Group/O2Physics/pull/491, previously failing analyses can be put back to `sim_challange.sh`

PR #8057, opened 2022-02-01
[MFT] Add timers to tracker workflow

rpezzi:
Add separate stopwatches for
- data preparation
- track finding
- track fitting
- MC label computation


shahor02 (approved):
fullCI failure is unrelated

PR #8060, opened 2022-02-02
update ShmManager::ResetContent to be able to reset after a crash

rbx:
Adapt ShmManager to recreate relevant shmem management segment contents after a "soft" shmem contents reset. Requires FairMQ 1.4.48.


ktf:
Restarted with the latest alidist.

PR #8064, opened 2022-02-02
DPL: timeout on STOP transition

ktf:


ktf:
@davidrohr @shahor02 this works for me on a local test. The `--exit-transition-timeout` option can be used to set a delay in seconds. If we get the end-of-stream message the timeout is ignored and we exit immediately.

ktf:
Tested on hyperloop. No side effects unless the option is provided on the command line.

PR #8065, opened 2022-02-02
Limit maximum memory usage during tracking to 32GB (tunable)

mpuccio:
Hi @davidrohr,

this should do what you asked, even if it is ugly.
I am still testing locally, as soon as it is ready I will remove the draft status.

Cheers,
Max﻿


shahor02:
Hi @mpuccio 
shall I merge this?

PR #8069, opened 2022-02-03
Truncate digitcontext output

sawenzel:
With QED this may be very large otherwise.
The full output can still be obtained by
inspection of the context file.


sawenzel:
@davidrohr : This should solve your issue/request.

davidrohr:
thx!

PR #8073, opened 2022-02-03
Use Cluster class for MID instead of Cluster2D and Cluster3D

dstocco:
I used two separate classes for the clusters in local coordinates in the detection element and the final ones in global coordinates.
This choice however introduces a level of complication which is not really needed.
With this PR, I go back to 1 single object for clusters.


Detectors/MUON/MID/Clustering/src/Clusterizer.cxx
```diff
@@ -200,22 +183,27 @@ bool Clusterizer::init(std::function<void(size_t, size_t)> func)
 }
 
 //______________________________________________________________________________
-void Clusterizer::makeCluster(const MpArea& areaBP, const MpArea& areaNBP, const int& icolumn, const int& deIndex)
+void Clusterizer::makeCluster(const MpArea& areaBP, const MpArea& areaNBP, uint8_t deId)
+{
+  double sqrt12 = 3.4641016;
```
aphecetche: I assume this is an optimisation backed-up by some measurement ? ;-) 
Anyway, while at it, I would constexpr it

aphecetche (dismissed):

aphecetche (approved):

dstocco:
Since this PR changes the MID dataformats, I think that @shahor02 review is mandatory. Notice that the build errors are unrelated (everything was green last week...)

shahor02 (approved):
Hi @dstocco 
as far as I can see, outside of the MID-specific code, only the RecoContainer is affected and there the change is straightforward, so everything looks fine.
Some CI's don't run, afaik, due to the gitlab problems, I would wait with merging until they are recovered.

dstocco:
Dear @ktf , it seems that the CI is failing with no errors on macOS and with omp errors in fullCI. Is it a known issue?

ktf:
There were indeed gitlab issues which could have resulted in errors without log. Now everything is ok and all the tests are green.

PR #8080, opened 2022-02-04
DCS filepush server emulator + documentation

shahor02:


shahor02:
nothing to test as the code should be compiled outside of the O2, merging

PR #8082, opened 2022-02-04
Add runType to GRPECS + its creator

shahor02:
Also, change the way the EOR time is registered: at SOR the ECS must call the executable
with EOR time set to 0 (default), which will be recorded in the GRPECS object but the object
will be stored with validity SOR_time : SOR_time + 2days (to be queriable for the online
processing, assuming longest run duration of 2 days).
o2-ecs-grp-create -r <run> -p <period> -s <SOR_TIME>

Then, at the EOR, the ECS should call the creator with the same options + EOR_time:
o2-ecs-grp-create -r <run> -p <period> -s <SOR_TIME> -e <EOR_TIME>
which will update the object with validity max(SOR_time + 2days, EOR_time+1h).

### DIFF
--- a/Detectors/GRP/workflows/src/create-grp-ecs.cxx
+++ b/Detectors/GRP/workflows/src/create-grp-ecs.cxx
@@ -102,7 +102,7 @@ int main(int argc, char** argv)
     add_option("help,h", "Print this help message");
     add_option("period,p", bpo::value<std::string>(), "data taking period");
     add_option("run,r", bpo::value<int>(), "run number");
-    add_option("run-type,t", bpo::value<int>->default_value(int(GRPECSObject::RunType::NONE)), "run type");
+    add_option("run-type,t", bpo::value<int>()->default_value(int(GRPECSObject::RunType::NONE)), "run type");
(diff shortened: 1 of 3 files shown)


knopers8 (dismissed):

Detectors/GRP/workflows/src/create-grp-ecs.cxx
```diff
@@ -96,6 +102,7 @@ int main(int argc, char** argv)
     add_option("help,h", "Print this help message");
     add_option("period,p", bpo::value<std::string>(), "data taking period");
     add_option("run,r", bpo::value<int>(), "run number");
+    add_option("run-type,t", bpo::value<int>(int(GRPECSObject::RunType::NONE)), "run type");
```
knopers8: `-t` is also for `triggering`, maybe you could use a different letter?

shahor02: thanks for spotting, fixed

shahor02:
fullCI failure is unrelated.

PR #8084, opened 2022-02-05
Fix for DCS emulator compilation

shahor02:


shahor02:
Only the script was modified, merging

PR #8087, opened 2022-02-07
GlobalFwd: fix Pairing Efficiency calculation

rpezzi:
- Use correct histogram to evaluate Pairing Efficiency
- Meaningful names
- Add configuration to scan matching score cuts

### DIFF
--- a/Detectors/GlobalTracking/src/MatchGlobalFwdAssessment.cxx
+++ b/Detectors/GlobalTracking/src/MatchGlobalFwdAssessment.cxx
@@ -605,18 +605,16 @@ void GloFwdAssessment::finalizePurityAndEff()
   auto minBin = Reco->GetYaxis()->FindBin(2.4);
   auto midBin = Reco->GetYaxis()->FindBin(3.0);
   auto maxBin = Reco->GetYaxis()->FindBin(3.6);
-  auto PairablePtProjInner = (TH1*)hTrue->ProjectionX("PairableInner", midBin, maxBin);
-  auto PairablePtProjOuter = (TH1*)hTrue->ProjectionX("PairableOuter", minBin, midBin);
+  auto PairablePtProjInner = (TH1*)hPairable->ProjectionX("PairableInner", midBin, maxBin);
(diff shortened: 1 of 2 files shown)


Detectors/GlobalTracking/src/MatchGlobalFwdAssessment.cxx
```diff
@@ -626,48 +624,48 @@ void GloFwdAssessment::finalizePurityAndEff()
...
+    hPurityOuter->SetMaximum(1.2);
+
+    auto& hPairingEffOuter = mPairingPtOuterVecTH1.emplace_back((std::unique_ptr<TH1D>)static_cast<TH1D*>(RecoPtProjOuter->Clone()));
+    hPairingEffOuter->Divide(PairablePtProjInner); // Pairing Efficiency = N_reco / N_Pairable
+    hPairingEffOuter->SetNameTitle(Form("GMTrackPairingEffOuterPtCut_%.2f", scoreCut), Form("%.2f cut", scoreCut));
+    hPairingEffOuter->GetYaxis()->SetTitle("Pairing Efficiency [ N_{Rec} / N_{pairable}]");
+    hPairingEffOuter->SetOption("COLZ");
+    hPairingEffOuter->SetMarkerStyle(kFullTriangleUp);
+    hPairingEffOuter->SetMinimum(0.0);
+    hPairingEffOuter->SetMaximum(1.8);
```
shahor02, line 665: For the record: the projections like RecoPtProjOuter are all leaking. You should delete this which are not needed (or make them unique_ptrs). BTW, instead of calling `Clone` which is extremely expensive call, you could do directly
```
auto& hPairingEffOuter = mPairingPtOuterVecTH1.emplace_back();
hPairingEffOuter.reset(RecoPtProjOuter);
```

rpezzi: Hi @shahor02 , thanks for the input. This part of the code is not executed in production, so it will not cause problems. Given that @RenEjima is ready to open a PR that depends on this one, may I ask to have this merged and fix memory leak and optimization in a separate PR?

shahor02 (approved):

PR #8090, opened 2022-02-08
Timestamp propagation and use in digitizer workflow

sawenzel:
The digitizer workflow is now calculating the timeframe's timestamp
based on the HBF parameters and is propagating this value through DPL
header properties. This is done is the same way as used in async
reconstruction.
To anchor digitization to a particular timestamp it suffices to set
`HBFUtils.startTime=xxx`.

Additional changes to use consistent timing in digitization:
* take out timestamp setting on BasicCCDBManager for CPV
  (the timing is set anyways by the master digitizer workflow)
* use new mechanism to retrieve and use timestamp in TRD


martenole (approved):
Changes for the TRD look good

bazinski (approved):

ktf (approved):

PR #8092, opened 2022-02-08
[O2-2778] Change the way we favor CONFIG variant of TBB find_package

aphecetche:


sawenzel (approved):
Looks ok to me. But I have no setup to test if a system TBB would make it in...

davidrohr:
I can test it tonight ...

davidrohr (approved):
fullCI error is unrelated.
Tried locally with system TBB and works correctly

PR #8095, opened 2022-02-08
Do not discard decoded chip data if decoding error was set

shahor02:
Also, preserve the sorted fired MFT chips after filling the digits, to be used for the clustering if both digits and clusters were requested from the decoder.


shahor02:
fullCI failure is unrelated, merging

PR #8105, opened 2022-02-09
[EMCAL-757] Fix return type in cell-reader-workflow

mfasDa:


shahor02 (approved):
fullCI failure is unrelated

PR #8110, opened 2022-02-10
DPL: properly handle pollers on start-stop-start transition (O2-2751)

ktf:


ktf:
This properly reinitialises socket pollers when the transition does not go through INIT again. In my local test this makes the o2-raw-proxy behave correctly.

ktf:
@matthiasrichter could you check this as well? I think it fixes O2-2751

ktf:
Merging this to get a build including it to try on the FLP. @vascobarroso

PR #8112, opened 2022-02-10
Update sim_challenge.sh

sawenzel:
take out broken analysis


sawenzel:
The full-CI should hopefully come back

PR #8115, opened 2022-02-10
TRD make fixsm1617 option always true

bazinski:
Make the fix for 16 and 17 always in.
Due to the requirement of TracklHCHeader of 2 its actually not always in.

This is in reference to jira O2-2767


martenole:
Thanks Sean, but would it not make more sense to simply change the default here?
https://github.com/AliceO2Group/AliceO2/blob/90277fe58abf89c3cc0b15ec1b9bcc0115442426/Detectors/TRD/reconstruction/src/DataReader.cxx#L55

bazinski:
Hi Ole, problem with that is then if you use the option --fixsm1617 you are actually unfixing it.
I actually meant to take the option out completely.

martenole:
Ah true. Ok then I would either take it out completely or rename the option to `no-fix-sm1617` or similar

PR #8117, opened 2022-02-10
CPV noise calibration

sevdokim:
Hello!

CPV NoiseCalibrator based on TimeSlot calibration is created. Also new CCDB DPL service is introduced to RawToDigitsConverterSpec.


shahor02:
@sevdokim the CI errors are genuine

sevdokim:
Dear experts,

I have funny error message from build/AliceO2/O2/o2/macOS
```
/System/Volumes/Data/build/ci_checks/alice_ci_0/o2/sw/SOURCES/O2/8117/0/Detectors/CPV/workflow/src/RawToDigitConverterSpec.cxx:87:65: error: object backing the pointer will be destroyed at the end of the full-expression [-Werror,-Wdangling-gsl]
  const o2::cpv::Pedestals* pedestals = mIsPedestalData ? 0x0 : (ctx.inputs().get<o2::cpv::Pedestals*>("peds")).get();
```
Any suggestions how to avoid this error? Strange that on other systems there is no such error.

shahor02:
@sevdokim, I think it means that the uniuq_ptr-like object returned by the `pc.inputs().get` is destroyed on the same line.
Can you try to assign it 1st, e.g.
```
auto pedPtr = ctx.inputs().get<o2::cpv::Pedestals*>("peds");
const o2::cpv::Pedestals* pedestals = nullptr;
if (mIsPedestalData) {
  pedestals = pedPtr.get();
}
```
Note that you have to use `nullptr` instead of `0`. Also, please check the logics, should the assignment to `pedestals` be done if `mIsPedestalData` null or in the opposite case?

sevdokim:
Hi @shahor02 

 The logic is correct: if mIsPedestalData then we do not use pedestals from CCDB. 
 The problem is that I cannot call ctx.inputs().get<o2::cpv::Pedestals*>("peds"); when it is pedestal run because corresponding input is not connected. Probably problem can be easily solved by connecting inputs.emplace_back("peds", "CPV", "CPV_Pedestals", 0, o2::framework::Lifetime::Condition, o2::framework::ccdbParamSpec("CPV/Calib/Pedestals")); always but not only when it is needed.

shahor02:
@sevdokim , can you try with e.g.
```
std::decay_t<decltype(ctx.inputs().get<o2::cpv::Pedestals*>("peds"))> pedPtr{};
const o2::cpv::Pedestals* pedestals = nullptr;
if (!mIsPedestalData) {
  pedPtr = ctx.inputs().get<o2::cpv::Pedestals*>("peds");
  pedestals = pedPtr.get();
}
```

sevdokim:
Oups. I did something which I didn't meant to do.

PR #8118, opened 2022-02-10
[EMCAL-760] Stand-alone publisher for digits

mfasDa:
Trivial stand-alone publisher for digits, based on the
publisher in the RecoWorkflow. Needed for the QC
in simulation.

### DIFF
--- a/Detectors/EMCAL/workflow/src/digit-reader-workflow.cxx
+++ b/Detectors/EMCAL/workflow/src/digit-reader-workflow.cxx
@@ -13,7 +13,7 @@
 #include <vector>
 #include "Framework/Variant.h"
 #include "Framework/ConfigParamSpec.h"
-#include "DataFormatsEMCAL/Cell.h"
+#include "DataFormatsEMCAL/Digit.h"
 #include "DetectorsRaw/HBFUtilsInitializer.h"
 #include "EMCALWorkflow/PublisherSpec.h"
 #include "CommonUtils/ConfigurableParam.h"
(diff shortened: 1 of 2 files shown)


shahor02 (commented):
@mfasDa I would suggest to add the HBFUtilsInitializer lines which allows to provide consistent firstTForbit and tf creation time both in the MC and real data digits injection. Just grep for this word, you will see the line to add.

mfasDa:
Hi @shahor02, thanks for pointing out! I took it from the EMCAL reco workflow - it should be done now (also for the cell reader workflow.

Detectors/EMCAL/workflow/src/digit-reader-workflow.cxx
```diff
@@ -0,0 +1,56 @@
...
+#include "CommonUtils/ConfigurableParam.h"
+
+using namespace o2::framework;
+using namespace o2::emcal;
+
+// we need to add workflow options before including Framework/runDataProcessing
+void customize(std::vector<ConfigParamSpec>& workflowOptions)
+{
+  std::vector<ConfigParamSpec> options{{"disable-mc", VariantType::Bool, false, {"Do not propagate MC labels"}},
+                                       {"configKeyValues", VariantType::String, "", {"Semicolon separated key=value strings"}}};
```
shahor02, line 28: You missed here `o2::raw::HBFUtilsInitializer::addConfigOption(options);`

mfasDa: Done, thanks for spotting!

mfasDa:
@shahor02 Do you think it can be merged now?

shahor02 (approved):

PR #8119, opened 2022-02-11
ctpdev: ctpproxy publish CTP scalers

lietava:


DataFormats/Detectors/CTP/src/Scalers.cxx
```diff
@@ -361,3 +361,8 @@ int CTPRunScalers::updateOverflows(const CTPScalerRaw& scal0, const CTPScalerRaw
   //std::cout << std::endl;
   return 0;
 }
+int CTPRunScalers::parseZMQScalers(std::string zmqscalers)
+{
+  std::vector<std::string> tokens = o2::utils::Str::tokenize(zmqscalers, ' ');
```
shahor02, line 366: this method does nothing, just creates/destroys a local vector. Is this intended?

lietava: yes, sorry about that. It is not finished. I wanted to pull to avoid too much divergence from dev.

shahor02 (approved):

PR #8122, opened 2022-02-11
Merge the MC and the data Specs for MID reconstruction

dstocco:
This simplifies the maintenance. 
Also, match all digits subSpecs in clusterizer. This allows to solve an issue when running the entropy encoder and the MID reconstruction at the same time.
The decoder indeed sends 3 subspecs. The entropy encoder matches all of them with a wildcard, but the reconstruction workflow only matched one of them.
This resulted in the entropy encoder to also match the leftover specs from the reconstruction workflow, thus resulting in a double processing.
This PR solves the issue.


aphecetche (approved):

dstocco:
@ktf : the issue with the fullCI seems to be unrelated with this PR. Can this be merged?

ktf:
@shahor02 @TimoWilken did we turn on alien usage or is it just automatic? I think the alien token is still missing?
@costing I understood you cached the missing objects? Are we still missing some or it's just due to the alien usage?

PR #8125, opened 2022-02-11
Use standard MID digits instead of dedicated digits for MC

dstocco:
Originally, the ColumnData were kept as independent as possible from root. Therefore a special ColumnDataMC was used for MC, where the independence was not enforces. ColumnDataMC derived from ColumnData, but in the end, basically no difference between the two classes existed.
Since the advantage of having different digits type for MC and data seems to be much smaller than the obvious disadvantage, the ColumnDataMC is removed and ColumnData is used everywhere.


sawenzel (approved):
Fine for me if @aphecetche approves.

aphecetche (approved):

PR #8127, opened 2022-02-11
DPL: make sure Condition<> object can be used

ktf:


ktf:
Obviously wrong and not actually used. Merging.

PR #8132, opened 2022-02-11
CPV noise calibration

sevdokim:
Hello! 
CPV NoiseCalibrator and CCDB DPL service for Raw2DigitsConverterSpec is implemented in this PR.


kharlov (dismissed):

Detectors/CPV/calib/testWorkflow/NoiseCalibratorSpec.h
```diff
@@ -0,0 +1,165 @@
...
+    auto tfcounter = o2::header::get<o2::framework::DataProcessingHeader*>(pc.inputs().get("digits").header)->startTime;
+    auto&& digits = pc.inputs().get<gsl::span<o2::cpv::Digit>>("digits");
+    auto&& trigrecs = pc.inputs().get<gsl::span<o2::cpv::TriggerRecord>>("trigrecs");
+
+    // read pedestal efficiencies, dead and high ped channels from pedestal run
+    // do it only once as they don't change during noise scan
+    if (!mCalibrator->mPedEfficiencies) {
+      const auto pedEffs = o2::framework::DataRefUtils::as<CCDBSerialized<std::vector<float>>>(pc.inputs().get("pedeffs"));
+      if (pedEffs) {
+        mCalibrator->mPedEfficiencies = new std::vector<float>(pedEffs->begin(), pedEffs->end());
```
shahor02: I see you are creating a local copy of the CCDB objects. The method I've suggested you did not work?

shahor02 (commented):
Hi @sevdokim, the fullCI fails since the CPV does not get its CCDB object due to the absence of the alien-token on the build server. The PR itself is fine, but if we merge it now, it will keep breaking the fullCI. 
Is it urgent for you to have it merged? If not, I would put it on hold until the alien-token is available in the CI, hopefully, very soon.

Detectors/CPV/calib/testWorkflow/NoiseCalibratorSpec.h
sevdokim: Hi @shahor02 
Your method is working very fine, thank you! I use it in Raw2DigitsConverterSpec. 
In this noise calibration I decided to make local copy in very beginning of calibration only once and don't ask for this objects in next TFs as they will never change during noise scan. Is it ok?

sevdokim:
Hi @shahor02 
Yes, let's wait until fullCI has alien token in there. This PR is not urgent.

Detectors/CPV/calib/src/NoiseCalibrator.cxx
```diff
@@ -0,0 +1,150 @@
...
+  mBadChannelMapVec.clear();
+}
+//_____________________________________________________________________________
+void NoiseCalibrator::finalizeSlot(NoiseTimeSlot& slot)
+{
+  NoiseCalibData* calibData = slot.getContainer();
+  LOG(info) << "NoiseCalibrator::finalizeSlot() : finalizing slot "
+            << slot.getTFStart() << " <= TF <= " << slot.getTFEnd() << " with " << calibData->mNEvents << " events.";
+  o2::cpv::BadChannelMap* badMap = new o2::cpv::BadChannelMap();
+  bool* badMapBool = new bool[Geometry::kNCHANNELS];
```
shahor02: Why don't you use `boll badMapBool[Geometry::kNCHANNELS]={};` ? In any case, if you use `new`, you should also use `delete[]`.

Detectors/CPV/calib/include/CPVCalibration/NoiseCalibrator.h
```diff
@@ -0,0 +1,77 @@
...
+    LOG(info) << "hasEnoughData() is being called";
+    return slot.getContainer()->mNEvents >= mMinEvents;
+  }
+  void initOutput() final;
+  void finalizeSlot(NoiseTimeSlot& slot) final;
+  NoiseTimeSlot& emplaceNewSlot(bool front, uint64_t tstart, uint64_t tend) final;
+
+  std::vector<float>* mPedEfficiencies = nullptr;
+  std::vector<int>* mDeadChannels = nullptr;
+  std::vector<int>* mHighPedChannels = nullptr;
```
shahor02: if you are using a class (rather than a struct) then move the datamembers to `protected:` part and have getters/setters. Also, you are assigning a to these data members arrays created with `new`, but then you never call `delete[]`. The best would be to use simply vectors (instead of pointers) or use e.g. `std::unique_ptr<std::vector<float>> mPedEfficiencies`, which you can set by external raw vector pointer by `mPedEfficiencies.reset(ptr)`.

Detectors/CPV/calib/testWorkflow/NoiseCalibratorSpec.h
shahor02: No problem to create them once, but the comment above about creating a vector with `new` and not deleting it afterwards.

shahor02 (commented):
Hi @sevdokim 
OK, then let's put it on hold until the token problem is solved. Meanwhile, I had a closer look on your code, see few comments below.

Detectors/CPV/calib/include/CPVCalibration/NoiseCalibrator.h
sevdokim: Hi @shahor02 
you are absolutely right, now I use unique_ptr for those objects. Can you please check again?

Detectors/CPV/calib/src/NoiseCalibrator.cxx
sevdokim: Not deleting array is clearly my mistake, thanks for pointing this out.

Detectors/CPV/calib/src/NoiseCalibrator.cxx
```diff
@@ -0,0 +1,150 @@
...
+  bool badMapBool[Geometry::kNCHANNELS] = {};
+
+  // handle data from pedestal run first
+  // check pedestal efficiencies
+  if (mPedEfficiencies) {
+    LOG(info) << "NoiseCalibrator::finalizeSlot() : checking ped efficiencies";
+    for (int i = 0; i < Geometry::kNCHANNELS; i++) {
+      badMapBool[i] = false;
+      if (mPedEfficiencies.get()->at(i) > mToleratedChannelEfficiencyHigh ||
+          mPedEfficiencies.get()->at(i) < mToleratedChannelEfficiencyLow) {
```
shahor02: do you really need to use `at(i)` which does a range check? `[i]` is faster.

Detectors/CPV/calib/include/CPVCalibration/PedestalCalibrator.h
```diff
@@ -87,14 +87,14 @@ class PedestalCalibrator final : public o2::calibration::TimeSlotCalibration<o2:
   std::vector<o2::ccdb::CcdbObjectInfo> getCcdbInfoEfficienciesVector() { return mCcdbInfoPedEfficienciesVec; }
   std::vector<std::vector<float>> getEfficienciesVector() { return mPedEfficienciesVec; }
```
shahor02: this returns a copy of the vectors. Is this what you want? Usually it is enough to do e.g.
```
const std::vector<o2::ccdb::CcdbObjectInfo>&  getCcdbInfoEfficienciesVector() const { return mCcdbInfoPedEfficienciesVec; }
```

Detectors/CPV/calib/src/NoiseCalibrator.cxx
sevdokim: That I didn't know. Thanks for pointing!

Detectors/CPV/calib/include/CPVCalibration/PedestalCalibrator.h
sevdokim: Let me try to use const std::vector<o2::ccdb::CcdbObjectInfo>&

sevdokim: So the answer is following: CcdbInfoVector cannot be const as it may be changed by createObjectImage() function.

shahor02: ok, then you can do 
```
std::vector<o2::ccdb::CcdbObjectInfo>&  getCcdbInfoEfficienciesVector() { return mCcdbInfoPedEfficienciesVec; }
```
my point was why do you need return a copy.

sevdokim: Hi @shahor02 
yes, I implemented return values as references. CCDBinfo as normal references and CCDB objects as const references. Just for my understanding, what is approximate time for alien token to be imlemented into fullCI?

shahor02: Hi @sevdokim, unfortunately, I don't know, it is followed here: https://alice.its.cern.ch/jira/browse/O2-2760

chiarazampolli:
Hi @shahor02 , 

Do we need to wait for the CI to be fixed (token issue) to merge? We would like to test this calibration.

Chiara

shahor02:
Will merge after local FST check

shahor02:
The local FST fails with 
```
[536108:internal-dpl-ccdb-backend]: [00:19:12][INFO] Fetching object for route  { CPV/CPV_Pedestals/0}
[536108:internal-dpl-ccdb-backend]: [00:19:12][ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch/CPV/Calib/Pedestals/1645990691385/runNumber=0/
[536108:internal-dpl-ccdb-backend]: [00:19:12][INFO] Fetching object for route  { CPV/CPV_Gains/0}
[536108:internal-dpl-ccdb-backend]: [00:19:12][ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch/CPV/Calib/Gains/1645990691385/runNumber=0/
[536108:internal-dpl-ccdb-backend]: [00:19:12][INFO] Fetching object for route  { CPV/CPV_BadMap/0}
[536108:internal-dpl-ccdb-backend]: [00:19:12][ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch/CPV/Calib/BadChannelMap/1645990691385/runNumber=0/
```
@ktf, any idea why the runNumber tag is used? These objects are requested with default, run-independent ccdbParamSpec spec:
https://github.com/AliceO2Group/AliceO2/blob/3ff35a831f140f67b6add0c08281eca00fb518d9/Detectors/CPV/workflow/src/RawToDigitConverterSpec.cxx#L241-L249

ktf:
I will check.

ktf:
#8242 should fix the issue.

shahor02 (approved):
After fixing the CCDBFetcher the local FST passed, merging.

PR #8133, opened 2022-02-11
Avoid quitting readout-proxy on a restart

ktf:


ktf:
Of course this does not work if it needs to quit on his own...

matthiasrichter (commented):
I think this s necessary for the termination of workflows started with the DPL driver, so that  workflows in the shell automatically terminate.

Isn't that control command only interpreted by the main process? Which is not there if controlled by AliECS or DDS.

If this control command is received also be the state machine of the process itself, it is probably better to ignore it on that level.

PR #8134, opened 2022-02-12
DPL: do not block on forwarding if a new state is available

ktf:


Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -1460,7 +1460,11 @@ bool DataProcessingDevice::tryDispatchComputation(DataProcessorContext& context,
         continue;
       }
       // in DPL we are using subchannel 0 only
-      device->Send(forwardedParts[fi], spec->forwards[fi].channel, 0);
+      // Notice that we cannot block, because we will not
+      // be unblocked in case of a new state
+      // device->Send(forwardedParts[fi], spec->forwards[fi].channel, 0);
+      while (device->Send(forwardedParts[fi], spec->forwards[fi].channel, 0, 100) == 0 || device->NewStatePending() == false) {
```
ktf: @rbx this does not do what I want, I guess. How can I make sure that all the parts have been sent?

rbx: Return value of `0` can be success, if messages are empty. I think for what you want you should use `< 0`, or `<= 0` if you are sure there are no empty messages.
The return values below than 0 are:
    error = -1,
    timeout = -2,
    interrupted = -3
If it is 0 or larger, then it means all parts have been sent out. There is no partial transfer, it's either all or none.

https://github.com/FairRootGroup/FairMQ/blob/master/fairmq/Device.h#L114-L130

ktf: Ok, thanks. I adapted the code accordingly. I think I will simply move that responsibility to libuv as well, as already done for the inputs. Do I understand correctly that indeed the state change will not unblock a blocking send?

rbx: A state change will unblock a blocking Send/Receive call, which will return `TransferCode::interrupted` (-3). After such unblock you can still issue further transfer calls - they will attempt transfer for ~100ms long and if not successful return `TransferCode::interrupted` again. If successful, they will return number of bytes as usual.

ktf: > A state change will unblock a blocking Send/Receive call

this I do not see happening, not with the static plugin, at least.

rbx: It does for me. All controllers interrupt in the same way, there should be no difference there.
Try to play with this simple device: https://github.com/rbx/FairMQ/blob/check-sink/fairmq/devices/runSink.cxx
Run it with static/interactive controller:
```
./fairmq/fairmq-sink --id sink1 --channel-config name=data,method=bind,type=pull,address=ipc://i --control static
```
And interrupt it.

ktf:
To be clear, it happens to me when there is nothing connected on the other side, using the static plugin and using Send() without the timeout.

rbx:
Ok, I took the Sink, changed `Receive` to `Send`, removed the timeout, changed socket type from `pull` to `push` on the command line and it still gets interrupted. Also tried with multipart as msg content:
https://github.com/rbx/FairMQ/blob/check-sink/fairmq/devices/runSink.cxx
I'm not saying there is no bug, but I don't see one now. And a simple device would be most useful and easy to reproduce it.

matthiasrichter (approved):
In any case it makes sense to use the timeout. This is actually implemented in the Send methods of ExternalFairMQDevice, which was used in the output proxy before we switched to using the forwarding inside DataProcessingDevice (which is there anyway).

Probably we need to add a metric for how long a message is pending. Or a warning, if the sending can't go on for some time. In ExternalFairMQDevice there is such a warning, but this needs to be revised.

ktf:
Yeah, but what would really make sense is to queue the messages to be sent and epoll the availability of the sending socket, so that we can do something else in the meanwhile.

PR #8143, opened 2022-02-13
Split of the ITS threshold wf: Calibrator + aggregator - ADDED support for no EoS case

iravasen:
Changes included:
- o2-its-threshold-calib-workflow (already existing): foreseen to run on every EPN, ship DCSconfigObjects_t to aggregator in the endOfStream.
- o2-its-threshold-aggregator-workflow (new): foreseen to run on 1 EPN aggregator node, collects and merge all the DCSConfigObjects_t coming from every EPN. Prepare CCDB wrapper and payload and can ship this to ccdb-populator-wf (if endOfStream exists) or directly to CCDB with CCDBApi (i.e. no ccdb-populator-wf)
- fixes for ITHR tuning analysis

### DIFF
--- a/Detectors/ITSMFT/ITS/workflow/src/ThresholdAggregatorSpec.cxx
+++ b/Detectors/ITSMFT/ITS/workflow/src/ThresholdAggregatorSpec.cxx
@@ -46,23 +46,23 @@ void ITSThresholdAggregator::init(InitContext& ic)
 void ITSThresholdAggregator::run(ProcessingContext& pc)
 {
 
-  //take run type and scan type at the beginning
+  // take run type and scan type at the beginning
   if (this->mRunType == -1) {
-    this->mRunType = pc.inputs().get<short int>("runtype");
-    this->mScanType = pc.inputs().get<char>("scantype");
-    this->mFitType = pc.inputs().get<char>("fittype");
-    this->updateRunID(pc); // run number
+    mRunType = pc.inputs().get<short int>("runtype");
+    mScanType = pc.inputs().get<char>("scantype");
+    mFitType = pc.inputs().get<char>("fittype");
+    updateRunID(pc); // run number
   }
 
-  if (this->mLHCPeriod.empty()) {
-    this->updateLHCPeriod(pc);
+  if (mLHCPeriod.empty()) {
+    updateLHCPeriod(pc);
   }
 
   // Read strings with tuning info
-  const auto tunString = pc.inputs().get<o2::dcs::DCSconfigObject_t>("tunestring");
-  std::string tmpString(tunString.begin(), tunString.end());
-  //Merge all strings coming from several sources (EPN)
-  std::copy(tmpString.begin(), tmpString.end(), std::back_inserter(tuningMerge));
+  const auto tunString = pc.inputs().get<gsl::span<char>>("tunestring");
+  // std::string tmpString(tunString.begin(), tunString.end());
+  // Merge all strings coming from several sources (EPN)
+  std::copy(tunString.begin(), tunString.end(), std::back_inserter(tuningMerge));
 
   return;
 }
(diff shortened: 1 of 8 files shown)


davidrohr (dismissed):
thx, will wait for the FullCI to pass and then merge for tomorrow's nightly.

Detectors/ITSMFT/ITS/workflow/src/ThresholdAggregatorSpec.cxx
```diff
@@ -0,0 +1,220 @@
...
+
+  if (this->mLHCPeriod.empty()) {
+    this->updateLHCPeriod(pc);
+  }
+
+  // Read strings with tuning info
+  const auto tunString = pc.inputs().get<o2::dcs::DCSconfigObject_t>("tunestring");
+  std::string tmpString(tunString.begin(), tunString.end());
+  //Merge all strings coming from several sources (EPN)
+  std::copy(tmpString.begin(), tmpString.end(), std::back_inserter(tuningMerge));
```
shahor02: Here you already create a` vector<char>`, then you create one more locally. Instead, you better do 
```
const auto tunString = pc.inputs().get<gsl::span<char>>("tunestring");
std::copy(tmpString.begin(), tmpString.end(), std::back_inserter(tuningMerge));
```
BTW, you don't need `this->` in the class methods.

Detectors/ITSMFT/ITS/workflow/src/ThresholdAggregatorSpec.cxx
```diff
@@ -0,0 +1,220 @@
...
+//////////////////////////////////////////////////////////////////////////////
+void ITSThresholdAggregator::finalize(EndOfStreamContext* ec)
+{
+  LOGF(info, "endOfStream report:", mSelfName);
+
+  // Below is CCDB stuff
+  long tstart, tend;
+  tstart = o2::ccdb::getCurrentTimestamp();
+  constexpr long SECONDSPERYEAR = 365 * 24 * 60 * 60;
+  tend = o2::ccdb::getFutureTimestamp(SECONDSPERYEAR);
```
shahor02: note that the `tstart = o2::ccdb::getCurrentTimestamp()` gives time in `ms`, while your `tend` is just 1 year in seconds, while it should be an abs. time-stamp in ms. If you want to assign 1y validity, you should have
```
tend = tstart + 365L*24*3600*1000;
```

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -603,16 +609,16 @@ void ITSThresholdCalibrator::setRunType(const short int& runtype)
...
+  return (this->mPixelHits[chipID][col][*(this->N_RANGE) - 1] >= (N_INJ - 2.)); // TODO: -2 is a safety factor for now
 }
 
 //////////////////////////////////////////////////////////////////////////////
 // Extract thresholds and update memory
 void ITSThresholdCalibrator::extractAndUpdate(const short int& chipID)
 {
   // In threshold scan case, reset mThresholdTree before writing to a new file
-  if ((this->mScanType == 'T') && ((this->mRowCounter)++ == N_ROWS_PER_FILE)) {
+  if ((!this->mCounter || this->mCheckEos) && (this->mScanType == 'T') && ((this->mRowCounter)++ == N_ROWS_PER_FILE)) {
```
shahor02: Not sure I understand the logic: how is the data sent to the aggregator in absence of EoS?
I think the only way to do this is not to way for the finalize called from the stop() but detect that the scan for a given chip is complete and send its output from the regular `run()` call. This means, obviously, that each EPN should see more TFs than is needed to complete the scan for every chip it serves to.

shahor02 (approved):

shahor02 (changes requested):
sorry, pressed `approve` by mistake while I wanted `request changes`.

iravasen:
@shahor02 thanks a lot for the comments. I implemented the first 2 while for the third I removed those checks in the if condition trying to simplify a bit the logic when operating without EoS. The logic remains anyway a bit hard to understand but it is as follows:
- INFO: in our thr scan, when the scan reaches the last row of a chip the software behind (ITSComm) puts a zero in one of the calibration word user fields. The zeros come for the last 50 injections in the last row of every chipID. 
- We read the calibration word every time (for every ROF) in the run() method
- At the end the run() method, we check if the RunType (namely the calibration word user field mentioned above) is 0. We have to count the zeros because we expect 50 per chip. Then if runType is 0, we call finalize() which fill the DCSConfigObject_t only for the chips for which we have recorded enough data and then we ship out the DCSconfigObject with ```pc.outputs(...)```. 
- We do step above in a way that at the next RunType = 0, the data of older chips are cleared so that they do not end up again in the DCSconfigObject_t. This explains some calls like ```mThresholds.erase()``` or ```mTuning.clear()```.

davidrohr:
ok, I am not merging it tonight for the nightly. You said you need it on Tuesday. So we have to make sure it goes in tomorrow's nightly.

shahor02 (approved):
Hi @iravasen 
OK, thanks for clarifications, merging.

PR #8149, opened 2022-02-13
[o2-sim] En-/disable hit creation per detector

benedikt-voelkel:
Be able to en-/disable hit creation of detectors while still building
the geometry. This is an extension to the -m/--skipModules options.

* introduce --readoutDetectors and --skipReadoutDetectors options

  if a passive module name is passed for either of the options it will
  be ignored

* --readoutDetectors det_1 ... det_N
  * if det_i specified, only the given detectors will be readout
  * if det_i not contained in active modules (see -m/--skipModules), it
    takes precedence and det_i is added to active modules
  * if not specified, all detectors will be readout which are in line
    with the -m and --skipModules options

* --skipReadoutDetectors det_1 ... det_N
  * given detectors will not produce any hits
  * has no effect on the constructed geometry

* take care of corner cases such as
  --skipModules m --readoutDetectors m --skipReadoutDetectors m
  in which case m is not built at all and therefore not readout

### DIFF
--- a/Common/SimConfig/src/SimConfig.cxx
+++ b/Common/SimConfig/src/SimConfig.cxx
@@ -28,8 +28,10 @@ void SimConfig::initOptions(boost::program_options::options_description& options
     "mcEngine,e", bpo::value<std::string>()->default_value("TGeant3"), "VMC backend to be used.")(
     "generator,g", bpo::value<std::string>()->default_value("boxgen"), "Event generator to be used.")(
     "trigger,t", bpo::value<std::string>()->default_value(""), "Event generator trigger to be used.")(
-    "modules,m", bpo::value<std::vector<std::string>>()->multitoken()->default_value(std::vector<std::string>({"all"}), "all modules"), "list of detectors")(
-    "skipModules", bpo::value<std::vector<std::string>>()->multitoken()->default_value(std::vector<std::string>({""}), ""), "list of detectors to skip (precendence over -m")(
+    "modules,m", bpo::value<std::vector<std::string>>()->multitoken()->default_value(std::vector<std::string>({"all"}), "all modules"), "list of modules included in geometry")(
+    "skipModules", bpo::value<std::vector<std::string>>()->multitoken()->default_value(std::vector<std::string>({""}), ""), "list of modules excluded in geometry (precendence over -m")(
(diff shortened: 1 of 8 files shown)


Common/SimConfig/src/SimConfig.cxx
```diff
@@ -28,8 +28,10 @@ void SimConfig::initOptions(boost::program_options::options_description& options
...
     "generator,g", bpo::value<std::string>()->default_value("boxgen"), "Event generator to be used.")(
     "trigger,t", bpo::value<std::string>()->default_value(""), "Event generator trigger to be used.")(
-    "modules,m", bpo::value<std::vector<std::string>>()->multitoken()->default_value(std::vector<std::string>({"all"}), "all modules"), "list of detectors")(
-    "skipModules", bpo::value<std::vector<std::string>>()->multitoken()->default_value(std::vector<std::string>({""}), ""), "list of detectors to skip (precendence over -m")(
+    "modules,m", bpo::value<std::vector<std::string>>()->multitoken()->default_value(std::vector<std::string>({"all"}), "all modules"), "list of modules included in geometry")(
+    "skipModules", bpo::value<std::vector<std::string>>()->multitoken()->default_value(std::vector<std::string>({""}), ""), "list of modules excluded in geometry (precendence over -m")(
+    "readoutDetectors", bpo::value<std::vector<std::string>>()->multitoken()->default_value(std::vector<std::string>(), ""), "list of detectors creating hits, all if not given; added to to active modules")(
```
benedikt-voelkel, line 33: maybe help message needs to be more clear, such as "list of detectors creating hits; if not given, all detectors compatible with -m/skipModules flag are readout; if given but not yet compatible with -m/--skipModules, detectors are added for geometry construction as well"

Detectors/ITSMFT/MFT/simulation/include/MFTSimulation/Detector.h
```diff
@@ -49,6 +49,11 @@ namespace mft
 class Detector : public o2::base::DetImpl<Detector>
 {
  public:
+  /// Name : Detector Name
+  /// Active: kTRUE for active detectors (ProcessHits() will be called)
+  ///         kFALSE for inactive detectors
+  Detector(Bool_t active);
```
benedikt-voelkel, line 55: Additional ctor since we would like to be able to disable hit creation. All other detectors had this kind of ctor already

sawenzel (approved):

PR #8153, opened 2022-02-15
Improving sensitive hit creation for FT0 in Detector.cxx

zovarga:
- For FT0 the `Detector::ProcessHits(FairVolume* v)` function had too many string comparisons which made it slow. The comparison was modified to be based on integer IDs instead. 
- The new integer IDs have been defined in Detector.h, then added to Detector.cxx (modified functions: `Detector::InitializeO2Detector()` and `Detector::ProcessHits(FairVolume* v)`).
- The calculation of some quantities (etot, enDep, parentID) in `Detector::ProcessHits(FairVolume* v)` were moved inside the relevant if() statement block (i.e. when the particle is a photon), so they are only calculated when necessary.


mslupeck (commented):
Hi @zovarga,
Thanks for this code. Alla is on holidays until Monday, so she'll be able to approve it only then. Have you understood what went wrong the last time? I remember we had to revert a very similar commit a few weeks ago...

zovarga:
Hi Maciej,
the problem was that I cached the initial volume IDs by using TVirtualMC::VolId(const char* name).  
Compared to Geant3, the fMC->CurrentVolID(copy) returns with different integers for Geant4 and when the current volIDs were compared to the initial volIDs in the if() statements, the condition was always false, and no hits were added.
Following Sandro's advice, I am now caching the initial volume IDs by using TGeoVolume and this fixed the issue.

benedikt-voelkel:
Hej @zovarga @mslupeck 
Indeed this is the case and the volume IDs are handled differently in G4 internally. So the caching of IDs via the volume names makes sense. The G4 volume IDs are also updated for the `FairVolumes` (see https://github.com/FairRootGroup/FairRoot/blob/master/base/sim/FairDetector.cxx#L114). So in particular the sensitive volumes are affected by this different G4 numbering scheme.

sawenzel (approved):

benedikt-voelkel:
That seems to be a problem related again to the missing `alien` token. I ran the FST locally and it works just fine with a token.
Checking the sim challenge part now as well

benedikt-voelkel:
The sim_challenge part works just as well

AllaMaevskaya (approved):

benedikt-voelkel:
Comparing the FT0 hits w/ and w/o this PR, I can see exactly the same. To make sure to exactly reproduce everything, I basically ran
```bash
o2-sim-serial -n 10 --skipModules ZDC --readoutDetectors FT0 -e TGeant3 --configKeyValues "SimCutParams.trackSeed=true" --seed 123 -g pythia8pp
```

PR #8154, opened 2022-02-15
[Global Fwd Tracking] Enable to export True Pairing Eff for assessment of GlobalFwdTrack

RenEjima:
This change enable to export these plots.

- True Pairing Efficiency vs pT_{MC} (Inner pseudo rapidity region 2.4 < \eta < 3.0)
- True Pairing Efficiency vs pT_{MC} (Outer pseudo rapidity region 3.0 < \eta < 3.6)
- 2D distribution of pT_{MC} and \eta_{MC} for True Pairing Efficiency
- True Pairing Efficiency vs Pairing Purity (pT = 0.25, 0.75, 2.25 ; \eta = Inner, Outer)

For the calculation of True Pairing Efficiency, MC parameter of True matched track and Pairable track is used. Pairing Efficiency is already merged into O2 but there is difference which parameter(MC or Fit) is used for between numerator and denominator. When reconstructed global muon track include many Fake matched track, Pairing Efficiency will not work well due to this problem. So True Pairing Efficiency will help you to understand efficiency of MFT-MCH track matching.

### DIFF
--- a/Detectors/GlobalTracking/include/GlobalTracking/MatchGlobalFwdAssessment.h
+++ b/Detectors/GlobalTracking/include/GlobalTracking/MatchGlobalFwdAssessment.h
@@ -138,12 +138,16 @@ class GloFwdAssessment
 
   std::unique_ptr<TEfficiency> mChargeMatchEff = nullptr;
   std::unique_ptr<TH2D> mPairingEtaPt = nullptr;
+  std::unique_ptr<TH2D> mTruePairingEtaPt = nullptr;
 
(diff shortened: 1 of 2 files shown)


rpezzi (approved):
@RenEjima , thanks for your contribution. @shahor02 , memory leaks will be fixed in a forthcoming PR.

rpezzi:
Hello, all tests passed. Please merge when possible. Thanks.

shahor02 (approved):

PR #8157, opened 2022-02-15
[EMCAL-753,EMCAL-763] Redesing DigitsWriteoutBuffer with past/future dequeue

hahassan7:
The digits writeout buffer has been redesigned. The basic data containers of the ringbuffer will be 2 std::dequeues, one for past and one for future. The future dequeue will be initialized with a size 15, the past dequeue will not be initialized with a fixed size but expanded until a constant size of 15 is reached. The entry 0 in the future dequeue always represents the current time. When we sample the digits we fill the dequeue in future direction always with the digit time sample (digit time + delay) relative to the current timesample.

### DIFF
--- a/Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsVectorStream.h
+++ b/Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsVectorStream.h
@@ -17,6 +17,7 @@
 #include <vector>
 #include <deque>
 #include <list>
+#include <optional>
 #include <gsl/span>
 #include "TRandom3.h"
 #include "DataFormatsEMCAL/Digit.h"
@@ -38,11 +39,10 @@ namespace emcal
 /// \date 16/02/2022
 
 struct DigitTimebin {
-  unsigned long mTimestamp = 0;
   bool mRecordMode = false;
   bool mEndWindow = false;
   bool mTriggerColl = false;
-  o2::InteractionRecord mInterRecord;
+  std::optional<o2::InteractionRecord> mInterRecord;
   std::shared_ptr<std::unordered_map<int, std::list<LabeledDigit>>> mDigitMap = std::make_shared<std::unordered_map<int, std::list<LabeledDigit>>>();
 };
 
@@ -90,12 +90,6 @@ class DigitsVectorStream
 
   const SimParam* mSimParam = nullptr;  ///< SimParam object
   TRandom3* mRandomGenerator = nullptr; ///< random number generator
-
-  unsigned int mLiveTime = 1500;       ///< EMCal live time (ns)
-  unsigned int mBusyTime = 35000;      ///< EMCal busy time (ns)
-  unsigned int mPreTriggerTime = 600;  ///< EMCal pre-trigger time (ns)
-  unsigned int mTimeWindowStart = 700; ///< The start of the time window (ns)
-  unsigned int mDelay = 700;           ///< Signal delay time (ns)
 
   ClassDefNV(DigitsVectorStream, 1);
 };
--- a/Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsVectorStream.h
+++ b/Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsVectorStream.h
@@ -1,0 +1,107 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef ALICEO2_EMCAL_DIGITSVECTORSTREAM_H_
+#define ALICEO2_EMCAL_DIGITSVECTORSTREAM_H_
+
+#include <memory>
+#include <unordered_map>
+#include <vector>
+#include <deque>
+#include <list>
+#include <gsl/span>
+#include "TRandom3.h"
+#include "DataFormatsEMCAL/Digit.h"
+#include "EMCALSimulation/LabeledDigit.h"
+#include "DataFormatsEMCAL/TriggerRecord.h"
+#include "SimulationDataFormat/MCTruthContainer.h"
+#include "EMCALSimulation/SimParam.h"
+
+namespace o2
+{
+namespace emcal
+{
+
+/// \class DigitsVectorStream
+/// \brief Container class for Digits, MC lebels, and trigger records
+/// \ingroup EMCALsimulation
+/// \author Hadi Hassan, ORNL
+/// \author Markus Fasel, ORNL
+/// \date 16/02/2022
+
+struct DigitTimebin {
+  unsigned long mTimestamp = 0;
+  bool mRecordMode = false;
+  bool mEndWindow = false;
+  bool mTriggerColl = false;
+  o2::InteractionRecord mInterRecord;
+  std::shared_ptr<std::unordered_map<int, std::list<LabeledDigit>>> mDigitMap = std::make_shared<std::unordered_map<int, std::list<LabeledDigit>>>();
+};
+
+class DigitsVectorStream
+{
+
+ public:
+  /// Default constructor
+  DigitsVectorStream() = default;
+
+  /// Destructor
+  ~DigitsVectorStream() = default;
+
+  /// clear the container
+  void clear();
+
+  /// Initialize the streamer
+  void init();
+
+  /// Fill all the containers, digits, labels, and trigger records
+  void fill(gsl::span<o2::emcal::DigitTimebin> digitlist, o2::InteractionRecord record);
+
+  /// Getters for the finals data vectors, digits vector, labels vector, and trigger records vector
+  std::vector<o2::emcal::Digit> getDigits() { return mDigits; }
+  std::vector<o2::emcal::TriggerRecord> getTriggerRecords() { return mTriggerRecords; }
+  o2::dataformats::MCTruthContainer<o2::emcal::MCLabel> getMCLabels() { return mLabels; }
+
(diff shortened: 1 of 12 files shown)


Detectors/EMCAL/simulation/src/DigitsWriteoutBuffer.cxx
```diff
@@ -20,62 +20,104 @@
...
-    startingPosition = 0;
-  }
 
   std::list<std::unordered_map<int, std::list<LabeledDigit>>> output;
 
-  for (int ientry = startingPosition; ientry < startingPosition + nsamples; ientry++) {
-    output.push_back(mTimedDigits[ientry]);
+  for (auto digitTimebin : mTimedDigitsPast) {
+    output.push_back(*digitTimebin.mDigitMap);
   }
```
mfasDa: Consider a gsl::span (non-owning as return type in order to avoid the copy process:
``` c++
const gsl::span<DigitsWriteoutBuffer::DigitTimeBin> DigitsWriteoutBuffer::getLastSamples() const 
{
   return const gsl::<DigitsWriteoutBuffer::DigitTimeBin>(mTimedDigitsPast);
}

File: Detectors/EMCAL/simulation/src/Digitizer.cxx [hunk 2]
```diff
@@ -196,14 +204,12 @@ void Digitizer::setEventTime(double t)
 
   mEventTime = t - mTriggerTime;
 
+  mDigits.forwardMarker(mEventTime);
```
mfasDa: Here I am not sure how it works, the intention was that the timing in the DigitsWriteoutBuffer is absolute, but this line makes it relative (to mTriggerTime), changing mTriggerTime you might get a time that is smaller than the current marker in the DigitsWriteoutBuffer

Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsWriteoutBuffer.h
```diff
@@ -36,37 +36,56 @@ namespace emcal
...
+  void setBusyTime(unsigned int busyTime) { mBusyTime = busyTime; }
+  void setPreTriggerTime(unsigned int pretriggerTime) { mPreTriggerTime = pretriggerTime; }
 
  private:
-  unsigned int mBufferSize = 30;                                             ///< The size of the buffer, it has to be at least 2 times the size of the readout cycle
-  unsigned int mTimeBinWidth = 100;                                          ///< The size of the time bin (ns)
-  unsigned int mNumberReadoutSamples = 15;                                   ///< The number of smaples in a readout window
-  double mLastEventTime = 1500;                                              ///< The event time of last collisions in the readout window
-  std::deque<std::unordered_map<int, std::list<LabeledDigit>>> mTimedDigits; ///< Container for time sampled digits per tower ID
+  unsigned int mBufferSize = 30;               ///< The size of the buffer, it has to be at least 2 times the size of the readout cycle
```
mfasDa: In principle a redundant information.

Detectors/EMCAL/simulation/src/DigitsWriteoutBuffer.cxx
```diff
@@ -20,62 +20,104 @@
...
+    towerEntry->second.push_back(dig);
   }
+}
 
-  towerEntry->second.push_back(dig);
+// When the current time is forwarded (every 100 ns) the entry 0 of the future dequeue
+// is removed and pushed at the end of the past dequeue. At the same time a new entry in
+// the future dequeue is pushed at the end, and - in case the past dequeue reached 15 entries
+// the first entry of the past dequeue is removed.
+void DigitsWriteoutBuffer::forwardMarker(double eventTime)
```
mfasDa: I was thinking to handle the live/busy mode here: we would add two more parameters first to check whether with the event we open the readout (600 ns before end of busy), this would set the digit timebins in the future dequeue to record mode, and then a second option for the trigger, which would  the last entry in the future dequeue to "close window". The you could get rid of the time window handling in the addDigit function.

Detectors/EMCAL/simulation/src/DigitsWriteoutBuffer.cxx
```diff
@@ -20,62 +20,104 @@
...
+      break;
+    }
+
+    mTimedDigitsPast.push_back(mTimedDigitsFuture.front());
+    mTimedDigitsFuture.pop_front();
+    mTimedDigitsFuture.push_back(DigitTimebin());
+
+    if (mTimedDigitsPast.size() > mBufferSize) {
+      mTimedDigitsPast.pop_front();
+    }
```
mfasDa, line 106: I was thinking to do the filling of the output vector here, because basically in this step, in case the last digit in the past dequeue has the close marker set, we know we have to stream the digits which are just created. The filling part would then be removed from the DigitizerSpec, which would simplify the Spec even more. Furthermore, it makes the object to what it is actually supposed to, indicated by its name: Writing out the digits created by the time sampling. One could do this in two ways, I have no strong preference here:

- Create a dedicated stream class which contains the digits vector, trigger records vector and MCLabelContainer, pass the entire std::dequeue to this
- Reference the 3 containers mentioned above - which are owned by the Spec and fill them inside the DigitsWriteoutBuffer in a dedicated fill function.

The first option would maybe have the small advantage that it itself the 3 containers, so it is self-contained.

mfasDa (changes requested):
See suggestions dedicated to certain code blocks.

Detectors/EMCAL/simulation/src/DigitsWriteoutBuffer.cxx
hahassan7: Done

Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsWriteoutBuffer.h
hahassan7: Done

Detectors/EMCAL/simulation/src/DigitsWriteoutBuffer.cxx
hahassan7: Done

Detectors/EMCAL/simulation/src/DigitsWriteoutBuffer.cxx
hahassan7: Done

Detectors/EMCAL/simulation/src/Digitizer.cxx
hahassan7: Done

Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsVectorStream.h
```diff
@@ -0,0 +1,107 @@
...
+  bool mRemoveDigitsBelowThreshold = true; ///< remove digits below threshold
+
+  const SimParam* mSimParam = nullptr;  ///< SimParam object
+  TRandom3* mRandomGenerator = nullptr; ///< random number generator
+
+  unsigned int mLiveTime = 1500;       ///< EMCal live time (ns)
+  unsigned int mBusyTime = 35000;      ///< EMCal busy time (ns)
+  unsigned int mPreTriggerTime = 600;  ///< EMCal pre-trigger time (ns)
+  unsigned int mTimeWindowStart = 700; ///< The start of the time window (ns)
+  unsigned int mDelay = 700;           ///< Signal delay time (ns)
```
mfasDa: For my ignorance: Why does the output stream need to know about all the timing? The only task to it is to stream the digits which are already time-ordered via the DigitTimebin into the output vector?

Detectors/EMCAL/simulation/src/DigitsVectorStream.cxx
```diff
@@ -0,0 +1,151 @@
...
+          continue;
+        }
+
+        outputList[tower].push_back(ld);
+      }
+    }
+  }
+
+  for (auto [tower, outdiglist] : outputList) {
+    for (auto d : outdiglist) {
```
mfasDa: I expect both could be `const auto &` in order to prevent copy.

Detectors/EMCAL/simulation/src/DigitsVectorStream.cxx
```diff
@@ -0,0 +1,151 @@
...
+      outdiglist.sort();
+
+      Digit digit = d.getDigit();
+      std::vector<MCLabel> labels = d.getLabels();
+      mDigits.push_back(digit);
+
+      Int_t LabelIndex = mLabels.getIndexedSize();
+      for (auto label : labels) {
```
mfasDa: Same here (`const auto &`)

mfasDa (changes requested):
At least some access-by-value in rage-based for loop could be changed in order to avoid copying potentially large objects.

The idea of the DigitsVectorStream was to just stream (serialize) the (15) time samples the DigitsWriteoutBuffer created into the output vectors (digits vector, trigger records vector and MC Label Container) once the readout of the window finishes, which the DigitsWriteoutBuffer can self-determine once the EndOfWindow marker reaches the end of the past buffer. I don't fully understand why we need the whole timing stuff (livetime, busy time , trigger time, pretrigger time, ...) there. The only thing the DigitsVector stream should worry about is to convert a vector of 15 timebins of ntowers into towers vectors of 15 timebins (spaw axis), remove digits below threshold, and add the underlying noise (that I like a lot that it is done here). Once they are on the vector(s) the DigitsVectorStream can forget about them. Things like busy time or similar shouldn't matter here, this should be part of the DigitsWriteoutBuffer. Or do I miss something?

Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsVectorStream.h
hahassan7: Done

Detectors/EMCAL/simulation/src/DigitsVectorStream.cxx
hahassan7: Done

Detectors/EMCAL/simulation/src/DigitsVectorStream.cxx
hahassan7: Done

mfasDa (approved):

mfasDa:
CI errors are unrelated

PR #8161, opened 2022-02-15
Fix HMP mc->raw output for for-cru option

shahor02:
name the files as DET_alio2-cr1-flpXXX_crorcYYY_lnkZ.raw format


shahor02:
failures are due to the gitlab issues, tested locally, mergiing.

PR #8163, opened 2022-02-15
Fixed runType in ITS calib workflow

iravasen:
Run type not converted correctly when DB version is different from 0. Added mask to work in all cases.


iravasen:
@shahor02 this fixes the runType issue observed before during the test.

shahor02 (approved):
trivial fix, merging

PR #8169, opened 2022-02-16
DPL: fix start-stop-start transition

ktf:


ktf:
@davidrohr @matthiasrichter I think the actual culprit is readout-proxy trying to exit when the EoS is received. The rest does not harm, but I think it's unrelated with the issue with start-stop-start. This fixes my local reproducer (although I suspect removing the quit might kill some of the readout-proxy related tests).

davidrohr:
For reference, what I commented to Giulio on Mattermost:
with that PR the situation changes:
- Before the EoS, I see data in both the input proxy and the output proxy.
- After the EoS, now I see data in the input proxy (there was none before), but I don't see data in the output proxy.
This is independent from whether I do a restart or not, i.e. EoS,Stop,Start behaves exactly the same as only EoS.

But there is another side effect of this PR: The FST with DD does not terminate correctly any more. I think the PR disables the termination at EoS. We will have to make it such that if it is controlled by the DPL driver, it should still terminate on EoS. Only if controlled by ECS or ODC, it shouldn't.

ktf:
@davidrohr @matthiasrichter I am convinced this will work in ECS/DDS, however it will make the restart button of the GUI not working correctly because once we leave the RUN state the GUI cannot communicate with a device anymore. I have tried to improve the GUI to behave closely to what ECS/DDS does, but it's not trivial so I suggest we try this online, if the tests pass.

ktf:
Ok, nevermind what I said. I think this should work...

ktf:
Merging this to get it in the build. Tested locally.

PR #8170, opened 2022-02-16
Add draft for TRD DCS processor

martenole:
Hi @tdietel 
this is a first version of the DCS data point processing in O2. So far I am only taking into account the gas parameters and storing one CCDB object for them. Could you please have a look and see if it makes sense like that? I have to admit I tested it now only with a simulation. I am not sure that the format of the DCS data points is really the same in data. This I need to check probably with Denise.
In order to run the processing with a local CCDB:
1. follow the instructions for a local instance [here](https://docs.google.com/document/d/1_GM6yY7ejVEIRi1y8Ooc9ongrGgZyCiks6Ca0OAEav8)
2. create CCDB entry with DCS data points by running `Detectors/TRD/calibration/macros/makeTRDCCDBEntryForDCS.C`
3. start workflow

```
o2-calibration-trd-dcs-sim-workflow --max-timeframes 5 --delta-fraction 0.8 -b | o2-calibration-trd-dcs-workflow --use-ccdb-to-configure -b --use-verbose-mode | o2-calibration-ccdb-populator-workflow --ccdb-path="http://localhost:8080" -b
```

Cheers,
Ole

### DIFF
--- a/Detectors/TRD/calibration/src/DCSProcessor.cxx
+++ b/Detectors/TRD/calibration/src/DCSProcessor.cxx
@@ -1,0 +1,241 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include "TRDCalibration/DCSProcessor.h"
+#include "DetectorsCalibration/Utils.h"
+
+using namespace o2::trd;
+using namespace o2::dcs;
+
+using DeliveryType = o2::dcs::DeliveryType;
+using DPID = o2::dcs::DataPointIdentifier;
+using DPVAL = o2::dcs::DataPointValue;
+
+void TRDDCSMinMaxMeanInfo::print() const
+{
+  LOG(info) << "Min value: " << minValue;
+  LOG(info) << "Max value: " << maxValue;
+  LOG(info) << "Mean value: " << meanValue;
+  LOG(info) << "Number of points added: " << nPoints;
+}
+
+void TRDDCSMinMaxMeanInfo::addPoint(float value)
+{
+  if (nPoints == 0) {
+    minValue = value;
+    maxValue = value;
+    meanValue = value;
+  } else {
+    if (value < minValue) {
+      minValue = value;
+    }
+    if (value > maxValue) {
+      maxValue = value;
+    }
+    meanValue += (value - meanValue) / (nPoints + 1);
+  }
+  ++nPoints;
+}
+
+//__________________________________________________________________
+
+void DCSProcessor::init(const std::vector<DPID>& pids)
+{
+  // fill the array of the DPIDs that will be used by TRD
+  // pids should be provided by CCDB
+
+  for (const auto& it : pids) {
+    mPids[it] = false;
+  }
+}
+
+//__________________________________________________________________
+
+int DCSProcessor::process(const gsl::span<const DPCOM> dps)
+{
+
+  // first we check which DPs are missing - if some are, it means that
+  // the delta map was sent
+  if (mVerbose) {
+    LOG(info) << "\n\n\nProcessing new TF\n-----------------";
+  }
+  if (!mStartTFset) {
+    mStartTF = mStartValidity;
+    mStartTFset = true;
+  }
+
+  std::unordered_map<DPID, DPVAL> mapin;
+  for (auto& it : dps) {
+    mapin[it.id] = it.data;
+  }
+  for (auto& it : mPids) {
+    const auto& el = mapin.find(it.first);
+    if (el == mapin.end()) {
+      LOG(info) << "DP " << it.first << " not found in map";
+    } else {
+      LOG(info) << "DP " << it.first << " found in map";
+    }
+  }
+
+  // now we process all DPs, one by one
+  for (const auto& it : dps) {
+    // we process only the DPs defined in the configuration
+    const auto& el = mPids.find(it.id);
+    if (el == mPids.end()) {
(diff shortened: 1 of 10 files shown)


Detectors/TRD/calibration/src/DCSProcessor.cxx
```diff
@@ -0,0 +1,246 @@
...
+#include "DetectorsCalibration/Utils.h"
+
+using namespace o2::trd;
+using namespace o2::dcs;
+
+using DeliveryType = o2::dcs::DeliveryType;
+using DPID = o2::dcs::DataPointIdentifier;
+using DPVAL = o2::dcs::DataPointValue;
+
+void TRDDCSMinMaxMeanInfo::print() const
```
tdietel, line 22: Don't use people ostream operators these days? ;-) Maybe it's not quite in line with O2 coding standards.

On a serious note: I guess this should also log the number of seen data points.

Detectors/TRD/calibration/src/DCSProcessor.cxx
```diff
@@ -0,0 +1,246 @@
...
+  if (mVerbose) {
+    if (type == RAW_DOUBLE) {
+      LOG(info) << "Processing DP = " << dpcom << ", with value = " << o2::dcs::getValue<double>(dpcom);
+    } else if (type == RAW_INT) {
+      LOG(info) << "Processing DP = " << dpcom << ", with value = " << o2::dcs::getValue<int32_t>(dpcom);
+    }
+  }
+  auto flags = val.get_flags();
+  if (processFlags(flags, dpid.get_alias()) == 0) {
+    if (type == RAW_DOUBLE) {
```
tdietel: I think DIM usually works with single precision float. Is the conversion to double intentional?

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
```diff
@@ -0,0 +1,52 @@
...
+// or submit itself to any jurisdiction.
+
+#include "DCStestWorkflow/DCSRandomDataGeneratorSpec.h"
+#include "Framework/runDataProcessing.h"
+
+o2::framework::WorkflowSpec defineDataProcessing(o2::framework::ConfigContext const& configcontext)
+{
+  std::vector<o2::dcs::test::HintType> dphints;
+  // for TRD
+  // for test, we use less DPs that official ones
```
tdietel: that -> than

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
```diff
@@ -0,0 +1,52 @@
...
+  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gasCO2", 0, 50.});
```
tdietel: What art the values 0 and 50.? A comment might help... 
If one of them is a default value, it would be good to put something non-sensical (e.g. -999.).

Detectors/TRD/calibration/src/DCSProcessor.cxx
```diff
@@ -0,0 +1,246 @@
...
+}
+
+//______________________________________________________________________
+
+void DCSProcessor::updateDPsCCDB()
+{
+  // here we create the object to then be sent to CCDB
+  LOG(info) << "Finalizing";
+  // TODO I do not know why this converter works, but I don't want to go deeper into the DCS code at the moment
+  union Converter {
```
tdietel: Hmm, this converter might be problematic. As I wrote elsewhere, I am not sure why we would get double instead of float.
I would have looked in the direction of a reinterpret_cast.
I am also wondering if endianness could be a problem? How is that on apple silicon / M1?

Detectors/TRD/calibration/src/DCSProcessor.cxx
```diff
@@ -0,0 +1,246 @@
...
+  for (const auto& it : mPids) {
+    const auto& type = it.first.get_type();
+    if (type == o2::dcs::RAW_DOUBLE) {
+      auto& trddcs = mTRDDCS[it.first];
+      if (it.second == true) { // we processed the DP at least 1x
+        auto& dpVect = mDpsDoublesmap[it.first];
+        for (const auto& dpVal : dpVect) {
+          converter.raw_data = dpVal.payload_pt1;
+          double value = converter.double_value;
+          if (trddcs.nPoints == 0) {
```
tdietel: Minor style issue: this could be moved to a method of TRDDCSMinMaxMeanInfo

tdietel (commented):
I have a few comments. The main one is the one around the converter. Do you want to test this before the merge?

Detectors/TRD/calibration/src/DCSProcessor.cxx
martenole: I have no idea about the precision of the data types.. All I have is the list from Denise. But if its usually single precision I'll use `RAW_FLOAT` instead

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
martenole: I'll put a comment. It means for simulations the provided value is in the range from 0 to 50

Detectors/TRD/calibration/src/DCSProcessor.cxx
martenole: It is not so clear to me, as in the DCS code the value is encapsulated in a 64 bit variable with a 56 bit payload (if I understand the code in `Detectors/DCS/include/DetectorsDCS/DataPointValue.h` correctly). The other detectors use the same conversion to get the payload from the data point. So I for now just copied that. Probably I need to ask someone from DCS about it, but I don't know whom.

Detectors/TRD/calibration/src/DCSProcessor.cxx
tdietel: Ok, I guess we will see it in the tests.

Detectors/TRD/calibration/src/DCSProcessor.cxx
martenole: This is changed now and uses the `getValue` method provided from DCS

Detectors/TRD/calibration/src/DCSProcessor.cxx
martenole: Done

tdietel (approved):

PR #8173, opened 2022-02-17
IDCs: various code optimisations

matthias-kleiner:
- creation of status map for pads which are dead or delivering high IDC values using IDC0 map
- applying pad status map on FLPs when calculating 1D-IDCs
- fix race condition when using multithreading during calculation of IDC1
- using the timestamp information instead of TFid when writing IDCs to CCDB
- printing of debug informations
- do not use weights during averaging/grouping in case no weights were set
- setting of the z-axis when drawing factorized IDCs

### DIFF
--- a/Detectors/TPC/base/include/TPCBase/CalArray.h
+++ b/Detectors/TPC/base/include/TPCBase/CalArray.h
@@ -278,7 +278,7 @@ inline bool CalArray<T>::operator==(const CalArray<T>& other) const
     LOG(error) << "pad subset type of the objects it not compatible";
     return false;
   }
-  bool isSame = mData == other.mData;
+  bool isSame = (mData == other.mData);
   return isSame;
 }
 
--- a/Detectors/TPC/base/include/TPCBase/CalArray.h
+++ b/Detectors/TPC/base/include/TPCBase/CalArray.h
@@ -122,6 +122,9 @@ class CalArray
   /// Divide this by other channel by channel
   const CalArray& operator/=(const CalArray& other);
 
+  /// check for equality
+  bool operator==(const CalArray& other) const;
+
   /// Add value to all channels
   const CalArray& operator+=(const T& val);
 
@@ -265,6 +268,18 @@ inline const CalArray<T>& CalArray<T>::operator/=(const CalArray<T>& other)
     }
   }
   return *this;
+}
+
+//______________________________________________________________________________
+template <class T>
+inline bool CalArray<T>::operator==(const CalArray<T>& other) const
+{
+  if (!((mPadSubset == other.mPadSubset) && (mPadSubsetNumber == other.mPadSubsetNumber))) {
+    LOG(error) << "pad subset type of the objects it not compatible";
+    return false;
+  }
+  bool isSame = mData == other.mData;
+  return isSame;
 }
 
 //______________________________________________________________________________
--- a/Detectors/TPC/workflow/include/TPCWorkflow/TPCFourierTransformAggregatorSpec.h
+++ b/Detectors/TPC/workflow/include/TPCWorkflow/TPCFourierTransformAggregatorSpec.h
@@ -58,7 +58,7 @@ class TPCFourierTransformAggregatorSpec : public o2::framework::Task
   {
     // set the min range of TFs for first TF
     if (mProcessedTFs == 0) {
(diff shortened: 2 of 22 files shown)


Detectors/TPC/workflow/include/TPCWorkflow/TPCFourierTransformAggregatorSpec.h
```diff
@@ -109,20 +107,22 @@ class TPCFourierTransformAggregatorSpec : public o2::framework::Task
...
   std::map<std::string, std::string> mMetadata; ///< meta data of the stored object in CCDB
   bool mWriteToDB{};                            ///< flag if writing to CCDB will be done
-  std::array<uint32_t, 2> mTFRange{};           ///< storing of first and last TF used when setting the validity of the objects when writing to CCDB
+  std::array<uint64_t, 2> mTimeStampRange{};    ///< storing of first and last time stamp used when setting the validity of the objects when writing to CCDB
   int mProcessedTFs{0};                         ///< number of processed time frames to keep track of when the writing to CCDB will be done
 
   /// \return returns TF of current processed data
   uint32_t getCurrentTF(o2::framework::ProcessingContext& pc) const { return o2::framework::DataRefUtils::getHeader<o2::header::DataHeader*>(pc.inputs().getFirstValid(true))->tfCounter; }
 
+  uint64_t getCurrentTimeStamp(o2::framework::ProcessingContext& pc) const { return DataRefUtils::getHeader<DataProcessingHeader*>(pc.inputs().getFirstValid(true))->creation; }
```
wiechula, line 116: Perhaps these two functions can be put in https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/TPC/workflow/include/TPCWorkflow/ProcessingHelpers.h
And I would prefer to rename `getCurrentTimeStamp` to `getCreationTime`.

wiechula (changes requested):
Just minor things.

Detectors/TPC/workflow/include/TPCWorkflow/TPCFourierTransformAggregatorSpec.h
matthias-kleiner: Thanks for the comment. I prefer to change this and some minor things (update the extracted pad status map only when it changed) in another PR. Can it then be merged now?

wiechula (dismissed):
Fine for me. Minor things can be done later.

wiechula:
@davidrohr @shahor02 , seems the error is unrelated. Can this be merged?

Detectors/TPC/base/include/TPCBase/CalArray.h
```diff
@@ -267,6 +270,18 @@ inline const CalArray<T>& CalArray<T>::operator/=(const CalArray<T>& other)
...
 
+//______________________________________________________________________________
+template <class T>
+inline bool CalArray<T>::operator==(const CalArray<T>& other) const
+{
+  if (!((mPadSubset == other.mPadSubset) && (mPadSubsetNumber == other.mPadSubsetNumber))) {
+    LOG(error) << "pad subset type of the objects it not compatible";
+    return false;
+  }
+  bool isSame = mData == other.mData;
```
wiechula: Readability braces.
```suggestion
  bool isSame = (mData == other.mData);
```

wiechula (dismissed):

Detectors/TPC/base/include/TPCBase/CalArray.h
matthias-kleiner: Thanks, should I fix it in this PR?
Otherwise I would create a new PR with some other minor changes and fix it there...

wiechula: I thinks this can be fixed in a separate PR. It would be good to get these modification in soon, I think.

shahor02: The FST fails due to the alien-token problem on the builder, but this PR cannot affect the FST, shall I merge this?

matthias-kleiner: From my side it can be merged

wiechula (approved):
@shahor02 , from my side this can be merged.

shahor02 (approved):

PR #8177, opened 2022-02-18
DPL Analysis: do not call operator* twice for slice index

aalkin:


Framework/Core/include/Framework/ASoA.h
```diff
@@ -1501,76 +1501,79 @@ void notBoundTable(const char* tableName);
...
+    template <typename T>                                                                   \
+    auto _Getter_##_as() const                                                              \
+    {                                                                                       \
+      if (O2_BUILTIN_UNLIKELY(mBinding == nullptr)) {                                       \
+        o2::soa::notBoundTable(#_Table_);                                                   \
+      }                                                                                     \
+      if (O2_BUILTIN_UNLIKELY(!has_##_Getter_())) {                                         \
+        return static_cast<T const*>(mBinding)->emptySlice();                               \
+      }                                                                                     \
+      auto a = *mColumnIterator;                                                            \
```
ktf, line 1547: this does a copy of the whole iterator, no?

ktf: also, are you sure the compiler does not optimize this out in any case?

ktf (commented):
See inline comments.

Framework/Core/include/Framework/ASoA.h
aalkin: `operator*` contains branching and copies several pointers, and this would only be a copy of a pair of int32s, would be also easier for the compiler to optimize.

ktf: Did you actually check the generated asm? I would be surprised if `*mColumnIterator` is not considered a constant expression and executed only once, assuming `operator*` is const.

aalkin: I haven't checked the asm, although `operator*` is indeed const.

PR #8178, opened 2022-02-18
Fix OB Half Staves relative position and U-leg length

mario6829:
The relative separation of the two half staves in a OB stave was erroneously set to a smaller value than the actual one (1.76mm instead of 3.5mm). One of the two U-legs holding the two half staves to the stave structure was shorter than the real one (5.0mm instead of 6.3mm), and the right and left U-legs were swapped. All these errors are fixed with this PR.


iouribelikov (approved):

iouribelikov:
@shahor02  The failure of one of the checks seems to be unrelated to these changes.  May be, we can merge this PR.

PR #8184, opened 2022-02-18
DPL Analysis: introduce particle-to-tracks reverse index

aalkin:
@jgrosseo I propose to add this sort of index centrally, to simplify tracking efficiency studies.


jgrosseo:
What do we do when this is not unique?

aalkin:
This is my main concern with this at the moment, I expected it to be *mostly* unique, but that is not the case. This can probably be resolved by grouping or introducing 1-to-array and 1-to-slice index tables. The latter are probably useful in other cases.

jgrosseo:
I think most studies will loop over tracks and mc particles separately to build an efficiency. Then special studies which look for duplicated tracks could profit from grouping. I would not introduce an additional index for this.

aalkin:
If we can guarantee that most of the particles have one track, then sure, we do not need this index. The ability to build it for specialized studies would be useful, though.

PR #8185, opened 2022-02-18
Fixed mask for runType once more + bug fixes for ITHR scan

iravasen:


iravasen:
@shahor02 I fixed once more the mask for the runType since I made a mistake with the last fix (was not universal for all run types). In addition I fixed some other bugs discovered with ITHR scan today.

shahor02 (approved):

PR #8187, opened 2022-02-19
DPL Analysis: allow index builder to use extended tables as input

aalkin:


ktf:
Tested on hyperloop. Merging.

PR #8191, opened 2022-02-20
CCDB fetcher remapping to different hosts or local files

shahor02:
Allows remapping certain conditions to different hosts or local directories, e.g.
```
 --condition-remap “http://alice-ccdb.cern.ch/user/username:TOF/Calib/LHCPhace,TOF/Calib/Diagnostics;\
http://ccdb-test.cern.ch:8080:ITS/Calib/Align;\
file:///<localDirectory>:GLO/Config/GRPECS” 
```

### DIFF
--- a/CCDB/include/CCDB/CcdbApi.h
+++ b/CCDB/include/CCDB/CcdbApi.h
@@ -76,6 +76,12 @@ class CcdbApi //: public DatabaseInterface
   std::string const& getURL() const { return mUrl; }
 
   /**
+   * Check if we are in a snapshot mode
+   *
+   */
+  bool isSnapshotMode() const { return mInSnapshotMode; }
+
+  /**
    * Create a binary image of the arbitrary type object, if CcdbObjectInfo pointer is provided, register there
    *
    * the assigned object class name and the filename
--- a/Framework/Core/src/CCDBHelpers.cxx
+++ b/Framework/Core/src/CCDBHelpers.cxx
@@ -44,11 +44,15 @@ struct CCDBFetcherHelper {
   std::unordered_map<std::string, DataAllocator::CacheId> mapURL2DPLCache;
   std::string createdNotBefore = "0";
   std::string createdNotAfter = "3385078236000";
-
-  o2::ccdb::CcdbApi api;
+  std::unordered_map<std::string, o2::ccdb::CcdbApi> apis;
   std::vector<OutputRoute> routes;
   std::map<int64_t, CCDBCacheInfo> cache;
   std::unordered_map<std::string, std::string> remappings;
+  o2::ccdb::CcdbApi& getAPI(const std::string& path)
+  {
+    auto entry = remappings.find(path);
+    return apis[entry == remappings.end() ? "" : entry->second];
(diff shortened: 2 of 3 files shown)


shahor02:
Don't merge yet!

ktf:
Wasn't  this supposed to be a feature of CCDB API?

shahor02:
@ktf if you mean using separate CCDBApi instance per host, yes, we could use just one but re-init it every time to a new host. Apart from some overhead of reinitialization, there will be at least 2 extra string comparisons per query (current vs queried host) + check if the host to query is a local dir. (in which case we can skip the query if the object was already cached, since there is no time validity involved).
Do you see a problem with that?

shahor02:
@ktf BTW, what is the purpose of allowing `/` in the remapping? If this is to refer to the local directory, the CcdbAPI expects `file://` for that, which I've added.

ktf:
> @ktf BTW, what is the purpose of allowing `/` in the remapping? If this is to refer to the local directory, the CcdbAPI expects `file://` for that, which I've added.

Yes, that was the idea.

shahor02:
Then, if you don't mind, I would remove it: the format accepted by the CcdbAPI is `file://`.

ktf:
> @ktf if you mean using separate CCDBApi instance per host, yes, we could use just one but re-init it every time to a new host. Apart from some overhead of reinitialization, there will be at least 2 extra string comparisons per query (current vs queried host) + check if the host to query is a local dir. (in which case we can skip the query if the object was already cached, since there is no time validity involved).

Well, the expensive part is the reconnect / SSL negotiation, I guess. So what I was thinking is that we could keep a map of host->curl_handle in the CCDB API and reuse accordingly, which at least would avoid the curl_init_handle reinitialisation. What would really be beneficial is using curl_multi_perform (so that we can have multiple fetches in fly, without using threads) or maybe even a shortlived websocket (but that would require backend support).

ktf:
> Then, if you don't mind, I would remove it: the format accepted by the CcdbAPI is `file://`.

Sure.

shahor02:
> Well, the expensive part is the reconnect / SSL negotiation, I guess. So what I was thinking is that we could keep a map of host->curl_handle in the CCDB API and reuse accordingly, which at least would avoid the curl_init_handle reinitialisation. What would really be beneficial is using curl_multi_perform (so that we can have multiple fetches in fly, without using threads) or maybe even a shortlived websocket (but that would require backend support).

Is there is difference between yet non-existing cashing of host->curl_handles in the CcdbApi or using a cache of CcdbApi instances? In my PR each instance (I would not expect more then 2-3) is initialized only once. Note that currently every change of host, i.e. `CcdbAPI::init(..)` also checks for the alien-token.

ktf:
Indeed, with the current CCDBApi i guess the minimal change is what you have here.

ktf:
Dataflow CS8 worked.

ktf:
@awegrzyn @davidrohr do we still need the dataflow build on CC7?

davidrohr:
For the epn we don't need it.

PR #8192, opened 2022-02-21
DPL GUI: allow tracing different code paths independently

ktf:


ktf:
This will allow the GUI to trace different codepaths separately, so that one can provide fine grained log level and have them enabled only for some "events". Useful to e.g. have debug information when OOB channels are polled, rather than the usual ones.

ktf:
image

PR #8194, opened 2022-02-21
[O2-2748] Enable Compiler Warnings

MichaelLettrich:
Enable most warnings in `- Wall` and `-Wextra` in O2, but make sure they don't lead to compile errors even if `-Werror` is enabled. 

DO NOT MERGE - this tests the different behaviors on clang and GCC.

### DIFF
--- a/dependencies/O2CompileFlags.cmake
+++ b/dependencies/O2CompileFlags.cmake
@@ -11,6 +11,64 @@
 
 include_guard()
 
+#
+# o2_build_warning_flags() builds a list of warning flags from their names.
+#
+set(O2_ENABLE_WARNINGS "OFF")
+if(DEFINED ENV{ALIBUILD_O2_WARNINGS})
+  set(O2_ENABLE_WARNINGS "ON")
+endif()
+
+function(o2_build_warning_flags)
+  cmake_parse_arguments(PARSE_ARGV 0 A "" "PREFIX;OUTPUTVARNAME" "WARNINGS")
+
+  if(A_UNPARSED_ARGUMENTS)
+    message(FATAL_ERROR "Unexpected unparsed arguments: ${A_UNPARSED_ARGUMENTS}")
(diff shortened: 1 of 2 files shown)


TimoWilken:
Looks like clang on Mac gets confused. Disabling the `unknown-warning-option` warning should silence this, if it's fine to have some of these only on Linux?

MichaelLettrich:
@TimoWilken thanks for noticing. It looks like this is a clang vs gcc missmatch of some warnings. I'll add the `-Wno-unknown-warning-option` to the cmake.

ktf:
What do we plan to do with this? I am against keeping this open indefinitely until people act on them.

ktf (changes requested):
Either someone is tasked to fix the errors or this will be closed.

MichaelLettrich:
Hi, I'm taking care of it till the next WP3 meeting.

davidrohr:
Shouldn't we generally disable `-Wpre-c++17-compat` ? We don't need pre c++17 compatibility? Same for `-Wpre-c++14-compat` etc.

MichaelLettrich:
@ktf  this should be it. 
As discussed with Timo, we would introduce the  `ALIBUILD_O2_WARNINGS` environment variable, which enables building O2 with a selected set of warnings, but not trearing them as errors. 
I enabled building with warnings for the CI, just to make sure it works on all platforms, I will remove the last commit before merging.

MichaelLettrich:
@ktf the CI passed and the warnings generated by Clang were comparable to the ones generated by GCC. As requested by you, warnings are disabled by default and are enabled by setting the environment variable `ALIBUILD_ENABLE_WARNINGS`. This is ready to merge. Please review and merge if you think this is OK.

Note: As `ALIBUILD_ENABLE_WARNINGS` is disabled by default, there will be no visible changes to the CI, so you can merge right away.

PR #8195, opened 2022-02-21
CCDB Api: avoid adding two '/' in retrieveBlob with preservePath==false

knopers8:


knopers8:
another separator is added in line 602

PR #8198, opened 2022-02-21
DPL: stop timer after EoS has been reached

ktf:


ktf:
@davidrohr

davidrohr:
I have repeated my EoS tests with this PR, and to me it seems that with this PR the `--exit-transition-timeout 20` option does not work at all any more.
What I am seeing is, when I send a STOP command (same problem for all processes of the sync workflow):
- There is an empty printout, e.g. `[19073:tpc-entropy-encoder]: [21:19:36][INFO]`
- Afterwards the process is stuck. Neither does it terminate after 20 seconds (which it did before), nor does it terminate when it receives an EoS.

For reference, the workflow I was using for tests is:
```
ARGS_ALL_EXTRA="--completion-policy=wait --exit-transition-timeout 20"  WORKFLOW_DETECTORS=TPC SYNCMODE=1 SHMSIZE=$((16<<30)) EXTINPUT=1 $O2_ROOT/prodtests/full-system-test/dpl-workflow.s
```

PR #8199, opened 2022-02-22
DPL: additional tracing / debug statements

ktf:
Now that we have an easy way to selectively enable debug messages, we can add more of them to simplify debugging.


ktf:
@TimoWilken I do not see any status update since over 3 ours here. Maybe an issue with the PR in ali-bot we just merged?

TimoWilken:
@ktf you're right. Fixed now! The statuses should be updated pretty quickly as the builds have already happened.

ktf:
@davidrohr @sawenzel is the error with the full system test known? Or maybe we are building without NDEBUG...

ktf:
Actually I see the error elsewhere as well.

PR #8200, opened 2022-02-22
ITS-CPU: Use o2::its::Timeframe as memory interface for seeding vertexer

mconcas:
First of a series of PRs that aims at using a uniform interface to data for both ITS PV seeding and tracking.
Tests to be performed before promoting to PR:
- [x] Test multiplicity cuts on Nclusters and Ncontributors
- [x] Fix CookedTracker


mconcas:
Hi @shahor02,
this pr is now ready to be reviewed.
It introduces a change in the approach that makes the primary seeding vertexer to process the whole Timeframe in one go.
The reason behind that is to have a common memory interface for both ITS reco steps, that should remove some for loops and making it potentially more efficient, also it will be very convenient for next GPU developments.

It also includes a fix in the tracklet selection that was leading to include **in some rare cases** (at least in _pp_) some duplicates of the same tracklet, thus leading to instabilities in the vertex calculation (DCA point of two parallel lines goes beyond the `float` domain in any of the x,y,z coordinates).
This effect made the workflow to not give the same result upon re-launching. Admittedly I was surprised not to observe the same effect before, as this was already there since years. 

Hi @iouribelikov,
I also adapted the CookedTracker to use it, I tested it and seems to work, but 4 eyes are better than 2...

iouribelikov:
@mconcas  Should not something be changed also in these two macros ?
https://github.com/AliceO2Group/AliceO2/blob/dev/macro/run_trac_ca_its.C
https://github.com/AliceO2Group/AliceO2/blob/dev/macro/run_trac_its.C

mconcas:
Ah, sure. I forgot...

Detectors/ITSMFT/ITS/tracking/include/ITStracking/ClusterLines.h
```diff
@@ -209,11 +181,28 @@ GPUhdi() void Line::getDCAComponents(const Line& line, const float point[3], flo
   destArray[4] = std::sqrt(destArray[3] * destArray[3] + destArray[5] * destArray[5]);
 }
 
-///
+inline bool Line::operator==(const Line& rhs) const
+{
+  bool val;
+  for (int i{0}; i < 3; ++i) {
+    val = this->originPoint[i] != rhs.originPoint[i] ? false : true;
+  }
```
shahor02: shoud not this be:
```
  bool val = true;
  for (int i{0}; i < 3; ++i) {
    val &= this->originPoint[i] != rhs.originPoint[i] ? false : true;
  }
```

Detectors/ITSMFT/ITS/tracking/include/ITStracking/ClusterLines.h
```diff
@@ -209,11 +181,28 @@ GPUhdi() void Line::getDCAComponents(const Line& line, const float point[3], flo
...
+  return val && this->isEmpty == rhs.isEmpty;
+}
+
+inline bool Line::operator!=(const Line& rhs) const
+{
+  bool val;
+  for (int i{0}; i < 3; ++i) {
+    val = this->originPoint[i] != rhs.originPoint[i] ? true : false;
```
shahor02: idem

Detectors/ITSMFT/ITS/tracking/include/ITStracking/ClusterLines.h
mpuccio: or even just `val &= this->originPoint[i] == rhs.originPoint[i];`

mconcas: Thanks

iouribelikov:
@mconcas   The test workflows seem to work fine.
But both run_trac_*its.C macros break with "segmentation violation", for me.
( I can not yet say why exactly. )

iouribelikov:
@mconcas   The test workflows seem to work fine.
But both run_trac_*its.C macros break with "segmentation violation", for me.
( I can not yet say why exactly. )

iouribelikov:
@mconcas  Sorry, Matteo.  I pushed a wrong button.  Re-open !

mconcas:
> @mconcas The test workflows seem to work fine. But both run_trac_*its.C macros break with "segmentation violation", for me. ( I can not yet say why exactly. )

I'll have a look, I admit I just checked the compilation not the usage.

mconcas:
> @mconcas The test workflows seem to work fine. But both run_trac_*its.C macros break with "segmentation violation", for me. ( I can not yet say why exactly. )

I'll have a look, I admit I just checked the compilation not the usage.

mconcas:
> @mconcas The test workflows seem to work fine. But both run_trac_*its.C macros break with "segmentation violation", for me. ( I can not yet say why exactly. )

Fixed now :) @iouribelikov

iouribelikov (dismissed):
@mconcas  Thank you, Matteo.  Now, it seems to work fine.

shahor02:
@mconcas my https://github.com/AliceO2Group/AliceO2/pull/8305 has created conflict here, will send a patch soon.

mconcas:
> @mconcas my https://github.com/AliceO2Group/AliceO2/pull/8305 has created conflict here, will send a patch soon.

I can also rebase and fix the conflicts myself, no worries

shahor02:
let me do this, the macros are still using dictionary from the file, will fix this also

shahor02:
I've opened https://github.com/mconcas/AliceO2/pull/30 in your fork

mconcas:
@shahor02 I think I fixed all the glitches. Now it seems that github is having issues in the web interface. Last time I checked the CI we had false positives in macOS build and usual token bug in fullCI. Indeed I would not merge before seeing the CI results.

As a side note: I was wondering whether this PR should be kept outside any version of O2 to be used for the anchored MC for pilot beam pass3, so to keep consistency between reconstruction code.

shahor02:
Hi @mconcas 
Yes, the github is unstable now, I guess macOS CI problems are related to this. I care more about the fullCI since your PR touches the reco. I've asked @TimoWilken to check why the alien-token has disappeared, let's wait for it to pass.
For the effect on MC vs apass3: does it change reconstruction results? I don't see how we can avoid having it used in the MC which will be done with the latest tag (I guess). The only way I see is to put its merging on hold.

shahor02:
Hi @mconcas 
Yes, the github is unstable now, I guess macOS CI problems are related to this. I care more about the fullCI since your PR touches the reco. I've asked @TimoWilken to check why the alien-token has disappeared, let's wait for it to pass.
For the effect on MC vs apass3: does it change reconstruction results? I don't see how we can avoid having it used in the MC which will be done with the latest tag (I guess). The only way I see is to put its merging on hold.

mconcas:
> Hi @mconcas Yes, the github is unstable now, I guess macOS CI problems are related to this. I care more about the fullCI since your PR touches the reco. I've asked @TimoWilken to check why the alien-token has disappeared, let's wait for it to pass. For the effect on MC vs apass3: does it change reconstruction results? I don't see how we can avoid having it used in the MC which will be done with the latest tag (I guess). The only way I see is to put its merging on hold.

> does it change reconstruction results?

It fixes a small bug in the vertexer that I would not expect to impact too much the reconstruction, especially given the misaligment we have at the moment. So perhaps, since this should go in a "better" direction, it is worthwhile to have it already in the anchored MC run. My doubt was more on the theoretical plane, although I guess this will not be the only part in the reconstruction that changed since the apass3.

This said, in my local tests I don't see anything weird, I would not put this on hold as the upcoming developments I am working on depends on this, so let's merge it as soon as fullCI is green.

mconcas:
Hi @TimoWilken,
any update on the fullCI state? Thanks a lot for your effort!

TimoWilken:
Hi @mconcas, the machine that ran the fullCI for this PR had a configuration issue that was fixed earlier this morning. I've restarted the build; the next rebuild should have an alien token.

mconcas:
> Hi @mconcas, the machine that ran the fullCI for this PR had a configuration issue that was fixed earlier this morning. I've restarted the build; the next rebuild should have an alien token.

Thanks!

mconcas:
@shahor02 we finally have the green lantern

PR #8203, opened 2022-02-22
Fix TRD tracking in standalone version for TFs

martenole:
Before the tracker could not handle a TF with multiple collisions. I did not notice it so far, since I only tested with the box generator and a single event..
@davidrohr for the GPU track type I need a time estimate, similar to the `TrackTPC::getTime0()`. Directly in `GPUTPCGMMergedTrack` I could not see it. I assume I need to translate the track's Z into a time coordinate (using a vertex estimate and track eta?) in order to find the matching TRD trigger times. Can I do this using the default TPC vDrift value? And then open the road rather wide in Z accordingly?

There are some unrelated minor code changes in some of the TRD libraries in this PR, since as I learned today we might enable some additional compiler warnings in the future (`-Wreorder` for example).

### DIFF
--- a/GPU/GPUTracking/Global/GPUChainTrackingTRD.cxx
+++ b/GPU/GPUTracking/Global/GPUChainTrackingTRD.cxx
@@ -37,6 +37,13 @@ int GPUChainTracking::RunTRDTracking()
   if (mIOPtrs.nTRDTracklets == 0) {
     return 0;
   }
+
+  bool isTriggeredEvent = (param().par.continuousMaxTimeBin == 0);
+
+  if (!isTriggeredEvent) {
+    Tracker.SetProcessPerTimeFrame(true);
+  }
+
   Tracker.SetGenerateSpacePoints(mIOPtrs.trdSpacePoints == nullptr);
 
   mRec->PushNonPersistentMemory(qStr2Tag("TRDTRACK"));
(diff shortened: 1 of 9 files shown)


GPU/GPUTracking/Global/GPUChainTrackingTRD.cxx
```diff
@@ -37,6 +37,7 @@ int GPUChainTracking::RunTRDTracking()
   if (mIOPtrs.nTRDTracklets == 0) {
     return 0;
   }
+  Tracker.SetProcessPerTimeFrame(true);
```
davidrohr: Do we want to enable this unconditionally? I mean, there is still the possibility that we process triggered events.
I think you should check that `GPUParam::par.continuousMaxTimeBin != 0` for time frames, = 0 means triggered event.

GPU/GPUTracking/Global/GPUChainTrackingTRD.cxx
```diff
@@ -52,7 +53,7 @@ int GPUChainTracking::RunTRDTracking()
       if (!Tracker.CheckTrackTRDCandidate(trktrd)) {
         continue;
       }
-
+      GPUWarning("It does not make sense to load input track without time window information, except when processing Run 2 data");
```
davidrohr: Also this should depend on whether we process a time frame or not.
The conversion to O2 tracks gets the time0 as:
```
time0 = tracks[i].GetParam().GetTZOffset();
```
The problem is that the deltaTFwd and deltaTBwd is not so obvious.
For this the tracker iterates over the TPC clusters. Perhaps for now you can just set 1 TPC drift time? That should at least work. And this mode of operation is anyway mostly for tests, the default will be using the o2 track I guess.

davidrohr (changes requested):

GPU/GPUTracking/Global/GPUChainTrackingTRD.cxx
martenole: Thanks I did not know that this option already existed. So I am using it now

GPU/GPUTracking/Global/GPUChainTrackingTRD.cxx
martenole: Does `GetTZOffset()` have a similar precision as the time of the `TrackTPC`? At least when running with a single thread and looking at the times of the GPU and O2 track types they did not really match (for the same track index). And then in O2 I check if the track is on the A/C side or if it crosses the CE. In the latter case the time should be precise and I check only one TRD trigger time. Trying to do the same now with the GPU track types I find much less tracks than with the O2 track version (roughly factor 10)

martenole: I had the wrong unit for the TPC drift time... This works now as expected

martenole:
ping @davidrohr

PR #8206, opened 2022-02-22
Revert "DPL: stop timer after EoS has been reached"

ktf:
Reverts AliceO2Group/AliceO2#8198


ktf:
Reverting since @davidrohr is saying it breaks things.

PR #8208, opened 2022-02-23
Improved testing of the DPL proxies

matthiasrichter:
The default input and output proxies can not be tested together in the
same workflow configuration because data descriptions on the producer
and consumer side can not be the same. There are now two workflow
configurations:
1. For testing the default input proxy, the producer is opening the
out of band channel directly and the output proxy is skipped.

2. For testing the default output proxy, the input proxy is created
with a custom converter function to adjust data description. A specific
unit test is added with the necessary command line option


ktf:
Issue in fullCI unrelated. Merging.

PR #8212, opened 2022-02-23
new TrackCuts class

alindner14:
header and implementation files for the new TrackCuts class that has to be implemented - something like the AliESDtrackCuts class in Run2; Please let me know if I have to move it elsewhere. Thanks!


alindner14:
Hi @shahor02 . If the fullCI test does not fail from my fault and the selected path for the new class is ok, can you please merge the PR?

PR #8213, opened 2022-02-23
Accesing payload size through DataRefUtils

matthiasrichter:


knopers8 (commented):
The changes look all correct. I wonder though if we should still expect a header and copy it for each part, like below:
https://github.com/AliceO2Group/AliceO2/blob/b9e24981b6ed4935aca1abb418508e59e2ffe908/Utilities/DataSampling/src/Dispatcher.cxx#L101

matthiasrichter:
@knopers8, I confirm that DPL Input record API stays the same, all payloads are exposed as header-payload pairs. For the format with payload sequences, DataRef for every of the payloads will contain the pointer to the common data header. Since Dispatcher is publishing a copy of all payloads through the DPL output API with a new data description, there is now change, if the payload size is retrieved correctly. The latter is done with this PR.

With the new format and also functionality we now have the foundation to further optimize the data flow through the dispatcher, e.g. we could think about an API extension for forwarding payloads without copy using the FairMQ feature to allow read-only instances and avoiding physical copies if possible.

I have not yet finished my tests of the QC in my setup on EPN test node, there have been other issues blocking this. I will resume this on Wednesday.

matthiasrichter:
I have finished testing, and confirm that QC is working as before with this PR also for the new data format.

Details:
Attached `o2-qc` workflow with basic configuration to the readout proxy, the proxy is configured to forward all TPC RAWDATA, the qc runs the skeleton task to histogram the payload size. Configuration has been adjusted to query TPC data in the data sampling policy.
```
  "dataSamplingPolicies": [
    { 
      "id": "tst-raw",
      "active": "true",
      "machines": [],
      "query": "data:TPC/RAWDATA",
      "samplingConditions": [
        { 
          "condition": "nConsecutive",
          "samplesNumber": "10",
          "cycleSize": "100"
        }
      ],
      "blocking": "false"
    }
  ]
```

```
o2-dpl-raw-proxy --session default --severity info --shm-segment-id 0 --shm-segment-size 8589934592  --early-forward-policy noraw --monitoring-backend no-op:// --no-cleanup --fairmq-rate-logging 0 --dataspec "dd:FLP/DISTSUBTIMEFRAME/0;TPCRAW:TPC/RAWDATA" --channel-config "name=readout-proxy,type=pull,method=connect,address=ipc://@stfb-to-dpl,transport=shmem,rateLogging=1" | o2-qc --config json://basic.json --session default --shm-segment-id 
```

StfBuilder test with both formats using the `DATADIST_NEW_DPL_CHAN` switch.

matthiasrichter:
@TimoWilken @ktf, the CI failures are unrelated, in one case there is not even a log. Please restart the CI.

knopers8 (approved):
Great, thank you for confirming

PR #8214, opened 2022-02-23
DPL: add method to rescan DataRelayer / TimesliceIndex

ktf:
This is needed to make sure external events, like OOB channels, or
certain completion policy related ones work correctly.


ktf:
Failure unrelated. Merging.

PR #8215, opened 2022-02-23
[MRRTF-146] MCH: Introduce CSV version of the Bad Channel List

aphecetche:
In addition the MCH calibration code was reorganized / renamed.

For instance the PedestalCalibrator was renamed BadChannelCalibrator as
the calibration output is about bad channel map, not about pedestal.

Some more renaming will have to be performed in stages, in order not to
break the QC : renamed methods are first marked deprecated here in O2,
they will then have to be renamed in QC and finally removed from O2.
In addition some members are temporarily added to match the expectation
from QC, but they will be removed soon.

Also removed custom DsChannelGroup class, which was nothing more than a
vector of DsChannelId (but keep the include for the moment, again not to
break the current QC)

@aferrero2707 the renaming of things (which I believe is for the better) is for the moment a bit of a gymnastic, in order to avoid breaking the current QC build (and/or to have to sync an O2 PR with a QC PR...), but that should clear up afterwards. 

As far as the features are concerned the main changes are : 
- two calibration objects are sent : the previous vector of DsChannelId, plus a "plain" TObjString containing the same information in CSV form, for DCS.
- the calibrator can now be instructed to send its output not only at EOS but also when a given stat is reached (controlled by BadChannelCalibratorParam parameters)


aphecetche:
~one point I forgot to mention : in this PR, the validity range of the calibration objects is not (yet) correct (it's set to 0-infinity simply), as I have yet to understand how to get the timestamp corresponding to the beginning of the slot...~

aphecetche:
@shahor02 assuming the fullCI won't recover anytime soon, would it be possible to merge this one ? We'd need it for next week pedestal calibration tests...

shahor02:
@aphecetche : from the code it looks like only calibration classes are modified, but I am going to run the FST locally, will let you know

shahor02 (approved):
local FST passed, merging

PR #8217, opened 2022-02-24
Sort AOD track daughters indices

shahor02:
@benedikt-voelkel : do you mean this?


benedikt-voelkel:
Hej @shahor02 
Thanks! I reckon, that would be it. I asked also for feedback in O2-2810.

PR #8218, opened 2022-02-24
MCH: add ability to write digits in Root format in the digits-writer

aphecetche:


aphecetche:
@shahor02 another case of unrelated failure I would say : could you please merge it ? Thanks.

Detectors/MUON/MCH/DevIO/Digits/digits-writer-workflow.cxx
```diff
@@ -142,6 +146,7 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
                                ConfigParamSpec::HelpString{"do not expect, in addition to digits and rofs, to get Orbits at the input"});
   workflowOptions.emplace_back(ConfigParamSpec{"input-digits-data-description", VariantType::String, "DIGITS", {"description string for the input digits data"}});
   workflowOptions.emplace_back(ConfigParamSpec{"input-digitrofs-data-description", VariantType::String, "DIGITROFS", {"description string for the input digit rofs data"}});
+  workflowOptions.emplace_back(ConfigParamSpec{"enable-root-output", VariantType::Bool, false, {"output in Root format"}});
```
shahor02, line 149: So far all workflows have global option `--disable-root-output`, which, if activated, simply does not attach the corresponding writer device. Can we have the same approach for the MCH, to be able to use common settings?

aphecetche: well, this `digits-writer` device has been used for debug mainly, with a binary (non Root) format (this PR only adds the ability to use the Root format) . It's not supposed to be used "routinely" I would say.

shahor02 (approved):
ok

PR #8219, opened 2022-02-24
Drop unhandled test requiring timestamps

ktf:


ktf:
Merging since currently breaking all the PRs.

PR #8220, opened 2022-02-24
Revert "Drop unhandled test requiring timestamps"

ktf:
Reverts AliceO2Group/AliceO2#8219


ktf:
@benedikt-voelkel do I understand  correctly that after your fix in O2DPG, this is no longer required?

benedikt-voelkel:
No, it's not. Since the analyses are based on `O2DPG`, the update of https://github.com/AliceO2Group/O2DPG/pull/241 fixes this issue, so you revert here is fine as long as things rely on the latest `O2DPG`. I tested that locally and all analyses pass

benedikt-voelkel:
@ktf This seems also be related to missing alien token. I can confirm that the TPC PID analysis went through in all the recent alibi nightlies

PR #8221, opened 2022-02-24
DPL: automatically include --exit-transition-timeout

ktf:


davidrohr:
I assume this is only for the FLP?
In general, I think we should also enable it for the EPN. Couldn't we make the default depend on the QuitPolicy?
I.e., when the Policy is quit, like in the DPL driver by default, we don't need it.
And when the policy is wait, like with DDS/ECS (I assume it is like that), we have the timeout by default?
Is the waiting for EoS fixed in the meantime?

ktf:
Mmm.. No, the GUI uses the wait policy as well. What really makes the difference is being run alone vs being run with the driver. Maybe I should add way to notify that to the executable.
The issue with the EoS is still there, because the readout-proxy is not really the first device in the chain (because the channel is managed out of bound) so I am still trying to see how to handle that.

knopers8 (commented):
In principle the PR seems correct, but won't it put `--exit-transition-timeout` twice if it is also added to a workflow command explicitely?

ktf:
Closing this in favour of #8245.

ktf:
@knopers8 as discussed elsewhere.

knopers8 (dismissed):
thank you

ktf:
Test fixed and tested locally. Merging to get it in the build.

PR #8222, opened 2022-02-25
BadMap calibrator added; get calibr. from CCDB added

peressounko:
- Add bad map calculation from several sources
- Read calibration and bad map from CCDB in clusterization
- Switch to avoid unfolding of huge corrupted clusters


Detectors/PHOS/calib/src/PHOSBadMapCalibDevice.cxx
```diff
@@ -9,372 +9,414 @@
...
+    if (mMode == 2) { // Pedestals
+      kind = "PHS/Calib/BadMapPed";
     }
-    clu.addDigit(absId, energy, 0., 0, 1.);
-    digitIt++;
-  }
-  if (digitIt == digitEnd && clu.getMultiplicity() == 0) {
-    return false;
+    std::map<std::string, std::string> md;
+    o2::ccdb::CcdbObjectInfo info(kind, "BadMap", flName, md, mRunStartTime, 99999999999999);
```
shahor02: Do you really need an infinite validity? Can it be more reasonable?

shahor02 (changes requested):
Hi @peressounko 
could you fix the formatting error reported by the fullCI?

Detectors/PHOS/calib/src/PHOSBadMapCalibDevice.cxx
peressounko: Idea is to calculate bad map from current time till the next calculation. For some methods (pedestal, chi2) this can be rare: once per week or even month, for occupancy ~ once per fill. I assume, that after next iteration a more fresh object will overwrite current one, but it is hard to predict when it happen.

shahor02:
@peressounko could you set the validity to the largest reasonable coverage, e.g. month or 2? In principle, infinite validity should be allowed only for the default objects. Costin asked to minimize amount of objects matching to given timestamp.

Now there is a clang-format problem.

peressounko:
Dear @shahor02, I changed validity range to 1 year. I don't think we should reduce it further. Clang errors should be fixed now

peressounko:
fullCl failed probably because of [18197:CPVDigitizer]: [10:12:51][ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch/download/9b4a8fe2-9638-11ec-982e-0aa202a21b9a

shahor02:
Hi @peressounko 
I run locally FST with your PR and get `o2-phos-digi2raw --file-for link -o raw/PHS had nonzero exit code 134`
```
[INFO] [RawWriter] getting calibration object from ccdb
[INFO] Is alien token present?: 1
[FATAL] [RawWriter] can not get calibration object from ccdb
For later analysis we write a core dump to core_dump_2427107
Aborted (core dumped)
```
Looking on the code of https://github.com/peressounko/AliceO2/blob/08a9be49a307f910de09dc5281fd969a797fd665/Detectors/PHOS/simulation/src/RawWriter.cxx#L68-L87, I don't see how cannot fail, since if the ccdb reading is requested, you are initializing the CCDBManager but even don't try to read the object. And, actually, you don't have any object in the 
http://alice-ccdb.cern.ch/browse/PHS/Calib

peressounko:
Dear @shahor02, I fixed RawWriter and other similar places. We would like to enter today's FLP suit reliase, could you please have a look and approve if possible?
   Best regards, Dmitri

shahor02 (approved):
Hi @peressounko 
will merge once CI is ok. In general, please consider moving DPL CCDB fetcher instead of using local ccdb manager.

PR #8223, opened 2022-02-25
Option to write ITS/MFT raw dump to custom directory

shahor02:
o2-itsmft-stf-decoder-workflow --raw-data-dumps-directory <dir=""> ...


shahor02:
tests (except fullCI failing due to the absent alien-token) were passed, merging after rebase.

PR #8224, opened 2022-02-25
Enable PHS clusterizer in dpl-workflow if QC is requested

shahor02:


shahor02:
Tests were passed (FST locally, FullCI fails due to the alien-token), merging after rebase

PR #8229, opened 2022-02-25
[EMCAL-630] Make decoder errors optional

mfasDa:
Add option to disable decoding errors
in case no consumer workflow is added
(mostly QC)


mfasDa:
Hi @shahor02, on the EPN we have a problem that the TF buffer is not released. After discussion with Gvozden we suspect the decoding errors to the the reason, because they are already created but the decoding error QC component is not yet ready, so the messages are not yet consumed. For this purpose we decided to add a switch to disable the decoding errors. Does the PR look reasonable to you?

shahor02:
Hi @mfasDa 
Looks reasonable.

mfasDa:
I tested locally that the reco works still fine. @shahor02 can you please merge once the CI finishes? FullCI unrelated.

shahor02:
sure

PR #8230, opened 2022-02-25
Hmpid fix decoder

fapfap69:


fapfap69:
Please, @shahor02 , could you verify the CI error, it isn't relate to HMPID code, but CPVDigitizer. 
I would like, if possible, this change to go into the next suite, thank you.

shahor02 (approved):
local FST test passed, merging

PR #8236, opened 2022-02-28
DPL: fix more than one attribute in query.

ktf:


ktf:
Passes enough tests, currently not working as advertised. Merging.

PR #8237, opened 2022-02-28
DPL GUI: improve color for debug messages

ktf:


ktf:
Compiles fine on  mac and CS8. Not really tested. Merging.

PR #8240, opened 2022-02-28
Avoid race condition in channel names

ktf:


ktf:
I suspect what is happening for the `failed to attach channel from_eta-and-cls-histograms_to_internal-dpl-aod-global-analysis-file-sink[0] (bind)` is that the jobs run in containers and they get assigned the same port ranges inside the container, however the unix domain socket end up having the same `@<hostname>_<uniqueport>` set of names, resulting a clash across containers. If that is the case, it might actually be a limitation / configuration of the container runtime...

PR #8242, opened 2022-03-01
DPL: honour correctly ccdb-run-dependent metadata

ktf:
Checking if the metadata is there is not enough.


ktf:
@shahor02

shahor02 (approved):
Thanks, fixes the problem.

ktf:
Currently broken. Merging.

PR #8243, opened 2022-03-01
[MFT] Add chipID mapping and updateL2GMatrixCache 

robincaron13:
- ChipID mapping from readout to geometry `mChipIDGeoToRO` is needed in the construction of alignable volume
- Method to update L2G matrices `updateL2GMatrixCache` update the cluster matrix cache, and it is used in the prealignment process
- Some fixes are also included (especially one missing element of the `ChipID2Layer` array)


rpezzi (approved):

shahor02:
test FST locally, merging.

PR #8245, opened 2022-03-01
Adapt exit timeout

ktf:


ktf:
@vascobarroso @knopers8 @teo @davidrohr @shahor02 timeout should now be 20s when running inside DDS / ECS.

PR #8250, opened 2022-03-02
MCH: fixed computation of ROF orbit for out-of-time digits

aferrero2707:
The commit fixes a bug in the computation of the ROF orbit in the case of digits that belong to the time-frame preceding the one being processed.


aphecetche (approved):

aphecetche:
@shahor02 @kft this one is another victim of the fullCI (unrelated) failure. Could it be merged ?

PR #8254, opened 2022-03-02
Optimising IDC processing

matthias-kleiner:
Optimising distribute IDCs spec
- first TF can be detected automatically by setting --firstTF -1
- adding possibillity to set an offset for skipping first TFs in case of automatically detecting the first TF
- skipping CRUs in case they are received, but not specified in input list
- adding check for not received CRUs
- generally avoiding crashes, instead printing fatal log messages

IDC factorization
- fixing IDC factorization in case not all CRUs are used

IDCs on FLPS:
- Automatically loading the pad status map on the FLPS from the CCDB in case the map was updated

CCDB:
- Use precise time stamp when writting IDCs and pad status map to the CCDB

### DIFF
--- a/Detectors/TPC/workflow/include/TPCWorkflow/TPCDistributeIDCSpec.h
+++ b/Detectors/TPC/workflow/include/TPCWorkflow/TPCDistributeIDCSpec.h
@@ -44,7 +44,7 @@ class TPCDistributeIDCSpec : public o2::framework::Task
 {
  public:
   TPCDistributeIDCSpec(const std::vector<uint32_t>& crus, const unsigned int timeframes, const unsigned int outlanes, const bool loadFromFile, const int firstTF)
-    : mCRUs{crus}, mTimeFrames{timeframes}, mOutLanes{outlanes}, mLoadFromFile{loadFromFile}, mProcessedCRU{{std::vector<unsigned int>(timeframes), std::vector<unsigned int>(timeframes)}}, mDataSend{std::vector<bool>(timeframes), std::vector<bool>(timeframes)}, mTFStart{{firstTF, firstTF + timeframes}}, mTFEnd{{firstTF + timeframes - 1, mTFStart[1] + timeframes - 1}}
+    : mCRUs{crus}, mTimeFrames{timeframes}, mOutLanes{outlanes}, mLoadFromFile{loadFromFile}, mProcessedCRU{{std::vector<unsigned int>(timeframes), std::vector<unsigned int>(timeframes)}}, mDataSent{std::vector<bool>(timeframes), std::vector<bool>(timeframes)}, mTFStart{{firstTF, firstTF + timeframes}}, mTFEnd{{firstTF + timeframes - 1, mTFStart[1] + timeframes - 1}}
   {
     // sort vector for binary_search
     std::sort(mCRUs.begin(), mCRUs.end());
@@ -190,9 +190,9 @@ class TPCDistributeIDCSpec : public o2::framework::Task
     }
 
     // check if all CRUs for current TF are already aggregated and send data
-    if ((mProcessedCRU[currentBuffer][relTF] == 2 * mCRUs.size() || mLoadFromFile) && !mDataSend[currentBuffer][relTF]) {
+    if ((mProcessedCRU[currentBuffer][relTF] == 2 * mCRUs.size() || mLoadFromFile) && !mDataSent[currentBuffer][relTF]) {
       LOGP(info, "All data for current TF received. Sending data...");
-      mDataSend[currentBuffer][relTF] = true;
+      mDataSent[currentBuffer][relTF] = true;
       ++mProcessedTFs[currentBuffer];
       sendOutput(pc, currentOutLane, currentBuffer, relTF);
     }
@@ -209,7 +209,7 @@ class TPCDistributeIDCSpec : public o2::framework::Task
 
       mProcessedTFs[currentBuffer] = 0; // reset processed TFs for next aggregation interval
       std::fill(mProcessedCRU[currentBuffer].begin(), mProcessedCRU[currentBuffer].end(), 0);
-      std::fill(mDataSend[currentBuffer].begin(), mDataSend[currentBuffer].end(), false);
+      std::fill(mDataSent[currentBuffer].begin(), mDataSent[currentBuffer].end(), false);
 
       // set integration range for next integration interval
       mTFStart[mBuffer] = mTFEnd[!mBuffer] + 1;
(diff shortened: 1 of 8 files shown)


Detectors/TPC/workflow/include/TPCWorkflow/TPCDistributeIDCSpec.h
```diff
@@ -102,47 +116,100 @@ class TPCDistributeIDCSpec : public o2::framework::Task
   void run(o2::framework::ProcessingContext& pc) final
   {
     // check which buffer to use for current incoming data
-    const auto tf = getCurrentTF(pc);
+    const auto tf = processing_helpers::getCurrentTF(pc);
+
+    // automatically detext firstTF in case firstTF was not specified
```
wiechula:
```suggestion
    // automatically detect firstTF in case firstTF was not specified
```

Detectors/TPC/workflow/src/ProcessingHelpers.cxx
```diff
@@ -61,3 +63,11 @@ uint64_t processing_helpers::getCreationTime(o2::framework::ProcessingContext& p
 {
   return DataRefUtils::getHeader<DataProcessingHeader*>(pc.inputs().getFirstValid(true))->creation;
 }
+
+uint64_t processing_helpers::getTimeStamp(o2::framework::ProcessingContext& pc, o2::ccdb::BasicCCDBManager& ccdbManager)
+{
+  const auto tfOrbitFirst = DataRefUtils::getHeader<o2::header::DataHeader*>(pc.inputs().getFirstValid(true))->firstTForbit;
+  const auto* tv = ccdbManager.getForTimeStamp<std::vector<Long64_t>>("CTP/Calib/OrbitReset", processing_helpers::getCreationTime(pc));
```
wiechula: I think the `CTP/Calib/OrbitReset` should also come from using something like
```c++
inputSpecs.emplace_back("tpcpadmap", "CTP", "OrbitReset", 0, Lifetime::Condition, ccdbParamSpec("CTP/Calib/OrbitReset"));
```
and then pass the `tv` from the workflow to this function. Using ccdb manager directly from a workflow should not be done any longer.
Perhaps you can check how it is done in the ccdb access in the framework. I could imagine that `CTP/Calib/OrbitReset` is already available default. But I don't know for sure.

Detectors/TPC/workflow/include/TPCWorkflow/TPCDistributeIDCSpec.h
```diff
@@ -166,24 +233,23 @@ class TPCDistributeIDCSpec : public o2::framework::Task
...
   std::array<std::array<std::vector<pmr::vector<float>>, CRU::MaxCRU>, 2> mOneDIDCs{}; ///< 1D IDCs for the whole TPC. CRU -> time frame -> IDCs. Buffer used in case one FLP delivers the TF after the last TF for the current aggregation interval faster then the other FLPs the last TF.
   std::array<std::vector<unsigned int>, 2> mProcessedCRU{};                            ///< counter of received data from CRUs per TF to merge incoming data from FLPs. Buffer used in case one FLP delivers the TF after the last TF for the current aggregation interval faster then the other FLPs the last TF.
-  std::array<uint32_t, 2> mTFStart{};                                                  ///< storing of first TF used when setting the validity of the objects when writing to CCDB
-  std::array<uint32_t, 2> mTFEnd{};                                                    ///< storing of last TF used when setting the validity of the objects when writing to CCDB
+  std::array<std::vector<bool>, 2> mDataSend{};                                        ///< to keep track if the data for a given tf has already been send
```
wiechula:
```suggestion
  std::array<std::vector<bool>, 2> mDataSent{};                                        ///< to keep track if the data for a given tf has already been sent
```

wiechula (changes requested):

Detectors/TPC/workflow/src/ProcessingHelpers.cxx
matthias-kleiner: Thanks for pointing that out. I just fixed it by loading the OrbitReset from the CCDB directly via  setting it as an input spec.

Detectors/TPC/workflow/include/TPCWorkflow/TPCFactorizeIDCSpec.h
```diff
@@ -191,6 +195,13 @@ class TPCFactorizeIDCSpec : public o2::framework::Task
   /// \return returns first time stamp for validity range when storing to IDCDelta CCDB
   auto getFirstTimeStampDeltaIDC(const unsigned int iChunk) const { return mTimeStampRangeIDCDelta[iChunk]; }
 
+  Long64_t getOrbitReset(o2::framework::ProcessingContext& pc) const
+  {
+    auto tv = pc.inputs().get<std::vector<Long64_t>*>("orbitreset");
+    const auto orbitReset = tv->front();
+    return orbitReset;
+  }
+
```
wiechula: shouldn't this also go to the `processing_helpers`?

Detectors/TPC/workflow/include/TPCWorkflow/TPCFactorizeIDCSpec.h
```diff
@@ -203,7 +214,7 @@ class TPCFactorizeIDCSpec : public o2::framework::Task
     auto it = std::find(mTFRangeIDCDelta.begin(), mTFRangeIDCDelta.end(), tf);
     if (it != mTFRangeIDCDelta.end()) {
       const int index = std::distance(mTFRangeIDCDelta.begin(), it);
-      mTimeStampRangeIDCDelta[index] = processing_helpers::getCreationTime(pc);
+      mTimeStampRangeIDCDelta[index] = processing_helpers::getTimeStamp(pc, getOrbitReset(pc)) / 1000;
```
wiechula: I guess the `getOrbitReset(pc)` can also go to `processing_helpers`, then one can simply have `getTimeStamp(pc)` and call the `getOrbitReset(pc)` from within.

Detectors/TPC/workflow/src/ProcessingHelpers.cxx
```diff
@@ -61,3 +62,10 @@ uint64_t processing_helpers::getCreationTime(o2::framework::ProcessingContext& p
 {
   return DataRefUtils::getHeader<DataProcessingHeader*>(pc.inputs().getFirstValid(true))->creation;
 }
+
```
wiechula, line 65: With my comments from above, moving code to `processing_helpers` please also add documentation, that the `InputSpec` for `ccdbParamSpec("CTP/Calib/OrbitReset")` must be defined.

wiechula (changes requested):

Detectors/TPC/workflow/src/ProcessingHelpers.cxx
matthias-kleiner: Yes you are right, I just fixed it and moved the function to the `processing_helpers`

wiechula (approved):

shahor02 (approved):

PR #8255, opened 2022-03-02
DPL: improve InfoLogger support

ktf:


martenole:
Shouldn't you also change
https://github.com/AliceO2Group/AliceO2/blob/30cce95a84020ee4cfe9f3bcacc8ab3c1075e47a/Framework/Core/src/runDataProcessing.cxx#L1027
to
```
("infologger-mode", bpo::value<std::string>()->default_value(defaultInfologgerMode), "O2_INFOLOGGER_MODE override");
```
?

ktf:
Indeed. Fixed and added a proper error message if InfoLogger fails to construct.

PR #8259, opened 2022-03-03
DPL: improve parsing of channel-config

ktf:
Now include parser which allows for creating a new set of `OutputChannelSpec`s. From a a provided --channel-config option string.


ktf:
o2-dataflow test obsolete. Merging.

PR #8263, opened 2022-03-04
Remove obsolete init, use std::set and document all parameters for MID preclustering

dstocco:


aphecetche (dismissed):

dstocco:
@aphecetche I have just rebased to trigger a rebuild of the fullCI which is in red state since 2 weeks now

aphecetche (approved):

PR #8273, opened 2022-03-05
Optionally use ITS/MFT NoiseMap to mask chips in digitization

shahor02:
If the file ITSNoiseMap.root (MFTNosieMap.root) is found on the path `o2::itsmft::DPLDigitizerParam<DETID>.noiseFilePath`,
then it will be used to avoid digitization in the chips which have `map.isFullChipMasked(i) == true`.


shahor02:
All tests were passed, merging after changing info message to debug.

PR #8275, opened 2022-03-05
TPC dEdx: automatically loading calib objects from CCDB

matthias-kleiner:
- adding options for disabling loading calibrations from CCDB
- using the gain map to get the original charge which was used during
  the zero supresison
- apply residual gain map on cluster charge


GPU/GPUTracking/DataTypes/CalibdEdxContainer.h
```diff
@@ -32,6 +32,22 @@
...
 {
 
+/// flags to set which corrections will be loaded from the CCDB
+enum class Calibsdedx : unsigned short {
+  calTopologySpline = 1 << 0,  ///< flag for a topology correction using splines
+  calTopologyPol = 1 << 1,     ///< flag for a topology correction using polynomials
+  calThresholdMap = 1 << 2,    ///< flag for using threshold map
+  calGainMap = 1 << 3,         ///< flag for using the gain map to get the correct cluster charge
+  calResidualGainMap = 1 << 4, ///< flag for applying residual gain map
+  calResidualGain = 1 << 5,    ///< flag for residual dE/dx gain correction
```
wiechula: Naming convention is capital case: https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html?showone=Enumerator_Names#Enumerator_Names
I'm wondering if we should change `calResidualGain` to `CalTimeGain`, to be consisten with the CCDB path.

GPU/GPUTracking/DataTypes/CalibdEdxContainer.h
```diff
@@ -32,6 +32,22 @@
 namespace o2::tpc
 {
 
+/// flags to set which corrections will be loaded from the CCDB
+enum class Calibsdedx : unsigned short {
```
wiechula: Perhaps
```suggestion
enum class CalibsdEdx : unsigned short {
```
to  be consistent with other names.

GPU/GPUTracking/DataTypes/CalibdEdxContainer.h
```diff
@@ -143,32 +169,75 @@ class CalibdEdxContainer : public o2::gpu::FlatObject
   /// \param fileName input file containing the correction
   void loadSplineTopologyCorrectionFromFile(std::string_view fileName);
 
+  /// set the polynomial track topology
+  /// \param calibTrackTopology polynomial track topology correction
+  void setPolTopologyCorrection(const CalibdEdxTrackTopologyPol& calibTrackTopology);
+
+  /// setting a default topology correction which just returns 1
+  void setdefaultPolTopologyCorrection();
```
wiechula: Perhapas
```suggestion
  void setDefaultPolTopologyCorrection();
```

GPU/GPUTracking/dEdx/GPUdEdx.h
```diff
@@ -145,7 +145,8 @@ GPUdnii() void GPUdEdx::fillCluster(float qtot, float qmax, int padRow, unsigned
   const int region = param.tpcGeometry.GetRegion(padRow);
   z = CAMath::Abs(z);
   const float threshold = calibContainer->getZeroSupressionThreshold(slice, padRow, padPos); // TODO: Use the mean zero supresion threshold of all pads in the cluster?
-  const float qTotIn = CAMath::Clamp(qtot, calibContainer->getMinqTot(), calibContainer->getMaxqTot());
+  const float fullGainMapGain = calibContainer->getGain(slice, padRow, padPos);
+  const float qTotIn = fullGainMapGain * CAMath::Clamp(qtot, calibContainer->getMinqTot(), calibContainer->getMaxqTot());
   const float qMaxTopologyCorr = calibContainer->getTopologyCorrection(region, o2::tpc::ChargeType::Max, tanTheta, snp, z, absRelPad, relTime, threshold, qTotIn);
```
wiechula: Perhaps it would be good to add an option to the workflow and `calibContainer` to be able to use the full gain map to correct `qmax` and `qtot` and then, of course, not use it for the `qTotIn`. This would e.g. be needed in case the full gain map was not applied in the cluster finder. As e.g. for the pilot beam data.

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -376,6 +406,55 @@ DataProcessorSpec getGPURecoWorkflowSpec(gpuworkflow::CompletionPolicyData* poli
...
+        updateCalibs = true;
+      }
+
+      if (updateCalibs) {
+        // updating the calibration object
+        GPUCalibObjectsConst newTopologyCalib;
+        newTopologyCalib.dEdxCalibContainer = dEdxCalibContainer;
+        auto& tracker = processAttributes->tracker;
+        tracker->UpdateCalibration(newTopologyCalib);
+      }
```
wiechula: I'm not sure this is the best place to trigger the `updateCalibration`, since if several calibrations are updated in the the same TF, this will be called multiple times. Perhaps the `updateCalibs` should be propagated to the `processingFct` and `UpdateCalibration` called just before `RunTracking`.

wiechula (changes requested):
Thanks a lot for implementing this!
Please have a look at my comment.

matthias-kleiner:
> Thanks a lot for implementing this! Please have a look at my comment.

Thanks for the comments. I implemented the suggestions in the latest commit

wiechula (dismissed):
@matthias-kleiner , very nice!

sawenzel:
So can the draft status be removed to trigger testing?

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -656,6 +732,39 @@ DataProcessorSpec getGPURecoWorkflowSpec(gpuworkflow::CompletionPolicyData* poli
...
+
+      if (dEdxCalibContainer->isCorrectionCCDB(CalibsdEdx::CalTopologyPol)) {
+        pc.inputs().get<o2::tpc::CalibdEdxTrackTopologyPolContainer*>("tpctopologygain");
+      }
+
+      if (dEdxCalibContainer->isCorrectionCCDB(CalibsdEdx::CalTimeGain)) {
+        pc.inputs().get<o2::tpc::CalibdEdxCorrection*>("tpctimegain");
+      }
+
+      if (processAttributes->updateCalibs) {
```
davidrohr, line 759: Do I understand correctly that we take this path only when the calib object has actually changed? We will not update with the same object over and over?

davidrohr (dismissed):
Looks good to me. Can you just reply to my comment below?
And I was waiting to approve and merge this until we have the current CCDB issues in the CI and on the EPNs fixed.

GPU/Workflow/src/GPUWorkflowSpec.cxx
matthias-kleiner: Yes this function triggers a check if the CCDB object has changed and if it has the `finaliseCCDB` function will be called.

wiechula (approved):
Fine from my side

PR #8276, opened 2022-03-06
Improve handling of ccdb / condition options

ktf:


Framework/Core/src/WorkflowCustomizationHelpers.cxx
```diff
@@ -41,13 +41,21 @@ namespace o2::framework
 
 std::vector<ConfigParamSpec> WorkflowCustomizationHelpers::requiredWorkflowOptions()
 {
+  auto defaultConditionBackend = std::getenv("DDS_SESSION_ID") ? "http://o2ccdb.internal" : "http://alice-ccdb.cern.ch";
```
shahor02: Should be `http://o2-ccdb.internal`.

ktf:
@TimoWilken I think there is something wrong with the builders. build/O2/o2 has been hanging since yesterday.

davidrohr:
@ktf : I just tried with this PR, but it fails with
```
[47345:internal-dpl-ccdb-backend]: [09:16:39][FATAL] error while setting up workflow in o2-cpv-reco-workflow: missing option: condition-backend (No such node (condition-backend))
```

ktf:
I am looking into it.

PR #8284, opened 2022-03-07
Introduce flag to run FST in in batch mode

oTTer-Chief:
Allow to run FST unattended by setting FST_TMUX_BATCH_MODE=1. Runs quietly if logging is enabled.
Additionally: 
  - improved logging in TMUX mode by showing output and logging to file simultaneously.
  - explicitly check cmd line parameters

Tested locally and via ssh. Somehow only works if a tty is allocated, otherwise only one numa domain starts but I don't see how this is caused by this script.

### DIFF
--- a/prodtests/full-system-test/start_tmux.sh
+++ b/prodtests/full-system-test/start_tmux.sh
@@ -49,15 +49,9 @@ if [ "0$FST_TMUX_KILLCHAINS" == "01" ]; then
 fi
 
 if [ "0$FST_TMUX_LOGPREFIX" != "0" ]; then
-  if [ "0$FST_TMUX_BATCH_MODE" == "01" ]; then
-    LOGCMD0=" &> ${FST_TMUX_LOGPREFIX}_0.log"
-    LOGCMD1=" &> ${FST_TMUX_LOGPREFIX}_1.log"
-    LOGCMD2=" &> ${FST_TMUX_LOGPREFIX}_2.log"
-  else
-    LOGCMD0=" |& tee ${FST_TMUX_LOGPREFIX}_0.log"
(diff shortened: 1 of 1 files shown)


prodtests/full-system-test/start_tmux.sh
```diff
@@ -49,8 +49,15 @@ if [ "0$FST_TMUX_KILLCHAINS" == "01" ]; then
...
-  LOGCMD0=" &> ${FST_TMUX_LOGPREFIX}_0.log"
-  LOGCMD1=" &> ${FST_TMUX_LOGPREFIX}_1.log"
+  if [ "0$FST_TMUX_BATCH_MODE" == "01" ]; then
+    LOGCMD0=" &> ${FST_TMUX_LOGPREFIX}_0.log"
+    LOGCMD1=" &> ${FST_TMUX_LOGPREFIX}_1.log"
+    LOGCMD2=" &> ${FST_TMUX_LOGPREFIX}_2.log"
+  else
+    LOGCMD0=" |& tee ${FST_TMUX_LOGPREFIX}_0.log"
+    LOGCMD1=" |& tee ${FST_TMUX_LOGPREFIX}_1.log"
+    LOGCMD2=" |& tee ${FST_TMUX_LOGPREFIX}_2.log"
```
davidrohr: I had this syntax before, and removed it since it was causing issues. Don't remember exactly what was the issue, need to check. But I would be hesitating to merge it as is without checking before. Might take me some days. If you want to get the rest in before, you can split it in 2 PRs.

oTTer-Chief: I removed it. I tested it once and it worked but there is a good chance I didn't catch the edge case.

PR #8287, opened 2022-03-07
add isPhysicalPrimary()

alindner14:
Hi @chiarazampolli , @shahor02. I added the isPhysicalPrimary() function. Can you please check it? Thank you!


chiarazampolli:
Hello @alindner14 , 

Cannot you use directly the function in https://github.com/AliceO2Group/AliceO2/blob/ae3454c2bc4d3338ac9766d94c125656fcba2061/DataFormats/simulation/include/SimulationDataFormat/MCUtils.h? This is a copy-paste of that code, which is never good practice. 
Also, these functions are general, not specific of the QC task, so they should anyway not go there. 
If MCUtils.h is not good as it is implemented now, then you should, in my opinion, change there. Adding @sawenzel , he might have suggestions.

Chiara

PR #8296, opened 2022-03-09
DPL: add our own signal handler for SEGV and alike

ktf:


davidrohr (changes requested):
Can we do this such that it catches the SEGV etc. only when `O2_NO_CATCHALL_EXCEPTIONS` is not set, otherwise we get the same debugging issues as we had with the exceptions. There, we have put this check before activating exception handling:
```
static bool noCatch = getenv("O2_NO_CATCHALL_EXCEPTIONS") && strcmp(getenv("O2_NO_CATCHALL_EXCEPTIONS"), "0");
```

ktf:
Like it is now, you mean?
For the record, I am not sure this will actually work without any issues, but at least we will be able to tell if the issue with the demangling is due to the fact we do not use the system compiler but use libSegFault from the system.

cc @jgrosseo

davidrohr:
yes, that is what I meant

ktf:
Ok, this seems to work (tested on hyperloop). It will only print the thread which had the exception. To do more than that we can simply fall back to the old libSegFault approach, which I leave for advanced usage.

PR #8298, opened 2022-03-09
DPL: handle hex and binary numbers as subspecs

ktf:


ktf:
tested locally. Merging.

PR #8299, opened 2022-03-09
[DPL Analysis] [Event mixing] Introducing Binning Policy for block combinations

saganatt:
Ciao @ktf,

The BinningPolicy classes replace the special hash task / table for block combinations.

Sorry for a huge PR, but I needed to modify in tests all lines that call block combinations, otherwise it won't compile.
The core changes are: new BinningPolicy.h and rewritten `groupTable()` in ASoAHelpers.h.

### DIFF
--- a/Framework/Core/include/Framework/ASoAHelpers.h
+++ b/Framework/Core/include/Framework/ASoAHelpers.h
@@ -63,10 +63,6 @@ inline bool diffCategory(std::pair<uint64_t, uint64_t> const& a, std::pair<uint6
   return a.first >= b.first;
 }
 
-// TODO: Store types, not labels in BinningPolicy, can be a framework::pack<...>
-// C::type would return uint64_t
-// Pass column type C, not the label, so you have the access to C::type and the rest out of the box
-// C::mColumnIterator.mColumn == arrow::ChunkedArray
 template <typename BinningPolicy, typename T>
 std::vector<std::pair<uint64_t, uint64_t>> groupTable(const T& table, const BinningPolicy& binningPolicy, int minCatSize, int outsider)
 {
@@ -86,7 +82,7 @@ std::vector<std::pair<uint64_t, uint64_t>> groupTable(const T& table, const Binn
     selectedRows = table.getSelectedRows(); // vector<int64_t>
   }
 
-  auto arrowColumns = binningPolicy.getArrowColumns(arrowTable);
+  auto arrowColumns = binning_arrow_helpers::getArrowColumns(arrowTable);
   auto chunksCount = arrowColumns[0]->num_chunks();
   // TODO: Are such checks needed or can we safely assume chunks are always the same?
   for (int i = 1; i < binningPolicy.mColumnsCount; i++) {
@@ -96,7 +92,7 @@ std::vector<std::pair<uint64_t, uint64_t>> groupTable(const T& table, const Binn
   }
 
   for (uint64_t ci = 0; ci < chunksCount; ++ci) {
-    auto chunks = binningPolicy.getChunks(arrowTable, ci);
+    auto chunks = binning_arrow_helpers::getChunks(arrowTable, ci);
     auto chunkLength = std::get<0>(chunks)->length();
     // TODO: Are such checks needed or can we safely assume chunks are always the same?
     for_<binningPolicy.mColumnsCount - 1>([&chunks, &chunkLength](auto i) {
@@ -119,7 +115,7 @@ std::vector<std::pair<uint64_t, uint64_t>> groupTable(const T& table, const Binn
         selInd = selectedRows[ind];
       }
 
-      auto rowData = binningPolicy.getRowData(arrowTable, ci, ai);
+      auto rowData = binning_arrow_helpers::getRowData(arrowTable, ci, ai);
       int val = binningPolicy.getBin(rowData);
       if (val != outsider) {
         groupedIndices.emplace_back(val, ind);
--- a/Framework/Core/include/Framework/ASoAHelpers.h
+++ b/Framework/Core/include/Framework/ASoAHelpers.h
@@ -174,12 +174,12 @@ std::vector<std::pair<uint64_t, uint64_t>> doGroupTable(const T& table, const Bi
     auto chunks = o2::framework::binning_helpers::getChunks(arrowTable, binningColumns, ci);
     auto chunkLength = std::get<0>(chunks)->length();
     // TODO: Are such checks needed or can we safely assume chunks are always the same?
-    constexpr auto cn = binningPolicy.mColumnsCount - 1;
-    for_<cn>([&chunks, &chunkLength](auto i) {
-      if (std::get<i.value + 1>(chunks)->length() != chunkLength) {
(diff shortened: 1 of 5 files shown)


saganatt:
Hi @jgrosseo,

in the next PR I will adjust event mixing interface. With the changes introduced here, event mixing will look like this:
```
std::vector<double> zBins{VARIABLE_WIDTH, -7.0, -5.0, -3.0, -1.0, 1.0, 3.0, 5.0, 7.0};
std::vector<double> yBins{VARIABLE_WIDTH, 0, 5, 10, 20, 30, 40, 50, 101};
PairBinningPolicy<aod::cent::CentEstV0M, aod::collision::PosZ> pairBinning{yBins, zBins};
SameKindPair<aod::Collisions, aod::Tracks> pair{pairBinning, 5, -1};
void process (soa::Join<aod::Collisions, aod::CentV0Ms>>& collisions, aodTracks const& tracks) {
    for (auto& [c1, tracks1, c2, tracks2] : pair) { ... }
}
```
"pair binning" means "calculate bins based on 2 columns". Similarly, there is SingleBinning and TripleBinning. We can think about a better naming, if it is confusing.

What do you think about this approach?

Framework/Core/include/Framework/ASoAHelpers.h
```diff
@@ -82,25 +82,44 @@ std::vector<std::pair<uint64_t, uint64_t>> doGroupTable(const T& table, const st
     selectedRows = table.getSelectedRows(); // vector<int64_t>
   }
 
-  for (uint64_t ci = 0; ci < chunkedArray->num_chunks(); ++ci) {
-    auto chunk = chunkedArray->chunk(ci);
+  auto binningColumns = binningPolicy.getColumns();
+  auto arrowColumns = o2::framework::binning_helpers::getArrowColumns(arrowTable, binningColumns);
+  auto chunksCount = arrowColumns[0]->num_chunks();
+  // TODO: Are such checks needed or can we safely assume chunks are always the same?
```
saganatt, line 166: @ktf?
If I don't need to check the alignments, then I can get rid of `getArrowColumns()` and `getChunks()` altogether.

Framework/Core/include/Framework/BinningPolicy.h
```diff
@@ -0,0 +1,331 @@
...
+
+namespace o2::framework
+{
+
+namespace binning_helpers
+{
+template <typename... Cs>
+std::array<arrow::ChunkedArray*, sizeof...(Cs)> getArrowColumns(arrow::Table* table, pack<Cs...>)
+{
+  static_assert(std::conjunction_v<typename Cs::persistent...>, "BinningPolicy: only persistent columns accepted (not dynamic and not index ones");
```
saganatt, line 29: An important point: one cannot do event mixing / block combinations based on values in dynamic columns. Simply because they do not exist as real columns.

jgrosseo:
Thanks for this work! The interface looks good to me. I let @ktf comment on the technicalities. 

About the naming. So we cannot have just `BinningPolicy` with a different number of template arguments? Then maybe simply `BinningPolicy1` `BinningPolicy2` ... ?

jgrosseo:
It breaks the compilation of O2Physics so we would need to force merge both in parallel. Is there an easy trick to avoid this?

saganatt:
> About the naming. So we cannot have just `BinningPolicy` with a different number of template arguments? Then maybe simply `BinningPolicy1` `BinningPolicy2` ... ?

BinningPolicyBase has a variable number of template arguments. I created Single, Pair, Triple policies so as to override `getBin()`. Perhaps this function could be generalized to any number of input columns, but I would need more time to think how.

On the other hand, are there many (any?) use cases of mixing based on 4 or more event properties?

jgrosseo:
> On the other hand, are there many (any?) use cases of mixing based on 4 or more event properties?

No, 1-3 is sufficient.

saganatt:
> It breaks the compilation of O2Physics so we would need to force merge both in parallel. Is there an easy trick to avoid this?

Ah, right. I changed the selfCombinations parameters. I can write a temporary binning class for backward compatibility but perhaps this is not worthy the effort? I see it is used in 4 places, in connection with event mixing. With my next PR that will need to change anyways.

saganatt:
I added the backward compability.

> About the naming. So we cannot have just BinningPolicy with a different number of template arguments? Then maybe simply BinningPolicy1 BinningPolicy2 ... ?

I implemented a single binning class. But I am curious whether this code looks good (wrt performance) to @ktf ? Or whether old implementation with multiple classes was better?

I also commented lines that caused the problem in macOS, as I am not sure we will need them at the end (see my review comments above, ASoAHelpers.h:88).

Framework/Core/include/Framework/ASoAHelpers.h
```diff
@@ -139,30 +141,129 @@ std::vector<std::pair<uint64_t, uint64_t>> doGroupTable(const T& table, const st
...
+struct is_string<std::basic_string<T, Traits, Alloc>> {
+  static const bool value = true;
+};
+
+template <std::size_t N>
+struct is_string<char[N]> {
+  static const bool value = true;
+};
+} // namespace old_interface
+
```
ktf, line 257: I think you could just use `is_specialization`, no?

Framework/Core/include/Framework/ASoAHelpers.h
```diff
@@ -13,6 +13,8 @@
 #define O2_FRAMEWORK_ASOAHELPERS_H_
 
 #include "Framework/ASoA.h"
+#include "Framework/BinningPolicy.h"
+#include "Framework/Logger.h"
```
ktf, line 17: Please, no Logger.h here.

Framework/Core/include/Framework/ASoAHelpers.h
```diff
@@ -139,30 +141,129 @@ std::vector<std::pair<uint64_t, uint64_t>> doGroupTable(const T& table, const st
...
-  auto columnIndex = table.asArrowTable()->schema()->GetFieldIndex(categoryColumnName);
-  auto dataType = table.asArrowTable()->column(columnIndex)->type();
-  if (dataType->id() == arrow::Type::UINT64) {
-    return doGroupTable<uint64_t, arrow::UInt64Array>(table, categoryColumnName, minCatSize, outsider);
+  arrow::Table* arrowTable = table.asArrowTable().get();
+
+  uint64_t ind = 0;
+  uint64_t selInd = 0;
+  gsl::span<int64_t const> selectedRows;
+  std::vector<std::pair<uint64_t, uint64_t>> groupedIndices;
```
ktf, line 152: I would use a struct rather than a pair.

ktf (approved):
I am afraid the changes are too large for me to review them thoroughly, but the idea looks fine, apart from a few nitpicks that can be fixed later (please make sure you drop Logger.h, in particular). Assuming you tested them and in particular you made sure grouping still works, I am fine with them.

Framework/Core/include/Framework/ASoAHelpers.h
ktf: For sure arrow arrows mixing chunks sizes / number of chunks. I would not assume they are necessarily the same (although it might be in our case so far).

PR #8267, opened 2022-03-04
DPL: automatically subscribe the ccdbBackend to the correct clock

ktf:
* If someone provides the DISTSUBTIMEFRAME message, use it
* If someone provides the analogous for analysis, use it
* If someone requires a timer (and only a timer) use the timer.
* If no one provides the DISTSUBTIMEFRAME message, add it to the
  first device which only gets an enumeration.

PR #8305, opened 2022-03-09
ITS and MFT fetch noise masks and clust.dict. from CCDB

shahor02:
Needs https://github.com/AliceO2Group/AliceO2/pull/8267

Cluster dictionaries used for the pilot beam run are uploaded to http://alice-ccdb.cern.ch/browse/ITS/Calib/NoiseMap as default (same for the MFT).
Empty noise map is uploaded to http://alice-ccdb.cern.ch/browse/ITS/Calib/NoiseMap as default and 2 chip-level only objects are added for specific pilot-beam time ranges.

### DIFF
--- a/Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/ClustererSpec.h
+++ b/Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/ClustererSpec.h
@@ -40,11 +40,15 @@ class ClustererDPL : public Task
   ~ClustererDPL() override = default;
   void init(InitContext& ic) final;
   void run(ProcessingContext& pc) final;
+  void finaliseCCDB(ConcreteDataMatcher& matcher, void* obj) final;
 
  private:
+  void updateTimeDependentParams(ProcessingContext& pc);
+
   int mState = 0;
   bool mUseMC = true;
   bool mPatterns = true;
+  bool mUseClusterDictionary = true;
   int mNThreads = 1;
   std::unique_ptr<std::ifstream> mFile = nullptr;
   std::unique_ptr<o2::itsmft::Clusterer> mClusterer = nullptr;
--- a/Detectors/ITSMFT/MFT/workflow/include/MFTWorkflow/ClustererSpec.h
+++ b/Detectors/ITSMFT/MFT/workflow/include/MFTWorkflow/ClustererSpec.h
@@ -35,11 +35,15 @@ class ClustererDPL : public Task
   ~ClustererDPL() override = default;
   void init(InitContext& ic) final;
   void run(ProcessingContext& pc) final;
+  void finaliseCCDB(ConcreteDataMatcher& matcher, void* obj) final;
 
  private:
+  void updateTimeDependentParams(ProcessingContext& pc);
+
   int mState = 0;
   bool mUseMC = true;
   bool mPatterns = true;
+  bool mUseClusterDictionary = true;
   int mNThreads = 1;
   std::unique_ptr<std::ifstream> mFile = nullptr;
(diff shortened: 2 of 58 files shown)


shahor02:
@ktf it looks like the `finaliseCCDB` callback stopped being called. I was testing this PR with https://github.com/AliceO2Group/AliceO2/pull/8267 cherry-picked, in the log of the FST I see the fetcher reporting query and caching the object, but the code like 
https://github.com/AliceO2Group/AliceO2/blob/597d5dd8483bbfbb3ca01dcb4f05bd02412d26b4/Detectors/ITSMFT/ITS/workflow/src/ClustererSpec.cxx#L127-L141
never acknowledges the callback.

Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/ClustererSpec.h
```diff
@@ -42,9 +42,13 @@ class ClustererDPL : public Task
   void run(ProcessingContext& pc) final;
 
  private:
+  void updateTimeDependentParams(ProcessingContext& pc);
+  void finaliseCCDB(ConcreteDataMatcher& matcher, void* obj);
```
ktf:
```suggestion
  void finaliseCCDB(ConcreteDataMatcher& matcher, void* obj) final;
```

ktf: Don't you need the final / override?

shahor02: @ktf: I do, but adding it does not help, I have also run with the `debug` in https://github.com/AliceO2Group/AliceO2/blob/d4de430a8c943a7de54587456e4b5c46b97c13f0/Framework/Core/include/Framework/Task.h#L104 changed to `info`, it is not invoked.

Detectors/ITSMFT/MFT/workflow/include/MFTWorkflow/ClustererSpec.h
```diff
@@ -37,9 +37,13 @@ class ClustererDPL : public Task
   void run(ProcessingContext& pc) final;
 
  private:
+  void updateTimeDependentParams(ProcessingContext& pc);
+  void finaliseCCDB(ConcreteDataMatcher& matcher, void* obj) final;
```
ktf: It should not be private either.

ktf: I confirm that if it's private it will not work, because the public, default, version has precedence in the lookup I do to see if the function is implemented in the derived class.

shahor02: Thanks, yes, with public it works. I still have problem to get it for all devices where I expect this, but it seems to be related to whole FST workflow getting stuck (due to the PR8275). Now rebuilding with #8275 reverted.

ktf: Cool. I will try to see if there is a way to at least notice if there is a private method and complain.

shahor02:
moving to WIP to fix all instances of loading noise and cluster dictionaries from files.

rpezzi (approved):

shahor02:
o2 CI seems to be stuck, merging to unblock nightly jobs.

`o2-eve-display` and `O2GPUDPLDisplay` still need local ITS,MFT dictionaries, fixing them is more tricky, will do in separate PR.

PR #8311, opened 2022-03-10
Make CMake fatalise if hipify-perl is missing

mconcas:
@davidrohr please let me know if we need something more sophisticated


mconcas:
ping @davidrohr

davidrohr:
@mconcas : I have 2 comments:

Could you use
```
find_program(HIPIFY_EXECUTABLE hipify-perl HINTS [SOME_PATH])
```
to find the executable, instead of using the hardcoded path, and then check if it was found?

second: I think the way you have it right now, you require hipify-perl to be present if `${HIP_PATH}` exists. That breaks the concept of autodetection, i.e. if someone has HIP installed without the hipify, the build will fail. Instead, it should just not use HIP if `ENABLE_HIP` is set to `AUTO`. I propose to move the check here:
```
if(hip_FOUND AND hipcub_FOUND AND rocthrust_FOUND AND rocprim_FOUND AND hip_HIPCC_EXECUTABLE)
```
And perhaps one could adjust the error message
```
"HIP requested but HIP_PATH=${HIP_PATH} does not exist"
```
to show what was not found, since now we are checking for multiple components, so if the check fails it would be helpful to print where it fails.

mconcas:
> @mconcas : I have 2 comments:
> 
> Could you use
> 
> ```
> find_program(HIPIFY_EXECUTABLE hipify-perl HINTS [SOME_PATH])
> ```
> 
> to find the executable, instead of using the hardcoded path, and then check if it was found?
> 
> second: I think the way you have it right now, you require hipify-perl to be present if `${HIP_PATH}` exists. That breaks the concept of autodetection, i.e. if someone has HIP installed without the hipify, the build will fail. Instead, it should just not use HIP if `ENABLE_HIP` is set to `AUTO`. I propose to move the check here:
> 
> ```
> if(hip_FOUND AND hipcub_FOUND AND rocthrust_FOUND AND rocprim_FOUND AND hip_HIPCC_EXECUTABLE)
> ```
> 
> And perhaps one could adjust the error message
> 
> ```
> "HIP requested but HIP_PATH=${HIP_PATH} does not exist"
> ```
> 
> to show what was not found, since now we are checking for multiple components, so if the check fails it would be helpful to print where it fails.

Ok, I restored the autodetection and set up a more granular set of messages in case one or more hip components are missing in the hip-forced scenario.

davidrohr (approved):

PR #8316, opened 2022-03-11
DPL: detect default condition-backend

ktf:
This should at least work around the issue on the EPN while
I try to get the workflow options to behave

### DIFF
--- a/Framework/Core/src/WorkflowHelpers.cxx
+++ b/Framework/Core/src/WorkflowHelpers.cxx
@@ -203,6 +203,15 @@ void WorkflowHelpers::addMissingOutputsToBuilder(std::vector<InputSpec> const& r
   }
 }
 
+// get the default value for condition-backend
+std::string defaultConditionBackend()
+{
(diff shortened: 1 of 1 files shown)


Framework/Core/src/WorkflowHelpers.cxx
```diff
@@ -203,6 +203,15 @@ void WorkflowHelpers::addMissingOutputsToBuilder(std::vector<InputSpec> const& r
   }
 }
 
+// get the default value for condition-backend
+std::string defaultConditionBackend()
+{
+  if (getenv("DDS_SESSION_ID") != nullptr || getenv("OCC_CONTROL_PORT") != nullptr) {
+    return "o2-ccdb.internal";
```
davidrohr: I think it should actually be `http://o2-ccdb.internal`
Can you also add an env variable that I can use to override it, while the command line parsing is not working?

davidrohr (changes requested):

davidrohr:
not sure what is the CI problem, tested locally and on EPN, merging

PR #8318, opened 2022-03-11
DPL: insert timeout to avoid OOM killer when runningh with DDS

ktf:


Framework/Core/src/DDSConfigHelpers.cxx
```diff
@@ -117,6 +116,8 @@ void dumpDeviceSpec2DDS(std::ostream& out,
     }
   }
 
+  float timeout = 0.0;
```
ktf, line 119: @davidrohr this starts from 0, in the chat you had 0.2, was that intended?

davidrohr (approved):
thx, trying on EPN

Framework/Core/src/DDSConfigHelpers.cxx
davidrohr: no, from 0 is fine of course :)

davidrohr (approved):
perfect, works, shall go into tomorrow's nightly

PR #8327, opened 2022-03-12
MCH: DCS DP change from RAW to DPVAL

aphecetche:
also make the DCSDPconfig valid up to infinity (instead of the default 1 day)

@chiarazampolli I hope I interpreted correctly the change that was necessary. 

Note that I've updated the MCH DP config in ccdb-test.cern.ch:8080 with this change

(if that works I'll do the same for MID during Monday's tests)

### DIFF
--- a/Detectors/MUON/Common/src/dcs-ccdb.cxx
+++ b/Detectors/MUON/Common/src/dcs-ccdb.cxx
@@ -109,7 +109,7 @@ void makeCCDBEntryForDCS(const std::string ccdbUrl, uint64_t timestamp)
   DPID dpidtmp;
   for (const auto& a : aliases) {
     auto legitName = o2::muon::replaceDotByUnderscore(a);
-    DPID::FILL(dpidtmp, legitName, o2::dcs::DeliveryType::RAW_DOUBLE);
+    DPID::FILL(dpidtmp, legitName, o2::dcs::DeliveryType::DPVAL_DOUBLE);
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Hi @aphecetche ,

Thanks! I must have overlooked this one, sorry. Yes, it is (hopefully) correct - we will know for sure only after testing, but GRP worked with this. 

Many thanks!

Chiara

Detectors/MUON/Common/src/dcs-ccdb.cxx
```diff
@@ -120,7 +120,10 @@ void makeCCDBEntryForDCS(const std::string ccdbUrl, uint64_t timestamp)
             << o2::muon::subsysname() << " data points to "
             << CcdbDpConfName() << "\n";
 
-  api.storeAsTFileAny(&dpid2DataDesc, CcdbDpConfName(), md, timestamp);
+  uint64_t endOfValidity = 99999999999999;
```
shahor02, line 123: @aphecetche why do you need infinite validity? In this way every next object will overlap will all previous, afaik, this will create overheads. Isn't e.g. 1-2 days validity enough?

chiarazampolli: Hi @shahor02 . 
This is the configuration object, determining which DPs to process. 
Chiara

shahor02 (approved):

PR #8334, opened 2022-03-13
DPL: set correct values in timingInfo

ktf:


ktf:
@shahor02 I guess you need something like that. Doing this would also allow you to rely on DataAllocator::make, rather than creating the messages yourself.

shahor02:
@ktf I've added these lines to https://github.com/AliceO2Group/AliceO2/pull/8332, extending them also to rawTF and CTF readers, closing this PR.

PR #8335, opened 2022-03-13
DPL: make sure proxies set the correct TimingInfo from out of band data.

ktf:


ktf:
@shahor02 this should be what is needed to handle automatically the 0xccdb for the raw-proxy.

PR #8339, opened 2022-03-13
TRD modify verbose output to debug grid

bazinski:
Grid async reconstruction of pilot beam data is dying it seems in rawreader.
It runs fine on the data in question on mine and other desktops and fine on epn.
Cant enable verbose as we lose the logs due to needing to much resources to write the logs in question.
Added a check if the incoming buffer is null, and some more detailed output for when rdh exceed the bounds of the mem.

Not sure how much this will help, will try run on grid now.


tdietel (approved):

bazinski:
That error in something in CPV, or at least appears to be. Not related to code in this PR

bazinski:
@martenole could approve so that we can get this merged ?

PR #8341, opened 2022-03-14
set timestamp in reconstruction

AllaMaevskaya:


mslupeck (dismissed):

AllaMaevskaya:
FT0 should be reconstructed in pass4 with this PR to have it calibrated

shahor02 (dismissed):
Merging, but you should consider switching to DPL CCDB support.

Detectors/FIT/common/calibration/include/FITCalibration/FITCalibrationApi.h
```diff
@@ -80,9 +80,9 @@ FITCalibrationApi::CalibObjWithInfoType FITCalibrationApi::doSerializationAndPre
   CalibObjWithInfoType result;
   auto clName = o2::utils::MemFileHelper::getClassName(calibrationObject);
   auto flName = o2::ccdb::CcdbApi::generateFileName(clName);
+  stopping = 999999999999;
```
shahor02: Actually, before I merge: what is the reason to set inf. end validity for every object? Overlapping objects create overhead in the query. Cannot you define one default object with large validity and upload calibration results with reasonable validity, e.g. end timestamp of the calibration data + some reasonable margin (e.g. 1 day) ?

AllaMaevskaya: The reason of so large validity is that if to use offsets from previous runs if something went wrong.

shahor02: But how old calibration you are ready to accept? I can understand a few days old, otherwise, why don't you put some reasonable default only once?
With the current infinite settings, all your objects will overlap.

AllaMaevskaya: OK, I will set validity as last timestamp + 1 day and upload default calibration with infinite settings - as you suggest

shahor02 (approved):
Thanks! Merging since the tests were passed before trivial change.

PR #8346, opened 2022-03-14
ITS THR SCAN: multiple decoder pipelines + fixed minor bugs

iravasen:
Changes included:

- added possibility to work with multiple decoder pipelines in o2-its-threshold-calib-workflow: now more rows per chip are stored in memory and as soon as a row is completed, thresholds are extracted and the hits of that row are deleted. 
- fixed minor bugs: return value of threshold calculation methods, repetitions in ROOT trees, simplification of finalize(). 
- removed multithreading for sum calculation. Not scaling with multi-thread. 
- Improved timing performance of run() method.


iravasen:
Full CI failing because of missing CCDB object for CPV. Does anyone know anything about this issue?

shahor02 (approved):

PR #8351, opened 2022-03-15
Remove LOGP message to avoid crash with long log texts.

pbuehler:


ktf:
Not sure I get the reason for the crash.

pbuehler:
This LOGP used to print out the contents of a json file - the --aod-writer-json. The entire text of the json was converted to a std::string and provided to LOGP as text to print. When the json file exceeded a given length, the LOGP started to crash.

ktf:
@rbx @dennisklein does this sound familiar? Do we have a limit for the formatting variables in fmt?

rbx:
I've not seen such a limitation. If there is one, it must be in fmt. LOGP passes it through to fmt directly:
https://github.com/FairRootGroup/FairLogger/blob/master/logger/Logger.h#L433

After `fmt::format`, the content is given to `std::ostringstream`, which at the end ends up in `fmt::print`. If there is a custom sink registered, something could go wrong there - it happens before `fmt::print`:
https://github.com/FairRootGroup/FairLogger/blob/master/logger/Logger.cxx#L245-L268

PR #8352, opened 2022-03-15
Add PHOS table

nburmaso:


nburmaso:
This is a draft code for PHOS in the AOD producer. I have tried adapting EventHandler code from EMCAL, but I'm not sure if it is a correct way to access information. @peressounko, could you please check the code and see if it is ok for you?

Also, I have a problem with accessing PHOS cells via RecoContainer -- I will create a JIRA ticket for this

shahor02:
@nburmaso with https://github.com/AliceO2Group/AliceO2/pull/8364 I was to access PHOS data via RecoContainer and run AODProducer from this branch, though did not check the result.

nburmaso:
Thank you @shahor02, I will try running with this fix too

nburmaso:
Hi @shahor02. I have tested this PR with your fix, and it seems it works fine now. Thank you!

@peressounko: I have updated the code to also add trigger information

DataFormats/Detectors/PHOS/src/EventHandler.cxx
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  return outputEvent;
+}
+
+template <class CellInputType>
+bool EventHandler<CellInputType>::compareInteractionRecords(const InteractionRecord& lhs, const InteractionRecord& rhs) const
+{
+  return lhs.bc == rhs.bc && lhs.orbit == rhs.orbit;
+}
```
shahor02, line 157: just for the record: InteractionRecord has operator==, so one can do directly `lhs==rhs`

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1574,7 +1568,15 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   if (mInputSources[GIndex::EMC]) {
     // fill EMC cells to tables
     // TODO handle MC info
-    fillCaloTable(caloEMCCells, caloEMCCellsTRGR, caloCellsCursor, caloCellsTRGTableCursor, bcsMap);
+    auto* caloEventHandler = new o2::emcal::EventHandler<o2::emcal::Cell>();
+    fillCaloTable(caloEventHandler, caloEMCCells, caloEMCCellsTRGR, caloCellsCursor, caloCellsTRGTableCursor, bcsMap, 1);
+    delete caloEventHandler;
```
shahor02, line 1573: can you avoid new/delete, using instead stack object, i.e.
```
o2::emcal::EventHandler<o2::emcal::Cell> caloEventHandler;
fillCaloTable(&caloEventHandler, caloEMCCells, caloEMCCellsTRGR, caloCellsCursor, caloCellsTRGTableCursor, bcsMap, 1);
```
same for the PHOS. Can be done in different PR, since this one already passed CIs

shahor02 (dismissed):

shahor02:
merging after resolving trivial conflict, all tests were passed.

PR #8353, opened 2022-03-15
[EMCAL-753,EMCAL-763] Redesing DigitsWriteoutBuffer with past/future dequeue

hahassan7:
The digits writeout buffer has been redesigned. The basic data containers of the ringbuffer will be 2 std::dequeues, one for past and one for future. The future dequeue will be initialized with a size 15, the past dequeue will not be initialized with a fixed size but expanded until a constant size of 15 is reached. The entry 0 in the future dequeue always represents the current time. When we sample the digits we fill the dequeue in future direction always with the digit time sample (digit time + delay) relative to the current timesample.

### DIFF
--- a/Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsVectorStream.h
+++ b/Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsVectorStream.h
@@ -1,0 +1,101 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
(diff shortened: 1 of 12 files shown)


mfasDa (dismissed):
The bug has been found in the trigger record storing start and end of a trigger instead of start and size as it is normal for the RangeReference.

mfasDa:
CI errors unrelated. @sawenzel do you think this can be merged?

Detectors/EMCAL/simulation/include/EMCALSimulation/DigitsVectorStream.h
```diff
@@ -0,0 +1,101 @@
...
+  /// Initialize the streamer
+  void init();
+
+  /// Fill all the containers, digits, labels, and trigger records
+  void fill(std::deque<o2::emcal::DigitTimebin>& digitlist, o2::InteractionRecord record);
+
+  /// Getters for the finals data vectors, digits vector, labels vector, and trigger records vector
+  std::vector<o2::emcal::Digit> getDigits() { return mDigits; }
+  std::vector<o2::emcal::TriggerRecord> getTriggerRecords() { return mTriggerRecords; }
+  o2::dataformats::MCTruthContainer<o2::emcal::MCLabel> getMCLabels() { return mLabels; }
```
sawenzel: I think this should be `const&` and `() const` function too.

PR #8355, opened 2022-03-15
DPL Analysis: use subspec for table versioning

aalkin:
@ktf @jgrosseo 
This (partially) enables using subspec as a version for AOD tables. 

- [x] use both binding and subspec when comparing inputs
- [x] define if extended/index/derived tables inherit the version of their constituents


Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -869,7 +869,7 @@ DECLARE_SOA_TABLE_FULL(StoredMcParticles_000, "McParticles", "AOD", "MCPARTICLE"
                        mcparticle::GetProcess<mcparticle::Flags, mcparticle::StatusCode>,
                        mcparticle::IsPhysicalPrimary<mcparticle::Flags>);
 
-DECLARE_SOA_TABLE_FULL(StoredMcParticles_001, "McParticles_001", "AOD", "MCPARTICLE_001", //! MC particle table, version 001
+DECLARE_SOA_TABLE_FULL(StoredMcParticles_001, "McParticles_001", "AOD", "MCPARTICLE_001", 1, //! MC particle table, version 001
```
jgrosseo: Thanks for this PR, @aalkin !
Could we drop in the table name here the _001 (-> "MCPARTICLE") and generate that somewhere from the version number 1 internally when the table is read from the file?

aalkin: The next step would be adjusting the reader, indeed.

jgrosseo:
- add version also for EXTENDED, INDEX table macro
- make backward compatible macros with version 0 (to be removed when O2Physics is adjusted)

jgrosseo:
@aalkin When do you think you can come back to this?

aalkin:
@jgrosseo this now is a working version, although it needs to be tested with special JSON input configurations (where it should work, but I'm not 100% sure it will).

aalkin:
I've tested this with customized writing/reading and it works in the base case.

Framework/Core/src/DataOutputDirector.cxx
```diff
@@ -55,12 +55,15 @@ DataOutputDescriptor::DataOutputDescriptor(std::string inString)
...
+  if (!std::string(tableItems[3]).empty() && std::atoi(std::string(tableItems[3]).c_str()) > 0) {
+    version = std::string{"_"}.append(std::string(3 - std::string(tableItems[3]).length(), '0')).append(std::string(tableItems[3]));
+  }
 
   // get the tree name
   // default tree name is the O2 + table name (lower case)
   treename = tablename;
   std::transform(treename.begin(), treename.end(), treename.begin(), [](unsigned char c) { return std::tolower(c); });
-  treename = std::string("O2") + treename;
+  treename = std::string("O2") + treename + version;
```
jgrosseo, line 66: Does this cover both the reading and writing case?

jgrosseo:
Apart from my question - which I think you answered already above - this is fine with me. @ktf do you want to test it on hyperloop and then merge?

Framework/Core/src/DataOutputDirector.cxx
aalkin: This is for writing, reading is covered by code in `AnalysisDataModelHelpers.cxx`.

ktf (dismissed):
LGTM

PR #8356, opened 2022-03-15
Fix O2-2843: handling empty rofs

mpuccio:


mpuccio:
@shahor02 , I confirm that I obtain the exact same result on MC and that the processing of the cosmics data you gave me does not crash. Ready to be merged as soon as the CI is green.

shahor02 (dismissed):
Thanks! Code looks fine, will merge after CI has passed.

mpuccio:
@shahor02 I think the fullCI is failing for a unrelated issue

shahor02:
The fullCI did not rerun yet after fixing the alien-token. Did you try the reconstruction locally?

mpuccio:
Yes, and I get the same results before and after (I used the simulations from alibi to check, thanks to Benedikt)

shahor02 (approved):
fullCI failure is unrelated

PR #8358, opened 2022-03-15
Fix IB and OB geometry to latest blueprints

mario6829:
This commit fixes some small but important geometry mismatches with latest blueprints. The IB chip position was not centered wrt to the stave, and the FPC and glue volumes were smaller and centered; whereas now the chip volume is centered in the X coordinate while the FPC and glue have the correct transverse dimension and are properly aligned to one end of the stave. Lastly the radial position of the stave of all OB layers were reduced to the correct value.


iouribelikov:
@mario6829  Looks good !   Let me just finish my usual formal checks...

iouribelikov (approved):
All checks pass !

shahor02:
will merge after CI

PR #8360, opened 2022-03-15
TPC interpolation use correct global track index

martenole:
@davidrohr this fixes the crash of the track interpolation. For ITS-TPC-TRD-TOF tracks the index in the `GlobalIDSet` which is filled by the `RecoContainer::getSingleDetectorRefs()` method puts the GID of the global track into the TRD slot, since no standalone TRD track exists. With that index one cannot request the TRD track. One has to use the reference for the ITS-TPC-TRD track instead.
I think at that time of the development there was an issue with the TOF matching which lead to the fact that there were no ITS-TPC-TRD-TOF tracks existing and that is why I did not see it.

### DIFF
--- a/Detectors/TPC/calibration/SpacePoints/src/TrackInterpolation.cxx
+++ b/Detectors/TPC/calibration/SpacePoints/src/TrackInterpolation.cxx
@@ -168,7 +168,7 @@ void TrackInterpolation::interpolateTrack(int iSeed)
   }
   if (gidTable[GTrackID::TRD].isIndexSet()) {
     LOG(debug) << "TRD available";
-    const auto& trkTRD = mRecoCont->getITSTPCTRDTrack<o2::trd::TrackTRD>(gidTable[GTrackID::TRD]);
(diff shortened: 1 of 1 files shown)


davidrohr:
thx, don't have time to double-check but fix seems trivial. Let's get this into tomorrow's build.

martenole:
> thx, don't have time to double-check but fix seems trivial. Let's get this into tomorrow's build.

Fine with me. I have tried with your tmp-fst13 data set and with a small simulation I did myself

martenole:
Hi @davidrohr could you maybe merge this? At least the fullCI is green. And I see that other PRs are also failing because of the track interpolation: https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/AliceO2/8362/f5e48d2b5a3f71475c00bf78308bf059ef328111/build_O2_fullCI/pretty.html
In general the CI seems to be very slow to me at the moment. E.g. https://github.com/AliceO2Group/AliceO2/pull/8170 is waiting for more than two days for the macOS build

shahor02 (approved):
Fixes are trivial, merging

PR #8361, opened 2022-03-15
Restructure init of sim magnetic field

sawenzel:
* move init code to common function and use
  multiple times
* allow field init from CCDB (with --field ccdb)


sawenzel:
@chiarazampolli  @shahor02 : After this, we can fetch the field from CCDB in anchored-sim.

shahor02 (dismissed):
Looks fine. I guess later the same SimFieldUtils can be used to generate a local clone of the GRPMagField, to which ccdb queries should be remapped.

PR #8363, opened 2022-03-15
TrackletTransformer uses DPL CCDB mechanism

martenole:
Hi @shahor02 
with this patch the tracklet transformer uses the DPL CCDB fetching mechanism. The time stamp option has been removed.
Cheers,
Ole

### DIFF
--- a/Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
+++ b/Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
@@ -16,7 +16,6 @@
 #include "DataFormatsTRD/Tracklet64.h"
 #include "DataFormatsTRD/CalibratedTracklet.h"
 #include "DataFormatsTRD/CalVdriftExB.h"
-#include "CCDB/BasicCCDBManager.h"
 
 namespace o2
 {
@@ -26,8 +25,10 @@ namespace trd
 class TrackletTransformer
 {
  public:
-  TrackletTransformer();
+  TrackletTransformer() = default;
   ~TrackletTransformer() = default;
+
+  void init();
 
   float getXCathode() { return mXCathode; }
   float getXAnode() { return mXAnode; }
@@ -39,12 +40,7 @@ class TrackletTransformer
   void setXDrift(float x) { mXDrift = x; }
   void setXtb0(float x) { mXtb0 = x; }
 
-  bool hasCalibration() { return mCalibration != nullptr; }
-
-  void loadPadPlane(int hcid);
-
-  // use 555555 for default calibration values
-  void loadCalibrationParameters(int timestamp);
+  void setCalVdriftExB(const CalVdriftExB* cal) { mCalVdriftExB = cal; };
 
   float calculateY(int hcid, int column, int position);
 
@@ -61,15 +57,15 @@ class TrackletTransformer
   double getTimebin(int detector, double x);
 
  private:
-  o2::trd::Geometry* mGeo;
-  const o2::trd::PadPlane* mPadPlane;
+  Geometry* mGeo{nullptr};
+  const PadPlane* mPadPlane{nullptr};
 
   float mXCathode;
   float mXAnode;
   float mXDrift;
   float mXtb0;
 
-  o2::trd::CalVdriftExB* mCalibration;
+  const CalVdriftExB* mCalVdriftExB{nullptr};
 };
 
 } // namespace trd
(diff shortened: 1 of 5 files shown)


bazinski (dismissed):

shahor02 (dismissed):

Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
```diff
@@ -61,15 +57,15 @@ class TrackletTransformer
...
+  Geometry* mGeo{nullptr};
+  const PadPlane* mPadPlane{nullptr};
 
   float mXCathode;
   float mXAnode;
   float mXDrift;
   float mXtb0;
 
-  o2::trd::CalVdriftExB* mCalibration;
+  const CalVdriftExB* mCalibration{nullptr};
```
tdietel: We will need some other calibrations in this class as well, mainly for PID: the krypton gain tables and the chamber gain. 
I would suggest to rename this member variable in preparation for that.
```suggestion
  const CalVdriftExB* mCalVdriftExB{nullptr};
```

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
```diff
@@ -125,6 +132,7 @@ o2::framework::DataProcessorSpec getTRDTrackletTransformerSpec(bool trigRecFilte
   auto& inputs = dataRequest->inputs;
   inputs.emplace_back("trdtracklets", "TRD", "TRACKLETS", 0, Lifetime::Timeframe);
   inputs.emplace_back("trdtriggerrec", "TRD", "TRKTRGRD", 0, Lifetime::Timeframe);
+  inputs.emplace_back("calibration", "TRD", "CALVDRIFTEXB", 0, Lifetime::Condition, ccdbParamSpec("TRD/Calib/CalVdriftExB"));
```
tdietel: Also here, we should prepare the inclusion of Kr and gain calibrations.

Detectors/TRD/base/src/TrackletTransformer.cxx
```diff
@@ -123,7 +102,8 @@ CalibratedTracklet TrackletTransformer::transformTracklet(Tracklet64 tracklet)
   auto slope = tracklet.getSlopeBinSigned();
 
   // calculate raw local chamber space point
-  loadPadPlane(detector);
+  mPadPlane = mGeo->getPadPlane(detector);
```
tdietel, line 105: I guess this no longer needs to be a member variable.
```suggestion
  auto padPlane = mGeo->getPadPlane(detector);
```

martenole: Well it is used in three other functions which are called from there, so one would need to pass it along every time. I would leave it as it is for now.

Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
martenole: Why would we actually need those PID calibrations in this class? The CalibratedTracklet only stores position information and one direction. I thought this would be covered by vDrift and ExB? But T0 might come at some point, so I will do the naming change nevertheless

martenole:
Hi @TimoWilken would it be possible for you to restart the fullCI test for this PR please? And could I get another approval from @bazinski perhaps? The only thing I changed wrt to previous version of this PR was the naming as suggested by Tom

bazinski (approved):

bazinski:
@martenole just out curiosity, and this is not a request to change, but why not std::shared_ptr ?

martenole:
The calibration objects which are obtained from the CCDB (e.g. `mCalVdriftExB`) use raw pointers. For `mPadPlane` the object is owned by the TRD geometry class. I did not want to change this now, but we could of course also switch to smart pointer in https://github.com/AliceO2Group/AliceO2/blob/e281554a65c23d00803b9459c5defdb5e9e59597/Detectors/TRD/base/include/TRDBase/GeometryBase.h#L49 and https://github.com/AliceO2Group/AliceO2/blob/e281554a65c23d00803b9459c5defdb5e9e59597/Detectors/TRD/base/include/TRDBase/Geometry.h#L33

TimoWilken:
Hi @martenole, the fullCI reran automatically, everything is green now.

shahor02 (approved):

martenole:
Hi @TimoWilken thanks a lot and sorry for the noise. I was not sure if it would be starting again automatically, since it already ran several times.

PR #8368, opened 2022-03-16
Mch dcs fix start validity time

aphecetche:


Detectors/MUON/Common/src/dcs-ccdb.cxx
```diff
@@ -120,7 +120,10 @@ void makeCCDBEntryForDCS(const std::string ccdbUrl, uint64_t timestamp)
             << o2::muon::subsysname() << " data points to "
             << CcdbDpConfName() << "\n";
 
-  api.storeAsTFileAny(&dpid2DataDesc, CcdbDpConfName(), md, timestamp);
+  uint64_t endOfValidity = 99999999999999;
```
shahor02, line 123: Looks like my yesterday comment was lost by the github...
Why do you need infinite end validity? This will make every new object to overlap with all previous, which, afaik, leads to overhead. Is not it enough to add e.g. 1-2 days margin to start-time?

chiarazampolli: Hi @shahor02 , 
I allow myself to jump in :-) I also think I replied to you this morning... Weird. 
Anyway, this is the configuration entry, where they define which DPs to process, it is not expected to change often (if ever).

shahor02: Ah, ok, thanks for pointing this out.

shahor02 (approved):
@chiarazampolli do you have comments on this, shall I merge it?

chiarazampolli:
> @chiarazampolli do you have comments on this, shall I merge it?

This is what we tested yesterday, fine with me to merge. 

Thanks!

PR #8370, opened 2022-03-16
STFDecoder requests cluster dict only if clustering is asked

shahor02:


shahor02:
macOS CI is stuck, merging.

PR #8371, opened 2022-03-16
Add TRD-TOF tracks in subscribed sources

njacazio:
@noferini @shahor02 this should fix the missing TRD-TOF tracks in the AO2D. I can test the first nightlies as soon as this is merged


noferini:
Adding also @chiarazampolli 
Shall we wait for all test passed?

shahor02:
This is not enough, they should be added also in the vertexing. I'll test it locally and open a PR once finish something else.

njacazio:
Understood, closing then

PR #8372, opened 2022-03-16
Fix new xcode

ktf:


ktf:
@shahor02 https://github.com/AliceO2Group/AliceO2/pull/8372/commits/f5b24f76d56caa6c06e0eaefdf01be6b3fd67662 is particularly nasty. Any reason to avoid the move constructor for those classes? Do we have other  similar case, as far as you can tell?

PR #8378, opened 2022-03-17
digitizer: Ability to fetch bcFilling scheme from CCDB

sawenzel:


Steer/DigitizerWorkflow/src/SimReaderSpec.cxx
```diff
@@ -121,9 +123,22 @@ DataProcessorSpec getSimReaderSpec(SubspecRange range, const std::vector<std::st
       mgr.getInteractionSampler().setFirstIR({0, o2::raw::HBFUtils::Instance().orbitFirstSampled});
       mgr.getDigitizationContext().setFirstOrbitForSampling(o2::raw::HBFUtils::Instance().orbitFirstSampled);
 
+      auto setBCFillingHelper = [](auto& sampler, auto& bcPatternString) {
+        if (bcPatternString == "ccdb") {
+          LOG(info) << "Fetch bcPattern information from CCDB";
+          // fetch the GRP Object
+          auto& ccdb = o2::ccdb::BasicCCDBManager::instance();
```
shahor02, line 130: Aren't you planning to move this class to DPL CCDB backend? Then this code cannot be in the init() but should move to run(). In any case, I think one should at least set the ccdb host (the GRPLHCIFs are in the http://alice-ccdb.cern.ch).
Also, with this code you get the object for `now()`, for the anchoring one should pass the run timestamp. 
Since you have already headers creation / firstTFOrbit fixed, this would work automatically with the DPL backend.

sawenzel: Everything is set up correctly for CCDBManager here (in the outside workflow creation) and I will use CCDBManager until the caching from snapshots is working correctly (which I think is not yet the case).

shahor02: OK, I did not find where it is set. What exactly does not work with caching? I am already working with `export ALICEO2_CCDB_LOCALCACHE=<xxx>` to create the cache and with `--condition-backend file://<xxx>` to use it directly (also can be done via `export DPL_CONDITION_BACKEND=file://<xxx>`)

PR #8379, opened 2022-03-17
TPC dEdx: automatically loading calib objects from CCDB

matthias-kleiner:
- adding options for disabling loading calibrations from CCDB
- disable loading of CCDB objects, when calib ist set from file
- using the gain map to get the original charge which was used during the zero supression
- possibillity to apply gain map during the reconstruction of the dE/dx
- 'smooth' zero supression threshold map
- apply residual gain map on cluster charge

- TPC clusterer: automatically loading the gain map from the CCDB


matthias-kleiner:
Hi @davidrohr , can you test if this PR is working now for you? @wiechula has uploaded all the calibrations to the official CCDB

davidrohr:
Yes, sure. This is for the PR I had to revert since it broke the TPC reco workflow?
So I would need to verify that TPC reco workflow is working, right?

matthias-kleiner:
> Yes, sure. This is for the PR I had to revert since it broke the TPC reco workflow? So I would need to verify that TPC reco workflow is working, right?

Yes, that was the reverted PR that broke the reco workflow

wiechula (approved):

davidrohr (approved):
tested and tpc tracking works correctly, can be merged when the CI is passed

PR #8380, opened 2022-03-17
DPL: fatal level also in the case of CCDB objects missing

ktf:


davidrohr (approved):
Gives the correct error for my testcase now, thx

PR #8383, opened 2022-03-17
DPL: handle condition-backend and completion-policy globally

ktf:


ktf:
@davidrohr this should fix the issue with the GUI and the wrong message when clicking on restart and it should make sure the condition-backend is correctly propagated to devices.

PR #8384, opened 2022-03-17
DPL: expire Optional inputs only when a given message is provided

ktf:


ktf:
@davidrohr @shahor02 not tested yet, but at least compiles. Notice that for this to work the optional input must come after the DISTSTF one.

davidrohr:
seems to work, merging to have it in tomorrow's build

shahor02:
> Notice that for this to work the optional input must come after the DISTSTF one.

Could you clarify this point? If you mean the ordering in the specs definition, in many devices this is not the case

davidrohr:
I just tried it in the FST and seemed to be OK. I thought it referred to arrival time, and the optional inputs are the raw data so they should arrive together with the DISTSTF. But @ktf should clarify what exactly he means.

ktf:
i actually meant what ruben meant . will check in a bit the actual details. if that's not the case the "dead beef" will simply not trigger.

davidrohr:
> i actually meant what ruben meant . will check in a bit the actual details. if that's not the case the "dead beef" will simply not trigger.

ok, but that is actually a bit error-prone. Can't this be changed? :)

shahor02:
@ktf cannot you sort the inputs internally? If this is problematic, I can fix in a couple of hours the specs, putting the diststf to the head of the input.

davidrohr:
> I can fix in a couple of hours the specs, putting the diststf to the head of the input.

I am doing this right now to test it.
There is anyway a problem with the new O2, so we anyway need a new build.
Would still be good to check internally first if the distsubtimeframe is there.

davidrohr:
https://github.com/AliceO2Group/AliceO2/pull/8387 should do it

ktf:
> > i actually meant what ruben meant . will check in a bit the actual details. if that's not the case the "dead beef" will simply not trigger.
> 
> ok, but that is actually a bit error-prone. Can't this be changed? :)

Yes, I will change it now. It's just that currently the expiration mechanism is configured as I add the routes to the device. I will do it in a separate step.

PR #8389, opened 2022-03-18
DPL: avoid reading in memory at random

ktf:


ktf:
@davidrohr this might explain the crashes you see.

davidrohr:
thx, this fixes the random crash for me. Merging since it also affects the CI.

PR #8391, opened 2022-03-18
Fix AO2Ds in case of negative first daughter 

mpuccio:
@jgrosseo

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -728,6 +728,9 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
       item = mToStore.find(Triplet_t(source, event, mcDaughterL));
       if (item != mToStore.end()) {
         daughters[1] = item->second;
+        if (daughters[0] < 0) {
+          daughters[0] = daughters[1]; /// Treat the case of first negative label (pruned in the kinematics)
+        }
       } else {
         daughters[1] = daughters[0];
       }


jgrosseo (dismissed):

jgrosseo:
@sawenzel This check will most likely break all ongoing productions (from @mpuccio investigation), it is still the right thing to do...

chiarazampolli:
Hi @jgrosseo , 

Do you mean that from the moment when it is merged, all jobs will fail, not passing the test?
See also my comment in https://alice.its.cern.ch/jira/browse/O2-2820

Chiara

jgrosseo:
This concerns only MC. Probably not all but quite some will fail. At least Max found this in a number of AODs. I do not have answers to your other questions in the ticket...

sawenzel:
I was always in favour of having such checks but rather on the analysis level. (Doing analysis regression tests on the alibi-nightlies). I am not in favour of introducing this now here, as it will

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -728,6 +728,9 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
       if (daughters[0] > daughters[1]) {
         std::swap(daughters[0], daughters[1]);
       }
+      if (daughters[0] < 0 && daughters[1] >= 0) {
+        LOG(fatal) << "Discordant daughters indices (" << daughters[0] << "," << daughters[1] << "): the AO2D would be nonconforming to the data model, aborting.";
+      }
```
sawenzel: so just to understand: if daughter index 0 is negative (== no daughter) we should not have a daughter in index 1 (second daughter)?

It appears to me that the 2 lines above will already cause such misbehaviour. We seem to swap the index whenever daughters[0] > daughters[1] which happens in particular when daughters[1] == -1. @shahor02 : Can you comment on this?

mpuccio:
Hello everyone, after discussing with @sawenzel we came to the conclusion that these cases might be due to pruned particles in the kinematics. We will have a look in details to check that this is the case and that no physics is harmed (there is already quite some certainty about this, as the `keepPhysics` function in Stack keeps primaries and their offsprings).
The current solution will make the AO2D conforming to the data model and will not disrupt current productions.

PR #8396, opened 2022-03-18
Integration of fast simulation module in O2 pipeline

SwirtaB:
Integration allows to to run fast simulation model on event. FastSim module is configured via parameters.


Detectors/ZDC/simulation/src/Detector.cxx
```diff
@@ -2398,6 +2430,22 @@ void Detector::BeginPrimary()
   resetHitIndices();
 
   mCurrentPrincipalParticle = *stack->GetCurrentTrack();
+
+#ifdef ZDC_FASTSIM_ONNX
+  if (o2::zdc::ZDCSimParam::Instance().useZDCFastSim && mFastSimModel != nullptr) {
```
sawenzel: and where do you invoke the model ? Is this not done yet?

SwirtaB: It is done at line 2446.

SwirtaB: mFastSimResults.push_back(mFastSimModel->getChannels(particle));

sawenzel: Ok I see. Thanks.

Detectors/ZDC/simulation/src/Detector.cxx
```diff
@@ -2398,6 +2430,22 @@ void Detector::BeginPrimary()
...
+    std::vector<float> particle = {static_cast<float>(mCurrentPrincipalParticle.Energy()),
```
sawenzel: So this looks like you are creating a new std::vector at each invocation which is overhead.
Consider using a different fixed-size type or struct that can be passed to the fast sim model, so that it can be constructed on the stack.

Detectors/ZDC/simulation/CMakeLists.txt
```diff
@@ -9,13 +9,23 @@
...
+                                        O2::DetectorsRaw O2::Headers O2::FASTSimulation)
+else ()
+        o2_add_library(ZDCSimulation
+                SOURCES src/Detector.cxx src/Digitizer.cxx src/SimCondition.cxx
+                src/ZDCSimParam.cxx src/SpatialPhotonResponse.cxx src/Digits2Raw.cxx
+                src/DigitizerTest.cxx
+                PUBLIC_LINK_LIBRARIES O2::SimulationDataFormat O2::ZDCBase
+                                        O2::DataFormatsZDC O2::CCDB O2::SimConfig
+                                        O2::DetectorsRaw O2::Headers)
+endif()
```
sawenzel, line 22: Here I think we can avoid code duplication by constructing the list of sources or list of link libraries before `o2_add_library` (with conditional list append). Or use some kind of CMake generator expression.

SwirtaB: List is probably a good idea.

Detectors/ZDC/simulation/src/Detector.cxx
SwirtaB: std::vector still will be created at each invocation inside fast sim model wrapper. What I think can do, is to change interface so the setData method will get mCurrentPrincipalParticle and extract necessery data itself. What do you think about it?

I think I will roll back to std::array for now and investigate more generic solution after gathering more information.

Detectors/ZDC/simulation/CMakeLists.txt
SwirtaB: Fixed in commit [490480f](https://github.com/AliceO2Group/AliceO2/pull/8396/commits/490480f565d3cb2ab5610f1e7d3b8c67801f2b6a)

Detectors/ZDC/simulation/src/Detector.cxx
SwirtaB: Fixed in commit [3e2161d](https://github.com/AliceO2Group/AliceO2/pull/8396/commits/3e2161d5bc10386f37db67ce3d2f8ad15cb1cea1)

SwirtaB:
@sawenzel could you please post why fullCI faild? Also if I rember correctly fullCI and o2-cs8 passed on friday. It's strange that it fails now because I haven't made any changes since then.

Detectors/ZDC/simulation/src/Detector.cxx
```diff
@@ -2387,6 +2404,21 @@ void Detector::FinishPrimary()
   // after each primary we should definitely reset
   mLastPrincipalTrackEntered = -1;
   flushSpatialResponse();
+#ifdef ZDC_FASTSIM_ONNX
+  if (o2::zdc::ZDCSimParam::Instance().useZDCFastSim && mFastSimModel != nullptr) {
+    std::fstream output("o2sim-FastSimResult", output.out | output.app);
+    if (!output.is_open()) {
+      LOG(error) << "Could not open file.";
+    }
```
sawenzel: ideally we shouldn't open this at every primary but leave the file handle open until end of simulation

sawenzel (approved):

sawenzel:
@SwirtaB : The errors in o2-dataflow are genuine. This is one example where FastSim is disabled so you need to add more protection.

sawenzel (changes requested):
Still some minor build issues to solve. Otherwise it looks good.

SwirtaB:
I have encountered strange runtime errors so whole O2 rebuild was required. That is why I have done push force.

Some summary:
 - I have greatly modified interface so that it is more generic and will be much easier to maintain in future.
 - According to my internal tests current implementation is on par with previous or slightly faster (due to droping unnecessery copies).
 - I had to roll back to open file on each FinishPrimary due to strange runtime error caused by including fstream in Detector.h. Some other solution is required.

SwirtaB:
Looks like dataflow is building correctly. Any idea why macOS fails?

sawenzel:
I think this is due to the changes that you did in your latest force-push to the models. In general, it's not a good idea to mix several new features into one PR. It would have been better to just work on the integration first.

SwirtaB:
All checks are green. Is it good for merge?

PR #8398, opened 2022-03-18
[DPL Analysis] [Event mixing] Upgrade event mixing to use BinningPolicy

saganatt:
... and get rid of the additional hash table messing around


ktf:
@saganatt it looks like some genuine issue...

ktf:
@saganatt it looks like some genuine issue...

ktf (changes requested):
Please fix the compilation issue.

saganatt:
I don't see it now...?

ktf:
Indeed, probably it simply needed a rebuild.

PR #8399, opened 2022-03-18
Expire when all present

ktf:


ktf:
@davidrohr should expire data when all the non optional inputs are there.

ktf:
Not needed anymore.

ktf:
included in #8405

PR #8400, opened 2022-03-18
[EMCAL-566]: Updated EMCal time calibration

jokonig:
- implemented arguments to optimize the time calibration: time window,
minimum number of entries needed for calibration
- Added option to store time distributions and calibration coefficients
in local root file
- Fixed initialization of EMCALCalibExtractor

### DIFF
--- a/Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALCalibExtractor.h
+++ b/Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALCalibExtractor.h
@@ -40,16 +40,23 @@ class EMCALCalibExtractor
   using boostHisto = boost::histogram::histogram<std::tuple<boost::histogram::axis::regular<double, boost::use_default, boost::use_default, boost::use_default>, boost::histogram::axis::integer<>>, boost::histogram::unlimited_storage<std::allocator<char>>>;
 
  public:
-  EMCALCalibExtractor() = default;
+  EMCALCalibExtractor()
+  {
+    LOG(info) << "initialized EMCALCalibExtractor";
+    if (!mGeometry) {
+      mGeometry = o2::emcal::Geometry::GetInstance();
+      if (!mGeometry) {
+        mGeometry = o2::emcal::Geometry::GetInstanceFromRunNumber(300000); // fallback option
+      }
+    }
+    mNcells = mGeometry->GetNCells();
+  };
   ~EMCALCalibExtractor() = default;
-
-  o2::emcal::Geometry* mGeometry = o2::emcal::Geometry::GetInstanceFromRunNumber(300000);
-  int NCELLS = mGeometry->GetNCells();
 
   int getNsigma() const { return mSigma; }
   void setNsigma(int ns) { mSigma = ns; }
 
-  void setNThreads(int n) { mNThreads = std::min(n, NCELLS); }
+  void setNThreads(int n) { mNThreads = std::min(n, mNcells); }
   int getNThreads() const { return mNThreads; }
 
   void setUseScaledHistoForBadChannels(bool useScaledHistoForBadChannels) { mUseScaledHistoForBadChannels = useScaledHistoForBadChannels; }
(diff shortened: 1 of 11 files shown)


Detectors/EMCAL/calibration/src/EMCALTimeCalibData.cxx
```diff
@@ -55,7 +56,7 @@ void EMCALTimeCalibData::merge(const EMCALTimeCalibData* prev)
   mTimeHisto += prev->getHisto();
 }
 //_____________________________________________
-bool EMCALTimeCalibData::hasEnoughData() const
+bool EMCALTimeCalibData::hasEnoughData(const int minNEntries) const
```
mfasDa: For primitive datatypes the const specifier is usually useless, they are in any case copied by-value so whatever modification a function would do doesn't matter outside, and for within the function constant is a lie as the value is set at runtime.

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALCalibExtractor.h
```diff
@@ -40,12 +40,14 @@ class EMCALCalibExtractor
   using boostHisto = boost::histogram::histogram<std::tuple<boost::histogram::axis::regular<double, boost::use_default, boost::use_default, boost::use_default>, boost::histogram::axis::integer<>>, boost::histogram::unlimited_storage<std::allocator<char>>>;
 
  public:
-  EMCALCalibExtractor() = default;
+  EMCALCalibExtractor()
+  {
+    LOG(info) << "initialized EMCALCalibExtractor";
+    mGeometry = o2::emcal::Geometry::GetInstanceFromRunNumber(300000);
```
mfasDa: Ideally the init should check if the singleton already exist and only if not initialize it with a run number. This will also solve the run number issue as we will initalize the geometry in the EMCALChannelCalibrator (Device) which at some point will hopefully have access to the run number.

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALCalibExtractor.h
```diff
@@ -40,12 +40,14 @@ class EMCALCalibExtractor
...
+    NCELLS = mGeometry->GetNCells();
```
mfasDa: All-capital is usually used for const, for members please use mNcells (members starting with "m")

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelCalibrator.h
```diff
@@ -102,16 +116,15 @@ void EMCALChannelCalibrator<DataInput, DataOutput, HistContainer>::initOutput()
 template <typename DataInput, typename DataOutput, typename HistContainer>
 bool EMCALChannelCalibrator<DataInput, DataOutput, HistContainer>::hasEnoughData(const o2::calibration::TimeSlot<DataInput>& slot) const
 {
-
   const DataInput* c = slot.getContainer();
-  LOG(info) << "Checking statistics";
```
mfasDa, line 107: Needs to be LOG(DEBUG), otherwise you will create a message flood far beyond what EMCAL is already doing now :)

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALTimeCalibData.h
```diff
@@ -73,7 +74,7 @@ class EMCALTimeCalibData
   void merge(const EMCALTimeCalibData* prev);
 
   /// \brief Check if enough data for calibration has been accumulated
-  bool hasEnoughData() const;
+  bool hasEnoughData(const int minNEntries) const;
```
mfasDa: Parameter should not be constant

mfasDa (changes requested):
Only a couple of changes requested from my side

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALTimeCalibData.h
jokonig: done

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelCalibrator.h
jokonig: This is deleted anyways. I removed it as it would not really provide any reasonable information when running the calibration

Detectors/EMCAL/calibration/src/EMCALTimeCalibData.cxx
jokonig: done

mfasDa (approved):
Looks good now. @sawenzel Could you please merge when ready?

shahor02 (approved):

PR #8404, opened 2022-03-19
By default do not send explicit FLP/DISTSUBTIMEFRAME/0xccdb from readers

shahor02:

PR #8405, opened 2022-03-20
Race condition handling

ktf:


ktf:
Updated. The topology is now correct. I only see an error in raw-reader sending out DISTSUBTIMEFRAME/0, which for some reason is not declared as output anymore. Fix in a bit.

ktf:
Updated again. This seems to work, although I get an issue in fv0 which looks similar to the one in fit we spotted yesterday.

shahor02:
> Updated again. This seems to work, although I get an issue in fv0 which looks similar to the one in fit we spotted yesterday.

But your fix was general for the FIT common LUT. Perhaps @afurs can tell if there are subdetector-specific maps where the hasher is provided w/o proper comparison operator?

ktf:
Nevermind, I cleaned up my area before opening the PR, so my fix was not there, actually. I can retry in a bit.

afurs:
Hello @shahor02 comparer and hasher are common for all three FIT sub-detectors(but I kept possibility to make custom ones). So, this PR fixes all three sub-detector's LUT.

ktf:
@afurs @shahor02 indeed it was my area to miss my fix due to git juggling. I see now it triggers deadbeef in a few cases. I will investigate.

ktf:
Ok, I actually can reproduce this now. Investigating.

ktf:
Ok, the new update does not trigger the deadbeef anymore and also TOF works. How do I trigger a deadbeef message on purpose? Can I simply tell it not to send the tof RAWDATA so that it crashes as before?

shahor02:
> How do I trigger a deadbeef message on purpose? Can I simply tell it not to send the tof RAWDATA so that it crashes as before?

As I wrote in the MM:  you can use --drop-tfoption of the raw-file-reader
```
With option --drop-tf "ITS,3,2;TPC,10" the reader will not send the output for ITS in TFs with (TFid%3)==2 and for TPC in TFs with (TFid%10)==0. Note that in order to acknowledge the TF, the message {FLP/DISTSUBTIMEFRAME/0} will still be sent even if all detector's data was dropped for a given TF.
```
https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/Raw#raw-data-file-reader-workflow

ktf:
@shahor02 ok, this seems to work for me, with and without missing data. Could you cross check?

shahor02:
As discussed over MM: works fine except the extra FLP/DISTSUBTIMEFRAME/xxx messages triggered in the run() call which just sets EOS.
BTW, since now you send in loop these messages both for the optional and for the ccdb fetcher, the https://github.com/AliceO2Group/AliceO2/pull/8404 will not work since it defines DPL origin for the ccdb fetcher. 
Shall I close my PR? We hardly will have 0xccdb devices with optional inputs attached

shahor02:
@ktf, your suggested patch
```
diff --git a/Framework/Core/src/CommonServices.cxx b/Framework/Core/src/CommonServices.cxx
index ceaaebe98d8..de02ced33db 100644
--- a/Framework/Core/src/CommonServices.cxx
+++ b/Framework/Core/src/CommonServices.cxx
@@ -497,6 +497,10 @@ o2::framework::ServiceSpec CommonServices::ccdbSupportSpec()
       const auto ref = pc.inputs().getFirstValid(true);
       const auto* dh = DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
       const auto* dph = DataRefUtils::getHeader<DataProcessingHeader*>(ref);
+      if (pc.services().get<DeviceState>().streaming == StreamingState::EndOfStreaming) {
+        LOGP(info, "We are in EoS, do not automatically add DISTSUBTIMEFRAME");
+        return;
+      }
       // For any output that is a FLP/DISTSUBTIMEFRAME with subspec != 0,
       // we create a new message.
       InputSpec matcher{"matcher", ConcreteDataTypeMatcher{"FLP", "DISTSUBTIMEFRAME"}};
```
fixes the problem of unnecessary DISTSTF sending at EoS, could you include it to this PR?
What remains is the double ccdbfetcher call per TF.

shahor02:
@ktf ok, the double ccdb fetcher call happen because the raw-reader sends explicit `DISTSUBTIMEFRAME/0xccdb` message and your mechanism sends another on top of it. 
I will suppress mine.

ktf:
Ok. I pushed again with the fix applied. Hopefully this will be it.

shahor02:
@ktf still not good: I've tried digits driven reconstruction, e.g. simply
`o2-its-reco-workflow --trackerCA --disable-mc --run | tee xx.log`
and with this checking for EoS no processing is done at all. The problem is that all O2 readers (digits, clusters etc.) are reading/injecting 1 tree entry per TF (in the MC we have just 1 entry) and on the last entry they also call 
```
  if (mTree->GetReadEntry() + 1 >= mTree->GetEntries()) {
    pc.services().get<ControlService>().endOfStream();
    pc.services().get<ControlService>().readyToQuit(QuitRequest::Me);
  }
```
So, the `if` in the 
```
if (pc.services().get<DeviceState>().streaming == StreamingState::EndOfStreaming) {
   LOGP(debug, "We are in EoS, do not automatically add DISTSUBTIMEFRAME to outgoing messages");
   return;
}
```
should check also if there are no messages sent by the device.

ktf:
Tested separately by both @shahor02 and myself. Merging.

sawenzel:
This PR seems to break all MC workflows (or at least make them break).
Since the nightly release of today, all productions exhibit an error message like so:
```
[7621:mft-digit-reader]: [03:26:20][INFO] fair::mq::Device running...
[7621:mft-digit-reader]: Error in <TTree::SetBranchAddress>: The pointer type given (o2::dataformats::IOMCTruthContainerView) does not correspond to the class needed (vector<char>) by the branch: MFTDigitMCTruth
[7621:mft-digit-reader]: [03:26:20][INFO] MFTDigitReader pushes 0 ROFRecords, 0 digits at entry 0
[7621:mft-digit-reader]:     /lib64/libc.so.6: ()
[7621:mft-digit-reader]:     /cvmfs/alice.cern.ch/el7-x86_64/Packages/O2/nightly-20220325-1/lib/libO2ITSMFTWorkflow.so: o2::itsmft::DigitReader::run(o2::framework::ProcessingContext&)
[7621:mft-digit-reader]:     /cvmfs/alice.cern.ch/el7-x86_64/Packages/O2/nightly-20220325-1/lib/libO2Framework.so: ()
```

In particular it seems to be caused by commit 1eb2fb477cee62. Actually the fullCI would probably have failed. Better to wait for it.

sawenzel:
@ktf @shahor02 : Any suggestion what to do? Roll back? Or investigate on ITS/MFT side? Unfortunately, this is blocking the production efforts for QM.

sawenzel:
To reproduce, it suffices to run https://github.com/AliceO2Group/AliceO2/blob/dev/prodtests/sim_challenge.sh.

ktf:
Indeed I think we are missing one commit.

Ciao,
Giulio

ktf:
No, the commit is there. I think #8439 is what is needed. It indeed skipped my local sim tests due to the log_done. Sorry about that. logs now deleted and testing. Can we add a dependency of the tests on the code? Maybe we should rethink sim_challenge as a set of cmake macros...

ktf:
For the record, https://github.com/AliceO2Group/AliceO2/commit/1eb2fb477cee62eef4a35963a1ca914fde4ae66b is known to be buggy, but https://github.com/AliceO2Group/AliceO2/pull/8405/commits/415dfc29f38e2696a0780fb4218b5e6119599d46 should have fixed it. Anyways #8439 is needed in any case, but I am not sure that's the actual issue.

shahor02:
Actually, I have tested it both with FST and with reconstruction from digit, but since I used the digits from FST  , where the MC labels are suppressed, I had to use --disable-mc...

ktf:
Indeed, it looks like the standard fullCI is not complaining about anything. Could you check if #8439 actually fixes the issue or provide me the commandline to reproduce?

davidrohr:
Hi, I have tried what Sandro posted: https://github.com/AliceO2Group/AliceO2/blob/dev/prodtests/sim_challenge.sh.
- works fine before this PR (except for some TRD CCDB problem).
- crashes in ITS or MFT reco with the issue Sandro posted with this PR.
- When I add #8439, the situation does not change. It still crashes in the ITS digit reader.

Besides, the other issue I reported via mail, that ITS raw files are empty, is also caused by this PR. This was not seen by the CI since it was just handled like an empty ITS event. I have improved the FST script here #8441 to throw an error in that case, but I am not merging that yet.
The behavior is like the issue reported by Sandro:
- works before this PR.
- Fails with this PR.
- Also fails when I add #8439

And I am quite sure the 2 problems are the same. Since also with `sim_challenge.sh` the ITS and the MFT digits are empty, and I assume the processing fails for that reason.

shahor02:
Hi,
Looks like the run() method of digitizer is not called due to the EoS sent together with data.
I am on the move now, can check only tomorrow

davidrohr:
Hi @shahor02 , yes, only ITS and MFT are empty. Although, running the check on some of my older datasets I found one with PHS empty, checking that, but perhaps that is just an old problem.

OK, that would actually be good if the problem is only with the EoS / data message not triggering `run()`, perhaps then we can solve it easily. Let's see tomorrow. In the meantime, I'll run a set of full 8 GPU tests with different detector combinations to double check, particularly those who were failing online, and also check for SHM memory leaks.

ktf:
Could you add:
```
diff --git a/Framework/Core/src/CommonServices.cxx b/Framework/Core/src/CommonServices.cxx
index f4888b908d5..f2aa51db3cc 100644
--- a/Framework/Core/src/CommonServices.cxx
+++ b/Framework/Core/src/CommonServices.cxx
@@ -496,9 +496,13 @@ o2::framework::ServiceSpec CommonServices::ccdbSupportSpec()
       }
       if (pc.services().get<DeviceState>().streaming == StreamingState::EndOfStreaming) {
         if (pc.outputs().countDeviceOutputs(true) == 0) {
-          LOGP(debug, "We are in EoS w/o outputs, do not automatically add DISTSUBTIMEFRAME to outgoing messages");
+          LOGP(info, "We are in EoS w/o outputs, do not automatically add DISTSUBTIMEFRAME to outgoing messages");
           return;
+        } else {
+          LOGP(info, "We are in EoS with outputs, automatically add DISTSUBTIMEFRAME to outgoing messages");
         }
+      } else {
+        LOGP(info, "We are not in EoS");
       }
       const auto ref = pc.inputs().getFirstValid(true);
       const auto* dh = DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
```

to understand which path is taken? Indeed https://github.com/AliceO2Group/AliceO2/pull/8439 should not matter for its/mft.

PR #8408, opened 2022-03-21
Do not check for alien token on macOS

ktf:
The check for TGrid::Connect(\"alien\") should be enough.


ktf:
@sawenzel any objections? I was actually discussing with @costing what is the case this check is covering which is not already covered by the `TGrid::Connect("alien")` (now that we use jalien). Maybe we could drop the whole thing...

PR #8411, opened 2022-03-21
Enable upgrade dets if requested

benedikt-voelkel:


benedikt-voelkel:
The removed check (this PR) for upgrade detectors was actually not correct. Since we require the detector to be in the module list, we only rely on that and readout any detector that is in line with the module list.

sawenzel (approved):

PR #8412, opened 2022-03-21
TPC simulation: fix bug where not all hits are stored when a track crosses the centre of a sector

matthias-kleiner:
When the shifted sector (centre of the sector) or the track ID changes during storage of the hits in the simulation, the first hit of the new sector/particle were not stored.
This led to lower cluster charge in the centre of a sector for tracks which crosses the centre of a sector.


wiechula (approved):
Very nice! Thanks a lot for the fix.

PR #8415, opened 2022-03-22
FIT: fix undefined behaviour in LookUpTable

ktf:


ktf:
Relying on the fact that equality will not be called in the case of a perfect hash is undefined behaviour. E.g. hashes can be bucketed and a linked list can be kept for a given bucket. This in particular breaks clang on macOS. Having the proper equality operators fixes the issue.

mslupeck:
@afurs - I think this is your area

afurs:
@mslupeck it is OK. There was an issue after MacOS compiler update(stdlib)

mslupeck (approved):

PR #8416, opened 2022-03-22
switch variable initialization order

zhaomr13:
To fix the Wrong-Order warning

### DIFF
--- a/Algorithm/test/parser.cxx
+++ b/Algorithm/test/parser.cxx
@@ -46,7 +46,7 @@ struct SizedTrailer {
   unsigned char flags = 0xaa;
   size_t payloadSize = 0;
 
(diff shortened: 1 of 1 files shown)


sawenzel:
Can you please be specific about "the wrong order warning"? Sometimes the orders are engineered to provide a specific data layout and size. This change will likely increase the size of the struct from 16 to 24 bytes which is most certainly not wanted.

sawenzel:
Can you please be specific about "the wrong order warning"? Sometimes the orders are engineered to provide a specific data layout and size. This change will likely increase the size of the struct from 16 to 24 bytes which is most certainly not wanted.

zhaomr13:
@sawenzel 
It is just the warning of "xx will be initialized after [-Werror=reorder]".
I did not notice the trick to reduce the size of the struct. 
Now I switch it back and change the order of initialization.

PR #8394, opened 2022-03-18
Update dpl-workflow.sh

sawenzel:
TRD is not fundamentally required for the TOF matcher workflow (because we can adjust the track sources)

PR #8419, opened 2022-03-22
Fix matching workflow creation logic

martenole:
Pinging @davidrohr  @shahor02 and @benedikt-voelkel 

I am not 100% sure, but I think this is the cleanest solution to the problem encountered today. No topology could be generated when the detectors ITS, TPC and TOF were included and the event display was active at the same time. The error which was observed was that the event display complained about missing input (ITS-TPC-TOF matches). Checking the workflow I saw that the `o2-tof-matcher-workflow` was not included.
With the current workflow generation TOF matching is only active when all of ITS, TPC, TRD and TOF are present. For the TRD matching it is similar, it only works when ITS and TPC are both present, although they are not both strictly needed.
Since `TRD_SOURCES` and `TOF_SOURCES` are only filled when there should be some matching to one of the detectors I think this change is OK.
What do you think?

InfoLogger message:
```
Severity	fatal
Date	22/03/2022
Time	16:12:40
Hostname	epn001
Rolename	
PID	4114354
Username	epn
System	ODC
Facility	ODC
Detector	
Partition	2X7BPyXvyZT
Run	
ErrCode	
ErrLine	74
ErrSource	InfoLogger.h
Incorrect topology provided: Topology script "/usr/bin/bash -c \"GEN_TOPO_HASH=1 GEN_TOPO_SOURCE='71739f0b27bf7718a8b8738878c404c5c9338c27' DDMODE='processing' GEN_TOPO_LIBRARY_FILE='production/production.desc' GEN_TOPO_WORKFLOW_NAME='synchronous-workflow-1numa' WORKFLOW_DETECTORS='CPV,EMC,ITS,MID,PHS,TOF,TPC' WORKFLOW_DETECTORS_QC='CPV,EMC,ITS,MID,PHS,TOF,TPC' WORKFLOW_DETECTORS_CALIB='CPV,EMC,ITS,MID,PHS,TOF,TPC' WORKFLOW_PARAMETERS='QC,CALIB,GPU,CTF,EVENT_DISPLAY' RECO_NUM_NODES_OVERRIDE=135 RECO_MAX_FAIL_NODES_OVERRIDE=0 MULTIPLICITY_FACTOR_RAWDECODERS=1 MULTIPLICITY_FACTOR_CTFENCODERS=1 MULTIPLICITY_FACTOR_REST=1 GEN_TOPO_WIPE_CACHE=0 BEAMTYPE='cosmic' NHBPERTF=128 GEN_TOPO_PARTITION='2X7BPyXvyZT' GEN_TOPO_ONTHEFLY=1 OVERRIDE_PDPSUITE_VERSION='O2PDPSuite/epn-20220314-0951-1' SET_QCJSON_VERSION='1' DD_DISK_FRACTION='100' SHM_MANAGER_SHMID='' SHM_MANAGER_SHM_RECREATE=0 /home/epn/pdp/gen_topo.sh\"" execution faile [...]
```

### DIFF
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -409,8 +409,8 @@ has_detector_reco ITS && add_W o2-its-reco-workflow "--trackerCA $ITS_CONFIG $DI
 has_detectors_reco ITS TPC && has_detector_matching ITSTPC && add_W o2-tpcits-match-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N itstpc-track-matcher MATCH REST TPCITS)" "$ITSMFT_FILES"
 has_detector_reco FT0 && add_W o2-ft0-reco-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N ft0-reconstructor FT0 REST)"
 has_detector_reco TRD && add_W o2-trd-tracklet-transformer "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC $TRD_FILTER_CONFIG --pipeline $(get_N TRDTRACKLETTRANSFORMER TRD REST TRDTRK)"
-has_detectors_reco TRD TPC ITS && [[ ! -z "$TRD_SOURCES" ]] && add_W o2-trd-global-tracking "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC $TRD_CONFIG $TRD_FILTER_CONFIG --track-sources $TRD_SOURCES" "$TRD_CONFIG_KEY;$ITSMFT_FILES"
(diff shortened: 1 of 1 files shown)


davidrohr:
Isn't it sufficient to just remove TRD from the has detector reco for the tof workflow?

martenole:
But you have different configurations. It could also be that ITS is not available and you do TPC-TOF matching or vice versa

martenole:
So you would need to repeat the logic which is already implemented above:
https://github.com/AliceO2Group/AliceO2/blob/8ca3f76ef6c9d46b9caa6399e88055e57ac59324/prodtests/full-system-test/dpl-workflow.sh#L201-L219

davidrohr:
Ok, point taken. But then i would leave at least the check for has detector reco trd (and tof respectively), although i assume when that is not set the sources are anyway empty.

Kind Regards
David Rohr

martenole:
Yes that check should be redundant, but I doesn't hurt to have it in. It is not urgent to have this in the nightly for tomorrow, since a software update is anyhow not foreseen.
Thanks for your feedback!

shahor02:
@martenole this PR is fine but it essentially replicates Sandro's https://github.com/AliceO2Group/AliceO2/pull/8394 (see also my comment there)

martenole:
That is a bit frustrating. I should watch the other PRs more in oder not to debug an already known problem.
I do prefer the way it is done in my PR, so maybe @sawenzel can adapt his and then I would close this?
Because during the commissioning where they try to run sometimes with what is available I could imagine that for tests also ITS and TOF (without TPC) might be included in a partition and then we would get the same error again. Unless the event display would not show ITS-TOF matches anyway (can't remember right now).

shahor02:
@martenole I agree that your PR is more appropriate.
We don't have ITS-TRD or ITS-TOF matching and I don't see why it would be useful.

sawenzel (approved):

PR #8420, opened 2022-03-23
[QC-751] Add new QC flags

knopers8:
These represent the typical problems in Run 2 and derive from the DPG Run 2 run list definitions.
ProcessingError, MissingQualityObject and MissingQuality are going to be removed after we stop using them in QC.


Barthelemy (dismissed):

Barthelemy:
there seems to be a legit problem

knopers8:
yes, a dot instead of a comma

PR #8422, opened 2022-03-23
Better protection against bad data

ktf:


ktf:
@preghenella any idea of what might cause rdh to be 0?

ktf:
ok, it was getting triggered by the deadbeef messages. Anyways, this still looks like a good idea. @noferini could you sign off?

noferini (approved):
done

PR #8424, opened 2022-03-23
AOD: Fix for inconsistent MC daughter problem

sawenzel:
This should fix a problem described in https://alice.its.cern.ch/jira/browse/O2-2820,
related to inconsistent MC daughter ids.

The problem with daughter ids was not observable for the
case when we used all MC particles (not just reconstructed) ones.
This meant that the filtering code with mRecoOnly was buggy.

Indeed, it then turned out that some `continue` statements should
not have been there, because they prevented correct registration of
mother/daughter particles. This is now corrected in this commit.


sawenzel:
@mpuccio

PR #8425, opened 2022-03-23
Add protection in TOF compressor against deadbeef input

shahor02:


noferini (approved):
Hi Ruben,
it is fine with me.
For my understanding.
What device can inject the 0xDEADBEEF, and what condition can trigger that?
adding also @preghenella to keep him informed

shahor02:
The deadbeef is generate internally by the framework in all raw data processing devices if only FLP/DISTSUBTIMEFRAME/0 input is received from the DD, in response the device should send valid output in order to not block the workflow.

noferini:
> The deadbeef is generate internally by the framework in all raw data processing devices if only FLP/DISTSUBTIMEFRAME/0 input is received from the DD, in response the device should send valid output in order to not block the workflow.

I see, thanks.
I guess that DD has an internal timeout if it doesn't receive data from readout

shahor02:
On FLP STFBuilder : yes. When the TOF compression is done on the FLP, this will lead simply to missing STF with CRAWDATA, and this is handled by the deadbeef protection in the decompressor. But in case of raw processing on the EPN or in the FST, we need to protect also the Compressor.

PR #8427, opened 2022-03-23
Force 1 thread for ITSMFT reconstruction on macOS

ktf:
Does not work otherwise: busy-looping waiting for threads to join.


ktf:
Trivial. Merging.

PR #8429, opened 2022-03-23
DPL: cleanup unneeded TimingInfo instance

ktf:


ktf:
Apparently I forgot to remove it when I switch to use a service for it.

PR #8430, opened 2022-03-23
Validity range fix; debug info added

peressounko:
ccdb path fixed;Validity range fix; debug info added

### DIFF
--- a/Detectors/PHOS/workflow/src/ClusterizerSpec.cxx
+++ b/Detectors/PHOS/workflow/src/ClusterizerSpec.cxx
@@ -43,8 +43,8 @@ void ClusterizerSpec::init(framework::InitContext& ctx)
     ccdbManager.setURL(o2::base::NameConf::getCCDBServer());
     LOG(info) << " set-up CCDB " << o2::base::NameConf::getCCDBServer();
 
-    BadChannelsMap* badMap = ccdbManager.get<o2::phos::BadChannelsMap>("PHS/BadMap");
-    CalibParams* calibParams = ccdbManager.get<o2::phos::CalibParams>("PHS/Calib");
+    BadChannelsMap* badMap = ccdbManager.get<o2::phos::BadChannelsMap>("PHS/Calib/BadMap");
+    CalibParams* calibParams = ccdbManager.get<o2::phos::CalibParams>("PHS/Calib/CalibParams");
     if (badMap) {
       mClusterizer.setBadMap(badMap);
     } else {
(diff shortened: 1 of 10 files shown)


Detectors/PHOS/workflow/src/ClusterizerSpec.cxx
```diff
@@ -43,8 +43,8 @@ void ClusterizerSpec::init(framework::InitContext& ctx)
     ccdbManager.setURL(o2::base::NameConf::getCCDBServer());
     LOG(info) << " set-up CCDB " << o2::base::NameConf::getCCDBServer();
 
-    BadChannelsMap* badMap = ccdbManager.get<o2::phos::BadChannelsMap>("PHS/BadMap");
-    CalibParams* calibParams = ccdbManager.get<o2::phos::CalibParams>("PHS/Calib");
+    BadChannelsMap* badMap = ccdbManager.get<o2::phos::BadChannelsMap>("PHS/Calib/BadMap");
+    CalibParams* calibParams = ccdbManager.get<o2::phos::CalibParams>("PHS/Calib/CalibParams");
```
shahor02, line 47: above you change to this: https://github.com/AliceO2Group/AliceO2/blob/9e546ae0d3a1199026eb910316b7bee97cd84bae/Detectors/PHOS/calib/src/PHOSBadMapCalibDevice.cxx#L137-L145
and here you again use PHS/Calib/BadMap. Could you cross-check?

shahor02 (commented):
@peressounko please see the comment below. In general, would be good to gradually switch to DPL CCDB fetcher instead of modifying the code using BasicCCDBManager.

Detectors/PHOS/workflow/src/ClusterizerSpec.cxx
peressounko: Thank you for careful reading. Idea is to store intermediate maps from different methods in PHS/BadMap and manually combine them to final map stored in PHS/Calib/BadMap.

peressounko:
Concerning switching to CCDB fetcher - I see your exchange with Sergey, probably I will switch when everything will be fixed/checked and I  can use his code as an example.

Detectors/PHOS/workflow/src/ClusterizerSpec.cxx
shahor02: Why manual (which we want to avoid). How often will you do this?

peressounko: Ingredients of bad map will be run automatically. Combination will be done once per modification of electronics ~ as in Run2 per period or few periods.

peressounko:
Can not find what is the problem in build/O2/o2-dataflow-cs8?

shahor02 (approved):

PR #8434, opened 2022-03-24
Fix dummy values in O2fwdtrack

nburmaso:
This is a fix addressing issue stated in [O2-2831](https://alice.its.cern.ch/jira/browse/O2-2831).

@rpezzi, could you please check the code and see if it is ok with you? Specifically these parts:

https://github.com/nburmaso/AliceO2/blob/3642ca5901d0717fe72ced1324feddb250ee6e3f/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L370-L382

https://github.com/nburmaso/AliceO2/blob/3642ca5901d0717fe72ced1324feddb250ee6e3f/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L400

https://github.com/nburmaso/AliceO2/blob/3642ca5901d0717fe72ced1324feddb250ee6e3f/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L490-L503


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -452,127 +435,142 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
...
+    fwdInfo.chi2matchmchmft = track.getMFTMCHMatchingChi2();
+    fwdInfo.matchscoremchmft = track.getMFTMCHMatchingScore();
+    fwdInfo.matchmfttrackid = track.getMFTTrackID();
+    fwdInfo.matchmchtrackid = track.getMCHTrackID();
+    fwdInfo.trackTime = track.getTimeMUS().getTimeStamp() * 1.e3;
+    fwdInfo.trackTimeRes = track.getTimeMUS().getTimeStampError() * 1.e3;
+
+    getMCHBitMap(fwdInfo.matchmchtrackid);
+
+    int midTrackID = track.getMIDTrackID();
```
dstocco, line 525: The information that you want to put in the AOD is now computed in the MID Track. 
So it's enough to do: midTrack.midBitMap = midTrack.getHitMap().
Concerning the second information, this was changed with respect to the midBoards information in a way that it is more similar to what was done in Run 2.
If you prefer, I can add a getEfficiencyWord method that you can use to define midBoards.

nburmaso: Hi @dstocco. Thank you for the PR #8495. I have modified the code using the method you added. Could you please take a look?

dstocco: I think it should be ok now. Thanks!

dstocco (dismissed):

sawenzel (dismissed):
Looks ok to me. Maybe @shahor02 can also comment.

zconesa:
I know you are busy, but is there a chance to get it reviewed / approved ? @shahor02 @sawenzel

PR #8435, opened 2022-03-24
TPC-QC: update of Tracks task

stheckel:
o Add histograms for q over pT
o Enlarge range of Ncluster histograms
o Change switch of A or C side from eta to Clusters only on that side

### DIFF
--- a/Detectors/TPC/qc/src/Tracks.cxx
+++ b/Detectors/TPC/qc/src/Tracks.cxx
@@ -91,14 +91,13 @@ bool Tracks::processTrack(const o2::tpc::TrackTPC& track)
   const auto phi = track.getPhi();
   const auto pt = track.getPt();
   const auto sign = track.getSign();
-  const auto charge = track.getCharge();
+  const auto qOverPt = track.getQ2Pt();
   const auto nCls = track.getNClusterReferences();
   const auto dEdxTot = track.getdEdx().dEdxTotTPC;
   const auto hasASideOnly = track.hasASideClustersOnly();
   const auto hasCSideOnly = track.hasCSideClustersOnly();
 
   double absEta = TMath::Abs(eta);
-  double qOverPt = 0.;
 
   // ===| histogram filling before cuts |===
   mHist1D[0].Fill(nCls);
@@ -110,10 +109,6 @@ bool Tracks::processTrack(const o2::tpc::TrackTPC& track)
 
   // ===| histogram filling including cuts |===
   if (absEta < 1. && nCls > 60 && dEdxTot > 20) {
-
-    if (pt > 0) {
-      qOverPt = charge / pt;
-    }
 
     // ===| 1D histogram filling |===
     mHist1D[1].Fill(nCls);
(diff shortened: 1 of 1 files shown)


Detectors/TPC/qc/src/Tracks.cxx
```diff
@@ -50,17 +50,21 @@ void Tracks::initializeHistograms()
   mHist1D.emplace_back("hPtPos", "Transverse momentum, pos. tracks;p_T", logPtBinning.size() - 1, logPtBinning.data());         //| mHist1D[14]
   mHist1D.emplace_back("hEtaBeforeCuts", "Pseudorapidity (before cuts);eta", 400, -2., 2.);                                     //| mHist1D[15]
   mHist1D.emplace_back("hPtBeforeCuts", "Transverse momentum (before cuts);p_T", logPtBinning.size() - 1, logPtBinning.data()); //| mHist1D[16]
+  mHist1D.emplace_back("hQOverPt", "Charge over transverse momentum;q/p_T", 400, -20., 20.);                                    //| mHist1D[17]
+  mHist1D.emplace_back("hPhiBothSides", "Azimuthal angle, both sides clusters;phi", 360, 0., 2 * M_PI);                         //| mHist1D[18]
```
chiarazampolli, line 54: Is this used? Maybe I just don't see it in the changes?

wiechula: It is in lines 127 and 132 below.

wiechula (dismissed):

Detectors/TPC/qc/src/Tracks.cxx
chiarazampolli: Ah yes, I did not see L127. Thanks!
BTW, we need @shahor02 or @davidrohr to approve...

Detectors/TPC/qc/src/Tracks.cxx
```diff
@@ -103,44 +111,53 @@ bool Tracks::processTrack(const o2::tpc::TrackTPC& track)
   // ===| histogram filling including cuts |===
   if (absEta < 1. && nCls > 60 && dEdxTot > 20) {
 
+    if (pt > 0) {
```
shahor02: Why don't you use directly `track.getQ2Pt()` ?

shahor02 (commented):
Hi, a small comment below, but since the changes are rather trivial, I can merge it w/o waiting for remaining CIs, let me know.

stheckel:
Thanks Chiara and Jens for checking this PR!

Concerning the usage of the histograms, honestly, the implementation with the vectors is a bit ugly at the moment and with more and more histograms its indeed getting difficult to keep track. We are already working on restructuring the PID task using maps instead of vectors, this could/should be also done for the Tracks (but not in this PR).

Detectors/TPC/qc/src/Tracks.cxx
stheckel: Good point, I will change that.

stheckel:
Hi @shahor02,
I updated this PR with your suggestion, which in principle is a trivial change. Before, almost all checks had been done and successful, now I see that the checks take much longer, most are running or not even started and one even failed - with the error "No errors found", so I guess unrelated. Would it maybe still be possible to merge?

shahor02 (approved):

stheckel:
Thanks a lot!

PR #8439, opened 2022-03-25
Extra fixes for the Race condition handling

ktf:


ktf:
@davidrohr @sawenzel @shahor02, I think this is it. We probably need something similar also for arrow, but I will add this in a separate PR.

davidrohr:
This works for me now, tested both with full_system_test.sh and sim_challenge.sh, and fixes the digitizer problem introduced with #8405. PLEASE DO NOT MERGE YET! First I'd like to understand why #8441 did not spot the issue, in order to be sure that the CI can find such problems in the future. I want to make the CI run #8443 first before this is merged. I'll merge later today when I understand this.

PR #8454, opened 2022-03-28
Switch to CCDB fetcher; Fix energy calibrator;

peressounko:
- Switch to CCDB fetcher; 
- Fix energy calibrator
- Send digits for later calibration to vector stream instead of file output

### DIFF
--- a/Detectors/PHOS/calib/include/PHOSCalibWorkflow/PHOSEnergyCalibrator.h
+++ b/Detectors/PHOS/calib/include/PHOSCalibWorkflow/PHOSEnergyCalibrator.h
@@ -28,7 +28,6 @@
 
 #include <TLorentzVector.h>
 #include <TVector3.h>
-#include <TFile.h>
 #include <array>
 
 using namespace o2::framework;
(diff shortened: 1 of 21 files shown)


peressounko:
Dear @shahor02, could you please have a look at this PR?

Detectors/PHOS/calib/include/PHOSCalibWorkflow/PHOSEnergyCalibrator.h
```diff
@@ -70,43 +69,44 @@ class PHOSEnergySlot
...
-  void fill(const gsl::span<const Cluster>& /*c*/){}; //not used
-  void merge(const PHOSEnergySlot* /*prev*/) {}       //not used
+  void fill(const gsl::span<const Cluster>& /*c*/){}; // not used
+  void merge(const PHOSEnergySlot* /*prev*/) {}       // not used
   void clear();
 
-  ETCalibHistos& getCollectedHistos() { return mHistos; }
-  std::vector<uint32_t>& getCollectedDigits() { return mDigits; }
+  ETCalibHistos* getCollectedHistos() { return mHistos.get(); }
+  const std::vector<uint32_t>& getCollectedDigits() { return mDigits; }
```
shahor02: this should be const method

Detectors/PHOS/calib/include/PHOSCalibWorkflow/PHOSEnergyCalibrator.h
```diff
@@ -70,43 +69,44 @@ class PHOSEnergySlot
...
+  std::unique_ptr<ETCalibHistos> mHistos; /// final histograms
   uint32_t mEvBC = 0;
   uint32_t mEvOrbit = 0;
   uint32_t mEvent = 0;
-  float mPtMin = 1.5; /// minimal energy to fill inv. mass histo
-  float mEminHGTime = 1.5;
-  float mEminLGTime = 5.;
-  std::vector<uint32_t> mDigits; /// list of calibration digits to fill
-
-  ClassDefNV(PHOSEnergySlot, 1);
```
shahor02, line 109: You've suppressed  the ClassDef for this class but did not remove it from the `Detectors/PHOS/calib/src/PHOSCalibWorkflowLinkDef.h`

Detectors/PHOS/calib/src/PHOSEnergyCalibDevice.cxx
```diff
@@ -64,18 +47,44 @@ void PHOSEnergyCalibDevice::init(o2::framework::InitContext& ic)
...
 {
-  // TODO! extract vertex information and send to Calibrator
+
+  // Do not use ccdb if already created
+  if (!mHasCalib) { // Default map and calibration was not set, use CCDB
+    LOG(info) << "Getting calib from CCDB";
+    // update BadMap and calibration if necessary
+    std::decay_t<decltype(pc.inputs().get<o2::phos::BadChannelsMap*>("bdmap"))> badMapPtr{};
+    badMapPtr = pc.inputs().get<o2::phos::BadChannelsMap*>("bdmap");
+    const o2::phos::BadChannelsMap* badMap = badMapPtr.get();
```
shahor02: why not simply 
```
auto badMapPtr = pc.inputs().get<o2::phos::BadChannelsMap*>("bdmap");
mCalibrator->setBadMap(badMapPtr.get());
```
You don't need to check if the object is not fetched, the DPL fetcher will produce Fatal in that case.

Detectors/PHOS/calib/src/PHOSEnergyCalibrator.cxx
```diff
@@ -92,16 +93,16 @@ void PHOSEnergySlot::fill(const gsl::span<const Cluster>& clusters, const gsl::s
         d.mCluster = (i - firstCluInEvent) % kMaxCluInEvent;
         mDigits.push_back(d.mDataWord);
         if (i - firstCluInEvent > kMaxCluInEvent) {
-          //Normally this is not critical as indexes are used "locally", i.e. are compared to previous/next
-          LOG(info) << "Too many clusters per event:" << i - firstCluInEvent << ", apply more strict selection; clusters with same indexes will appear";
+          // Normally this is not critical as indexes are used "locally", i.e. are compared to previous/next
+          LOG(error) << "Too many clusters per event:" << i - firstCluInEvent << ", apply more strict selection; clusters with same indexes will appear";
```
shahor02: if this should attract attention in the infologger, please use `LOG(important)`. `error` is for the algorithmic errors.

shahor02 (changes requested):
Hi @peressounko 

What is the reason to keep `useCCDB` options? The standard way should be to have  reasonable default objects overridden by specific objects and always use CCDB.
Then, I see that you are controlling the 1st update of CCDB objects by `mHasCalib`. Why don't you use the finaliseCCDB method, which is called each time the object is updated?
It is much more universal, see e.g.:
https://github.com/AliceO2Group/AliceO2/blob/ba7db323bd148725212e318980612998d9e4e5ef/Detectors/ITSMFT/common/workflow/src/STFDecoderSpec.cxx#L253-L267
with the check triggered by 
https://github.com/AliceO2Group/AliceO2/blob/ba7db323bd148725212e318980612998d9e4e5ef/Detectors/ITSMFT/common/workflow/src/STFDecoderSpec.cxx#L242-L249
called from
https://github.com/AliceO2Group/AliceO2/blob/ba7db323bd148725212e318980612998d9e4e5ef/Detectors/ITSMFT/common/workflow/src/STFDecoderSpec.cxx#L137

peressounko:
Dear @shahor02, thank you for careful reading, I am implementing your suggestions. Concerning reading or not CCDB objects, there are 2 arguments: 1) with my connection it takes ~50% of time to check if there is a new objects and clusterization itself. I did not make real timing, but looked at the speed of output. We need object once at the very beginning but fetcher checks it again and again. Probably, with finaliseCCDB one can avoid it - I am checking. 2) We need possibility to apply empty bad map e.g. for re-calculation of bad map. Normally fetcher will provide the most recent and complete bad map. So, for clusterizer at EPN for QC and calibration we will need one map, for cluterizer for AOD analysis - another (stricter) bad map and calibration. I will think, should we provide two kinds of bad map "definitely bad" and "one for physics analysis".

shahor02:
Hi @peressounko 
The DPL fetcher will anyway check the validity of the object (w/o fetching it if the cached one is still valid) for every TF, this is required behaviour. The finaliseCCDB (where you can do assignments) will be called only when the new object is fetched, effectively only once for long-lived objects.
For the empty maps: ok, if for certain operations you need to disable loading the object, this justifies the switch.

peressounko:
Dear @shahor02, should we then return to BasicCCDBManager to get object once instead of fetching it each TF?

shahor02:
> Dear @shahor02, should we then return to BasicCCDBManager to get object once instead of fetching it each TF?

Why? The DPL fetcher does not fetch the object at every TF. In the normal mode, when you need calibration from the CCDB, you should use DPL fetcher and delegate all updates to finaliseCCDB.
In the calibration mode, where you don't need external objects, you should simply not subscribe to these objects, like you already do in https://github.com/AliceO2Group/AliceO2/blob/5dcd383353f1de0919898e96199d044d97b5b962/Detectors/PHOS/calib/src/PHOSEnergyCalibDevice.cxx#L126-L129 *AND* do not quere the data provided by these inputs.
E.g. in the ITS, when I don't need certain object, I have this check:
https://github.com/AliceO2Group/AliceO2/blob/a686cf43c7fd57e5b91a21213ee43b3326d13e48/Detectors/ITSMFT/common/workflow/src/STFDecoderSpec.cxx#L299-L302
followed by the check https://github.com/AliceO2Group/AliceO2/blob/a686cf43c7fd57e5b91a21213ee43b3326d13e48/Detectors/ITSMFT/common/workflow/src/STFDecoderSpec.cxx#L246-L248 done for every TF. Note that it is `pc.inputs().get...` for the CCDB input which triggers finaliseCCDB (provided that the object was updated by the fetcher).

peressounko:
Dear @shahor02, it there any difference in terms of performance between current implementation where I once pick up (permanent) calib object checking with switch mHasCalib and approach with finaliseCCDB which you implemented e.g. in CTFDecoderSpec? Fetcher will anyway check updates of calib Object in both cases? If there is no diff, I would prefer current implementation.

shahor02:
@peressounko it is important to use DPL fetcher to minimize the number of connections. As for using finaliseCCDB or taking the object only once from `pc.inputs().get...`, this is up to you.

peressounko:
Dear @shahor02, I think we can commit current version. It uses Fetcher or no CCDB access at all. I implemented other your comments

sawenzel:
This PR leads to crashes in unanchored MC (and will likely affect GRID productions). See for instance https://mattermost.web.cern.ch/aliperf-pdp/pl/bp1p6jzjijgudr3fot3gx54c9a. It would be good providing a default object on CCDB to which we can always fall-back.

PR #8457, opened 2022-03-28
Workaround FST to work on mac with raw-reader and NUMA

ktf:


davidrohr (commented):
Both changes make sense, but I'd prefer to do it slightly differently, since it is actually not only for MacOS :). But please leave it open and I'll create a similar PR for comparison soon.

PR #8462, opened 2022-03-29
Sanitize metadata keys, avoiding forbidden characters

chiarazampolli:
@sawenzel , @costing

### DIFF
--- a/CCDB/include/CCDB/CcdbApi.h
+++ b/CCDB/include/CCDB/CcdbApi.h
@@ -503,6 +503,12 @@ class CcdbApi //: public DatabaseInterface
 
   std::string getHostUrl(int hostIndex) const;
 
+  /**
+   * Function to sanitize the keys for metadata
+   * see https://developers.cloudflare.com/rules/transform/request-header-modification/reference/header-format/
+   */
+  std::map<std::string, std::string> sanitizeMetaData(std::map<std::string, std::string> const& metadata) const;
+
   /// Base URL of the CCDB (with port)
   std::string mUrl{};
   std::vector<std::string> hostsPool{};
--- a/CCDB/include/CCDB/CcdbApi.h
+++ b/CCDB/include/CCDB/CcdbApi.h
@@ -504,10 +504,10 @@ class CcdbApi //: public DatabaseInterface
   std::string getHostUrl(int hostIndex) const;
 
   /**
-   * Function to sanitize the keys for metadata
+   * Function to check the keys for metadata
    * see https://developers.cloudflare.com/rules/transform/request-header-modification/reference/header-format/
    */
-  std::map<std::string, std::string> sanitizeMetaData(std::map<std::string, std::string> const& metadata) const;
+  void checkMetadataKeys(std::map<std::string, std::string> const& metadata) const;
 
   /// Base URL of the CCDB (with port)
   std::string mUrl{};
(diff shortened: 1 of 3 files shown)


costing:
Ciao Chiara,

I'm of two minds about this. It's good to avoid problems later on, but it silently modifies what the user has passed, so that `my key` becomes `mykey` (if I understand it correctly). But from then on they would have to use `mykey` in all queries instead of the original string. Wouldn't it be better to throw a fatal error in this case and force a correct, consistent string to be used in both setting and retrieving later on?

Cheers,

.costin

chiarazampolli:
I don't have a strong opinion, I was asked if it was possible to fix this on the framework side. 
If the query will be done with the wrong key, then the object won't be found, and there will be fatal (unless disabled).

ktf:
Objects won't be found unless there is a matching key without the forbidden characters. IMHO this is dangerous, I would simply throw / FATAL if the key is considered to contain invalid characters.

chiarazampolli:
> Objects won't be found unless there is a matching key without the forbidden characters. IMHO this is dangerous, I would simply throw / FATAL if the key is considered to contain invalid characters.

I don't understand the objection: if the key that is passed is different from the one in CCDB, the object will not be found, and the code will have to be fixed. 
What you mean is that you prefer that even before that the user notices the wrong characters, while uploading the object, right? But this could also break the calibration workflow if the wrong key goes in by mistake. 

Anyway, I don't have a strong preference, I can change that there is fatal if there is a match instead of a replacement of the key. 
Should I do this, or we wait for the S&C meeting to see if there are further feedback?

sawenzel:
One could still offer this function to the user, so that one sanitizes before calling CcdbApi??

chiarazampolli:
@sawenzel , do you mean as an optional feature?

costing:
As there are other potential clients to ccdb, there is no guarantee that all calls would be sanitized via this utility. 

I am in favor of the fatal error in case of malformed strings. This would be caught immediately in the first test that the user will perform and corrected even before committing it.

I'll add a similar check on the server side, to reject uploads when invalid metadata keys are passed. That doesn't replace this functionality as an HTTP error code is too obscure to the user, the explicit error message is good to have.

jgrosseo:
I am also in favour of fatal in case of invalid strings.

chiarazampolli:
Agreed, i will change to fatal, then.

thanks for the feedback!

chiarazampolli:
Fatal implemented, is this ok?

sawenzel (dismissed):

sawenzel:
We have a small code checker complaint: `/sw/SOURCES/O2/8462-slc8_x86-64/0/CCDB/src/CcdbApi.cxx:1674:39: error: escaped string literal can be written as a raw string literal [modernize-raw-string-literal]`

chiarazampolli:
@sawenzel , thanks, it should be fixed.

chiarazampolli:
Hello, 

Any further comment, or can this be merged?

Chiara

chiarazampolli:
Hello, 

Can we merge this?

Chiara

costing (approved):

chiarazampolli:
No useful logs in the reported errors... The PR was green (unless I remember wrong) this morning.

shahor02:
This happened with a few PRs today, looks like a few builders failed to start the job. Merging

PR #8467, opened 2022-03-29
Material map

amorsch:
A new executable  _o2-sim-evalmat_  _allows_ to calculate maps of path integrated material properties (radiation length, absorption length and density). The calculation is performed during the transport of geantinos.
The geantinos are generated in the class _GeneratorGeantinos_.
The map is calculated in a helper class _MaterialBudgetMap_ owned by a specific _O2MCApplication_ class, _O2MCApplicationEvalMat_.
Three types of maps are supported:
_theta-phi
eta-phi
zv-phi_


sawenzel:
@amorsch : There are a few codechecker error (see log in full-system-test). Otherwise looking good.

sawenzel (approved):

PR #8469, opened 2022-03-29
TRD fix bounds error in mc to raw

bazinski:
- Memory over run found by @shahor02 fixed. while loop was not checking for exceeding the bounds of the tracklets and digits
- bounds checking also added to isTrackletOnLink and isDigitOnLink.

- mFEEID in Trap2Raw class changed from uint64_t to uint16_t, causing error on m1 for @ktf

### DIFF
--- a/Detectors/TRD/simulation/src/Trap2CRU.cxx
+++ b/Detectors/TRD/simulation/src/Trap2CRU.cxx
@@ -52,9 +52,9 @@ namespace trd
 {
 
 struct TRDCRUMapping {
-  uint32_t flpid;       // hostname of flp
-  uint32_t cruHWID = 0; // cru ID taken from ecs
-  uint32_t HCID = 0;    // hcid of first link
+  int32_t flpid;       // hostname of flp
+  int32_t cruHWID = 0; // cru ID taken from ecs
+  int32_t HCID = 0;    // hcid of first link
 };
 
 //this should probably come from ccdb or some authoritive source.
@@ -382,6 +382,9 @@ uint32_t Trap2CRU::buildHalfCRUHeader(HalfCRUHeader& header, const uint32_t bc, 
 bool Trap2CRU::isTrackletOnLink(const int linkid, const int currenttrackletpos)
 {
   //hcid is simply the halfcru*15+linkid
+  if (currenttrackletpos == mTracklets.size()) {
+    return false;
+  }
   int link = HelperMethods::getLinkIDfromHCID(mTracklets[currenttrackletpos].getHCID());
   if (linkid == link) {
     // this tracklet is on this link.
@@ -393,6 +396,9 @@ bool Trap2CRU::isTrackletOnLink(const int linkid, const int currenttrackletpos)
 bool Trap2CRU::isDigitOnLink(const int linkid, const int currentdigitpos)
 {
   Digit* digit = &mDigits[mDigitsIndex[currentdigitpos]];
+  if (currentdigitpos == mDigits.size()) {
+    return false;
+  }
   int link = HelperMethods::getLinkIDfromHCID(digit->getDetector() * 2 + (digit->getROB() % 2));
   if (link == linkid) {
     return true;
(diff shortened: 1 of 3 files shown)


martenole (dismissed):

Detectors/TRD/simulation/src/Trap2CRU.cxx
```diff
@@ -659,7 +659,7 @@ void Trap2CRU::convertTrapData(o2::trd::TriggerRecord const& triggerrecord, cons
   //set startdigit and starttracklet relative to the trigger.
   //
   int starttrackletindex = triggerrecord.getFirstTracklet();
-  int endtrackletindex = triggerrecord.getFirstTracklet() + triggerrecord.getNumberOfTracklets();
+  int endtrackletindex = triggerrecord.getFirstTracklet() + triggerrecord.getNumberOfTracklets() - 1;
```
shahor02: Don't you skip the last tracklet in this way, taking into account the `<` in the https://github.com/AliceO2Group/AliceO2/blob/f7b4ceb7e4ff96158d9ee41420d4b4b17125a2b4/Detectors/TRD/simulation/src/Trap2CRU.cxx#L726 ?

bazinski: yes, it just seemed more elegant to put the minus 1 at the top. Would you prefer it there?

shahor02: I meant with this definition of `endtrackletindex` you are checking all but the last tracklet, why the last one should not be checked?

bazinski: The problem is actually at line 494 and its associated loop. The -1 solved my initial tests. So was hoping it was the actual problem.

bazinski: @shahor02 ok that fixes it properly, it fixes the valgrind memcheck, there is no "invalid read" but I cant test mac of course.

ktf: it still breaks on mac.

bazinski: @TimoWilken the CI runs the full system test on mac as well or only centos8 ?

bazinski:
@ktf uint64_t removed which should fix the error you were seeing about 
[INFO] EP:0 CRU:0x0000 Link:15  FEEID:0x0008 SrcID:TRD[4] Packet:0   MemSize:0     OffsNext:0      prio.:0 HS:64 HV:6
[INFO] EP:1 CRU:0x0000 Link:15  FEEID:0x00e1 SrcID:TRD[4] Packet:0   MemSize:0     OffsNext:0      prio.:0 HS:64 HV:6 
vs for me : 
[INFO] EP:0 CRU:0x0000 Link:15  FEEID:0x0000 SrcID:TRD[4] Packet:0   MemSize:0     OffsNext:0      prio.:0 HS:64 HV:6
[INFO] EP:1 CRU:0x0000 Link:15  FEEID:0x0001 SrcID:TRD[4] Packet:0   MemSize:0     OffsNext:0      prio.:0 HS:64 HV:6 

FEEID is/was stored as a uint64_t.

The valgrind memory overrun that @shahor02 saw was fixed as well, although not related.

ktf:
These are all fine for me, but the actual culprit was #8481.

ktf:
@martenole @bazinski @shahor02 can this be merged as well?

bazinski:
oui, it solves the bounding error as well as some attempts at the ultimate #8481 solution of @ktf

Detectors/TRD/simulation/src/Trap2CRU.cxx
```diff
@@ -530,19 +537,22 @@ int Trap2CRU::buildTrackletRawData(const int trackletindex, const int linkid)
         destroytracklets = true;
         break;
     }
+    //this loop should really be redone but this will work for now.
     trackletcounter++;
+    if (trackletcounter + trackletindex >= mTracklets.size()) {
```
martenole, line 542: Is this a paranoia check or is this expected to happen? If it is a paranoia check maybe one could add a log message? Similar for the link checks for digits and tracklets

PR #8471, opened 2022-03-29
ITS calib: added possibility to run on different chips in parallel

iravasen:
Added flags `chip-mod-selector` and `chip-mod-base` (global flags) to run in parallel on different chips. Workflow name set accordingly (adding chip-mode-selector value to the name). 

Removed files ThresholdCalibrationWorkflow.cxx and .h since not used anymore with this last modification.


iravasen:
@shahor02 this addresses your suggestion on splitting the chip processing. Thanks a lot

iravasen:
@shahor02 if everything looks good, can you please merge this PR so that we can test it next week at P2? Sorry for the pressure, thanks a lot.

shahor02 (approved):

PR #8472, opened 2022-03-30
MCH: changed default time-reco-mode to bcreset

aferrero2707:
The new default corresponds to the case in which the SAMPA counters are reset at the beginning of each TF, which is the preferred option for the physics data taking.


aphecetche (approved):

aferrero2707:
@ktf @davidrohr @shahor02 some of the CI jobs have failed for reasons completely unrelated to the changes in the PR. Would it be possible to merge it such that we can have the changes included in the next software release? Thanks!

aferrero2707:
@shahor02 do you think we can force-merge this one as well?

shahor02 (approved):
trivial, merging

PR #8475, opened 2022-03-30
Fixing updating of TPC calibs from CCDB

matthias-kleiner:
- moving TPC code to separate functions

### DIFF
--- a/GPU/Workflow/src/GPUWorkflowSpec.cxx
+++ b/GPU/Workflow/src/GPUWorkflowSpec.cxx
@@ -89,6 +89,45 @@ using namespace o2::tpc;
 
 namespace o2::gpu
 {
+
+using ClusterGroupParser = o2::algorithm::ForwardParser<ClusterGroupHeader>;
+struct ProcessAttributes {
+  std::unique_ptr<ClusterGroupParser> parser;
+  std::unique_ptr<GPUO2Interface> tracker;
+  std::unique_ptr<GPUDisplayFrontend> displayFrontend;
+  std::unique_ptr<TPCFastTransform> fastTransform;
+  std::unique_ptr<TPCPadGainCalib> tpcPadGainCalib;
+  std::unique_ptr<TPCPadGainCalib> tpcPadGainCalibBufferNew;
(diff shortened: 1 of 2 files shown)


wiechula:
@davidrohr , wouldn't it in general be nicer to convert the workflow to a class inheriting from `Task`? This might look more structured.

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -89,6 +89,45 @@ using namespace o2::tpc;
...
+void initFunctionTPC(ProcessAttributes* processAttributes, const GPUSettingsO2& confParam);
+
+/// storing new calib objects in buffer
+void finaliseCCDBTPC(ProcessAttributes* processAttributes, gpuworkflow::Config const& specconfig, ConcreteDataMatcher& matcher, void* obj);
+
+/// asking for newer calib objects
+void fetchCalibsCCDBTPC(ProcessAttributes* processAttributes, gpuworkflow::Config const& specconfig, ProcessingContext& pc);
+
+/// storing the new calib objects by overwritting the old calibs
+void setCalibsTPC(ProcessAttributes* processAttributes);
```
davidrohr: perhaps we can find a better name here, like `storeUpdatedCalibsTPCPtrs` instead of `setCalibsTPC` to make it clear what happens here.

davidrohr (changes requested):
This is exactly what I meant in the mail exchange, except for one minor comment on a misleading function name below.

matthias-kleiner:
> This is exactly what I meant in the mail exchange, except for one minor comment on a misleading function name below.

Okay, thanks for checking the code. I renamed the function and squashed the commits

davidrohr:
> @davidrohr , wouldn't it in general be nicer to convert the workflow to a class inheriting from `Task`? This might look more structured.

Yes, I agree, but I'd try to get this in first, then we can try to change it perhaps. But migrating to a task was not a priority yet, since I don't see any benefit except for being more in line with other workflows. Note this one was developed before this task class was available.

PR #8478, opened 2022-03-30
DPL: fix CCDB fetcher verbosity

ktf:


ktf:
@Barthelemy this should fix the flooding.

@davidrohr @rbx @dennisklein do we have anything which is severity below info but not compiled out in optimised builds?

dennisklein:
> do we have anything which is severity below info but not compiled out in optimised builds?

Not sure, if I can follow, does https://github.com/FairRootGroup/FairLogger#3-severity and https://github.com/FairRootGroup/FairLogger#31-compile-time-severity-switch answer your question?

ktf:
No, I think it does not, because between info and debug there is no intermediate log levels. In general `LOG(debug)` statements have been put under the assumption they will be completely excluded if NDEBUG is set. What I want is an "almost-info-but-more-than-debug" ("details"?) level which can be used as minimun enabled severity at compile time, yet it would be generally discarded at runtime (so that it does so clobber infologger). The usecase would be to use it for all those cases where in general we do not want to see those messages, yet they are cheap enough to format that we leave them in for the optimised builds.

Barthelemy:
thank you

PR #8481, opened 2022-03-31
Fix uninitialised unused bits in FEEID

ktf:


ktf:
Obviously undefined behaviour and working by accident. Merging.

PR #8482, opened 2022-03-31
its-vertexer: Use single vector to store number of found tracklets

mconcas:
Using a single flat buffer to store this info will be more impactful on the GPU code, this pr is to align the two approaches.

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/include/ITStracking/TimeFrame.h
+++ b/Detectors/ITSMFT/ITS/tracking/include/ITStracking/TimeFrame.h
@@ -136,11 +136,11 @@ class TimeFrame final
   // Vertexer
   void computeTrackletsScans();
   std::vector<int>& getIndexTableL0(int tf);
-  std::array<std::vector<int>, 2>& getNTrackletsCluster(int tf);
   int& getNTrackletsROf(int combId, int tf);
   std::vector<Line>& getLines(int tf);
   std::vector<ClusterLines>& getTrackletClusters(int tf);
   gsl::span<const Tracklet> getFoundTracklets(int rofId, int combId) const;
+  gsl::span<int> getNTrackletsCluster(int rofId, int combId);
   // \Vertexer
 
   void initialiseRoadLabels();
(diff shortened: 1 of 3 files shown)


iouribelikov:
Hi @mconcas,
It works with our normal workflow.  But if I try to run it with the \<O2\>/macro/run_trac_ca_its.C macro, not a single vertex is found.
 Should not some parameters of the macro be adjusted ?

mconcas:
> Hi @mconcas, It works with our normal workflow. But if I try to run it with the <O2>/macro/run_trac_ca_its.C macro, not a single vertex is found. Should not some parameters of the macro be adjusted ?

It's weird, I just tried with the macro and it finds the vertices...

iouribelikov:
> > Hi @mconcas, It works with our normal workflow. But if I try to run it with the /macro/run_trac_ca_its.C macro, not a single vertex is found. Should not some parameters of the macro be adjusted ?
> 
> It's weird, I just tried with the macro and it finds the vertices...

Weird. Could not it be that you ran another macro, the run_trac_its.C ?   That one is OK.
But the run_trac_ca_its.C does not work for me.  I'll try to look inside...

mconcas:
> Hi @mconcas, It works with our normal workflow. But if I try to run it with the <O2>/macro/run_trac_ca_its.C macro, not a single vertex is found. Should not some parameters of the macro be adjusted ?

ERRATA CORIGE: you are right. 
It seems the vertices are not correctly stored in the output file. I'll have a look to it later, but I would not like to rerun the whole CI to fix a bug in this macro.
Especially seeing that tracks are found and stored correctly.

Can we agree I'll fix it in next PR?

iouribelikov (approved):

iouribelikov:
> > Hi @mconcas, It works with our normal workflow. But if I try to run it with the /macro/run_trac_ca_its.C macro, not a single vertex is found. Should not some parameters of the macro be adjusted ?
> 
> ERRATA CORIGE: you are right. It seems the vertices are not correctly stored in the output file. I'll have a look to it later, but I would not like to rerun the whole CI to fix a bug in this macro. Especially seeing that tracks are found and stored correctly.
> 
> Can we agree I'll fix it in next PR?



> > Hi @mconcas, It works with our normal workflow. But if I try to run it with the /macro/run_trac_ca_its.C macro, not a single vertex is found. Should not some parameters of the macro be adjusted ?
> 
> ERRATA CORIGE: you are right. It seems the vertices are not correctly stored in the output file. I'll have a look to it later, but I would not like to rerun the whole CI to fix a bug in this macro. Especially seeing that tracks are found and stored correctly.
> 
> Can we agree I'll fix it in next PR?

OK. No problem.  We can fix this macro later.

mconcas:
@shahor02 hi, I don't know what happend to the CI but this used to be green for the past 4 days. I think it's ready to be merged.

shahor02 (approved):
merging as the tests were previously passed.

PR #8483, opened 2022-03-31
Make IRFrame reader/writer global devices and workflows

shahor02:
to be used for CTF skimming


shahor02:
failure is unrelated, merging.

PR #8485, opened 2022-03-31
Suppress unnecessary and wrong check of CCDB obj. pointer

shahor02:
@peressounko As I wrote to you before, you don't need to check the pointer of the ccdb object returned by the DPL fetcher: if the fetching fails, the fatal will be produced already on the fetcher level. You also don't need to use complicated 
```
    std::decay_t<decltype(ctx.inputs().get<o2::phos::BadChannelsMap*>("badmap"))> badMapPtr{};
    badMapPtr = ctx.inputs().get<o2::phos::BadChannelsMap*>("badmap");
    const o2::phos::BadChannelsMap* badMap = badMapPtr.get();
```
you can simply do
```
auto badMapPtr = ctx.inputs().get<o2::phos::BadChannelsMap*>("badmap");
const o2::phos::BadChannelsMap* badMap = badMapPtr.get();
```

Merging since currently the reconstruction is broken

### DIFF
--- a/Detectors/PHOS/workflow/src/CellConverterSpec.cxx
+++ b/Detectors/PHOS/workflow/src/CellConverterSpec.cxx
@@ -69,11 +69,7 @@ void CellConverterSpec::run(framework::ProcessingContext& ctx)
 
   // get BadMap from CCDB, once
   if (!mHasCalib) {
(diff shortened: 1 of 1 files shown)


peressounko:
Dear @shahor02, I remember I implemented your suggestion. But then It looks like I reverted older version of my code and committed it. I will remove remnants in next commit.

PR #8489, opened 2022-04-01
DPL: introduce ConsumeAndRescan as a completion policy

ktf:


ktf:
@davidrohr this should allow the rescanning needed by #7536. Just return `ConsumeAndRescan` there when this is merged.

ktf:
All green, and in any case not triggered by anyone.

PR #8493, opened 2022-04-03
Optionally check cached ccdb objects only every N TFs

shahor02:
with `--condition-tf-per-query N (def=1)` the validity of cached objects for TF timestamp will
be checked every `N` TFs. If N <= 0, never check validity once they are cached.

@peressounko

### DIFF
--- a/Framework/Core/src/CCDBHelpers.cxx
+++ b/Framework/Core/src/CCDBHelpers.cxx
@@ -50,6 +50,7 @@ struct CCDBFetcherHelper {
   std::vector<OutputRoute> routes;
   std::map<int64_t, CCDBCacheInfo> cache;
   std::unordered_map<std::string, std::string> remappings;
+  size_t queryDownScaleRate = 1;
   o2::ccdb::CcdbApi& getAPI(const std::string& path)
   {
     auto entry = remappings.find(path);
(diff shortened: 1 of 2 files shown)


shahor02:
Tests were green before, merging.

sevdokim:
Hi @shahor02 
this nice option changes checking rate for all the objects in the workflow, right?
It would be cool if one can change checking rate for different ccdb objects separately. For example, check CPV/Calib/Pedestals only once at beginning of run but CPV/Calib/BadChannelMap can be checked once per 10000 TFs. 
Also it will be useful for huge workflows such as dpl-workflow.sh to change checking rates individually for different detectors.

shahor02:
@sevdokim theoretically this can be done but would require passing the command line options for plenty of objects. Before diving to this: what is the use case? Are you going to upload new BadChannel objects online during the sync. processing? Note that with localhost:8084 the additional benefit from the prescaling the validity checks can be only on the FLPs, where the TF rate is order of magnitude higher than on the EPN. Though, even on the EPN in the sync workflow update only at start (or ar very low rate) should be enough.

sevdokim:
Hi @shahor02 I'm thinking about using this option for CPV processing: it would be enough not to check Pedestals, Gains and BadChannelMap objecgts for CPV during the run. However I think no one from other detectors will agree on that. 
For now I see this option can be used only for FLP processing detectors where standalone workflow is running.

sevdokim:
Now I see this option re-opens door for CPV FLP processing.

shahor02:
@sevdokim in general, in the sync. workflow we don't need to check the ccdb at high rate. And the EPN workflows anyway have much lower TF rate than those on the FLP. 
I agree that it would be good to have this as an option, but it may require quite some changes, so I would keep it at lower priority at the moment. But do you still see a backpressure with localhost if you don't use the feature added by this PR?

sevdokim:
Hi @shahor02 
to be honest I didn't check processing on flp recently. My thoughts are based on ~1 month late expirience. I'll check.

PR #8503, opened 2022-04-05
[WIP][EMCAL-687] Declaring BC as index column for Calorimeter tables

mfasDa:


mfasDa:
The build currently crashes in test_AnalysisTask.cxx when creating a join of the two calorimeter tables:
```c++
void process(o2::aod::Collision const&, o2::soa::Join<o2::aod::Tracks, o2::aod::TracksExtra, o2::aod::TracksCov> const&, o2::aod::AmbiguousTracks const&, o2::soa::Join<o2::aod::Calos, o2::aod::CaloTriggers> const&)
```
Because both declare separate index column. According to my understanding this would mean that calos and CaloTrigger tables are supposed to be related in rows (same objects in both tables, only different parts of the information). They are not, they come from different readouts, represent different segmentations (for EMCAL the trigger has only 1/4 of the amount of channels compared to the cells) and must therefore never be joined but treated as separate objects.

For EMCAL we definitely want to declare index columns so the cells and FastORs are filtered per BC ID.

jgrosseo:
I think the title of the PR is misleading. Adding `o2::soa::Index<>` does not declare anything as index column. It just adds some functions to get the row number etc. @aalkin may want to comment

If is see correctly you don't change any other table definition, right?

mfasDa:
@jgrosseo I see. The problem I want to fix is that for the emcal correction task subscribing to o2::emcal::calos it obviously did not get only the cells from a single bunch crossing but from all bunch crossings in the time frame, see for example
```
[340:emcal-correction-task]: [00:57:10][INFO] Converted cells. Contains: 279. Originally 2241. About to run clusterizer.
[340:emcal-correction-task]: [00:57:10][INFO] Converted cells. Contains: 264. Originally 2241. About to run clusterizer.
[340:emcal-correction-task]: [00:57:10][INFO] Converted cells. Contains: 287. Originally 2241. About to run clusterizer.
[340:emcal-correction-task]: [00:57:10][INFO] Converted cells. Contains: 265. Originally 2241. About to run clusterizer.
[340:emcal-correction-task]: [00:57:10][INFO] Converted cells. Contains: 308. Originally 2241. About to run clusterizer.
[340:emcal-correction-task]: [00:57:10][INFO] Converted cells. Contains: 285. Originally 2241. About to run clusterizer.
[340:emcal-correction-task]: [00:57:10][INFO] Converted cells. Contains: 290. Originally 2241. About to run clusterizer.
[340:emcal-correction-task]: [00:57:10][INFO] Converted cells. Contains: 263. Originally 2241. About to run clusterizer.
```
where this printout is called for each BC, "Contains" refers to the cells with the same BC ID and "Originally" is the size of o2::aod::calos, so if BC and calos would be matched the numbers should be identical.

jgrosseo:
I guess you refer to 
https://github.com/AliceO2Group/O2Physics/blob/master/PWGJE/TableProducer/emcalCorrectionTask.cxx#L129

Calo cannot be directly grouped to Collisions because you need the indirection through the BC. Then you loop by hand in your code and build a new list but that is not efficient. If you want to start from Collision, we need to do a similar trick to this here:
https://github.com/jgrosseo/AliceO2/blob/dev/Framework/Core/include/Framework/AnalysisDataModel.h#L1039

You can check how this is used e.g. here:
https://github.com/AliceO2Group/O2Physics/blob/master/Tutorials/src/ZDCVZeroIteration.cxx#L80

Let me know if this helps.

mfasDa:
@jgrosseo Indeed this new list I would like to get rid of. 

In principle what I would like to do is to group both collision and Calos to BC, since the index column should be the BC for the calorimenter, and the collision is only needed getting the primary vertex for the cluster kinematics, which we could set to 0 in case the bc doesn't have a primary vertex assigned. However not sure this is possible.

jgrosseo:
Try process(BC, Collisions, Calos)

mfasDa:
@jgrosseo Thanks! Also found out myself in the mean time, see https://github.com/AliceO2Group/O2Physics/pull/661

jgrosseo:
I will merge this but adjust the commit message

jgrosseo (approved):

PR #8504, opened 2022-04-05
fix for laser data

AllaMaevskaya:


mslupeck (approved):

AllaMaevskaya:
build/O2/fullCI  error not relate to my code

davidrohr (commented):
Should we generally disable this?
If this is corrupt data, do you want to at least print an alarm in case of non-laser data?

Detectors/FIT/FT0/calibration/src/FT0ChannelTimeTimeSlotContainer.cxx
```diff
@@ -63,8 +63,6 @@ void FT0ChannelTimeTimeSlotContainer::fill(const gsl::span<const FT0CalibrationI
       mHistogram[chID]->Fill(chTime);
       ++mEntriesPerChannel[chID];
       LOG(debug) << "@@@@entries " << mEntriesPerChannel[chID] << " chID " << int(chID) << " time " << chTime << " tiestamp " << uint64_t(entry.getTimeStamp());
-    } else {
-      LOG(fatal) << "Invalid channel data";
```
shahor02, line 67: is this protection against coding error or corrupted data? Normally, the latter should not lead to fatal but raise an alarm (LOG(alarm)) and pass over it.

AllaMaevskaya: This is protection was done before look-up table implementation, now its not actual. Later reading of dead channel map will be add

shahor02: So, shall I close this PR?

AllaMaevskaya: Please  merge it. We have problem to test calibration @P2 with laser data with addition channels for reference PMTs.

PR #8505, opened 2022-04-05
[WIP]  Kr calibration

jcrkovsk:


martenole:
Hi @jcrkovsk 

Thanks for opening this PR. So this class is supposed to do only the upload of the calibration results to the CCDB. Can the code for the actual calibration also be added? And then the upload itself could also be just a simple macro. I don't see a need for a class, since this will be executed only manually in any case I assume.
And it would be very interesting to see the effect of the new calibration object you have uploaded. Can you run a simulation and use once the default object for the local gain factor and once the new one and then check the differences in the output?

Cheers,
Ole

ping also to @tdietel because it seems notifications were not sent automatically by github.

jcrkovsk:
> 

Hi @martenole ,

I pushed new changes to the code.
The KrCalibration class now contains methods to fill and normalize the gain per each detector with at least 1 hit.
I also added a macro that runs the calibration and can upload to ccdb.

Regarding the simulation, I don't understand what exactly you ask me to do.

Cheers,
Jana

(ping @tdietel)

Detectors/TRD/calibration/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ add_subdirectory(macros)
 o2_add_library(TRDCalibration
                SOURCES src/TrackBasedCalib.cxx
                        src/CalibratorVdExB.cxx
+                       src/KrCalibration.cxx
```
tdietel: I find the name KrCalibration misleading- I would expect it runs the actual calibration. It seems this class prepares the data for the upload, and the name should reflect that.

Detectors/TRD/calibration/src/KrCalibration.cxx
```diff
@@ -0,0 +1,67 @@
...
+ClassImp(o2::trd::KrCalibration);
+
+namespace o2::trd
+{
+
+void KrCalibration::writeToCCDB(TTree* tree)
+{
+    ////write map to CCDB
+    o2::ccdb::CcdbApi ccdb;
+    ccdb.init("http://ccdb-test.cern.ch:8080");
```
tdietel: The test CCDB should not be hardcoded, as it would require recompiling O2 to upload to the production DB.

Detectors/TRD/calibration/include/TRDCalibration/KrCalibration.h
```diff
@@ -0,0 +1,62 @@
...
+class KrCalibration
+{
+ public:
+  KrCalibration() {};
+  ~KrCalibration() {};
+
+  float ComputeDetectorAverage(TH2F* hDet);
+  TH2F* CreateNormalizedMap(TH2F* hDet, TString sNewName = "");
+  void FillInTheGap(TH2F* hDet, int column, int row , float newGain);
+  TH2F* FillTheMap(TH2F* hDet, TString sNewName = ""); // TODO change to void?
```
tdietel: What is the TODO about? Does this method return a new TH1D, or modify the original one?

Detectors/TRD/calibration/include/TRDCalibration/KrCalibration.h
```diff
@@ -0,0 +1,62 @@
...
+  std::vector<std::vector<int>> FindEmpty(TH2F* hDetectorMap);
+  float GetAverageFromNeighbors(TH2F* hDet, int column, int row); // TODO float
```
tdietel: What's the TODO comment here?

Detectors/TRD/calibration/src/KrCalibration.cxx
```diff
@@ -0,0 +1,225 @@
...
+}
+
+
+TH2F* KrCalibration::GetDetectorMap( TTree* tree, int nDet, float mingain, float maxgain, TString sDetName )
+{ // creates a th2 map of a detector nDet
+  // allows to limit the range of accepted adc gain by setting 
+  // mingain and maxgain (default range 0-10k)
+    if ( sDetName == "" ) sDetName = Form("hDet%i",nDet);
+    int nTrdRows = NROWC1;
+    int detRow = nDet % 30; // set the correct # of rows for stack 2
```
tdietel: I was confused by detRow - I would find something like detInSupermodule more readable.

Detectors/TRD/calibration/src/KrCalibration.cxx
```diff
@@ -0,0 +1,225 @@
...
+        tree->GetEntry(ientry);
+        if( adc < 0 || adc > 10'000 ) continue; // limit the accepted gain
+        calObject.setPadValue((int)det,(int)col,(int)row,adc);
+    }
+    std::cout << "created and filled the calibration object" << std::endl;
+
+    ////upload temporarily
+    auto timeStamp = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+    auto timeStampEnd = timeStamp;
+    timeStampEnd += 1e3 * 60 * 60 ; // set validity to 1 hour
```
tdietel: I think the lifetime should be configurable. The start time would typically be the time when the krypton run was taken, and the end time would be undefined (and superseded by the next krypton run).

It also seems like this function is duplicated in the macro. Consider removing it here, or only keep the part that builds the LocalGainFactor object and returns it to the macro, so the macro can upload it to the CCDB.

Detectors/TRD/macros/calibrateKr.C
```diff
@@ -0,0 +1,38 @@
...
+
+    TFile* file = TFile::Open( sOpenFile );
+    TTree* tree = (TTree*) file->Get("nt_Krypton");
+    LocalGainFactor calObject; 
+    TH2F* hDet[540];
+
+    for( int idet = 0; idet < 540; idet++ ) {
+        TH2F* hDetDef = krCalibration->GetDetectorMap( tree, idet);
+        TH2F* hDetFilled = krCalibration->FillTheMap(hDetDef);
+        hDet[idet] = krCalibration->CreateNormalizedMap(hDetFilled);
```
tdietel: I don't see why there is a need to keep all 540 hDet histograms in memory. I suggest to not create them as an array - the LocalGainFactor object should still be the same.

Detectors/TRD/macros/calibrateKr.C
```diff
@@ -0,0 +1,38 @@
...
+        for(int irow = 0; irow < hDet[idet]->GetNbinsY(); irow++ ) {
+            for(int icol = 0; icol < hDet[idet]->GetNbinsX(); icol++ ) {
+                float relativeGain = hDet[idet]->GetBinContent(hDet[idet]->GetXaxis()->FindBin(icol), hDet[idet]->GetYaxis()->FindBin(irow));
+                calObject.setPadValue(idet, icol, irow, relativeGain);
+            }
+        }
+    }
```
tdietel: By default, you no longer need hDetDef, hDetFilled, hDet after this loop, so delete them at the end, or reuse them.

Detectors/TRD/macros/calibrateKr.C
```diff
@@ -0,0 +1,38 @@
...
+
+    ///write map to CCDB
+    o2::ccdb::CcdbApi ccdb;
+    ccdb.init("http://ccdb-test.cern.ch:8080");
+    // o2::base::NameConf::getCCDBServer()
+    std::map<std::string, std::string> metadata;
+
+    ///upload temporarily
+    auto timeStamp = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
```
tdietel: Same comments about validity as in the class, but in the macro at least it's easy to adjust without a need to recompile O2.

Detectors/TRD/macros/calibrateKr.C
```diff
@@ -0,0 +1,38 @@
+#include "TRDCalibration/KrCalibration.h"
+
+using namespace o2::trd;
+
+void calibrateKr( TString sOpenFile )
```
tdietel: As we discussed, this macro does not run the actual calibration, but - as I understand it - reads the values for the calibrated pads from a tree, fills any gaps, subtracts the average, and fills the LocalGainFactor object. A name like CreateCCDBLocalGainFactor might be more appropriate.

tdietel (changes requested):
Hello,

I have a few comments, and I think renaming the class and the macro would make the purpose clearer. 

Otherwise there are a few minor comments that should be quick to address.

Detectors/TRD/calibration/include/TRDCalibration/KrCalibration.h
jcrkovsk: It creates a new TH2F. I forgot to delete the TODO comment before committing.

Detectors/TRD/calibration/include/TRDCalibration/KrCalibration.h
jcrkovsk: I forgot to delete the TODO comment before committing.

Detectors/TRD/calibration/src/KrCalibration.cxx
jcrkovsk: OK

Detectors/TRD/calibration/src/KrCalibration.cxx
jcrkovsk: I removed this method. This can and will be handled by a macro.

Detectors/TRD/calibration/src/KrCalibration.cxx
jcrkovsk: I removed this method. This can and will be handled by a macro.

Detectors/TRD/macros/calibrateKr.C
jcrkovsk: OK

Detectors/TRD/macros/calibrateKr.C
jcrkovsk: OK

Detectors/TRD/calibration/CMakeLists.txt
jcrkovsk: Renamed to KrCalibUploadToCCDB.

Detectors/TRD/macros/calibrateKr.C
jcrkovsk: Done

Detectors/TRD/macros/calibrateKr.C
jcrkovsk: OK, will do that. For now I keep them hardcoded and set starting date as the (afaik) first date with Kr run and end date as Jan 1, 2040.
I tried to create a function which would take a std::string with date and time and create the stamp from there but I keep getting unexpected dates this way. This will be resolved but for now the priority is to have something that functions the way I want it to.

Detectors/TRD/calibration/include/TRDCalibration/KrCalibUploadToCCDB.h
```diff
@@ -31,20 +31,19 @@ namespace o2
 namespace trd
 {
 
-class KrCalibration
+class KrCalibUploadToCCDB
```
tdietel: Actually, the name is still misleading: this class does not upload to the CCDB, it prepares the CCDB object. A better name might be something like PadCalibCCDBBuilder.
@bazinski, @martenole  is there some convention on naming such classes?

Detectors/TRD/macros/calibrateKr.C
tdietel: Ok. As long as the date is in the .C macro that's fine. Might be good to have that under version control anyways, rather than relying on remembering the right parameters after the next krypton run.

Detectors/TRD/calibration/include/TRDCalibration/KrCalibUploadToCCDB.h
tdietel: Hello. I think we are close to merging this PR. Should we invent a new name, or stick with the current one? @martenole , @bazinski ?

martenole:
Hi, concerning the name of the class I would also try to use something which reflects more what is actually done. So maybe something like `PadGainMapBuilder`?
The name also needs to be changed in the CMakeLists file where its still `KrCalibration`.
@jcrkovsk could you please also apply the formatting rules? This can be done automatically with the command explained here: https://github.com/AliceO2Group/CodingGuidelines#check-commits-formatting
Then you could mark the PR as ready to review which will also trigger other code checks.
Cheers,
Ole

bazinski:
can you collapse some of those commits, particularly the "meh" ones :-D

PR #8509, opened 2022-04-06
Write metafile for TPC residuals

martenole:
And make output directory for residuals configurable

At the moment this is WIP. Will add the TF statistics information in an additional commit.

### DIFF
--- a/Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
+++ b/Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
@@ -112,11 +112,12 @@ void ResidualsContainer::fill(const o2::dataformats::TFIDInfo& ti, const gsl::sp
     auto& residVecOut = residuals[sec];
     auto& statVecOut = stats[sec];
     std::array<unsigned char, TrackResiduals::VoxDim> bvox;
+    float xPos = param::RowX[residIn.row];
     float yPos = residIn.y * param::MaxY / 0x7fff;
     float zPos = residIn.z * param::MaxZ / 0x7fff;
-    if (!trackResiduals->findVoxelBin(sec, param::RowX[residIn.row], yPos, zPos, bvox)) {
+    if (!trackResiduals->findVoxelBin(sec, xPos, yPos, zPos, bvox)) {
       // we are not inside any voxel
-      LOGF(debug, "Dropping residual in sec(%i), x(%f), y(%f), z(%f)", sec, param::RowX[residIn.row], yPos, zPos);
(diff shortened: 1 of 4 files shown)


martenole:
Waiting with this until #8533 is merged which will give me easy access to the TF ID information

martenole:
Hi @shahor02 
this would be ready for review now.
Cheers,
Ole

Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
```diff
@@ -182,7 +205,7 @@ bool ResidualAggregator::hasEnoughData(const Slot& slot) const
 
 void ResidualAggregator::initOutput()
 {
-  // nothing to be done here, but method needs to be overwritten
+  mFileMetaData = std::make_unique<o2::dataformats::FileMetaData>();
```
shahor02: Why do you need this? It is anyway assigned at L227. You could actually use a local FileMetaData object at that line, w/o any smart pointers. I will also change this in the CTFWriter.

Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
```diff
@@ -195,15 +218,41 @@ void ResidualAggregator::finalizeSlot(Slot& slot)
...
   cont->treeOutStats.reset();
+  cont->treeOutRecords->Write();
+  cont->treeOutRecords.reset();
   cont->fileOut->Close();
   cont->fileOut.reset();
+  std::filesystem::rename(o2::utils::Str::concat_string(mOutputDir, cont->fileName, ".part"), mOutputDir + cont->fileName);
+  if (mStoreMetaData) {
+    mFileMetaData = std::make_unique<o2::dataformats::FileMetaData>();
+    mFileMetaData->fillFileData(mOutputDir + cont->fileName);
+    // mFileMetaData->run = getTFrunNumber();
```
shahor02: Why did you comment setting the run number?

Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
martenole: Sorry I did not really think about it when I took that from the CTF writer.. It's changed now to a local object

Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
martenole: I actually forgot to fix this, that is why. I am taking the run number now from the TimeSlot which in turn gets it through the TFIDInfo. Thanks for spotting!

martenole:
Hi @shahor02 do you think this could be merged?
Cheers, Ole

shahor02 (approved):

PR #8510, opened 2022-04-06
[EMCAL-769] Catch invalid DDL IDs in Mapper

mfasDa:
- Use unsigned integers for DDL ID in
  mapping handler and check for DDL
  ID < 40
- Catch DDLInvalid in RawToCellConverter
  and print error message in infoBrowser

### DIFF
--- a/Detectors/EMCAL/base/include/EMCALBase/Mapper.h
+++ b/Detectors/EMCAL/base/include/EMCALBase/Mapper.h
@@ -345,13 +345,13 @@ class MappingHandler
   /// \param ddl ID of the DDL for which to get the mapping
   /// \return Mapping for the DDL (if valid)
   /// \throw DDLInvalid if DDL is invalid for EMCAL
-  Mapper& getMappingForDDL(int ddl);
+  Mapper& getMappingForDDL(unsigned int ddl);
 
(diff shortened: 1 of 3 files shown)


mfasDa:
Issue spotted on EPN today indecated that the mapping is called for a corruped DDL ID (must have been implicitly casted to a negative value). In order to prevent such cases I changed the ddl ID to unsigned int and throw a DDLInvalid exception which is caught in the RawToCellConverter leaving a message (level "error") in the infoBrowser. A dedicated QC component monitoring such raw data errors in order to decrease verbosity on the infoBrowser.

davidrohr (approved):
thx, will merge once CI passes

mfasDa:
This must be an extremely rare data corruption - we haven't seen this before. Based on the stack trace the hwaddress cannot be the issue, because it is valid and the corresponding functions are protected with exceptions which we catch in the raw to cell converter. The DDL was the only place a corruption could sneak in as internally an array is used to handle the mapping, which could go out-of-bounds, and here the protection was indeed weak as implicit casting could lead to out-of-bounds in case of corrupted DDL ID. The protection is now reflecting the the DDL index range.

PR #8514, opened 2022-04-07
its-gpu: use TimeframeGPU for GPU vertexer

mconcas:
In this PR we have:
 - [x] adaptation of the its-vertexer to use the `o2::its::gpu::TimeFramegpu` memory layout
 - [x] cleanup of old classes (namely: `deviceStore*` and `standalonedebugger`)
 - [x] streamlining of its gpu code


mconcas:
@davidrohr : I think this can now be processed. As I suspected there is still something to be fixed in the GPU vertexer as it does not find a vertex anymore 😞 . Since combinatorial trackleting and tracklet matching is working, I am currently investigating where it misses the point in the vertex finding phase.

@mpuccio : due to the extensive changes to other tracking files and to the fact that we are not yet using ITS GPU code, we agreed with David on splitting the work on the porting of GPU vertexing to TimeFrame based approach in multiple PRs.
 
Notable changes in this PR:
 - Removal of `TrackerTraitsCPU.h` in favour of a default implementation of `TrackerTraits.h`. 
 - The standalone debugging class is gone.
 - We get tracker traits from GPUreco.
 - Data store for vertexer (symmetric to the `devicestoreNV`) is superseded by `TimeFrameGPU`, so it will be the gpu-tracker one.

Common/MathUtils/include/MathUtils/CartesianGPU.h
```diff
@@ -26,8 +26,8 @@ template <typename T, int I>
 struct GPUPoint2D {
   GPUdDefault() GPUPoint2D() = default;
   GPUd() GPUPoint2D(T a, T b) : xx(a), yy(b) {}
-  GPUd() float X() const { return xx; }
-  GPUd() float Y() const { return yy; }
+  GPUhd() float X() const { return xx; }
```
mconcas, line 29: @davidrohr please see these kind of changes. Here I think they are harmless.

DataFormats/Reconstruction/include/ReconstructionDataFormats/Vertex.h
```diff
@@ -40,7 +40,7 @@ class VertexBase
                         kCovYZ,
                         kCovZZ };
   static constexpr int kNCov = 6;
-  GPUdDefault() VertexBase() = default;
+  GPUhd() VertexBase() = default;
```
mconcas: @davidrohr  Whereas here I'm not 100% sure.

Common/MathUtils/include/MathUtils/CartesianGPU.h
davidrohr: In principle they should be harmless for inline functions, not to be used for non-inline functions.
But basically, the host version is only needed if it is called by the host part during GPU compilation, which should usually not be the case?

DataFormats/Reconstruction/include/ReconstructionDataFormats/Vertex.h
davidrohr: Not sure if this will pass the fullCI, let's see. Again, this means you are creating a VertexBase() in the host part during GPU compilation. Is that needed?

mconcas: E.g.: when you want to pull back vertices from GPU to plug into CPU tracker. I just had to add this for that reason.

davidrohr: OK, so you have also host code that works on the data in the `.cu` / `.hip` files. This is fine with me as long as the FullCI is green.

mconcas:
The failure seems to be not genuine (CI has been green over the entire weekend) @davidrohr

PR #8516, opened 2022-04-07
[MCH MID] enable input/output tracks readers/writer from/to root file

pillot:
This will allow to add this workflow to the full system test with the standard input/output options

### DIFF
--- a/Detectors/MUON/Workflow/src/tracks-matcher-workflow.cxx
+++ b/Detectors/MUON/Workflow/src/tracks-matcher-workflow.cxx
@@ -16,6 +16,7 @@
 
 #include <string>
 #include <vector>
+#include "Framework/CompletionPolicyHelpers.h"
 #include "Framework/ConfigParamSpec.h"
 #include "Framework/ConfigContext.h"
 #include "Framework/WorkflowSpec.h"
@@ -27,6 +28,12 @@
 #include "TrackWriterSpec.h"
 
 using namespace o2::framework;
+
+void customize(std::vector<CompletionPolicy>& policies)
+{
+  // ordered policies for the writers
+  policies.push_back(CompletionPolicyHelpers::consumeWhenAllOrdered(".*(?:MUON|muon).*[W,w]riter.*"));
+}
 
 // we need to add workflow options before including Framework/runDataProcessing
 void customize(std::vector<ConfigParamSpec>& workflowOptions)
--- a/Detectors/MUON/Workflow/src/tracks-matcher-workflow.cxx
+++ b/Detectors/MUON/Workflow/src/tracks-matcher-workflow.cxx
@@ -20,7 +20,11 @@
 #include "Framework/ConfigContext.h"
 #include "Framework/WorkflowSpec.h"
 #include "CommonUtils/ConfigurableParam.h"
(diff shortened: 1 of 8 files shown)


pillot:
@aphecetche I switched the PR to draft because I just realized that these changes in the workflow will also require changes in sim_challenge.sh and o2dpg_sim_workflow.py that are both using it. Simply the track reader and writer workflows will have to be removed since they will be both activated by default in the matching workflow itself. I will do that after vacation and also update the doc by the way.

aphecetche (dismissed):

aphecetche:
@pillot can this one be squashed or do you want to keep the 3 commits separated ?

pillot:
It can be squashed.

aphecetche:
@pillot ok (I asked because I've seen the PR was green and thought I could merge it but in fact I can not...)

pillot:
@sawenzel can you merge this PR please ?

Detectors/MUON/Workflow/src/tracks-matcher-workflow.cxx
```diff
@@ -36,6 +46,30 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
...
+  if (!disableRootInput) {
+    specs.emplace_back(o2::mch::getTrackReaderSpec(useMC, "mch-track-reader"));
+    specs.emplace_back(o2::mid::getTrackReaderSpec(useMC, "mid-track-reader"));
+  }
+
+  specs.emplace_back(o2::muon::getTrackMatcherSpec("muon-track-matcher"));
+
+  if (!disableRootOutput) {
+    specs.emplace_back(o2::muon::getTrackWriterSpec("muon-track-writer", "muontracks.root"));
+  }
```
shahor02, line 73: Would be better to preserve the detector name in the device name. In any case, the writers must be subjected to completion policy ensuring that the TFs are stored in the ordered way (in case of pipe-lining processing).
You should add to the workflow definition a block like:
https://github.com/AliceO2Group/AliceO2/blob/f153b1e14e688b7dcbb47a0a2f9422c9e9f671b6/Detectors/MUON/MCH/Workflow/src/clusters-writer-workflow.cxx#L19-L23
Note that if you preserve the name `muon-track-writer`, you will need either to adapt the line to
```
policies.push_back(CompletionPolicyHelpers::consumeWhenAllOrdered("muon-track-writer.*"));
```
or modify the regex accordingly .

shahor02 (changes requested):
@pillot could you adapt the writer to ordered policy, see below.
This will be needed for all writers!

Detectors/MUON/Workflow/src/tracks-matcher-workflow.cxx
pillot: Sorry, I missed this recent modification in the writer workflow.

I think it's better to preserve the name with "muon", to do not mix with the MCH track writer. Here it writes the matched MCH-MID tracks (ReconstructionDataFormats/TrackMCHMID.h).

Unfortunately I had to update alidist as well after updating O2, so it will take a while to recompile...

shahor02 (approved):

pillot:
@shahor02 all checks have passed. Can you merge it please?

PR #8519, opened 2022-04-07
MID afterburner for digits synchronization with collision BC

dstocco:
There is an offset between the collision BC and the BC that can be obtained from the electronics clock.
This offset was in principle accounted for, but a bug was found in the correction.
This PR fixes the bug and also introduces an afterburner to change the offset on-the-fly if needed.

The PR consists of 3 commits:
- The first one removes obsolete decoders, that would not compile anymore after this fix. Since they are not used, it is easier to remove them rather than correct them.
- The second improves the way the digits are synchronized in the decoding phase, and fixes a bug.
- The third one introduces an afterburner, which allows to change the offset on-the-fly if needed, instead of re-generating the CTFs.


aphecetche (approved):

dstocco:
Errors for fullCI are once again unrelated...

dstocco:
When merging, please do not squash...

dstocco:
@aphecetche : all checks are passed but the merging is blocked. Did you allow automatic merge in your review?

dstocco:
Plese do not squash...

PR #8523, opened 2022-04-07
TPC: validity timestamp for time slot residual gain calibration

badarots:
Also, better default values for fit selection

### DIFF
--- a/Detectors/TPC/workflow/src/ProcessingHelpers.cxx
+++ b/Detectors/TPC/workflow/src/ProcessingHelpers.cxx
@@ -9,6 +9,7 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
+#include <cstdint>
 #include <string>
 
 #include <fairmq/Device.h>
@@ -20,6 +21,7 @@
 #include "Framework/InputRecord.h"
 #include "Framework/ServiceRegistry.h"
 #include "CommonConstants/LHCConstants.h"
+#include "DetectorsRaw/HBFUtils.h"
 
 #include "TPCWorkflow/ProcessingHelpers.h"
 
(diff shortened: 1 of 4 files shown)


Detectors/TPC/workflow/src/ProcessingHelpers.cxx
```diff
@@ -76,3 +78,15 @@ Long64_t processing_helpers::getOrbitReset(o2::framework::ProcessingContext& pc)
   const auto orbitReset = tv->front();
   return orbitReset;
 }
+
+uint64_t processing_helpers::getAbsoluteTimeFrame(o2::framework::ProcessingContext& pc)
+{
+  static const double TFlength = 1e-3 * o2::raw::HBFUtils::Instance().getNOrbitsPerTF() * o2::constants::lhc::LHCOrbitMUS; // in ms
+  return static_cast<uint64_t>(processing_helpers::getCreationTime(pc) / TFlength);
```
wiechula: would either
```suggestion
  static const auto TFlength = static_cast<uint64_t>(1e-3 * o2::raw::HBFUtils::Instance().getNOrbitsPerTF() * o2::constants::lhc::LHCOrbitMUS); // in ms
  return processing_helpers::getCreationTime(pc) / TFlength;
```
or
```suggestion
  static const double TFlength = 1e-3 * o2::raw::HBFUtils::Instance().getNOrbitsPerTF() * o2::constants::lhc::LHCOrbitMUS; // in ms
  return static_cast<uint64_t>(double(processing_helpers::getCreationTime(pc)) / TFlength);
```
make sense?

wiechula (commented):
Just a minor suggestion.

Detectors/TPC/workflow/src/ProcessingHelpers.cxx
badarots: It makes, I would go for option 2. Option 1 will truncate the TFlengh from 11.4 ms to 11 ms.

badarots: Does this way of setting the validity time make sense for you? The drawback is that the validity time will not be exactly equal to the creation time, it will be the previous timestamp (in ms) multiple of the time frame interval... this is the same thing [LHCClockCalibrator](https://github.com/AliceO2Group/AliceO2/blob/bdf24162f716f1bb486c55683e9d543f631d9ab0/Detectors/TOF/calibration/testWorkflow/LHCClockCalibratorSpec.h#L88) does.

The other option I could think of was to pass the time stamp as the TF identifier, which would set the validity to the exact creation timestamp, but this would change the semantics of CLI arguments like `--tfs-per-slot` to milliseconds per slot.

PR #8526, opened 2022-04-08
ZDC - Tower intercalibration workflow

cortesep:


cortesep:
Test on Detectors/ZDC/macro/CreateInterCalibConfig.C is failing. However the macro executes with no problems on local computer. Cannot reproduce the failure.

shahor02:
@cortesep The CI macro tests do not execute them, just load. Particularly, in this case loading in the interpreted mode fails.
Are you able to do locally  (from root session) both
`.L CreateInterCalibConfig.C`
and 
`.L CreateInterCalibConfig.C+`
?
From the https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/AliceO2/8526/bebe7dcda73abadc8d4843b8f94fc7430fed2e18/build_AliceO2_O2_o2_macOS/fullLog.txt it looks like you are exposing to interpreter some boost header, which should not be done.

cortesep:
> 

The weird behaviour was because no header was loaded in interpreted mode. The problem is solved by always including this:
#include "ZDCBase/Constants.h"
#include "ZDCCalib/InterCalibConfig.h"

shahor02 (approved):

PR #8528, opened 2022-04-08
[EMCAL-645] Propagate trigger bits to EventData

mfasDa:
- Add field for trigger bits to EventData
- Add propagation of trigger bits in
  EventHandler
- Add handling in case trigger bits in
  Cell and Cluster TriggerRecord would
  differ


mfasDa:
Error unrelated. @shahor02 Do you think this can be merged?

shahor02:
@mfasDa You have forcepushed something and now it is testing again. I would wait for the end of tests.

shahor02 (approved):

PR #8533, opened 2022-04-09
Changes in the time-slot based calibration

shahor02:
Some calibrations need access from the time-slot container to detailed TF info for every processed TF. 
They can do this by defining `fill(o2::dataformats::TFIDInfo& ti, const <data>)` method instead of current `fill(const <data>)`.
This method, if available, will be automatically called with [TFIDInfo](https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/common/include/CommonDataFormat/TFIDInfo.h) filled from the TF DataHeader and DataProcessingHeader in the run method of the calibrator device. Otherwise, the old `fill` method will be used.

Also, the `<calibrator>.process(TFType, <data>)`  was changed to `<calibratort>.process(<data>)`: the process methods defined in the TimeSlotCalibration class will access TFIDInfo internally to decide which slot must be processed.


Detectors/TOF/calibration/testWorkflow/LHCClockCalibratorSpec.h
```diff
@@ -71,28 +71,23 @@ class LHCClockCalibDevice : public o2::framework::Task
...
     if (!mcalibTOFapi) {
       mcalibTOFapi = new o2::tof::CalibTOFapi(long(0), &mPhase, &mTimeSlewing); // TODO: should we replace long(0) with tfcounter defined at the beginning of the method? we need the timestamp of the TF
       mCalibrator->setCalibTOFapi(mcalibTOFapi);
     }
-
     if (data.size() == 0) {
       return;
     }
-
-    tfcounter = uint64_t(data[0].getTimestamp() * TFlengthInv);
```
shahor02, line 88: @chiarazampolli @noferini : what is the meaning of this line? The timestamp you extract is integer in seconds, how you can convert it to "tfcounter" and why?
Also, we should not use o2::raw::HBFUtils in the production workflows, it is not guaranteed to have correct values there.

Detectors/TOF/calibration/testWorkflow/TOFDiagnosticCalibratorSpec.h
```diff
@@ -47,16 +47,10 @@ class TOFDiagnosticCalibDevice : public o2::framework::Task
...
   void run(o2::framework::ProcessingContext& pc) final
   {
-
-    static const double TFlengthInv = 1E6 / o2::raw::HBFUtils::Instance().getNOrbitsPerTF() / o2::constants::lhc::LHCOrbitMUS;
-
-    auto tfcounter = o2::header::get<o2::framework::DataProcessingHeader*>(pc.inputs().get("input").header)->startTime;
-
+    o2::base::TFIDInfoHelper::fillTFIDInfo(pc, mCalibrator->getCurrentTFInfo());
     auto const data = pc.inputs().get<o2::tof::Diagnostic*>("input");
-    tfcounter = uint64_t(data->getTimeStamp() * TFlengthInv);
```
shahor02, line 58: @chiarazampolli @noferini same question here

Detectors/TPC/workflow/include/TPCWorkflow/LaserTracksCalibratorSpec.h
```diff
@@ -49,14 +49,12 @@ class LaserTracksCalibratorDevice : public o2::framework::Task
...
     const auto dph = o2::header::get<o2::framework::DataProcessingHeader*>(pc.inputs().get("laserTracks").header);
-    const auto startTime = dph->startTime;
-    const auto endTime = dph->startTime + dph->duration - 1;
 
+    o2::base::TFIDInfoHelper::fillTFIDInfo(pc, mCalibrator->getCurrentTFInfo());
     auto data = pc.inputs().get<gsl::span<TrackTPC>>("laserTracks");
-    LOGP(info, "Processing TF with start time {} and {} tracks", startTime, data.size());
+    LOGP(info, "Processing TF {} and {} tracks", mCalibrator->getCurrentTFInfo().tfCounter, data.size());
 
-    mCalibrator->getSlotForTF(startTime).getContainer()->setTFtimes(startTime, endTime);
```
shahor02, line 58: @wiechula why do you need to set these startTime and endTime (which are not really times but counters), I don't see where they are used. In any case, does not your container cover multiple TFs? Here the span of 1 event only is set.

PR #8535, opened 2022-04-09
Write only TRU digits to AOD trig table...

peressounko:
- Write only TRU digits to AOD trig table; 
- Wrong DDL fix in rawConverter; 
- Remove Decompose fail message in clusterizer; 
- fix validity range

### DIFF
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -442,6 +442,16 @@ class CellHelper
   {
     return cell.getAbsId();
   }
+  // If this cell - trigger one?
+  static bool isTRU(const o2::emcal::Cell& cell)
+  {
+    return cell.getTRU();
+  }
+
+  static bool isTRU(const o2::phos::Cell& cell)
+  {
+    return cell.getTRU();
+  }
 
   static int16_t getFastOrAbsID(const o2::emcal::Cell& cell)
   {
@@ -450,7 +460,7 @@ class CellHelper
 
   static int16_t getFastOrAbsID(const o2::phos::Cell& cell)
   {
-    return cell.getAbsId();
+    return cell.getTRUId();
   }
 
   static float getAmplitude(const o2::emcal::Cell& cell)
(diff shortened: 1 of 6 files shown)


peressounko:
Dear @shahor02, core team, could you please have a look at this PR? Thanks, Dmitri

Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
```diff
@@ -442,6 +442,16 @@ class CellHelper
...
+  // If this cell - trigger one?
+  static bool isTRU(const o2::emcal::Cell& cell)
+  {
+    return true; // To be checked with Markus
+  }
+
+  static bool isTRU(const o2::phos::Cell& cell)
+  {
+    return cell.getTRU();
+  }
```
shahor02: should not these methods be provided by the Cell classes? 
Do we need to wait for the reply of @mfasel ?

peressounko: Ideally both PHOS and EMCAL should provide consistent naming to avoid this Helper class. But this is current status. Comment from @mfasDa on filling this AOD table would be very welcome! Note that commit does not change anything from EMCAL side.

mfasDa: For EMCAL we currently plan only propagate STU L1 TimeSums to the AOD, not TRU digits, due to noise reason. TRU digits will only be used for QC where we will probably use the cell class as well for, the resolution should be sufficient.

In the cell class we defined (DataFormats/EMCAL/include/DataFormatsEMCAL/Cell.h, L155)
``` c++
  Bool_t getTRU() const { return isChannelType(ChannelType_t::TRU); }
 ``` 

Ideally we would combine the function names even though we have different implementations, for the 2 detectors, then a lot of the code can be combine with templates.


Not yet decided whether we will reuse the same class also for the L1 timesum from the STU.

shahor02: @mfasDa thanks for reply, just to be sure: it is ok to merge this PR with `isTRU(const o2::emcal::Cell& cell)` returning always `true`?

mfasDa: @shahor02 From how I understand the code no, as it would flag also FEC cells as TRU cells, it must return getTRU() from o2::emcal::Cell in the same way as it does for PHOS.

shahor02: @peressounko can you adapt the `isTRU(const o2::emcal::Cell& cell)` ?

peressounko: Dear @mfasDa, @shahor02, I added TRU selection for EMCAL too, please have a look. Please note, that in current O2 neither PHOS nor EMCAL do not have such selection, so I left it for EMCAL as is.

peressounko:
@shahor02, checks passed. Probably we can merge

shahor02 (approved):

PR #8536, opened 2022-04-09
ITS Digital Scan: added features for full matrix counts 

iravasen:
Added (mainly):
- full matrix per chip into memory (only for digital and analogue scan) to consider extra hits also in ROOT trees and to better spot bad dcols


iravasen:
Hi @shahor02, can I please ask you to merge also this one by today after successful tests? I added some features for the dig/ana scans.

shahor02 (approved):

PR #8538, opened 2022-04-10
[MFT] Config local or CCDB cluster dictionary

rpezzi:
Disables CCDB fetcher if local cluster dictionary is provided.


shahor02:
@rpezzi why do you need this? Note that you can tell the fetcher to read particular objects from the local files by adding at the end of the workflow e.g.
```
--condition-remap “file://myccdb=MFT/Calib/ClusterDictionary"
```
The fetcher will look for the file `myccdb/MFT/Calib/ClusterDictionary/snapshot.root` with the object stored under the name "ccdb_object".
One can also remap multiple conditions as `--condition-remap “file://dir1=X/Y/Z;A/B/C;file://dir2=Q/W/E;R/T/Y;" etc.

rpezzi:
Hi Ruben, I was not aware of option `--condition-remap`. Thanks.

PR #8348, opened 2022-03-15
Data model for analysis track propagation to PV

jgrosseo:
Illustrating a solution how we could make the track propagation from the reconstruction output (innermost update) to the collision vertex.
This requires to allow extended tables to be written directly from an analysis task (bypassing the spawner) and avoiding a circular dependency.

O2Physics PR here: https://github.com/AliceO2Group/O2Physics/pull/614

image

PR #8542, opened 2022-04-11
DPL Analysis: prevent spawner from creating DYN tables that are already provided

aalkin:
Supporting #8348


jgrosseo:
Great. Thanks!

PR #8543, opened 2022-04-11
[FIT] Separate Trigger class and make it common for FT0, FV0 and FDD.…

mslupeck:
… Fix Trigger class initial values.

### DIFF
--- a/DataFormats/Detectors/FIT/CMakeLists.txt
+++ b/DataFormats/Detectors/FIT/CMakeLists.txt
@@ -9,7 +9,8 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
+add_subdirectory(common)
 add_subdirectory(FDD)
 add_subdirectory(FT0)
 add_subdirectory(FV0)
-add_subdirectory(common)
+
--- a/DataFormats/Detectors/FIT/FDD/CMakeLists.txt
+++ b/DataFormats/Detectors/FIT/FDD/CMakeLists.txt
@@ -10,19 +10,20 @@
 # or submit itself to any jurisdiction.
 
 o2_add_library(DataFormatsFDD
-               SOURCES  src/RawEventData.cxx
-                        src/CTF.cxx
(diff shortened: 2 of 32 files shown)


AllaMaevskaya (commented):
I like your changes. The only thing I don't agree is  
 static const int16_t DEFAULT_TIME = -127; // for average/8 of one side (A or C)
1. average time should not be divided by 8, only sum amplitude does;
2. -127 is too sharp threshold, if we want vertex cut +-50cm we will be out of this. Time channels are from -2048 to +2048, so let set number out of this

mslupeck:
> I like your changes. The only thing I don't agree is static const int16_t DEFAULT_TIME = -127; // for average/8 of one side (A or C)
> 
> 1. average time should not be divided by 8, only sum amplitude does;

Good catch - I modified the comment. Thanks!

> 2. -127 is too sharp threshold, if we want vertex cut +-50cm we will be out of this. Time channels are from -2048 to +2048, so let set number out of this

Actually, I reverted it back to the previous value (-5000)

AllaMaevskaya (dismissed):

AllaMaevskaya (dismissed):

mslupeck:
Hi @shahor02 
Could you merge it?

shahor02 (approved):
@mslupeck before I merge this: given the changes in the format and CTFCoder, did you check that the old, e.g. pilot beam data are readable with this PR?

mslupeck:
@shahor02 - thanks for the hint. Now I've tested it using:
`o2-ctf-reader-workflow --ctf-input o2_ctf_run00505582_orbit0316488698_tf0000167535.root --onlyDet FV0 | o2-fv0-digits-writer-workflow --disable-mc -b`
`o2-ctf-reader-workflow --ctf-input o2_ctf_run00505582_orbit0316488698_tf0000167535.root --onlyDet FT0 | o2-ft0-digits-writer-workflow --disable-mc -b`
They were executed without errors and the output digit files are the same as in dev branch (except the sum of amplitude default value, which was modified from -5000 to 0 in this PR). I think it can be merged.

PR #8544, opened 2022-04-11
MCH: use 3-levels CCDB path for bad channel calib object

aphecetche:


Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
```diff
@@ -89,7 +89,7 @@ ccdb::CcdbObjectInfo createCcdbInfo(const T& object, uint64_t timeStamp, std::st
   auto flName = o2::ccdb::CcdbApi::generateFileName(clName);
   std::map<std::string, std::string> md;
   md["upload-reason"] = reason;
-  return o2::ccdb::CcdbObjectInfo("MCH/BadChannelCalib", clName, flName, md, timeStamp, 99999999999999);
+  return o2::ccdb::CcdbObjectInfo("MCH/Calib/BadChannel", clName, flName, md, timeStamp, 99999999999999);
```
shahor02: Could you reduce the infinite end validity to something more reasonable (e.g. timestamp + some margin)?

aphecetche: is there a definite drawback in using infinity ? 

Because the calibration is indeed valid from the point where the pedestal run was taken up to the next pedestal run, that is an unknown date.  In normal running conditions pedestal calibration should be taken "frequently" (on the order of once per fill or per day I would say), but unless we impose a strict policy on when to perform those, that's not a guarantee.

@aferrero2707 what's your take on this ?

aferrero2707:
@aphecetche  I absolutely agree. The frequency of the pedestals calibration is not precisely known, and moreover as far as we can see the noise pattern is quite stable, hence I think that giving an infinite validity to the last available calibration is the good approach, and I personally don't see potential drawbacks. But I might be overlooking something...

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
shahor02: I let @costing to comment: as far as I understood him, overlaps cost an overhead.
With infinite end-validity every new object will overlap will all previous ones.
Are you going to run with 1 year (or 1 month old object) if you expect it to change so frequently that it requires daily calibration? Cannot you have a reasonable default with inf. validity and upload specific calibrations with more reasonable end-validity margin?

costing: Indeed, the first cut, by the timestamp when the object should be valid, is the indexed column / search criteria. If all objects have the same inf. validity, you invalidate this index and force a sequential scan of the entire namespace, that is much slower. Ruben is right, you should create the objects with as small an interval of validity as possible so that the overlap is minimal between them.

aphecetche: well, as stated above, the end of validity is intrinsically unknown for this calibration. 
But maybe what we could do is use something like 2 days **AND** mandate the shift crew to perform the pedestal calibration at least once a day ? @aferrero2707 what do you think ?

aferrero2707: I am not 100% sure that we will perform calibrations every day. Maybe ~5 days would be more reasonable (or whatever the expected duration of one fill is)?

In case of validity gaps between the available calibrations, would it still be possible to use the most appropriate one (last calibration preceding the data being processed), and issue a warning?

shahor02: > In case of validity gaps between the available calibrations, would it still be possible to use the most appropriate one (last calibration preceding the data being processed), and issue a warning?

No, only objects valid for the query timestamp can be fetched. So, you can have a single "infinite" validity default object (if it is tagged as default, the ccdb fetcher will notify about that in the logs) and short validity, e.g. a few days, week (the fill, from the orbit rewinding to 0 point of view, can last ~4 days) specific objects.

aphecetche:
@shahor02 @aferrero2707 just forced-pushed this one where I changed the validity to be 5 days.

aphecetche:

shahor02 (approved):

PR #8547, opened 2022-04-11
Revert "DPL: prevent OutputProxy from using 100% CPU usage when idle"

ktf:
Reverts AliceO2Group/AliceO2#7313


davidrohr:
Could you comment what was the problem with this PR?

ktf:
According to @shahor02 it breaks the EoS with QC. I suspect this prevents the EoS to be dispatched. We agreed in the meeting to revert this to see if that's really the error.

PR #8550, opened 2022-04-11
DPL: Add support for propagating oldest possible timeslice

ktf:


ktf:
@davidrohr this works with my synthetic example which I added (stage/bin/o2-testworkflows-diamond-workflow-leaky) I still have some crash at the end, probably because the special message carrying the information is not pruned correctly.

davidrohr:
ok, thx. I'll wait until you manage to fix that crash before I test it in the FST :)

ktf:
yes, sure. I was also thinking that this is basically the fix for the QC rate limiting. We just use oldestPossibleTimeslice as number of processed timeframe.

ktf:
Ok, all the crashes understood and fixed in my local build. There was a bad assert which I did not see because of the optimised build.

ktf:
@davidrohr I have improved this to provide better feedback when the messages get dropped or not and why. This should allow us to immediately see if some wrong decision was taken by the logic. There is also an error message in case timeframe data is dropped unexpectedly.

ktf:
This should finally handle the forwarding and the fact that by default the oldest possible input should be 0. In retrospect also timers should work the same way (and they should with this), because we do not want to go back and forward with the oldest possible input. It required quite extensive reshuffling of how the inputs are produced, kinda like I already did for the outputs, so that we can correctly account for  "fake" channels like those of timers. The advantage is that we can probably simplify (in a subsequent PR) how data sending works.

davidrohr:
This fails compilation:
```
2022-04-15@14:38:19:DEBUG:O2:O2:0: /home/qon/alice/sw/SOURCES/O2/dev/0/Framework/Core/src/CommonDataProcessors.cxx: In lambda function:
2022-04-15@14:38:19:DEBUG:O2:O2:0: /home/qon/alice/sw/SOURCES/O2/dev/0/Framework/Core/src/CommonDataProcessors.cxx:517:23: error: 'timesliceIndex' was not declared in this scope; did you mean 'TimesliceIndex'?
2022-04-15@14:38:19:DEBUG:O2:O2:0:   517 |           *consumed = timesliceIndex.getOldestPossibleOutput().timeslice.value;
2022-04-15@14:38:19:DEBUG:O2:O2:0:       |                       ^~~~~~~~~~~~~~
2022-04-15@14:38:19:DEBUG:O2:O2:0:       |                       TimesliceIndex
```
Could you check?

ktf:
So there is two problems:
1) The first one  is fixed by the extra validChannels++ and its due to the fact that I should not skip channels in the indexing, even if I do not expect data from them.
2) The second one looks like the extra message created by the dispatcher is not correctly parsed on the receiving side.

ktf:
@davidrohr Ok, all known problems understood and hopefully fixed. With this your reproducer (using raw-reader.sh to read the timeframe). 2) was merely due to the too strict sampling policy which was not sending the oldestPossibleTimeslice messages often enough. Now we propagate that information regardless of the sampling frequency.

@Barthelemy @knopers8 could you also have a look at this?

ktf:
For the record, I have tested this on hyperloop as well and it does not seem to have side effects.

PR #8551, opened 2022-04-11
Subtract BC from collisioncontext in TOF calib raw time

noferini:
tested with 1 CTF of pilot beam (66 TFs)
All fine

### DIFF
--- a/Detectors/GlobalTrackingWorkflow/src/tof-matcher-workflow.cxx
+++ b/Detectors/GlobalTrackingWorkflow/src/tof-matcher-workflow.cxx
@@ -26,6 +26,9 @@
 #include "DetectorsRaw/HBFUtilsInitializer.h"
 #include "Framework/CallbacksPolicy.h"
 #include "GlobalTrackingWorkflowHelpers/InputHelper.h"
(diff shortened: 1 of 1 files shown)


noferini:
Can it be merged?

Detectors/GlobalTrackingWorkflow/src/tof-matcher-workflow.cxx
```diff
@@ -123,6 +126,23 @@ WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
...
+      auto bcf = context->getBunchFilling();
+      std::bitset<3564> isInBC = bcf.getBCPattern();
+      for (unsigned int i = 0; i < isInBC.size(); i++) {
+        if (isInBC.test(i)) {
+          o2::tof::Utils::addInteractionBC(i, true);
+        }
+      }
+    }
+  }
+
```
shahor02, line 145: Note that in general the bunch filling will not be available in the workflow definition but should be extracted from the GRP at the 1st TF processing. At the moment merging as is.

shahor02 (approved):

PR #8552, opened 2022-04-11
Add filename to mermaid argument

Barthelemy:
Take a filename as argument, use "-" to force output to go to stdout.


Framework/Core/src/runDataProcessing.cxx
```diff
@@ -2388,36 +2393,36 @@ int doMain(int argc, char** argv, o2::framework::WorkflowSpec const& workflow,
...
+    ("min-failure-level", bpo::value<LogParsingHelpers::LogLevel>(&minFailureLevel)->default_value(LogParsingHelpers::LogLevel::Fatal),                                //                                                                                                                          //
+     "minimum message level which will be considered as fatal and exit with 1")                                                                                        //                                                                                                                            //
+    ("graphviz,g", bpo::value<bool>()->zero_tokens()->default_value(false), "produce graphviz output")                                                                 //                                                                                                                              //
+    ("mermaid", bpo::value<std::string>()->default_value(""), "produce graph output in mermaid format in file under specified name or on stdout if argument is \"-\"") //                                                                                                                              //
```
ktf, line 2415:
```suggestion
    ("mermaid", bpo::value<std::string>()->default_value("-"), "produce graph output in mermaid format in file under specified name or on stdout if argument is \"-\"") //                                                                                                                              //
```

Barthelemy: Wouldn't that enable the `mermaid` mode when `--mermaid` is not specified ?

ktf: ah, good point. Ok, then...

PR #8555, opened 2022-04-11
Adopt namespaced FairMQ API

dennisklein:
and replace deprecated usage of `FairRun::SetOutputFile()`.


davidrohr:
Could you please apply the clang-format changes? This seems mostly indentation of comments after your string replace.
In general, I would only merge this after the new FairMQ is in production at P2, which will (considering Eastern...) still take at least 2 weeks. We can see whether there are compile errors, but afterwords I would mark this as Draft, to avoid having it dangling around.

dennisklein:
> Could you please apply the clang-format changes?

Will do.

> We can see whether there are compile errors, but afterwords I would mark this as Draft, to avoid having it dangling around.

:+1: - no urgency from my side anyways.

ktf:
Seems to have conflicts (probably due to recent github woes, actually). Could you please rebase?

ktf (changes requested):
Please do not update the Copyright statement unless you want to fix the script checking it as well.

davidrohr:
should we perhaps rebase this now and then merge it? The new FMQ is in alidist since a while

dennisklein:
> should we perhaps rebase this now and then merge it?

It is on my todo list for this week.

> The new FMQ is in alidist since a while

It is not needed for FairMQ v1.4 though. Only in v1.5 (unreleased) the old names/headers will be deprecated (but still work).

dennisklein:
> Please do not update the Copyright statement unless you want to fix the script checking it as well.

Removed those changes.

> Seems to have conflicts (probably due to recent github woes, actually). Could you please rebase?

Done. Tested it on centos 8 stream with `aliBuild --defaults o2 build O2` against future FairMQ with deprecations enabled (compilation only).

PR #8560, opened 2022-04-12
DPL: provide per channel metric when backpressure happens

ktf:


ktf:
@Barthelemy as you requested. This will publish a `backpressure_<channel-name>` metric whenever a given channel has back pressure detected / recovered.

Barthelemy:
thank you !

ktf:
@awegrzyn @Barthelemy @davidrohr this will post one metric per device when there is back pressure, and one metric per device when the back pressure is recovered. These will reach the production influxDB, because I am not marking them as DPL metrics (which we drop, IIRC). Please go ahead and merge this if you agree this can be sustained by the infrastructure.

ktf:
@awegrzyn @Barthelemy @davidrohr this will post one message per device when there is back pressure, and one message per device when the back pressure is recovered. These will reach the production influxDB, because I am not marking them as DPL metrics (which we drop, IIRC). Please go ahead and merge this if you agree this can be sustained by the infrastructure.

Barthelemy (approved):

awegrzyn (approved):

ktf:
@davidrohr agreed out of band.

PR #8561, opened 2022-04-12
Prototyping a demonstrator workflow for extended DataDescriptor matching

matthiasrichter:
There is a request to support input specs matching all blocks of
a particular origin/description with all but a specific sub spec.
E.g. a processor with input spec {TST/SAMPLE/!0} matching all blocks
of {TST/SAMPLE} with a subspec other than '0', and the output
{TST/SAMPLE/0}.

A custom matcher can be created for this purpose using xor operation
to invert the result of the embedded SubSpecificationTypeValueMatcher.

### DIFF
--- a/Framework/Core/test/test_DataDescriptorMatcherWorkflow.cxx
+++ b/Framework/Core/test/test_DataDescriptorMatcherWorkflow.cxx
@@ -1,0 +1,148 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include "Framework/DataDescriptorMatcher.h"
+#include "Framework/DataDescriptorQueryBuilder.h"
+#include "Framework/InputSpec.h"
+#include "Framework/DataDescriptorMatcher.h"
+#include "Framework/InputRecordWalker.h"
+#include "Framework/DataProcessorSpec.h"
+#include "Framework/DataRefUtils.h"
+#include "Framework/ControlService.h"
+#include "Framework/InputRecord.h"
+#include "Framework/DataAllocator.h"
+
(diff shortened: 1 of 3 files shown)


Framework/Core/src/DataSpecUtils.cxx
```diff
@@ -330,10 +330,10 @@ MatcherInfo extractMatcherInfo(DataDescriptorMatcher const& top)
...
       // For now we do not support extracting a type from things
       // which have an OR, so we reset all the uniqueness attributes.
-      if (action.node->getOp() == ops::Or || action.node->getOp() == ops::Xor) {
-        state.hasError = true;
-        return VisitNone;
-      }
+      //if (action.node->getOp() == ops::Or || action.node->getOp() == ops::Xor) {
+      //  state.hasError = true;
+      //  return VisitNone;
+      //}
```
matthiasrichter: @ktf, I guess this was added to allow quick implementation without considering too many corner cases. Maybe we can solve this now. At some point we want to support this anyhow.

Framework/Core/test/test_DataDescriptorMatcherWorkflow.cxx
```diff
@@ -0,0 +1,150 @@
...
+    data_matcher::OriginValueMatcher{"TST"},
+    std::make_unique<DataDescriptorMatcher>(
+      DataDescriptorMatcher::Op::And,
+      data_matcher::DescriptionValueMatcher{"SAMPLE"},
+      std::make_unique<DataDescriptorMatcher>(
+        DataDescriptorMatcher::Op::And,
+        std::make_unique<DataDescriptorMatcher>(
+          DataDescriptorMatcher::Op::Xor,
+          data_matcher::SubSpecificationTypeValueMatcher{0},
+          data_matcher::ConstantValueMatcher{true}),
```
matthiasrichter: Since it's two bool variables which will be logically xor'ed, this corresponds to a negation.

It's a demonstrator and it's working. I will implement the `DataDescriptorMatcher::Op::Not` alongside `Op::Just`, this will probably  not require the other change above, but the 'Not' operator might have similar implications on the uniqueness of the matcher parameters.

Framework/Core/src/DataSpecUtils.cxx
ktf: yes, exactly. The problem is extracting a ConcreteDataType from anything which is not a query. We basically need to revisit all the places where DataSpecUtils::asConcreteDataType is used and see if we cannot change it to something that matches the whole query. In particular I think the place where we serialise the workflow needs to have a recursive parser / serialiser to serialise and read back the query.

PR #8564, opened 2022-04-12
DPL: do not allow dummy creation times to be used for CCDB fetching

ktf:


ktf:
@shahor02 @chiarazampolli with this we should be able to spot immediately if the creation time for some object was not correctly initialised.

davidrohr:
How will we test this? Deploy on the EPN as alternatve version, run and see how it fails at STOP?
But I assume this should not go into a build that is really running? Or do we accept this?
Alternatively, couldn't we just throw an error message, and keep the last time?

ktf:
Well, I guess we could start by having an error and that's it. I would prefer we do not try to correct the time at the CCDB fetcher level, though, because it's simply working around a bug somewhere else, even in the case the correct time to use is the current time, so it really needs to be fixed.

davidrohr:
I also don't want to correct it, but for now, if we say at the previous time, we would probably avoid the crashes we have at the moment, and the error message will hopefully help us to spot where it comes from.
Would there be any chance to print some data descriptor/subspec together with the error, to identify where it comes from?

ktf:
Ok, this should behave as you asked. Notice however that this is under the assumption that the bad firstTFOrbit comes from a non-initialised DataHeader. If that's not the case, I have no way to spot if the firstTFOrbit was bogus.

shahor02:
can you spot where the TimingInfo.tfCounter is assigned the same value as TimingInfo.timeSlice, i.e. DPH.trStart?

davidrohr:
looks reasonable for me. I'd merge only this today, since it should be safe in any case. We can deploy tomorrow as an alternative version, and then run some test partitions to investigate. Then I'd merge the other PRs afterwards.

davidrohr:
merging, to have it in the build tomorrow, and this is almost trivial and has no real chance to fail in the CI

PR #8565, opened 2022-04-12
Adding DataDescriptionMatcher operation 'Not'

matthiasrichter:
Need to check if using this negator operation can have implications
on the handling of concrete data types. No impact to existing code
however.

### DIFF
--- a/Framework/Core/src/DataDescriptorMatcher.cxx
+++ b/Framework/Core/src/DataDescriptorMatcher.cxx
@@ -243,6 +243,8 @@ bool DataDescriptorMatcher::match(char const* d, VariableContext& context) const
   //     return std::visit(eval, mLeft) ^ std::visit(eval, mRight);
   //   case Op::Just:
   //     return std::visit(eval, mLeft);
+  //   case Op::Not:
+  //     return !std::visit(eval, mLeft);
   // }
   //  When we drop support for macOS 10.13
   if (auto pval0 = std::get_if<OriginValueMatcher>(&mLeft)) {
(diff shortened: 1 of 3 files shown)


matthiasrichter:
This will disentangle #8561 from the mapping of DataMatcher to ConcreteDataMatchers as we do not need to use the xor operation.

On the other hand, introducing the 'Not'-operator might have a similar effect on the handling of concrete data descriptions.

Framework/Core/src/DataDescriptorMatcher.cxx
```diff
@@ -386,6 +393,11 @@ bool DataDescriptorMatcher::operator==(DataDescriptorMatcher const& other) const
     return true;
```
matthiasrichter: @ktf shouldn't we return `leftValue` here?

Framework/Core/src/DataDescriptorMatcher.cxx
```diff
@@ -386,6 +393,11 @@ bool DataDescriptorMatcher::operator==(DataDescriptorMatcher const& other) const
...
   }
 
+  if (mOp == Op::Not) {
+    // FIXME: don't know if that is the correct logic
+    return true;
```
matthiasrichter: think the correct logic is to return `leftValue`

Framework/Core/src/DataDescriptorMatcher.cxx
```diff
@@ -383,7 +390,11 @@ bool DataDescriptorMatcher::operator==(DataDescriptorMatcher const& other) const
   }
 
   if (mOp == Op::Just) {
-    return true;
+    return leftValue;
```
matthiasrichter, line 393: I think when having two matchers with just left condition, the two should should be tested for equality and thus we need to return `leftValue`.

Actually, `leftResult` and `rightresult` would be better naming, shall we apply this?

matthiasrichter:
moving on with this and merging, the CI error is unrelated

PR #8571, opened 2022-04-13
Correcting the indices to MCH and MFT tracks in FwdTracks table

mcoquet642:
Introduce a function "fillIndexTablesPerCollision" which fills two vectors (mIndexTableMFT, mIndexTableFwd) which map the trackIndices to the index inside the fwdtrack table. It respect the same order and is called before "fillTrackTablesPerCollision".
The two vectors are initialized respectively with sizes equal to "recoData.getMFTTracks().size()" and "recoData.getMCHTracks().size() * 3", taking an upperbound to the size of the FwdTrack table.

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -286,7 +286,7 @@ void AODProducerWorkflowDPL::addToMFTTracksTable(mftTracksCursorType& mftTracksC
   if (needBCSlice) {
     ambigMFTTracksCursor(0, mTableTrMFTID, bcSlice);
   }
-  mGIDToTableMFTID.emplace(trackID, mTableTrMFTID);
+//  mGIDToTableMFTID.emplace(trackID, mTableTrMFTID);
   mTableTrMFTID++;
 }
 
@@ -365,7 +365,6 @@ void AODProducerWorkflowDPL::fillIndexTablesPerCollision(const o2::dataformats::
           }
 
           mGIDToTableMFTID.emplace(trackIndex, mIndexMFTID);
-          //          mIndexTableMFT.emplace(trackIndex.getIndex(), mIndexMFTID);
           mIndexTableMFT[trackIndex.getIndex()] = mIndexMFTID;
           mIndexMFTID++;
 
@@ -376,7 +375,6 @@ void AODProducerWorkflowDPL::fillIndexTablesPerCollision(const o2::dataformats::
 
           mGIDToTableFwdID.emplace(trackIndex, mIndexFwdID);
           if (src == GIndex::Source::MCH) {
-            //            mIndexTableFwd.emplace(trackIndex.getIndex(), mIndexFwdID);
             mIndexTableFwd[trackIndex.getIndex()] = mIndexFwdID;
           }
           mIndexFwdID++;
@@ -613,7 +611,7 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
   if (needBCSlice) {
     ambigFwdTracksCursor(0, mTableTrFwdID, bcSlice);
   }
-  mGIDToTableFwdID.emplace(trackID, mTableTrFwdID);
+//  mGIDToTableFwdID.emplace(trackID, mTableTrFwdID);
   mTableTrFwdID++;
 }
 
@@ -1533,9 +1531,6 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
     collisionID++;
   }
 
-  mGIDToTableFwdID.clear();
-  mGIDToTableMFTID.clear();
-
   // filling unassigned tracks first
   // so that all unassigned tracks are stored in the beginning of the table together
   auto& trackRef = primVer2TRefs.back(); // references to unassigned tracks are at the end
@@ -1668,6 +1663,11 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   mTableCollID = 0;
   mV0ToTableID.clear();
   mTableV0ID = 0;
+
+  mIndexTableFwd.clear();
+  mIndexFwdID=0;
+  mIndexTableMFT.clear();
+  mIndexMFTID=0;
 
   originCursor(0, tfNumber);
 
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -352,7 +352,6 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
 
 void AODProducerWorkflowDPL::fillIndexTablesPerCollision(const o2::dataformats::VtxTrackRef& trackRef, const gsl::span<const GIndex>& GIndices)
 {
-  //  for (int src = GIndex::NSources; src--;) {
   for (int src : {GIndex::Source::MFTMCH, GIndex::Source::MCH, GIndex::Source::MFT}) {
     int start = trackRef.getFirstEntryOfSource(src);
     int end = start + trackRef.getEntriesOfSource(src);
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -352,7 +352,7 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
 
 void AODProducerWorkflowDPL::fillIndexTablesPerCollision(const o2::dataformats::VtxTrackRef& trackRef, const gsl::span<const GIndex>& GIndices)
 {
-//  for (int src = GIndex::NSources; src--;) {
+  //  for (int src = GIndex::NSources; src--;) {
   for (int src : {GIndex::Source::MFTMCH, GIndex::Source::MCH, GIndex::Source::MFT}) {
     int start = trackRef.getFirstEntryOfSource(src);
     int end = start + trackRef.getEntriesOfSource(src);
@@ -365,8 +365,8 @@ void AODProducerWorkflowDPL::fillIndexTablesPerCollision(const o2::dataformats::
           }
 
           mGIDToTableMFTID.emplace(trackIndex, mIndexMFTID);
-//          mIndexTableMFT.emplace(trackIndex.getIndex(), mIndexMFTID);
-    mIndexTableMFT[trackIndex.getIndex()] = mIndexMFTID;
+          //          mIndexTableMFT.emplace(trackIndex.getIndex(), mIndexMFTID);
+          mIndexTableMFT[trackIndex.getIndex()] = mIndexMFTID;
           mIndexMFTID++;
 
         } else if (src == GIndex::Source::MCH || src == GIndex::Source::MFTMCH) {
@@ -375,9 +375,9 @@ void AODProducerWorkflowDPL::fillIndexTablesPerCollision(const o2::dataformats::
           }
 
           mGIDToTableFwdID.emplace(trackIndex, mIndexFwdID);
-          if (src == GIndex::Source::MCH){
-//            mIndexTableFwd.emplace(trackIndex.getIndex(), mIndexFwdID);
-        mIndexTableFwd[trackIndex.getIndex()] = mIndexFwdID;
+          if (src == GIndex::Source::MCH) {
+            //            mIndexTableFwd.emplace(trackIndex.getIndex(), mIndexFwdID);
+            mIndexTableFwd[trackIndex.getIndex()] = mIndexFwdID;
           }
           mIndexFwdID++;
         }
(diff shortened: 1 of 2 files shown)


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -499,8 +535,8 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
     chi2matchmchmid = track.getMIDMatchingChi2();
     chi2matchmchmft = track.getMFTMCHMatchingChi2();
     matchscoremchmft = track.getMFTMCHMatchingScore();
-    matchmfttrackid = track.getMFTTrackID();
-    matchmchtrackid = track.getMCHTrackID();
+    matchmfttrackid = mIndexTableMFT[track.getMFTTrackID()];
```
shahor02, line 535: here you fill the remapped index of the MFT track contributing to the match, but you actually not registering it in the fillIndexTablesPerCollision.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -350,6 +350,42 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
...
+      auto& trackIndex = GIndices[ti];
+      if (GIndex::includesSource(src, mInputSources)) {
+        if (src == GIndex::Source::MFT) {
+          if (trackIndex.isAmbiguous() && mGIDToTableMFTID.find(trackIndex) != mGIDToTableMFTID.end()) {
+            continue;
+          }
+
+          mGIDToTableMFTID.emplace(trackIndex, mIndexMFTID);
+          //          mIndexTableMFT.emplace(trackIndex.getIndex(), mIndexMFTID);
+          mIndexTableMFT[trackIndex.getIndex()] = mIndexMFTID;
```
shahor02: don't you need to fill mGIDToTableMFTID and mIndexTableMFT also for the MFT contributors of global forward tracks? See comment on L 538.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -499,8 +535,8 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
     chi2matchmchmid = track.getMIDMatchingChi2();
     chi2matchmchmft = track.getMFTMCHMatchingChi2();
     matchscoremchmft = track.getMFTMCHMatchingScore();
-    matchmfttrackid = track.getMFTTrackID();
-    matchmchtrackid = track.getMCHTrackID();
+    matchmfttrackid = mIndexTableMFT[track.getMFTTrackID()];
+    matchmchtrackid = mIndexTableFwd[track.getMCHTrackID()];
```
shahor02, line 536: also here: mIndexTableFwd is filled in the fillIndexTablesPerCollision only for standalone MCH tracks, while here you are using it as remapping for matched one.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -350,6 +350,42 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
...
+          if (trackIndex.isAmbiguous() && mGIDToTableFwdID.find(trackIndex) != mGIDToTableFwdID.end()) {
+            continue;
+          }
+
+          mGIDToTableFwdID.emplace(trackIndex, mIndexFwdID);
+          if (src == GIndex::Source::MCH) {
+            //            mIndexTableFwd.emplace(trackIndex.getIndex(), mIndexFwdID);
+            mIndexTableFwd[trackIndex.getIndex()] = mIndexFwdID;
+          }
+          mIndexFwdID++;
```
shahor02: why do you update mIndexTableFwd for MCH standalone only but always increment mIndexFwdID?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1484,6 +1520,22 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+  fillIndexTablesPerCollision(trackReffwd, primVerGIs);
+  collisionID = 0;
+  for (auto& vertex : primVertices) {
+    auto& trackReffwd = primVer2TRefs[collisionID];
+    fillIndexTablesPerCollision(trackReffwd, primVerGIs); // this function must follow the same track order as 'fillTrackTablesPerCollision' to fill the map of track indices
+    collisionID++;
+  }
+
+  mGIDToTableFwdID.clear();
+  mGIDToTableMFTID.clear();
```
shahor02: Why are you resetting these maps here? They are used in the fillTrackTablesPerCollision which is called below, right? Should not it be e.g. here: https://github.com/AliceO2Group/AliceO2/blob/d7e376507aa3a4cc5efd7874aca27b48cf452851/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L1671
Also, I think you should set to 0 the mIndexFwdID and mIndexMFTID,

shahor02 (commented):
Hi @mcoquet642 @rpezzi please check the comments below, I have some doubts in the logic of the changes. 
Also, rechecking the treatment of forward tracks in https://github.com/AliceO2Group/AliceO2/blob/ca93a6955100ff8151c1c5e2ad705be8cd2dc3e2/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L61 (used for track vertex association) I've got extra questions:

https://github.com/AliceO2Group/AliceO2/blob/ca93a6955100ff8151c1c5e2ad705be8cd2dc3e2/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L283-L285 flags consumed `MFTMCH` tracks (though there is no much point to do this: they are not constituents of a more global track, so there is no ambiguity to resolve) but you are not flagging contributing `MFT` and `MCH` tracks as consumed.

Then, when handling MFT: https://github.com/AliceO2Group/AliceO2/blob/ca93a6955100ff8151c1c5e2ad705be8cd2dc3e2/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L333-L335 and MCH https://github.com/AliceO2Group/AliceO2/blob/ca93a6955100ff8151c1c5e2ad705be8cd2dc3e2/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L358-L361 tracks, you check if they are already used (and the check will never yield `true` since the usage of standalone tracks was not filled above)
So, effectively, you don't have redundancy reduction, particularly, vertex->tracks association will account for both global tracks and their contributors. Note that for the barrel tracks the contributors are not accounted.
In case this is the wanted behaviour  for you, it is still better to rectify the code: do a proper accounting of MFTMCH contributors *if* the `creator`  at https://github.com/AliceO2Group/AliceO2/blob/ca93a6955100ff8151c1c5e2ad705be8cd2dc3e2/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L283 returns `true` but in the creator implementation of e.g. https://github.com/AliceO2Group/AliceO2/blob/ca93a6955100ff8151c1c5e2ad705be8cd2dc3e2/Detectors/Vertexing/src/VertexTrackMatcher.cxx#L198 return false if the GIndex is `MFT-MCH`

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mcoquet642: Corrected, thanks. Also removed the filling of mGIDToTableFwdID and mGIDToTableMFTID in addToFwdTracksTable and addToMFTTracksTable, since they are already filled in fillIndexTablesPerCollision.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mcoquet642: I think here we only need the map to retrieve the correct indices for MCH standalone. But I increment mIndexFwdID so that we have the correct index in the fwdtrack table, i.e. we count all fwdtracks not only MCH standalone.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mcoquet642: Here I use the map mIndexTableFwd to retrieve the correct indices for MCH standalone, since 'matchmchtrackid' is supposed to be pointing to a MCH track in the table.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mcoquet642: I am not sure I understand the comment. Here we fill the maps for all MFT tracks, including contributors to global forward tracks

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mcoquet642: Same, not sure I understand the comment. All MFT tracks are registered in fillIndexTablesPerCollision, including the ones contributing to the match. Maybe I am missing something.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: I mean the only place you are filling the `mIndexTableFwd` is https://github.com/AliceO2Group/AliceO2/blob/b2699abec433f6ed70585650719058455cd61e1f/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L377-L379, i.e. only for standalone MCH tracks but not for the MCH contributor of the matched track. Because of the way the CreateTracksVariadic treats forward tracks, this may work now (since the MCH tracks which are matched are present also in the standalone MCH tracks pool, so you effectively account them in the mIndexTableFwd) but if you switch to non-redundant version, the mIndexTableFwd for matched MCH track will not be filled.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: In the fillIndexTablesPerCollision you fill the mGIDToTableMFTID only for standalone MFT tracks: https://github.com/AliceO2Group/AliceO2/blob/b2699abec433f6ed70585650719058455cd61e1f/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L362, see the comment below.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mcoquet642: I see, I am not aware of this. I can only say that right now it effectively works (when I test this implementation the indices are correct).

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mcoquet642: From what I tested this effectively seems to fill all MFT tracks. The structure inside this `fillIndexTablesPerCollision` is the same as in the `fillTrackTablesPerCollision`function for MFT tracks.

rpezzi:
Hi @shahor02 , thanks for checking this PR. I think most of your comments considers that MCH and/or MFT tracks consumed by global forward tracks will not be stored. However, it has been agreed that MCH  and MFT tracks will be stored independently of their matching. The current index counting provided by Maurice will remain valid.
This PR should be ready to be merged.

shahor02:
Hi @rpezzi 

My comment https://github.com/AliceO2Group/AliceO2/pull/8571#pullrequestreview-941569882 meant that the current logic and usage of the RecoContainerCreateTracksVariadic by the MFT and MOUN is wrong: the method is supposed to account for the redundancy (i.e. to not provide simpler tracks if they are contributing to tagged global tracks). 
As I wrote, it does not mean that you cannot have redundancy in your selection: you just have to return false from your `creator` function, in this case the `createTracksVariadic` should not flag the contributors of the global track as used and will be provided to the `creator` method at following iterations. 
So, both the MFT/MUON part of the `createTracksVariadic` and its usage in the VertexTrackMatcher.cxx should be corrected.
Do you prefer to do this in a separate PR?

rpezzi:
> So, both the MFT/MUON part of the `createTracksVariadic` and its usage in the VertexTrackMatcher.cxx should be corrected.
> Do you prefer to do this in a separate PR?

Hi @shahor02. Thanks for identifying this inconsistency. This will be fixed in a separate PR. This is my understanding of this issue:
* No redundancy reduction in expected for forward tracks, then there is no need to `flagUsed2` any forward track in  https://github.com/AliceO2Group/AliceO2/blob/ca93a6955100ff8151c1c5e2ad705be8cd2dc3e2/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h#L61
* Forward tracks do not contribute to the primary vertex, then, for any  MFT, MCH, MFTMCH* 
  * we can skip this test test : https://github.com/AliceO2Group/AliceO2/blob/ca93a6955100ff8151c1c5e2ad705be8cd2dc3e2/Detectors/Vertexing/src/VertexTrackMatcher.cxx#L176-L178 
  * creator always return false

Does this address the inconsistency?

shahor02:
@rpezzi even if all current tasks don't need redundancy reduction  for the forward tracks, the RecoContainerCreateTracksVariadic behaviour must be to 
1) set the flagUsed2 for the constituents of the track *IF* the creator lambda returned true for this track.
2) skip the track it is flagged as used.
Simply, to avoid the redundancy reduction your creator lambda must return `false` on these tracks,

For the VertexTrackMatcher: in principle, yes.

PR #8576, opened 2022-04-13
DPL DataDescriptorQueryBuilder: Implementing support for property modifiers

matthiasrichter:
First implemented operation is negation of the subSpecification matcher to
realize "all but specification", e.g. `EMC/RAW/!0`

### DIFF
--- a/Framework/Core/src/DataDescriptorQueryBuilder.cxx
+++ b/Framework/Core/src/DataDescriptorQueryBuilder.cxx
@@ -41,6 +41,7 @@ enum QueryBuilderState {
   IN_END_QUERY,
   IN_STRING,
   IN_NUMBER,
+  IN_NEGATION,
   IN_ERROR,
   IN_BEGIN_ATTRIBUTES,
   IN_END_ATTRIBUTES,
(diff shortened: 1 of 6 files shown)


Framework/Core/src/DataDescriptorQueryBuilder.cxx
```diff
@@ -217,28 +246,29 @@ std::vector<InputSpec> DataDescriptorQueryBuilder::parse(char const* config)
           nodes.push_back(DescriptionValueMatcher{*currentDescription});
         } else if (*next == '?' && assignLastStringMatch("description", 16, currentDescription, IN_BEGIN_ATTRIBUTES)) {
           nodes.push_back(DescriptionValueMatcher{*currentDescription});
-        } else if (*next == '\0' && assignLastStringMatch("description", 16, currentDescription, IN_BEGIN_ATTRIBUTES)) {
+        } else if (*next == '\0' && assignLastStringMatch("description", 16, currentDescription, IN_END_MATCHER)) {
```
matthiasrichter, line 249: @ktf I think this was a typo, should go to END_MATCHER if we are on the last character of the string

Framework/Core/src/DataDescriptorQueryBuilder.cxx
```diff
@@ -217,28 +246,29 @@ std::vector<InputSpec> DataDescriptorQueryBuilder::parse(char const* config)
...
         } else if (*next == '\0' && assignLastNumericMatch("subspec", currentSubSpec, IN_END_MATCHER)) {
-          nodes.push_back(SubSpecificationTypeValueMatcher{*currentSubSpec});
         } else {
           error("Expected a number");
+          break;
         }
+        auto backup = states.back();
+        states.pop_back();
+        pushMatcher(SubSpecificationTypeValueMatcher{*currentSubSpec});
+        states.push_back(backup);
```
matthiasrichter, line 271: I'm checking a better solution here, we need to check the state before the one just added in `assignLastNumericMatch`. Have not yet disentangled because the method is in itself checking if the given state can be added.

ktf:
For the record, I think we are now (since ROOT 6.24) allowed to have std::visit in headers, it might actually simplify code a bit...

ktf:
Is this ready to go? It LGTM.

matthiasrichter:
Yes, that's ready to go if there is no further comments, so I will merge. I can look into using std::visit in a follow up

PR #8578, opened 2022-04-13
Add theta calculation to Fwd track

pnwkw:


shahor02 (approved):
CI failure unrelated, merging

PR #8580, opened 2022-04-13
Implemented reading MFT tracks from AOD files

pnwkw:


EventVisualisation/Workflow/include/EveWorkflow/AO2DConverter.h
```diff
@@ -38,14 +38,14 @@ namespace o2::event_visualisation
 class TPCFastTransform;
 
 struct AO2DConverter {
-  o2::framework::Configurable<std::string> jsonPath{"jsons-folder", "./jsons", "name of the folder to store json files"};
+  o2::framework::Configurable<std::string> jsonPath{"jsons-folder", "./json", "name of the folder to store json files"};
```
shahor02, line 41: actually, also the directory on /home/epn/ where the jsons are stored is `jsons`.

EventVisualisation/Workflow/src/EveWorkflowHelper.cxx
```diff
@@ -298,7 +298,33 @@ void EveWorkflowHelper::drawAOD(EveWorkflowHelper::AODFullTrack const& track, fl
 
   auto const tr = o2::track::TrackParCov(track.x(), track.alpha(), arraypar, covpar);
 
-  addTrackToEvent(tr, GID::ITSTPCTRDTOF, trackTime, 0., GID::ITSTPCTRD);
+  addTrackToEvent(tr, detectorMapToGID(track.detectorMap()), trackTime, 0.);
```
shahor02: The `addTrackToEvent` takes GlobalTrackID (track type a.k.a Source + index in the source container + optional 2 bits) as an argument, while detectorMapToGID returns just a Source. Since here you don't care about the index, you can rename the detectorMapToGID to detectorMapToGIDSource and use it as 
```
addTrackToEvent(tr, GID{0, detectorMapToGIDSrc(track.detectorMap())}, trackTime, 0.);
```

EventVisualisation/Workflow/include/EveWorkflow/AO2DConverter.h
pnwkw: I changed this to follow the default directory from which ED reads them:

https://github.com/AliceO2Group/AliceO2/blob/191fc132d064ed8bfc256c9024bb2ed4eca1f95f/EventVisualisation/View/src/Options.cxx#L59

In any case I think the best solution would be to extract this value to a header, so it is in sync in both ED and the AOD workflow.

EventVisualisation/Workflow/src/EveWorkflowHelper.cxx
pnwkw: Thanks, this is now fixed.

shahor02 (approved):

PR #8581, opened 2022-04-13
Pass headers in case of redirection in navigateURL...

shahor02:
@afurs this was the problem.


costing (approved):

sawenzel (approved):

afurs:
> @afurs this was the problem.

@shahor02 ok, thanks for fixing!

PR #8588, opened 2022-04-14
DPL Analysis: workaround for setting self-index binding before invoking process()

aalkin:
@fgrosa this explicitly sets the self-index binding, so that your code works with https://github.com/AliceO2Group/O2Physics/pull/668 reverted.


fgrosa:
> [AliceO2Group/O2Physics#668](https://github.com/AliceO2Group/O2Physics/pull/668)

thanks a lot @aalkin!

PR #8594, opened 2022-04-15
Fix for not throwing exception in cpv::RawReaderMemory when processin…

sevdokim:
…g corrupted RDH
Hello @davidrohr @shahor02 
This is a fix to catch exception which is thrown when reading corrupted cpv raw data.

### DIFF
--- a/Detectors/CPV/reconstruction/src/RawReaderMemory.cxx
+++ b/Detectors/CPV/reconstruction/src/RawReaderMemory.cxx
@@ -66,7 +66,7 @@ RawErrorType_t RawReaderMemory::next()
         e == RawErrorType_t::kRDH_DECODING ||       // incorrect rdh -> fatal error
         e == RawErrorType_t::kPAYLOAD_INCOMPLETE || // we reached end of mRawMemoryBuffer but payload size from rdh tells to read more
         e == RawErrorType_t::kSTOPBIT_NOTFOUND) {   //new HBF orbit started but no stop bit found, need to return
-      return e;                                     //some principal error occured -> stop reading.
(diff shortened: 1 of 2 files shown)


shahor02:
@sevdokim the https://github.com/AliceO2Group/AliceO2/blob/302104609ee006de31a28241bb675c0243ac0fcf/Detectors/CPV/reconstruction/src/RawReaderMemory.cxx#L59 will now detect the error and return a proper error code (w/o rethrowing) but the caller function does not check the returned value, it just catches exceptions. 
You should check it and in case of error abandon the TF processing (still sending output to not block the workflow and logging corresponding message, like you do for the deadbeef).

sevdokim:
Hi @shahor02 
is it really necessary to abandon whole TF in case of bad rdh catched? Currently it just skips page with bad rdh and process the rest of TF.

shahor02:
@sevdokim strictly speaking it is not necessary to skip whole TF, but RDH errors usually mean memory problems, likely the whole TF will be corrupted.
In any case, I don't see where do you skip the page with bad RDH: in https://github.com/AliceO2Group/AliceO2/blob/302104609ee006de31a28241bb675c0243ac0fcf/Detectors/CPV/workflow/src/RawToDigitConverterSpec.cxx#L129 you are not checking the return code of the `next()` and proceed to decoding unless the exception was caught, but it will not be caught there since you have a catch in the `nextPage()`.

sevdokim:
Hi @shahor02 
you are absolutely right! Thank you very much for careful review of the code. Skipping of corrupted tf is implemented now. It seems to be working fine with randomly generated rdh versions.

shahor02 (approved):

PR #8603, opened 2022-04-17
Properly disable numa configuration if numactl is not there

ktf:
The current code does not work on system where which prints out
on stdour "numactl not found".

### DIFF
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -87,7 +87,7 @@ fi
 [[ $SHMTHROW == 0 ]] && ARGS_ALL+=" --shm-throw-bad-alloc 0"
 [[ ! -z $SHM_MANAGER_SHMID ]] && ARGS_ALL+=" --shm-no-cleanup on --shmid $SHM_MANAGER_SHMID"
 [[ $NORATELOG == 1 ]] && ARGS_ALL+=" --fairmq-rate-logging 0"
-if [[ $EPNSYNCMODE == 1 ]] || [[ `which numactl 2> /dev/null | wc -l` != "0" && `numactl -H | grep "node . size" | wc -l` -ge 2 ]]; then
+if [[ $EPNSYNCMODE == 1 ]] || type numactl >/dev/null 2>&1 && [[ `numactl -H | grep "node . size" | wc -l` -ge 2 ]]; then
   [[ $NUMAGPUIDS != 0 ]] && ARGS_ALL+=" --child-driver 'numactl --membind $NUMAID --cpunodebind $NUMAID'"
 fi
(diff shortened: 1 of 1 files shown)


ktf:
@davidrohr before I forget about it again...

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -87,7 +87,7 @@ fi
 [[ $SHMTHROW == 0 ]] && ARGS_ALL+=" --shm-throw-bad-alloc 0"
 [[ ! -z $SHM_MANAGER_SHMID ]] && ARGS_ALL+=" --shm-no-cleanup on --shmid $SHM_MANAGER_SHMID"
 [[ $NORATELOG == 1 ]] && ARGS_ALL+=" --fairmq-rate-logging 0"
-if [[ $EPNSYNCMODE == 1 ]] || [[ `which numactl 2> /dev/null | wc -l` != "0" && `numactl -H | grep "node . size" | wc -l` -ge 2 ]]; then
+if [[ $EPNSYNCMODE == 1 ]] || type numactl >/dev/null 2>&1 && [[ `numactl -H | grep "node . size" | wc -l` -ge 2 ]]; then
```
davidrohr, line 90: No output should go to stdout (It regularly ends up in JSON and XML files), why do you have 2>&1? Can't we just remove that?

davidrohr (changes requested):

prodtests/full-system-test/dpl-workflow.sh
ktf: Because BSD `type` / `which` like the one on macOS prints "numactl not found" on stdout. I think the only thing you can rely on is that when the argument does not exist they will exit with a non zero exit code.

ktf: Also notice that here I do `>/dev/null 2>&1` which means both go to `/dev/null` (the logic is: first you redirect stdout to dev null and then you redirect stderr to the **redirected** stdout).

davidrohr: Ahh, ok, didn't know that. I assumed it would send stdout to /dev/null and afterwards make stderr write to stdout. Wasn't aware that the `> /dev/null` also applied to that. Usually, to send everything to `/dev/null` I use `&> /dev/null`. :)

ktf: Does `&>` actually work in bash? I thought it was zsh only and `>/dev/null 2>&1` is for sure the "POSIX" way. Anyways, given we have other bash-isms I am also fine if you want to change to `&>`.

davidrohr: `&>` for sure works in bash.
What I find a bit confusing is that for a pipe one needs `|&` ...
Anyway, I already merged it anyway.

PR #8607, opened 2022-04-18
o2-eve-display workflow: evenly sampling tracks

jmyrcha:


davidrohr:
I am not so sure if this is really what we want. As discussed in the meeting, we should optionally limit the extraction of tracks to selected ITS readout frames. But then "thinning" the tracks via sampling in this way might also give a false impression of the event. It is kind of hiding in a weird way that the ED cannot process more tracks. The hard cut from the beginning as we have now is actually more clear. It shows that the tracks were cut, and at least the part that is drawn is correct. We can discuss at today's RC meeting what RC prefers.

davidrohr:
closing, since we'll do it in a different way as discussed.

PR #8608, opened 2022-04-18
[EMCAL-548] Adapting the cell converter to handle mc labels

hahassan7:
The digit to cell converter using the raw fitter has been adapted to handle mc labels. When converting the sampled digits to bunches the mc labels of the different digits are added using weighted average. The sampled digits are converted into two bunches, one high gain and one low gain. The difference between these bunches is the noise added to the sampled digits. Then these bunches will be fitted separately. First we fit the HG bunch, if is saturated we fit the LG bunch and the cell will be LG. If the HG bunch is not saturated then the cell will be high gain.

Another change is the changing the format of the Digit object. Now the digit object stores the GeV energy in "double" instead of ADC int. The reason for that is because the digit is originally initialized with GeV and then it gets converted to ADC in which it will be stored which will lead to the loss of precision of the energy. 
Other change on the Digit object is that, now it stores two kinds of noise, HG noise and LG noise (in uint16_t). The reason for that is to make the simulation pretty similar to data, in which the sampled digits will be converted into two bunches HG and LG.
Also the channel type object in the digit has been removed, and the type will be determined on fly (when getting it).

### DIFF
--- a/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Constants.h
+++ b/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Constants.h
@@ -90,14 +90,16 @@ std::ostream& operator<<(std::ostream& stream, ChannelType_t chantype);
 namespace constants
 {
 
-constexpr int OVERFLOWCUT = 950;             ///< sample overflow
-constexpr int ORDER = 2;                     ///< Order of shaping stages of the signal conditioning unit
-constexpr double TAU = 2.35;                 ///< Approximate shaping time
-constexpr Double_t EMCAL_TIMESAMPLE = 100.;  ///< Width of a timebin in nanoseconds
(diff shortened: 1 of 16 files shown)


mfasDa:
@sawenzel we changed the logic in the way that we don't add the noise value to the digit amplitude but carry them as separate values inside the digit as 16 bit unsigned int. As this we can properly simulate the two ADCs we have in hardware, for which unfortunately we see different noise values, and decide on which to take as we do it in data, based on the peak amplitude of the calo bunch which needs the full bunch to be reconstructed. Like this we simulated the behaviour in data, because the raw data always has the ADCs from the two digitizers. However this will increase the digit size by 32 bit. I hope this increase is tolerable. I think it is still better than duplicating the bunch as high and low gain bunch, because the signal is common an can be shared (only conversion factor differs but is multiplicative), only the noise is different, so if we would duplicate the bunch we would need to duplicate the 32 bit energy.

mfasDa (approved):

PR #8612, opened 2022-04-19
Fixes for MID decoding and data access

shahor02:


shahor02:
@dstocco

dstocco (approved):

DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
```diff
@@ -118,12 +118,12 @@ void o2::globaltracking::RecoContainer::createTracksVariadic(T creator) const
   usedData[GTrackID::TPCTRDTOF].resize(getTPCTRDTOFMatches().size());       // to flag used ITSTPC-TOF matches
   usedData[GTrackID::ITSTPCTRDTOF].resize(getITSTPCTRDTOFMatches().size()); // to flag used ITSTPC-TOF matches
 
+  static int BCDiffErrCount = 0;
+  constexpr int MAXBCDiffErrCount = 5;
   auto getBCDiff = [startIR = this->startIR](const o2::InteractionRecord& ir) {
-    static int BCDiffErrCount = 0;
-    constexpr int MAXBCDiffErrCount = 5;
     auto bcd = ir.differenceInBC(startIR);
     if (uint64_t(bcd) > o2::constants::lhc::LHCMaxBunches * 256 && BCDiffErrCount < MAXBCDiffErrCount) {
```
dstocco, line 125: Are you assuming that the number of orbits per TF is 256? Isn't it 128 now?

shahor02: 256 is the max we can have though in real data production we will likely have only 128 (but e.g. MC currently uses 256). To be rigorous, one should take the real NHBFperTF from the GRP, but I did not want to make RecoContainerCreateTracksVariadic on CCDB

PR #8614, opened 2022-04-19
Reduce default max memory to 12GB

mpuccio:
@shahor02 this is a more sensible default for normal workstations.


iouribelikov (approved):

shahor02 (approved):
thanks, merging as the change is trivial.

PR #8615, opened 2022-04-19
[EMCAL-630] Move log level of EMCAL decoding errors to "alarm"

mfasDa:


mfasDa:
Errors unrelated. @davidrohr do you think it can be merged, so it will appear in the next FLP suite?

PR #8617, opened 2022-04-19
Add option to change tracker algorithm only keeping the best track

dstocco:
This allows to run the old tracker algorithm, which results in a reduced efficiency, but it is faster and can be used to reconstruct TED data.
Just launch with o2-mid-reco-workflow --mid-tracker-keep-best


dstocco:
@pillot

pillot:
looks fine to me.

PR #8620, opened 2022-04-20
DPL: work with ROOT master

ktf:


Framework/Core/src/RCombinedDS.cxx
```diff
@@ -42,7 +42,9 @@ double loop on the entries of the two RDataSources.
 #endif
 
 #include <ROOT/TSeq.hxx>
+#if __has_include(<ROOT/RMakeUnique.hxx>)
 #include <ROOT/RMakeUnique.hxx>
```
pzhristov: I propose simply to replace RMakeUnique.hxx by memory. With c++17 they are equivalent according to the comment and content of the file.

ktf: I agree, but I could not find the suggestion. Where is it?

ktf: Actually I suspect it was already done and I simply did not drop the include.

pzhristov: See commit 8044f652ec11ffe96b9d8e1127f42dd2ce4a2524 in Root.

ktf: Indeed it seems to work without the header....

ktf:
Compiles fine where it matters. Merging.

PR #8626, opened 2022-04-21
Fix Partition<Filtered>, add separate sliceByCached for joins, filtered and nested filtered

saganatt:
Hi @ktf @aalkin 

the Join method is a copy of the Table `sliceByCached()`, and Filtered is a copy of FilteredBase method. The problem is in returning exact O2 table type, what apparently cannot be done without copies or making `sliceByCached()` a standalone function and changing the UI. I need the exact type so as to use `sliceByCached()` for faster event mixing, which will come in another PR.

The changes in Partition constructor fix a newly discovered bug in creating Partitions on Filtered. The buggy methods were called when creating `Partitions<Filtered>` inside `process()` or calling `sliceByCached()` on `Partitions<Filtered>`.

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -2133,6 +2133,20 @@ struct Join : JoinBase<Ts...> {
   using const_iterator = iterator;
   using filtered_iterator = typename table_t::template RowViewFiltered<Join<Ts...>, Ts...>;
   using filtered_const_iterator = filtered_iterator;
+
+  auto sliceByCached(framework::expressions::BindingNode const& node, int value)
+  {
+    uint64_t offset = 0;
+    std::shared_ptr<arrow::Table> result = nullptr;
+    auto status = this->getSliceFor(value, node.name.c_str(), result, offset);
+    if (status.ok()) {
+      auto t = Join<Ts...>({result}, offset);
+      this->copyIndexBindings(t);
+      return t;
+    }
+    o2::framework::throw_error(o2::framework::runtime_error("Failed to slice table"));
+    O2_BUILTIN_UNREACHABLE();
+  }
 };
 
 template <typename... Ts>
@@ -2321,6 +2335,11 @@ class FilteredBase : public T
     std::shared_ptr<arrow::Table> result = nullptr;
     auto status = ((table_t*)this)->getSliceFor(value, node.name.c_str(), result, offset);
     if (status.ok()) {
+      if (offset >= this->tableSize()) {
+        self_t fresult{{result}, SelectionVector{}, 0}; // empty slice
+        this->copyIndexBindings(fresult);
+        return fresult;
+      }
       auto start = offset;
       auto end = start + result->num_rows();
       auto start_iterator = std::lower_bound(mSelectedRows.begin(), mSelectedRows.end(), start);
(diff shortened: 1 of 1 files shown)


Framework/Core/include/Framework/ASoA.h
```diff
@@ -2133,6 +2133,20 @@ struct Join : JoinBase<Ts...> {
...
+    uint64_t offset = 0;
+    std::shared_ptr<arrow::Table> result = nullptr;
+    auto status = this->getSliceFor(value, node.name.c_str(), result, offset);
+    if (status.ok()) {
+      auto t = Join<Ts...>({result}, offset);
+      this->copyIndexBindings(t);
+      return t;
+    }
+    o2::framework::throw_error(o2::framework::runtime_error("Failed to slice table"));
+    O2_BUILTIN_UNREACHABLE();
```
ktf:
```suggestion
    if (this->getSliceFor(value, node.name.c_str(), result, offset).ok() == false) {
      o2::framework::throw_error(o2::framework::runtime_error("Failed to slice table"));
    }
    auto t = Join<Ts...>({result}, offset);
    this->copyIndexBindings(t);
    return t;
```

Framework/Core/include/Framework/ASoA.h
```diff
@@ -2627,10 +2676,34 @@ class Filtered<Filtered<T>> : public FilteredBase<typename T::table_t>
...
-  using FilteredBase<typename T::table_t>::sliceByCached;
+  auto sliceByCached(framework::expressions::BindingNode const& node, int value)
+  {
+    uint64_t offset = 0;
+    std::shared_ptr<arrow::Table> result = nullptr;
+    auto status = ((table_t*)this)->getSliceFor(value, node.name.c_str(), result, offset);
+    if (status.ok()) {
+      auto start = offset;
+      auto end = start + result->num_rows();
+      auto start_iterator = std::lower_bound(this->getSelectedRows().begin(), this->getSelectedRows().end(), start);
```
ktf: Isn't this equal_range?

saganatt: I don't think so. We have different values of row numbers, not all of them selected, and the `start_iterator` might point at some number higher than `offset`. Then the `stop_iterator` is compared with `end`, which is a totally different number.

Framework/Core/include/Framework/ASoA.h
```diff
@@ -2521,7 +2536,33 @@ class Filtered : public FilteredBase<T>
...
+      this->copyIndexBindings(fresult);
+      return fresult;
+    }
+    auto start = offset;
+    auto end = start + result->num_rows();
+    auto start_iterator = std::lower_bound(this->getSelectedRows().begin(), this->getSelectedRows().end(), start);
+    auto stop_iterator = std::lower_bound(start_iterator, this->getSelectedRows().end(), end);
+    SelectionVector slicedSelection{start_iterator, stop_iterator};
+    std::transform(slicedSelection.begin(), slicedSelection.end(), slicedSelection.begin(),
+                   [&](int64_t idx) {
```
ktf: In general try to avoid `[&]` but specify the actual elements you want to capture. I guess `[start]` here.

Framework/Core/include/Framework/ASoA.h
aalkin: We have two lower bounds here, while equal_range is lower and upper. Unrelated, I think we should move this sort of selection vector manipulation in a separate function, since it is now used in multiple places.

ktf (approved):
LGTM if it compiles.

saganatt:
CI errors unrelated, can we merge?

PR #8639, opened 2022-04-23
Use tfCounter in time-slot calib, provide real time.

shahor02:
don't merge


shahor02:
@TimoWilken the fullCI again fails due to the missing alien-token, any idea why?

martenole (approved):

PR #8640, opened 2022-04-24
CCDBPopulator optionally fatals on failed upload (def=on)

shahor02:


Detectors/Calibration/workflow/CCDBPopulatorSpec.h
```diff
@@ -84,13 +85,17 @@ class CCDBPopulator : public o2::framework::Task
 
       LOG(info) << "Storing in ccdb " << wrp->getPath() << "/" << wrp->getFileName() << " of size " << pld.size()
                 << " Valid for " << wrp->getStartValidityTimestamp() << " : " << wrp->getEndValidityTimestamp();
-      mAPI.storeAsBinaryFile(&pld[0], pld.size(), wrp->getFileName(), wrp->getObjectType(), wrp->getPath(),
-                             *md, wrp->getStartValidityTimestamp(), wrp->getEndValidityTimestamp());
+      int res = mAPI.storeAsBinaryFile(&pld[0], pld.size(), wrp->getFileName(), wrp->getObjectType(), wrp->getPath(),
+                                       *md, wrp->getStartValidityTimestamp(), wrp->getEndValidityTimestamp());
+      if (res && mFatalOnFailure) {
+        LOGP(fatal, "failed on uploading to {} / {}", mAPI.getURL(), wrp->getPath());
```
chiarazampolli, line 91: Hi, 
In my PR, I explicitly mention the error - see https://github.com/AliceO2Group/AliceO2/pull/8641. But since you now use "LOGP(alarm)" it will also appear in the infologger, right?

shahor02: yes, now instead of writing to cerr, the CcdbApi will produce a message seen by the infologger.

chiarazampolli: Thanks, I didn’t like it either that it wrote to stderr, but didn’t want to change it as I understood that it was better to leave the CcdbApi as it was (but I would have brought this up).

PR #8641, opened 2022-04-24
Optional FATAL in case CCDB populator fails in uploading object (def …

chiarazampolli:
…= true)

@shahor02 , @davidrohr , @sawenzel , @Barthelemy


chiarazampolli:
Closing, since @shahor02  was faster than me (https://github.com/AliceO2Group/AliceO2/pull/8640)

PR #8642, opened 2022-04-24
Provide MCH clusters via global-track-cluster-reader and RecoContainer

shahor02:


shahor02:
@dstocco could you please add a cluster reader for the MID? Something like the `MUON/MCH/Workflow/ClusterReaderSpec.cxx`

shahor02:
The MID clusters are accessible from the RecoContainer but can be requested only if provided by the upstream workflow: at the moment there is no MID cluster reader to. Once provided by the MID, will add to InputHelper and the clusters will be accessible via `o2-global-track-cluster-reader`.

PR #8646, opened 2022-04-25
DPL: ignore DomainInfoHeader in input proxy

ktf:


Framework/Core/src/ExternalFairMQDeviceProxy.cxx
```diff
@@ -229,6 +230,10 @@ InjectorFunction dplModelAdaptor(std::vector<OutputSpec> const& filterSpecs, DPL
     static int64_t dplCounter = -1;
     dplCounter++;
     for (int msgidx = 0; msgidx < parts.Size(); msgidx += 2) {
+      const auto dih = o2::header::get<DomainInfoHeader*>(parts.At(msgidx)->GetData());
+      if (dih == nullptr) {
+        continue;
+      }
```
knopers8: If you want to ignore it, don't you need `if (dih != nullptr) continue`?

ktf: yes, indeed. Changed. I am actually not sure it's correct, given it might happen that it's attached to data..

ktf:
#8649 should avoid this in the first place. Closing.

PR #8648, opened 2022-04-25
TRD fix tracklethcheader 1 option, should not be used.

bazinski:
- due to the ambiguity of tracklethcheader and digithcheader, which option 1 enables it should not be used.
- however the data is in fact that format.
- the code now works for that option, up to the ambiguity.
 
Raw data tests will be finished before this finishes in the ci.


davidrohr:
What is the outcome of the raw data tests? Successful? Can this be merged?

bazinski:
@davidrohr oops forgot to come and update. Yes tests all passed.

PR #8652, opened 2022-04-25
DPL: refactoring of WorkflowSerializationHelpers

matthiasrichter:
Preparing for better serialization of DataMatchers
- new states added to importer
- json import of InputSpec now based on stack of DataMatchers

### DIFF
--- a/Framework/Core/src/WorkflowSerializationHelpers.cxx
+++ b/Framework/Core/src/WorkflowSerializationHelpers.cxx
@@ -56,6 +56,10 @@ struct WorkflowImporter : public rapidjson::BaseReaderHandler<rapidjson::UTF8<>,
     IN_INPUT_DESCRIPTION,
     IN_INPUT_SUBSPEC,
     IN_INPUT_LIFETIME,
+    IN_INPUT_MATCHER,
+    IN_INPUT_MATCHER_OPERATION,
+    IN_INPUT_LEFT_MATCHER,
+    IN_INPUT_RIGHT_MATCHER,
     IN_INPUT_OPTIONS,
     IN_OUTPUT,
     IN_OUTPUT_BINDING,
@@ -146,6 +150,18 @@ struct WorkflowImporter : public rapidjson::BaseReaderHandler<rapidjson::UTF8<>,
         break;
       case State::IN_INPUT_SUBSPEC:
         s << "IN_INPUT_SUBSPEC";
+        break;
+      case State::IN_INPUT_MATCHER:
+        s << "IN_INPUT_MATCHER";
+        break;
+      case State::IN_INPUT_MATCHER_OPERATION:
+        s << "IN_INPUT_MATCHER_OPERATION";
+        break;
+      case State::IN_INPUT_LEFT_MATCHER:
+        s << "IN_INPUT_LEFT_MATCHER";
+        break;
+      case State::IN_INPUT_RIGHT_MATCHER:
+        s << "IN_INPUT_RIGHT_MATCHER";
         break;
       case State::IN_INPUT_LIFETIME:
         s << "IN_INPUT_LIFETIME";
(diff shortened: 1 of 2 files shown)


Framework/Core/src/WorkflowSerializationHelpers.cxx
```diff
@@ -277,34 +298,28 @@ struct WorkflowImporter : public rapidjson::BaseReaderHandler<rapidjson::UTF8<>,
...
-          std::make_unique<DataDescriptorMatcher>(
-            DataDescriptorMatcher::Op::And,
-            DescriptionValueMatcher{ContextRef{1}},
-            std::make_unique<DataDescriptorMatcher>(
-              DataDescriptorMatcher::Op::And,
-              SubSpecificationTypeValueMatcher{ContextRef{2}},
-              std::make_unique<DataDescriptorMatcher>(DataDescriptorMatcher::Op::Just,
-                                                      StartTimeValueMatcher{ContextRef{0}})))};
-
-        dataProcessors.back().inputs.push_back(InputSpec({binding}, std::move(expectedMatcher00)));
```
matthiasrichter, line 294: @ktf in this case, the `lifetime` and `inputOptions` parameters were missing, was this a bug? The two other cases below are using those two additional parameters.

Framework/Core/src/WorkflowSerializationHelpers.cxx
```diff
@@ -277,34 +298,28 @@ struct WorkflowImporter : public rapidjson::BaseReaderHandler<rapidjson::UTF8<>,
...
-      } else if (inputHasDescription) {
-        if (inputHasSubSpec) {
-          dataProcessors.back().inputs.push_back(InputSpec(binding, origin, description, subspec, lifetime, inputOptions));
-        } else {
-          dataProcessors.back().inputs.push_back(InputSpec(binding, {origin, description}, lifetime, inputOptions));
+      auto buildMatcher = [](auto& nodes) -> std::unique_ptr<DataDescriptorMatcher> {
+        auto lastMatcher =
+          std::make_unique<DataDescriptorMatcher>(DataDescriptorMatcher::Op::Just,
+                                                  StartTimeValueMatcher(ContextRef{0}));
```
matthiasrichter: `ContextRef{0}` is always reserved for start time value matching, right?

ktf: Yes. We should probably change it to `data_matcher::STARTTIME_POS`.

matthiasrichter: ok, will do

ktf:
Issue looks genuine?

matthiasrichter:
Need to look into the error of the QC test
```
[ERROR] invalid workflow in o2-qc: InputSpec for Enumeration must be fully qualified
```

matthiasrichter:
The issue comes from the fact, that some input specs are required to be fully qualified, and this is implemented by requiring the input spec to be of type `ConcreteDataMatcher`. This is not the case with this PR, this only places `DataDescriptorMatcher` objects in the input spec when restoring the workflow from json configuration. The matchers still describe fully qualified specs, but in a different representation. The logic needs to be revised.

Since this is becoming very complex, as a simple solution a helper method is needed which creates a `ConcreteDataMatcher` from a `DataDescriptorMatcher` object if this is describing a fully qualified data spec with unique origin/description/subspec and only `and` operations. The available helpers, e.g. `extractMatcherInfo` are not handling more complex matcher objects. To move on with this task, the first implementation will only support a fixed DataDescriptorMatcher hierarchy, and later we can extend also the other helpers.

matthiasrichter:
with merge #8677, the original behavior of adding simplified ConcreteDataMatcher objects has been restored in this PR.

matthiasrichter:
The remaining CI issue is unrelated,
```
[ERROR] Requested resource does not exist: http://ccdb-test.cern.ch:8080/Test/CachingA/500/
```
makes `CCDB/test/testBasicCCDBManager.cxx` failing

PR #8656, opened 2022-04-26
MUON: dots in DCS aliases are legit

aphecetche:


chiarazampolli:
Hello @ktf , @shahor02 ,
Who should approve this one?
Chiara

shahor02 (approved):

ktf:
@aphecetche himself, that's why it does not show, because you cannot approve yourself... I think he could have appointed someone else though. Next time could you try?

aphecetche:
> @aphecetche himself, that's why it does not show, because you cannot approve yourself... I think he could have appointed someone else though. Next time could you try?

@ktf I'm not so sure I understand : am I supposed to appoint someone for a review (that I guess I can do) or someone with approval right, in which case I guess I'll get back to the usual suspects (as there's no muoner with merging rights) won't I ?

ktf:
Well, the best would be to have two codeowners per package, so that one can approve the PRs of the other. There is apparently no "fallback to usual suspects" for this particular case and you need to pick a reviewer by hand.

ktf:
Well, the best would be to have two codeowners per package, so that one can approve the PRs of the other. There is apparently no "fallback to usual suspects" for this particular case and you need to pick a reviewer by hand.

aphecetche:
@ktf @shahor02 one option for muon would indeed be to appoint a second merger. Don't know what's the policy to decide on whether or not to add more people with merging rights, but if that's possible at all, I think @pillot would be a good candidate to review and merge my PRs :wink:

PR #8657, opened 2022-04-26
Only GLO objects can be run-dependent

chiarazampolli:
@shahor02 , @ktf , what do you think?

### DIFF
--- a/Framework/Core/src/CCDBHelpers.cxx
+++ b/Framework/Core/src/CCDBHelpers.cxx
@@ -185,11 +185,24 @@ auto populateCacheWith(std::shared_ptr<CCDBFetcherHelper> const& helper,
     std::map<std::string, std::string> headers;
     std::string path = "";
     std::string etag = "";
+    bool isGLOobject = false;
+    for (auto& meta : route.matcher.metadata) {
+      if (meta.name == "ccdb-path") {
+        path = meta.defaultValue.get<std::string>();
+        if (path.find("GLO/Config") != std::string::npos) {
+          isGLOobject = true;
+        }
+      }
+    }
     for (auto& meta : route.matcher.metadata) {
       if (meta.name == "ccdb-path") {
         path = meta.defaultValue.get<std::string>();
       } else if (meta.name == "ccdb-run-dependent" && meta.defaultValue.get<bool>() == true) {
-        metadata["runNumber"] = dtc.runNumber;
+        if (isGLOobject) {
+          metadata["runNumber"] = dtc.runNumber;
+        } else {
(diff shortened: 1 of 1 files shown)


shahor02:
What is the reason to have such a strong restriction? And why only GLO/Config?

chiarazampolli:
Don't we want that CCDB objects are timestamp dependent? See e.g. when MID had this flag on, then it was not working. 

One could also have all GLO objects, I was thinking that these are those that will be run dependent (in fact, only the ECS one will).

shahor02:
The metadata (e.g. run) dependence comes on top of the time-dependence. Simply, if the query comes with run request, one should be sure that the object was uploaded with the runNumber tag.
Have I misunderstood what you   have said?

chiarazampolli:
I understand that the latest matching object would be returned, but do we really want to allow that objects are picked by run, if they are not supposed to be run dependent? I think we should just go with the timestamp only (in the fetcher, not the CCDB api) - even if the run number is attached to every object. 
In addition, for the CcdbApi, maybe it could be useful to have a function that returns all objects valid for a run.

shahor02:
> I understand that the latest matching object would be returned, but do we really want to allow that objects are picked by run, if they are not supposed to be run dependent? I think we should just go with the timestamp only (in the fetcher, not the CCDB api) - even if the run number is attached to every object.

The selection on the runNumber will be involved only if author explicitly asked this in the `ccdbParamSpec` (or if this is hardcoded in the fetcher), why not leaving to the author to decide if his object can be run-dependent (which would mean that no default object is possible for that type)?

>  In addition, for the CcdbApi, maybe it could be useful to have a function that returns all objects valid for a run.

do you mean w/o providing a time-stamp? Only @costing knows how to search for such objects

chiarazampolli:
Ok, if you think that this is not needed (but I remind you that MID had this option on by mistake), then I won't insist. For me, no object other than GLO should be accessed with the run number, so this PR made sense as a protection. CCDB should be timestamp-based, we always said it. 

`do you mean w/o providing a time-stamp? Only @costing knows how to search for such objects`

Yes. It might be useful that we can retrieve all objects valid for a run, not to be used in sim/reco, but for debug. Don't you agree?

shahor02:
The ccdb is anyways always timestamp-based. I am just afraid that by hardcoding restrictions we will shoot to our own leg. At the moment likely only GLO objects may overlap in time for parallel runs, but I cannot exclude that later we will not have run-specific CTP or analysis objects.

For the utility to retrieve all objects of given run, I agree this could be useful.

chiarazampolli:
Ok, since you are against this change, I will close the PR, and I will look into the other utility in the coming days (getting all objects matching a run).

PR #8659, opened 2022-04-26
Printing log when we find a default CCDB entry

chiarazampolli:
@shahor02 , @ktf , is this the correct place? From a local test it works, but I don't know if it covers all possibilities. 
I would allow the three keys "default", "Default", "DEFAULT", let me know what you think.

### DIFF
--- a/Framework/Core/src/CCDBHelpers.cxx
+++ b/Framework/Core/src/CCDBHelpers.cxx
@@ -23,6 +23,7 @@
 #include <TError.h>
 #include <TMemFile.h>
 #include <functional>
+#include <boost/algorithm/string/predicate.hpp>
 
 namespace o2::framework
 {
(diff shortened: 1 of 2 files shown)


shahor02:
@chiarazampolli the place is correct, but I would not consider different cases, just `default` (can be added https://github.com/pnwkw/AliceO2/blob/f5d53e06caa8d9224dcbe7d0bcb6e0d01af8c63f/Common/Utils/include/CommonUtils/NameConf.h#L69 and used for uploading). Also, I would not check the `true`, `1` etc.: non-default objects should not have this tag at all, just presence of `default` should be enough. So, I would change it simply to:
```
if (headers.find("default")!=headers.end()) {
  LOGP(info, "******** Default entry used for {} ********", path);
}
```

chiarazampolli:
Ok, I can remove the various options for writing "default", but what if someone used "default = false" to specify that it is not a default? I would keep the check on the value.

shahor02:
Who can use it? The default objects should be uploaded manually and this can be done by special operators only. If we can avoid passing the value at all, I would prefer this.

chiarazampolli:
Ok, I will change it.

chiarazampolli:
Done.

Framework/Core/src/CCDBHelpers.cxx
```diff
@@ -211,6 +212,10 @@ auto populateCacheWith(std::shared_ptr<CCDBFetcherHelper> const& helper,
         // FIXME: I should send a dummy message.
         continue;
       }
+      // printing in case we find a default entry
+      if (headers.find("default") != headers.end()) {
+        LOGP(info, "******** Default entry used for {} ********", path);
```
ktf:
```suggestion
        LOGP(detail, "******** Default entry used for {} ********", path);
```

Framework/Core/src/CCDBHelpers.cxx
```diff
@@ -23,6 +23,7 @@
 #include <TError.h>
 #include <TMemFile.h>
 #include <functional>
+#include <boost/algorithm/string/predicate.hpp>
```
ktf: Why do you need predicate?

chiarazampolli: It is indeed not needed anymore, after the changes following Ruben's comments, I will remove it. Thanks for spotting.

Framework/Core/src/CCDBHelpers.cxx
chiarazampolli: What is the effect of "detail"?

ktf: It's like info, but it does not go to infologger.

shahor02:
@chiarazampolli I've checked, the metadata with empty value does not work, so we have to provide something but should look on the presence of the tag only.

chiarazampolli:
Ciao, 

Yes, I had not understood that you proposed to not add the value at all. We should just make sure that the value is not misleading when we do the manual upload. 
Anyway, for the changes here, it does not matter. 
Chiara

chiarazampolli:
If there are no more comments, can this be approved?

shahor02 (approved):

PR #8661, opened 2022-04-26
Gain map using tracks: Loading topology correction from CCDB

matthias-kleiner:
other changes:
- adding normalization of extracted gain map per stack,region or none
- adding drawing of reference gain map
- some small fixes and code documentation
- debugging: dumping extracted maps to tree, creating merged gain map
- add option to performing refit of the tracks instead of propagating the track to extract precise track parameters
- add option to set charge type used for gain map extraction
- optimising default parameters
- option to use the last extracted gain map as a reference gain map

### DIFF
--- a/Detectors/TPC/calibration/include/TPCCalibration/CalibPadGainTracksBase.h
+++ b/Detectors/TPC/calibration/include/TPCCalibration/CalibPadGainTracksBase.h
@@ -19,7 +19,6 @@
 
 // o2 includes
 #include "TPCBase/CalDet.h"
-#include "TPCBase/Mapper.h"
 #include "TPCCalibration/FastHisto.h"
 
 #include <gsl/span>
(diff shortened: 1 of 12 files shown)


Detectors/TPC/calibration/include/TPCCalibration/CalibPadGainTracksBase.h
```diff
@@ -146,6 +158,14 @@ class CalibPadGainTracksBase
...
   void dumpToFile(const char* outFileName = "calPadGainTracksBase.root", const char* outName = "calPadGain") const;
 
+  /// dump extracted gain map to tree
+  /// \filename name of the output file
+  void dumpGainMapToTree(const std::string filename = "gainmap_debug.root") const;
+
+  /// dump extracted sigma map to tree
+  /// \filename name of the output file
+  void dumpSigmaMapToTree(const std::string filename = "sigmamap_debug.root") const;
+
```
wiechula: Why dump them separately, not in the same tree?

wiechula (dismissed):

Detectors/TPC/calibration/include/TPCCalibration/CalibPadGainTracksBase.h
matthias-kleiner: You are right, it would be better to dump them into the same tree

wiechula (approved):

wiechula:
Error seems unrelated, can this be merged?

PR #8663, opened 2022-04-27
Event Display: Add filter to only visualize tracks matching ITS readout frames & standardize workflow options

pnwkw:
Requires update of the PDP workflow script to include the `--disable-root-input` for ED.

### DIFF
--- a/EventVisualisation/Workflow/src/O2DPLDisplay.cxx
+++ b/EventVisualisation/Workflow/src/O2DPLDisplay.cxx
@@ -47,15 +47,15 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
     {"number-of_files", VariantType::Int, 300, {"maximum number of json files in folder"}},
     {"number-of_tracks", VariantType::Int, -1, {"maximum number of track stored in json file (-1 means no limit)"}},
     {"time-interval", VariantType::Int, 5000, {"time interval in milliseconds between stored files"}},
-    {"enable-mc", o2::framework::VariantType::Bool, false, {"enable visualization of MC data"}},
-    {"disable-mc", o2::framework::VariantType::Bool, false, {"disable visualization of MC data"}}, // for compatibility, overrides enable-mc
+    {"disable-mc", o2::framework::VariantType::Bool, false, {"disable visualization of MC data"}},
     {"display-clusters", VariantType::String, "ITS,TPC,TRD,TOF", {"comma-separated list of clusters to display"}},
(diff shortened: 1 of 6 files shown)


EventVisualisation/Workflow/src/O2DPLDisplay.cxx
```diff
@@ -174,17 +185,21 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
...
 
-  if (cfgc.options().get<bool>("read-from-files")) {
-    InputHelper::addInputSpecs(cfgc, specs, srcCl, srcTrk, srcTrk, useMC);
+  auto filterITSROF = cfgc.options().get<bool>("filter-its-rof");
+
+  if (filterITSROF) {
+    dataRequest->requestIRFramesITS();
   }
 
+  InputHelper::addInputSpecs(cfgc, specs, srcCl, srcTrk, srcTrk, useMC);
```
shahor02, line 194: Here you request reading data from the files, this can be done only if you are running display in a standalone mode over the data produced in advance. Using InputHelper should be conditioned:
```
if (!cfgc.options().get<bool>("disable-root-input")) {
  InputHelper::addInputSpecs(cfgc, specs, srcCl, srcTrk, srcTrk, useMC);
}
```
and `--disable-root-input` must be  requested when o2-eve-display runs with the direct input from the reco workflows., please add it to https://github.com/AliceO2Group/AliceO2/blob/6fb55ea45025b352b710e2bebdcc28b4cadbe556/prodtests/full-system-test/dpl-workflow.sh#L499

shahor02 (changes requested):

EventVisualisation/Workflow/src/O2DPLDisplay.cxx
pnwkw: According to:
https://github.com/AliceO2Group/AliceO2/blob/b8ff87d0fe54417fa1f7827127c16a1160112e6b/Detectors/GlobalTrackingWorkflow/helpers/src/InputHelper.cxx#L56=

the `addInputSpects` will itself do nothing if the option `--disable-root-input` is present in the `cfgc`, which is why I removed that `if` logic completely.

I'll fix the `dpl-workflow.sh`.

shahor02: Right, I forgot that. Then you just need to update the `dpl-workflow.sh`.

shahor02 (approved):

PR #8666, opened 2022-04-27
Fixing compilation warnings in DataFormat/Headers

matthiasrichter:
```
DataFormats/Headers/include/Headers/DataHeader.h:302:25: warning: moving a local object in a return statement prevents copy elision [-Wpessimizing-move]
DataFormats/Headers/include/Headers/NameHeader.h:60:12: warning: 'char* strncpy(char*, const char*, size_t)' output truncated before terminating nul copying 8 bytes from a string of the same length [-Wstringop-truncation]
DataFormats/Headers/include/Headers/NameHeader.h:60:12: warning: 'char* strncpy(char*, const char*, size_t)' output truncated copying 7 bytes from a string of length 8 [-Wstringop-truncation]
```

### DIFF
--- a/Algorithm/test/headerstack.cxx
+++ b/Algorithm/test/headerstack.cxx
@@ -41,7 +41,7 @@ BOOST_AUTO_TEST_CASE(test_headerstack)
   dh.payloadSize = 0;
 
   using Name8Header = o2::header::NameHeader<8>;
-  Name8Header nh("NAMEDHDR");
+  Name8Header nh("NAMEHDR");
 
   o2::header::Stack stack(dh, nh);
 
--- a/DataFormats/Headers/include/Headers/NameHeader.h
+++ b/DataFormats/Headers/include/Headers/NameHeader.h
@@ -42,6 +42,7 @@ struct NameHeader : public BaseHeader {
   static const o2::header::SerializationMethod sSerializationMethod;
   NameHeader() : BaseHeader(sizeof(NameHeader), sHeaderType, sSerializationMethod, sVersion), name()
   {
+    static_assert(N > 0, "Can not construct zero-length NameHeader");
     // set length of padding in the last byte
     uint8_t* lastByte = reinterpret_cast<uint8_t*>(this) + sizeof(NameHeader) - 1;
     *lastByte =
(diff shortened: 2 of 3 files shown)


DataFormats/Headers/include/Headers/NameHeader.h
```diff
@@ -57,7 +57,7 @@ struct NameHeader : public BaseHeader {
     *lastByte =
       reinterpret_cast<const uint8_t*>(this) + sizeof(NameHeader) - reinterpret_cast<const uint8_t*>(&name[N]);
     // here we actually want a null terminated string
-    strncpy(name, in.c_str(), N);
+    strncpy(name, in.c_str(), N < in.length() ? N : in.length());
```
ktf: Is this actually right? Don't we zero pad?

matthiasrichter: Thanks, but I think the padding is not a problem in this `NameHeader` example. But the string must be zero terminated, and that might be missing, so that would be a real bug.

ktf:
```suggestion
    strncpy(name, in.c_str(), N);
```

matthiasrichter: Absolutely correct, the actual warning is also correct, it is pointing to where NameHeader is used and initialized with strings of the same or bigger length than the template parameter N. Thanks for spotting this.

matthiasrichter: fixed

Algorithm/test/headerstack.cxx
```diff
@@ -41,7 +41,7 @@ BOOST_AUTO_TEST_CASE(test_headerstack)
   dh.payloadSize = 0;
 
   using Name8Header = o2::header::NameHeader<8>;
-  Name8Header nh("NAMEDHDR");
+  Name8Header nh("NAMDHDR");
```
matthiasrichter: @ktf I think the warning is actually pointing here where a string of the same length as the NameHeader string length is used, so there is no space for zero termination, and this is what the warning says.

ktf: yes, agreed. Maybe we should throw if the Name8Header string has length > 7...

ktf: or assert...

matthiasrichter: Then I would prefer to throw, since we build with NODEBUG

DataFormats/Headers/include/Headers/NameHeader.h
```diff
@@ -64,8 +69,12 @@ struct NameHeader : public BaseHeader {
   const char* getName() const { return name; }
   size_t getNameLength() const
   {
-    const uint8_t* lastByte = reinterpret_cast<const uint8_t*>(this) + size() - 1;
-    return (lastByte - reinterpret_cast<const uint8_t*>(name)) - *lastByte + 1;
+    if constexpr (N > 0) {
```
matthiasrichter: this is not working, because it is intended to retrieve a NamedHeader of unknown size from a raw buffer and call `getNameLength`. Going to remove it. The buffer is self consistent since the header length is stored in the corresponding  header member. The important protection is at creation time.

Algorithm/test/headerstack.cxx
matthiasrichter: I have changed the type of the argument so that we can use static_assert

DataFormats/Headers/include/Headers/NameHeader.h
matthiasrichter: Not needed anymore, since we make sure in the constructor `N > 0`` at creation time of the object.

davidrohr (approved):

PR #8667, opened 2022-04-27
Add/use createDirectoriesIfAbsent method, robust against concurrent calls

shahor02:


Common/Utils/src/FileSystemUtils.cxx
```diff
@@ -53,4 +55,16 @@ std::vector<std::string> listFiles(std::string const& searchpattern)
   return listFiles("./", searchpattern);
 }
 
+void createDirectoriesIfAbsent(std::string const& path)
+{
+  if (!std::filesystem::exists(path)) {
+    if (!std::filesystem::create_directories(path)) {
```
ironMann: Hi @shahor02, I would simplify this as:
```c++
std::filesystem::create_directories(path);
if (!std::filesystem::is_directory(path)) {
    throw std::runtime_error(fmt::format("Failed to create {} directory", path));
}
```

shahor02: Thanks, indeed, I see that it is not throwing on the existing directory! Changed to even shorter
```
if (!std::filesystem::create_directories(path) && !std::filesystem::is_directory(path)) {
   throw std::runtime_error(fmt::format("Failed to create {} directory", path));
}
```

PR #8668, opened 2022-04-27
TRD vDrift and ExB calibration updated

martenole:
To send CCDB object without relying on the CCDB manager.

Not sure if we want to merge the second commit. This would enable the calibration per default in the FST.

### DIFF
--- a/prodtests/full-system-test/calib-workflow.sh
+++ b/prodtests/full-system-test/calib-workflow.sh
@@ -8,6 +8,7 @@ fi
 if [[ $BEAMTYPE != "cosmic" ]]; then
   has_detector_calib TPC && has_detectors TPC ITS TRD TOF && add_W o2-tpc-scdcalib-interpolation-workflow "$DISABLE_ROOT_OUTPUT --disable-root-input --pipeline $(get_N tpc-track-interpolation TPC REST)" "$ITSMFT_FILES"
   has_detector_calib ITS && has_detectors ITS && has_detectors_reco ITS && has_detector_matching PRIMVTX && [[ ! -z "$VERTEXING_SOURCES" ]] && add_W o2-calibration-mean-vertex-calibration-workflow
+  has_detector_calib TRD && has_detector ITS TPC TRD && add_W o2-calibration-trd-vdrift-exb
 fi
 
 true # everything OK up to this point, so the script should return 0 (it is !=0 already if a has_detector check fails)
--- a/Detectors/TRD/calibration/include/TRDCalibration/CalibratorVdExB.h
+++ b/Detectors/TRD/calibration/include/TRDCalibration/CalibratorVdExB.h
@@ -20,6 +20,8 @@
 #include "DetectorsCalibration/TimeSlot.h"
 #include "DataFormatsTRD/Constants.h"
 #include "DataFormatsTRD/AngularResidHistos.h"
+#include "CCDB/CcdbObjectInfo.h"
+#include "DataFormatsTRD/CalVdriftExB.h"
 
 #include "Rtypes.h"
 #include "TProfile.h"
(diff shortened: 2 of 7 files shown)


shahor02:
> Not sure if we want to merge the second commit. This would enable the calibration per default in the FST.

Actually, since the workflows from calib-workflow.sh are attached to the (sync) processing workflow, running there aggregators (e.g. vdrift and meanvertex) makes sense only in the FST. On the EPN they should be substituted by the output proxies sending the data requested by the aggregators + separate workflow started on the aggregator node.
Since at the moment we are not running online calibrations on the EPNs, I think the current code is fine, but we should eventually substitute these aggregators by the corresponding output proxy and move the aggregators to separate script, like in O2DPG/DATA/production/calib. Then they can also be started in the FST as separate workflows on the same PC. 
Don't know what plans @davidrohr has on this.

davidrohr:
Fine with me in general, I am also working on some automation to be able to run the aggregators locally, though still via 2 proxies, in order to run the full calib workflow in the FST.
Just, before merging I want to test the performance and stability in 8 GPU FST on the EPNs, will do that next week.

chiarazampolli:
Ciao @davidrohr ,

I am also testing the aggregator with calibrations run locally. Just to not duplicate the efforts, you are just working on the FST framework or also integrating the calibrations? I am working on the latter. 

Chiara

davidrohr:
Just working on the framework, not on any individual aggregators

chiarazampolli:
Ok, I need to fix something, but then I will commit the first version of the scripts (hopefully tomorrow), so that you can take a look if it fits. 
Thanks!!!

davidrohr (changes requested):
With this PR the full system test on the EPN with 8 GPUs produces backpressure, and I see that the SHM runs full. (I retried the exact same software without this PR and then it works.)

To reproduce, on the EPN:
```
GEN_TOPO_WORKDIR=/home/drohr/tmp4/ QC_REDIRECT_MERGER_TO_LOCALHOST=1 WORKFLOW_PARAMETERS=QC,CALIB,EVENT_DISPLAY WORKFLOW_DETECTORS_QC=FT0,FV0,FDD,ZDC,TOF,ITS,MFT,MID,MCH,EMC,PHS,CPV CONFIG_EXTRA_PROCESS_o2_gpu_reco_workflow="GPU_global.benchmarkMemoryRegistration=1;" TFDELAY=5 NTIMEFRAMES=100 $O2_ROOT/prodtests/full-system-test/start_tmux.sh dd
```

To check the SHM:
```
fairmq-shmmonitor -v -i
```

martenole:
Thanks a lot for checking @davidrohr  I am having a look

martenole:
So the backpressure was traced back to a bug in the workflow. The output from the TrackBasedCalib was sent only for every 200th TF, but the output was not declared sporadic. This is now fixed and with the current status this PR does not cause backpressure anymore on the FST on the EPNs.
Currently I send the output for every TF. I tried also to declare it Lifetime::Sporadic but this then again lead to backpressure.

shahor02:
> So the backpressure was traced back to a bug in the workflow. The output from the TrackBasedCalib was sent only for every 200th TF, but the output was not declared sporadic. This is now fixed and with the current status this PR does not cause backpressure anymore on the FST on the EPNs. Currently I send the output for every TF. I tried also to declare it Lifetime::Sporadic but this then again lead to backpressure.

yes, the Sporadic is not working as expected, @ktf is checking

martenole:
Hi @davidrohr is it OK for you to merge this as it is now? As mentioned above there is no more backpressure anymore now that the output is sent for every TF.

martenole:
Hi @davidrohr is it OK for you to merge this as it is now? As mentioned above there is no more backpressure anymore now that the output is sent for every TF.

prodtests/full-system-test/calib-workflow.sh
```diff
@@ -8,6 +8,7 @@ fi
 if [[ $BEAMTYPE != "cosmic" ]]; then
   has_detector_calib TPC && has_detectors TPC ITS TRD TOF && add_W o2-tpc-scdcalib-interpolation-workflow "$DISABLE_ROOT_OUTPUT --disable-root-input --pipeline $(get_N tpc-track-interpolation TPC REST)" "$ITSMFT_FILES"
   has_detector_calib ITS && has_detectors ITS && has_detectors_reco ITS && has_detector_matching PRIMVTX && [[ ! -z "$VERTEXING_SOURCES" ]] && add_W o2-calibration-mean-vertex-calibration-workflow
+  has_detector_calib TRD && has_detector ITS TPC TRD && add_W o2-calibration-trd-vdrift-exb
```
chiarazampolli, line 11: Here it should be `has_detectors`, plural, no? Anyway, see my other comment https://github.com/AliceO2Group/AliceO2/pull/8668#issuecomment-1119614489

chiarazampolli:
Hello, 

Please see:

https://github.com/AliceO2Group/O2DPG/pull/375
https://github.com/AliceO2Group/AliceO2/pull/8736

Chiara

prodtests/full-system-test/calib-workflow.sh
martenole: Yes, you are right. Thanks! I see you have already corrected that in https://github.com/AliceO2Group/O2DPG/pull/375

chiarazampolli: Yes, I did, but of course the other PR is a complete change with respect to yours, and your calibration won't run for now in the FST. Is it ok?

martenole: For me in principle yes. If I understand correctly with your changes none of the aggregators would be tested in the FST per default. I am wondering if we should not test them in the CI or then at least enable them when we install a new O2 version at P2 and let the FST run for a longer time to check the stability. But here I let @davidrohr comment (probably better directly in #8736)

PR #8672, opened 2022-04-28
changes for macOS

pillot:
@davidrohr I have a problem using this script on macOS because the string comparison of the `wc` command result does not work for me (probably because of the formatting of the command output). I propose these little modifications, which I suppose will work on all platforms. Do you think it's ok ?

I also conditioned the test of StfBuilder existence to the "DD" mode usage, since I suppose it is needed only in this mode.


pillot:
If there is no objection, could this be merged please?

PR #8673, opened 2022-04-28
TRD bug fix for mac m1

bazinski:
This fixes a couple of things.
Handles the case of processHBF returning false (should never happen)
Correctly checks the bounds of the incoming rdh payload to ensure it fits in the allocated buffer for the half cru in question.

@chiarazampolli it appears to fix grid problem, I could not replicate grid problem with nightly and this executable and library running on top of the nightly last night.

@pzhristov Shout if this fixes your problems. I still running it through my series of timeframes i use for testing.

### DIFF
--- a/Detectors/TRD/reconstruction/src/CruRawReader.cxx
+++ b/Detectors/TRD/reconstruction/src/CruRawReader.cxx
@@ -78,9 +78,9 @@ void CruRawReader::OutputHalfCruRawData()
     LOG(info) << "link " << z << " length : " << mCurrentHalfCRULinkLengths[z] << " (256bit rows)";
   }
   int linkcount = 0;
-  uint64_t linkzsum = 0;
+  uint32_t linkzsum = 0;
   int bufferoffset = 0;
-  uint64_t totalhalfcrulength = std::accumulate(mCurrentHalfCRULinkLengths.begin(),
+  uint32_t totalhalfcrulength = std::accumulate(mCurrentHalfCRULinkLengths.begin(),
                                                 mCurrentHalfCRULinkLengths.end(),
                                                 decltype(mCurrentHalfCRULinkLengths)::value_type(0));
   totalhalfcrulength *= 8; //convert from 256 bits to 32 bits.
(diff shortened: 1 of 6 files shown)


bazinski:
This is currently rebuilding on linux(intel) and mac(m1), passed before, i assume still will. The problem of the singular tf i have with a config in it, remains, but I suspect corruption in the tf, comparing to dev in a 3rd build. Answers to that will come before the ci finishes with this.

bazinski:
@martenole regarding your query, the actual fix for m1 is line 196(new) vs 160(old) of CruRawReader.cxx, together with the unit64_t move to uint32_t, together with correctly handling the return case of false from processHBFs.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -114,26 +114,45 @@ void CruRawReader::OutputHalfCruRawData()
...
+
 bool CruRawReader::processHBFs(int datasizealreadyread, bool verbose)
 {
   if (mHeaderVerbose) {
-    LOG(info) << "PROCESS HBF starting at " << std::hex << (void*)mDataPointer;
+    LOG(info) << "PROCESS HBF starting at " << std::hex << (void*)mDataPointer << " already read in : " << datasizealreadyread;
   }
   mDataRDH = reinterpret_cast<const o2::header::RDHAny*>(mDataPointer);
-  mOpenRDH = reinterpret_cast<o2::header::RDHAny*>((char*)mDataPointer);
+  mOpenRDH = reinterpret_cast<const o2::header::RDHAny*>((const char*)mDataPointer);
```
martenole, line 139: Ok, as I said I don't really understand most of this pointer casting, like here the `reinterpret_cast<>` with an intermediate cast to `char*`. And also the rest here is very hard to understand for me. But if you have tested this and it fixes an actual issue then it is fine with me.

martenole (approved):

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: ;-) that is actually not used, a legacy from testing early on and how I was intending to go through the rdhs comprising the relevant data. I ended up simply combining the payload parts for each rdh into a singular cru payload and working with that.
The change in this pr is actually accidental. New reader is way way more clear as to what is going on, and faster.

pzhristov (approved):
I confirm that these changes fix the issue on Mac M1. I propose to merge them.

PR #8686, opened 2022-04-29
Misc fixes3

ktf:


ktf:
Already tested. Only change was some clang-format issue.

PR #8688, opened 2022-04-30
Applied walker to aggregator inputs + adapted dcs output to needs

iravasen:
Changes:

- Added walker over subspecs to avoid race condition in case of multiple its-calibrators
- Added confDB chipID to DCS CCDB output to speed-up confDB operations
- Added confDB ID to DCS CCDB metadata
- In Dig/Ana scan, dump every single bad pix/dcol by repeating chip location. This facilitate confDB operations. 
- Added output for QC to know when chip scans are completed (userfriendly plots)
- Removed cut at 48 zeros at the EOR. Now ask 50 zeros (i.e. all of them).


iravasen:
Hi @shahor02, I consider this completed on my side. Let me know if you have any suggestion for improvement. I would like to point the attention on a new file (ChipDBmap.h) which is a big map that allow us to convert a chip into a chipID which is relevant for confDB operations (this motivates the +24k lines of this PR). I don't know if this is the best way to proceed within the O2 framework.

Detectors/ITSMFT/ITS/workflow/src/ThresholdAggregatorSpec.cxx
```diff
@@ -45,31 +45,60 @@ void ITSThresholdAggregator::init(InitContext& ic)
...
+      // Merge all strings coming from several sources (EPN)
+      std::copy(tunString.begin(), tunString.end(), std::back_inserter(tuningMerge));
+    }
+    if (DataRefUtils::match(inputRef, "chipdonestring")) {
+      // Read strings with list of completed chips
+      const auto chipDoneString = pc.inputs().get<gsl::span<char>>(inputRef);
+      // Merge all strings coming from several sources (EPN)
+      std::copy(chipDoneString.begin(), chipDoneString.end(), std::back_inserter(chipDoneMerge));
+    }
+  } // end for inputRef
```
shahor02: You don't need to check other `DataRefUtils::match` cases if the inputRef already matched to something, so you can do
```
    if (DataRefUtils::match(inputRef, "runtype")) {
      if (mRunType == -1) {
        mRunType = pc.inputs().get<short int>(inputRef);
        updateRunID(PC);
        updateLHCPeriod(PC); 
      }
    } else if (...
```
But in fact, you don't even need to loop over all subspecs except for `tunestring` and `chipdonestring`: it is much faster to do
```
for (auto const& inputRef : InputRecordWalker(pc.inputs(), {{"check", ConcreteDataTypeMatcher{"ITS", "RUNT"}}})) {
   mRunType = pc.inputs().get<short int>(inputRef);
   updateRunID(pc);
   updateLHCPeriod(PC);
   break;
}
for (auto const& inputRef : InputRecordWalker(pc.inputs(), {{"check", ConcreteDataTypeMatcher{"ITS", "SCANT"}}})) {
   mScanType = pc.inputs().get<char>(inputRef);
   break;
}
for (auto const& inputRef : InputRecordWalker(pc.inputs(), {{"check", ConcreteDataTypeMatcher{"ITS", "FITT"}}})) {
   mFitType = pc.inputs().get<char>(inputRef);
   break;
}
for (auto const& inputRef : InputRecordWalker(pc.inputs(), {{"check", ConcreteDataTypeMatcher{"ITS", "CONFDBV"}}})) {
   mDBversion = pc.inputs().get<short int>(inputRef);
   break;
}
for (auto const& inputRef : InputRecordWalker(pc.inputs(), {{"check", ConcreteDataTypeMatcher{"ITS", "TSTR"}}})) {
   // Read strings with tuning info
   const auto tunString = pc.inputs().get<gsl::span<char>>(inputRef);
   // Merge all strings coming from several sources (EPN)
   std::copy(tunString.begin(), tunString.end(), std::back_inserter(tuningMerge));
}
for (auto const& inputRef : InputRecordWalker(pc.inputs(), {{"check", ConcreteDataTypeMatcher{"ITS", "QCSTR"}}})) {
   // Read strings with list of completed chips
   const auto chipDoneString = pc.inputs().get<gsl::span<char>>(inputRef);
   // Merge all strings coming from several sources (EPN)
   std::copy(chipDoneString.begin(), chipDoneString.end(), std::back_inserter(chipDoneMerge));
}
```

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -909,52 +928,65 @@ void ITSThresholdCalibrator::addDatabaseEntry(
...
+    o2::dcs::addConfigItem(this->mChipDoneQc, "Stave", std::string(stave));
+    o2::dcs::addConfigItem(this->mChipDoneQc, "Hs_pos", std::to_string(ssta));
+    o2::dcs::addConfigItem(this->mChipDoneQc, "Hic_Pos", std::to_string(mod));
+    o2::dcs::addConfigItem(this->mChipDoneQc, "ChipID", std::to_string(chipInMod));
+    return;
+  }
+
+  // Get ConfDB id for the chip chipID
+  std::string hs = ssta == 0 ? "L" : "U";
+  int confDBid = mConfDBmap.getConfigDBid(lay < 3 ? Form("L%d_%02d_M%d_C%d", lay, sta, mod, chipInMod) : Form("L%d_%02d%s_M%d_C%d", lay, sta, hs.c_str(), mod, chipInMod));
```
shahor02: Note that this format differs from that used in the `o2::itsmft::ChipMappingITS::getChipNameHW`:
I use e.g.
```
root [0] o2::itsmft::ChipMappingITS mp;
root [1] mp.getChipNameHW(1234);
(std::string) "L3_07L_M2_C04"
```
while you would get `"L3_07L_M2_C4"`.
Which one is correct?

The implementation of the mapping in mConfDBmap is indeed too heavy. What is the meaning of this `confDBid`?
Can't it be implemented algorithmically?

shahor02: In fact, the O2 `chipID` is a contiguous number from 0 to 24119, to map it on some integer confDBid, you don't need to decompose it to ` lay, sta, mod, chipInMod`, then create a string then to query a map, you can simply define a 
`std::array<int, 24120> confDBidMap` with the values you need and query it simply as
`int confDBid = confDBidMap[chipID];`
Still, instead of putting into the code such a huge const array (which I guess may change one day), better to have CCDB entry with such an array and update it when needed.

Detectors/ITSMFT/ITS/workflow/src/ThresholdAggregatorSpec.cxx
iravasen: Done

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
iravasen: @shahor02 I would stick to your format for the chip location. Anyway now I tried the workflow with the CCDB retrieve of the object containing the map of confDB id. It works (the vector<int> is readout properly) but when running the workflow I see a huge number of messages like:
```
[40890:its-calibrator_0]: [19:13:30][ALARM] Backpressure on channel from_its-stf-decoder_to_its-calibrator_0[0]. Waiting.
[40890:its-calibrator_0]: [19:13:30][INFO] Back to normal on channel from_its-stf-decoder_to_its-calibrator_0[0].
```
At some point the workflow freezes so I cannot fully test locally. The problem is due to the ccdb retrieve of the object because removing it, all the messages above disappear.

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -909,52 +940,64 @@ void ITSThresholdCalibrator::addDatabaseEntry(
...
+  if (isQC) {
+    o2::dcs::addConfigItem(this->mChipDoneQc, "Stave", std::string(stave));
+    o2::dcs::addConfigItem(this->mChipDoneQc, "Hs_pos", std::to_string(ssta));
+    o2::dcs::addConfigItem(this->mChipDoneQc, "Hic_Pos", std::to_string(mod));
+    o2::dcs::addConfigItem(this->mChipDoneQc, "ChipID", std::to_string(chipInMod));
+    return;
+  }
+
+  // Get ConfDB id for the chip chipID
+  int confDBid = mConfDBmap->at(chipID);
```
shahor02: use `(*mConfDBmap)[chipID]`, `->at()` has an overhead due to the range check.

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
shahor02: I guess you are doing local test, right? What is the exact workflow you are running?

iravasen: yes, this is the workflow:
```
o2-raw-tf-reader-workflow --input-data 514354.dat --max-tf -5 --onlyDet ITS --shm-segment-size 40000000000 | \
o2-itsmft-stf-decoder-workflow  --ignore-noise-map --nthreads 1 --pipeline its-stf-decoder:1 --no-clusters --no-cluster-patterns --enable-calib-data --digits --shm-segment-size 40000000000 | \
o2-its-threshold-calib-workflow -b  --meta-output-dir "./" --nthreads 1 --fittype derivative | \
o2-its-threshold-aggregator-workflow -b --ccdb-url "http://ccdb-test.cern.ch:8080"
```
where the 514354.dat contains 3 local TF files.

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
iravasen: done

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
shahor02: OK, the ITSThresholdCalibrator does not send on every TF while the DPL fetcher sends at every TF and fills the pipeline. 
1st of all, you should have the output types of getITSThresholdCalibratorSpec not 
`Lifetime::Timeframe` but `Lifetime::Sporadic`.
Could you try if this changes something? It might be that one more modification will be needed, but I will need to clarify it with Giulio, yesterday I had a similar situation and it did not work.

iravasen: Ah ok I see, unfortunately even with `Lifetime::Sporadic`, the issue is still there.

shahor02: I see, in principle, https://github.com/shahor02/AliceO2/blob/9d0663a7dc1a34eead2973f6bff1fe4e8ec2a5a0/Detectors/MUON/MCH/Workflow/src/cru-page-reader-workflow.cxx#L493-L508 (with `return broadCastLatest();` issued at every TF, even if the run() does not send output) is supposed to fix such situations, but it did not work and @ktf is checking why.

Note that most probably, in the real EPN workflow, where the aggregator is running on a separate host and does not see the clock tick of every TF, you will not have this 
situation, since the CCDB fetcher will get the object only when the input proxy gets something. But to not rely on that, perhaps at the moment you better comment out ccbParamSpec stuff and load the object via ccdbmanager...

iravasen: ok so you would implement the ccdbmanager solution also for P2 of just for local tests? Yes at P2 I do not expect that as well because we don't see issues with the NOISEMAP in the STFDecoderSpec which uses the ccdbparamspec.

shahor02: STFDecoderSpec is different, it receives both data and CCDB objects  for every TF, so there is no problem. The problem appears when the workflow does not receive data for the TF for which it already got ccdb object: then the workflow gets stuck. 
This is what happening with the aggregator in your local test (because its receiving of the CCDB object is triggered by the 1st device in the workflow, i.e. the reader). This should not happen on the real epn, where the CCDB query will be triggered by the input-proxy only when it receives data. But, I guess you don't want to use different CCDB access mechanisms in the local tests and on the EPN, so I would suggest at the moment to switch to the CCDBManager, unless you can wait a few days: this is a general problem for all aggregators, we (@ktf) should find a solution for it.

ktf: If data is sporadic, then it should be marked as `Lifetime::Sporadic`. That will guarantee that spurious timeframes are ignored and CCDB data is dropped not matched by detector data will be dropped. If this is not the case, it's a bug and we (I) need to fix it.

shahor02: @ktf Ivan has already tried with `Lifetime::Sporadic` (see discussion above), but it does not work. I guess because the aggregator has both this sporadic input and normal one from the CCDBFetcher.

ktf: In principle it should have excluded the Lifetime::Condition. I am checking this.

iravasen: @ktf @shahor02 thanks for the explanation. I can wait till Sunday. The idea would be to have this PR for next week at P2.

iravasen: in addition I'm doing the `pc.inputs(..)` of the CCDB object only when the first TF arrives (when `mRunType==-1`). Checking the STFdecoder more in detail I see that this is done for every TF in the run() method. I guess I should do the same. But then the issue with aggregator remains, as you said.

iravasen:
@ktf @shahor02 I applied the temporary solution with the ccdb manager for now since I would like to merge this PR by tonight. After successful test, for me this is ready to go.

shahor02 (approved):

shahor02:
@iravasen @ktf The checks are passed but looks like there are again git problem: all PRs report `This branch has conflicts that must be resolved` and one cannot force-merge them.

iravasen:
@shahor02 thanks, ok I can wait for the git issue resolution. We will need this PR at P2 towards the end of the week so I hope you can do an upgrade of the EPNs during the week so to have this code at P2. Thanks a lot.

ktf:
For the record, I merged this by hand.

PR #8690, opened 2022-04-30
Fix a bunch of pedantic warnings

mconcas:
I was trying to fix some warnings in ITS for the --pedantic mode, I also tried to follow the ones related to the dependencies (triggered by re-compilation). The process diverged quite quickly, so I truncated it to some files.

I plan to open other PRs if this kind of work is well accepted by other CODEOWNERS.
Please have a look to the changes if you have time.


DataFormats/Reconstruction/include/ReconstructionDataFormats/Vertex.h
```diff
@@ -143,8 +143,8 @@ class Vertex : public VertexBase
   GPUd() void setTimeStamp(const Stamp& v) { mTimeStamp = v; }
 
  protected:
-  float mChi2 = 0;           ///< chi2 or quality of tracks to vertex attachment
   ushort mNContributors = 0; ///< N contributors
+  float mChi2 = 0;           ///< chi2 or quality of tracks to vertex attachment
```
shahor02: Not sure we need to respect this stupid `-Wreorder` warning, I think since even c+11 the ordering is not required.
In any case, please don't change the class layout (moreovere from 4B,2B,2B to 2B,4B,2B) but swap the order of initialization in the https://github.com/AliceO2Group/AliceO2/blob/cdb68743ea3d7122f7eb25d286a9032d0e34e6ae/DataFormats/Reconstruction/include/ReconstructionDataFormats/Vertex.h#L125

Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
```diff
@@ -98,7 +98,7 @@ bool MatchITSTPCQC::init()
   Double_t xlogminPt = TMath::Log10(xminPt);
   Double_t xlogmaxPt = TMath::Log10(xmaxPt);
   Double_t dlogxPt = (xlogmaxPt - xlogminPt) / nbinsPt;
-  Double_t yPt = (xlogmaxPt - xlogminPt) / nbinsPt;
+  // Double_t yPt = (xlogmaxPt - xlogminPt) / nbinsPt;
```
shahor02: better to remove unused stuff

Detectors/ITSMFT/ITS/calibration/include/ITSCalibration/NoiseCalibratorSpec.h
```diff
@@ -55,11 +55,11 @@ class NoiseCalibratorSpec : public Task
   void sendOutput(DataAllocator& output);
   void updateTimeDependentParams(ProcessingContext& pc);
   std::unique_ptr<CALIBRATOR> mCalibrator = nullptr;
+  bool mUseClusters = false;
   std::shared_ptr<o2::base::GRPGeomRequest> mCCDBRequest;
   size_t mDataSizeStat = 0;
   size_t mNClustersProc = 0;
   int mValidityDays = 3;
-  bool mUseClusters = false;
```
shahor02: same comment on ordering warnings: change c-tor rather than the layout

Detectors/ITSMFT/ITS/simulation/src/Detector.cxx
```diff
@@ -1220,83 +1220,4 @@ Hit* Detector::addHit(int trackID, int detID, const TVector3& startPos, const TV
   return &(mHits->back());
 }
 
-void Detector::Print(std::ostream* os) const
```
shahor02, line 1231: It's ok to suppress these methods, but I think you did not remove them from the header

shahor02 (changes requested):
Thanks for taking care of this, see some comments below.

DataFormats/Reconstruction/include/ReconstructionDataFormats/Vertex.h
mconcas: Ups, I completely forgot about padding! sure I'll do that.

Detectors/GlobalTracking/src/MatchITSTPCQC.cxx
mconcas: Done and removed all similar cases

Detectors/ITSMFT/ITS/simulation/src/Detector.cxx
mconcas: done

mconcas:
As a general comment:

I found some spots where I think it's not possible to easily fix warnings, I'm compiling the list hereafter as soon as I encounter something I cannot fix:

1. Geometry objects construction that creates entries in static `gGeoManager`. This is quite subtle as I think it creates errors at runtime but not at compile time. e.g.: https://github.com/AliceO2Group/AliceO2/blob/90c2cf415b903c22b430925c68d1a31ce402ff40/Detectors/ITSMFT/ITS/simulation/src/V3Layer.cxx#L2462
2. Unused variables in `assert()` statements with no assert flag set. e.g. https://github.com/AliceO2Group/AliceO2/blob/f58d3f693fd4112bc822d46c9a4a50f41119e3e4/Framework/Core/include/Framework/MessageContext.h#L541
3. Hits added to static stack (gStack?) e.g.: https://github.com/AliceO2Group/AliceO2/blob/f58d3f693fd4112bc822d46c9a4a50f41119e3e4/Detectors/ITSMFT/ITS/simulation/src/Detector.cxx#L354

davidrohr:
@mconcas : could you rebase this?

mconcas:
@davidrohr I figured out why the test was failing (it was ITS-related one, so somehow it was pertinent). See point 3 in previous [comment](https://github.com/AliceO2Group/AliceO2/pull/8690#issuecomment-1114014620). If this passes the tests can be merged.
Next PRs will be more restricted to ITS code, as discussed. Stand still the problems found so far.

Framework/Core/include/Framework/MessageContext.h
```diff
@@ -503,7 +503,7 @@ class MessageContext
             outputsPerChannel[mProxy.getOutputChannelIndex(message->route()).value].AddPart(std::move(part));
           }
         }
-        for (int ci = 0; ci < mProxy.getNumOutputChannels(); ++ci) {
+        for (int ci = 0; ci < (int)mProxy.getNumOutputChannels(); ++ci) {
```
ktf:
```suggestion
        for (size_t ci = 0; ci < mProxy.getNumOutputChannels(); ++ci) {
```

ktf (changes requested):
I think the correct way to handle this kind of warnings is to try to fix the original type, not to force the cast (which might go wrong).

Framework/Core/include/Framework/MessageContext.h
mconcas: I generally agree with you. I opted for a cast to avoid clashes with the `ChannelIndex` construction, so to limit the propagation of the change.

https://github.com/AliceO2Group/AliceO2/blob/a67bba64517e95f125d2c582d8056a0657485c98/Framework/Core/include/Framework/MessageContext.h#L511

On the other hand I understand this is not code I developed, so I see two options here:
 1)  I follow the chain of changes to shut the warnings down to a correct fix that then you can check and approve
 1) I revert this fix and let owners taking care of warnings at their convenience

Both are fine with me, I see this is an annoying task, I wanted to contribute to the effort.

Detectors/GlobalTracking/src/MatchTPCITS.cxx
```diff
@@ -1825,14 +1822,13 @@ int MatchTPCITS::followABSeed(const o2::track::TrackParCov& seed, const ITSChipC
...
     const auto& chipC = lad.chips[chipIDguess];
-    bool res = mFieldON ? chipC.xyEdges.circleCrossParam(trcCircle, t) : chipC.xyEdges.lineCrossParam(trcLinPar, t);
     chipC.xyEdges.eval(t, xCross, yCross);
     float dx = xCross - xCurr, dy = yCross - yCurr, dst2 = dx * dx + dy * dy, dst = sqrtf(dst2);
     // Z-step sign depends on radius decreasing or increasing during the propagation
     float zCross = seedC.getZ() + seedC.getTgl() * (dst2 < 2 * (dx * xCurr + dy * yCurr) ? dst : -dst);
 
-    for (int ich = -1; ich < 2; ich++) {
-      int chipID = chipIDguess + ich;
+    for (size_t ich = -1; ich < 2; ich++) {
```
davidrohr: This one is probably wrong: `size_t ich = -1`

davidrohr (changes requested):

Detectors/GlobalTracking/src/MatchTPCITS.cxx
mconcas: Sure, my bad.

PR #8691, opened 2022-04-30
CPV: fix channel mapping

sevdokim:
Hello! This is a fix in CPV channel mapping suggested by Artem Shangaraev.


peressounko:
Hi @sevdokim, I see that mapping was actually changed, not only beautified. Was it checked with real data?

sevdokim:
Hi @peressounko 
yes, it was proposed and checked by Artem. Also I checked it on pedestal data: wide pedestal sigma edge effect is now really at the edge of cpv modules. Also I checked it with MC for self-consistence: it is OK.

peressounko (approved):

PR #8694, opened 2022-05-01
Temporary restore CTF encoding buffers defaults extra margins

shahor02:


shahor02:
trivial and tested locally, merging

PR #8696, opened 2022-05-02
Added feature: local density scaling (per module).

zovarga:
- SimParams.h: Adding the local density factor.
- MaterialManager.h: Defining the map between the module names and the density values. Modifying the constructor to create the map. Error handling of adding negative values and non-existing module names as input. Removing the old `setDensityScalingFactor()` and `setDensityScalingFactor() `as they are not needed anymore.
- MaterialManager.cxx: Implementing the creation of `MaterialManager::createDensityMap()`. Modifying `MaterialManager::Material()`and `MaterialManager::Mixture()` to use the given density values.
- o2sim.C: Removing the usage of `setDensityScalingFactor()` as it doesn't exist anymore.

### DIFF
--- a/Common/SimConfig/include/SimConfig/SimParams.h
+++ b/Common/SimConfig/include/SimConfig/SimParams.h
@@ -41,6 +41,7 @@ struct SimCutParams : public o2::conf::ConfigurableParamHelper<SimCutParams> {
 // parameter influencing material manager
 struct SimMaterialParams : public o2::conf::ConfigurableParamHelper<SimMaterialParams> {
   float globalDensityFactor = 1.f;
+  std::string localDensityFactor;
 
   O2ParamDef(SimMaterialParams, "SimMaterialParams");
 };
--- a/Common/SimConfig/include/SimConfig/SimParams.h
+++ b/Common/SimConfig/include/SimConfig/SimParams.h
@@ -40,8 +40,9 @@ struct SimCutParams : public o2::conf::ConfigurableParamHelper<SimCutParams> {
 
 // parameter influencing material manager
 struct SimMaterialParams : public o2::conf::ConfigurableParamHelper<SimMaterialParams> {
+  // Local density value takes precedence over global density value, i.e. local values overwrite the global value.
   float globalDensityFactor = 1.f;
-  std::string localDensityFactor;
+  std::string localDensityFactor; // Expected format: "SimMaterialParams.localDensityFactor=<mod1>:<value1>,<mod2>:<value2>,..."
 
   O2ParamDef(SimMaterialParams, "SimMaterialParams");
 };
--- a/Detectors/Base/include/DetectorsBase/MaterialManager.h
+++ b/Detectors/Base/include/DetectorsBase/MaterialManager.h
@@ -204,17 +204,16 @@ class MaterialManager
   // print out all registered media
   void printMedia() const;
 
-  /// set the density scaling factor
-  void setDensityScalingFactor(float f) { mDensityFactor = f; }
-  /// set the density scaling factor
-  float getDensityScalingFactor() const { return mDensityFactor; }
-
   /// print all tracking media inside a logical volume (specified by name)
   /// and all of its daughters
   static void printContainingMedia(std::string const& volumename);
 
  private:
-  MaterialManager() = default;
+  std::unordered_map<std::string, float> mDensityMap;
+
+  void createDensityMap();
+
+  MaterialManager() { createDensityMap(); }
 
   // Hide details by providing these private methods so it cannot happen that special settings
   // are applied as default settings by accident using a boolean flag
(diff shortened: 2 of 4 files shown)


Detectors/Base/src/MaterialManager.cxx
```diff
@@ -91,21 +93,66 @@ const std::unordered_map<ECut, const char*> MaterialManager::mCutIDToName = {
...
+    inputDensityValues.push_back(token.substr(pos + 1));
+  }
+  std::vector<std::string> allModuleNames = {
+    "ABSO", "CAVE", "COMP", "DIPO", "FRAME", "HALL", "MAG", "PIPE",
+    "SHIL", "CPV", "EMC", "FDD", "FT0", "FV0", "HMP", "ITS",
+    "MCH", "MFT", "MID", "PHS", "TOF", "TPC", "TRD", "ZDC",
+    "ALPIDE", "IT3", "TRK", "FT3", "A3IP"};
+  if (o2::conf::SimMaterialParams::Instance().globalDensityFactor < 0) {
+    LOG(fatal) << "Negative value "
+               << o2::conf::SimMaterialParams::Instance().globalDensityFactor
```
benedikt-voelkel: I guess you could put 
`auto& globalDensityfactor = o2::conf::SimMaterialParams::Instance().globalDensityFactor;`
somewhere at the beginning of this function because you reuse it a few times.
In addition you should check its validity and `LOG(fatal)` as early as possible in this function before you do anything else

zovarga: Thank you very much Benedikt! I implemented your suggestions.

Common/SimConfig/include/SimConfig/SimParams.h
```diff
@@ -41,6 +41,7 @@ struct SimCutParams : public o2::conf::ConfigurableParamHelper<SimCutParams> {
 // parameter influencing material manager
 struct SimMaterialParams : public o2::conf::ConfigurableParamHelper<SimMaterialParams> {
   float globalDensityFactor = 1.f;
+  std::string localDensityFactor;
```
sawenzel: could you please add a comment to this data member describing the format expected to be given by the user? Also, we should clarify if localDensityFactor takes precedence over globalDensityFactor (or if they act in a multiplicative way).

Detectors/Base/src/MaterialManager.cxx
```diff
@@ -91,21 +93,66 @@ const std::unordered_map<ECut, const char*> MaterialManager::mCutIDToName = {
...
+  if (globalDensityFactor < 0) {
+    LOG(fatal) << "Negative value "
+               << globalDensityFactor
+               << " found for global material density!\n";
+  }
+  std::vector<std::string> allModuleNames = {
+    "ABSO", "CAVE", "COMP", "DIPO", "FRAME", "HALL", "MAG", "PIPE",
+    "SHIL", "CPV", "EMC", "FDD", "FT0", "FV0", "HMP", "ITS",
+    "MCH", "MFT", "MID", "PHS", "TOF", "TPC", "TRD", "ZDC",
+    "ALPIDE", "IT3", "TRK", "FT3", "A3IP"};
```
sawenzel: do we really need this static list of names? I think we better should avoid this because otherwise we might run into troubles when someone adds more detectors.
It should be enough to create the map dynamically from the given string. One can then create another helper and return `globalDensityFactor` whenever something is not in the map, or in the other case the content of the map.

Detectors/Base/include/DetectorsBase/MaterialManager.h
```diff
@@ -204,17 +204,16 @@ class MaterialManager
...
   /// and all of its daughters
   static void printContainingMedia(std::string const& volumename);
 
  private:
-  MaterialManager() = default;
+  std::unordered_map<std::string, float> mDensityMap;
+
+  void createDensityMap();
+
+  MaterialManager() { createDensityMap(); }
```
sawenzel: not sure this is a good idea. Initialization might happen during construction of the singleton MaterialManager, at which moment the ConfigParams might not be fully initialized.

It may be better to init this *late*. For instance we could have a function 
```
float getDensity(std::string const& modname) {
  if (!initialized) {
     initDensityMap();
  }
  if (mDensityMap.find(modname) != mDensityMap.end()) {
    return mDensityMap[modname]
  }
  return mGlobalDensityFactor;
}
```

Common/SimConfig/include/SimConfig/SimParams.h
zovarga: Hi Sandro, thank you very much for the review! I added the comments to help the user.

Detectors/Base/src/MaterialManager.cxx
zovarga: I implemented your suggestion.

Detectors/Base/include/DetectorsBase/MaterialManager.h
zovarga: The suggested function has been implemented in the code.

benedikt-voelkel:
Failing CIs so far unrelated to changes in this PR.

sawenzel (approved):

PR #8703, opened 2022-05-02
TPC: fix DCS DP type and reset, throttle error messages

wiechula:


wiechula:
@davidrohr , can this be merged?

PR #8704, opened 2022-05-02
MCH: moved time calibration to digits filter

aferrero2707:
The time calibration is controlled by the `MCHDigitFilterParam.timeOffset` configuration parameter, that specifies the global offset (in BC units) applied to the time stamp of digits and ROFs.

### DIFF
--- a/Detectors/MUON/MCH/DigitFiltering/src/DigitFilteringSpec.cxx
+++ b/Detectors/MUON/MCH/DigitFiltering/src/DigitFilteringSpec.cxx
@@ -52,6 +52,21 @@ class DigitFilteringTask
     // resolution will suffer.
     // That's why we only apply the "reject background" filter, which
     // is a loose background cut that does not penalize the signal
+
+    mTimeCalib = DigitFilterParam::Instance().timeOffset;
+  }
+
+  void shiftDigitsTime(gsl::span<ROFRecord> rofs, gsl::span<Digit> digits)
+  {
+    for (auto i = 0; i < rofs.size(); i++) {
+      ROFRecord& rof = rofs[i];
+      rof.getBCData() += mTimeCalib;
+    }
+
+    for (auto i = 0; i < digits.size(); i++) {
+      Digit& d = digits[i];
+      d.setTime(d.getTime() + mTimeCalib);
+    }
   }
 
   void run(ProcessingContext& pc)
@@ -116,6 +131,10 @@ class DigitFilteringTask
          oDigits.size(), iDigits.size(),
          labelMsg);
 
+    if (mTimeCalib != 0) {
+      shiftDigitsTime(oRofs, oDigits);
+    }
+
     if (abort) {
       LOGP(error, "Sanity check failed");
     }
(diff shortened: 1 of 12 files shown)


aphecetche:
@aferrero2707, if you agree, I would propose that : 

1 - we merge https://github.com/AliceO2Group/AliceO2/pull/8737 to get the `DigitFilterParam` in place
2 - we introduce the digit noise filtering (PR to be opened after 1 is merged, see https://github.com/AliceO2Group/AliceO2/commit/65b0fb905883fbc501cab7e2d763996030145149) 
3 - we rebase this one on (1 or 2) to benefit from the DigitFilterParam

or equivalently, 1 - 3 -2 (2 and 3 being different filtering features)

what do you think ?

aphecetche:
@aferrero2707, if you agree, I would propose that : 

1 - we merge https://github.com/AliceO2Group/AliceO2/pull/8737 to get the `DigitFilterParam` in place
2 - we introduce the digit noise filtering (PR to be opened after 1 is merged, see https://github.com/AliceO2Group/AliceO2/commit/65b0fb905883fbc501cab7e2d763996030145149) 
3 - we rebase this one on (1 or 2) to benefit from the DigitFilterParam

or equivalently, 1 - 3 -2 (2 and 3 being different filtering features)

what do you think ?

aferrero2707:
I think 1 and 2 are more urgent than this PR, so we can rebase this one once the other two are merged.

shahor02:
> I think 1 and 2 are more urgent than this PR, so we can rebase this one once the other two are merged.

@aferrero2707 OK, would you check/validate https://github.com/AliceO2Group/AliceO2/pull/8737 ? I'll merge it once fullCI is passed

aferrero2707:
@aphecetche all checks are green, could you please have a look if the modifications are OK for you? meanwhile I will commit a small update of the documentation.
We should probably also consider a unit test dedicated to the digits filter, but I would propose to make a dedicated PR for that.

Detectors/MUON/MCH/DigitFiltering/src/DigitFilteringSpec.cxx
```diff
@@ -125,6 +144,8 @@ class DigitFilteringTask
   bool mSanityCheck;
   bool mUseMC;
   DigitFilter mIsGoodDigit;
+  std::string mTimeCalibSource;
```
pillot: this data member is not used

pillot (commented):
Hi @aferrero2707 ,
Appart from the little change below, shouldn't the configurable parameter CoDecParam::Instance().sampaBcOffset be removed?
In which case it also has to be removed from the few places where it is still used (e.g. testSampaBunchCrossingCounter.cxx, codec-closure-test.sh and testClosureCoDecDigit.cxx).

aferrero2707:
@pillot you are absolutely right! It seems I stopped in the middle of the code cleanup to remove the `sampaBcOffset`.

Detectors/MUON/MCH/DigitFiltering/src/DigitFilteringSpec.cxx
aferrero2707: Removed.

aphecetche (approved):

PR #8706, opened 2022-05-03
Fix in the creation of subdirectory

pzhristov:


Detectors/CTP/simulation/src/digi2raw.cxx
```diff
@@ -109,7 +109,7 @@ void digi2raw(const std::string& inpName, const std::string& outDir, int verbosi
   }
   // if needed, create output directory
   if (!std::filesystem::exists(outDirName)) {
-    if (!std::filesystem::create_directories(outDirName)) {
+    if (std::filesystem::create_directories(outDirName)) {
```
shahor02, line 112: @pzhristov why is that? This breaks the fullCI. If create_directories returns true this means that the directory was created. Reverting this.

PR #8708, opened 2022-05-03
o2-sim: Automatic timestamp setup from run number

sawenzel:
When a run number is given, o2-sim can now deduce a
consistent timestamps automatically, and propagate it
to digitization via GRP (and later via HBFUtils).

This is fixing an issue where arbitrary "now" timestamps
where used in un-unchored MC for which no CCDB objects were available.

### DIFF
--- a/prodtests/full_system_test.sh
+++ b/prodtests/full_system_test.sh
@@ -46,6 +46,7 @@ NHBPERTF=${NHBPERTF:-128}
 RUNFIRSTORBIT=${RUNFIRSTORBIT:-0}
 FIRSTSAMPLEDORBIT=${FIRSTSAMPLEDORBIT:-0}
 OBLIGATORYSOR=${OBLIGATORYSOR:-false}
+RUNNUMBER=300000 # a default un-anchored run number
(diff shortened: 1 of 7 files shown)


prodtests/full_system_test.sh
```diff
@@ -46,6 +46,7 @@ NHBPERTF=${NHBPERTF:-128}
 RUNFIRSTORBIT=${RUNFIRSTORBIT:-0}
 FIRSTSAMPLEDORBIT=${FIRSTSAMPLEDORBIT:-0}
 OBLIGATORYSOR=${OBLIGATORYSOR:-false}
+RUNNUMBER=300000 # a default un-anchored run number
```
shahor02, line 49: can you add to https://github.com/AliceO2Group/AliceO2/blob/e8ae9ffc0b8559a2001729b4b23511d65936a110/prodtests/full_system_test.sh#L78
`;HBFUtils.runNumber=${RUNNUMBER}`

shahor02 (changes requested):

PR #8712, opened 2022-05-03
MCH: remove custom completion policy from pedestal calibrator

aphecetche:


shahor02 (approved):
Trivial and anyways not tested by the fullCI, merging

PR #8715, opened 2022-05-03
Add missing HBFUtilsInitializer calls + MCH cru-page-reader time/run assignment

shahor02:
@aphecetche when testing your workflow, the ccdb fetcher was failing since the reader was not injecting the TF time. So, I am adding this commit on top of few cases of incomplete HBFUtilsInitializer usage.
As I wrote before, the reader has a problem that it does not send the output for every `run()` call. Therefore I changed its output to `Lifetime::Sporadic`, though still something prevents to use it in the workflow with the CCDB fetcher.

* Minimal changes for cru-page-reader to inject timestamp and runNumber
In case the cru-page-reader should inject to the DataHeader and DataProcessingHeader
the info like runNumber, orbitFirst and creation time, it can be imposed via e.g.
`--configKeyValues HBFUtils.startTime=1651562466196;HBFUtils.runNumber=514769;HBFUtils.orbitFirst=128`
Note that this info will be added as is, i.e. DataProcessingHeader.creation will be always the same as
passed startTime. If needed, one should apply more elaborate logic depending what is read out from the file.
Thie is minimum needed for the DPL CCDB fetcher functionality.

### DIFF
--- a/Detectors/EMCAL/workflow/src/cell-reader-workflow.cxx
+++ b/Detectors/EMCAL/workflow/src/cell-reader-workflow.cxx
@@ -13,6 +13,7 @@
 #include <vector>
 #include "Framework/Variant.h"
 #include "Framework/ConfigParamSpec.h"
+#include "Framework/CallbacksPolicy.h"
 #include "DataFormatsEMCAL/Cell.h"
 #include "DetectorsRaw/HBFUtilsInitializer.h"
 #include "EMCALWorkflow/PublisherSpec.h"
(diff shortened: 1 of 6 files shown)


aphecetche:
@aferrero2707 could you x-check the proposed changes to `mch-cru-page-reader` ?

aferrero2707:
@shahor02 @aphecetche I will need to run some tests locally to confirm that everything is working as expected in the `mch-cru-page-reader` case.

shahor02:
@aphecetche @aferrero2707 shall I merge this? Note that the only unconditional change it does to cru_reader is declaring the output sporadic (since it is). The rest is optional injection of orbit / timestamp / run to the headers.

aferrero2707:
@shahor02 ok for me to merge. I have not been able to test the code so far because all the workflows get stuck at initialization phase on my development machine, but whatever the problem it does not seem to be related to this PR.

shahor02:
ok, thanks, just removed some debug stuff.

PR #8716, opened 2022-05-03
[EMCAL-774] Add EMCAL volumes for the simple ED geometry

mfasDa:


mfasDa:
@pnwkw These would be the EMCAL volumes used for the visualization

shahor02 (approved):
trivial, merging

mfasDa:
For EMCAL we keep the single supermodules as volumes in the simplified geometry. Going to the single tower level would make the visualization too complicated also to monitor for the user. We have to use the single supermodules as for EMCAL we cannot provide a mothervolume as it would have overlap with PHOS.

PR #8718, opened 2022-05-03
Produce in o2sim the triplet of GRPs, update GRPECS in GRPUpdater

shahor02:


Steer/DigitizerWorkflow/src/SimReaderSpec.cxx
```diff
@@ -143,6 +197,10 @@ DataProcessorSpec getSimReaderSpec(SubspecRange range, const std::vector<std::st
 
       mgr.getInteractionSampler().init();
       mgr.getInteractionSampler().print();
+
+      // update local GRPLHCIF with bunch filling schema
+      updateGRPLHCIF(o2::base::NameConf::getGRPLHCIFFileName(simprefixes[0]), mgr.getInteractionSampler().getBunchFilling());
```
sawenzel: what speaks against doing this also in the GRPUpdater device? Might not need the code duplication.

shahor02: GRPUpdater collects inputs from all digitizer devices and updates the GRPECS (and still old monolithic GRP also). `updateGRPLHCIF` just adds the bunchFilling to GRPLHCIF, this does not depend on other detectors / digitizers.
Which code is duplicated (apart from the interprocess semaphor which is already duplicated in a few places, perhaps it should be extracted to separate utility)?

sawenzel: It is true that up until now GRPUpdater updates fields coming from detectors. But I don't see a reason not to use it for other updates as well. We could simple add a connection directly from SimReader to the GRPUpdater?

shahor02: ok, can change it bit later, now on other branch. In any case, I will need https://github.com/AliceO2Group/AliceO2/pull/8708 merged 1st.

sawenzel: @shahor02 : I think we can merge #8708 only after we have default CCDB objects for everything (which might take a while). But I can rework the other PR also after this PR has been merged.

shahor02: Isn't it only imposing default runNumber 300000 in sim_challenge and full_system_test which requires a retrofitted CCDB?
If we would set a the moment some pilot beam run by default (and make it settable via an env.var) it should work, right?
I just want to have correct values set to the GRPs, i.e. the runNumber (via HBFUtils) and run start.

sawenzel: True. We could set a different default run number for sim.

shahor02: @sawenzel moved GRPLHCIF update to GRPUpdater. While doing this, got a question: why do we send tens of copies of the same collision context with different subspecs,  
https://github.com/AliceO2Group/AliceO2/blob/3c5523758f34bf9c50a640348c900fabf79e312a/Steer/DigitizerWorkflow/src/SimReaderSpec.cxx#L209-L216 ?

Will wait for https://github.com/AliceO2Group/AliceO2/pull/8708 to set correct runNumber and time.

Steer/DigitizerWorkflow/src/SimpleDigitizerWorkflow.cxx
```diff
@@ -739,7 +739,7 @@ WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
   }
   // GRP updater: must come after all detectors since requires their list
```
shahor02, line 740: @sawenzel fixing the conflict after rebasing your PR I see that the CTP specs do not follow the same    digitizerSpecs/ writerSpecs pattern, was it intended? https://github.com/AliceO2Group/AliceO2/pull/8718/commits/2175faeaa7968107b31aa1f6eaa713977993337c#diff-b1fedbaacd6c752678b71e5f22d308e35c2c3c8d4b9841013d900ba4faffdc7dR736-R739

PR #8719, opened 2022-05-04
Revert "Fix in the creation of subdirectory"

shahor02:
Reverts AliceO2Group/AliceO2#8706

### DIFF
--- a/Detectors/CTP/simulation/src/digi2raw.cxx
+++ b/Detectors/CTP/simulation/src/digi2raw.cxx
@@ -109,7 +109,7 @@ void digi2raw(const std::string& inpName, const std::string& outDir, int verbosi
   }
   // if needed, create output directory
   if (!std::filesystem::exists(outDirName)) {
-    if (std::filesystem::create_directories(outDirName)) {
+    if (!std::filesystem::create_directories(outDirName)) {
       LOG(fatal) << "could not create output directory " << outDirName;
     } else {
       LOG(info) << "created output directory " << outDirName;


davidrohr:
thx for fixing, was just preparing the same :)

pzhristov:
@shahor02 @davidrohr What is wrong with my fix? The return value on Mac is exactly as in my modification, otherwise it creates the directory, prints error message, exits and you have to restart the test that will work since the directory was created in the previous attempt. Is this what you want?

davidrohr:
Can it be that the return value on mac and on linux are different?

Kind Regards
David Rohr

On 4 May 2022 11:25:08 CEST, Peter Hristov ***@***.***> wrote:
***@***.*** @davidrohr What is wrong with my fix? The return value on Mac is exactly as in my modification, otherwise it creates the directory, prints error message, exits and you have to restart the test that will work since the directory was created in the previous attempt. Is this what you want?

shahor02:
@pzhristov something is wrong with your Mac :) https://en.cppreference.com/w/cpp/filesystem/create_directory

pzhristov:
@shahor02 Indeed something is wrong with my Macs... A standalone test shows that my fix is indeed wrong, but when I run the full system test, it always fails as I explained above and needs inverse logic.

shahor02:
@pzhristov can you try https://github.com/AliceO2Group/AliceO2/blob/2fef6d55f05b46ddfe4f59e1d0aafc8bab3cd20b/Common/Utils/include/CommonUtils/FileSystemUtils.h#L33 function on Mac, both in the cases when the directory is new and when it exists in advance?

PR #8720, opened 2022-05-04
ITS: Fix GPU vertexer

mconcas:
Functioning of ITS GPU vertexer is restored, will later test on EPN AMD gpus.
@davidrohr FYI


mconcas:
@davidrohr @shahor02 failures in CI should be unrelated, can we merge?

PR #8721, opened 2022-05-04
DPL: update what "now" means before setting the timer

ktf:
Due to the fact we do not invoke uv_run unless there is
some idle time between some data and the other, the concept
of "now" in libuv is not updated. This means that when
we setup a timer, libuv things it's still the past. However
when we run uv_run e.g. because data is not
arriving anymore, the time suddenly switches to the correct one,
triggering the callback immediately, because the 20 seconds from
the moment in the past have already passed.

This explains why end of stream is missed (FairMQ simply switched
already to the READY state) or why some race condition which should not
be so frequent (like the one for START-STOP-START), is so frequent.


davidrohr (approved):

ktf:
Merging. This is for sure correct, anything which breaks because of this is actually a separate race condition and we should understand it sooner rather than later.

PR #8724, opened 2022-05-04
Make assignHeaderVersion virtual to allow to detector to change version

shahor02:
@afurs


shahor02:
trivial + checked locally, merging.

PR #8727, opened 2022-05-04
Use TRandom::GetSeed to get what was set

mpuccio:
As documented in ROOT (yet completely misleading!)
```
Starting ROOT version 6.26/02.
root [0] gRandom->GetSeed()
(unsigned int) 624
root [1] gRandom->SetSeed(1)
root [2] gRandom->GetSeed()
(unsigned int) 624
```
but
```
root [16] gRandom->TRandom::GetSeed()
(unsigned int) 1
```

### DIFF
--- a/Generators/src/GeneratorFactory.cxx
+++ b/Generators/src/GeneratorFactory.cxx
@@ -71,7 +71,7 @@ void GeneratorFactory::setPrimaryGenerator(o2::conf::SimConfig const& conf, Fair
       LOG(info) << "Reading \'Pythia8\' base configuration: " << config << std::endl;
       gen->readFile(config);
     }
-    auto seed = (gRandom->GetSeed() % 900000000);
+    auto seed = (gRandom->TRandom::GetSeed() % 900000000);
     LOG(info) << "Using random seed from gRandom % 900000000: " << seed;
     gen->readString("Random:setSeed on");
     gen->readString("Random:seed " + std::to_string(seed));


Generators/src/GeneratorFactory.cxx
```diff
@@ -71,7 +71,7 @@ void GeneratorFactory::setPrimaryGenerator(o2::conf::SimConfig const& conf, Fair
       LOG(info) << "Reading \'Pythia8\' base configuration: " << config << std::endl;
       gen->readFile(config);
     }
-    auto seed = (gRandom->GetSeed() % 900000000);
+    auto seed = (gRandom->TRandom::GetSeed() % 900000000);
```
sawenzel, line 74: ok. This is of course confusing. Does the ROOT team know about this? I think we should open a ticket.

sawenzel (approved):

Generators/src/GeneratorFactory.cxx
sawenzel: ... so you are saying this is documented?

mpuccio: Yes, this is explicit in the [documentation of TRandom3](https://root.cern.ch/doc/master/classTRandom3.html#a6f01d53341006cbdbfa7ff98e29c5709)... I find it quite confusing too. It's even such that if you actually set 0 as seed the state is changed (and this is one of the reason we did not spot this before) while if you set any other number (or you don't set any seed) you will always get 624.

PR #8728, opened 2022-05-05
Fix for walking through DataDescriptorMatcher structure

matthiasrichter:
- Only visit the left node for operation 'not'.
- Additional small change for DataDescriptorMatcher created from
ConcreteDataTypeMatcher to be in line with the structure used
in all other cases.

Follow up to #8565

### DIFF
--- a/Framework/Core/src/DataSpecUtils.cxx
+++ b/Framework/Core/src/DataSpecUtils.cxx
@@ -334,7 +334,8 @@ MatcherInfo extractMatcherInfo(DataDescriptorMatcher const& top)
         state.hasError = true;
         return VisitNone;
       }
-      if (action.node->getOp() == ops::Just) {
+      if (action.node->getOp() == ops::Just ||
+          action.node->getOp() == ops::Not) {
         return VisitLeft;
       }
       return VisitBoth;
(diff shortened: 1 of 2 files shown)


Framework/Core/src/DataSpecUtils.cxx
```diff
@@ -510,7 +511,8 @@ DataDescriptorMatcher DataSpecUtils::dataDescriptorMatcherFrom(ConcreteDataTypeM
   auto timeDescriptionMatcher = std::make_unique<DataDescriptorMatcher>(
     DataDescriptorMatcher::Op::And,
     DescriptionValueMatcher{dataType.description.as<std::string>()},
-    StartTimeValueMatcher(ContextRef{0}));
+    std::make_unique<DataDescriptorMatcher>(DataDescriptorMatcher::Op::Just,
```
ktf, line 514: Why would you add the extra node?

matthiasrichter: this is to make it the same structure as in the 3 other `dataDescriptorMatcherFrom` methods, e.g. https://github.com/AliceO2Group/AliceO2/blob/e7320674a230d05f10ca9ce983b2bc14c32627a1/Framework/Core/src/DataSpecUtils.cxx#L503

ktf: yes, understood but then should't we support both cases? the matcher with three ands is a perfectly valid one.

ktf: yes, understood but then should't we support both cases? the matcher with three ands is a perfectly valid one.

matthiasrichter: ok, my change was more in terms of code, not functionality. Its not required and not part of the actual fix in this PR. It only adopts the code structure of the other methods.

matthiasrichter: both cases are still supported, the same functionality can also be expressed with a different layout of the DataDescriptorMatcher

PR #8729, opened 2022-05-05
Ctpdev: config and scalers to CCDB

lietava:


lietava:
Looking into logs I do not see any problem in those which failed. Please, let me know if otherwise.

shahor02:
FullCI says: 
```
/sw/SOURCES/O2/8729-slc8_x86-64/0/DataFormats/Detectors/CTP/src/Configuration.cxx:639:25: error: statement should be inside braces [readability-braces-around-statements]
```

Detectors/CTP/macro/CreateCTPConfigRun2.C
```diff
@@ -0,0 +1,112 @@
...
+2 cl_1khz bcd10 \n \
+3 cluster clu4 emc cpv \n \
+4 cl_5khz bcd20 \n";
+  // ctpcfg.loadConfiguration(cfgstr);
+  ctpcfg.loadConfigurationRun3(cfgRun3str);
+  ctpcfg.printStream(std::cout);
+  std::cout << "Going write to db" << std::endl;
+  // return;
+  ///
+  /// add to database
```
sawenzel: I would find it more natural to return just the CTP config object from this function (or store it in a file).
The upload to CCDB could then be done with an outside tool (o2-ccdb-uploadfile or curl). In this way, one could create the object also locally without an accompanying CCDB upload.

lietava:
all tests ok, please, merge

PR #8731, opened 2022-05-05
DPL Analysis: Add pack selection based on condition with multiple parameters

saganatt:
@ktf @aalkin 

This will be used to automagically extract correct index columns from tables to supply them to sliceByCached() in event mixing.

### DIFF
--- a/Framework/Foundation/include/Framework/Pack.h
+++ b/Framework/Foundation/include/Framework/Pack.h
@@ -100,46 +100,28 @@ template <typename P1, typename P2>
 using interleaved_pack_t = decltype(interleave_pack(P1{}, P2{}));
 
 /// Selects from the pack types that satisfy the Condition
-template <template <typename> typename Condition, typename Result>
-constexpr auto select_pack(Result result, pack<>)
-{
-  return result;
-}
-
-template <template <typename> typename Condition, typename Result, typename T, typename... Ts>
-constexpr auto select_pack(Result result, pack<T, Ts...>)
-{
-  if constexpr (Condition<T>()) {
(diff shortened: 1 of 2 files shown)


Framework/Foundation/include/Framework/Pack.h
```diff
@@ -100,24 +100,28 @@ template <typename P1, typename P2>
...
+    return select_pack<Condition>(result, pack<Ts...>{}, condPack);
   }
 }
 
-template <template <typename> typename Condition, typename... Types>
-using selected_pack = std::decay_t<decltype(select_pack<Condition>(pack<>{}, pack<Types...>{}))>;
+template <template <typename...> typename Condition, typename... Types>
+using selected_pack = std::decay_t<decltype(select_pack<Condition>(pack<>{}, pack<Types...>{}, pack<>{}))>;
+template <template <typename...> typename Condition, typename CondPack, typename Pack>
+using selected_pack_multicondition = std::decay_t<decltype(select_pack<Condition>(pack<>{}, Pack{}, CondPack{}))>;
```
ktf, line 124: I am not sure  why you would need a "multicondition" version. Can't you simply do it with some and operator?

saganatt: This is my use case: https://github.com/saganatt/AliceO2/blob/mixing-slice-cached/Framework/Core/include/Framework/GroupedCombinations.h#L43

I need somewhat to pass the `G` type. With previous select there was no way to use a comparison that needs more than 1 template argument, like `is_index_to_g` (and `is_binding_compatible_v`)

PR #8738, opened 2022-05-06
DPL: fix issue with 100% CPU usage when output proxy is idling

ktf:


ktf:
@shahor02 @davidrohr @chiarazampolli this should work correctly now, the issue with the missing end of stream should be fixed. How about we merge it and we do more tests with detectors next week?

davidrohr:
Agreed. But Can you merge it tomorrow? Ibwant tobhave it in a separate build not in tomorrow morning's nightly. So we can easily enable/disable it.

Kind Regards
David Rohr

On 7 May 2022 00:26:57 CEST, Giulio Eulisse ***@***.***> wrote:
***@***.*** @davidrohr @chiarazampolli this should work correctly now, the issue with the missing end of stream should be fixed. How about we merge it and we do more tests with detectors next week?

ktf:
yes, sure.

PR #8739, opened 2022-05-06
DPL: drop messages when they are too old

ktf:


ktf:
@shahor02 @aferrero2707 @aphecetche this fixes the MCH test we did the other day for me.

shahor02:
@ktf for me `o2-mch-cru-page-reader-workflow --infile data-158-ped-20220414.raw --configKeyValues "HBFUtils.startTime=1651562466196;HBFUtils.runNumber=514769" --full-tf | o2-mch-pedestal-decoding-workflow --input-spec "readout:RDT/RAWDATA" |  o2-calibration-mch-badchannel-calib-workflow  --shm-segment-size 16000000000` still gets stuck. Do I need anything else apart from this PR?

ktf:
@davidrohr this implements what we discussed today and it actually seems to fix @shahor02 / @aphecetche / @aferrero2707 mch issue. It still needs a bit of cleanup of log messages / squashing, but apart from that it should be ready.

I am now testing this with the FLP workflows.

davidrohr:
Thx, that would be great! :)
Once you are ready with the cleanup let me know, then I'll try FST with QC and TF throttling enabled.
Hopefully that is fixed as well.
Once it works and is merged, we should try to deploy at P2 as soon as possible.

ktf:
@TimoWilken do you understand why it started testing CS8 only this morning? Was there any issue with it?

ktf:
o2-dataflow test failing due to CCDB. o2-dataflow-cs8 works, so I guess that's what matters.

Merging this to go forward and since it fixes a few obvious issues with QC / Aggregator workflow. Will handle the fallout separately.

vkucera:
@ktf This PR breaks table reading/processing from the AO2D files. Reported here https://alice.its.cern.ch/jira/browse/O2-2963

PR #8740, opened 2022-05-06
Warning fixes

peressounko:


sawenzel:
There is a failure in CPV raw data reading in the full-CI.

shahor02:
Something was wrong with CPV digits tee, just run FST locally, did not see any problems.

peressounko:
Indeed writing for(;;) instead of range based for( a: b)  causes crush. Probably this is related to compiler optimization?
Reverted previous form. Not clear how to fix warning (unused evnt) here?
// First version of for loop causes fault Optimizer???
//  for (auto evnt = treereader->begin(); evnt != treereader->end(); ++evnt) {
  for (auto evnt : *treereader) {
    rawwriter.digitsToRaw(*digitbranch, *triggerbranch);
  }
Also reverted CPVPedestalCalibrator where few lines were erroneously removed.

peressounko:
dear @shahor02, @sawenzel could you please look and merge this PR?

shahor02 (approved):

PR #8741, opened 2022-05-06
ShmManager: do a full reset at the start

rbx:
@davidrohr am I forgetting a use case where you need a reconnection to existing segments by this tool?


davidrohr:
I don't see a need to reconnect to an existing segment. If the tool has such a feature, I'd only make it optional and disable it by default.

davidrohr (approved):

PR #8742, opened 2022-05-06
DPL Analysis: Event mixing: Simplify mixing codes with sliceByCached. Mixing on Filtered and Join

saganatt:
Hi @ktf , @jgrosseo ,

this change replaces GroupSlicer with `sliceByCached` for slicing tables. I need yet to see in the benchmarks whether and how much it will improve mixing performance (I expect that caching will help).
Additionally, this mixing already works well with Filtered and Joins, I will check how it can be used with partitions. So at least in this regard it is better than the previous version.

### DIFF
--- a/Framework/Core/include/Framework/AnalysisManagers.h
+++ b/Framework/Core/include/Framework/AnalysisManagers.h
@@ -40,16 +40,15 @@ struct GroupedCombinationManager {
   }
 };
 
-template <typename T1, typename GroupingPolicy, typename BP, typename G, typename... Us, typename... As>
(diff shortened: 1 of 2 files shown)


saganatt:
There is one side change - @jgrosseo, canonically, do we want to skip collision pairs with no tracks available, or do the users expect to receive all collision pairs?

jgrosseo:
> There is one side change - @jgrosseo, canonically, do we want to skip collision pairs with no tracks available, or do the users expect to receive all collision pairs?

Yes, analyzers should receive all pairs so that they can count correctly. If you can allow a flag to skip those pairs, that is a good feature but the default should be to also retrieve pairs without tracks.

jgrosseo:
@ktf I merge this if you have no objections

PR #8744, opened 2022-05-06
new plots for ITSTrackTask / ClusterSize vs Chip (O2 modification, Tr…

delico18:
…ackITSExt Class Usage)

(TH2D) Corrected cluster size for track clusters vs Chip by Layer

(TH2D) Corrected cluster size for track clusters vs Chip Full Detector

O2 modification is implemented.
There is one modification for O2/Detectors/ITSMFT/ITS/workflow/src/TrackerSpec.cxx
TrackITS ->TrackITSExt (L196)


shahor02 (changes requested):
@delico18 this cannot be done: the TrackITS is what all downstream devices expect. 
Wahr it the reason of this change? The track->cluster association is accessible via references, see e.g. 
https://github.com/AliceO2Group/AliceO2/blob/6593894a7b7c91b46ae5a71f30ccf7f4066d4c8d/Detectors/GlobalTracking/src/MatchTPCITS.cxx#L1259-L1274

PR #8746, opened 2022-05-06
Matchinfohmp

gvolpe79:


sawenzel:
@gvolpe79 : Would you be able to provide a reasonable description for this PR? We should also squash the ~130 commits most likely.

gvolpe79:
Hi Sandro, 
I created the matching info for the HMPID, and modified the RecoContainer adding the HMPID matching info and the clusters. 
There was anyway a problem! The student we have here at cern, that is doing some development in parallel for the HMPID calibration, committed directly on the dev of my fork end not in a separate branch as I requested. Once I realized this, I updated the dev removing all his commits and started from there for the matchinfohmp branch. I think it should be ok.

DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
```diff
@@ -213,6 +220,7 @@ struct DataRequest {
   void requestTRDTracklets(bool mc);
   void requestMCHClusters(bool mc);
   void requestMIDClusters(bool mc);
+  void requestHMPClusters(bool mc);
```
shahor02, line 223: You should also add `requestHMPMatches(bool mc)`

DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
```diff
@@ -580,6 +595,11 @@ struct RecoContainer {
   auto getTOFClusters() const { return getSpan<o2::tof::Cluster>(GTrackID::TOF, CLUSTERS); }
   auto getTOFClustersMCLabels() const { return mcTOFClusters.get(); }
 
+  // HMPID clusters
+  auto getHMPClusters() const { return getSpan<o2::hmpid::Cluster>(GTrackID::HMP, CLUSTERS); }
+  auto getHMPTriggers() const { return getSpan<o2::hmpid::Trigger>(GTrackID::HMP, CLUSREFS); }
```
shahor02: Your matches are also grouped per trigger, right? Then there should be separate 
```
auto getHMPClusterTriggers() const { return getSpan<o2::hmpid::Trigger>(GTrackID::HMP, CLUSREFS); }
auto getHMPClusters() const { return getSpan<o2::hmpid::Cluster>(GTrackID::HMP, CLUSTERS); }
```
and e.g.
```
auto getHMPMatchTriggers() const { return getSpan<o2::hmpid::Trigger>(GTrackID::HMP, TRACKREFS); }
auto getHMPMatches() const { return getSpan<o2::hmpid::MatchInfoHMP>(GTrackID::HMP, MATCHES); }
```

DataFormats/Reconstruction/include/ReconstructionDataFormats/MatchInfoHMP.h
```diff
@@ -0,0 +1,69 @@
...
+  void print() const;
+
+ private:
+  int mIdxHMPClus;    // Idx for HMP cluster
+  GTrackID mIdxTrack; // Idx for track
+  float mCkovAngle;
+  float mMipCluQ = 0.0;                 // MIP cluster charge
+  int mMipCluSize = 0.0;                // MIP cluster size
+  int mNPhots = 0.0;                    // number of candidate photo-electrons
+  std::vector<float>* mPhotsClusCharge; // list of cluster charge of candidate photo-electrons
```
shahor02: we are trying to avoid dynamical size data formats, moreover with raw pointer, such objects are not messageable via shared memory and need require expensive root streaming. 
Do you really need a vector of charges? In such cases we usually add yet another vector which would keep the charges and the match would just keep the 1st entry index and the number of entries it has in this vector.

DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
gvolpe79: ok, and  I have to keep this then?

const o2::dataformats::MatchInfoHMP& getHMPMatch(GTrackID id) const { return getObject<o2::dataformats::MatchInfoHMP>(id, MATCHES); }

DataFormats/Reconstruction/include/ReconstructionDataFormats/MatchInfoHMP.h
gvolpe79: It refers to the charge of the cherenkov photon clusters for the given track. This is a information very useful and I would like to not lose it! So adding an other vector, but it is not clear to me where?

DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
shahor02: As you prefer, the same can be obtained as 
```
auto mtspan = recoCont.getHMPMatches();
auto& match = mtspan[id.getIndex()];
```
Note that each `getHMPMatch(GTrackID id) ` would go via new span creation. It is not very expensive but still more expensive than getting span once then iterating over its elements.

DataFormats/Reconstruction/include/ReconstructionDataFormats/MatchInfoHMP.h
shahor02: E.g. HMPClusterCharges via
```
auto getHMPClusterCharges() const { return getSpan<float>(GTrackID::HMP, PATTERNS); }
```
The idea is to refer to the range of entries in this vector from the HMPCluste in the same way as you refer from HMPTriggers to HMPClusters.

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
```diff
@@ -266,6 +266,17 @@ void DataRequest::requestMCHClusters(bool mc)
   requestMap["clusMCH"] = mc;
 }
 
+void DataRequest::requestHMPClusters(bool mc)
+{
+  addInput({"hmpidcluster", "HMP", "CLUSTERS", 0, Lifetime::Timeframe});
+  addInput({"hmpidtriggers", "HMP", "CLUSREFS", 0, Lifetime::Timeframe});
```
shahor02, line 272: The special vector of charges is part of the clusters info, should be added here

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
```diff
@@ -941,6 +967,15 @@ void RecoContainer::addTOFClusters(ProcessingContext& pc, bool mc)
   }
 }
 
+//__________________________________________________________
+void RecoContainer::addHMPClusters(ProcessingContext& pc, bool mc)
+{
+  commonPool[GTrackID::HMP].registerContainer(pc.inputs().get<gsl::span<o2::hmpid::Cluster>>("hmpidcluster"), CLUSTERS);
+  commonPool[GTrackID::HMP].registerContainer(pc.inputs().get<gsl::span<o2::hmpid::Trigger>>("hmpidtriggers"), CLUSREFS);
```
shahor02, line 994: here also the cluster charges vector must be added.

DataFormats/Detectors/HMPID/include/DataFormatsHMP/Cluster.h
```diff
@@ -120,6 +120,16 @@ class Cluster
 };
 
 } // namespace hmpid
+
```
shahor02, line 123: please fix the Cluster class, it should not contain the pointer on the vector, instead the charges should be accessed from the separate vector.

shahor02 (changes requested):
Hi @gvolpe79 

See comments below. In short: your clusters still allocate the memory via pointer to the vector, they should not (they are not messageable now, even if you declare them to be so). The variable size info should come via separate vector, the cluster will refer to the entries in this vector.
For the matches info (which should go the AOD together with matched clusters info?): it should either contain to the selected clusters vector which is also stored together with refilled charges vector or (preferable to not increase the number of vectors) refer directly on the charges vector of selected clusters.
In any case, if I understood correctly what you need, there will be 2 vector of charges: one produced together with clusters and another one only for charges of matched clusters

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
shahor02: discard this comment, it was wrong

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
shahor02: discard this comment, it was wrong

DataFormats/Detectors/HMPID/include/DataFormatsHMP/Cluster.h
shahor02: discard this comment, it was wrong

shahor02 (changes requested):
Hi @gvolpe79 

Please discard my previous comments, they were wrong:  I've probably misread your Cluster class, thinking that the mDigs vector you have there is also for the charges. Now I realize that you have if only at the clustering stage and then don't use it, is it so?

But there are another problems:
1) you declared/implemented `requestHMPClusters()` but you should invoke it in from the `requestClusters`, e.g. here (see the same for other detectors):
https://github.com/AliceO2Group/AliceO2/blob/183cdd3e763b2197bc63d832e25467e3b835614a/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L512

2) You have declared `requestHMPMathes` but you neither implement it nor invoked it from the  requestTracks, should add somewhere here:
https://github.com/AliceO2Group/AliceO2/blob/183cdd3e763b2197bc63d832e25467e3b835614a/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L475

3) You have declared/implemented `addHMPMatches` but you should also invoke it e.g. in https://github.com/AliceO2Group/AliceO2/blob/183cdd3e763b2197bc63d832e25467e3b835614a/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L701

gvolpe79:
Hi @shahor02, 
I just committed the requested modifications!

PR #8747, opened 2022-05-06
Fix in ccdb snapshot creation/usage with CCDBFetcher

shahor02:


sawenzel:
Could you elaborate a bit what is "fixed" or improved?

shahor02:
With `CcdbApi::loadFileToMemory` method (used by the DPL fetcher) the validity of cached object was not checked when the snapshot creation was requested. 
Also, fixed the directory creation for snapshot files:
https://github.com/AliceO2Group/AliceO2/pull/8747/files#diff-2355211fcd8f9232fbeddc19348089f206c655459237baffc0252dbd544dd025L560-L563 is error prone: it will fail in case the directory is created by concurrent process after the 1st `if` and and before the 2nd.

sawenzel (approved):

PR #8751, opened 2022-05-06
Fix ITS CTF format and subspecs in CTF reader

shahor02:


shahor02:
all was green except formatting which is fixed, merging.

PR #8753, opened 2022-05-07
MC: Optionally apply device-combine to a couple of workflows

sawenzel:
The novel ability to combine multiple DataProccesorSpecs (devices)
into a single one is applied (optionally) in a couple of workflows for
MC:
- digitization
- TOC-TPC-ITC matching
- primary vertex finding
- secondary vertex finding
- AOD creation

In all cases, we batch together multiple ROOT reader or writer devices.
In case of digitization, we also combine multiple detector digitizers together.

The benefits are reduction of load time, less CPU system usage, and importantly
a significant reduction in memory (~1GB for AOD creation alone). We also share
more objects like geometry/field between multiple digitizers.

Nothing changes by default: The combinations need to be activated via special
`--combine-devices` (or similar) flags.

In order to make this work, a couple of things needed to be changed/adapted:
- small changes in RootTreeWriter to allow treatment of multiple TTree-TFiles
  pairs in the same process
- usage of unique InputSpec binding-labels in some detector specs
  ... otherwise these InputSpecs cannot be combined into a single device
- for now use of a custom device-combine method until native DPL
  framework::workflow::combine can support combination of devices that
  have same option keys (such as multiple RootTreeWriters together).

### DIFF
--- a/Detectors/Base/include/DetectorsBase/DPLWorkflowUtils.h
+++ b/Detectors/Base/include/DetectorsBase/DPLWorkflowUtils.h
@@ -1,0 +1,264 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+// some user-space helpers/utilities for DPL workflowspecs
+
+//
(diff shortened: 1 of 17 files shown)


Detectors/Base/include/DetectorsBase/DPLWorkflowUtils.h
```diff
@@ -0,0 +1,264 @@
...
+  return std::string("");
+}
+
+// a utility combining multiple specs into one
+// (with some checking that it makes sense)
+// spits out the combined spec (merged inputs/outputs and AlgorithmSpec)
+// Can put policies later whether to multi-thread or serialize internally etc.
+// fills a "remaining" spec container for things it can't simply merge
+// (later on we could do a full topological sort / spec minimization approach)
+o2::framework::DataProcessorSpec specCombiner(std::string const& name, std::vector<DataProcessorSpec> const& speccollection,
```
ktf, line 108: Why not using / adapting  o2::framework::combine?

PR #8755, opened 2022-05-08
o2-eve: calorimeters display, fixes in json serialisation

jmyrcha:


EventVisualisation/View/src/Initializer.cxx
```diff
@@ -123,19 +123,18 @@ void Initializer::setupGeometry()
     string detName = gVisualisationGroupName[det];
     LOG(info) << detName;
 
-    if (detName == "TPC" || detName == "MCH" || detName == "MID" || detName == "MFT") { // don't load MUON+MFT and AD and standard TPC to R-Phi view
```
pnwkw: A mistake during merge? All changed lines in Initializer.cxx are a regression.

pnwkw (changes requested):

EventVisualisation/View/src/Initializer.cxx
jmyrcha: fixed

shahor02 (approved):

PR #8760, opened 2022-05-09
Omit token check if internal CCDB is used

martenole:
Hi @davidrohr 
should we maybe avoid the check for the token in case we are using the internal CCDB on the EPNs? Otherwise one gets an error when trying to run the `start_tmux.sh` script, because this does not set the EPNSYNCMODE.

### DIFF
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -5,7 +5,7 @@
 MYDIR="$(dirname $(realpath $0))"
 source $MYDIR/setenv.sh
 
-if [[ $EPNSYNCMODE == 0 ]]; then
+if [[ $EPNSYNCMODE == 0 ]] && [[ $DPL_CONDITION_BACKEND != "http://o2-ccdb.internal" ]]; then
   alien-token-info >& /dev/null
   if [[ $? != 0 ]]; then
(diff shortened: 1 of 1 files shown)


davidrohr (changes requested):
fine with me, but the check should be
`[[ "0$DPL_CONDITION_BACKEND" != "0http://..." ]] since the variable is not guaranteeed to be set.

martenole:
But even if the variable is not set then the check as I have it `[[ $DPL_CONDITION_BACKEND != "http://o2-ccdb.internal" ]] == true`. It is the same results as if I would do as you suggest. So I would leave it as it is, no?

davidrohr:
> But even if the variable is not set then the check as I have it `[[ $DPL_CONDITION_BACKEND != "http://o2-ccdb.internal" ]] == true`. It is the same results as if I would do as you suggest. So I would leave it as it is, no?

ah, just learned something:
```
qon@qon ~ $ unset FOO
qon@qon ~ $ if [ $FOO != "aaa" ]; then echo a; fi
bash: [: !=: unary operator expected
qon@qon ~ $ if [[ $FOO != "aaa" ]]; then echo a; fi
a
```
Normally in posix style the string must not be empty, so the 0 in front ensures that. Apparently the bash `[[` syntax allows also empty strings.

PR #8763, opened 2022-05-09
Fixed bitshifts in getQFromRaw from raw MCM data

f3sch:
The charge distribution from TRD tracklet data I investigated were nonsense.
The reason is that some bitshifts in the function for getting the tracklet charge from the raw MCM data were wrong.
Q0 is unaffected.

 A detailed description of the format can be found in https://alicetrd.web.cern.ch/alicetrd/tdp/main.pdf under 17.2.1.

Signed-off-by: Felix Schlepper <felix.schlepper@cern.ch>

### DIFF
--- a/DataFormats/Detectors/TRD/src/RawData.cxx
+++ b/DataFormats/Detectors/TRD/src/RawData.cxx
@@ -132,7 +132,7 @@ uint32_t getQFromRaw(const o2::trd::TrackletMCMHeader* header, const o2::trd::Tr
   /*
    * Q0/1 are 7 bits, Q2 is 6 bits.
    * Q0 is completely in TrackletMCMData::data and Q2 is completely in TrackletHCHeader::pid*,
-   * while Q1 is split with the lower 5 bits in the header and the upper 2 bits in the data.
+   * while Q1 is split with the lower 5 bits in the data and the upper 2 bits in the header.
    *
    * A detailed description of the format can be found in
    * https://alicetrd.web.cern.ch/alicetrd/tdp/main.pdf under 17.2.1
--- a/DataFormats/Detectors/TRD/src/RawData.cxx
+++ b/DataFormats/Detectors/TRD/src/RawData.cxx
@@ -130,19 +130,34 @@ uint32_t getQFromRaw(const o2::trd::TrackletMCMHeader* header, const o2::trd::Tr
       break;
   }
   /*
-   * Q0/1 are 7 bits, Q2 is 6 bits
+   * Q0/1 are 7 bits, Q2 is 6 bits.
    * Q0 is completely in TrackletMCMData::data and Q2 is completely in TrackletHCHeader::pid*,
    * while Q1 is split with the lower 5 bits in the header and the upper 2 bits in the data.
+   *
+   * A detailed description of the format can be found in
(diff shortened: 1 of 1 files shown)


tdietel:
Hello Felix,

Thanks for investigating this. It looks like you found the bug.

Could you add a reference in your comment (which is very helpful) where you found the bit mappings? Did you talk to Venelin, or did you get it from the TDP? 

@bazinski, do you have any comments?

Cheers,
Tom.

f3sch:
Hi Tom,

I updated my PR and added a more detailed description on the calculation of Q1.
Additionally, I added the TDP as a resource, however I am unsure about adding hard links in code?
I did not had the chance to talk to Venelin, yet, just @bazinski.
Maybe the hardware does some weird bitmanipulation, but even then the calculation of Q1/2 would be wrong in my opinion.

Cheers
Felix

bazinski:
@tdietel  Yes @f3sch emailed me before doing this PR.
The contents of this pr is relative to 1 part of the TDP, and correct according 1 part of the asm (DYN_FP_Q==0).


Problem goes back to a discussion we had about using the 4 bit format to define the charge windows.
The actual format is defined by DYN_FP_Q
if DYN_FP_Q == 1 then we have a different format to the one defined in this PR.
line 174 onwards of fit_fit.asm which defines the 2 bit off set and then all Q's being 6 bit, giving Q0 and Q1 of 8 bits(effectively) and Q2 still of 6 bits.

DYN_FP_Q == 0  we have : 
line 244 of fit_fit.asm builds the complete PID of " (Q0 << 13) | (Q1 << 6) | Q2" and then splits it into hi and low at lines 254 to 256
lower pid (LPID) is 12 bit and the HPID is 8bit for a specific tracklet.
which is what is correctly defined in this pr, not sure what i was thinking re the 0x2f, but its interesting that nobody has picked it up till now.

The asm encodes DYN_FP_Q into the lsb of the format bits at line 363 of run_post_const.tcs

tdietel:
Ok, looks good. The only issue is if we ever switch on the DYN_FP feature - but I think using the fixed precision is actually cleaner. 

So let's merge this PR.

(I wish I was surprised that nobody has picked this up yet...)

tdietel (dismissed):

martenole:
Hi Felix,
first of all kudos for digging deep into the code and for this first PR which contains already a substantial change. Well done! As you can see there are some CI tests ongoing which need to be passed first, before we can merge. In addition to these checks I am also doing some manual tests to compare your version of the raw reader to the one we currently have. This will take some more time, but as soon as I have the results I'll let you know. And then we could merge your PR.
Cheers,
Ole

tdietel (approved):

tdietel:
The CI failed in building O2, but I do not really see what the problem is. Could it be that it's a transient/unrelated problem?

bazinski:
@tdietel ja that is all o2physics and some qc testing stuff that is failing, not related to this code as far as i could see.

martenole (approved):
Indeed CI failures are unrelated, fullCI was green before. Plus I tested it locally and it did not break the processing.
@sawenzel could you please merge this? I cannot merge myself unless all required checks are green

PR #8765, opened 2022-05-09
MCH: fixed BC rollover in digits merger

aferrero2707:
The BC rollover constant is set to OrbitsInTF * 3564 in the case of the BC reset at each TF.


aferrero2707:
@aphecetche this PR fixes the computation of the digit time at the very beginning of the time-frames, particularly for digits that are recorded before the BC reset signal is received by the front-end SAMPA chips.

Incidentally, it also solves many of the "multiple digits on the same pad" messages in the pre-clustering.

aphecetche (approved):

PR #8766, opened 2022-05-09
mis-calibration in digitization

AllaMaevskaya:
- mis-calibration in digitization

- move FT0DigParam to base


Detectors/FIT/FT0/workflow/src/ReconstructionSpec.cxx
```diff
@@ -55,11 +58,11 @@ void ReconstructionDPL::run(ProcessingContext& pc)
...
-  // auto caliboffsets = pc.inputs().get<o2::ft0::FT0ChannelTimeCalibrationObject*>("ft0offsets");
-  //  mReco.SetChannelOffset(caliboffsets.get());
-  auto caliboffsets = mCCDBManager.get<o2::ft0::FT0ChannelTimeCalibrationObject>("FT0/Calibration/ChannelTimeOffset");
-  mReco.SetChannelOffset(caliboffsets);
-  LOG(debug) << " RecoSpec  mReco.SetChannelOffset(&caliboffsets)";
+  if (mUpdateCCDB) {
+    auto caliboffsets = pc.inputs().get<o2::ft0::FT0ChannelTimeCalibrationObject*>("ft0offsets");
+    mReco.SetChannelOffset(caliboffsets.get());
+    LOG(info) << "RecoSpec  mReco.SetChannelOffset(&caliboffsets)";
+  }
```
shahor02, line 65: This will work since you have just 1 CCDBobject, in general, you don't need to set again the pointer of the calib object here. The idea of using the `finaliseCCDB` method is:
1) here you simply call
```
if (mUpdateCCDB) {
  pc.inputs().get<o2::ft0::FT0ChannelTimeCalibrationObject*>("ft0offsets");
}
```
Note that you ignore the return value: this will check if the object needs to be updated (which is always a case at the 1st call) and if it is, will call the `finaliseCCDB` with the matcher and pointer of this object.
2) in the finaliseCCDB (called for each CCDB object and only when it is updated) you just do
```
if (matcher == ConcreteDataMatcher("FT0", "TimeOffset", 0)) {
 mReco.SetChannelOffset((o2::ft0::FT0ChannelTimeCalibrationObject*)obj)
 LOG(info) << "TimeOffset updated";
  mUpdateCCDB = false; // only if you don't want to check anymore the validity of the object (though it costs nothing) and note that this is dangerous if you have multiple ccdb objects, it is safer to keep this line in the `if (mUpdateCCDB) ...` above.
return;
}
```

shahor02 (approved):

TimoWilken:
Hi @AllaMaevskaya @shahor02, since this PR was merged, O2 compilation in CI is failing with the error:

```
+ cmake --build . -- -j16 install
ninja: error: dependency cycle: cmake_object_order_depends_target_O2lib-FT0Calibration -> Detectors/FIT/FT0/calibration/G__O2FT0Calibration.cxx -> stage/lib64/G__O2FT0Simulation_rdict.pcm -> stage/lib64/libO2FT0Calibration.so -> Detectors/FIT/FT0/calibration/CMakeFiles/O2lib-FT0Calibration.dir/src/FT0ChannelTimeTimeSlotContainer.cxx.o -> cmake_object_order_depends_target_O2lib-FT0Calibration
```

It seems to me that the reordering of includes in this PR probably introduced a circular dependency. I'm not sure why this wasn't picked up in the CI for this PR -- perhaps another PR that was merged at a similar time is interacting badly with this one?

shahor02:
Hi @TimoWilken 
yes, I've also got this and had to rebuild with alibuild (passed w/o problem). Before the FT0Calibration depended on the O2FT0Simulation and this was reflected in the generated pcm files, now the dependence was reversed (@AllaMaevskaya may comment if it is really needed). 
Apparently, our build does not update the pcm files (not sure why) so the suppressed old dependency conflicts with newly introduced one. Could you try to recompile after removing `G__O2FT0Simulation_rdict.pcm` and `G__O2FT0Calibration_rdict.pcm`?

PR #8771, opened 2022-05-10
CTP: Make the NCOUNTERS variable constexpr

pnwkw:
This small change is required to fix O2 building under my CLion dev environment. What is strange is that an aliBuild build was successful - maybe the compiler parameters the IDE is using by default are different?


shahor02 (approved):
trivial, merging

PR #8776, opened 2022-05-10
FT0: new decoder, based on AVX512

afurs:
New decoder for FT0, contains a lot of low level code based on AVX512 CPU instructions. Can be used only at production FLP machines. Best option if this code can be compiled only  for FLPSuite package(via predefinitions?).

TODO:
- BC-orbit protection
- empty raw data protection(in case if data block contains only header, w/o data)
- additional enhancement via pre-caching for buffers(make more cache friendly code).

This decoder is tested(full chain readout+StfB+DPL+StfS) at FLP machine on dummy data samples which emulates maximum 600Mb/(link*second) data flow(1 TF ~ 250Mb).

### DIFF
--- a/Detectors/FIT/FT0/workflow/src/ft0-flp-workflow.cxx
+++ b/Detectors/FIT/FT0/workflow/src/ft0-flp-workflow.cxx
@@ -17,6 +17,11 @@
 #include "FT0Raw/RawReaderFT0Base.h"
 #include "SimulationDataFormat/MCTruthContainer.h"
 #include "Framework/CompletionPolicyHelpers.h"
+#if defined __has_include
+#if defined(__linux__) && (defined(__x86_64) || defined(__x86_64__)) && __has_include(<emmintrin.h>) && __has_include(<immintrin.h>)
+#include "FT0Workflow/FT0DataDecoderDPLSpec.h"
+#endif
+#endif
 
 using namespace o2::framework;
 
@@ -26,7 +31,6 @@ void customize(std::vector<o2::framework::CompletionPolicy>& policies)
   // ordered policies for the writers
   policies.push_back(CompletionPolicyHelpers::consumeWhenAllOrdered(".*(?:FT0|ft0).*[W,w]riter.*"));
 }
-
 // we need to add workflow options before including Framework/runDataProcessing
 void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
 {
@@ -58,6 +62,12 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
                     o2::framework::VariantType::Bool,
                     false,
                     {"do not subscribe to FLP/DISTSUBTIMEFRAME/0 message (no lost TF recovery)"}});
+
+  workflowOptions.push_back(
+    ConfigParamSpec{"new-decoder",
+                    o2::framework::VariantType::Bool,
+                    false,
+                    {"New decoder which uses AVX-512 CPU instractions"}});
 }
 
 // ------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


afurs:
It looks like there are unrelated issue in build/O2/fullCI

shahor02:
@costing the fetcher fails on
```
[10762:internal-dpl-ccdb-backend]: Info in <TJAlienFile::Open>: Accessing file /alice/data/CCDB/CPV/Calib/CTFDictionary/08/64436/d33bd720-515f-11ec-8357-2a010e0a0b16 in SE <ALICE::CERN::EOS>
[10762:internal-dpl-ccdb-backend]: Info in <TJAlienFile::Open>: Accessing file /alice/data/CCDB/TRD/Calib/NoiseMapMCM/10/03511/119c250c-cae8-11ec-981d-200114580202 in SE <ALICE::CERN::OCDB>
[10762:internal-dpl-ccdb-backend]: [09:12:33][ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch/GLO/Config/GRPECS/1652259380260/
[10762:internal-dpl-ccdb-backend]: [09:12:33][FATAL] Unable to find object GLO/Config/GRPECS/0
```

But there is a matching object:
```
root [0] auto& cm = o2::ccdb::BasicCCDBManager::instance();
root [1] cm.setURL("http://alice-ccdb.cern.ch");
root [2] cm.setTimestamp(1652259380260);
root [3] cm.get<o2::parameters::GRPECSObject>("GLO/Config/GRPECS");
Info in <TJAlienConnectionManager>: Opening connection to JCentral. Please wait
Info in <TJAlienConnectionManager>: Opening connection to 137.138.99.140
Info in <TJAlienConnectionManager>: Successfully connected to 137.138.99.140
Info in <TJAlienFile::Open>: Accessing file /alice/data/CCDB/GLO/Config/GRPECS/09/25116/72b550cc-d10b-11ec-8000-0aa145271b9a in SE <ALICE::CERN::OCDB>
(o2::parameters::GRPECSObject *) 0xd34a3a0
```

Any idea why it fails?

costing:
The object indeed exists (tested with `curl -i 'http://alice-ccdb.cern.ch/GLO/Config/GRPECS/1652259380260/'`) , but the error message claims that 
`Unable to find object GLO/Config/GRPECS/0`
could it be that the timestamp is not passed correctly ?

shahor02:
> The object indeed exists (tested with `curl -i 'http://alice-ccdb.cern.ch/GLO/Config/GRPECS/1652259380260/'`) , but the error message claims that `Unable to find object GLO/Config/GRPECS/0` could it be that the timestamp is not passed correctly ?

No, `0` here is not the timestamp but the timeslice (TF) ID for which the query was done. The query to the CcdbApi was done with ts = 1652259380260.

costing:
Then I have no idea why it fails. @ktf ?

ktf:
are you sure you are not querying by run number? What's the input spec for that object?

shahor02:
At the moment it is not run-dependent: 
https://github.com/AliceO2Group/AliceO2/blob/4aaa476a917bf40d314aaa009ca7b5202a661440/Detectors/Base/src/GRPGeomHelper.cxx#L62-L64

costing:
Hi Ruben,

As discussed in person, the requests are now logged. Next time we should at least be able to look at the exact request that was made to the server. It would help if you could add some unique identifier in the form of a "User-Agent" HTTP header (i.e. with https://curl.se/libcurl/c/CURLOPT_USERAGENT.html).

Cheers,

.costin

Detectors/FIT/FT0/workflow/src/ft0-flp-workflow.cxx
```diff
@@ -71,27 +81,36 @@ WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
...
+      if (!disableRootOut) {
+        specs.emplace_back(o2::fit::FITDigitWriterSpecHelper<RawReaderFT0, MCLabelCont>::getFITDigitWriterSpec(false, false, dataOrigin));
+      }
     }
   } else {
-    specs.emplace_back(o2::fit::getFITDataReaderDPLSpec(RawReaderFT0{dataOrigin, dumpReader}, askSTFDist));
-    if (!disableRootOut) {
-      specs.emplace_back(o2::fit::FITDigitWriterSpecHelper<RawReaderFT0, MCLabelCont>::getFITDigitWriterSpec(false, false, dataOrigin));
-    }
+#if defined(__has_include)
```
shahor02, line 109: perhaps you should not offer the option `new-decoder` or at least print some error message when you are not in the correct compiler environment?

shahor02 (commented):
Is the output supposed to be the same as with the old decoder? 
Particularly, does it work with the current CTF version?

Detectors/FIT/FT0/workflow/src/ft0-flp-workflow.cxx
afurs: > Is the output supposed to be the same as with the old decoder?
> Particularly, does it work with the current CTF version?

Yes, output is the same as from old decoder, except Trigger inputs and Extended mode for TCM(specific mode which is considered to use time to time, is implemented to old decoder). Unfortunately I didn't check it with current CTF version yet. But I checked with other devices(Digit writer), and this decoder has sorted by BC-orbit IDs digits within output container. And actually I think it will not work correctly with current CTF version because ChIDs is unsorted(we discussed it). In parallel I'm preparing and testing such PR for CTFCoder.

> perhaps you should not offer the option `new-decoder` or at least print some error message when you are not in the correct compiler environment?

Yes, I'm planning to implement this in next patch(+ some other kind of protections within new decoder, in TODO list). But first we would like to make stress tests for this decoder at the FLP next week. Just to be sure that it can handle with highest rates in real life(I tested it only by using emulated readout.exe->StfB->DPL->StfS chain w/o ECS at FLP machine) And also I was thinking about implementation of runtime check for AVX512(compilation by using compiler which supports avx512 is only for FLPsuite package preparation, but running machine code on machines which are not support avx512 is another story)

PR #8779, opened 2022-05-10
DPL: revert serialization of DataDescriptorMatcher

ktf:


ktf:
@matthiasrichter it looks like your code breaks the serialization of InputSpec with a ConcreteDataTypeMatcher. The test I add in this PR shows the issue in a simple manner. I think you either are not stopping early enough when serialising, or you simply should add the case in which you handle a matcher with a context ref inside. I am now trying to actually fix the problem but maybe we should just revert this for now.

ktf:
Ok, merging this since we are getting close to nightly time. I think I actually have a proper fix in #8780, but besides the unit test I have not checked it.

PR #8780, opened 2022-05-10
DPL: fix serialisation context ref

ktf:


ktf:
Besides having all the tests passing, I also verified that @shahor02 reproducer works as expected.

```
o2-its-cluster-reader-workflow --hbfutils-config none | o2-its-noise-calib-workflow --use-clusters | o2-calibration-ccdb-populator-workflow --ccdb-path http://ccdb-test.cern.ch:8080
```

actually produced:

http://ccdb-test.cern.ch:8080/browse/ITS/Calib/NoiseMap?report=true

correctly.

Merging.

PR #8737, opened 2022-05-06
MCH: make reco more parametrizable

aphecetche:
@aferrero2707 @pillot 
I've added a `DigitFilterParam` and `TimeClusterizerParam` in order to ease the configuration of those components from reco scripts. Could you please have a look ?

PR #8781, opened 2022-05-11
MCH: add digit noise filtering options

aphecetche:
This is a follow-up of #8737 

We introduce two digit filtering options : 

- rejectBackground : to kill background while not touching the signal (i.e. loose background rejection to ensure signal is not touched)
- selectSignal : kill as much background as possible, even at the cost of loosing some signal 

Both cuts were tuned by @pillot on the (oct 21) pilot beam data.

The first option (rejectBackground) is intended to be used at the digit filtering stage and is relatively safe. 

The second one should be used with more **caution** : the idea is to use it in the (@aferrero2707 ) time clustering (in a to-be-done PR) just to help it pick the region of interest more easily, but the set of digits that are then passed to the (pre)clustering must remain the input one (in particular low charge digits must *not* be removed prior to clustering, otherwise the clustering resolution will suffer).

### DIFF
--- a/Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
+++ b/Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
@@ -48,7 +48,7 @@ double backgroundCut(double* x, const double* p)
   }
 }
 
-o2::mch::DigitFilter createMinAdcCut(int minADC)
+o2::mch::DigitFilter createMinAdcCut(uint32_t minADC)
 {
   return [minADC](const o2::mch::Digit& digit) -> bool {
     if (digit.getADC() < minADC) {
@@ -60,7 +60,7 @@ o2::mch::DigitFilter createMinAdcCut(int minADC)
 
 o2::mch::DigitFilter createRejectBackground()
 {
-  int minNSamplesBackground = 14;
+  uint16_t minNSamplesBackground = 14;
   double backgroundParam[4] = {18., 24., -20., 7.0};
 
   auto backgroundCut = [backgroundParam](double* x) {
@@ -78,7 +78,7 @@ o2::mch::DigitFilter createRejectBackground()
 
 o2::mch::DigitFilter createSelectSignal()
 {
-  int minNSamplesSignal = 17;
+  uint16_t minNSamplesSignal = 17;
   double signalParam[4] = {80., 16., 12., 1.2};
 
   auto signalCut = [signalParam](double* x) {
--- a/Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
+++ b/Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
@@ -96,11 +96,13 @@ o2::mch::DigitFilter createSelectSignal()
 
 namespace o2::mch
 {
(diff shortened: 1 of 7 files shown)


Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
```diff
@@ -0,0 +1,120 @@
...
+    if (digit.getADC() < minADC) {
+      return false;
+    }
+    return true;
+  };
+}
+
+o2::mch::DigitFilter createRejectBackground()
+{
+  int minNSamplesBackground = 14;
```
pillot: nSample is a uint16_t. Won't the comparison issue a warning?

Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
```diff
@@ -0,0 +1,120 @@
...
+    if (digit.getNofSamples() < minNSamplesBackground || digit.getADC() < backgroundCut(&nSample)) {
+      return false;
+    }
+    return true;
+  };
+}
+
+o2::mch::DigitFilter createSelectSignal()
+{
+  int minNSamplesSignal = 17;
```
pillot: same question as above

Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
```diff
@@ -0,0 +1,120 @@
...
+}
+} // namespace
+
+namespace o2::mch
+{
+DigitFilter createDigitFilter(int minADC, bool rejectBackground, bool selectSignal)
+{
+  std::vector<DigitFilter> parts;
+
+  parts.emplace_back(createMinAdcCut(minADC));
```
pillot: minADC is a uint32_t

Maybe the addition of this selection should be conditioned to minADC>0, so it not applied to every digits if it is not expected to reject any. I am thinking about the usage of the filtering in the time clustering, where (a priori) we only want to apply the signal selection.

Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
aphecetche: good point. will change.

Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
aphecetche: same answer as above :smile:

Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
```diff
@@ -0,0 +1,122 @@
...
+  }
+}
+
+/** function used to reject the background.
+ * might not kill all background, focus here is ensure
+ * we're not killing some signal along the way.
+ */
+double backgroundCut(double* x, const double* p)
+{
+
```
aferrero2707, line 42: There is a spurious space here...

aferrero2707 (approved):

aferrero2707:
Ok for me to merge.

aphecetche:
@shahor02 could you please merge this one ? thanks.

PR #8783, opened 2022-05-11
DPL Analysis: simple marker column template

aalkin:
@jgrosseo as we have discussed, this is a simple service column, that can be added to a table's column list to make it distinguishable. Provides `.mark()` method, that returns the number set in the table definition (as `o2::soa::Marker<#>`)


jgrosseo:
Very nice. Thanks!

jgrosseo (approved):

PR #8786, opened 2022-05-11
Adding missing resets of tables used for filling of fwdtrack and MFT tracks

mcoquet642:


mcoquet642:
Hi @shahor02, could you please merge this if this is ok for you ? I think this should be fixed as soon as possible.

shahor02 (approved):

PR #8788, opened 2022-05-11
Fix: MFT tracking must update ROF also for empty ROFs

shahor02:


rpezzi (approved):
Thanks for spotting this, @shahor02 .
Empty ROFs had an uninitialised value for `firstROFTrackEntry`. Did you spot any problem somewhere else in the reconstruction due to this?

shahor02:
> Thanks for spotting this, @shahor02 . Empty ROFs had an uninitialised value for `firstROFTrackEntry`. Did you spot any problem somewhere else in the reconstruction due to this?

Actually, the values for the `mFirstEntry` and `mNEntries` were copied from the clusters ROF vector (when making `pc.outputs().make` and were not overridden if the tracking was not done due to the  nClusters = 0. 
The track-vertex matcher was failing to extract non-existing MFT tracks mentioned in the ROFRecord vector.

PR #8790, opened 2022-05-11
proper parsing of VERSIONED tables

pbuehler:
. improved recognition of header files


pbuehler:
@jgrosseo An improved version of the data model parser

PR #8791, opened 2022-05-11
Fixes in MID raw data encoder

dstocco:
This PR fixes some issues in the MID raw data encoder. In particular it fixes the treatment of the delays between the HB trigger and the data.
Also, the regional response to CTP triggers was not correct for the GBT link in the regional board that only reads 1 LOC.
Finally, the last commit ensures that we always generate raw data in simulation, also in the case where we have no muon in the TF.

### DIFF
--- a/Detectors/MUON/MID/Raw/include/MIDRaw/Encoder.h
+++ b/Detectors/MUON/MID/Raw/include/MIDRaw/Encoder.h
@@ -54,10 +54,12 @@ class Encoder
 
  private:
   void completeWord(std::vector<char>& buffer);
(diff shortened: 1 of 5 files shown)


aphecetche (approved):

dstocco:
@sawenzel @shahor02 : is this PR ok for you?

shahor02 (approved):

dstocco:
@sawenzel : is this PR ok for you?

dstocco:
@sawenzel @shahor02 : can this be merged?

PR #8793, opened 2022-05-11
DPL Analysis: Remove backward compability from combinations

saganatt:
No use of string column names in block policies. BinningPolicy becomes obligatory.

### DIFF
--- a/Framework/Core/include/Framework/ASoAHelpers.h
+++ b/Framework/Core/include/Framework/ASoAHelpers.h
@@ -72,85 +72,8 @@ inline bool diffCategory(BinningIndex const& a, BinningIndex const& b)
   return a.bin >= b.bin;
 }
 
-template <typename T2, typename ARRAY, typename T>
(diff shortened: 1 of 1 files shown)


ktf (changes requested):
Could you check the error? it seems genuine, although I do not understand how it could be cc8 specific.

saganatt:
It seems cs8 used an older version of O2Physics. The error is caused by using old version of combinations with string.
But I removed that in https://github.com/AliceO2Group/O2Physics/pull/733/

ktf:
Ah, ok, maybe the test started before I updated the tag. Let me restart it.

jgrosseo:
> Ah, ok, maybe the test started before I updated the tag. Let me restart it.

cs8 does not build against master. see https://mattermost.web.cern.ch/alice/pl/5yeqd6wtzidyxrzh53w6b3qmea

ktf:
It should build now, since I merged the PR for the nightly build.

saganatt:
It passed cs8 check, could you merge, please? :-)

jgrosseo (approved):

PR #8794, opened 2022-05-11
Synchronize MID data orbit with RDH orbit at TF limit

dstocco:
The CRU UL ensures that all data coming before the answer to an orbit trigger are flushed only at the TF limit.
There is no such check within a TF.
The decoder is modified accordingly. At the beginning of the TF we can trust that the data belong to the orbit written in the RDH.
Within a TF, on the other hand, we increment the orbit number based on the information in the payload.

### DIFF
--- a/Detectors/MUON/MID/Raw/src/ELinkDataShaper.cxx
+++ b/Detectors/MUON/MID/Raw/src/ELinkDataShaper.cxx
@@ -17,7 +17,7 @@
 #include "MIDRaw/ELinkDataShaper.h"
 
 #include "CommonConstants/LHCConstants.h"
-// #define ORBITFROMMIDRO
+#include "CommonConstants/Triggers.h"
 
 namespace o2
 {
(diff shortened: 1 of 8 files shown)


Detectors/MUON/MID/Raw/src/ELinkDataShaper.cxx
```diff
@@ -43,27 +43,33 @@ ELinkDataShaper::ELinkDataShaper(bool isDebugMode, bool isLoc, uint8_t uniqueId,
...
     mIR.bc = 0;
-// The reset changes depending on the way we synch with the orbit
-// (see processOrbitTrigger for details)
-// FIXME: pick one of the two
-#ifdef ORBITFROMMIDRO
-    mIR.orbit = orbit - 1; // with orbit increase
-#else
-    mIR.orbit = orbit;   // with reset to RDH
-#endif
+    mIR.orbit = orbit - 1;
```
aphecetche, line 57: what happens in the boundary case where orbit=0 ? Is that guaranteed to never happen in this part of the code ?

dstocco: This code is called only the first time the method is called, which typically means at SOR. In any case, it will be for the first orbit of the first TF processed. The CRU UL ensures that this will start with an answer to the HB trigger. This means that the first method that will be called is: processHBTrigger() where we have ++mIR.orbit. So this initialisation is actually consistent with what I do at HB trigger, meaning that we should never really have orbit - 1 in the end.

aphecetche (approved):

PR #8801, opened 2022-05-11
Fix unsigned DPL option in TPC raw2digits

martenole:
Hi @wiechula .
I opened this to possibly get it in the next nightly.
Unfortunately, DPL seems to be **very** sensitive with respect to the default value type. I could not even manually set the parameter, because then I still get
```
[marten@epn155 ~]$ o2-tpc-raw-to-digits-workflow -b --sync-offset-reference 144u
[ERROR] invalid workflow in o2-tpc-raw-to-digits-workflow: Mismatch between declared option type (19) and default value type (0) for sync-offset-reference in DataProcessorSpec of tpc-raw-to-digits-0
```


martenole:
Superseded by #8802

PR #8802, opened 2022-05-11
TPC: fix parameter type for sync-offset-reference

wiechula:


wiechula:
@davidrohr , since this is a trivial fix, can this be merged right away?

shahor02 (approved):
trivial, merging

PR #8810, opened 2022-05-12
[EMCAL-651] Don't initialize geometry in the ClusterFactory

mfasDa:
Calling GetInstanceFromRunNumber
in header leads to the geometry
being constructed every time a
ClusterFactory is constructed.
Instead the Geometry should be
handled outside, and ClusterFactory
should only throw a error in case the
geometry is not initialized.

### DIFF
--- a/Detectors/EMCAL/base/src/ClusterFactory.cxx
+++ b/Detectors/EMCAL/base/src/ClusterFactory.cxx
@@ -25,9 +25,17 @@
 
 using namespace o2::emcal;
 
+template <typename InputType>
+ClusterFactory<InputType>::ClusterFactory()
+{
+  mGeomPtr = o2::emcal::Geometry::GetInstance();
+}
+
 template <class InputType>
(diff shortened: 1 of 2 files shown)


Detectors/EMCAL/base/src/ClusterFactory.cxx
```diff
@@ -50,6 +58,8 @@ o2::emcal::AnalysisCluster ClusterFactory<InputType>::buildCluster(int clusterIn
   if (clusterIndex >= mClustersContainer.size()) {
     throw ClusterRangeException(clusterIndex, mClustersContainer.size());
   }
+  if (!mGeomPtr)
+    throw GeometryNotSetException();
```
sawenzel, line 62: we need to use braces for if-statements (see full-CI).

Secondly, I think we prefer (previous discussion) the simpler variant of using LOG(fatal) for these kinds of problems unless the exception can actually be handled in a good way. It is also a lot less code.

mfasDa: Yes, in many case they can, i.e. the task can (and should) be delegated to an outer component. Therefore I would prefer keeping the exception.

I will fix the braces.

shahor02 (approved):

PR #8812, opened 2022-05-12
Allow proper parsing of VERSIONED tables

pbuehler:


ktf:
It looks like there is a conflict. Could you please fix it?

PR #8814, opened 2022-05-12
Option to ignore object validity check/refetching if cache is available

shahor02:
When used with the DPL CCDB fetcher (i.e. loadFileToMemory is called) in order to prefer the available
snapshot w/o its validity check an extra variable `IGNORE_VALIDITYCHECK_OF_CCDB_LOCALCACHE` must be defined
otherwhise the object will be fetched from the server after the validity check and new snapshot will be created if needed.
`IGNORE_VALIDITYCHECK_OF_CCDB_LOCALCACHE` should be always accompanied by the `ALICEO2_CCDB_LOCALCACHE` env.var indicating the snapshot directory (if the directory of particular snapshots are absent, they will be populated at 1st call).

@sawenzel

### DIFF
--- a/CCDB/include/CCDB/CcdbApi.h
+++ b/CCDB/include/CCDB/CcdbApi.h
@@ -516,7 +516,9 @@ class CcdbApi //: public DatabaseInterface
   std::string mUniqueAgentID{}; // Unique User-Agent ID communicated to server for logging
   std::string mUrl{};
   std::vector<std::string> hostsPool{};
-  std::string mSnapshotTopPath{};
+  std::string mSnapshotTopPath{};    // root of the snaphot in the snapshot backend mode, i.e. with init("file://<dir>) call
+  std::string mSnapshotCachePath{};  // root of the local snapshot (to fill or impose, even if not in the snapshot backend mode)
+  bool mPreferSnapshotCache = false; // if snapshot is available, don't try to query its validity even in non-snapshot backend mode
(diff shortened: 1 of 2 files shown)


sawenzel:
It appears to be working. Although, the logging to `ALICEO2_CCDB_LOCALCACHE/log` seems no longer happening.
Before, I could see who is reading from the cache.

shahor02:
Do you want to create a semaphore just to update the log file with the info on reading? Why not simply producing a `LOG(info)`, probably conditioned by the env.var?
BTW, the semaphore you use associated with the object path being updated and the log updater is in the same scope. This will not protect against the simultaneous log file update if two concurrent processes create semaphores for 2 different objects, right?

sawenzel:
I am just asking about features/behaviour which were there before and on which I was relying. The technical implementation is a different matter.

sawenzel:
So I take it you no longer put a semaphore at all? So what happens if 2 processes download the same object at the same time?

shahor02:
The semaphore is there for storing the object, reading does not need it.
As I wrote, I am not against the logging, but why do you need it to go to separate file which would cost a dedicated semaphore?

ktf:
why using a semaphore? moving files is guaranteed to be atomic. can't you just download the file somewhere unique and move it in place?

shahor02:
@sawenzel shall I merge this?

sawenzel:
Ok for me.

PR #8817, opened 2022-05-13
Utility for detector run status check in DCS processing + EMC example

shahor02:
Utility class to check the status of the run with particular detectors participating. 
Usage: 
```
  // first create an instance for the mask of detectors which must be in the run, e.g.
  RunStatusChecker runChecker{ o2::detectors::DetID::getMask("EMC,PHS") };
  const o2::parameters::GRPECSObject* myGRPECS = nullptr;

  // then, periodically check:
  myGRPECS = runChecker.check();
  auto dets = o2::detectors::DetID::getNames(checker->getDetectorsMask());
  using RunStatus = o2::dcs::RunStatusChecker::RunStatus;

  // the check will set the status of the run with selected detectors (can be inspected by getRunStatus() method).
  if (check.getRunStatus() == RunStatus::NONE) {
    LOGP(info, "No run with {} is ongoing or finished", dets);
  }
  else if (check.getRunStatus() == RunStatus::START) { // saw new run with wanted detectors
    LOGP(info, "Run {} with {} has started", checker.getFollowedRun(),  dets);
  }
  else if (check.getRunStatus() == RunStatus::ONGOING) { // run which was already seen is still ongoing
    LOGP(info, "Run {} with {} is still ongoing", checker.getFollowedRun(), dets);
  }
  else if (check.getRunStatus() == RunStatus::STOP) { // run which was already seen was stopped (EOR seen)
    LOGP(info, "Run {} with {} was stopped", checker.getFollowedRun(), dets);
  }
  // In all cases except RunStatus::NONE a const non-null pointer on the GRP of the followed run will be returned.
  // By default the check will be done for the current timestamp, for test purposes, one can call it with arbitrary increasing timestamps
```

@chiarazampolli @gvolpe79

### DIFF
--- a/Detectors/DCS/CMakeLists.txt
+++ b/Detectors/DCS/CMakeLists.txt
@@ -24,6 +24,7 @@ o2_add_library(
           src/GenericFunctions.cxx
           src/StringUtils.cxx
           src/Clock.cxx
+   src/RunStatusChecker.cxx
   PUBLIC_LINK_LIBRARIES O2::Headers O2::CommonUtils O2::CCDB
                         O2::DetectorsCalibration Microsoft.GSL::GSL)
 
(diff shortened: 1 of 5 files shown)


chiarazampolli:
Thanks @shahor02 !

shahor02:
FullCI failure unrelated (sim_challenge), merging

PR #8819, opened 2022-05-13
Fix Fwd and MFT MC label filling

benedikt-voelkel:


benedikt-voelkel:
@mcoquet642 @shahor02 

MFT and Fwd MC track labels were not filled anymore after https://github.com/AliceO2Group/AliceO2/pull/8786 ash been merged.

Added filling of `mGIDToTableMFTID` and `mGIDToTableFwdID` to `fillTrackTablesPerCollision` because right now they just stay empty after the call of `fillIndexTablesPerCollision` and so no mc track labels are filled for MFT and Fwd.

This might need more refinement in the future because now the above mentioned maps are filled twice.

Without this PR the MFT/Fwd tracks and associated MC labels simply don't match and the latter stay empty

mcoquet642:
Thanks @benedikt-voelkel indeed if we reset the maps we have to fill them again. The lines you added indeed correspond to the wrongly commented lines [for MFT](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L291) and [Fwd tracks](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L605).

mcoquet642:
Thanks @benedikt-voelkel indeed if we reset the maps we have to fill them again. The lines you added indeed correspond to the wrongly commented lines [for MFT](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L291) and [Fwd tracks](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L605).

sawenzel (approved):

PR #8826, opened 2022-05-13
DPL Analysis: Event mixing: Add a generic getter to dynamic columns, support dynamic in mixing

saganatt:
Ciao @ktf, @jgrosseo, @aalkin 

the iteration on underlying arrow array works only with persistent columns, so the users could not use, e.g., multiplicity for event mixing.

I added also a bit of support for index columns in the binning policy, though I am not sure we would ever use index columns in making bins for block combinations / mixing...?

### DIFF
--- a/Framework/Core/include/Framework/BinningPolicy.h
+++ b/Framework/Core/include/Framework/BinningPolicy.h
@@ -21,62 +21,6 @@
 namespace o2::framework
 {
 
-namespace binning_helpers
-{
-template <typename... Cs>
-std::array<arrow::ChunkedArray*, sizeof...(Cs)> getArrowColumns(arrow::Table* table, pack<Cs...>)
-{
-  static_assert(std::conjunction_v<typename Cs::persistent...>, "BinningPolicy: only persistent columns accepted (not dynamic and not index ones");
-  return std::array<arrow::ChunkedArray*, sizeof...(Cs)>{o2::soa::getIndexFromLabel(table, Cs::columnLabel())...};
-}
-
-template <typename... Cs>
-std::array<std::shared_ptr<arrow::Array>, sizeof...(Cs)> getChunks(arrow::Table* table, pack<Cs...>, uint64_t ci)
-{
-  static_assert(std::conjunction_v<typename Cs::persistent...>, "BinningPolicy: only persistent columns accepted (not dynamic and not index ones");
-  return std::array<std::shared_ptr<arrow::Array>, sizeof...(Cs)>{o2::soa::getIndexFromLabel(table, Cs::columnLabel())->chunk(ci)...};
-}
-
-template <typename C>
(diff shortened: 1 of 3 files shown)


jgrosseo:
I don't think index columns are ever needed for mixing because the index value itself can change in merging.

This PR is fine for me. @ktf @aalkin any comments?

Framework/Core/include/Framework/BinningPolicy.h
```diff
@@ -37,11 +37,43 @@ std::array<std::shared_ptr<arrow::Array>, sizeof...(Cs)> getChunks(arrow::Table*
...
+template <typename C>
+typename C::type getSingleRowPersistentData(arrow::Table* table, uint64_t ci, uint64_t ai)
 {
-  static_assert(std::conjunction_v<typename Cs::persistent...>, "BinningPolicy: only persistent columns accepted (not dynamic and not index ones");
-  return std::make_tuple(std::static_pointer_cast<o2::soa::arrow_array_for_t<typename Cs::type>>(o2::soa::getIndexFromLabel(table, Cs::columnLabel())->chunk(ci))->raw_values()[ai]...);
+  return std::static_pointer_cast<o2::soa::arrow_array_for_t<typename C::type>>(o2::soa::getIndexFromLabel(table, C::columnLabel())->chunk(ci))->raw_values()[ai];
+}
+
+template <typename T, typename C>
+typename C::type getSingleRowDynamicData(T& rowIterator, uint64_t globalIndex)
```
aalkin: This is not universal - there are dynamic columns that require free arguments and attempting to call them will not compile. Probably we can add a dependent type or constexpr int in the dynamic column definition that determines if this particular column needs extra arguments. Then we'll be able to exclude it from this call.

jgrosseo: Isn't it OK if in the case someone requests such a column, the code doesn't compile?

jgrosseo: I think it is OK if we do not support those cases for mixing (people can always create a transient table if those columns should really be needed for mixing)

aalkin: This code is useful not only for mixing, getting the content from the whole row can also be used for skimming or for feeding the table to a model. I agree, however, that It would make sense to have a special case for columns with arguments and not just skip them.

saganatt: Well, if this code would be used also outside mixing, then I can move these functions to ASoA.h, for example.
Then, in another PR one could add the support for columns with arguments.

jgrosseo: I agree

saganatt: OK, so I moved the codes, let me know @aalkin if you'd prefer to see them in another place in ASoA.h

jgrosseo (approved):

PR #8829, opened 2022-05-14
Options and fixes for use of CCDB in LHCphase and channel calib 

chiarazampolli:
To be checked (@noferini ):
* for the channel calib, did we set the slots boundaries according to the data? I could not find it if it is so. So I changed the time used for the CCDB entry to use the one in MS from the slot;
* we need to add some settings for the LHCphase and slewing object to the api, to avoid destroy and recreate (also for the clusterer)

---> for the first point above: the data was used to define the timestamp for the LHCphase object read by the calibTOFapi --> now changed, it uses DPL timestamp and OrbitReset
---> setters implemented, but for now the clusterer is not touched.

### DIFF
--- a/Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
+++ b/Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
@@ -413,7 +413,7 @@ void TOFChannelData::resetAndReRange(float range)
   mV2Bin = mNBins / (2 * mRange);
   for (int isect = 0; isect < 18; isect++) {
     mHisto[isect] = boost::histogram::make_histogram(boost::histogram::axis::regular<>(mNBins, -mRange, mRange, "t-texp"),
-						   boost::histogram::axis::integer<>(0, mNElsPerSector, "channel index in sector" + std::to_string(isect)));
+                                                     boost::histogram::axis::integer<>(0, mNElsPerSector, "channel index in sector" + std::to_string(isect)));
   }
   return;
 }
--- a/Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
+++ b/Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
@@ -403,6 +403,21 @@ float TOFChannelData::integral(int ch) const
   return mEntries.at(ch);
 }
 
+//_____________________________________________
+void TOFChannelData::resetAndReRange(float range)
+{
+  // empty the container and redefine the range
+
+  setRange(range);
+  std::fill(mEntries.begin(), mEntries.end(), 0);
+  mV2Bin = mNBins / (2 * mRange);
+  for (int isect = 0; isect < 18; isect++) {
+    mHisto[isect] = boost::histogram::make_histogram(boost::histogram::axis::regular<>(mNBins, -mRange, mRange, "t-texp"),
+						   boost::histogram::axis::integer<>(0, mNElsPerSector, "channel index in sector" + std::to_string(isect)));
+  }
+  return;
+}
+
 //-------------------------------------------------------------------
 // TOF Channel Calibrator
 //-------------------------------------------------------------------
--- a/Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
+++ b/Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
@@ -609,7 +609,11 @@ void TOFChannelCalibrator<T>::finalizeSlotWithCosmics(Slot& slot)
 
   auto clName = o2::utils::MemFileHelper::getClassName(ts);
   auto flName = o2::ccdb::CcdbApi::generateFileName(clName);
-  mInfoVector.emplace_back("TOF/Calib/ChannelCalib", clName, flName, md, slot.getTFStart(), o2::ccdb::CcdbObjectInfo::INFINITE_TIMESTAMP);
+  auto startValidity = slot.getStartTimeMS();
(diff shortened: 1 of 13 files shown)


Detectors/TOF/calibration/testWorkflow/TOFChannelCalibratorSpec.h
```diff
@@ -153,16 +141,20 @@ class TOFChannelCalibDevice : public o2::framework::Task
...
+      // startTimeChCalib = pc.inputs().get<long>("startTimeChCalib");
     }
 
     LOG(debug) << "startTimeLHCphase = " << startTimeLHCphase << ",  startTimeChCalib = " << startTimeChCalib;
 
-    mcalibTOFapi = new o2::tof::CalibTOFapi(long(0), &mPhase, &mTimeSlewing); // TODO: should we replace long(0) with tfcounter defined at the beginning of the method? we need the timestamp of the TF
+    if (!mcalibTOFapi) {
+      mcalibTOFapi = new o2::tof::CalibTOFapi(long(0), &mPhase, &mTimeSlewing); // TODO: should we replace long(0) with tfcounter defined at the beginning of the method? we need the timestamp of the TF
+    } else {
+      if (mUseCCDB && mUpdateCCDB) {
```
shahor02, line 161: mUpdateCCDB is reset and set in the finaliseCCDB for each object, here you don't know which one was updated, is this what you want (trigger on any update)?

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
```diff
@@ -609,7 +609,7 @@ void TOFChannelCalibrator<T>::finalizeSlotWithCosmics(Slot& slot)
 
   auto clName = o2::utils::MemFileHelper::getClassName(ts);
   auto flName = o2::ccdb::CcdbApi::generateFileName(clName);
-  mInfoVector.emplace_back("TOF/Calib/ChannelCalib", clName, flName, md, slot.getTFStart(), o2::ccdb::CcdbObjectInfo::INFINITE_TIMESTAMP);
+  mInfoVector.emplace_back("TOF/Calib/ChannelCalib", clName, flName, md, slot.getStartTimeMS(), o2::ccdb::CcdbObjectInfo::INFINITE_TIMESTAMP);
```
shahor02: does this really need INFINITE_TIMESTAMP ?

Detectors/TOF/calibration/testWorkflow/TOFChannelCalibratorSpec.h
chiarazampolli: Yes, as soon as one changes, we need to update mcalibTOFapi. 
This, though, is probably not the best way - deleting and getting a new pointer. I will implement the setters.

Detectors/TOF/calibration/src/TOFChannelCalibrator.cxx
chiarazampolli: Yes, I know. The problem are long TS, also over Xmas. We don't want to restart from scratch. So either we set here 1 month (it should be enough for TSs, right? unless something breaks dramatically), and then do a manual upload when data taking restarts (but we need to remember it!), or better this limit. Even if it will not be convenient...
Maybe we could use 1 year?

shahor02: If this is frequently updated object, then even 1 year will create lot of overlaps (and if it indeed changes so often, is 1 year old object any better than the nominal default one?). I would consider 1-2 months margin at most, and if needed, would extend the validity of last (non-default) entry, perhaps we should have a generic tool for that, which check a list of objects for which such operation makes sense.

chiarazampolli: Now 2 months are used.

chiarazampolli:
@noferini : tomorrow I will test everything thoroughly, then if all is fine, we can remove the [WIP] and proceed with the review.

noferini (approved):
It looks fine with me

chiarazampolli:
Dear @shahor02 , whenever you have some time, could you review this one?
Thanks!

Detectors/TOF/calibration/testWorkflow/LHCClockCalibratorSpec.h
```diff
@@ -54,39 +57,69 @@ class LHCClockCalibDevice : public o2::framework::Task
...
-    for (int ich = 0; ich < TimeSlewing::NCHANNELS; ich++) {
-      mTimeSlewing.addTimeSlewingInfo(ich, 0, 0);
-      int sector = ich / TimeSlewing::NCHANNELXSECTOR;
-      int channelInSector = ich % TimeSlewing::NCHANNELXSECTOR;
-      mTimeSlewing.setFractionUnderPeak(sector, channelInSector, 1);
+    if (!mUseCCDB) {
+      // calibration objects set to zero
+      mPhase.addLHCphase(0, 0);
+      mPhase.addLHCphase(o2::ccdb::CcdbObjectInfo::INFINITE_TIMESTAMP_SECONDS, 0);
+      if (gSystem->AccessPathName("localTimeSlewing.root") == false) {
```
shahor02, line 64: where this local file supposed to come from?

shahor02 (approved):
Just 1 question below.

Detectors/TOF/calibration/testWorkflow/LHCClockCalibratorSpec.h
chiarazampolli: It was used for local testing only.

PR #8833, opened 2022-05-15
MCH: some refactoring (mainly moving things around)

aphecetche:
This commit does not introduce any new feature, but just reorganizes a
bit where some files are located. In particular :

- the TrackerParam is moved to MCHBase to be useable from e.g.
  TimeClustering w/o requiring TimeClustering to depend on MCHTracking (@pillot)
- the ROF filtering used in TimeClustering (based on the trackable condition)
is moved to its own module (in preparation for the addition of a new ROF
filter based on IRFrames soon, @aferrero2707)

(note that this PR contributes to the goal of making the Workflow directory leaner)


Detectors/MUON/MCH/TimeClustering/README.md
```diff
@@ -0,0 +1,24 @@
...
+```shell
+o2-mch-digits-to-timeclusters-workflow
+```
+
+Take as input the list of all digits ([Digit](/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/Digit.h)) in the current time frame, with the data description "DIGITS", and the list of ROF records ([ROFRecord](../../../../DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/ROFRecord.h)) pointing to the digits associated to each interaction, with the data description "DIGITROFS". Send a new list of ROF records that combine all the digits correlated in time within a user-defined time window, with the data description "TIMECLUSTERROFS".
+
+The option `--max-cluster-width xxx` allows to set the width of the time correlation window.
+
+The time clustering is based on a brute-force peak search algorithm, which arranges the input digits into coarse time bins. The number of bins in one time cluster window can be set via the `--peak-search-nbins` option.
+
```
pillot: These options no longer exist. You may profit from this PR to update the doc or do it later. As you prefer.

Detectors/MUON/MCH/Base/include/MCHBase/Trackable.h
```diff
@@ -56,38 +56,6 @@ std::array<int, 5> perStation(gsl::span<const T> items)
   return st;
```
pillot, line 55: Should ++i in the loop and st[i] = ch[2 * i] + ch[2 * i + 1];

Detectors/MUON/MCH/Base/include/MCHBase/Trackable.h
```diff
@@ -20,27 +20,27 @@
 namespace o2::mch
```
pillot, line 19: you can remove #include "DataFormatsMCH/ROFRecord.h"
also Cluster.h is not needed (yet) in the cxx.

Detectors/MUON/MCH/ROFFiltering/include/MCHROFFiltering/TrackableFilter.h
```diff
@@ -0,0 +1,58 @@
...
+  createTrackableFilter(gsl::span<const T> items, bool onlyTrackable,
+                        std::array<bool, 5> requestStation = {true, true, true, true, true},
+                        bool moreCandidates = false)
+{
+  if (!onlyTrackable) {
+    // return an "always true" filter
+    return [](const ROFRecord&) {
+      return true;
+    };
+  }
```
pillot, line 49: Why would one like to create a filter that do not filter ? ;-). This filter is also pretty specific (according to its name) so it's hard to imagine that it could perform other type of filtering.

Detectors/MUON/MCH/Base/include/MCHBase/Trackable.h
aphecetche: removed.

Detectors/MUON/MCH/Base/include/MCHBase/Trackable.h
aphecetche: good catch. thanks.

Detectors/MUON/MCH/ROFFiltering/include/MCHROFFiltering/TrackableFilter.h
aphecetche: agreed it's a bit convoluted. Might change it in a future PR.

aferrero2707:
Ok for me with the changes proposed by @pillot.

pillot (approved):

PR #8835, opened 2022-05-16
Adding owners of data model documentation

jgrosseo:
@ktf @sawenzel Any objections?


ktf:
No objections but I need @pbuehler to be added to the writers for this to be effective. 1 sec.

PR #8837, opened 2022-05-16
CPV: add gain calibrator

sevdokim:
Hello!

cpv gain calibrator created. It picks up CALIBDIGITS produced by Clusterizer, prepares GainCalibData (which is struct representing 23040 histos with 1000 bins each, bin content is stored as uint32_t) and produces CalibParams object with gains when collected statistics is enough. GainCalibData is saved to CCDB at the end of each run and is read read back from CCDB when new run starts. When statistics is enough to calibrate 10000 channels then calibration is produces for ready channels, statistics in those is resetted. In channels which were not ready, statistics continues to be collected. 

Also several warnings for cpv are fixed.

### DIFF
--- a/DataFormats/Detectors/CPV/src/Hit.cxx
+++ b/DataFormats/Detectors/CPV/src/Hit.cxx
@@ -43,14 +43,21 @@ Hit& Hit::operator+=(const Hit& rhs)
   return *this;
 }
 
-Hit Hit::operator+(const Hit& rhs) const
+// Hit Hit::operator+(const Hit& rhs) const
+// {
+//   Hit result(*this);
+//   if (rhs.GetEnergyLoss() > result.GetEnergyLoss()) {
+//     result.SetTime(rhs.GetTime());
+//   }
+//   result.SetEnergyLoss(result.GetEnergyLoss() + rhs.GetEnergyLoss());
+//   return result;
+// }
+Hit operator+(const Hit& lhs, const Hit& rhs)
 {
-  Hit result(*this);
-  if (rhs.GetEnergyLoss() > result.GetEnergyLoss()) {
-    result.SetTime(rhs.GetTime());
-  }
-  result.SetEnergyLoss(result.GetEnergyLoss() + rhs.GetEnergyLoss());
-  return result;
+  // Hit::Hit(int trackID, int detID, const math_utils::Point3D<float>& pos, double tof, double qLoss)
+  return Hit(lhs.GetTrackID(), lhs.GetDetectorID(), lhs.GetPos(),
+             lhs.GetEnergyLoss() > rhs.GetEnergyLoss() ? lhs.GetTime() : rhs.GetTime(),
+             lhs.GetEnergyLoss() + rhs.GetEnergyLoss());
 }
 
 std::ostream& operator<<(std::ostream& stream, const Hit& p)
--- a/DataFormats/Detectors/CPV/src/Hit.cxx
+++ b/DataFormats/Detectors/CPV/src/Hit.cxx
@@ -43,15 +43,6 @@ Hit& Hit::operator+=(const Hit& rhs)
   return *this;
 }
 
-// Hit Hit::operator+(const Hit& rhs) const
-// {
-//   Hit result(*this);
-//   if (rhs.GetEnergyLoss() > result.GetEnergyLoss()) {
-//     result.SetTime(rhs.GetTime());
-//   }
-//   result.SetEnergyLoss(result.GetEnergyLoss() + rhs.GetEnergyLoss());
-//   return result;
-// }
 Hit operator+(const Hit& lhs, const Hit& rhs)
 {
   // Hit::Hit(int trackID, int detID, const math_utils::Point3D<float>& pos, double tof, double qLoss)
--- a/Detectors/CPV/base/include/CPVBase/CPVCalibParams.h
+++ b/Detectors/CPV/base/include/CPVBase/CPVCalibParams.h
@@ -25,34 +25,44 @@ namespace cpv
 struct CPVCalibParams : public o2::conf::ConfigurableParamHelper<CPVCalibParams> {
 
   // Parameters used in pedestal calibration
-  uint16_t mPedToleratedGapWidth = 5;    ///< Tolerated gap between bins: if |bin1 - bin2| < width -> bin1 and bin2 belongs to same peak
-  uint32_t mPedMinEvents = 100;          ///< Minimal number of events to produce calibration
-  float mPedSuspiciousPedestalRMS = 10.; ///< Take additional care for channel if its RMS >  mPedSuspiciousPedestalRMS
-  float mPedZSnSigmas = 3.;              ///< Zero Suppression threshold
+  uint16_t pedToleratedGapWidth = 5;    ///< Tolerated gap between bins: if |bin1 - bin2| < width -> bin1 and bin2 belongs to same peak
+  uint32_t pedMinEvents = 100;          ///< Minimal number of events to produce calibration
+  float pedSuspiciousPedestalRMS = 10.; ///< Take additional care for channel if its RMS >  mPedSuspiciousPedestalRMS
+  float pedZSnSigmas = 3.;              ///< Zero Suppression threshold
 
   // Parameters used in noise scan calibration
-  uint32_t mNoiseMinEvents = 100;                    ///< Minimal number of events to produce calibration
-  float mNoiseToleratedChannelEfficiencyLow = 0.9;   ///< Tolerated channel efficiency (lower limit)
-  float mNoiseToleratedChannelEfficiencyHigh = 1.01; ///< Tolerated channel efficiency (upper limit)
-  uint16_t mNoiseThreshold = 10;                     ///< ADC threshold
-  float mNoiseFrequencyCriteria = 0.5;               ///< Appearance frequency of noisy channels
+  uint32_t noiseMinEvents = 100;                    ///< Minimal number of events to produce calibration
+  float noiseToleratedChannelEfficiencyLow = 0.9;   ///< Tolerated channel efficiency (lower limit)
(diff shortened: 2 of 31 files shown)


kharlov (dismissed):

DataFormats/Detectors/CPV/src/Hit.cxx
```diff
@@ -43,14 +43,21 @@ Hit& Hit::operator+=(const Hit& rhs)
...
-Hit Hit::operator+(const Hit& rhs) const
+// Hit Hit::operator+(const Hit& rhs) const
+// {
+//   Hit result(*this);
+//   if (rhs.GetEnergyLoss() > result.GetEnergyLoss()) {
+//     result.SetTime(rhs.GetTime());
+//   }
+//   result.SetEnergyLoss(result.GetEnergyLoss() + rhs.GetEnergyLoss());
+//   return result;
+// }
```
shahor02: do you need commented code?

Detectors/CPV/base/include/CPVBase/CPVCalibParams.h
```diff
@@ -0,0 +1,58 @@
...
+#include "CommonUtils/ConfigurableParamHelper.h"
+#include "CommonUtils/NameConf.h"
+
+namespace o2
+{
+namespace cpv
+{
+// parameters used in responce calculation and digitization
+// (mostly used in GEANT stepping and Digitizer)
+struct CPVCalibParams : public o2::conf::ConfigurableParamHelper<CPVCalibParams> {
```
shahor02, line 25: for structs one should not prepend the data members by `m`, just start from the lower case.

Detectors/CPV/calib/src/GainCalibrator.cxx
```diff
@@ -0,0 +1,281 @@
...
+  if (mFitRangeL < AmplitudeSpectrum::lRange) {
+    mFitRangeL = AmplitudeSpectrum::lRange;
+  }
+  if (mFitRangeR > AmplitudeSpectrum::rRange) {
+    mFitRangeR = AmplitudeSpectrum::rRange;
+  }
+  double binWidth = (AmplitudeSpectrum::rRange - AmplitudeSpectrum::lRange) / AmplitudeSpectrum::nBins;
+  mFitRangeL = AmplitudeSpectrum::lRange + std::floor((mFitRangeL - AmplitudeSpectrum::lRange) / binWidth) * binWidth;
+  mFitRangeR = AmplitudeSpectrum::lRange + std::floor((mFitRangeR - AmplitudeSpectrum::lRange) / binWidth) * binWidth;
+  LOG(info) << "Parameters used: ";
```
shahor02: note that the cpvParams (as ConfigurableParam) has printKeyValues method, which can print also the provenance.

Detectors/CPV/calib/testWorkflow/GainCalibratorSpec.h
```diff
@@ -0,0 +1,190 @@
...
+  void run(o2::framework::ProcessingContext& pc) final
+  {
+    o2::base::GRPGeomHelper::instance().checkUpdates(pc);
+    o2::base::TFIDInfoHelper::fillTFIDInfo(pc, mCalibrator->getCurrentTFInfo());
+    TFType tfcounter = mCalibrator->getCurrentTFInfo().startTime;
+    auto&& digits = pc.inputs().get<gsl::span<o2::cpv::Digit>>("calibdigs");
+
+    // previous gains
+    if (!mCalibrator->isSettedPreviousGains()) {
+      const auto previousGains = o2::framework::DataRefUtils::as<CCDBSerialized<o2::cpv::CalibParams>>(pc.inputs().get("gains"));
```
shahor02: please always use `pc.inputs().get<o2::cpv::CalibParams*>("gains")` rather than `as<...>` which does not invoke finaliseCCDB callback and may miss other features in the future.

Detectors/CPV/calib/testWorkflow/GainCalibratorSpec.h
```diff
@@ -0,0 +1,190 @@
...
+      if (previousGains) {
+        mCalibrator->setPreviousGains(new o2::cpv::CalibParams(*previousGains));
+        LOG(info) << "GainCalibratorSpec()::run() : I got previous gains";
+      }
+    }
+
+    // previous calib data
+    if (!mCalibrator->isSettedPreviousGainCalibData()) {
+      const auto previousGainCalibData = o2::framework::DataRefUtils::as<CCDBSerialized<o2::cpv::GainCalibData>>(pc.inputs().get("calibdata"));
```
shahor02: idem

shahor02 (changes requested):
Hi @sevdokim 
please see the comments below, particularly on the fetching CCDB objects from the context.

I see that you are reassigning one to one ConfigurableParam data members to other class data members, in principle, you can use them directly.
Also, if your configurable params should be eventually fetched from the CCDB (it depends if you will have different period-dependent settings, if they never change than you don't need to do this) better implement this at once to avoid further changes (such objects still can be overridden from the command line), see my email ("Fetching ConfigurableParam-based objects from the CCDB via DPL fetcher") to wp13 from today.

Detectors/CPV/calib/testWorkflow/GainCalibratorSpec.h
sevdokim: Hi @shahor02 
thank you for your comments! I'm implementing as you proposed. 
Question regarding "Fetching ConfigurableParam-based objects": as I understood, it's enought just to fetch ConfigParam object once and do nothing with it: it will be automatically loaded into memory. Is it correct? Then I just use it as before. 
Is it possible to fetch ConfigParams in void init(o2::framework::InitContext& ic) method?

shahor02: Hi @sevdokim 
Yes, it is enough to get once `pc.inputs.get<configurable*>(XXX)` to have it initialized.
But all accesses to CCDB and following initialisations can be don from the run() method only, when there is a data with the timestamp. The init() method has no timestamp.
See e.g. how I do inits in https://github.com/AliceO2Group/AliceO2/blob/acc00cd728be26ab7c022b764e9ab3be5382ce30/Detectors/GlobalTrackingWorkflow/src/PrimaryVertexingSpec.cxx#L186

sevdokim: Hi @shahor02 
I'm trying to fetch o2::cpv::CPVCalibParams as
```
pc.inputs().get<o2::cpv::CPVCalibParams*>("calibparams");
```
however it does not compile:
```
/home/evdokimov/alice/sw/SOURCES/O2/dev/0/Framework/Core/include/Framework/InputRecord.h: In instantiation of 'decltype(auto) o2::framework::InputRecord::get(R, int) const [with T = o2::cpv::CPVCalibParams*; R = const char*]':
/home/evdokimov/alice/sw/SOURCES/O2/dev/0/Detectors/CPV/calib/testWorkflow/GainCalibratorSpec.h:76:62:   required from here
/home/evdokimov/alice/sw/SOURCES/O2/dev/0/Framework/Core/include/Framework/InputRecord.h:450:46: error: static assertion failed: T is not a supported type
  450 |       static_assert(always_static_assert<T>::value, "T is not a supported type");
      |                                              ^~~~~

```
`o2::framework::DataRefUtils::as<CCDBSerialized<o2::cpv::CPVCalibParams>>(pc.inputs().get("calibparams"));` also does not work. It compiles normally but do not invoke methods responsible for updating singleton class o2::cpv::CPVCalibParams at runtime, just like you said. Do you know what can be a problem?

shahor02: As I wrote here: https://docs.google.com/presentation/d/1WPNixo57zAn_gM6vMCIBOHs-DZqcKAkLX0CwUgO_VYg/edit#slide=id.g12aeb8209f1_3_0, the class should be declared messageable.

sevdokim: Hi @shahor02 thanks, managed to get things work (approximately). Now another problem appear:
I pushed to ccdb-test modified CPVCalibParams in which gainDesiredLandauMPV is 201 instead of default 200. I run following code to fetch it:
```
const auto par = pc.inputs().get<o2::cpv::CPVCalibParams*>("calibparams");
      par->printKeyValues(); // print what we get from query
      o2::cpv::CPVCalibParams::Instance().printKeyValues(); // print what we really have as parameters
```
and that is what I get:
```
[2138535:cpv-noise-calibration]: [18:20:47][INFO] Caching in 140547698162288 ptr to CPV/CPV_CalibPars/0 (0x2e1af80)
...
[2138535:cpv-noise-calibration]: CPVCalibParams.gainDesiredLandauMPV : 201		[ CODE ]
...
[2138535:cpv-noise-calibration]: CPVCalibParams.gainDesiredLandauMPV : 200		[ CODE ]

```
Looks like CPVCalibParams is not a singleton?

sevdokim: Looks like I need to wait until https://github.com/AliceO2Group/AliceO2/pull/8834 is merged.

shahor02: @sevdokim as the page I've referred to says, one should NOT assign the output of the ConfigurableParam fetching. 
It is a singleton by its usage but the since the ROOT persistency requires the class to have a public c-tor, it has it.

Yes, you need to wait for the https://github.com/AliceO2Group/AliceO2/pull/8834 to be merged (or rebase to it).

sevdokim:
Hi @shahor02 
can you please look again at my code? I implemented your suggestions. CI "build/AliceO2/O2/o2/macOS" is failed but I think the error is not related.

shahor02 (approved):
failure is unrelated, merging

PR #8838, opened 2022-05-16
aod-producer: make out-of-bound access to MCH tracks non fatal

sawenzel:
Accessing mchTracks is sometimes out of bounds.
Let's avoid crashing the productions for now by
catching the situation and issuing an error messgae.

To reproduce, use /alice/cern.ch/user/a/aliprod/LHC21i1/O2DPG_pp_minbias_testbeam_v4.sh

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -405,11 +405,17 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
   // helper lambda for mch bitmap -- common for global and standalone tracks
   auto getMCHBitMap = [&](int mchTrackID) {
     if (mchTrackID != -1) { // check matching just in case
-      const auto mchTrack = mchTracks[mchTrackID];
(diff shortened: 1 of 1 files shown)


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -405,11 +405,17 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
   // helper lambda for mch bitmap -- common for global and standalone tracks
   auto getMCHBitMap = [&](int mchTrackID) {
     if (mchTrackID != -1) { // check matching just in case
-      const auto mchTrack = mchTracks[mchTrackID];
+      // check to for JIRA-TICKET https://alice.its.cern.ch/jira/browse/O2-2939
+      // remove once underlying problem solved
+      if (mchTrackID < 0 || mchTrackID >= mchTracks.size()) {
```
shahor02, line 410: This may make the AOD broken. I  think it is better to solve the problem itself.

The wrong track is anyway accessed on https://github.com/AliceO2Group/AliceO2/pull/8838/files#diff-6dc278732c5ed5f43d5567ad2c469ac11acf0e85339ae708bdf977ddd8fba392R440
BTW, there the track also should be fetched by reference.

PR #8841, opened 2022-05-16
MCH: filter input digits for time-clustering peak search

aferrero2707:
The commit introduces the possibility to only use signal-like digits as input for the peak searching step.
The digits filtering is enable via the `MCHTimeClusterizerParam.peakSearchSignalOnly` configurable parameter (enabled by default).

### DIFF
--- a/Detectors/MUON/MCH/TimeClustering/CMakeLists.txt
+++ b/Detectors/MUON/MCH/TimeClustering/CMakeLists.txt
@@ -11,7 +11,7 @@
 
 o2_add_library(MCHTimeClustering
         SOURCES src/ROFTimeClusterFinder.cxx src/TimeClusterizerParam.cxx src/TimeClusterFinderSpec.cxx
-        PUBLIC_LINK_LIBRARIES O2::MCHBase O2::Framework O2::MCHROFFiltering)
+        PUBLIC_LINK_LIBRARIES O2::MCHBase O2::Framework O2::MCHDigitFiltering O2::MCHROFFiltering)
 
 o2_target_root_dictionary(MCHTimeClustering HEADERS include/MCHTimeClustering/TimeClusterizerParam.h)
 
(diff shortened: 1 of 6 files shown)


aphecetche:
@aferrero2707 the code looks fine, but if I understand correctly with this change, as it is, the selected ROFs will be narrower, won't they ? And thus we risk loosing some digits ?

Anyway, I'm also trying in // to use your QC ROF branch to try to visualise the effect of those changes...

aferrero2707:
@aphecetche in fact the ROFs will not be necessarily smaller. Here is how the algorithm works:

1. the time-frame is divided in time bins (see [here](https://github.com/AliceMCH/AliceO2/blob/mrrtf-169-improve-digit-selection-for-tc-window-search/Detectors/MUON/MCH/TimeClustering/src/ROFTimeClusterFinder.cxx#L40)), and the resulting histogram is filled with the number of digits in each bin (`timeBin.mNDigits`), as well as the number of digits for the peak searching (`timeBin.mNDigitsPS`)
2. the code searches for peaks in the number of (signal-like) digits in each time bin (see [here](https://github.com/AliceMCH/AliceO2/blob/mrrtf-169-improve-digit-selection-for-tc-window-search/Detectors/MUON/MCH/TimeClustering/src/ROFTimeClusterFinder.cxx#L100))
3. for each peak, a window of +/- `mNbinsInOneWindow / 2` maximum is opened, and all digits falling in this window are assigned to a new ROF (see [here](https://github.com/AliceMCH/AliceO2/blob/mrrtf-169-improve-digit-selection-for-tc-window-search/Detectors/MUON/MCH/TimeClustering/src/ROFTimeClusterFinder.cxx#L240))
4. the (optional) gaps between the peaks are filled with additional ROFs of `mNbinsInOneWindow` width (see [here](https://github.com/AliceMCH/AliceO2/blob/mrrtf-169-improve-digit-selection-for-tc-window-search/Detectors/MUON/MCH/TimeClustering/src/ROFTimeClusterFinder.cxx#L231)). Those ROFs are likely to be non-physical and be removed by the "trackable" criteria.

aphecetche (approved):

PR #8844, opened 2022-05-16
Executable to query or upload MID bad channels

dstocco:


dstocco:
@chiarazampolli : this PR is used to generate a default bad channel object for MID. For simplicity, the object can be found also here: http://ccdb-test.cern.ch:8080/browse/MID/Calib/BadChannels can you pleas upload it as a default object in the standard CCDB?
Notice that I had to start it from timestamp 1 since I have the impression that timestamp 0 is considered as "now".

Detectors/MUON/MID/Calibration/exe/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+# v3 (GPL Version 3), copied verbatim in the file "COPYING".
+#
+# In applying this license CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization or
+# submit itself to any jurisdiction.
+
+o2_add_executable(
+  bad-channels-ccdb
+  COMPONENT_NAME mid
+  SOURCES bad-channels-ccdb.cxx
```
sawenzel, line 15: I believe the code is fine. 

However, I am slightly worried about such development since I see an ever growing number of executables that we link ... and this costs a lot of (link) time and other resources. The problem will grow if we have an exec for each single calib object.

Alternatively, CCDB objects could be produced by ROOT macros (no linking) and uploaded using existing tools such as `o2-ccdb-upload`. Downloading a file is possible using `o2-ccdb-downloadccdbfile` from where it could again be read using a macro or other tools. At the very least, it would be good trying to constrain to a single `swiss-army` CCDB tool on a detector level.

This comment should not block the PR. I just wanted to give my thoughts.

aphecetche (approved):

Detectors/MUON/MID/Calibration/exe/CMakeLists.txt
aphecetche: @sawenzel I agree the linking times (and the sheer size of our executables) are probably too large. While side-tracking this issue using ROOT macros is certainly one option (that I don't particularly like personally) , I think we should probably also tackle the root (pun intended) of the issue and check maybe if our libraries are not simply exporting too many symbols (thus increasing linking time) and/or why our executables are so large.
That being said, for the case at hand, I could indeed [try to see](https://alice.its.cern.ch/jira/browse/MRRTF-172) if merging the MCH and MID ccdb-related executables into a single executable would be feasible

PR #8850, opened 2022-05-16
DPL: add printout to report downstream congestion

ktf:


Framework/Core/src/SendingPolicy.cxx
```diff
@@ -35,6 +35,15 @@ std::vector<SendingPolicy> SendingPolicy::createDefaultPolicies()
             .matcher = [](DeviceSpec const&, ConfigContext const&) { return true; },
             .send = [](FairMQDeviceProxy& proxy, FairMQParts& parts, ChannelIndex channelIndex) { 
               auto *channel = proxy.getOutputChannel(channelIndex);
-              channel->Send(parts); }}};
+              auto timeout = 5000;
```
davidrohr: Thx for the PR, looks good, though I would reduce the timeout to 500ms

ktf:
```suggestion
              auto timeout = 1000;
```

Let's make it 1s so that I do not have to change the printout. I also worry we start having a bunch of false positives / flip / flopping...

PR #8852, opened 2022-05-16
[MFT] Restore full tracking by default

rpezzi:
This PR restores MFT tracking default option to run full Kalman filter.

It renders obsolete O2DPG workflow option `--mft-reco-full`. Which will be suppressed once once this PR is merged.

Note that reconstruction of pilot beam data and TED shots will then require the corresponding adaptation to deal with a misaligned detector:  `MFTTracking.LTFclsRCut=0.2; MFTTracking.forceZeroField=true;`. @chiarazampolli, @catalinristea, @shahor02 and @julekm am keeping this PR as a draft while changes are made for pass4 and any other reconstruction with data that may be going on (such as ED).


davidrohr:
I don't really understand how the changes in this PR would restore the full tracking. If I understand correctly it makes a cut more strict and enables the B-Field? You are sure this is the correct PR.
Also, in combination with https://github.com/AliceO2Group/O2DPG/pull/410, I assume the prodtests/full-system-test/dpl-workflow.sh should also be changed, to run the full tracking in pp but not in Pb-Pb?

rpezzi:
Hi @davidrohr , thanks. This is the correct PR, but I am closing it in favour of using configuration parameters from CCDB per period.

PR #8858, opened 2022-05-17
Applying Branson corrections to MCH track in FwdMatching workflow

mcoquet642:
During matching between MFT & MCH, the MCH tracks which are matched to an MFT track are propagated to the last MFT to the matching plane, applying Branson corrections to account for MCS going through the absorber. This has been tested to improve the transverse momentum resolution of the GlobalMuonTracks.

### DIFF
--- a/Detectors/GlobalTracking/include/GlobalTracking/MatchGlobalFwd.h
+++ b/Detectors/GlobalTracking/include/GlobalTracking/MatchGlobalFwd.h
@@ -325,6 +325,7 @@ class MatchGlobalFwd
   bool mMCTruthON = false;      ///< Flag availability of MC truth
   bool mUseMIDMCHMatch = false; ///< Flag for using MCHMID matches (TrackMCHMID)
   int mSaveMode = 0;            ///< Output mode [0 = SaveBestMatch; 1 = SaveAllMatches; 2 = SaveTrainingData]
+  bool mUseBranson = true;      ///< Flag for applying Branson corrections
   MatchingType mMatchingType = MATCHINGUNDEFINED;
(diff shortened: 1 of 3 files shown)


shahor02 (commented):
@mcoquet642 Applying the Branson correction with the position of the MFT track used as a vertex means effectively constraining the MCH track to this point (i.e. consider it as a measured point). But then you update this already "updated" track by the clusters of the MFT track. In a way, the update by the last point of the MFT track will be done twice. 
Formally this will improve the resolution (decrease the cov.matrix) but in reality I don't see how it can improve the track. Does not it affects the pulls?

Note that this is different from the classical Branson correction with the primary vertex position used as a constraint: the PV is normally not used in the track fit, so there is no double usage of the same info.

rpezzi:
Hi @shahor02 , Branson correction improves energy loss correction for track propagation though the absorber, resulting in a better momentum estimate as compared to ideal propagation pathlength-based correction. @mcoquet642 then uses this improved track-momentum to start fitting global forward tracks with MFT clusters. Maurice showed that this approach significantly improves momentum resolution.

But indeed, we did not check pulls, and, in fact, there may be one duplicated measurement for the variance of q/pt. We have to check pulls. @shahor02 , thanks for your feedback.

shahor02:
Hi @rpezzi, actually, the Branson plane correction improves the errors due to the MS in the absorber, not e.loss, effectively by constraining the track to a known point before the absorber. This is ok to do if the point is not used in the track fit, then one indeed uses additional information so that the track improves. 
But here we speak about using the point which is anyway constraining the track during the fit, I don't see where the additional info comes.

PR #8860, opened 2022-05-17
DPL: fix out of bound access in DataDescriptorQueryBuilder

ktf:


ktf:
@matthiasrichter could you confirm this was the intended behaviour?

ktf:
Currently undefined behaviour. Merging this since it can only improve the situation.

ktf:
Currently undefined behaviour. Merging this since it can only improve the situation.

PR #8861, opened 2022-05-17
Again disable p.vertexin if ITS is absent

shahor02:
But keep --skip mode for cosmic.
We observed that running the p.vertexing with the CCDB objects being the
only inputs leads to random timestamp used in the DPL CCDB fetcher query
(which is upstream wrt the vertexer).


shahor02:
Tested locally, merging since current version creates problems on EPN

PR #8865, opened 2022-05-17
Add simcuts file for ITS

fmazzasc:
Cuts inherited from Run 2 file (AliRoot/data/galice.cuts). Delta-ray production tested in local simulations (see plot in attachment)
deltasim


mconcas (approved):

chiarazampolli:
Thanks, merging.

PR #8868, opened 2022-05-17
Fixes in TPC-ITS matching: obsolete reference + out-of-bound access [O2-2909]

shahor02:
Should fix [O2-2909](https://alice.its.cern.ch/jira/browse/O2-2909)


shahor02:
almost trivial, tested locally, merging

PR #8870, opened 2022-05-17
Configurable param class for CTP triggers offsets

shahor02:
the hack of FT0 decoder is unrelated, will be removed before merging


shahor02:
Failure unrelated

PR #8871, opened 2022-05-18
DPL Analysis: Event mixing: time performance validation

saganatt:
Benchmark that I prepared for QM, showing that mixing with O2 interface has the same performance as manual looping.


ktf (dismissed):

saganatt:
Sorry, I had just a code format nitpick. Can we merge? :-) Thank you!

PR #8872, opened 2022-05-18
[EMCAL-551] Use std::round for energy and time truncation

mfasDa:
Brute force trunctation leading to
precision loss resulting in approx
2% shift of the pi0 mass peak.
The usage of std::round before
conversion to int recovers the
precision loss within the range
of precision of the detector.


mfasDa:
Results of the studies concerning the effect of the truncation demonstrating the need for changing can be found here https://alice.its.cern.ch/jira/browse/EMCAL-737 and here https://alice.its.cern.ch/jira/browse/EMCAL-551

shahor02 (approved):

shahor02:
failure unrelated, merging

PR #8873, opened 2022-05-18
Event Display: Change the default minimum amount of tracks for JSON production to at least one

pnwkw:
… least one


shahor02 (approved):
trivial, merging

PR #8875, opened 2022-05-18
Do not use MCTruthContainer for MID reconstructed track labels

dstocco:


dstocco:
Errors are unrelated

shahor02 (approved):
thanks!

PR #8881, opened 2022-05-18
[MID] add configurable params and protection against too many tracks

pillot:
The 1st commit moves the MID tracking parameters that may need to be adjusted into configurable params.

The 2nd commit add a protection to abort the tracking if too many tracks (number is configurable) are found in the same ROF. I added this protection only for the algorithm tracking every possible candidates (the one to be used for pp and PbPb data in principle), because it cannot cope with extreme multiplicities like in the TED shots. The other algorithm, tracking only the best candidates, already has an intrinsic limitation and shows no problem to reconstruct the TED shots.

@dstocco is it ok for you?

### DIFF
--- a/Detectors/MUON/MID/Tracking/CMakeLists.txt
+++ b/Detectors/MUON/MID/Tracking/CMakeLists.txt
@@ -11,8 +11,12 @@
 
 o2_add_library(
   MIDTracking
-  SOURCES src/HitMapBuilder.cxx src/Tracker.cxx
+  SOURCES src/HitMapBuilder.cxx src/Tracker.cxx src/TrackerParam.cxx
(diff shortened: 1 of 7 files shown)


dstocco:
Looks fine to me

aphecetche (approved):

aphecetche:
@TimoWilken this one seems to be stuck in the formatting checks...

TimoWilken:
Hi @aphecetche, I changed the way those are run slightly, so the name changed. Sorry about the inconvenience!

You can make GitHub pick up the new names by adding an empty commit or amending the last one with no changes.

If this PR should just be merged, let me know, and I can do that.

aphecetche:
@TimoWilken if you can merge it, yes please do. All the other checks are green, and it's been reviewed.

PR #8889, opened 2022-05-19
Changing CcdbApi to ccdb-populator for uploading to CCDB

matthias-kleiner:


wiechula (approved):
Thanks for the quick implementation!

wiechula:
seems some of the cis are hanging

matthias-kleiner:
> seems some of the cis are hanging

Should I push again to trigger the tests?

PR #8890, opened 2022-05-19
Use HBFUtilsInitializer in FIT digit reader workflows

shahor02:
@mslupeck not all FIT digit readers follow the standard behaviour for all readers: 
1) the files from MC can be read as is, correct timestamp/orbit  will be injected: o2simdigitizerworkflow_configuration.ini created in digitization will be used (equivalent to default `--hbfutils-config o2simdigitizerworkflow_configuration.ini `
2) files from raw data processing using o2-tfidinfo-writer-workflow (which creates o2_tfidinfo.root file) should be read with `--hbfutils-config o2_tfidinfo.root` 
3) files from raw data processing produced w/o  o2-tfidinfo-writer-workflow: at the moment you will need to create an ini file containing a block like
```
[HBFUtils]
nHBFPerTF=128
obligatorySOR=true
orbitFirst=0
runNumber=300001
orbitFirstSampled=12800
maxNOrbits=4294967295
startTime=1546300800000
```
and use it with `--hbfutils-config <ini file>` .
The i-th TF will be injected with timestamp `startTime+i*nHBFPerTF*orbit_duration` and the injected TF 1st orbit will be `1st_orbit_of_TF_containig_orbitFirstSamled + i*nHBFPerTF`
4) to ignore time/orbit injection (i.e. recover old behaviour of your readers): use `--hbfutils-config none` option.

### DIFF
--- a/Detectors/FIT/FDD/workflow/src/digits-reader-workflow.cxx
+++ b/Detectors/FIT/FDD/workflow/src/digits-reader-workflow.cxx
@@ -15,13 +15,20 @@
 /// \author ruben.shahoyan@cern.ch
 
 #include "Framework/CallbackService.h"
+#include "Framework/CallbacksPolicy.h"
 #include "Framework/ControlService.h"
 #include "Framework/ConfigParamRegistry.h"
 #include "Framework/Task.h"
 #include "FDDWorkflow/DigitReaderSpec.h"
+#include "DetectorsRaw/HBFUtilsInitializer.h"
 #include "CommonUtils/ConfigurableParam.h"
 
 using namespace o2::framework;
+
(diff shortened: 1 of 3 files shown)


mslupeck:
>

mslupeck:
Sorry, I didn't mean to close it...
What do you mean by the first sentence? After this PR will the behaviour of the digit readers be consistent with other readers?

mslupeck (approved):
@afurs @AllaMaevskaya - FYI

shahor02:
> What do you mean by the first sentence? After this PR will the behaviour of the digit readers be consistent with other readers?

All o2 workflows which start by reading data from root files are injecting the timestamp/1st_orbit for every TF.

PR #8896, opened 2022-05-19
Reduce how many timeframes the faster channel can be ahead

ktf:


ktf:
Merging to get it in the nightly...

PR #8901, opened 2022-05-20
Make FV0 reco use DPL CCDB fetcher, fix Calib path

shahor02:
FV0 reconstruction is changed to use the DPL CCDB fetcher, all old paths FV0/Calibration/...
are changed to FV0/Calib/...

@AllaMaevskaya @mslupeck @mbroz84 @jotwinow could you please check if there are other instances (e.q. in QC) of access to <FIT>/Calibration, if so, should be changed to /Calib and the objects should be relocated. Please also change from CCDBManager to DPL fetcher, like this PR does.
Also, see the question about old objects in FV0/Calibration: https://alice.its.cern.ch/jira/browse/O2-2917?focusedCommentId=285348&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-285348

### DIFF
--- a/Detectors/FIT/FV0/calibration/macros/makeChannelTimeOffsetFV0CalibObjectInCCDB.C
+++ b/Detectors/FIT/FV0/calibration/macros/makeChannelTimeOffsetFV0CalibObjectInCCDB.C
@@ -26,6 +26,6 @@ int makeChannelTimeOffsetFV0CalibObjectInCCDB(const std::string url = "http://al
   for (auto& dummyCalCoeff : obj.mTimeOffsets) {
     dummyCalCoeff = 0;
   }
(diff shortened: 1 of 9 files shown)


mslupeck:
@shahor02 
Hi, I confirm there are no /Calibration paths hardcoded in QC for neither of the FIT detectors. This PR handles it for FV0. FDD never had calibration implemented, and for FT0 @AllaMaevskaya agreed to make the needed adjustments.

shahor02:
Thanks @mslupeck!

AllaMaevskaya:
Hello
are wait me to make similar changes for FT0?

shahor02:
yes, please, we should switch to DPL CCDB fetcher in the workflows.

AllaMaevskaya:
Fetcher was implemented week ago, now I only changed path name to FT0/Calib

PR #8903, opened 2022-05-20
DPL: do not send the oldest possible timeframe on a per route basis

ktf:


ktf:
@jgrosseo this should fix analysis
@davidrohr this might explain the issue of last night...

davidrohr:
So you are saying last night. Does this fix the problem introduced by https://github.com/AliceO2Group/AliceO2/pull/8739/files ?

ktf:
I think there are two problems actually. The race condition fixed last night occurs when there is a channel that produces data faster then the others and we have pipelining.

This PR fixes the case in which FairMQ decides to deliver messages in multiple bunches rather that all together, even if they belong to the same computation.

This PR fixes for sure a regression in analysis.

PR #8908, opened 2022-05-20
Fix a bunch of pedantic warnings

mconcas:
Re-proposing: #8690. @davidrohr

### DIFF
--- a/run/checkStack.cxx
+++ b/run/checkStack.cxx
@@ -118,6 +118,7 @@ int main(int argc, char** argv)
     }
 
     if (hitbr) {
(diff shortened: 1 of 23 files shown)


run/checkStack.cxx
```diff
@@ -118,6 +118,7 @@ int main(int argc, char** argv)
     }
 
     if (hitbr) {
+      LOG(info) << "trackidsinITS.size() " << trackidsinITS.size() << " trackidsinITS_fromhits.size() " << trackidsinITS_fromhits.size();
```
sawenzel: do you need this? Probably has nothing to do with "fixing warnings".

mconcas: No, indeed. I used it to fix a failing test. Removing it soon.

sawenzel (approved):

PR #8909, opened 2022-05-20
MRRTF-173: cluster transformer now reads geometry from CCDB by default

aphecetche:
The old behavior of reading from file (using the `--geometry` option) is retained
 in the `o2-mch-clusters-transformer-workflow` executable, but must then
 be used together with the `--mch-disable-geometry-from-ccdb` option.

Note this PR also moves the cluster-transformer out of the Workflow directory (one more step into making this directory leaner)

### DIFF
--- a/Detectors/MUON/MCH/Geometry/Transformer/CMakeLists.txt
+++ b/Detectors/MUON/MCH/Geometry/Transformer/CMakeLists.txt
@@ -21,3 +21,15 @@ o2_add_executable(
   PUBLIC_LINK_LIBRARIES ROOT::Geom Microsoft.GSL::GSL Boost::program_options
                         O2::MCHGeometryTransformer
   COMPONENT_NAME mch)
+
+o2_add_executable(
+        clusters-transformer-workflow
+        SOURCES src/clusters-transformer-workflow.cxx src/ClusterTransformerSpec.cxx
+        COMPONENT_NAME mch
+        PUBLIC_LINK_LIBRARIES
+           O2::DetectorsBase
+           O2::Framework
+           O2::MCHBase
+           O2::MCHGeometryTransformer
+        )
+
--- a/Detectors/MUON/MCH/Geometry/Transformer/CMakeLists.txt
+++ b/Detectors/MUON/MCH/Geometry/Transformer/CMakeLists.txt
@@ -11,9 +11,10 @@
 
 o2_add_library(
   MCHGeometryTransformer
-  SOURCES src/Transformations.cxx src/VolumePaths.cxx
+  SOURCES src/Transformations.cxx src/VolumePaths.cxx src/ClusterTransformerSpec.cxx
   PUBLIC_LINK_LIBRARIES ROOT::Geom O2::MathUtils Microsoft.GSL::GSL
-                        RapidJSON::RapidJSON)
+                        RapidJSON::RapidJSON O2::Framework O2::DetectorsBase
+                        O2::DataFormatsMCH)
 
 o2_add_executable(
   convert-geometry
@@ -24,7 +25,7 @@ o2_add_executable(
 
 o2_add_executable(
         clusters-transformer-workflow
-        SOURCES src/clusters-transformer-workflow.cxx src/ClusterTransformerSpec.cxx
+        SOURCES src/clusters-transformer-workflow.cxx
         COMPONENT_NAME mch
         PUBLIC_LINK_LIBRARIES
            O2::DetectorsBase
--- a/Detectors/MUON/MCH/Geometry/Transformer/src/ClusterTransformerSpec.cxx
+++ b/Detectors/MUON/MCH/Geometry/Transformer/src/ClusterTransformerSpec.cxx
@@ -9,7 +9,7 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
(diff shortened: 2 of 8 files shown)


Detectors/MUON/MCH/Geometry/Transformer/CMakeLists.txt
```diff
@@ -21,3 +21,15 @@ o2_add_executable(
...
+        clusters-transformer-workflow
+        SOURCES src/clusters-transformer-workflow.cxx src/ClusterTransformerSpec.cxx
+        COMPONENT_NAME mch
+        PUBLIC_LINK_LIBRARIES
+           O2::DetectorsBase
+           O2::Framework
+           O2::MCHBase
+           O2::MCHGeometryTransformer
+        )
+
```
pillot, line 36: I think you have to add the spec to the library (instead of the executable) to make it visible to the reco-workflow

Detectors/MUON/MCH/Geometry/Transformer/src/ClusterTransformerSpec.cxx
```diff
@@ -96,16 +122,27 @@ class ClusterTransformerTask
 
  public:
   o2::mch::geo::TransformationCreator transformation;
+  std::shared_ptr<base::GRPGeomRequest> mCcdbRequest;
+  bool mDisableCcdb;
```
pillot: mDisableCcdb is not used in the end

Detectors/MUON/MCH/Workflow/README.md
```diff
@@ -186,20 +186,7 @@ The decoding is done automatically by the `o2-ctf-reader-workflow`.
...
-
-```
-o2-mch-clusters-sampler-workflow
-    -b --nEventsPerTF 1000 --infile someclusters.data |
-o2-mch-clusters-transformer-workflow
-    -b --geometry Detectors/MUON/MCH/Geometry/Test/ideal-geometry-o2.json |
-o2-mch-clusters-sink-workflow
-    -b --txt --outfile global-clusters.txt --no-digits --global
-```
+Converts the clusters coordinates from local (2D within detection element plane) to global (3D within Alice reference frame) [more...](/Detectors/MUON/MCH/Geometry/Transformer/src/clusters-transformer.cxx)
```
pillot: don't you want to point to the README for more details rather than the class?

Detectors/MUON/MCH/Geometry/Transformer/src/clusters-transformer-workflow.cxx
```diff
@@ -23,12 +23,17 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
   workflowOptions.emplace_back("configKeyValues",
                                o2::framework::VariantType::String, "",
                                o2::framework::ConfigParamSpec::HelpString{"Semicolon separated key=value strings"});
+  workflowOptions.emplace_back("mch-disable-geometry-from-ccdb",
+                               o2::framework::VariantType::Bool, false,
+                               o2::framework::ConfigParamSpec::HelpString{"do not reead geometry from ccdb"});
```
pillot: reead --> read ;-)

pillot:
Why don't you activate the reading from the CCDB by default also in the reco-workflow?

Detectors/MUON/MCH/Geometry/Transformer/CMakeLists.txt
aphecetche: you're perfectly right. Hum, don't know how I could miss that it in my tests... probably did not ran exactly the code I thought I ran...

Detectors/MUON/MCH/Geometry/Transformer/src/ClusterTransformerSpec.cxx
aphecetche: indeed.

Detectors/MUON/MCH/Geometry/Transformer/src/clusters-transformer-workflow.cxx
aphecetche: rrright ;-)

Detectors/MUON/MCH/Workflow/README.md
aphecetche: good point.

aphecetche:
> Why don't you activate the reading from the CCDB by default also in the reco-workflow?

well, I am  : 

https://github.com/aphecetche/AliceO2/blob/b04366d148207ef7d5cfafc536629a488d60bb56/Detectors/MUON/MCH/Workflow/src/reco-workflow.cxx#L98

?

pillot:
My bad, I read "false" and forgot it was a disabling flag! ':)

pillot (approved):

aphecetche:
@shahor02 @davidrohr the fullCI test is consistently failing for this one, but I believe the failure to be unrelated to the PR itself (just tried a FST on a FLP-like machine to x-check that and it went through).

shahor02:
Hi @aphecetche ,
the test failed on `Mon 23 May 2022, 11:10:23 UTC` while the object for the MC range was loaded `23 May 2022 15:27 UTC`:  /browse/GLO/Config/GeometryAligned
http://alice-ccdb.cern.ch/browse/GLO/Config/GeometryAligned

Could you push some dummy commit to retrigger CI?

PR #8910, opened 2022-05-20
Enable TTree support in CcdbApi and upload tool + fixes

shahor02:
@Barthelemy : I've kept your compression settings but override them in case of uploading a TTree to avoid too large files.

@sawenzel I had to modify the retrieveBlob since the previous version with `string fullUrl = getHostUrl(i) + "/download/" + ETag;` was failing on local CCDB server (with localhost:8080)


shahor02:
@chiarazampolli , the CcdbApi was crashing on reading, uploader on uploading

PR #8912, opened 2022-05-20
Adding TPC calib

chiarazampolli:
@wiechula , @davidrohr 
This needs: https://github.com/AliceO2Group/AliceO2/pull/8889 (since I removed the option for the ccdb path for the upload).

@wiechula : in case you have some settings that should change according to pp or PbPb, can you let me know, or just open a PR on this one (or in O2 if this is merged) for aggregator-workflow.sh, doing as for TOF?

It goes with 
https://github.com/AliceO2Group/O2DPG/pull/416


wiechula:
@chiarazampolli , thanks for implementing it! For the moment, there are no differences between pp and PbPb.

PR #8913, opened 2022-05-20
[EMCAL-534] Skip pages which throw exceptions in page decoding

mfasDa:
In case the next function throws an exception the
payload must be skipped for the link, as the payload
extraction and separation failed. The relevant continue statement was missing and is added in this commit.
The consequence is that subsequent functions will
rethrow the error, as they depend on uninitialized data,
which are unprotected.


mfasDa:
Fixing the crash observed this evening on the EPN (run 516468)

davidrohr:
trivial, merging
thx for fixing!

PR #8914, opened 2022-05-20
DPL: make sure any output-proxy is last in the topology

ktf:
Notice that this relies on the name, but in the future
it should be extended to use labels.

Also notice that this will fail if there is a dependency
between two output proxies... Maybe we should find a better
approach when those become part of the reco workflow.


davidrohr:
thx, this seems to work. At least I can start up the workflow locally. Need to test on the EPNs at next occasion.

chiarazampolli:
Thanks!!

davidrohr (approved):

PR #8919, opened 2022-05-21
DPL: always broadcast the oldest possible timeframe when forwarding

ktf:


ktf:
Tested on my mac, on alihyperloop,  on the FLP staging with the ITS+QC workflow and on the EPN by @davidrohr. Seems to fix the issues with backpressure in the entropy encoders and similar problems.

PR #8922, opened 2022-05-22
MCH: electronic mapping fixes

aphecetche:
@aferrero2707 here's the re-generation of the electronic mapping following your fixes on the master file. In addition to the expected changes in CH9 and CH10 there are also some in CH1L : could you please x-check whether those are actually expected as well ?


aferrero2707:
> In addition to the expected changes in CH9 and CH10 there are also some in CH1L : could you please x-check whether those are actually expected as well ?

Yes, we swapped two SOLAR IDs in CH1L. Since the change was done both in the FEC and CRU mappings it does not affect the overall conversion from FEEID,LINK,DS to DE,PAD.

pillot (approved):

PR #8923, opened 2022-05-22
CTFReader: do not use usleep

ktf:
FairMQ can hang if one waits using usleep and we trigger a state
transition in between.


ktf:
@shahor02 this is probably unrelated, but without the CTF writer might hang on e.g. exit. Does it happen at all?

shahor02:
Have never seen it, will test in 20 min on the same workflow

PR #8932, opened 2022-05-23
DPL: allow for 64bit DPL_MAX_CHANNEL_AHEAD to exclude clocks

ktf:


davidrohr (approved):
just tried it, works as it should. Should go into the nightly tag!

ktf:
Tested also on the FLP staging with the ITS workflows. No regressions AFAICT.

PR #8935, opened 2022-05-23
[EMCAL-534] Skip pages with inconsistent trailers

mfasDa:
Pages must be skipped if the number of found trailer
words identified by the trailer marker is smaller than
the trailersize reported by the trailer itself encoded
in the last trailer word. In such cases the consistency
of the payload is also questionable.


mfasDa:
@shahor02 @davidrohr We have recently found cases in which SRUs were sending corrupted trailers, meaning trailers in which several words were missing the trailer pattern. In this case the payload size (number of ALTRO words) might be corrupted. A bug in the SRU firmware resulting in such cases was fixed, however we  decided that in case the same condition reappears these pages must be dropped as the consistency is not guaranteed.

PR #8939, opened 2022-05-23
[rANS] Fix Frequency Table Overflow

MichaelLettrich:
Solves an overflow issue  that is encountered when building a dictionary from many timeframes. In certain cases the `uint32_t` data type used to count the frequencies of symbols can overflow. The result is skewed distributions resulting in bad compression (usual case) or unrecoverable data loss due to wrong encoding (unlikely but possible).

Changes:
* Use `uint64_t` when building cumulative frequencies during renorming process.
* Introduce overflow checks when adding two frequency tables.
* Fault tolerance in `CTFWriterSpec`, in case addition of frequency tables fails.

### DIFF
--- a/Detectors/CTF/workflow/src/CTFWriterSpec.cxx
+++ b/Detectors/CTF/workflow/src/CTFWriterSpec.cxx
@@ -43,6 +43,7 @@
 #include "DataFormatsCTP/CTF.h"
 #include "rANS/rans.h"
 #include <vector>
+#include <stdexcept>
 #include <array>
 #include <TStopwatch.h>
 #include <vector>
@@ -118,8 +119,8 @@ class CTFWriterSpec : public o2::framework::Task
   bool mCreateRunEnvDir = true;
   bool mStoreMetaFile = false;
   int mVerbosity = 0;
-  int mSaveDictAfter = 0; // if positive and mWriteCTF==true, save dictionary after each mSaveDictAfter TFs processed
-  int mFlagMinDet = 1;    // append list of detectors to LHC period if their number is <= mFlagMinDet
+  int mSaveDictAfter = 0;          // if positive and mWriteCTF==true, save dictionary after each mSaveDictAfter TFs processed
+  int mFlagMinDet = 1;             // append list of detectors to LHC period if their number is <= mFlagMinDet
   uint32_t mPrevDictTimeStamp = 0; // timestamp of the previously stored dictionary
   uint32_t mDictTimeStamp = 0;     // timestamp of the currently stored dictionary
   uint64_t mRun = 0;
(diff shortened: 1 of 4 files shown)


Detectors/CTF/workflow/src/CTFWriterSpec.cxx
```diff
@@ -264,12 +265,22 @@ size_t CTFWriterSpec::processDet(o2::framework::ProcessingContext& pc, DetID det
     for (int ib = 0; ib < C::getNBlocks(); ib++) {
```
shahor02, line 267: Once a given block has overflown once, there is no point in continuing accumulation for it at producing warnings.
Could you add:
```
static uint64_t overflowStatus = 0;
for (int ib = 0; ib < C::getNBlocks(); ib++) {
  if (overflowStatus & (0x1UL<<ib)) {
     continue;
  }
```
and set in the catch after the warning 
```
overflowStatus |= (0x1UL<<ib);
```

Utilities/rANS/include/rANS/internal/helper.h
```diff
@@ -142,6 +145,15 @@ class JSONArrayLogger
   bool mReverse{false};
 };
 
+inline uint32_t safeadd(uint32_t a, uint32_t b)
+{
+  uint32_t result;
+  if (rans_unlikely(__builtin_uadd_overflow(a, b, &result))) {
```
shahor02, line 151: is this portable?

MichaelLettrich: Yes. I checked for all compilers we support (GCC>7, CLANG), however they might perform differently between compilers and architectures (like arm vs x86). The reason why I am using this, is because compilers have low cost access to CPU flags that I cannot access in a portable way (like the carry bit that is set when overflow occurs).

Detectors/CTF/workflow/src/CTFWriterSpec.cxx
MichaelLettrich: OK, I wanted to leave the benefit of doubt in case one tf overflows yet another does not. But if you prefer, I can change it.

shahor02: TFs should be rather uniform

MichaelLettrich: If that is the case I'll change that.

shahor02 (dismissed):
thanks!

shahor02:
@MichaelLettrich could you account for CI complaints?

MichaelLettrich:
@shahor02 done. Seems like `std::plus<>` is doing the correct upcasting of the operation if the accumulation type is explicitly specified in `std::inclusive_scan`.

shahor02 (approved):

shahor02:
failures unrelated.

PR #8941, opened 2022-05-23
[EMCAL-556] minor improvements in EMCAL DCS DP processing

mpoghos:
- possibility for using the run number from EMC DCS or from GRP as a trigger for storing the FEE config at SOR
- store the run number in FeeDCS
- minor fixes


Detectors/EMCAL/calib/include/EMCALCalib/CalibDB.h
```diff
@@ -256,19 +257,33 @@ class CalibDB
...
 
-  /// \brief Store Trigger DCS data in the CCDB
-  /// \param dcs trigger DCS data to be stored
+  /// \brief Store FEE DCS data in the CCDB
+  /// \param dcs FEE DCS data to be stored
   /// \param metadata Additional metadata that can be used in the query
   /// \param timestart Start of the time range of the validity of the object
   /// \param timeend End of the time range of the validity of the object
-  void storeTriggerDCSData(TriggerDCS* dcs, const std::map<std::string, std::string>& metadata, ULong_t timestart, ULong_t timeend);
+  void storeFeeDCSData(FeeDCS* dcs, const std::map<std::string, std::string>& metadata, ULong_t timestart, ULong_t timeend);
```
mfasDa, line 265: What will happen with the trigger DCS DP? I would guess at some point they will have to be added back. Would have been easier in this case to add new functions for the FeeDCS besides the existing ones for the trigger DCS, in case they are anyhow needed both. This interface change makes the macro tests crashing (since in O2 macros are tested at least for whether they compile).

mfasDa (commented):
Interface change TriggerDCS/FeeDCS in o2::EMCAL::CalibDB to revised.

Detectors/EMCAL/calib/include/EMCALCalib/CalibDB.h
mpoghos: TriggerDCS is included in FeeDCS

shahor02:
@mpoghos @mfasDa, it is  ready to be merged?

mfasDa (approved):
Looks good. @shahor02 when merging could you please squash-merge the PR keeping only the commit message of the first commit?

mpoghos:
> @mpoghos @mfasDa, it is ready to be merged?

@shahor02 yes, please

PR #8943, opened 2022-05-24
Fix for EPN ctf writer ordering which works on FLP

ktf:


ktf:
@davidrohr this seems to work for both the CTF writer issue and the ITS workflow on the FLP in staging. Could you cross check?

davidrohr:
Unfortunately, this seems to break the FST on the EPNs, with ctf-writer to writing all CTFs.
Actually the situation is even worse than before.
Before I needed to introduce fluctuations in the publishing rate by stalling the readout-proxy with gdb to produce this behavior.
With this PR, it failed by itself after only 5 TFs written.

ktf:
Updated. This seems to work now on my box (in previous attempts I forgot to revert back to use consumeOrdered) when I inject artificial delays via lldb and on the FLP.

ktf:
Error in QC unrelated. This does not break things more than they are already broken so we merge it so that we can have a smaller "fix" PR.

PR #8945, opened 2022-05-24
DPL Analysis: add int8 and int16 to Variant/Configurable

aalkin:
* add `int8` and `int16` to Variant
* update tests
* update parsing of small ints in config/workflow serialization and deserialization

### DIFF
--- a/Framework/Core/include/Framework/Variant.h
+++ b/Framework/Core/include/Framework/Variant.h
@@ -28,8 +28,6 @@ namespace o2::framework
 
 // Do NOT insert entries in this enum, only append at the end (before "Empty"). Hyperloop depends on the order.
 enum class VariantType : int { Int = 0,
-                               Int8,
-                               Int16,
                                Int64,
                                Float,
                                Double,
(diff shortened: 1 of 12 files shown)


Framework/Core/include/Framework/Variant.h
```diff
@@ -28,6 +28,8 @@ namespace o2::framework
 
 // Do NOT insert entries in this enum, only append at the end (before "Empty"). Hyperloop depends on the order.
 enum class VariantType : int { Int = 0,
+                               Int8,
+                               Int16,
```
jgrosseo: Do NOT insert entries in this enum, only append at the end (before "Empty"). Hyperloop depends on the order.
(as it says 3 lines above) :)

jgrosseo:
Thanks a lot @aalkin 
Apart from the order (to be fixed), it is good for me.

Framework/Core/include/Framework/Variant.h
aalkin: Indeed, that was quite silly of me. Fixed.

jgrosseo (approved):

PR #8949, opened 2022-05-24
Adding Dead and Inefficienct pixel information

pdhankhe:
1. Adding Dead and Inefficient Pixel Map in the workflow. 
2. Fixing a bug related to saving data in the tree for Noisy pixel.


Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -402,20 +402,52 @@ bool ITSThresholdCalibrator::findThresholdHitcounting(
...
+
+      else if (vThreshold[col_i] > N_INJ) {
         this->mNoisyPixID[chipID].push_back(col_i * 1000 + row);
       }
+
+      else if (vThreshold[col_i] > 0) {
+        this->mIneffPixID[chipID].push_back(col_i * 1000 + row);
+      }
+
+      isDead = this->isPixelDead(&(this->mPixelHits[chipID][row][col_i][0]), *(this->N_RANGE));
```
iravasen: I do not understand this part. Can't you simply add at the end (to the chain of if/elseif):
```
else {
  this->mDeadPixID[chipID].push_back(col_i * 1000 + row);
}
```
Note that `extractThresholdRow` for the digital and analog scan is called only at the end (`finalize` method) once all the hits of a given `chipID` have been collected.

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -402,20 +402,52 @@ bool ITSThresholdCalibrator::findThresholdHitcounting(
...
+    int nDead = 0;
     // Loop over all columns (pixels) in the row
     for (short int col_i = 0; col_i < this->N_COL; col_i++) {
       vChipid[col_i] = chipID;
       vRow[col_i] = row;
       vThreshold[col_i] = this->mPixelHits[chipID][row][col_i][0];
-      if (vThreshold[col_i] > N_INJ) {
+
+      if (vThreshold[col_i] == N_INJ) {
+        this->mCorrectPixID[chipID].push_back(col_i * 1000 + row);
```
iravasen: we don't need to track these pixels. They will be billions so the object we create will be too heavy to be shipped to the ccdb.

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -1124,6 +1185,9 @@ void ITSThresholdCalibrator::finalize(EndOfStreamContext* ec)
       if (this->mVerboseOutput) {
         LOG(info) << "Chip " << itchip->first << " done";
       }
+      if (!mNoisyPixID.count(itchip->first)) {
+        mRunTypeChip[itchip->first] = 0; // to avoid multiple writes into the tree
+      }
```
iravasen, line 1150: let's remove the `if` and keep only: `mRunTypeChip[itchip->first] = 0`.

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -1145,6 +1210,40 @@ void ITSThresholdCalibrator::finalize(EndOfStreamContext* ec)
         ++it;
       }
     }
```
iravasen, line 1168: We have to modify this while loop like this (and the same for the dead and inefficient ones):
```
while (it != this->mNoisyPixID.cend()) {
      if (!this->mCheckEos) {
        ++it;
        continue;
      }
      PixelType = "Noisy";
      this->addDatabaseEntry(it->first, name, 0, 0, 0, 0, 0, false); // all zeros are not used here
      if (this->mVerboseOutput) {
        LOG(info) << "Chip " << it->first << " done";
      }
      if (!this->mCheckEos) {
        it = this->mNoisyPixID.erase(it);
      } else {
        ++it;
      }
    }
```

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -988,19 +1023,45 @@ void ITSThresholdCalibrator::addDatabaseEntry(
...
+      for (int i = 0; i < v_Dead.size(); i++) {
         o2::dcs::addConfigItem(this->mTuning, "Stave", std::string(stave));
         o2::dcs::addConfigItem(this->mTuning, "Hs_pos", std::to_string(ssta));
         o2::dcs::addConfigItem(this->mTuning, "Hic_Pos", std::to_string(mod));
         o2::dcs::addConfigItem(this->mTuning, "ChipID", std::to_string(chipInMod));
         o2::dcs::addConfigItem(this->mTuning, "ChipDbID", std::to_string(confDBid));
-        o2::dcs::addConfigItem(this->mTuning, "Dcol", ds);
-        o2::dcs::addConfigItem(this->mTuning, "Row", std::to_string(v[i] % 1000));
-        o2::dcs::addConfigItem(this->mTuning, "Col", std::to_string(int((v[i] - v[i] % 1000) / 1000)));
+        o2::dcs::addConfigItem(this->mTuning, "Dead_Row", std::to_string(v_Dead[i] % 1000));
```
iravasen: here and also for the inefficient pixels your are adding the info to `mTuning`. Unfortunately we cannot do this because this data are not needed for the detector configuration but only for reference in the production ccdb. So the solution I see is:
- create a clone of `mTuning` (let's call it `mPixStat` in this example) that you fill with every pixel type: dead, ineff, noisy.
- mPixStat will go to an additional output of the task. 
- The aggregator receives it and makes the same kind of aggregation we do for `mTuning`. 
- The aggregator will ship the merged object to its output (an additional one). 
- The aggregator will need to have the possibility to ship the object to production ccdb directly (without the ccdb-populator wf) so I think here we can add an additional flag to the wf (say `--ccdb-url-prod`) that will contain the correct link (or an empty string if we do not want to ship to ccdb).

Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -988,19 +1023,45 @@ void ITSThresholdCalibrator::addDatabaseEntry(
...
+          o2::dcs::addConfigItem(this->mTuning, "ChipID", std::to_string(chipInMod));
+          o2::dcs::addConfigItem(this->mTuning, "ChipDbID", std::to_string(confDBid));
+          o2::dcs::addConfigItem(this->mTuning, "Dcol", ds);
+          o2::dcs::addConfigItem(this->mTuning, "Row", std::to_string(v[i] % 1000));
+          o2::dcs::addConfigItem(this->mTuning, "Col", std::to_string(int((v[i] - v[i] % 1000) / 1000)));
         }
+      }
+    }
+
+    if (PixelType == "Dead") {
```
iravasen: be sure that this `if` and the next one for ineff pix is inside the `if(this->mTagSinglePix)`

iravasen:
@pdhankhe please modify also this line: https://github.com/AliceO2Group/AliceO2/blob/55cf5dedf712968332487de5256a37af37f983cf/Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx#L140 

Please add a parameter to the workflow (a string basically) with which we can configure this link. At P2 we have to use a different link essentially so having it as parameter will simply our life when testing locally and at P2.

iravasen:
Hi @shahor02, this is a PR to calibration task of ITS which introduces the tracking of dead and inefficient pixels in digital/analog scans. In addition it fixes some bugs here and there. For us it is good to go.

shahor02 (approved):
conditionally approving to trigger CI

iravasen:
Hi @shahor02 , the test which fails does not seem to be related to this code modifications, right? Thanks.

PR #8954, opened 2022-05-25
Add MCH-MID tracks to FwdTracks table

nburmaso:


nburmaso:
Hi @rpezzi. This PR concerns the discussion we had via e-mail some time ago. While doing simulations and analysis of forward tracks, I noticed that there are no MCH-MID tracks in AOD even though MFT-MCH-MID matches are processed, so I added it by checking MCH-MID match info for MCH tracks manually. Could you please take a look at the code and see if it is ok with you?

rpezzi:
Hi @nburmaso , thanks for your contribution. The best person to review this code is @pillot , who told me yesterday he was looking at this.

pillot:
Thanks @nburmaso . I will have a closer look next week, but I already see one problem here. The MCH-MID tracks have a different timing than the MCH-only tracks, meaning that their association with collisions can be different. So they should be first introduced in the VertexTrackMatcher with their own time bracket, then filled into the AODs as a third type of FwdTracks.

pillot:
FYI, I am working on introducing the MCH-MID tracks in the VertexTrackMatcher. Will get back to this PR afterwards...

nburmaso:
Hi @pillot. Thank you for the news. In this case, I am converting this PR into a draft for now, as it will need corrections later.

pillot:
Hi @nburmaso. Everything should be in place to resume this PR.
The MCH-MID matched tracks are now associated with collisions like other forward tracks. The corresponding MCH-standalone tracks are skipped so they are not accounted twice.
MCH-MID tracks also have MC labels now, which are accessible via the RecoContainer like for other tracks.

nburmaso:
Hi @pillot. I have corrected this PR according to your changes in the VertexTrackMatcher. Could you please check the code?

Also, I noticed that you set a very large time bracket in the VertexTrackMatcher for MCH-MID tracks (~0.5 mus). If I understand correctly, MID timings are very precise, so this seems strange to me. In particular, such a large time window may cause problems for reconstruction of events without good vertices, e.g. genuine UPC tracks may be wrongly matched to collision vertices.

If you are fine with it, I would suggest skipping an additional 0.5 mus margin here

https://github.com/AliceO2Group/AliceO2/blob/c3eea64b43281c9a7b70bf4c3b88b9d1ab9b33ce/Detectors/Vertexing/src/VertexTrackMatcher.cxx#L199

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -372,7 +373,7 @@ void AODProducerWorkflowDPL::fillIndexTablesPerCollision(const o2::dataformats::
           mIndexTableMFT[trackIndex.getIndex()] = mIndexMFTID;
           mIndexMFTID++;
 
-        } else if (src == GIndex::Source::MCH || src == GIndex::Source::MFTMCH) {
+        } else if (src == GIndex::Source::MCH || src == GIndex::Source::MFTMCH || src == GIndex::Source::MCHMID) {
```
pillot, line 376: The MCHMID source must be added to the loop too.

Also, in this case, the position in mIndexTableFwd where to store mIndexFwdID is not trackIndex.getIndex() but the index of the associated MCH track, i.e. mchmidMatch.getMCHRef().getIndex()

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -417,8 +419,17 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
     }
   };
 
-  if (trackID.getSource() == GIndex::MCH) { // This is a MCH track
-    fwdInfo.trackTypeId = o2::aod::fwdtrack::MCHStandaloneTrack;
+  auto getMIDBitMapBoards = [&](int midTrackID) {
+    if (midTrackID != -1) { // check matching just in case
+      const auto midTrack = midTracks[midTrackID];
```
pillot: You can use a reference instead of copying the track.
Same for MCH tracks, MCHMID tracks and MCH clusters elsewhere in this function.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -498,14 +508,38 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
...
 
-    const auto& rof = data.getMCHTracksROFRecords()[mMCHROFs[trackID.getIndex()]];
+    return true;
+  };
+
+  if (trackID.getSource() == GIndex::MCH) { // This is an MCH track
+    int mchTrackID = trackID.getIndex();
+    getMCHBitMap(mchTrackID);
+    if (!extrapMCHTrack(mchTrackID)) {
+      return;
```
pillot: If this ever happen, the addition of the track cannot be skipped, otherwise the synchronization with other tables (mIndexTableFwd and MClabels) will be lost. A track with dummy parameters should be added instead, with an error message.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -498,14 +508,38 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
...
     errGaussian = false;
-
-  } else {
-    // This is a GlobalMuonTrack or a GlobalForwardTrack
+  } else if (trackID.getSource() == GIndex::MCHMID) { // This is an MCH-MID track
+    fwdInfo.trackTypeId = o2::aod::fwdtrack::MuonStandaloneTrack;
+    const auto mchmidMatch = mchmidMatches[trackID.getIndex()];
+    const auto& mchTrackID = mchmidMatch.getMCHRef().getIndex();
+    if (!extrapMCHTrack(mchTrackID)) {
+      return;
```
pillot: same as above

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -2001,6 +2029,9 @@ DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src, bool enableSV, boo
   auto dataRequest = std::make_shared<DataRequest>();
 
   dataRequest->requestTracks(src, useMC);
+  if (src[GID::MCHMID]) {
+    dataRequest->requestMCHMIDMatches(useMC);
+  }
```
pillot: This is not needed. It is already done in requestTracks.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Thank you for the comment, I have added MCH-MID to the loop. 

As for the indices, (if I understand correctly) `mIndexFwdID` is used for `MFT-MCH-MID` and `MFT-MCH` tracks to add references to matched `MCH-only` tracks in the ForwardTracks table. In the case of `MCH-MID` tracks MCH tracks are not duplicated so these references shouldn't be needed at all.

pillot:
> Also, I noticed that you set a very large time bracket in the VertexTrackMatcher for MCH-MID tracks (~0.5 mus). If I understand correctly, MID timings are very precise, so this seems strange to me. In particular, such a large time window may cause problems for reconstruction of events without good vertices, e.g. genuine UPC tracks may be wrongly matched to collision vertices.
> 
> If you are fine with it, I would suggest skipping an additional 0.5 mus margin here
> 
> https://github.com/AliceO2Group/AliceO2/blob/c3eea64b43281c9a7b70bf4c3b88b9d1ab9b33ce/Detectors/Vertexing/src/VertexTrackMatcher.cxx#L199

I did not change this part. This is done for every tracks, including e.g. tracks with TOF with an even better time resolution. I don't know the reason behind this common increase of time bracket (which is also done to define BC slices for ambiguous tracks in the AOD creator by the way), so I am not sure if this would be a good idea to skip it for forward tracks. @shahor02 what do you think ?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Thank you for noticing, I switched to references where applicable

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: I removed `return` and added a warning message here. When an `MCH` track can't be extrapolated, dummy parameters from the structures (`FwdTrackInfo` and `FwdTrackCovInfo`) will be used.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Same changes as in lines 515-518 here

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
nburmaso: Removed double `MCH-MID` request

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
pillot: Yes they are need for the `MFT-MCH-MID` tracks to refer to the matched `MCH-MID` tracks.

nburmaso: Sorry, I misunderstood the second part of your comment. I have added reference IDs for `MCH` parts of `MCH-MID` tracks as well

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -357,7 +358,7 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
 
 void AODProducerWorkflowDPL::fillIndexTablesPerCollision(const o2::dataformats::VtxTrackRef& trackRef, const gsl::span<const GIndex>& GIndices)
 {
-  for (int src : {GIndex::Source::MFTMCH, GIndex::Source::MCH, GIndex::Source::MFT}) {
+  for (int src : {GIndex::Source::MFTMCH, GIndex::Source::MCH, GIndex::Source::MFT, GIndex::Source::MCHMID}) {
```
pillot: If I am not wrong, we must loop over sources in the same order as in fillTrackTablesPerCollision, so MCHMID should be first.

nburmaso: Thank you, fixed

pillot (dismissed):

pillot:
Hi @shahor02 , I think this PR is ready to be reviewed. The error in o2-dataflow looks unrelated.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -498,15 +512,39 @@ void AODProducerWorkflowDPL::addToFwdTracksTable(FwdTracksCursorType& fwdTracksC
...
     fwdInfo.trackTimeRes = o2::constants::lhc::LHCBunchSpacingNS * bcWidth / 2; // half interval
     errGaussian = false;
-
-  } else {
-    // This is a GlobalMuonTrack or a GlobalForwardTrack
-    const auto track = data.getGlobalFwdTrack(trackID);
+  } else if (trackID.getSource() == GIndex::MCHMID) { // This is an MCH-MID track
+    fwdInfo.trackTypeId = o2::aod::fwdtrack::MuonStandaloneTrack;
+    const auto& mchmidMatch = mchmidMatches[trackID.getIndex()];
+    const auto& mchTrackID = mchmidMatch.getMCHRef().getIndex();
```
shahor02: creating a reference on int is more expensive (8B instead of 4B + dereferencing) than simple copy, i.e.
```
auto mchmidMatch = mchmidMatches[trackID.getIndex()];
auto mchTrackID = mchmidMatch.getMCHRef().getIndex();
```
The same below.

pillot: I agree for the indices, e.g. `mchTrackID`, but `mchmidMatch` is an object, so better getting a reference rather than copying it.

nburmaso: Hi @shahor02, @pillot. Thank you for noticing the bug, I have removed the references now.

shahor02: well, it does not matter if it is an object or a simple type: being POD object its copy constructor is exactly the same as an assignment of int, so it is cheaper than the references. The reference would be make sense only if you want to modify it and preserve the modifications in the source.

nburmaso: Reference from MCH-MID match is now removed as well

pillot (dismissed):

pillot:
Hi @shahor02 , can you merge this please ? The error is unrelated, everything was green yesterday.

shahor02 (approved):
checks were passed, just made cosmetic change

PR #8962, opened 2022-05-25
[EMCAL-610] Adapting the raw encoder to encode HG and LG bunches

hahassan7:
The raw writer has been adapted to encode both the HG and LG bunches for the same sampled digits.
A time shift has been added to the Digits to Cell converter.

### DIFF
--- a/Detectors/EMCAL/reconstruction/src/CaloRawFitterGamma2.cxx
+++ b/Detectors/EMCAL/reconstruction/src/CaloRawFitterGamma2.cxx
@@ -98,7 +98,7 @@ float CaloRawFitterGamma2::doFit_1peak(int firstTimeBin, int nSamples, float& am
 
   float chi2(0.);
 
-  // fit using gamma-2 function 	(ORDER =2 assumed)
+  // fit using gamma-2 function   (ORDER =2 assumed)
   if (nSamples < 3) {
     throw RawFitterError_t::FIT_ERROR;
   }
--- a/Detectors/EMCAL/reconstruction/src/CaloRawFitterGamma2.cxx
+++ b/Detectors/EMCAL/reconstruction/src/CaloRawFitterGamma2.cxx
@@ -87,7 +87,7 @@ CaloFitResults CaloRawFitterGamma2::evaluate(const gsl::span<const Bunch> bunchl
     time = time * constants::EMCAL_TIMESAMPLE;
     time -= mL1Phase;
 
-    return CaloFitResults(maxADC, pedEstimate, mAlgo, amp, time, (int)time, chi2, ndf);
+    return CaloFitResults(maxADC, pedEstimate, 0, amp, time, (int)time, chi2, ndf);
   }
   // Fit failed, rethrow error
   throw RawFitterError_t::FIT_ERROR;
--- a/Detectors/EMCAL/reconstruction/src/CaloRawFitterStandard.cxx
+++ b/Detectors/EMCAL/reconstruction/src/CaloRawFitterStandard.cxx
@@ -96,7 +96,7 @@ CaloFitResults CaloRawFitterStandard::evaluate(const gsl::span<const Bunch> bunc
     time = time * constants::EMCAL_TIMESAMPLE;
     time -= mL1Phase;
 
-    return CaloFitResults(maxADC, pedEstimate, mAlgo, amp, time, (int)time, chi2, ndf);
+    return CaloFitResults(maxADC, pedEstimate, 0, amp, time, (int)time, chi2, ndf);
   }
   throw RawFitterError_t::FIT_ERROR;
 }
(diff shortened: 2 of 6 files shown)


Detectors/EMCAL/reconstruction/src/CaloRawFitterGamma2.cxx
```diff
@@ -87,7 +87,7 @@ CaloFitResults CaloRawFitterGamma2::evaluate(const gsl::span<const Bunch> bunchl
     time = time * constants::EMCAL_TIMESAMPLE;
     time -= mL1Phase;
 
-    return CaloFitResults(maxADC, pedEstimate, mAlgo, amp, time, (int)time, chi2, ndf);
+    return CaloFitResults(maxADC, pedEstimate, 0, amp, time, (int)time, chi2, ndf);
```
mfasDa, line 90: What is this?

Detectors/EMCAL/reconstruction/src/CaloRawFitterStandard.cxx
```diff
@@ -96,7 +96,7 @@ CaloFitResults CaloRawFitterStandard::evaluate(const gsl::span<const Bunch> bunc
     time = time * constants::EMCAL_TIMESAMPLE;
     time -= mL1Phase;
 
-    return CaloFitResults(maxADC, pedEstimate, mAlgo, amp, time, (int)time, chi2, ndf);
+    return CaloFitResults(maxADC, pedEstimate, 0, amp, time, (int)time, chi2, ndf);
```
mfasDa, line 99: Also here

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -419,21 +419,19 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
...
                   // High gain cell: Flag as low gain if above threshold
                   if (amp / CONVADCGEV > o2::emcal::constants::OVERFLOWCUT) {
-                    flagChanType = ChannelType_t::LOW_GAIN;
                     hgOutOfRange = true;
                   }
                   hwAddressHG = chan.getHardwareAddress();
                 }
                 int fecID = mMapper->getFEEForChannelInDDL(feeID, chan.getFECIndex(), chan.getBranchIndex());
-                currentCellContainer->push_back({o2::emcal::Cell(CellID, amp, fitResults.getTime() - timeshift, flagChanType),
+                currentCellContainer->push_back({o2::emcal::Cell(CellID, amp, fitResults.getTime() - timeshift, chantype),
```
mfasDa, line 434: Also here I am surprise to see this. If the high gain cells are above threshold they must be flagged as low gain. This is absolutely needed for data, to find data corruptions. The change in RawToCellConverter must not be merged.

mfasDa (approved):
There are few changes in the reconstruction part (raw fitter, raw->cell converter) which I am surprised to see. Particularly the change in the raw->cell converter cannot be merged.

Detectors/EMCAL/reconstruction/src/CaloRawFitterGamma2.cxx
hahassan7: The third argument in CaloFitResults takes the fit status. While in the old version we were giving the third argument the fitting algorithm instead of the fitting status which wrong. So now it is fixed, and 0 status means the fit is successful.

Detectors/EMCAL/reconstruction/src/CaloRawFitterStandard.cxx
hahassan7: Same as above.

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
hahassan7: Yes true, but the problem if we set the cell to LG then it will not take the LG bunch fit results written above due to a bug.

mfasDa: I see, you are right this is actually a bug. In AliRoot we were flagging saturated HG without LG as LG cells, but in the raw to cell converter this shadows the the LG cells as the merge condition requires cell type to be different (see block above). Since the HG out-of-range with no LG is rejected in any case in O2 the cell type can stay HG with having the OOR flag set.

Detectors/EMCAL/reconstruction/src/CaloRawFitterStandard.cxx
mfasDa: OK, agreed.

Detectors/EMCAL/reconstruction/src/CaloRawFitterGamma2.cxx
mfasDa: OK, agreed.

shahor02 (approved):

PR #8969, opened 2022-05-25
TRD DCS processor updated

martenole:
This should fix https://alice.its.cern.ch/jira/browse/TRD-109

ping @tdietel 

The compiler fixes are minor and I did not touch the raw reader parts, since Sean is working on them currently

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/DcsCcdbObjects.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/DcsCcdbObjects.h
@@ -1,0 +1,43 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file DcsCcdbObjects.h
+/// \brief Objects which are created from DCS DPs and stored in the CCDB
+/// \author Ole Schmidt, ole.schmidt@cern.ch
+
+#ifndef ALICEO2_DCSCCDBOBJECTSTRD_H
+#define ALICEO2_DCSCCDBOBJECTSTRD_H
+
+#include "DataFormatsTRD/Constants.h"
+#include "Rtypes.h"
+#include <array>
+
+namespace o2
+{
+namespace trd
+{
+
+struct TRDDCSMinMaxMeanInfo {
+  float minValue{0.f};  // min value seen by the TRD DCS processor
+  float maxValue{0.f};  // max value seen by the TRD DCS processor
+  float meanValue{0.f}; // mean value seen by the TRD DCS processor
+  int nPoints{0};       // number of values seen by the TRD DCS processor
+
+  void print() const;
+  void addPoint(float value);
+
+  ClassDefNV(TRDDCSMinMaxMeanInfo, 1);
+};
+
+} // namespace trd
+} // namespace o2
+
+#endif // ALICEO2_DCSCCDBOBJECTSTRD_H
--- a/Detectors/TRD/calibration/README.md
+++ b/Detectors/TRD/calibration/README.md
@@ -16,3 +16,11 @@ If you want to run the calibration from a local file with residuals, trdangreshi
     o2-calibration-trd-vdrift-exb -b --enable-root-input --calib-vdexb-calibration '--tf-per-slot 1 --min-entries 50000'
 
 ## DCS data points
+
+To process the DCS data points for the TRD the list of aliases as for example "trd_gaschromatographXe" has to be available in the CCDB. This can be achieved with the macro `Detectors/TRD/calibration/macros/makeTRDCCDBEntryForDCS.C`. The full list of aliases is available in jira (https://alice.its.cern.ch/jira/browse/TRD-109).
+With the list of aliases defined one can run the `o2-calibration-trd-dcs-sim-workflow` which provides DCS DPs for all possible aliases and sends them on via DPL. Attaching the `o2-calibration-trd-dcs-workflow` will include the processing of these data points. For testing purposes this is sufficient. In case also the CCDB should be populated the `o2-calibration-ccdb-populator-workflow` has to be appended. Via the `--ccdb-path` flag this can also be configured to write to a local CCDB for testing.
+
+So, in order to test the workflow independent of the actual CCDB using a local instance one can do:
+
+    root $O2_ROOT/share/macro/makeTRDCCDBEntryForDCS.C+
+    o2-calibration-trd-dcs-sim-workflow -b --delta-fraction 0.5 --max-timeframes 10 | o2-calibration-trd-dcs-workflow -b --ccdb-path http://localhost:8080 --use-ccdb-to-configure --processor-verbosity 1 | o2-calibration-ccdb-populator-workflow -b --ccdb-path http://localhost:8080
--- a/Detectors/TRD/calibration/macros/readTRDDCSentries.C
+++ b/Detectors/TRD/calibration/macros/readTRDDCSentries.C
@@ -9,12 +9,14 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-// macro to read the TOF DCS information from CCDB
-// default ts is very big: Saturday, November 20, 2286 5:46:39 PM
+// macro showing how to read a CCDB object created from DCS data points
 
 #if !defined(__CLING__) || defined(__ROOTCLING__)
+#include "CCDB/BasicCCDBManager.h"
 #include "CCDB/CcdbApi.h"
 #include "DetectorsDCS/DataPointIdentifier.h"
+
+#include "TRDCalibration/DCSProcessor.h"
 
 #include <string>
 #include <unordered_map>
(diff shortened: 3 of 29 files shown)


DataFormats/Detectors/TRD/include/DataFormatsTRD/DcsCcdbObjects.h
```diff
@@ -0,0 +1,43 @@
...
+#include "DataFormatsTRD/Constants.h"
+#include "Rtypes.h"
+#include <array>
+
+namespace o2
+{
+namespace trd
+{
+
+struct TRDDCSMinMaxMeanInfo {
```
tdietel, line 28: I'd suggest to add getters for min,max,mean.
This would allow us to change the implementation, e.g. if we want to add rms (maybe in a derived class). But maybe I'm overthinking the OOP design.

Detectors/TRD/calibration/macros/readTRDDCSentries.C
```diff
@@ -9,60 +9,56 @@
...
+#include "TRDCalibration/DCSProcessor.h"
+
 #include <string>
 #include <unordered_map>
 #include <chrono>
 #include <bitset>
 #endif
 
-void readTRDDCSentries(long ts = 9999999999000, const char* ccdb = "http://localhost:8080")
+void readTRDDCSentries(std::string ccdb = "http://ccdb-test.cern.ch:8080", long ts = -1)
```
tdietel, line 27: This could be a useful non-expert tool. Could you add some usage remarks:
- URL of the production CCDB (maybe even make it the default)
- a good value for ts, or a hint how to find it

A use case would be to get the DCS data for a given run or wall time period ("1-15 May"). But maybe that should go to a readme or a separate macro / PR.

Detectors/TRD/calibration/src/KrClusterFinder.cxx
```diff
@@ -29,7 +29,7 @@ double KrClusterFinder::LandauChi2Functor::operator()(const double* par) const
   // par[1] : location parameter (approximately most probable value)
   // par[2] : sigma
   double retVal = 0;
-  for (unsigned int i = xLowerBound; i <= xUpperBound; ++i) {
+  for (int i = xLowerBound; i <= xUpperBound; ++i) {
```
tdietel, line 32: Seems unrelated, but minor change.

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
```diff
@@ -15,38 +15,36 @@
...
   dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvDriftImon[00..539]", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvDriftUmon[00..539]", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<int32_t>{"trd_fedChamberStatus", 0, 255});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<int32_t>{"trd_runNo", 0, 255});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<int32_t>{"trd_runType", 0, 255});
-  */
-  o2::framework::WorkflowSpec specs;
-  specs.emplace_back(o2::dcs::test::getDCSRandomDataGeneratorSpec(dphints, "TRD"));
-  return specs;
+  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvDriftUmon[00..539]", 49., 50.});
```
tdietel, line 32: Can we add more realistic values here? I am thinking about plotting and auto-ranging, and possible problems if we have very small or large values.
And I don't know if the anode/drift currents are stored in A/mA/uA/nA. I think ampere (A) would make sense, but I remember it's something different.

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
```diff
@@ -15,38 +15,36 @@
...
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gasH2O", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gasO2", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gaschromatographCO2", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gaschromatographN2", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gaschromatographXe", 0, 50.});
+
+  // HV parameters
   dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvAnodeImon[00..539]", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvAnodeUmon[00..539]", 0, 50.});
+  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvAnodeUmon[00..539]", 49., 50.});
```
tdietel, line 30:
```suggestion
  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvAnodeUmon[00..539]", 1549., 1550.});
```

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
```diff
@@ -15,38 +15,36 @@
...
   dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvDriftImon[00..539]", 0, 50.});
```
tdietel, line 31:
```suggestion
  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvDriftImon[00..539]", 100.e-6, 150.e-6});
```

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
```diff
@@ -15,38 +15,36 @@
...
-  /*
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gasCO2", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gasH2O", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gasO2", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gaschromatographCO2", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gaschromatographN2", 0, 50.});
-  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_gaschromatographXe", 0, 50.});
+
+  // HV parameters
   dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvAnodeImon[00..539]", 0, 50.});
```
tdietel, line 29:
```suggestion
  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"trd_hvAnodeImon[00..539]", 0, 20.e-9});
```

Detectors/TRD/calibration/README.md
```diff
@@ -16,3 +16,11 @@ If you want to run the calibration from a local file with residuals, trdangreshi
     o2-calibration-trd-vdrift-exb -b --enable-root-input --calib-vdexb-calibration '--tf-per-slot 1 --min-entries 50000'
 
 ## DCS data points
+
+To process the DCS data points for the TRD the list of aliases as for example "trd_gaschromatographXe" has to be available in the CCDB. This can be achieved with the macro `Detectors/TRD/calibration/macros/makeTRDCCDBEntryForDCS.C`. The full list of aliases is available in jira (https://alice.its.cern.ch/jira/browse/TRD-109).
+With the list of aliases defined one can run the `o2-calibration-trd-dcs-sim-workflow` which provides DCS DPs for all possible aliases and sends them on via DPL. Attaching the `o2-calibration-trd-dcs-workflow` will include the processing of these data points. For testing purposes this is sufficient. In case also the CCDB should be populated the `o2-calibration-ccdb-populator-workflow` has to be appended. Via the `--ccdb-path` flag this can also be configured to write to a local CCDB for testing.
+
+So, in order to test the workflow independent of the actual CCDB using a local instance one can do:
+
+    root $O2_ROOT/share/macro/makeTRDCCDBEntryForDCS.C+
```
tdietel, line 25: Interesting, does this need to be compiled?

tdietel (commented):
Looks pretty good. I have a few minor comments, mostly on clarifications.

I noticed a lot of cleanup (signed/unsigned, size_t/int replacements). This made this PR longer than necessary. But it's probably not worth the effort to split it now.

martenole:
> I noticed a lot of cleanup (signed/unsigned, size_t/int replacements). This made this PR longer than necessary. But it's probably not worth the effort to split it now.

It is two different commits. If you go to the "Commits" tab of this PR and then select the DCS processor one you can review only the changes related to the DCS processor. Of course I could have and probably should have just opened two PRs...

Detectors/TRD/calibration/README.md
martenole: Actually no, also works without the `+`

DataFormats/Detectors/TRD/include/DataFormatsTRD/DcsCcdbObjects.h
martenole: I think this will be a read-only object and then it's a struct so there are no getters needed for the members. In case we would like to add getters later I believe that this does not change the ROOT streamer information and thus would still be possible. Only if we add/remove members then the streamer information would need to be updated

Detectors/TRD/calibration/src/KrClusterFinder.cxx
martenole: This belongs to the other commit which addresses some of the compiler warnings which are sent around via a weekly mail now.

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
martenole: According to the jira ticket the currents are stored in uA. So the numbers are not very small or very large and I don't think there will be issues with the processor because of this. Anyways, I can update the values of course

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
tdietel: Ok, if we have the prefix units, it's probably ok. Only if we want to distinguish what's where, but the key test will anyways be with real data.

Detectors/TRD/calibration/workflow/trd-dcs-sim-workflow.cxx
tdietel: I actually just noticed the `00..539`. I would expect either `0..539` or `000..539` - but I don't know anything about the convention here, and possible leading zeros in the expansion. Will this be handled correctly?

martenole: Yes, this is done correctly by the alias expansion inside the DCS framework

Detectors/TRD/calibration/macros/readTRDDCSentries.C
martenole: This will be done in a followup PR

PR #8975, opened 2022-05-26
o2-eve-workflow: filter on sorted track list

jmyrcha:


davidrohr (approved):

davidrohr (changes requested):
Actually, one small change request:
Could we add a command line option to enable/disable the sorting (should default to true).
And in case the TotalNTracks filter is not set, the sorting should be skipped in any case.

Finally, a minor comment: Instead of the foreach with the check of the TotalNTracks filter inside the loop, you could use a normal for loop and just compute ahead of time to run either till trackTimeNodes.size() or mMaxNTracks.

davidrohr (approved):

PR #8981, opened 2022-05-26
fix TOF matching source in dpl workflow (sync)

noferini:
This is required for https://github.com/AliceO2Group/O2DPG/pull/426
in order to include TOF matching QC (with and without TRD depending on the detector sources)
both option tested (assuming both PRs: O2 and O2DPG)

### DIFF
--- a/prodtests/full-system-test/workflow-setup.sh
+++ b/prodtests/full-system-test/workflow-setup.sh
@@ -56,9 +56,9 @@ has_detectors_reco ITS TPC && has_detector_matching ITSTPC && add_comma_separate
 has_detectors_reco TPC TRD && has_detector_matching TPCTRD && { add_comma_separated TRD_SOURCES TPC; add_comma_separated TRACK_SOURCES "TPC-TRD"; }
 has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRD && { add_comma_separated TRD_SOURCES ITS-TPC; add_comma_separated TRACK_SOURCES "ITS-TPC-TRD"; }
 has_detectors_reco TPC TOF && has_detector_matching TPCTOF && { add_comma_separated TOF_SOURCES TPC; add_comma_separated TRACK_SOURCES "TPC-TOF"; }
-has_detectors_reco ITS TPC TOF && has_detector_matching ITSTPCTOF && { add_comma_separated TOF_SOURCES ITS-TPC; add_comma_separated TRACK_SOURCES "ITS-TPC-TOF"; }
-has_detectors_reco TPC TRD TOF && has_detector_matching TPCTRDTOF && { add_comma_separated TOF_SOURCES TPC-TRD; add_comma_separated TRACK_SOURCES "TPC-TRD-TOF"; }
-has_detectors_reco ITS TPC TRD TOF && has_detector_matching ITSTPCTRDTOF && { add_comma_separated TOF_SOURCES ITS-TPC-TRD; add_comma_separated TRACK_SOURCES "ITS-TPC-TRD-TOF"; }
+has_detectors_reco ITS TPC TOF && has_detector_matching ITSTPC && has_detector_matching ITSTPCTOF && { add_comma_separated TOF_SOURCES ITS-TPC; add_comma_separated TRACK_SOURCES "ITS-TPC-TOF"; }
+has_detectors_reco TPC TRD TOF && has_detector_matching TPCTRD && has_detector_matching TPCTRDTOF && { add_comma_separated TOF_SOURCES TPC-TRD; add_comma_separated TRACK_SOURCES "TPC-TRD-TOF"; }
+has_detectors_reco ITS TPC TRD TOF && has_detector_matching ITSTPCTRD && has_detector_matching ITSTPCTRDTOF && { add_comma_separated TOF_SOURCES ITS-TPC-TRD; add_comma_separated TRACK_SOURCES "ITS-TPC-TRD-TOF"; }
 has_detectors_reco MFT MCH && has_detector_matching MFTMCH && add_comma_separated TRACK_SOURCES "MFT-MCH"
 has_detectors_reco MCH MID && has_detector_matching MCHMID && add_comma_separated TRACK_SOURCES "MCH-MID"
 for det in `echo $LIST_OF_DETECTORS | sed "s/,/ /g"`; do


prodtests/full-system-test/workflow-setup.sh
```diff
@@ -55,12 +55,15 @@ TRACK_SOURCES=
 has_detectors_reco ITS TPC && has_detector_matching ITSTPC && add_comma_separated TRACK_SOURCES "ITS-TPC"
 has_detectors_reco TPC TRD && has_detector_matching TPCTRD && { add_comma_separated TRD_SOURCES TPC; add_comma_separated TRACK_SOURCES "TPC-TRD"; }
 has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRD && { add_comma_separated TRD_SOURCES ITS-TPC; add_comma_separated TRACK_SOURCES "ITS-TPC-TRD"; }
-has_detectors_reco TPC TOF && has_detector_matching TPCTOF && { add_comma_separated TOF_SOURCES TPC; add_comma_separated TRACK_SOURCES "TPC-TOF"; }
+has_detectors_reco TPC TOF && { add_comma_separated TOF_SOURCES TPC; add_comma_separated TRACK_SOURCES "TPC-TOF"; }
 has_detectors_reco ITS TPC TOF && has_detector_matching ITSTPCTOF && { add_comma_separated TOF_SOURCES ITS-TPC; add_comma_separated TRACK_SOURCES "ITS-TPC-TOF"; }
-has_detectors_reco TPC TRD TOF && has_detector_matching TPCTRDTOF && { add_comma_separated TOF_SOURCES TPC-TRD; add_comma_separated TRACK_SOURCES "TPC-TRD-TOF"; }
+has_detectors_reco TPC TRD TOF && has_detector_matching TPCTRD && { add_comma_separated TOF_SOURCES TPC-TRD; add_comma_separated TRACK_SOURCES "TPC-TRD-TOF"; }
```
shahor02: Hi @noferini 
in this way you have unconditionally TPC-TOF matching (provided detectors are there) while the TPC-TRD-TOF is conditioned by the TPC-TRD matching and ITS-TPC-TRD-TOF by ITS-TPC-TRD.
What is the reason of such an asymmetric request? 
Note that the TRD matching in the syncmode is limited to those IRframes which where selected by the ITS so we can downscale the amount of tracks processed by the TRD. If TOF matching it conditioned by the ITS-TPC or ITS-TPC-TRD, then the TOF is also automatically downscaled. But enabling unconditional TOF matching to TPC w/o any downscaling may be to CPU consuming at high load. Why do you need to enable it by default?

noferini: Hi Ruben,
I tried to follow the logic I found.

For TPC I needed to assume it is always there since I cannot add "TPC" to WORKFLOW_DETECTORS_MATCHING (it triggers strange behaviors in other workflow)

For other combinations I added the extra requirement if included in WORKFLOW_DETECTORS_MATCHING

Note that if ITS-TPC tracks are not there TOF_SOURCE is not used by any TOF workflow

shahor02: But what was wrong with looking on the `has_detector_matching TPCTOF` etc ?

noferini: Ok, I restored the check with TPCTOF.

My doubt is the following.
With this logic (checking *track*TOF instead of *track*) it is in principle possible that somebody add
TPCTRDTOF to WORKFLOW_DETECTORS_MATCHING but not TPCTRD

This will trigger an error since it will add a TOF_SOURCE TPCTRD in TOF workflow which is not provided as input since dependencies are not requested by WORKFLOW_DETECTORS_MATCHING .

shahor02: Well, you can set as a condition of TPCTRDTOF `  has_detector_matching TPCTRD && has_detector_matching TPCTRDTOF`, though there are many other ways to screw up the workflow.

But why do you want to add unconstrained TPC matching to TOF in the sync. workflow. Did you check up to which occupancy is it feasible? Note that even TPC-TRD is not enables by default, although it is able to work with prescaled data.

noferini: I started from the same TOF matching QC json.
If we decid we are not interested to it in sync phase I can easily remove it

shahor02: Before enabling unconstrained TPC track matching in the sync phase, I would make TOF matching able to use IRFrames like TRD matching is doing, this would at least allow to control the fraction of tracks being processed.

noferini: Ok, I'll comment it for the moment. Is it ok?

noferini: It is a bit more complicated. I need TPC in TOF_SOURCE for QC. I have to work a bit to decouple QC from matcher

shahor02: ok

prodtests/full-system-test/workflow-setup.sh
```diff
@@ -56,11 +56,14 @@ has_detectors_reco ITS TPC && has_detector_matching ITSTPC && add_comma_separate
...
-has_detectors_reco TPC TRD TOF && has_detector_matching TPCTRDTOF && { add_comma_separated TOF_SOURCES TPC-TRD; add_comma_separated TRACK_SOURCES "TPC-TRD-TOF"; }
-has_detectors_reco ITS TPC TRD TOF && has_detector_matching ITSTPCTRDTOF && { add_comma_separated TOF_SOURCES ITS-TPC-TRD; add_comma_separated TRACK_SOURCES "ITS-TPC-TRD-TOF"; }
+has_detectors_reco ITS TPC TOF && has_detector_matching ITSTPC && has_detector_matching ITSTPCTOF && { add_comma_separated TOF_SOURCES ITS-TPC; add_comma_separated TRACK_SOURCES "ITS-TPC-TOF"; }
+has_detectors_reco TPC TRD TOF && has_detector_matching TPCTRD && has_detector_matching TPCTRDTOF && { add_comma_separated TOF_SOURCES TPC-TRD; add_comma_separated TRACK_SOURCES "TPC-TRD-TOF"; }
+has_detectors_reco ITS TPC TRD TOF && has_detector_matching ITSTPCTRD && has_detector_matching ITSTPCTRDTOF && { add_comma_separated TOF_SOURCES ITS-TPC-TRD; add_comma_separated TRACK_SOURCES "ITS-TPC-TRD-TOF"; }
 has_detectors_reco MFT MCH && has_detector_matching MFTMCH && add_comma_separated TRACK_SOURCES "MFT-MCH"
 has_detectors_reco MCH MID && has_detector_matching MCHMID && add_comma_separated TRACK_SOURCES "MCH-MID"
+
+echo "TOF_SOURCES=$TOF_SOURCES"
```
davidrohr: Please remove the printout, there should be no messages. And if at all they must be forwarded to stderr.

prodtests/full-system-test/workflow-setup.sh
```diff
@@ -8,7 +8,7 @@ SOURCE_GUARD_WORKFLOW_SETUP=1
 #Some additional settings used in this workflow
 
 if [[ $SYNCMODE == 1 ]]; then
-  if [[ -z "${WORKFLOW_DETECTORS_MATCHING+x}" ]]; then export WORKFLOW_DETECTORS_MATCHING="ITSTPC,ITSTPCTRD,ITSTPCTOF,ITSTPCTRDTOF,PRIMVTX"; fi # Select matchings that are enabled in sync mode
+  if [[ -z "${WORKFLOW_DETECTORS_MATCHING+x}" ]]; then export WORKFLOW_DETECTORS_MATCHING="ITSTPC,ITSTPCTRD,ITSTPCTOF,ITSTPCTRDTOF,TPCTRD,TPCTRDTOF,TPCTOF,PRIMVTX"; fi # Select matchings that are enabled in sync mode
```
davidrohr: Here I agree with Ruben. Do we really want unconditional matching enabled. Before we do so, we should test in a synthetic run with full rate whether this works. The WORKFLOW_DETECTORS_MATCHING can be overridden in the extra env variables for this.
For reference, there is also no unconstrained TPCTRD matching enabled in sync mode, only in async.

davidrohr (commented):
The additional `has_detector_matching` check makes sense, but I am not so sure about running more matchings in sync mode. Ther performance must be checked in Pb-Pb.

prodtests/full-system-test/workflow-setup.sh
noferini: I just keep TPC in track sources since it is needed by QC

prodtests/full-system-test/workflow-setup.sh
noferini: I just keep TPC in track sources since it is needed by QC

prodtests/full-system-test/workflow-setup.sh
noferini: I just keep TPC in track sources since it is needed by QC

noferini:
I was wrong, there is no need to insert TPC in TRACK_SOURCE
I restored as it is, I added just the additional check to for consistency whit TOF workflow

PR #8982, opened 2022-05-26
Multiple changes for DCS DP processing

shahor02:
* dcs-proxy: group outputs per channel, option to send IBF first.    
The output from the proxy will be grouped per channel. Unless `--may-send-delta-first` option is explicitly requested, the dcs-proxy will delay its 1st output until the FBI is seen
* Make sure GRPMagField data is complete before uploading to CCDB
* Suppress unused instances of CcdbApi
* Change INFINITE_TIMESTAMP to MONTH in TOF, EMC and MFT processors 

@chiarazampolli @noferini @mpoghos @syano0822 Note that I've changed in your processors the validity time from INFINITE to 1 month. If needed, the objects validity may be extended later. I have some ideas on how to limit the number of overlaps w/o the risk of creating holes in the coverage.

### DIFF
--- a/Detectors/DCS/testWorkflow/src/DCStoDPLconverter.h
+++ b/Detectors/DCS/testWorkflow/src/DCStoDPLconverter.h
@@ -19,6 +19,7 @@
 #include "DetectorsDCS/DataPointIdentifier.h"
 #include "DetectorsDCS/DataPointValue.h"
 #include "DetectorsDCS/DataPointCompositeObject.h"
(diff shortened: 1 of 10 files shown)


chiarazampolli:
Hi Ruben,

thank you! I will implement a forced storing for TOF every 2 weeks, so we should not have holes.

chiara

shahor02:
All tests (except o2-dataflow-cs8  failing on testBasicCCDBManager, which is unrelated) were green, merging after custom squashing.

PR #8989, opened 2022-05-27
Configurable pipeline length

ktf:


davidrohr (approved):

ktf:
In principle the completion policy allows overriding this already, but the extra env var is probably better...

ktf:
Merged because it fixes hyperloop (with a pipeline of 16).

davidrohr:
Do we understand why it fails on hyperloop with a longer pipeline length?
This could hint to another bug somewhere?

ktf:
I think I need to move the rate limiting inside the loop which creates new timeslices, otherwise on the first iteration will end up creating all of them and do the rate limiting too late.

davidrohr:
ok, that makes sense

ktf:
Notice this only affect the rate limiting on shared memory usage, not the one on the timeframes (at least based  on how it's implemented right now).

PR #8990, opened 2022-05-27
Use DataTakingContext and TimingInfo provided by the framework

shahor02:


ktf (dismissed):

Framework/Core/src/CommonServices.cxx
```diff
@@ -185,7 +185,7 @@ o2::framework::ServiceSpec CommonServices::datatakingContextSpec()
       }
       auto extEnvId = services.get<RawDeviceService>().device()->fConfig->GetProperty<std::string>("environment_id", "unspecified");
       if (extEnvId != "unspecified") {
-        context.runType = extEnvId;
+        context.envId = extEnvId;
       }
```
shahor02, line 189: @ktf There was actually a typo in the assignment of the envId. 
In the end, I left it by default at `unknown` and checking it before using.

PR #8991, opened 2022-05-27
Update ADM for TracksIU/Tracks idx equivalence

ddobrigk:


Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -965,6 +965,7 @@ using McParticle = McParticles::iterator;
 namespace soa
 {
 DECLARE_EQUIVALENT_FOR_INDEX(aod::StoredMcParticles_000, aod::StoredMcParticles_001);
+DECLARE_EQUIVALENT_FOR_INDEX(aod::StoredTracks, aod::StoredTracksIU);
```
jgrosseo, line 968: Shouldn't we do the same for the Cov matrix?

ddobrigk: I asked the same question to @aalkin but apparently the counter is based solely on StoredTracks (and that's what is used for the cov matrices) so that's not necessary! @aalkin , could you confirm, to be sure? Thanks!

aalkin: Yes, the base for the index is the same as the base for the join, it is enough to only declare equivalence for the tracks. TracksCovs and others are never (expected to be) used not in a join with Tracks.

jgrosseo: I see. Thanks

jgrosseo (approved):

PR #8994, opened 2022-05-27
Fix TOF DCS DP processing

shahor02:
Reset DP acknowledgment bit at the start of new period

@chiarazampolli I saw your comment in https://github.com/AliceO2Group/AliceO2/pull/8992 only after merging it. 
Indeed, the same problem was in TOF processor, applying the same fix. 
It would crashe only if some DP seen in some interval, does not arrive during the following interval, apparently it does not happen with TOF.

Since the fix is trivial, merging.


chiarazampolli:
Hi @shahor02 ,

ok, thanks. I don’t understand why the simulation did never crash, but the logic now is correct. Maybe it is because the FBI is sent in between and i never saw it.

I will also think about a change to not store if not all DPs are seen.

chiara

PR #8998, opened 2022-05-28
fix in TOF CalibApi initialization

noferini:
Array initialization is no longer working (really not clear for me).
I couldn't find any way to attribute a value to all array elements within the constructor implementation.
I implemented initialization for all arrays in the header.
Now it works.

NB: the TOF clusterization was "broken" for this reason in the last tags (it seems a recent problem) since the majority of the channels were flagged as trmError or noisy leading to a very low TOF ME.

This fix is required to resume LHC21k6 production
@chiarazampolli

### DIFF
--- a/Detectors/TOF/base/include/TOFBase/CalibTOFapi.h
+++ b/Detectors/TOF/base/include/TOFBase/CalibTOFapi.h
@@ -39,9 +39,9 @@ class CalibTOFapi
 
  public:
   void resetDia();
-  CalibTOFapi();
+  CalibTOFapi() = default;
   CalibTOFapi(const std::string url);
-  CalibTOFapi(long timestamp, o2::dataformats::CalibLHCphaseTOF* phase, o2::dataformats::CalibTimeSlewingParamTOF* slew, Diagnostic* dia = nullptr) : mTimeStamp(timestamp), mLHCphase(phase), mSlewParam(slew), mDiaFreq(dia) { CalibTOFapi(); }
+  CalibTOFapi(long timestamp, o2::dataformats::CalibLHCphaseTOF* phase, o2::dataformats::CalibTimeSlewingParamTOF* slew, Diagnostic* dia = nullptr) : mTimeStamp(timestamp), mLHCphase(phase), mSlewParam(slew), mDiaFreq(dia) {}
   ~CalibTOFapi()
   {
     if (mLHCphase) {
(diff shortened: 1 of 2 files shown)


Detectors/TOF/base/include/TOFBase/CalibTOFapi.h
```diff
@@ -107,17 +117,17 @@ class CalibTOFapi
...
+  float mEmptyTOF = 0;                               ///< probability to have TOF fully empty
+  float mEmptyCrateProb[Geo::kNCrate] = {INITF_ALL}; ///< probability to have an empty crate in the current readout window
+  std::vector<std::pair<int, float>> mNoisy;         ///< probTRMerror
+  std::vector<std::pair<int, float>> mTRMerrorProb;  ///< probTRMerror
+  std::vector<int> mTRMmask;                         ///< mask error for TRM
+
+  bool mIsErrorCh[Geo::NCHANNELS] = {INITB_ALL}; ///< channels in error (TRM)
+  std::vector<int> mFillErrChannel;              ///< last error channels filled
+  bool mIsOffCh[Geo::NCHANNELS] = {INITB_ALL};   ///< channels in error (TRM)
+  bool mIsNoisy[Geo::NCHANNELS] = {INITB_ALL};   ///< noisy channels
```
shahor02: @noferini You did not have array initialization at all, otherwise it would work :)
The code like 
```
float mEmptyCrateProb[Geo::kNCrate];
```
leaves array non-initilized. To set all elements to 0 it is enough to do
```
float mEmptyCrateProb[Geo::kNCrate] = {};
```
The same for all arrays you have (also the std::arrays)

noferini: I did a memset in the constructor

noferini: and it worked

noferini: and it still works outside the constructor

shahor02:
see my reply to your email. Before it worked by chance only.

Detectors/TOF/base/include/TOFBase/CalibTOFapi.h
```diff
@@ -107,17 +117,17 @@ class CalibTOFapi
...
-  std::vector<std::pair<int, float>> mTRMerrorProb; ///< probTRMerror
-  std::vector<int> mTRMmask;                        ///< mask error for TRM
-
-  bool mIsErrorCh[Geo::NCHANNELS];  ///< channels in error (TRM)
-  std::vector<int> mFillErrChannel; ///< last error channels filled
-  bool mIsOffCh[Geo::NCHANNELS];    ///< channels in error (TRM)
-  bool mIsNoisy[Geo::NCHANNELS];    ///< noisy channels
+  int mNoisyThreshold = 1;                           ///< threshold to be noisy
+  float mEmptyTOF = 0;                               ///< probability to have TOF fully empty
+  float mEmptyCrateProb[Geo::kNCrate] = {INITF_ALL}; ///< probability to have an empty crate in the current readout window
```
shahor02: you don't need these `{XXX}` with defines, just use `{}`.

Detectors/TOF/base/include/TOFBase/CalibTOFapi.h
noferini: After email discussion the bug was understood (it was a my fault in previous implementation)
Thanks!

Detectors/TOF/base/include/TOFBase/CalibTOFapi.h
noferini: Sorry, I didn't understand the comment. What are you proposing?

shahor02: I mean remove all these `#define`s and simple declare your datamembers as
```
float mEmptyCrateProb[Geo::kNCrate] = {};
...
bool mIsErrorCh[Geo::NCHANNELS] = {};
bool mIsOffCh[Geo::NCHANNELS] = {};
bool mIsNoisy[Geo::NCHANNELS] = {};
```

noferini: done

shahor02 (dismissed):

noferini:
I'm sorry I forgot clang. Now it is in. It needs to be re-approved

shahor02 (approved):

shahor02:
failure is unrelated

PR #9008, opened 2022-05-28
DPL GUI: display metrics for input and output channels

ktf:
This adds labels to each of the channels in input and in output.

In particular the current implementation allows to display the
oldest possible timeslice as seen by the inputs and the one notified
on the output.

Notice that due to the way sending works, the metric on the output
channel might be somewhat delayed wrt its real value.

### DIFF
--- a/Framework/Core/CMakeLists.txt
+++ b/Framework/Core/CMakeLists.txt
@@ -111,6 +111,7 @@ o2_add_library(Framework
                        src/ResourcePolicyHelpers.cxx
                        src/SendingPolicy.cxx
                        src/ServiceRegistry.cxx
+                       src/ServiceSpec.cxx
                        src/SimpleResourceManager.cxx
                        src/SimpleRawDeviceService.cxx
                        src/StreamOperators.cxx
--- a/Framework/Core/include/Framework/CommonServices.h
+++ b/Framework/Core/include/Framework/CommonServices.h
@@ -81,6 +81,7 @@ struct CommonServices {
   static ServiceSpec ccdbSupportSpec();
   static ServiceSpec decongestionSpec();
   static ServiceSpec asyncQueue();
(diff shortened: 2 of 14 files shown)


ktf:
@davidrohr as discussed, this provides the oldest possible timeframe as labels to the channels. It should also significantly speedup some GUI related processing for metrics, but I would still advise to run with DPL_DEFAULT_PIPELINE_LENGTH=32...

davidrohr:
I would propose to merge at least the speedup of the GUI soon.
The display of the oldestPossible values could be made optional, I don't think they are always needed.
Also, there is still the problem that the GUI gets messed up when too many devices are present.

ktf:
New PR with just the speedup part at https://github.com/AliceO2Group/AliceO2/pull/9082

ktf:
@matthias-kleiner could you please apply this PR and send me a screenshot of the multiple lanes workflow you have which is getting stuck?

matthias-kleiner:
> @matthias-kleiner could you please apply this PR and send me a screenshot of the multiple lanes workflow you have which is getting stuck?

Yes sure, I will do it. Thanks for your help

davidrohr:
I would opt for making this an optional feature enabled on command line / via env variable, and then merge it in O2 :)

ktf:
Ok, since I came handy also to debug the forwarding, I added the generic support for enabling / disabling services. This can be done by setting either DPL_OVERRIDE_SERVICES or DPL_DRIVER_OVERRIDE_SERVICES, where the former only apply to the dataprocessors. They take a comma separate list of `<device-name>:<enable/disable>` pairs. For this specific case:

```bash
DPL_OVERRIDE_SERVICES=gui-metrics:enable # or 1
```

will behave as requested.

ktf:
Missing nullptr fixed.

PR #9009, opened 2022-05-28
DPL: have timers start at a random point within an interval

ktf:


Framework/Core/src/LifetimeHelpers.cxx
```diff
@@ -98,7 +99,11 @@ ExpirationHandler::Creator LifetimeHelpers::enumDrivenCreation(size_t start, siz
 
 ExpirationHandler::Creator LifetimeHelpers::timeDrivenCreation(std::chrono::microseconds period)
 {
-  auto start = getCurrentTime();
+  std::random_device r;
+  std::default_random_engine e1(r());
+  std::uniform_int_distribution<uint64_t> dist(0, period.count());
```
davidrohr: I thought about it again, and perhaps we could use `(-period.count() / 2, period.count() / 2)` as random range?

ktf: My  worry is that then `delta` below overflows and we end up with a triggered timer in the future. Maybe the correct thing to do is: `-rand(0, period.count() / 2) - period.count() / 2` so that we still guarantee at least half of the period for the first timer duration, but we are still random enough to make it sustainable on the receiving side...

davidrohr: but then you do not have proper randomization, you randomize only over half the period. I would not do that.

ktf: Well, we can tweak that to include 3/4 of the period of something like that. I think triggering a QC task with a time in the future is more dangerous if the value of the timestamp is actually used e.g. for a running average or things like that.

davidrohr:
ok, then let's randomize over 90% and add 10% as offset.

ktf:
@davidrohr deal. Updated accordingly.

ktf:
@Barthelemy @knopers8 could you green light this? I think it goes in the direction you were asking as well.

Barthelemy (approved):

PR #9014, opened 2022-05-29
ctf-writer logs per-det. data sizes every --report-data-size-interval TFs

shahor02:
@davidrohr e.g. for FST reports:
```
[IMPORTANT] CTF#0 size: ITS:1156776 TPC:243714102 TRD:14806105 TOF:355787 PHS:87223 CPV:479847 EMC:109134 HMP:646 MFT:390180 MCH:879173 MID:267335 ZDC:502826 FT0:35119 FV0:1476 FDD:3090 CTP:3602
```

### DIFF
--- a/Detectors/CTF/workflow/include/CTFWorkflow/CTFWriterSpec.h
+++ b/Detectors/CTF/workflow/include/CTFWorkflow/CTFWriterSpec.h
@@ -24,7 +24,7 @@ namespace ctf
 {
 
 /// create a processor spec
-framework::DataProcessorSpec getCTFWriterSpec(o2::detectors::DetID::mask_t dets, uint64_t run, const std::string& outType, int verbosity);
+framework::DataProcessorSpec getCTFWriterSpec(o2::detectors::DetID::mask_t dets, uint64_t run, const std::string& outType, int verbosity, int reportInterval);
 
 } // namespace ctf
 } // namespace o2
--- a/Detectors/CTF/workflow/src/CTFWriterSpec.cxx
+++ b/Detectors/CTF/workflow/src/CTFWriterSpec.cxx
@@ -92,7 +92,7 @@ class CTFWriterSpec : public o2::framework::Task
 {
  public:
   CTFWriterSpec() = delete;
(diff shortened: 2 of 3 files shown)


davidrohr:
Thx, this looks good. Do we actually have a possibility with FairLogger to format numbers with thousand separators? That would helpful?
@rbx

shahor02:
> Thx, this looks good. Do we actually have a possibility with FairLogger to format numbers with thousand separators? That would helpful? @rbx

fmt::format supports this, but this requires setting a locale and either DPL or FMQ has have problems with that so had to apply the workaround https://github.com/AliceO2Group/AliceO2/blob/8d6382c5b02b2f9982db322f068aed5195db2465/Detectors/CTF/workflow/src/CTFWriterSpec.cxx#L308-L314
Now reports as 
```
grep 'size report' reco_NOGPU.log
[2818207:ctf-writer]: [17:34:34][IMPORTANT] CTF#0 size report: ITS:486,036 TPC:197,311,602 TRD:14,630,269 TOF:211,483 PHS:62,115 CPV:469,135 EMC:85,662 HMP:646 MFT:134,976 MCH:4,378,093 MID:262,795 ZDC:500,006 FT0:21,519 FV0:24,864 FDD:25,966 CTP:3,978
```

rbx:
FairLogger just passes output to cout (with the `LOG` macro), fmt's printf (`LOGF` macro) or fmt's format (`LOGP` macro).
https://github.com/FairRootGroup/FairLogger#2-additional-macros

For cout you can try:
https://stackoverflow.com/questions/43482488/how-to-format-a-number-with-thousands-separator-in-c-c/43482688#43482688
for fmt:
https://stackoverflow.com/a/68226736

rbx:
Also for fmt, you can try `group_digits` from here:
https://fmt.dev/latest/api.html#utilities
Supposed to be locale independent

shahor02:
@rbx thanks, switched to `fmt::group_digits`. But I find it error-prone that setting the locale breaks shmem access.

rbx:
@shahor02 I agree.

PR #9019, opened 2022-05-30
Populator may to adjust previous EOV to SOV of new CCDB object

shahor02:
If the CcdbObjectInfo of the object sent to the populator was created with
adjustableEOV flag (default), then, after uploading the new object populator will
override the EOV of the previous overlapping object (if any) to the SOV of
the new object. This is done only provided the object to override has metadata
key \"adjustableEOV\" defined (value is irrelevant) and has no \"default\" key.
The newly uploaded object will have this \"adjustableEOV\" assigned in its metadata.

@costing did in a somewhat different way, could you check. Tested on the http://ccdb-test.cern.ch:8080/browse/TST/Override?report=true, I guess on the http://alice-ccdb.cern.ch it should not work since credentials will be requested.
What would be the curl API call equivalent to `curl -X PUT http://ccdb-test.cern.ch:8080/TST/Override/1499/1500`

### DIFF
--- a/CCDB/include/CCDB/CCDBTimeStampUtils.h
+++ b/CCDB/include/CCDB/CCDBTimeStampUtils.h
@@ -16,12 +16,14 @@
 #ifndef O2_CCDBTIMESTAMPUTILS_H
 #define O2_CCDBTIMESTAMPUTILS_H
 
-/// a couple of static helper functions to create timestamp values for CCDB queries
+/// a couple of static helper functions to create timestamp values for CCDB queries or override obsolete objects
 
 namespace o2
 {
 namespace ccdb
 {
+class CcdbApi;
+class CcdbObjectInfo;
 
 /// returns the timestamp in long corresponding to "now + secondsInFuture"
 long getFutureTimestamp(int secondsInFuture);
@@ -31,6 +33,9 @@ long getCurrentTimestamp();
 
 /// \brief Converting time into numerical time stamp representation
 long createTimestamp(int year, int month, int day, int hour, int minutes, int seconds);
+
+/// \brief set EOV of overriden objects to SOV-1 of overriding one if it is allowed
+int adjustOverriddenEOV(CcdbApi& api, const CcdbObjectInfo& infoNew);
 
 } // namespace ccdb
 } // namespace o2
--- a/CCDB/include/CCDB/CcdbObjectInfo.h
+++ b/CCDB/include/CCDB/CcdbObjectInfo.h
@@ -33,12 +33,19 @@ class CcdbObjectInfo
   static constexpr long YEAR = 364 * DAY;
   static constexpr long INFINITE_TIMESTAMP = 9999999999999;
   static constexpr long INFINITE_TIMESTAMP_SECONDS = 2000000000; // not really inifinity, but close to std::numeric_limits<int>::max() till 18.05.2033
+  static constexpr const char* AdjustableEOV = "adjustableEOV";
+  static constexpr const char* DefaultObj = "default";
 
   CcdbObjectInfo() = default;
   CcdbObjectInfo(std::string path, std::string objType, std::string flName,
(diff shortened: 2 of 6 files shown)


costing:
@shahor02 the [updateMetadata](https://github.com/AliceO2Group/AliceO2/blob/dev/CCDB/src/CcdbApi.cxx#L1283) method could be used, but it needs an extra timestamp argument, the new EOV. To keep the compatibility with the current implementation it could be defined with a default value of `0`, in which case it would not be passed on to the `curl` call. Something like:

```
-     void CcdbApi::updateMetadata(std::string const& path, std::map<std::string, std::string> const& metadata, long timestamp, std::string const& id)
+     void CcdbApi::updateMetadata(std::string const& path, std::map<std::string, std::string> const& metadata, long timestamp, std::string const& id, long newEOV = 0)
...
      fullUrl << getHostUrl(hostIndex) << "/" << path << "/" << timestamp;
+     if (newEOV>0)
+        fullUrl << "/" << newEOV;
```
plus it would benefit from automatically following redirects with
```
+    curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
```

The extra read call of `prevHeader = api.retrieveHeaders(..)` is to log the action? I am thinking of simply calling `updateMetadata` and log the returned code (in particular `304` = `modified`, `204` = `no changes`, `400` = `rejected due to invalid arguments`, `404` = `no object matching`) and headers (the modified object's info is returned as headers also for `PUT`). That would require further modifications in `updateMetadata`: an extra argument for the metadata constraints (in addition to the new values to set) and a different return value. On the up side it would avoid an extra read call for every new object.

CCDB/include/CCDB/CCDBTimeStampUtils.h
```diff
@@ -32,6 +34,9 @@ long getCurrentTimestamp();
 /// \brief Converting time into numerical time stamp representation
 long createTimestamp(int year, int month, int day, int hour, int minutes, int seconds);
 
+/// \brieve set EOV of overriden objects to SOV-1 of overriding one if it is allowed
```
chiarazampolli: brieve --> brief

CCDB/include/CCDB/CcdbObjectInfo.h
```diff
@@ -66,8 +92,8 @@ class CcdbObjectInfo
   std::map<std::string, std::string> mMD; // metadata
   long mStart = 0;                        // start of the validity of the object
   long mEnd = 0;                          // end of the validity of the object
-
-  ClassDefNV(CcdbObjectInfo, 1);
+  bool mAdjustableEOV = false;            // each now object may override EOV of object it overrides to its own SOV
```
chiarazampolli: now --> new

CCDB/include/CCDB/CcdbObjectInfo.h
```diff
@@ -51,7 +58,26 @@ class CcdbObjectInfo
   void setPath(const std::string& path) { mPath = path; }
 
   [[nodiscard]] const std::map<std::string, std::string>& getMetaData() const { return mMD; }
-  void setMetaData(const std::map<std::string, std::string>& md) { mMD = md; }
+  void setMetaData(const std::map<std::string, std::string>& md)
+  {
+    mMD = md;
+    if (AdjustableEOV) {
```
chiarazampolli: should this be 
```
if (mMD.find(AdjustableEOV) != mMD.end()) {
```

shahor02: actually, should be `if (mAdjustableEOV) {`, thanks for spotting

shahor02:
@costing switched to updateMetaData.

shahor02:
@costing if there are no more comments I want to merge this

costing (approved):

costing:
Looks good, thank you

PR #9022, opened 2022-05-30
TRD fix handling bad cru halfchamber headers, and errorneous data at end of buffers

bazinski:
- check for vaguely valid cruhalfchamber headers.
- skip data at the end if rdh and cruheaders dont match up on length, only seen crh header < rdh but both are handled.
- hack around going back onto an rdh stop.
- cruhalf chamber header is checked that :
-- lengths are < link maximal data
-- errors are  < known errors of 0x0 0x1 0x2 (checks <0x4)
-- checks sum of all lengths < payload, or remaining payload.
-- checks that endpoint is < 0x2, not sure why its 4 bits.
-- next reader also checks that the orbit is increasing, but cant trivially do that here.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
@@ -85,6 +85,7 @@ constexpr int MAXLINKERRORHISTOGRAMS = 10;     // size of the array holding the 
 constexpr int MAXPARSEERRORHISTOGRAMS = 60;    // size of the array holding the parsing error plots from the raw reader
 constexpr int ETYPEPHYSICSTRIGGER = 0x2;       // CRU Half Chamber header eventtype definition
 constexpr int ETYPECALIBRATIONTRIGGER = 0x3;   // CRU Half Chamber header eventtype definition
+constexpr int MAXCRUERRORVALUE = 0x2;          // Max possible value for a CRU Halfchamber link error. As of may 2022, can only be 0x0, 0x1, and 0x2, at least that is all so far(may2022).
 
 } //namespace constants
 } // namespace trd
--- a/DataFormats/Detectors/TRD/src/RawData.cxx
+++ b/DataFormats/Detectors/TRD/src/RawData.cxx
@@ -333,6 +333,36 @@ std::ostream& operator<<(std::ostream& stream, const HalfCRUHeader& halfcru)
   return stream;
 }
 
+bool halfCRUHeaderSanityCheck(o2::trd::HalfCRUHeader& header, std::array<uint32_t, 15>& lengths, std::array<uint32_t, 15>& eflags)
+{
+  // check the sizes for less than max value
+  // check the errors for either < 0x3, for now (may 2022) there is only no error, 1, or 2.
+  //
+  bool goodheader = true;
+  for (int lengthindex = 0; lengthindex < 15; ++lengthindex) {
+    if (lengths[lengthindex] > o2::trd::constants::MAXDATAPERLINK256) {
+      // something has gone insane.
+      //LOG(info) << "AAA dumping half cru as : half cru link length > max possible! : " << lengths[lengthindex] << " ?? " << o2::trd::constants::MAXDATAPERLINK256;
+      goodheader = false;
+    }
+  }
+  for (int eflagindex = 0; eflagindex < 15; ++eflagindex) {
+    if (eflags[eflagindex] > o2::trd::constants::MAXCRUERRORVALUE) {
+      // something has gone insane.
+      // LOG(info) << "AAA dumping half cru as : half cru link eflag > max possible! : " << std::hex << eflags[eflagindex] << " ?? " << o2::trd::constants::MAXCRUERRORVALUE;
+      goodheader = false;
+    }
+    if (header.EndPoint > 1) {
+      // end point can only be zero or 1, for ach of the 2 pci end points in the cru
+      goodheader = false;
+    }
+    //LOG(info) << "Header sanity check is : " << goodheader;
+    goodheader = false;
+  }
+
+  return goodheader;
+}
+
 bool trackletMCMHeaderSanityCheck(o2::trd::TrackletMCMHeader& header)
 {
   // a bit limited to what we can check.
--- a/Detectors/TRD/reconstruction/src/CruRawReader.cxx
+++ b/Detectors/TRD/reconstruction/src/CruRawReader.cxx
@@ -117,7 +117,7 @@ void CruRawReader::OutputHalfCruRawData()
 void CruRawReader::dumpRDHAndNextHeader(const o2::header::RDHAny* rdh)
 {
   LOG(info) << "######################### Dumping RDH incoming buffer ##########################";
-  o2::raw::RDHUtils::printRDH(rdh);
+  //  o2::raw::RDHUtils::printRDH(rdh);
   LOG(info) << "Now for the buffer breakdown";
   auto offsetToNext = o2::raw::RDHUtils::getOffsetToNext(rdh);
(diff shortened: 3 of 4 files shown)


DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
```diff
@@ -85,6 +85,7 @@ constexpr int MAXLINKERRORHISTOGRAMS = 10;     // size of the array holding the
 constexpr int MAXPARSEERRORHISTOGRAMS = 60;    // size of the array holding the parsing error plots from the raw reader
 constexpr int ETYPEPHYSICSTRIGGER = 0x2;       // CRU Half Chamber header eventtype definition
 constexpr int ETYPECALIBRATIONTRIGGER = 0x3;   // CRU Half Chamber header eventtype definition
+constexpr int MAXCRUERRORVALUE = 0x4;          // Max possible value for a CRU Halfchamber link error. As of may 2022, can only go up to 0x2
```
tdietel: The numeric value and comment don't agree.

bazinski:
@tdietel @martenole given the urgency if you could check this as well.
I ran it against the run @davidrohr quoted in an email on sat, ran it against a standalone run Nastia took on sat, and currently going through a global run from friday and sat. So far so good.

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -333,6 +333,35 @@ std::ostream& operator<<(std::ostream& stream, const HalfCRUHeader& halfcru)
...
+  bool goodheader = true;
+  for (int lengthindex = 0; lengthindex < 15; ++lengthindex) {
+    if (lengths[lengthindex] > o2::trd::constants::MAXDATAPERLINK256) {
+      // something has gone insane.
+      //LOG(info) << "AAA dumping half cru as : half cru link length > max possible! : " << lengths[lengthindex] << " ?? " << o2::trd::constants::MAXDATAPERLINK256;
+    }
+    goodheader = false;
+  }
+  for (int eflagindex = 0; eflagindex < 15; ++eflagindex) {
+    if (eflags[eflagindex] > o2::trd::constants::MAXCRUERRORVALUE) {
```
tdietel, line 350: It looks like at the moment, this rather pathetic error does not produce any output - is this correct?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -140,6 +140,9 @@ bool CruRawReader::processHBFs(int datasizealreadyread, bool verbose)
   auto rdh = mDataRDH;
   auto preceedingrdh = rdh;
   uint32_t totaldataread = 0;
+  if (mHeaderVerbose) {
+    LOG(info) << " mem : " << o2::raw::RDHUtils::getMemorySize(rdh) << " headersize : " << o2::raw::RDHUtils::getHeaderSize(rdh);
```
tdietel, line 144:
```suggestion
    LOG(info) << " mem size: " << o2::raw::RDHUtils::getMemorySize(rdh) << " headersize : " << o2::raw::RDHUtils::getHeaderSize(rdh);
```

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -233,17 +234,22 @@ bool CruRawReader::processHBFs(int datasizealreadyread, bool verbose)
...
+          //          return true;
       }
     }
     //take care of the case where there is an "empty" rdh containing all 0xeeeeeeee as payload.
-    if (mTotalHBFPayLoad / 4 - mHBFoffset32 == 8 && mHBFPayload[mHBFoffset32 + 7] == o2::trd::constants::CRUPADDING32) {
-      mHBFoffset32 += 8;
+    //    if (mTotalHBFPayLoad / 4 - mHBFoffset32 == 8 && mHBFPayload[mHBFoffset32 + 7] == o2::trd::constants::CRUPADDING32) {
+    //      mHBFoffset32 += 8;
+    //    }
+    if (halfcruprocess == -2) {
```
tdietel, line 250: Why is this not in the switch/case statement above? It seems like the case 2, but with a minus sign.

tdietel (commented):
Looks like mostly debugging output to investigate the problem. Is this correct?

DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
bazinski: its an upperbound of 4, I have only seen 0,1,2.

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: if the eflag is greater than the known ones we have its a bad cruheader. I dont follow your comment?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: actually I can take that one out, that was my debugging to track down the empty rdhs.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: The stuff above that is actually only for verbose style outputs.

bazinski: -2 is basically the case of dump and run. I would put in an enum and a pretty name, but this is hopefully that last time I look at this code ;-)

DataFormats/Detectors/TRD/src/RawData.cxx
tdietel: You are checking if the `eflags[eflagindex]` are outside the usual region, and if they are, you don't do anything (the block in lines 351-353 is all commented out). So it is treated exactly like a 'valid' error (i.e. LME type 1 or 2), and there is not red flag about this.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
tdietel: Ok.

bazinski:
@tdietel to answer your query about it all being debug output, most of it is error checking on the cruheader for sane values. No full proof way to say, yay or nay for a cruhalfchamber header, so lengths are checked to be below a full even, errors are checked for less than known maximal values, end point is checked, sum of all lenghts is checked to fit in the rdh payload. A fix to check for empty rdhs which was causing a inf loop. Associated error messages and warnings, and possibly some tidy up.

davidrohr:
Tried with some raw data files and some MC files. Looks good. Will merge once CI passes.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -446,6 +452,22 @@ int CruRawReader::processHalfCRU(int cruhbfstartoffset)
   mTotalHalfCRUDataLength = mTotalHalfCRUDataLength256 * 32; //convert to bytes.
   int mTotalHalfCRUDataLength32 = mTotalHalfCRUDataLength256 * 8; //convert to bytes.
 
+  //can this half cru length fit into the available space of the rdh accumulated payload
+  if (mTotalHalfCRUDataLength32 > mTotalHBFPayLoad - mHBFoffset32) {
+    if (mMaxErrsPrinted > 0) {
+      LOG(error) << "Next HalfCRU header says it contains more data than in the rdh payloads! " << mTotalHalfCRUDataLength32 << " < " << mTotalHBFPayLoad << "-" << mHBFoffset32;
```
martenole, line 458: Hi @bazinski 
I tried to run over some raw data from Sunday with the current O2 and this error turns up very frequently until the limit is reached. Is this expected to be so often or only when there was a corruption in the data?

PR #9028, opened 2022-05-30
fix track length for TOF-TPC tracks

noferini:
This is to compute properly track length for TPC only track in TOF matching. Thanks to @shahor02 for the instructions.
The test in MC looks fine (min track length found consistent with TOF radius)
@njacazio : to be studied more extensively


noferini:
All checks have passed

PR #9030, opened 2022-05-30
DPL: fix oldest possible timeslice with forwarding

ktf:


ktf:
@davidrohr here is what we were discussing.

ktf:
Updated. Rate limiting with QC seems to work and tested on FLP staging and hyperloop as well.

davidrohr (approved):
Works for me.
FST with QC and CALIB still gets stuck with TIMEFRAME_RATE_LIMIT=5 as communicated over mattermost, but I assume it is a different problem, so we can go ahead and merge this already.

PR #9032, opened 2022-05-31
Fix HG overflow; 

peressounko:
- Overflow in HG channels stored in Cells fixed; 
 - Trigger decoding in raw data implemented; 
 - Mapping and geometry of trig. tiles fixed; 
 - LM-L0 trigger offset implemented in raw data decoding and MC writing


peressounko:
Error in build/O2/o2-cs8 looks unrelated to PR

PR #9036, opened 2022-05-31
ZDC waveform calibration

cortesep:


DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEventFlat.h
```diff
@@ -99,6 +100,32 @@ struct RecEventFlat { // NOLINT: false positive in clang-tidy !!
...
+
+  void allocate(int isig)
+  {
+    if (inter[isig].size() != NIS) {
+      inter[isig].resize(NIS);
+      for (int iis = 0; iis < NIS; iis++) {
+        inter[isig][iis] = 0;
+      }
+    }
+  }
```
shahor02: If the vector size should be always NIS, why don't you use arrays instead? 
Note that vector<float>.resize() will automatically fill by 0 new elements, you don't need to set them to 0.

Detectors/ZDC/base/include/ZDCBase/Constants.h
```diff
@@ -353,6 +366,12 @@ constexpr int toDet(int channel, int& tower)
   }
 }
 
+// Calibration workflows
+static constexpr int WaveformCalib_NBB = 3;
+static constexpr int WaveformCalib_NBA = 6;
+static constexpr int WaveformCalib_NBT = WaveformCalib_NBB + WaveformCalib_NBA + 1;
+static constexpr int WaveformCalib_NW = WaveformCalib_NBT * NIS;
```
shahor02: these are not class members, no need for static.

Detectors/ZDC/calib/include/ZDCCalib/CalibParamZDC.h
```diff
@@ -0,0 +1,35 @@
...
+/// \brief ZDC calibration common parameters
+/// \author P. Cortese
+
+namespace o2
+{
+namespace zdc
+{
+struct CalibParamZDC : public o2::conf::ConfigurableParamHelper<CalibParamZDC> {
+  int debug_output = -1; // Debug output
+  void print();
```
shahor02, line 29: Note that there is a generic printKeyValues which also provides the provenance of values (default, CCDB, command line)

Detectors/ZDC/calib/include/ZDCCalib/WaveformCalibConfig.h
```diff
@@ -0,0 +1,81 @@
...
+  static constexpr int NBT = WaveformCalib_NBT;
+  static constexpr int NW = WaveformCalib_NW;
+
+  WaveformCalibConfig();
+
+  double cutLow[NChannels];         /// Amplitude cut low
+  double cutHigh[NChannels];        /// Amplitude cut high
+  double min_e[NChannels] = {0.};   /// Minimum entries to compute waveform
+  double cutTimeLow[NTDCChannels];  /// TDC cut low
+  double cutTimeHigh[NTDCChannels]; /// TDC cut high
```
shahor02: please initalize all with `{}`

Detectors/ZDC/calib/include/ZDCCalib/WaveformCalibQueue.h
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  int mFirst = 0;           // First bunch of waveform w.r.t. bunch of signal peak
+  int mLast = 0;            // Last bunch of waveform w.r.t. bunch of signal peak
+  int mPk = 0;              // Bunch position of peak w.r.t. first bunch
+  int mN = 1;               // Number of bunches in waveform
+  int mPeak = 0;            // Peak position (interpolated samples) w.r.t. first point
+  int mNP = 0;              // Number of interpolated points in waveform
+  int mTimeLow[NChannels];  /// Cut on position difference low
+  int mTimeHigh[NChannels]; /// Cut on position difference high
```
shahor02, line 49: init arrays with `{}`

Detectors/ZDC/calib/src/InterCalibSpec.cxx
```diff
@@ -114,29 +114,29 @@ void InterCalibSpec::updateTimeDependentParams(ProcessingContext& pc)
...
-  mInterCalib.setTowerParam(towerParam.get());
-  mInterCalib.setInterCalibConfig(interConfig.get());
+  mWorker.setEnergyParam(energyParam.get());
+  mWorker.setTowerParam(towerParam.get());
+  mWorker.setInterCalibConfig(interConfig.get());
 }
 
 void InterCalibSpec::run(ProcessingContext& pc)
 {
   updateTimeDependentParams(pc);
```
shahor02: You are calling updateTimeDependentParams for every TF, and for each TF you are setting extracted values to `worker`.
Normally these values will not change during the run. Note  that if a new object is picked, a `finaliseCCDB` callback will be called if you define it, in the way you can update the mWorker only on change, see https://docs.google.com/presentation/d/1WPNixo57zAn_gM6vMCIBOHs-DZqcKAkLX0CwUgO_VYg/edit#slide=id.g112eef84963_0_3 and grep O2 for finaliseCCDB to see details.
If you are sure the objects never change, you can put in the updateTimeDependentParams
```
static bool updateOnceDone = false;
if (!updateOnceDone) {
   // do what you do now at every TF
   updateOnceDone = true;
}
```

Note: you don't need to checks like https://github.com/AliceO2Group/AliceO2/blob/42aba96b47898f45409b0cb5999b942697f0abde/Detectors/ZDC/calib/src/InterCalibSpec.cxx#L91-L94: if the object is not found, `fatal` will be called by the fetcher.

Detectors/ZDC/calib/src/WaveformCalibData.cxx
```diff
@@ -0,0 +1,202 @@
...
+    LOG(error) << "Cannot create file: " << fn;
+    return 1;
+  }
+  for (int32_t is = 0; is < NChannels; is++) {
+    if (mWave[is].mEntries > 0) {
+      TString n = TString::Format("h%d", is);
+      TString t = TString::Format("Waveform %d %s", is, ChannelNames[is].data());
+      int nbx = mWave[is].mLastValid - mWave[is].mFirstValid + 1;
+      int min = mWave[is].mFirstValid - mPeak - 0.5;
+      int max = mWave[is].mLastValid - mPeak + 0.5;
```
shahor02: int or float?

Detectors/ZDC/calib/src/WaveformCalibEPNSpec.cxx
```diff
@@ -0,0 +1,144 @@
...
+{
+  mTimer.Stop();
+  mTimer.Reset();
+}
+
+void WaveformCalibEPNSpec::init(o2::framework::InitContext& ic)
+{
+  mVerbosity = ic.options().get<int>("verbosity-level");
+  mWorker.setVerbosity(mVerbosity);
+  mTimer.CpuTime();
```
shahor02: why is this here? It just returns CPU time.

Detectors/ZDC/calib/src/WaveformCalibEPNSpec.cxx
```diff
@@ -0,0 +1,144 @@
...
+  mTimer.Start(false);
```
shahor02: If you want to measure the time when your code was actually running, you should stop the timer here, not start.
And then, you should `Start(false)` it at the entrance to run() and `Stop()` on exit from it.

Detectors/ZDC/calib/src/WaveformCalibSpec.cxx
```diff
@@ -0,0 +1,156 @@
...
+      LOG(fatal) << "Missing calibration object: " << ct;
+      return;
+    } else {
+      loadedConfFiles += " ";
+      loadedConfFiles += ct;
+      mWorker.setConfig(config.get());
+    }
+    LOG(info) << loadedConfFiles;
+    mTimer.CpuTime();
+    mTimer.Start(false);
```
shahor02, line 87: same comment on the timer.

Detectors/ZDC/calib/src/zdc-waveformcalib-epn-workflow.cxx
```diff
@@ -0,0 +1,47 @@
...
+#include "Framework/runDataProcessing.h"
+
+WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
+{
+  // Update the (declared) parameters if changed from the command line
+  o2::conf::ConfigurableParam::updateFromString(configcontext.options().get<std::string>("configKeyValues"));
+  WorkflowSpec specs;
+  specs.emplace_back(o2::zdc::getWaveformCalibEPNSpec());
+  // configure dpl timer to inject correct firstTFOrbit: start from the 1st orbit of TF containing 1st sampled orbit
+  o2::raw::HBFUtilsInitializer hbfIni(configcontext, specs);
```
shahor02: this is needed only if your workflow reads data from root files, which it does not.

Detectors/ZDC/calib/src/zdc-waveformcalib-workflow.cxx
```diff
@@ -0,0 +1,47 @@
...
+  o2::raw::HBFUtilsInitializer::addNewTimeSliceCallback(policies);
+}
+
+// we need to add workflow options before including Framework/runDataProcessing
+void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
+{
+  // option allowing to set parameters
+  std::string keyvaluehelp("Semicolon separated key=value strings ...");
+  workflowOptions.push_back(ConfigParamSpec{"configKeyValues", VariantType::String, "", {keyvaluehelp}});
+  o2::raw::HBFUtilsInitializer::addConfigOption(workflowOptions);
```
shahor02, line 31: idem

shahor02 (changes requested):
Hi @cortesep 

Obviously, I could do only a very superficial review. Do you have a way to validate such a complex code?

Detectors/ZDC/calib/include/ZDCCalib/WaveformCalibConfig.h
cortesep: I introduced the initialization, however they are initialized also in the default constructor. Is it ok to do that?

https://github.com/cortesep/AliceO2/blob/42aba96b47898f45409b0cb5999b942697f0abde/Detectors/ZDC/calib/src/WaveformCalibConfig.cxx#L17

shahor02: yes

Detectors/ZDC/calib/src/WaveformCalibData.cxx
cortesep: float! Thanks!

DataFormats/Detectors/ZDC/include/DataFormatsZDC/RecEventFlat.h
cortesep: Thanks for the suggestion about initialization.
I use vectors because in normal runs the inter[isig] data is not filled and therefore I prefer not to allocate space for them. They are quite large arrays with 2400 elements each.

shahor02: ok, then just remove 0 assignments.

Detectors/ZDC/calib/include/ZDCCalib/CalibParamZDC.h
cortesep: I noted the suggestion. This debug_output enables writing debug histograms.

cortesep:
Hi,

I implemented the corrections. When running on a small data sample everything works as expected, when running on 10000 Pb-Pb events the program crashes because reconstruction cannot deal with all this waveform data. 

Is there a way to split the reconstruction over more processes? Is it possible to reduce the size of the time frame?

Thanks.

[38643:zdc-digi-reco]: [17:25:17][INFO] Reconstructed 66729 signal TDCs and 159219 ZDC energies and 429 info messages in 20087/20087 b.c. and 372173 waveform chunks
[38643:zdc-digi-reco]: [17:25:17][ERROR] Exception caught: Requested message size (3574349564) exceeds segment size (2000000000)

shahor02:
@cortesep what do you mean by more processes? The single TF of 128 orbits contains ~550 pbpb events (hadronic ones, you will see more from the QED). But, obviously, you will need to integrate over many TFs, I don't know where do you produce so much data?

cortesep:
Ok. Thanks. I didn't know it was lowered to 128. In any case we need to find a way to simulate correctly the ZDC signal. If we scale for the cross sections the number of ZDC events including QED would be:
550/7.67*420.5=30153

If I simulate separately hadronic, QED single side ZNA, QED single side ZNC, QED double side, is there a way to merge them with correct cross sections?

cortesep:
@shahor02 
I was trying with 10000 hadronic events in a single reconstruction and the reconstruction is not able to send the output of the reconstruction if full interpolated waveforms are dispatched to the calibration process or written to file. The size of the message becomes too big. However in case of 550 hadronic + 30000 QED events the message size could be smaller. We should try to produce a more reasonable simulation output.. 

In addition the actual firmware will readout sequences of 10 bunch crossings instead of sequences of 2 bunch crossings therefore producing 5 times more data. If there is a limitation on the message size we have a problem. We might need to strongly reduce the number of BC in a TF for the special ZDC run.

shahor02:
Hi @cortesep 
The message size is limited by the shared memory and the bandwidth, but whatever they are, it is not normal that ZDC calibration sends >3GB of data. Is this per TF?
Can you produce a specific number on how much data per TF ZDC is going to produce and what will be the output in the CTF and calibrator (per TF and total) for realistic collisions?

> If I simulate separately hadronic, QED single side ZNA, QED single side ZNC, QED double side, is there a way to merge them with correct cross sections?

What generators do you need to simulate these different types? The simulation framework allows to combine different generators with different weights, but I've never did this, better to ask @sawenzel

cortesep:
Hi @shahor02,

this concerns the message size between two processes running on the same EPN node and is per time frame. The traffic from the EPN node to the aggregator will be much lower because all the waveforms of the events in the time frame are aligned and stacked onto one.

In my test, I realize now, the message size it is greater than 3 GB
10000 events * 2 BC * 26 channels * 2400 interpolated samples * 4 bytes = 5 GB
The size of the message towards the aggregator would be 1/10000 of this and therefore around 500 k

If my calculation is correct, by putting a reasonable estimate for the EMD event size, in the standard condition of data taking, for the special run (5 minutes every change of HV that correspond roughly to the LHC period) the message size would reach 18 GB. Therefore we should use shorter TF. I would say 8 orbits would be ok.

Concerning the generators I need the electromagnetic dissociation input from ntuples by Igor Pschenichnov. I asked @coppedis  to prepare Kine trees as input for O2 simulation. I will ask @sawenzel how to merge the different simulations

cortesep:
BTW this is a reconstruction special output that does not go to CTF. It is just consumed by the calibration. During normal data taking all of this is turned off.

shahor02:
Hi @cortesep 

Theoretically, 18GB messages should work on the EPN if ZDC runs in a standalone workflow, but how are you going to test such a code. But I don't understand `10000 events * 2 BC * 26 channels * 2400 interpolated samples * 4 bytes` estimate, can you explain what is the input raw data rate from the ZDC and where it grows to this size?

cortesep:
10000 events * 2 BC * 26 channels * 2400 interpolated samples * 4 bytes = 5 GB**

- 10000 events * 2 BC:  was just a check on a more substantial amount of data
- * 26 channels: because in hadronic collisions all 26 channels usually have data
- * 2400 interpolated samples: here is where the size of the reconstructed event grows. Each of the 12 samples is interpolated by 200 samples
- * interpolation output is float: 4 bytes

During normal operation the interpolation is done only around the signal peak and is not saved since the output arrays are not allocated

During data taking the input data rate will be 30000 events per time frame but the average occupancy will be about 6-7 channels

cortesep:
I tested the code in simulation for sequences of two bunches and I get nice interpolated waveforms in output. Currently the simulation is not yet configured to switch to 10 BCs per event for special runs. It is in the TODO list. This is one of the three updates needed to match the ZDC firmware V1.0. I'm working on the most urgent one that will be in the Baseline calibration workflow.

We will likely be able to test the code in technical or laser runs in a few weeks from now, before I'm able to update digitization.

shahor02:
ok, not sure I understand what means to interpolate a sample by 200 samples and what `DigiReco::interpolate` is doing, but even if you could manage to work with ~20GB per TF, how long it takes to process such a TF? I doubt this can be done online. Can you process only small fraction of the TF, e.g. 1 HBF only?

shahor02:
@cortesep shall I merge it as it now?

cortesep:
The problem of the speed of synchronous reconstruction is quite important. We will need to estimate it with a typical timeframe and will need to allocate resources or sample the events.
I think you can merge it as it is.

shahor02:
ok, will merge once CI is passed.

shahor02:
FullCI failing due to https://github.com/AliceO2Group/QualityControl/pull/1269 (@dstocco ) 
Merging.

PR #9038, opened 2022-05-31
[EMCAL-614] Adding parameter that turns off the L1 phase

hahassan7:
A parameter has been added for switching off the L1 phase simulation. Since in data the L1 phase might be subtracted in the hardware.


mfasDa (approved):
Trivial commit

PR #9047, opened 2022-06-01
ZDC FastSim: Generate Hits as final output of fast simulation

SwirtaB:
SpatialPhotonResponse as return type from fast simulation will provide easy conversion to hits.
Since zdc::Detector already is using it, it will be relatively easy to use it for fast simulation. 
I propose function to easly convert fast simulation output to SpatialPhotonResponse - pixel by pixel.


SwirtaB:
I have implemented another function to explore another solution which is to directly create hits from ML model response.
The function is actully copy of createHitsFromImage, modified to work with Ort::Value.

PR #9051, opened 2022-06-01
Make TGeant4 the default

benedikt-voelkel:


benedikt-voelkel:
@sawenzel as discussed in the meeting. Also that is a pretty trivial change

sawenzel (approved):

PR #9053, opened 2022-06-01
[EMCAL-614] Changing the default delay to 600

hahassan7:
Changing the default delay to 600 ns.


mfasDa (approved):

shahor02 (commented):
trivial, merging.

mfasDa:
As shown in EMCAL-614 the time peak is shifted in the simulation, the delay initially assumed was too much. We need to reduce the delay back to 600 ns, the same what we normally observe in data.

PR #9054, opened 2022-06-01
[WIP] Update TRD validation macro for FST

martenole:
Hi Sean,
I think it is better to do a one to one comparison of the digits and tracklets in the FST. Since I am not sure about the ordering I have two double loops now where I try to find a match for each digit and tracklet. 
For the tracklets this works fine now when I ignore the PID values. So HCId, position etc are identical. For the digits however I can find only 15 matches out of almost 1 million digits. I don't really understand this, because when I compare the different spectra for e.g. ADC values they look identical. This is the case for all members of the digits class. Can it be that there is some mixup in during the conversion?
Cheers,
Ole

### DIFF
--- a/Detectors/TRD/simulation/macros/CheckTRDFST.C
+++ b/Detectors/TRD/simulation/macros/CheckTRDFST.C
@@ -24,6 +24,9 @@
 // Then run this script.
 // the convert-raw-to-tf-file.sh must be run on a machine with >200G the rest can be run anywhere.
 //
+// WARNING: This should be used only for rather small data sets (e.g. the 20 PbPb events as suggested above).
+//          Otherwise, due to the quadratic dependency on the number of digits/tracklets this takes very long,
+//          especially in case there are no matches found.
 
 #if !defined(__CLING__) || defined(__ROOTCLING__)
 #include <TFile.h>
@@ -43,9 +46,7 @@
 
 using namespace o2::trd;
 
-constexpr int kMINENTRIES = 100;
-
-void CheckTRDFST(std::string fstbasedir = "./",
+void CheckTRDFST(bool ignoreTrkltPid = true, std::string fstbasedir = "./",
                  std::string digitfile = "trddigits.root", std::string trackletfile = "trdtracklets.root",
                  std::string recodigitfile = "trddigits.root", std::string recotrackletfile = "trdtracklets.root")
 {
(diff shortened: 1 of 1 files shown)


bazinski:
@martenole I still like the idea of pretty pictures ;-) to do the verification, it allows for easier debugging if something is wrong, maybe leave it in as an option? But yes a 1 to 1 comparison is probably better. As for the digits, looking now.

martenole:
But then you would actually need plots for everything, not just the detector number. And even then I am not sure how much the plots would help, if there were differences? In principle I expect the exact same number of tracklets and for every digit which is in the TF file an exact match in the simulation (there might be less digits, because of the calibration triggers). So I could also just break on the first tracklet/digit which does not have a match in simulation, right?

bazinski:
Yes, I suppose what I am really proposing is heading into qc. Maybe I should consider the task at hand, testing the sim to raw to reco chain and 1 to 1 comparison is what is needed. It would however be nice to plot a pile of things in one macro and know kind of definitively if something is off with the digits and tracklets.

tdietel:
Hello,
It sounds like we have quite a problem with the matching here. I agree that we probably need to do a 1:1 comparision - but with 1 million digits that's no fun. Can we run a FST with a single pp collisions? That should reduce the ambiguities quite a bit...
@bazinski, I hear you and look forward to many pretty pictures in your next presentations ;-)
But I think in this case we have a more fundamental problem that needs the 1:1 comparison.

bazinski:
1 or a million, I think I know what is wrong. At the moment just trying to sort out why the tracklets in the data taken over the weekend are test tracklet format. Then will get to this.

martenole:
Sure, the number of collisions in the FST is configurable. The comparison would be faster, if some matches would be found ;)
@bazinski what do you mean that the tracklets were test tracklet format? Are they not usable at all?

bazinski:
@martenole chat on mattermost ;-)

davidrohr:
what is the status of this? Can this be merged?

martenole:
Sorry, not yet. Will remove WIP once ready

bazinski:
Ok, digits problem is fixed in #PR9187.
As for this script, I think its better to sort the digits and do the tracklet and digit comparison per trigger record and then fail on the first missed one as opposed to going through the entire 2d comparison. I have a version you can merge into this.

martenole:
Closing, macro functionality included in https://github.com/AliceO2Group/AliceO2/pull/9373

PR #9057, opened 2022-06-01
FIT: modification for CTF encoding/decoding procedure

afurs:
Procedure for Channel ID encoding/decoding modified. For new decoding one has to use CTF dictionary with 1.1 version. For old data it should be 1.0


Detectors/FIT/FT0/reconstruction/src/CTFCoder.cxx
```diff
@@ -91,7 +100,8 @@ void CTFCoder::compress(CompressedDigits& cd, const gsl::span<const Digit>& digi
     }
     uint8_t prevChan = 0;
     for (uint8_t ic = 0; ic < cd.nChan[idig]; ic++) {
-      cd.idChan[ccount] = chanels[ic].ChId - prevChan;
+      // cd.idChan[ccount] = chanels[ic].ChId - prevChan; //Old method, lets keep it for a while
+      cd.idChan[ccount] = chanels[ic].ChId;
```
shahor02: Here you unconditionally using abs. value instead of the difference, but if the compression goes with old version dictionary, then (1) decoding will be wrong since will be assuming differences, (2, minor point) compression will be inefficient.
Same comment for FDD.

afurs: For old decoding method will be used dictionary 1.0v here https://github.com/AliceO2Group/AliceO2/blob/71b4f176348dada587aadbdf503b5d8c812bd788/Detectors/FIT/FT0/reconstruction/include/FT0Reconstruction/CTFCoder.h#L182

and for new decoding - dictionary version > 1.0 here https://github.com/AliceO2Group/AliceO2/blob/71b4f176348dada587aadbdf503b5d8c812bd788/Detectors/FIT/FT0/reconstruction/include/FT0Reconstruction/CTFCoder.h#L185

afurs:
Unrelated issue in build/O2/fullCI  Should I restart it by using dummy commit?

Detectors/FIT/FT0/reconstruction/src/CTFCoder.cxx
shahor02: I am talking about encoding, not decoding. It is always done with abs value, regardless the version of the imposed dictionary.

afurs: yes, but we need to store chIDs directly, w/o using differences. chIDs will be unsorted within container.

shahor02: @afurs: if you encode the data is v1.0 dictionary imposed and using abs values, then, during the decoding it will detect the old dictionary and decode the data using differences algorithm, i.e. decoding will be wrong.
With the old dictionary you should still difference at encoding, it might be inefficient in terms of compression level, but at least the encoding/decoding will be consistent.

The fullCI failure is unrelated and general, but before merging this PR, please make sure that 
1) you can read old CTFs
2) you can write and read CTFs using old dictionary
3) same for the new dictionary.

afurs: If I understand correctly, all new FIT CTF dicts will be produced with 1.1v https://github.com/AliceO2Group/AliceO2/blob/71b4f176348dada587aadbdf503b5d8c812bd788/Detectors/FIT/FT0/reconstruction/src/CTFCoder.cxx#L41-L49

1. I already checked it
2. Actually we cannot write with old dict, it will be new by default. I was thinking about keeping opportunity to write with old version, but compression https://github.com/AliceO2Group/AliceO2/blob/71b4f176348dada587aadbdf503b5d8c812bd788/Detectors/FIT/FT0/reconstruction/include/FT0Reconstruction/CTFCoder.h#L86
stands before assigning dict version https://github.com/AliceO2Group/AliceO2/blob/71b4f176348dada587aadbdf503b5d8c812bd788/Detectors/FIT/FT0/reconstruction/include/FT0Reconstruction/CTFCoder.h#L96
3. Already tested

afurs: Ok, I will try to solve p.2 As I understand, I can use `mExtHeader`

shahor02: For (2), I don't understand what do you mean `we cannot write with old dict, it will be new by default`? Unless the encoding is done with option `--ctf-dict none` (in this case for every TF a dictionary of default version is created in stored in the TF itself), the dictionary is an object fetched from the CCDB or from the local file. And the encoding must correspond to the version of the imposed dictionary. E.g. if we do the MC for the 2021 pilot beam period, it will use for the encoding http://alice-ccdb.cern.ch/browse/FT0/Calib/CTFDictionary (in fact, both dictionaries there are the same, one of the "unanchored MC" period of 2019 and other is for the productions) having version 1.0.

shahor02: > Ok, I will try to solve p.2 As I understand, I can use `mExtHeader`

yes

shahor02 (approved):

PR #9058, opened 2022-06-01
add MCH-MID tracks to VertexTrackMatcher

pillot:
MCH-MID tracks are basically MCH tracks with MID timing information. The object TrackMCHMID only contains matching information, timing and references to the corresponding MCH-standalone and MID-standalone tracks.

The MCH-MID matched tracks are treated first and the corresponding MCH-standalone tracks are discarded.

Since the timing information is stored in the TrackMCHMID object, there is not need to access the MID tracks here.

### DIFF
--- a/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
+++ b/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
@@ -32,6 +32,8 @@
 #include "DataFormatsMID/MCClusterLabel.h"
 #include "DataFormatsMID/MCLabel.h"
 
+#include "ReconstructionDataFormats/TrackMCHMID.h"
+
 #include "DataFormatsTPC/TrackTPC.h"
 #include "DataFormatsTOF/Cluster.h"
 #include "DataFormatsHMP/Cluster.h"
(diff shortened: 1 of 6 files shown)


pillot:
Hi @shahor02 
I solved the problem spotted by the fullCI, so all checks are green and this PR is now ready for review :)

DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
```diff
@@ -287,6 +290,27 @@ void o2::globaltracking::RecoContainer::createTracksVariadic(T creator) const
     }
   }
 
+  // MCH-MID matches
+  {
+    currentSource = GTrackID::MCHMID;
+    if (matchesMCHMID.size() && !tracksMCH.size()) {
```
shahor02, line 296: why don't you put in the RecoContainer::requestMCHMIDMatches request of MCH tracks, so that the MCH tracks are loaded whenever MCHMID is requested?

pillot: I was not sure about the policy so I tried to separate things as much as possible, but I can do it. To be consistent I can also add the corresponding MCHTrackReaderSpec in the InputHelper when MCHMID is requested. Should I ?

shahor02:
See e.g. how the TOF clusters are requested if there is a request for TOF matches:
https://github.com/AliceO2Group/AliceO2/blob/77424afd79744899f77f15ad4139914659bc50fb/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L473-L474
The parallel request of the TOF clusters reader when reading TOF matches is requested was actually incomplete, just fixed in https://github.com/AliceO2Group/AliceO2/pull/9134

shahor02 (approved):

PR #9061, opened 2022-06-02
[EMCAL-565,EMCAL-566] Remove symbol duplication

mfasDa:
Defining the EMCALChannelScaleFactors in
both LinkDefs of EMCALCalib and EMCALCalibration
leading to symbol duplication in due to dictionries
of the class in both libraries.


sawenzel (approved):

mfasDa:
OK, EMCAL-related fullCI errors are indeed gone.

PR #9068, opened 2022-06-02
Implementing class for nD piecewise polynomials

matthias-kleiner:
nD piecewise polynomials can be used in a similar way to nD splines, except that the polynomials are not guaranteed to be smooth at the edges.
The main advantage compared to splines is that there is not such a large drop in performance when higher dimensions are used.

- Adding test for picewise polynomials

other changes
MultivariatePolynomial:
- adding fit functionality for runtime polynomials
- adding interaction_only feature (see:
https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.PolynomialFeatures.html)
- adding test case for interaction only
- make runtime polynomials CPU only (for GPU use compile time version)


matthias-kleiner:
@davidrohr can this be merged?

PR #9069, opened 2022-06-02
Disable RDH packet jumps check at start of TF in ITS/MFT decoder

shahor02:
As this check makes no sense if TFs are not contiguous, i.e. on EPN

@nicolovalle


shahor02:

PR #9076, opened 2022-06-03
Fix: export of ITS seeding vertices was broken

shahor02:
The vertices were found and used in ITS tracking but the exporting
to the output and logging was broken by the recent change in sync.reco
selection algorithm.


shahor02:
tested locally, merging

PR #9079, opened 2022-06-03
Fix reporting of used ITS/MFT Alpide param used

shahor02:


shahor02:
tested locally, merging

PR #9080, opened 2022-06-03
TRD charge handling in tracklets

bazinski:
- fix charge handling for all charge formats in the tracklets.
- this include the dynamic range one that seems to be present in current data.
- better identification of garbage at end of link data as opposed to simply assuming its default halfcruheader.
- macro toviw charge windows per sector.
- macro to view tracklets (option to view only 1 sector)
- merged some code from new raw reader, as it evident by the rather large changes to rawdata.cxx.

This was tested on 517134, the old krypton data, the noise data 514419, will check some more runs tomorrow.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
@@ -14,6 +14,8 @@
 
 #include "fairlogger/Logger.h"
 
+#include <bitset>
+
 /// \class TRDRDH
 /// \brief Header for TRD raw data header
 //  this is the header added by the CRU
@@ -22,7 +24,7 @@ namespace o2
 {
 namespace trd
 {
-
+class Tracklet64;
 /// \structure HalfCRUHeader
 /// \brief Header for half a cru, each cru has 2 output, 1 for each pciid.
 //         This comes at the top of the data stream for each event and 1/2cru
@@ -125,22 +127,22 @@ struct TrackletHCHeader {
   union {
     //             10987654321098765432109876543210
     // uint32_t:   33222222222211111111110000000000
-    //                 cccccccccccccccX LLL   SSSSS
-    //             ffff|              |y|  sss|
-    //             |   |              |||  |  |-----  0-4  supermodule
-    //             |   |              |||  |--------  5-7  stack
-    //             |   |              ||------------  8-10 layer
-    //             |   |              |------------- 11    always 0x1
-    //             |   |              |------------- 12 side of chamber
+    //                 cccccccccccccccXSSSSS   SSSy
+    //             ffff|              |    |sss  ||
+    //             |   |              |    |  |  ||--  0    side
+    //             |   |              |    |  |  |---  1-3  stack
+    //             |   |              |    |  |------  4-6  layer
+    //             |   |              |    |--------  7-11 sector
+    //             |   |              |------------- 12 always 0
     //             |   ----------------------------- 13-27 MCM Clock counter
     //             --------------------------------- 28-31 tracklet data format number
     uint32_t word;
     struct {
-      uint32_t supermodule : 5;
-      uint32_t stack : 3;
-      uint32_t layer : 3;
-      uint32_t one : 1;   //always 1
       uint32_t side : 1;  // side of chamber
+      uint32_t layer : 3;
+      uint32_t stack : 3;
+      uint32_t supermodule : 5;
+      uint32_t one : 1;   //always 0
       uint32_t MCLK : 15; // MCM clock counter 120MHz ... for simulation -- incrementing, and uniform across an event
       uint32_t format : 4;
       //  0 baseline PID 3 time slices, 7 bit each
@@ -417,13 +419,9 @@ struct DigitMCMData {
   };
 };
 
-void buildTrackletHCHeader(TrackletHCHeader& header, int sector, int stack, int layer, int side, int chipclock, int format);
-void buildTrackletHCHeaderd(TrackletHCHeader& header, int detector, int rob, int chipclock, int format);
-uint16_t buildTRDFeeID(int supermodule, int side, int endpoint);
 uint32_t setHalfCRUHeader(HalfCRUHeader& cruhead, int crurdhversion, int bunchcrossing, int stopbits, int endpoint, int eventtype, int feeid, int cruid);
 uint32_t setHalfCRUHeaderLinkData(HalfCRUHeader& cruhead, int link, int size, int errors);
 void buildTrackletMCMData(TrackletMCMData& trackletword, const uint slope, const uint pos, const uint q0, const uint q1, const uint q2);
-uint32_t unpacklinkinfo(const HalfCRUHeader& cruhead, const uint32_t link, const bool data);
 uint32_t getlinkerrorflag(const HalfCRUHeader& cruhead, const uint32_t link);
 uint32_t getlinkdatasize(const HalfCRUHeader& cruhead, const uint32_t link);
 uint32_t getlinkerrorflags(const HalfCRUHeader& cruheader, std::array<uint32_t, 15>& linkerrorflags);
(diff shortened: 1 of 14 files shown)


Detectors/TRD/reconstruction/macros/plotAllCharges.C
```diff
@@ -0,0 +1,96 @@
...
+  std::vector<TH1F*> hCharges;
+  for (int sector = 0; sector < 18; ++sector) {
+    hCharges.push_back(new TH1F(Form("Charge0_%i", sector), Form("Sector%i hCharge window 0;charge;counts", sector), 64, 0, 63));
+    hCharges.push_back(new TH1F(Form("Charge1_%i", sector), Form("Sector%i Charge window 1;charge;counts", sector), 256, 0, 255));
+    hCharges.push_back(new TH1F(Form("Charge2_%i", sector), Form("Sector%i Charge window 2;charge;counts", sector), 256, 0, 255));
+  }
+  return hCharges;
+}
+
+void plotAllCharges()
```
martenole: Did you upload the correct macros? Basically this does the same as `checkTrackletCharges.C`, right?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -429,11 +429,11 @@ int CruRawReader::processHalfCRU(int cruhbfstartoffset)
   uint32_t cruwordsread = 9;
   //reject halfcru if it starts with padding words.
   //this should only hit that instance where the cru payload is a "blank event" of o2::trd::constants::CRUPADDING32
-  if (mHBFPayload[cruhbfstartoffset] == o2::trd::constants::CRUPADDING32 && mHBFPayload[cruhbfstartoffset + 1] == o2::trd::constants::CRUPADDING32) {
+  if (mHBFPayload[cruhbfstartoffset] == o2::trd::constants::CRUPADDING32) { //} && mHBFPayload[cruhbfstartoffset + 1] == o2::trd::constants::CRUPADDING32) {
     if (mVerbose) {
       LOG(info) << "blank rdh payload data at " << cruhbfstartoffset << ": 0x " << std::hex << mHBFPayload[cruhbfstartoffset] << " and 0x" << mHBFPayload[cruhbfstartoffset + 1];
     }
-    mHBFoffset32 += 2;
+    mHBFoffset32++; //= 2;
```
martenole: Is this correct?

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
```diff
@@ -14,6 +14,10 @@
 
 #include "fairlogger/Logger.h"
 
+#include <bitset>
+
+#include <bitset>
```
martenole: typo, included twice

martenole (commented):
I did not go through it in detail yet, only minor comments. But would it not be possible to add also the rest of the new raw reader?

Detectors/TRD/reconstruction/macros/plotAllCharges.C
bazinski: yip just plots them all one, not really clear though, too many plots. 2d version was messy as well.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: yip only reading the first 0xeeeeeeee now not demanding 2, but about to change.

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
bazinski: oops fixed.

Detectors/TRD/reconstruction/macros/plotAllCharges.C
bazinski: I removed that macro, its pointless, you can see anything.

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -84,253 +103,265 @@ uint32_t getlinkerrorflag(const HalfCRUHeader& cruhead, const uint32_t link)
...
+  header.word = 0;
+
+  header.format = format;
+  header.supermodule = ~sector;
+  header.stack = ~stack;
+  header.layer = ~layer;
+  header.side = ~side;
+
+  header.MCLK = chipclock;
+  header.one = 1;
```
f3sch: According to the TDP, which might be wrong, this fields name is 'header.zero' and is always 0?

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -84,253 +103,265 @@ uint32_t getlinkerrorflag(const HalfCRUHeader& cruhead, const uint32_t link)
...
+      int dynamicq = hcheader.format & 0x1;       // last bit of format (lsb) defines the version of tracklet charge calculation
+      uint32_t pidoffset = (pidword >> 18) & 0x3; // used for dynamic ranged charge windows.
+      //pidword is here to make this code more readible and less error prone.
+      switch (pidindex) {
+        case 2: //Q2
+          if (!dynamicq) {
+            pid = (pidword >> 13) & 0x7f; // 7 bits at the top of all of pid (MSB)
+          } else {
+            pid = (pidword >> 12) & 0x3f; // 6 bits of Q2 and a shift
+            pid += pidoffset;
```
f3sch: I still don't quite get this scale factor... This would add at most 3 to any pid

bazinski: other end, hence the bitshift of pidoffset a few lines above the quote of your comment.

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -84,253 +103,265 @@ uint32_t getlinkerrorflag(const HalfCRUHeader& cruhead, const uint32_t link)
...
+    //length of q1 and q2 are 6 with a 2 bit offset.
+    //What happens when q2 and q1 dont share the same offset (upper 2 bits ?) ?
+    LOG(error) << "This tracklet format has not been tested yet";
+    trackletword.pid = (q0 & 0x3f) & ((q1 & 0x1) << 6);
+    //TODO check the 2 bit offset is still valid ... ??
+  } else {
+    trackletword.pid = (q0 & 0x3f) & ((q1 & 0x1) << 6);
+  }
+  //q2 sits with upper 2 bits of q1 in the header pid word, hence the 0x1f so 5 bits are used here.
+  trackletword.checkbit = 1;
```
f3sch: Shouldn't this also be a trailing zero, or is the TDP here wrong as well?

bazinski: yes, but that function I need to fix in lieu of all these tracklet formats, else it will fail the fst.

DataFormats/Detectors/TRD/src/RawData.cxx
f3sch: Ah missed that, thanks for clarifying.

DataFormats/Detectors/TRD/src/RawData.cxx
martenole: why would it fail the fst? that's just running the simulation + raw conversion + reconstruction the way we define it, so it should be happy either way, right?

bazinski: the format field is strict now, in reconstruction. The fst wont fail now, as the trap2raw builds its own trackeltmcmdata, instead of using the one in rawdata.cxx.

bazinski: basically in fixing the acceptance of the tracklet formats. the format field in trap2raw needs to be properly done. I dont know that I am making this any clearer though :-)

DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
```diff
@@ -30,6 +30,7 @@
 #include "GPUCommonDef.h"
 #include "GPUCommonRtypes.h"
 #include "DataFormatsTRD/Constants.h"
+#include <iostream>
```
martenole: seems this include is not used. I would remove it, since it breaks the GPU compilation. If its needed it would need to be protected from inclusion when compiling the GPU code (see fullCI failure)

bazinski: was there for debugging at some point, will remove

PR #9081, opened 2022-06-03
DPL: have environment variables correctly populated in the ECS YAML

ktf:


ktf:
@vascobarroso @davidrohr this should take care of adding the correct environment variables when you specify them via `--environment`.

davidrohr:
@ktf : Could you clarify what is `env`? Is this just the string provided to `--environment`? I assume this will not work in case 2 environment variables are provided.
@vascobarroso : In the patch that was merged for TOF : We defined a new variable under defaults that was used as value for the env var `FOO = {{ new_var }}`, instead of putting `FOO=1` directly, since Gvozden said `FOO=1` might not work. Could you check what is the correct syntax for putting environment variables there?

ktf:
I just checked and it seems to work provided you space separate the variables on the command line:

```
--environment "FOO=bar BAR=foo"
```

translates to:

```
env:
- FOO=bar
- BAR=foo
```

ktf:
The code which actually creates execution.environ is at DeviceSpecHelpers.cxx:1277 and subsequent lines. I confirm we split by space the string provided by `--environment` and add each part to the vector.

ktf:
This should be all it's needed. Merging since in any case we need to adapt the yaml right now, so it cannot be worse than that.

vascobarroso:
> @vascobarroso : In the patch that was merged for TOF : We defined a new variable under defaults that was used as value for the env var `FOO = {{ new_var }}`, instead of putting `FOO=1` directly, since Gvozden said `FOO=1` might not work. Could you check what is the correct syntax for putting environment variables there?

@davidrohr @ironMann just checked with @teo, the `FOO=1` should also work.

davidrohr:
@vascobarroso : ok, thx for checking. So this PR should be fine.
It is in O2 2022-06-05 and later. So we can test it with the next (not today) FLPSuite update.

PR #9082, opened 2022-06-03
DPL GUI: speedup metric processing

ktf:


ktf:
Okey, apparently I did not split it correctly...

PR #9087, opened 2022-06-04
mch-reco-workflow stores cluster if root output is enabled

shahor02:
Hi @aphecetche 
All detector reco workflows are storing their cluster trees unless `--disable-root-output` is passed. So, adding the global-cluster-writer to the mch-reco-workflow. 
But I did not find who can produce `MCH/CLUSTERLABELS/0` output, therefore at the moment disabled the writing and reading of cluster MClabels. Could you clarify how the cluster labels should be produced?


pillot:
Hi @shahor02 
The MC labels are not produced for the clusters for the moment.

shahor02:
Thanks @pillot , ok, then it is correct to disable their request at the moment. Shall I merge this?

pillot:
Looks ok to me

pillot (approved):

PR #9091, opened 2022-06-04
Sending to CCDB only after all DPs have been seen.

chiarazampolli:
At end of stream, what is there is sent, even if not complete.

@noferini , @shahor02 : this follows a comment by @shahor02 concerning what happens is not all DPs are seen. 
Since the FBI is sent every 5 seconds, and TOF integrates at least over a minute (of course, via command line option one could set shorter intervals), and in real data now every 10 minutes, this situation should never happen, but here is a fix. 
At end of stream, of course, we could be in the situation that not everything is there, but here there is not much we can do. 
Another option, though, could be that we keep the previous value till the new one arrives. If this solution is better, then I will implement it that way (it came to my mind after I implemented this one).

### DIFF
--- a/Detectors/TOF/calibration/include/TOFCalibration/TOFDCSProcessor.h
+++ b/Detectors/TOF/calibration/include/TOFCalibration/TOFDCSProcessor.h
@@ -115,6 +115,16 @@ class TOFDCSProcessor
     mTOFDCS.clear();
   }
 
+  bool areAllDPsFilled()
+  {
+    for (auto& it : mPids) {
+      if (!it.second) {
+        return false;
+      }
+    }
+    return true;
+  }
+
  private:
   std::unordered_map<DPID, TOFDCSinfo> mTOFDCS;                // this is the object that will go to the CCDB
   std::unordered_map<DPID, bool> mPids;                        // contains all PIDs for the processor, the bool
--- a/Detectors/TOF/calibration/testWorkflow/TOFDCSDataProcessorSpec.h
+++ b/Detectors/TOF/calibration/testWorkflow/TOFDCSDataProcessorSpec.h
@@ -105,6 +105,7 @@ class TOFDCSDataProcessor : public o2::framework::Task
   void run(o2::framework::ProcessingContext& pc) final
   {
     auto startValidity = DataRefUtils::getHeader<DataProcessingHeader*>(pc.inputs().getFirstValid(true))->creation;
+    LOG(debug) << "startValidity = " << startValidity;
     auto dps = pc.inputs().get<gsl::span<DPCOM>>("input");
     auto timeNow = HighResClock::now();
     if (startValidity == 0xffffffffffffffff) {                                                                   // it means it is not set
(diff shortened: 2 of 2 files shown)


Detectors/TOF/calibration/include/TOFCalibration/TOFDCSProcessor.h
```diff
@@ -115,8 +115,17 @@ class TOFDCSProcessor
     mTOFDCS.clear();
   }
 
- private:
-  std::unordered_map<DPID, TOFDCSinfo> mTOFDCS;                // this is the object that will go to the CCDB
+  bool areAllDPsFilled()
+  {
+    if (!it.second) {
+      return false;
```
shahor02: what is `it`? Was something cut out?

shahor02 (commented):
Hi @chiarazampolli 

Not sure which comment you refer to, what I remember telling is that processors which write to CCDB at fixed periods would risk to write incomplete data. For this reason I've changer dcs-proxy to not send anything until 1st FBI is seen. 
Now the processors are guaranteed to see the full set, so incomplete data is possible only if the corresponding DP is simple not sent by the server. 

As far as can understand, there is no EoS in dcs-dp workflow: for that the DCS server would need to send some signal telling that there will be no more data.

Detectors/TOF/calibration/include/TOFCalibration/TOFDCSProcessor.h
chiarazampolli: This was my typo when removing some logs, I messed up the parenthesis. Sorry.

chiarazampolli:
> Hi @chiarazampolli
> 
> Not sure which comment you refer to, what I remember telling is that processors which write to CCDB at fixed periods would risk to write incomplete data. For this reason I've changer dcs-proxy to not send anything until 1st FBI is seen. Now the processors are guaranteed to see the full set, so incomplete data is possible only if the corresponding DP is simple not sent by the server.
> 
> As far as can understand, there is no EoS in dcs-dp workflow: for that the DCS server would need to send some signal telling that there will be no more data.

I am not referring to the beginning of the processing, but to the fact that as soon as we store an object, we clear the information on the DPs; so if we update again CCDB but a new FBI has not yet been sent, we might store in the CCDB an incomplete object. This comment was for MFT, which kind of copied from TOF, so the comment applies also here, see https://github.com/AliceO2Group/AliceO2/pull/8992#issue-1251285952. 

Of course DCS DP processing will not receive an end of stream, but when for any reason we will stop processing, we will get one.

shahor02:
> Of course DCS DP processing will not receive an end of stream, but when for any reason we will stop processing, we will get one.

Not sure, who can send it?

chiarazampolli:
Will not ctrl-c trigger it? In case, as i said, we stop the process.

shahor02:
> Will not ctrl-c trigger it? In case, as i said, we stop the process.

I don't think so but not 100% sure, better @ktf tells if stopping the proxy by ^C will initiate EoS.
But it any case, who can send such a signal on flp199 to a process running in background?

ktf:
Yes, it should initiate the EoS on ^C.

Ciao,
Giulio

chiarazampolli:
Thanks @ktf for confirming. 
Yes, this is an automatic process, but there might be the need to stop it somehow (via systemd now, or nomad later), which, I assume, would result in calling EoS.
As I said, this should not happen often, and most likely not during a run, so it will not be a problem. 
The PR here is a kind of protection, to be sure that all CCDB entries will have data for all DPs.

shahor02 (approved):

PR #9097, opened 2022-06-05
DPL: dummy sync should always rescan to avoid deadlocks

ktf:


ktf:
Somehow the dummysink ends up with messages  with time based timeslice, which result in dummy entries being created in the DataRelayer, that after a while provoke backpressure. I am not sure what is causing them, but rescanning seems to be the correct thing to do in any case.

ktf:
@davidrohr @shahor02 FYI.

ktf:
Tested on hyperloop and the FLP staging. The only thing I do not understand is that now staging seems to top at 344MB/s rather than > 500MB/s like before (independently of this PR).

ktf:
@jgrosseo Merging this as it might also fix analysis getting stuck.

PR #9100, opened 2022-06-06
TRD TRAP floating point arithmetics for tracklet charges

martenole:
Add option to enable them from the command line


bazinski (approved):

martenole:
At some point this might come out of the TRAP config. Until then merging to have the option available for simulations

PR #9101, opened 2022-06-06
TRD variable timebins for digits

bazinski:
- Enable somewhat variable timebins for digits.
- It must be below 30, if not it will be set to 30 via o2::trd::constants::TIMEBINS, the storing structure remains the same 30 timebins.
- Any bins less than 30 are stored blank. This therefore only affects the parsing. The objects sent back still have 30 timesbins in the ADC struct.
- This is needed for possibly improving the pulseheight spectra. Some runs were taken with 24 time bins.

### DIFF
--- a/Detectors/TRD/reconstruction/include/TRDReconstruction/CruRawReader.h
+++ b/Detectors/TRD/reconstruction/include/TRDReconstruction/CruRawReader.h
@@ -76,6 +76,7 @@ class CruRawReader
     mEnableTimeInfo = options[TRDEnableTimeInfoBit];
     mEnableStats = options[TRDEnableStatsBit];
     mOptions = options;
+    mTimeBins = constants::TIMEBINS; // set to value from constants incase the DigitHCHeader1 header is not present.
   }
 
   void setMaxErrWarnPrinted(int nerr, int nwar)
@@ -235,6 +236,7 @@ class CruRawReader
   HalfCRUHeader mCurrentHalfCRUHeader; // are we waiting for new header or currently parsing the payload of on
   DigitHCHeader mDigitHCHeader;        // Digit HalfChamber header we are currently on.
   DigitHCHeader1 mDigitHCHeader1;      // this and the next 2 are option are and variable in order, hence
+  uint16_t mTimeBins;
   DigitHCHeader2 mDigitHCHeader2;      // the individual seperation instead of an array.
   DigitHCHeader3 mDigitHCHeader3;
   TrackletHCHeader mTrackletHCHeader;  // Tracklet HalfChamber header we are currently on.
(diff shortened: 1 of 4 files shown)


Detectors/TRD/reconstruction/include/TRDReconstruction/CruRawReader.h
```diff
@@ -76,6 +76,7 @@ class CruRawReader
     mEnableTimeInfo = options[TRDEnableTimeInfoBit];
     mEnableStats = options[TRDEnableStatsBit];
     mOptions = options;
+    mTimeBins = constants::TIMEBINS; // set to value from constants incase the DigitHCHeader1 header is not present.
```
tdietel, line 79: The header should always be present. It should be a (fatal) error if it is missing. But if this is the 'old' raw reader, it might make sense to keep this option, for the new one, we should probably remove it.

bazinski: Well I dont want the thing to crash. Bad data should get flagged and acted on. The flagging is done somewhere else. DigitHCHeader1 is optional though, only the DigitHCHeader is not optional according to the code, as I understand it. Yes I agree its pretty useless with out.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -357,6 +357,7 @@ int CruRawReader::parseDigitHCHeader()
           //LOG(alarm) << "Digit HC Header 1 reserved : " << std::hex << mDigitHCHeader1.res << " raw: 0x" << mDigitHCHeader1.word;
           increment2dHist(TRDParsingDigitHeaderWrong1, mFEEID.supermodule * 2 + mHalfChamberSide[0], mStack[0], mLayer[0]);
         }
+        mTimeBins = mDigitHCHeader1.numtimebins;
```
tdietel, line 360: I might be paranoid, but I'd suggest a check here if the number of timebins from the HCH1 is the same as in the preset. We could use a preset of -1 to indicate that we trust the HC1.

bazinski: mmm good point. So you dont trust the contents of the headers ;-)

tdietel: I trust the contents of the headers, but I do not trust the person that enters them. And I wonder what will happen if 539 chambers are configured with 24 and 1 with 27 timebins...
Did I mention that I might be paranoid?

bazinski: 24 and 27 wont matter so long as its <30. It does not even have to be a multiple of 3, although I am not sure what the trap will do with a number not divisible by 3. Of course if its >30 then those adc values are going to get dumped after the first one with >30 up to the next digit end marker. Paranoid is sometimes bad, for this type of code, running on line paranoid is the wiser stance.

davidrohr:
What is the status of this PR, can it be merged?

bazinski:
@davidrohr  Its busy running through test runs at the moment. When(if) that finishes happily it can be merged.

bazinski:
@tdietel and @martenole all tested on 30bin runs and runs with 24 time bins, interestingly more digits in the code that accomodates 24 timebins than the current dev. Not a lot more, but more 17.794 m vs 17.787 million for 517134

martenole (approved):
Checked with @bazinski that this is fine to be merged

PR #9103, opened 2022-06-07
test for propagation to PV

jgrosseo:


jgrosseo:
(discussion on JIRA): https://alice.its.cern.ch/jira/browse/ATO-592

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -109,6 +110,8 @@ DECLARE_SOA_EXPRESSION_COLUMN(Eta, eta, float, //! Pseudorapidity
                               -1.f * nlog(ntan(PIQuarter - 0.5f * natan(aod::track::tgl))));
 DECLARE_SOA_EXPRESSION_COLUMN(Pt, pt, float, //! Transverse momentum of the track in GeV/c
                               nabs(1.f / aod::track::signed1Pt));
+DECLARE_SOA_DYNAMIC_COLUMN(IsAtPV, isAtPV, //! Is the track within the beam pipe (= propagated to a collision vertex)
+                           [](float x) -> bool { return (std::fabs(x) < o2::constants::geom::XBeamPipeInnerRef); });
```
shahor02: The actual outer radius is 1.6cm. Is this to check that the track enters to the beam pipe or that the propagation has succeeded? The X might be ~0 but the DCA still can be very large.

jgrosseo: It would be the check that the propagation has succeeded. So it would be true also for all HF decays etc.

jgrosseo: We could call it  `isPropagated` or `isPropagatedToPV`. What do you think?

shahor02: But the propagation may succeed and still the X at DCA can be large... If this is just to check that the propagation succeeded, you should check for abs(dca)<998.
To make sure it enters to beam pipe, one should check for x^2+y^2<1.6^2.

jgrosseo: We cannot check the DCA because it is in a different table...
I was intending to check the radius at which the track is

shahor02: Ah, wait, sorry, what I wrote is wrong: if the vertex is within the BP, then the track successfully taken to DCA will have X = radius of the vertex, even if the DCA itself is very large, so, the check is correct.

jgrosseo: OK, cool. Then I adjust it to 1.6cm. Is it worth to distinguish the inner and outer radius of the BP? Otherwise I drop the inner as well.

jgrosseo: And we could call it `isWithinBeamPipe` to make it totally clear what it is :)

PR #9104, opened 2022-06-07
MCH: New Clustering Refactoring

grasseau:


Detectors/MUON/MCH/Clustering/include/MCHClustering/ClusterConfig.h
```diff
@@ -0,0 +1,74 @@
...
+namespace o2
+{
+namespace mch
+{
+
+typedef int PadIdx_t;   // Pad index type
+typedef short Groups_t; // Groups/sub-cluster type
+typedef short Mask_t;   // To build mask
+
+struct ClusterConfig {
```
aphecetche, line 28: For the record (not to be modified for this PR), when we'll get in production with this new clustering (not for this PR either, unfortunately ;-) ), this configuration should probably be integrated with the existing [MCHClusterizerParam](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/MUON/MCH/Clustering/include/MCHClustering/ClusterizerParam.h) in order to be steerable "from the outside"

Detectors/MUON/MCH/Clustering/include/MCHClustering/ClusterFinderGEM.h
```diff
@@ -136,7 +133,11 @@ class ClusterFinderGEM
   void setClusterResolution(Cluster& cluster) const;
   std::unique_ptr<MathiesonOriginal[]> mMathiesons; ///< Mathieson functions for station 1 and the others
   // GG MathiesonOriginal* mMathieson = nullptr;          ///< pointer to the Mathieson function currently used
-
+  //
+  // GG Introduced for run3
+  // function to reinterpret digit ADC as charge
+  std::function<double(uint32_t)> mADCToCharge = [](uint32_t adc) { return static_cast<double>(adc); };
```
aphecetche, line 139: seems to be a rather complicated way of doing a simple cast. Or is it placeholder for a foreseen more complex ADC to charge conversion ?

aphecetche (approved):

Detectors/MUON/MCH/Clustering/include/MCHClustering/ClusterConfig.h
grasseau: The configuration of new algo is different from the old one but this configuration could inherit from the  ClusterizerParam or maybe better, having a subset of ClusterConfig.h steerable from outside (with the same variable name for common values).
Add in my long TODO list ...

Detectors/MUON/MCH/Clustering/include/MCHClustering/ClusterFinderGEM.h
grasseau: I agree.
If I remember, I used the cast function in old implementation.
I don't really focus on it.
Thanks

PR #9105, opened 2022-06-07
Change path for CCDB entry for GRP/LHCIF datapoints

chiarazampolli:
Now the GRP Data Points processing will store:
- GLO/Config/GRPMagField --> magnetic field
- GLO/Config/LHCIFDataPoints --> beam mode, BPTX, luminosity... 
- GLO/Config/EnvVars --> pressure and temperature
- GLO/Config/Collimators --> collimators

In addition, as a reminder, we have: 
- GLO/Config/GRPECS --> for ECS info (star/end run, detectors...)
- GLO/Config/GRPLHCIF --> for LHC file-push (beam type, energy, filling scheme...)

If we want to have "GRP" in front of the 3rd level path of all objects, those filled in the Pilot Beam till now will be invalidated. For LHCIFDataPoints it does not matter, as the data stored there were not meaningful (@shahor02 )


shahor02 (approved):
trivial, merging

PR #9109, opened 2022-06-07
Rdev tofcompressor fixes

preghenella:
This PR fixes a bug in the TOF compressor when running with the `decoderParanoid` option.

It also changes the decoding strategy for the LTM, removing the endless in favour of a checked-loop against the declared event size. It will result in slower but safer execution.

### DIFF
--- a/Detectors/TOF/compression/src/Compressor.cxx
+++ b/Detectors/TOF/compression/src/Compressor.cxx
@@ -538,6 +538,7 @@ bool Compressor<RDH, verbose, paranoid>::processLTM()
   /** process LTM **/
 
   mDecoderSummary.ltmDataHeader = mDecoderPointer;
+  uint32_t eventWords = GET_LTMDATAHEADER_EVENTWORDS(*mDecoderPointer);
   if (verbose && mDecoderVerbose) {
     auto ltmDataHeader = reinterpret_cast<const raw::LTMDataHeader_t*>(mDecoderPointer);
     auto eventWords = ltmDataHeader->eventWords;
@@ -551,7 +552,6 @@ bool Compressor<RDH, verbose, paranoid>::processLTM()
   }
 
   /** loop over LTM payload **/
-  uint32_t eventWords = GET_LTMDATAHEADER_EVENTWORDS(*mDecoderPointer);
   for (int i = 0; i < eventWords; ++i) {
     if (verbose && mDecoderVerbose) {
       printf(" %08x LTM Data \n", *mDecoderPointer);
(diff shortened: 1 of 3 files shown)


Detectors/TOF/compression/src/Compressor.cxx
```diff
@@ -549,26 +547,28 @@ bool Compressor<RDH, verbose, paranoid>::processLTM()
...
+    mDecoderSummary.ltmDataTrailer = mDecoderPointer;
     if (verbose && mDecoderVerbose) {
-      printf(" %08x LTM data \n", *mDecoderPointer);
+      printf(" %08x LTM Global Trailer \n", *mDecoderPointer);
     }
     decoderNext();
     if (paranoid && decoderParanoid()) {
       return true;
     }
+    break;
```
noferini: This break is no longer in the while.
It breaks now compilation 
Please remove 571 line, I did and everything looks in a good shape. 3 readout window recovered in my local tests with paranoid option

noferini (dismissed):
Fix just formatting errors

noferini (dismissed):
It is fine for me.
Since there are many fixes for tof compressor it is important to have changes available tomorrow before the discussion of FLP software updates for next week

noferini (dismissed):

noferini:
o2-cs8 has an error I am not able to understand.


The following tests FAILED:
	 91 - test_Framework_test_CallbackService (Failed)
Errors while running CTest

I don't think it is related to this PR but not 100% sure

noferini (approved):

noferini:
Two tests failing but not related to this PR, can we merge it?

davidrohr:
@noferini @preghenella : The FST is failing for unrelated alien token issue.
FST is stable locally.
If you want we can merge it, to get a new O2 build for the FLPs with the fix.

However, in my FST on simulated raw data I see a lot of errors:
```
processTRMchain: nsteps in while loop = 21, infity loop?
```
Is this expected? Will this appear for all data and produce a log flood?
I am fine with merging it for now, but either we have to remove the error message again afterwards, or the MC2raw conversion must be fixed such that this error does not appear.

noferini:
@davidrohr , if possible I would keep this logging for the moment which should help for debugging other possible issue with compressor. Indeed I tested threshold in real data not Pb-Pb simulation and what you see it is a case I didn't check.
I will remove the logs in any case in future.
Can you privately send me one log you get so that I can see the maximum number of steps reached?

davidrohr:
ok, merging as is to create a tag with it.
Please follow it up soon to remove the error messages.

PR #9111, opened 2022-06-07
optionally refresh CCDB server cache after uploading GRPECS

shahor02:
@costing could you please check?

@knopers8 adding `--refresh` to `o2-ecs-grp-create` will do the job, will you add it once the version with this PR is deployed?
(are you using FLP suite?)


knopers8:
@shahor02 Yes, I will add it to Control once this is merged, then both changes should be there in one or another FLP suite.
As I understand, I should add it for both SOR and EOR, right?

shahor02:
@knopers8 thanks, should be enough to add it at SOR only

PR #9113, opened 2022-06-08
DPL: improve websocket encoding

ktf:
Apparently the "leak" which is observed comes from some weird allocation pattern from the metrics. If I pool allocations like done in this PR it seems to behave correctly also at low rates. Not sure what is going on, but this fixes the issue in my case.


ktf:
Apparently the "leak" which is observed comes from some weird allocation pattern from the metrics. If I pool allocations like done in this PR it seems to behave correctly also at low rates. Not sure what is going on, but it fixes the issue in my case.

ktf:
@shahor02 @davidrohr could you cross check?

shahor02:
indeed, this eliminates the mem.leak.

ktf:
Ok, great. Let me fix the test and then I will merge it.

ktf:
While adapting the test I found a couple of issues with large frames. Most likely not actually used anywhere (but maybe in the remote GUI, which might explain some of the crashes we see in that PR).

ktf:
Tested also with the FST and hyperloop. Merging.

PR #9117, opened 2022-06-08
dev: fixes in config/scalers

lietava:


lietava:
@shahor02 Please, pull.

DataFormats/Detectors/CTP/src/Configuration.cxx
```diff
@@ -567,7 +567,8 @@ void CTPRunManager::init()
 }
 int CTPRunManager::startRun(const std::string& cfg)
 {
-  LOG(info) << "Starting run: " << cfg;
+  LOG(info) << "Starting run: \n"
+            << cfg;
```
shahor02: Note that the cfg will not go to the logs

lietava:
@shahor02 : database http://ccdb-test.cern.ch:8080/browse/CTP/ is unreachable now ?

shahor02:
Yes, also http://alice-ccdb.cern.ch is unreachable. Pinging to @costing .
I think there is a powercut or network problem in out building since I also cannon reach my desktop.

costing:
Yes, there's a building power cut affecting the network to the servers

lietava:
@shahor02 : thanks for ccdb info. Now I already touched that question: how can I extract from CCDB the files corresponding to particular run ? I need to use validity range ?

shahor02:
Hi, all CCDB objects are accessed via timestamp, not run number. There is only 1 artificially created exception: `RCT/Info/RunInformation` which stores run SOR/EOR times, using the run number as timestamp, at the moment can be accessed via
```
o2::ccdb::CcdbApi cdb;
cdb.init("http://alice-ccdb.cern.ch");
int runNumber = 518131;
auto hd = cdb.retrieveHeaders("RCT/Info/RunInformation", {}, runNumber);
std::cout << stol(hd["SOR"]) << "\n";
```

Not that the EOR info was not stored correctly, the code if fixed but the objects need to be retrofited.

lietava:
@shahor02 : thanks. I do example of run 517732
o2::ccdb::CcdbApi cdb;
  cdb.init("http://alice-ccdb.cern.ch");
  int runNumber = 517732;
  auto hd = cdb.retrieveHeaders("RCT/Info/RunInformation", {}, runNumber);
  std::cout << stol(hd["SOR"]) << "\n";
  long timestamp = stol(hd["SOR"]);
  //timestamp = 1654408907757;
  mng.getScalersFromCCDB(timestamp);

for the code as above it says
[ERROR] Requested resource does not exist: http://ccdb-test.cern.ch:8080/CTP/Scalers/1654408899141/
but if I use timestamp directly from web interface to database:
//timestamp = 1654408907757;
it works ok.

shahor02:
Hi @lietava , for this run SOR indeed gives 1654408899141, where the 1654408907757 comes from?
There is indeed a  http://ccdb-test.cern.ch:8080/CTP/Scalers object created at 1654408907757, but why it is created at that time stamp? Note that the GRPECS creation is not moved to the beginning of the SOR sequences, to have some time margin to propagate the object to  the CCDB. How and when your object is created and what defines its timestamps?

BTW, (1) why are you using http://ccdb-test.cern.ch:8080 and not production server ? (2) we are using 3 level paths, the CTP/Scalers should be adjusted to that, e.g. CTP/Calib/Scalers

lietava:
Hi @shahor02 
1. 1654408907757 comes from http://ccdb-test.cern.ch:8080/browse/CTP/Scalers 06 Jun 2022 08:01 CEST but I see that it gives time Sun Feb 18 06:55:57 54396.  Is it timestamp I created ?
2. Please, let me know which production database I should use.

shahor02:
>     1. 1654408907757 comes from http://ccdb-test.cern.ch:8080/browse/CTP/Scalers 06 Jun 2022 08:01 CEST but I see that it gives time Sun Feb 18 06:55:57 54396.  Is it timestamp I created ?

I guess so, who else could create it?
 
>     2. Please, let me know which production database I should use.

The  http://ccdb-test.cern.ch:8080 is for tests only, for production we are using http://alice-ccdb.cern.ch
The upload of objects there goes via requests on jira (O2 project, CCDB upload type), unless done from the EPN network to  http://o2-ccdb.internal (these 2 databases are syncing each other).

lietava:
1. I meant there is correct date to it. OK, I debug. For start I am using following:
const auto now = std::chrono::system_clock::now();
const long timeStamp = std::chrono::duration_cast<std::chrono::milliseconds>(now.time_since_epoch()).count();
I greped it from some other file in o2.
Maybe the miliseconds is wrong ?

shahor02:
But when are you creating this object? Is this in SOR sequence and where you upload it from? Can you upload it to  http://o2-ccdb.internal ?

lietava:
For the run we discuss I created this object at:
[1924081:ctp-proxy]: [08:02:17][INFO] CTP scalers saved in ccdb, run:517732
I am running it at the moment on my cern virtual machine: ctpo2cs8 as I can apply fixes immediatelly.
 Later I can run it at ctp control machine in P2 or anywhere we agree.

shahor02:
OK, but I don't understand when this object is created, is this done per run and is part of the SOR sequence? Why do you expect it to see for particular run?

lietava:
yes, it is created per run. I am not sure what you mean exactly 'part of SOR sequence'. Itis  following:
ctp control daemon (ctpd)  publishes CTP counters all the time.
ctpd receives from ecs command loadRun.
ctpd sends to ctp-proxy ctp config
ctp-proxy saves config to ccdb
ctpd receives from ecs command startRun.
ctpd sends to ctp-proxy message sox (inside counters)
ctpproxy starts to collect countres for this run 
ctpd receives from ecs command stopRun
ctpd sends to ctp-proxy message eox (inside counters)
ctpproxy close counters and saves them in CCDB.

shahor02:
@vascobarroso or @knopers8 can explain better than me the details of SOR procedure. Now the GRPECS object is created in the very beginning. Apparently `ctpd receives from ecs command loadRun` after that, so the configuration you upload has later timestamp: `1654408907757 - 1654408899141 -> 8.6 s`. I guess you should ask them to provide the timestamp they used to create GRPECS object and use it as a start of validity for your object.
And, to make sure that your object appears in time for the online processing (if it is needed), you should upload to  http://o2-ccdb.internal.

vascobarroso:
@shahor02 currently `trg.RunLoad` is executed by AliECS before `ccdb.RunStart()`.
So I'm not sure why the trg configuration has a later timestamp.  Do you have a concrete run number we can look at ?

shahor02:
You can take any run in last 3 days, when http://alice-ccdb.cern.ch/browse/GLO/Config/GRPECS  creation was moved to the beginning of the SOR sequence. E.g. run 518299 : http://alice-ccdb.cern.ch/browse/GLO/Config/GRPECS/1654850194141/33022a9a-e89a-11ec-8000-0aa145271b9a
BTW, if you know the run number at the time of the ccdb upload, for the object created per run better to put it into the metadata in the format `"runNumber=<number>"`

lietava:
Hi @shahor02 ,
I am looking for 'CCDB upload type' in O2 project and can not find anything relevant. Can you, please, send me issue number ?
I know run number at creation. If I add it to metadata it would solve the search problem ?

shahor02:
For instance: https://alice.its.cern.ch/jira/browse/O2-2992?filter=-2

> I know run number at creation. If I add it to metadata it would solve the search problem ?

No, e.g. it will facilitate tagging your object in the browser. To access any object you still need the timestamps.

lietava:
@shahor02 
1.  I created jira issue 02-3002 to get access to production database.
2. For the moment I plan to run from my virtual machine and write to production database. When debugged/stable we can discuss to run at epn (or is it enough at  alioo2-cr1-ctpctr ?). That would be needed when ctpreadout is ready
3. database validity issue: it seems a bit complicated to send timestamp from ecs to ctpd and so on ... , not using run number seems a bit unnatural. Is there not some other way, e.g. I give ctp objects validity -1, +3 days with respect to sor and  we use metadata runnumber ? Or some other trick ? Otherwise we need to discuss with @vascobarroso .
Cheers, Roman.

shahor02:
@lietava you should not write to the production database, it is not automatically synchronized to online one. Instead, you better directly write to http://o2-ccdb.internal: then your object will be automatically broadcasted to all nodes.

For the validity: if your objects are produced for every SOR, then, like GRPECS, multiple objects may be valid at given time. In this case you must tag them with `runNumber=xxx` in the metadata so that the query to this object is done passing the timestamp and run number at the same time.
I don't see a problem with defining your validity start as e.g. `now() - 1min` to guarantee that it is compatible with the SOR. For the end of validity you can indeed put +3 days, at the EOR I can redefine it to real EOR like we do for the GRPECS.

lietava:
OK, but you said that to http://o2-ccdb.internal/ I have to be on epn network ?
Creating metadata like this:
map<string, string> metadata; // can be empty
metadata["runNumber"] = std::to_string(cfg->getRunNumber());
is correct ?

shahor02:
> OK, but you said that to http://o2-ccdb.internal/ I have to be on epn network ? 

Yes, what host you are using?

Metadata is correct.

lietava:
At the moment I use ctpo2cs8.cern.ch. Later, when more stable I may use: alio2-cr1-ctpctr.cern.ch

shahor02:
> At the moment I use ctpo2cs8.cern.ch. Later, when more stable I may use: alio2-cr1-ctpctr.cern.ch

from the alio2-cr1-ctpctr.cern.ch you should be able to upload to the  http://o2-ccdb.internal, so it will be automatically broadcasted. For the  ctpo2cs8.cern.ch, better @costing tells. Note that if you upload to http://alice-ccdb.cern.ch, then after that 
a ` curl -I -i -s "http://o2-ccdb.internal/latest/<your_object_path>/`date +%s000`/?prepare=sync"` should be called on a machine seeing ` http://o2-ccdb.internal` to trigger a broadcast. 
In principle, it is already done each time the GRPECS is uploaded, but it seems to me that your object is uploaded after the GRPECS.

costing:
`ctpo2cs8.cern.ch` can only reach the production instance, and to write there it would need an identity and permissions on a folder. It might not be worth the effort if this is anyway a temporary setup. I would indeed suggest using `alio2-cr1-ctpctr.cern.ch` to write to `o2-ccdb.internal`, that doesn't need any special privileges and would immediately propagate all new objects to all FLP and EPN nodes.

lietava:
@costing @shahor02 : on the other hand running on ctpctr requires to do software changes via github which slows down development at this stage. Then for tomorrow  I write to test database. Writing to production instance seems to me as reasonable step before going to o2-ccdb.internal. On the other hand object from test can be entered later in production, if needed ?

shahor02:
Either the whole directory on the test server can be synced to production server or particular objects need to retrofitted one by one.

lietava:
@shahor02 : I do not see any issues here ? If so, please, pull.

shahor02 (approved):

PR #9120, opened 2022-06-08
add MC labels for MCH-MID matched tracks

pillot:


Detectors/MUON/Workflow/src/TrackMatcherSpec.h
```diff
@@ -24,7 +24,7 @@ namespace o2
 namespace muon
 {
 
-framework::DataProcessorSpec getTrackMatcherSpec(const char* name = "TrackMatcher");
+framework::DataProcessorSpec getTrackMatcherSpec(bool useMC, const char* name = "TrackMatcher");
```
aphecetche, line 27: not necessarily for this PR, but I'd change the default spec name to something more descriptive, e.g. mch-mid-track-matcher, as "TrackMatcher" is a bit generic for searching in log files ;-)

aphecetche (approved):

Detectors/MUON/Workflow/src/TrackMatcherSpec.h
pillot: Given the time to pass all the checks I would prefer to do that in another PR ;-)
And it is currently replaced by "muon-track-matcher" in the corresponding workflow.

PS: I am still not allowed to merge..

PR #9121, opened 2022-06-08
Added TRDTrackQC parameters

f3sch:
As preliminary parameters for my thesis, I added some Track parameters to the QC workflow.
I need the charges for each layer, the momentum, the number of layers in which matching tracklets have been found.
Additionally I will start my analysis using the TPC PID, but will later look at specific decays.

Also I ran clang-format on the relevant files

### DIFF
--- a/Detectors/TRD/qc/include/TRDQC/Tracking.h
+++ b/Detectors/TRD/qc/include/TRDQC/Tracking.h
@@ -44,13 +44,13 @@ namespace trd
 struct TrackQC {
   int type;          ///< 0 TPC-TRD track; 1 ITS-TPC-TRD track
   int nTracklets;    ///< number of attached TRD tracklets
-  int nLayers; //< Number of associated Layers of a Track
+  int nLayers;       //< Number of associated Layers of a Track
   float chi2;        ///< total chi2 value for the track
   float reducedChi2; ///< chi2 total divided by number of layers in which track is inside TRD geometrical acceptance
   float pt;          ///< the transverse momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
   float ptSigma2;    //< Sigma2 of pt
   track::PID pid;    //< assigned particle id from TPC
-  float phi; //< Phi 0:2Pi value of Track
+  float phi;         //< Phi 0:2Pi value of Track
 
   // layer-wise information for seeding track and assigned tracklet (if available)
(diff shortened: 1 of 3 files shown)


Detectors/TRD/qc/src/Tracking.cxx
```diff
@@ -55,9 +55,12 @@ void Tracking::checkTrack(const TrackTRD& trkTrd, bool isTPCTRD)
   TrackQC qcStruct;
   qcStruct.type = isTPCTRD ? 0 : 1;
   qcStruct.nTracklets = trkTrd.getNtracklets();
+  qcStruct.nLayers = trkTrd.getNlayersFindable();
```
martenole, line 58: Note that "number of associated layers of a track" is actually the number of layers in which the track extrapolation was inside the geometrical acceptance of the TRD. It does not mean that a tracklet was added. This is the `nTracklets` variable

martenole (dismissed):
Maybe you can make the comment on nLayers a bit more precise, otherwise this looks good

f3sch:
I added a better description, requesting new review.

martenole (dismissed):

bazinski (dismissed):

martenole:
Hi Felix, could you squash your commits into a single one and then force-push again? This will trigger a restart of the CI which failed due to reasons unrelated to your changes

martenole:
> Hi Felix, could you squash your commits into a single one and then force-push again? This will trigger a restart of the CI which failed due to reasons unrelated to your changes

Hi @f3sch could you please do the squash? E.g. with `git rebase -i HEAD~4` and then create only a single commit with a short description of your changes

Detectors/TRD/qc/include/TRDQC/Tracking.h
```diff
@@ -75,6 +74,8 @@ struct TrackQC {
   std::array<int, constants::NLAYER> trackletMcm{};                                      ///< the MCM number of the tracklet
   std::array<float, constants::NLAYER> trackletChi2{};                                   ///< estimated chi2 for the update of the track with the given tracklet
   std::array<std::array<int, constants::NCHARGES>, constants::NLAYER> trackletCharges{}; ///< charges of tracklets
+  std::array<float, constants::NLAYER> trackletPhi{};         //< Phi 0:2Pi value of Tracklet
```
martenole: @f3sch this should be `trackPhi` and `trackEta`. Could you please add it to the paragraph with the track parameters?

martenole (approved):

PR #9123, opened 2022-06-08
optional blocking mode for CCDB refresh from o2-ecs-grp-create

shahor02:
option --refresh changed from bool to string with:
\"none\" (or default, no --refresh passed): do not refresh
\"async\" or implicit (--refresh w/o arguments): refresh in non-blocking mode
\"sync\" : refresh in blocking mode

@knopers8


shahor02:
trivial, tested locally

PR #9128, opened 2022-06-08
Use raw string to fix check

ktf:


ktf:
@shahor02 to fix the codechecker...

shahor02:
thanks, but where it was failing? https://github.com/AliceO2Group/AliceO2/pull/9123 passed all formatting checks before I merged it.

ktf:
fullCi, where we run clang static checks

Ciao,
Giulio

shahor02:
OK, but then this also should be fixed : 
https://github.com/AliceO2Group/AliceO2/blob/387b76446d374a6e8c141cb85e847a78b5e51718/Detectors/GRP/workflows/src/create-grp-ecs.cxx#L192

ktf:
Indeed... Merging since it's currently broken. Tested locally.

PR #9133, opened 2022-06-09
GRP-ECS: Fix modernize-raw-string-literal CI failure

sawenzel:


sawenzel:
@davidrohr : This should bring the fullCI back to green.

sawenzel:
Ok. Seems to have been fixed meanwhile.

sawenzel:
The option was taken out

PR #9136, opened 2022-06-09
store vectors for EnvVars, Collimators, change a bit the policy

chiarazampolli:
To test:
```
o2-calibration-grp-dcs-dps-sim-workflow --max-timeframes 100 --delta-fraction 0.5 --max-cycles-no-full-map 3 | o2-calibration-grp-dcs-dps-workflow --use-verbose-mode -b > out.out
```

### DIFF
--- a/Detectors/Calibration/include/DetectorsCalibration/Utils.h
+++ b/Detectors/Calibration/include/DetectorsCalibration/Utils.h
@@ -25,6 +25,7 @@
 #include "CCDB/CcdbObjectInfo.h"
 #include "CommonUtils/MemFileHelper.h"
 #include "CCDB/CcdbApi.h"
+#include <vector>
 
 namespace o2
 {
@@ -32,11 +33,58 @@ namespace calibration
 {
 
 struct Utils {
+
+  enum ValueType { Invalid = 0,
+                   Interpolation,
+                   ClosestAvailableFromBelow,
+                   ClosestAvailableFromAbove,
+                   SameAsRequested };
+
   static constexpr o2::header::DataOrigin gDataOriginCDBPayload{"CLP"}; // generic DataOrigin for calibrations payload
   static constexpr o2::header::DataOrigin gDataOriginCDBWrapper{"CLW"}; // generic DataOrigin for calibrations wrapper
   template <typename T>
   static void prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const std::string& path,
                                     const std::map<std::string, std::string>& md, long start, long end = -1);
+
+  static std::pair<int, double> findPair(const std::vector<std::pair<uint64_t, double>>& vect, uint64_t timestamp)
+  {
+    // function to find the pair in the vector with the timestamp closest in the past to the one passed as argument
+    // if two entries in the vector have the same timestamp, the first that is found is used
+
+    // let's first check that the elements are sorted in ascending order
+    if (!std::is_sorted(vect.begin(), vect.end(), [](const std::pair<uint64_t, double>& lhs, const std::pair<uint64_t, double>& rhs) -> bool { return lhs.first < rhs.first; })) {
+      LOG(fatal) << "Vector is not sorted, we cannot execute the findPair function";
+    }
+    auto lower = std::lower_bound(vect.begin(), vect.end(), timestamp, [](const std::pair<uint64_t, double>& p, uint64_t value) { return p.first < value; });
+    if ((*lower).first == timestamp) {
+      LOG(debug) << "We found the element for the exact timestamp";
+      return std::make_pair(SameAsRequested, vect[std::distance(vect.begin(), lower)].second);
+    } else if (lower == vect.end()) {
+      LOG(debug) << "All values are smaller than the queried one " << timestamp << ", we return the closest available from below: " << vect.back().first;
+      return std::make_pair(ClosestAvailableFromBelow, vect.back().second);
+    } else if (lower == vect.begin()) {
+      LOG(debug) << "All values are greater that the queried one " << timestamp << ", we return the closest available from above: " << (*vect.begin()).first;
+      return std::make_pair(ClosestAvailableFromAbove, vect.begin()->second);
+    } else {
+      // doing interpolation
+      const auto& p1 = vect[std::distance(vect.begin(), lower) - 1];
+      const auto& p2 = vect[std::distance(vect.begin(), lower)];
+      auto t1 = p1.first;
+      auto t2 = p2.first;
+      auto val1 = p1.second;
+      auto val2 = p2.second;
+      if (t1 == t2) {
+        LOG(debug) << "times are the same, cannot interpolate, returning closest from below";
+        return std::make_pair(ClosestAvailableFromBelow, val1);
+      }
+      // in case values are not ordered in timestamp
+      double val = t2 > t1 ? (timestamp - t1) * (val2 - val1) / (t2 - t1) + val1 : (timestamp - t1) * (val1 - val2) / (t1 - t2) + val1;
+      LOG(debug) << "Doing interpolation between (" << t1 << ", " << val1 << ") and (" << t2 << ", " << val2 << ") for t = " << timestamp << " --> " << val;
+      return std::make_pair(Interpolation, val);
+    }
+    LOG(error) << "Something went wrong!";
+    return std::make_pair(Invalid, 0);
+  }
 };
 
 template <typename T>
--- a/Detectors/GRP/calibration/include/GRPCalibration/GRPDCSDPsProcessor.h
+++ b/Detectors/GRP/calibration/include/GRPCalibration/GRPDCSDPsProcessor.h
@@ -41,25 +41,16 @@ inline unsigned long llu2lu(std::uint64_t v) { return (unsigned long)v; }
 
 struct GRPEnvVariables {
 
-  std::pair<uint64_t, double> mCavernTemperature;
-  std::pair<uint64_t, double> mCavernAtmosPressure;
-  std::pair<uint64_t, double> mSurfaceAtmosPressure;
-  std::pair<uint64_t, double> mCavernAtmosPressure2;
-
-  GRPEnvVariables()
-  {
-    mCavernTemperature = std::make_pair(0, -999999999);
-    mCavernAtmosPressure = std::make_pair(0, -999999999);
-    mSurfaceAtmosPressure = std::make_pair(0, -999999999);
-    mCavernAtmosPressure2 = std::make_pair(0, -999999999);
-  }
+  std::unordered_map<std::string, std::vector<std::pair<uint64_t, double>>> mEnvVars;
 
   void print()
   {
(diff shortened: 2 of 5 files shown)


Detectors/GRP/calibration/src/GRPDCSDPsProcessor.cxx
```diff
@@ -59,8 +60,7 @@ int GRPDCSDPsProcessor::process(const gsl::span<const DPCOM> dps)
     }
   }
   mMagFieldHelper.updated = false;
```
shahor02, line 50: why https://github.com/AliceO2Group/AliceO2/pull/9136/files#diff-9f1a769e89aa2a999c2f76dedb0239499ab2afbec8a0a95e9123e6861cbf8144R50-R61 is needed, it does nothing but consumes CPU.
Also https://github.com/AliceO2Group/AliceO2/blob/606d34794df1fd64829e79a81b0f89989fdd42af/Detectors/GRP/calibration/src/GRPDCSDPsProcessor.cxx#L69-L73 can be suppressed

Detectors/GRP/calibration/include/GRPCalibration/GRPDCSDPsProcessor.h
```diff
@@ -109,31 +121,27 @@ struct MagFieldHelper {
...
-    mleft_downstream = std::make_pair(0, -999999999);
-    mleft_upstream = std::make_pair(0, -999999999);
-    mright_downstream = std::make_pair(0, -999999999);
-    mright_upstream = std::make_pair(0, -999999999);
+    mCollimators["LHC_CollimatorPos_TCLIA_4R2_lvdt_gap_downstream"].emplace_back(0, -999999999);
+    mCollimators["LHC_CollimatorPos_TCLIA_4R2_lvdt_gap_upstream"].emplace_back(0, -999999999);
+    mCollimators["LHC_CollimatorPos_TCLIA_4R2_lvdt_left_downstream"].emplace_back(0, -999999999);
+    mCollimators["LHC_CollimatorPos_TCLIA_4R2_lvdt_left_upstream"].emplace_back(0, -999999999);
+    mCollimators["LHC_CollimatorPos_TCLIA_4R2_lvdt_right_downstream"].emplace_back(0, -999999999);
+    mCollimators["LHC_CollimatorPos_TCLIA_4R2_lvdt_right_upstream"].emplace_back(0, -999999999);
```
shahor02: same for collimators

Detectors/GRP/calibration/include/GRPCalibration/GRPDCSDPsProcessor.h
```diff
@@ -41,25 +41,24 @@ inline unsigned long llu2lu(std::uint64_t v) { return (unsigned long)v; }
...
   GRPEnvVariables()
   {
-    mCavernTemperature = std::make_pair(0, -999999999);
-    mCavernAtmosPressure = std::make_pair(0, -999999999);
-    mSurfaceAtmosPressure = std::make_pair(0, -999999999);
-    mCavernAtmosPressure2 = std::make_pair(0, -999999999);
+    mEnvVars["CavernTemperature"].emplace_back(0, -999999999);
+    mEnvVars["CavernAtmosPressure"].emplace_back(0, -999999999);
+    mEnvVars["SurfaceAtmosPressure"].emplace_back(0, -999999999);
+    mEnvVars["CavernAtmosPressure2"].emplace_back(0, -999999999);
```
shahor02: why do you need this dummy element? If just to have something to pass to the `processPairD` you can simply pass the vector itself instead of `back()` and always update if it is empty.

Not urgent, but we would need also some easy method to get the value of the point for given timestamp, given that now we store a vector.

Detectors/GRP/calibration/src/GRPDCSDPsProcessor.cxx
chiarazampolli: done

Detectors/GRP/calibration/include/GRPCalibration/GRPDCSDPsProcessor.h
chiarazampolli: I wanted to avoid to have sometimes something empty (even if it should not happen, since the first processing will be on a FBI). So I used a dummy first value. I can surely remove it.

shahor02: What is wrong with having empty entry?

chiarazampolli: Nothing, just thought that it was better to see that the entry was a dummy value. I am changing it.

chiarazampolli: It was also to initialize the items of the map. Now it is done in the code with the access to the items themselves, l.147:

```
  auto& vect = mapToUpdate[alias]; // we also create at this point the vector in the map if it did not exist yet
```

chiarazampolli:
Hi @shahor02 . 

I added the function to retrieve the value from the vector. You can test with a macro like this:

```
#include "DetectorsCalibration/Utils.h"
#include "CommonUtils/StringUtils.h"
#include <vector>
#include <utility>

void findTimeUtils(uint64_t timestamp) {

  std::vector<std::pair<uint64_t, double>> vect;
  vect.emplace_back(1230, 10.1);
  vect.emplace_back(1240, 20.1);
  vect.emplace_back(1250, 30.1);
  vect.emplace_back(1260, 40.1);
  vect.emplace_back(1260, 40.1);
  vect.emplace_back(1270, 50.1);

  std::pair<uint64_t, double> p = o2::calibration::Utils::findPair(vect, 1);
  std::cout << "p.first " << p.first << ", p. second " << p.second << std::endl;
  std::cout << "\n";
  p = o2::calibration::Utils::findPair(vect, 1230);
  std::cout << "p.first " << p.first << ", p. second " << p.second << std::endl;
  std::cout << "\n";
  p = o2::calibration::Utils::findPair(vect, 1235);
  std::cout << "p.first " << p.first << ", p. second " << p.second << std::endl;
  std::cout << "\n";
  p = o2::calibration::Utils::findPair(vect, 1260);
  std::cout << "p.first " << p.first << ", p. second " << p.second << std::endl;
  std::cout << "\n";
  p = o2::calibration::Utils::findPair(vect, 1290);
  std::cout << "p.first " << p.first << ", p. second " << p.second << std::endl;
  std::cout << "\n";

 }
```

Detectors/Calibration/include/DetectorsCalibration/Utils.h
```diff
@@ -25,18 +25,65 @@
...
+                   ClosestAvailableFromAbove,
+                   SameAsRequested };
+
   static constexpr o2::header::DataOrigin gDataOriginCDBPayload{"CLP"}; // generic DataOrigin for calibrations payload
   static constexpr o2::header::DataOrigin gDataOriginCDBWrapper{"CLW"}; // generic DataOrigin for calibrations wrapper
   template <typename T>
   static void prepareCCDBobjectInfo(T& obj, o2::ccdb::CcdbObjectInfo& info, const std::string& path,
                                     const std::map<std::string, std::string>& md, long start, long end = -1);
+
+  static std::pair<int, double> findPair(std::vector<std::pair<uint64_t, double>> vect, uint64_t timestamp)
```
shahor02: pass by const ref, i.e.
```
static std::pair<int, double> findPair(const std::vector<std::pair<uint64_t, double>>& vect, uint64_t timestamp)
```

Detectors/Calibration/include/DetectorsCalibration/Utils.h
```diff
@@ -25,18 +25,65 @@
...
+  {
+    // function to find the pair in the vector with the timestamp closest in the past to the one passed as argument
+    // if two entries in the vector have the same timestamp, the first that is found is used
+    std::vector<uint64_t> times;
+    for (const auto& el : vect) {
+      times.push_back(el.first);
+    }
+    auto lower = std::lower_bound(times.begin(), times.end(), timestamp);
```
shahor02: use lambda instead of creating a vector:
```
auto lower = std::lower_bound(times.begin(), times.end(), timestamp, [](const std::pair<uint64_t, double>& p, uint64_t value) { return p.first < value; } );
```

Respectivelt, use directly the `vect` instead of `times` below.

Detectors/Calibration/include/DetectorsCalibration/Utils.h
```diff
@@ -25,18 +25,65 @@
...
+      auto t1 = p1.first;
+      auto t2 = p2.first;
+      auto val1 = p1.second;
+      auto val2 = p2.second;
+      if (t1 == t2) {
+        LOG(debug) << "times are the same, cannot interpolate, returning closest from below";
+        return std::make_pair(ClosestAvailableFromBelow, val1);
+      }
+      // in case values are not ordered in timestamp
+      double val = t2 > t1 ? (timestamp - t1) * (val2 - val1) / (t2 - t1) + val1 : (timestamp - t1) * (val1 - val2) / (t1 - t2) + val1;
```
shahor02, line 81: how values can be not ordered? If this is possible, then the whole method does not work

Detectors/GRP/calibration/include/GRPCalibration/GRPDCSDPsProcessor.h
```diff
@@ -153,20 +143,20 @@ struct GRPLHCInfo {
...
 
-  void reset()
+  void resetAndKeepLast()
   {
     for (int i = 0; i < 2; ++i) {
-      mIntensityBeam[i].clear();
-      mBPTXPhase[i].clear();
-      mBPTXPhaseRMS[i].clear();
-      mBPTXPhaseShift[i].clear();
+      mIntensityBeam[i].erase(mIntensityBeam[i].begin(), mIntensityBeam[i].end() - 1);
```
shahor02: you should probably check if beging == end

shahor02 (changes requested):

Detectors/Calibration/include/DetectorsCalibration/Utils.h
chiarazampolli: Yes, of course.

Detectors/Calibration/include/DetectorsCalibration/Utils.h
chiarazampolli: Done.

Detectors/Calibration/include/DetectorsCalibration/Utils.h
chiarazampolli: You are right, I wanted to have a protection. 
Ok, I can add a check at the very beginning.

chiarazampolli:
Done, comments implemented with the check on the fact that the vector is sorting in the function "findPair".

shahor02 (approved):
thanks, the failures are unrelated. Before merging: did you test it? It will be picked by the processor...

chiarazampolli:
Hi @shahor02 , 

Yes, I tested it with the emulator, as reported in the first comment in this PR. You can also try.

Cheers,

Chiara

PR #9138, opened 2022-06-09
[EMCAL-778] Implementation of LM-L0 delay in BC

hahassan7:
The delay of the L0 (observed BC by FEE) with respect to LM (true BC of the collision) needs to be implemented in the raw creator. For more information see https://alice.its.cern.ch/jira/browse/CTP-133.

### DIFF
--- a/Detectors/EMCAL/simulation/src/RawWriter.cxx
+++ b/Detectors/EMCAL/simulation/src/RawWriter.cxx
@@ -14,6 +14,7 @@
 #include <fmt/core.h>
 #include <gsl/span>
 #include <TSystem.h>
+#include "DataFormatsCTP/TriggerOffsetsParam.h"
 #include "DataFormatsEMCAL/Constants.h"
 #include "EMCALBase/Geometry.h"
 #include "EMCALBase/RCUTrailer.h"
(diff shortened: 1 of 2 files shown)


mfasDa:
@shahor02 After adding the delay, does the RawFileWriter handle IRs which exceed the timeframe or do we need to exclude them in our part of the RawWriter?

shahor02:
> @shahor02 After adding the delay, does the RawFileWriter handle IRs which exceed the timeframe or do we need to exclude them in our part of the RawWriter?

It will store whatever IR you supply, the new TF will be added if needed.

mfasDa:
Thanks @shahor02! Then according to my understanding it should be fine.

mfasDa (approved):

Detectors/EMCAL/simulation/src/RawWriter.cxx
```diff
@@ -152,7 +153,7 @@ bool RawWriter::processTrigger(const o2::emcal::TriggerRecord& trg)
     auto ddlid = srucont.mSRUid;
     auto [crorc, link] = mGeometry->getLinkAssignment(ddlid);
     LOG(debug1) << "Adding payload with size " << payload.size() << " (" << payload.size() / 4 << " ALTRO words)";
-    mRawWriter->addData(ddlid, crorc, link, 0, trg.getBCData(), payload, false, trg.getTriggerBits());
+    mRawWriter->addData(ddlid, crorc, link, 0, trg.getBCData() + o2::ctp::TriggerOffsetsParam::Instance().LM_L0, payload, false, trg.getTriggerBits());
```
shahor02, line 156: Don't you need to account for it in the decoding?

mfasDa: See the second PR: https://github.com/AliceO2Group/AliceO2/pull/9137

shahor02 (approved):

shahor02:
@TimoWilken the fullCI has failed due to the missing alien-token, could you check, please?

PR #9141, opened 2022-06-10
DPL Analysis: Event mixing: Get bins counts, nitpick for (unused yet) binning with overflows

saganatt:
Get number of all bins and along each axis for easier debugging / histogramming.

Binning with overflows has not been used by anyone yet, so it didn't impact any analysis. I just noted the bins calculations in such case would not cover all possible bins.


saganatt:
Hi, can we merge this?

jgrosseo (approved):

PR #9147, opened 2022-06-10
Split proxies depending on lifetime of specs

chiarazampolli:
@davidrohr 

For now, I don't know which specs have Lifetime Sporadic, so this is WIP.
Goes with https://github.com/AliceO2Group/O2DPG/pull/460

### DIFF
--- a/prodtests/full-system-test/aggregator-workflow.sh
+++ b/prodtests/full-system-test/aggregator-workflow.sh
@@ -44,67 +44,124 @@ if [[ $BEAMTYPE == "PbPb" ]]; then
 fi
 
 # Calibration workflows
-if ! workflow_has_parameter CALIB_LOCAL_INTEGRATED_AGGREGATOR; then WORKFLOW=; fi
+if ! workflow_has_parameter CALIB_LOCAL_INTEGRATED_AGGREGATOR; then
+  WORKFLOW=
+else
+  if [[ -z $AGGREGATOR_TASKS ]]; then
+    AGGREGATOR_TASKS=ALL
+  fi
+fi
+
+if [[ -z $AGGREGATOR_TASKS ]]; then
+  echo "ERROR: AGGREGATOR_TASKS is empty, you need to define it to run!" 1>&2
+  exit 1
+fi
+
+# Which calibrations are we aggregating
+echo "AGGREGATOR_TASKS = $AGGREGATOR_TASKS" 1>&2
 
 # adding input proxies
 if workflow_has_parameter CALIB_PROXIES; then
-  if [[ ! -z $CALIBDATASPEC_BARREL ]]; then
-    add_W o2-dpl-raw-proxy "--dataspec \"$CALIBDATASPEC_BARREL\" $(get_proxy_connection barrel input)" "" 0
-  fi
-  if [[ ! -z $CALIBDATASPEC_CALO ]]; then
-    add_W o2-dpl-raw-proxy "--dataspec \"$CALIBDATASPEC_CALO\" $(get_proxy_connection calo input)" "" 0
+  if [[ $AGGREGATOR_TASKS == BARREL_TF ]]; then
+    if [[ ! -z $CALIBDATASPEC_BARREL_TF ]]; then
+      add_W o2-dpl-raw-proxy "--dataspec \"$CALIBDATASPEC_BARREL_TF\" $(get_proxy_connection barrel_tf input)" "" 0
+    fi
+  elif [[ $AGGREGATOR_TASKS == BARREL_SPORADIC ]]; then
+    if [[ ! -z $CALIBDATASPEC_BARREL_SPORADIC ]]; then
+      add_W o2-dpl-raw-proxy "--dataspec \"$CALIBDATASPEC_BARREL_SPORADIC\" $(get_proxy_connection barrel_sp input)" "" 0
+    fi
+  elif [[ $AGGREGATOR_TASKS == CALO_TF ]]; then
+    if [[ ! -z $CALIBDATASPEC_CALO_TF ]]; then
+      add_W o2-dpl-raw-proxy "--dataspec \"$CALIBDATASPEC_CALO_TF\" $(get_proxy_connection calo_tf input)" "" 0
+    fi
+  elif [[ $AGGREGATOR_TASKS == CALO_SPORADIC ]]; then
+    if [[ ! -z $CALIBDATASPEC_CALO_SPORADIC ]]; then
+      add_W o2-dpl-raw-proxy "--dataspec \"$CALIBDATASPEC_CALO_SPORADIC\" $(get_proxy_connection calo_sp input)" "" 0
+    fi
+  elif [[ $AGGREGATOR_TASKS == MUON_TF ]]; then
+    if [[ ! -z $CALIBDATASPEC_MUON_TF ]]; then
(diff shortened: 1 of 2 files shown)


chiarazampolli:
@davidrohr , can you check if this is what satisfies the requirements? 
The aggregator needs to be called with `AGGREGATOR_TASKS` defined as one of:
```
BARREL_TF
BARREL_SPORADIC
CALO_TF
CALO_SPORADIC
MUON_TF
MUON_SPORADIC
```

prodtests/full-system-test/aggregator-workflow.sh
```diff
@@ -44,67 +44,120 @@ if [[ $BEAMTYPE == "PbPb" ]]; then
...
 # Calibration workflows
-if ! workflow_has_parameter CALIB_LOCAL_INTEGRATED_AGGREGATOR; then WORKFLOW=; fi
+if ! workflow_has_parameter CALIB_LOCAL_INTEGRATED_AGGREGATOR; then
+    WORKFLOW=
+else
+    if [[ -z $AGGREGATOR_TASKS ]]; then
+  ech HERE1
+    AGGREGATOR_TASKS=ALL
+  fi
+fi
```
davidrohr: Seems to be a typo `ech`, but anyhow this seems to be debug output that should go.
Second, I would always run the check for the aggregator task, also in case `CALIB_LOCAL_INTEGRATED_AGGREGATOR` is set

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -412,7 +412,7 @@ fi
 # ---------------------------------------------------------------------------------------------------------------------
 # Calibration workflows
 workflow_has_parameter CALIB && { source ${CALIB_WF:-$MYDIR/calib-workflow.sh}; [[ $? != 0 ]] && exit 1; }
-workflow_has_parameters CALIB CALIB_LOCAL_INTEGRATED_AGGREGATOR && { source ${CALIB_AGGREGATOR_WF:-$MYDIR/aggregator-workflow.sh}; [[ $? != 0 ]] && exit 1; }
+workflow_has_parameters CALIB CALIB_LOCAL_INTEGRATED_AGGREGATOR && { export AGGREGATOR_TASKS=ALL; source ${CALIB_AGGREGATOR_WF:-$MYDIR/aggregator-workflow.sh}; [[ $? != 0 ]] && exit 1; }
```
davidrohr: see comment above, I'd do the check for missing `AGGREGATOR_TASKS` inside `aggregator-workflow.sh`

davidrohr (changes requested):

prodtests/full-system-test/aggregator-workflow.sh
```diff
@@ -44,67 +44,120 @@ if [[ $BEAMTYPE == "PbPb" ]]; then
...
-    add_W o2-dpl-raw-proxy "--dataspec \"$CALIBDATASPEC_CALO\" $(get_proxy_connection calo input)" "" 0
+  if [[ $AGGREGATOR_TASKS == BARREL_TF ]]; then
+      if [[ ! -z $CALIBDATASPEC_BARREL_TF ]]; then
+    add_W o2-dpl-raw-proxy "--dataspec \"$CALIBDATASPEC_BARREL_TF\" $(get_proxy_connection barrel_tf input)" "" 0
+      fi
+  elif [[ $AGGREGATOR_TASKS == BARREL_SPORADIC ]]; then
+      if [[ ! -z $CALIBDATASPEC_BARREL_SPORADIC ]]; then
+    add_W o2-dpl-raw-proxy "--dataspec \"$CALIBDATASPEC_BARREL_SPORADIC\" $(get_proxy_connection barrel_sp input)" "" 0
+      fi
+  elif [[ $AGGREGATOR_TASKS == CALO_TF ]]; then
```
davidrohr, line 73: Also the indentation seems strange. Why isn't the `add_W ...` indented?

prodtests/full-system-test/aggregator-workflow.sh
chiarazampolli: Oh sorry, I wanted to remove the debug printout when I did the commit. Will fix it, sure. 
Ok, what if AGGREGATOR_TASKS is not set? if the aggregator is local, then I know I want ALL, but if not...?

prodtests/full-system-test/aggregator-workflow.sh
chiarazampolli: This is the result from clang-format + `sed -i 's/[[:space:]]*$//'` and `sed -i 's/\t/  /g'`
I can check again.

davidrohr: > This is the result from clang-format + `sed -i 's/[[:space:]]*$//'` and `sed -i 's/\t/ /g'`
> I can check again.

Can it be that your editor inserts TABs instead of 8 spaces, and then your `sed` command replaces the TAB with 2 spaces?

prodtests/full-system-test/aggregator-workflow.sh
davidrohr: Well, if it is not set, it will also default to `ALL`.

prodtests/full-system-test/aggregator-workflow.sh
chiarazampolli: I will check.

prodtests/full-system-test/aggregator-workflow.sh
chiarazampolli: Only if we have the aggregator running locally. Otherwise, mixing proxies will not work.

davidrohr: I don't understand what you mean. This setting specifies which input proxies are created, and which calib tasks are running.
So in my opinion, it makes sense with and without proxies.
Without proxies, it will simply run only a subset of the aggregation tasks. The default is `AGGREGATOR_TASKS=ALL`, which you already have, so by default it would run all tasks.
So if you make the default here independent from the `if ! workflow_has_parameter CALIB_LOCAL_INTEGRATED_AGGREGATOR; then`, then you don't need to explicitly set it in `dpl-workflow.sh`.

chiarazampolli: I am then not understanding. I thought that we can run all the tasks only if we have no proxy, otherwise there will be too many input proxies in the aggregator.

prodtests/full-system-test/aggregator-workflow.sh
chiarazampolli: Emacs was putting 4 spaces - no idea why, it always worked. Now I changed to 2, let's see if it works...

prodtests/full-system-test/aggregator-workflow.sh
davidrohr: ok, I see that point.
But what will happen if AGGREGATOR_TASK is not set in that case? Will it just produce an empty workflow? I think that would be ok.
Or we add a check and throw an error message if `AGGREGATOR_TASK` is not set.

Still, the `export AGGREGATOR_TASKS=ALL` in `dpl-workflow.sh` is not needed, since it will default to `ALL`  in `aggregator-workflow.sh`

chiarazampolli: I was thinking about adding an error if "AGGREGATOR_TASK" is not set. 
And you are right on the second point, it is not needed in dpl-workflow - I added it there before adding it in aggregator-workflow.

chiarazampolli: Done. Let me know if you have more comments.

PR #9148, opened 2022-06-10
Keeping last values for TOF DPs

chiarazampolli:
Being tested.


chiarazampolli:
@noferini : like this, we will keep always the last value that was stored for each DP. 
Note that:
- at the moment, there is no check whether all values are filled, not even at the very beginning; if you want it can be added; if some DPs are not there, dummy values will be passed;
- the situation above should not happen because the first data that we will see will always be a Full Buffer Image with all DPs.

noferini (dismissed):

chiarazampolli:
I want to make one more change, sorry. So i moved this to WIP again.

chiarazampolli:
Ok, ready for review @noferini .

Now I store also the min value and max value seen, and the max change stores the timestamps of the measurements that gave the change.

noferini (approved):

noferini:
fullCI failing because of
Error in <TJAlienFile::Open>: No JAliEn Grid connection available!

not related to this PR

PR #9149, opened 2022-06-10
FIT: empty TF dummy protection + some fixes for new FT0 decoder

afurs:


jotwinow (dismissed):

shahor02:
@ironMann this PR allows to avoid TFs with empty payload, but have not you fixed DD to be able to process empty STFs (payload size = 0)?

afurs:
Hello @shahor02 , fullCI failed, seems to be unrelated issue. Shall I make new commit and restart it?

afurs:
it looks like unrelated failure build/O2/o2

shahor02:
Hi @afurs, the failure is indeed unrelated, but for the empty TF protection, frankly speaking, I don't like such a hack, the DD should be able to cope with empty payload and I thought this is fixed. Let me ping @ironMann  again.

shahor02:
@afurs, @ironMann just wrote me that the fix for DD to cope with empty payload is there since 2 weeks. If your reconstruction sends to StfSender a header + empty payload, it will be propagated to EPN. Could you suppress sending dummy channels in this PR? And you should probably check that propagating empty payload indeed works.

afurs:
@shahor02 there are option `disable-empty-tf-protection` which allows to suppress this protection. And I also tested on ECS in two cases:
- 1 Hz of artificial data (1 Digit object within container per 44 TFs) - failures were observed at readout
- 44 Hz (1 Digit object per 1 TFs) - success, stable run

shahor02:
@afurs when did you test it, do you remember O2PDPSuite version?

afurs:
@shahor02 I tested it on Friday , 10th June

shahor02 (approved):
ok, thanks, merging but once @ironMann is back we should clarify this problem and remove this hack.

PR #9154, opened 2022-06-12
fix for TOF async calib

noferini:


noferini:
@chiarazampolli 
I am currently running a test on the pilot beam 2021 calibration outputs of apass4 to validate all the steps (test ccdb).
If fine I will open a JIRA asking you to re-run it with production ccdb.

Note that for the next runs (apass1) I added a flag in tof-reco-workflow to specify that calibrations are not yet done (to be done in apass1). This will switch off the filter on problematic channels (loaded from ccdb of previous runs) since problematic channels should be defined after calibration.

noferini:
PR is ready for the review.
Tests with 2021 pilot beam look fine.

noferini:
failing checks not related to this PR

Detectors/TOF/base/src/CalibTOFapi.cxx
```diff
@@ -272,6 +272,28 @@ float CalibTOFapi::getTimeCalibration(int ich, float tot)
 
 //______________________________________________________________________
 
+float CalibTOFapi::getTimeCalibration(int ich, float tot, float phase)
```
shahor02, line 275: this should be const right? Can be changed in next PR

shahor02 (approved):

Detectors/TOF/base/src/CalibTOFapi.cxx
noferini: Yes, I forgot sorry

PR #9156, opened 2022-06-12
example macro of TPCFastTransform initialisation from voxels

sgorbuno:
A macro with an example of TPCFastTransform initialization from voxels.

@martenole, please have a look.

### DIFF
--- a/GPU/TPCFastTransformation/CMakeLists.txt
+++ b/GPU/TPCFastTransformation/CMakeLists.txt
@@ -83,12 +83,14 @@ if(${ALIGPU_BUILD_TYPE} STREQUAL "O2")
           SplineDemo.C
           SplineRecoveryDemo.C
           fastTransformQA.C
+          TPCFastTransformInit.C
           generateTPCCorrectionNTuple.C)
     o2_add_test_root_macro(macro/${m}
                            PUBLIC_LINK_LIBRARIES O2::TPCFastTransformation
                                                  O2::DataFormatsTPC
                                                  O2::TPCSimulation
                                                  O2::TPCReconstruction
+                                                 O2::SpacePoints
                           PUBLIC_INCLUDE_DIRECTORIES
                                                  ${CMAKE_BINARY_DIR}/stage/include
                            LABELS gpu tpc COMPILE_ONLY)
--- a/GPU/TPCFastTransformation/macro/TPCFastTransformInit.C
+++ b/GPU/TPCFastTransformation/macro/TPCFastTransformInit.C
@@ -1,0 +1,155 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file  generateTPCCorrection.C
+/// \brief A macro for generating TPC fast transformation
+///        out of set of space charge correction voxels
+///
+/// \author  Sergey Gorbunov <sergey.gorbunov@cern.ch>
+///
+
+/// how to run the macro:
+///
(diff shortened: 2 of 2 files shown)


GPU/TPCFastTransformation/CMakeLists.txt
```diff
@@ -83,6 +83,7 @@ if(${ALIGPU_BUILD_TYPE} STREQUAL "O2")
           SplineDemo.C
           SplineRecoveryDemo.C
           fastTransformQA.C
+          TPCFastTransformInit.C
           generateTPCCorrectionNTuple.C)
     o2_add_test_root_macro(macro/${m}
                            PUBLIC_LINK_LIBRARIES O2::TPCFastTransformation
```
martenole, line 89: I think here you need to add `O2::SpacePoints` to the PUBLIC_LINK_LIBRARIES

GPU/TPCFastTransformation/macro/TPCFastTransformInit.C
```diff
@@ -0,0 +1,159 @@
...
+
+void TPCFastTransformInit(const char* fileName = "debugVoxRes.root")
+{
+  // open file with space-charge density histograms
+  // location: alien:///alice/cern.ch/user/e/ehellbar/TPCSpaceCharge/RUN3/ SCDensityHistograms/InputSCDensityHistograms.root
+  // The file contains several sc densities for epsilon=20 scenario
+  // Naming convention: inputSCDensity3D<nPileUpEvents>
+  //          - nPileUpEvents = 0,2,4,6,8,10,12
+  //                - number of events (in units of 1000) contributing to the ion density. 0 = 130000 events, 2 = 2000 events, 4 = 4000 events, etc.
+  //                - expected default map with 8000 pileup events: inputSCDensity3D8
```
martenole: This is probably leftover from a different macro?

GPU/TPCFastTransformation/macro/TPCFastTransformInit.C
```diff
@@ -0,0 +1,159 @@
...
+
+    auto x = trackResiduals.getX(xBin);             // radius of the pad row
+    auto y2x = trackResiduals.getY2X(xBin, y2xBin); // y/x coordinate of the bin ~-0.15 ... 0.15
+    auto z2x = trackResiduals.getZ2X(z2xBin);       // z/x coordinate of the bin 0.1 .. 0.9
+    float y = x * y2x;
+    float z = x * z2x;
+
+    float correctionX = v->D[o2::tpc::TrackResiduals::ResX];
+    float correctionY = v->D[o2::tpc::TrackResiduals::ResY];
+    float correctionZ = v->D[o2::tpc::TrackResiduals::ResZ];
```
martenole, line 92: Eventually I think we would use the smoothed residuals as input (`v->DS`). But in the example which I sent you these are not filled. I also need to clean up the `VoxRes` struct, which I simply copied from AliRoot. E.g. the member `DC` for the Chebyshev parametrization will probably go. So for now should be absolutely fine to use `v->D` since its toy data only anyways.

GPU/TPCFastTransformation/macro/TPCFastTransformInit.C
```diff
@@ -0,0 +1,159 @@
...
+      sumDiff[i] += d[i] * d[i];
+    }
+    nDiff++;
+  }
+  for (int i = 0; i < 3; i++) {
+    sumDiff[i] = sqrt(sumDiff[i]) / nDiff;
+  }
+
+  std::cout << "Max difference in x,y,z :  " << maxDiff[0] << " " << maxDiff[1] << " " << maxDiff[2] << endl;
+  std::cout << "Mean difference in x,y,z : " << sumDiff[0] << " " << sumDiff[1] << " " << sumDiff[2] << endl;
```
martenole, line 152: Even with the toy data this gives tiny mean differences and also very small max differences. It looks very good, thanks a lot!

martenole (dismissed):
Hi @sgorbuno,
this looks very good, thank you for preparing this. The results with toy data are already very promising.
Cheers,
Ole

GPU/TPCFastTransformation/CMakeLists.txt
sgorbuno: ok, done

GPU/TPCFastTransformation/macro/TPCFastTransformInit.C
sgorbuno: yes, i fixed it

PR #9157, opened 2022-06-13
TRD update trigger record message logic

martenole:
Important is the last commit 9a0d4413c39e646023323b0446275093da51d64d. With that I changed the logic of the trigger record messages. Now all trigger record messages are of the type `TRKTRGRD`. I would have liked it better to change it to `TRIGREC` everywhere, but this would then require a parallel update of QC which I'd like to avoid.
This allows us to move the downscaling of digits into the TRAP simulator. So this fixes https://alice.its.cern.ch/jira/browse/O2-2977. The idea is to simply use different subspecs. In case both the TRD digitization and the TRAP simulator are contained in a workflow the digitizer sends its digits and trigger records with subspec 1 and the TRAP simulator performs the downscaling of the digitst and sends the digits with subspec 0. The digit writer takes the messages with subspec 0 and thus avoids writing the large vector with all digits.
As a bonus when we attach the `o2-trd-digittracklet-writer` to the CTF reader workflow both the digits and the tracklets file will contain the trigger records and not only the trdtracklets.root file as it was the case before.

### DIFF
--- a/Detectors/TRD/qc/include/TRDQC/Tracking.h
+++ b/Detectors/TRD/qc/include/TRDQC/Tracking.h
@@ -63,13 +63,16 @@ struct TrackQC {
   std::array<float, constants::NLAYER> trackEta{}; //< Eta value of Track
 
   // tracklet position is also given in sector coordinates
-  std::array<float, constants::NLAYER> trackletYraw{}; ///< y-position of tracklet without tilt correction
-  std::array<float, constants::NLAYER> trackletZraw{}; ///< z-position of tracklet without tilt correction
-  std::array<float, constants::NLAYER> trackletY{};    ///< y-position of tracklet used for track update (including correction)
-  std::array<float, constants::NLAYER> trackletZ{};    ///< z-position of tracklet used for track update (including correction)
-  std::array<float, constants::NLAYER> trackletDy{};   ///< tracklet deflection over drift length obtained from CalibratedTracklet
-  std::array<int, constants::NLAYER> trackletSlope{};  ///< the raw slope from Tracklet64 (signed integer)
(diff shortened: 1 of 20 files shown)


bazinski:
@martenole maybe change the default for trap2raw to be all digits, to avoid mistakes of digits every 1 in 10k.

martenole:
> @martenole maybe change the default for trap2raw to be all digits, to avoid mistakes of digits every 1 in 10k.

Done (would be only 1 in 1M actually ;) )

davidrohr:
@TimoWilken @ktf @costing : The CI is again failing due to missing alien credentials.
Could someone have a look?
```
[INFO] CCDB: Reading o2::cpv::CalibParams from CPV/Calib/Gains
Error in <TJAlienConnectionManager>: Failed to get any credentials
Error in <TJAlienFile::Open>: No JAliEn Grid connection available!
[INFO] ccdb reads http://alice-ccdb.cern.ch/CPV/Calib/Gains/1627542202422/fdbfdde0-5441-11ec-8357-200114580202 for 1655188109798 (retrieve, agent_id: alientest06-1655188109-FjlTE2), 
[INFO] CCDB: Reading o2::cpv::Pedestals from CPV/Calib/Pedestals
Error in <TJAlienFile::Open>: No JAliEn Grid connection available!
[FATAL] Got nullptr from CCDB for path CPV/Calib/Pedestals and timestamp 1655188109798
```

ktf:
The issue seems to be CVMFS not working on the machine, with subsequent lack of CAs. Rebooting in progress.

martenole:
This would be ready to be merged

PR #9159, opened 2022-06-13
ITS/MFT decoding: fix race condition in PHY trigger decoding

shahor02:


shahor02:
tested locally, current code is broken, merging

PR #9164, opened 2022-06-13
DPL GUI: speedup drawing of many rects

ktf:


ktf:
@davidrohr could you cross check this works also on Linux with OpenGL?

davidrohr:
works on Linux for me

ktf:
Not tested in CI. Works on macOS and Linux, merging.

PR #9168, opened 2022-06-13
DPL: error message when FLP/DISTSTF used without /0.

ktf:


ktf:
@davidrohr as discussed.

PR #9171, opened 2022-06-13
HMPID condition paramters calibration using DCS DPs

gvolpe79:


Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
```diff
@@ -0,0 +1,229 @@
...
+    auto timeNow = HighResClock::now();
+
+    // process datapoints:
+    auto dps = pc.inputs().get<gsl::span<DPCOM>>("input");
+    mProcessor->process(dps);
+    if (startValidity == 0xffffffffffffffff) {                                                                   // it means it is not set
+      startValidity = std::chrono::duration_cast<std::chrono::milliseconds>(timeNow.time_since_epoch()).count(); // in ms
+    }
+
+    mProcessor->setStartValidity(startValidity);
```
shahor02: you are setting startvalidity each time you receive something, i.e. at least every 5 s (when the full DPs image is sent even if nothing was changed). Should not you use as a start validity the beginning of the integration period?

Detectors/HMPID/calibration/src/HMPIDDCSProcessor.cxx
```diff
@@ -0,0 +1,759 @@
...
+  }
+  LOG(info) << Form("No entries in HV for ch%isec%i", iCh, iSec);
+  return pHvTF;
+}
+
+// Process all arrays of DPCOM-vectors, perform fits
+// Called At the end of the Stream of DPs, in
+// HMPIDDCSDataProcessorSpec::endOfStream() function
+void HMPIDDCSProcessor::finalize()
+{
```
shahor02, line 693: As far as I can see, you create a CCDB object only in the endOfStream, but this means never, as the workflow is running permanently, w/o any correlation with data taking (SOR/EOR).
In case you need to perform some action at the EOR, you can use https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/DCS/include/DetectorsDCS/RunStatusChecker.h, see how EMCAL does this: https://github.com/AliceO2Group/AliceO2/blob/3d5a983fbadf879127fbde5d17e6a85b350c985d/Detectors/EMCAL/workflow/include/EMCALWorkflow/EMCALDCSDataProcessorSpec.h#L130-L146

Detectors/HMPID/calibration/include/HMPIDCalibration/HMPIDDCSProcessor.h
```diff
@@ -0,0 +1,409 @@
...
+  }
+
+  // get methods for time-ranges
+  // ===============================================================================
+  // const auto& getTimeQThresh() const { return mTimeQThresh; }
+  // const auto& getTimeArNmean() const { return mTimeArNmean; }
+
+  /// / return timestamp of first fetched datapoint for a given ID (Tin/Tout,
+  /// Environment pressure, HV, chamber pressure)
+  TimeStampType getMinTime(const std::vector<DPCOM> dps)
```
shahor02: pass by ref. instead of copy: `const std::vector<DPCOM>& dps`

Detectors/HMPID/calibration/include/HMPIDCalibration/HMPIDDCSProcessor.h
```diff
@@ -0,0 +1,409 @@
...
+    TimeStampType firstTime = std::numeric_limits<uint64_t>::max();
+    for (const auto& dp : dps) {
+      const auto time = dp.data.get_epoch_time();
+      firstTime = std::min(firstTime, time);
+    }
+    return firstTime;
+  }
+  // return timestamp of last fetched datapoint for a given ID (Tin/Tout,
+  // Environment pressure, HV, chamber pressure)
+  TimeStampType getMaxTime(const std::vector<DPCOM> dps)
```
shahor02: idem.

Detectors/HMPID/calibration/include/HMPIDCalibration/HMPIDDCSProcessor.h
```diff
@@ -0,0 +1,409 @@
...
+
+      // check if tme of DP is greater (i.e. later) than previously latest
+      // fetched DP:
+      const auto time = dp.data.get_epoch_time();
+      lastTime = std::max(lastTime, time);
+    }
+    return lastTime;
+  }
+
+  void checkEntries(const std::vector<TF1> arQthresh,
```
shahor02: idem.

Detectors/HMPID/calibration/include/HMPIDCalibration/HMPIDDCSProcessor.h
```diff
@@ -0,0 +1,409 @@
...
+    bool arQthreFull = true;
+
+    LOG(info) << " ";
+    LOG(info) << "======================================== ";
+    LOG(info) << "All entries Processed";
+    LOG(info) << " checking if CCDB objects are filled : ";
+
+    for (int iCh = 0; iCh < 7; ++iCh) {
+      for (int iSec = 0; iSec < 6; ++iSec) {
+        auto tf = arQthresh.at(6 * iCh + iSec);
```
shahor02: use `[i]` instead of `.at(i)`, also below

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
```diff
@@ -0,0 +1,304 @@
...
+    "HMP_TRANPLANT_MEASURE_[0..29]_C6F14CELL"
+    "HMP_RUNNUMBER"};
+
+  bool isRunStarted = false;
+  bool mCheckRunStartStop = false;
+  o2::dcs::RunStatusChecker mRunChecker{o2::detectors::DetID::getMask("HMP")};
+
+  std::unique_ptr<HMPIDDCSProcessor> mProcessor;
+  HighResClock::time_point mTimer;
+  int64_t mDPsUpdateInterval;
```
shahor02: I don't see where you set or use update interval. Please avoid very frequent updates of CCDB at every change. See how other detectors, e.g. FIT/FT0/calibration/testWorkflow/FT0DCSDataProcessorSpec.h do

shahor02 (changes requested):

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
gvolpe79: Hi @shahor02, variable mDPsUpdateInterval is not used indeed, is there by mistake, we are going to remove it! We update CCDB just at the end of RUN (see endOfStream method). I think this is correct! At least if I have not misunderstood something!

shahor02: Hi @gvolpe79 
The endOfStream for the DP processor is rather meaningless: the DP processing workflow runs continuously and is not synchronized with SOR/EOR. The only way to detect run start/stop is to use `mRunChecker.getRunStatus()` as you do and send run-related output from the run(..) method. That said, note that it is not 100% reliable: if the AliECS crashes, the EOR will be never flagged in the CCDB. @knopers8 works on fixing this but the ETA is not clear.

Detectors/HMPID/calibration/src/HMPIDDCSProcessor.cxx
```diff
@@ -0,0 +1,803 @@
...
+
+  std::unique_ptr<TF1> pEnv = finalizeEnvPressure();
+  for (int iCh = 0; iCh < 7; iCh++) {
+
+    std::unique_ptr<TF1> pChPres = finalizeChPressure(iCh);
```
shahor02, line 698: What is the role of these histos? They are created and deleted when going out of scope w/o being used

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
```diff
@@ -0,0 +1,315 @@
...
+    ConcreteDataTypeMatcher{o2::calibration::Utils::gDataOriginCDBWrapper,
+                            "RefIndex"},
+    Lifetime::Sporadic);
+
+  return o2::framework::DataProcessorSpec{
+    "hmp-dcs-data-processor", Inputs{{"input", "DCS", "HMPDATAPOINTS"}},
+    outputs, AlgorithmSpec{adaptFromTask<o2::hmpid::HMPIDDCSDataProcessor>()},
+    Options{{"ccdb-path",
+             VariantType::String,
+             "localhost:8080",
```
shahor02: please use `o2::base::NameConf::getCCDBServer()` as a default.

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
```diff
@@ -0,0 +1,315 @@
...
+             {"Path to CCDB"}},
+            {"use-ccdb-to-configure",
+             VariantType::Bool,
+             false,
+             {"Use CCDB to configure"}},
+            {"follow-hmpid-run",
+             VariantType::Bool,
+             false,
```
shahor02, line 312: if I get it right, with default setting (no run following) you will not produce any output? Is this what you want?

Detectors/HMPID/calibration/src/HMPIDDCSProcessor.cxx
gvolpe79: Indeed, are not in the CCDB but used to create the objects (charge threshold) to be stored in the CCDB (see lines 717 and 718 ) and then deleted.

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
gvolpe79: Do you mean to replace the local host?

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
gvolpe79: Yes, this is what we want.

Detectors/HMPID/calibration/src/HMPIDDCSProcessor.cxx
shahor02: I see that you are using the name only, why do you need to create a histo?

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
shahor02: Yes, the default server should be `o2::base::NameConf::getCCDBServer()`, so it can be globally changed.

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
shahor02: ok, but then it is not clear why this is an option, moreover set to false by default ?

Detectors/HMPID/calibration/src/HMPIDDCSProcessor.cxx
gvolpe79: They are not an histos, they are TF1 objects coming from the fit of histos.

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
gvolpe79: ok, we are going to modify the line

Detectors/HMPID/calibration/testWorkflow/HMPIDDCSDataProcessorSpec.h
gvolpe79: You are right, the idea was to optimize the code later on! Anyway, since it is simple to modify it, we are doing to do it now.

Detectors/HMPID/calibration/src/HMPIDDCSProcessor.cxx
shahor02: OK, still don't fully understand the logic, but given that it is done once per run at finalization, the overhead should not be important.

shahor02 (approved):

PR #9173, opened 2022-06-14
support structure without overlaps

AllaMaevskaya:


sawenzel:
Given that ~1800 lines are changed here, would you be able to give some more information on the commit?
Is this introducing additional material or are you fixing overlaps in the existing geometry?

Detectors/FIT/FT0/base/src/Geometry.cxx
```diff
@@ -1,122 +1,126 @@
...
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+#include "FT0Base/Geometry.h"
+#include "TSystem.h"
+#include "Framework/Logger.h"
+#include <sstream>
+#include <iomanip>
+#include <string>
+#include <iostream>
```
sawenzel, line 18: I don't understand why these lines appear as "changed".

sawenzel: You have probably introduced additional white space characters. Would you be able to fix this and provide a cleaned up version?

AllaMaevskaya:
Dear Sandro,

detector geometry was totally re-written by out CalPoly colleges. I asked Jennifer to take cake about their code.

Alla

Detectors/FIT/FT0/base/src/Geometry.cxx
dedx: I am not sure why it triggered these lines as "changed".  I looked into the file and they do not appear to have additional white space.

dedx:
This update is a major overhaul of the geometry code for the FT0-A aluminum frame, which was re-written to eliminate overlaps, and new code for the FT0-C aluminum frame was added.

AllaMaevskaya:
Jennifer, could you have a look why build crashed?

dedx:
Hi all.  The fix is relatively simple.  On lines 727, 728, 847, 857, and 867 of Detector.cxx, the variables should be initialized as "static constexpr".  When I made that change and recompiled with tests ON, 100% passed.

sawenzel:
@dedx : The fix should be provided as part of this PR.

mslupeck:
Since Alla is away for a longer time, we have decided to prepare another PR to fix the minor errors present here. PR #9406 replaces this one.

PR #9176, opened 2022-06-14
[WIP] TRD fix XOR handling of tracklet word

martenole:
With this change the Tracklet64 class is expected to always store the position and slope without any inverted bits. The XOR takes place in 2 places now:
1) When the raw reader creates the Tracklet64 which are stored in the CTFs it applies the XOR
2) Within the Sim->Raw conversion used for creating synthetic data sets the XOR is applied to the simulated Tracklet64 before the the raw data is created.

That way we don't need to worry at all inside reconstruction, calibration or QC about flipping bits.

I have added an option to the tracklet transformer to apply the XOR. This would only be needed for reprocessing the CTFs from the pilot beam in 2021. Any data taken after cb42172601cd4bc43764894cb457eec837039b28 was merged in January will have the bit flip corrected position and slope stored in the Tracklet64


martenole:
Moving to WIP for now, since my tests are failing at the moment

PR #9178, opened 2022-06-14
STFDecoder fetches ITS/MFT Alpide params from CCDB

shahor02:


shahor02:
FullCI failure unrelated, other CI's were already green, tested locally, merging.

PR #9179, opened 2022-06-14
Selective CTF compression

shahor02:


shahor02:
Don't merge yet, even if passed CI

PR #9184, opened 2022-06-15
TPCFastTransform: implementing possibility to set ref. corrections (CPU only)

matthias-kleiner:
Make SpacheCharge independent from TPCFastTransform to avoid cyclic
dependency

Move omp includes from Tricubic interpolator to src to avoid problems

The reference correction can be used in simulations to study the impact of the uncertainty of the splines in the TPCFastTransform object on the track parameters


wiechula (dismissed):

wiechula:
@shahor02 @davidrohr , the failure seems unrelated. Can this be merged?

davidrohr:
The CI problem is fixed, let's wait for it to rerun

wiechula:
Ok, looks good now.

GPU/TPCFastTransformation/TPCFastTransform.h
```diff
@@ -253,6 +272,14 @@ class TPCFastTransform : public FlatObject
   float mTOFcorr;
 
   float mPrimVtxZ; ///< Z of the primary vertex, needed for the Time-Of-Flight correction
+
+#if !defined(GPUCA_GPUCODE) && !defined(GPUCA_STANDALONE)
+  /// Note: if this is defined above, it will give errors during the GPU reconstruction?!
+  /// Correction of (x,u,v) with tricubic interpolator on a regular grid
+  std::unique_ptr<o2::tpc::SpaceCharge<float>> mCorrectionSlow; ///<! reference space charge corrections
+  bool mUseSlowCorrection{false};                               ///<! flag whether to use the slow correction
+#endif
```
davidrohr: This breaks GPU processing, since it relies on the fact that memory layouts on host and GPU are identical.
You must not have data members that exist only on the GPU.
The GPU should not use bool in shared data members, you can use char instead.
I also don't like that the flat object has a unique ptr.

GPU/TPCFastTransformation/TPCFastTransform.h
```diff
@@ -370,7 +397,35 @@ GPUdi() void TPCFastTransform::Transform(int slice, int row, float pad, float ti
...
   if (mApplyCorrection) {
     float dx, du, dv;
-    mCorrection.getCorrection(slice, row, u, v, dx, du, dv);
+
+#if !defined(GPUCA_GPUCODE) && !defined(GPUCA_STANDALONE)
+    if (!mUseSlowCorrection) {
+#endif
+      mCorrection.getCorrection(slice, row, u, v, dx, du, dv);
+#if !defined(GPUCA_GPUCODE) && !defined(GPUCA_STANDALONE)
+    } else {
```
davidrohr: If possible, could we reduce the usage of ifdefs a bit. E.g., the mUseSlowCorrection can be a normal member, then you can also just use it on the GPU.

davidrohr (changes requested):

matthias-kleiner:
thanks for the comments @davidrohr. I just made some small modifications which should fix the issues. Is it ok like this?

GPU/TPCFastTransformation/TPCFastTransform.h
```diff
@@ -78,8 +101,17 @@ class TPCFastTransform : public FlatObject
...
   TPCFastTransform& operator=(const TPCFastTransform&) CON_DELETE;
 
-  /// Destructor
+/// Destructor
+#if !defined(GPUCA_GPUCODE) && !defined(GPUCA_STANDALONE)
+  ~TPCFastTransform()
+  {
+    if (mCorrectionSlow) {
+      delete mCorrectionSlow;
+    }
```
davidrohr: delete checks for a nullptr and does nothing on it, you don't need to check yourself

matthias-kleiner: Thanks, I did not know that.

PR #9185, opened 2022-06-15
[MCH] new option to exclude corners when looking for fired neighbouring pads

pillot:
The option can be enabled through configurable parameters. It is disabled by default, leaving the behavior of the preclustering unchanged.

The preclustering spec and workflow have also been moved to the PreClustering directory to continue the cleanup of the Workflow directory.

The README have been updated accordingly.


aphecetche (approved):

pillot:
@aphecetche checks are green, you can merge now :-)

PR #9187, opened 2022-06-15
TRD fix time bin ordering in sim to raw

bazinski:
- fixes the ordering of the timebins in sim to raw.
- the digits are written in z y x z y x, not in x y z x y z x y z order.


martenole (approved):
Tested with FST and indeed tracklets and digits are matching now :+1:

PR #9188, opened 2022-06-15
TRD XOR cleanup

martenole:
By adding the bit flipping to the constructor of Tracklet64 and the getters for position and slope in January I created a big mess. This way the two bits were flipped multiple times: in the raw reader, in the CTF encoder and in the CTF decoder.
This PR fixes this mess by applying the bit shifts explicitly in the only two places where it is needed: when we read the raw data and when we convert simulated data into raw data for synthetic running. This way the position and slope stored in the Tracklet64 class can always be used without any additional bit manipulation.

Since now the bits are actually inverted in all CTFs taken so far I have added one commit which adds an option to the tracklet transformer to apply the bit flipping again. I tested this with the last data taken with beam and it works for me.

Just to explain a bit the mess with the current state of the code before this PR: the raw reader flips the bits so they are actually OK. Our QC plots show `Tracklet64::getPosition()` and `Tracklet64::getSlope` which flip the bits again resulting in funny looking plots although the data is actually OK. Then the CTF encoder flips the bits again storing the data XORed in the CTFs. The CTF decoder does another XOR, now position and slope are OK again, but since we apply the XOR in the getters we see again wrong data for slope and position. Clearly I hadn't thought #7898 through completely

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
@@ -44,9 +44,10 @@ Word 0  |   Format  |              HCID              |  padrow   | col |        
 Word 0  |  slope                |    Q2                 |    Q1                 |         Q0            |
         -------------------------------------------------------------------------------------------------
 
-        Note: In the tracklet word both the position and the slope have one bit inverted. This avoids the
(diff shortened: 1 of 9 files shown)


bazinski (approved):
Just to say it the nth time. Tracklet64 now stores the *actual* position and slope, in its raw units from the electronics.

PR #9191, opened 2022-06-15
FV0 trigger bits 

jotwinow:


jotwinow:
@AllaMaevskaya @mslupeck could you approve it

AllaMaevskaya (approved):

afurs:
@jotwinow could you confirm that FV0 trg bits have the same positions in raw data? this is important, otherwise decoding for FV0 trg bits will be wrong

jotwinow:
> @jotwinow could you confirm that FV0 trg bits have the same positions in raw data? this is important, otherwise decoding for FV0 trg bits will be wrong

Yes, they are properly set @afurs

PR #9195, opened 2022-06-16
[EMCAL-566] Added option to time-profile calibration

jokonig:
- Added switch to obtain the time spend in the run function of the time
and bad channel calib. Needed because very high CPU consumption was
observed in online tests at point2 which cannot be reproduced
offline/locally. If option enabled (via EMCALCalibParams), time in ns is
printed out. Only for debug purposes
- Added options to set slot length and restriction to trigger
calibration only at end of run


mfasDa (approved):
Looks good. @shahor02 could you please merge?

PR #9196, opened 2022-06-16
[O2-2213] Allow command line options to be specified multiple times

matthiasrichter:
The last specified option will make the race. This implementation is
using `unordered_map` to assemble unique command line options, and later
occurrences will simply overwrite earlier ones.

This makes it also possible to set some options globally and override
for dedicated processors, e.g.
   `--severity error --producer '--severity debug'`


davidrohr:
Thx, this looks good, but haven't really tested it yet. Let's see if it passes the CI, then I'll experiment a bit with overriding options for the FST on the EPN.

davidrohr:
@matthiasrichter : This fails the ctests for DDS and for OCC export. COuld you have a look.
The failure in the fullCI should be unrelated.

ktf:
We should also check this on hyperloop as well before merging. I was thinking that using a vector of pairs and retaining the order of the options might be a good idea (and I suspect that's why some of the tests are failing).

davidrohr:
@matthiasrichter : I have tried with this PR, and indeed the overriding (e.g. `--severity info --tpc-tracker '--severity error'`) works correctly, but specifying an option twice fails.
```
[O2PDPSuite/latest] ~/alitest/tmp-pp-fst01 $> o2-dpl-run -b -b
[ERROR] error parsing options of o2-dpl-run: option '--batch' cannot be specified more than once
```
Could you have a look?

davidrohr:
and on another note: I have seen that you have added some magic for the channel config, such that if it is provided multiple times it gets appended and not overridden.
Perhaps we would want that behavior as well for selected other cases, particularly the `--[device_name] "device setting"`. Could this be added easily?

matthiasrichter:
ok, I'm going to check if there is yet another specific way of assembling the options.

matthiasrichter:
Some of the unit test had to be adjusted because the sequence of options changed. Looks a bit better now, but I did not catch all places.

Regarding the additional parsing of the channel-config option this was added earlier by @shahor02. We can think about some generalization for this.

I just added a comment because I think that there is an inconsistency when adding this to `control.options`, so far I have not checked what the options added to the `control` object are used for.

matthiasrichter:
for the remaining CI failures, it looks like `unordered_map` on arm produces a different sequence and this makes the unit tests still fail on those two platforms.

ktf:
You mean on macOS? That's possible. unordered_map is not required to be implemented as an hashmap and in particular I know (because it was causing issues in some FIT code) that the macOS on ARM implementation seems to use buckets and uses the equality operator regardless of wether or not the hashing is perfect. As I said I would simply rely on sorted vectors.

matthiasrichter:
Yes, sorry, mac os. I'm simply storing the sequence of option keys in addition now to force deterministic behavior.

Now `test_Framework_test_SingleDataSource` is failing on macos, but that seems to be unrelated.
```
[3686:A]: [11:13:51][INFO] Correctly handshaken websocket connection.
[INFO] Quitting
[INFO] QUIT_REQUESTED
[ERROR] pid 3687 (internal-dpl-injected-dummy-sink) crashed with 1
[3687:internal-dpl-injected-dummy-sink]: Assertion failed: (info.parts.Size() > 0), function handleData, file DataProcessingDevice.cxx, line 1171.
[3687:internal-dpl-injected-dummy-sink]: *** Segmentation fault (O2)
[3687:internal-dpl-injected-dummy-sink]: Backtrace:
```

ktf:
I have seen that also elsewhere, it looks like something spurious, because it then disappear. I still need to investigate.

davidrohr:
working correctly, and now passing also all ctests.
The `o2-dpl-run -b -b` still fails, and I would still like if one could use `--[process_name]` twice and it would be concatenated. But can be followed up in a separate PR.

PR #9201, opened 2022-06-16
o2-eve-export-workflow: added EMCAL export

jmyrcha:


jmyrcha:
fullCI failed because of internal errors - not related to the code

shahor02:
Failure is unrelated

PR #9204, opened 2022-06-16
[O2-2679] Fix bug in TPCITS Afterburner

mconcas:
Bug was introduced in #8908 by me.
It affects also: https://alice.its.cern.ch/jira/browse/O2-2679

@shahor02


shahor02 (approved):
Thanks! Merging, since trivial fix restores old behaviour.

PR #9206, opened 2022-06-16
Fix in vertex-track association

shahor02:
If checked non-contributor track time-bracket was above vertex time-bracket
by good margin, such a track could be declared as an orphan instead of comparing
with the next vertex.


chiarazampolli:
Can this be merged? Tests passed.

PR #9214, opened 2022-06-17
MCH: select only non empty IRFrames

aphecetche:


shahor02 (approved):
Thanks, was just going to write about that.

pillot (approved):

PR #9216, opened 2022-06-18
fix for logs and consts

noferini:


noferini:
Can we merge this?
It should be proted in the next flp suite

noferini:
Hi @shahor02 ,
can we merge this PR (yesterday all checks were green)?
It seems it is quite urgent for flp-suite.
Cheers
Francesco

shahor02 (approved):

PR #9217, opened 2022-06-18
DPL: improve forward handling

ktf:
* handle forwarding in FairMQDeviceProxy
* refactor forwardign itself so that it can be done via a free
  function.

Should simplify the forwarding code and allow it to be done
already at the level of DataRelayer, so that we can handle
the issue with forwarding of obsolete timeframes.


ktf:
New version passes all tests on my box.

davidrohr:
@ktf : what is actually with this PR? Do we still need it / want it?

ktf:
It's actually included in the other PR for the fix-calib.

davidrohr:
closing as superseded by #9395

PR #9218, opened 2022-06-19
ctpdev: two more params to ctp-proxy

lietava:


lietava:
@shahor02 : Hi Ruben, it appears it would be useful to add two more parameters to https://github.com/lietava/AliceO2/blob/ctpdev/Detectors/CTP/workflowScalers/src/ctp-proxy.cxx:
1.) CCDBHost
2.) switch ( I called it mQC which) which would modify functionality of RunManager to be used in two different context: as now and for QC  without database operations ( either through parameter or I define base class runmanager)

shahor02:
@lietava but waht CTPConfiguration has with ccdb host?

lietava:
@shahor02 : CTPRunManager is in files Configuration files.
run manager follows runs and  save run configuration and run counters in CCDB. For that reason it would be good to have option to specify CCDB.
Then CTP run counters would be monitored in o2 standard QC. There the very similar functionality for run managment counters is required. Therefore we want to make ctp-proxy universal for 1. CTP config and counters usage (which is already there) and for 2.) QC counters monitoring.
@bombara

shahor02:
@lietava OK, I thought it is in the config class. Note that your PR is still in the draft mode, if it is ready, please press on "Ready for review"

lietava:
@shahor02 : I went through build/O2/o2 log but can  not find any issue related to ctp ?

shahor02 (approved):

PR #9220, opened 2022-06-20
Use GRPGeomHelper to fetch GRP and Geometry related stuff from CCDB

shahor02:
This PP is for the not yet final transition of the workflows from local-file based geom/grp/field to fetching from CCDB using GRPGeomHelper.
I could validate only the workflows which are in the FST and sim_challenge. So I would ask authors to check their code and if possible, validate it.
Some specific comments below.

@noferini : in the workflows below you are extracting the bunchfilling from the dig.context directly in the defineDataProcessing:
```
GlobalTrackingWorkflow/src/tof-eventtime-checker-workflow.cxx:    auto context = mcReader->getDigitizationContext();
GlobalTrackingWorkflow/src/tof-matcher-workflow.cxx:    auto context = mcReader->getDigitizationContext();
```
Calib objects should be set per device, in this case I am even not sure for which device the bunch filling is set. 
Could you remove it from the defineDataProcessing and set in the relevant devices like I did in the  Detectors/TOF/workflow/src/TOFClusterizerSpec.cxx in this PR
pendentParam

TOFEventTimeChecker: I did not find why it loads geometry, GRP and inits field, so I've suppressed them.

@mfasDa :
EMCAL AnalysisClusterSpec : uses hardcoded Geometry::GetInstanceFromRunNumber(223409)

@chiarazampolli : I did not modify MatchITSTPCQC since loads geometry from file not in the DPL device, also called from QualityControl/Modules/GLO/src/ITSTPCMatchingTask.cxx which does not support GRPGeomHelper, we need to fix QC first

@mconcas in the ITS TrackerSpec the matLUT is always provided by the CCDB, so the switch for the mat. corrType depending on the presence of the file does not make sense anymore, hence eliminated. If you want, you can add a switch via configurable params.

@aphecetche : MCH ClusterTransformerTask loading geometry might be explicitly requested from the json file or from TGeo, not sure how to handle it. In any case, as far as I understand, you started the migration to CCDB initialization, so I will not touch MCH part.

### DIFF
--- a/Detectors/MUON/MID/Workflow/src/TrackerSpec.cxx
+++ b/Detectors/MUON/MID/Workflow/src/TrackerSpec.cxx
@@ -32,6 +32,7 @@
 #include "MIDTracking/Tracker.h"
 #include "MIDSimulation/TrackLabeler.h"
 #include "CommonUtils/NameConf.h"
+#include "DetectorsBase/GRPGeomHelper.h"
 
 namespace of = o2::framework;
 
@@ -42,26 +43,13 @@ namespace mid
 class TrackerDeviceDPL
 {
  public:
-  TrackerDeviceDPL(bool isMC) : mIsMC(isMC) {}
+  TrackerDeviceDPL(std::shared_ptr<o2::base::GRPGeomRequest> gr, bool isMC) : mGGCCDBRequest(gr), mIsMC(isMC) {}
   ~TrackerDeviceDPL() = default;
 
   void init(o2::framework::InitContext& ic)
   {
-
-    if (!gGeoManager) {
-      o2::base::GeometryManager::loadGeometry();
-    }
-
-    auto keepAll = !ic.options().get<bool>("mid-tracker-keep-best");
-
-    auto geoTrans = createTransformationFromManager(gGeoManager);
-
-    mTracker = std::make_unique<Tracker>(geoTrans);
-    if (!mTracker->init(keepAll)) {
-      LOG(error) << "Initialization of MID tracker device failed";
-    }
-
-    mHitMapBuilder = std::make_unique<HitMapBuilder>(geoTrans);
+    o2::base::GRPGeomHelper::instance().setRequest(mGGCCDBRequest);
+    mKeepAll = !ic.options().get<bool>("mid-tracker-keep-best");
 
     auto stop = [this]() {
       LOG(info) << "Capacities: ROFRecords: " << mTracker->getTrackROFRecords().capacity() << "  tracks: " << mTracker->getTracks().capacity() << "  clusters: " << mTracker->getClusters().capacity();
@@ -74,6 +62,7 @@ class TrackerDeviceDPL
   void run(o2::framework::ProcessingContext& pc)
   {
     auto tStart = std::chrono::high_resolution_clock::now();
+    updateTimeDependentParams(pc);
 
     auto clusters = pc.inputs().get<gsl::span<Cluster>>("mid_clusters");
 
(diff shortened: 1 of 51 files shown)


shahor02:
@TimoWilken the fullCI fatals due to the alien-token, could you please check?

aphecetche:
> @aphecetche : MCH ClusterTransformerTask loading geometry might be explicitly requested from the json file or from TGeo, not sure how to handle it. 

A priori there's nothing to handle :wink: :  the option is there only for the cases where we use the cluster transformer as a separate workflow, it is _not_ used in the regular `o2-mch-reco-workflow`.

shahor02:
> A priori there's nothing to handle wink : the option is there only for the cases where we use the cluster transformer as a separate workflow, it is _not_ used in the regular `o2-mch-reco-workflow`.

you mean in this case the geometry does not need to be loaded? Anyway, are modifying MCH to use GRPGeomHelper or I should add it to this PR?

Detectors/MUON/MID/Workflow/src/TrackerSpec.cxx
```diff
@@ -111,9 +100,30 @@ class TrackerDeviceDPL
...
+    o2::base::GRPGeomHelper::instance().checkUpdates(pc);
+
+    auto geoTrans = createTransformationFromManager(gGeoManager);
+    mTracker = std::make_unique<Tracker>(geoTrans);
+    if (!mTracker->init(mKeepAll)) {
+      LOG(error) << "Initialization of MID tracker device failed";
+    }
+    mHitMapBuilder = std::make_unique<HitMapBuilder>(geoTrans);
+  }
+
```
pillot, line 125: Why not (re)initializing of `mTracker` and `mHitMapBuilder` in `finaliseCCDB()`, so that it is done only when the geometry is updated?

shahor02: @pillot thanks for spotting, actually, I have missed there the usual clause 
```
static bool initOnceDone = false;
if (!initOnceDone) {
  initOnceDone = true;
  // do one-time initializations
}
```
I prefer not to do detector initializations in the finaliseCCDB since sometimes this requires (or may require in future) multiple ccdb objects while the finaliseCCDB is called per object. 
Will fix this

aphecetche:
> > A priori there's nothing to handle wink : the option is there only for the cases where we use the cluster transformer as a separate workflow, it is _not_ used in the regular `o2-mch-reco-workflow`.
> 
> you mean in this case the geometry does not need to be loaded? Anyway, are modifying MCH to use GRPGeomHelper or I should add it to this PR?

When `--mch-disable-geometry-from-ccdb` is used in `o2-mch-cluster-transformer-workflow` (and only in this case, and only for this workflow) the geometry is loaded from file. In all other cases it's loaded from CCDB already. Or do I miss something here ?

shahor02:
> When `--mch-disable-geometry-from-ccdb` is used in `o2-mch-cluster-transformer-workflow` (and only in this case, and only for this workflow) the geometry is loaded from file. In all other cases it's loaded from CCDB already. Or do I miss something here ?

Ah, I did not notice that you've already implemented reading from the CCDB, just reacted `o2::base::GeometryManager::loadGeometry(geoFile)`.
But now I see possible potential problem: you are calling `transformation = o2::mch::geo::transformationFromTGeoManager(*gGeoManager)` in every call of finaliseCCDB(). If tomorrow you add another calibration object, the `finaliseCCDB` will be called also for it, again triggering `transformation` assignment.
That's why I do https://github.com/AliceO2Group/AliceO2/pull/9220#discussion_r901721306

What about  `Workflow/src/TrackAtVertexSpec.cxx`, which also loads geometry from file?

shahor02:
don't merge yet, needs to be matched to some developments on MC side.

aphecetche:
> Ah, I did not notice that you've already implemented reading from the CCDB, just reacted `o2::base::GeometryManager::loadGeometry(geoFile)`. But now I see possible potential problem: you are calling `transformation = o2::mch::geo::transformationFromTGeoManager(*gGeoManager)` in every call of finaliseCCDB(). If tomorrow you add another calibration object, the `finaliseCCDB` will be called also for it, again triggering `transformation` assignment. That's why I do [#9220 (comment)](https://github.com/AliceO2Group/AliceO2/pull/9220#discussion_r901721306)

@shahor02 ok, good point. But for the moment we are safe in the cluster-transformer as we only have one calib object. But noted for a future upgrade of that class : we've been discussing this with Philippe and we'll remove the option to read from a file altogether and work from a custom CCDB for the dev cases covered by the reading from file (I've added a ticket in order not the forget  [MRRTF-187](https://alice.its.cern.ch/jira/browse/MRRTF-187)
 
> What about `Workflow/src/TrackAtVertexSpec.cxx`, which also loads geometry from file?

Better to leave that one alone for the moment. It's not used in the regular reco and is only used by experts.

noferini:
Hi @shahor02 
compilation fails because I forgot an include in my patch
#include "TOFBase/Utils.h"
It should be added to Detectors/GlobalTrackingWorkflow/src/TOFMatcherSpec.cxx

shahor02:
tests were passed, merging after rebase

davidrohr:
@shahor02 : does this now need any change for SYNTHETIC runs / FST tests? Do I need to resimulate something, to be sure geom and GRPs are in sync?

shahor02:
@davidrohr At the moment there is no need for any change. Most of paths settings in the dpl-worklow will be simply ignored but the TPC reco and QC still need aux. files.

After this PR we will not be able to cheat synthetic run using `now` timestamp by relinking geometry and collision context, instead, we will need to use --condition-remap for geom and GRP to fetch MC-specific objects from alternative (user) space on CCDB (available also on internal).

davidrohr:
ok, but I assume that we have objects uploaded already for the unanchored MC run ranges. So once we can override the time, this will work out of the box?

shahor02:
Yes, with correct timestamp everything should be fetched automatically w/o any hacks.

PR #9230, opened 2022-06-21
TPC: Adjust DCS upload times for CCDB

wiechula:


shahor02 (approved):

shahor02:
@wiechula I've restarted the dcs-dpc on flp199 with your fix, could you check if now the uploads are as expected?

wiechula:
@shahor02 , yes looks good!

PR #9234, opened 2022-06-21
TRD correctly shift the trigger bc time to take account of the delay

bazinski:
Correctly shift the trigger bunch crossing according to https://alice.its.cern.ch/jira/browse/CTP-133

### DIFF
--- a/Detectors/TRD/reconstruction/src/CruRawReader.cxx
+++ b/Detectors/TRD/reconstruction/src/CruRawReader.cxx
@@ -31,12 +31,12 @@
 #include "Framework/DataSpecUtils.h"
 #include "Framework/Output.h"
 #include "Framework/InputRecordWalker.h"
+#include "DataFormatsCTP/TriggerOffsetsParam.h"
 
 #include <cstring>
 #include <string>
 #include <vector>
 #include <array>
-#include <iostream>
 #include <numeric>
 
 namespace o2::trd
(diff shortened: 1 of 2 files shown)


Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -484,6 +484,11 @@ int CruRawReader::processHalfCRU(int cruhbfstartoffset, int numberOfPreviousCRU)
 
   //get eventrecord for event we are looking at
   mIR.bc = mCurrentHalfCRUHeader.BunchCrossing; // correct mIR to have the physics trigger bunchcrossing *NOT* the heartbeat trigger bunch crossing.
+  //shift accordingly
+  mIR.bc = mCurrentHalfCRUHeader.BunchCrossing - o2::ctp::TriggerOffsetsParam::Instance().LM_L0;
+  if (mIR.bc < 0) {
+    mIR.bc = 0;
+  }
```
shahor02, line 491: Why setting BC to 0 ? The data corresponding to the BC which does not belong to TF should be simply discarded.

bazinski: good point, fixing.

shahor02: Also, the inverse (adding the bias) should be applied in the mc2raw.

bazinski: even better point ;-) I want to get the safety net fixed for the data decoder first, then will come back to this. I dont want to dump the entire dataframe rather only the halfcruheader in question, and with the codes current state its rather delicate.

shahor02 (approved):
Hi @bazinski 
merging as is since currently the TRD BCs are shifted in the reconstruction, but please add the changes discussed in the comments as additional PR (rejecting trigger data in the BC becomes negative after correction and emulating shift in the mc2raw).

PR #9236, opened 2022-06-21
Improve to avoid typos in DP names, possibility to keep latest value, change in startValidity

chiarazampolli:
@shahor02 
Now there are some enum's used. It is not the best probably, but the other option is to change the GRPLHCInfo struct to have a map of vectors where the key is the alias. The disadvantage is to lose direct access to the different information, while becoming closer to teh Collimators. What do yo think?

### DIFF
--- a/Detectors/GRP/calibration/src/GRPDCSDPsProcessor.cxx
+++ b/Detectors/GRP/calibration/src/GRPDCSDPsProcessor.cxx
@@ -32,6 +32,57 @@ void GRPDCSDPsProcessor::init(const std::vector<DPID>& pids)
     mPids[it] = false;
   }
   mMagFieldHelper.verbose = mVerbose;
+
+  // initializing vector of aliases for LHC IF DPs
+  int lastFilledElement = 0;
+  for (int i = 0; i < GRPLHCInfo::CollimatorAliases::NCollimatorAliases; ++i) {
+    mArrLHCAliases[i] = GRPLHCInfo::collimatorAliases[i];
+  }
+  lastFilledElement += GRPLHCInfo::CollimatorAliases::NCollimatorAliases;
+
+  for (int i = 0; i < GRPLHCInfo::BeamAliases::NBeamAliases; ++i) {
+    mArrLHCAliases[i + lastFilledElement] = GRPLHCInfo::beamAliases[i];
+  }
+  lastFilledElement += GRPLHCInfo::BeamAliases::NBeamAliases;
+
+  for (int i = 0; i < GRPLHCInfo::BkgAliases::NBkgAliases; ++i) {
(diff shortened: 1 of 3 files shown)


shahor02:
Thanks! Enums are fine, the access via map is much slower.

chiarazampolli:
Ok, then let's see what the CI says. 
Note that the default is now to keep the latest measurement in the vectors, but with an option, it can be changed (`--clear-vectors`)

chiarazampolli:
Moving to WIP, I want to change the way the startValidity is determined, using the earlier timestamp in the vectors for EnvVariables, Collimators etc.

chiarazampolli:
CI failure not related to this PR. Let's see the rest.

chiarazampolli:
Hi @shahor02 , do yo think you could review this one?

Detectors/GRP/calibration/src/GRPDCSDPsProcessor.cxx
```diff
@@ -230,55 +277,62 @@ bool GRPDCSDPsProcessor::processLHCIFDPs(const DPCOM& dpcom)
...
+      if (aliasStr == static_cast<std::string>(GRPLHCInfo::bptxPhaseRMSAliases[ibeam])) {
+        updateVector(dpid, mLHCInfo.mBPTXPhaseRMS[ibeam], aliasStr, dpcomdata.get_epoch_time(), val);
         return true;
-      } else if (aliasStr == fmt::format("{}{}", "BPTX_Phase_Shift_B", ibeam)) {
-        updateVector(dpid, mLHCInfo.mBPTXPhaseShift[ibeam - 1], aliasStr, dpcomdata.get_epoch_time(), val);
+      }
+    }
+    for (int ibeam = 0; ibeam < GRPLHCInfo::BPTXPhaseShiftAliases::NBPTXPhaseShiftAliases; ++ibeam) {
+      if (aliasStr == static_cast<std::string>(GRPLHCInfo::bptxPhaseShiftAliases[ibeam])) {
+        LOG(info) << "aliasStr = " << aliasStr << " alias to check = " << static_cast<std::string>(GRPLHCInfo::bptxPhaseShiftAliases[ibeam]);
```
shahor02: this one is logged even in non-verbose mode, could you just suppress it?  Will merge w/o another set of CIs

chiarazampolli: Ah sorry, leftover, I will fix it.

chiarazampolli: Done

shahor02 (approved):
Tests were passed, afterwards just 1 LOG.. was suppressed, merging

PR #9238, opened 2022-06-21
Protect LTFast calculation agains tracks with too high pT [O2-3023]

shahor02:
Tracks with anomalously high TPC pT (~500 GeV) are assigned 0 helix curvature but
were still treated as helices, leading to NAN.


sawenzel (approved):

shahor02:
failures are unrelated, merging

PR #9242, opened 2022-06-21
DPL: introduce an AsyncQueue

ktf:
Useful to postpone actions which need to happen outside of the
data processing loop and that might have to be debounced, like
sending the oldest possible timeframe.


ktf:
Passing tests and not actually enabled. Merging.

PR #9244, opened 2022-06-22
CPV: add LM-L0 delay subtraction to RawToDigitConverterSpec

sevdokim:
Hello! LM-L0 delay subtraction is added to RawToDigitConverterSpec. Also it is taken into account in simulation/RawWriter.cxx.
Also ccdb timestamp is fetched  from hbfutils (was 'now' previously) in there.

### DIFF
--- a/Detectors/CPV/simulation/src/RawWriter.cxx
+++ b/Detectors/CPV/simulation/src/RawWriter.cxx
@@ -42,33 +42,35 @@ void RawWriter::init()
     mRawWriter->registerLink(link.feeId, link.cruId, link.linkId, link.endPointId, rawFileName.data());
   }
 
-  // Lm-L0 delay
-  mLM_L0_delay = o2::ctp::TriggerOffsetsParam::Instance().LM_L0;
-
   // CCDB setup
   const auto& hbfutils = o2::raw::HBFUtils::Instance();
   LOG(info) << "CCDB Url: " << mCcdbUrl;
   auto& ccdbMgr = o2::ccdb::BasicCCDBManager::instance();
   ccdbMgr.setURL(mCcdbUrl);
   bool isCcdbReachable = ccdbMgr.isHostReachable(); // if host is not reachable we can use only dummy calibration
-  if (!isCcdbReachable) {
+  if (!isCcdbReachable) {                           // dummy calibration
     if (mCcdbUrl.compare("localtest") != 0) {
       LOG(error) << "Host " << mCcdbUrl << " is not reachable!!!";
     }
-    LOG(info) << "Using dummy calibration";
+    LOG(info) << "Using dummy calibration and default Lm-L0 delay";
+    mLM_L0_delay = o2::ctp::TriggerOffsetsParam::Instance().LM_L0;
     mCalibParams = new o2::cpv::CalibParams(1);
     mBadMap = new o2::cpv::BadChannelMap(1);
     mPedestals = new o2::cpv::Pedestals(1);
-  } else {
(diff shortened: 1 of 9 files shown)


sevdokim:
hello @shahor02 
can you please have a look at the ci errors? they look unrelated to cpv code. should I worry about them?

Detectors/CPV/workflow/src/RawToDigitConverterSpec.cxx
```diff
@@ -58,8 +59,47 @@ void RawToDigitConverterSpec::init(framework::InitContext& ctx)
...
+  if (!updateOnlyOnce) {
+    ctx.inputs().get<o2::ctp::TriggerOffsetsParam*>("trigoffset");
+
+    std::decay_t<decltype(ctx.inputs().get<o2::cpv::Pedestals*>("peds"))> pedPtr{};
+    std::decay_t<decltype(ctx.inputs().get<o2::cpv::BadChannelMap*>("badmap"))> badMapPtr{};
+    std::decay_t<decltype(ctx.inputs().get<o2::cpv::CalibParams*>("gains"))> gainsPtr{};
+
+    if (!mIsPedestalData) {
+      pedPtr = ctx.inputs().get<o2::cpv::Pedestals*>("peds");
+      mPedestals = pedPtr.get();
```
shahor02: same as before: why don't you simply use 
```
auto pedPtr = ctx.inputs().get<o2::cpv::Pedestals*>("peds");
mPedestals = pedPtr.get();
```
?

Detectors/CPV/simulation/src/RawWriter.cxx
```diff
@@ -40,7 +42,11 @@ void RawWriter::init()
     mRawWriter->registerLink(link.feeId, link.cruId, link.linkId, link.endPointId, rawFileName.data());
   }
 
+  // Lm-L0 delay
+  mLM_L0_delay = o2::ctp::TriggerOffsetsParam::Instance().LM_L0;
```
shahor02: I don't see where you fetch the TriggerOffsetsParam from the CCDB, seem you are always using default in digi2raw while for decoding you use it from CCDB?

sevdokim: Yes, it is correct. If you suggest how to fetch it using BasicCCDBManager then I'll insert it.

Detectors/CPV/workflow/src/RawToDigitConverterSpec.cxx
sevdokim: There was a reason for that when code was organized in different manner. I'll try to change it now.

Detectors/CPV/simulation/src/RawWriter.cxx
shahor02: Exactly like you do for `CPV/Calib/Gains`, using the timestamp from the hbfutils. Simply, for the ConfigurableParams you don't need to assign the result of `ccdbMgr.get<o2::ctp::TriggerOffsetsParam>("CTP/Config/TriggerOffsets")`

sevdokim: thanks, got it!

sevdokim:
Please don't merge yet. I spotted an error.

sevdokim:
hi! the bug is fixed. please merge if possible

sevdokim:
hi @shahor02 
this PR has no updates for a while. can it be merged or we need to change something?

shahor02 (approved):

PR #9245, opened 2022-06-22
[FT0] Add workflow for processing FT0 DCS DPs

andreasmolander:
Based on https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/TOF/calibration/testWorkflow#dcs-dp-processing


shahor02 (dismissed):
@andreasmolander looks fine to me, just word of caution: like in the TOF device from @chiarazampolli,  which you used as a prototype, you are creating the CCDB object in the end of the integration window with the start of validity timestamp corresponding to last DCS input, essentially, the end of the integration window. So, the data of last 10 min. is stored with the validity of next 10 min. I guess this is ok for FT0.

chiarazampolli:
I will check for TOF, if this is the case, I will change it.

shahor02:
not sure why the Detectors/FIT/FT0/calibration/macros/makeChannelOffsetCalibObjectInCCDB.C macro which was not touched here and excluded from tests triggers an error here, but https://github.com/AliceO2Group/AliceO2/pull/9248 should fix it.

jotwinow (commented):
Dear Andres, 

As Ruben mentioned, the validity of the CCDB object should correspond to real detector readout time. In our case from the first DP to inf. (or the last DP).

andreasmolander:
Currently, the CCDB objects have a three day validity. The validity starts at the moment when new DCS DPs are pushed to the workflow (the first new DPs after the latest CCDB upload). The CCDB object is uploaded 1 second after this start validity, the object contains the latest values seen by the workflow during this 1 second. E.g. inside a CCDB object with start validity at epoch second 1, there can be DPs that arrived to the workflow only at epoch second 1.5. Inside the object, there is also a separate time stamp for each DP, this time comes from DCS, so it's more accurate than the start validity of the CCDB object.

This means there's a smoothing in this workflow as well; upload is not allowed more frequently than 1 Hz. This is also the frequency expected from DCS. But we could disable this to rely only on the DCS smoothing, or e.g. allow an update interval of 0 seconds, which would make the workflow upload to CCDB every time there's new DP values. Doing this, there would be no overwriting of DPs in case a DP is updated multiple times in the 1 second period (but this should not happen anyway thanks to the DCS smoothing).

I don't really understand @shahor02's word of caution; I would say the start of validity corresponds to the *first* DCS input?

As for the validity end time, I understood too long validity times are to be avoided due to DB performance reasons? So then we should decide if three days is good or if it should be adjusted (@jotwinow)? In one CCDB object, fetched based on a timestamp, there would not be values for all DPs anyway, only the ones that were updated most recently before this timestamp. Currently, fetching up to date values for *all* DPs is the responsibility of the software reading these values from CCDB, and not this workflow.

Let me know if I have misunderstood something, or if we need some modifications! The workflow is a bit rough around the edges maybe, I kept it like that due to time constraints, but since there will be some time before deployment on the FLP, we can still make it more sophisticated.

mslupeck:
The validity time of 3 days is OK for now, but we should test if it can be shortened in the future. We should also make sure that we have a single default object with infinite validity time uploaded to CCDB. @andreasmolander have you done it already? 

Do I understand correctly that when querying an object from CCDB, there may be a 1-cycle offset of the validity start time, but inside this object we will have a precise timestamp for each DP, corresponding to real time?

shahor02:
@andreasmolander you are storing by default in 10min intervals, indeed assigning the last value for every DP you saw during these 10 mins for the object valid for the next 10 mins (in opposite to TOF whose object refers to min/max values seen during this interval). I was just referring that you are using past values for the interval in future, up to you to decide if this is ok,
For the validity duration: 3 months is ok, in any case, every next object will override the end of validity of previous object to its start (if they are overlapping).

andreasmolander:
@mslupeck Ok, I will make a JIRA ticket for the default objects. You're second point is correct...

...however, I discussed with Ruben, and 1 Hz is too often to upload stuff to CCDB. The interval of 10 mins is better. So I need to make another structure of the object that is uploaded, that will contain some kind of time seriers of the values during this 10 min period (similar to the way TOF does it). What to store in this object (time spacing between values) we can discuss today on the software meeting (cc @jotwinow).

So I guess we shouldn't merge this just yet.

mslupeck:
> @mslupeck Ok, I will make a JIRA ticket for the default objects. You're second point is correct...
> 
> ...however, I discussed with Ruben, and 1 Hz is too often to upload stuff to CCDB. The interval of 10 mins is better. So I need to make another structure of the object that is uploaded, that will contain some kind of time seriers of the values during this 10 min period (similar to the way TOF does it). What to store in this object (time spacing between values) we can discuss today on the software meeting (cc @jotwinow).
> 
> So I guess we shouldn't merge this just yet.

OK. Let's discuss it later today, so we can make a decision by tomorrow. To me, the proposed solution, to upload the object containing internal DP timestamps every 10 mins is OK for now. It is certainly sufficient for our QC at first. Once (if) we start using it at synch reconstruction, it may become a bottleneck, but we can worry about it only if it becomes an issue.

andreasmolander:
Modified so that every CCDB object contains all DP values that the workflow recieves from DCS. Start validity of the CCDB object corresponds to the first DCS input (during the 10 min interval).

From my side, this can be taken into use on the FLP. The needed DCS DP config: https://alice.its.cern.ch/jira/browse/O2-3043

shahor02 (approved):
CI did not start, apparently due to the 1st submission, approving provisionary to start it.

PR #9247, opened 2022-06-22
Event visualisation / DPL Analysis: Move event display analysis task to .cxx

saganatt:
Hi @jmyrcha @ktf @pnwkw 

I need to move the analysis task to the .cxx file, otherwise, with my Framework updates, I get DWARF errors about repeated symbols during linking.

It is actually enough to remove the redundant `#include Framework/AnalysisTask.h` in `DetectorData.h` but let's stay further on the safe side by keeping this include in .cxx only.

With curiosity:
Actually, I have no idea why this problem appeared just now and in this way. It is enough to add any dummy function to `binning_helpers` in `Framework/BinningPolicy.h` to trigger DWARF, even though there are already 2 other functions in that namespace without any problem. And all header files have `ifndef... define` guards, so it is even more strange the symbols get repeated for the debugger.


ktf:
Do you have the actual error / warning?

saganatt:
Please see the compilation log in the file.

[dwarf.log](https://github.com/AliceO2Group/AliceO2/files/8965709/dwarf.log)

PR #9249, opened 2022-06-22
[EMCAL-784] Provide only EMV-EMCAL sampling factor for Geant4

mfasDa:
Geant4 tune for EMCAL identical to the one used in AliDPG,
however the tag "EMV-EMCAL" was removed from the title.
Therefore the sampling factor "EMV" was selected as the
tag is part of the physics list and it is the second in
priorty. Choosing the EMV sampling factor leads to an
overscaling. In the current stage we only provide the
EMV-EMCAL scaling factor as it is adapted to the usual
physics list, in a next iteration we will provide dedicated
sampling factors for the other two physics lists as well.


sawenzel (approved):

mfasDa:
For the record: As shown in EMCAL-784 now GEANT4 (10.7) shows the same and expected non-linear behaviour as what we have in GEANT3 which was tuned against the testbeam. This still doesn't explain the shift we see with GEANT4 11

PR #9250, opened 2022-06-22
FIT hotfix: fix for protection against single header in event

afurs:
At extremely high rates FIT FEE produce only 1 GBT word(16 bytes) for some events. Such data blocks should be excluded from further processing.


shahor02 (approved):
tested locally by @afurs , merging

PR #9252, opened 2022-06-23
[EMCAL-630] Correction for BC mod

mfasDa:
Due to clock difference between LHC (40 MHz) and
ALTRO (10 MHz) 4 bunch crossings are within one
ALTRO sample. Consequently the fitted time of the trigger
of the collision for the trigger BC is shifted by 25, 50 and 75
ns for the different BC % 4 (and consequently also
translates into the pileup peaks). In order to align the time
peaks for the differnt trigger BC the convention is used to
shift the times for trigger in BC % 4 != 0 to BC %4 = 0.

### DIFF
--- a/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
@@ -395,6 +395,10 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
             if (fitResults.getTime() < 0) {
               fitResults.setTime(0.);
             }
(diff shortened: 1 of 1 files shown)


mfasDa:
@mpoghos for reference

@shahor02 BC phases and their ordering were validated during PB2, now a correction can be applied. This could be either done in the reconstruction directly via the formula time_corr = time - 25 * (bc %4) or it could be absorbed in the time calibration. As the correction will stay the same we think it is better to apply it directly in the reconstruction, because absorbing in the time calibrartion would require preforming and applying the time calibration for each BC % 4, so 4 times instead of 1 time, making the time calibration even slower. In run2 this was necessary due to the interplay with the L1 phase, however since
this is fixed the ordering will stay fixed and we can apply the correction directly.

shahor02 (approved):

PR #9257, opened 2022-06-23
Muon dcs processor: change way to set validity range

aphecetche:
1st commit : Getting the validity range from the datapoints themselves does not seem doable (as some datapoints seem to get "stuck" in the past), so changing to a method that is used in other DCSProcessor : get the start of validity from the first input message received (and an end of validity = start + 5 days) 

2nd commit: add an option to `mch|mid-dcs-ccdb` to read from local Root file.

### DIFF
--- a/Detectors/MUON/Common/src/dcs-ccdb.cxx
+++ b/Detectors/MUON/Common/src/dcs-ccdb.cxx
@@ -27,11 +27,22 @@
 #include <unordered_map>
 #include <vector>
 #include "subsysname.h"
+#include <TFile.h>
 
 namespace po = boost::program_options;
 using DPID = o2::dcs::DataPointIdentifier;
 using DPVAL = o2::dcs::DataPointValue;
 using DPMAP = std::unordered_map<DPID, std::vector<DPVAL>>;
+
+float sum(float s, o2::dcs::DataPointValue v)
+{
+  union Converter {
+    uint64_t raw_data;
+    double value;
+  } converter;
+  converter.raw_data = v.payload_pt1;
+  return s + converter.value;
+};
 
 std::string CcdbDpConfName()
 {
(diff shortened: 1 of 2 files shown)


chiarazampolli:
Hi @aphecetche , 

I just opened a PR for TOF that should do the opposite of what you did: from setting the validity according to the time when the data are received, to the earliest measurement. After your comment, I am thinking that maybe it is the end validity that should be set to not 3 days after the start, but 3 days after the current time. Would this work better?

@noferini , @shahor02 .

shahor02:
Isn't the difference 10 min. anyway?

shahor02:
I would always set the start of integration period, the internal timestamp of earliest datapoint is indeed misleading since it is likely not to valid for other points

aphecetche:
can't comment on the TOF datapoints, but as an example of the issue I have with the current MCH datapoints (and that this PR ought to solve) can be found in the first comment of [MRRTF-175](https://alice.its.cern.ch/jira/browse/MRRTF-175?focusedCommentId=285745&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-285745) : in that example two datapoints have timestamps that are 9 hours before the other ones...

chiarazampolli:
The lines here in the current version for TOF:
https://github.com/AliceO2Group/AliceO2/blob/74b7feb0319f339fc19b3c550df1c0c711827059/Detectors/TOF/calibration/src/TOFDCSProcessor.cxx#L72-L74
had the purpose to use the start of the integration time. Then I guess I just forgot to use them. 
Indeed setting the start validity as the earliest measurement is wrong: in the end, if a DP has the value of 9h ago, it means that this value is still valid, so the correct timestamp is the start of the integration. 
I will revise the PR for TOF and use what exists. 
@shahor02 : we wanted to change also GRP. Also in this case the correct is the start of the integration. In addition, note that with vectors one can call the function https://github.com/AliceO2Group/AliceO2/blob/74b7feb0319f339fc19b3c550df1c0c711827059/Detectors/Calibration/include/DetectorsCalibration/Utils.h#L46, and if the timestamp with which it is called is later than the elements in the vector, the latest is anyway given, so it will be fine (unless I am forgetting some corner cases).

pillot (approved):

PR #9258, opened 2022-06-23
DPL: demote info message to debug

ktf:


ktf:
@davidrohr any objections?

davidrohr:
sure this is fine, also some more should be demoted, but let's have a look after the current issue with CALIB is fixed

PR #9259, opened 2022-06-23
Fix in the "tree" of possible track prolongations

iouribelikov:


shahor02 (approved):
failure is unrelated. merging

PR #9264, opened 2022-06-24
correct and standardize the definition of MCH/MID-based track time

pillot:
This corrects the calculation of the MCH/MID-based track time. Also it is now done in one single place for each type of track (MCH-standalone, MID-standalone and MCH-MID), thus avoiding having different definitions in different places.

The fonctions return the track time with error, in µs, wrt to the TF start given as a parameter. They also return a flag to assess its validity, i.e. if the first BC of the time bracket is inside the TF, whose number of orbits is also a parameter of the functions. Finally they have a third parameter to optionally print an error message if it is not valid.

I also profit from this PR to fix the calculations of muon track parameter covariances and pDCA as well as a harmless bug in MFT and fwd track index recording in the AOD producer.

### DIFF
--- a/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
+++ b/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
@@ -299,12 +299,13 @@ void o2::globaltracking::RecoContainer::createTracksVariadic(T creator) const
     }
     for (unsigned i = 0; i < matchesMCHMID.size(); i++) {
       const auto& match = matchesMCHMID[i];
+      auto [trcTime, isInTF] = match.getTimeMUS(startIR, 256, BCDiffErrCount < MAXBCDiffErrCount);
+      if (!isInTF && BCDiffErrCount < MAXBCDiffErrCount) {
+        BCDiffErrCount++;
+      }
       auto gidxMCH = match.getMCHRef();
       const auto& trc = tracksMCH[gidxMCH.getIndex()];
-      float t0err = 0.0005;
-      auto bcdiff = getBCDiff(match.getIR());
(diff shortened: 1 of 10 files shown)


DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainerCreateTracksVariadic.h
```diff
@@ -299,12 +299,13 @@ void o2::globaltracking::RecoContainer::createTracksVariadic(T creator) const
     }
     for (unsigned i = 0; i < matchesMCHMID.size(); i++) {
       const auto& match = matchesMCHMID[i];
+      auto [trcTime, isInTF] = match.getTimeMUS(startIR, 256, BCDiffErrCount < MAXBCDiffErrCount);
```
aphecetche, line 302: why are you using a hard-coded number for the number of orbits and why 256 ? (knowing that the default of the function, 128, seems more in line with the current data taking conditions ?)

pillot: I use the same hardcoded number as L 133 above to keep the current behavior. If we take this value from somewhere (?) I imagine it should be done consistently within this function.

shahor02: At the moment I used constant 256 orbits as max possible duration of the TF. To be rigorous one should use run specific value (usually 128 but e.g. MC sometimes uses 256) but this would require a guarantee that the GRPECS was loaded, which is not trivial with such a helper method as RecoContainer.

pillot:
I think this PR is ready to be merged, isn't it ? :)

davidrohr:
@aphecetche @shahor02 : can we merge this?

aphecetche (approved):

shahor02 (approved):

PR #9273, opened 2022-06-26
DPL: improve color dark red

ktf:


ktf:
Difficult to read exceptions with the old value.

PR #9280, opened 2022-06-27
TF rate limiting for CTF, raw-file and raw-tf readers

shahor02:
The reader will throttle injection of the TF to not more than N TFs in processing if
--timeframes-rate-limit <N> --timeframes-rate-limit-ipcid <IPCID>
options are provided too all workflows (e.g. via ARGS_ALL)
The IPCID is the NUMA domain ID (usually 0 on non-EPN workflow).

Additionally, one may throttle on the free SHM by providing an option to the reader
--timeframes-shm-limit <shm-size>

### DIFF
--- a/Detectors/CTF/README.md
+++ b/Detectors/CTF/README.md
@@ -172,3 +172,8 @@ by `ctrl-C`. Periodic incremental saving of so-far accumulated dictionary data d
 ``--save-dict-after <N>``.
 
 When decoding CTF containing dictionary data (i.e. encoded w/o external dictionaries), externally provided dictionaries will be ignored.
+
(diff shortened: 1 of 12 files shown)


shahor02:
@davidrohr FullCI fails with 
```
reco_NOGPU.log:[8639:gpu-reconstruction]: [21:35:18][FATAL] error while setting up workflow in o2-gpu-reco-workflow: Need MC information to create QA plots
```
Any idea why?  I don't see any change in the `GPU/Workflow/src/GPUWorkflowSpec.cxx` which prints this message.

davidrohr:
Pretty weird, I don't see what could have changed.
Will check...

davidrohr:
Sorry, my fault, fixed in #9283 (yesterday accidentally merged the wrong PR, but since it looked non-critical I didn't bother...)

shahor02:
@davidrohr  Thanks!

PR #9281, opened 2022-06-27
Add TPC residual aggregator CALIB workflow

martenole:
Add aggregator workflow for TPC residuals. Currently without defining `--meta-output-dir` which means that the output is not shipped to alien. But one could nevertheless check with the created local files if they are reasonable.
Goes together with https://github.com/AliceO2Group/O2DPG/pull/496


chiarazampolli (dismissed):

davidrohr:
That seems to be a genuine error: `/sw/slc8_x86-64/O2/9281-slc8_x86-64-local1/prodtests/full-system-test/dpl-workflow.sh: line 470: o2-tpc-residual-aggregator: command not found`

martenole:
> That seems to be a genuine error: `/sw/slc8_x86-64/O2/9281-slc8_x86-64-local1/prodtests/full-system-test/dpl-workflow.sh: line 470: o2-tpc-residual-aggregator: command not found`

Indeed, its called `o2-calibration-residual-aggregator`...

PR #9284, opened 2022-06-28
TPC: SACs, add missing call to send data, bugfixes

wiechula:
* Add logic to only sent data, if all FEs were decoded for a time stamp
* Clean up


wiechula:
The error seems unrelated, can this be merged?

shahor02 (approved):

PR #9286, opened 2022-06-28
TRD small reco update

martenole:
- digits and tracklets print methods updated to include sector_stack_layer
- allow to set externally additional search road in time for ITS-TPC tracks (to account for possible offset between TRD trigger time and ITS-TPC track times observed in recent pilot beam runs)


bazinski (approved):

martenole:
@davidrohr is this change fine with you? Is it correct that I don't need to update the `ClassDef` version for `Digit` and `Tracklet64` in case I only change the member functions for printing?

PR #9291, opened 2022-06-29
DPL Analysis: Event mixing: Refactor binning policy to accept lambdas

saganatt:
Ciao @ktf @aalkin, @jgrosseo 

this allows using custom lambda functions for binning and mixing. Let me know if you have any coding suggestions.

ColumnBinningPolicy preserves the old interface for using columns only. However, these changes might still break O2Physics. I can work around this by a series of interleaved O2 and O2Physics PRs, but maybe let's first agree on the O2 implementation. And I will then split this PR.


saganatt:
Split into other PRs.

PR #9293, opened 2022-06-29
TRD fix bad rdh headers, add numerous data integrity checks

bazinski:
- fix rdh bad headers, in particular 0xfff? for the FEEID, data gets dumped.
- caused a sector of 255 instead of [0,17] and was used as an index into an array.
- ensure sequential rdh have identical relevant information
- ensure halfcru headers on a link all have same informaiton.
- add checks for various other parameters to make sure they are inrange.
- change stats to not have 2d plots associated with the 1d plot in those cases where you cant know the sector,side,stack,layer.
- add sanity checks to volume of data to attempt to parse.
- remove all the pre qc, qc where various plots were produced by the decoder. All that info is now passed to qc via a rawstats message.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
@@ -73,9 +73,23 @@ enum ParsingErrors { TRDParsingNoError,
                      TRDParsingTrackletIgnoringDataTillEndMarker,              // for some reason we are bouncing to the end word by word, this counts those words
                      TRDParsingGarbageDataAtEndOfHalfCRU,                      // if the first word of the halfcru is wrong i.e. side, eventype, the half cru header is so wrong its not corrupt, its other garbage
                      TRDParsingHalfCRUSumLength,                               // if the HalfCRU headers summed lengths wont fit into the buffer, implies corruption, its a faster check than the next one.
-
-                     TRDParsingHalfCRUCorrupt, // if the HalfCRU headers has values out of range, corruption is assumed.
-                     TRDLastParsingError       // here as place holder for trivial sizing of structures.
+                     TRDParsingBadRDHFEEID,                                    // RDH parsing failure for reasons in the word
+                     TRDParsingBadRDHEndPoint,                                 // RDH parsing failure for reasons in the word
+                     TRDParsingBadRDHOrbit,                                    // RDH parsing failure for reasons in the word
+                     TRDParsingBadRDHCRUID,                                    // RDH parsing failure for reasons in the word
+                     TRDParsingBadRDHPacketCounter,                            // RDH parsing failure for reasons in the word
+                     TRDParsingHalfCRUCorrupt,                                 // if the HalfCRU headers has values out of range, corruption is assumed.
+                     TRDParsingDigitHCHeader1,                                 // multiple instances of Digit HC Header 1
+                     TRDParsingDigitHCHeader2,                                 // multiple instances of Digit HC Header 2
+                     TRDParsingDigitHCHeader3,                                 // multiple instances of Digit HC Header 3
+                     TRDProcessingBadPayloadOrOffset,                          // if something is off with the HBFPayload array or its offset into it.
+                     TRDParsingDigitHCHeaderSVNMismatch,                       // svn version information has changed in the DigitHCHeader3.
+                     TRDParsingBadLinkstartend,                                // end - start of tracklet is greater than the maximal length stored in the cru half chamber header field.
+                     TRDParsingTrackletsReturnedMinusOne,                      // trackletparsing returned -1, data was dumped;
+                     TRDFEEIDIsFFFF,                                           // RDH is in error, the FEEID is 0xffff
+                     TRDFEEIDBadSector,                                        // RDH is in error, the FEEID.supermodule is not a valid value.
+                     TRDParsingDigitHCHeaderPreTriggerPhaseOOB,                // pretrigger phase in Digit HC header has to be less than 12, it is not.
+                     TRDLastParsingError
 };
 
 extern std::vector<std::string> ParsingErrorsString;
@@ -99,8 +113,12 @@ enum OptionBits {
   TRDM1Debug
 };
 
+//Data to be stored and accumulated on an event basis.
+//events are spread out with in the data coming in with a halfcruheader per event, per ... half cru.
+//this is looked up via the interaction record (orbit and bunchcrossing).
+//this permits averaging in the data that gets senton per timeframe
 class TRDDataCountersPerEvent
-{ //thisis on a per event basis
+{
  public:
   //TODO this should go into a dpl message for catching by qc ?? I think.
   double mTimeTaken;             // time take to process an event (summed trackletparsing and digitparsing) parts not accounted for.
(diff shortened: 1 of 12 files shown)


Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -129,6 +129,88 @@ void CruRawReader::dumpRDHAndNextHeader(const o2::header::RDHAny* rdh)
...
 }
 
+bool CruRawReader::checkRDH(const o2::header::RDHAny*  rdh)
+{
+// first check for bogus FEEID
+  TRDFeeID feeid;
+  feeid.word=0;
+  feeid.word= o2::raw::RDHUtils::getFEEID(rdh);
+  if (feeid.word==0xffff){
+    LOG(info) << "failed due to 0xffff : " << std::hex << feeid.word << " whole feeid : " << std::hex << (unsigned int)feeid.word;
```
martenole: If I understood Guido correctly this would happen in case a CRU is not configured. And if everything is properly configured this should never happen. Thus I would throw an `error` here so that the operators see it immediately and call the TRD oncall. I assume we would want to configure the CRU correctly immediately when this happens, right?

bazinski: oops, yes you are right, that was actually there as a debug statement and meant to be changed to error message, let me go through and check, there might be more of those.
I also just realised I should fix the accounting, so that the dumped data is allocated to the dumped data count as opposed to be ignored.

bazinski: I am also not too happy with the mechanism for limiting errors via mMaxErrsPrinted, as things get missed. I think some errors need to be made warnings and possibly alarms.

bazinski: Also need to expand the qc shifter notes to include watching closely the last bins in the parsing errors plots.

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -535,13 +543,17 @@ void printDigitMCMData(o2::trd::DigitMCMData& digitmcmdata)
...
+  //  LOG(info) << "getDigitHCHeaderwordtype : " << std::hex << word;
   if ((word & 0x3f) == 0b110001) {
+    //  LOG(info) << "getDigitHCHeaderwordtype  2 : " << std::hex << word << " returning 2 for :" << std::hex << (word&0x3f);
     return 2;
   }
   if ((word & 0x3f) == 0b110101) {
+    //  LOG(info) << "getDigitHCHeaderwordtype 3 : " << std::hex << word << " returning 3 for :" << std::hex << (word&0x3f);
     return 3;
   }
   if ((word & 0x3) == 0b01) {
```
tdietel, line 555: This check could be stricter: the upper 4 bits, which are the pretrigger phase in HCH1,  can only take values of 0-11. 12 and 13 are for HCH2 and HCH3, respectively, and 14,15 should never happen.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -129,6 +129,91 @@ void CruRawReader::dumpRDHAndNextHeader(const o2::header::RDHAny* rdh)
   LOG(info) << "######################### Finished RDH incoming buffer ##########################";
 }
 
+bool CruRawReader::checkRDH(const o2::header::RDHAny* rdh)
+{
+  // first check for bogus FEEID
+  TRDFeeID feeid;
+  feeid.word = 0;
```
tdietel: Unnecessary -> remove.

tdietel (commented):
Looks fine, I only have to minor comments.

davidrohr:
The failed checks seem to be unrelated. But could you rebase?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: done

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: This is not where the checks are done, that is just to identify which digit half chamber it is.
But yes will implement those checks in the relevant places.

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
```diff
@@ -114,7 +127,7 @@ class TRDDataCountersPerEvent
 };
 
 class TRDDataCountersPerTimeFrame
-{ //thisis on a per event basis
+{ //this is on a per event basis
```
martenole: isn't it per TF?

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
```diff
@@ -66,6 +65,16 @@ int TrackletsParser::Parse(std::array<uint32_t, o2::trd::constants::HBFBUFFERMAX
   mIgnoreTrackletHCHeader = options[TRDIgnoreTrackletHCHeaderBit];
   mEventRecord = eventrecord;
   mEventRecords = eventrecords;
+  if (std::distance(start, end) > o2::trd::constants::MAXDATAPERLINK32) { // full event is all digits and 3 tracklets per mcm,
+    //sanity check that the length of data to scan is less the possible maximum for a link
```
martenole, line 69: Would this be worth a log message or is that anyhow coming higher in the hierarchy in case the parsing fails?

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
```diff
@@ -159,6 +165,10 @@ int TrackletsParser::Parse()
   int trackletloopcount = 0;
   int headertrackletcount = 0;
   bool ignoreDataTillTrackletEndMarker = false;             // used for when we need to dump the rest of the tracklet data.
+  if (std::distance(mStartParse, mEndParse) > o2::trd::constants::MAXDATAPERLINK32) { // full event is all digits and 3 tracklets per mcm,
+    //sanity check that the length of data to scan is less the possible maximum for a link
```
martenole, line 171: same question here as above

Detectors/TRD/reconstruction/src/CruRawReader.cxx
martenole: Not done yet?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -129,6 +129,91 @@ void CruRawReader::dumpRDHAndNextHeader(const o2::header::RDHAny* rdh)
...
+  feeid.word = o2::raw::RDHUtils::getFEEID(rdh);
+  if (((feeid.word) >> 4) == 0xfff) { // error condition is 0xfff? as the end point is known to the cru the "?" but the rest is configured.
+    if (mVerbose) {
+      LOG(error) << "failed due to 0xffff : " << std::hex << feeid.word << " whole feeid : " << std::hex << (unsigned int)feeid.word;
```
martenole: I still think it would be better to always show this error, not only in verbose mode. Maybe with an additional unique counter such that it is not hidden by other error messages? And once we have the QC with automatic checks and a large red box in the plot saying please call oncall then we can remove the error message again

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -334,27 +428,43 @@ int CruRawReader::parseDigitHCHeader()
...
+        }
         mDigitHCHeader1.word = headers[headerwordcount];
+        headersfound.set(0);
         if (mDigitHCHeader1.res != 0x1) {
-          //LOG(alarm) << "Digit HC Header 1 reserved : " << std::hex << mDigitHCHeader1.res << " raw: 0x" << mDigitHCHeader1.word;
-          increment2dHist(TRDParsingDigitHeaderWrong1, mFEEID.supermodule * 2 + mHalfChamberSide[0], mStack[0], mLayer[0]);
+          LOG(alarm) << "Digit HC Header 1 reserved : " << std::hex << mDigitHCHeader1.res << " raw: 0x" << mDigitHCHeader1.word;
+          incrementErrors(TRDParsingDigitHeaderWrong1);
+        }
+        //if
```
martenole: ?

martenole (commented):
From what I can tell this looks ok, only minor comments I think

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
bazinski: no, this is per event, its looked up via the interaction record to get the right event across various crus.

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
bazinski: It should be caught by line 729 of CruRawReader.cxx, this is just a cover my ass, but yes its an error so will put in an error message.

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
bazinski: same reply as above ;-)

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: Its caught by the calling method, the return of false triggers an error message. The FEEID is output in that error message. The verbose version is just so one can see which error is actually caught with out having to sit and decode it.

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
martenole: Then the class name `TRDDataCountersPerTimeFrame` confuses me

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: must have been planning to write a comment and got distracted and never came back. removed.

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
bazinski: Its 2 sets of numbers, some numbers are accumulated into the timeframe number as well, some keep their granularity so an average can be taken for the whole time frame. One message per time frame.
It took me a while to figure it out, will comment accordingly to make it all clear (hopefully).

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: done

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: done now.

shahor02:
Is this ready to be merged?

bazinski:
@shahor02 I just noticed that  a warning was not wrapped. Can merge after the ci is happy.

bazinski:
@tdietel @martenole you happy to have this merged ?

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
martenole: Possibly add the check `mMaxErrorsPrinted` here?

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
martenole: same here

PR #9295, opened 2022-06-30
DPL Analysis: Improve tests for persistent and index columns

saganatt:
Hi @ktf @aalkin 

this is the first split from #9291 . I need `is_persistent_t` to work also with types that do not contain the `persistent` field.
I also improved the functions for getting index columns data to properly recognize different index kinds.

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -541,8 +541,23 @@ struct Index : o2::soa::IndexColumn<Index<START, END>> {
 template <typename T>
 using is_dynamic_t = framework::is_specialization<typename T::base, DynamicColumn>;
 
+namespace persistent_type_helper
+{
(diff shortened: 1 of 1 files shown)


saganatt:
Hi, if you have time could you please look at this? Thanks a lot!

Framework/Core/include/Framework/ASoA.h
```diff
@@ -541,8 +541,23 @@ struct Index : o2::soa::IndexColumn<Index<START, END>> {
 template <typename T>
 using is_dynamic_t = framework::is_specialization<typename T::base, DynamicColumn>;
 
+namespace persistent_type_helper
```
ktf, line 544: Not sure I like the idea of such a specific namespace. Do you really need it?

saganatt: Then instead would you prefer the included functions to have a more specific name instead of `test`?

PR #9297, opened 2022-06-30
GRP tool for simulation

sawenzel:
Introduce a GRP tool for simulation
    
    Introducing a GRP command line tool for MC, which can
    
    - create the set of ECS,MagField,LHCIF objects locally
      **or** download them from CCDB, for the use in MC simulations.
    
      (creation is fast and does not need firing up the simulation executable
       as the case up until now)
    
    - publish or put these objects into directories where they can be picked
      up through CCDB queries (during reco)
    
    - edit these files to change settings (prior to digitization)
      * for now only RO mode for GRPECS is supported
    
    - print GRPs to screen
    
    The tool offers command and sub-command options.
    
    The final goal is to allow MC processing to proceed the same way
    irrespective of whether the mode is anchored, unachored, ad-hoc, etc.
    
    The tool achieves this by enabling a common workflow for all these cases.
    
    Note that the present version has some basic functionality implemented to give
    the idea. Extensions are to follow.
    
    The sim_challenge.sh script is modified to show usage of the new tool.

### DIFF
--- a/Common/SimConfig/include/SimConfig/SimConfig.h
+++ b/Common/SimConfig/include/SimConfig/SimConfig.h
@@ -117,6 +117,9 @@ class SimConfig
   static void determineActiveModules(std::vector<std::string> const& input, std::vector<std::string> const& skipped, std::vector<std::string>& active);
   static void determineReadoutDetectors(std::vector<std::string> const& active, std::vector<std::string> const& enabledRO, std::vector<std::string> const& skippedRO, std::vector<std::string>& finalRO);
 
+  // helper to parse field option
+  static bool parseFieldString(std::string const& fieldstring, int& fieldvalue, o2::conf::SimFieldMode& mode);
+
   // get selected generator (to be used to select a genconfig)
   std::string getGenerator() const { return mConfigData.mGenerator; }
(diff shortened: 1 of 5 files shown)


sawenzel:
This would be a first step. More tasks need to follow, such as

- [ ] use of GRP information inside digitization (instead of via command line arguments)
- [ ] deprecating GRPUpdater during digitization
- [ ] make use of GRP inside simulation instead of recreating files
- [ ] convert all workflows to pickup GRP like pvfinder does
- [ ] adjust all production scripts (O2DPG).

shahor02 (dismissed):
Thanks! 
For the "todo" list:
1) I would add storing the the GRPLHCIF generator parameters (beam type, energy), reconstruction needs them.
2) `convert all workflows to pickup GRP like pvfinder does`: https://github.com/AliceO2Group/AliceO2/pull/9220 implements it for O2 workflows. For the QC, the extension of the GRPGeomHelper to o2-qc is being discussed, then the QC tracks also can be modified.

Not directly related, but I would consider also moving o2sim_geometry.root and o2sim_geometry-aligned.root to CCDB-like directories, to allow access via --remap-condition.

PR #9300, opened 2022-06-30
improve MCH chambers response

pillot:
- Extracts the improvements made in #7010 on chamber response in term of integrated charge distribution (leaving aside the modifications on the noise and time distribution for the moment)
- Add further improvements on this response and mutualize the code also used in the clustering
- Add a response in term of number of ADC samples corresponding to the simulated integrated ADC value (the best I can do for now without simulating the full electronic signal)

The later allows to apply the same digit filtering in MC as in data, thus allowing to run the data reconstruction workflow on simulated data (apart from the TF duration limitation in the time clustering which is another topic).

The parameterization of the response 'nSamples vs ADC' has been done using the same data as the one used to tune the filtering selections, to be consistent. They might need some adjustments with better data later on.


Detectors/MUON/MCH/Simulation/include/MCHSimulation/Response.h
```diff
@@ -25,77 +26,52 @@ enum class Station {
...
 
-  //parameter for station number
-  Station mStation;
-  //proper parameter in aliroot in AliMUONResponseFactory.cxx
-  float mQspreadX; //charge spread in cm
-  float mQspreadY;
+  /** Converts energy deposition into a charge.
+   *
+   * @param edepos deposited energy from Geant (in GeV)
+   * @returns an equivalent charge (roughyl in ADC units)
```
aphecetche, line 43: (for a next PR) : roughyl => roughly.

aphecetche (approved):

aphecetche:
Looks good to me. Thanks.

PR #9305, opened 2022-06-30
DPL: make sure severity flag is used ASAP

ktf:


ktf:
@davidrohr this should be enough. I will check in a bit...

ktf:
Tested locally and behaves as expected.

PR #9308, opened 2022-07-01
ZDC - Checks in DigitReaderSpec

cortesep:


Detectors/ZDC/workflow/src/ZDCDataReaderDPLSpec.cxx
```diff
@@ -78,13 +78,24 @@ void ZDCDataReaderDPLSpec::run(ProcessingContext& pc)
...
+      if (it.data() == nullptr) {
+        LOG(alarm) << "ZDCDataReaderDPLSpec::run - Null payload pointer on page " << count;
+      } else if (it.size() == 0) {
+        LOG(alarm) << "ZDCDataReaderDPLSpec::run - No payload on page " << count;
+      } else {
+        gsl::span<const uint8_t> payload(it.data(), it.size());
+        mRawReader.processBinaryData(payload, rdhPtr->linkID);
+      }
+    }
+    count++;
```
shahor02, line 114: If raw data is corrupted, it will be likely like that for the whole TF, I would simply abandon the TF sending empty output.
Producing alarm on empty payload will overflow the infologger in case the problem persists.
Please add a counter for the alarm messages printed and stop printing them after some count (not very high, e.g. 5).

Detectors/ZDC/workflow/src/ZDCDataReaderDPLSpec.cxx
```diff
@@ -78,16 +78,56 @@ void ZDCDataReaderDPLSpec::run(ProcessingContext& pc)
     mRawReader.setVerifyTrigger(mVerifyTrigger);
     LOG(info) << "Check of trigger condition during conversion is " << (mVerifyTrigger ? "ON" : "OFF");
   }
+
   uint64_t count = 0;
+  uint64_t nErr[3] = {0};
```
shahor02: this should be `static uint64_t nErr[3] = {0};`

shahor02 (changes requested):

Detectors/ZDC/workflow/src/ZDCDataReaderDPLSpec.cxx
cortesep: Should it survive different run calls to drop all the remaining part of the timeframe?

shahor02: yes, otherwise the logs will be swamped by these messages. If you want to check if the errors are persistent, you can add a check when last message was printed and reset the counter e.g. after 1 min.

shahor02 (approved):

PR #9310, opened 2022-07-01
Avoid OOB access

ktf:


ktf:
Maybe it's not what causes the crash online, but shouldn't it be this way?

kharlov (approved):

peressounko (approved):

ktf:
@shahor02 @davidrohr merging this as it passes all tests and in any case fixes a possible issue.

PR #9315, opened 2022-07-01
MeanVertex X and Y coordinate depend on Z

chiarazampolli:
And make MeanVertex object derive from VertexBase.

@shahor02 , I am still testing, this is why there is the WIP.

### DIFF
--- a/DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
+++ b/DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
@@ -15,29 +15,38 @@
 #include <array>
 #include "Rtypes.h"
 
+#include "Framework/Logger.h"
+#include "ReconstructionDataFormats/Vertex.h"
+
 namespace o2
 {
 namespace dataformats
 {
-class MeanVertexObject
+class MeanVertexObject : public VertexBase
 {
 
  public:
-  MeanVertexObject(float x, float y, float z, float sigmax, float sigmay, float sigmaz)
+  MeanVertexObject(float x, float y, float z, float sigmax, float sigmay, float sigmaz, float slopeX, float slopeY)
   {
-    mPos[0] = x;
-    mPos[1] = y;
-    mPos[2] = z;
-    mSigma[0] = sigmax;
-    mSigma[1] = sigmay;
-    mSigma[2] = sigmaz;
+    math_utils::Point3D<float> p(x, y, z);
+    gpu::gpustd::array<float, kNCov> cov;
+    cov[CovElems::kCovXX] = sigmax;
+    cov[CovElems::kCovYY] = sigmay;
+    cov[CovElems::kCovZZ] = sigmaz;
+    VertexBase(p, cov);
+    mSlopeX = slopeX;
+    mSlopeX = slopeY;
   }
-  MeanVertexObject(std::array<float, 3> pos, std::array<float, 3> sigma)
+  MeanVertexObject(std::array<float, 3> pos, std::array<float, 3> sigma, float slopeX, float slopeY)
   {
-    for (int i = 0; i < 3; i++) {
-      mPos[i] = pos[i];
-      mSigma[i] = sigma[i];
-    }
+    math_utils::Point3D<float> p(pos[0], pos[1], pos[2]);
+    gpu::gpustd::array<float, kNCov> cov;
+    cov[CovElems::kCovXX] = sigma[0];
+    cov[CovElems::kCovYY] = sigma[1];
+    cov[CovElems::kCovZZ] = sigma[2];
+    VertexBase(p, cov);
+    mSlopeX = slopeX;
+    mSlopeX = slopeY;
   }
   MeanVertexObject() = default;
   ~MeanVertexObject() = default;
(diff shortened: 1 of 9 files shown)


chiarazampolli:
Ready to start review. Note that there are quite many changes, hopefully I did not miss anything.

DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
```diff
@@ -46,42 +55,69 @@ class MeanVertexObject
...
+  float getYAtZ(float z) { return getY() + mSlopeY * (z - getZ()); }
+
+  void print() const;
+  std::string asString() const;
 
  private:
-  std::array<float, 3> mPos;   // position of mean vertex
-  std::array<float, 3> mSigma; // sigma of mean vertex
+  float mSlopeX; // slope of x = f(z)
+  float mSlopeY; // slope of y = f(z)
```
shahor02: init to 0 by default: `mSlopeX{0.f};`

Detectors/Calibration/src/MeanVertexCalibrator.cxx
```diff
@@ -35,137 +36,280 @@ void MeanVertexCalibrator::initOutput()
...
-      LOG(error) << "Y: Fit failed with result = " << fitres;
-    }
-    mSMAMVobj.setY(fitValues[1]);
-    mSMAMVobj.setSigmaY(fitValues[2]);
+    lf.AddPoint(&meanZvect[i], fitResSlicesX[i][1]);
+  }
+  lf.Eval();
+  double slopeX = lf.GetParameter(0);
+  mvo.setSlopeX(slopeX);
+  mvo.setX(mvo.getZ() * slopeX + lf.GetParameter(1));
```
shahor02: isn't it opposite: `offset = lf.GetParameter(0)` and `slope = lf.GetParameter(1)` ?

DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
```diff
@@ -15,29 +15,38 @@
...
 
+#include "Framework/Logger.h"
+#include "ReconstructionDataFormats/Vertex.h"
+
 namespace o2
 {
 namespace dataformats
 {
-class MeanVertexObject
+class MeanVertexObject : public VertexBase
```
shahor02, line 25: we need also getters for mean vertex at given Z, i.e.

```
VertexBase getMeanVertex(float z) 
{
  VertexBase v = *this;
  float dz = getZ() - z;
  v.setXYZ(v.getX()+mSlopeX*dz, v.getY()+mSlopeY*dz, z);
}

const VertexBase& getMeanVertex() const {
  return (const VertexBase&)(*this);
}
```

shahor02 (changes requested):
Hi, sorry for the late review. Note that there are conflicts to resolve.

DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
chiarazampolli: Done

Detectors/Calibration/src/MeanVertexCalibrator.cxx
chiarazampolli: Good catch, thanks!

DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
chiarazampolli: Done, using the functions already in the class.

chiarazampolli:
Hi @shahor02 , 
Errors are not related to this PR. 
Do you think you will have more comments? If so, the PR can wait till you are back, of course, otherwise, do you think we could merge it?
Thanks, 
Chiara

DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
```diff
@@ -46,42 +56,81 @@ class MeanVertexObject
...
-    for (int i = 0; i < 3; i++) {
-      mPos[i] = val[i];
+    if (icoord == 0) {
+      setX(val);
+    } else if (icoord == 1) {
+      setY(val);
+    } else if (icoord == 2) {
+      setZ(val);
+    } else {
+      LOG(fatal) << "Coordinate out of bound to set vtx " << icoord << ", should be in [0, 2]";
```
ktf: It's probably a good idea to avoid using Logger in an inlined function in an header. Why do you need to change the API?

chiarazampolli: which API do you mean?

ktf: Why setting things one by one rather than using the array?

chiarazampolli: This sets only 1 coordinate.

chiarazampolli:
Hi @shahor02 , 
Do you have further comments? otherwise, could we merge it? We can also wait that you are back, of course, but if you think it is fine, maybe we could merge. 
@ktf , I addressed one of your comments. For the logger, if it is forbidden, I can move the function to the source file.
Cheers,
Chiara

DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
ktf: Yes, I get that, but why? Where is it actually used?

ktf:
Yes, the LOG should not be part of a header, IMHO. I also do not like the API for setting one component at the time. If you really need that, I would simply return a reference to the actual container class.

DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
chiarazampolli: See line 293 of Detectors/Calibration/src/MeanVertexCalibrator.cxx in this PR.
And anyway, I would not see any issue in having such method.

chiarazampolli:
> Yes, the LOG should not be part of a header, IMHO. I also do not like the API for setting one component at the time. If you really need that, I would simply return a reference to the actual container class.

I can move the implementation to the source file, but I would like to keep this method as it is. I don't see any drawback in having it.

ktf:
Yes, please, at least move it out of line, we should not have LOG statements in headers.

The drawback is that you need to check at runtime for something which is actually known at compile time since AFAICT you only have the version with "2" hardcoded. I would keep the separate methods and have a switch(icoord) in fitMeanVertexCoord...

chiarazampolli:
> Yes, please, at least move it out of line, we should not have LOG statements in headers.
> 
> The drawback is that you need to check at runtime for something which is actually known at compile time since AFAICT you only have the version with "2" hardcoded. I would keep the separate methods and have a switch(icoord) in fitMeanVertexCoord...

Ciao @ktf , 

I use now the switch, to stay general. Is this ok?

Detectors/Calibration/src/MeanVertexCalibrator.cxx
```diff
@@ -35,137 +36,289 @@ void MeanVertexCalibrator::initOutput()
...
+  double fitres = fitGaus(nbins, array, minRange, maxRange, fitValues);
+  if (fitres != -4) {
+    LOG(info) << "coordinate " << icoord << ": Fit result of full statistics => " << fitres << ". Mean = " << fitValues[1] << " Sigma = " << fitValues[2];
+  } else {
+    LOG(error) << "coordinate " << icoord << ": Fit failed with result = " << fitres;
+  }
+  switch (icoord) {
+    case 0:
+      mvo.setX(fitValues[1]);
+      mvo.setSigmaX(fitValues[2]);
```
ktf, line 296: you are missing `break;`.

chiarazampolli: What a stupidity... Sorry. I did not compile because I am working on a different branch, and wanted to do this fast.

chiarazampolli: Note that I did not add a default, as these are the only options, and default should not be there.

chiarazampolli:
Hi @shahor02 , @ktf ,
If there are no major problems with this PR, can we merge it? I would like to have it for the next O2 upgrade at P2.

shahor02 (approved):

PR #9317, opened 2022-07-02
add extra time tolerance for TRD tracks

noferini:
@shahor02 
I added an extra time tolerance of 500 ns (20 BC) for TRD tracks.
For the moment this is hardcoded, I will add the possibility to configure it via dpl


shahor02 (approved):
Thanks! Indeed, better to do this settable.

PR #9322, opened 2022-07-04
DPL GUI: prepare ancillary classes to enable remote GUI

ktf:
* Allow CORS connection on the websocket
* Setup the remote GUI if a non DPL connection on the websocket
  port happens.
* Invoke pollGUIPostRender and protect against empty frames in the
  beginning.


ktf:
Tested locally and no regression in the GUI. It should be still inactive in any case.

PR #9325, opened 2022-07-04
TPC cluster residuals + track information are dumped by aggregator

martenole:
In order to use `TimeSlotCalibration::process()` with a custom data object which is not of type `const gsl::span<const Input>` I had to remove the redundant method and the final keyword in the ITS calibration, since templated methods cannot be virtual.

The residual aggregator is now writing binned residuals, unbinned residuals and track data to 3 different trees. For each tree I create one entry per TF.

### DIFF
--- a/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
+++ b/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
@@ -102,7 +102,6 @@ class TimeSlotCalibration
 
   template <typename DATA>
   bool process(const DATA& data);
-  virtual bool process(const gsl::span<const Input> data);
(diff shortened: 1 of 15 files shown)


martenole:
Needs to be merged together with https://github.com/AliceO2Group/O2DPG/pull/512

martenole:
In order to run locally and dump everything (binned/unbinned residuals) and track data:
`o2-tpc-scdcalib-interpolation-workflow -b --send-track-data  | o2-calibration-residual-aggregator -b --output-type trackParams,unbinnedResid,binnedResid --enable-track-input`

davidrohr:
@martenole : This failed with
```
[ERROR] error parsing options of o2-calibration-residual-aggregator: unrecognised option '--outputType'
```

martenole:
Thanks for pinging me. I made a typo there, its called `--output-type`... :(

PR #9339, opened 2022-07-06
Fix for track-lenght calculation

shahor02:
For tracks almost parallel to Y axis operations with acos(phi) are prone
to produce NAN, use straight line approximation in that case


shahor02:
don't merge, it seems there is a problem in the calculation.

PR #9341, opened 2022-07-06
FIT: hotfix for trg bits

afurs:
Fixing trg bit positions. Confirmed with FEE experts


afurs:
Trivial, no need in waiting for CICD to be finished

shahor02:
@afurs have you tested it locally with FST?

afurs:
> @afurs have you tested it locally with FST?

we checked raw data from FEE. Documentation which describes bit positions was wrong.

jotwinow (approved):

PR #9344, opened 2022-07-06
DPL: avoid internal channels for oldest possible timeslice

ktf:


ktf:
@davidrohr this might explain other hangings you see.

ktf:
Test failure was just a side effect of the mocking. Fixed.

ktf:
This is for sure wrong right now, so I am merging it.

PR #9349, opened 2022-07-07
FST enable TRD digit downscaling always

martenole:
also when `SPLITTRDDIGI=0` is set. This error was introduced by me.

But of course this does not fix the issue that we are loosing tracklets when reading simulated raw data.


davidrohr (approved):
should be squash merged (I fixed some spaces, they are not required right now, but in case other options would be added to DIGIOPT in the future).
Currently github seems to have problems again, cannot merge it.
Should be squash-merger when github works again.

davidrohr:
cloned this PR in #9366 and merged that instead. For some reason I could not merge this one. When clicking on the merge button still nothing happens. Seems to be a github issue...

PR #9350, opened 2022-07-07
TRD fix tracklethcheader in mc to raw

bazinski:
- the parser is now very strict on the tracklethcheader causing a lot of the tracklets to be rejected.
- mc to raw now uses the methods in rawdata.cxx to construct the tracklethcheader correctly.

- remove option for fixsm1617
- fix pid in mc to raw to reconstruction
- change level of some error messages, bad adc mask seems to appear rather often, now info message.
- fix the construction of adcmasks in the mc to raw.
- remove the strict limit on pretrigger phase, it is above 11 far too often, code is still there and the parsing error graph updated for now.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
@@ -133,7 +133,7 @@ struct TrackletHCHeader {
     //             |   |              |    |  |  |---  1-3  stack
     //             |   |              |    |  |------  4-6  layer
     //             |   |              |    |--------  7-11 sector
-    //             |   |              |------------- 12 always 0
+    //             |   |              |------------- 12 always 1
     //             |   ----------------------------- 13-27 MCM Clock counter
     //             --------------------------------- 28-31 tracklet data format number
     uint32_t word;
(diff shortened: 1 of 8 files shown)


bazinski:
I have not finished testing this, but figured due to urgency, should get it started through the ci while I complete the tests.

martenole:
With this we are loosing not as many tracklets anymore. Now I see 23642 tracklets before conversion to raw and 21228 tracklets after conversion to raw and reading the raw files.
There are still many warnings
`[1154492:trd-datareader]: [12:36:51][WARN]  sanity check failure on TracklHCHeader of 0xc0ec4838`
And
`[1154492:trd-datareader]: [12:36:51][WARN] ***DigitMCMADCMask bad bit count maskcount:-32::31 bitscounting:4 bp: 0xf`
Seems the breaking changes were introduced in https://github.com/AliceO2Group/AliceO2/pull/9080

bazinski:
@martenole the bad adc mask will cause the digits to get dumped not the tracklets. Will investigate the error message though

martenole:
> @martenole the bad adc mask will cause the digits to get dumped not the tracklets. Will investigate the error message though

The with the digits getting dumped I understood, still that should of course not happen when we are using simulated data. And the tracklet HC header failures probably mean that tracklets are dropped, right?

bazinski:
indeed, but the tracklethcheaders are now fixed by the initial contents of this pr. trying to reproduce the raw files now

martenole:
I was using the changes from this PR for the FST and still saw the tracklet HC header check failures

bazinski:
oui it failed, but was never enforced in the decoder of raw data.

bazinski:
c3
All sector, stack and layer now accounted for.

martenole:
FST this one is OK, although the tracklet PID is still not fully matching the simulation
```
[INFO] Number of tracklets in simulation: 34542. In reco: 34542. Number of tracklets from simulation for which a match in reco was found: 28855
[INFO] Number of digits in simulation: 76622. In reco: 76622. Number of digits from simulation for which a match in reco was found: 76622
```
If I ignore the tracklet PID I get matching results
```
[INFO] Number of tracklets in simulation: 34542. In reco: 34542. Number of tracklets from simulation for which a match in reco was found: 34542
[INFO] Number of digits in simulation: 76622. In reco: 76622. Number of digits from simulation for which a match in reco was found: 76622
[WARN] The PID values stored inside the tracklets have been ignored for this comparison

```
I processed 100 raw TFs of run 519905 with this and this looks OK. Although I have not yet finished cross checking with the O2 version we were running at P2

bazinski:
That is odd, the plots of pid look decent, maybe its the spikes at the end.

bazinski:
tracklet-charges
For reference those were the plots from last night. Will go through the pid stuff again now.

bazinski:
tracklet-charges
For reference those are the plots of pid from last night.

PR #9359, opened 2022-07-07
Produce ALARM instead of exception on empty bunch filling

shahor02:


davidrohr:
Merged #9367 instead. Github seems to have some issues merging PRs. Not clear if a rebase was needed or what was the readon. Had the same issue with #9349 .

PR #9360, opened 2022-07-07
Add extra info to TPC VDrift correction objects

shahor02:
* Store in LtrCalibData reference VDrift and creation time 
* Move TPC VDtgl calibration to TPC/calibration
Use VDriftCorrFact object for CCDB, which apart from the correction/error provides reference VDrift used and stores the timestamp of the calibration

@wiechula since we are working with correction factors rather than abs. vdrift, to keep track of what was the reference vdrift used for the calibration I add it both to LtrCalibData and to the object created by the tgl-calibration.
Also adding the timestamp of creation since at run-time will need to select among different sources.

@martenole one thing we have not discussed: the residuals should be extracted wrt well-defined reference VDrift, currently it is whatever value of `GasParam.VDrift` you use to create the transform map when working with clusters. 
Eventually, we should stop playing with `GasParam.VDrift` from the command line and instead use whatever reference was stored in the LtrCalibData or VDriftCorrFact objects, I'll provide a helper for that.

### DIFF
--- a/DataFormats/Detectors/TPC/include/DataFormatsTPC/VDriftCorrFact.h
+++ b/DataFormats/Detectors/TPC/include/DataFormatsTPC/VDriftCorrFact.h
@@ -9,12 +9,12 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-/// \file LtrCalibData.h
+/// \file VDriftCorrFact.h
 /// \brief calibration data from laser track calibration
 ///
-/// This class holds the calibration output data of CalibLaserTracks
+/// This class holds the calibration output data ITS/TPC tgl difference calibration
 ///
-/// \author Jens Wiechula, Jens.Wiechula@ikf.uni-frankfurt.de
+/// \author ruben.shahoyan@cern.ch
 
 #ifndef AliceO2_TPC_VDRIFT_CORRFACT_H
 #define AliceO2_TPC_VDRIFT_CORRFACT_H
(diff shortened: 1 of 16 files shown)


martenole:
> @martenole one thing we have not discussed: the residuals should be extracted wrt well-defined reference VDrift, currently it is whatever value of GasParam.VDrift you use to create the transform map when working with clusters.
Eventually, we should stop playing with GasParam.VDrift from the command line and instead use whatever reference was stored in the LtrCalibData or VDriftCorrFact objects, I'll provide a helper for that.

Ok thanks. Indeed, at the moment it just takes the fixed GasParam.VDrift..

wiechula:
Thanks @shahor02 , this looks good!

DataFormats/Detectors/TPC/include/DataFormatsTPC/VDriftCorrFact.h
```diff
@@ -0,0 +1,36 @@
...
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file LtrCalibData.h
+/// \brief calibration data from laser track calibration
+///
+/// This class holds the calibration output data of CalibLaserTracks
+///
+/// \author Jens Wiechula, Jens.Wiechula@ikf.uni-frankfurt.de
```
wiechula:
```suggestion
/// \file VDriftCorrFact.h
/// \brief generic TPC drift velocity correction value
```

wiechula (changes requested):

wiechula (dismissed):

PR #9361, opened 2022-07-07
MRRTF-190: decrease precluster finder verbosity

aphecetche:
just store the number of times the "multiple digits on the same pad"
error is encountered and show the summary at de-initialization phase.

@pillot @aferrero2707 we might consider reusing the newly introduced ErrorMap class in other places as well. To be seen.
Note that in this PR I just changed the way the reporting is done : from "right away" to only once at the end, but we probably should actually sending the ErrorMap in the corresponding DeviceSpec instead (and quality-control it for instance), as the logs are not really that useful on the long term.

### DIFF
--- a/Detectors/MUON/MCH/Base/src/ErrorMap.cxx
+++ b/Detectors/MUON/MCH/Base/src/ErrorMap.cxx
@@ -1,0 +1,73 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include "MCHBase/ErrorMap.h"
+
+namespace o2::mch
+{
+
+uint64_t encode(uint32_t a, uint32_t b)
+{
+  uint64_t r = a;
+  r = (r << 32) & 0xFFFFFFFF00000000 | b;
+  return r;
+}
+
+std::pair<uint32_t, uint32_t> decode(uint64_t x)
+{
+  uint32_t a = static_cast<uint32_t>((x & 0xFFFFFFFF00000000) >> 32);
+  uint32_t b = static_cast<uint32_t>(x & 0xFFFFFFFF);
+  return std::make_pair(a, b);
+}
+
+void ErrorMap::add(uint32_t errorType, uint32_t id0, uint32_t id1)
+{
+  mErrorCounts[errorType][encode(id0, id1)]++;
+}
+
+void ErrorMap::forEach(ErrorFunction f) const
+{
+  for (auto errorType : mErrorCounts) {
+    for (auto errorCounts : errorType.second) {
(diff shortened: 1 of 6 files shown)


Detectors/MUON/MCH/Base/src/ErrorMap.cxx
```diff
@@ -0,0 +1,73 @@
...
+
+void ErrorMap::forEach(ErrorFunction f) const
+{
+  for (auto errorType : mErrorCounts) {
+    for (auto errorCounts : errorType.second) {
+      uint64_t count = errorCounts.second;
+      uint64_t id = errorCounts.first;
+      auto p = decode(id);
+      uint32_t id0 = p.first;
+      uint32_t id1 = p.second;
```
pillot: to be more concise you could have used:
auto [id0, id1] = decode(id);

Detectors/MUON/MCH/Base/src/ErrorMap.cxx
```diff
@@ -0,0 +1,73 @@
...
+  };
+  em.forEach(countErrorTypes);
+  return errorTypes.size();
+}
+
+uint64_t totalNumberOfErrors(const ErrorMap& em)
+{
+  uint64_t n{0};
+  auto countErrors = [&n](uint32_t errorType,
+                          uint32_t id0, uint32_t id1, uint64_t count) {
```
pillot: shouldn't the unused variables be commented like above to avoid warning?

Detectors/MUON/MCH/PreClustering/src/PreClusterFinder.cxx
```diff
@@ -67,6 +67,11 @@ void PreClusterFinder::init()
 void PreClusterFinder::deinit()
 {
   /// clear the internal structure
+  auto print = [](uint32_t errorType, uint32_t deId, uint32_t padid,
+                  uint64_t count) {
+    LOGP(warning, "multiple digits on the same pad (DE {} pad {}): seen {} time{}", deId, padid, count, count > 1 ? "s" : "");
```
pillot, line 72: also here errorType should be commented I think

pillot (commented):
I agree, the ErrorMap should probably be propagated along the workflow. Also the error code should probably be defined globally so that we could end up with one single message cumulating all the errors from the entire reconstruction chain.

Concerning this particular error, note that this is a real error that is expected to disappear at some point (or become a fatal). Currently the pileup digits are just skipped in the preclustering, which is not the correct way to handle them. Instead, they should be merged properly in the time clustering.

Detectors/MUON/MCH/Base/src/ErrorMap.cxx
aphecetche: yep. will change.

Detectors/MUON/MCH/Base/src/ErrorMap.cxx
aphecetche: yes, they should.

pillot (approved):

davidrohr:
@aphecetche : This crashes the FST with
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  chargeSum should be between 0 and 1048575 but is 1147179
```

pillot:
> @aphecetche : This crashes the FST with
> 
> ```
> terminate called after throwing an instance of 'std::invalid_argument'
>   what():  chargeSum should be between 0 and 1048575 but is 1147179
> ```

Hi @davidrohr , this is not related to this PR. It should be fixed by #9365.

PR #9362, opened 2022-07-07
Fix ITS aggregator for EoS operations at P2

iravasen:
@shahor02 this should apply the fixes we discuss to operate with EoS at P2. I have one question:
- this part https://github.com/AliceO2Group/AliceO2/blob/3c74bf64226e93eb5ca102a115432a44a9d29dc1/Detectors/ITSMFT/ITS/workflow/src/ThresholdAggregatorSpec.cxx#L69-L87 take pieces of strings from calibrators and merge them in 1 big string. Here I did not put any protection for the garbage arriving at EoS. Might this be an issue? In principle these are DCSConfigObject_t which are not filled at EoS time by calibrators and hence I guess the aggregator should not do anything. 

I keep it as WIP because I want to test it locally just to be sure that everything works as expected (at least locally)

### DIFF
--- a/Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/ThresholdAggregatorSpec.h
+++ b/Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/ThresholdAggregatorSpec.h
@@ -92,10 +92,10 @@ class ITSThresholdAggregator : public Task
   // confDB version
   short int mDBversion = -1;
 
-  // DataTakingContext used to get lhcperiod 
+  // DataTakingContext used to get lhcperiod
   o2::framework::DataTakingContext mDataTakingContext{};
 
-  // Timing info used to get run number 
+  // Timing info used to get run number
   o2::framework::TimingInfo mTimingInfo{};
 };
 
--- a/Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/ThresholdAggregatorSpec.h
+++ b/Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/ThresholdAggregatorSpec.h
@@ -33,6 +33,8 @@
 #include "CCDB/CcdbApi.h"
 #include "DataFormatsDCS/DCSConfigObject.h"
 #include "Framework/InputRecordWalker.h"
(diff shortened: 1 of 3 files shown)


shahor02:
Hi @iravasen 
If input appears to be bogus in the case of EoS, I am not sure what it will contain. To be on the safe side, I would simply avoid processing the input if e.g. `if (pc.services().get<o2::framework::TimingInfo>().firstTFOrbit == -1U) {...`

iravasen:
Hi @shahor02, thanks, implemented and tested. I have only one last doubt which came to my mind. With the end-of-stream, the DPL channels `ITS/TSTR/<subspec>` and `ITS/PIXTYP/<subspec>` sends good data to aggregator when endOfStream function is called on the calibrators. These data are averages of chip parameters which get merged then in the final ccdb object. Can it be that with that protection for `firstTFOrbit`, these data are skipped when doing the merge? Or they come to aggregator before this bogus timeslice?

shahor02:
Hm, this I am not sure, better @ktf explains what TimingInfo is supposed to be in the DataHeader and DataProcessingHeader of the message sent explicitly from the endOfStream method.

ktf:
So the flow is as follow:

1. Input data with the associated data header populates the TimingInfo
2. user creates some output data
3. the information in the TimingInfo is used to create the header for the object found in 2.

In principle the message with the bogus timeslice is always the end of stream, so it would come last.

shahor02:
> Input data with the associated data header populates the TimingInfo

But what is the input data in the EndOfStreamContext, where it takes the TimingInfo from?

ktf:
good point…

Indeed there is no InputRecord in the EndOfStreamContext. However, the way
things are done right now, it will simply reuse the TimingInfo associated
to the last timeframe **processed**.


Ciao,
Giulio

shahor02:
ok, in this case there should be no problem with skipping the input if supplied orbit is -1.

iravasen:
Thanks a lot @ktf and @shahor02, removing WIP

davidrohr:
@shahor02 : I cannot squash-merge again. Can you try?

PR #9363, opened 2022-07-07
DPL: allow overriding orbit reset time

ktf:


ktf:
@davidrohr @shahor02 as discussed the other day...

davidrohr:
@ktf : Thx for the PR!

Though, after thinking about it again today, I am no longer sure if overriding the orbitResetTime is enough.
There might be devices requesting the orbitResetTime object directly. Can we make sure that in that case they get an object with the overridden orbitResetTime from the internal-dpl-ccdb-backend?

Second, but unrelated, today I found out that I have the same problem with the FST from DD as in SYNTHETIC run, but there I cannot set the FMQ property. Could we add an env variable as alternative to override the time?

ktf:
> Though, after thinking about it again today, I am no longer sure if overriding the orbitResetTime is enough. There might be devices requesting the orbitResetTime object directly. Can we make sure that in that case they get an object with the overridden orbitResetTime from the internal-dpl-ccdb-backend?

Is it just a matter of always adding the OrbitResetTime as object which is always produced by CCDBFetcher? I think the object is already in the cache in any case...

> Second, but unrelated, today I found out that I have the same problem with the FST from DD as in SYNTHETIC run, but there I cannot set the FMQ property. Could we add an env variable as alternative to override the time?

In principle we could have that every device has a --orbit-reset-time ConfigParamSpec like we do for other stuff. That would fix it, no? That way you could even override it on a per device basis. Of course if you prefer the environment variable, we can do that as well.

shahor02:
Will there be a performance penalty if we add `dph.creation` override option in the raw-proxy?

ktf:
@shahor02 I guess not, assuming we fetch it only at the beginning of the processing and not for each timeframe.

davidrohr:
> In principle we could have that every device has a --orbit-reset-time ConfigParamSpec like we do for other stuff. That would fix it, no? That way you could even override it on a per device basis. Of course if you prefer the environment variable, we can do that as well.

I don't fully understand how that would work. If different devices set different --orbit-reset-time values: I assume there is still only one ccdb fetcher. What orbit reset time would it use? Or would it fetch the objects twice with 2 different times?



> Will there be a performance penalty if we add `dph.creation` override option in the raw-proxy?

I don't think there will be a large overhead.
But if we override dph, we might have a problem on the FLP, since then stfsender will receive a different dpl creation time compared to what stfbuilder injected.

shahor02:
>> But if we override dph, we might have a problem on the FLP, since then stfsender will receive a different dpl creation time compared to what stfbuilder injected.

Does the StfSender check the creation time? It may receive from the DPL outputs which it StfBuilder has never injected, what it does in this case?

davidrohr:
@ironMann can comment, but IIRC it is used for computing delays.

I think we have to decide between 3 paths:
1. we adjust the creation time for replay (might need changes in DD).
2. we have another field in the DPH for the runTime, so the creation time is untouched, and used only by the data transport, and we use the run time for everything processing related. The raw-proxy could clone the creationTime into the runTime, and optionally override it.
3. we use another mechanism, like env variable, or FMQ properties, like @ktf has done in this PR. And I thought we had decided to go this way, but principally I am open for either solution.

ktf:
> I don't fully understand how that would work. If different devices set different --orbit-reset-time values: I assume there is still only one ccdb fetcher. What orbit reset time would it use? Or would it fetch the objects twice with 2 different times?

No, indeed it does not makes much sense. We can just add it as an option to the CCDB fetcher. I think it's merely not exposed as a command line argument.

PR #9365, opened 2022-07-08
MCH: fix digit charge sum exceeding limit

pillot:
I also decoupled a bit the response and the digitization for the number of samples.

### DIFF
--- a/Detectors/MUON/MCH/Simulation/src/DEDigitizer.cxx
+++ b/Detectors/MUON/MCH/Simulation/src/DEDigitizer.cxx
@@ -13,6 +13,7 @@
 
 #include "MCHSimulation/DigitizerParam.h"
 #include "DetectorsRaw/HBFUtils.h"
+#include <algorithm>
 #include <cmath>
 #include <random>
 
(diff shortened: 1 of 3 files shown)


Detectors/MUON/MCH/Simulation/src/DEDigitizer.cxx
```diff
@@ -112,11 +113,18 @@ void DEDigitizer::extractDigitsAndLabels(std::vector<Digit>& digits,
...
-      digits.emplace_back(mDeId, padid, adc, time, nSamples);
+      nSamples = std::min(nSamples, 0x3FFU); // the number of samples must fit within 10 bits
+      bool saturated = false;
+      // the charge sum must fit within 20 bits
+      // FIXME: we should better handle charge saturation here
+      if (adc > 0xFFFFFU) {
+        adc = 0xFFFFFU;
+        saturated = true;
+      }
+      digits.emplace_back(mDeId, padid, adc, time, nSamples, saturated);
```
aphecetche, line 127: hum, now that nSamples is a uint32_t there's a mismatch in the call to the Digit actor which expects a uint16_t, isn't it ?

pillot: same for the time. Is it needed to do an explicit cast ?

aphecetche: Will play a bit with clang warnings to see. But that does not prevent to merge this one I would say.

aphecetche (approved):

PR #9368, opened 2022-07-08
[EMCAL-558] important bugfix for EMCal clusterizer

fjonasALICE:
missing cast by reference caused usage of local copy of digit which goes out of scope. This means that previous version of clusterizer clustered all cells in proximity to each other with E>cellthreshold and seed cell E>cellseed, but all subsequent comparison always return true (no gradient cut, no requirement subsequent cell has to have less energy than seed etc)

Clusterizer is now fixed and validated with comparisons to AliRoot clusterizer

### DIFF
--- a/Detectors/EMCAL/reconstruction/include/EMCALReconstruction/Clusterizer.h
+++ b/Detectors/EMCAL/reconstruction/include/EMCALReconstruction/Clusterizer.h
@@ -59,7 +59,7 @@ class Clusterizer
   };
 
   struct InputwithIndex {
-    InputType* mInput;
+    const InputType* mInput;
     ClusterIndex mIndex;
   };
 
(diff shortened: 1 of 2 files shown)


mfasDa (dismissed):
Looks good! @shahor02 could you please merge once the CI passes?

mfasDa:
As can be seen in https://alice.its.cern.ch/jira/browse/EMCAL-558 the clusterizer in AliRoot (AliEMCALClusterizerV3) and O2 (o2::emcal::Clusterizer) now give identical results on converted data from run2.

davidrohr:
@fjonasALICE : could you squash your 2 commits and force-push?
There is a problem with github right now, and I cannot squash-merge the PR.

davidrohr (approved):

shahor02:
Hi @davidrohr 
Actually I can squash merge.

PR #9372, opened 2022-07-08
o2-eve-export-workflow: change defaults for json file location and nu…

jmyrcha:
…mber of stored files


EventVisualisation/Workflow/src/O2DPLDisplay.cxx
```diff
@@ -50,10 +50,10 @@ void customize(std::vector<o2::framework::CallbacksPolicy>& policies)
 void customize(std::vector<ConfigParamSpec>& workflowOptions)
 {
   std::vector<o2::framework::ConfigParamSpec> options{
-    {"jsons-folder", VariantType::String, "jsons", {"name of the folder to store json files"}},
+    {"jsons-folder", VariantType::String, "/scratch/services/ed/jsons", {"name of the folder to store json files"}},
```
davidrohr: I would not change the default for the storage here. It will break when users run the export, since that folder will not exist / they will not be allowed to write there.
The export folder is already set by the topology generation script anyway.

PR #9377, opened 2022-07-08
TPC digit: some optimisation for PadRegion calculations

sawenzel:
The "findPad" function is called billions of times in digitization
and amounts to ~10% of time according to valgrind.

It therefore makes sense to optimise this a bit:
a) avoid divisions in favour of faster multiplications
b) avoid calling a non-inlined wrapper of the findPad function


sawenzel:
@wiechula @davidrohr : I tried to directly make `PadRegionInfo::findPad(LocalPointXX)` inline but failed because of some GPU code. Maybe you have an idea how to do it.

davidrohr:
OK, so you want to move findPad from the .cxx to .h?
I can have a look.

PR #9383, opened 2022-07-09
MatchGlobalFwd fix: do not throw on empty bunch filling (unused)

shahor02:
Apparently setting of BunchFilling was copy/pasted from other matchers, although currently it is not
used. Substituted exception by error message in case empty BunchFilling is provided


shahor02:
trivial, w/o this synthetic run may crash, merging

PR #9386, opened 2022-07-10
Add protection against huge number of cells

mpuccio:
Hi @shahor02,

still to be tested locally (running it now...), but you can get the gist.
Let me know if you have better ideas.

Cheers,
Max﻿

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/src/Tracker.cxx
+++ b/Detectors/ITSMFT/ITS/tracking/src/Tracker.cxx
@@ -55,15 +55,29 @@ void Tracker::clustersToTracks(std::function<void(std::string s)> logger, std::f
     total += evaluateTask(&Tracker::initialiseTimeFrame, "Timeframe initialisation",
                           logger, iteration, mMemParams[iteration], mTrkParams[iteration]);
     total += evaluateTask(&Tracker::computeTracklets, "Tracklet finding", logger);
+    logger(fmt::format("\t- Number of tracklets: {}", mTimeFrame->getNumberOfTracklets()));
     if (!mTimeFrame->checkMemory(mTrkParams[iteration].MaxMemory)) {
       error("Too much memory used during trackleting, check the detector status and/or the selections.");
       break;
     }
+    float trackletsPerCluster = mTimeFrame->getNumberOfClusters() > 0 ? float(mTimeFrame->getNumberOfTracklets()) / mTimeFrame->getNumberOfClusters() : 0.f;
+    if (trackletsPerCluster > mTrkParams[iteration].TrackletsPerClusterLimit) {
+      error(fmt::format("Too many tracklets per cluster ({}), check the detector status and/or the selections.", trackletsPerCluster));
+      break;
+    }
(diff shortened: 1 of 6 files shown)


mpuccio:
OK, locally it does not break anything on a single TF, I will now try to reconstruct several. We should just bear in mind that the limit I put (factor 10) it is not tuned, we could have a look only at the timings and not at the number of artefacts so I put something out of gut feeling. The parameter is tunable with the command line interface in case it is not effective (or it is disruptive) during data taking.

Detectors/ITSMFT/ITS/tracking/src/Tracker.cxx
```diff
@@ -64,12 +65,21 @@ void Tracker::clustersToTracks(std::function<void(std::string s)> logger, std::f
       error("Too much memory used during cell finding, check the detector status and/or the selections.");
       break;
     }
+    cellsPerCluster = mTimeFrame->getNumberOfClusters() > 0 ? float(mTimeFrame->getNumberOfCells()) / mTimeFrame->getNumberOfClusters() : 0.f;
+    if (mNumberOfRuns > 9 && cellsPerCluster > mTrkParams[iteration].CellsPerClusterLimit * mAverageCellsPerClusters) {
```
shahor02: Is clustersToTracks call meant per whole TF (with possibly preselected ROFs for reject)? In PbPb we can have quite large variation of multiplicity, which may also change the non-linear cells/clusters. Should not we have a cut on the abs. fraction of cells/cluster (then, if you want, also very relaxed cut of its factor wrt average)? Note that this protection will not be effective if the anomalous TF happens to the within first 9 iterations.

Also, if I get it right, the iteration is done with a particular set of cuts after suppressing the clusters of the previous iteration, right? Then each level iteration may have quite different cellsPerCluster. Should not per-iteration-type mAverageCellsPerClusters be calculated?
Finally, you apply the cut on the cell level, but already for the trackleting the time jumps by factor 100. Should we apply cut on this level?

mpuccio:
Hi @shahor02, I made the cuts "absolute" and added the string output necessary to tune it in case it does not work out of the box (right now I put 10 times the ratios I see in a local pp simulation). I added the selections also at the level of tracklets.

shahor02:
> right now I put 10 times the ratios I see in a local pp simulation

Thanks, pp simulations at which rate? Would not it affect synthetic pbpb runs? Do have numbers for e.g. 400 kHz pp and 50 kHz pbpb?

mpuccio:
No, I don't have those number as I don't have such simulations, the local sim is pp@50kHz. However with the new printout we can set easily the threshold.
Note that this will not make any run fail, we would just get errors from ITS and the tracking would not run.

mpuccio:
OK, I tried locally with 500kHz and the values put there are in line with those of the previous simulation. I updated the values since I miscalculated them... now the tolerance is factor 20 wrt 500kHz pp. I will simulate some Pb-Pb events and check also in that case.

shahor02 (approved):

PR #9387, opened 2022-07-10
Enable configuration of maxTrackletsPerCluster

mconcas:
@shahor02: the option to limit the per-cluster combinatorics was there already, now can be set via command line.
I think `ITSVertexerParam.maxTrackletsPerCluster=??` should do the work.

Still waiting for building on my local pc to test. But just to understand if this is what we are talking about.

### DIFF
--- a/Detectors/ZDC/simulation/src/Detector.cxx
+++ b/Detectors/ZDC/simulation/src/Detector.cxx
@@ -2630,7 +2630,7 @@ bool Detector::FastSimToHits(const Ort::Value& response, const TParticle& partic
       int currentMediumid = determineMediumID(detector, x, y);
       // LOG(info) << " x " << x << " y " << y << " sec " << sector << " medium " << currentMediumid;
       // Model output needs to be converted with exp(x)-1 function to be valid
-      int nphe = (int)std::expm1(pixels[Nx * x + y]);
+      int nphe = (int)std::expm1(pixels[Nx * y + x]);
 
       if (nphe > 0) {
         float trackenergy = 0; // energy of the primary (need to fill good value)


Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
```diff
@@ -143,15 +143,17 @@ void VertexerTraits::computeTracklets()
...
     trackleterKernelSerial<TrackletMode::Layer1Layer2>(
       mTimeFrame->getClustersOnLayer(rofId, 2),
       mTimeFrame->getClustersOnLayer(rofId, 1),
       mTimeFrame->getIndexTable(rofId, 2).data(),
       mVrtParams.phiCut,
       mTimeFrame->getTracklets()[1],
       mTimeFrame->getNTrackletsCluster(rofId, 1),
-      mIndexTableUtils);
+      mIndexTableUtils,
+      mVrtParams.maxTrackletsPerCluster);
```
shahor02: Thanks, yes, but should not you substitute https://github.com/AliceO2Group/AliceO2/blob/b8c5a0925f225d26a80aabed0cd68d95a3ce95cb/Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx#L45 by the `mVrtParams.maxTrackletsPerCluster` ?

BTW, is 2e3 tracklets/cluster based on some study? Looks quite high.

mconcas: > Thanks, yes, but should not you substitute
> 
> https://github.com/AliceO2Group/AliceO2/blob/b8c5a0925f225d26a80aabed0cd68d95a3ce95cb/Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx#L45
> 
> by the `mVrtParams.maxTrackletsPerCluster` ?

Isn't it done at line 156, or am I missing the point?

> BTW, is 2e3 tracklets/cluster based on some study? Looks quite high.

It is high indeed on purpose as the number actually comes from some super-safe assumption done in case of limited memory (gpu case).
There the plan was to have a hard limit but to keep all the tracklets.

If we now want to have a reasonable default indeed this must be studied/estimated.

Also this number is quite obsolete as even the GPU approach now does not afford anymore so many tracklets (due to the rof->tf change of approach)

shahor02: > Isn't it done at line 156, or am I missing the point?

Right, sorry, confused it with local contant.

PR #9393, opened 2022-07-11
dev: fake ctp readout

lietava:
Hi @shahor02 , so what I understand that can be done is to replace line:
https://github.com/AliceO2Group/AliceO2/blob/912313ec7f515b1b77a66dadc0dd2220dd7190ae/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L1213
to function call which takes information from trigger detector recPoints and calculates ctpDigits.
Can it work like that ?
Cheers, Roman.
cc: @mpuccio


shahor02:
yes, you will need to convert digit to trigger inputs and emulate CTP decision.

lietava:
@shahor02 : can it be done like this (just function is there) ?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -92,6 +92,13 @@ using SMatrix55Sym = ROOT::Math::SMatrix<double, 5, 5, ROOT::Math::MatRepSym<dou
 namespace o2::aodproducer
 {
 
+void AODProducerWorkflowDPL::CTPReadout(const o2::globaltracking::RecoContainer& recoData, gsl::span<o2::ctp::CTPDigit> ctpDigits)
```
shahor02: ctpDigits is what you fill when emulating the trigger, right? Then it should be a `std::vector<o2::ctp::CTPDigit> &ctpDigits`.
The rest is ok.

lietava: Hi @shahor02 ,
yes, ctpDigita are filled.
I have three questions:
1)  can I get runnumber inside AODProducerWorkflowSpec ? ( I need it for getting CTPconfig from CCDB)
2)  to know if PHS and EMC are triggered I use 
const auto& caloEMCCellsTRGR = recoData.getEMCALTriggers();
const auto& caloPHOSCellsTRGR = recoData.getPHOSTriggers();
I guess these are produced only for NC which were also readout ?
3) what would be the simplest way to know if TRD is readout ?
Thanks, Roman.

lietava: Hi @shahor02 , no need to comment on my questions in previous post. With the help of @mpuccio it is answered. Cheers, R.

shahor02: Sorry, actually missed your message...

lietava:
HI @shahor02 , I implemented software ctpreadout - so far simplified version which should create readout for classes which contains TRD.
It looks simple -  please, have a look and let me know any comments and how to test.
Cheers, Roman.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -92,6 +94,43 @@ using SMatrix55Sym = ROOT::Math::SMatrix<double, 5, 5, ROOT::Math::MatRepSym<dou
...
+    if (t0triggers.getVertex()) {
+      uint64_t globalBC = ft0RecPoint.getInteractionRecord().toLong();
+      uint64_t classmask = ctpcfg.getClassMaskForInput("MTVTX");
+      bcsMapT0triggers[globalBC] = classmask;
+    }
+  }
+  // find trd redaout and add CTPDigit if trigger there
+  for (auto& trdrec : triggerrecordTRD) {
+    uint64_t globalBC = trdrec.getBCData().toLong();
+    bcsMapTRDreadout[globalBC] = 1;
```
shahor02: Currently TRD decoder produces lot of spurious triggers (with impossible BC, decreasing orbit, mostly 0 tracklets).
In the CTF decoding I am filtering them out, but the AOD can be created also from the raw data directly. I would suggest to check and skip such triggers. See https://github.com/AliceO2Group/AliceO2/blob/5d97c8b4764e3075da632ecb0f87e9c49089f8a8/Detectors/TRD/reconstruction/include/TRDReconstruction/CTFCoder.h#L89-L108.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -92,6 +94,43 @@ using SMatrix55Sym = ROOT::Math::SMatrix<double, 5, 5, ROOT::Math::MatRepSym<dou
...
+    if (bcsMapT0triggers.count(globalBC) > 0) {
+      o2::ctp::CTPDigit ctpdig;
+      ctpdig.intRecord.setFromLong(globalBC);
+      ctpdig.CTPClassMask = bcsMapT0triggers[globalBC];
+      ctpDigits.push_back(ctpdig);
+    } else {
```
shahor02: should be faster to do
```
auto t0entry = bcsMapT0triggers.find(globalBC);
if (t0entry != bcsMapT0triggers.end()) {
   auto& ctpdig = ctpDigits.emplace_back();
   ctpdig.intRecord.setFromLong(globalBC);
   ctpdig.CTPClassMask = t0entry->second;
}
```

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -92,6 +94,43 @@ using SMatrix55Sym = ROOT::Math::SMatrix<double, 5, 5, ROOT::Math::MatRepSym<dou
 namespace o2::aodproducer
 {
 
+void AODProducerWorkflowDPL::createCTPReadout(const o2::globaltracking::RecoContainer& recoData, std::vector<o2::ctp::CTPDigit>& ctpDigits, int runNumber)
+{
+  // Extraxt CTP Config from CCDB
+  o2::ctp::CTPConfiguration ctpcfg = o2::ctp::CTPRunManager::getConfigFromCCDB(-1, std::to_string(runNumber)); // how to get run
```
shahor02: getConfigFromCCDB still uses BasicCCDBManager which we are eliminating from the production workflows. Instead of using this method, use DPL CCDB functionality, see e.g. https://docs.google.com/presentation/d/1WPNixo57zAn_gM6vMCIBOHs-DZqcKAkLX0CwUgO_VYg/edit#slide=id.g10e79f44b99_0_7

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1209,9 +1249,14 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
   auto caloPHOSCells = recoData.getPHOSCells();
   auto caloPHOSCellsTRGR = recoData.getPHOSTriggers();
-
   auto ctpDigits = recoData.getCTPDigits();
+  const auto& tinfo = pc.services().get<o2::framework::TimingInfo>();
 
+  if (mCTPReadout == 1) {
+    std::vector<o2::ctp::CTPDigit> ctpDigitsCreated;
+    createCTPReadout(recoData, ctpDigitsCreated, tinfo.runNumber);
+    ctpDigits = ctpDigitsCreated;
```
shahor02: `ctpDigitsCreated` is deleted once you go outside `{}` so the span ctpDigits will be invalid. You should move it before the `if`.

shahor02 (changes requested):
Hi @lietava 
Looks fine but see comments below.

lietava:
@shahor02 : I addressed all your issues. Thanks, R.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1209,9 +1262,14 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
   auto caloPHOSCells = recoData.getPHOSCells();
   auto caloPHOSCellsTRGR = recoData.getPHOSTriggers();
-
   auto ctpDigits = recoData.getCTPDigits();
+  const auto& tinfo = pc.services().get<o2::framework::TimingInfo>();
 
+  if (mCTPReadout == 1) {
+    std::vector<o2::ctp::CTPDigit> ctpDigitsCreated;
+    createCTPReadout(recoData, ctpDigitsCreated, pc);
+    ctpDigits = std::move(ctpDigitsCreated);
```
shahor02: Not sure this will work: ctpDigits is a gsl/span while ctpDigitsCreated is a vector. Did you compile and test how it works? I would simple take the ctpDigitsCreated outside of the `if` and assign its span to `ctpDigits`

shahor02 (commented):
Please see the comment below.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
lietava: It did compiled. I did not test. I modified as you suggested.

lietava:
@shahor02 : in build/O2/fullCl I got error
Unable to find object CTP/Config/Config/1550600800000.
How can I debug the full thing ?
Cheers, Roman.

shahor02:
@lietava It is what it says: you should upload (via jira upload request) a CCDB object for "un-anchored" MC ranges used by the tests, see the end of https://twiki.cern.ch/twiki/bin/viewauth/ALICE/O2DPGMCSamplingSchema

lietava:
Hi @mpuccio and @shahor02 , I created jira ticket for CTPConfiguration for pp MC.
https://alice.its.cern.ch/jira/browse/O2-3101
Cheers, Roman.

lietava:
@shahor02 it seems to me errors are not on ctp side ?

shahor02:
Hi @lietava 

Actually, fullCI error is genuine: FST for PbPb does not find the CTP/Config/Config object for non-anchored MC.
I see that there is an object for pp MC from https://alice.its.cern.ch/jira/browse/O2-3101, but we need objects for the https://twiki.cern.ch/twiki/bin/viewauth/ALICE/O2DPGMCSamplingSchema?sortcol=0;table=1;up=0 periods.

lietava:
> Hi @lietava
> 
> Actually, fullCI error is genuine: FST for PbPb does not find the CTP/Config/Config object for non-anchored MC. I see that there is an object for pp MC from https://alice.its.cern.ch/jira/browse/O2-3101, but we need objects for the https://twiki.cern.ch/twiki/bin/viewauth/ALICE/O2DPGMCSamplingSchema?sortcol=0;table=1;up=0 periods.

Hi @shahor02 , ok, I see I created only the one with validity for pp. Thanks, R.

lietava:
@shahor02 : it went through fullCl, waiting only for O2/o2 ?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mfasDa: @lietava Sorry, I see this only now, would have been good if you would have pinged me and the PHOS experts directly, I did not receive any type of notification. For EMCAL yes, we create one trigger record per event which is read out, independent of the payload size. Note that we correct the interaction record in the synchronous reconstruction by the L0_LM delay. Since the CTF  for EMCAL contains also the calibration triggers we have also the trigger type in our trigger record object, but since you correlate later with TVX this information is probably not needed.

PR #9404, opened 2022-07-12
Adding debug streamer for reconstruction

matthias-kleiner:
Multithreading:
each thread creates its own debug file.
These files can be merged using hadd.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,11 @@ if (DPL_ENABLE_TRACING)
   add_definitions(-DDPL_ENABLE_TRACING)
 endif()
 
+if(DEFINED ENV{ENABLE_DEBUG_STREAMER})
+  add_definitions(-DDEBUG_STREAMER)
+  MESSAGE(STATUS "Enable debug streamer")
+endif()
+
 # include macros and functions that are used in the following subdirectories'
 # CMakeLists.txt
 include(O2AddExecutable)
--- a/Common/Utils/CMakeLists.txt
+++ b/Common/Utils/CMakeLists.txt
@@ -23,6 +23,7 @@ o2_add_library(CommonUtils
                        src/BoostHistogramUtils.cxx
                        src/NameConf.cxx
                        src/IRFrameSelector.cxx
+                       src/DebugStreamer.cxx
                PUBLIC_LINK_LIBRARIES ROOT::Hist ROOT::Tree Boost::iostreams O2::CommonDataFormat O2::Headers
                                      FairLogger::FairLogger)
 
@@ -45,7 +46,8 @@ o2_target_root_dictionary(CommonUtils
                                   include/CommonUtils/VerbosityConfig.h
                                   include/CommonUtils/FileFetcher.h
                                   include/CommonUtils/NameConf.h
-                                  include/CommonUtils/IRFrameSelector.h)
+                                  include/CommonUtils/IRFrameSelector.h
+                                  include/CommonUtils/DebugStreamer.h)
 
 o2_add_test(TreeStream
             COMPONENT_NAME CommonUtils
--- a/Common/Utils/include/CommonUtils/DebugStreamer.h
+++ b/Common/Utils/include/CommonUtils/DebugStreamer.h
@@ -1,0 +1,139 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file DebugStreamer.h
+/// \brief Definition of class for writing debug informations
+/// \author Matthias Kleiner <mkleiner@ikf.uni-frankfurt.de>
+
+#ifndef ALICEO2_TPC_DEBUGSTREAMER_H_
+#define ALICEO2_TPC_DEBUGSTREAMER_H_
+
+#include "GPUCommonDef.h"
+#if !defined(GPUCA_GPUCODE) && !defined(GPUCA_STANDALONE)
+#include "CommonUtils/ConfigurableParamHelper.h"
+#if defined(DEBUG_STREAMER)
+#include "CommonUtils/TreeStreamRedirector.h"
+#endif
+#endif
+
+namespace o2::utils
+{
+
+/// struct defining the flags which can be used to check if a certain debug streamer is used
+enum StreamFlags {
+  streamdEdx = 1 << 0, ///< stream corrections and cluster properties used for the dE/dx
+};
+
+#if !defined(GPUCA_GPUCODE) && !defined(GPUCA_STANDALONE)
+
+inline StreamFlags operator|(StreamFlags a, StreamFlags b)
+{
+  return static_cast<StreamFlags>(static_cast<int>(a) | static_cast<int>(b));
+}
+inline StreamFlags operator&(StreamFlags a, StreamFlags b) { return static_cast<StreamFlags>(static_cast<int>(a) & static_cast<int>(b)); }
+inline StreamFlags operator~(StreamFlags a) { return static_cast<StreamFlags>(~static_cast<int>(a)); }
+
+/// struct for setting and storing the streamer level
+struct ParameterDebugStreamer : public o2::conf::ConfigurableParamHelper<ParameterDebugStreamer> {
+  StreamFlags StreamLevel{}; /// flag to store what will be streamed
+  O2ParamDef(ParameterDebugStreamer, "DebugStreamerParam");
+};
+
+#endif
+
+/// class to enable streaming debug information to root files
+class DebugStreamer
+{
+
+// CPU implementation of the class
+#if !defined(GPUCA_GPUCODE) && !defined(GPUCA_STANDALONE) && defined(DEBUG_STREAMER)
+ public:
+  /// set the streamer i.e. create the output file and set up the streamer
+  /// \param outFile output file name without .root suffix
+  /// \param option RECREATE or UPDATE
+  void setStreamer(const char* outFile, const char* option);
+
+  /// \return returns if the streamer is set
+  bool isStreamerSet() const { return mTreeStreamer ? true : false; }
+
+  /// \return returns streamer object
+  o2::utils::TreeStreamRedirector& getStreamer() { return *mTreeStreamer; }
+
+  /// \return returns streamer level i.e. what will be written to file
+  static StreamFlags getStreamFlags() { return ParameterDebugStreamer::Instance().StreamLevel; }
+
+  ///< return returns unique ID for each CPU thread to give each thread an own output file
+  static size_t getCPUID();
+
+  /// \return returns number of trees in the streamer
+  int getNTrees() const;
+
+  /// \return returns an unique branch name which is not already written in the file
+  /// \param tree name of the tree for which to get a unique tree name
+  std::string getUniqueTreeName(const char* tree) const;
+
+  /// set directly the debug level
+  static void setStreamFlags(const StreamFlags streamFlags) { o2::conf::ConfigurableParam::setValue("DebugStreamerParam", "StreamLevel", static_cast<int>(streamFlags)); }
+
+  /// enable specific streamer flag
+  static void enableStream(const StreamFlags streamFlag);
+
+  /// disable a specific streamer flag
+  static void disableStream(const StreamFlags streamFlag);
+
+  /// check if streamer for specific flag is enabled
+  static bool checkStream(const StreamFlags streamFlag) { return ((getStreamFlags() & streamFlag) == streamFlag); }
+
+  /// merge trees with the same content structure, but different naming
+  /// \param inpFile input file containing several trees with the same content
+  /// \param outFile contains the merged tree from the input file in one branch
+  /// \param option setting which is used for the merging
+  static void mergeTrees(const char* inpFile, const char* outFile, const char* option = "fast");
+
+ private:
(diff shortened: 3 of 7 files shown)


matthias-kleiner:
@miranov25 this is a simple example of the debug streaming we discussed. I implemented it for now only for the dE/dx to debug the applied corrections, but it could be put to other places in the code in a similar way. The debug flags can be set from the command line when running the reconstruction e.g. for the dE/dx: `o2-tpc-reco-workflow -infile tpcdigits.root --output-type "tracks" --configKeyValues "DebugStreamerParam.StreamLevel=1"`

@davidrohr is something like this okay for you? I guess it has no impact on the GPU reconstruction performance and for the CPU reconstruction it is just an `if` that is checked.

davidrohr:
Hi @matthias-kleiner : will comment in more detail later after the TPC GPU ZS decoding is fixed.
But some comments already here:

We have already https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/GPUTracking/Debug/GPUROOTDump.h, which has a similar functionality. Would that work for your?

Otherwise: I think something generic should rather be moved to CommonUtils or so, not inside the GPU subfolder.

And I'd prefer if one can disable it completely at compile time if possible, so that it does not affect the runtime also on the CPU.

And lastly, I'd prefer an implementation where the `#ifdef`s are not spread through the application code, but are locallized in the class implementation. Basically, when disabled the header would return an empty class which does nothing.

matthias-kleiner:
> Hi @matthias-kleiner : will comment in more detail later after the TPC GPU ZS decoding is fixed. But some comments already here:
> 
> We have already https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/GPUTracking/Debug/GPUROOTDump.h, which has a similar functionality. Would that work for your?
> 
> Otherwise: I think something generic should rather be moved to CommonUtils or so, not inside the GPU subfolder.
> 
> And I'd prefer if one can disable it completely at compile time if possible, so that it does not affect the runtime also on the CPU.
> 
> And lastly, I'd prefer an implementation where the `#ifdef`s are not spread through the application code, but are locallized in the class implementation. Basically, when disabled the header would return an empty class which does nothing.

Thanks for the comments. I wasnt aware that something like this already exists. I will take a look at it and see if we can use it.

miranov25:
Hallo @matthias-kleiner  and @davidrohr 

As we discussed last Friday, we would like to have a function to optionally stream some data to check the algorithms, the assumptions used in the algorithm, and the input/output parameters of the algorithm.

We have implemented this approach in Run1 and Run2 and would like to use it in Run3 as well. To this end, we have used TTreeSRedirector in simulation, reconstruction, and calibration. **TTreeSRedirector is much more flexible than the functionality proposed by David**. In case needed, implementation can be further optimized, I have a few possible changes, to improve caching. 

**Streaming in Run1-Run2 was well established and we used it regularly. I would like to point out the functionality, referring to the Run1-Run2 code.** 

Streamer parameterization:
* enums for enabling streamers:
  * https://github.com/miranov25/AliRoot/blob/2f249b4d1cff4e54e0b17f2b23ddb4d9b5e4bd9e/TPC/TPCrec/AliTPCtracker.h#L42-L66
* streamers are self-documenting -see e.g streamer for common mode and ion tail streaming
  * https://github.com/miranov25/AliRoot/blob/2f249b4d1cff4e54e0b17f2b23ddb4d9b5e4bd9e/TPC/TPCrec/AliTPCtracker.cxx#L2718-L2729 
* example usage in the Run2 simulation:
  * https://github.com/miranov25/AliRoot/blob/2f249b4d1cff4e54e0b17f2b23ddb4d9b5e4bd9e/TPC/TPCsim/AliTPCDigitizer.cxx#L1106-L1133
* O2 example usage for skimming
   * https://github.com/AliceO2Group/O2Physics/blob/047b31b823c585b7704e13b39fc66752c3329721/Common/Tasks/TPCQCskimming.cxx#L333

The streamers should be not used in the production, they are aimed to be used for the development or for the subset of the data.

Regards
Marian

miranov25:
Hello.

As for the #define statements suggested by David, I have no preference. To motivate their use, perhaps it will become clear once we have more optional streamers. 

In my previous comment, I forgot to mention navigation. Normally we write the indexes in the TTree streamers that allow links to other trees/tables. It is like a relational database with object. This is a very important feature that is further used in the queries and in the RootInteractive visualisation. Now in the Run3  once we will integrate it with RDataFrame<->awkward the functionality will be even more improved, in respect to Run1,2.

matthias-kleiner:
Hello @davidrohr and @miranov25 ,

I tried to implement the suggestions by @davidrohr. I think the compiler should now optimize out the streaming without the  `ifdefs`, since `checkStream()` should always be `false` when compiled without `DEBUG_STREAMER=1` https://github.com/AliceO2Group/AliceO2/blob/ef6dc2af8453354907800da7fee80305bdfb6a14/GPU/GPUTracking/dEdx/GPUdEdx.h#L189 Is it fine like this?

@miranov25 I also changed the interface of how the objects are streamed. You can look here how it is done https://github.com/AliceO2Group/AliceO2/blob/ef6dc2af8453354907800da7fee80305bdfb6a14/GPU/GPUTracking/dEdx/GPUdEdx.h#L191 Is it fine for you?

Do you have some more suggestions?

davidrohr (dismissed):
Thx, looks much cleaner in my opinion

matthias-kleiner:
> Thx, looks much cleaner in my opinion

ok :)

I am now getting an error when compiling. Do you know how to fix this?

[960/1448] Generating GPUReconstructionOCL2Code.bc
FAILED: ....
/data/mkleiner/alice/sw/SOURCES/O2/debug_streamer/0/GPU/GPUTracking/dEdx/GPUdEdx.h:190:27: error: cannot initialize a parameter of type 'const __generic char *' with an lvalue of type 'const __constant char [11]'
    mStreamer.setStreamer("debug_dedx", "UPDATE");

/data/mkleiner/alice/sw/SOURCES/O2/debug_streamer/0/Common/Utils/include/CommonUtils/DebugStreamer.h:131:38: note: passing argument to parameter here
  GPUd() void setStreamer(const char*, const char*){};

davidrohr:
this problem usually appears with templated references and opencl address spaces. You can try to add a GPUgeneric() in front of the reference/ptr type in the argument list. But will need to check in detail to give a precise answer.

matthias-kleiner:
> this problem usually appears with templated references and opencl address spaces. You can try to add a GPUgeneric() in front of the reference/ptr type in the argument list. But will need to check in detail to give a precise answer.

Unfortunately that didnt help.
I could only get it compiling by defining the function as`GPUd() void setStreamer(const char[], const char[]){};` and pass the `char`s as
```
const char name[] = "debug_dedx";
const char type[] = "UPDATE";
mStreamer.setStreamer(name, type);
```
But I guess there must be a better way to do it...

davidrohr:
ok, let me have a look. I actually am currently looking at a very similar problem for Ruben anyway...

davidrohr:
You can just template everything away (see patch below), and please also add the include patch from my patch, otherwise the standalone benchmark breaks.
```
diff --git a/Common/Utils/include/CommonUtils/DebugStreamer.h b/Common/Utils/include/CommonUtils/DebugStreamer.h
index b9ee64a3d14..a7eab88621c 100644
--- a/Common/Utils/include/CommonUtils/DebugStreamer.h
+++ b/Common/Utils/include/CommonUtils/DebugStreamer.h
@@ -128,21 +128,23 @@ class DebugStreamer
   // empty implementation of the class for GPU or when the debug streamer is not build for CPU
  public:
   /// empty for GPU
-  GPUd() void setStreamer(const char*, const char*){};
+  template <typename... Args>
+  GPUd() void setStreamer(Args... args){};
 
   /// always false for GPU
   GPUd() static bool checkStream(const StreamFlags) { return false; }
 
   /// empty for GPU
-  template <typename T, typename... Types>
-  GPUd() void stream(T var1, Types... var2)
+  template <typename... Types>
+  GPUd() void stream(Types... var)
   {
   }
 
   /// empty struct for GPU
   template <typename Type>
   struct StreamerObject {
-    GPUd() StreamerObject(const Type&, const char*){};
+    template <typename... Args>
+    GPUd() StreamerObject(const Type&, Args... args){};
   };
 
 #endif
diff --git a/GPU/GPUTracking/Standalone/CMakeLists.txt b/GPU/GPUTracking/Standalone/CMakeLists.txt
index 6b2c8dd4783..758bd68c8f1 100644
--- a/GPU/GPUTracking/Standalone/CMakeLists.txt
+++ b/GPU/GPUTracking/Standalone/CMakeLists.txt
@@ -164,6 +164,7 @@ include_directories(${GPUTRACKING_DIR}/TPCClusterFinder
                     ${O2_DIR}/Common/Field/include
                     ${O2_DIR}/Common/Constants/include
                     ${O2_DIR}/Common/MathUtils/include
+                    ${O2_DIR}/Common/Utils/include
                     ${O2_DIR}/DataFormats/common/include
                     ${O2_DIR}/DataFormats/Detectors/Common/include
                     ${O2_DIR}/DataFormats/Detectors/ITSMFT/common/include
```

miranov25:
Hello @matthias-kleiner 

I checked the implementation, resp. your example. I did not understand why, having already used TreeSRedirector, you do not use the native functionality?

Why?
```
      Streamer::StreamerObject{region, "region"},
      Streamer::StreamerObject{padRow, "padRow"},
      Streamer::StreamerObject{tanTheta, "tanTheta"},
```
Instead of the native equivalent:
```
"region="<<region<<    
"padRow="<<padRow<<
"tanTheta="<<tanTheta<<
```
I prefer to reduce the number of templates and interfaces when they are not needed. Was that really necessary? Obviously, I do not understand the motivation.

Regards
Marian

matthias-kleiner:
> You can just template everything away (see patch below), and please also add the include patch from my patch, otherwise the standalone benchmark breaks.
> 
> ```
> diff --git a/Common/Utils/include/CommonUtils/DebugStreamer.h b/Common/Utils/include/CommonUtils/DebugStreamer.h
> index b9ee64a3d14..a7eab88621c 100644
> --- a/Common/Utils/include/CommonUtils/DebugStreamer.h
> +++ b/Common/Utils/include/CommonUtils/DebugStreamer.h
> @@ -128,21 +128,23 @@ class DebugStreamer
>    // empty implementation of the class for GPU or when the debug streamer is not build for CPU
>   public:
>    /// empty for GPU
> -  GPUd() void setStreamer(const char*, const char*){};
> +  template <typename... Args>
> +  GPUd() void setStreamer(Args... args){};
>  
>    /// always false for GPU
>    GPUd() static bool checkStream(const StreamFlags) { return false; }
>  
>    /// empty for GPU
> -  template <typename T, typename... Types>
> -  GPUd() void stream(T var1, Types... var2)
> +  template <typename... Types>
> +  GPUd() void stream(Types... var)
>    {
>    }
>  
>    /// empty struct for GPU
>    template <typename Type>
>    struct StreamerObject {
> -    GPUd() StreamerObject(const Type&, const char*){};
> +    template <typename... Args>
> +    GPUd() StreamerObject(const Type&, Args... args){};
>    };
>  
>  #endif
> diff --git a/GPU/GPUTracking/Standalone/CMakeLists.txt b/GPU/GPUTracking/Standalone/CMakeLists.txt
> index 6b2c8dd4783..758bd68c8f1 100644
> --- a/GPU/GPUTracking/Standalone/CMakeLists.txt
> +++ b/GPU/GPUTracking/Standalone/CMakeLists.txt
> @@ -164,6 +164,7 @@ include_directories(${GPUTRACKING_DIR}/TPCClusterFinder
>                      ${O2_DIR}/Common/Field/include
>                      ${O2_DIR}/Common/Constants/include
>                      ${O2_DIR}/Common/MathUtils/include
> +                    ${O2_DIR}/Common/Utils/include
>                      ${O2_DIR}/DataFormats/common/include
>                      ${O2_DIR}/DataFormats/Detectors/Common/include
>                      ${O2_DIR}/DataFormats/Detectors/ITSMFT/common/include
> ```

Thanks for the help and the nice fix

matthias-kleiner:
> Hello @matthias-kleiner
> 
> I checked the implementation, resp. your example. I did not understand why, having already used TreeSRedirector, you do not use the native functionality?
> 
> Why?
> 
> ```
>       Streamer::StreamerObject{region, "region"},
>       Streamer::StreamerObject{padRow, "padRow"},
>       Streamer::StreamerObject{tanTheta, "tanTheta"},
> ```
> 
> Instead of the native equivalent:
> 
> ```
> "region="<<region<<    
> "padRow="<<padRow<<
> "tanTheta="<<tanTheta<<
> ```
> 
> I prefer to reduce the number of templates and interfaces when they are not needed. Was that really necessary? Obviously, I do not understand the motivation.
> 
> Regards Marian

Hello @miranov25 ,
I implemented this interface so that it will compile on the GPU. I can check if it is possible to use the original interface by implementing the `operator<< ` in the debug streamer class.

regards Matthias

miranov25:
Hello @matthias-kleiner 

Concerning your response:
https://github.com/AliceO2Group/AliceO2/pull/9404#issuecomment-1192706673

Do I understand correctly that you cannot compile the code I showed with the internal TTreeSRedirector, but were forced to write a wrapper that calls the same function at the end? Before I implement a new function, I want to be sure I understand the problem.

Marian

matthias-kleiner:
> Hello @matthias-kleiner
> 
> Concerning your response: [#9404 (comment)](https://github.com/AliceO2Group/AliceO2/pull/9404#issuecomment-1192706673)
> 
> Do I understand correctly that you cannot compile the code I showed with the internal TTreeSRedirector, but were forced to write a wrapper that calls the same function at the end? Before I implement a new function, I want to be sure I understand the problem.
> 
> Marian

Yes, you are right. But now I changed it back so that the TTreeSRedirector interface will be used and for the GPU it will use some dummy implementation. You can see here https://github.com/AliceO2Group/AliceO2/blob/8e60c829b065fd744532a91bd40c6007ec22b566/GPU/GPUTracking/dEdx/GPUdEdx.h#L200 that for the CPU `Streamer.getStreamer()` returns the `TreeStreamRedirector` and for GPU or if the debugger is not compiled it returns some dummy class. I think now it is ok for you?

miranov25:
@matthias-kleiner 
Now it is fine for me

PR #9406, opened 2022-07-12
major revision: updated FT0 frame geometry for both A and C side

dedx:
This update includes a re-write of the FT0-A frame geometry to address challenges of complex volumes introducing overlaps and inclusion of the FT0-C side frame geometry.  Due to the re-write this update changes a large number of lines of code but cleans up the logic so that it is easier to follow and compactly coded inside a smaller number of functions.

### DIFF
--- a/Detectors/FIT/FT0/base/src/Geometry.cxx
+++ b/Detectors/FIT/FT0/base/src/Geometry.cxx
@@ -33,7 +33,7 @@ void Geometry::setAsideModules()
 {
   //These are coordinate positions for the sensitive elements within the FIT mother volume
   //measured from the CAD drawings.  The positive/negative X values are referenced
-  //from the back side of the frame lookinmg toward the interaction point 
+  //from the back side of the frame lookinmg toward the interaction point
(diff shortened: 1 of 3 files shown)


dedx:
Resolved the whitespace issues and the clang-format test.

mslupeck (commented):
Hi @dedx 
Thanks for preparing this PR. Since this is a significant code modification, let's wait with merging, until we run more thorough tests comparing hits with, and without this PR.

mslupeck (approved):
As expected, I see only a minor 2% increase in multiplicity on the C-side, where aluminum structure was implemented. Otherwise, timing, charge and lateral spectra look reasonable. Approving to initiate tests. We should still get a bit deeper comment about secondary production in the newly implemented frame.

sawenzel:
@mslupeck : What is the way forward here? Is someone studying secondary production in the new material or should this be merged now?

mslupeck:
Hi @sawenzel 
We have converged (just yesterday afternoon) that this is OK to be merged (once the CI errors are fixed).
We've checked that the behaviour of secondary production in the frame and in sensitive elements behaves according to expectations.
@dedx Could you take care of fixing the CI errors?

PR #9408, opened 2022-07-12
DPL options parsing: allow multiple options `--batch`

matthiasrichter:
Implementing a special handling for the batch option. This is on request
and some kind of a hack. Need to use a vector-type to allow to specify
the same option multiple times on the command line. Default values
can not be handled in the option definition for the moment, but that
problem should be solved somewhere.

For now, this is tailored to only the `batch` option but should be generalized.
Secondly, the boost option parsing does not fill the vector elements,
although code is following the example.


matthiasrichter:
also the `--no-batch` option did not work as intended with the first version. Can not retrieve the correct value from the varmap, fixed now by simply checking if the option was specified.

PR #9414, opened 2022-07-13
[EMCAL-630, EMCAL-789] EMCAL Reco Params and LGnoHG suppression 

mfasDa:
Use 3 sigma noise threshold with a sigma of 0.4 for
the LG digitizer to separate faulty cells from simple
noise: If the LG cell ADC is < 3 sigma the cell is
considered as pure noise cell where the error handling
is discarded, otherwise it is treated as a data corruption.

In addition moving converion from energy to ADC
from internal conversion constant to common constant. This has
a slightly better precision and will result in ~2% increase of
the energy.

[EMCAL-789] Create parameter handler for reconstruction params

Currently implemented:
- Time shift at cell level
- Noise cut LGnoHG

### DIFF
--- a/Detectors/EMCAL/reconstruction/CMakeLists.txt
+++ b/Detectors/EMCAL/reconstruction/CMakeLists.txt
@@ -10,60 +10,61 @@
 # or submit itself to any jurisdiction.
 
 o2_add_library(EMCALReconstruction
-               SOURCES src/RawReaderMemory.cxx
-                       src/RawBuffer.cxx
-                       src/RawHeaderStream.cxx
-                       src/RawPayload.cxx
-                       src/AltroDecoder.cxx
-                       src/Bunch.cxx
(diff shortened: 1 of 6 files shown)


mfasDa:
WIP, don't merge yet!

mfasDa:
For the record: This commit also includes a change of the conversion constant used from to convert the ADC counts to energy, from 0.016 GeV / ADC used so far hard coded in the reconstruction tasks to the more precise 0.0162 GeV/ADC used in run2 as reference. This is of relevance since the current data use the old energy calibration for reference, which was done with respect to 0.0162 GeV/ADC. This leads to an increase of the cell energy by ~1.25% explaining part of the remaining 4% of the mass shift of the pi0 peak we are still observing.

mfasDa:
PR is ready to merge

shahor02 (approved):

PR #9415, opened 2022-07-13
Service work

AlexBigO:
Implementation of new charge-collection tables for Vbb=0V back-bias. Possibility to switch between Vbb=0V and -3V via `o2-sim-digitizer-workflow` command line. Default value put to 0V for ITS and kept to -3V for MFT.


sawenzel:
From looking at this PR, I am wondering if we actually read ~800 text files when we perform a digitization for ITS/MFT?

If this is the case, I believe it to be quite an IO burden/contributor especially in conjunction with CVMFS. Would it be possible to have these parameters converted to a single binary (ROOT) file which is actually used at run-time?

PR #9422, opened 2022-07-14
TRD updates

martenole:
- small fix for the tracking QC which needs to get the GRP now from CCDB
- Trap2CRU now sends different trigger depending on whether digits were suppressed (physics) or included (calib)

### DIFF
--- a/Detectors/TRD/simulation/src/Trap2CRU.cxx
+++ b/Detectors/TRD/simulation/src/Trap2CRU.cxx
@@ -20,16 +20,10 @@
 
 #include "Headers/RAWDataHeader.h"
 #include "CommonDataFormat/InteractionRecord.h"
-#include "DataFormatsTRD/TriggerRecord.h"
-#include "DataFormatsTRD/LinkRecord.h"
-#include "DataFormatsTRD/RawData.h"
-#include "DataFormatsTRD/Tracklet64.h"
-#include "DataFormatsTRD/Constants.h"
 #include "DataFormatsCTP/TriggerOffsetsParam.h"
 #include "DetectorsRaw/HBFUtils.h"
 #include "DetectorsRaw/RawFileWriter.h"
 #include "TRDSimulation/Trap2CRU.h"
-#include "TRDSimulation/TrapSimulator.h"
 #include "CommonUtils/StringUtils.h"
 #include "TFile.h"
 #include "TTree.h"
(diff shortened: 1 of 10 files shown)


Detectors/TRD/simulation/src/Trap2CRU.cxx
```diff
@@ -613,32 +611,35 @@ void Trap2CRU::convertTrapData(o2::trd::TriggerRecord const& triggerrecord, cons
...
-          int firstDigitMCM = mCurrentDigit;
-          while (mDigits[mDigitsIndex[mCurrentDigit]].getMCM() == currentMCM &&
-                 mDigits[mDigitsIndex[mCurrentDigit]].getROB() == currentROB &&
-                 mDigits[mDigitsIndex[mCurrentDigit]].getHCId() == hcid) {
-            mCurrentDigit++;
-            digitcounter++;
-            if (mCurrentDigit == enddigitindex) {
-              break;
+
+        if (isCalibTrigger) {
```
martenole, line 615: @bazinski if I send a physics trigger and still write the digit HC headers and raw end markers then the raw reader gets quite confused:
```
[40125:trd-datareader]: [20:41:02][INFO] 2 current endpont : 1 previous end point : 0
[40125:trd-datareader]: [20:41:02][INFO] 1 current endpont : 1 previous end point : 0
[40125:trd-datareader]: [20:41:02][INFO] 2 current endpont : 1 previous end point : 0
[40125:trd-datareader]: [20:41:02][INFO] 2 current endpont : 1 previous end point : 0
[40125:trd-datareader]: [20:41:02][INFO] 2 current stopbit: 5 previous stopbit: 1
[40125:trd-datareader]: [20:41:02][INFO] 1 current stopbit: 5 previous stopbit: 1
[40125:trd-datareader]: [20:41:02][INFO] 2 current stopbit: 5 previous stopbit: 1
[40125:trd-datareader]: [20:41:02][INFO] 2 current stopbit: 5 previous stopbit: 1
```
But the way it is implemented here it works fine

martenole:
Here I also changed that no more triggers without any digits or tracklets are written out. So in synthetics we should not see the "bogus trigger" messages for the TRD anymore. But I still need to test it with a larger 500kHz pp TF to check that also the BC and orbit are not nonsense.

Detectors/TRD/simulation/src/Trap2CRU.cxx
bazinski: It should just reject the data after the tracklet end marker, and then go in search of the next tracklethcheader. you might end up with more warnings/info messages than you would care to deal with.

bazinski: The raw data off the detector does not have digit half chamber headers for physics triggers.

bazinski (dismissed):

martenole:
Failure is unrelated. This is ready to be merged

PR #9427, opened 2022-07-15
 [EMCAL-790] Fix correction of the cell time for BC mod 4

mfasDa:
The correction method used to merged the bunches
of the 4 trigger BC into 1 assumes the permutation
(0 1 2 3), otherwise it will only merge either 2 or 3 trigger
BCs into 1. Two possible issues can spoil the correction
a) Initial permutation differs from (0 1 2 3), i.e. BC 3 comes
   first, BC 0 second ...
b) Number of BCs in L0-LM correction not a multiple of 4
In order to allow for minimal intervention we treat the two
cases separately
a) We introduce a rollback parameter (default 0) rotating
   the initial permutation
b) We add again L0-LM delay % 4 BCs to the BC used in
   the correction to obtain the default parameterization
In the default case we will get the permutation (0 1 2 3)
and stay with a rollback of 0, non-0 rollback parameters
are needed only of the BC obtained would be shifted with
respect to expectation.

### DIFF
--- a/Detectors/EMCAL/reconstruction/include/EMCALReconstruction/RecoParam.h
+++ b/Detectors/EMCAL/reconstruction/include/EMCALReconstruction/RecoParam.h
@@ -37,8 +37,8 @@ class RecoParam : public o2::conf::ConfigurableParamHelper<RecoParam>
 
  private:
   double mNoiseThresholdLGnoHG = 10.;  ///< Noise threshold applied to suppress LGnoHG error
-  double mCellTimeShiftNanoSec = 600.; ///< Time shift applied on the cell time to center trigger peak around 0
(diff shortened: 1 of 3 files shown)


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/9414

mfasDa:
More information concerning the problem can be found here: https://alice.its.cern.ch/jira/browse/EMCAL-790

mfasDa:
@shahor02 @davidrohr We agreed with @mpoghos that we set update the reco params with the proper params for the current detector configuration so that they don't need to be handled by the shift crew as EXTRA_ENV variables. We have updated the PR accordingly. With the default parameters committed above the 4 BCs merge in 1 peak and is shifted roughly to 0.

mfasDa:
CS8 errors unrelated

PR #9429, opened 2022-07-15
ITS DCS parser for CCDB

ezradlesser:
Added new workflow which can be used for parsing DCS files and pushing them to the CCDB.

Example usage:
```o2-its-dcs-parser-workflow -b --ccdb-url="http://ccdb-test.cern.ch:8080"```

For testing offline one can use the DCS emulator:
```
CHANFROM="type=sub,method=connect,address=tcp://127.0.0.1:5556,rateLogging=1,transport=zeromq"
CHANACK="type=push,method=connect,address=tcp://127.0.0.1:5557,Logging=1,transport=zeromq"
o2-dcs-config-proxy --subscribe-to "$CHANFROM" --acknowlege-to "$CHANACK" | ...
```

### DIFF
--- a/Detectors/ITSMFT/ITS/workflow/CMakeLists.txt
+++ b/Detectors/ITSMFT/ITS/workflow/CMakeLists.txt
@@ -23,6 +23,8 @@ o2_add_library(ITSWorkflow
                        src/ThresholdCalibratorSpec.cxx
                        src/ThresholdAggregatorWorkflow.cxx
                        src/ThresholdAggregatorSpec.cxx
+                       src/DCSParserWorkflow.cxx
+                       src/DCSParserSpec.cxx
          PUBLIC_LINK_LIBRARIES O2::Framework
                                O2::SimConfig
                                O2::DataFormatsITS
@@ -60,6 +62,11 @@ o2_add_executable(threshold-calib-workflow
 
 o2_add_executable(threshold-aggregator-workflow
                   SOURCES src/its-threshold-aggregator-workflow.cxx
+                  COMPONENT_NAME its
+                  PUBLIC_LINK_LIBRARIES O2::ITSWorkflow)
+
+o2_add_executable(dcs-parser-workflow
+                  SOURCES src/its-dcs-parser-workflow.cxx
                   COMPONENT_NAME its
                   PUBLIC_LINK_LIBRARIES O2::ITSWorkflow)
 
--- a/Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/DCSParserSpec.h
+++ b/Detectors/ITSMFT/ITS/workflow/include/ITSWorkflow/DCSParserSpec.h
@@ -1,0 +1,107 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by ezradlesser** [comment]
Hi @shahor02  -- this is the workflow which parses the DCS files for ITS.
(Edit: not DataPoints)

**Comment by shahor02** [comment]
@ezradlesser I've just asked DCS to redirect ITS DPs to the test channel and got reply :`ITS doesn’t have ADAPOS at all, nor they requested it yet`. Could you clarify what are you processing ? (pinging to @iravasen )

**Comment by shahor02** [comment]
Ah, wait, this is not the DCS DataPoints  but the file-push service processor. Did you test it with the `dcssend` DCS server emulator, https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/DCS/testWorkflow#dcs-config-proxy ?

**Comment by ezradlesser** [comment]
> Ah, wait, this is not the DCS DataPoints but the file-push service processor. Did you test it with the `dcssend` DCS server emulator, https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/DCS/testWorkflow#dcs-config-proxy ?

Ah, yes, sorry for the wrong wording... and yes I have tested with `o2-dcs-config-proxy` as above and it is working, and pushing entries to the CCDB

**Comment by iravasen** [comment]
Hi @shahor02, this data (processed then by this wf) are shipped directly from DCS to alidcsfps.cern.ch:3760 via file push without using ADAPOS. This data are sent at the EOR and are quite a lot because they summarized the list of disabled double columns / pixels for every chip. 
Better to ask Markus but I think it was decided like that because adapos is just for small amount of data (and it pushes data every a given amount of time which is not what we need here). 
Markus is also preparing the sending of the strobe_length value and there was the doubt if to use ADAPOS or again file push. It seems adapos is better in this case (i.e. 1 data point only) but do you know if we can use the same workflow or if the logic to receive data from adapos (on O2 side) is completely different?

(now we used `dcssend` on one terminal and dcs-config-proxy + this workflow on another terminal --> the next step is to test it via real ITS DCS to check if the file push works)

**Comment by shahor02** [comment]
@iravasen My 1st message mentioning ADAPIS was a reaction to "Data Points" in the description (they come via adapos). 
Since it appears to process the files from the file-push service, it should be tested as described in the link of the 2nd message. 

If I understood you correctly, it should not be run on the flp199 (where other file-push processors are running) but on your server?
Still, the procedure is the same.

**Comment by iravasen** [comment]
@shahor02 yes sorry, I read your newer messages in delay. We followed that instructions and used dcssend in one terminal (we have a local file with the data to be processed) and dcs-proxy + this workflow on a second terminal. File is received and processed correctly. 

Let me ask Markus about the machine. DCS support told him to send it to alidcsfps but probably can be sent also to flp199 given that DCS can access this machine.

**Comment by iravasen** [comment]
@shahor02 from one of your latest emails about the issues with this workflow you said that on FLP199 you use:
```
SERVER=ALIDCSFPS001.cern.ch
MSGFROM=3764
MSGTO=3766
CHANFROM=\""type=sub,method=connect,address=tcp://$SERVER:$MSGFROM,rateLogging=1,transport=zeromq"\"
CHANACK=\""type=push,method=connect,address=tcp://$SERVER:$MSGTO,rateLogging=1,transport=zeromq"\"
``` 
does it mean that flp199 receives the data from alidcsfps (hence the machine which receives the data from our DCS)?

shahor02:
> does it mean that flp199 receives the data from alidcsfps (hence the machine which receives the data from our DCS)?

yes

iravasen:
ok! So in principle it is fine to run the processors on flp199 (as far as I understand).

shahor02 (approved):
Merging. @ezradlesser @iravasen I guess it should be eventually added to https://gitlab.cern.ch/AliceO2Group/system-configuration/-/blob/dev/ansible/roles/calib-dcs/templates/o2-calib-dcs-file-push-proxy.sh.j2 workflow. Are there some options to set?

PR #9431, opened 2022-07-15
ZDC: pimpl and fwd declare for fast sim

sawenzel:
The fastsim headers should defintely no longer
be seen by ROOT.


sawenzel:
@davidrohr

davidrohr:
Thx, I have unfortunately no good setup to test it directly. I propose we merge it before zonight, and I'll check tomorrow's RPMs.

PR #9433, opened 2022-07-15
Fix: reference invalidated by vector expansion was used

shahor02:
fixes https://alice.its.cern.ch/jira/browse/O2-3068


shahor02:
trivial, merging

PR #9435, opened 2022-07-15
Add additional TimeFrame dump for debug

mpuccio:
@qgp check if this is enough for your needs. To enable it:
```
o2-its-reco-workflow --clusters-from-upstream --trackerCA --tracking-mode sync --configKeyValues "ITSCATrackerParam.dumpTimeFrames=true" --disable-mc --run -b
```


shahor02:
Hi @mpuccio , could you resolve conflicts?

PR #9436, opened 2022-07-15
Limit error messages from DPLRawParser

shahor02:
@davidrohr something like this? Do you want all classes using DPLRawParser to be patched by this?


davidrohr:
Hi Ruben, I'd do it only for the detectors which regularly send such errors, which are currently ITS and MFT.
But actually, my point was to reduce the severity from alarm to warning, and not to throttle after a certain amount of alarms.
I think since every run with ITS or MFT has this kind of messages, it doesn't really help to throttle after the first n per process. It will just send few thousand alarms to the InfoLogger at the start of every run, which every one just ignored. So could we add also a severity flag to the constructor of the raw parser, which defaults to alarm, and in case of ITS and MFT we set it to warning?

PR #9441, opened 2022-07-16
Fixes for tof-calibinfo-reader, more robust TimeSlice methods

shahor02:
@noferini your reader was not using HBFUtilsInitializer and on top of this, was setting all orbits to constant (0).
This was making the `getRunStartOrbit() const { return mCurrentTFInfo.firstTForbit - o2::base::GRPGeomHelper::getNHBFPerTF() * mCurrentTFInfo.tfCounter; }` negative and TF dependent, hence the result were wrong.

This is fixed, but I've put the protection against negative `getRunStartOrbit()` if orbit is not set for some reason.

Now with your workflow I get
```
 grep -E 'closing slot|Sending object' xx.log
[1986551:tof-diagnostic-calibration]: [02:49:44][INFO] closing slot 0 <= TF <= 26399
[1986551:tof-diagnostic-calibration]: [02:49:44][INFO] Sending object TOF/Calib/Diagnostic/o2-tof-Diagnostic_1657932584879.root of size 1024 bytes, valid for 1654239385946 : 1654239706440
[1986551:tof-diagnostic-calibration]: [02:49:51][INFO] closing slot 26400 <= TF <= 52799
[1986551:tof-diagnostic-calibration]: [02:49:51][INFO] Sending object TOF/Calib/Diagnostic/o2-tof-Diagnostic_1657932591353.root of size 1024 bytes, valid for 1654239686440 : 1654240006934
[1986551:tof-diagnostic-calibration]: [02:49:57][INFO] closing slot 52800 <= TF <= 79199
[1986551:tof-diagnostic-calibration]: [02:49:57][INFO] Sending object TOF/Calib/Diagnostic/o2-tof-Diagnostic_1657932597169.root of size 1024 bytes, valid for 1654239986934 : 1654240307428
```

### DIFF
--- a/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
+++ b/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
@@ -151,7 +151,15 @@ class TimeSlotCalibration
 
  protected:
   auto& getSlots() { return mSlots; }
-  int getRunStartOrbit() const { return mCurrentTFInfo.firstTForbit - o2::base::GRPGeomHelper::getNHBFPerTF() * mCurrentTFInfo.tfCounter; }
+  int getRunStartOrbit() const {
+    int orb = mCurrentTFInfo.firstTForbit - o2::base::GRPGeomHelper::getNHBFPerTF() * mCurrentTFInfo.tfCounter;
+    if (orb<0) {
+      LOGP(alarm, "Negative runStartOrbit = {} deduced for from tfCounter={} and firstTForbit={}, enforcing runStartOrbit to 0", orb, mCurrentTFInfo.tfCounter, mCurrentTFInfo.firstTForbit);
+      orb = 0;
+    }
+    return orb;
+  }
+
   TFType tf2SlotMin(TFType tf) const;
 
   std::deque<Slot> mSlots;
--- a/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
+++ b/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
@@ -151,9 +151,10 @@ class TimeSlotCalibration
 
  protected:
   auto& getSlots() { return mSlots; }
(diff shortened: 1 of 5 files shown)


noferini:
Hi Ruben,
thanks.
I did a quick check for run 517616. There is still something I don't understand (update objects to ccdb have a validity shifted  of almost 1 hour early wrt the one I expect from the run Start/Stop and tree timetamps) but I can debug it only next week

shahor02:
The time for sure will be shifted unless you correctly match the 1st orbit and pass it via `HBFUtils.orbitFirst` and `HBFUtils.orbitFirstSampled`. To get correct results you should not rely on HBFUtils but use
```
o2-tof-calib-reader --hbfutils-config o2_tfidinfo.root ...
```
with o2_tfidinfo.root produced in the reco workflow by the `o2-tfidinfo-writer-workflow`

noferini:
Hi Ruben,
I will have a look. If tfidinfo contains all the information needed to process the proper time I could replace current timestamp in calibInfo with it.
I would prefer not to have dependencies on other files, but CalibInfo, in async TOF chain.

shahor02:
Hi @TimoWilken the o2-dataflow-cs8 CI fails with `No space left on device`

shahor02:
@noferini the o2-tfidinfo is a standard way to inject correct O2 headers, what is the problem to use them? BTW, your tof-calibinfo-reader in case of ninstances>1 will inject the same timeslice from different instances which will likely break DPL logic in many places. 
Shall I merge this?

noferini:
Hi Ruben,
yes, please merge this.
About your question, any dependency could limit/make harder the possibility to play with different configuration, for instance replay  from files multiple runs for time slewing calibration, ...

shahor02:
Well, if you want to merge your trees from different runs, you can also merge the o2-tfinfo files in the same order. Or, you should have the content  of DataFormats/common/include/CommonDataFormat/TFIDInfo.h in your class and set it to timing info

PR #9442, opened 2022-07-16
Fix PVertexer: wrong protection against empty bunchfilling

shahor02:


shahor02:
trivial, merging (currently the vertexing is broken).

PR #9454, opened 2022-07-17
DPL: check precondition only once

ktf:


martenole (approved):
Works fine in the FST on the EPNs

PR #9456, opened 2022-07-17
Possibility to set condition query rate via DPL_CONDITION_QUERY_RATE

shahor02:


shahor02:
trivial, checked locally, merging

PR #9458, opened 2022-07-17
Possibility to set query rate per ccdbParamSpec

shahor02:
`ccdbParamSpec` helper is extended to take an extra parameter for the validity check rate:
```
ccdbParamSpec(std::string const& path, bool runDependent, std::vector<CCDBMetadata> metadata = {}, int64_t rate=0);
```
or
```
ccdbParamSpec(std::string const& path, std::vector<CCDBMetadata> metadata = {},int64_t rate=0);
```
with the following meaning:

- `rate>0`: the validity of cached object will be checked for given TF timeslice if `(tfSlice%rate)==0`.

- `rate=0` (default): validity check of the cached object follows the global policy of the DPL fetcher set by the option
`--condition-tf-per-query <N>` (or env.var. `DPL_CONDITION_QUERY_RATE=<N>`) with same meaning that the check is done
for `tfSlice%N==0`. I.e. if you want your object validity check to be done every (or every 3d) TF seen by the workflow, use the `rate = 1 (or 3)`.

- `rate<0`: disable validity checks after caching the object, can be used for objects which are guaranteed to not change during workflow lifetime.

Note: in the online processing on the EPN the tfSlice is a dynamic counter depending on the number of EPNs and workflows per EPN, hence the result of `tfSlice%rate` has nothing to do with the check of `tfCounter%rate` except for the `rate=1`.
Effectively, the values which make sense are 0: follow global policy, 1: check every TF and –1 (or any negative value): don’t check after caching.

### DIFF
--- a/Detectors/Base/src/GRPGeomHelper.cxx
+++ b/Detectors/Base/src/GRPGeomHelper.cxx
@@ -65,7 +65,7 @@ GRPGeomRequest::GRPGeomRequest(bool orbitResetTime, bool GRPECS, bool GRPLHCIF, 
     addInput({"grplhcif", "GLO", "GRPLHCIF", 0, Lifetime::Condition, ccdbParamSpec("GLO/Config/GRPLHCIF")}, inputs);
   }
   if (askGRPMagField) {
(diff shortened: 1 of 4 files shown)


ktf (approved):

shahor02:
Sim_challenge Failed for unrelated reason, will be fixed in separate PR, tested locally, merging

PR #9459, opened 2022-07-18
DPL: use 15s as metric interval

ktf:


ktf:
@awegrzyn this should help with the influxdb overload.

awegrzyn (approved):

PR #9460, opened 2022-07-18
Avoid empty aggregator workflows

martenole:


davidrohr:
I think this will not work. If the workflow generates an empty topology without processes, it is anyway stripped. So this check is not needed. Probably it will even break it since it would not create the output file at all.

Instead, I think the problem is that if tue upload url is set, the ccdb populator is added unconditionally.

Kind Regards
David Rohr

On 18 July 2022 11:28:17 EEST, Ole Schmidt ***@***.***> wrote:
***@***.*** requested review from @AliceO2Group/core on: AliceO2Group/AliceO2#9460 Avoid empty aggregator workflows as a code owner.

martenole:
Hm, ok. This was not meant to fix the problem with the empty input for the ccdb-populator, but rather in the FST I am seeing an error in case TPC DEDX calib is not set. Because then the BARREL_SPORADIC aggregator workflow is empty and in the console I see `ERROR: Empty workflow provided while running in batch mode`. But this does not affect the other processes it seems so it is probably OK to leave it as that. What I don't understand yet is why I have to put for the calibration workflows `GLOBALDPLOPT=-b` in order to avoid warnings that it wants to run in batch mode and for the normal reco workflows this is doing that automatically

PR #9461, opened 2022-07-18
Fix TPC to ITS propagation + add extra Z tolerance

fmazzasc:


shahor02:
@fmazzasc Thanks! Could you please apply clang-format?

shahor02 (dismissed):

Detectors/Vertexing/src/SVertexer.cxx
```diff
@@ -513,18 +539,23 @@ int SVertexer::checkCascades(float rv0, std::array<float, 3> pV0, float p2V0, in
...
+      LOG(debug) << "DCA: " << dca.getY() << " " << dca.getZ();
       mCascadesTmp[ithread].pop_back();
       continue;
     }
-    casc.setCosPA(cosPA);
-    casc.setVertexID(v0.getVertexID());
+
+    LOG(debug) << "Casc successfully added";
+    casc.setCosPA(bestCosPA);
+    casc.setVertexID(cascVtxID);
```
shahor02, line 558: The cascade contains the V0 ID, e.g. in the AOD it is stored simply as a reference on V0 and bachelor track indices: https://github.com/AliceO2Group/AliceO2/blob/a9d875e3b1082d218489380b536036b1d103b45a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L972.
With this fix you may have a situation when V0 is registered for one PV while the cascade using this v0 is for a different PV. I think you need to readjust also the PV index and cosPA of the V0. 
But given that you may add different cascades for different PVs using the same V0, I think you will need to replicate the V0 if the PV used by the cascade is not the same as the PV used by the original V0.

Detectors/Vertexing/src/SVertexer.cxx
```diff
@@ -421,10 +421,10 @@ bool SVertexer::checkV0(const TrackCand& seedP, const TrackCand& seedN, int iP,
   if (checkForCascade) {
     int nCascAdded = 0;
     if (hypCheckStatus[HypV0::Lambda] || !mSVParams->checkCascadeHypothesis) {
-      nCascAdded += checkCascades(rv0, pV0, p2V0, iN, NEG, ithread);
+      nCascAdded += checkCascades(rv0, pV0, p2V0, iN, NEG, vlist, ithread);
     }
     if (hypCheckStatus[HypV0::AntiLambda] || !mSVParams->checkCascadeHypothesis) {
-      nCascAdded += checkCascades(rv0, pV0, p2V0, iP, POS, ithread);
+      nCascAdded += checkCascades(rv0, pV0, p2V0, iP, POS, vlist, ithread);
```
shahor02, line 427: alternatively, one can add V0s not only for the vertex with the best cosPA value but for each PV with respect to which cosPA is acceptable, then to look for the cascades in old way, for this PV only.

fmazzasc: but this in principle could leave us with many duplicates that are not even associated with a cascade, right? If what I am saying is correct maybe the first strategy is the best one, and I will just duplicate the V0 if its corresponding cascade is found

shahor02:
yes, there will be duplicates but this is a matter of policy if a given track pair is allowed to make only 1 V0 with the PV is has the best convergence to, or any acceptable (passing cuts) V0 is counted.

fmazzasc:
> yes, there will be duplicates but this is a matter of policy if a given track pair is allowed to make only 1 V0 with the PV is has the best convergence to, or any acceptable (passing cuts) V0 is counted.

I see, any preference then? 
strategy 1) --> store the v0 with best cosPA and duplicate it if the corresponding cascade has a different PV
strategy 2) --> store all the v0s passing the cosPA cut and continue with the old cascade association

mpuccio:
I would go for strategy 1 as it preserves all the info for both V0s and cascade analysis without creating dangerous duplicates for people looking at V0s only.

shahor02:
It depends if the current "best cosPA" selection incurs an inefficiency for non-cascade V0s. One can probably add to SVertexerParams.h a data member 
```
bool selectBestV0 = true; // if true, given pair of tracks can make only 1 V0 with PV if has best convergence to
```
then in the loop https://github.com/AliceO2Group/AliceO2/blob/118804595c3ece7de24f7830d0cae500be3924ca/Detectors/Vertexing/src/SVertexer.cxx#L395 add all v0s passing cuts and move the cascade search for this v0, https://github.com/AliceO2Group/AliceO2/blob/118804595c3ece7de24f7830d0cae500be3924ca/Detectors/Vertexing/src/SVertexer.cxx#L421 inside this loop. Then, after to loop, if `selectBestV0==true`, flag non-best V0s (which did not produce cascade) for removal, e.g. by setting setCosPA(-1).
And finally, in the merging of results from different threads https://github.com/AliceO2Group/AliceO2/blob/118804595c3ece7de24f7830d0cae500be3924ca/Detectors/Vertexing/src/SVertexer.cxx#L80-L117, skip those V0s which were marked for removal.

In this way the cascades check will be exhaustive and we can switch from one V0 mode to another via configurable param.

fmazzasc:
Closing and opening two separate pull requests for Afterburner and SVertexer

PR #9464, opened 2022-07-18
MC: add special particles

sawenzel:
fixes https://alice.its.cern.ch/jira/browse/O2-3050


mpuccio:
Hi @sawenzel, we checked with @fmazzasc and for the hypertriton 3 body decay it indeed works! Thanks a lot!

fmazzasc:
Hi @sawenzel ,  could we merge this PR? we would need it to start a full simulation campaign for hypernuclei

PR #9466, opened 2022-07-18
Fix HBFUtilsInitializer usage in TRD TrackletTransformer and TrapSimulator

shahor02:
Trivial, tested locally, merging.


martenole:
Thanks!

PR #9469, opened 2022-07-18
Define CTF metadata file data type depending on the runType

shahor02:


Detectors/CTF/workflow/src/CTFWriterSpec.cxx
```diff
@@ -259,9 +261,33 @@ void CTFWriterSpec::init(InitContext& ic)
 void CTFWriterSpec::updateTimeDependentParams(ProcessingContext& pc)
 {
   static bool initOnceDone = false;
+  using GRPECS = o2::parameters::GRPECSObject;
   if (!initOnceDone) {
     initOnceDone = true;
     mDataTakingContext = pc.services().get<DataTakingContext>();
+    // determine the output type for the CTF metadata
```
ironMann, line 268: Hi @shahor02, if I understand the raw TFs should have the same EOS metadatatype as CTFs. If so, could you please turn this switch statement into a helper function inside of `DataFormats/Parameters` library, so that DD can use the same code? Ideally, it would convert the fmq property to the metadata value. Thanks

shahor02: @ironMann I can convert it to helper function with signature
```
std::string getMetaDataFileType(const std::string& runType, bool rawImposed)
```
where the rawImposed will come from https://github.com/AliceO2Group/ControlWorkflows/pull/423.
But in O2 the parsing FMQ properties is done centrally by the framework, filling DataTakingContext. For DD you will need to parse them yourself and pass to this helper, ok?

shahor02: @ironMann done: https://github.com/AliceO2Group/AliceO2/pull/9469/files#diff-200da31558fb6e70a67c414204f2c72ecbfcc015d8d1d8f10893abf2ec20730dR153

ironMann: Thanks @shahor02, looks good to me.

shahor02: @ironMann shall I merge it once tests are passed? We should have it deployed on the EPN when also DD is ready

ironMann: @shahor02 please go ahead and merge. Unfortunately, in this state, I cannot disentangle the method from ROOT and GPU libraries, which I don't want to link in DD. I'll open a new PR with a header only implementation later.
Thanks

shahor02:
test were passed, merging after rebase

PR #9470, opened 2022-07-18
DPL Analysis: add simple testcase for ARRAY_INDEX

ktf:


ktf:
@mfasDa this is a trivial example of how you can use array of indices, as you requested in your "point 4" today.

* You need to have something where you declare that  you want an array of indices (pointGroup in this case) for a given separate table (Points3D here) `DECLARE_SOA_ARRAY_INDEX_COLUMN(Points3D, pointGroup);`.
* Then you include the array of indices by adding the `<your-type>`Ids column to your table (Points3DIds, in this case).
* Then, given an iterator, you can refer to the objects referred  by the array of indices via `it.<your accessor>_as<YourType>()`. You can then iterate on the returned object and call methods from that object.

@ddobrigk can we find someone to write a proper example for this in the O2Physics code?
@aalkin did I miss anything? Do you have a pointer to your tutorial where this is explained?

mfasDa:
Adding @fjonasALICE

PR #9471, opened 2022-07-18
Fix: no need to load old GRP from file

shahor02:


shahor02:

PR #9472, opened 2022-07-19
Tentative fix for fetching OrbitReset object

shahor02:


Framework/Core/src/CommonServices.cxx
```diff
@@ -154,24 +154,24 @@ o2::framework::ServiceSpec CommonServices::datatakingContextSpec()
...
-      // Only on the first message
-      if (context.source == OrbitResetTimeSource::Data) {
-        return;
-      }
-      // Only if we do not have already the proper number from CTP
-      if (context.source == OrbitResetTimeSource::CTP) {
-        return;
-      }
-      context.source = OrbitResetTimeSource::Data;
-      context.orbitResetTime = -1;
```
shahor02, line 166: Why all this is needed? The `context.source` seems to be nowhere set except at L165 and it is always setting `context.orbitResetTime` to -1.
I think before it was simply a dead code since the OrbitReset object was unconditionally fetched from the server. 
Now, once it is fetched only if the string `context.orbitResetTime` contains `ccdb://CTP/Calib/OrbitReset`, assigning -1 here leads to skipping the OrbitReset fetching and message `[WARN] timestamp 1727494790 for orbit 787466286 and orbit reset time 1657469667 is well behind TF creation time 1657469667915, use the latter`

Framework/Core/src/CommonServices.cxx
```diff
@@ -154,24 +154,24 @@ o2::framework::ServiceSpec CommonServices::datatakingContextSpec()
...
       for (auto const& ref : processingContext.inputs()) {
         const o2::framework::DataProcessingHeader *dph = o2::header::get<DataProcessingHeader*>(ref.header);
         const auto* dh = o2::header::get<o2::header::DataHeader*>(ref.header);
         if (!dph || !dh) {
           continue;
         }
-        LOGP(debug, "Orbit reset time from data: {} ", dph->creation);
-        context.orbitResetTime = dph->creation;
+        // LOGP(debug, "Orbit reset time from data: {} ", dph->creation);
+        // context.orbitResetTime = std::to_string(dph->creation);
```
shahor02, line 174: also this assignment invalidates the fetching from the server. Actually, I would eliminate all other ways of getting the orbit reset time but reading from the CCDB, for synthetic runs the trick with DPL_RAWPROXY_OVERRIDE_ORBITRESET is fully enough and still leads to reading correct OrbitReset objects from the CCDB. 
In opposite, I realize that hijacking the `orbit-reset-time` at https://github.com/AliceO2Group/AliceO2/pull/9472/files#diff-75f33822fa1864e701f55c79b380b588f9d8d571edf106891acb84a469d4ea7eL209 is prone to create problems: it will provide some orbit reset time generate a timestamp for reading other objects, but many workflows request the orbir_reset object itself, so it should be anyway read. Shall we eliminate this line 209 (it is ok to keep `>GetProperty<std::string>("orbit-reset-time"...` in the proxy) and revert conditional reading of the orbitReset object?

davidrohr:
For synthetic runs wr have to respect orbit-reset-time.
But it ahoukd be enough to read it in the proxy and set the creationtime accordingly, indeed.

Kind Regards
David Rohr

Framework/Core/src/CommonServices.cxx
ktf: Yes, indeed. I think it's just obsolete code which was used by the old, per device, fetching strategy.

Framework/Core/src/CommonServices.cxx
ktf: Yes, sounds good to me.

Framework/Core/src/CommonServices.cxx
shahor02: Then, shall I simply 
1) remove DataTakingContext.source and line https://github.com/AliceO2Group/AliceO2/blob/3361566da33221b8ea2d7384ed5794a55c4432ca/Framework/Core/src/CommonServices.cxx#L209
2) revert https://github.com/AliceO2Group/AliceO2/blob/3361566da33221b8ea2d7384ed5794a55c4432ca/Framework/Core/src/CCDBHelpers.cxx#L324 to what it was before?
3) convert DataTakingContext.orbitResetTime to long and simply assign to it the value extracted from the CTP object

For the synthetic runs creation time hack the current proxy version is fine, it does not need any accompanying change on the CCDBHelpers or CommonServices side.

shahor02:
superseded by https://github.com/AliceO2Group/AliceO2/pull/9477

PR #9473, opened 2022-07-19
[EMCAL-551][WIP] improvement of emcal cell energy compression

fjonasALICE:
- this commit changes the compression for the emcal cell energy. Studies reported here [EMCAL-558](https://alice.its.cern.ch/jira/browse/EMCAL-558) found that using 14bit over an energy range of 0 to 250GeV for cell energy does not give sufficient precision on emcal cluster level. In order to not change the number of bits stored in the CTF (14bit) a separate treatment of high gain and low gain cells in the conversion was implemented, increasing the effective precision achievable with 14bit significantly (e.g. for HG cells 14bit can be used to store only energies of 0 to 15.39GeV)
- fix in the emcal constants, where the constants EMCAL_HGLGTRANSITION and OVERFLOWCUT were swapped. I restored the proper meaning and fixed occurences accordingly
- **this fix is marked WIP!** While the number of bits stored in the CTF does not change, the new conversion does **break backwards compatibility**. We have to discuss how we move forward and if e.g. reproduction of the CTFs is an option. However, this change is necessary in any case to achieve suitable performance on emc cluster level

### DIFF
--- a/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Constants.h
+++ b/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Constants.h
@@ -89,15 +89,14 @@ std::ostream& operator<<(std::ostream& stream, ChannelType_t chantype);
 
 namespace constants
 {
-
-constexpr int OVERFLOWCUT = 950;               ///< sample overflow
+constexpr int EMCAL_HGLGTRANSITION = 950;      ///< sample overflow
 constexpr int LG_SUPPRESSION_CUT = 880;        ///< LG bunch suppression ADC value
 constexpr int ORDER = 2;                       ///< Order of shaping stages of the signal conditioning unit
 constexpr double TAU = 2.35;                   ///< Approximate shaping time
 constexpr Double_t EMCAL_TIMESAMPLE = 100.;    ///< Width of a timebin in nanoseconds
 constexpr Double_t EMCAL_ADCENERGY = 0.0162;   ///< Energy of one ADC count in GeV/c^2
 constexpr Int_t EMCAL_HGLGFACTOR = 16;         ///< Conversion from High to Low Gain
-constexpr Int_t EMCAL_HGLGTRANSITION = 1024;   ///< Transition from High to Low Gain
+constexpr Int_t OVERFLOWCUT = 1024;            ///< Transition from High to Low Gain
 constexpr Int_t EMCAL_MAXTIMEBINS = 15;        ///< Maximum number of time bins for time response
 constexpr int MAX_RANGE_ADC = 0x3FF;           ///< Dynamic range of the ADCs (10 bit ADC)
 constexpr double EMCAL_TRU_ADCENERGY = 0.0786; ///< resolution of the TRU digitizer, @TODO check exact value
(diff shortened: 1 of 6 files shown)


fjonasALICE:
@shahor02 we have changes to the emcal cell object as stated in the description above. We do not change the total number of bits and also do not change the number of bits assigned to each property, only the conversion changes, which breaks backwards compatibility since it affects the encoding/decoding. Do you have any suggestion on how to proceed in this case? Do you know of some way we could use to check with what O2 version a CTF was produced so we could catch these cases and use the old conversion? Otherwise the only option i could think of would be to convert emcal standalone ctfs to the new format.

shahor02:
Hi,
The methods `cell.getPackedXXX` and `setPacked` don't change, right? In this case, no change is needed on the CTF side, though the dictionary should be eventually regenerated.

From the data, there is no way to know which O2 was used to create it. The CTF header contains a dictionary ID, what you can do is 
1) for the new data use a dictionary with incremented e.g. minor version (currently v1.0 is used), e.g. like in  https://github.com/AliceO2Group/AliceO2/blob/a9d875e3b1082d218489380b536036b1d103b45a/Detectors/FIT/FV0/reconstruction/src/CTFCoder.cxx#L41-L49
2) recalculate energy during decoding of the old data at 
https://github.com/AliceO2Group/AliceO2/blob/a9d875e3b1082d218489380b536036b1d103b45a/Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h#L161
if data was encoded with old dictionary:
```
if (static_cast<const o2::ctf::CTFDictHeader&>(header).majorVersion==1 && static_cast<const o2::ctf::CTFDictHeader&>(header).minorVersion==0) { ...}
```
In this case you must regenerate a new dictionary for the new data.

The problem is that to be consistent, you still have to foresee encoding with the old dictionary and in this case you should impose the old way of coding.

mfasDa:
Hi @shahor02,

the packed bits don't change (getPacked and setPacked), only the en-/de-coding of the energy into tho the packed bits.

I was also wondering whether we can implement a method in the decoding which unpacks the energy with the old method and sets it again using the new method. That would probably need to be implemented in the CTFCoder for EMCAL. In this case we would still be able to read the cell from the old data and just have to take into account the worse precision at low pt.

Attention: In the new version we make use of the cell type for the energy coding defining different encodings for high and low gain cells based on the energy range which is distinct for the two (separated by the HG-LG transition at 950 ADC counts -> ~15.4 GeV). Therefore in case the function setEnergy is used the cell type must be set before (should not affect setPacked/getPacked).

For what concerns the coding: New data we will not want to code the old way, because the new method has better precision. The only thing I would be worried about are anchored simulations to the data collected so far. However they should not use the CTF but rather the emccells.root with the cells in a tree, so we need to provide a cell coding for the old coding method, but directly at cell level.

shahor02:
Hi @mfasDa 

If from the data itself you cannot detect if old or new energy coding is used, then the choice is limited:

1) pass an option to decoder to telling that this old data and after decoding CTF it has to rewrite the energy. In this case no dictionary change is needed.
2) Use a higher version of dictionary for the new data decoding. The old data can be recognized by the dictionary version v1.0.
This will require a dictionary update. Note that since to decode old data we have to keep the current dictionary for the time coverage from Oct. 2021 till the validity start of the new dictionary, the current dictionary will be used also if you run encoding with the timestamp before new dict. validity start. So, this data will look like an old energy packing, even if it is not.

PR #9474, opened 2022-07-19
add tfidinfo in tof calibinfo

noferini:


noferini:
Hi,
I added TFIDInfo to allow calib-reader to reconstruct properly the time and torun async calibration s sync one.
TFIDInfo are added when diagnostic are created after raw and ctf decoding.
Shall/can I set also it in simulation (digitization)?

shahor02:
> Shall/can I set also it in simulation (digitization)?

Hi, where do you want to add it in sim/dig and for which purpose? There the TF is even not defined yet.

BTW, apart from adding this info to your class, you should also inject it into DPL.

noferini:
> BTW, apart from adding this info to your class, you should also inject it into DPL.

This will run on reconstruction cpass (now in apass1) and I assumed that the info is already injected after reading ctf.
Is this correct or the info is added later?

shahor02:
Yes, I was referring to injection by your custom reader

noferini:
> Yes, I was referring to injection by your custom reader

I see, yes I will take care of it

noferini:
> > Yes, I was referring to injection by your custom reader
> 
> I see, yes I will take care of it

Now added

noferini:
Hi Ruben,
can we merge this PR?

PR #9475, opened 2022-07-19
DPL Analysis: improve the error message when a known wagon is missing.

ktf:


Framework/Core/src/DataInputDirector.cxx
```diff
@@ -576,7 +576,14 @@ TTree* DataInputDirector::getDataTree(header::DataHeader dh, int counter, int nu
     treename = fileAndFolder.folderName + "/" + treename;
     tree = (TTree*)fileAndFolder.file->Get(treename.c_str());
     if (!tree) {
-      throw std::runtime_error(fmt::format(R"(Couldn't get TTree "{}" from "{}". Please check https://aliceo2group.github.io/analysis-framework/docs/troubleshooting/treenotfound.html for more information.)", treename, fileAndFolder.file->GetName()));
+      std::map<std::string, std::string> likelyMissingTasks = {
+        {"O2cfcollision_001", "FDDConverter"},
```
jgrosseo, line 580: `O2cfcollision_001` does not make sense here. We were playing with it to validate something and it is a CF derived format (actually mine) :)

jgrosseo: Also the FDD converter does not produce it.

jgrosseo:
We could suggest the missing workflow here but not the wagon name because they are unrelated (yes, we call the wagon containing o2-analysis-fdd-converter FDDConverter but this is of course random...

PR #9477, opened 2022-07-19
Cleanup of orbitReset extraction

shahor02:
@ktf I propose this, more cardinal clean up instead of https://github.com/AliceO2Group/AliceO2/pull/9472


davidrohr:
Looks good to me

PR #9479, opened 2022-07-19
DPL: add instructions for the remote GUI

ktf:


ktf:
image
image

ktf:
@shahor02 @davidrohr could you check if you have the same rendering issues with the GUI when running in "remote" mode, over websocket? That would tell us if the problem is at the level of the renderer or already at the level of ImGUI. Notice you might have to allow for CORS in Firefox.

mvala:
@ktf is it possible to specify port as enviroment variable? Something like
```
export DPL_DRIVER_REMOTE_GUI_PORT=12345
```

ktf:
Have you tried with DPL_REMOTE_GUI_PORT?

Ciao,
Giulio

mvala:
No i didn't and i've just tried and it is working. Maybe it can be documented. Thanks

ktf:
I have updated  the documentation in #10072 . Thank you for trying things out.

PR #9480, opened 2022-07-20
FIT: Add DCS configuration workflows

andreasmolander:
Add workflows for processing bad channel maps for FT0, FV0 and FDD


mslupeck (dismissed):

mslupeck (approved):

davidrohr:
@shahor02 @chiarazampolli : Shall we merge this? Or is someone still checking?

chiarazampolli:
For me it is ok to merge. We will then have to test it before putting it in production.

andreasmolander:
We are not 100% ready to test this on FLP with real DCS data just yet (probably will be after a day or two, I just need to make sure we know how to send the DCS data for the testing), and some questions about this code came up today that we could discuss in the meantime (which may result in small changes still). So from my side it can be merged, but it's also fine to wait a couple of days.

PR #9481, opened 2022-07-20
TRD fix error message to only have information that is certain to be there.

bazinski:
- weird text seen in error messages.
- change the information to be sector:side:layer so the link cru in question and not the actual link, which we sometimes cant know definitively.
- removed 2 checks on stack and layer in the digit hcheader as they are not constant for a given half cru of 15 links, did not produce an error merely updated the graphs which was somewhat confusing, assuming trd qc is actually running.
- fix the datarates that appear in the logs, rejected and accepted.

### DIFF
--- a/Detectors/TRD/reconstruction/include/TRDReconstruction/CruRawReader.h
+++ b/Detectors/TRD/reconstruction/include/TRDReconstruction/CruRawReader.h
@@ -154,19 +154,24 @@ class CruRawReader
   bool checkRDH(const o2::header::RDHAny* rdh);
   bool skipRDH();
   void updateLinkErrorGraphs(int currentlinkindex, int supermodule_half, int stack_layer);
+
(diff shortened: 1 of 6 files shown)


bazinski:
@martenole all cool for 200tf of 520886 and 100 of 521257

martenole (dismissed):
Thanks, then this is fine with me of course. I assume you saw the HalfCRU Header alarm somewhere in your tests?

bazinski:
Yip a couple of times. At 280ms per timeframe, its *alot* slower on my desktop than the epn (60ms) ;-)

bazinski:
@martenole I fixed the stats in the log files, they are sent to qc as well, as being in the logs on the epn.

bazinski:
@martenole dont merge this, I will change error and alarm to warn where applicable, else in another PR ?

bazinski:
@davidrohr this now includes changes to deprecate errors/alarms to warn and includes rate limiting on them.

PR #9493, opened 2022-07-21
Fix V0-bachelor association

fmazzasc:
Follow up of P.R. #9461

I implemented strategy 1) --> if the v0 and the bachelor match with a different vertexID, a new v0 clone is created, the cosPA is recomputed and the new PV is set. Then the new v0 is pushed into the v0 vector. I have also added a couple of flags in the SVertexerParams file:

- `marginBachPV` : it's the lower margin from where the bachelor loop starts: since bachelors are sorted in bracket.GetMin() order, the loops should not start here `mVtxFirstTrack[posneg][v0.getVertexID()]` as this excludes all the bachelors w/ bachelor.bracket.GetMin() < than the v0.getVertexID() one. So I added a margin that is subtracted to the minimum vertex index of the V0. The value of the margin is chosen by looking at the maximum number of vertices in which a bachelor is present 

- `selectBestV0` : if the bachelor bracket does not include the v0 best vertex, the cascade candidate is rejected. If the v0 best vertex is in the bachelor bracket, this vertex is the one used. With this flag `True` no clones should be generated

### DIFF
--- a/Detectors/Vertexing/src/SVertexer.cxx
+++ b/Detectors/Vertexing/src/SVertexer.cxx
@@ -574,7 +574,7 @@ int SVertexer::checkCascades(float rv0, std::array<float, 3> pV0, float p2V0, in
       v0clone.setCosPA(cosPA);
       v0clone.setVertexID(cascVtxID);
       mV0sTmp[ithread].push_back(v0clone);
-      casc.setV0ID(mV0sTmp[ithread].size() - 1);  // set the new V0 index in the cascade
+      casc.setV0ID(mV0sTmp[ithread].size() - 1); // set the new V0 index in the cascade
     }
   }
 
--- a/Detectors/Vertexing/src/SVertexer.cxx
+++ b/Detectors/Vertexing/src/SVertexer.cxx
@@ -471,7 +471,7 @@ int SVertexer::checkCascades(float rv0, std::array<float, 3> pV0, float p2V0, in
     v0Idx++;
     firstIdx = mVtxMaxLUT[posneg][v0Idx];
   }
-  auto firstTr = mTrackSortVtxMax[posneg][firstIdx];
+
   for (int it = firstIdx; it < tracks.size(); it++) {
     auto& trackInd = mTrackSortVtxMax[posneg][it];
     auto& bach = tracks[trackInd];
--- a/Detectors/Vertexing/src/SVertexer.cxx
+++ b/Detectors/Vertexing/src/SVertexer.cxx
@@ -312,8 +312,10 @@ void SVertexer::buildT2V(const o2::globaltracking::RecoContainer& recoData) // a
       if (vtxFirstT[t.vBracket.getMin()] == -1) {
         vtxFirstT[t.vBracket.getMin()] = i;
       }
-      if (vtxLastLUT[t.vBracket.getMax()] == -1) {
-        vtxLastLUT[t.vBracket.getMax()] = i;
+      const auto& sortIdx = vtxLastT[i];
+      const auto& trcSort = tracksPool[sortIdx];
+      if (vtxLastLUT[trcSort.vBracket.getMax()] == -1) {
+        vtxLastLUT[trcSort.vBracket.getMax()] = i;
       }
     }
   }
@@ -469,10 +471,8 @@ int SVertexer::checkCascades(float rv0, std::array<float, 3> pV0, float p2V0, in
     v0Idx++;
     firstIdx = mVtxMaxLUT[posneg][v0Idx];
   }
-
-  int firstTr = mTrackSortVtxMax[posneg][firstIdx], nTr = tracks.size();
-
-  for (int it = firstTr; it < nTr; it++) {
+  auto firstTr = mTrackSortVtxMax[posneg][firstIdx];
+  for (int it = firstIdx; it < tracks.size(); it++) {
     auto& trackInd = mTrackSortVtxMax[posneg][it];
     auto& bach = tracks[trackInd];
 
@@ -480,7 +480,7 @@ int SVertexer::checkCascades(float rv0, std::array<float, 3> pV0, float p2V0, in
       continue; // skip the track used by V0
     }
     if (bach.vBracket.getMin() > v0vlist.getMax()) {
-      LOG(info) << "Skipping";
+      LOG(debug) << "Skipping";
       break; // all other bachelor candidates will be also not compatible with this PV
     }
     auto cascVlist = v0vlist.getOverlap(bach.vBracket); // indices of vertices shared by V0 and bachelor
@@ -503,7 +503,7 @@ int SVertexer::checkCascades(float rv0, std::array<float, 3> pV0, float p2V0, in
     const auto& cascXYZ = fitterCasc.getPCACandidatePos(candC);
 
     // make sure the cascade radius is smaller than that of the mean vertex
-    float dxc = cascXYZ[0] - mMeanVertex.getX(), dyc = cascXYZ[1] - mMeanVertex.getY(), dzc = cascXYZ[2] - mMeanVertex.getZ(), r2casc = dxc * dxc + dyc * dyc;
+    float dxc = cascXYZ[0] - mMeanVertex.getX(), dyc = cascXYZ[1] - mMeanVertex.getY(), r2casc = dxc * dxc + dyc * dyc;
     if (rv0 * rv0 - r2casc < mMinR2DiffV0Casc || r2casc < mMinR2ToMeanVertex) {
       continue;
     }
@@ -521,11 +521,7 @@ int SVertexer::checkCascades(float rv0, std::array<float, 3> pV0, float p2V0, in
     trNeut.getPxPyPzGlo(pNeut);
     trBach.getPxPyPzGlo(pBach);
     std::array<float, 3> pCasc = {pNeut[0] + pBach[0], pNeut[1] + pBach[1], pNeut[2] + pBach[2]};
-    auto prodPPos = pV0[0] * dxc + pV0[1] * dyc + pV0[2] * dzc;
-    if (prodPPos < 0.) { // causality cut
-      LOG(debug) << "Casc not causally compatible";
-      continue;
-    }
+
     float pt2Casc = pCasc[0] * pCasc[0] + pCasc[1] * pCasc[1], p2Casc = pt2Casc + pCasc[2] * pCasc[2];
     if (pt2Casc < mMinPt2Casc) { // pt cut
       LOG(debug) << "Casc pt too low";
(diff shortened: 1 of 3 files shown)


Detectors/Vertexing/src/SVertexer.cxx
```diff
@@ -436,29 +436,38 @@ bool SVertexer::checkV0(const TrackCand& seedP, const TrackCand& seedN, int iP,
...
   // check last added V0 for belonging to cascade
   auto& fitterCasc = mFitterCasc[ithread];
   const auto& v0 = mV0sTmp[ithread].back();
   auto& tracks = mTracksPool[posneg];
-  const auto& pv = mPVertices[v0.getVertexID()];
+
   int nCascIni = mCascadesTmp[ithread].size();
   // start from the 1st track compatible with V0's primary vertex
-  int firstTr = mVtxFirstTrack[posneg][v0.getVertexID()], nTr = tracks.size();
+  int firstIdx = v0vlist.getMin() - mSVParams->marginBachPV > 0 ? v0vlist.getMin() - mSVParams->marginBachPV : 0;
```
shahor02: Could you clarify why this marginBachPV is needed? By taking the `firstIdx = v0vlist.getMin()` you are already picking the 1st track compatible with the earliest vertex compatible with v0. Why do you need to go to earlier tracks?

Detectors/Vertexing/src/SVertexer.cxx
```diff
@@ -513,20 +546,38 @@ int SVertexer::checkCascades(float rv0, std::array<float, 3> pV0, float p2V0, in
...
-    casc.setVertexID(v0.getVertexID());
+
+    LOG(debug) << "Casc successfully added";
+    casc.setCosPA(bestCosPA);
+    casc.setVertexID(cascVtxID);
     casc.setDCA(fitterCasc.getChi2AtPCACandidate());
+
+    // clone the V0, set new cosPA and VerteXID, add it to the list of V0s
+    if (cascVtxID != v0.getVertexID()) {
+      auto v0clone = v0;
```
shahor02, line 595: This is not cloning: the clone should be added to the end of the  `mV0sTmp[ithread]`, i.e. 
```
auto& v0clone = mV0sTmp[ithread].push_back(v0);
```
But there is 1 more subtle point: some V0s are added only to check if there is a cascade for them, they are flagged by https://github.com/AliceO2Group/AliceO2/blob/5ce5ee631b75399add8a0869bf0edc8c8de99301/Detectors/Vertexing/src/SVertexer.cxx#L386. Imagine that you find a cascade for such a v0 but had to clone it because the cascade prefers different vertex. Then the v0 which should have been removed (since it has bad cosPA for the non-cascade V0 and is not used by any V0) will remain in the mV0sTmp[ithread].

I think at https://github.com/AliceO2Group/AliceO2/blob/5ce5ee631b75399add8a0869bf0edc8c8de99301/Detectors/Vertexing/src/SVertexer.cxx#L429 one should check if ant of the added cascades is using this V0 and if not, one should effectively copy the last added clone (in reality, redefine the vertexID and cosPA) to this v0, remove the clone and fix the reference from the cascade to removed cloned V0.

One more point :): you are cloning V0 for every cascade which prefers different vertex than original V0. In this case you may have multiple clones for the same V0/PV, while one should do this only once.

shahor02 (changes requested):

Detectors/Vertexing/src/SVertexer.cxx
fmazzasc: Because bachelors are sorted by minimum vtxID.  If the array starts here: `mVtxFirstTrack[posneg][v0.getMin()]`  you miss all the bachelors having a minimum vtx id < v0.getMin(). Then when you take the overlap between the two brackets is already too late, and you are getting the overlap only for bachelors w/ min vertexID >= v0.getMin(). Do you see my point?

shahor02: OK, I think I understood, but then any fixed offset will be either insufficient starting from some interaction rate (vertices per millisecond) or will kill the advantage of using sorted indices. Would not it better in this case to prepare in advance an extra index vector for bachelors with inverse sorting (by the highest vertex ID compatible with this track), so that all the tracks below `mVtxFirstTrack[posneg][ invBachIndex[v0vlist.getMin] ]` have no chance to be compatible with this vertex and those after?

fmazzasc: Yes, you are right! I am working on it

Detectors/Vertexing/src/SVertexer.cxx
fmazzasc: > This is not cloning: the clone should be added to the end of the mV0sTmp[ithread], i.e.

But actually this is what I am doing: creating a clone here: 

> https://github.com/AliceO2Group/AliceO2/blob/5ce5ee631b75399add8a0869bf0edc8c8de99301/Detectors/Vertexing/src/SVertexer.cxx#L570

and pushing_back it here:

> https://github.com/AliceO2Group/AliceO2/blob/5ce5ee631b75399add8a0869bf0edc8c8de99301/Detectors/Vertexing/src/SVertexer.cxx#L576

or am I missing something?

-------------------------------------------------

> one should check if ant of the added cascades is using this V0 and if not, one should effectively copy the last added clone (in reality, redefine the vertexID and cosPA) to this v0

I see your point, working on it

-------------------------------------------------

> One more point :): you are cloning V0 for every cascade which prefers different vertex than original V0. In this case you may have multiple clones for the same V0/PV, while one should do this only once.

not sure I am getting your point. To me it looks that for every v0 you check for the cascade, and eventually you clone the v0 if the cascade is added with a different PV. But only the PV that guarantees the best CosPA is chosen. So in principle you can have at maximum two identical V0s w/ different PV. Isn't it?

shahor02: But what if you create a few cascades with the same V0 and PV different from the PV used by the V0? Would you add a V0 clone for each of these cascades, even though they use the same PV?

Sorry, you are right about cloning (missed `push_back` in  the end).

fmazzasc:
Hi @shahor02 , last commit should address your comments.
1) Max Vtx ID of the bachelor is matched with the Min Vtx ID of the V0. I had to generate a LUT for doing it without cloning the tracks. I checked the fix in local and the matches look correct
2) Added a map in order to avoid useless duplication of the V0.

Detectors/Vertexing/src/SVertexer.cxx
```diff
@@ -436,29 +448,49 @@ bool SVertexer::checkV0(const TrackCand& seedP, const TrackCand& seedN, int iP,
...
+  while (firstIdx < 0 || firstIdx >= mTrackSortVtxMax[posneg].size()) {
+    v0Idx++;
+    firstIdx = mVtxMaxLUT[posneg][v0Idx];
   }
+
+  int firstTr = mTrackSortVtxMax[posneg][firstIdx], nTr = tracks.size();
+
   for (int it = firstTr; it < nTr; it++) {
-    if (it == avoidTrackID) {
+    auto& trackInd = mTrackSortVtxMax[posneg][it];
```
shahor02: @fmazzasc The `firstTr` is first extracted from the mTrackSortVtxMax, then you pass it again to `mTrackSortVtxMax`, why? Should not you simply loop like?
```
for (int it = firstIdx; it < nTr; it++)
{
  int firstTr = mTrackSortVtxMax[posneg][it];
}
```

Detectors/Vertexing/src/SVertexer.cxx
```diff
@@ -469,8 +501,9 @@ int SVertexer::checkCascades(float rv0, std::array<float, 3> pV0, float p2V0, in
     }
     int candC = 0;
     const auto& cascXYZ = fitterCasc.getPCACandidatePos(candC);
-    // make sure the cascade radius is smaller than that of the vertex
-    float dxc = cascXYZ[0] - pv.getX(), dyc = cascXYZ[1] - pv.getY(), dzc = cascXYZ[2] - pv.getZ(), r2casc = dxc * dxc + dyc * dyc;
+
+    // make sure the cascade radius is smaller than that of the mean vertex
+    float dxc = cascXYZ[0] - mMeanVertex.getX(), dyc = cascXYZ[1] - mMeanVertex.getY(), dzc = cascXYZ[2] - mMeanVertex.getZ(), r2casc = dxc * dxc + dyc * dyc;
```
shahor02: why using `mMeanVertex.getZ()` here? It can be many cm away from the real vertex.

fmazzasc: dzc was employed only for the causality cut. Now I moved the causality cut after the proper vertex computation.

Detectors/Vertexing/src/SVertexer.cxx
fmazzasc: You are right, thanks! I also spotted a bug in the LUT table, that I have fixed. You can see the impact of the whole P.R. in the attached plot
c1

shahor02:
Thanks! And for the timing, does it suffer too much?

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

fmazzasc:
> Thanks! And for the timing, does it suffer too much?>



Tested on a pp MB sample @500kHz, 600 signal events + 4 Xi injection per event:
new --> Found 1181 V0s and 129 cascades, timing: CPU: 2.29 Real: 2.36894 s
old -->   Found 1143 V0s and 61 cascades, timing: CPU: 1.76 Real: 1.84048 s

I am trying also to test it on a Pb-Pb MB event, but for the moment the simulation crashes during the its-tpc-matcher workflow. 
@shahor02

shahor02:
Thanks, the time increase is acceptable, at least in pp.
Could you provide more details on matching crash?

fmazzasc:
> Could you provide more details on matching crash?

sure, the error is: 
[3296760:tpc-clusters-sharing-map-producer]: [14:01:51][ERROR] Exception caught: Must provide clusters access and preallocated recepient for shared map 

log in attachment
[itstpcMatch_1.log](https://github.com/AliceO2Group/AliceO2/files/9209918/itstpcMatch_1.log)

shahor02:
It looks like the TPC clusters are corrupted. Is this private MC? Could you put the files on the cernbox?

fmazzasc:
Hi @shahor02 , time estimate for PbPb MB event @ 50kHz, 30 events.
before the fix ---> Found 6189 V0s and 94 cascades, timing: CPU: 3.71 Real: 3.78549 s 
after the fix ---> Found 6229 V0s and 167 cascades, timing: CPU: 7.56 Real: 7.63396 s
What do you think?

shahor02:
Hi @fmazzasc 

Thanks for the estimate, doubling processing time is not good but I was afraid of the worst... I think at the moment we have no other choice but to accept it. Unless we can optimise it further, which is not obvious, the reduction of the CPU time must be achieved by minimizing the amount of ambiguous track-vertex assignments as much as possible.

shahor02 (approved):

PR #9497, opened 2022-07-21
o2-eve, o2-eve-export-workflow, o2-eve-convert: serialisation to root…

jmyrcha:
… files, conversion tool

### DIFF
--- a/EventVisualisation/Base/include/EventVisualisationBase/FileWatcher.h
+++ b/EventVisualisation/Base/include/EventVisualisationBase/FileWatcher.h
@@ -18,6 +18,7 @@
 
 #include <string>
 #include <deque>
+#include <vector>
 
 namespace o2
 {
@@ -33,10 +34,11 @@ class FileWatcher
   std::string prevItem(const std::string& item) const;
   std::string mDataFolder;  ///< folder being observed
   std::string mCurrentFile; ///< "current" file name
+  std::vector<std::string> mExt; ///< extensions of files to be observed
   bool currentFileExist();
 
  public:
-  FileWatcher(const std::string& path);
+  FileWatcher(const std::string& path, std::vector<std::string> ext);
   void changeFolder(const std::string& path);                         ///< switch to observe other folder
   void saveCurrentFileToFolder(const std::string& destinationFolder); ///< copies
   int getSize() const; ///< include guards (so >=2 )
--- a/EventVisualisation/DataConverter/src/VisualisationEventROOTSerializer.cxx
+++ b/EventVisualisation/DataConverter/src/VisualisationEventROOTSerializer.cxx
@@ -40,7 +40,7 @@ std::string VisualisationEventROOTSerializer::readString(TFile& f, const char* n
 {
   TNamed* v = (TNamed*)f.Get(name);
   if (v == nullptr) {
-    return 0;
+    return "";
   }
   std::string result = v->GetTitle();
   free(v);
@@ -213,7 +213,7 @@ bool VisualisationEventROOTSerializer::fromFile(VisualisationEvent& event, std::
   int track_points;   // number of track points
   int track_clusters; // number of track clusters
   int track_source;
-  std::string* track_GID = 0;
+  std::string* track_GID = nullptr;
   float track_time;
   int track_charge;
   float track_theta;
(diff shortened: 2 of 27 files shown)


jmyrcha:
there is no trailing spaces in O2DPLDisplay.cxx

EventVisualisation/Base/include/EventVisualisationBase/FileWatcher.h
```diff
@@ -33,10 +34,11 @@ class FileWatcher
   std::string prevItem(const std::string& item) const;
   std::string mDataFolder;  ///< folder being observed
   std::string mCurrentFile; ///< "current" file name
+  std::vector<std::string> mExt; ///< extensions of files to be observed
   bool currentFileExist();
 
  public:
-  FileWatcher(const std::string& path);
+  FileWatcher(const std::string& path, std::vector<std::string> ext);
```
shahor02, line 41: why passing the vector by copy?

jmyrcha:
fullCI failed on irrelevant problems

EventVisualisation/DataConverter/src/VisualisationEventROOTSerializer.cxx
```diff
@@ -0,0 +1,329 @@
...
+  calo.Branch("GID", &calo_GID);
+  calo.Branch("PID", &calo_PID);
+
+  for (auto calorimeter : event.getCalorimetersSpan()) {
+    calo_source = calorimeter.getSource();
+    calo_time = calorimeter.getTime();
+    calo_energy = calorimeter.getEnergy();
+    calo_eta = calorimeter.getEta();
+    calo_phi = calorimeter.getPhi();
+    calo_GID = calorimeter.getGIDAsString();
```
shahor02, line 178: what is the reason of using string for GID, which is a simple integer?

shahor02 (commented):
Hi,
Please see comments below. Also in plenty of places you are passing string by copy, please use `const std::string&` instead.
If you prefer, I can merge it as is now and you address questions later.

EventVisualisation/Base/include/EventVisualisationBase/FileWatcher.h
jmyrcha: It is executed once in a program lifetime (may be executed again on source switch, but seldom done) so I prefer code readibility over efficiency - passing by reference will require creating variable just over place where it is called. If you request I can refactor it in next PR (so please merge it as it is now with request to refactor)

EventVisualisation/DataConverter/src/VisualisationEventROOTSerializer.cxx
jmyrcha: It was preserved from the previous version to keep interoperability with previous *.json files. I will address that in next (separate to that) PR when I will be able to test this change only. Please merge as it is and I will fix it soon.

EventVisualisation/Base/include/EventVisualisationBase/FileWatcher.h
shahor02: Not sure what you mean, now you are creating a copy to pass a vector, by passing the reference you create nothing.

jmyrcha: And such variable will be visible in outer space - when it has no usage.
something like:
static std::vector<std::string> extensions = ...

shahor02: there is some misunderstanding: as it is now you are creating 2 clones of the vector: 1st one just to pass the vector, 2nd as a data member of the class.
To get rid of the 1st copy you just need to define your constructor as:
```
 FileWatcher(const std::string& path, const std::vector<std::string>& ext) : mExt(ext) ... {}
```
To get rid of the 2nd one: define your class as 
```
class FileWatcher
{
....
 const std::vector<std::string>& mExt;
};
```

shahor02 (approved):
Merging but please get rid of excessive copies I mentioned.

shahor02 (changes requested):
there are actually conflicts, please rebase to dev and resubmit.

shahor02 (approved):
Tests were passed, merging after rebase.

PR #9498, opened 2022-07-21
Use TPC VDrift from CCDB

shahor02:


Detectors/GlobalTracking/include/GlobalTracking/MatchCosmics.h
```diff
@@ -27,6 +27,7 @@
 #include "GlobalTracking/MatchCosmicsParams.h"
 #include "CommonUtils/TreeStreamRedirector.h"
 #include "TPCFastTransform.h"
+#include "DataFormatsTPC/VDriftCorrFact.h"
```
wiechula: Use also forward declaration here, as below?

davidrohr (dismissed):

wiechula (commented):
Looks fine for me.

wiechula (dismissed):

PR #9502, opened 2022-07-22
[EMCAL-566] Added StatAccumulator for fast time calib

jokonig:
- Standard calibration is based on boost histograms and fitting the time
distribution for each EMCal cell
- New, fast, procedure, calculates mean and sigma on the fly and
therefore, no fitting is needed.
- StatAccumulators could be stored in ccdb for each run to later
calibrate based on multiple runs if needed (for example for the low gain
cells above ~16GeV)


mfasDa (approved):
Looks good to me!

PR #9510, opened 2022-07-23
matchers do not set bunch filling for cosmic runs

shahor02:


shahor02:
all CI's were green, merging after rebase.

PR #9514, opened 2022-07-23
Event Display: Calorimeter proper scaling

mchwes:


shahor02 (approved):
FullCI fails for unrelated and unknown reasons: I've seen already a few times this pattern: HMPID and MID are crashing then ITS, but could not identify the reason. Particularly, it was a reason of the hack: https://github.com/AliceO2Group/AliceO2/commit/52cca075dbc12e59bbd39bcc32f4d6c6ef35a2c0

PR #9515, opened 2022-07-24
tof calib reader loading TFIDInfo

noferini:
Fix for tof calib info reader and matching when calibInfo are asked (diagnostics are now mandatory)
Also sorting TF (per run) when reading them from file


shahor02 (approved):

noferini:
FullC failing because HMP-clusterization and MCH-tracking, not related to this PR

shahor02:
there is a general problem with async which I still don't understand: https://github.com/AliceO2Group/AliceO2/pull/9514#pullrequestreview-1048651964

Looks like memory corruption when creating topology, since e.g. the crash in HMPID or MID depends on the presence of detectors which are not related on these two.

shahor02 (approved):

noferini:
I see, thanks

PR #9517, opened 2022-07-24
TPC residual aggregator removes obsolete files

martenole:
Was not working before, because I forgot to delete one TTree. Thanks a lot for letting me know @shahor02


martenole:
Could this be merged @shahor02 or @davidrohr ?

PR #9521, opened 2022-07-25
Fixes for HMP clusterization

shahor02:
Hi @gvolpe79 I rewrote part of loops but I think there is a lot of room for optimization, there are too many loops over digits and the same conversion (e.g. pad2Absolute) are done over and over. 
The FST passes with these changes, so will merge this PR once any CI has passed. Optimizations can be done later.


shahor02:
FST now pased, merging.  But sim_challenge fails due to the 
```
./sim-challenge.log[INFO] Diagnostic switched on since required for calibInfo time
./sim-challenge.log[ERROR] invalid workflow in o2-tof-matcher-workflow: No matching output found for TOF/DIAFREQ/0 as requested by data processor "TOFCalibWriter". Candidates:
```
@noferini could you please check, should be related to yesterday https://github.com/AliceO2Group/AliceO2/pull/9515

PR #9522, opened 2022-07-26
remove calib output in tof matching by default

noferini:
This should  fix a problem with FST.
calibInfo should be not requested if matching is performed in a different chain wrt digit decoding (MC)


noferini:
All checks passed including FST.
Can we merge? The fix is needed for all other PRs otherwise FST will fail (because ot tof matching)

shahor02 (approved):

PR #9525, opened 2022-07-26
ZDC - Fix raw data checks, better debug output, speedup reconstruction

cortesep:
Found bug in checks performed on ALICE triggers (..were not present in simulation). Improved debug output in case of inconsistent raw data.

### DIFF
--- a/Detectors/ZDC/raw/include/ZDCRaw/DumpRaw.h
+++ b/Detectors/ZDC/raw/include/ZDCRaw/DumpRaw.h
@@ -39,6 +39,7 @@ class DumpRaw
  private:
   void setStat(TH1* h);
   int mVerbosity = 1;
+  TH2* mTransmitted = nullptr;
   TH1* mBaseline[NDigiChannels] = {nullptr};
   TH1* mCounts[NDigiChannels] = {nullptr};
   TH2* mSignal[NDigiChannels] = {nullptr};
--- a/Detectors/ZDC/raw/include/ZDCRaw/DumpRaw.h
+++ b/Detectors/ZDC/raw/include/ZDCRaw/DumpRaw.h
@@ -41,10 +41,13 @@ class DumpRaw
   int mVerbosity = 1;
   TH2* mTransmitted = nullptr;
   TH2* mBits = nullptr;
+  TH2* mBitsH = nullptr;
   TH1* mBaseline[NDigiChannels] = {nullptr};
   TH1* mCounts[NDigiChannels] = {nullptr};
-  TH2* mSignal[NDigiChannels] = {nullptr};
-  TH2* mBunch[NDigiChannels] = {nullptr};
+  TH2* mSignalA[NDigiChannels] = {nullptr};
+  TH2* mSignalT[NDigiChannels] = {nullptr};
+  TH2* mBunchA[NDigiChannels] = {nullptr};
+  TH2* mBunchT[NDigiChannels] = {nullptr};
   EventChData mCh;
 };
 } // namespace zdc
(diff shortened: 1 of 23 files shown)


Detectors/ZDC/raw/src/DumpRaw.cxx
```diff
@@ -56,7 +57,57 @@ void DumpRaw::init()
...
+  double xmax = 2 * NTimeBinsPerBC - 0.5;
+  if (mTransmitted == nullptr) {
+    mTransmitted = (TH2*)gROOT->FindObject("ht");
+  }
+  if (mTransmitted == nullptr) {
+    mTransmitted = new TH2F("ht", "Transmitted channels", NModules, -0.5, NModules - 0.5, NChPerModule, -0.5, NChPerModule - 0.5);
+  }
+  if (mBits == nullptr) {
+    mBits = (TH2*)gROOT->FindObject("hb");
+  }
```
shahor02: @cortesep could you explain where the raw-parser using this class is running and where goes this histo file?
In any case, what is the reason of creating a "stateless" device https://github.com/AliceO2Group/AliceO2/blob/226842d256ae68032c9a82fe1b0fc07511246a6c/Detectors/ZDC/raw/src/raw-parser.cxx#L55 which in fact needs to book a bunch of histos and update them from TF to TF? Also, writing the updated histos to a file at every TF is an overkill. Is this production workflow?

cortesep: @shahor02 This is not a production workflow. It is used on raw data files that are produced at P2 and copied outside the ALICE network. Usually the histograms I develop into Detectors/ZDC/raw/include/ZDCRaw/RawReaderZDC.h are later introduced in the QC worflow (that has, by the way, an excessively long upload latency).

For what concerns the executable raw-parser I just copied an example some time ago and I don't understand well what it is doing. It seems it is launched at every time frame and therefore I have to call gROOT->FindObject at every step in order to continue filling the histograms. Do you have a suggestion how can I improve this? Thanks.

shahor02: > For what concerns the executable raw-parser I just copied an example some time ago and I don't understand well what it is doing. It seems it is launched at every time frame and therefore I have to call gROOT->FindObject at every step in order to continue filling the histograms. Do you have a suggestion how can I improve this? Thanks.

well, the suggestion would be to convert it to usual DPL task which defines histos at init then fills them in the run(..).
But, if this is only for private debugging, I don't really care about its efficiency :)

cortesep: Ok. I will try to modify this.

cortesep: Thanks for the hint. Speed improvement is huge. Now 1 GB raw data file is analyzed in just one minute.

cortesep:
Hi Ruben,

do you think it is ok now?
Thanks,
Pietro

Detectors/ZDC/raw/include/ZDCRaw/DumpRaw.h
```diff
@@ -39,13 +40,17 @@ class DumpRaw
  private:
   void setStat(TH1* h);
   int mVerbosity = 1;
+  TH2* mTransmitted = nullptr;
```
shahor02: Hi @cortesep 
Please, either use unique_ptr for pointers (preferred) or add a destructor where you cleanup the histos

shahor02 (changes requested):

Detectors/ZDC/raw/include/ZDCRaw/DumpRaw.h
cortesep: Hi @shahor02,
Introduced unique_ptr for the histograms

shahor02 (approved):

PR #9529, opened 2022-07-27
TRD: TrackletsParser: Proper zero-ing of array

f3sch:
Small fix to properly zero-out of an array.
Before only the first 3 bytes of a 4*3 byte array where zero-ed.
Now, independent of the size of the array, everything is zero-ed.

Also includes some minor obvious spelling fixes in comments and changes provided by clang-format.

### DIFF
--- a/Detectors/TRD/reconstruction/src/TrackletsParser.cxx
+++ b/Detectors/TRD/reconstruction/src/TrackletsParser.cxx
@@ -21,7 +21,7 @@
 #include "TRDReconstruction/EventRecord.h"
 #include "fairlogger/Logger.h"
 
-//TODO come back and figure which of below headers I actually need.
+// TODO come back and figure which of below headers I actually need.
 #include <cstring>
 #include <string>
 #include <vector>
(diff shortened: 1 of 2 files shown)


Detectors/TRD/reconstruction/src/TrackletsParser.cxx
```diff
@@ -263,13 +263,13 @@ int TrackletsParser::Parse()
...
           if (headertrackletcount > 0) {
-            mState = StateTrackletMCMData; // afrter reading a header we should then have data for next round through the loop
+            mState = StateTrackletMCMData; // after reading a header we should then have data for next round through the loop
           } else {
             mState = StateTrackletMCMHeader;
           }
 
           mcmtrackletcount = 0;
-          std::memset(mTrackletMCMData.data(), 0, 3);
+          std::fill(mTrackletMCMData.begin(), mTrackletMCMData.end(), TrackletMCMData{0});
```
martenole, line 272: Nice catch! Though although this memset was wrong, I am not sure whether this will have an effect, since for each tracklet word, before any data is read from it, it is first overwritten with what comes from the raw data. Is not this reset obsolete?

f3sch: I am not convinced this does anything useful either, but I am unsure about the code paths stuff could take, since it is a bit convoluted.
Can delete on request.

martenole: Fine with me to zero it out like you did. Maybe @bazinski can also comment?

bazinski: I dont think it does anything either, it was there as a cover my ass, when it was crashing for odd reasons online. The crash was assumed to be uninitialised variables (first guess) and hence the zero here.

bazinski (approved):
I dont understand all these clang formatting changes to some code that was correct a short while ago, for the last pr that had this in it.
@f3sch you want to start working on branches not your dev. but of course up to you.

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
f3sch: Should I just get rid of it, or do we accept it as is?

martenole: I'd say we can leave it as you have it here

martenole (approved):

PR #9530, opened 2022-07-27
TRD TRAP charge scaling configurable

martenole:
Should come from TRAP config eventually, but for now to try different settings in simulations this is added as command line parameter


bazinski (approved):

bazinski:
@martenole forgot to add to my approval ... 2 days to go .... ;-)

PR #9535, opened 2022-07-27
DCS2CCDB for ZDC DCS data - attempt n

coppedis:
Code to test DCS2CCDB for ZDC DCS DP

### DIFF
--- a/Detectors/ZDC/calib/macros/CMakeLists.txt
+++ b/Detectors/ZDC/calib/macros/CMakeLists.txt
@@ -14,5 +14,5 @@ o2_add_test_root_macro(
   PUBLIC_LINK_LIBRARIES O2::DetectorsDCS O2::CCDB)
 
 install(
-  FILES makeZDCCCDBEntryForDCS
+  FILES makeZDCCCDBEntryForDCS.C
   DESTINATION share/macro/)
--- a/Detectors/ZDC/calib/macros/CMakeLists.txt
+++ b/Detectors/ZDC/calib/macros/CMakeLists.txt
@@ -1,0 +1,18 @@
+# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+# See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+# All rights not expressly granted are reserved.
+#
+# This software is distributed under the terms of the GNU General Public
+# License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+#
+# In applying this license CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+o2_add_test_root_macro(
+  makeZDCCCDBEntryForDCS.C
+  PUBLIC_LINK_LIBRARIES O2::DetectorsDCS O2::CCDB)
+
+install(
+  FILES makeZDCCCDBEntryForDCS
+  DESTINATION share/macro/)
--- a/Detectors/ZDC/calib/macros/CMakeLists.txt
+++ b/Detectors/ZDC/calib/macros/CMakeLists.txt
@@ -11,8 +11,5 @@
 
 o2_add_test_root_macro(
   makeZDCCCDBEntryForDCS.C
-  PUBLIC_LINK_LIBRARIES O2::DetectorsDCS O2::CCDB)
-
-install(
-  FILES makeZDCCCDBEntryForDCS.C
-  DESTINATION share/macro/)
+  PUBLIC_LINK_LIBRARIES O2::DetectorsDCS O2::CCDB
+  LABELS zdc)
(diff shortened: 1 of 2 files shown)


coppedis:
from the output of fullCI, it seems that the failure doesn't depend on ZDC code, as it says:
## sw/BUILD/o2checkcode-latest/log
--
/sw/SOURCES/O2/9535-slc8_x86-64/0/Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h:39:3: error: use '= default' to define a trivial destructor [modernize-use-equals-default]
--

What am I supposed to do then? 
Thanks for any input!

Detectors/ZDC/calib/src/ZDCDCSProcessor.cxx
```diff
@@ -0,0 +1,410 @@
...
+    mapin[it.id] = it.data;
+  }
+  for (auto& it : mPids) {
+    const auto& el = mapin.find(it.first);
+    if (el == mapin.end()) {
+      LOG(debug) << "DP " << it.first << " not found in map";
+    } else {
+      LOG(debug) << "DP " << it.first << " found in map";
+    }
+  }
```
shahor02: this is never used, please suppress.

Detectors/ZDC/calib/testWorkflow/ZDCDCSConfigProcessorSpec.h
```diff
@@ -0,0 +1,111 @@
...
+    LOG(info) << " ************************* Verbose?" << mVerbose;
+  }
+
+  //---------------------------------------------------------
+
+  void run(o2::framework::ProcessingContext& pc) final
+  {
+    auto configBuff = pc.inputs().get<gsl::span<char>>("inputConfig");
+    auto configFileName = pc.inputs().get<std::string>("inputConfigFileName");
+    auto timer = std::chrono::duration_cast<std::chrono::milliseconds>(HighResClock::now().time_since_epoch()).count();
```
shahor02: please use `std::chrono::system_clock`

Detectors/ZDC/calib/testWorkflow/ZDCDCSDataProcessorSpec.h
```diff
@@ -0,0 +1,224 @@
...
+    }
+    mProcessor->init(vect);
+    mTimer = HighResClock::now();
+  }
+
+  void run(o2::framework::ProcessingContext& pc) final
+  {
+    auto startValidity = DataRefUtils::getHeader<DataProcessingHeader*>(pc.inputs().getFirstValid(true))->creation;
+    auto dps = pc.inputs().get<gsl::span<DPCOM>>("input");
+    auto timeNow = HighResClock::now();
```
shahor02: std::chrono::system_clock

shahor02 (commented):
Hi @coppedis 

The CI failures are indeed unrelated. Please see minor comments below.
Did you test the DCS DP processor with your own generator and the DCS config processor as it is described in https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/DCS/testWorkflow#dcs-config-proxy ?

Detectors/ZDC/calib/testWorkflow/ZDCDCSDataProcessorSpec.h
coppedis: I think I changed what you asked for (I am pushing those right now). I did some test, using o2-calibration-zdc-dcs-sim-workflow, and it seemed to work. Unfortunately I could not use the DCS server emulator since I am having troubles in compiling the emulator...

coppedis:
This is highly needed! I will try to understand what is going wrong here. 
Please, don't close it.
Thanks.

chiarazampolli:
Hello @TimoWilken , @ktf , 
There are plenty of errors recently from this PR (and others). Is something not ok in the CI?
Cheers,
Chiara

TimoWilken:
Hi @chiarazampolli, in this case the error looks genuine to me:

```
Macro Detectors/ZDC/calib/macros/makeZDCCCDBEntryForDCS.C should be tested
```

I expect the solution would be to add a test for this macro, which is added by this pull request.

chiarazampolli:
Hello @TimoWilken , 
Ok, but why did I receive mails every few minutes? 
Cheers,
Chiara

TimoWilken:
The CI process posts a separate comment per commit (as it resets every time something is pushed to this PR) and per check (of which there are 7). Subsequent re-runs of the same checks against the same commit will then "edit" the existing comment instead of posting a new one.

coppedis:
Hi all,

I apologize (and I confess I am a bit lost now). I will wait for Pietro 
next week before updating anything else.

Kind regards, ch.

coppedis:
Will make a brand new one

PR #9539, opened 2022-07-27
TPC IDCs: fixing parallelisation

matthias-kleiner:
- Testing IDCs: add option to load IDCs from file and perform randomization


matthias-kleiner:
@davidrohr can this be merged? This is a temporary fix to avoid that the processing gets stuck when using more than one lane

PR #9540, opened 2022-07-27
ITS/MFT decoder should read ClustererParams from CCDB

shahor02:


shahor02:
The http://alice-ccdb.cern.ch/browse/MFT/Config/ClustererParam and http://alice-ccdb.cern.ch/browse/ITS/Config/ClustererParam objects disabling masking were uploaded to the CCDB but we read from there only by the `its/mft-reco-workflow` but not by the `itsmf-stf-decoder` so the masking with N=1 was applied.

PR #9543, opened 2022-07-28
Add MC labels to ITS seeding vertices

mconcas:
@shahor02 : I restored a mechanism to associate MC labels to seeding vertices. For the time being my goal is to keep all the labels associated to tracklets contributing to a vertex.
In this way it should be easy to track merged vertices too.

~~This still has to be tested (compiling locally), I firstly wanted to ask if the syntax to store them into file is correct and safe.~~

~~**edit**: I tested locally, I see no labels in output file. I am definitely missing something on how to store data with DPL. Investigating.~~

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/include/ITStracking/TimeFrame.h
+++ b/Detectors/ITSMFT/ITS/tracking/include/ITStracking/TimeFrame.h
@@ -65,6 +65,7 @@ class TimeFrame
   const Vertex& getPrimaryVertex(const int) const;
   gsl::span<const Vertex> getPrimaryVertices(int tf) const;
   gsl::span<const Vertex> getPrimaryVertices(int romin, int romax) const;
+  gsl::span<const std::vector<MCCompLabel>> getPrimaryVerticesLabels(const int rof) const;
   gsl::span<const std::array<float, 2>> getPrimaryVerticesXAlpha(int tf) const;
   void fillPrimaryVerticesXandAlpha();
   int getPrimaryVerticesNum(int rofID = -1) const;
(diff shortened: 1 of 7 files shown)


shahor02:
well, dealing with vector of vectors of vectors is already bad for memory and sending vectors of vectors will by quite inefficient. I would simply restrict the tracklet label to the triplet of corresponding cluster labels (there is a small fraction of clusters with multiple labels contributing, but you can select only one of them: if the majority label contributes to the cluster, take it, otherwise it does not matter which wrong cluster label you take). 
But why don't you use simply 1 label per tracklet, assigning the majority MC track index and a flags isFake telling if all the contributors to the tracks share this label?

mconcas:
> well, dealing with vector of vectors of vectors is already bad for memory and sending vectors of vectors will by quite inefficient. 

Agree, I just wanted to see if this went trhough in first place, I'll flatten the structure.

> I would simply restrict the tracklet label to the triplet of corresponding cluster labels (there is a small fraction of clusters with multiple labels contributing, but you can select only one of them: if the majority label contributes to the cluster, take it, otherwise it does not matter which wrong cluster label you take). 
But why don't you use simply 1 label per tracklet, assigning the majority MC track index and a flags isFake telling if all the contributors to the tracks share this label?

Not sure I understood the comment, shouldn't this block do what you mention?
https://github.com/AliceO2Group/AliceO2/blob/e3182f29d7a693a41030b2b49affd10d08e8913c/Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx#L167-L186
There I essentially assign a label if the two clusters of a tracklet share the same label. Default `unset` is left otherwise.
Then I store labels only for L0-L1 tracklets, so that if they are later validated with L1-L2 ones we anyways keep only the labels of tracklets used to "fit" the vertex.
Then my plan is to store the assigned label for each of the contributors of each vertex.

shahor02:
Ah, ok, I did not dig so deep in the assignment code, so, the inner vector in the output vector of vectors is for tracklet labels of single vertex tracklets. Then it is fine provided it is eventually flattened.

mconcas:
> Ah, ok, I did not dig so deep in the assignment code, so, the inner vector in the output vector of vectors is for tracklet labels of single vertex tracklets. Then it is fine provided it is eventually flattened.

Took a bit of of time, now it seems to be working

mconcas:
@shahor02 ok, this should be ready, if you agree we can merge

shahor02 (approved):

PR #9550, opened 2022-07-29
[EMCAL-566] Added parameters for constraining fit range

jokonig:
- Configurable parameters in EMCALCalibParams for constraining the time
window for the fit performed on the cell time distributions
- Needed as previously the time window was hardcoded and the main time
peak was partly outside the window
- Also added parameter to constrain fit range to a window around the
maximum entry in the histogram


mfasDa (approved):
Looks good!

PR #9562, opened 2022-07-30
trace int BC used and candidates in TOF calibInfo

noferini:
This is needed to trace the interaction BC assumed/used to store calib info and possible candidates (int mask).
This will allow to identify shift larger than 1 BC and correct for in calibration workflows also for small (25 ns) bunch spacing


noferini:
All checks passed. Are there comments? Cheers,
Francesco

PR #9578, opened 2022-08-01
Full rapidjson dev

TrifleMichael:
Added two functions to CCDBApi:

browse() - retrieves the /browse/ result from all servers in hostsPool and concatenates them into one string in JSON format. In case of duplicate ids the function takes the object from the server that comes earlier in the hostsPool, in order to mimick a /browse/ response from all servers.

latest() - same as browse() but for the /latest/ call. Removes duplicate ids and paths, prioritizing objects coming from servers earlier in the hostsPool.

If needed, contact me at michal.dominik.trzebuniak@cern.ch


ktf:
Thank you for your contribution. Would you be so kind to squash your commits, please?

TrifleMichael:
Of course, the code should be more readable now.

PR #9581, opened 2022-08-01
Add add function to work with strings

njacazio:
This allows to switch from e.g. 
`histos.add(hnsigma[id].data(), axisTitle, kTH2F, {pAxis, nSigmaAxis});`
to 
`histos.add(hnsigma[id], axisTitle, kTH2F, {pAxis, nSigmaAxis});`
when using arrays to initialize histograms.
This can also be done locally with helper functions but probably it's best to have it in the parent class

### DIFF
--- a/Framework/Core/include/Framework/HistogramRegistry.h
+++ b/Framework/Core/include/Framework/HistogramRegistry.h
@@ -93,7 +93,7 @@ class HistogramRegistry
   HistPtr add(const HistogramSpec& histSpec);
   HistPtr add(char const* const name, char const* const title, const HistogramConfigSpec& histConfigSpec, bool callSumw2 = false);
   HistPtr add(char const* const name, char const* const title, HistType histType, const std::vector<AxisSpec>& axes, bool callSumw2 = false);
-  HistPtr add(std::string_view name, char const* const title, HistType histType, const std::vector<AxisSpec>& axes, bool callSumw2 = false);
+  HistPtr add(const std::string& name, char const* const title, HistType histType, const std::vector<AxisSpec>& axes, bool callSumw2 = false);
 
   template <typename T>
   std::shared_ptr<T> add(char const* const name, char const* const title, const HistogramConfigSpec& histConfigSpec, bool callSumw2 = false);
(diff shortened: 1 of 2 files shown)


njacazio:
@mario-krueger

Framework/Core/include/Framework/HistogramRegistry.h
```diff
@@ -93,6 +93,7 @@ class HistogramRegistry
   HistPtr add(const HistogramSpec& histSpec);
   HistPtr add(char const* const name, char const* const title, const HistogramConfigSpec& histConfigSpec, bool callSumw2 = false);
   HistPtr add(char const* const name, char const* const title, HistType histType, const std::vector<AxisSpec>& axes, bool callSumw2 = false);
+  HistPtr add(const std::string_view& name, char const* const title, HistType histType, const std::vector<AxisSpec>& axes, bool callSumw2 = false);
```
ktf: I would drop the reference and const. string_view is cheap.

Framework/Core/src/HistogramRegistry.cxx
```diff
@@ -114,6 +114,11 @@ HistPtr HistogramRegistry::add(char const* const name, char const* const title,
   return insert({name, title, {histType, axes}, callSumw2});
 }
 
+HistPtr HistogramRegistry::add(const std::string_view& name, char const* const title, HistType histType, const std::vector<AxisSpec>& axes, bool callSumw2)
+{
+  return insert({name.data(), title, {histType, axes}, callSumw2});
```
ktf: a string_view is not necessarily null terminated. Is this taken into account?

Framework/Core/include/Framework/HistogramRegistry.h
njacazio: Thanks, done!

Framework/Core/src/HistogramRegistry.cxx
njacazio: Indeed, I moved to std:: string

njacazio:
Hi @ktf if you agree with the changes I would have this merged

PR #9582, opened 2022-08-01
 [EMCAL-776] Make bad channel scale factors streamable

jokonig:
- To be able to write an object of EMCALChannelScaleFactors to a root
file, add map<EnergyIntervals, float> to LinkDef
- Make class EnergyIntervals independent of EMCALChannelScaleFactors


mfasDa (approved):
Looks good

PR #9580, opened 2022-08-01
Suppress massive sorting in SVertexer

shahor02:
Hi @fmazzasc 

Here is it, could you check if suffers in efficiency wrt your version?  Note that with unconstrained TPC tracks enabled it is still quite slow in high rate pp, but at least it can profits from the multithreding: 
```
o2-secondary-vertexing-workflow $GLO_ARG  --vertexing-sources ITS-TPC,ITS-TPC-TRD,ITS-TPC-TOF,ITS-TPC-TRD-TOF,ITS,TPC,TOF,FT0,TRD --configKeyValues "keyval.output_dir=/dev/null;$VDRIFT;" --threads 6 --hbfutils-config o2_tfidinfo.root

[2791441:secondary-vertexing]: [14:53:57][INFO] Collected 4719724 positive and 3514131 negative seeds
[2791441:secondary-vertexing]: [14:57:32][INFO] Found 87135 V0s and 56093 cascades, timing: CPU: 1289.3 Real: 217.468 s
[2791443:secondary-vertex-writer]: [14:57:32][INFO] SecondaryVertexWriter pulled 87135 v0s
[2791441:secondary-vertexing]: [14:57:32][INFO] Secondary vertexing total timing: Cpu: 1.289e+03 Real: 2.175e+02 s in 1 slots, nThreads = 6
[2791443:secondary-vertex-writer]: [14:57:32][INFO] SecondaryVertexWriter pulled 56093 cascades
```

PR #9588, opened 2022-08-02
Release cuts in SVertexer

fmazzasc:
PR #9580 already rebased.

In this P.R. few cuts are released:
- DCAtoPV of daughter tracks from 0.1 to 0.05 (most intensive operation from computational point of view)
- DCAz and DCAXY of the cascade from 0.3 to 0.5
- maxDCAXYToMeanVertexV0Casc from 0.5 to 2
- Causality cut for bachelor removed as it is wrong since it does not take into account the geometry of the ITS. For example the 16 cm gap from L2 to L3 causes the loss of more than 50% of the bachelors, when the cut is applied.

The increase in CPU time has been evaluated in pp and Pb-Pb collisions. In pp the increase is negligible

**pp collisions at 500kHz**
- old cuts: Secondary vertexing total timing: Cpu: 1.900e+00 Real: 1.976e+00 s in 1 slots, nThreads = 1
- new cuts: Secondary vertexing total timing: Cpu: 1.920e+00 Real: 1.992e+00 s in 1 slots, nThreads = 1

**Pb-Pb collisions at 50kHz**
- old cuts: Secondary vertexing total timing: Cpu: 4.300e+00 Real: 4.376e+00 s in 1 slots, nThreads = 1
- new cuts: Secondary vertexing total timing: Cpu: 9.720e+00 Real: 9.797e+00 s in 1 slots, nThreads = 1

-----------------

The improvement in reconstruction efficiency for the Xi baryon is shown in the attached plot (more than a factor 2 increase)
c1

### DIFF
--- a/Detectors/Vertexing/include/DetectorsVertexing/SVertexer.h
+++ b/Detectors/Vertexing/include/DetectorsVertexing/SVertexer.h
@@ -122,8 +122,6 @@ class SVertexer
   std::vector<std::vector<Cascade>> mCascadesTmp;
   std::array<std::vector<TrackCand>, 2> mTracksPool{}; // pools of positive and negative seeds sorted in min VtxID
   std::array<std::vector<int>, 2> mVtxFirstTrack{};    // 1st pos. and neg. track of the pools for each vertex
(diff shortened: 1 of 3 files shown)


shahor02 (approved):
failures unrelated.

PR #9590, opened 2022-08-03
Alpide simulation response for Vbb=0V and Vbb=-3V

AlexBigO:
Implementation of the Vbb=0V charge-collection tables (on top of the Vbb=-3V ones already there).
Conversion of charged-collection tables from 810 .txt files to one .json file for each Vbb value.


iouribelikov:
@ktf @sawenzel @shahor02 
Hi Giulio, Sandro, Ruben.  Could not you help us with the "whitespace" check, please ?
There are two kinds of reported errors: "Indent code using spaces instead of tabs" and "Remove the trailing spaces at the end of the line".   Yes, one may guess what it actually means.   But is there a way to see the line number(s) where these errors happen ?  Or, what kind of "grep" should be applied, in order to find the problematic lines ?
Thank you very much,
Yura.

iouribelikov (dismissed):
For additional details about this PR, have a look at these slides:
https://indico.cern.ch/event/1186795/contributions/4986682/attachments/2487941/4272233/its_meeting_02-08-22.pdf

sawenzel:
Hi. I understand the motivation behind this transformation (reading many files from CVMFS over the network or local file system is not ideal) and thanks for reacting to my initial comments !

However, I am personally not convinced that the new proposition is a step forward. Essentially I see that

a) we have **4 million** of new lines of text to treat in git (maybe not critical but still maybe not necessary)

b) have text files of considerable size (22MB) that still need to be fetched from CVMFS ...**and also parsed** with a json parser during runtime. Note that the size of the new json file is also considerably larger than all the original 810 files taken
together.

Wouldn't it be more useful to convert these textual data into actual compressed binary objects ? I think we can by all means have a hybrid approach here:

a) keep text files that you can modify in git

b) convert these text files to a small compressed binary file for instance containing ROOT objects (by calling a conversion step during build)

c) install binary files as part of installation, not text files (eventually one could also install them via CCDB)

d) use these ROOT objects directly in the code, without needing to parse large text files

I cannot judge how critical this PR is now ... but maybe this is something that could be considered.

shahor02:
Hi, I agree with @sawenzel , I would consider converting it to CCDB object.

ktf:
@sawenzel I see there is a number of PRs which fail all with:

```
## sw/BUILD/O2-latest/log
448/448 Test #404: o2sim_hepmc ..........................................................................***Timeout 400.02 sec
```

I assume this is some timeout CCDB related? Or are we at the limit for that test?

PR #9592, opened 2022-08-03
DPL: fix InputProxy EoS

ktf:
An input proxy can be configured so that the out of band channel
has many connections for the same channel. We therefore need to
account for it when we try to propagate the end of stream, doing
so only when we have received as many end of streams as we have open
connections on the input channel.

### DIFF
--- a/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
+++ b/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
@@ -279,7 +279,8 @@ InjectorFunction dplModelAdaptor(std::vector<OutputSpec> const& filterSpecs, DPL
         if (DataSpecUtils::match(spec, OutputSpec{{header::gDataOriginAny, header::gDataDescriptionAny}}) ||
             DataSpecUtils::match(spec, query)) {
           channelName = channelRetriever(query, dph->startTime);
-          if (channelName.empty()) {
+          // We do not complain about DPL/EOS/0, since it's normal not to forward it.
+          if (channelName.empty() && DataSpecUtils::describe(query) != "DPL/EOS/0") {
             LOG(warning) << "can not find matching channel, not able to adopt " << DataSpecUtils::describe(query);
           }
           break;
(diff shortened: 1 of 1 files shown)


ktf:
@Barthelemy @knopers8 @dennisklein @teo @davidrohr this is part 1 of what we discussed. How can we check it?

knopers8:
thanks! if you build also QC and testing without a control system is enough, you can do this in three separate terminals:
```
o2-qc-run-producer --message-amount 100 | o2-qc --config json:///absolute/path/to/config.json --local --host localnode1 -b

o2-qc-run-producer --message-amount 100 | o2-qc --config json:///absolute/path/to/config.json --local --host localnode2 -b

o2-qc --config json://o2-qc --config json:///absolute/path/to/config.json --remote -b
```
where the json is:
```
{
  "qc": {
    "config": {
      "database": {
        "implementation": "CCDB",
        "host": "ccdb-test.cern.ch:8080",
        "username": "not_applicable",
        "password": "not_applicable",
        "name": "not_applicable"
      },
      "Activity": {
        "number": "42",
        "type": "2"
      },
      "monitoring": {
        "url": "infologger:///debug?qc"
      },
      "consul": {
        "url": ""
      },
      "conditionDB": {
        "url": "ccdb-test.cern.ch:8080"
      }
    },
    "tasks": {
      "MultiNodeLocal": {
        "active": "true",
        "className": "o2::quality_control_modules::skeleton::SkeletonTask",
        "moduleName": "QcSkeleton",
        "detectorName": "TST",
        "cycleDurationSeconds": "10",
        "maxNumberCycles": "-1",
        "dataSource_comment": "The other type of dataSource is \"direct\", see basic-no-sampling.json.",
        "dataSource": {
          "type": "dataSamplingPolicy",
          "name": "rnd-many"
        },
        "taskParameters": {},
        "location": "local",
        "localMachines": [
          "localnode1",
          "localnode2"
        ],
        "remoteMachine": "localhost",
        "remotePort": "30132",
        "mergingMode": "delta"
      }
    },
    "checks": {}
  },
  "dataSamplingPolicies": [
    {
      "id": "rnd-many",
      "active": "true",
      "machines": [
        "localnode1",
        "localnode2"
      ],
      "query": "random:TST/RAWDATA",
      "samplingConditions": [
        {
          "condition": "random",
          "fraction": "0.5",
          "seed": "1234"
        }
      ],
      "blocking": "false"
    }
  ]
}
```

ktf:
So, I just tried @knopers8 suggested test:

```
o2-qc-run-producer --message-amount 100 | o2-qc --config json:///$PWD/config.json --local --host localnode1 -b --completion-policy=wait
o2-qc-run-producer --message-amount 300 | o2-qc --config json:///$PWD/config.json --local --host localnode2 -b --completion-policy=wait
o2-qc --config json:///$PWD/config.json --remote -b
```

in three different shells and indeed the third workflow quits after 30s and not after 10. What I do not understand is the failing unit tests, as the same test works fine on my box. I am now trying to check with the static analyser.

Framework/Core/src/ExternalFairMQDeviceProxy.cxx
```diff
@@ -435,8 +436,21 @@ DataProcessorSpec specifyExternalFairMQDeviceProxy(char const* name,
       });
     };
     ctx.services().get<CallbackService>().set(CallbackService::Id::Start, channelConfigurationChecker);
+    static int numberOfEoS = 0;
+    numberOfEoS = 0;
```
knopers8, line 440: What if we put two proxies into one process? IIRC, this is/will be possible.

ktf: Ok, I can make the number of EoS received a property of the ChannelInfo. Let me refactor this a  bit.

ktf:
Merging this for now, since it solves the issue with the input proxy shutting down too early. I have added the support for multiple input proxies on my todo list.

PR #9599, opened 2022-08-04
Add verbose option for ITS DCS parser

ezradlesser:
Adding verbose option to the ITS DCS parser workflow which prints out the string being parsed by the workflow. This will be useful for debugging strings which are not properly parsed by the workflow.


ezradlesser:
Tagging @shahor02 for review -- many thanks :)

chiarazampolli:
Adding also @davidrohr , since @shahor02  is away.

ezradlesser:
The fullCI build error does not seem to be related to this PR.

PR #9600, opened 2022-08-04
Correctly flag FET for MID REG boards

dstocco:
In self-triggered events, the regional response is delayed compared to the local board one.
This was correctly accounted for in most cases, except when flagging the FET event.
Since the regional response is not used, the bug is harmless for physics, but it is of course better to fix it.


pillot (approved):

dstocco:
Errors seems unrelated. @ktf is it a known issue?

ktf:
I thought it was a  known issue in QC which @Barthelemy already fixed? Bart?

Barthelemy:
it is fixed but it seems that the test ran last Saturday.

dstocco:
Should I rebase and push to trigger a check? But I see that there are unrelated failures on other PRs...so not sure that the check will be successful...

ktf:
I just retriggered them.

PR #9601, opened 2022-08-05
[EMCAL-794] Added cell data producer

jokonig:
- data producer has emcal cells and emcal trigger as outputs
- cell energy and time can either be randomly created from exponential
and normal distributions or via a root input histogram from which random
values are taken
- the data producer is written for the bad channel and time calibration,
can however be used for all worflows which take emcal cells as input
- small fix in channel calibrator to recreate local root file if no file
exists yet


mfasDa:
CI error unrelated. @shahor02 @chiarazampolli @davidrohr @ktf could this be merged?

mfasDa (approved):
Sorry, forgot to submit the review and just approved as comment. Done now.

PR #9602, opened 2022-08-05
[EMCAL-769, EMCAL-795] Handle cases where bunch size would be 0

mfasDa:
In this case the bunch header is present (bunch word 0
check passed) but no payload is reported.

### DIFF
--- a/Detectors/EMCAL/reconstruction/src/AltroDecoder.cxx
+++ b/Detectors/EMCAL/reconstruction/src/AltroDecoder.cxx
@@ -122,6 +122,14 @@ void AltroDecoder::readChannels()
         // we must break here as well, the bunch is cut and the pointer would be set to invalid memory
         break;
       }
+      if (bunchlength == 0) {
+        // skip bunches with bunch size 0, they don't contain any payload
+        // Map error type to NULL header since header word is null
+        mMinorDecodingErrors.emplace_back(MinorAltroDecodingError::ErrorType_t::BUNCH_HEADER_NULL, channelheader, 0);
+        // Forward position by bunch header size
+        currentsample += bunchlength + 2;
+        continue;
+      }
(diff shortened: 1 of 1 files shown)


mfasDa:
Assessing the source of the crash described in https://github.com/AliceO2Group/QualityControl/pull/1393. Marking PR as WIP as we are not fully sure the fix cures the problem.

mfasDa:
Luckily over the weekend we got data with the same symptom - run 522805. Indeed we found bunch lengths 0 in this data in one FEC and indeed they are responsible for the crash we observed. Offline tests showed that the protection indeed works. The PR is ready to be merged and needs to be deploy as hotfix the earliest possible.

martenole (approved):

martenole:
I am not sure why the CI is taking so long here. fullCI running already since 3 hours...
@mfasDa the `AltroDecoder` is runs on the EPNs, not on the FLPs, right? So this is not something which must be included in the FLP suite, but it should be included in the O2 version which runs on the EPNs? Just wanted to clarify, still think we can merge this today. But the fullCI I'd like to have green

mfasDa:
@martenole Yes, the AltroDecoder runs on the EPN. It should enter the next tag deployed, it is quite critical for us. Though the PR in QC linked above and already tagged as emergency patch release prevents the QC from merging this commit is also urgently needed since we need to exclude these data in general and need to monitor the card responsible for the data corruption. Therefore we would prefer having this available as soon as possible and not wait for sneaking in with other higher priority fixes.

I think the fullCI is done now. The other tests failures are I think unrelated.

martenole:
Indeed, all CI failures are unrelated and fullCI is green.
@pzhristov could you please merge this PR?

pzhristov:
The failing tests are unrelated, merging.

PR #9604, opened 2022-08-06
IDCs: Correctly handle dropped data

matthias-kleiner:
- removing not needed code
- code optimisations
- perform correct factorisation of IDCs in case of dropped data
- using reasonable time stamp when using IDC generator


matthias-kleiner:
Can this be merged? The error seems unrelated

PR #9610, opened 2022-08-09
Improve MFT linear track fitting + configurable misalignment residuals

rpezzi:
This PR enables Kalman filter for linear MFT tracks and introduces a misalignment residual parameter.


shahor02 (approved):
at the next PR better change std::cout to LOG, even if the verbosity needs to activated from the command line.

PR #9612, opened 2022-08-09
Fix compatibility issue with omp

ktf:


ktf:
@mpuccio @mconcas could you check this? someone on the support mailing list was having troubles compiling the current version, and indeed I think old versions of omp do not support range loops. Since the fix is trivial enough, would you mind it, even if less elegant? Besides that, it looks like you lack a const version of `Road::operator[]` which I guess it might make sense.

mconcas (approved):

mpuccio:
It's fine, but Catalina is not supported anymore and people should really upgrade.

ktf:
Yes, agreed, that's what I wrote on the mailing list.

PR #9613, opened 2022-08-09
[EMCAL-784] Sampling factor for different physics list for Geant4

mfasDa:
Sampling factors for all supported Geant4 physics lists in
O2. Determined with Geant4 v10.7.


mfasDa:
We determined the sampling factors for all physics lists based on the non-linearity functions for electrons and photons using Geant4 v10.7. All physics lists show similar responses for EMCAL, resulting in similar scaling factors. Results are also consistent between electrons and photons. (More information: https://alice.its.cern.ch/jira/browse/EMCAL-784)

We will also now have a look again at Geant4 v11.0.2 in order to see whether the issue with the energy shift observed previously is resolved, and determine the sampling factors again with the new version

mfasDa:
Error unrelated

shahor02 (approved):

PR #9615, opened 2022-08-10
[EMCAL-566] Add is_messagable to calib params

jokonig:
- is_messagable added to calib params in order to prepare to read them
from ccdb (this will need additional changes in the CalibratorSpec which
will be added as soon as the ccdb object exists)
- Changed the default value of the minimum number of events needed to
trigger the calibration. The 10 million events was tested at point 2 and
gives reasonable results


mfasDa (approved):
Looks good!

mfasDa:
@shahor02 @chiarazampolli @davidrohr @ktf this commit should be ready

PR #9616, opened 2022-08-10
[EMCAL-566] print cellID if fit in time calib fails

jokonig:
- cellID is now printed if the fit of the time distribution of this cell
fails.
- Can be usefull for debugging etc.
- Additionally the time calibration parameter that is assigned to this
cell is printed (per default the value of the previous cell is used)


mfasDa (dismissed):
Looks good!

mfasDa (approved):
Looks good!

mfasDa:
Finally the CI finished - forwarding to @shahor02, @davidrohr or @chiarazampolli for final merge.

This is a rather simple commit, it just adds more information about the channel ID in case a fit fails.

PR #9617, opened 2022-08-10
getNbins, typos, warning for too many bins

jgrosseo:


jgrosseo:
@mario-krueger @njacazio Is this ok for you?

mario-krueger:
Looks good to me. (probably .has_value() is not necessary)

jgrosseo:
I liked it more than the cast to bool. Or what do you have in mind?

mario-krueger:
There should be no need to cast it. If you just put if(nBins) it should be the same as if(nBins.has_value())
But it since both are the same, it does not really matter what is used.

jgrosseo:
I see (ah yes, because *nBins is the value in any case). Anyway, like this it is very clear. I leave it.

njacazio:
Looking good for me, thanks!

PR #9618, opened 2022-08-10
DPL: drain messages in READY

ktf:


ktf:
@Barthelemy @knopers8 @davidrohr here is the rest of the code we discussed.
One question: what do we want this only when transitioning from Running to Ready or in general?

davidrohr:
Don't fully understand. You mean only in the READY state if we came from running before, or in the READY state in general? I think in that case we could drop message in the ready state in general.

Barthelemy:
yes, I was also thinking that we drain in the READY state, it is not related to a transition but a state.

Barthelemy:
By the way, I am not sure that it is configurable as it is. Is it ?

ktf:
Well, you can enter ready from either the startup sequence or for running. Basically the question is: do we want to drain the first time we enter in ready?

ktf:
> By the way, I am not sure that it is configurable as it is. Is it ?

right, I need to add that as well...

davidrohr:
I think when the draining is enabled, it should always be enabled - to be consistent.
Could you please make the default to "not drain", so we set it explicitly for the QC input proxies and nowhere else?

ktf:
Ok, deal. coming in a bit.

ktf:
Optional and disabled by default now.

ktf:
Error unrelated. Merging.

PR #9624, opened 2022-08-11
Fix bug in the ROF records of MID dead channels

dstocco:
There was a mismatch between the dead channels and the corresponding ROF records. This was not spotted because teh ROF records were not used downstream. But the bug was crashing the QC. Now fixed.


pillot (approved):

dstocco:
All checks green. Can this be merged?

PR #9625, opened 2022-08-11
IDCs: Updating IDC documentation

matthias-kleiner:


martenole (approved):
Thanks, this looks good. Once the fullCI is through this should be merged

martenole:
Not sure why the fullCI won't start here :(

martenole:
@ktf do you know why the CI is not starting here? Could you manually start it? It would be good to have this in tomorrow morning for testing at P2

matthias-kleiner:
> @ktf do you know why the CI is not starting here? Could you manually start it? It would be good to have this in tomorrow morning for testing at P2

Would it make sense if I just put the fixes in the aggregator-workflow.sh into a separate PR? Maybe not all tests are required then?

chiarazampolli:
I don't think that the test depend on which file is touched, but better that @ktf confirms.

ktf:
I am looking, but it it  touches only the aggregator, I do not think it really matters.

martenole:
I would not split only the aggregator-workflow part. With the current O2@dev, if you try `o2-tpc-idc-factorize --help` you get a FATAL error complaining `o2tpcaveragegroupidc_configuration.ini : config file does not exist!`. This is also fixed in this PR by setting the default to no config file.
I tried the FST manually with the changes from this PR on the EPN, it was fine. I would suggest to force merge if it is fine with @ktf and/or @sawenzel

PR #9626, opened 2022-08-11
DPL: drop wrong assertion

ktf:


ktf:
For some reason now handleData can be invoked with 0 entries under rare circumstances. We therefore drop the assert since the code is in any case a no-op for parts.Size() == 0.

ktf:
Merging this. I've verifies that running with 0 parts works just fine and will commit a test in a subsequent PR..

PR #9629, opened 2022-08-12
O2DatabasePDG service

sawenzel:
This commit concerns the PDG database.
It modifies/restructures existing code (from Generator/PDG) and wraps it into a new
service "O2DatabasePDG". The new service returns a modified
TDatabasePDG instance, in which ALICE particles have been added.

Existing code can simply transform from the use of "TDatabasePDG::Instance"
to "O2DatabasePDG::Instance". This transition is made easy via this commit.

Also touching the DPL service that returns TDatabasePDG, so that O2Physics, Analysis
automatically get the correctly augmended object.

MCTrack has been modified to use the new service, fixing issues with wrong
masses.


sawenzel:
@mpuccio

aalkin:
There is already PDG service (used in analysis), defined as 
`o2::framework::ServiceSpec CommonAnalysisServices::databasePDGSpec()`
in CommonServices.cxx:942.

Can you use it?

sawenzel:
@aalkin : I am not introducing a new service. Merely modifying the existing one (the one you mention) to include special ALICE particles (which TDatabasePDG does not have).

aalkin:
@sawenzel we should probably merge these services into one at some point. Adding particles should be a matter of customizing the init callback for the analysis service (and we'll probably need this customizations in analysis anyway).

PR #9631, opened 2022-08-14
DPL: use array rather than vector in homogeneous_apply_ref

ktf:
Should be simpler and might even be constexpr if the
lambda is constexpr.


jgrosseo:
This is faster?

ktf:
Seems so, and it uses a bit less memory as well, although it's hard to tell... For sure it avoids one allocation at runtime...

PR #9634, opened 2022-08-15
TRD enforce tracklet mcm header sanity check

bazinski:
- correct the sanity check of the tracklet mcm header, and now enforce it.

### DIFF
--- a/DataFormats/Detectors/TRD/src/RawData.cxx
+++ b/DataFormats/Detectors/TRD/src/RawData.cxx
@@ -171,12 +171,28 @@ uint32_t getHCIDFromTrackletHCHeader(const TrackletHCHeader& header)
   return header.layer * 2 + header.stack * constants::NLAYER * 2 + header.supermodule * constants::NLAYER * constants::NSTACK * 2 + header.side;
 }
 
-uint32_t getChargeFromRawHeaders(const o2::trd::TrackletHCHeader& hcheader, const o2::trd::TrackletMCMHeader* header, const std::array<o2::trd::TrackletMCMData, 3>& data, int pidindex, int trackletindex)
+int getNumberOfTrackletsFromHeader(const o2::trd::TrackletMCMHeader* header, bool verbose)
+{
+  int headertrackletcount = 0;
+  if (header->pid0 != 0xff) { // pid of cpu0
+    headertrackletcount++;
+  }
+  if (header->pid1 != 0xff) { // pid of cpu1
+    headertrackletcount++;
+  }
+  if (header->pid2 != 0xff) { // pid of cpu2
+    headertrackletcount++;
+  }
+  return headertrackletcount;
+}
+
+int getChargesFromRawHeaders(const o2::trd::TrackletHCHeader& hcheader, const o2::trd::TrackletMCMHeader* header, const std::array<o2::trd::TrackletMCMData, 3>& data, std::array<uint8_t, 3>& q, int trackletindex)
 {
   uint32_t pid = 0;
   uint32_t highPID = 0; // highPID holds the 8 bits from the mcmheader
   uint32_t lowPID = 0;  // lowPID holds the 12 bits from mcmdata
   uint32_t datatype = (hcheader.format) >> 2;
+  q.fill(0xff);
   switch (datatype) {
     case 0: //Cosmic
             // LOG(warn) << "This is a problem cosmic format tracklets ";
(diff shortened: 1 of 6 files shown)


martenole:
Why do you say you correct the sanity check? Aren't you just taking out the check on the upper PID bits here? Or is it now understood why that check so often fails?

bazinski:
@martenole well i am fixing it, before there was the assumption that the pid came in sequential order, when in fact it is cpu0 cpu1 cpu2 order in the header. As it is now correct, the sanity check is actually enforced.

martenole:
Ok so you say that if a MCM sends only 1 tracklet, it might come from any CPU, not just the first one?

bazinski:
exaclty. lines 370 to 458 of fit_tracklets.asm

martenole:
Very good that this is found! But then we should now also fix the tracklet parsing, since then this is also wrong, no?
https://github.com/AliceO2Group/AliceO2/blob/4f958e2bd6a3dbf0f70a659850dfbca7bce384ce/DataFormats/Detectors/TRD/src/RawData.cxx#L195-L208

bazinski:
ah good point, thought I had missed something.

f3sch:
I actually thought, the tracklets get sorted according to the number of hits and then get distributed on the cpu0-2 for integration?
This at least what I recall from the trap simulator in trap_trap.cpp from venelin.
But I do not find anything related to that in the asm.
So ... good job? :+1:

martenole (dismissed):

bazinski:
@martenole its late will finalise the unit tests later this morning.

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -171,12 +171,28 @@ uint32_t getHCIDFromTrackletHCHeader(const TrackletHCHeader& header)
   return header.layer * 2 + header.stack * constants::NLAYER * 2 + header.supermodule * constants::NLAYER * constants::NSTACK * 2 + header.side;
 }
 
-uint32_t getChargeFromRawHeaders(const o2::trd::TrackletHCHeader& hcheader, const o2::trd::TrackletMCMHeader* header, const std::array<o2::trd::TrackletMCMData, 3>& data, int pidindex, int trackletindex)
+int getNumberOfTrackletsFromHeader(const o2::trd::TrackletMCMHeader* header, bool verbose)
```
f3sch, line 174: We have this nice verbose flag, maybe we should use it?

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -192,60 +208,57 @@ uint32_t getChargeFromRawHeaders(const o2::trd::TrackletHCHeader& hcheader, cons
...
-        case 2:
-          highPID = header->pid2;
-          break;
-        default:
-          LOG(warn) << " unknown trackletindex of " << trackletindex << " to " << __func__ << " : " << pidindex;
-          break;
+      //trackletindex is the index into the [0:2] range of tracklets that can be attached to the mcm.
+      //jump to cpu:
+      uint32_t hpid = header->pid0 | (header->pid1 << 8) | (header->pid2 << 16);
+      std::array<uint16_t, 3> hpidvalues;
```
f3sch, line 214: Should rather be uint8_t.

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -192,60 +208,57 @@ uint32_t getChargeFromRawHeaders(const o2::trd::TrackletHCHeader& hcheader, cons
...
+      int dynamicq = hcheader.format & 0x1;         // last bit of format (lsb) defines the version of tracklet charge calculation
+      uint32_t pidoffset = ((pidword >> 18) & 0x3); // used for dynamic ranged charge windows, may or may not be used below.
+      if (!dynamicq) {
+        q[2] = (pidword >> 14) & 0x3f; // 6 bits at the top of all of pid (MSB)
+      } else {
+        q[2] = (pidword >> 12) & 0x3f; // 6 bits of Q2 and a shift
+        q[2] |= pidoffset << 6;
+        // LOG(info) << "Q2 pid : " << std::hex << pid << " pidoffset: "  << pidoffset;
+      }
           if (!dynamicq) {
```
f3sch, line 246: Why is the indentation weird here?
Also wouldnt a single if(!dynamicq) suffice?

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -192,60 +208,57 @@ uint32_t getChargeFromRawHeaders(const o2::trd::TrackletHCHeader& hcheader, cons
...
+        hpidvalues[counter++] = header->pid0;
+      }
+      if (header->pid1 != 0xff) {
+        hpidvalues[counter++] = header->pid1;
+      }
+      if (header->pid2 != 0xff) {
+        hpidvalues[counter++] = header->pid2;
       }
+      //hpidvalues now holds the sequential list of high pid values related to the subsequent sequence of trackletmcmdata words
+      highPID = hpidvalues[trackletindex];
```
martenole, line 227: This is not correct. In case CPU0 does not have a tracklet, but CPU1 does, you would still take the header PID information for the first tracklet from CPU0 (which in that case would be 0xff). I have a suggestion on how one could do it, we can check together later today

martenole (changes requested):

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: but 0xff in pid first pid means the cpu0 does not have a tracklet. The first trackletword after the header is from cpu1.

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: if would, but I did wanted to change the code as little as possible.

DataFormats/Detectors/TRD/src/RawData.cxx
martenole: Indeed, since you fill the highPID only with the PID information when its not `0xff` this is also fine

martenole (approved):

PR #9638, opened 2022-08-15
TPC dE/dx: using nD-piecewise polynomials instead of polynomials

matthias-kleiner:
The following new calibration object is needed to apply this correction:
[Tracktopology_piecewise_scaled.root.zip](https://github.com/AliceO2Group/AliceO2/files/9338706/Tracktopology_piecewise_scaled.root.zip)

### DIFF
--- a/Detectors/TPC/base/include/TPCBase/CDBInterface.h
+++ b/Detectors/TPC/base/include/TPCBase/CDBInterface.h
@@ -101,7 +101,7 @@ const std::unordered_map<CDBType, const std::string> CDBTypeMap{
   {CDBType::CalGas, "TPC/Calib/Gas"},
   {CDBType::CalTemperature, "TPC/Calib/Temperature"},
   {CDBType::CalHV, "TPC/Calib/HV"},
-  {CDBType::CalTopologyGain, "TPC/Calib/TopologyGain"},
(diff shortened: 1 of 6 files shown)


matthias-kleiner:
@wiechula can you test on the EPNs the impact on the performance of the new correction?

matthias-kleiner:
Hello @chiarazampolli,  

with this PR the static calibration which is used for the TPC dE/dx (http://alice-ccdb.cern.ch/browse/TPC/Calib/TopologyGain) will be changed, which means that with this PR the old object cannot be loaded and applied anymore.

This would also mean that the http://alice-ccdb.cern.ch/browse/TPC/Calib/TimeGain calibration objects created with the old TopologyGain calibration object would not perfectly match the new TopologyGain correction. Reconstructing the dE/dx with the new TopologyGain calibration and the objects currently in the TimeGain folder would therefore result in a slight mismatch and the dE/dx could deteriorate. 

Do you think it would be reasonable to just put the new calibration object in a new folder in the CCDB that will be loaded and applied for the TPC dE/dx when this PR is merged, or do you think there is a better way to handle this? It would be also possible to upload the new TopologyGain calib object with an appropriate timestamp range, but then the reconstruction wouldnt apply the TopologyGain correction if one wants to run the reconstruction on newer timestamps with an older o2 version...

PR #9643, opened 2022-08-16
[FDD] Use of TVDC bit in average time calculation in reconstruction

arvindkhuntia:


Detectors/FIT/FDD/reconstruction/src/Reconstructor.cxx
```diff
@@ -39,26 +40,31 @@ o2::fdd::RecPoint Reconstructor::process(o2::fdd::Digit const& digitBC,
...
     Float_t adc = outChData[ich].mChargeADC;
     Float_t time = outChData[ich].mTime;
+    bool inTime = inChData[ich].getFlag(ChannelData::EEventDataBit::kIsEventInTVDC);
     if (time == o2::InteractionRecord::DummyTime) {
       continue;
     }
     Float_t timeErr = 1;
-    if (adc > 1) {
+    if (adc > 3) {
       timeErr = 1 / adc;
```
mslupeck, line 49: Did you forget a `.` to make sure that 1/adc is calculated as float and not integer?
timeErr = 1. / adc;

mslupeck (approved):

Detectors/FIT/FDD/reconstruction/src/Reconstructor.cxx
arvindkhuntia: Hi @mslupeck, thanks for pointing out this, however the ADC here is a float type that makes the return value of timeErr as float.

arvindkhuntia:
Hi @shahor02  Could you please merge this branch.

Thanks and regards,
Arvind

shahor02 (approved):

PR #9644, opened 2022-08-16
Fixing mUpdateCCDB flag, otherwise it invalidates the current object

chiarazampolli:
@noferini 
@martenole : this fixes the crash that you saw some days ago at P2
@davidrohr : this should also fix the OOM issue with the LHCclock calibration.


martenole:
Thanks a lot for looking into this and for the fix of course!

PR #9646, opened 2022-08-16
[EMCAL-797] Read calibration params from ccdb

jokonig:
- Added functionality to read calib params from ccdb in he calibratorSpec
- Object is already uploaded to ccdb (O2-3116)
- This will overwrite the default values in the CalibParams but can be
overwritten if secified via the configurableParams
- By default, the object is taken from the ccdb, one can switch it of by
adding the workflow argument --no-loadCalibParamsFromCCDB

### DIFF
--- a/Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
+++ b/Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
@@ -31,6 +31,8 @@
 #include "CCDB/CcdbObjectInfo.h"
 #include "CommonUtils/NameConf.h"
 #include "DetectorsBase/GRPGeomHelper.h"
+#include "Framework/CCDBParamSpec.h"
+
 // for time measurements
 #include <chrono>
 
@@ -47,7 +49,7 @@ class EMCALChannelCalibDevice : public o2::framework::Task
   using EMCALCalibParams = o2::emcal::EMCALCalibParams;
 
  public:
-  EMCALChannelCalibDevice(std::shared_ptr<o2::base::GRPGeomRequest> req) : mCCDBRequest(req) {}
+  EMCALChannelCalibDevice(std::shared_ptr<o2::base::GRPGeomRequest> req, bool params) : mCCDBRequest(req), mLoadCalibParamsFromCCDB(params) {}
 
   void init(o2::framework::InitContext& ic) final
   {
@@ -87,8 +89,13 @@ class EMCALChannelCalibDevice : public o2::framework::Task
   void finaliseCCDB(o2::framework::ConcreteDataMatcher& matcher, void* obj) final
   {
     o2::base::GRPGeomHelper::instance().finaliseCCDB(matcher, obj);
+    // check if calib params need to be updated
+    if (matcher == ConcreteDataMatcher("EMC", "EMCALCALIBPARAM", 0)) {
+      LOG(info) << "EMCal CalibParams updated";
+    }
   }
 
+  //_________________________________________________________________
   void run(o2::framework::ProcessingContext& pc) final
   {
     if (EMCALCalibParams::Instance().enableTimeProfiling) {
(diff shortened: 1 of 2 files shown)


Detectors/EMCAL/calibration/testWorkflow/emc-channel-calib-workflow.cxx
```diff
@@ -36,7 +36,8 @@ using namespace o2::emcal;
 void customize(std::vector<ConfigParamSpec>& workflowOptions)
 {
   std::vector<ConfigParamSpec> options{
-    {"configKeyValues", VariantType::String, "", {"Semicolon separated key=value strings"}}};
+    {"configKeyValues", VariantType::String, "", {"Semicolon separated key=value strings"}},
+    {"no-loadCalibParamsFromCCDB", VariantType::Bool, false, {"if true, load the EMCalCalibParams from the ccdb. If false, use the default values given in the struct"}}};
```
mfasDa: I think the description is now swapped with respect to the program logics.

mfasDa (changes requested):
Only one small issue in the description has to be fixed.

Detectors/EMCAL/calibration/testWorkflow/emc-channel-calib-workflow.cxx
jokonig: Yes you are right. I corrected it. Thanks for catching that!

mfasDa (dismissed):
Looks good now

mfasDa:
Forwarding to @davidrohr and @shahor02 for final merging. The relevant CCDB object was uploaded on Aug. 12 to alice-ccdb.cern.ch (O2-3116, EMCAL-797).

shahor02 (dismissed):

shahor02:
@mfasDa fullCI fails with 
```
[3263:internal-dpl-ccdb-backend]: [18:49:43][ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch/EMC/Config/CalibParam/1550600800000/
```
Could you open an upload request for the object covering ranges in the end of the https://twiki.cern.ch/twiki/bin/viewauth/ALICE/O2DPGMCSamplingSchema ?

jokonig:
@shahor02 I think the error in the build is unrelated (timeout of o2sim_hepmc). If you agree this PR can be merged.
The previous issue which was holding this PR back was to upload the EMCalCalibParams for a specific TF range which was now done (O2-3116)

Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
```diff
@@ -227,6 +240,10 @@ DataProcessorSpec getEMCALChannelCalibDeviceSpec()
   std::vector<InputSpec> inputs;
   inputs.emplace_back(device::getCellBinding(), o2::header::gDataOriginEMC, "CELLS", 0, o2::framework::Lifetime::Timeframe);
   inputs.emplace_back(device::getCellTriggerRecordBinding(), o2::header::gDataOriginEMC, "CELLSTRGR", 0, o2::framework::Lifetime::Timeframe);
+  // for loading the channelCalibParams from the ccdb
+  if (loadCalibParamsFromCCDB) {
```
shahor02, line 244: it would be enough to call this only once, since params don't change during the run

shahor02 (approved):
merging, but please see comment below, can be changed with different PR, e.g. like
```
static bool firstTF = true;
if (firstTF) {
  firstTF = false;
  // put here all one-time queries.
  if (loadCalibParamsFromCCDB) {
     ....
  }
}
```

Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
jokonig: Okay that makes sense. I will add that!

PR #9647, opened 2022-08-17
TRD raw reader update

martenole:
This leaves the logic and functionality of the raw reader as it was before. It contains lots of cleanup, fixing of compiler warnings (unused variables, signed/unsigned comparisons) and adds more comments to the raw reader header to all its methods. The parsing of tracklets and digits data is left untouched as much as possible.
Still WIP, since I find 1.8% less tracklets with these changes and that should actually be constant, unless one check was not fully working before and is now dumping input data as it should. This I am checking now together with @bazinski

### DIFF
--- a/Detectors/TRD/reconstruction/src/CruRawReader.cxx
+++ b/Detectors/TRD/reconstruction/src/CruRawReader.cxx
@@ -39,94 +39,48 @@
 #include <array>
 #include <numeric>
 
+using namespace o2::trd::constants;
+
 namespace o2::trd
 {
 
-bool CruRawReader::skipRDH()
+void CruRawReader::configure(int tracklethcheader, int halfchamberwords, int halfchambermajor, std::bitset<16> options)
 {
-  // check rdh for being empty or only padding words.
-  if (o2::raw::RDHUtils::getMemorySize(mOpenRDH) == o2::raw::RDHUtils::getHeaderSize(mOpenRDH)) {
-    //empty rdh so we want to avoid parsing it for cru data.
-    if (mVerbose) {
-      LOG(info) << " skipping rdh (empty) packetcount of: " << std::hex << o2::raw::RDHUtils::getPacketCounter(mOpenRDH);
-    }
-    return true;
-  } else {
-
-    if (mHBFPayload[0] == o2::trd::constants::CRUPADDING32 && mHBFPayload[0] == o2::trd::constants::CRUPADDING32) {
-      //event only contains paddings words.
-      if (mVerbose) {
-        LOG(info) << " skipping rdh (padding) with packetcounter of: " << std::hex << o2::raw::RDHUtils::getPacketCounter(mOpenRDH);
-      }
-      // mDataPointer+= o2::raw::RDHUtils::getOffsetToNext()/4;
-      auto rdh = reinterpret_cast<const o2::header::RDHAny*>(mDataPointer);
-      mDataPointer += o2::raw::RDHUtils::getOffsetToNext(rdh) / 4;
-      //mDataPointer=reinterpret_cast<const uint32_t*>(reinterpret_cast<const char*>(rdh) + o2::raw::RDHUtils::getOffsetToNext(rdh));
-      return true;
-      return true;
-    } else {
-      return false;
-    }
-  }
+  mVerbose = options[TRDVerboseBit];
+  mHeaderVerbose = options[TRDHeaderVerboseBit];
+  mDataVerbose = options[TRDDataVerboseBit];
+  mFixDigitEndCorruption = options[TRDFixDigitCorruptionBit];
+  mTrackletHCHeaderState = tracklethcheader;
+  mHalfChamberWords = halfchamberwords;
+  mHalfChamberMajor = halfchambermajor;
+  mOptions = options;
+  mTimeBins = TIMEBINS; // set to value from constants incase the DigitHCHeader1 header is not present.
+  mPreviousDigitHCHeadersvnver = 0xffffffff;
+  mPreviousDigitHCHeadersvnrver = 0xffffffff;
 }
 
-void CruRawReader::OutputHalfCruRawData()
+void CruRawReader::incrementErrors(int hist, int sector, int side, int stack, int layer)
 {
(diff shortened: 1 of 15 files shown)


Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -623,27 +437,31 @@ int CruRawReader::processHalfCRU(int cruhbfstartoffset, int numberOfPreviousCRU,
...
-    mHBFoffset32 += mTotalHalfCRUDataLength32; // go to the end of this halfcruheader and payload.
+    mWordsRejected += (mTotalHBFPayLoad / 4) - mHBFoffset32;
     return -2;
   }
 
   //get eventrecord for event we are looking at
   mIR.bc = mCurrentHalfCRUHeader.BunchCrossing; // correct mIR to have the physics trigger bunchcrossing *NOT* the heartbeat trigger bunch crossing.
-  //shift accordingly
-  mIR.bc = mCurrentHalfCRUHeader.BunchCrossing - o2::ctp::TriggerOffsetsParam::Instance().LM_L0;
-  if (mIR.bc < 0) {
```
martenole, line 635: @bazinski I found the reason for the slightly different numbers: `mIR.bc` is unsigned, so the check you had hear never failed and we were not rejecting triggers which would be shifted to a negative bc. This is fixed with the changes in here

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
```diff
@@ -145,7 +145,7 @@ int TrackletsParser::Parse()
       // digit half chamber header ends with 01b and has the supermodule in position (9-13).
       // this of course can conflict with a tracklet hc header, hence should not be used!
       // NBNBNBNB
-      if ((tmpheader & 0x3) == 0x1 && (((tmpheader >> 9) & 0x1f) == mHCID / 30)) {
+      if ((tmpheader & 0x3) == 0x1 && (((tmpheader >> 9) & 0x1f) == mHCID / constants::NHCPERSEC)) {
```
f3sch, line 148: Could we not get rid of tmpheader and use the hcheader fields.
This would get rid of the bitshifting and a useless variable?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -198,214 +160,113 @@ bool CruRawReader::compareRDH(const o2::header::RDHAny* firstrdh, const o2::head
...
+    rdhPrevious = rdh;
+    firstRdh = false;
     auto headerSize = o2::raw::RDHUtils::getHeaderSize(rdh);
     auto memorySize = o2::raw::RDHUtils::getMemorySize(rdh);
-    if (memorySize == 0) {
-      LOG(warn) << "rdh memory size is zero";
-      return false; // get out of here if the rdh says it has nothing.
-    }
     auto offsetToNext = o2::raw::RDHUtils::getOffsetToNext(rdh);
     auto rdhpayload = memorySize - headerSize;
```
f3sch, line 190: Could memorySize < headerSize at any point? Then the memcpy gets problematic.

martenole: No, because the memory size contains the header size. Only of the RDH itself would be corrupted this could not be true, but then the above `checkRDH()` should fail and the whole data for that RDH is dumped

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
martenole: There are quite a few things I would like to change in the tracklet and digit parsers as well. But as I said above I wanted to leave them as they are for the moment so that this PR does not get too big. Sean also has some changes in the pipeline and after that I will look in more detail also through the parsing part

martenole:
The failure of the fullCI seems to be unrelated
```
[9961:internal-dpl-injected-dummy-sink]: [08:12:41][STATE] INITIALIZING DEVICE ---> INITIALIZED
[9961:internal-dpl-injected-dummy-sink]: [08:12:41][STATE] INITIALIZED ---> BINDING
[9961:internal-dpl-injected-dummy-sink]: [08:12:41][ERROR] Failed binding socket internal-dpl-injected-dummy-sink.metric-feedback[0].push, address: ipc://@metric-feedback-0, reason: Address already in use
[9961:internal-dpl-injected-dummy-sink]: [08:12:41][ERROR] failed to attach channel metric-feedback[0] (bind)
[9961:internal-dpl-injected-dummy-sink]: [08:12:41][ERROR] 1 of the binding channels could not initialize. Initial configuration incomplete.
```
Nevertheless, this PR should pass the check, so will squash the last three commits and force-push again

bazinski (approved):

Detectors/TRD/reconstruction/src/TrackletsParser.cxx
bazinski: badly named variable yes, useless no.

martenole:
I think I have seen this fullCI error quite often recently
```
[24026:BadMapCalibSpec]: [16:39:41][ERROR] Insufficient statistics: 6 entries in lowE histo, do nothing
```
Can this be denoted as warning, so that the CI does not fail because of this when it simulates only 5 events in one TF @mfasDa ?

martenole:
> I think I have seen this fullCI error quite often recently
> 
> ```
> [24026:BadMapCalibSpec]: [16:39:41][ERROR] Insufficient statistics: 6 entries in lowE histo, do nothing
> ```
> 
> Can this be denoted as warning, so that the CI does not fail because of this when it simulates only 5 events in one TF @mfasDa ?

Sorry @mfasDa , the error is actually coming from a PHOS workflow. I was mixing that with the `calib-emcalchannel-calibration`, which crashed. Unfortunately I can't see in the logs why. But maybe this was just a connection timeout issue. Sorry for the noise

martenole:
I ran the FST with these changes on top of `dev` locally without issues. Could this be merged or otherwise the FST restarted @TimoWilken ?

TimoWilken:
@martenole Checks are green now!

martenole:
Great, thanks a lot for pinging!

PR #9648, opened 2022-08-17
DPL Analysis: Event mixing: Mixing counters

saganatt:
Ciao @ktf @jgrosseo 

here come mixing variables needed by @jgrosseo for weighted correlations.

I return the same number for strictly upper / upper / full combinations, so I delegate the responsibility to the user to interpret it correctly. For example, in upper combinations the 1st event will be mixed with `currentWindowNeighbors() + 1` events, including itself. In strictly upper combinations, the number is simply `currentWindowNeighbors()` and that's the most common use case. Let me know if you prefer some differentiation here.

Analysis task example in the PR mentioned below.

### DIFF
--- a/Framework/Core/include/Framework/ASoAHelpers.h
+++ b/Framework/Core/include/Framework/ASoAHelpers.h
@@ -505,7 +505,7 @@ struct CombinationsBlockIndexPolicyBase : public CombinationsIndexPolicyBase<Ts.
     if (this->mIsEnd) {
       return 0;
     }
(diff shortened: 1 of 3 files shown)


jgrosseo:
Thanks! This looks good to me. I think your notice above "treat with care" is fine, please copy it also into the tutorial example and in the function `currentWindowNeighbours` as comment

jgrosseo (approved):

jgrosseo (approved):

saganatt:
Failed checks unrelated

PR #9649, opened 2022-08-17
Cleanup ASoA

ktf:
* Use _v suffix in place of ::value, as in C++14 and subsequent.
* Use std::array rather than std::vector when the number of columns is known.
* Avoid gymnastic between  columns and tables and back to columns.

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -1015,20 +1015,15 @@ constexpr bool is_binding_compatible_v()
   return are_bindings_compatible_v<T>(originals_pack_t<B>{});
 }
 
+//! Helper to check if a type T is an iterator
 template <typename T>
-using is_soa_iterator_t = typename framework::is_base_of_template<RowViewCore, T>;
-
-template <typename T>
-constexpr bool is_soa_iterator_v()
-{
-  return is_soa_iterator_t<T>::value || framework::is_specialization_v<T, RowViewCore>;
-}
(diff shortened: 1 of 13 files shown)


Framework/Core/include/Framework/ASoA.h
```diff
@@ -1015,20 +1015,15 @@ constexpr bool is_binding_compatible_v()
...
+//! Helper to check if a type T is an iterator
 template <typename T>
-using is_soa_iterator_t = typename framework::is_base_of_template<RowViewCore, T>;
-
-template <typename T>
-constexpr bool is_soa_iterator_v()
-{
-  return is_soa_iterator_t<T>::value || framework::is_specialization_v<T, RowViewCore>;
-}
+inline constexpr bool is_soa_iterator_v = framework::is_base_of_template_v<RowViewCore, T>;
```
aalkin: I doubt that this is equivalent to the previous code.

ktf: I have my doubts about it as well, but why would:

```c++
framework::is_specialization_v<T, RowViewCore>;
```

and

```C++
framework::is_base_of_template_v<RowViewCore, T>
```

differ?

ktf: ok, I have aligned the two versions. As discussed, I am not sure it actually matters, since hyperloop seems fine with it.

ktf:
Errors unrelated. Merging.

ktf:
Tested on hyperloop as well.

PR #9651, opened 2022-08-17
Fix for PVertexer stall + debug reco method

shahor02:

PR #9654, opened 2022-08-18
[EMCAL-565] Use proper calibrator in fillTFIDInfo

mfasDa:
Code was using hard-coded mTimeCalibrator in
fillTFIDInfo, which doesn't exist for the bad channel
calibration. Instead the bad channel calibrator must be
used here. Handling like the rest of the code is exclusive.


mfasDa:
@shahor02 This should fix the CI errors reported in #9651

PR #9659, opened 2022-08-19
[Event Display] Adds primary vertex selection based on its position

maoderos:
This PR adds x,y,z range parameters for selecting primary vertices in primary vertex mode based on its position. Jira ticket ED-90.


davidrohr (approved):
approving such that the CI runs

PR #9660, opened 2022-08-19
ShmManager: error & exit if memory is removed unexpectedly

rbx:
Add a option (on by default) to check periodically if all shmem segments/regions are still present, and error & exit if they are not.
To catch and report cases where a misconfigured device or anything else would remove the memory.

For https://alice.its.cern.ch/jira/browse/EPN-106.

### DIFF
--- a/Utilities/ShmManager/src/ShmManager.cxx
+++ b/Utilities/ShmManager/src/ShmManager.cxx
@@ -161,6 +161,21 @@ struct ShmManager {
     }
   }
 
+  bool CheckPresence()
+  {
+    for (const auto& sc : segmentCfgs) {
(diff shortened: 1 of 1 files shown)


rbx:
This requires FairMQ v1.4.52

davidrohr:
we are already at fmq 1.4.52. Why did you mark it as draft?

rbx:
Only want to make sure it compiles before putting it up for review.
I've only tested this in a FairMQ environment and then copied the necessary changes without compiling O2.

davidrohr:
ok, I see. Note that our CI will not run on draft PRs.

rbx:
I'll undraft it then.

ktf (approved):

PR #9665, opened 2022-08-19
Separating executable for Run 3 and Run 5 (the different modules will not mix)

zovarga:
Separating sim executable for Run3 / Run5.

- `o2-sim` and `o2-sim-serial` work only with Run 3 modules, while 
- `o2-sim-run5` and `o2-sim-serial-run5` work only with Run 5 modules.

We added the logic in **SimConfig.cxx** which checks whether the correct modules are used with the `-m` flag and the code returns with FATAL ERROR if the wrong modules are used. If no module specified, the executable builds all the Run3 or Run5 modules, depending on which executable is used.


mconcas:
Hi @zovarga @sawenzel, thanks a lot for taking care of this.

I have a general question: could not be possible to have a more generic separation between run3 (production) and all other upgrades?

For instance: IT3 => ITS 3, it's not a run 5 detector, it will eventually require almost everything from run 3 but ITS, at least standing thins as they are now.
In later upgrades indeed we might need to redefine even some of the hardcoded volumes (e.g. pipe, magnet, shields, etc.) which I see now are included only if this is not run5. But I guess something like cave and hall will more or less stay the same (please correct me if I am wrong).

sawenzel:
@mconcas : Yes, I agree we'll need more configurability in any case. The present PR is a first step/investigation that was motivated from the goal to not load
upgrade geometry libraries in the sim, when they are not needed (and vice versa). And separate run3 and run5 executable would wrap some of the configurations consistently so that the user would not need to pass them all the time to a generic executable.  Finally, a proper plugin mechanism with some json-config files, describing a possible consistent configuration, might need to be targeted.

In any case, at the moment the PR does not compile for all CI configs. We need to rethink a bit where this is going.

zovarga:
Hello Matteo & Sandro.
The current modifications have been implemented with the assumption, that `ENABLE_UPGRADES` is always true.
One of the tests fails because it runs the code with the above variable being false. I have to revise the logic to be consistent with that case, too. 
We should get rid of `ENABLE_UPGRADES` eventually anyway, it was only left in the code because it occurs at too many places, and we concluded that it might be better to leave it there for now and advance in small steps.
Otherwise the code works as intended.

sawenzel (approved):

PR #9672, opened 2022-08-22
[WIP] TRD raw reader tracklet parsing

martenole:
Not to be merged yet, will rebase once Sean's next PR is through.

- Simplify tracklet parsing and move it into CruRawReader
- Update the HCID from the TrackletHCHeader in case there is a mismatch with the RDH

TODO: For the wrongly connected links there is still a mismatch between the HCID assigned to the digits and the HCID assigned to the tracklets. This I will clean up after the rebase when I'll also go through the digit parsing and see if that can be simplified as well.


martenole:
Superseded by #9691

PR #9675, opened 2022-08-22
TRD reformulate debug output

bazinski:
change the logging, to prevent erroneously large logs on the epn.
make them more readiable and usable.
adds some options : 

1. --verbose 
2. --verbosehalfcru   dump the half cru on receiving.
3. --verboselink dump the link's data on error
4. --verboseword dump each word of data as its recognise and unpack it for what it is thought to be.
5. --verboseerrors make sure error messages appear for those that ordinarily simply go to histograms.
6. --onlycalibrationtriggers filter out non calibration triggers, making debugging easier when you want digits an tracklets together.
a later update will limit the number of errors on a per error basis.


Detectors/TRD/reconstruction/include/TRDReconstruction/CruCompressorTask.h
```diff
@@ -40,9 +40,6 @@ class CruCompressorTask : public Task
  private:
   CruRawReader mReader; // this will do the parsing.
   std::array<uint64_t, o2::trd::constants::HBFBUFFERMAX> mOutBuffer;
-  bool mVerbose{false};
-  bool mDataVerbose{false};
-  bool mHeaderVerbose{false};
```
martenole, line 45: Do we need to keep the `CruCompressorTask` at all or should we remove it completely?

Detectors/TRD/reconstruction/include/TRDReconstruction/DataReaderTask.h
```diff
@@ -45,11 +44,10 @@ class DataReaderTask : public Task
...
   void updateTimeDependentParams(framework::ProcessingContext& pc);
-  CruRawReader mReader;                  // this will do the parsing, of raw data passed directly through the flp(no compression)
-  CompressedRawReader mCompressedReader; //this will handle the incoming compressed data from the flp
-                                         // in both cases we pull the data from the vectors build message and pass on.
-                                         // they will internally produce a vector of digits and a vector tracklets and associated indexing.
-                                         // TODO templatise this and 2 versions of datareadertask, instantiated with the relevant parser.
+  CruRawReader mReader; // this will do the parsing, of raw data passed directly through the flp(no compression)
+                        // we pull the data from the vectors build message and pass on.
+                        // they will internally produce a vector of digits and a vector tracklets and associated indexing.
+                        // TODO templatise this and 2 versions of datareadertask, instantiated with the relevant parser.
```
martenole: I think the TODO can be removed here ;)

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
```diff
@@ -120,6 +120,11 @@ struct HelperMethods {
     return ori;
   }
 
+  static int getLinkFromSectorSideEndPointLink(uint32_t sector, uint32_t side, uint32_t endpoint, int link)
+  {
+    return (sector * 2 + side) * 30 + (15 * endpoint) + link;
```
martenole: I don't like this method, as it is very hard to read with the magic numbers and it's not really clear what is returned. I added a comment where it is used below with suggested change

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -59,27 +56,95 @@ void CruRawReader::configure(int tracklethcheader, int halfchamberwords, int hal
...
+    outputstring << z << ":" << mHBFPayload[bufferoffset + z] << " ";
+  }
+  LOG(info) << "CRH1 " << outputstring.str();
+  int outputoffset = 2;
+  bufferoffset += 8;
+  int olengthoffset0 = 0;
+  for (int link = 0; link < 15; ++link) {
+    uint32_t linkid = 0, hcid = 0;
+    uint32_t cruindex = mFEEID.supermodule * 2 + mFEEID.side;
+    uint32_t ori = HelperMethods::getLinkFromSectorSideEndPointLink(mFEEID.supermodule, mFEEID.side, mFEEID.endpoint, link);
```
martenole: you already have the cruindex (which is currently not used afterwards btw), can you not do the same as I do in `processHalfCRU()`?
```
    int cruIdx = mFEEID.supermodule * 2 + mFEEID.side;                    // 2 CRUs per SM, side defining A/C-side CRU
    int halfCruIdx = cruIdx * 2 + mFEEID.endpoint;                        // endpoint (0 or 1) defines half-CRU
    int linkIdxGlobal = halfCruIdx * NLINKSPERHALFCRU + currentlinkindex; // global link ID [0..1079]
    int halfChamberId = HelperMethods::getHCIDFromLinkID(linkIdxGlobal);
```

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -59,27 +56,95 @@ void CruRawReader::configure(int tracklethcheader, int halfchamberwords, int hal
...
+    }
+    bufferoffset += 8;
+    LOG(info) << wordcount++ << ":" << outputstring.str();
   }
-  mEventRecords.incParsingError(hist, sector, side, stack * NLAYER + layer);
-  if (mVerbose) {
-    LOG(info) << "Parsing error: " << hist << " sector:" << sector << " side:" << side << " stack:" << stack << " layer:" << layer;
+}
+
+void CruRawReader::OutputHalfCruRawData(uint32_t offset)
```
martenole: Actually, not sure if we need these two methods at all or if not simply `CruRawReader::dumpInputPayload()` would suffice?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -122,47 +193,60 @@ bool CruRawReader::compareRDH(const o2::header::RDHAny* rdhPrev, const o2::heade
...
+    incrementErrors(TRDParsingBadRDHFEEID);
     return false;
   }
   if (o2::raw::RDHUtils::getEndPointID(rdhPrev) != o2::raw::RDHUtils::getEndPointID(rdhCurr)) {
     if (mMaxWarnPrinted > 0) {
-      LOG(warn) << "ERDH  EndPointID are not identical in rdh.";
       checkNoWarn();
     }
-    incrementErrors(TRDParsingBadRDHEndPoint, 0, 0, 0, 0);
+    if (mOptions[TRDVerboseErrorsBit]) {
```
martenole, line 146: Instead of adding everywhere the `mOptions[TRDVerboseErrorsBit]` plus a duplicate of the log message, could you not just change the `checkNoWarn()` and `checkNoErr()` methods to **not** decrement the mMaxErrsPrinted if the verbose bit is set? Would make the code much cleaner

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -195,23 +279,29 @@ int CruRawReader::processHBFs()
 
     // copy the contents of the current RDH into the buffer to be parsed, RDH payload is memory size minus header size
     std::memcpy((char*)&mHBFPayload[0] + mTotalHBFPayLoad, ((char*)rdh) + headerSize, rdhpayload);
+    int packetCount = o2::raw::RDHUtils::getPacketCounter(rdh);
```
martenole: is this variable used for anything?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -260,19 +353,30 @@ int CruRawReader::parseDigitHCHeader(int hcid)
   if (mDigitHCHeader.major == 0 && mDigitHCHeader.minor == 0 && mDigitHCHeader.numberHCW == 0) {
     //hack this data into something resembling usable.
     mDigitHCHeader.major = mHalfChamberMajor;
-    mDigitHCHeader.minor = 42; // to keep me entertained
+    mDigitHCHeader.minor = 42;
     mDigitHCHeader.numberHCW = mHalfChamberWords;
+    if (mHalfChamberWords == 0 || mHalfChamberMajor == 0) {
+      if (mMaxWarnPrinted > 0) {
+        LOG(warn) << "halfchamber header is corrupted and you have only set the halfchamber command line option to zero, hex dump of data and revisit what it should be.";
```
martenole, line 286: where is the hex dump?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -426,37 +562,57 @@ int CruRawReader::processHalfCRU(int iteration)
...
     incrementErrors(TRDParsingHalfCRUCorrupt);
     mWordsRejected += (mTotalHBFPayLoad / 4) - mHBFoffset32;
     return -2;
   }
 
   //get eventrecord for event we are looking at
-  mIR.bc = mCurrentHalfCRUHeader.BunchCrossing; // correct mIR to have the physics trigger bunchcrossing *NOT* the heartbeat trigger bunch crossing.
+  mIR.bc = mCurrentHalfCRUHeader.BunchCrossing; // correct mIR to have the physics trigger
+                                                // bunchcrossing *NOT* the heartbeat trigger
+                                                // bunch crossing.
```
martenole, line 497: Did you enforce now the maximum line length in the editor? I am not sure if it makes it so much more readable

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -523,29 +678,40 @@ int CruRawReader::processHalfCRU(int iteration)
...
-      if (mHeaderVerbose) {
+      if (mOptions[TRDVerboseErrorsBit]) {
         LOGF(info, "Tracklet parser starting at offset %u and processing up to %u words", mHBFoffset32, currentlinksize32);
       }
-
       // for now we are using 0 i.e. from rdh FIXME figure out which is authoritative between rdh and ori tracklethcheader if we have it enabled.
       int trackletWordsRead = mTrackletsParser.Parse(&mHBFPayload, linkstart, linkend, mFEEID, side, detectorId, stack, layer, mCurrentEvent, &mEventRecords, mOptions, mTrackletHCHeaderState); // this will read up to the tracklet end marker.
+      if (mTrackletsParser.dumpLink()) {
+        //dump the link that cause the error
+        // the call to dumpLink resets the boolean to false;
```
martenole, line 591: but with every call to `incParsingError()` the boolean is set to true again. That way you would dump the data all the time and not only once upon error per run. Or was it intended like that?

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -60,12 +60,16 @@ std::ostream& operator<<(std::ostream& stream, const TrackletMCMHeader& mcmhead)
 
 void dumpHalfCRUHeader(o2::trd::HalfCRUHeader& halfcru)
```
martenole, line 61: Why not just remove this method and use `printHalfCRUHeader()` instead?

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -322,6 +324,10 @@ void printHalfCRUHeader(o2::trd::HalfCRUHeader& halfcru)
   LOG(info) << "Raw: " << std::hex << halfcru.word0 << " " << halfcru.word12[0] << " " << halfcru.word12[1] << " " << halfcru.word3 << " " << halfcru.word47[0] << " " << halfcru.word47[1] << " " << halfcru.word47[2] << " " << halfcru.word47[3];
 }
 
+void printDigitADCMask(o2::trd::DigitMCMADCMask& mask)
+{
+}
```
martenole: Remove this empty method?

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -582,6 +593,21 @@ void printDigitHCHeader(o2::trd::DigitHCHeader& header, uint32_t headers[3])
   }
 }
 
+int getNumberOfTrackletsFromHeader(o2::trd::TrackletMCMHeader* header, bool verbose)
```
martenole: Don't we already have this method? Duplicate?

Detectors/TRD/reconstruction/include/TRDReconstruction/CruRawReader.h
```diff
@@ -115,18 +109,24 @@ class CruRawReader
...
   // - probably enough to only pass the half-chamber ID, if we know it
   // - or what about being more granular? dump ROB or MCM, if we know it?
-  void incrementErrors(int hist, int sector = -1, int side = 0, int stack = 0, int layer = 0);
+  void incrementErrors(int error, int hcid = -1);
 
   // helper function to dump the whole input payload including RDH headers
   void dumpInputPayload() const;
 
+  // dump out and entire half cru buffer.
+  void OutputHalfCruRawData(uint32_t offset);
```
martenole: A tiny thing, but still... Class methods in O2 should start with lower case letter...

Detectors/TRD/reconstruction/include/TRDReconstruction/EventRecord.h
```diff
@@ -152,16 +152,23 @@ class EventStorage
...
-  void incParsingError(int error, int sm, int side, int stacklayer)
+  void incParsingError(int error, int hcid)
   {
+    bool detailGraphs = true;
+    if (hcid < 0)
+      detailGraphs = false;
+    unsigned int sm = HelperMethods::getSector(hcid / 2);
+    unsigned int stack = HelperMethods::getStack(hcid / 2);
+    unsigned int layer = HelperMethods::getLayer(hcid / 2);
+    unsigned int side = hcid % 2;
```
martenole: probably all these calculations should only be done if `hcid>=0`

martenole (commented):
Did not go through everything yet, especially not the parsers and I also did not try to run it yet. Still I have some comments already now

Detectors/TRD/reconstruction/include/TRDReconstruction/CruCompressorTask.h
bazinski: huh i thought i did remove it.

Detectors/TRD/reconstruction/include/TRDReconstruction/DataReaderTask.h
bazinski: oui

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: sure

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: oops

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: done

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: removed we already have printDigitMCMADCMask, not sure what that was left in.

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: odd, and gone.

Detectors/TRD/reconstruction/include/TRDReconstruction/CruRawReader.h
bazinski: fixed

Detectors/TRD/reconstruction/include/TRDReconstruction/EventRecord.h
bazinski: actually dont need those, gone.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: true

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: used to be, but check is somewhere else for consecutive packetcount.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: certainly not there.

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
bazinski: not needed anymore.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: removed.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -122,47 +140,49 @@ bool CruRawReader::compareRDH(const o2::header::RDHAny* rdhPrev, const o2::heade
       LOG(warn) << "ERDH FEEID are not identical in rdh.";
       checkNoWarn();
     }
-    incrementErrors(TRDParsingBadRDHFEEID, 0, 0, 0, 0);
+    incrementErrors(TRDParsingBadRDHFEEID);
     return false;
   }
   if (o2::raw::RDHUtils::getEndPointID(rdhPrev) != o2::raw::RDHUtils::getEndPointID(rdhCurr)) {
-    if (mMaxWarnPrinted > 0) {
+    if (mMaxErrsPrinted > 0) {
```
martenole: Should be `mMaxWarnPrinted`

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -93,23 +108,26 @@ bool CruRawReader::checkRDH(const o2::header::RDHAny* rdh)
       LOG(error) << "RDH check failed due to 0xfff. FLP not configured, call TRD on call. Whole feeid = " << std::hex << (unsigned int)feeid.word;
       checkNoErr();
     }
-    incrementErrors(TRDFEEIDIsFFFF, 0, 0, 0, 0);
+    if (mOptions[TRDVerboseErrorsBit]) { // ignore the count limit
```
martenole: not needed to ignore the count limit here

Detectors/TRD/reconstruction/src/CruRawReader.cxx
martenole: then I would just remove this unused variable...

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -220,8 +243,8 @@ int CruRawReader::processHBFs()
   int iteration = 0;
   mHBFoffset32 = 0;
   while ((mHBFoffset32 < ((mTotalHBFPayLoad) / 4))) {
-    if (mHeaderVerbose) {
-      LOGP(info, "Current half-CRU iteration {}, current offset in the HBF payload {}, total payload in number of 32-bit words {}", iteration, mHBFoffset32, mTotalHBFPayLoad / 4);
+    if (mOptions[TRDVerboseBit]) {
+      LOG(info) << "Looping over cruheaders in HBF, loop count " << iteration << " current offset is" << mHBFoffset32 << " total payload is " << mTotalHBFPayLoad / 4 << "  raw :" << mTotalHBFPayLoad;
```
martenole: can you please use the original `LOGP` message?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
```diff
@@ -242,7 +265,10 @@ void CruRawReader::checkDigitHCHeader(int halfChamberIdRef)
 
   if (!mOptions[TRDIgnoreDigitHCHeaderBit]) {
     if (halfChamberIdHeader != halfChamberIdRef) {
-      incrementErrors(TRDParsingDigitHCHeaderMismatch, mFEEID.supermodule, halfChamberIdRef % 2, HelperMethods::getStack(halfChamberIdRef / 2), HelperMethods::getLayer(halfChamberIdRef / 2));
+      incrementErrors(TRDParsingDigitHCHeaderMismatch, halfChamberIdRef);
+      if (mOptions[TRDVerboseErrorsBit]) {
+        LOGF(info, "Sector mismatch in Digit HCHeader : currentsector : %d halfChamberIdHeader: %d halfChamberId : %d", mFEEID.supermodule, halfChamberIdHeader, halfChamberIdRef);
```
martenole: Its a HCID mismatch and I would just dump both HCIDs

martenole (commented):
Some more smaller points and one general thing:
Why do we not change the method `void incrementErrors(int error, int hcid = -1);` to take in addition a string `void incrementErrors(int error, int hcid = -1, std::string message="");` and then we use `fmt::format` to generate the message in place upon the call to the method. In the method we can decide if we want to print the message based on the verbosity settings or not. That way we could separate nicely the processing code from the logging and avoid all `if (mOptions[TRDVerboseErrorsBit])` blocks.
Also in RawDataStats I would be in favor of using an `unordered_map` to have a simple conversion of error number to error string which is not sensitive to having the same order in the `enum` and the `std::vector<string>`.
If you want I can also see if I manage to do that today on top of your changes here. What do you think?

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: no hex dump, you are going to have to take a hex dump of a timeframe and work through it to figure out what the data is.

Detectors/TRD/reconstruction/src/CruRawReader.cxx
bazinski: Intended like that. In the initial comments for this pr, I will implement a counter per error, but want this through now. Can fix that later.

martenole (approved):
Merging as is now, more changes will follow

PR #9678, opened 2022-08-22
[EMCAL-794] Add event counter and end of stream to data producer

jokonig:
- As we wan to terminate the process after a certain amount of events is
reached (typically the number of events needed to trigger the
calibration), an event counter is introduced
- number of events added as workflow argument


jokonig:
@shahor02 the error in build/O2/o2 seems unrelated. I think this can be merged.
It says that its waiting on code owner review by @mfasDa but I should also be code owner here now.

shahor02 (approved):
According to CODEOWNERS file, you are, but I guess, codeowners are not supposed to review themselves...

PR #9680, opened 2022-08-23
DPL GUI: improve GUI using special glyphs

ktf:
Now that both macOS and linux behave correctly, we can try to have
nicer GUI using glyphs.


ktf:
Error unrelated. Merging.

PR #9681, opened 2022-08-23
bug fix due to wrong truncation in TOF digitizer/encoder

noferini:
@sawenzel this is the fix you need for large value of  orbitFirstSampled


sawenzel (approved):

sawenzel:
checked compile locally. Fixing an urgent problem.

PR #9683, opened 2022-08-23
Ctpdev: ctp aod readout

lietava:


lietava:
HI @shahor02 , I am debugging 'ctpreadout' creation in AOD.
To generate data and run workflow I use macro derived from:
https://github.com/AliceO2Group/O2DPG/blob/master/MC/run/examples/O2DPG_pp_minbias_one_tf_qc.sh.
However it seems that trd trigger info:
https://github.com/lietava/AliceO2/blob/40293d7932b02075876abbf180f4f42404580cef/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L1181
is not available at:
https://github.com/lietava/AliceO2/blob/40293d7932b02075876abbf180f4f42404580cef/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L115
How can I get it ? Or should I use different container ?

shahor02:
@lietava You can add TRD to https://github.com/AliceO2Group/AliceO2/blob/6cf03ec14f1b080df1351c7e21223444232af8be/Detectors/AOD/src/aod-producer-workflow.cxx#L55

lietava:
Hi @shahor02 , I added TRD to the line you suggested. I also added TRD to the line 587 (o2-aod-producer-workflow) in 
[workflowrl.json.txt](https://github.com/AliceO2Group/AliceO2/files/9414568/workflowrl.json.txt).
I still get:
[626342:aod-producer-workflow]: I am here 01
[626342:aod-producer-workflow]: [10:41:11][ERROR] No TRD triggers
[626342:aod-producer-workflow]: I am here 1
R.

shahor02:
This means that the workflow does not require TRD as an input source. I think you should add to 
https://github.com/AliceO2Group/O2DPG/blob/7def6eb42c8f883d3823f3adde22b019927b1437/MC/bin/o2dpg_sim_workflow.py#L1082

```
aodinfosources += ',TRD'
```

lietava:
@shahor02 @mpuccio : so in summary I added TRD (see above):
1.) aod-producer-worlflow.cxx
2.) workflowrl.json 
3.) o2dpg_sim_workflow.py
Still 
[636075:aod-producer-workflow]: [10:40:15][ERROR] No TRD triggers
What else ?
R.

shahor02:
@lietava you did not need to touch the json file, it is anyway generated on the fly. Are you sure you have installed the O2DPG after modifying the `o2dpg_sim_workflow.py`? If so, could you put the tarball with your simulation on the cernbox?

lietava:
@shahor02 : I just modified o2dpg_sim_workflow.py in my /home/rl/O2DPG/MC/bin.
How I reinstalled O2DPG ? Recompile O2Physics ? R.

shahor02:
If you are modifying master branch of O2DPG, it should be enough to run `alibuild build O2DPG`, otherwise, you should build whole O2PDPSuite. But, for the fast check, you can simply modify the `$O2DPG_ROOT/MC/bin/o2dpg_sim_workflow.py` (will be overwritten at the rebuild)

lietava:
"But, for the fast check, you can simply modify the $O2DPG_ROOT/MC/bin/o2dpg_sim_workflow.py (will be overwritten at the rebuild)" -- but this is what I just did ? R.

shahor02:
Isn't ` /home/rl/O2DPG/MC/bin` your source directory? What is you $O2DPG_ROOT ?

lietava:
export O2DPG_ROOT=/home/rl/O2DPG

shahor02:
$O2DPG_ROOT as well as other base dirs. are set by the alienv. Anyway, could you put one of your simulated TFs on the cernbox?

lietava:
@shahor02 please, see email

shahor02:
OK, the TRD info is read but the AODProducer does not use it since it should be requested as a "cluster" rather than "track" type.
Please add:
```
if (src[GID::TRD]) {
   dataRequest->requestTRDTracklets(false);
}
```
at 
https://github.com/AliceO2Group/AliceO2/blob/ae867a6afd003a4fd9bc48794c38f6816d0d8fe8/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L2111

lietava:
@shahor02 @mpuccio : it seems it works now. Thanks. We can try on real data. But I think there is still at least one problem: access to CCDB does not use run number.

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
```diff
@@ -1180,7 +1180,12 @@ gsl::span<const o2::trd::CalibratedTracklet> RecoContainer::getTRDCalibratedTrac
 
 gsl::span<const o2::trd::TriggerRecord> RecoContainer::getTRDTriggerRecords() const
 {
-  return inputsTRD->mTriggerRecords;
+  if (inputsTRD == nullptr) {
+    LOG(ERROR) << "No TRD triggers";
```
shahor02: if the TRD is not there, this is not an error, you can probably log a warning once, no need to do this at every TF.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1662,6 +1667,7 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   // helper map for fast search of a corresponding class mask for a bc
   std::unordered_map<uint64_t, uint64_t> bcToClassMask;
   if (mInputSources[GID::CTP]) {
+    LOG(info) << "CTP input available";
```
shahor02: These logs at every TF is also too much, printing at 1st TF only should be enough.

shahor02 (commented):
Could you clarify where `access to CCDB does not use run number` problem is?

lietava:
> Could you clarify where `access to CCDB does not use run number` problem is?

it is line https://github.com/lietava/AliceO2/blob/db4affa28cbf0a6087f4598c2e86d27b59d81b9a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L2094
where for MC I do not need specify run as validity is fir all MCs.
For real data It needs to be changed to something like:
ccdbParamSpec("CTP/Config/Config", bool runDependent =1, std::vector<CCDBMetadata> metadata = {{"runNumber","12345"}}, int64_t rate=0);
?
R.

shahor02:
> it is line https://github.com/lietava/AliceO2/blob/db4affa28cbf0a6087f4598c2e86d27b59d81b9a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L2094 where for MC I do not need specify run as validity is fir all MCs. For real data It needs to be changed to something like: ccdbParamSpec("CTP/Config/Config", bool runDependent =1, std::vector metadata = {{"runNumber","12345"}}, int64_t rate=0); ? R.

You just need to change it to 
```
dataRequest->inputs.emplace_back("ctpconfig", "CTP", "CTPCONFIG", 0, Lifetime::Condition, ccdbParamSpec("CTP/Config/Config", true));
```
But this will make it run dependent both for data and MC. Using `useMC` flag would not be reliable as we may run the AODproducer on MC run w/o this flag. So, we would either need an extra option telling that this is an MC data or to have per-run CTP objects also for CTP, as we have for e.g. GRPECS.

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
lietava: so I added warning counter to RecoContainr. h. OK ?

shahor02: yes, but it seems you did not push it.

lietava: yes, I will, thanks.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
lietava: info changed to debug

DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
lietava: I pushed. Hmm there is problem - I try to increase counter in const function:
https://github.com/lietava/AliceO2/blob/a68ed6692d911e8da791b4f4751510500af938f2/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L1186
Shall I give up this message ?
R.

shahor02: you can use local static counter.

lietava: done

lietava:
@shahor02 : for CTP config I created parameter:
https://github.com/lietava/AliceO2/blob/0843fc6bb0f7aef7ae0ea79e12b2207b81a7e8a7/Detectors/AOD/src/aod-producer-workflow.cxx#L42

shahor02:
note that your PR is still in draft mode some the CI is not running

lietava:
@shahor02 : can we merge ?

shahor02 (approved):

PR #9684, opened 2022-08-23
[EMCAL-687] (WIP) Added cell calibration to Event handler

jokonig:
- Time, Bad channel, Energy and temperature calibration have to be
applied going from CTF -> AO2D
- The calibration objects are stored in the ccdb and are now loaded into
the dedicated component: EMCALCalibrationHandler which can be steered
via the EMCALCalibCCDBHelper where one can set which calibrations should
be applied.
- In the EventHandler, before the calibrations are stored in the
EventData object, the EMCALCalibrationHandler decides if the cell is
good or bad (the cell is only stored  if it is good), and if it is a
good cell, applies the calibrations.
- In order to only accept good cells, the cell list in the EventData had
to be changed to a vector instead of a span. The span does not own the
objects in the span and we could therefore only look at the existing
cell array of all cells in the event. To avoid filtering out cells
additionally in the AODProducer, a vector was used to store the
calibrated cells in the eventData.
- The EMCALCalibCCDBHelper is instanciated in the AODProducer directly
and can be accessed from all tasks afterwards
- The EventHandler was moved from DataFormats to reconstruction in order
to avoid circular dependencies

### DIFF
--- a/DataFormats/Detectors/EMCAL/CMakeLists.txt
+++ b/DataFormats/Detectors/EMCAL/CMakeLists.txt
@@ -10,20 +10,20 @@
 # or submit itself to any jurisdiction.
 
 o2_add_library(DataFormatsEMCAL
-               SOURCES src/EMCALBlockHeader.cxx 
+               SOURCES src/EMCALBlockHeader.cxx
                        src/TriggerRecord.cxx
                        src/Constants.cxx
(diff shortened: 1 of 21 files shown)


mfasDa:
Superseeded by #9755, can be closed.

PR #9685, opened 2022-08-24
[FIT] Enforce use of getters in fit::Triggers, FV0 and FT0 reco align…

mslupeck:
…ment, FV0 reco cuts revision


mslupeck:
Hi @shahor02 
Could you merge it? It's just a cleanup with the only significant change in cuts for collision time reconstruction for FV0

shahor02 (approved):

PR #9686, opened 2022-08-24
[FIT] Change of CODEOWNERS

mslupeck:
This is to reflect the changes in FIT software responsibles:
- FT0: Artur replaces Alla (Alla has already left)
- FV0: Andreas replaces myself (for the transition period, until December, let's keep both accounts)
- FDD: Arvind takes over Michal (for some reason Michal was never a codeowner)


ktf:
Merging this. Notice that @afurs @andreasmolander @arvindkhuntia will have to accept my invitation to be writers of the repository before they will be allowed to leave reviews of the code they own.

PR #9688, opened 2022-08-24
Cleanup table builder

ktf:


ktf:
@aalkin could you have a second look at this? All the unit tests seem to pass and compilation time seems to be greatly improved.

Framework/Core/include/Framework/TableBuilder.h
```diff
@@ -664,8 +671,8 @@ class TableBuilder
 
   void validate() const;
 
-  template <typename... ARGS>
-  auto makeBuilders(std::array<char const*, sizeof...(ARGS)> const& columnNames, size_t nRows)
+  template <typename... ARGS, size_t I = sizeof...(ARGS)>
```
aalkin, line 674: Does this improve compilation time somehow? Is it not equivalent to `sizeof...` in the array declaration?

aalkin:
Other than the question above this seems very reasonable to me.

Framework/Core/include/Framework/TableBuilder.h
ktf: What I see is that the sizeof...(ARGS) is left unexpanded (e.g. `sizeof...(int, float, ...)`), at least in some error messages, duplicating the already long list of template parameters. By having `size_t I` as a template parameter, it simply matches on the size of the array, when possible, resulting in a much smaller template name.

ktf:
Checked on hyperloop as well. Merging.

PR #9689, opened 2022-08-24
TPC QC: Workaround for memory corruption in CalDet filling

bulukutlu:
Fix on TPC Cluster QC task ensuring that hits with out-of-range pad positions don't lead to run failure observed in async productions.

It is observed that clusters can have an assigned padNumber which exceeds the number of pads in IROCs (5280) and OROCs (9280) accordingly. Ignoring such clusters mitigates the seg faults observed in async qc.

### DIFF
--- a/Detectors/TPC/qc/src/Clusters.cxx
+++ b/Detectors/TPC/qc/src/Clusters.cxx
@@ -11,12 +11,12 @@
 
 #include <string>
 
-//root includes
+// root includes
 #include "TH1.h"
 #include "TH2.h"
 #include "TFile.h"
 
-//o2 includes
+// o2 includes
 #include "TPCQC/Clusters.h"
 #include "TPCBase/Painter.h"
 #include "TPCBase/ROC.h"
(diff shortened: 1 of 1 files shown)


Detectors/TPC/qc/src/Clusters.cxx
```diff
@@ -38,11 +38,27 @@ bool Clusters::processCluster(const T& cluster, const o2::tpc::Sector sector, co
...
+
   const int nROC = row < 63 ? int(sector) : int(sector) + 36;
   const int rocRow = row < 63 ? row : row - 63;
 
   const float pad = cluster.getPad();
 
+  size_t position = mapper.getPadNumber(mNClusters.getCalArray(nROC).getPadSubset(), mNClusters.getCalArray(nROC).getPadSubsetNumber(), rocRow, pad);
+
+  if ((nROC < 36)) { // IROC
+    if (position > 5279) {
```
chiarazampolli: Do we really to hardcode this numbers? Are they not some constants of some TPC class? Same for the lines below.

bulukutlu: they are indeed already defined in Detectors/TPC/base/include/TPCBase/Mapper.h.
I will update that they are taken from there

Detectors/TPC/qc/src/Clusters.cxx
```diff
@@ -38,11 +38,27 @@ bool Clusters::processCluster(const T& cluster, const o2::tpc::Sector sector, co
...
   const float pad = cluster.getPad();
 
+  size_t position = mapper.getPadNumber(mNClusters.getCalArray(nROC).getPadSubset(), mNClusters.getCalArray(nROC).getPadSubsetNumber(), rocRow, pad);
+
+  if ((nROC < mapper.getNumberOfIROCs())) { // IROC
+    if (position >= mapper.getPadsInIROC()) {
+      LOG(error) << "IROC out of range: pad position " << position << "\tROC " << nROC << "\tIROC has " << mapper.getPadsInIROC() << " pads";
+      return false;
+    }
+  } else if ((nROC >= mapper.getNumberOfIROCs())) { // OROC
```
chiarazampolli: The number is the same, but should this not be `getNumberOfOROCs`? See https://github.com/AliceO2Group/AliceO2/blob/3e46ea52b652e4c706b6dda8398fc355c536dc67/Detectors/TPC/base/include/TPCBase/Mapper.h#L399-L400

bulukutlu: the nROC here is defined so that from zero to 36 it is IROCs and 36 to 72 it designates OROCs, so here the 
`(nROC >= mapper.getNumberOfIROCs()` 
is accurate. If wanted, to make it more clear I can change it to 
`(nROC >= mapper.getNumberOfIROCs() && nROC < mapper.getNumberOfIROCs()+mapper.getNumberOfOROCs()`
but shouldn't be necessary for the logic.

chiarazampolli: Ah true, I overlooked that it is just an `else` to the condition on line 50. But then, isn't just `else {` enough? without the check `if ((nROC >= mapper.getNumberOfIROCs()))`?

bulukutlu: yes you are right, `else if` instead of `else` is redundant here

chiarazampolli: Hello @bulukutlu , @wiechula , 
It woudl be good to have this PR finalized. If you agree to remove the `else`, can you please do so and then we can habve the PR merged? In this way we can have back the TPC cluster QC in production. 
Thanks!

chiarazampolli:
Hello @wiechula , @shahor02 , @bulukutlu , 
Let's take my comment later, and merge this, if you agree. We need the TPC QC back in business. 
Cheers,
Chiara

shahor02 (approved):
`else` was indeed enough, fixing abd approving to trigger CI

PR #9690, opened 2022-08-24
Adding list of FLPs to GRPECS

chiarazampolli:
@shahor02 , @davidrohr , @ktf

### DIFF
--- a/DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
+++ b/DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
@@ -112,6 +112,7 @@ class GRPECSObject
   // methods to manipulate the list of FLPs in the run
   const std::vector<unsigned short>& getListOfFLPs() const { return mFLPs; }
   void addFLP(unsigned short flp) { mFLPs.push_back(flp); }
+  void setListOfFLPs(std::vector<unsigned short> listFLPs) { mFLPs = listFLPs; }
   bool getFLPStatus(unsigned short flp) const
   {
     return std::find(mFLPs.begin(), mFLPs.end(), flp) == mFLPs.end() ? false : true;
--- a/DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
+++ b/DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
@@ -19,6 +19,7 @@
 #include <Rtypes.h>
 #include <cstdint>
 #include <ctime>
+#include <bitset>
 #include "DataFormatsParameters/ECSDataAdapters.h"
 #include "DetectorsCommonDataFormats/DetID.h"
 
@@ -109,6 +110,12 @@ class GRPECSObject
   /// same with comma-separate list of detector names
   DetID::mask_t getDetsReadOut(const std::string& only, const std::string& skip = "") const { return getDetsReadOut(DetID::getMask(only), DetID::getMask(skip)); }
 
+  // methods to manipulate the list of FLPs in the run
+  std::bitset<202> getListOfFLPs() const { return mFLPs; }
+  void setFLPStatus(size_t flp, bool status) { mFLPs.set(flp, status); }
+  bool getFLPStatus(size_t flp) const { return mFLPs.test(flp); }
+  bool listOfFLPsSet() const { mFLPs.count() > 0 ? true : false; }
+
   /// print itself
   void print() const;
 
@@ -128,11 +135,12 @@ class GRPECSObject
   int mRun = 0;                     ///< run identifier
   RunType mRunType = RunType::NONE; ///< run type
   std::string mDataPeriod{};        ///< name of the period
+  std::bitset<202> mFLPs{};         ///< to store which FLPs were in the processing
 
   // detectors which are always readout in triggered mode. Others are continuous by default but exceptionally can be triggered
   static constexpr DetID::mask_t DefTriggeredDets = DetID::getMask(DetID::TRD) | DetID::getMask(DetID::PHS) | DetID::getMask(DetID::CPV) | DetID::getMask(DetID::EMC) | DetID::getMask(DetID::HMP);
 
-  ClassDefNV(GRPECSObject, 4);
+  ClassDefNV(GRPECSObject, 5);
 };
 
 } // namespace parameters
--- a/DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
+++ b/DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
@@ -19,7 +19,6 @@
 #include <Rtypes.h>
 #include <cstdint>
 #include <ctime>
-#include <bitset>
 #include "DataFormatsParameters/ECSDataAdapters.h"
 #include "DetectorsCommonDataFormats/DetID.h"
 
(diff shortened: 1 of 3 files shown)


shahor02 (commented):
Hi @chiarazampolli 
not sure the bitset (change of N is not supported by schema evolution) is appropriate: what if in the future the range of IDs changes? I assume the ECS will provide a string of FLP hostnames, would store it directly.

davidrohr:
Hi, not sure if we want hostnames? Then we do need to do a lot of string comparison. How about a vector of integers with the flp numbers?

chiarazampolli:
I assumed that we will never have more than the present 202 FLPs. I would also not store the hostnames, and I hope FLP does not provide that. I can change it to a vector of integers, fine with me. The way we will fill it depends on what AliECS will pass.

shahor02:
vector<int> would also work, provided the hostname format is carved in stone. The string comparisons would need to be done only once, at initialization.

shahor02:
> I assumed that we will never have more than the present 202 FLPs.

The number of FLPs will probably not increase, but the IDs, theoretically, may.

davidrohr:
I don't see why we depend on the hostname format. Each hostname should contain the FLP number in one or the other form, so if we would just extract that once at the beginning, aliecs can also just put the numbers in directly. In case we would add more FLPs, it would work transparently, as long as they get a number.
Only problem would be if the numbering would change, i.e. TPC would suddenly have other FLP numbers. Then we'd need to ajust the code to connect to the correct FLPs.

shahor02:
@davidrohr if you are using integers, don't you need to compose the hostname? Is it guaranteed that names will be always `alio2-cr1-flp<XXX>` ?

davidrohr:
Ok, point, that is what you meant. But also, we need to filter the host names to find zhe tpc flps to connect to. So I think we need the host name scheme in the source code anyway. Or we filter for just all numbers in the string, but don't really like that.

Kind Regards
David Rohr

On 24 August 2022 18:27:18 CEST, Ruben Shahoyan ***@***.***> wrote:
***@***.*** if you are using integers, don't you need to compose the hostname? Is it guaranteed that names will be always `alio2-cr1-flp<XXX>` ?

chiarazampolli:
We can also have vector of strings, but this will not help in knowing which FLP is for which detector. Such mapping should not be hardcoded, if you really think that it can change.
Additionally, remember that the GRP won’t be used for IDCs, but just in case this info is useful in reco.

DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
```diff
@@ -109,6 +110,18 @@ class GRPECSObject
   /// same with comma-separate list of detector names
   DetID::mask_t getDetsReadOut(const std::string& only, const std::string& skip = "") const { return getDetsReadOut(DetID::getMask(only), DetID::getMask(skip)); }
 
+  // methods to manipulate the list of FLPs in the run
+  std::vector<unsigned short> getListOfFLPs() const { return mFLPs; }
```
shahor02: should be 
`const std::vector<unsigned short>& getListOfFLPs() const { return mFLPs; }`

shahor02 (changes requested):
thanks, just a small comment

DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
chiarazampolli: Sure, will change it ASAP.

chiarazampolli: done.

chiarazampolli:
Hello @shahor02 , 
Can this be merged now? Or are there additional comments?
Thanks!
Chiara

DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
```diff
@@ -109,6 +110,18 @@ class GRPECSObject
   /// same with comma-separate list of detector names
   DetID::mask_t getDetsReadOut(const std::string& only, const std::string& skip = "") const { return getDetsReadOut(DetID::getMask(only), DetID::getMask(skip)); }
 
+  // methods to manipulate the list of FLPs in the run
+  const std::vector<unsigned short> getListOfFLPs() const { return mFLPs; }
```
shahor02: please return `const std::vector<unsigned short>&`

DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h
```diff
@@ -19,6 +19,7 @@
 #include <Rtypes.h>
 #include <cstdint>
 #include <ctime>
+#include <bitset>
```
shahor02: not needed

shahor02 (changes requested):
sorry, forgot about this PR, see below

chiarazampolli:
Done.

chiarazampolli:
Maybe at this point we can add here the change to the executable too. So let's wait to merge. I will take care.

chiarazampolli:
Hello @TimoWilken , @ktf ,
The error does not point to this PR, can you check?
```
[INFO] 1st reading of A: testObjectO
[INFO] ccdb reads http://ccdb-test.cern.ch:8080/Test/CachingB/1000/febcfe9c-2add-11ed-a816-c0a80209250c for 1500 (retrieve, agent_id: aliflow01.cern.ch-1662136846-teP1Bu), 
[INFO] Reading of cached and modified A: Cached
[ERROR] Requested resource does not exist: http://ccdb-test.cern.ch:8080/Test/CachingA/500/
[INFO] Read for wrong time, expect null: 0
[INFO] ccdb reads http://ccdb-test.cern.ch:8080/Test/CachingA/1000/febc01dc-2add-11ed-a816-c0a80209250c for 1500 (retrieve, agent_id: aliflow01.cern.ch-1662136846-teP1Bu), 
[INFO] Reading of A from scratch after error: testObjectO
--
[INFO] Reading same A, expect cached and modified value: Cached
[INFO] ccdb reads http://ccdb-test.cern.ch:8080/Test/CachingB/1000/fee59254-2add-11ed-a816-c0a80209250c for 1500 (retrieve, agent_id: aliflow01.cern.ch-1662136846-teP1Bu), 
[INFO] Reading B, expect cached since only A cache was cleaned: testObjectO
/sw/SOURCES/O2/9690-slc7_x86-64/0/CCDB/test/testBasicCCDBManager.cxx(97): [1;31;49merror: in "TestBasicCCDBManager": check objB && (*objB) == hack has failed[0;39;49m
[INFO] ccdb reads http://ccdb-test.cern.ch:8080/Test/CachingB/1000/fee59254-2add-11ed-a816-c0a80209250c for 1500 (retrieve, agent_id: aliflow01.cern.ch-1662136846-teP1Bu), 
[INFO] Reading B after cleaning cache completely: testObjectO
[ERROR] Requested resource does not exist: http://ccdb-test.cern.ch:8080/Test/CachingA/1500/
[ERROR] Requested resource does not exist: http://ccdb-test.cern.ch:8080/Test/CachingA/1500/
[INFO] ccdb reads http://ccdb-test.cern.ch:8080/Test/CachingA for 1500 (retrieve, agent_id: aliflow01.cern.ch-1662136846-teP1Bu), 
[INFO] Reading A after disabling the cache: testObjectO
[INFO] ccdb reads http://ccdb-test.cern.ch:8080/Test/CachingA for 1500 (retrieve, agent_id: aliflow01.cern.ch-1662136846-teP1Bu), 
--
[26336:simple]: [16:49:50][INFO] Correctly handshaken websocket connection.
[INFO] Quitting
[INFO] QUIT_REQUESTED
[ERROR] pid 26337 (internal-dpl-injected-dummy-sink) crashed with 1
[26335:internal-dpl-clock]: [16:49:50][INFO] Received device shutdown request (signal 15).
[26335:internal-dpl-clock]: [16:49:50][INFO] Waiting for graceful device shutdown. Hit Ctrl-C again to abort immediately.
[26335:internal-dpl-clock]: [16:49:50][INFO] New state requested. No timeout set, quitting immediately as per --completion-policy
--
[26336:simple]: [16:49:51][STATE] IDLE ---> EXITING
[26336:simple]: [16:49:51][STATE] Exiting FairMQ state machine
[INFO] Dumping used configuration in dpl-config.json
[ERROR] SEVERE: Device internal-dpl-injected-dummy-sink (26337) returned with 1

        Start 103: test_Framework_test_SimpleRDataFrameProcessing
 89/106 Test #103: test_Framework_test_SimpleRDataFrameProcessing ..........   Passed    7.10 sec
--
```

chiarazampolli:
Hello @shahor02 , 
Can this be merged?
Chiara

shahor02:
actually, all CIs were failed due to the typo (perhaps in my fix)

chiarazampolli:
Hello @shahor02 ,

now it is really unrelated, the error. I don’t know what is happening with CCDB, but I would consider merging this.

chiara

PR #9691, opened 2022-08-24
TRD raw reader update

martenole:
- Tracklet parsing moved into CruRawReader
- prepared Link ID to HCID mapping in CCDB

Still to be done are digit parsing and the event record stuff, this will follow after this is through. Then one can also activate the new link mapping.
Also a few error messages might need to be adapted, but this can also be done with the next PR


martenole:
A few more words, since this turned out to become a rewrite of most of the actual parsing of the TRD raw data.
1. Due to many hotfixes and workarounds over the last months there were quite some parts of code which were not used anymore or which were never. A lot of those parts are removed now in order to make the code clearer and better maintainable.
2. The digit and trackler parsing was done in separate classes. I did not see any benefit in that. The actual parsing is not that long. And having it all in the `CruRawReader` gives I think a better overview (no need to carry around all different kinds of options and the error reporting can all be done in a single place).
3. So far the `TrackletHCHeader`s were not properly checked. In addition to the wrongly connected links which were not noticed before also a bug in the Sim->Raw conversion was uncovered by the CI (see changes in `constructTrackletHCHeader`). This means that at some point we should regenerate the synthetic data samples. Until then a workaround has been put in place, so that the TrackletHCHeaders can still be ignored by the raw reader. This should be activated until the new data samples are available (but only for synthetics, not for real data).
4. The raw reader now reports for every TF it has seen the total data size of its input (including all headers, so also the RDH), the data which was rejected by the digit and tracklet parsing (here only the words count which could not be read) and the total number of digits and tracklets. Comparing the size of the dropped data to the sizes of digits and tracklets gives an idea on how many digits/tracklets were dropped.

In general I tried to still follow the same logic which was already in place. I think there is still some optimization potential , but the first priority should be to have a stable and maintainable raw data parsing. How stable it is with the changes from this PR still has to be proven in production of course... But my tests look promising.

After this is merged we need to create the CCDB object for the HCID to Link mapping and then add a few changes which are all already prepared in the code. For the synthetic data sample we might need to use the CCDB object as well. Here I need to check which timestamp we are using during running. Maybe we can also keep the default mapping for synthetics.

martenole:
When going through the data from the noise run 520813 I noticed the current digit parsing was too restrictive. In case there is invalid ADC data discovered I try to find the next MCM header now instead of rejecting the rest of the data from the link completely.
Invalid ADC data is present in all half chambers. @bazinski I think it would make sense to add some error counting not only on the level of HCID, but also in the level of individual MCMs. What do you think about it? How hard would it be to add this to QC? Adding it to the event records is very easy and straight forward

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
```diff
@@ -54,6 +54,13 @@ struct HelperMethods {
     printf("%02i_%i_%i\n", det / constants::NCHAMBERPERSEC, (det % constants::NCHAMBERPERSEC) / constants::NLAYER, det % constants::NLAYER);
   }
 
+  static void printSectorStackLayerSide(int hcid)
+  {
+    // for a given half-chamber number prints SECTOR_STACK_LAYER_side
+    int det = hcid / 2;
+    printf("%02i_%i_%i side: %i\n", det / constants::NCHAMBERPERSEC, (det % constants::NCHAMBERPERSEC) / constants::NLAYER, det % constants::NLAYER, hcid % 2);
+  }
+
```
tdietel: I have been wondering if we should have some general formatting goodies, either for fmt::format or C++ ostreams.
I think we have some ostream operators for digits (and tracklets?). 
I would also suggest to write this as e.g. '01_2_3A'. or '04_5_1B'.

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
```diff
@@ -398,48 +401,50 @@ struct trdTestPattern6 {
...
-      uint32_t c : 2; // c is wrong I cant remember name, but not a concern at the moment.
+      uint32_t f : 2;
       uint32_t z : 10;
       uint32_t y : 10;
       uint32_t x : 10;
     } __attribute__((__packed__));
   };
 };
 
+struct LinkToHCIDMapping {
```
tdietel, line 417: I guess this is a placeholder for the CCDB object that we want to have? Can you add a comment that this struct will be replaced soon?

DataFormats/Detectors/TRD/test/testRawData.cxx
```diff
@@ -109,15 +109,15 @@ BOOST_AUTO_TEST_CASE(TRDRawDataHeaderInternals)
...
+  // auto trackletcount = getNumberOfTrackletsFromHeader(&header);
 
-  BOOST_CHECK(trackletcount == 1);
-  BOOST_CHECK(invalidheader == 0);
-  //std::cout << std::hex << " charges[0]:0x"<<(int)charges[0];
-  //std::cout << std::hex << " charges[1]:0x"<<(int)charges[1];
-  //std::cout << std::hex << " charges[2]:0x"<<(int)charges[2];
-  BOOST_CHECK((int)charges[2] == 0x02);
+  // BOOST_CHECK(trackletcount == 1);
+  // BOOST_CHECK(invalidheader == 0);
```
tdietel, line 116: Is there a reason why these checks were removed?

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
tdietel: I just saw the macro to create this object below. Having two maps as the final solution does not seem elegant - although I see the simplicity. 
One idea would be to create one map from the other in the constructor - or at least verify the mapping.
I also found boost::bimap - we are obviously not the first ones to have this problem.

tdietel: And is there a reason not to implement swapLinks as a method of this class? Maybe because we do not want the user to see/use this method?

Detectors/TRD/reconstruction/macros/createLinkToHCIDMapping.C
```diff
@@ -0,0 +1,93 @@
...
+
+#include <chrono>
+#include <map>
+
+#endif
+
+using namespace o2::trd;
+
+// linkA and linkB refer to the global ORI index and not to the half-chamber ID
+void swapLinks(int linkA, int linkB, std::map<int, int>& linkIDToHCID, std::map<int, int>& hcIDToLinkID)
```
tdietel: As mentioned before, this could be implemented as a method in the CCDB object.

Detectors/TRD/reconstruction/src/EventRecord.cxx
```diff
@@ -44,192 +40,58 @@
...
-
-void EventRecord::addTracklets(std::vector<Tracklet64>& tracklets)
-{
-  for (auto tracklet : tracklets) {
-    mTracklets.push_back(tracklet);
-  }
-}
-
-void EventRecord::sortByHCID()
+void EventRecord::sortTrackletsByHCID()
```
tdietel: I might have seen a comment in the corresponding header:
Why do we sort by HCID? For things like tracking, it might be more natural to sort by detector and then by pad row and pad column.  But that would interweave the tracklets from the two half-chambers.
Why do we sort at all? Is the order relevant at some point?

Detectors/TRD/reconstruction/src/EventRecord.cxx
```diff
@@ -44,192 +40,58 @@
...
-  digits.reserve(digitcount);
   std::vector<TriggerRecord> triggers;
-  triggers.reserve(triggercount);
   for (auto& event : mEventRecords) {
-    //sort tracklets
-    event.sortByHCID();
-    //TODO do this sort in parallel over the events
-    tracklets.insert(std::end(tracklets), std::begin(event.getTracklets()), std::end(event.getTracklets()));
-    digits.insert(std::end(digits), std::begin(event.getDigits()), std::end(event.getDigits()));
+    event.sortTrackletsByHCID();
```
tdietel: It looks inefficient to copy all digits and tracklets from the event vectors to TF vectors. Wouldn't it be more efficient to build the TF vectors from the start?
This is certainly the less invasive change - not sure if we should address this now.

tdietel: Obviously, the add_tracklet() and add_digit() methods would have to move from the EventRecord to the EventRecordContainer class.

Detectors/TRD/reconstruction/README.md
```diff
@@ -4,17 +4,18 @@
...
+## Introduction
 
-- Parts to reconstruction :
-    - flp :
-        - o2-trd-compressor
-            This takes in the data from the CRU, optionally compresses it or sends it raw out to the epn via workflows.
-    - epn :
-        - o2-trd-rawreader
-            Take the piped in raw data and unpacks it to a vector tracklets/digits/triggerrecords.
+12 FLPs are receiving the data from the TRD detector. Each FLP is equipped with 3 CRUs. In total we thus have 36 CRUs. The CRUs pack the data into the O2 raw data format. This means that Raw Data Headers (RDH) are added to the data stream and the data from each link (one link receives the data from one TRD half-chamber) are aligned to 256 bits. If there is less data for one link then padding words (0xeeeeeeee) are added to keep the alignment.
```
tdietel, line 9: It certainly helps to have this in the README. But I am not an expert how the CRU packs the link data, and at which level the RDH headers are inserted. It would be good if Guido could have a look at this. 

I think of the CRU processing in the following order:
- first the links are aligned to 256 bits
- then the Half-CRU headers are created, with data sizes for each link that can be used to calculate the offsets
- then all the data blocks (header + 15x link data) for one HCRU are concatenated for one time frame
- then this TF-HCRU data block is split into a number of 8kiB pages, with an RDH header at the beginning of each page.

Detectors/TRD/reconstruction/include/TRDReconstruction/EventRecord.h
```diff
@@ -34,100 +30,63 @@ class TriggerRecord;
...
-  TRDDataCountersPerEvent mEventStats{};
-  //statistics stuff these get passed to the per tf data at the end of the timeframe,
-  //but as we read in per link, events are seperated hence these counters
+  TRDDataCountersPerEvent mEventStats{}; /// statistics, for this trigger
 };
 
-class EventStorage
+/// \class EventRecordContainer
+/// \brief Stores the TRD data for one TF i.e. a vector of EventRecords and some statistics
+class EventRecordContainer
```
tdietel, line 79: Is EventRecordContainer the right name here?
Including 'TimeFrame' might make it clearer what is actually in here.
And 'EventRecord' does not sound like digits and tracklets are actually included here.

Detectors/TRD/reconstruction/include/TRDReconstruction/CruRawReader.h
```diff
@@ -104,6 +110,21 @@ class CruRawReader
   // sanity check on individual RDH (can find unconfigured FLP or invalid data)
   bool checkRDH(const o2::header::RDHAny* rdh);
 
+  // given the total link size and the hcid from the RDH
+  // parse the tracklet data. Overwrite hcid from TrackletHCHeader if mismatch is detected
+  // trackletWordsRejected:  count the number of words which were skipped (subset of words read)
+  // returns total number of words read (no matter if parsed successfully or not)
+  int parseTrackletLinkData(int linkSize32, int& hcid, int& trackletWordsRejected);
+
+  int parseDigitLinkData(int maxWords32, int hcid, int& digitWordsRejected);
```
tdietel: This is a very general comment.
The merged parser processes data from two different sources: tracklets and digits from the FEE, and RDH and HCRU headers from the CRU. I have not yet looked at the actual parser, but this might make it difficult to separate the different domains. 
A complication is that the the CRU data stream is split into 8kiB blocks (incl. RDH headers). The RDH headers then appear in the middle of the CRU/FEE data blocks.

tdietel: A more specific comment arising from the one above.
It might be useful to be able to parse a half-chamber data link outside of a CRU/RDH data stream. For example, if we use the TRD chamber at UCT to read out test data. I do not know if this will ever be used, but now it's probably easy to make this possible.

DataFormats/Detectors/TRD/src/RawData.cxx
```diff
@@ -292,25 +183,25 @@ DigitMCMADCMask constructBlankADCMask()
   return mask;
 }
 
-void printTrackletHCHeader(o2::trd::TrackletHCHeader& halfchamber)
+void printTrackletHCHeader(const o2::trd::TrackletHCHeader& halfchamber)
```
tdietel, line 212: This seems to duplicate the functionality of `std::ostream& operator<<(std::ostream& stream, const TrackletHCHeader halfchamberheader)` in line 36.

tdietel (commented):
I have gone through most of the periphery, but it's too late for the CruRawReader. 
Most of what I've seen makes sense, I have added a few points inline in the code.
To review the tracklets/digits parsers, it would be useful to see a state diagram. This part is probably correct, so we might not learn too much from it - but it would be two nice figures in @bazinski's thesis.
I think the critical question is what finer level errors there are (you revised the enum that lists all of them), how they are handled (and if this is the best we can do), how often the error actually occurs (we have the histograms), and if we can actually fix the problems upstream.

DataFormats/Detectors/TRD/test/testRawData.cxx
martenole: Yes, it is because the functions they rely on were removed. And I removed them since they are not needed anymore by the new parsing. If one would want to test that here one would need to create the data for one link and then use `CruRawReader::parseTrackletLinkData()` and compare input/output.
As a test I used the chain from simulation to raw data and then again to digits and tracklets and made sure that that is consistent.

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
martenole: Actually I was planning on using this struct as CCDB object. I agree it is not super elegant and googling for bidirectional maps you do get quite some examples on how one could do it alternatively. But in the end this is an object which is loaded once in the beginning by the raw reader and which afterwards never changes during the run. It is 7kBytes which are kept in memory all the time. This should really not impact the performance. I am not sure if we should spent much time thinking about an elegant solution as long as what we have works and is also readable.

martenole: We would update this object only when we have physically swapped links or if we notice in the future that more cables were not correctly plugged. We might see this from chambers which might be recovered in the future for example. Therefore we have this macro where we can just add a line for swapped links. Sure I could also put `swapLinks()` to the struct, there is no reason for keeping it in the macro.
The user will in the end anyhow only work with a `const` object received from the CCDB. So only `int getHCID(int link) const` and `int getLink(int hcid) const` will be used by the user

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
martenole: Ok

Detectors/TRD/reconstruction/src/EventRecord.cxx
martenole: The TRD tracking expects the tracklets ordered by detector ID. And it would be good to have the correct order in the output of the raw reader, because then we can move the tracklets around and don't need to duplicate them downstream. This we would need to do if we decided e.g. to do the sorting in the tracklet transformer. So indeed for me its fine to sort by detector and not HCID. In the header I made a comment that we might want to think about the sorting later. For this PR I was just doing what was already there

Detectors/TRD/reconstruction/README.md
martenole: I don't know more about the processing in the CRU. Here we should wait for Guido to comment. One correction to your points above: the half-CRU data are concatenated for one HBF and not for one TF. Currently one TF contains 128 HBFs.

Detectors/TRD/reconstruction/include/TRDReconstruction/EventRecord.h
martenole: Well, before we had the classes `EventRecord` and `EventStorage`. And one was holding all the data for one trigger/event and one was holding a vector of objects of the other type. You can guess which one was what...
I guessed wrong 50% of the time and I think `EventRecord` for digits, tracklets, InteractionRecord for a single event and `EventRecordContainer` do make more sense.

DataFormats/Detectors/TRD/src/RawData.cxx
martenole: There were many many duplicates and I tried to remove most of them. Probably I have missed some. It might make sense to have both an `operator<<` and a `print()` method. With the streamer you can assemble a log message without line breaks. I am not using the `operator<<` though so for me it would be fine to remove this

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
tdietel: Absolutely valid points. The only mis-leading item is then that the `isOK` method suggests a more thorough check: I would expect a forward and reverse mapping check for all links. Can you add a comment that this is a minimal check?

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
```diff
@@ -30,136 +30,83 @@ namespace o2::trd
...
+  DigitHCHeaderSVNMismatch,            // svn version information has changed in the DigitHCHeader3.
+  TrackletsReturnedMinusOne,           // trackletparsing returned -1, data was dumped;
+  FEEIDIsFFFF,                         // RDH is in error, the FEEID is 0xffff
+  FEEIDBadSector,                      // RDH is in error, the FEEID.supermodule is not a valid value.
+  HalfCRUBadBC,                        // the BC in the half-CRU header is so low that the BC shift would make it negative
+  TRDLastParsingError                  // This is to keep QC happy until we can change it there as well.
 };
 
-static std::unordered_map<int, std::string> ParsingErrorsString = {
+static const std::unordered_map<int, std::string> ParsingErrorsString = {
```
f3sch, line 71: Going through this a bit.
Isnt ParsingErrors::BadRDHMemSize missing in the map?
This would cause an exception if the rdh memory is invalid, which we probably do not want :smile: 
```cxx
  if (o2::raw::RDHUtils::getMemorySize(rdh) <= 0) {
    if (mMaxWarnPrinted > 0) {
      LOG(warn) << "Received RDH header with invalid memory size (<= 0) ";
      checkNoWarn();
    }
    incrementErrors(BadRDHMemSize, -1, fmt::format("Received RDH header with invalid memory size (<= 0) "));
    return false;
  }
}
```

Detectors/TRD/reconstruction/src/EventRecord.cxx
martenole: Not sure if this could be done so easily. The data comes in HBFs from single half CRUs. So in one chunk we always have multiple triggers for different links. If we would want to keep only single vectors of digits and tracklets we would need to do multiple checks when inserting (for trigger number and for detector ID). So this might turn out more processing intensive in the end. But that's just an assumption, I have not tested it.

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawData.h
martenole: I added this more thorough check now to the mapping

DataFormats/Detectors/TRD/include/DataFormatsTRD/RawDataStats.h
martenole: Thanks, indeed. I have added a runtime check on the size of the map and the number of entries in the enum now. A compile time check with static assert is unfortunately not possible with the map since it does not have a `constexpr` constructor

martenole:
@bazinski are the changes here fine with you?

I just let it run over a single TF for a couple of times with the version of the raw reader we have right now and with the one from this PR.
With the version in dev:
```
[57948:trd-datareader]: [14:28:53][INFO] Digits: 36469 (36469 TF), Tracklets: 141789 (141789 TF), DataRead in: 28870.314 MB, Rejected: 34943.840 MB for TF 25651 in 170 ms
```
New version:
```
[56579:trd-datareader]: [14:28:03][INFO] Digits: 36233, Tracklets: 141789, DataRead in: 13.470 MB, Rejected: 130.695 kB for TF 25651 in 30 ms
```
The reported processing time with the changes from this PR is a factor 5.6 faster. Also the reported data sizes are now corrected.

bazinski:
> A few more words, since this turned out to become a rewrite of most of the actual parsing of the TRD raw data.
> 
> 1. Due to many hotfixes and workarounds over the last months there were quite some parts of code which were not used anymore or which were never. A lot of those parts are removed now in order to make the code clearer and better maintainable.
> 2. The digit and trackler parsing was done in separate classes. I did not see any benefit in that. The actual parsing is not that long. And having it all in the `CruRawReader` gives I think a better overview (no need to carry around all different kinds of options and the error reporting can all be done in a single place).
> 3. So far the `TrackletHCHeader`s were not properly checked. In addition to the wrongly connected links which were not noticed before also a bug in the Sim->Raw conversion was uncovered by the CI (see changes in `constructTrackletHCHeader`). This means that at some point we should regenerate the synthetic data samples. Until then a workaround has been put in place, so that the TrackletHCHeaders can still be ignored by the raw reader. This should be activated until the new data samples are available (but only for synthetics, not for real data).
> 4. The raw reader now reports for every TF it has seen the total data size of its input (including all headers, so also the RDH), the data which was rejected by the digit and tracklet parsing (here only the words count which could not be read) and the total number of digits and tracklets. Comparing the size of the dropped data to the sizes of digits and tracklets gives an idea on how many digits/tracklets were dropped.
> 
> In general I tried to still follow the same logic which was already in place. I think there is still some optimization potential , but the first priority should be to have a stable and maintainable raw data parsing. How stable it is with the changes from this PR still has to be proven in production of course... But my tests look promising.
> 
> After this is merged we need to create the CCDB object for the HCID to Link mapping and then add a few changes which are all already prepared in the code. For the synthetic data sample we might need to use the CCDB object as well. Here I need to check which timestamp we are using during running. Maybe we can also keep the default mapping for synthetics.

1. agreed.
2. of course having the seperate classes is a pain, ergo that would imply a good reason. The code is shared with the compressor, so as not to have 2 pieces of code parsing the respective data. It also permits trivial root macros to parse data.
3. not checked as it was not necesarily there so could not be relied on, now it is guaranteed to be there.
4. not sure why the numbers went wrong, those are sent to qc.

bazinski:
> When going through the data from the noise run 520813 I noticed the current digit parsing was too restrictive. In case there is invalid ADC data discovered I try to find the next MCM header now instead of rejecting the rest of the data from the link completely. Invalid ADC data is present in all half chambers. @bazinski I think it would make sense to add some error counting not only on the level of HCID, but also in the level of individual MCMs. What do you think about it? How hard would it be to add this to QC? Adding it to the event records is very easy and straight forward

It will get complicated as its a simple blob of data sent across to qc, with that granularity it would make sense to rather use a sparse array. It could still be done as a blob as its currently done, just have to change the calculations, and not sure how one syncs that up with qc, similar to the problem with the last error position name having to stay for qc to pass.

Detectors/TRD/reconstruction/src/EventRecord.cxx
bazinski: I dont remember the reason for hcid sorting, but it was probably mid last year. I thought it was to make it the same as the output of the simulation or someone required the digits/tracklets to be in that order, its neither here nor there, the sort could be changed to anything, or moved out and done later.

Detectors/TRD/reconstruction/src/EventRecord.cxx
bazinski: data is read is in as : 
1a 1b 1c 1d 1e 1f 2a 2c 2d 2e 2f 3b 3c 3d 3e 3f ..... 1079 1079c 1079f 1g 1h 1i 1j 2g 2h 2i 2j .... 
a being event and integer being hcid, the hcid order is not sequential, but for the purposes of this it will suffice.

You dont know how big each block is, so either you bin them by 1,2,3,4 or you bin them by a,b,c,d,e,f
leading to a store of each event, which when finished is then merged into the message memory space via the copies above.
Now one could change the trigger record  format to include a range of ranges, but that would mess with a lot of code.
Or one could have an additional dpl device to merge that all together, but why.
Now there are other copies that one can remove but I doubt you will get past this one.

Detectors/TRD/reconstruction/include/TRDReconstruction/EventRecord.h
bazinski: I fail to see the need for sarcasm if you don't understand something ask.
Event Record holds an event, in line with the convention naming the trigger record.
EventStorage holds an entire timeframe. as per the comment in the class : //store a timeframes events for later collating sending on as a message

DataFormats/Detectors/TRD/src/RawData.cxx
bazinski: the duplicates were to make a one liner and a more verbose version as far as i remember.

Detectors/TRD/reconstruction/README.md
bazinski: > @bazinski are the changes here fine with you?
> 
> I just let it run over a single TF for a couple of times with the version of the raw reader we have right now and with the one from this PR. With the version in dev:
> 
> ```
> [57948:trd-datareader]: [14:28:53][INFO] Digits: 36469 (36469 TF), Tracklets: 141789 (141789 TF), DataRead in: 28870.314 MB, Rejected: 34943.840 MB for TF 25651 in 170 ms
> ```
> 
> New version:
> 
> ```
> [56579:trd-datareader]: [14:28:03][INFO] Digits: 36233, Tracklets: 141789, DataRead in: 13.470 MB, Rejected: 130.695 kB for TF 25651 in 30 ms
> ```
> 
> The reported processing time with the changes from this PR is a factor 5.6 faster. Also the reported data sizes are now corrected.

I am not sure why the speed is a surprise, it was one of the 2 reasons for the rewrite that I was doing.

martenole:
> of course having the seperate classes is a pain, ergo that would imply a good reason. The code is shared with the compressor, so as not to have 2 pieces of code parsing the respective data. It also permits trivial root macros to parse data.

The parsing methods are `public`, so nothing is preventing from using them in trivial macros.

> not checked as it was not necesarily there so could not be relied on, now it is guaranteed to be there.

Even if they are not always there, when they are, I think they should be checked.

Detectors/TRD/reconstruction/include/TRDReconstruction/EventRecord.h
martenole: If you want I can change the names back. I thought the ones proposed here are more intuitive, that's all

bazinski: I was replying to your comment. I have not opinions on what is a good name or bad.

Detectors/TRD/reconstruction/include/TRDReconstruction/EventRecord.h
```diff
@@ -34,100 +30,63 @@ class TriggerRecord;
...
-  }
-  void clear()
-  {
-    mDigits.clear();
-    mTracklets.clear();
-  }
-
-  void incStats(int tracklets, int digits, int wordsread, int wordsrejected);
+  // OS: Do we need to keep event statistics at all? Are they not anyhow accumulated for a whole TF?
+  //     or are they also used on a per event basis somewhere?
```
bazinski, line 68: The original reason was to have the actual digits per event. But it proves onerous to send that data, and more importantly for what purpose? Currently, it is averaged over a timeframe.

bazinski: not only digits of course.

Detectors/TRD/reconstruction/src/DataReader.cxx
```diff
@@ -33,18 +33,12 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
 {
 
   std::vector<o2::framework::ConfigParamSpec> options{
-    {"output-desc", VariantType::String, "TRDTLT", {"Output specs description string."}},
     {"verbose", VariantType::Bool, false, {"Enable verbose epn data reading."}},
-    {"verbosehalfcru", VariantType::Bool, false, {"Enable verbose for a halfcru, the halfcru contents are dumped out in hex."}},
```
bazinski, line 38: why is the dumping out of halfcru, link, parse words dropped?
It's invaluable when things go unexpectedly wrong. Clearly not useable in a run online.

Detectors/TRD/reconstruction/src/EventRecord.cxx
```diff
@@ -44,192 +40,58 @@
...
   int sumtracklettime = 0;
-  int sumtime = 0;
   uint64_t sumwordsrejected = 0;
   uint64_t sumwordsread = 0;
   for (auto event : mEventRecords) {
     sumtracklets += event.getEventStats().mTrackletsFound;
     sumdigits += event.getEventStats().mDigitsFound;
     sumtracklettime += event.getEventStats().mTimeTakenForTracklets;
     sumdigittime += event.getEventStats().mTimeTakenForDigits;
+    // OS: the two counters below are not even used anymore, are they needed?
```
bazinski, line 94: no they are not

Detectors/TRD/reconstruction/src/DataReader.cxx
martenole: I don't see an advantage over only using the `verbose` option. It contains everything needed for debugging. It might be good at some point to add an option to only dump the full link data once in case of an error. Since this could then also be used in production. This I would add with a followup PR

Detectors/TRD/reconstruction/src/EventRecord.cxx
martenole: good, then they'll also go with the next PR

bazinski (approved):

PR #9695, opened 2022-08-25
ITS/MFT Clusterers were using local GRP file

shahor02:


Detectors/Vertexing/include/DetectorsVertexing/PVertexer.h
```diff
@@ -38,7 +38,7 @@
 
 //TODO: MeanVertex and parameters input from CCDB
 
-//#define _PV_DEBUG_TREE_ // if enabled, produce dbscan and vertex comparison dump
+#define _PV_DEBUG_TREE_ // if enabled, produce dbscan and vertex comparison dump
```
mconcas:
```suggestion
//#define _PV_DEBUG_TREE_ // if enabled, produce dbscan and vertex comparison dump
```

maybe a leftover?

shahor02: Yes,  thanks for spotting!

rpezzi (approved):

PR #9696, opened 2022-08-25
MRRTF-176: Fix mix-up between milliseconds and seconds in duration check

aphecetche:


aphecetche:
@shahor02 this solves the issue of the MCH and MID DCS processor uploading to the CCDB every second or so. Should now respect their `--max-duration` options...

shahor02:
ok, thanks, shall we enable MCH once this is merged / deployed? What about MID?

aphecetche:
> ok, thanks, shall we enable MCH once this is merged / deployed? What about MID?

@shahor02 yes, please. 

MID works as well. 

Note that for MCH we still miss part of the LV datapoints (while we have all HV ones), but that's not due to the processor (will be followed-up with our DCS experts).

shahor02:
@aphecetche So, I will ask Dominik to redirect the MCH and MID back to the adapos server, so that these inputs will not be available on the alicers05 (once your PR is enabled in the next FLPSuite)?

pillot (approved):

PR #9697, opened 2022-08-25
[EMCAL-565]: Fix for bad channel histo range, reorganize EMCALCalibParams

jokonig:
- The bad channel calibration crashed due to the histogram axis beeing
initialized with a range from 0-1GeV while the analysis went up to 4GeV
which leads to undefined behavior
- Now the histogram ranges are stored in the CalibParams and could be
modified if needed. The maximum energy was increased to 10GeV
- Reorganized the EMCALCalibParams to enable setting different values
for the bad channel and the time calibration
- Delete the CalibInitParams and instead moved the values to the
EMCALCalibParams to unify all calibration parameters in one class


jokonig:
@shahor02 I think this can be merged

PR #9698, opened 2022-08-25
Fixes for Mean Vertex calibration

chiarazampolli:
Dear @martenole , @davidrohr , @shahor02 ,

This PR fixes the issues in the mean vertex calibration - at least locally. 
@shahor02 : there was also a typo, and the fitGaus was considered as failing when the return code was 10 instead of -10. But I also see that for the tgLamdba based vdrift calib, you consider a failure if the result is `<-2` (https://github.com/AliceO2Group/AliceO2/blob/825cb83467cd666ed96dd8f1457e5cd726d9150d/Detectors/TPC/calibration/src/TPCVDriftTglCalibration.cxx#L55). Can you confirm that -10 is correct?

Thanks,

Chiara

### DIFF
--- a/Detectors/Calibration/include/DetectorsCalibration/MeanVertexData.h
+++ b/Detectors/Calibration/include/DetectorsCalibration/MeanVertexData.h
@@ -30,7 +30,7 @@ struct MeanVertexData {
   std::vector<std::array<float, 3>> histoVtx{0};
   bool mVerbose = false;
 
-  MeanVertexData();
+  MeanVertexData() = default;
 
   ~MeanVertexData()
   {
--- a/Detectors/Calibration/src/MeanVertexCalibrator.cxx
+++ b/Detectors/Calibration/src/MeanVertexCalibrator.cxx
@@ -69,6 +69,9 @@ void MeanVertexCalibrator::binVector(std::vector<float>& vectOut, const std::vec
       continue;
     }
     int bin = (vectIn[i] - min) * binWidthInv;
+    if (bin >= vectOut.size()) {
+      continue;
+    }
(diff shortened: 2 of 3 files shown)


shahor02:
Hi @chiarazampolli 
-10 is definitive failure, the next one, -3 means that the fit per se has failed (e.g. data was too noisy) and the value comes from the recovery procedure (mean and rms), up to the user to accept the mean of a noisy data as a substitute for the gaussian fit. 
-2 and -1 means that there was only 1 or 2 entries in the sample, in the case also the value comes from the mean / rms.
In the dTgl I don't use -3 as I have many bins in Z and the loss of a few dubious ones should not be a problem.

In principle, with enough statistics, you should never have a negative result, but yes, it is of to declare failure on -10 only.

shahor02 (approved):

chiarazampolli:
> Hi @chiarazampolli -10 is definitive failure, the next one, -3 means that the fit per se has failed (e.g. data was too noisy) and the value comes from the recovery procedure (mean and rms), up to the user to accept the mean of a noisy data as a substitute for the gaussian fit. -2 and -1 means that there was only 1 or 2 entries in the sample, in the case also the value comes from the mean / rms. In the dTgl I don't use -3 as I have many bins in Z and the loss of a few dubious ones should not be a problem.
> 
> In principle, with enough statistics, you should never have a negative result, but yes, it is of to declare failure on -10 only.

Hi @shahor02 , 
I had seen in the fitGaus implementation the meaning of the codes, I was wondering why you don't accept anything below -2. Now it is clearer, you prefer to keep better fits only.
Thanks!

PR #9704, opened 2022-08-26
DPL: drop all messages when cleanup property increases

ktf:
This is an attempt at doing what we discussed. Frankly speaking
I am more and more convinced this is a hack (if it works at all)
and we should either have a proper DRAIN state, or we discard messages
based on the run number.


davidrohr:
What is the problem with this? actually looks quite straight-forward

ktf:
First of all, it's not clear to me how doing something in Subscribe() will avoid the state machine to advance. Properties are notified from a separate thread and the mutex is released before the callback is invoked.
Secondly, it relies on the fact that Receive returns 0 parts, but there is no way I can know if they are really 0, or it just happened to be 0.
If this works reliably, there is something I do not understand.

PR #9705, opened 2022-08-26
add one option for TOF workflow in epnsync mode

noferini:


noferini:
Is there any objection to merge this PR?

shahor02:
Hi, not sure what `to disable check on problematic, otherwise masked for new calibrations` means, what it does?

noferini:
problematic channels are defined at the end of TOF channel offset calibration if the fraction of entries below the peak/fit for a given channel is too low or if the width of the Gaussian fit is unrealistic (Run-2 criterium).
Then the list of problematics is read from the calibration object during reconstruction.
Problematic channels are filter during clusterization and before the matching both in data and MC by default.
However, when moving to a new run or a new calibration the new list of problematic should be redone from scratch (it could be we are able to recover channels not working in the past). The only way to do that is not to filter them in the matching since matching provides the new calib info (if a channel is filtered it will not be calibrated).
In sync calibration (or in a ppass) this flag is then required.

shahor02 (approved):
ok, thanks, merging.

PR #9706, opened 2022-08-26
DCSConfigObject to noise calib workflow and extra cut for ITS IB

iravasen:
This PR introduces the DCSConfigObject (for DCS CCDB) for the noise calibration workflow. It's a copy of `o2::itsmft::NoiseMap` which can be read out from DCS. 

In WIP just because I need confirmation from Hartmut about the format of the object. Can be reviewed anyway (cc @shahor02 and @chiarazampolli and @iouribelikov )

New option introduced (`--cut-ib`) which allow to apply a special cut for Inner Barrel. So far, we are cutting at 10^-2 for IB and 10^-6 for OB and this is what we would like to keep (maybe 10^-5 can be considered for OB). 

Workflow tested in the following way:
```
o2-ctf-reader-workflow -b --ctf-input /data/iravasen/513464/513464_random.dat --remote-regex "^alien:///alice/data/.+" --copy-cmd no-copy --onlyDet ITS --its-digits | \
o2-its-noise-calib-workflow -b --prob-threshold 1e-4 --cut-ib 1e-2 --nthreads 2 | \
o2-calibration-ccdb-populator-workflow -b --ccdb-path "http://ccdb-test.cern.ch:8080" --sspec-min 0 --sspec-max 0 | \
o2-calibration-ccdb-populator-workflow -b --ccdb-path "http://ccdb-test.cern.ch:8080" --sspec-min 1 --sspec-max 1 --name-extention dcs
```

It has to be noted that even if `NoiseMap.h` has been changed, it is NOT needed to change MFT workflow. In real life, the first calib workflow will ship to prod CCDB while the second to DCS CCDB.

### DIFF
--- a/DataFormats/Detectors/ITSMFT/common/include/DataFormatsITSMFT/NoiseMap.h
+++ b/DataFormats/Detectors/ITSMFT/common/include/DataFormatsITSMFT/NoiseMap.h
@@ -93,9 +93,10 @@ class NoiseMap
     return dumpAboveThreshold(p * mNumOfStrobes);
   }
 
-  void applyProbThreshold(float t, long int n, float relErr = 0.2f)
+  void applyProbThreshold(float t, long int n, float relErr = 0.2f, int minChipID = 0, int maxChipID = 24119)
   {
     // Remove from the maps all pixels with the firing probability below the threshold
+    // Apply the cut only for chips between minChipID and maxChipID (included)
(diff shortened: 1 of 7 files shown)


iravasen:
we agreed on the format so the PR is ready from my side.

iravasen:
This time it seems the test errors are not related to this PR

PR #9707, opened 2022-08-26
Improve processing arrow metrics

ktf:


ktf:
Tested also on hyperloop.

PR #9708, opened 2022-08-27
DPL: improve metric processing

ktf:
This should drastically reduce the memory needed for metrics, in
particular those which are associated to the GUI, by keeping
only a fraction of the history for things where we do not care /
it does not make sense to show it.

Moreover it speeds up metric processing by removing obsolete code and avoiding a memory allocation when parsing.


ktf:
Not yet tested...

ktf:
Works on hyperloop.

ktf:
Tested on hyperloop.

jgrosseo:
Seems not to have a sizeable effect in the nightly Hyperloop tests:
image

PR #9710, opened 2022-08-28
DPL Analysis: do not use string lookup for finding columns

aalkin:
@ktf this removes string column lookup reducing runtime costs of table construction significantly, but probably increases compile time. Could you take a look?


ktf:

PR #9713, opened 2022-08-29
DPL Analysis: invoke endOfStream method for services, if present

ktf:
If a service defines an "endOfStream" method, either taking a
EndOfStreamContext or void, we will invoke it right before stopping
a run.

PR #9715, opened 2022-08-29
Report CCDBManagerInstance statistice via endOfStream method

shahor02:


CCDB/src/BasicCCDBManager.cxx
```diff
@@ -43,5 +43,30 @@ std::pair<uint64_t, uint64_t> CCDBManagerInstance::getRunDuration(int runnumber)
...
+    }
+    res += fmt::format(" for {} objects", nfailObj);
+  }
+  res += fmt::format("), instance: {}", mCCDBAccessor.getUniqueAgentID());
+  return res;
+}
+
+void CCDBManagerInstance::endOfStream()
+{
+  LOG(info) << getSummaryString();
```
jgrosseo:
```suggestion
  LOG(info) << "CCDB summary: " << getSummaryString();
```

jgrosseo: In this way I can grep it easier... (I don't think `LOG(info)` adds a prefix, or does it?)

shahor02: OK, accepted, the `LOG(info) ` will add only the name of the parent process.

CCDB/src/BasicCCDBManager.cxx
```diff
@@ -43,5 +43,30 @@ std::pair<uint64_t, uint64_t> CCDBManagerInstance::getRunDuration(int runnumber)
...
+    }
+    res += fmt::format(" for {} objects", nfailObj);
+  }
+  res += fmt::format("), instance: {}", mCCDBAccessor.getUniqueAgentID());
+  return res;
+}
+
+void CCDBManagerInstance::endOfStream()
+{
+  LOG(info) << "CCDB summary: " << getSummaryString();
```
sawenzel: will this function be called from somewhere automatically?

sawenzel: I would maybe be more specific and say "CCDBManager summary" instead of CCDB summary (which may give a false impression).

shahor02: @sawenzel yes, from the analyses: https://github.com/AliceO2Group/AliceO2/pull/9713.
OK for the log.

shahor02:
tests were passed, merging after cosmetic change in the log message.

jgrosseo:
@ktf @shahor02 Thanks! This works very nicely and is displayed on our Hyperloop dev instance already:
https://alihyperloop.cern.ch/dev/wagon-test/5362 (within CERN only) --> test results
We will roll it out in the next days.

PR #9717, opened 2022-08-29
adjust TOF calib collector (speed up)

noferini:
@chiarazampolli 
This apply a fix (validated on LHC22m run) to speed up the calib collector.
I removed emplace in random position and I added a sorting on the channel number when finalizing.
This is very much faster, I just needed to replace CalibInfoShort with CalibInfo since the channel number is needed for the sorting.


chiarazampolli:
Thanks @noferini !

shahor02 (approved):

PR #9718, opened 2022-08-29
[FDD] (Good amplitude only for reconstruction (use of IsInAdcGate bit))

arvindkhuntia:
@mslupeck  Please have a look on this.

### DIFF
--- a/Detectors/FIT/FDD/reconstruction/src/Reconstructor.cxx
+++ b/Detectors/FIT/FDD/reconstruction/src/Reconstructor.cxx
@@ -22,25 +22,28 @@
 using namespace o2::fdd;
 
 //_____________________________________________________________________
-o2::fdd::RecPoint Reconstructor::process(o2::fdd::Digit const& digitBC,
-                                         gsl::span<const o2::fdd::ChannelData> inChData,
-                                         gsl::span<o2::fdd::ChannelDataFloat> outChData)
+void Reconstructor::process(o2::fdd::Digit const& digitBC, gsl::span<const o2::fdd::ChannelData> inChData,
+                            std::vector<o2::fdd::RecPoint>& RecPoints, std::vector<o2::fdd::ChannelDataFloat>& outChData)
+// gsl::span<o2::fdd::ChannelDataFloat> outChData)
 {
   // Compute charge weighted average time
   Double_t timeFDA = 0, timeFDC = 0;
   Double_t weightFDA = 0.0, weightFDC = 0.0;
   Int_t nInTimeA = 0, nInTimeC = 0;
-
+  int firstEntry = outChData.size();
+  int nStored = 0;
   int nch = inChData.size();
   for (int ich = 0; ich < nch; ich++) {
-    outChData[ich] = o2::fdd::ChannelDataFloat{inChData[ich].mPMNumber,
-                                               (inChData[ich].mTime) * timePerTDC,
-                                               (double)inChData[ich].mChargeADC,
-                                               inChData[ich].mFEEBits};
-
-    Float_t adc = outChData[ich].mChargeADC;
-    Float_t time = outChData[ich].mTime;
     bool inTime = inChData[ich].getFlag(ChannelData::EEventDataBit::kIsEventInTVDC);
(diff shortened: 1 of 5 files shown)


Detectors/FIT/FDD/reconstruction/src/Reconstructor.cxx
```diff
@@ -22,25 +22,28 @@
...
   int nch = inChData.size();
   for (int ich = 0; ich < nch; ich++) {
-    outChData[ich] = o2::fdd::ChannelDataFloat{inChData[ich].mPMNumber,
-                                               (inChData[ich].mTime) * timePerTDC,
-                                               (double)inChData[ich].mChargeADC,
-                                               inChData[ich].mFEEBits};
-
-    Float_t adc = outChData[ich].mChargeADC;
-    Float_t time = outChData[ich].mTime;
     bool inTime = inChData[ich].getFlag(ChannelData::EEventDataBit::kIsEventInTVDC);
```
mslupeck, line 37: Is this used anywhere?

Detectors/FIT/FDD/reconstruction/src/test-raw-conversion.cxx
```diff
@@ -35,8 +35,8 @@ int main(int argc, char* argv[])
   }
   std::vector<o2::fdd::Digit> digitsBC, *fddBCDataPtr = &digitsBC;
   std::vector<o2::fdd::ChannelData> digitsCh, *fddChDataPtr = &digitsCh;
-  tree->SetBranchAddress("FDDDIGITSBC", &fddBCDataPtr);
-  tree->SetBranchAddress("FDDDIGITSCH", &fddChDataPtr);
+  tree->SetBranchAddress("FDDDigit", &fddBCDataPtr);
+  tree->SetBranchAddress("FDDDigitCh", &fddChDataPtr);
```
mslupeck: I'm not sure if you can change these just like that. These are the names of branches set in your digitizer (or any piece of code that produces digits, also raw converter). If you want to change it, you need to do it in all places

Detectors/FIT/FDD/reconstruction/src/Reconstructor.cxx
```diff
@@ -22,25 +22,28 @@
...
+    bool inAdcGate = inChData[ich].getFlag(ChannelData::EEventDataBit::kIsCFDinADCgate);
+    if (inAdcGate) {
```
mslupeck, line 39: This condition is OK, but it's quite relaxed. So the question is: do you want to get more precise time sacrificing efficiency, or vice-versa? This requires a lot of testing, starting with simulations, but ultimately with the experimental data. Have you done any tests?

Detectors/FIT/FDD/reconstruction/src/Reconstructor.cxx
arvindkhuntia: Hi @mslupeck, inTime is used to calculate the avg time of A and c side. At least one of the channels should have TVDC bit true.

Detectors/FIT/FDD/reconstruction/src/test-raw-conversion.cxx
arvindkhuntia: This is not necessary at the moment. I will revert these changes. Thank you.

Detectors/FIT/FDD/reconstruction/src/Reconstructor.cxx
arvindkhuntia: Hi @mslupeck,  inAdcGate bit is only used to store charge and not in the calculation of the avg time. It will not affect the estimation of avg time which is propagated to AO2D.

mslupeck:
Other than the active comments, it looks good. But you should definitely test it before merging

arvindkhuntia:
Hi @mslupeck, thank you.  I have checked with run3 pp data and after implementing this adcInGate, I don't see the negative charges and the charge distribution looks ok. The time distribution has no impact on these implementations.

arvindkhuntia:
Hi @shahor02 Could you please merge this branch.

Thank you.

shahor02 (approved):

PR #9720, opened 2022-08-30
Treat mismatches of MC labels for TPC-TOF tracks

nburmaso:


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -698,6 +698,17 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
                                                   const o2::globaltracking::RecoContainer& data,
                                                   const std::map<std::pair<int, int>, int>& mcColToEvSrc)
 {
+  auto setLabelForSrc = [&](std::array<GID, GID::NSources>& contributorsGID, uint8_t src) {
+    auto mcTruthITS = data.getTrackMCLabel(contributorsGID[src]);
```
shahor02: please change to mcLabel, otherwise the name is misleading

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -717,20 +728,12 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
...
-              event = mcTruthTPC.getEventID();
-              particle = mcTruthTPC.getTrackID();
-              mToStore[Triplet_t(source, event, particle)] = 1;
-            }
+            setLabelForSrc(contributorsGID, GIndex::Source::ITS);
+            setLabelForSrc(contributorsGID, GIndex::Source::TPC);
+          }
+          if (contributorsGID[GIndex::Source::TOF].isIndexSet() && contributorsGID[GIndex::Source::TPC].isIndexSet()) {
+            setLabelForSrc(contributorsGID, GIndex::Source::TOF);
+            setLabelForSrc(contributorsGID, GIndex::Source::TPC);
```
shahor02: For the ITS-TPC-TOF tracks this will set the TPC info twice. What is wrong with calling setLabelForSrc for ITS, TPC, TOF after the separate check for each of these detectors?

shahor02 (commented):
Hi @nburmaso 

Thanks for spotting the inconsistency, please see some comments below. But I think it is better to fix the problem on the source level.
@noferini, is there a  reason to depart from the common convention for the barrel global track labels:
1) the trackID is always that of TPC.
2) during the matching to particular detector the label is flagged as `isFake` if the label of this detector contribution is compatible with tpc label.
As it is now, the labels from the TOF matching are referring to TOF MC track id (with arbitrary only 1 st contributor selected): https://github.com/AliceO2Group/AliceO2/blob/706d2f15d3f5dc58954c18a3371b3108c2ef78b0/Detectors/GlobalTracking/src/MatchTOF.cxx#L1128
I think it should be 
```
mOutTOFLabels[trkTypeSplitted].push_back(labelTrack).setFakeFlag(fake);
```

nburmaso:
Hi @shahor02. Thank you for your comments. I tried improving the code a bit, please take a look.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
shahor02: And this?

nburmaso: Sorry, I didn't commit the code changes for this. Should be ok now

noferini:
> Hi @nburmaso
> 
> Thanks for spotting the inconsistency, please see some comments below. But I think it is better to fix the problem on the source level. @noferini, is there a reason to depart from the common convention for the barrel global track labels:
> 
>     1. the trackID is always that of TPC.
> 
>     2. during the matching to particular detector the label is flagged as `isFake` if the label of this detector contribution is compatible with tpc label.
>        As it is now, the labels from the TOF matching are referring to TOF MC track id (with arbitrary only 1 st contributor selected): https://github.com/AliceO2Group/AliceO2/blob/706d2f15d3f5dc58954c18a3371b3108c2ef78b0/Detectors/GlobalTracking/src/MatchTOF.cxx#L1128
>        
>        I think it should be
> 
> 
> ```
> mOutTOFLabels[trkTypeSplitted].push_back(labelTrack).setFakeFlag(fake);
> ```

Hi @shahor02,
I am not familiar with this piece of code (I don't know who set it) but I cannot immagine any reason to have a different convention (adding also @njacazio, it may be he knows).
About the label I agree that is wrong to consider only the first one also in the matching (MatchTOF) I check all the possible labels before to set a matching to fake. Can we simply inherit it from matching infos?

shahor02:
Hi @noferini 

I am referring to line https://github.com/AliceO2Group/AliceO2/blob/f757cc5defab87b181d9afab15ad7f8fd64acf7d/Detectors/GlobalTracking/src/MatchTOF.cxx#L1128 and 
https://github.com/AliceO2Group/AliceO2/blob/f757cc5defab87b181d9afab15ad7f8fd64acf7d/Detectors/GlobalTracking/src/MatchTOF.cxx#L1237: the MC particle track ID should correspond to TPC (the only contributor which is common for all global tracks and which defines the kinematics) while the correctness of the match to other detector defines the return value of the `label.isFake()` flag.

noferini:
> Hi @noferini
> 
> I am referring to line
> 
> https://github.com/AliceO2Group/AliceO2/blob/f757cc5defab87b181d9afab15ad7f8fd64acf7d/Detectors/GlobalTracking/src/MatchTOF.cxx#L1128
> and
> 
> https://github.com/AliceO2Group/AliceO2/blob/f757cc5defab87b181d9afab15ad7f8fd64acf7d/Detectors/GlobalTracking/src/MatchTOF.cxx#L1237
> : the MC particle track ID should correspond to TPC (the only contributor which is common for all global tracks and which defines the kinematics) while the correctness of the match to other detector defines the return value of the `label.isFake()` flag.

Hi @shahor02 ,
now I got your point. I thought you were referring to the AODproducer part (sorry to many hours spent in meeting today).
I can replace the trackID with the one of the TPC, I used the one of the matched combined tracks for internal tests in the past but I don't need to trace it now.
About the setting of all TOF labels I can also do that. I have only a doubt. Shall I set all the labels with the same fake flags or shall I flag as not fake only the one which really matches TPC label pushing label inside the loop performing the check? It depends on how it will be used in the next steps.

shahor02:
Hi @noferini 

For what concerns the global track MC label, the loop https://github.com/AliceO2Group/AliceO2/blob/f757cc5defab87b181d9afab15ad7f8fd64acf7d/Detectors/GlobalTracking/src/MatchTOF.cxx#L1232-L1236 is fully enough as  it covers all cluster contributors. Why do you need to other label(s)?  If needed, the cluster label can be accessed from the cluster's container.

noferini:
> Hi @noferini
> 
> For what concerns the global track MC label, the loop
> 
> https://github.com/AliceO2Group/AliceO2/blob/f757cc5defab87b181d9afab15ad7f8fd64acf7d/Detectors/GlobalTracking/src/MatchTOF.cxx#L1232-L1236
> is fully enough as it covers all cluster contributors. Why do you need to other label(s)? If needed, the cluster label can be accessed from the cluster's container.

I was still confused with the many TOF labels. Please discard my last comment

nburmaso:
Hi @shahor02, @noferini

If I understand correctly, code changes in this PR are independent from the upcoming changes in the TOF matching code. Can we merge this PR for now? Then fix for TOF labels in `MatchTOF.cxx` could be also done in a separate PR.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -716,21 +727,14 @@ void AODProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReader&
...
-              mToStore[Triplet_t(source, event, particle)] = 1;
-            }
+          if (contributorsGID[GIndex::Source::TPC].isIndexSet()) {
+            markMCTrackForSrc(contributorsGID, GIndex::Source::TPC);
+          }
+          if (contributorsGID[GIndex::Source::ITS].isIndexSet()) {
+            markMCTrackForSrc(contributorsGID, GIndex::Source::ITS);
+          }
+          if (contributorsGID[GIndex::Source::TOF].isIndexSet()) {
+            markMCTrackForSrc(contributorsGID, GIndex::Source::TOF);
```
shahor02: @nburmaso sorry, I got some doubts about this call: the `contributorsGID[GIndex::Source::TOF]` refers to TOF cluster, then in the `markMCTrackForSrc` you are using `data.getTrackMCLabel` with this reference. But the TOF cluster's MC labels are not coming from the span with 1 label per cluster as  https://github.com/AliceO2Group/AliceO2/blob/9e0753e352b7b8279d80a1efae3af30493d20edd/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h#L438 expects but are stored in https://github.com/AliceO2Group/AliceO2/blob/9e0753e352b7b8279d80a1efae3af30493d20edd/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h#L305 with possibly multiple labels per cluster. That's why the loop https://github.com/AliceO2Group/AliceO2/blob/f757cc5defab87b181d9afab15ad7f8fd64acf7d/Detectors/GlobalTracking/src/MatchTOF.cxx#L1232-L1236 is used to decide if the match is fake or not.
I think the right thing would be to have a fix by Francesco and then to use isFake result of the TOF match.
Otherwise, one has to use https://github.com/AliceO2Group/AliceO2/blob/9e0753e352b7b8279d80a1efae3af30493d20edd/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h#L607 to access TOF clusters and implement the loop over cluster contributors.

noferini:
I opened a PR in parallel
https://github.com/AliceO2Group/AliceO2/pull/9730

shahor02 (dismissed):

shahor02:
Hi @nburmaso 
Unfortunately the aod-producer fails in the sim_challenge, the error looks genuine. Note the the RecoContainer usually does not load TOF clusters MC labels: if TOF-matched tracks are involved, it automatically requests also the clusters but w/o MC info: https://github.com/AliceO2Group/AliceO2/blob/53008b365ae36315919810cadd0cc256e9bccedc/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L489-L492. Is it possible that in the sample you were testing there were no TOF tracks?
You can try to change this line https://github.com/AliceO2Group/AliceO2/blob/53008b365ae36315919810cadd0cc256e9bccedc/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L491 to `requestTOFClusters(useMC);` and test with `sim_challenge.sh`

shahor02 (approved):

PR #9722, opened 2022-08-30
Trd qc update

f3sch:
I removed some silly variables and added other useful ones.


Detectors/TRD/qc/include/TRDQC/Tracking.h
```diff
@@ -44,12 +44,12 @@ namespace trd
 struct TrackQC {
   int type;          ///< 0 TPC-TRD track; 1 ITS-TPC-TRD track
   int nTracklets;    ///< number of attached TRD tracklets
-  int nLayers;       //< Number of Layers of a Track in which the track extrapolation was in geometrical acceptance of the TRD
```
martenole, line 47: why do you say this is silly? might be interesting to compare the number of tracklets which could be attached to the number of layers were we expect the track to be found, no?

f3sch: Silly in the sense that I added in #9121 but did not use it.
Will drop this change.

martenole: still, I think it could be useful in the future so I would just leave it in. the rest of the changes is fine with me of course

martenole (approved):

PR #9728, opened 2022-08-30
[MFT] solve problem of chip id from RO to geo when making sym names

arakotoz:
Modification in the MFT standalone alignment code:
- fix wrong symbolic name of the sensor in the geometry by using a dedicated method to retrieve the chip id Geo from a given chip id RO
- for each sensor, the symbolic name is used to correctly store the computed align param of this sensor in the geometry


rpezzi (approved):

arakotoz:
Hi,
is the failure in test_Framework_test_ParallelProducer reported by build/O2/o2 related to the PR ?
Cheers

PR #9731, opened 2022-08-31
DPL: move away from CallbackRegistry

ktf:
The recursive template is simply slow to compile, accounting alone for 10% of the
compilation time in analysis. While this version does not allow
to invoke callbacks from the id anymore, we never used such a feature,
so we can simply drop it.


ktf:
@costing @Barthelemy the multinode issue is back. Did anything change?

```
27/39 Test #33: multinode_test ..........................***Failed   35.75 sec
97% tests passed, 1 tests failed out of 39
```

Barthelemy:
@knopers8 see previous comment from Giulio. @ktf is it something you saw in many places or just here ?

knopers8:
> Did anything change?

Not in QC.

PR #9733, opened 2022-08-31
[EMCAL-786, EMCAL-565, EMCAL-566] Different spec names for Bad Channel Calib and Time Calib

mfasDa:
Define differnet names for the EMCALChannelCalibratorSpec
depending on what calibration is running.
Needed in order to run the two calibrations
in one workflow at the same time.


mfasDa:
CI errors are probably unrelated. @davidrohr do you think it can be merged.

shahor02 (approved):

PR #9734, opened 2022-08-31
Update ITS DCS string parser

ezradlesser:
Two bug fixes for the ITS DCS string parser:
- work with both DOS and Unix file endings;
- work correctly with ampersand delineation.

Code has already been tested -- tagging @shahor02 for merging.

### DIFF
--- a/Detectors/ITSMFT/ITS/workflow/src/DCSParserSpec.cxx
+++ b/Detectors/ITSMFT/ITS/workflow/src/DCSParserSpec.cxx
@@ -56,7 +56,7 @@ void ITSDCSParser::run(ProcessingContext& pc)
   // Check for DOS vs. Unix line ending
   std::string line_ending = "\n";
   size_t newline_pos = inStringConv.find(line_ending);
-  if (newline_pos && inStringConv[newline_pos-1] == '\r') {
+  if (newline_pos && inStringConv[newline_pos - 1] == '\r') {
     line_ending = "\r\n";
   }
 
--- a/Detectors/ITSMFT/ITS/workflow/src/DCSParserSpec.cxx
+++ b/Detectors/ITSMFT/ITS/workflow/src/DCSParserSpec.cxx
@@ -175,8 +175,8 @@ void ITSDCSParser::updateMemoryFromInputString(const std::string& inString)
 //////////////////////////////////////////////////////////////////////////////
 // Update pos and npos and check for validity. Return false if there is error
 bool ITSDCSParser::updatePosition(unsigned int& pos, unsigned int& npos,
-  const std::string& delimiter, const char* word,
-  const std::string& inString, bool ignoreNpos/*=false*/)
+                                  const std::string& delimiter, const char* word,
+                                  const std::string& inString, bool ignoreNpos /*=false*/)
 {
   pos += npos + delimiter.length() + std::string(word).length();
   if (!ignoreNpos) {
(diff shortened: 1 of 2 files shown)


Detectors/ITSMFT/ITS/workflow/src/DCSParserSpec.cxx
```diff
@@ -52,7 +52,16 @@ void ITSDCSParser::run(ProcessingContext& pc)
   const auto inString = pc.inputs().get<gsl::span<char>>("inString");
   std::string inStringConv;
   std::copy(inString.begin(), inString.end(), std::back_inserter(inStringConv));
-  for (const std::string& line : this->vectorizeStringList(inStringConv, "\n")) {
+
+  // Check for DOS vs. Unix line ending
+  std::string line_ending = "\n";
+  size_t newline_pos = inStringConv.find(line_ending);
+  if (newline_pos && inStringConv[newline_pos - 1] == '\r') {
```
shahor02: should  be:
```
if (newline_pos != std::string::npos && inStringConv[newline_pos - 1] == '\r') {
```
since this is a value returned on failure.
I see that in the code you have many `npos = string.find...` used w/o checking the return value of npos. This is not safe, could you add everywhere verification `npos !=  std::string::npos` ?

shahor02 (changes requested):
Hi @ezradlesser 
please see the comment below.
Cheers,
 Ruben

Detectors/ITSMFT/ITS/workflow/src/DCSParserSpec.cxx
ezradlesser: Hi @shahor02 ,

Thanks for checking this in such detail!

Line 59 is meant to check that `newline_pos != 0` (the case of inStringConv having 0 length, or for some reason there is a newline at the beginning of the file) -- in this case trying to access `[newline_pos - 1]` will not work since `newline_pos - 1` will be a negative number. Is your suggested edit doing the same thing?

The `string.find...` functions are meant to find characters inside of the string so that I can separate long strings into substrings. From what I see on a quick Google search, `std::string::npos` seems to have to do something with the end of the "long" string itself -- maybe I am misunderstanding this or your recommended implementation?

Thanks again,
Ezra

ezradlesser: Oh I think I understand what you mean with the second point -- to check that the "finding" works, and does not just return `std::string::npos`. But if the `find()` does not work then I think the code will not work in any case?

shahor02: `string.find` returns `std::string::npos` (==max_ulong) if the pattern is not found. This is what you will get if inStringConv does not contain '\n' and which current code you will have out-of-bound access. 
If needed, you can leave also check for 0. But don't take for granted that the string.find returns a value within the string length, even if this is meant by your file format, the code should not crash if e.g. the file is truncated for some reason.

ezradlesser: Ah, understood -- thanks. Yes, I will add this additional check to prevent out-of-bounds access.

shahor02: If an anomaly in the data is detected, you should just produce an error message but avoid exceptions or segfaults.

ezradlesser: Hi @shahor02 , I just pushed a new commit adding some checks. Please take a look and let me know if you see any issue. Thanks again!

shahor02 (approved):
looks fine, thanks!

shahor02 (changes requested):
@ezradlesser the error is genuine, could you always use size_t for string::find results ?

shahor02:
@ezradlesser I've allowed myself to fix the code to make sure it passes CI by the next tag.

ezradlesser:
Hi @shahor02 , sorry I missed this until now -- thanks a lot for recognizing and fixing the issue!

PR #9737, opened 2022-09-01
Put 10M basket size

jgrosseo:
This reduces the AOD size dramatically because the compression for large tables goes up a factor 10. 

Ideally this basket size is tuned based on the known size of the table * column type at writing.


jgrosseo:
Superseded by https://github.com/AliceO2Group/AliceO2/pull/9745

PR #9744, opened 2022-09-02
EMCAL-612

mfasDa:
Publisher can publish cells on different subspec then 0.
Default subspec 0, settable via ConfigParamSpec. Needed
when running processor consuming cells and publishing
new cell containers (i.e. in order to apply cell-level
calibration)


mfasDa:
@shahor02 As soon as the recalibration workflow will be in place it will most likely also run on MC. Therefore the cell reader must be able to publish as well on subspec 1. For simplicity I added the option
also to the writer, mainly for testing purpose.

shahor02 (approved):
@mfasDa With https://github.com/AliceO2Group/AliceO2/pull/9740/commits/bb753cc665bd89ffb8d35740ef498c9e129f5aa2 of https://github.com/AliceO2Group/AliceO2/pull/9740 `--emcal-decoded-subspec <S=0>` option passed to ctf-reader will impose subspec 0 on decoded triggers/cells

PR #9746, opened 2022-09-02
Store start time of slot in ms when it is created.

chiarazampolli:
Needed when you have time slots that span over more than one run: if you
don't cache this value, you risk that the OrbitReset used to
calculate the start time is different from the correct one (namely,
if the last run that is processed has a different OrbitReset from the
run when the slot started).

@noferini

### DIFF
--- a/Detectors/Calibration/include/DetectorsCalibration/TimeSlot.h
+++ b/Detectors/Calibration/include/DetectorsCalibration/TimeSlot.h
@@ -34,10 +34,7 @@ class TimeSlot
 {
  public:
   TimeSlot() = default;
-  TimeSlot(TFType tfS, TFType tfE) : mTFStart(tfS), mTFEnd(tfE)
-  {
-    mTFStartMS = getStartTimeMS();
-  }
+  TimeSlot(TFType tfS, TFType tfE) : mTFStart(tfS), mTFEnd(tfE) {}
   TimeSlot(const TimeSlot& src) : mTFStart(src.mTFStart), mTFEnd(src.mTFEnd), mContainer(std::make_unique<Container>(*src.getContainer())) {}
   TimeSlot& operator=(const TimeSlot& src)
   {
@@ -65,6 +62,7 @@ class TimeSlot
 
   void setTFStart(TFType v) { mTFStart = v; }
   void setTFEnd(TFType v) { mTFEnd = v; }
+  void setStaticStartTimeMS(long t) { mTFStartMS = t; }
   void setRunStartOrbit(long t) { mRunStartOrbit = t; }
   auto getRunStartOrbit() const { return mRunStartOrbit; }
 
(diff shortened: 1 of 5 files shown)


shahor02 (commented):
Hi, how it is supposed to work over multiple runs? The TF counter is reset at every run, so it may work only if working with a single slot of 0 - infinity TF validity.

chiarazampolli:
Hi @shahor02 , 
This is for manual calibrations only. 
Chiara

Detectors/Calibration/include/DetectorsCalibration/TimeSlot.h
```diff
@@ -34,13 +34,17 @@ class TimeSlot
 {
  public:
   TimeSlot() = default;
-  TimeSlot(TFType tfS, TFType tfE) : mTFStart(tfS), mTFEnd(tfE) {}
+  TimeSlot(TFType tfS, TFType tfE) : mTFStart(tfS), mTFEnd(tfE)
+  {
+    mTFStartMS = getStartTimeMS();
```
shahor02: This cannot work: getStartTimeMS() uses mRunStartOrbit which is set by the TimeSlotCalibration once the slot is created. Perhaps you can modify setRunStartOrbit method to also assign mTFStartMS whenever the mRunStartOrbit is set.

shahor02 (changes requested):

Detectors/Calibration/include/DetectorsCalibration/TimeSlot.h
chiarazampolli: Hi @shahor02 ,
Yes, I noticed. When I opened the PR, we were still testing, I forgot to add [WIP]. I will do so now. 
Thanks, 
Chiara

chiarazampolli:
Hello @shahor02 , 
The code is now fixed. Please, check the change in TimeSlotCalibration, 
```
-      TFType ntf = mSlotLengthInSeconds / (long(o2::base::GRPGeomHelper::getNHBFPerTF()) * o2::constants::lhc::LHCOrbitMUS / 1000000);
+      TFType ntf = mSlotLengthInSeconds / (o2::base::GRPGeomHelper::getNHBFPerTF() * o2::constants::lhc::LHCOrbitMUS * 1e-6);
```
we think it is better (not sure why you had the casting to `long`, and the multiplication should be faster).

shahor02 (approved):

chiarazampolli:
Hello @TimoWilken , @ktf , 
The issue is not related to this PR, unless I am overlooking something. Can you check?
Chiara

ktf:
@Barthelemy this is a CCDB issue, right?

```
23/42 Test #25: testCheckWorkflow .......................***Failed   11.78 sec
```

chiarazampolli:
Hello, 
The errors are not related to this PR. Can it be merged?
Chiara

PR #9747, opened 2022-09-02
ITS - change format for dcs object + fixed few bugs

iravasen:
- changed format for DCS object as in noise calibrator
- fixed bug for `mPixelStat`
- change subspec of second object shipped by aggregator to ccdb-populator when running dig scan

@shahor02 can we please have this merged by Sunday so to have this PR at P2 next week? Thanks a lot in advance.


shahor02:
will merge as soon as CIs are passed

PR #9748, opened 2022-09-02
Create HBFUtilsInitializer after fully defining workflow

shahor02:
The o2::raw::HBFUtilsInitializer hbfIni(configcontext, specs) needs to scan all devices in the workflow
therefore it should be invoked in the defineDataProcessing only once the workflow is fully defined.


martenole (approved):
Thanks a lot Ruben, this indeed fixes the issue I had

PR #9749, opened 2022-09-02
DPL: speed up metric processing

ktf:


ktf:
This is only relevant  for the case when the DPL driver is actually active.

ktf:
Tested on hyperloop.

PR #9750, opened 2022-09-04
Fix MeanVertexObject and use in the PVertexer

shahor02:


shahor02:
Merging after the fix and local test.

PR #9753, opened 2022-09-05
DPL Analysis: prevent nullptr access on empty filter groups

aalkin:
@jgrosseo this fixes the crash on empty table when using `SmallGroups`


jgrosseo:
Thanks!

PR #9754, opened 2022-09-05
WP Check the contribution from EMC active links

mpoghos:
Check the contribution of all active links during the building of a given event defined by {orbit,BC}.
The list of (expected) active links to be taken from CCDB or from a config file: variable  bitSetActiveLinks.
Currently the value is hard codded and all 40 links are assumed to be active


mfasDa:
Not to be merged unless it gets a ticket ID in the EMCAL JIRA where the problem fixed by this commit is documented!

PR #9755, opened 2022-09-05
[EMCAL-830] Recalibrator workflow for EMCAL cells

mfasDa:
Applying cell level calibrations
- Bad channel masking
- Cell time
- Absolute energy
Calibrations can be enabled and disabled one-by-one
(default: all enabled). Input and output specs mut be
specified by the user. Calibration objects are taken from
teh CCDB.

### DIFF
--- a/Detectors/EMCAL/workflow/CMakeLists.txt
+++ b/Detectors/EMCAL/workflow/CMakeLists.txt
@@ -10,73 +10,81 @@
 # or submit itself to any jurisdiction.
 
 o2_add_library(EMCALWorkflow
-               SOURCES src/EMCALDigitWriterSpec.cxx
-                       src/EMCALDigitizerSpec.cxx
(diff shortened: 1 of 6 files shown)


mfasDa:
Needs #9740

mfasDa:
For the CCDB access I use now a similar approach as what is implemented in the GRPRequest. The main motivation for this is to allow for static CCDB objects, mainly for testing purpose, to bypass the CCDB query and load files statically in the init function, otherwise the ccdb request is delegated to finalizeCCDB.

mfasDa:
Case sensitivity fixed

shahor02:
@mfasDa the fillCI failure is genuine

mfasDa:
Missing bracket issue fixed, should be the last remaining issue from the codechecker.

shahor02 (approved):

PR #9757, opened 2022-09-05
TPC SCD calib: store correct angle in residuals

martenole:
- Add TRD chi2 to the track data
- Geometry is not needed by the static map creator and can thus be removed from the workflow

### DIFF
--- a/Detectors/TPC/calibration/SpacePoints/src/TrackResiduals.cxx
+++ b/Detectors/TPC/calibration/SpacePoints/src/TrackResiduals.cxx
@@ -393,7 +393,8 @@ void TrackResiduals::doOutlierRejection(bool writeToFile)
         continue;
       }
       ++nClValidated;
-      mUnbinnedResiduals.emplace_back(clRes[clIdx].dy, clRes[clIdx].dz, clRes[clIdx].tgl, clRes[clIdx].y, clRes[clIdx].z, iRow, clRes[clIdx].sec);
+      float tgPhi = std::tan(std::asin(clRes[clIdx].phi));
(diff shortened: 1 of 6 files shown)


Detectors/TPC/calibration/SpacePoints/src/TrackResiduals.cxx
```diff
@@ -393,7 +393,8 @@ void TrackResiduals::doOutlierRejection(bool writeToFile)
         continue;
       }
       ++nClValidated;
-      mUnbinnedResiduals.emplace_back(clRes[clIdx].dy, clRes[clIdx].dz, clRes[clIdx].tgl, clRes[clIdx].y, clRes[clIdx].z, iRow, clRes[clIdx].sec);
+      float tgPhi = std::tan(std::asin(clRes[clIdx].phi));
```
shahor02: it is probably negligible overhead compared to the bulk of the processing, but the asin is quite an expensive function, the calculation `tg = snp/sqrt((1.-snp)*(1+snp))` should be faster. BTW, the comment of `TPCClusterResiduals.phi` says it is a phi angle of the track. The name of variable is already confusing enough, perhaps one should change the comment to "sin of the phi angle between the padrow and the track"

martenole: I agree. And since the calibration has to be redone in any case, I also changed the variable name to make it clearer

martenole:
Hi @shahor02 would it be ok to merge this as is (the commits squashed)?

shahor02 (approved):

PR #9758, opened 2022-09-05
Rebase to dev, Mon 05.09.2022

chiarazampolli:


davidrohr:
@chiarazampolli : The idea was not to do pull requests for stable-sync and stable-async branches.
Instead, we can merge all required changes locally, then push the branch, and then tag.

chiarazampolli:
> @chiarazampolli : The idea was not to do pull requests for stable-sync and stable-async branches. Instead, we can merge all required changes locally, then push the branch, and then tag.

Ciao @davidrohr , I am not sure I understand what you mean with "push the branch", sorry. I was just trying to rebase to dev, as I understood that we should do it from time to time.

davidrohr:
Yes, but we don't need to open a PR for it. We just do the rebase locally, and then we can update the branch via `git push origin-with-write-permissions stable-async:stable-async`.
If we use the github workflow, there will either be an additional merge commit, or everything is rebased and we loose the history, or it must be squashed. This is documented here: http://alisw.github.io/infrastructure-o2-branches

PR #9759, opened 2022-09-05
DPL: speedup arrow metrics processing

ktf:


ktf:
Tested also on hyperloop, merging.

PR #9760, opened 2022-09-05
More metric processing improvements

ktf:


ktf:
Tested on hyperloop. Merging.

PR #9761, opened 2022-09-05
[EMCAL-565]: Change offline calibrator to add option to store calibration objects locally.

hjbossi:
Previously objects were written to the test ccdb. Added the option here to write locally, which allows for later upload to the production ccdb and keeps the test ccdb clear for tests at P2 for example.

### DIFF
--- a/Detectors/EMCAL/calibration/run/runCalibOffline.cxx
+++ b/Detectors/EMCAL/calibration/run/runCalibOffline.cxx
@@ -44,6 +44,7 @@ int main(int argc, char** argv)
   std::string ccdbServerPath;
   bool doBadChannelCalib;
   bool debugMode = false;
+  bool doLocal = false;
   std::string nameCalibInputHist; // hCellIdVsTimeAbove300 for time, hCellIdVsEnergy for bad channel
   std::string namePathStoreLocal; // name for path + histogram to store the calibration locally in root TH1 format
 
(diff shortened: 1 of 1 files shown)


Detectors/EMCAL/calibration/run/runCalibOffline.cxx
```diff
@@ -187,22 +193,29 @@ int main(int argc, char** argv)
...
-    std::map<std::string, std::string> metadata;
-    calibdb.storeTimeCalibParam(&TCparams, metadata, rangestart, rangeend);
+    if (!doLocal) {
+      // store parameters in ccdb via emcal calibdb
+      std::map<std::string, std::string> metadata;
+      calibdb.storeTimeCalibParam(&TCparams, metadata, rangestart, rangeend);
+    }
 
-    if (namePathStoreLocal.find(".root") != std::string::npos) {
+    if (namePathStoreLocal.find(".root") != std::string::npos && doLocal) {
```
mfasDa: For the time calib this is actually no the `local` mode. Recall: For the local mode we export the calibration object to a file which can then be uploaded manually to the CCDB. What was implemented for the time calib however is to export a ROOT hist to file (for visualization), which is not the calibration object. Like this we cannot post the time calib object to the CCDB. I suggest to implement like for the bad channel calib, because once the object is stream to a ROOT file it is trivial to extract from this the ROOT histogram for visualization.

mfasDa (changes requested):
For the bad channel calibration it looks fine but for the time calibration you don't store the CCDB object but the histogram for visualization. I suggest to store in both cases the calibration object, because extracting the visualization object from it is trivial.

Detectors/EMCAL/calibration/run/runCalibOffline.cxx
hjbossi: Thanks for the comments, I have fixed this and re-pushed!

mfasDa (approved):

shahor02 (approved):

PR #9767, opened 2022-09-06
new features for TOF calib scheme (slewing and track quality flags)

noferini:


noferini:
build/O2/o2-cs8 is failing but not because of this PR.
If it is fine with you I would propose to merge it.
Changes were tested with a macro extracting and uploading the timeslewing calibration from the output of the accumulator workflow.
The macro has to be converted to an O2 class (after some tuning) so it is not yet in this PR (it will come soon)

shahor02 (approved):

PR #9768, opened 2022-09-06
DPL: reset the number of end of stream received before quitting

ktf:


ktf:
This is clearly wrong, passes tests and ends up with a different problem in staging.

PR #9769, opened 2022-09-06
[EMCAL-757, QC-851] Fix type for subspec in EMCAL cell reader

mfasDa:
Mark default argument as unsigned int since unsinged int
is used as data type.


shahor02 (approved):
trivial, merging

PR #9771, opened 2022-09-07
TRD: AODProducer: Fix trdPattern

f3sch:
In noticed sometime that 7 tracklets were counted as contributing to the track, which is physically impossible since we only have 6 layers.

I used this function to convert from the pattern to nTracklets:

  // Count the number of contributor tracklets from the trd pattern.
  inline uint8_t nTrackletsFromPattern(uint8_t pattern) {
    return std::bitset<8>{pattern}.count();
    }

Looking at AODProducerWorkflowDPL:getTRDPattern(), we are looping over the layers _7_ times instead of 6 (kNLayers=6 and break condition is >=0).
In this function we call track.getTrackletIndex(6) on our first iteration, which indexes into mAttachedTracklets[kNLayers]. This is UB!

This patch fixes it by decrementing the loop iterations by one.

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -1114,7 +1114,7 @@ void AODProducerWorkflowDPL::countTPCClusters(const o2::tpc::TrackTPC& track,
 uint8_t AODProducerWorkflowDPL::getTRDPattern(const o2::trd::TrackTRD& track)
 {
   uint8_t pattern = 0;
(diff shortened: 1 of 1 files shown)


martenole (approved):
Thanks a lot for spotting this @f3sch 
For me it is fine to merge this once the checks are through. Currently this could indeed lead to a segfault during AOD creation.

f3sch:
@martenole I do not see how this could cause test_Framework_test_SingleDataSource to fail?

sawenzel (approved):

martenole:
All good, sometimes tests run into a timeout and cause the checks to fail. Here it was clearly unrelated. Usually the checks are after some time automatically restarted.

PR #9772, opened 2022-09-07
Don't use latest libuv API

jgrosseo:


jgrosseo:
as discussed with @ktf

ktf (approved):

PR #9775, opened 2022-09-07
Drew final pr

ktf:


ktf:
Everything now is in a plugin, so that we do not risk anything by including it. The code will be enabled only if you run with `DPL_LOAD_SERVICES=O2FrameworkGUISupport:SpyGUIPlugin`.

ktf:
Not actually enabled until you load the associated plugin.

martenole:
@ktf can it be that this breaks the CI? Although I don't understand why here the fullCI was successful. In #9875 the fullCI fails with the error
```
/sw/SOURCES/O2/9875-slc8_x86-64/0/Framework/GUISupport/src/SpyService.cxx:31:35: error: redundant void argument list in function definition [modernize-redundant-void-arg]
## End sw/BUILD/o2checkcode-latest/log
```

PR #9776, opened 2022-09-07
TRD raw reader update

martenole:
Won't pass the CI until the CCDB object `TRD/Config/LinkToHCIDMapping` is available. But when it is we can enable link to HCID mapping from CCDB.
In addition some warnings have been replaced by sending error counters to QC. And currently the CRU sends quite often bogus link data which would always lead to warnings in the log file. Therefore the warnings are disabled for now and should be re-enabled when the CRU is updated. We still send errors counters for this to QC.
Finally, some more compiler warnings are addressed.

### DIFF
--- a/Detectors/TRD/simulation/src/trap2raw.cxx
+++ b/Detectors/TRD/simulation/src/trap2raw.cxx
@@ -40,7 +40,7 @@ namespace bpo = boost::program_options;
 
 void trap2raw(const std::string& inpDigitsName, const std::string& inpTrackletsName,
               const std::string& outDir, int verbosity, std::string filePerLink,
-              uint32_t rdhV = 6, bool noEmptyHBF = false, int tracklethcheader = 0, int superPageSizeInB = 1024 * 1024);
+              uint32_t rdhV = 6, bool noEmptyHBF = false, int tracklethcheader = 0, int superPageSizeInB = 1024 * 1024, long startTime = 1547590800000);
 
 int main(int argc, char** argv)
 {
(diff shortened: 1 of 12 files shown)


Detectors/TRD/simulation/src/trap2raw.cxx
```diff
@@ -89,12 +89,14 @@ int main(int argc, char** argv)
   }
   o2::conf::ConfigurableParam::updateFromString(vm["configKeyValues"].as<std::string>());
 
-  trap2raw(vm["input-file-digits"].as<std::string>(), vm["input-file-tracklets"].as<std::string>(), vm["output-dir"].as<std::string>(), vm["verbosity"].as<int>(), vm["file-per"].as<std::string>(), vm["rdh-version"].as<uint32_t>(), vm["no-empty-hbf"].as<bool>(), vm["tracklethcheader"].as<int>());
+  const auto& hbfu = o2::raw::HBFUtils::Instance(); // we need the creation time for the link mapping from CCDB
```
martenole, line 92: Hi @shahor02 do you know how I can correctly get the `startTime` from `o2simdigitizerworkflow_configuration.ini`? I tried here simply with `o2::raw::HBFUtils::Instance().startTime` but that gives me 1547590800000 while in the ini file I have `startTime=1662576045021`.
A few lines above I thought the HBFUtils parameters should be updated from the file
```
  std::string confDig = vm["hbfutils-config"].as<std::string>();
  if (!confDig.empty() && confDig != "none") {
    o2::conf::ConfigurableParam::updateFromFile(confDig, "HBFUtils");
  }
```

martenole: Ah, sorry for the noise I think everything is fine. I was using the default parameter by accident

f3sch (approved):
LGTM! :thumbsup:

bazinski (approved):

PR #9778, opened 2022-09-07
DPL: avoid race condition when resetting output proxy

ktf:
Due to the consumeWhenAllOrdered policy of the output proxy, if the startTime on the input proxy and the counter in the policy go out of sync, it will result in no messages being processed.

To obviate this, we reset the dplCounter whenever we have a new run and we reset the policy counter when we see a 0 for the startTime.


ktf:
Error unrelated. For sure fixes things.

PR #9780, opened 2022-09-08
MCH: some more debug options for reco-workflow

aphecetche:
add a couple of debug options to stop the reco before clustering or tracking stage


Detectors/MUON/MCH/Workflow/src/reco-workflow.cxx
```diff
@@ -67,8 +69,8 @@ WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
   auto digits = configcontext.options().get<bool>("digits");
   auto triggered = configcontext.options().get<bool>("triggered");
   auto useMC = !configcontext.options().get<bool>("disable-mc");
-
-  o2::conf::ConfigurableParam::updateFromString(configcontext.options().get<std::string>("configKeyValues"));
```
pillot, line 71: Is this removed on purpose?

aphecetche: hum, no, I don't think so :wink:

pillot (approved):

PR #9783, opened 2022-09-09
Add the LOG(warn) for HMPID header reported errors

fapfap69:


Detectors/HMPID/reconstruction/src/HmpidDecoder2.cxx
```diff
@@ -1098,6 +1098,7 @@ void HmpidDecoder2::dumpHmpidError(int ErrorField)
       std::cout << "HMPID Decoder2 : [ERROR] "
                 << "HMPID Error field = " << ErrorField << " : " << printbuf << std::endl;
     }
+    LOG(warn) << "HMPID Header Error Field =" << ErrorField << " [" << printbuf << "]";
```
sawenzel, line 1101: this is the same line as 2 lines above. I think it makes sense to simplify this. What do you think?

fapfap69: Is the same, but the above are directed to the standard out and are linked to the verbosity mode. The added line will be logged every time an error occurred. I would reserve the 'verbosity' case to future improvements and expansion of the message/status.

sawenzel: LOG(warn) also writes to std::out in a file. I don't see why you would want to have it twice in case of verbosity switched on. But for me this is not a blocker.

PR #9784, opened 2022-09-09
IDCs: use default parameter for first TF

matthias-kleiner:
Using the first TF as start TF otherwise the first 100 TFs are skipped and not processed


matthias-kleiner:
Can this be merged?

PR #9785, opened 2022-09-09
Moving calibrations to production CCDB when in PHYSICS data taking

chiarazampolli:
@martenole , @davidrohr , @shahor02

### DIFF
--- a/prodtests/full-system-test/aggregator-workflow.sh
+++ b/prodtests/full-system-test/aggregator-workflow.sh
@@ -17,7 +17,11 @@ if [[ -z ${CCDB_POPULATOR_UPLOAD_PATH+x} ]]; then
   if [[ $RUNTYPE == "SYNTHETIC" ]]; then
     CCDB_POPULATOR_UPLOAD_PATH="http://ccdb-test.cern.ch:8080"
   elif [[ $RUNTYPE == "PHYSICS" ]]; then
-    CCDB_POPULATOR_UPLOAD_PATH="http://ccdb-test.cern.ch:8080"
+    if [[ $$EPNSYNCMODE == 1 ]]; then
+      CCDB_POPULATOR_UPLOAD_PATH="http://localhost:8084"
+    else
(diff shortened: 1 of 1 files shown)


martenole (dismissed):
I think this is OK or should it actually be the `http://o2-ccdb.internal` instead?

prodtests/full-system-test/aggregator-workflow.sh
```diff
@@ -17,7 +17,7 @@ if [[ -z ${CCDB_POPULATOR_UPLOAD_PATH+x} ]]; then
   if [[ $RUNTYPE == "SYNTHETIC" ]]; then
     CCDB_POPULATOR_UPLOAD_PATH="http://ccdb-test.cern.ch:8080"
   elif [[ $RUNTYPE == "PHYSICS" ]]; then
-    CCDB_POPULATOR_UPLOAD_PATH="http://ccdb-test.cern.ch:8080"
+    CCDB_POPULATOR_UPLOAD_PATH="http://alice-ccdb.cern.ch"
```
shahor02: Should not it be `http://o2-ccdb.internal` and only in `EPNSYNCMODE==1` ?

chiarazampolli: Or even `http://localhost:8084`, right @costing ?
And yes, only in `EPNSYNCMODE==1`

costing: If this runs in the Online environment, then yes, `http://localhost:8084` should be the target. Or `http://o2-ccdb.internal` (where the redirects will go to anyway). Clearly **not** `alice-ccdb.cern.ch` because that endpoint will require a valid X.509 certificate to upload objects to.

chiarazampolli: Changed, please x-check.

shahor02 (approved):

davidrohr (approved):

PR #9786, opened 2022-09-09
Remove option to ignore bogus TRD headers

martenole:
The current synthetic data samples contain more bogus data so it is not worth trying to circumvent all this in the raw reader. Instead, we need to create new synthetic data samples in order to exercise the TRD processing without beam.


bazinski (approved):

martenole:
Failures unrelated, merging

PR #9788, opened 2022-09-10
MUON: dcs processor fix (do not add several times the very same datapoint)

aphecetche:
and change slightly the output format of the helper `dcs-ccdb` program

### DIFF
--- a/Detectors/MUON/Common/src/dcs-processor-workflow.cxx
+++ b/Detectors/MUON/Common/src/dcs-processor-workflow.cxx
@@ -266,10 +266,16 @@ void processDataPoints(o2::framework::ProcessingContext& pc,
   }
   auto dps = pc.inputs().get<gsl::span<o2::dcs::DataPointCompositeObject>>("input");
   for (auto dp : dps) {
-    // FIXME: check we're not adding twice the same dp (i.e. check timestamp ?)
     for (auto i = 0; i < NOBJECTS; i++) {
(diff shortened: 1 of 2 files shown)


Detectors/MUON/Common/src/dcs-processor-workflow.cxx
```diff
@@ -266,10 +266,16 @@ void processDataPoints(o2::framework::ProcessingContext& pc,
...
     for (auto i = 0; i < NOBJECTS; i++) {
       if (std::find(aliases[i].begin(), aliases[i].end(), dp.id.get_alias()) != aliases[i].end()) {
-        dataPoints[i][dp.id].emplace_back(dp.data);
+        auto& v = dataPoints[i][dp.id];
+        bool shouldAdd{true};
+        if (v.size() > 0 && v.back() == dp.data) {
+          shouldAdd = false;
+        }
+        if (shouldAdd) {
+          dataPoints[i][dp.id].emplace_back(dp.data);
```
pillot: you could have used v.emplace_back(dp.data);

pillot (dismissed):

Detectors/MUON/Common/src/dcs-processor-workflow.cxx
aphecetche: indeed. now fixed.

pillot (approved):

PR #9790, opened 2022-09-12
[EMCAL-565]: Integrate scaling in the bad channel calibration

hjbossi:
Scaling of the energy vs. cellID histogram is now integrated into the CalibExtractor. Reads scale factors from the CCDB. 

- Additionally change the writing of root histograms in the offline calibrator to have a more realistic name.


mfasDa (approved):
Looks good!

mfasDa:
@shahor02 @chiarazampolli  Can you please merge this commit? This is needed both for offline calibration and online calibration at P2. It is a correction we apply for effects coming from the material in the TRD support structures, without we would flag most of the cells behind the TRD sector/stack boundaries as bad.

shahor02 (approved):

PR #9799, opened 2022-09-12
Fix indentation, and typo

chiarazampolli:
@davidrohr , @martenole , @shahor02 : I fixed the indentation, but there was also a typo:
```
FLP_ADDRESS**+**="tcp://alicr1-flp-ib${flp}:${TPC_IDC_FLP_PORT}"
```
should have been:
```
FLP_ADDRESS="tcp://alicr1-flp-ib${flp}:${TPC_IDC_FLP_PORT}"
```

### DIFF
--- a/prodtests/full-system-test/aggregator-workflow.sh
+++ b/prodtests/full-system-test/aggregator-workflow.sh
@@ -110,7 +110,7 @@ if workflow_has_parameter CALIB_PROXIES; then
   # expand FLPs; TPC uses from 001 to 145, but 145 is reserved for SAC
   for flp in $(seq -f "%03g" 1 144)
   do
-    FLP_ADDRESS+="tcp://alicr1-flp-ib${flp}:${TPC_IDC_FLP_PORT}"
+    FLP_ADDRESS="tcp://alicr1-flp-ib${flp}:${TPC_IDC_FLP_PORT}"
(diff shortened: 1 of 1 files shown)


martenole:
Thanks, but why is then `FLP_ADDRESS` defined inside the for loop? Then it will always be `tcp://alicr1-flp-ib144:${TPC_IDC_FLP_PORT}`
If its always flp144 I think that should be defined outside the loop

chiarazampolli:
> Thanks, but why is then `FLP_ADDRESS` defined inside the for loop? Then it will always be `tcp://alicr1-flp-ib144:${TPC_IDC_FLP_PORT}` If its always flp144 I think that should be defined outside the loop

I don't think so. We need to have one channel per FLP, so we will append to `CHANNELS_LIST` every time a different FLP. Maybe I am blind... I will try it in a shell.

martenole:
Ah sorry, never mind. It's only used inside the loop.. I think its fine as it is now

chiarazampolli:
Error is not related to the PR, can it be merged? @shahor02

davidrohr:
merging, but this seems to fix only the typo, not the indentation.

chiarazampolli:
Weird, i pushed yesterday the fix, after the first one didn’t work. I will check.

PR #9800, opened 2022-09-12
Kr calibration - normalized gain maps

jcrkovsk:
Class to prepare upload of normalized gain per chamber, where inter/extrapolated gain is negative.

Code has been tested and should be good to go. Some minor details will be worked out later.

### DIFF
--- a/Detectors/TRD/base/include/TRDBase/PadParameters.h
+++ b/Detectors/TRD/base/include/TRDBase/PadParameters.h
@@ -25,6 +25,7 @@
 
 #include "DataFormatsTRD/Constants.h"
 #include "DataFormatsTRD/HelperMethods.h"
+#include "Framework/Logger.h"
 
 namespace o2
 {
--- a/Detectors/TRD/calibration/include/TRDCalibration/PadCalibCCDBBuilder.h
+++ b/Detectors/TRD/calibration/include/TRDCalibration/PadCalibCCDBBuilder.h
@@ -1,0 +1,76 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file PadCalibCCDBBuilder.h
+/// \brief Krypton calibration
+/// \author Jana Crkovska
+
+#ifndef O2_TRD_KRCALIBRATION_H
+#define O2_TRD_KRCALIBRATION_H
+
+#include "DataFormatsTRD/Constants.h"
+#include "Rtypes.h"
+#include "TH2.h"
+#include "TTree.h"
+#include <cstdlib>
+#include <numeric>
+#include <vector>
+
+namespace o2
+{
+namespace trd
+{
+
+class PadCalibCCDBBuilder
+{
+ public:
+  PadCalibCCDBBuilder() = default;
+  ~PadCalibCCDBBuilder() = default;
+
+  void checkIfIsolatedHotPadCandidate(TH2F* hDet, std::vector<int> coordinates, float upperLimit = 1.5, int areaContainedWithin = 4);
+  void checkIfSmallerCloserToCenter(TH2F* hDet, std::vector<int> coordinates, float allowedDifference);
+  std::vector<int> compareGain(TH2F* hDet, int column, int row, int shiftcolumn, int shiftrow, float allowedDifference);
+  float computeDetectorAverage(TH2F* hDet);
+  float computeDistance(std::vector<float> pad1, std::vector<float> pad2);
+  TH2F* createNormalizedMap(TH2F* hDet, TString sNewName = "");
+  void fillInTheGap(TH2F* hDet, int column, int row, float newGain);
+  TH2F* fillTheMap(TH2F* hDet, TString sNewName = "", int nbuffer = 3);
+  std::vector<std::vector<int>> findEmpty(TH2F* hDetectorMap);
+  std::vector<std::vector<int>> findInhomogeneities(TH2F* hDet, float allowedDifference);
+  float getAverageFromNeighbors(TH2F* hDet, int column, int row, int nbuffer = 3);
+  TH2F* getDetectorMap(TTree* tree, int nDet, float mingain = 0, float maxgain = 10'000, TString sDetName = "");
+  bool isHotAreaIsolated(TH2F* hDet, int column, int row, int matrixSize = 1);
+  int isolatedHotPadsContainmentSize(TH2F* hDet, int column, int row);
+  void populateEmptyNormalizedMap(TH2F* hDet, float valueToSet = -1);
+  void removeEdges(TH2F* hDet, int nsize = 2);
+  void removeExtremePads(TH2F* hDet, float upperLimit = 2., float lowerLimit = 0.5);
+  void replacePadCloserToCenter(TH2F* hDet, int column, int row);
+  void replaceIsolatedHotPads(TH2F* hDet, int column, int row, int nsize);
+  void setTreeBranches(TTree* tree);
+  void smoothenTheDetector(TH2F* hDet, float allowedDifference = 1000);
+  TH2F* transformMapIntoAbsoluteValues(TH2F* hDet, TString sName = "");
+
+ private:
+  float mDet;
+  float mCol;
+  float mRow;
+  float mAdc;
+  float mChi;
+  float mSgm;
+  float mAmp;
+
+  ClassDefNV(PadCalibCCDBBuilder, 1);
+};
+
+} // namespace trd
+} // namespace o2
+
+#endif // O2_TRD_KRCALIBRATION_H
--- a/Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
+++ b/Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
@@ -1,0 +1,552 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file PadCalibCCDBBuilder.cxx
+/// \brief Krypton calibration - class to smoothen and inter/extrapolate gain in chambers and calculate normalized ADC gain per chamber
+/// \author Jana Crkovska
+
+#include "TRDCalibration/PadCalibCCDBBuilder.h"
+#include <iostream>
+#include <string>
+#include <map>
+#include <memory>
+#include "TFile.h"
+#include "TH2.h"
+#include "TMath.h"
+#include "TString.h"
+#include "TTree.h"
+
+using namespace o2::trd::constants;
+
+namespace o2::trd
+{
+
+void PadCalibCCDBBuilder::checkIfIsolatedHotPadCandidate(TH2F* hDet, std::vector<int> coordinates, float upperLimit, int areaContainedWithin)
+{
+  auto numberOfCoordinates = coordinates.size();
+  if (numberOfCoordinates != 4) {
+    std::cerr << "Invalid size of the coordinates vector!" << std::endl;
+    return;
+  }
+  float averageGain = computeDetectorAverage(hDet);
+
+  if (hDet->GetBinContent(coordinates[0], coordinates[1]) > upperLimit * averageGain) {
+    int sizeOfHotArea = isolatedHotPadsContainmentSize(hDet, coordinates[0], coordinates[1]);
+    if (sizeOfHotArea > 0 && sizeOfHotArea < areaContainedWithin) {
+      replaceIsolatedHotPads(hDet, coordinates[0], coordinates[1], sizeOfHotArea);
+    }
+  } else if (hDet->GetBinContent(coordinates[2], coordinates[3]) > upperLimit * averageGain) {
+    int sizeOfHotArea = isolatedHotPadsContainmentSize(hDet, coordinates[2], coordinates[3]);
+    if (sizeOfHotArea > 0 && sizeOfHotArea < areaContainedWithin) {
+      replaceIsolatedHotPads(hDet, coordinates[2], coordinates[3], sizeOfHotArea);
+    }
+  } else if (hDet->GetBinContent(coordinates[0], coordinates[1]) == 0 && hDet->GetBinContent(coordinates[2], coordinates[3]) != 0) {
+    int sizeOfHotArea = isolatedHotPadsContainmentSize(hDet, coordinates[2], coordinates[3]);
+    if (sizeOfHotArea > 0 && sizeOfHotArea < areaContainedWithin) {
+      replaceIsolatedHotPads(hDet, coordinates[2], coordinates[3], sizeOfHotArea);
+    }
+  }
+}
+
+void PadCalibCCDBBuilder::checkIfSmallerCloserToCenter(TH2F* hDet, std::vector<int> coordinates, float allowedDifference)
+{
+  if (hDet->GetBinContent(coordinates[0], coordinates[1]) == 0 || hDet->GetBinContent(coordinates[2], coordinates[3]) == 0)
+    return;
+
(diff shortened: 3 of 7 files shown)


Detectors/TRD/calibration/include/TRDCalibration/PadCalibCCDBBuilder.h
```diff
@@ -0,0 +1,80 @@
...
+namespace o2
+{
+namespace trd
+{
+
+class PadCalibCCDBBuilder
+{
+ public:
+  PadCalibCCDBBuilder(){};
+  ~PadCalibCCDBBuilder(){};
```
martenole: default constructor and destructor should be marked with `= default` instead of  an empty block `{}`

Detectors/TRD/calibration/include/TRDCalibration/PadCalibCCDBBuilder.h
```diff
@@ -0,0 +1,80 @@
...
+
+  void CheckIfIsolatedHotPadCandidate(TH2F* hDet, std::vector<int> coordinates, float upperLimit = 1.5, int areaContainedWithin = 4);
```
martenole: all class methods should start with a lower case letter

Detectors/TRD/calibration/include/TRDCalibration/PadCalibCCDBBuilder.h
```diff
@@ -0,0 +1,80 @@
...
+  void RemoveEdges(TH2F* hDet, int nsize = 2);
+  void RemoveExtremePads(TH2F* hDet, float upperLimit = 2., float lowerLimit = 0.5);
+  void ReplacePadCloserToCenter(TH2F* hDet, int column, int row);
+  void ReplaceIsolatedHotPads(TH2F* hDet, int column, int row, int nsize);
+  void SetTreeBranches(TTree* tree);
+  void SmoothenTheDetector(TH2F* hDet, float allowedDifference = 1000);
+  TH2F* TransformMapIntoAbsoluteValues(TH2F* hDet, TString sName = "");
+
+ private:
+  float det;
```
martenole: all class member variables should start with `m`, so `mDet` in this case for example

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,580 @@
...
+#include "TString.h"
+#include "TTree.h"
+
+// int NCOLUMN = 144;
+// int NROWC0 = 12;
+// int NROWC1 = 16;
+
+using namespace o2::trd::constants;
+
+ClassImp(o2::trd::PadCalibCCDBBuilder);
```
martenole: this `ClassImp` line is not needed

martenole (changes requested):
Hi Jana, just some very general comments without going through it in more detail. Please take a look at the coding guidelines https://github.com/AliceO2Group/CodingGuidelines

Detectors/TRD/calibration/include/TRDCalibration/PadCalibCCDBBuilder.h
jcrkovsk: ok

Detectors/TRD/calibration/include/TRDCalibration/PadCalibCCDBBuilder.h
jcrkovsk: ok

Detectors/TRD/calibration/include/TRDCalibration/PadCalibCCDBBuilder.h
jcrkovsk: ok

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
jcrkovsk: removed

Detectors/TRD/base/include/TRDBase/PadParameters.h
```diff
@@ -25,6 +25,7 @@
 
 #include "DataFormatsTRD/Constants.h"
 #include "DataFormatsTRD/HelperMethods.h"
+#include "Framework/Logger.h"
```
tdietel, line 28: If logging is not used in the header file, it should only be included in the .cxx file.

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+
+namespace o2::trd
+{
+
+void PadCalibCCDBBuilder::checkIfIsolatedHotPadCandidate(TH2F* hDet, std::vector<int> coordinates, float upperLimit, int areaContainedWithin)
+{
+  if ((int)coordinates.size() != 4) {
+    std::cerr << "Invalid size of the coordinates vector!" << std::endl;
+    return;
+  }
```
tdietel: Using an array or a struct would remove the need to test the size of the coordinates vector.
The struct would also clarify the meaning of the 4 elements of the vector. The code looks like there are two pairs of pad row/column coordinates, but I'm not sure what the difference is.

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+      int colNeighbor = (column - offset) + icol;
+      if (colNeighbor < 0 || colNeighbor >= hDet->GetNbinsX())
+        continue; // avoids under and overflow
+      float tempGain = TMath::Abs(hDet->GetBinContent(hDet->GetXaxis()->FindBin(colNeighbor), hDet->GetYaxis()->FindBin(rowNeighbor)));
+      if (tempGain <= 0)
+        continue; // exclude empty/negative bins
+      gainNeighbor.push_back(tempGain);
+    }
+  }
+  int numberOfNeighbors = (int)gainNeighbor.size();
```
tdietel: It looks like you have a few type casts from sizes to signed integers. Try to avoid them by using size_t (or auto) as the type.

tdietel (changes requested):
The code looks good, with a few suggestions for further cleanup.
But especially the header file for the PadCalibCCDBBuilder should get some more comments. For example, the vector<int> coordinates are not self-explanatory - it would be better to replace them with a struct.

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+{                       // clones the filled 2d map
+                        // computes an average
+                        // replaces each bin conetnt with ratio of gain over the average
+  if (sNewName == "") { // create a default name
+    sNewName = hDet->GetName();
+    sNewName += "_normalized";
+  }
+  TH2F* hDetNormalized = (TH2F*)hDet->Clone(sNewName.Data());
+
+  if (!hDet || hDet->GetEntries() == 0)
```
f3sch: Checking if hDet is nullptr is unnecessay and should be moved up before calling ->Clone().
Since calling Clone() on nullptr would throw.

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+                        // will fill any map with at least 1 hit!
+                        // the map is cloned to limit propagation of neighboring bin content
+  if (sNewName == "") { // create default name
+    sNewName = hDet->GetName();
+    sNewName += "_filled";
+  }
+  TH2F* hDetFilled = (TH2F*)hDet->Clone(sNewName);
+
+  if (!hDet || hDet->GetEntries() == 0)
+    return hDetFilled;
```
f3sch: Same here

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+
+  TH2F* hDetTemp = (TH2F*)hDet->Clone("hDetTemp"); // use as intermediate det in the filling process
```
f3sch, line 211: I think you are leaking memory here since ->Clone() calls new and never gets deleted.
Maybe use auto hDetTemp = std::make_unique<TH2F>(hDet->Clone("hDetTemp")); ?

Detectors/TRD/macros/CreateCCDBLocalGainFactor.C
```diff
@@ -0,0 +1,43 @@
...
+#include <string>
+
+using namespace o2::trd;
+
+void CreateCCDBLocalGainFactor(TString sOpenFile)
+{
+  PadCalibCCDBBuilder* krCalibToCCDB = new PadCalibCCDBBuilder();
+
+  TFile* file = TFile::Open(sOpenFile);
+  TTree* tree = (TTree*)file->Get("nt_Krypton");
```
f3sch, line 29: The file will stay open after the function call and the tree will persist as well.
Either drop both at the end or use:
```
std::unique_ptr<TFile> file( TFile::Open(sOpenFile) );
auto tree = file->Get<TTree>("nt_Krypton");
```

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+
+  return average;
+}
+
+float PadCalibCCDBBuilder::computeDistance(std::vector<float> pad1, std::vector<float> pad2)
+{
+  float distance = -1.;
+
+  if (pad1.size() != pad2.size()) {
+    std::cerr << "Something fishy with the pad coordinates!" << std::endl;
```
f3sch, line 153: Is using this for logging okay in calibration code or is LOG(error) the way to go here? @tdietel @martenole

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+  }
+
+  return coordinates;
+}
+
+float PadCalibCCDBBuilder::computeDetectorAverage(TH2F* hDet)
+{ // computes an average over filled cells
+  // cells are accessed through their values, not bin coordinates
+  // the average os computed over absolute values as inter/extrapolated cells
+  // have negatiove values
```
f3sch, line 128: I don't think that doxygen pick this up as a valid format to build docs? @tdietel @martenole 
I think you should move these above the functions.

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+  // check for edges - this operation goes before filling
+  // edges along X:
+  int xFirst = hDet->FindFirstBinAbove();
+  int xLast = hDet->FindLastBinAbove(); // stop the process at penultimate vs ultimate row/column
+  // // edges along Y:
+  int yFirst = hDet->FindFirstBinAbove(0, 2);
+  int yLast = hDet->FindLastBinAbove(0, 2); // stop the process at penultimate vs ultimate row/column
+
+  // int nbins_x = hDet->GetNbinsX();
+  // int nbins_y = hDet->GetNbinsY();
```
f3sch, line 296: delete dead code

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+  for (int irow = 0; irow < hDet->GetNbinsY(); irow++) {
+
+    int thisRow = hDet->GetXaxis()->FindBin(irow);
+    // if( thisRow < yFirst || thisRow > yLast ) continue;
```
f3sch, line 294: delete

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+  // and averages the gain in its (up to) 8 neighbours
+  float average = 0.;
+  std::vector<float> gainNeighbor;
+  int offset = nbuffer / 2;
+
+  for (int irow = 0; irow < nbuffer; irow++) {
+
+    int rowNeighbor = (row - offset) + irow;
+    if (rowNeighbor < 0 || rowNeighbor >= hDet->GetNbinsY())
+      continue; // avoids under and overflow
```
f3sch: Going from your comment here.
If rowNeighbor == 0 this would be the underflow in hDet for any bin and
rowNeigbor >= hDet->GetNBinsY would not only exclude the overflow bin but also the the before it.
Would also apply to colNeighbor
See: https://root.cern/doc/master/classTH1.html#binning
Note: Double check this, might very well be wrong.

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
```diff
@@ -0,0 +1,551 @@
...
+  return isIsolated;
+}
+
+int PadCalibCCDBBuilder::isolatedHotPadsContainmentSize(TH2F* hDet, int column, int row)
+{
+  int nsize = 0;
+  bool isContained = kFALSE;
+  while (!isContained) {
+    nsize++;
+    isContained = isHotAreaIsolated(hDet, column, row, nsize);
```
f3sch, line 453: Personal preference but maybe eliminate isContained? (Pretty sure the compiler drops this anyway and turns it into this)
```
do{
if (++nsize == 4)
return -1;
} while(! isHotAreaIsolated(hDet, column, row, nsize));

f3sch (changes requested):

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
jcrkovsk: In this very case, I use row/colNeighbor as hDet->GetXaxis()->FindBin(colNeighbor), so if colNeighbor is 0 it will point to bin 1. Similarly for colNeighbor = GetNbinsX-1, it will call the last bin, i.e., GetNbinsX.
So it should work alright but is a part of another larger issue where I use inconsistently row and column to either refer to bin coordinates or pad coordinates (i.e. as they come from TRD).
As also pointed in another comment, this needs fixing so that the use is non-ambiguous and consistent in all methods.

f3sch: I am not convinced this is true.
If you call ->FindBin(0)  this returns the underflow bin and ->FindBin(GetNbinsX-1) returns the second to last bin.
Maybe I am misunderstanding the doc for it ([https://root.cern/doc/master/classTH1.html#ae0895b66e993b9e9ec728d828b9336fe](FindBin) but I think this _may_ even have unintended side effects (e.g. extending your axis)

> This function tries to extend the axis if the given point belongs to an under-/overflow bin AND if CanExtendAllAxes() is true.

This refers to TH1::GetBin() which list the problem quite clearly I think?

Note: 
TH1::FindBin() internally calls TAxis::FindBin(), so hist->GetXaxis()->FindBin() is the same as hist->FindBin() but nicer wrapped.

jcrkovsk: No. FindBin(0) returns the bin which holds the value 0. In this case it gives you the first bin, i.e. 1. It is GetBin(0) which gets you the underflow.

bazinski:
@jcrkovsk you still need to run clang-format against this pr, and need to rebase to dev as well.

martenole:
Hi @jcrkovsk could you please solve the conflicts here and run clang-format as is described in https://github.com/AliceO2Group/CodingGuidelines and as Sean suggested?

f3sch:
You also need to rebase against dev and please squash your commits into a single one, otherwise the git log looks a bit weird with two commits titled 'clang' :)

Detectors/TRD/macros/CreateCCDBLocalGainFactor.C
jcrkovsk: deleted after the loop

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
jcrkovsk: hDetFilled is now deleted at the end

Detectors/TRD/base/include/TRDBase/PadParameters.h
jcrkovsk: Log is used in the header file.

martenole:
Hi @jcrkovsk 
the code checker is complaining about a few trailing spaces at end of lines. Could you please run `clang-format` again push the changes?
And then there are a few comments by Tom and Felix. Could you comment if they are solved?

Detectors/TRD/calibration/src/PadCalibCCDBBuilder.cxx
jcrkovsk: The two pairs give coordinates of two neighboring pads in either vertical or horizontal direction, which are compared together to see if the transition from pad to pad is smooth enough.
While a struct would probably be better, it would also mean a major rewrite, which is not a priority at the moment and will be considered in future updates.

jcrkovsk:
> 

Hi @martenole ,
I  removed all the trailing spaces. Comments have been dealt with.

martenole:
Thanks, then I'll merge this once the CI is through

PR #9802, opened 2022-09-12
ZDC FIX digitization, add noise calibration, add option to prescale reconstruction

cortesep:


shahor02:
@cortesep the error is genuine:
```
--
/sw/SOURCES/O2/9802-slc8_x86-64/0/Detectors/ZDC/calib/src/NoiseCalibData.cxx:191:19: error: statement should be inside braces [readability-braces-around-statements]
--

```

Detectors/ZDC/workflow/src/DigitRecoSpec.cxx
```diff
@@ -165,7 +174,20 @@ void DigitRecoSpec::run(ProcessingContext& pc)
...
+    double frac = std::rand() / double(RAND_MAX);
+    if (frac > mRecoFraction) {
+      toProcess = false;
+    }
+  }
+
+  if (toProcess) {
+    mWorker.process(peds, bcdata, chans);
+  }
   const std::vector<o2::zdc::RecEventAux>& recAux = mWorker.getReco();
```
shahor02: The DigitReco::mReco returned by `mWorker.getReco()` is reset in the `mWorker.process` call, so for the TFs which you don't want to process you will get the mReco of the previous TF. 
One should clear mReco in every TF.

shahor02 (changes requested):

Detectors/ZDC/workflow/src/DigitRecoSpec.cxx
cortesep: Corrected the compilation error. I overlooked it. Apologies.
Concerning this there was a protection in the loop by inserting a break if the event is stale. However the logic was indeed cumbersome and the printed info message was not clear. I changed to protect everything inside an if(toProcess). Now if the TF is skipped there is an explicit message and mReco is not used.

shahor02 (approved):

PR #9804, opened 2022-09-13
[EMCAL-830] Add minmal information messages

mfasDa:
- Loading of calibration objects in static_load
- Received and accepted cells in the recalibrator spec


mfasDa:
Trivial commit, only leaving information at low log levels which are needed in order to follow the program flow

PR #9809, opened 2022-09-13
DPL: add ability to load services from plugins

ktf:


ktf:
No op right now. Will use it soon.

PR #9811, opened 2022-09-13
[EMCAL-566] Add calib type as workflow argument

jokonig:
- calibType was used in the EMCALCalibParams. This class is however shared between the bad channel and time calibration, meaning the two calibrations cannot run in parallel
- Changed the calibType to a workflow argument to allow the parallel running of the two calibrations
- Made changes in the aggregator-worflow accordingly


jokonig:
@davidrohr could you have a quick look at the small change in the aggregator workflow? I was not able to test it locally but I only changed the workflow arguments.
If that looks okay I think it can be merged @shahor02 (The crash in the build seems unrelated I think)

davidrohr (approved):
looks ok for the aggregator workflow

PR #9812, opened 2022-09-13
ctf-writer --output-dir=/dev/null with disable CTF and metafile writing

shahor02:
@martenole tested with
```
CTF_DIR=/dev/null $O2_ROOT/prodtests/full_system_test.sh
```


martenole (approved):
Great, thanks a lot!

PR #9813, opened 2022-09-13
[EMCAL-565]: Add NHits as an additional condition for bad channel calibration.

hjbossi:
- In order to make run 3 algorithm compatible with the run 2 algorithm, add an additional check on the number of hits for the bad channel calibration (in addition to the energy per hit). 
- Create a good Nhit window by calculating the mean number of hits and specifying the sigma this should be away from the mean to be considered "good".


mfasDa (approved):
Looks good

shahor02 (approved):

PR #9814, opened 2022-09-13
Drop unneeded dependency

ktf:


ktf:
Do we really need this? It compiles on my box without.

ktf:
@sawenzel @shahor02

shahor02:
I don't know why it was added but I don't see any use of it, to me it can be removed, as well as FairMQ.

ktf:
Indeed... FairMQ can probably just be FairLogger.

PR #9819, opened 2022-09-14
[EMCAL-798] Reject calib. events in offline calib

jokonig:
- calibration events are currently mixed with physics events in the emcal offline calibrator
- implemented switch to reject these calibration events
- rejection is on by default (cases where one wants the calib events are rare)


mfasDa (approved):
Looks good!

shahor02 (approved):

PR #9822, opened 2022-09-14
TRD raw data should be one file per CRU end point

martenole:


davidrohr (dismissed):
makes sense. So we need new datasets for SYNTHETIC?

martenole:
Depends.. Do you still have the simulation files lying around? Then I could run `o2-trd-trap2raw` manually again to get the correct files. Otherwise we would need new synthetics, yes. But for the single TFs I could also produce them

martenole:
Not to be merged yet

davidrohr:
for reference, all files in `/home/drohr/raw/new*`/TRD*` have been regenerated, with 1 file per endpoint, for usage in SYNTHETIC runs.

martenole:
To be consistent with the other detectors, the option `o2-trd-trap2raw --file-per cru` produces now one file per CRU end point. This is what the readout is expecting and what is done in the FST

bazinski (approved):
I think this option change is confusing, but apparently we must change to be confusing like everyone else ...

martenole:
> I think this option change is confusing, but apparently we must change to be confusing like everyone else ...

I agree :(

davidrohr:
> > I think this option change is confusing, but apparently we must change to be confusing like everyone else ...
> 
> I agree :(

we can also change it for all detectors, or just rename the cru option to cruendpoint everywhere...

PR #9824, opened 2022-09-14
[EMCAL-548, O2-3169] Adding protection if there is no MC

hahassan7:
A protection has been added to the CellConverterSpec if the option --disable-mc was used.


sawenzel (approved):

mfasDa (approved):
Should be fine

PR #9828, opened 2022-09-14
DPL: add back Pss metrics when resources-monitoring is enabled

ktf:


jgrosseo (approved):

jgrosseo:
Thanks! This is a blocker for Hyperloop processing at present...

ktf:
@Barthelemy @knopers8 do I need to worry about the message?

ktf:
Merged. This means people will have to update to the new monitoring in order for this to work. @awegrzyn in retrospect having backward compatibility would have been better. Could you change that?

awegrzyn:
Hi, I was explicitly asked to break backwards compatibility, but indeed doing it in 2 stages would be way better. Do you still want to bring it back? Which other modules need those metric? I thought it's just Analysis? If not, I'll bring it back ASAP.

jgrosseo:
@awegrzyn I was not asked about it. Grid analysis is down at the moment...

jgrosseo:
> Merged. This means people will have to update to the new monitoring in order for this to work. @awegrzyn in retrospect having backward compatibility would have been better. Could you change that?

@ktf Can you explain what we have to do for the "new monitoring" ?

ktf:
@awegrzyn My bad, I was under the assumption all it would have broken is a missing PSS metric for one day, until I fixed the code on my side.

@jgrosseo we use the AliceO2Group/Monitoring library to get the system stats. The new version dropped PSS calculations because it's very onerous on CC7.

ktf:
This is also forcing everyone to recompile. @awegrzyn could we quickly add the old default so that I revert this?

awegrzyn:
Done, sorry for the mess, I'll keep it as it is and patch all other modules not to use "Smaps" (Readout, DD, ...).

jgrosseo:
Thanks! The daily tag is OK with @ktf 's fix in this PR. Let me know in case we should test something (@ktf has a test line for analysis on Hyperloop without having an explicit tag

PR #9834, opened 2022-09-15
FST MC->Raw files are produced by CRU end point

martenole:
When creating new synthetic data samples we noticed inconsistencies in the option naming scheme of different detectors.
By default, when creating raw data for readout at P2 we require the files per CRU end point. Thus the option `cru` used by most detectors to describe this has been changed to `cruendpoint`.
For HMPID I use `crorcendpoint` instead of `cru`. I did not use `link` as the other CROC detectors do, because that option is used by something else for HMPID.
Please let me know in case you have comments or prefer to keep the options as they are now. There was some initial discussion in https://github.com/AliceO2Group/AliceO2/pull/9822


bazinski (approved):

martenole:
If there are no comments or objections I would merge this tomorrow

rpezzi (approved):

davidrohr (approved):

martenole:
Failure unrelated and all checks were green before, merging

davidrohr:
Thx, this makes sense to me. Eventually one should also check why hmpid uses the link differently and homogenize that as well. But can be done as follow-up.

PR #9839, opened 2022-09-15
[EMCAL-835] General error handling for invalid access in CCDB objects

mfasDa:
Error souces very similar for all objects so they can be
implemented in the same file. Currently only invalid indices
are handled.


mfasDa:
Simple bug in range check, made more robust taking the range limits from the array size where possible. Now all unit tests pass.

shahor02 (approved):

PR #9840, opened 2022-09-15
DPL: improve configuation parser

ktf:
It now handles spaces via backtracking, rather than via special characters.

### DIFF
--- a/Framework/Core/src/DataProcessingDevice.cxx
+++ b/Framework/Core/src/DataProcessingDevice.cxx
@@ -355,7 +355,7 @@ void DataProcessingDevice::Init()
     } else {
       str = entry.second.get_value<std::string>();
     }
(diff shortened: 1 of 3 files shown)


ktf:
@jgrosseo this should allow you to have ; in the options.

jgrosseo:
Thanks!

ktf:
This was passing all tests before. I have also checked it on hyperloop and I have a further commit which extends the unit test to explicitly use something with a space in it. Merging.

@sawenzel @costing I see everything is now failing in CCDB. Is this a problem with the backend or with the reworked CCDB tests?

costing:
Looks like the only object in that path was created one year ago, with one year of validity, and has expired on 17 Sep 2022 18:21 CEST :
http://ccdb-test.cern.ch:8080/browse/GLO/Config/PVertexer/

I've extended it for one more year. Hope that's ok. Please let me know if the tests pass now or if I should do the same for other objects too.

Cheers,

.costin

PR #9841, opened 2022-09-15
Improve information for MID chamber efficiency calculation

dstocco:
This PR consists of two commits:
- the first one changes the information stored in the MID track that is used to identify a local board. The legacy ID is replaced by something that it is more easily usable in the Run 3 code.
- the second one declares as bad for efficiency calculations those tracks that would have matched a masked strip. This avoids a double counting of the masked channel effect, since they would also lower the efficiency otherwise.

### DIFF
--- a/DataFormats/Detectors/MUON/MID/include/DataFormatsMID/Track.h
+++ b/DataFormats/Detectors/MUON/MID/include/DataFormatsMID/Track.h
@@ -143,19 +143,26 @@ class Track
   /// Gets the word allowing to compute the chamber efficiency
   uint32_t getEfficiencyWord() const { return mEfficiencyWord; }
 
-  /// Sets the fired local board for efficiency calculation
-  /// \param locId local board ID in the range 1-234
-  void setFiredLocalBoard(int locId) { setEfficiencyWord(8, 0xFF, locId); }
+  /// Sets the fired FEE ID
+  void setFiredFEEId(int uniqueFeeId) { setEfficiencyWord(8, 0x7FFF, uniqueFeeId); }
+
+  /// Gets the fired FEE ID
+  int getFiredFEEId() const { return (mEfficiencyWord >> 8) & 0x7FFF; }
 
   /// Gets the fired local board for efficiency calculation
-  int getFiredLocalBoard() const { return (mEfficiencyWord >> 8) & 0xFF; }
+  [[deprecated]] int getFiredLocalBoard() const { return (mEfficiencyWord >> 8) & 0xFF; }
 
-  /// Sets the fired detection element ID for efficiency calculation
-  /// \param deId detection element ID in the range 1-234
-  void setFiredDeId(int deId) { setEfficiencyWord(16, 0xFF, deId); }
+  /// Gets the fired Detection Element ID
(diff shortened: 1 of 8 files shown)


DataFormats/Detectors/MUON/MID/include/DataFormatsMID/Track.h
```diff
@@ -143,19 +143,28 @@ class Track
...
-  /// \param locId local board ID in the range 1-234
-  void setFiredLocalBoard(int locId) { setEfficiencyWord(8, 0xFF, locId); }
+  /// Sets the fired FEE ID
+  void setFiredFEEId(int uniqueFeeId) { setEfficiencyWord(8, 0x7FFF, uniqueFeeId); }
+
+  /// Gets the fired FEE ID
+  int getFiredFEEId() const { return (mEfficiencyWord >> 8) & 0x7FFF; }
 
   /// Gets the fired local board for efficiency calculation
+  /// FIXME: Obsolete: remove after changing QC
```
aphecetche: Note that instead of a FIXME (that might be forgotten...) you might prefer to use https://en.cppreference.com/w/cpp/language/attributes/deprecated that will result in actual warnings (that might be ignored, I get it :wink: )

Detectors/MUON/MID/Tracking/src/HitMapBuilder.cxx
```diff
@@ -46,61 +67,89 @@ bool HitMapBuilder::crossCommonElement(const std::vector<int>& fired, const std:
...
+  bool badForEff = false;
   for (int ich = 0; ich < 4; ++ich) {
     auto icl = track.getClusterMatchedUnchecked(ich);
     if (icl >= 0) {
       auto& cl = clusters[icl];
-      firedDeIds.emplace_back(cl.deId);
-      firedRPCLines.emplace_back(detparams::getRPCLine(cl.deId));
-      firedLocIds.emplace_back(getLocId(cl.xCoor, cl.yCoor, cl.deId));
+      firedFEEIdMT11.emplace_back(getFEEIdMT11(cl.xCoor, cl.yCoor, cl.deId));
       // auto localPt = mHitFinder.getGeometryTransformer().globalToLocal(cl.deId, cl.xCoor, cl.yCoor, cl.zCoor);
```
aphecetche: this commented code should be removed, shouldn't it ?

DataFormats/Detectors/MUON/MID/include/DataFormatsMID/Track.h
dstocco: I did not know about this attribute. I'll give it a try

Detectors/MUON/MID/Tracking/src/HitMapBuilder.cxx
dstocco: Indeed. Will be removed in the next version. I will rewrite the history though so that the two commits will stay separate

dstocco:
Modification is done, but I need to modify QC accordingly and I have some issues (probably related to the change in compiler, better to start from scratch). Please do not merge before receiving confirmation on my side.

aphecetche (approved):

PR #9843, opened 2022-09-15
[EMCAL-830] Adapt CalibLoader to fully rely on CCDB-backend

mfasDa:
- Remove static loading of calibration objects from file in
  favour of condition-remap from the framework
- Remove ManagedObject as it is no longer needed due to
  ownership handling in the CCDB backend
- Add checkUpdate function triggering the request of the
  calibration objects from the ProcessingContext

### DIFF
--- a/Detectors/EMCAL/workflow/include/EMCALWorkflow/CalibLoader.h
+++ b/Detectors/EMCAL/workflow/include/EMCALWorkflow/CalibLoader.h
@@ -9,9 +9,8 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-#include <string>
-#include <string_view>
 #include <vector>
+#include <Framework/ProcessingContext.h>
 #include "Framework/ConcreteDataMatcher.h"
 #include "Framework/InputSpec.h"
 
@@ -49,15 +48,15 @@ class CalibLoader
 
   /// \brief Access to current bad channel map
   /// \return Current bad channel map (nullptr if not loaded)
-  BadChannelMap* getBadChannelMap() const { return mBadChannelMap.getObject(); }
+  BadChannelMap* getBadChannelMap() const { return mBadChannelMap; }
 
   /// \brief Access to current time calibration params
   /// \return Current time calibration params
-  TimeCalibrationParams* getTimeCalibration() const { return mTimeCalibParams.getObject(); }
+  TimeCalibrationParams* getTimeCalibration() const { return mTimeCalibParams; }
 
   /// \brief Access to current gain calibration factors
   /// \return Current gain calibration factors
-  GainCalibrationFactors* getGainCalibration() const { return mGainCalibParams.getObject(); }
+  GainCalibrationFactors* getGainCalibration() const { return mGainCalibParams; }
 
   /// \brief Check whether the bad channel map is handled
   /// \return True if the bad channel map is handled, false otherwise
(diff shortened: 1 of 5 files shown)


mfasDa:
The reason why not implementing the assignment or move operator and stay as by value is that in certain conditions the destructor would still be called leading to the object being deleting even though it is still in used. Since the object should basically only handle the owenrship status (owned in case of read from file, non-owned in case of handled via the CCDB API) the memory management is move to the std::shared_ptr and the ManagedObject is handling only the ownership status.

shahor02 (commented):
@mfasDa I don't think you can do this: the pointers provided to `finaliseCCDB` are owned by the framework, while your shared_ptr will destroy it either at the next assignment or in destructor.

mfasDa:
@shahor02 This was actually the idea behind the `ManagedObject`: The shared pointer deletes the managed object, not the CCDB object directly. The ManagedObject deletes in the destructor the captured object only if the ownership status is true, which is the case for objects loaded statically from ROOT files but not for the objects handled by the CCDB. Therefore shared pointer should not delete objects handled via the CCDB API and the code should be OK.

The problem was that owned objects were deleted in the assignment operator because the ManagedObject was copied by value. Instead a shared pointer is needed here to prevent ambiguity in case the CalibLoader would be shared between objects.

shahor02:
OK, in this case it should work, I don't see the point of the ManagedObject at all: we should anyway eliminate loading conditions from custom local files, using only DPL fetcher. It still allows to work with local files via `--remap-condition file://<dir>/<PATH>`

mfasDa:
@shahor02 The managed object would be only needed for memory management in case of loading from local file. This option should not be the default way to go but a bypass for test we are doing at the moment. I was not aware of the `--remap-condition` option. This would indeed make our code a lot easier, though I would probably keep the CalibFetcher which is adapted from the GRPHelper.

shahor02:
see https://docs.google.com/presentation/d/1WPNixo57zAn_gM6vMCIBOHs-DZqcKAkLX0CwUgO_VYg/edit#slide=id.g10e79f44b99_0_22 for the fetcher options.

mfasDa:
@shahor02 Thanks a lot for your help! I have adapted the PR accordingly to fully rely on the ccdb-backend via the condition-remap function and remove all internal static handling of the calibration object loading from file. The CalibLoader as helper for loading calibration object I still would like to keep because I believe it can be helpful.

shahor02 (dismissed):

Detectors/EMCAL/workflow/include/EMCALWorkflow/CalibLoader.h
```diff
@@ -146,55 +103,13 @@ class CalibLoader
...
-  std::string mPathGainCalib;                                        ///< Path of local file with gain calibration params (optional)
-  ManagedObject<o2::emcal::BadChannelMap> mBadChannelMap;            ///< Container of current bad channel map
-  ManagedObject<o2::emcal::TimeCalibrationParams> mTimeCalibParams;  ///< Container of current time calibration object
-  ManagedObject<o2::emcal::GainCalibrationFactors> mGainCalibParams; ///< Container of current gain calibration object
+  bool mEnableBadChannelMap;                           ///< Switch for enabling / disabling loading of the bad channel map
+  bool mEnableTimeCalib;                               ///< Switch for enabling / disabling loading of the time calibration params
+  bool mEnableGainCalib;                               ///< Switch for enabling / disabling loading of the gain calibration params
+  o2::emcal::BadChannelMap* mBadChannelMap;            ///< Container of current bad channel map
+  o2::emcal::TimeCalibrationParams* mTimeCalibParams;  ///< Container of current time calibration object
+  o2::emcal::GainCalibrationFactors* mGainCalibParams; ///< Container of current gain calibration object
```
shahor02: You better init them with nullptr, they will be undefined unless finalizeCCDB assigns them.

mfasDa: I see, sure, will be fixed immediately.

PR #9844, opened 2022-09-16
Fix: use raw string instead of escaping

shahor02:
@ktf fullCI was failing with `/sw/SOURCES/O2/9802-slc8_x86-64/0/Framework/Core/src/WorkflowHelpers.cxx:292:80: error: escaped string literal can be written as a raw string literal [modernize-raw-string-literal]`


shahor02:
trivial, merging

PR #9848, opened 2022-09-16
TPC: Several unrelated updates and fixes

wiechula:
* Simulation: load zero suppressions from a local file for testing purposes
* Reconstruction: load ion tail pad-by-pad parameters from file
* Calibration: Unify creation of CRU calibration value text files and add macros for common mode and ion tail
* Minor fixes


davidrohr:
@wiechula : I think we should clean up applying the ZS threshold. Currently the digit2raw conversion applies a ZS threshold of 2. If that is already applied in the digitization, we should no longer apply it during digit2raw conversion by default.

PR #9853, opened 2022-09-18
CPV: add persistent bad channels reading from CCDB

sevdokim:
Hello! Reading of persistent bad channels is added to CPV NoiseCalibrator. vector<int> is list of bad channels absIds is stored in CCDB with path "CPV/Config/PersistentBadChannels".

### DIFF
--- a/Detectors/CPV/calib/include/CPVCalibration/NoiseCalibrator.h
+++ b/Detectors/CPV/calib/include/CPVCalibration/NoiseCalibrator.h
@@ -48,12 +48,15 @@ class NoiseCalibrator final : public o2::calibration::TimeSlotCalibration<o2::cp
   std::vector<o2::ccdb::CcdbObjectInfo>& getCcdbInfoBadChannelMapVector() { return mCcdbInfoBadChannelMapVec; }
   const std::vector<o2::cpv::BadChannelMap>& getBadChannelMapVector() const { return mBadChannelMapVec; }
 
-  void setPedEfficiencies(std::vector<float>* pedEffs) { mPedEfficiencies.reset(pedEffs); }
-  void setDeadChannels(std::vector<int>* deadChs) { mDeadChannels.reset(deadChs); }
-  void setHighPedChannels(std::vector<int>* highPeds) { mHighPedChannels.reset(highPeds); }
-  bool isSettedPedEfficiencies() { return mPedEfficiencies.get() == nullptr ? false : true; }
-  bool isSettedDeadChannels() { return mDeadChannels.get() == nullptr ? false : true; }
-  bool isSettedHighPedChannels() { return mHighPedChannels.get() == nullptr ? false : true; }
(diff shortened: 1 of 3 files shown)


kharlov (dismissed):

Detectors/CPV/calib/include/CPVCalibration/NoiseCalibrator.h
```diff
@@ -51,9 +51,12 @@ class NoiseCalibrator final : public o2::calibration::TimeSlotCalibration<o2::cp
   void setPedEfficiencies(std::vector<float>* pedEffs) { mPedEfficiencies.reset(pedEffs); }
   void setDeadChannels(std::vector<int>* deadChs) { mDeadChannels.reset(deadChs); }
   void setHighPedChannels(std::vector<int>* highPeds) { mHighPedChannels.reset(highPeds); }
+  void setPersistentBadChannels(std::vector<int>* persBadChs) { mPersistentBadChannels.reset(persBadChs); }
```
shahor02: One cannot make unique_ptr out of pointer on CCDB object since the latter is owned by the framework.  Please use simple pointer.
The same is true for other objects

shahor02 (changes requested):
Hi @sevdokim 
Please see below

Detectors/CPV/calib/include/CPVCalibration/NoiseCalibrator.h
sevdokim: hi @shahor02 
yes, I also was thinking about this issue. However it worked fine. Thanks for pointing this out.

shahor02: It worked only because not all DPL tasks are calling destructor, also, it should fail in case of updating the object by more recent one...

sevdokim: I see. Now changed to simple pointers. But there is one thing that I do not understand a bit:
when new object is fetched and old is deleted then I need somehow propagate new one into noise calibrator. But currently I access ccdb object only once at the beginning:
```
    static bool isConfigFetched = false;
    if (!isConfigFetched) {
      pc.inputs().get<std::vector<float>*>("pedeffs");
      pc.inputs().get<std::vector<int>*>("persbadchs");
      pc.inputs().get<std::vector<int>*>("deadchs");
      pc.inputs().get<std::vector<int>*>("highpeds");
      ...
      isConfigFetched = true;
    }
```
What to do in this case? Maybe use shared_ptr in order to be sure that old object is not deleted?
Actually this objects are not expected to change during noise and physics runs. However things must be done correctly in order to avoid unexpected problems in future.

shahor02: no, shared_ptr is also deleted once only 1 instance remains. Currently the objects in the `if (!isConfigFetched) {` are not supposed to change during the run lifetime. If they do, then you should remove this if and in e.g. finaliseCCDB pass them to the calibrator.

shahor02 (approved):

PR #9860, opened 2022-09-19
[EMCAL-798] Add low and high gain histograms

jokonig:
- Low gain and high gain cell time has to be calibrated differently
- In order to estimate the difference between the two, two new histograms for low gain and high gain cells (cell time vs. cell ID) have been added to the offline calibrator


mfasDa (approved):
Looks good. @jokonig Recall to update the offline recalibration executable.

shahor02 (approved):

PR #9863, opened 2022-09-19
TRD: Calibration update and additonal macros

f3sch:
I updated the README and added two macros:
- plotVdriftExB.C plots the VDrift and ExB values in the ccdb for a given duration for each chamber
- manualCalibFit.C performs the calibration fits for VDrift and ExB manually

### DIFF
--- a/Detectors/TRD/calibration/README.md
+++ b/Detectors/TRD/calibration/README.md
@@ -23,6 +23,17 @@ If you want to run the calibration from a local file with residuals, trdangreshi
 
     o2-calibration-trd-vdrift-exb -b --enable-root-input --calib-vdexb-calibration '--tf-per-slot 1 --min-entries 50000'
 
+Additionally it is possible to perform the calibrations fit manually per chamber if you have TPC-TRD or ITS-TPC-TRD tracks, you can run:
+   
+   o2-trd-global-tracking -b --enable-trackbased-calib
+   
+This produces `trdangreshistos.root` which holds the residuals of the angles and differences.
+Then run the macro `Detectors/TRD/calibration/macros/manualCalibFit.C`.
+This produces a file of similar name with the fitted data and prints out the fit results.
+These can be compared afterwards by running:
+
+    o2-calibration-trd-vdrift-exb -b --enable-root-input
+
 ## DCS data points
 
 To process the DCS data points for the TRD the list of aliases for example "trd_gaschromatographXe" has to be available in the CCDB. 
--- a/Detectors/TRD/calibration/README.md
+++ b/Detectors/TRD/calibration/README.md
@@ -36,13 +36,6 @@ These can be compared afterwards by running:
 
 You can plot the calibration values for VDrift and ExB for a given Run by using the macro at `Detectors/TRD/cailbration/macros/plotVdriftExB.C`.
 This produces a root file of similar name which holds the time-series plots (reference the macro).
-The Run start time needs to be given in Unix Epoch in ms.
-Convert the time/date by using these commands:
-
-    # Date to seconds (multiply 1000 for ms)
-    date -d "01:22 2022-06-27" +%s
-    # Seconds to Date
-    date -d @1656285720
 
 ## DCS data points
 
--- a/Detectors/TRD/calibration/README.md
+++ b/Detectors/TRD/calibration/README.md
@@ -16,17 +16,17 @@ The histograms are send to aggregator nodes which calculate for each TRD chamber
 To run the calibration one can start the following workflow:
 
     o2-trd-global-tracking -b --disable-root-output | o2-calibration-trd-vdrift-exb -b --calib-vdexb-calibration '--tf-per-slot 5 --min-entries 50000 --max-delay 90000'
-    
-Hint: You can get information on the meaning of the parameters by running `o2-calibration-trd-vdrift-exb -b --help full`
+
+*Hint: You can get information on the meaning of the parameters by running `o2-calibration-trd-vdrift-exb -b --help full`*
 
 If you want to run the calibration from a local file with residuals, trdangreshistos.root, you can run:
 
     o2-calibration-trd-vdrift-exb -b --enable-root-input --calib-vdexb-calibration '--tf-per-slot 1 --min-entries 50000'
 
 Additionally it is possible to perform the calibrations fit manually per chamber if you have TPC-TRD or ITS-TPC-TRD tracks, you can run:
-   
-   o2-trd-global-tracking -b --enable-trackbased-calib
-   
+
+    o2-trd-global-tracking -b --enable-trackbased-calib
+
 This produces `trdangreshistos.root` which holds the residuals of the angles and differences.
 Then run the macro `Detectors/TRD/calibration/macros/manualCalibFit.C`.
 This produces a file of similar name with the fitted data and prints out the fit results.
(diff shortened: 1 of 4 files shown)


Detectors/TRD/calibration/README.md
```diff
@@ -6,19 +6,53 @@
...
+The drift velocity and ExB angle are calibrated using input from the global tracking, namely ITS-TPC-TRD and TPC-TRD tracks.
+The latter is currently disabled, but can be enabled to have more statistics easily.
+From the global tracks we obtain the association of TRD tracklets to tracks.
+In the calibration procedure a refit is performed to get TRD only tracks and the angular deviations between those TRD tracks and the associated tracklets are stored in histograms (see class TrackBasedCalib).
+This is done on the EPNs during the reconstruction.
+The histograms are send to aggregator nodes which calculate for each TRD chamber the drift velocity and ExB angle based on those histograms (see class CalibratorVdExB).
 
 To run the calibration one can start the following workflow:
+
     o2-trd-global-tracking -b --disable-root-output | o2-calibration-trd-vdrift-exb -b --calib-vdexb-calibration '--tf-per-slot 5 --min-entries 50000 --max-delay 90000'
```
martenole: Actually this is wrong I think (mistake made by me). Can you please change this line to
`o2-trd-global-tracking -b --disable-root-output --enable-trackbased-calib | o2-calibration-trd-vdrift-exb -b --calib-vdexb-calibration '--tf-per-slot 5 --min-entries 50000 --max-delay 90000'`?

martenole (dismissed):
Thanks a lot, this is quite helpful. I think it should also be possible to retrieve for a given run number the GRP object from the CCDB. But then one needs to use not only the time, since multiple runs can be ongoing in parallel. @shahor02 do you know how one can request the GRP object from a macro using the CCDB manager? Is this supported?
Then we could get the run start and stop time from that object and don't need to get it manually from bookkeeping.

shahor02:
> Thanks a lot, this is quite helpful. I think it should also be possible to retrieve for a given run number the GRP object from the CCDB. But then one needs to use not only the time, since multiple runs can be ongoing in parallel. @shahor02 do you know how one can request the GRP object from a macro using the CCDB manager? Is this supported? Then we could get the run start and stop time from that object and don't need to get it manually from bookkeeping.

Hi, one should provide the run number via metadata:
```
  auto& ccdbMgr = o2::ccdb::BasicCCDBManager::instance();
  ...
  std::map<std::string, std::string> md;
  md["runNumber"] = std::to_string(runNumber);
  const auto* grp = ccdbMgr.getSpecific<o2::parameters::GRPECSObject>("GLO/Config/GRPECS", timestamp, md);
```

Detectors/TRD/calibration/macros/CMakeLists.txt
```diff
@@ -32,8 +32,19 @@ o2_add_test_root_macro(readTRDVoltages.C
                                             O2::CCDB
                       LABELS trd COMPILE_ONLY)
 
+o2_add_test_root_macro(manualCalibFit.C
+                      PUBLIC_LINK_LIBRARIES O2::DataFormatsTRD
```
martenole, line 36: I think you need to add `O2::TRDCalibration` here as well. The CI tests are failing with `/System/Volumes/Data/build/nomad/alloc/627e7c3c-41b1-6bbb-b5b0-3fab21115acf/ci/local/o2/sw/SOURCES/O2/9863/0/Detectors/TRD/calibration/macros/manualCalibFit.C:32:10: fatal error: 'TRDCalibration/CalibratorVdExB.h' file not found`

martenole:
> Hi, one should provide the run number via metadata:
> 
> ```
>   auto& ccdbMgr = o2::ccdb::BasicCCDBManager::instance();
>   ...
>   std::map<std::string, std::string> md;
>   md["runNumber"] = std::to_string(runNumber);
>   const auto* grp = ccdbMgr.getSpecific<o2::parameters::GRPECSObject>("GLO/Config/GRPECS", timestamp, md);
> ```

Thanks a lot Ruben! @f3sch could you add this method such that one does not need to specify time stamps anymore? You can take a look at `DataFormats/Parameters/include/DataFormatsParameters/GRPECSObject.h`. With `grp->getTimeStart()` and `grp->getTimeEnd()` you can get beginning and end time of the run.

martenole:
> Hi, one should provide the run number via metadata:
> 
> ```
>   auto& ccdbMgr = o2::ccdb::BasicCCDBManager::instance();
>   ...
>   std::map<std::string, std::string> md;
>   md["runNumber"] = std::to_string(runNumber);
>   const auto* grp = ccdbMgr.getSpecific<o2::parameters::GRPECSObject>("GLO/Config/GRPECS", timestamp, md);
> ```
Hi again @shahor02 
But this means we still have to provide a timestamp from somewhere between SOR and EOR, right? Or can this also be fetched automatically somehow, e.g. from bookkeeping or is there a helper in O2 already for this?
Edit: what I would like to achieve is that only the run number needs to be provided to automatically get the SOR and EOR times from CCDB

f3sch:
Dont merge yet, waiting for @shahor02 to reply to @martenole.

shahor02:
In the macros, where we can use BasicCCDBManager, one can do:
```
auto& ccdbMgr = o2::ccdb::BasicCCDBManager::instance();
...
auto rr = ccdbMgr.getRunDuration(runNumber);
std::map<std::string, std::string> md;
md["runNumber"] = std::to_string(runNumber);
const auto* grp = ccdbMgr.getSpecific<o2::parameters::GRPECSObject>("GLO/Config/GRPECS", (rr.first+rr.second)/2, md);
```
PS: provided the run was registered in http://alice-ccdb.cern.ch/browse/RCT/Info/RunInformation

f3sch:
Works like a charm, thank you @shahor02.
If CI passes now ready for merge once @martenole approves.

Detectors/TRD/calibration/macros/plotVdriftExB.C
```diff
@@ -0,0 +1,133 @@
...
+
+using timePoint = o2::parameters::GRPECSObject::timePoint;
+static std::array<std::vector<std::tuple<double, double, timePoint>>, 540> vmap; // Map holding all values for all detectors.
+static std::array<bool, 540> good;                                               // Marks wether or not a chamber was 'good' for the entire duration.
+
+// Download the values and populate the map.
+void ccdbDownload(unsigned int runNumber, std::string ccdb, timePoint timeOffset)
+{
+  auto& ccdbMgr = o2::ccdb::BasicCCDBManager::instance();
+  ccdbMgr.setURL("http://alice-ccdb.cern.ch/");
```
martenole, line 42: probably you should set the URL to `ccdb` here

Detectors/TRD/calibration/macros/plotVdriftExB.C
```diff
@@ -0,0 +1,133 @@
...
+  mg->GetXaxis()->SetTitle("time since RunStart (ms)");
+
+  return std::move(mg);
+}
+
+// Plot the calibration values for a particular duration.
+// The times must be given in unix epoch ms format.
+// The default Run is 523677 for the test ccdb.
+// The offset for the next calibration is 15 Minutes.
+void plotVdriftExB(unsigned int runNumber = 523677, std::string ccdb = "http://ccdb-test.cern.ch:8080", timePoint timeOffset = 899991)
```
martenole: Why do you not take 15 minutes (900000) for the `timeOffset`? Also something like `queryInterval` might be more intuitive, but this is up to you

Detectors/TRD/calibration/macros/plotVdriftExB.C
f3sch: Didn't you tell me the GRPECSObject are only alice-ccdb not anywhere
else?
Maybe adding a second parameter for the ccdb for GRPECSObjects is better?

martenole: Yes. Ah, ok, you are changing the URL below to `ccdb` to get the TRD object from the specified path and only the GRP always from alice-ccdb. This I missed. Then this is fine.

Detectors/TRD/calibration/macros/plotVdriftExB.C
f3sch: I do not recall the reason for this exactly but I think in the ccdb the offsets are like this.
But changing this is works just as well.
Renaming is probably better.
_There are only two hard things in Computer Science: cache invalidation, naming things and off by one errors._

martenole (approved):

PR #9867, opened 2022-09-19
MID chamber efficiency objects to be stored in CCDB

dstocco:
This PR prepares the MID chamber efficiency objects that will be stored in CCDB and used in simulations.
The PR consists of 3 commits:
- The first one creates the efficincy counters
- The second one uses the counters in the ChamberEfficiency class, which is now moved to a dedicated directory
- The third one implements a macro to easily create the default CCDB object


Detectors/MUON/MID/Efficiency/macros/makeMIDEfficiencyCCDBEntry.C
```diff
@@ -0,0 +1,29 @@
...
+#include "MIDEfficiency/ChamberEfficiency.h"
+
+int makeMIDEfficiencyCCDBEntry(const char* url = "http://localhost:8080")
+{
+  auto chamberEff = o2::mid::createDefaultChamberEfficiency();
+  auto data = chamberEff.getCountersAsVector();
+  o2::ccdb::CcdbApi api;
+  api.init(url); // or http://localhost:8080 for a local installation
+  std::map<std::string, std::string> md;
+  api.storeAsTFileAny(&data, "MID/Calib/ChamberEfficiency", md, 1, 99999999999999);
```
aphecetche: might want to use `o2::ccdb::CcdbObjectInfo::INFINITE_TIMESTAMP` instead of the bunch of nines ?

Detectors/MUON/MID/Efficiency/src/ChamberEfficiency.cxx
```diff
@@ -25,65 +25,63 @@ namespace o2
...
-  int idx = indexToInt(deId, columnId, line);
+  auto idx = detparams::getUniqueFEEId(deId, columnId, lineId);
   auto entryTot = mCounters.find(idx);
   if (entryTot == mCounters.end()) {
-    std::cerr << "Warning: no efficiency found for deId: " << deId << "  column: " << columnId << "  line: " << line << "  type: " << typeToIdx(type) << "\n";
+    std::cerr << "Warning: no efficiency found for deId: " << deId << "  column: " << columnId << "  line: " << lineId << "  type: " << static_cast<int>(type) << "\n";
     return -1.;
   }
-  return static_cast<double>(entryTot->second[typeToIdx(type)]) / static_cast<double>(entryTot->second[3]);
+  return static_cast<double>(entryTot->second.getCounts(convert(type))) / static_cast<double>(entryTot->second.getCounts(EffCountType::AllTracks));
```
aphecetche, line 66: can we get a division by zero here ?

dstocco: In principle no, due to the check in the previous lines:
```
if (entryTot == mCounters.end()) {
    std::cerr << "Warning: no efficiency found for deId: " << deId << "  column: " << columnId << "  line: " << lineId << "  type: " << static_cast<int>(type) << "\n";
    return -1.;
  }
```
If the counter was added at some point, it means that it has at least 1 entry. So, either the counter does not exist, or the total number of counts is >=1

Detectors/MUON/MID/Efficiency/macros/makeMIDEfficiencyCCDBEntry.C
dstocco: Indeed

aphecetche (dismissed):

aphecetche:
@TimoWilken @ktf just to understand : why is the approval of `AliceO2Group/core` needed to merge this one ? As far as I can tell according to  [CODEOWNERS](https://github.com/AliceO2Group/AliceO2/blob/dev/CODEOWNERS) all the files should be under the `@AliceO2Group/muon-experts` realm ? 
Could the fact that GitHub complains about errors in that file explain that ? Or is it something else completely ?
Thanks,

TimoWilken:
@aphecetche I'm not sure; the current list of changed files is all owned by the "muon-experts" group. But a previous commit that has since been force-pushed out (https://github.com/AliceO2Group/AliceO2/commit/3bb01e02ac07f4a9a82aa79d326628d3a7538f4e) changed `cmake/O2RootMacroExclusionList.cmake`, which is not owned by "muon-experts"; perhaps that triggered the review request from "core".

aphecetche (dismissed):

aphecetche:
@TimoWilken Thanks. Indeed I'm now able to remove the review request to "core" (just I just did).

aphecetche (approved):

PR #9869, opened 2022-09-19
Use prod ccdb

ktf:


ktf:
Let's see if this works. There is a bunch of tests from the detectors which also use the test instance, but hopefully they are not part of the CI.

jgrosseo:
Excellent! And let's extend all the validity of the accessed objects until 2050, so we are good...

CCDB/README.md
```diff
@@ -51,7 +51,7 @@ std::map<std::string, std::string> headers = f.api.retrieveHeaders("FOO/DeadPixe
 // init
 CcdbApi api;
 map<string, string> metadata; // can be empty
-api.init("http://ccdb-test.cern.ch:8080"); // or http://localhost:8080 for a local installation
+api.init("http://alice-ccdb.cern.ch:8080"); // or http://localhost:8080 for a local installation
```
jgrosseo: Drop port

jgrosseo: Everywhere :)

ktf: Hopefully fixed.

PR #9871, opened 2022-09-20
DPL: use production instance for CCDB test

ktf:
ccdb-test is the development instance, so it should not be used for the production tests.


ktf:
Change only in a comment. Merging.

PR #9872, opened 2022-09-20
Allow custom GUIs via services and for devices

ktf:


ktf:
Tested on hyperloop, tested standalone, passes all tests.

PR #9874, opened 2022-09-20
Sign of sigma shouldn't flip with bz

mpuccio:
Hi @shahor02,

we spotted this with @wang-yuanzhe, as it is currently used in the SVertexing it suppresses the efficiency at large momenta if bz is negative. However I am not sure if there is some parameter change that we miss to make it working as intended.

Let us know,
Cheers,
Max


shahor02 (approved):
Thanks for spotting, the fix is correct!

shahor02:
trivial, merging

PR #9875, opened 2022-09-20
TRD: calibration: Added output flag and included deflection correlation macro

f3sch:
I included a macro to plot the deflection correlation and added a flag `--enable-root-output` to dump calibration plots and write/&log fit results.


shahor02:
The fullCI failure is bogus, is this ready to be merged?

martenole (approved):

PR #9876, opened 2022-09-20
DPL Analysis: Rework index builder

aalkin:
* Does not create `soa::Table` objects anymore, reducing overhead
* Some code moved to `IndexBuilderHelpers.h/cxx`
* Groundwork done to allow slice and array index columns for index tables (requires implementing search algos)

### DIFF
--- a/Framework/Core/include/Framework/AnalysisHelpers.h
+++ b/Framework/Core/include/Framework/AnalysisHelpers.h
@@ -20,6 +20,9 @@
 #include "Framework/OutputObjHeader.h"
 #include "Framework/StringHelpers.h"
 #include "Framework/Output.h"
+#include "Framework/IndexBuilderHelpers.h"
+#include <arrow/table.h>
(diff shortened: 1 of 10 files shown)


ktf:
This is just formatting changes, no?

Framework/Core/include/Framework/AnalysisHelpers.h
```diff
@@ -20,6 +20,9 @@
 #include "Framework/OutputObjHeader.h"
 #include "Framework/StringHelpers.h"
 #include "Framework/Output.h"
+#include "Framework/IndexBuilderHelpers.h"
+#include <arrow/table.h>
+#include <arrow/util/key_value_metadata.h>
```
ktf: Can't we hide these in the .cxx via some forward declaration?

aalkin: Yes, I think I can move the whole table creation into .cxx. I will update this later today.

PR #9878, opened 2022-09-20
add FORWARD_TF (ZDC) specs to aggregator proxies

shahor02:
@chiarazampolli


martenole (approved):
Looks good. In the `prodtests/full-system-test/calib-workflow.sh` the output for zdc_tf is already included

PR #9880, opened 2022-09-21
[EMCAL-565] Add Time information for bad channel calib

jokonig:
- In Run2, the time information for the identification of bad channels could only be used in a second step (analysis-level QA) because the time calibration, including the BC%4 correction, was not done at the stage of the bad channel calibration
- In Run3, the BC%4 correction is already performed before the bad channel calibration. Due to that, we can use the width of the cell time distribution to identify bad channels.
- The width (std. deviation) is calculated for each cell. The values for all cells are stored and a mean and sigma of this new distribution is calculated using the TRobustEstimator. Cells above mean + nSigmas*sigma are then flagged as bad (the nSigmas can be modified in the EMCalCalibParams)
- The calibration can be switched on/off using the EMCALCalibParams
- In the offline calibrator and channel-data-producer, the time information can be added via an additional input histogram

### DIFF
--- a/Common/Utils/include/CommonUtils/BoostHistogramUtils.h
+++ b/Common/Utils/include/CommonUtils/BoostHistogramUtils.h
@@ -423,6 +423,32 @@ double getMeanBoost1D(boost::histogram::histogram<axes...>& inHist1D)
   return stats.getMean();
 }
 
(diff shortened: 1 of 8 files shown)


mfasDa (approved):
Looks good!

PR #9882, opened 2022-09-21
FIT: Change data type of FIT DCS CCDB objects

andreasmolander:
Make the map key a string instead of a DataPointIdentifier. This way the DP type is not needed when getting the map value for a DP alias key.


andreasmolander:
I changed my mind, let's keep it as it is.

PR #9885, opened 2022-09-21
DPL: allow customization of driver GUI via services

ktf:


ktf:
Tested also locally, merging.

PR #9886, opened 2022-09-21
DPL: allow loading ThreadPool as plugin

ktf:


ktf:
Should simplify testing the threading support.

PR #9887, opened 2022-09-21
ZDC - Do not produce a fatal error for inconsistent events: just skip the TF

cortesep:


shahor02 (approved):
Hi @cortesep 
will merge once passes CI, but as far as I understand, the errors were quite frequent, in this case the infologger will be flooded by messages. To avoid this we usually silence the error reporting after certain number (a few) of messages. Please open later another PR with the counter of reported errors and foresee a mechanism to mute them after reaching some limit (can be done per error type).

cortesep:
> reported

@shahor02 It is confirmed that the error is consequence of a firmware bug. It will be fixed in the next week therefore we leave the error message. Hopefully it should not be frequent after the fix.

PR #9889, opened 2022-09-21
DPL: make sure MultiOutputProxy can be restarted

ktf:


ktf:
This only appears when using the EPNs, which I did not manage to do on staging at the time.

ktf:
@martenole @davidrohr can we include this in some test EPN build?

PR #9893, opened 2022-09-21
ctpdev: workflowLumi

lietava:
Hi @shahor02 , I created workflowLumi from workflow (=normal reconstruction).
It is supposed to create one lumi record per HB - these can be merged as needed.
Please, have a look and let me know your comments.
Cheers, Roman.


Detectors/CTP/macro/GetScalers.C
```diff
@@ -40,7 +40,8 @@ void GetScalers(std::string srun, long time, std::string ccdbHost = "http://ccdb
   scl = mng.getScalersFromCCDB(time, srun, ok);
   if (ok == 1) {
     scl.convertRawToO2();
-    scl.printStream(std::cout);
+    sc
+      l.printStream(std::cout);
```
shahor02: the line slipped.

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
```diff
@@ -0,0 +1,154 @@
...
+
+void RawToLumiConverterSpec::run(framework::ProcessingContext& ctx)
+{
+  mOutputLumiPoints.clear();
+  const gbtword80_t bcidmask = 0xfff;
+  using InputSpec = o2::framework::InputSpec;
+  using ConcreteDataTypeMatcher = o2::framework::ConcreteDataTypeMatcher;
+  using Lifetime = o2::framework::Lifetime;
+  // mOutputHWErrors.clear();
+  std::vector<InputSpec> filter{InputSpec{"filter", ConcreteDataTypeMatcher{"CTP", "RAWDATA"}, Lifetime::Timeframe}};
```
shahor02: you don't need a filter here since you the workflow subscribes to CTP/RAWDATA only.

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
```diff
@@ -0,0 +1,154 @@
...
+      orbit0 = triggerOrbit;
+      first = false;
+    }
+    auto feeID = o2::raw::RDHUtils::getFEEID(rdh); // 0 = IR, 1 = TCR
+    auto linkCRU = (feeID & 0xf00) >> 8;
+    if (linkCRU == o2::ctp::GBTLinkIDIntRec) {
+      payloadCTP = o2::ctp::NIntRecPayload;
+    } else {
+      return;
+    }
```
shahor02: You are using data of only 1 link, why don't you request only this link in the https://github.com/AliceO2Group/AliceO2/blob/4b086e5f8c575ffea1ee45f59e57a395d8620a8b/Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx#L140 as 
```
inputs.emplace_back("TF", {"CTP", "RAWDATA", <linkID>}, o2::framework::Lifetime::Optional);
```
then you don't need DPLRawParser and this for loop, instead you can directly read 
```
auto ref = inputs.get("TF",<subspec_of_your_link>);
auto rdh =  o2::framework::DataRefUtils::getHeader<o2::header::RAWDataHeader*>(ref);
...
auto payload = pc.inputs().get<const uint8_t>>(ref);
```
Note that in this case if the real input is absent, you will get instead of DEADBEEF stuff the input with the subspec you requested but with empty payload.

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
```diff
@@ -0,0 +1,154 @@
...
+
+  std::vector<o2::framework::OutputSpec> outputs;
+  outputs.emplace_back("CTP", "LUMI", 0, o2::framework::Lifetime::Timeframe);
+
+  return o2::framework::DataProcessorSpec{
+    "CTP-RawStreamDecoder",
+    inputs,
+    outputs,
+    o2::framework::AlgorithmSpec{o2::framework::adaptFromTask<o2::ctp::lumi_workflow::RawToLumiConverterSpec>()},
+    o2::framework::Options{{"result-file", o2::framework::VariantType::String, "/tmp/lumiCTPDecodeResults", {"Base name of the decoding results files."}}}};
```
shahor02: I don't see where this used. In any case, you should not write files on the FLP.

Detectors/CTP/workflowLumi/src/lumiWorkflowSpec.cxx
```diff
@@ -0,0 +1,42 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
```
shahor02: please rename this file to lumiWorkflow.cxx, the "Spec" is for devices, while this one defines the workflow.
also, you don't need corresponding header file.

shahor02 (changes requested):
Hi @lietava 
thanks, see the comments below. The general one: we discussed that the luminosity should be provided as an instantaneous luminosity integrated for ~1s, I don't see much use for per-HB lumi, nor it can be precise. It is fine to have it, but could you please add on top of it e.g. 1s. (should be a parameter of the workflow) running average?

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
lietava: HI @shahor02 ,

>``` inputs.emplace_back("TF", {"CTP", "RAWDATA", <linkID> }, o2::framework::Lifetime::Optional);```

does not compile - there is no such function.
I changed it to
```inputs.emplace_back("TF", "CTP", "RAWDATA", <linkID>, o2::framework::Lifetime::Optional);```
Then when I try to run on MC it can not find data:
[O2/latest-ctpdev-o2] ~/testlumi0 $> o2-raw-file-reader-workflow --input-conf CTPraw.cfg | o2-ctp-lumi-workflow -b 
[ERROR] invalid workflow in o2-ctp-lumi-workflow: No matching output found for CTP/RAWDATA/0 as requested by data processor "CTP-RawStreamDecoder". Candidates:
-DPL/ENUM/1301969186
-CTP/RAWDATA
-FLP/DISTSUBTIMEFRAME/0
-CTP/LUMI/0
? 
Cheers, Roman.

Comment (shahor02) [reply Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx]: sorry, there was a typo: `inputs.emplace_back("TF", "CTP", "RAWDATA", 0, o2::framework::Lifetime::Optional);`

Comment (lietava) [reply Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx]: Yes, I understood you did typo and I fixed it.  Please, read also after 'I changed it ...' in my post above.

Comment (shahor02) [reply Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx]: Ah, right, my comment was wrong: since the RAWDATA is sent with wildcard, the DPL will fail to build a topology with input asking for specific subspec... Indeed, one should use DPLRawParser. Sorry for misleading comment.

Comment (lietava) [reply Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx]: ok, no problem, thanks. 
Is there a way to specify interaction rate in o2-sim ?

Comment (shahor02) [reply Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx]: It is an option of the `o2-sim-digitizer-workflow`: `--interactionRate <rate_in_Hz>`

File: Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx (hunk 2 above)
Comment (lietava) [reply Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx]: I am not sure if I need to keep filter.

File: Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx (hunk 4 above)
Comment (lietava) [reply Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx]: LIne 153 removed.

**Comment by lietava** [comment]
> Hi @lietava thanks, see the comments below. The general one: we discussed that the luminosity should be provided as an instantaneous luminosity integrated for ~1s, I don't see much use for per-HB lumi, nor it can be precise. It is fine to have it, but could you please add on top of it e.g. 1s. (should be a parameter of the workflow) running average?

There is parameter --nHBF-to-IRaverage which specifies number of HB frames to average.

File: Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx (hunk 2 above)
Comment (shahor02) [reply Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx]: You don't need it.

**Comment by lietava** [comment]
@shahor02 : I removed filter. It seems I 've done everything you suggested. Also I tested with MC and works.

**Comment by shahor02** [comment]
@TimoWilken @davidrohr looks like switching to `https` for `alice-ccdb.cern.ch` leads to failure in ccdb access in CI. Is the token set only for fullCI? If so, can we have it for all builders?

**Comment by lietava** [comment]
@shahor02 : anything on my side ?

File: Detectors/CTP/workflowLumi/include/CTPWorkflowLumi/RawToLumiConverterSpec.h [hunk 6]
```diff
@@ -0,0 +1,76 @@
...
+
+namespace lumi_workflow
+{
+struct lumiPoint {
+  lumiPoint() = default;
+  InteractionRecord ir;                                 // timestamp of start of lumi interval
+  float_t nHBFs = 1;                                    // length of interval in HB
+  float_t counts = 0;                                   //  counts in the interval
+  float_t getLumi() { return counts / nHBFs / 88e-6; }; // rate in Hz
+  float_t getFractErrorLumi() { return 1. / sqrt(counts); };
```
shahor02: need protection against counts=0, e.e. `return counts>0. ? 1. / sqrt(counts) : -1;`

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
```diff
@@ -0,0 +1,176 @@
...
+      lp.ir = lumiPointsHBF1[ihb].ir;
+    } else {
+      countHBFn -= lumiPointsHBF1[ihb - mHBFsToAverage].counts;
+      lp.nHBFs = mHBFsToAverage;
+      lp.ir = lumiPointsHBF1[ihb - mHBFsToAverage + 1].ir;
+    }
+    lp.counts = countHBFn;
+    // LOG(info) << "lp orbit:" << lp.ir.orbit << " lp.countes:" << lp.counts << " T:" << lp.nHBFs;
+    mOutputLumiPoints.push_back(lp);
+  }
```
shahor02: Not sure this is amounts for ~1s running average we need: the accumulation is done only within 1 TF. I would simply define a data members
```
float mCounts = 0.;
size_t mNTFToIntegrate = 0;
size_t mNHBIntegrated = 0;
std::deque<float> mHistory;
```
and the number of TFs to integrate as an option (instead of the NHB) and assign it as:
```
float tfCounts = 0; // counts in this TF
for (int ihb = 0; ihb < lumiPointsHBF1.size(); ihb++) {
  tfCounts += lumiPointsHBF1[ihb].counts;
}
mHistory.push_back(tfCounts);
mCounts += tfCounts;
if (mHistory.size() < mNTFToIntegrate) { // still did not reach target NTFs
  mNHBIntegrated += lumiPointsHBF1.size(); // Assuming that we always have the same size in all TFs
} else {
  mCounts -= mHistory.front();
  mHistory.pop_front();
}
// normalize
float meanLumi = mNHBIntegrated ? mCounts / (mNHBIntegrated*o2::contacts::lhc::LHCOrbitMUS*1e-6) : 0;
// send meanLumi
```

shahor02 (commented):
Hi @lietava 
The CI failures are unrelated, the reason was already fixed, CI will be redone. But please check comments below.

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
lietava: @shahor02 : ok, I see . But you do not know mCounts and so error ? Is it not better to specify how many mCounts do we need then ?

shahor02: @lietava  What do you mean by `how many mCounts do we need then`? We should request only counting period, above I put `mNTFToIntegrate` for that. I assume that lumiPointsHBF1.size() is the same for all TFs and each entry always corresponds to 1 orbit. 
For the return value, indeed better to return a simple struct like
```
struct LumiInfo {
  size_t mCounts = 0;
  size_t mNHBFCounted = 0; 
};
```

lietava: ok, I did averging over TF as you suggested.
git: ctpdev was modified by you. I did
git pull ctpdev
git push ctpdev 
I restored  prodtests/full-system-test/dpl-workflow.sh (commit: adding CTP to reconstruction)

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
```diff
@@ -0,0 +1,183 @@
...
+      const auto dh = o2::framework::DataRefUtils::getHeader<o2::header::DataHeader*>(ref);
+      auto payloadSize = o2::framework::DataRefUtils::getPayloadSize(ref);
+      if (payloadSize == 0) {
+        auto maxWarn = o2::conf::VerbosityConfig::Instance().maxWarnDeadBeef;
+        if (++contDeadBeef <= maxWarn) {
+          LOGP(alarm, "Found input [{}/{}/{:#x}] TF#{} 1st_orbit:{} Payload {} : assuming no payload for all links in this TF{}",
+               dh->dataOrigin.str, dh->dataDescription.str, dh->subSpecification, dh->tfCounter, dh->firstTForbit, payloadSize,
+               contDeadBeef == maxWarn ? fmt::format(". {} such inputs in row received, stopping reporting", contDeadBeef) : "");
+        }
+        ctx.outputs().snapshot(o2::framework::Output{"CTP", "LUMI", 0, o2::framework::Lifetime::Timeframe}, mOutputLumiInfo);
```
shahor02, line 57: please create an empty local copy for sending dummy output, so that we know it has to be ignored.

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
```diff
@@ -70,6 +70,7 @@ void RawToDigitConverterSpec::run(framework::ProcessingContext& ctx)
   for (auto it = parser.begin(); it != parser.end(); ++it) {
     auto rdh = it.get_if<o2::header::RAWDataHeader>();
     auto triggerOrbit = o2::raw::RDHUtils::getTriggerOrbit(rdh);
+    std::cout << "==================>" << std::hex << triggerOrbit << std::endl;
```
shahor02: this will create a huge loge, please remove.

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
```diff
@@ -0,0 +1,183 @@
...
+      orbit0 = triggerOrbit;
+      first = false;
+    }
+    LOG(debug) << "RDH FEEid: " << feeID << " CTP CRU link:" << linkCRU << " Orbit:" << triggerOrbit;
+    // TF in 128 bits words
+    gsl::span<const uint8_t> payload(it.data(), it.size());
+    gbtword80_t gbtWord = 0;
+    int wordCount = 0;
+    std::vector<gbtword80_t> diglets;
+    if (orbit0 != triggerOrbit) {
```
shahor02, line 92: I am a little bit confused by the logic. In this `if` you are assigning the counts of the previous orbit, and aft the for loop below you are pushing to the lumiPointsHBF1 the counts of the last orbit, right? Why not simply counting for each orbit and assigning in the end of the `for` loop?

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
```diff
@@ -0,0 +1,183 @@
...
+  // LOG(info) << "lumiPoints size:" << lumiPointsHBF1.size() << " History size:" << mHistory.size();
+  //
+  size_t tfCounts = 0.;
+  for (auto const& lp : lumiPointsHBF1) {
+    tfCounts += lp.mCounts;
+  }
+  mHistory.push_back(tfCounts);
+  mCounts += tfCounts;
+  // std::cout << tfCounts << " " << mCounts << std::endl;
+  if (mHistory.size() <= mNTFToIntegrate) {
```
shahor02, line 150: Why `<=` and not `<` ?

lietava: when mNHBToIntegrate = 1
mHistory.size() = 1  (was pushed in line 140)
1 < 1  is false(line 150)  and mCounts / nHBIntegrated  are always 0.

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
lietava: Sorry, forgot it. Now removed.

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
lietava: Sorry, I do not understand context. Please, explain a bit more.
Local copy of what ? To be ignored by what ?

shahor02: I meant that usually when detecting 0xdeadbeef we send dummy (empty) output to signal that there was no input, while you are sending the counts from the last non-empty TF. But, after second thinking, we should probably leave it as it is: since the accumulator carries the orbit of the last update, the consumer can detect that it was not updated.

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
shahor02: ok, right

Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx
shahor02: It seems you did not push updated code, please do so, if this is the only change, I'll merge it w/o waiting for the new CI

shahor02 (approved):
Apart from the `std::cout` to remove it is fine. Once merged, can you suggest me some run to test it? Since building such a counter from the CTF is not possible due to TFs there not being consecutive, I will add this counter to the CTP CTF.

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
lietava: Thanks. So I keep it as it is.
For my understanding:
line https://github.com/lietava/AliceO2/blob/001cd99590ad937a693d61132abd1032dda6b50b/Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx#L57
sends always last lumiInfo
while
https://github.com/lietava/AliceO2/blob/001cd99590ad937a693d61132abd1032dda6b50b/Detectors/CTP/workflow/src/RawToDigitConverterSpec.cxx#L58
send empty payload , and so you wanted me to create empty dummy lumiInfo ?

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
lietava: Yes, I see. But if there is more than one page for HB then orbit will not change between RDHs ?

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
shahor02: Not sure what do you mean by L58. https://github.com/lietava/AliceO2/blob/001cd99590ad937a693d61132abd1032dda6b50b/Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx#L57 is currently sending last lumiInfo in case CTP did not provide any input (hence the deadbeef was generated). Initially I was proposing to send in this case a dummy object, but then I've realized that since the lumiInfo is not really TF-specific and it has a timestamp, it is better to keep it as is. 
https://github.com/lietava/AliceO2/blob/001cd99590ad937a693d61132abd1032dda6b50b/Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx#L163 sends the output updated with the last TF CTP info.

shahor02 (changes requested):
Sorry, I've just noticed that the `struct lumiInfo` is defined in your workflow header. Other workflows will not be able to use it... Could you please define it in the separate file in the DataFormats/Detectors/CTP (and rename it to LumiInfo to follow the naming conventions).

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
shahor02: No, multi-page HBs keep the same RDH, only page counter is incremented.

lietava: @shahor02 : you did not comment here. I kept it as it is.

lietava:
> Sorry, I've just noticed that the `struct lumiInfo` is defined in your workflow header. Other workflows will not be able to use it... Could you please define it in the separate file in the DataFormats/Detectors/CTP (and rename it to LumiInfo to follow the naming conventions).

I'll do

Detectors/CTP/workflowLumi/src/RawToLumiConverterSpec.cxx
shahor02: sorry, comment on what?

lietava:
@shahor02 : i'll send you some runs with ctpreadout

lietava:
@shahor02 : there is today short run 526638. (One can see it in logbook in environment : ctp_readout_enabled="true", e.g. in logbook.)

lietava:
HI @shahor02 : did you test with data ?

shahor02:
yes, after some fixes (in https://github.com/AliceO2Group/AliceO2/pull/10008 which I still need to check thoroughly and add some fixes for the CTF dictionary loading) it works.

lietava:
> yes, after some fixes (in #10008 which I still need to check thoroughly and add some fixes for the CTF dictionary loading) it works.

@shahor02 : thank you. I 'll try to check consistency  if readout in raw  as we have discussed in email.

PR #9895, opened 2022-09-21
DPL: allow resetting the oldest possible timeframe mechanism

ktf:


davidrohr:
@ktf : What happens actually at stop/start. Will the timeSlice start again at 0? Then it must also be reset at several other places?

vascobarroso:
@davidrohr for the O2 tasks (readout/DD/reco/QC/calib) a stop/start reusing an existing environment should be indistinguishable from a fresh start using a new environment.

davidrohr:
well, depends. If we want to use that stop/start as PAR perhaps not?

vascobarroso:
Indeed. But I think the current stop/start being developed is closer to the fast SOR/EOR we had during Run 2. PAR will come later.

vascobarroso:
@ktf @davidrohr any reason why this has not yet been merged ?

vascobarroso:
@ktf @davidrohr @shahor02 any chance this will be merged this week ? There is a scheduled START/STOP/START test at P2 on Tuesday next week, but if this will not be there I think it would be better to postpone.

ktf:
Sorry this week was complicated. I just fixed the conflict and I am trying this again.

davidrohr:
@ktf : shall we merge this now?

davidrohr:
this fixes issues I see on the EPN, tested locally, only fixed clang format, merging

PR #9903, opened 2022-09-22
IDCs: fix warning message

matthias-kleiner:


matthias-kleiner:
can this be merged, please? When the warning was triggered before this PR it caused a crash of the workflow due to the missing parameter

chiarazampolli:
@shahor02 , @martenole , could you merge?

Detectors/TPC/workflow/include/TPCWorkflow/TPCFactorizeIDCSpec.h
```diff
@@ -81,19 +81,21 @@ class TPCFactorizeIDCSpec : public o2::framework::Task
...
+      LOGP(warning, "firstTF not Found!!! Found valid inputs {}. Setting {} as first TF", pc.inputs().countValidInputs(), mTFFirst);
     }
 
+    const unsigned int relTF = (mTFFirst == -1) ? 0 : currTF - mTFFirst;
+
     // loop over input data
     for (auto& ref : InputRecordWalker(pc.inputs(), mFilter)) {
       ++mProcessedCRUs;
-      if (relTF >= mIDCFactorization.getNTimeframes()) {
+      if ((relTF >= mIDCFactorization.getNTimeframes()) || (relTF < 0)) {
```
martenole, line 98: `relTF` is unsigned and will never be smaller than 0. Is it also OK without the check or do you want to check above whether `currTF >= mTFFirst`?

matthias-kleiner: Thanks I missed that, I will do a check than for `currTF >= mTFFirst`

matthias-kleiner: I just changed the datatype from `unsigned int` to `long`, which would be the proper solution, also for the other places where `relTF` is used

martenole (approved):

matthias-kleiner:
Can this be merged now? The error in the fullCI seems unrelated

PR #9905, opened 2022-09-23
Fix coding conventions

ktf:


ktf:
@martenole Not sure how come this made it through. I thought I fixed all of them.

martenole:
Yes, I also don't understand why your PR passed the fullCI and in checks of later PRs its caught then.
Anyhow, thanks for the quick fix!

martenole:
For reference, this should have been caught in the fullCI of https://github.com/AliceO2Group/AliceO2/pull/9775 but was in turn only discovered when the fullCI failed here: https://github.com/AliceO2Group/AliceO2/pull/9875

PR #9907, opened 2022-09-23
TRD tracking QC is a proper DPL message

martenole:
See https://alice.its.cern.ch/jira/projects/TRD/issues/TRD-122 for more details


martenole:
I don't understand this. The errors reported in the fullCI
```
/sw/SOURCES/O2/9907-slc8_x86-64/0/GPU/GPUTracking/TPCClusterFinder/GPUTPCCFDecodeZS.cxx:753:77: error: statement should be inside braces [readability-braces-around-statements]
/sw/SOURCES/O2/9907-slc8_x86-64/0/GPU/GPUTracking/TPCClusterFinder/GPUTPCCFDecodeZS.cxx:753:77: error: statement should be inside braces [readability-braces-around-statements]
/sw/SOURCES/O2/9907-slc8_x86-64/0/GPU/GPUTracking/TPCClusterFinder/GPUTPCCFDecodeZS.cxx:753:77: error: statement should be inside braces [readability-braces-around-statements]
```
were actually introduced in https://github.com/AliceO2Group/AliceO2/pull/9892 but there the fullCI did not fail. We had this also yesterday with a different error: https://github.com/AliceO2Group/AliceO2/pull/9905#issuecomment-1256042085
Any idea why this might be @TimoWilken or @davidrohr ?

davidrohr:
@martenole : It is caused by incremental builds. The codechecker checks only the files that were rebuilt in the last incremental build.
If the file that offends the codechecker did compile, but the build failed at another file, the incremental build will not rebuild the offending file when the other file was fixed, and then the codechecker will not check it.
It has happened several times in the past.
No idea how to solve that properly. We could discuss with @sawenzel who set up the codechecker initially.

martenole:
> @martenole : It is caused by incremental builds. The codechecker checks only the files that were rebuilt in the last incremental build. If the file that offends the codechecker did compile, but the build failed at another file, the incremental build will not rebuild the offending file when the other file was fixed, and then the codechecker will not check it. It has happened several times in the past. No idea how to solve that properly. We could discuss with @sawenzel who set up the codechecker initially.

Ok thanks so it is at least understood. I think you mentioned this problem before, but I forgot about it. Now that I saw this twice on one day I will not forget again

Detectors/TRD/workflow/include/TRDWorkflow/TRDGlobalTrackingQCSpec.h
```diff
@@ -19,17 +19,113 @@
...
+    if (o2::base::GRPGeomHelper::instance().finaliseCCDB(matcher, obj)) {
+      return;
+    }
+  }
+
+ private:
+  void updateTimeDependentParams(framework::ProcessingContext& pc)
+  {
+    o2::base::GRPGeomHelper::instance().checkUpdates(pc);
+    static bool initOnceDone = false;
```
f3sch, line 73: Does this need to be static or can it be turned into a member var?

f3sch:
I noticed that we do not have README for qc, maybe it would be worthwhile to start it by at least referencing the jira ticket?

f3sch (approved):

Detectors/TRD/workflow/include/TRDWorkflow/TRDGlobalTrackingQCSpec.h
martenole: I think you are right it could also be a member variable. At the point when this DPL CCDB mechanism was added to all workflows it was probably easier to just make it `static` in all devices. If you grep for it you can see that it's done like this everywhere, so I would just leave it like that here as well.

PR #9909, opened 2022-09-24
[EMCAL-565] Fix condition for scalefactors in online mode

mfasDa:
Faulty comparison leading to scale factors being loaded for
the wrong calibration method (time calibration instead of
bad channel calibration)


mfasDa:
Unfortunately the wrong comparator leads in loading the scale factors for energy adjustment behind the TRD support structure for the time calibration instead of the bad channel calibration. This will not lead to a crash, the scalefractors are just ignored in the time calib and not applied in the bad channel calib, but we see the TRD support structures in the bad channel maps created since last night.

shahor02 (approved):

PR #9910, opened 2022-09-24
Optionally use Logger for printKeyValues

shahor02:
The output is sent to the logger instead of the std::cout if printKeyValues is called with 2nd parameter set to true


sawenzel (approved):
Ok. (Alternatively we could just always use the logger and have simpler interfaces).

PR #9916, opened 2022-09-25
Add L1phase calibration

peressounko:


peressounko:
Dear @shahor02, could you please have a look at this PR? I added another histo to calibrate L1phase shift (time between 10 mHz PHOS clock and 40 MHz LHC clock)

peressounko:
@shahor02, current (sporadic) error on macOS-arm is not related to commit. Do we continue running tests even once all tests were successful?

shahor02:
@timo @ktf here too, the connection with https to ccdb systematically fails on macOS-arm. Is this a token problem?

PR #9920, opened 2022-09-26
Fix Diamond workflow example

awegrzyn:
Currently Diamond workflow fails with fatal from FMQ:
```
[FATAL] error while setting up workflow in o2-testworkflows-diamond-workflow: Cannot update configuration. Socket method (bind/connect) for channel metric-feedback[0] not specified.
```


ktf (commented):
Sounds like a good idea, indeed... This will not open a new port, though, will it? If yes, maybe we should use inproc as a transport?

ktf:
@costing this is the CCDB error I was referring to. It's a bit everywhere.
@sawenzel any idea?

ktf:
@chiarazampolli did ITS/Calib/Align got deleted?

awegrzyn:
@ktf I don't fully understand what supposed to be pushed to that socket, my changes are minimal to get the device running, I guess I wanted me to add `address=ipc://...` ?

PR #9924, opened 2022-09-26
Move info to debug message in TPC raw to digits wf

martenole:
To avoid clogging the EPN disks: https://alice.its.cern.ch/jira/browse/O2-3195


wiechula:
@martenole , if this message is seen too often, then something must be wrong in the processing and should be investigated. Just silencing is not the best, I think. I can implement some throttling.

martenole:
That would be great, thanks! This was just thought as a poor man's temporary solution and to be reverted once the issue is understood. But message throttling is the proper workaround for now. So closing this one

PR #9925, opened 2022-09-26
[EMCAL-566] Add shift for low gain time calibration

jokonig:
- In run3, the statistics accumulated before the asynchornus reconstruction is (most likely) not sufficient to calibrate the low gain cells (> ~16GeV)
- From previous run2 studies and new data from LHC22m, we know that the low gain calibration factors should be nearly the same as the high gain calib factors, however they are shifted by a few nanoseconds
- Implemented that the low gain calib factors are taken from the high gain factors and added a constant shift for all cells
- Shift is settable via the EMCALCalibParams


mfasDa (approved):
Looks good!

shahor02 (approved):

PR #9926, opened 2022-09-26
TPC: do not depend on TPCMonitor for the reconstruction workflow

ktf:


ktf:
`O2::TPCMonitor` depends on the kitchen sink, so we should not include it in the default TPC reconstruction workflow.

ktf:
@chiarazampolli FYI

chiarazampolli:
@ktf , by how much will this reduce the number of opened files?

ktf:
@shahor02 @davidrohr is:

```
## sw/BUILD/O2-latest/log
 26/434 Test  #24: GPU/TPCFastTransformation/test/testMultivarPolynomials.cxx ...........................***Failed    9.05 sec
99% tests passed, 1 tests failed out of 434
100% tests passed, 0 tests failed out of 102
```

a known issue?

@sawenzel @costing I think there is some issue with CCDB.

ktf:
@chiarazampolli I need to check. It depends if this also reduces the huge amount of .pcm files we are currently loading as a side effect.

ktf:
ARM mac issue unrelated. Merging.

PR #9927, opened 2022-09-26
TPC dEdx residual correction load B-field from CCDB

matthias-kleiner:
- MIP track filter:
   - add option for setting maximum number of MIPs per TF
   - add option for skipping every nth TF

### DIFF
--- a/Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
+++ b/Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
@@ -27,6 +27,8 @@
 #include "Framework/ConfigParamRegistry.h"
 #include "TPCCalibration/CalibdEdx.h"
 #include "TPCWorkflow/ProcessingHelpers.h"
+#include "TPCBase/CDBInterface.h"
+#include "DetectorsBase/GRPGeomHelper.h"
 
 using namespace o2::framework;
 
@@ -36,8 +38,11 @@ namespace o2::tpc
 class CalibdEdxDevice : public Task
 {
  public:
+  CalibdEdxDevice(std::shared_ptr<o2::base::GRPGeomRequest> req) : mCCDBRequest(req) {}
+
   void init(framework::InitContext& ic) final
   {
+    o2::base::GRPGeomHelper::instance().setRequest(mCCDBRequest);
     const auto minEntriesSector = ic.options().get<int>("min-entries-sector");
     const auto minEntries1D = ic.options().get<int>("min-entries-1d");
     const auto minEntries2D = ic.options().get<int>("min-entries-2d");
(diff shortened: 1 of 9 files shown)


Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -51,34 +56,39 @@ class CalibdEdxDevice : public Task
...
+    o2::base::GRPGeomHelper::instance().checkUpdates(pc);
     const auto tfcounter = o2::header::get<DataProcessingHeader*>(pc.inputs().get("tracks").header)->startTime;
     const auto tracks = pc.inputs().get<gsl::span<TrackTPC>>("tracks");
 
     LOGP(info, "Processing TF {} with {} tracks", tfcounter, tracks.size());
-    mRunNumber = processing_helpers::getRunNumber(pc);
     mCalib->fill(tracks);
+
+    // store run number and CCDB time only once
+    if (mTimeStampStart == 0) {
```
shahor02: in view of upcoming changes due to the stop/start functionality which may lead to the run number change it is worth to  change this to `if (mTimeStampStart == 0 || pc.services().get<o2::framework::TimingInfo>().timeslice)`

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -51,34 +56,39 @@ class CalibdEdxDevice : public Task
...
     mCalib->set1DFitThreshold(minEntries1D);
     mCalib->set2DFitThreshold(minEntries2D);
-    mCalib->setField(field);
     mCalib->setElectronCut(fitThreshold, fitPasses);
   }
 
+  void finaliseCCDB(o2::framework::ConcreteDataMatcher& matcher, void* obj) final
+  {
+    if (o2::base::GRPGeomHelper::instance().finaliseCCDB(matcher, obj)) {
+      const auto field = (5.00668f / 30000.f) * o2::base::GRPGeomHelper::instance().getGRPMagField()->getL3Current();
```
shahor02: Note that this `if` might be entered when orbitResetTime is shipped and before the field value is ready.
Therefore, you should either check  additionally 
```
if (matcher == ConcreteDataMatcher("GLO", "GRPMAGFIELD", 0)) {
```
or move this field setting to the `run` method after https://github.com/AliceO2Group/AliceO2/blob/ae900e89c6975193657e82b87cfde4776e96d1c1/Detectors/TPC/workflow/src/CalibdEdxSpec.cxx#L79

Also, I see that in the CalibdEdx.cxx you are using `track.propagateTo(x,bz)` method. The nominal field (at IP) you are extracting by this scaling is not good for propagation in the TPC. If you want more or less precise propagation, the recommended is using e.g. (if you don't need material corrections): 
https://github.com/AliceO2Group/AliceO2/blob/66165fcad643a61ecacbe94e9b37cbc3599d0699/Detectors/GlobalTracking/src/MatchTPCITS.cxx#L1138
Note that in that case you don't need to pass the field, it will be set automatically in the GRPGeomHelper provided you pass after the `input` at https://github.com/AliceO2Group/AliceO2/blob/ae900e89c6975193657e82b87cfde4776e96d1c1/Detectors/TPC/workflow/src/CalibdEdxSpec.cxx#L136 extra 2 `true` arguments (to set up the propagator).

Detectors/TPC/workflow/src/CalibratordEdxSpec.cxx
```diff
@@ -86,9 +76,15 @@ class CalibratordEdxDevice : public Task
       mCalibrator->enableDebugOutput("calibratordEdx.root");
     }
   }
+
   void finaliseCCDB(o2::framework::ConcreteDataMatcher& matcher, void* obj) final
   {
-    o2::base::GRPGeomHelper::instance().finaliseCCDB(matcher, obj);
+
+    if (o2::base::GRPGeomHelper::instance().finaliseCCDB(matcher, obj)) {
```
shahor02: same comment here

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
shahor02: One more comment: I see that you don't use the covariance matrix of propagated tracks, in this case I would suggest to create a parame only clone of the track at https://github.com/AliceO2Group/AliceO2/blob/66165fcad643a61ecacbe94e9b37cbc3599d0699/Detectors/TPC/calibration/src/CalibdEdx.cxx#L70
i.e. `o2::track::TrackPar cpTrack = track;`
Then either use corresponding propagator method or `cpTrack.propagateParamTo...`

matthias-kleiner:
Many thanks for your comments and suggestions @shahor02 . I changed the track propagation to `PropagateToXBxByBz` as you suggested. Could you take a look at the changes I made and check if everything is correct?

wiechula (dismissed):

Detectors/TPC/workflow/src/CalibdEdxSpec.cxx
```diff
@@ -51,34 +56,35 @@ class CalibdEdxDevice : public Task
...
+    o2::base::GRPGeomHelper::instance().checkUpdates(pc);
     const auto tfcounter = o2::header::get<DataProcessingHeader*>(pc.inputs().get("tracks").header)->startTime;
     const auto tracks = pc.inputs().get<gsl::span<TrackTPC>>("tracks");
 
     LOGP(info, "Processing TF {} with {} tracks", tfcounter, tracks.size());
-    mRunNumber = processing_helpers::getRunNumber(pc);
     mCalib->fill(tracks);
+
+    // store run number and CCDB time only once
+    if (mTimeStampStart == 0 || pc.services().get<o2::framework::TimingInfo>().timeslice) {
```
shahor02: sorry, here there should be `if (mTimeStampStart == 0 || pc.services().get<o2::framework::TimingInfo>().timeslice==0) {`
This still may change as the framework will provide a method to detect run start after stop/start procedure.

shahor02 (commented):
Hi @matthias-kleiner 
Looks fine except small comment below (which was due to my typo).

shahor02 (approved):

PR #9930, opened 2022-09-26
Drop bogus deps

ktf:


ktf:
Better handled in #9979.

PR #9935, opened 2022-09-27
Reconstruction of displaced 3 body vertices

mpuccio:
Hi @shahor02,

This PR contains the development we did with @wang-yuanzhe to include the reconstruction of displaced three body vertices in the SVertexer. The code is ready, but I am still in the process of rebasing, and I am opening the PR to start the discussion.

Cheers,
Max


Detectors/GlobalTrackingWorkflow/src/SecondaryVertexingSpec.cxx
```diff
@@ -91,12 +93,17 @@ void SecondaryVertexingSpec::run(ProcessingContext& pc)
...
   auto& cascRefs = pc.outputs().make<std::vector<RRef>>(Output{"GLO", "PVTX_CASCREFS", 0, Lifetime::Timeframe});
+  auto& vtx3body = pc.outputs().make<std::vector<DecayNbody>>(Output{"GLO", "DECAYS3BODY", 0, Lifetime::Timeframe});
+  auto& vtx3bodyRefs = pc.outputs().make<std::vector<RRef>>(Output{"GLO", "PVTX_3BODYREFS", 0, Lifetime::Timeframe});
 
+  LOG(info)<<"SecondaryVertexingSpec::run process Start";
   mVertexer.process(recoData);
-  mVertexer.extractSecondaryVertices(v0s, v0Refs, cascs, cascRefs);
+  LOG(info)<<"SecondaryVertexingSpec::run process Finish";
+  mVertexer.extractSecondaryVertices(v0s, v0Refs, cascs, cascRefs, vtx3body, vtx3bodyRefs);
+  LOG(info)<<"SecondaryVertexingSpec::run extractSecondaryVertices Finish";
```
shahor02: is this leftover?

Detectors/GlobalTrackingWorkflow/src/secondary-vertexing-workflow.cxx
```diff
@@ -82,9 +82,11 @@ WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
   specs.emplace_back(o2::vertexing::getSecondaryVertexingSpec(src, enableCasc));
 
   // only TOF clusters are needed if TOF is involved, no clusters MC needed
+        LOG(info) << " addInputSpecs Start ";
   WorkflowSpec inputspecs;
   o2::globaltracking::InputHelper::addInputSpecs(configcontext, inputspecs, srcClus, src, src, useMC, srcClus);
   o2::globaltracking::InputHelper::addInputSpecsPVertex(configcontext, inputspecs, useMC); // P-vertex is always needed
+        LOG(info) << " addInputSpecs Finish";
```
shahor02: this too?

Detectors/Vertexing/src/SVertexer.cxx
```diff
@@ -389,25 +437,59 @@ bool SVertexer::checkV0(const TrackCand& seedP, const TrackCand& seedN, int iP,
...
 
+  // we want to reconstruct the 3 body decay of hypernuclei starting from the V0 of a proton and a pion (e.g. H3L->d + (p + pi-), or He4L->He3 + (p + pi-)))
+  bool checkFor3BodyDecays = mEnable3BodyDecays && (!mSVParams->checkV0Hypothesis || good3bodyV0Hyp) && (pt2V0 > 0.5);
+  bool rejectAfter3BodyCheck = false; // To reject v0s which can be 3-body decay candidates but not cascade or v0
   bool checkForCascade = mEnableCascades && r2v0 < mMaxR2ToMeanVertexCascV0 && (!mSVParams->checkV0Hypothesis || (hypCheckStatus[HypV0::Lambda] || hypCheckStatus[HypV0::AntiLambda]));
   bool rejectIfNotCascade = false;
+
+  if (!goodHyp && mSVParams->checkV0Hypothesis) {
+    LOG(debug) << "RejHypo";
+    if (!checkFor3BodyDecays){
```
shahor02: likely the checkForCascade will be false here, but better have it explicitly checked

shahor02 (commented):
Hi @mpuccio 
At first sight, everything seems right, see minor comments. In your 3body, do you consider the vertices made of 3 TPC-only tracks?

Detectors/Vertexing/src/SVertexer.cxx
mpuccio: Added!

Detectors/GlobalTrackingWorkflow/src/secondary-vertexing-workflow.cxx
mpuccio: Fixed!

Detectors/GlobalTrackingWorkflow/src/SecondaryVertexingSpec.cxx
mpuccio: Fixed!

mpuccio:
Hi @shahor02, for the time being, we use all the tracks in the pool as "bachelors". We did not put any specific requirement to avoid it; we can if required. In the future, we might add the additional requirement of having a "nuclei" PID in tracking, which will reject most of the possible bachelors and speed up the processing.

mpuccio:
I think this is as green as it gets... @shahor02, do you think we can merge?

shahor02 (dismissed):
sure, but I see you set it on by default. Before I merge: did you try to benchmark what is the overhead in pp and pbpb?

mpuccio:
Argh, you are right! For the time being, I would keep it off until we get numbers for Pb-Pb. It takes ~ the same time as the cascades in pp high rates. However, we heavily injected the events with hypertritons, so the timing might be pessimistic.

shahor02:
OK, but now you cannot enable it, could you add/pass an option like 
https://github.com/AliceO2Group/AliceO2/blob/80ea2bbb2fc296e1232fb6604f6033e465005456/Detectors/GlobalTrackingWorkflow/src/secondary-vertexing-workflow.cxx#L54
https://github.com/AliceO2Group/AliceO2/blob/80ea2bbb2fc296e1232fb6604f6033e465005456/Detectors/GlobalTrackingWorkflow/src/secondary-vertexing-workflow.cxx#L75
but with opposite action, i.e. `enable-3body` ?

mpuccio:
Good point; with the new modification, the command line option is in place.

mpuccio:
The full CI became red, but on an unrelated error. If there are no further suggestions, I would try to merge this to move on with the rest of the developments @shahor02

shahor02 (approved):

PR #9938, opened 2022-09-27
[EMCAL-612] Make workflow name optional for cell writer

mfasDa:
Needed in order to run different cell writers in parallel. As
default the current workflow name is used so the standard
behaviour is unchanged.


mfasDa:
Mainly needed for debugging (i.e. optional cell trees with calibrated and uncalibrated cells) in dpl-workflow.sh.

mfasDa:
Errors unrelated

shahor02 (approved):

PR #9940, opened 2022-09-27
Use CorrectionMapsHelper to allow TPC scaled transform

shahor02:
Hi @davidrohr 

Here is it. The code would be more robust and future changes easier if we could get rid of https://github.com/shahor02/AliceO2/blob/a53471588e2bb7561a09c5a6cbc5577c9dc8b62f/GPU/GPUTracking/DataTypes/GPUDataTypes.h#L216-L217 and the fastTransform would be accessed via https://github.com/shahor02/AliceO2/blob/a53471588e2bb7561a09c5a6cbc5577c9dc8b62f/GPU/GPUTracking/DataTypes/GPUDataTypes.h#L218.
Do you see anything against this?


sgorbuno (dismissed):

GPU/GPUTracking/DataTypes/GPUDataTypes.h
```diff
@@ -211,7 +212,10 @@ struct ConstPtr {
 
 template <template <typename T> class S>
 struct GPUCalibObjectsTemplate {
+  float transformScale = 0;
```
davidrohr: The transformScale must not be in here, this struct must contain only pointers to POD or to FlatObjects. Could you move it to CorrectionMapsHelper? And perhaps leave a comment in this struct to only have pointers to either POD or FlatObjects in here?
(The correctionMapsHelper will be a special case. It is not a POD but has to pointers, but these point to 2 flatobjects inside the struct, so will be fine with some manual handling.)

GPU/TPCFastTransformation/CorrectionMapsHelper.cxx
```diff
@@ -0,0 +1,69 @@
...
+// or submit itself to any jurisdiction.
+
+#include "CorrectionMapsHelper.h"
+
+using namespace o2::gpu;
+
+//________________________________________________________
+void CorrectionMapsHelper::clear()
+{
+#ifndef GPUCA_GPUCODE_DEVICE
```
davidrohr: The cxx will not be compiled for GPU anyway, so you could remove the protections here.

davidrohr (changes requested):

davidrohr:
> Hi @davidrohr
> 
> Here is it. The code would be more robust and future changes easier if we could get rid of https://github.com/shahor02/AliceO2/blob/a53471588e2bb7561a09c5a6cbc5577c9dc8b62f/GPU/GPUTracking/DataTypes/GPUDataTypes.h#L216-L217 and the fastTransform would be accessed via https://github.com/shahor02/AliceO2/blob/a53471588e2bb7561a09c5a6cbc5577c9dc8b62f/GPU/GPUTracking/DataTypes/GPUDataTypes.h#L218. Do you see anything against this?

@shahor02 : I'd like all objects in this struct to be either POD or derived from FlatObject. As we said we don't want to make the CorrectionHelper a flatObject, I would at least need to keep the pointers to the FlatObjects https://github.com/shahor02/AliceO2/blob/a53471588e2bb7561a09c5a6cbc5577c9dc8b62f/GPU/GPUTracking/DataTypes/GPUDataTypes.h#L216-L217 inside.

I totally agree on using the fastTransform only via the https://github.com/shahor02/AliceO2/blob/a53471588e2bb7561a09c5a6cbc5577c9dc8b62f/GPU/GPUTracking/DataTypes/GPUDataTypes.h#L218, but I'd still like to have the other pointer. We could rename them into something `...Internal` or so.

GPU/GPUTracking/DataTypes/GPUDataTypes.h
shahor02: thanks for spotting, this is actually a leftover from an older version, not used.

PR #9941, opened 2022-09-28
ZDC - Fix centroid and digitization, add o2-zdc-digits-writer

cortesep:


Detectors/ZDC/workflow/src/digits-writer-workflow.cxx
```diff
@@ -0,0 +1,46 @@
...
+
+using namespace o2::framework;
+
+// ------------------------------------------------------------------
+// we need to add workflow options before including Framework/runDataProcessing
+void customize(std::vector<ConfigParamSpec>& workflowOptions)
+{
+  // option allowing to set parameters
+  std::vector<ConfigParamSpec> options{
+    {"disable-mc", o2::framework::VariantType::Bool, false, {"disable MC propagation even if available"}},
```
shahor02: You are defining this option but then pass hardcoded `false` to getZDCDigitWriterDPLSpec, why? I this way you will not be able to read MC data with labels.

cortesep: The program is intended to be piped to a CTF reader in order convert experimental CTF data to digits. This was useful in order to inspect them manually to understand reconstruction issues. No MC truth data is therefore available. I removed the option in order not to make confusion.

lucaquaglia15 (approved):

shahor02 (approved):

PR #9942, opened 2022-09-28
TRD calib: Protection against NaN input

martenole:
Changing the logic to avoid a segfault in case `impactAngle == NaN`.
`std::fabs(NaN) ` returns `NaN` and >= comparison with `NaN` is always false, so it would try to fill an invalid bin. So instead of testing whether `impactAngle` is invalid I am testing now whether `impactAngle` is valid.

### DIFF
--- a/DataFormats/Detectors/TRD/src/AngularResidHistos.cxx
+++ b/DataFormats/Detectors/TRD/src/AngularResidHistos.cxx
@@ -31,14 +31,14 @@ bool AngularResidHistos::addEntry(float deltaAlpha, float impactAngle, int chamb
   // add entry for given angular residual
   // returns 0 in case of success (impact angle is in valid range)
   int chamberOffset = chamberId * NBINSANGLEDIFF;
-  if (std::fabs(impactAngle) >= MAXIMPACTANGLE) {
-    LOG(debug) << "Under-/overflow entry detected for impact angle " << impactAngle;
-    return 1;
-  } else {
+  if (std::fabs(impactAngle) < MAXIMPACTANGLE) {
     int iBin = (impactAngle + MAXIMPACTANGLE) * INVBINWIDTH;
     mHistogramEntries[chamberOffset + iBin] += deltaAlpha;
     ++mNEntriesPerBin[chamberOffset + iBin];
     ++mNEntriesTotal;
+  } else {
+    LOG(debug) << "Under-/overflow entry detected for impact angle " << impactAngle;
+    return 1;
   }
   return 0;
 }


shahor02:
But why `nan` appears at all?

martenole:
I don't know, it happens very rarely. Today it happened in high rate synthetic tests made by Johannes and Massimo. I also have
```
    if (std::isnan(trkWork.getSnp())) {
      LOG(alarm) << "Track with invalid parameters found: " << trkWork.getRefGlobalTrackId();
      continue;
    }
```
in the calibration before, but I haven't seen that warning in the InfoLogger yet. When I checked last time I saw this NaN in some track parameters after GPU track refitting

DataFormats/Detectors/TRD/src/AngularResidHistos.cxx
```diff
@@ -31,14 +31,14 @@ bool AngularResidHistos::addEntry(float deltaAlpha, float impactAngle, int chamb
...
-    LOG(debug) << "Under-/overflow entry detected for impact angle " << impactAngle;
-    return 1;
-  } else {
+  if (std::fabs(impactAngle) < MAXIMPACTANGLE) {
     int iBin = (impactAngle + MAXIMPACTANGLE) * INVBINWIDTH;
     mHistogramEntries[chamberOffset + iBin] += deltaAlpha;
     ++mNEntriesPerBin[chamberOffset + iBin];
     ++mNEntriesTotal;
+  } else {
+    LOG(debug) << "Under-/overflow entry detected for impact angle " << impactAngle;
```
f3sch, line 40: Not sure how meaningful logging nan here really is, maybe also catching before hand getting the Snp which caused this in TrackBasedCalib.cxx?
Something like this:
```
      float trkAngle = o2::math_utils::asin(trkWork.getSnp()) * TMath::RadToDeg();
      if (std::isnan(trkAngle)) {
        LOG(debug) << "Caught NaN value form asin domain error: Snp=" << trkWork.getSnp();
      }
```

f3sch:
Not sure if this is relevant at all but I tried to find out what could cause this error and the only thing that stood out to me were these lines in TrackParametrization.cxx:
```
  value_t crv = (math_utils::detail::abs<value_T>(b) < constants::math::Almost0) ? 0.f : getCurvature(b);
  value_t x2r = crv * dx;
  value_t f1 = getSnp(), f2 = f1 + x2r;
  if ((math_utils::detail::abs<value_T>(f1) > constants::math::Almost1) || (math_utils::detail::abs<value_T>(f2) > constants::math::Almost1)) {
    return false;
  }
// ...
  mP[kY] += dx * dy2dx;
  mP[kSnp] += x2r;
```

Compared to TrackParametrizationWithError.cxx:
```
  value_t crv = this->getCurvature(b);
  value_t x2r = crv * dx;
  value_t f1 = this->getSnp(), f2 = f1 + x2r;
  if ((gpu::CAMath::Abs(f1) > constants::math::Almost1) || (gpu::CAMath::Abs(f2) > constants::math::Almost1)) {
    return false;
  }
// ...
  dP[kY] = dx * dy2dx;
  dP[kSnp] = x2r;
```
So in the former we add to Snp in the latter we just set it.
Which seems odd.
Take this with a heap of salt, I do not understand what happens here fully.

DataFormats/Detectors/TRD/src/AngularResidHistos.cxx
martenole: The check for nan is in another place as I wrote above. Here I actually just check if the impact angle is in the valid range, which it can easily not be for more curved tracks with lower pT. The change I hope just gives a better protection from crashes, it's not meant for debugging the actual issue

martenole:
Hi Felix,

thanks a lot for looking into this. In the two functions you are looking at the calculations are done a bit differently. In `GPUd() bool TrackParametrizationWithError<value_T>::propagateTo(value_t xk, value_t b)` the changes to the track parameters are calculated first, stored in `dP` and applied a bit below  in `this->updateParams(dP); // apply corrections`.
In `GPUd() bool TrackParametrization<value_T>::propagateParamTo(value_t xk, value_t b)`, which should be used for small distances only as its written in the comment, the track parameters are directly updated.

Unfortunately, I do not have a reproducer anymore. Yesterday, when running the FST with our current synthetic dataset the error appeared actually once. But when I reran again it was not there anymore. Without a reproducer it will be hard to track down I fear.

shahor02:
> So in the former we add to Snp in the latter we just set it.
> Which seems odd.
> Take this with a heap of salt, I do not understand what happens here fully.

Hi, both methods in principle should be used for small steps (over which the field can be considered constant since it is queried only once at  the beginning of the propagation). As @martenole wrote, the `TrackParametrization` updates the parameters directly,
while the `TrackParametrizationWithError` first creates the increments to the param vector and then applies them via `updateParams` method. The reason is that the `mP` vector is a private member of the `TrackParametrization`.

f3sch:
Thank you both for explaining!

PR #9944, opened 2022-09-28
DPL: move responsibility to get services

ktf:
Responsibility to access services is now moved to
the newly introduced ServiceRegistryRef. This will cache the information about the current stream / thread (and in the future the multiplexed DataProcessor) and it will eventually avoid concurrent access to the shared services.


ktf:
Merged as part of a different PR. Closing

PR #9953, opened 2022-09-29
[EMCAL-789] Fix namespace of streaming operator for EMCAL reco params

mfasDa:


mfasDa:
Explicit namespace needed for streaming operator, otherwise the definition will be put in the global namespace leading to undefined symbol.

mfasDa:
Errors are unrelated

shahor02 (approved):

PR #9954, opened 2022-09-29
parameter to explicitly allow parent access

jgrosseo:


jgrosseo:
@ktf @pzhristov This should avoid using it "by accident"

ktf (approved):

PR #9955, opened 2022-09-29
TRD add noise calibration

martenole:
Barebone implementation for TRD pad noise and status information.
Added one device `NoiseCalibSpec` which should run on the EPNs and extract the relevant information per pad from the digits and one device `NoiseCalibAggregatorSpec` which should run on the aggregator (EPN calibration node) and create from the pad information the relevant CCDB objects.
 @Archita-Dash

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/NoiseCalibration.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/NoiseCalibration.h
@@ -16,6 +16,7 @@
 #include <bitset>
 #include "DataFormatsTRD/Constants.h"
 #include "DataFormatsTRD/Tracklet64.h"
+#include "DataFormatsTRD/Digit.h"
 
 namespace o2
 {
@@ -52,11 +53,21 @@ class NoiseStatusMCM
   bool getIsNoisy(int mcmIdxGlb) const { return mNoiseFlag.test(mcmIdxGlb); }
   auto getNumberOfNoisyMCMs() const { return mNoiseFlag.count(); }
   bool isTrackletFromNoisyMCM(const Tracklet64& trklt) const { return getIsNoisy(trklt.getHCID(), trklt.getROB(), trklt.getMCM()); }
+  bool isDigitFromNoisyMCM(const Digit& d) const { return getIsNoisy(d.getHCId(), d.getROB(), d.getMCM()); }
 
  private:
   std::bitset<constants::MAXHALFCHAMBER * constants::NMCMHCMAX> mNoiseFlag{};
 
   ClassDefNV(NoiseStatusMCM, 1);
+};
+
+struct PadAdcInfo {
+  // Struct for holding the relevant ADC information
+  // This is what is send to the aggregator
+  void fill(const gsl::span<const PadAdcInfo> input) {}
+  void merge(const PadAdcInfo* prev) {}
+  void print() {}
+  ClassDefNV(PadAdcInfo, 1);
 };
 
 } // namespace trd
--- a/Detectors/TRD/calibration/macros/checkNoisyMCMs.C
+++ b/Detectors/TRD/calibration/macros/checkNoisyMCMs.C
@@ -106,11 +106,12 @@ void checkNoisyMCMs(std::string inpFile = "trdtracklets.root", bool useMean = tr
   auto n = nTrkltsPerActiveMcm.size() / 2;
   auto median = nTrkltsPerActiveMcm[n]; // the distinction between odd/even number of entries in the vector is irrelevant
   mean90 /= nActiveMcms90;
-  auto noiseThreshold = (useMean) ? 100.f * mean90 : (float)totalTriggerCounter / 10.f;
+  auto noiseThreshold = (useMean) ? 30.f * mean90 : (float)totalTriggerCounter / 5.f;
 
   setlocale(LC_NUMERIC, "en_US.utf-8");
   printf("Info: Found in total %'lu MCMs which sent tracklets with a median of %i tracklets per MCM\n", nTrkltsPerActiveMcm.size(), median);
   printf("Info: Excluding the 10%% of MCMs which sent the highest number of tracklets %'i MCMs remain with on average %.2f tracklets per MCM\n", nActiveMcms90, mean90);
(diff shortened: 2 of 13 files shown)


DataFormats/Detectors/TRD/include/DataFormatsTRD/NoiseCalibration.h
```diff
@@ -52,13 +53,23 @@ class NoiseStatusMCM
   bool getIsNoisy(int mcmIdxGlb) const { return mNoiseFlag.test(mcmIdxGlb); }
   auto getNumberOfNoisyMCMs() const { return mNoiseFlag.count(); }
   bool isTrackletFromNoisyMCM(const Tracklet64& trklt) const { return getIsNoisy(trklt.getHCID(), trklt.getROB(), trklt.getMCM()); }
+  bool isDigitFromNoisyMCM(const Digit& d) const { return getIsNoisy(d.getHCId(), d.getROB(), d.getMCM()); }
 
  private:
```
f3sch, line 58: Don't you need to update ClassDefNV?

Detectors/TRD/workflow/src/trd-calib-workflow.cxx
```diff
@@ -33,10 +37,23 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
...
+    specs.emplace_back(getTRDVdAndExBCalibSpec());
   }
-  specs.emplace_back(getTRDVdAndExBCalibSpec());
+
+  if (configcontext.options().get<bool>("noise")) {
+    if (enableRootInp) {
+      specs.emplace_back(o2::trd::getTRDDigitReaderSpec(false));
+      specs.emplace_back(o2::trd::getTRDNoiseCalibSpec());
+    }
+    // specs.emplace_back(o2::trd::getTRDNoiseCalibAggregatorSpec());
```
f3sch, line 55: Just checking if this a commenting fluke. Probably not, right?

Detectors/TRD/calibration/macros/checkNoisyMCMs.C
```diff
@@ -106,11 +106,12 @@ void checkNoisyMCMs(std::string inpFile = "trdtracklets.root", bool useMean = tr
   auto n = nTrkltsPerActiveMcm.size() / 2;
   auto median = nTrkltsPerActiveMcm[n]; // the distinction between odd/even number of entries in the vector is irrelevant
   mean90 /= nActiveMcms90;
-  auto noiseThreshold = (useMean) ? 100.f * mean90 : (float)totalTriggerCounter / 10.f;
+  auto noiseThreshold = (useMean) ? 30.f * mean90 : (float)totalTriggerCounter / 5.f;
```
f3sch, line 109: Why are you tightening the definition of a noisyMCM, but at the same time loosening if useMean=false?
So in the first case, a noisyMCM is now one which produced 30x the mean of the lower 90 perecent.
In the second case, a noisyMCM is one which triggered on twice as much as before.

DataFormats/Detectors/TRD/include/DataFormatsTRD/NoiseCalibration.h
martenole: That is only required if the class layout changes. Since I am not adding/removing a class member its not needed

Detectors/TRD/calibration/macros/checkNoisyMCMs.C
martenole: Am I not tightening both? if useMean==false then a MCM which fired for at least 20% of the triggers is noisy, before it was only in case it fired 10% of the time. And I did that, because with the previous settings it looked to me we were not strict enough (looking at how many tracklets were sent by MCMs which were previously not marked noisy in comparison to the ones marked as noisy). I am trying to find the spot were there is a sudden jump for the number of tracklets sent per MCM

Detectors/TRD/workflow/src/trd-calib-workflow.cxx
martenole: Sure, but this is also WIP and some things can still be found by @Archita-Dash who should add to this the actual calibration part ;)

Detectors/TRD/calibration/macros/checkNoisyMCMs.C
f3sch: I looked a bit further into this and now I am not entirely convinced this is as effective as we think (given I do not fully understand what _really_ is a `noisy` MCM.

Below I plotted the _sorted_ vector with trackletCounter[i].first as bin content e.g. the number of tracklets. (This is to say that the shown xaxis is not the actual mcm number but in fact the idx after sort, hence should be disregarded for my argument, also the xaxis is zoomed since most of the mcm produce roughly an even number of tracklets, i used 10000tfs)
Additionally I drew the value for noiseThreshold as a red line.

There seems to be this small tail under red line we cannot really get rid of using these methods.
Like 200 MCMs.
Is this acceptable or should we tighten this even more?
Also below is the same for another Run with different triggers, there it looks like roughly 500 MCM in this tail.

526055,useMean=true:
image

526055,useMean=false:
image

Reference Run 523677, depends on triggers:
image

---
This is how these plots look not zoomed in, most of the stuff happens at the end:
image

martenole: Hi Felix,
in the end its not really MCMs which are noisy, at least as far as I can tell, but individual pads instead. See for example my comment here: https://github.com/AliceO2Group/AliceO2/pull/7615#issuecomment-970052729
Once we are able to take proper noise runs we should be able to detect noisy pads automatically and mask them. Afterwards no noisy MCM should remain.
The noise threshold we have set on the level as we are doing now is always kind of arbitrary. But this should be OK, since its used only for QC to clear the spectra so that we are able to see other peculiarities of the data. Without it e.g. the plots for charges would be polluted by the tracklets from only few MCMs.
I don't think it is worth to put much effort into optimizing the noisy MCM finder further. Instead, we should focus on the pad level analysis which @Archita-Dash is looking into

martenole:
Noise calibration is still dummy and won't work. But this could be merged in any case to prepare for the gain calibration which @f3sch is going to implement

PR #9956, opened 2022-09-29
[EMCAL-838] Load RecoParams from the CCDB in EMCAL reconstruction workflow

mfasDa:
- RecoParams are by default loaded from the CCDB
- Option to disable loading from CCDB available in
  RawToCellConverterSpec (will take the default
  parameters in the constructor)

### DIFF
--- a/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
@@ -17,6 +17,7 @@
 
 #include "CommonConstants/Triggers.h"
 #include "CommonDataFormat/InteractionRecord.h"
+#include "Framework/CCDBParamSpec.h"
 #include "Framework/ConfigParamRegistry.h"
 #include "Framework/ControlService.h"
 #include "Framework/InputRecordWalker.h"
@@ -92,8 +93,6 @@ void RawToCellConverterSpec::init(framework::InitContext& ctx)
   mDisablePedestalEvaluation = ctx.options().get<bool>("no-evalpedestal");
 
   LOG(info) << "Running gain merging mode: " << (mMergeLGHG ? "yes" : "no");
-  LOG(info) << "Using time shift: " << RecoParam::Instance().getCellTimeShiftNanoSec() << " ns";
-  LOG(info) << "Using BCshfit phase:" << RecoParam::Instance().getPhaseBCmod4() << " BCs";
   LOG(info) << "Using L0LM delay: " << o2::ctp::TriggerOffsetsParam::Instance().LM_L0 << " BCs";
 
   mRawFitter->setAmpCut(mNoiseThreshold);
@@ -103,6 +102,9 @@ void RawToCellConverterSpec::init(framework::InitContext& ctx)
 void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
 {
   LOG(debug) << "[EMCALRawToCellConverter - run] called";
+  // for reading the reco param object from the ccdb
+  ctx.inputs().get<o2::emcal::RecoParam*>("EMC_RecoParam");
+
   double timeshift = RecoParam::Instance().getCellTimeShiftNanoSec(); // subtract offset in ns in order to center the time peak around the nominal delay
   constexpr auto originEMC = o2::header::gDataOriginEMC;
   constexpr auto descRaw = o2::header::gDataDescriptionRawData;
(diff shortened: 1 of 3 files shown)


mfasDa:
WIP - don't merge yet!

mfasDa:
@shahor02 Can you please have a look if the changes look good? I tested it locally and the parameters were loaded as expected.

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -591,6 +602,16 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
   sendData(ctx, mOutputCells, mOutputTriggerRecords, mOutputDecoderErrors);
 }
 
+void RawToCellConverterSpec::finaliseCCDB(o2::framework::ConcreteDataMatcher& matcher, void* obj)
+{
+  // check if calib params need to be updated
+  if (matcher == framework::ConcreteDataMatcher("EMC", "EMCALRECOPARAM", 0)) {
+    LOG(info) << "RecoParams updated";
+    LOG(info) << o2::emcal::RecoParam::Instance();
```
shahor02: Why don't you use generic method `o2::emcal::RecoParam::Instance().printKeyValues(true, true);` ?
The 2nd true will ensure that the output goes to the LOG(info) instead of the stdout. The advantage is that also the provenance of the data member will be reported (e.g. in case you override some parameter from the command line).

Also, with your streaming operator I think only 1st line (data member) will be preceded by the [INFO], did you check it?

Detectors/EMCAL/workflow/src/emc-reco-workflow.cxx
```diff
@@ -53,6 +53,7 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
     {"configKeyValues", o2::framework::VariantType::String, "", {"Semicolon separated key=value strings"}},
     {"disable-mc", o2::framework::VariantType::Bool, false, {"disable sending of MC information"}},
     {"disable-decoding-errors", o2::framework::VariantType::Bool, false, {"disable propagating decoding errors"}},
+    {"no-loadrecoparamfromccdb", o2::framework::VariantType::Bool, false, {"disable loading RecoParam object from CCDB"}},
```
shahor02: Do you really need this option? Note that the values passed via `--configKeyValues ...` will override corresponding data-members in the object loaded from the CCDB.

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
mfasDa: I was not aware of this, I will try this and modify the PR accordingly.

Detectors/EMCAL/workflow/src/emc-reco-workflow.cxx
mfasDa: We have tested this already for the calib params (which also have a no-load... option). The plan is indeed to fully rely on the CCDB, the option was just a fallback for testing. I can remove it again, will simplify a bit the code.

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
mfasDa: Code is changed to use `o2::emcal::RecoParam::Instance().printKeyValues(true, true);`, thanks a lot for the hint!

Detectors/EMCAL/workflow/src/emc-reco-workflow.cxx
mfasDa: I removed the option to disable loading from the CCDB, indeed I agree it is not really needed.

shahor02: You removed check for this option but the option definition is still there.

mfasDa:
CI errors unrelated

Detectors/EMCAL/workflow/src/emc-reco-workflow.cxx
mfasDa: Done

mfasDa:
For the record: I have tested modifying the parameters separately via the command line and they were handled as expected, meaning the subset of values modified manually got overwritten while the other values were the ones store in the CCDB. All expected features can be considered as tested for this PR.

shahor02 (approved):

PR #9959, opened 2022-09-29
[EMCAL-791] Adapting the SimParam to be a CCDB object

hahassan7:
The EMCAL SimParam has been adapted to work as a CCDB object, so we can store some MC configuration in the CCDB and load them when needed.


mfasDa (approved):
Looks good!

mfasDa:
Needed for https://alice.its.cern.ch/jira/browse/O2-3202

sawenzel (approved):

PR #9964, opened 2022-09-30
Remove dependency on FairRoot::Base from Common/MathUtils

ktf:
Avoid loading Proof as a core dependency.


ktf:
Issues in O2Physics fixed. Issue with testMonitorObject unrelated. Merging.

PR #9967, opened 2022-09-30
AOD: Optimization of search in bunch crossings (fillBCSlice)

sawenzel:
fillBCSlice was doing two binary searches in the std::map holding (sorted) bunch crossing timestamps.

Since this function is called for many tracks T, the cost of fillBCSlice actually dominated the processing time for AOD creation in async reconstruction (O(T*log(N)), where N is the number of bunch crossings).

Based on this observation, it makes sense to setup a dedicated lookup structure that makes searching in the bunch crossings faster.

This commit provides such a structure, which follows the idea to bucketize bunch crossing times into cells (or time windows). In addition, cells know about their nearest filled neighbours for very fast navigation. So to get bunch crossing close to some time "t", one merely calculates an appropriate cell and just searches within the times located in that cell.

### DIFF
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -56,6 +56,154 @@ using DataRequest = o2::globaltracking::DataRequest;
 namespace o2::aodproducer
 {
 
+/// A structure or container to organize bunch crossing data of a timeframe
+/// and to facilitate fast lookup and search within bunch crossings.
+class BunchCrossings
+{
+ public:
+  /// Constructor initializes the acceleration structure
+  BunchCrossings() {}
+
+  /// initialize this container (to be ready for lookup/search queries)
+  void init(std::map<uint64_t, int> const& bcs)
+  {
+    clear();
+    // init the structures
+    for (auto& key : bcs) {
+      mBCTimeVector.emplace_back(key.first);
+    }
+    initTimeWindows();
+  }
+
+  /// return the sorted vector of increaing BC times
+  std::vector<uint64_t> const& getBCTimeVector() const { return mBCTimeVector; }
+
+  /// Performs a "lower bound" search for timestamp within the bunch crossing data.
+  /// Returns the smallest bunch crossing (index and value) equal or greater than timestamp.
+  /// The functions is expected to perform much better than
+  /// a binary search in the bunch crossing data directly. Expect O(1) instead of O(log(N)) at the cost
+  /// of the additional memory used by this class.
+  std::pair<size_t, uint64_t> lower_bound(uint64_t timestamp) const
+  {
+    // a) determine the timewindow
+    const auto NofWindows = mTimeWindows.size();
+    const auto smallestBC = mBCTimeVector[0];
+    const auto largestBC = mBCTimeVector.back();
+    auto timeindex = std::max((int)0, (int)((timestamp - smallestBC) / mWindowSize));
+
+    if (timeindex >= NofWindows) {
+      // there is no next greater; so the bc index is at the end of the vector
+      return std::make_pair<int, uint64_t>(mBCTimeVector.size(), 0);
+    }
+
+    const auto* timewindow = &mTimeWindows[timeindex];
+    while (timeindex < NofWindows && (!timewindow->isOccupied() || mBCTimeVector[timewindow->to] < timestamp)) {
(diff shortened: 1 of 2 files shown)


sawenzel:
@shahor02 @davidrohr : This addresses the observation made during async reco. Some adjustments may still be necessary. Not sure if I completely replicated the iterator +- logic that was there previously (but in all tests I got the same results so far).

Also we need to decide whether to use mBCLookup as a class member variable or if we want to pass it to the fillBCSlice function as an argument.

shahor02:
Hi @sawenzel 

Most of the CPU should be spent on calls like `bcsMap.find(globalBC);` 
Given that in the TF we can have at most 456192 = 128*3564  global BC, I am wondering if it will not be better to map the original map in the collectBCs to plain vector and use it for direct lookup, i.e.
```
std::vector<int> mBCLookup(MaxHB*LHCMaxBunches,-1);
auto mOffset = mStartIR.toLong();
for (auto& key : bcs) {
  mBCLookup[key.first - mOffset] = key.second;
}
```
Then, all calls like 
```
auto item = bcsMap.find(globalBC);
    int bcID = -1;
    if (item != bcsMap.end()) {
      bcID = item->second;
    } else {
      LOG(fatal) << "Error: could not find a corresponding BC ID for a collision; BC = " << globalBC << ", collisionID = " << collisionID;
    }
```
can be substituted by
```
int bcID = mBCLookup[globalBC - mOffset];
if (bcID<0) {
   LOG(fatal)  ...
}
```

davidrohr:
the lookup table for the BC seems quite appealing to me. Memory usage would be quite moderate.

sawenzel:
@shahor02 : So far, analysis with perf in async reco has shown that fillBCSlice was using close to 100% of the CPU time. The present idea removes this hotspot and for this we need a sorted container of BC times. The sorting is constructed/maintained automatically by constructing the BCs with a std::map.

For what concerns the remaining lookups to this std::map, I completely agree that we can also make them faster (and not O(log(N)) and I was planning to do this in a future improvement.

shahor02:
Hi @sawenzel 

The same linear vector can be used also for lower_bound and upper_bound direct lookup, one should just fill the unused elements by the reference on the next used one. Something like this:
```
  const int MaxRange = MaxHB*LHCMaxBunches;
  std::vector<int> mBCLookup(MaxRange,-1);

  void init(std::map<uint64_t, int>& bcsMap) {
    auto mOffset = mStartIR.toLong();
    for (auto& key : bcs) {
      mBCLookup[key.first - mOffset] = key.second;
    }    
    int up = MaxRange+1; // equivalent of end() iterator
    for (int i=MaxRange;i--;) { // fill lookup for lower_bound
      if (mBCLookup[i]<0) {
	mBCLookup[i] = -up; // store -(1+entry_of_upper_bound)
      } else {
	up = i+1; // update entry of upper_bound
      }
    }
  }

  std::pair<size_t, uint64_t> lower_bound(uint64_t globalBC) const
  {
    int loc = globalBC < mOffset ? 0 : (globalBC - mOffset);
    if (loc >= MaxRange) {
      return {-1, -1UL}; // equivalent of end() iterator
    }
    int v = mBCLookup[globalBC - mOffset];
    if (v<0) {
      v = -(v+1); // recover reference index
      return {mBCLookup[v], v+mOffset};
    }
    return {v, globalBC};
  }
  
  std::pair<size_t, uint64_t> upper_bound(uint64_t globalBC) const
  {
    return lower_bound(globalBC+1);
  }

  std::pair<size_t, uint64_t> find(uint64_t globalBC) const
  {
    int loc = globalBC < mOffset ? 0 : (globalBC - mOffset);
    if (loc<0 || loc>=MaxRange || mBCLookup[loc]<0) {
      return {-1, -1UL};
    }
    return {mBCLookup[loc], globalBC};
  }

```

sawenzel:
@shahor02: Are you suggesting that this PR be changed and implemented differently or is your comment about further optimizations? As I was saying, I would like to decouple the change of fillBCSlice with other improvements. But in principle this should be driven by a profile.

shahor02:
@sawenzel I was suggesting a faster implementation of your BunchCrossing helper class, which avoids any search or std::map operations and provides the functionality needed for the aod-producer, particularly by the fillBCSlice: methods find, lower/upper_bound.

davidrohr:
This works for me in the async tests, and it is already such a large improvements that we should have it in. We can still try to improve it further in a follow-up PR. Thx @sawenzel for looking into this!

sawenzel:
Ok. I will attack further improvements (such as described by @shahor02) in another round.

PR #9968, opened 2022-09-30
Fix rotation of station 3 when in dipole

javierecc:


sawenzel:
Can you give slightly more detail please? Was this wrong  in the current geometry? How did you find out?

javierecc:
Sure. 
Station 3 of MCH  is placed inside the dipole volume. The dipole (DDIP) is coded with z>0 convention and then rotated. MCH geometry is coded with z<0 convention. So when the station 3 is placed inside the dipole (not the case if creating a test standalone geometry) one must rotate it by 180 degrees around y. The way it was done was wrong because it neglected the fact that stations 3,4,5 have a slight tilt around x (angle between the beam pipe and the floor). 
The problem was spotted when working on mch alignment, when going from global to local coordinates, the local z coordinate was different than zero for all detection elements of chambers 5 and 6 (station 3).
So bottom line, in the current geometry the chambers of station 3 are tilted but in the opposite direction.

aphecetche (approved):

PR #9969, opened 2022-09-30
Protect MeanVertex calibrator against upload of wrong values

shahor02:


Detectors/Calibration/src/MeanVertexCalibrator.cxx
```diff
@@ -320,7 +332,9 @@ void MeanVertexCalibrator::finalizeSlot(Slot& slot)
   mTmpMVobjDqTime.emplace_back(slot.getStartTimeMS(), slot.getEndTimeMS());
```
chiarazampolli: This should not be done if the fit return value is false. It should be moved below

Detectors/Calibration/src/MeanVertexCalibrator.cxx
```diff
@@ -201,21 +211,18 @@ void MeanVertexCalibrator::fitMeanVertex(o2::calibration::MeanVertexData* c, Mea
 
   // now we update the error on x
   double sumX = 0, sumY = 0, weightSumX = 0, weightSumY = 0;
-  for (int iFit = 0; iFit < mNPointsForSlope; ++iFit) {
+  for (int iFit = 0; iFit < nBinsOK; ++iFit) {
```
chiarazampolli, line 230: Should we check somewhere if `nBinsOk` is > `mNPointsForSlope`?

shahor02: It cannot be ` >mNPointsForSlope`, it is incremented in the loop `0:mNPointsForSlope` for successful fits.

Detectors/Calibration/src/MeanVertexCalibrator.cxx
shahor02: right, fixed.

Detectors/Calibration/src/MeanVertexCalibrator.cxx
chiarazampolli: I expressed myself wrongly: what if we have less nBinsOk than the needed number to calculate the slope?

shahor02: `nBinsOK < 2` leads to failed fit

chiarazampolli: ok, thanks.

martenole (dismissed):
Thanks! At the moment we are running the mean vertex calibration without this additional protection. I hope we don't hit the issue of last night again, otherwise we will disable until this is employed at P2 I'd suggest

shahor02:
I was not able to reproduce the failing fits in the mean-vertex calibrator neither with small statistics I could process locally nor with high stat. fake data. Adding an option to dump non-empty bins to check what is wrong with fits and another one to skip sending objects to populator.
We will need to take a short run with `CONFIG_EXTRA_PROCESS_o2_calibration_mean_vertex_calibration_workflow="MeanVertexCalib.dumpNonEmptyBins=true;MeanVertexCalib.skipObjectSending=true"`.

chiarazampolli:
> I was not able to reproduce the failing fits in the mean-vertex calibrator neither with small statistics I could process locally nor with high stat. fake data. Adding an option to dump non-empty bins to check what is wrong with fits and another one to skip sending objects to populator. We will need to take a short run with `CONFIG_EXTRA_PROCESS_o2_calibration_mean_vertex_calibration_workflow="MeanVertexCalib.dumpNonEmptyBins=true;MeanVertexCalib.skipObjectSending=true"`.

Hi Ruben, 
The verbose mode prints the content of the bins already. 
Maybe that would then be too verbose for you? It prints it such that you can just copy-paste to create a histogram, or a vector. 
To debug, we can check if the async reco that I asked Catalin to start can help. 
A short run might not be useful, since the problem does not always appear.
Chiara

shahor02:
Indeed, the verbose mode is too verbose as it will print millions of vertices, we can not run it on the epn

Detectors/Calibration/src/MeanVertexCalibrator.cxx
```diff
@@ -236,13 +259,17 @@ void MeanVertexCalibrator::fitMeanVertex(o2::calibration::MeanVertexData* c, Mea
     LOG(info) << "SigmaX for MeanVertex = " << sigmaX;
     LOG(info) << "SigmaY for MeanVertex = " << sigmaY;
   }
+  if (sigmaX == 0 || sigmaY == 0 || mvo.getSigmaZ() == 0 || nBinsOK < 2) {
```
martenole, line 262: @shahor02 To be on the safe side would it not be better to do the same check as in the primary vertexer? So test for `sigmaX * sigmaX * sigmaY * sigmaY < PVertexer::kAlmost0D`?

shahor02: Hi @martenole 
The check I put prevents from the object update if one of the values above were not set, in this case they will be strictly 0, the fit itself cannot produce value !=0 but < kAlmost0D.

martenole: Ok, thanks for explaining. Then we can reinclude the mean vertex calibration after the next software update

PR #9972, opened 2022-10-01
Fix missing dependency

ktf:


ktf:
Not sure how come the builds do work but the CI does not.

ktf:
Ok, apparently this is only used in O2Physics. Merging and rebuilding daily, actually.

PR #9974, opened 2022-10-01
Drop need for Graf3d in detectors Align

ktf:


ktf:
@shahor02 @chiarazampolli any reason for the dependency?

shahor02:
no, this is a leftover of Aliroot dependencies

PR #9980, opened 2022-10-03
[EMCAL-565, EMCAL-566] Fix CCDB access for calibrators

mfasDa:
- Trigger loading scalefactors for bad channel calib in
  run function
- Display calib params in finalizeCCDB
- Move configuration of the calibrators based on
  calibParams from init to dedicated configure function

### DIFF
--- a/Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
+++ b/Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
@@ -63,10 +63,6 @@ class EMCALChannelCalibDevice : public o2::framework::Task
         mTimeCalibrator = std::make_unique<o2::emcal::EMCALChannelCalibrator<o2::emcal::EMCALTimeCalibData, o2::emcal::TimeCalibrationParams>>();
       }
       mTimeCalibrator->SetCalibExtractor(mCalibExtractor);
-      mTimeCalibrator->setSlotLength(EMCALCalibParams::Instance().slotLength_tc);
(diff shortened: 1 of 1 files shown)


mfasDa:
Extending #9909: The scale factors were assigned to the correct calibrator, however the loading of the scale factors was not triggered since the corresponding `pc.inputs().get<...>...` was not called in the run function, which consequently didn't trigger finalizeCCDB. The trigger is added only in case of bad channel mode for scaling enabled (tested offline and working).

When adding these lines I found however a second problem: The EMCALCalibParams were used in the init function to configure the calibrators. Since we take the CalibParams now from the CCDB the values from the CCDB are not available during the init stage, leading to the default values being used. The configure part has been moved to a dedicated configure functions called for the first time only in the run function after the CCDB access. At that stage the updated values from the CCDB are available.

mfasDa:
The rescaling is important since the current mask has visible structures from the TRD support structures. In areas behind the TRD structures the energy distribution is different compared to other parts in the acceptance due to the different material budget. The algorithm realizes that the spectrum is different in those regions and therefore considers the towers as bad. The scale factors reweight the spectra based on the tower ID in order to achieve a uniform distribution where outliers really correspond to bad channels and not to differences in the material budget. The scale factors were determined in Run2, since they represent a geometric effect we expect them to be valid also in run3. Tests performed offline validated this assumption.

Mask_526080

mfasDa:
Last rebase contained additional switch to trigger the update only for the first iteration.

shahor02 (approved):

PR #9982, opened 2022-10-03
Adding number of pixel type per chip information

pdhankhe:
The output string is modified to store the information of number of different pixel flavors per chip instead of pixel id. The following example string shows the format of the output string:

O2ChipID:7379,PixelType:Noisy,PixelNos:1,DcolNos:0,O2ChipID:7362,PixelType:Dead,PixelNos:293028,DcolNos:-1,O2ChipID:7359,PixelType:Ineff,PixelNos:2048,


Detectors/ITSMFT/ITS/workflow/src/ThresholdCalibratorSpec.cxx
```diff
@@ -954,6 +955,7 @@ void ITSThresholdCalibrator::addDatabaseEntry(
 
           // Noisy pixel IDs
           pixIDs_Noisy += std::to_string(v[i]);
+          nNoisy++;
```
iravasen: Do you need `nNoisy`? It should be the same as `n_pixels = v.size()` when the pixel type is "noisy" right?

iravasen:
Thanks a lot @pdhankhe.

@shahor02 for us this can be merged (after fullCI validation)

PR #9983, opened 2022-10-03
DPL: move TPDGDatabase customization to plugin

ktf:
This will avoid that every single device links against libPhysics & co.


ktf:
@sawenzel @aalkin could you cross check this behaves as expected?

ktf:
Tests already fully passed, trivial conflict fix triggered rebuild. Tested again on hyperloop and everything seems fine. Merging.

PR #9985, opened 2022-10-04
Make DataFormats/MemoryResources header only

ktf:


ktf:
We could even move this to DataFormats/Headers, since anything which uses it also uses the latter...

PR #9987, opened 2022-10-04
[EMCAL-839]  Integration CCDB objects in the digitizer

hahassan7:
Many things were implemented here:
1. The BC phase is now calculated from the trigger record bc instead of the event time.
2. Added the BC phase swap as in data.
3. The DigitizerSpec has been adapted to load CCDB objects (SimParam)
4. The CellConverterSpec has also been adapted to load the RecoParam as a CCDB objects
5. Added a function in run() that loads only once in order to initialize the digitizer after the SimParam are loaded from the CCDB.


mfasDa (dismissed):

mfasDa:
In principle looks good and tests can be started - removing WIP label

hahassan7:
Pull request can be merged now !

mfasDa (approved):

PR #9989, opened 2022-10-04
[EMCAL-630] Send mapping errors to DecoderErrorContainer

mfasDa:
Mapping errors were missing in the decoder error container,
treated with existing error code for decoder error in
major ALTRO decoding error. As additional information the
hardware address raising the errror is sent, FEC ID
cannot be determined as the HW address is invalid.

In addition log hardware address as hex value in error
message of AddressNotFoundException


mfasDa:
We saw today mapping errors in one DDL sending hardware addresses that were not registerd for the given DDL. The errors were handled and sent to the InfoLogger where Martin found them but they were not sent to the QC for monitoring by the shift crew since the exception handler was the only one left that did not send an error code to the error container. This is fixed by the pull request - the error is sent under the already existing errror code.

shahor02 (approved):

PR #9992, opened 2022-10-04
Add asynchronous downloader that reuses connections to CCDB

TrifleMichael:
This change adds a downloader capable of parallel and asynchronous curl calls, with the ability to make callback functions after a request is finished.

A lot of the code is base on an example from the curl manual which combines libuv with libcurl (https://curl.se/libcurl/c/multi-uv.html)

The main differences between the example and this code are:
- a mechanism that closes sockets after a set period of time,
- the loop runs a parallel thread to the rest of the program.

In case of any questions or remarks please contact me at trzebuniak@student.agh.edu.pl

### DIFF
--- a/CCDB/include/CCDB/CCDBDownloader.h
+++ b/CCDB/include/CCDB/CCDBDownloader.h
@@ -1,0 +1,358 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <uv.h>
+#include <curl/curl.h>
+#include <string>
+#include <vector>
+#include <iostream>
+#include <thread>
+#include <mutex>
+#include <condition_variable>
+#include <unordered_map>
+
+#ifndef ALICEO2_CCDBDOWNLOADER_H
+#define ALICEO2_CCDBDOWNLOADER_H
+
+using namespace std;
+
+namespace o2
+{
+namespace ccdb
+{
+
+/*
+ Some functions below aren't member functions of CCDBDownloader because both curl and libuv require callback functions which have to be either static or non-member.
+ Because non-static functions are used in the functions below, they must be non-member.
+*/
+
+/**
+ * uv_walk callback which is used to close passed handle.
+ *
+ * @param handle Handle to be closed.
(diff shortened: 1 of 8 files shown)


CCDB/include/CCDB/CCDBDownloader.h
```diff
@@ -0,0 +1,342 @@
...
+  /**
+   * Timer starts for each socket when its respective transfer finishes, and is stopped when another transfer starts for that handle.
+   * When the timer runs out it closes the socket. The period for which socket stays open is defined by socketTimoutMS.
+   */
+  std::unordered_map<curl_socket_t, uv_timer_t*> socketTimerMap;
+
+  /**
+   * The UV loop which handles transfers.
+   */
+  uv_loop_t loop;
```
ktf: Please, make it settable from the outside or at least make sure this will not interfere with a separate libuv loop which will for sure be there.

ktf (changes requested):
libuv loop should not interfere with the DPL one.

CCDB/include/CCDB/CCDBDownloader.h
TrifleMichael: I have added the functionality of using an external uv_loop, which stores handles for the downloader and is cleaned of them (and only them) when the downloader is destroyed.

Could you please also elaborate what you mean by interefence between the uv_loop and the DPL one?

costing:
@ktf bumping this thread not to loose track of it

costing:
@ktf , did you have a chance at checking out the updated pr ?

CCDB/src/CcdbApi.cxx
```diff
@@ -65,6 +65,8 @@ CcdbApi::~CcdbApi()
 
 void CcdbApi::curlInit()
 {
+  std::cout << "Creating downloader\n";
```
ktf, line 68: No cout.

CCDB/src/CcdbApi.cxx
```diff
@@ -1110,7 +1113,9 @@ bool CcdbApi::isHostReachable() const
       curl_easy_setopt(curl, CURLOPT_URL, mUrl.data());
       curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_data);
       curlSetSSLOptions(curl);
-      res = curl_easy_perform(curl);
+      std::cout << "mUrl " << mUrl.data() << "\n";
```
ktf, line 1116: Please use the LOG macro

CCDB/test/testCcdbApiDownloader.cxx
```diff
@@ -0,0 +1,286 @@
...
+
+
+using namespace std;
+
+namespace o2{
+namespace ccdb{
+
+size_t writeToString(void *contents, size_t size, size_t nmemb, std::string *dst)
+{
+  char *conts = (char *)contents;
```
ktf, line 34: I am not sure if using an std::string for the payload is a good idea.

ktf:
Sorry for taking so long to have a look at this. Apart from fixing the clang-format and dropping the various cout, I think the main issue is that you are using a std::string for the payload. Why is that? Could you move to a preallocated buffer with the correct size?

PR #9993, opened 2022-10-04
load TOF geo from geometry.root

noferini:
Minor refinements needed.
Rotation and Translation (but one sign, see below) now loaded from geometry.root and equal to the ideal one (checked).

To be checked: strip misalignment in the tangent direction (sector reference frame) is zero in ideal. I have to find a way to check the if the sign of possible misalignment is traced correctly

### DIFF
--- a/Detectors/TOF/base/include/TOFBase/Geo.h
+++ b/Detectors/TOF/base/include/TOFBase/Geo.h
@@ -65,7 +65,7 @@ class Geo
   static void rotate(Float_t* xyz, Double_t rotationAngles[6]);
 
   static void rotateToSector(Float_t* xyz, Int_t isector);
-  static void rotateToStrip(Float_t* xyz, Int_t iplate, Int_t istrip);
+  static void rotateToStrip(Float_t* xyz, Int_t iplate, Int_t istrip, Int_t isector);
   static void antiRotateToSector(Float_t* xyz, Int_t isector);
 
   static void antiRotate(Float_t* xyz, Double_t rotationAngles[6]);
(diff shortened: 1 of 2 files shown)


noferini:
PR is now ready for the review.
Changes were tested in one CTF of PB22 (full outputs available on the GRID).
The comparison of matching performances  with old(ideal geo) and new(geo from o2sim_geometry) is fine.
Also the sign of X misalignment (in sec reference) was checked by introducing and artificial misalignment (1 cm) and sign is fine.

shahor02 (commented):
Hi @noferini 
Difficult to review new transformations given that I cannot understand the (simpler) previous ones :). Is this ready to be merged?
It still does not solve my problem: for the alignment framework I need a method to convert `clus.getMainContributingChannel()` to ideal coordinates in the strip local frame, to which I can apply `base::GeometryManager::getOriginalMatrix(strip->getSymName())` matrix for further transformations.

noferini:
> Hi @noferini Difficult to review new transformations given that I cannot understand the (simpler) previous ones :). Is this ready to be merged? It still does not solve my problem: for the alignment framework I need a method to convert `clus.getMainContributingChannel()` to ideal coordinates in the strip local frame, to which I can apply `base::GeometryManager::getOriginalMatrix(strip->getSymName())` matrix for further transformations.

Hi @shahor02,
this PR is ready. It simply apply the previous scheme but now reading geometry from file (then alignment is applied).
The results are the same if loading  (ideal) geo from file or pre-defined geometry.
About you request I'm looking into that and I will come with an example to be cross-checked with you.
Cheers,
Francesco

shahor02 (approved):
Hi @noferini 

Thanks for looking into such method. At the moment I managed to do what I need in an brute force way, going from the channel id to sector coordinates, transforming to local, correcting for alignment then transforming back to the sector: https://github.com/AliceO2Group/AliceO2/blob/278df1e0e6c2b82e8e3b6e020ecba05471b79288/Detectors/Align/src/AlignableDetectorTOF.cxx#L89-L108
This is too heavy and I probably lose some precision on so many transformations, but it works since for the alignment I load separately ideal geom and alignment objects, so the TOF methods I am using operate with ideal geom.

PR #9998, opened 2022-10-05
Fix CPV errors in InfoLogger

martenole:
In order to reproduce with current O2:
```
root [0] o2::cpv::CalibParams* foo = new o2::cpv::CalibParams(1)
(o2::cpv::CalibParams *) 0x80c12d0
root [1] o2::utils::MemFileHelper::getClassName(foo)
[ERROR] Could not retrieve ROOT dictionary for type PN2o23cpv11CalibParamsE
(std::string) ""
root [2] o2::utils::MemFileHelper::getClassName(*foo)
(std::string) "o2::cpv::CalibParams"
root [3] 
```
pinging just for information also to @chiarazampolli


chiarazampolli:
Thanks @martenole !

PR #9999, opened 2022-10-05
Update list of CALIB_TASKS for FST

martenole:
Currently `BARREL_SPORADIC` is empty, unless `has_processing_step TPC_DEDX` is true. This is not the case by default, so I have taken it out.
`FORWARD_TF` is by default not empty in case ZDC is present.
ping @davidrohr


davidrohr:
I'd leave `BARREL_SPORADIC` in, even if it is empty by default.
Otherwise one runs into problem when one enables DEDX.

martenole:
Actually, `FORWARD_TF` is not needed, unless ZDC_RECO is in the list of extra processing steps. So similar situation as with `BARREL_SPORADIC`. Should this still be merged?

davidrohr:
yes, let's just have the aggregator enabled by default. Worst case it is empty, doesn't really matter...

martenole:
Ok, will do. But do you know why tmux is not putting 4 equal columns in the last row for the aggregators?

tmux-layout

davidrohr:
I think they are created one after another horizontally, but not rearranged afterwards. The problem is: I don't know if you can rearrange only the lower ones in a horizontal way. By default it would rearrange everything horizontally, it would break the previous vertical arrangement. But perhaps it is possible. You can dig in the tmux docu :)

martenole:
Well, I found that I could hardcode it four 4 calibration windows with `select-layout  a2d3,282x76,0,0[282x18,0,0,0,282x18,0,19,1,282x18,0,38,2,282x19,0,57{72x19,0,57,3,73x19,73,57,4,66x19,147,57,5,68x19,214,57,6}]`. But its not bothering me so much, so I'd just leave it as it is :)

PR #10001, opened 2022-10-05
MCH: add constants header-only library

aphecetche:
The list of detection element IDs were defined in a few places (and we have some in QC also, to be followed-up).
Replace those with a "single source of truth".


pillot (approved):

aphecetche:
@ktf @TimoWilken seems the `test_Framework_test_SingleDataSource` is failing quite often. Is the root cause known ?

ktf:
I think it's a wrong assert which might end up being triggered. Dropping it as part of #10010.

PR #10003, opened 2022-10-05
TRD raw reader catch error in verbose mode

martenole:
Otherwise we crash in verbose mode in case an invalid RDH is detected which does not make much sense, since in verbose mode we would like to understand bogus data


bazinski (dismissed):

f3sch (dismissed):
LGTM! :thumbsup:

PR #10005, opened 2022-10-05
Allow parsing of most severe message with reason

jgrosseo:


jgrosseo:
@ktf @matthiasrichter 
In case of an error, DPL writes at the end of the log the line " [ERROR] SEVERE: Device internal-dpl-aod-reader (42019) had at least one message above severity fatal: Unhandled o2::framework::runtime_error reached the top of main of o2-analysistutorial-histograms-full-tracks, device shutting down." which is the most severe error message found and allows easy inspection of what happened in batch mode by just checking the end of the file.

In case of an exception this does not include the reason though because of the newline where it in the log goes to the next line.

I propose to simply drop the first \n

Do you see any negative side effects of this?

Please forward as needed...

ktf:
No objections on my side. @davidrohr are we harvesting those online?

That said, why not simply using `grep -A2  | tail -n1` when harvesting?

jgrosseo:
It is not me or `grep` who is harvesting, it is a part of DPL...

jgrosseo:
```
      if (logLevel >= driverInfo.minFailureLevel) {
        info.lastError = token;
        if (info.firstSevereError.empty() || maxLogLevelIncreased) {
          info.firstSevereError = token;
        }
      }
```
in runDataProcessing.cxx and this only keeps one line...

ktf (approved):

davidrohr (commented):
for reference since I was asked, also no objection from my side. I think we might be grepping for "reached the top of main" at some point. Probably good to have the reason in the same line in any case.

PR #10006, opened 2022-10-05
Reduce verbosity of DCSDP processors, promote CCDBPopulator logs to important

shahor02:


shahor02:
@mpoghos @gvolpe79 please note that I've introduces some prescaling in the per-query logs of the followed run status, in order to reduce verbosity of the workflow. Some other logs were moved to the verbose mode.

PR #10008, opened 2022-10-06
Fix CTP lumi workflow, add it to dpl-workflow, add lumi to CTP CTF

shahor02:
don't merge yet


shahor02:
Hi @lietava 

There was a problem with 2 workflows sharing the same raw input (one of them always sees 0xdeadbeef) while in the FST we should produce both CTP digits and lumi in the same workflow. Therefore, I've suppressed the separate lumi workflow and addid its functionality to the former Raw2Digits (renaming it to RawDecoding).

Now o2-ctp-reco-workflow produces digits and lumi output (or only 1 of them with options `--no-lumi` and `--no-digits` resp.).
In the online processing we will have on the CTP FLP only the workflow producing the lumi info (which is a moving average over consecutive TFs), i.e. `o2-ctp-reco-workflow --no-digits`, which will also forward CTP raw data for the digits extraction on the EPN.
In this case on the EPN we will run `o2-ctp-reco-workflow --no-lumi --ntf-to-average 1`, which will output only digits (`--ntf-to-average 1` will be ignored). 
But in the FST or in case for some reason the FLP workflow is not running, we will need to run on the EPN also the lumi extraction part, i.e. simply `o2-ctp-reco-workflow --ntf-to-average 1`. In this case, the `ntf-to-average` option is essential since EPN does not see consecutive TFs (hence the precision of the lumi will be limited to statistics of 1 TF).

Is CTP now routinely writing raw data? If so, we should add asap the FLP the `o2-ctp-reco-workflow --no-digits`, forwarding the EPN lumi and original raw data, then we will have everything stored in the CTF.

lietava:
HI @shahor02 ,
yes, CTP now provides data routinely. ```we should add asap the FLP the o2-ctp-reco-workflow --no-digits``` : if no more tests are needed than Pippo et al will manage it ?

shahor02:
@lietava Yes, I've tested it as much as I could. Once this PR propagates to FLPSuite, you can add on the FLP 
`o2-ctp-reco-workflow --no-digits` workflow. Please don't forget to forward both lumi and raw data, i.e. it should look like:
```
o2-dpl-raw-proxy -b --session default --dataspec 'A:CTP/RAWDATA;dd:FLP/DISTSUBTIMEFRAME/0' --readout-proxy '--channel-config "name=readout-proxy,type=pull,method=connect,address=ipc:///tmp/stf-builder-dpl-pipe-0,transport=shmem,rateLogging=10"' | \
o2-ctp-reco-workflow -b --session default --no-digits | \
o2-dpl-output-proxy --environment DPL_OUTPUT_PROXY_ORDERED=1 -b --session default --dataspec 'A:CTP/RAWDATA;lumi:CTP/LUMI/0;dd:FLP/DISTSUBTIMEFRAME/0' --dpl-output-proxy '--channel-config "name=downstream,type=push,method=bind,address=ipc:///tmp/stf-pipe-0,rateLogging=10,transport=shmem"'
```
Note that no pipelining should be used to ensure consecutive order of TFs going to the moving average. If processing will happen to create backpressure, we will need to parallelize it internally, but, given that that code is short and simple, I don't expect backpressure.

Everything was green, merging after a trivial fix in the name of the device.

shahor02:
@lietava one more point: once this PR propagates to the EPN, for the runs with CTP we should have `CTP` in the `WORKFLOW_DETECTORS` PDP variable, and once the processing on FLP starts, one should add `CTP` also should be added to `WORKFLOW_DETECTORS_FLP_PROCESSING` (in the O2DPG/DATA/common/setenv.sh).
Anyway, please ping me when you will be ready to start processing on the FLP.

lietava:
@shahor02 : I communicate with Pippo at al about the deployment in P2. They should come back t0 us soon.

PR #10009, opened 2022-10-06
Point out about the environment

ktf:


ktf:
@awegrzyn the hint about `-b` was actually there, I am adding the part about the environment.

PR #10010, opened 2022-10-06
DPL: avoid crashes in case there is no actual data in a set of parts

ktf:


ktf:
We can most likely drop the message as well, as it should be completely fine that e.g. we get an End of Stream without any data.

PR #10014, opened 2022-10-06
Avoid issue with empty tables in Preslice and sliceBy

pbuehler:
When using Preslice / sliceBy with an empty table the processing crashed. This PR introduces changes to avoid the crash.

PR #10020, opened 2022-10-07
DPL Analysis: Protect Preslice in empty table case

aalkin:
@pbuehler @jgrosseo I had something else in mind, this should be a smaller change than #10014

### DIFF
--- a/Framework/Core/include/Framework/ASoA.h
+++ b/Framework/Core/include/Framework/ASoA.h
@@ -65,7 +65,7 @@ struct Preslice {
     arrow::Status status;
     if (fullSize == 0) {
       offset = 0;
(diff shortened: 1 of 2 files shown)


jgrosseo:
@pbuehler is this also fine for you?

pbuehler:
@jgrosseo @aalkin I tested both solutions and both solve the issue with the empty table.  So this PR can be closed if PR 10014 is accepted - or vice versa.

aalkin:
@pbuehler this one also fixes the function call in `PresliceManager` that was duplicating code, and does not touch `Kernels.cxx` -  the functions in it should not have extra runtime checks.

PR #10021, opened 2022-10-07
MCH: speedup elecmap by using static in some key methods

aphecetche:
The speedup on affected MCH tests is :

- ~30X for `testElectronicMapper`
- ~4X for `testClosureCoDecDigit`
- ~10X for `testNumberOfPadsPerFeeId`

Tested on a MacBook Pro M1.
On a FLP-like machine the results are comparable.

The total test time for mch (`ctest -L mch`) has been decreased by 1.5 on the MacBook, but not on the FLP-like machine where tests reading from disk (actual spinning disk vs SSD on the MacBook) dominate the total time.

@aferrero2707 FYI


aferrero2707 (approved):
Looks ok for me.

pillot (approved):

PR #10023, opened 2022-10-07
DPL: move CCDB support to a plugin

ktf:


ktf:
This gets rid of a bunch of dependencies in the framework itself. Only the CCDBFetcher will need to load JAliEn and alike...

PR #10024, opened 2022-10-07
DPL: introduce ServiceRegistryRef as a name to simplify migration

ktf:


ktf:
Will be useful to migrate QC to the multithreading API of the service registry. QC itself will not be multithreaded but this way I can (somewhat) easily migrate to the new API everywhere.

PR #10025, opened 2022-10-07
[EMCAL-565, EMCAL-566] Reject calib triggered events in online calib

jokonig:
- Calibration triggered events need to be rejected during the online calibration
- Previously these events did notas this was not working on CTP level
- Calibration triggers are rejected by default and can be accepted if special flag is set (no-rejectCalibTrigger)


mfasDa (approved):
Looks good!

mfasDa:
Calib triggers must be rejected, otherwise they distort the time distribution like what we see in supermodule 13 (attached) as they trigger peak is displaced in time specifically. Even though the LED events are triggered with low rate they are sufficient to influence the time fit.

We would keep the option to include the calib triggers mainly for future tests of the workflow in technical runs where calib triggers are needed to generate a signal pulse fo the test. In physics runs they must be rejeceted.

Bildschirmfoto 2022-10-08 um 16 13 49

shahor02 (approved):

PR #10028, opened 2022-10-07
Multithreading preparations

ktf:


ktf:
This actually requires #10024 and some extra changes in QualityControl before it can be merged.

PR #10031, opened 2022-10-08
FIT: add definitions for new DCS DPs

andreasmolander:
Trigger and background rates in DCS are configured for ADAPOS. Include these in CCDB upload workflows and related macros.

### DIFF
--- a/Detectors/FIT/FDD/dcsmonitoring/workflow/fdd-dcs-sim-workflow.cxx
+++ b/Detectors/FIT/FDD/dcsmonitoring/workflow/fdd-dcs-sim-workflow.cxx
@@ -20,22 +20,25 @@
 o2::framework::WorkflowSpec defineDataProcessing(o2::framework::ConfigContext const& configcontext)
 {
   std::vector<o2::dcs::test::HintType> dphints;
-  // for testing, we use less DPs than the official ones
   dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/SIDE_A/HV_A9/[I,V]MON", 250, 350});
+  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/SIDE_C/HV_C[9,32]/[I,V]MON", 250, 350});
+  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/SIDE_C/LAYER0/PMT_0_[0..3]/[I,V]MON", 250, 350});
+  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/SIDE_C/LAYER1/PMT_1_[0..3]/[I,V]MON", 250, 350});
+  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/SIDE_A/LAYER2/PMT_2_[0..3]/[I,V]MON", 250, 350});
+  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/SIDE_A/LAYER3/PMT_3_[0..3]/[I,V]MON", 250, 350});
   dphints.emplace_back(o2::dcs::test::DataPointHint<uint>{"FDD/PM/SIDE_A/PMT_A_9/ADC[0,1]_BASELINE", 30, 150});
-  // Official list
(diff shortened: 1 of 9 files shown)


Detectors/FIT/FDD/dcsmonitoring/workflow/fdd-dcs-sim-workflow.cxx
```diff
@@ -23,6 +23,7 @@ o2::framework::WorkflowSpec defineDataProcessing(o2::framework::ConfigContext co
   // for testing, we use less DPs than the official ones
   dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/SIDE_A/HV_A9/[I,V]MON", 250, 350});
   dphints.emplace_back(o2::dcs::test::DataPointHint<uint>{"FDD/PM/SIDE_A/PMT_A_9/ADC[0,1]_BASELINE", 30, 150});
+  dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/Trigger1_Central/CNT_RATE", 0, 1000000});
```
mslupeck: Assuming 1000000 is the max expected value, I'd say it should be increased, at least by a factor of 5. Same for FT0 and FV0

mslupeck (dismissed):

Detectors/FIT/FDD/dcsmonitoring/workflow/fdd-dcs-sim-workflow.cxx
andreasmolander: Ok I changed it. These are only the range of randomly generated values if one wants to simulate ADAPOS input locally. The production workflow sets no limits on anything.

Detectors/FIT/FDD/dcsmonitoring/workflow/fdd-dcs-sim-workflow.cxx
```diff
@@ -36,6 +37,13 @@ o2::framework::WorkflowSpec defineDataProcessing(o2::framework::ConfigContext co
   // dphints.emplace_back(o2::dcs::test::DataPointHint<uint>{"FDD/PM/SIDE_C/LAYER1/PMT_1_[0..3]/ADC[0,1]_BASELINE", 30, 150});
   // dphints.emplace_back(o2::dcs::test::DataPointHint<uint>{"FDD/PM/SIDE_A/LAYER2/PMT_2_[0..3]/ADC[0,1]_BASELINE", 30, 150});
   // dphints.emplace_back(o2::dcs::test::DataPointHint<uint>{"FDD/PM/SIDE_A/LAYER3/PMT_3_[0..3]/ADC[0,1]_BASELINE", 30, 150});
+  // dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/Trigger1_Central/CNT_RATE", 0, 5000000});
+  // dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/Trigger2_SemiCentral/CNT_RATE", 0, 5000000});
+  // dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/Trigger3_Vertex/CNT_RATE", 0, 5000000});
+  // dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/Trigger4_OrC/CNT_RATE", 0, 5000000});
+  // dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/Trigger5_OrA/CNT_RATE", 0, 5000000});
+  // dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/Background/[0..9]/CNT_RATE", 0, 50000});
+  // dphints.emplace_back(o2::dcs::test::DataPointHint<double>{"FDD/Background/[A,B,C,D,E,F,G,H]/CNT_RATE", 0, 50000});
```
shahor02: why are those commented? If not needed, can they be removed? Same comment for other FIT detectors.

andreasmolander: Just to make local testing a bit easier, it's less stuff to keep track of. If one works they would (probably) all work. But I uncommented all of them, one can comment out as needed instead. For sure nicer this way.

PR #10032, opened 2022-10-09
MCH: add DCS aliases parsing

aphecetche:
This (rather large, sorry about that) PR adds some functions to parse the various DCS aliases (HV voltages and currents, LV analog, digital and Solar).

The first commit is adding to the `elecmap.py` python script the ability to generate the mapping DCS alias -> solar crate number (from the master google sheet mapping).

The second commit (scope=MCHConditions lib) adds `aliasToXXX(string alias)` functions to extract various informations (chamber, side, plane, etc...) from  the DCS aliases.
 
The third commit (scope=MCHGlobalMapping lib) mainly adds the `aliasToDsIndices(string alias)` which gives, for a given alias, the list of dual sampa indices (from 0 to 16819). That function is used by the new `--dcs` option of the `o2-mch-global-mapper` executable to output a JSON mapping file (alias->[dsindices]) that is in turn used by MchView. 

The 3 commits are intended to be atomic, please do not squash.


Detectors/MUON/MCH/Conditions/include/MCHConditions/Chamber.h
```diff
@@ -0,0 +1,55 @@
...
+  Ch00,
+  Ch01,
+  Ch02,
+  Ch03,
+  Ch04,
+  Ch05,
+  Ch06,
+  Ch07,
+  Ch08,
+  Ch09,
```
pillot: there is a spurious comma after Ch09, which does not seem to be a problem.

Detectors/MUON/MCH/Conditions/src/DCSAliases.cxx
```diff
@@ -0,0 +1,92 @@
...
+
+namespace o2::mch::dcs
+{
+
+std::vector<std::string> aliases(std::vector<MeasurementType> types)
+{
+  std::vector<std::string> aliasesVector;
+
+  if (in(MeasurementType::HV_V, types)) {
+    aliasesVector.insert(begin(aliasesVector),
```
pillot: This is likely a detail, but is there a particular reason to insert at the front of the vector? It might be faster to insert at the back.

Detectors/MUON/MCH/Conditions/src/DetectionElement.cxx
```diff
@@ -0,0 +1,91 @@
...
+    return slat::detElemId(chamber, side, number);
+  } else {
+    return quadrant::detElemId(chamber, number);
+  }
+}
+
+std::optional<int> aliasToDetElemId(std::string_view dcsAlias)
+{
+  const auto m = aliasToMeasurementType(dcsAlias);
+  const auto chamber = aliasToChamber(dcsAlias);
```
pillot: Remove this line as it duplicates L78 in the if statement below.

Detectors/MUON/MCH/Conditions/src/Plane.cxx
```diff
@@ -0,0 +1,83 @@
...
+{
+  static constexpr auto chambersOfStation12 = {
+    Chamber::Ch00,
+    Chamber::Ch01,
+    Chamber::Ch02,
+    Chamber::Ch03};
+
+  auto chamber = aliasToChamber(dcsAlias);
+  if (std::find(begin(chambersOfStation12),
+                end(chambersOfStation12), chamber) != chambersOfStation12.end()) {
```
pillot: you could use the function isQuadrant(...)

Detectors/MUON/MCH/Conditions/src/Plane.cxx
```diff
@@ -0,0 +1,83 @@
...
+    if (side == Side::Left) {
+      switch (number) {
+        case 1:
+        case 4:
+          return Plane::Bending;
+        case 2:
+        case 3:
+          return Plane::NonBending;
+        default:
+          throw std::invalid_argument("invalild number");
```
pillot: typo: "invalid"

Detectors/MUON/MCH/Conditions/src/Plane.cxx
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+      switch (number) {
+        case 2:
+        case 3:
+          return Plane::Bending;
+        case 1:
+        case 4:
+          return Plane::NonBending;
+        default:
+          throw std::invalid_argument("invalild number");
```
pillot: typo: "invalid"

pillot (commented):
This is a first set of review (of MCH/Conditions). I will try to finish tomorrow...

As a general comment, in case of failure, the alias decoding sometimes return an invalid value, sometimes throw an exception and sometimes return a std::optional(nullopt) (which will eventually result in throwing an exception when used). I am wondering if the same thing could be done everywhere, e.g. returning an std::optional? Also, I haven't looked at the rest of the PR yet, but are the exceptions always caught?

Detectors/MUON/MCH/Conditions/src/DCSAliases.cxx
aphecetche: no special reason indeed. will change.

Detectors/MUON/MCH/Conditions/src/Plane.cxx
aphecetche: indeed.

Detectors/MUON/MCH/Conditions/src/Plane.cxx
aphecetche: fixed

Detectors/MUON/MCH/Conditions/src/Plane.cxx
aphecetche: fixed

Detectors/MUON/MCH/Conditions/src/DetectionElement.cxx
aphecetche: fixed

aphecetche:
> This is a first set of review (of MCH/Conditions). I will try to finish tomorrow...
> 
> As a general comment, in case of failure, the alias decoding sometimes return an invalid value, sometimes throw an exception and sometimes return a std::optional(nullopt) (which will eventually result in throwing an exception when used). I am wondering if the same thing could be done everywhere, e.g. returning an std::optional? Also, I haven't looked at the rest of the PR yet, but are the exceptions always caught?

The intent was indeed to throw an exception (either directly or when a std::nullopt will be dereferenced) when the alias is not valid. I've now stated that fact explicitly in the function headers and throw an exception in a case where I was returning a dummy value (-1).

As for catching the exception, so far the code in this PR is not used in the reconstruction part (but only in the `global-mapper` executable, so there's no need to "properly" catch them for the moment I would say.

Detectors/MUON/MCH/Conditions/src/DCSAliases.cxx
```diff
@@ -0,0 +1,92 @@
...
+                         begin(expectedLVAliasesSolar),
+                         end(expectedLVAliasesSolar));
+  }
+  return aliasesVector;
+}
+
+std::vector<std::string> allAliases()
+{
+  std::vector<std::string> validAliases;
+  validAliases.insert(validAliases.begin(), expectedHVAliasesVoltages.begin(),
```
pillot: here as well you could insert at back

Detectors/MUON/MCH/Conditions/include/MCHConditions/Plane.h
```diff
@@ -0,0 +1,36 @@
...
+  Bending,
+  NonBending,
+  Both
+};
+
+/** return a string representation of plane */
+std::string name(Plane plane);
+
+/** extract the plane information from the alias */
+Plane aliasToPlane(std::string_view alias);
```
pillot, line 33: you forgot the comment saying that it throws an exception if the alias is invalid

Detectors/MUON/MCH/Conditions/include/MCHConditions/DCSAliases.h
```diff
@@ -0,0 +1,53 @@
...
+#define O2_MCH_CONDITIONS_DCS_ALIASES_H
+
+#include "MCHConditions/Cathode.h"
+#include "MCHConditions/Chamber.h"
+#include "MCHConditions/DetectionElement.h"
+#include "MCHConditions/MeasurementType.h"
+#include "MCHConditions/Number.h"
+#include "MCHConditions/Plane.h"
+#include "MCHConditions/Side.h"
+#include <optional>
```
pillot: this include is not needed

Detectors/MUON/MCH/Conditions/src/DCSAliases.cxx
```diff
@@ -0,0 +1,92 @@
...
+#include "MCHConditions/DCSAliases.h"
+
+#include "DetectionElements.h"
+#include "HVAliases.h"
+#include "LVAliases.h"
+#include <array>
+#include <fmt/core.h>
+#include <iostream>
+#include <map>
+#include <optional>
```
pillot: this include is not needed

Detectors/MUON/MCH/GlobalMapping/include/MCHGlobalMapping/DsIndex.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include "MCHRawElecMap/DsDetId.h"
+#include <cstdint>
+#include <set>
+#include <string>
+#include <utility>
+
+namespace o2::mch
+{
+/** DsIndex is an integer, from 0 to 16719, which uniquely identifies
```
pillot: 16719 --> 16819, as NumberOfDualSampas = 16820 ?

aphecetche:
To complement my answer above, the choice of exception (vs optional for instance) is motivated by the fact that in the intended usage of those functions the origin of the alias strings is the CCDB (`MCH/Calib/HV` or `LV` objects), in which (in principle at least) we only have valid aliases. So getting an invalid alias should indeed be _exceptional_.

Detectors/MUON/MCH/GlobalMapping/src/DsIndex.cxx
```diff
@@ -0,0 +1,84 @@
...
+      const auto& seg = o2::mch::mapping::segmentation(deId);
+      std::vector<int> dsids;
+      seg.forEachDualSampa([&](int dsid) {
+        dsids.emplace_back(dsid);
+      });
+      // ensure dual sampa are sorted by dsid
+      std::sort(dsids.begin(), dsids.end());
+      for (auto dsId : dsids) {
+        raw::DsDetId det{deId, dsId};
+        m.emplace(encode(det), dsIndex);
```
pillot, line 64: How is it possible that you don't need to specify the namespace to use this function: raw::encode(...)?

Detectors/MUON/MCH/GlobalMapping/src/DsIndex.cxx
```diff
@@ -0,0 +1,84 @@
...
+      }
+    });
+  }
+  return m;
+}
+
+DsIndex getDsIndex(const o2::mch::raw::DsDetId& dsDetId)
+{
+  static std::map<uint32_t, uint16_t> m = buildDetId2DsIndexMap();
+  return m[encode(dsDetId)];
```
pillot, line 75: same here?

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
```diff
@@ -0,0 +1,113 @@
...
+  static auto allDsElecId = raw::getAllDs<raw::ElectronicMapperGenerated>();
+  static auto elec2det = raw::createElec2DetMapper<raw::ElectronicMapperGenerated>();
+  std::set<raw::DsDetId> allDs;
+  for (auto elecId : allDsElecId) {
+    allDs.emplace(elec2det(elecId).value());
+  }
+  return allDs;
+}
+
+double findXmin(const o2::mch::mapping::Segmentation& seg, int dsId)
```
pillot: This function is not used, right?

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
```diff
@@ -0,0 +1,113 @@
...
+#include "MCHConditions/DetectionElement.h"
+#include "MCHConditions/Number.h"
+#include "MCHGlobalMapping/DsIndex.h"
+#include "MCHMappingInterface/Segmentation.h"
+#include "MCHRawElecMap/DsDetId.h"
+#include "MCHRawElecMap/Mapper.h"
+
+namespace o2::mch::dcs
+{
+std::set<raw::DsDetId> allDsDetId()
```
pillot: Is there a reason to build this list from the electronic mapping? Otherwise, I am not sure if it is an easier, faster or just equivalent way to do the same thing, but I am wondering if this function could not be part of  DsIndex.h and make use of the static map already built with buildDetId2DsIndexMap(). In the end this is just the list of all the (decoded) keys in this map, no?
Actually looping directly on this map (if made available) in the function hvAliasToDsIndices(...) below would (i) remove the need for extracting the list of all DsDetId and (ii) remove the need to find the corresponding index for each Id with getDsIndex(dd).

pillot (commented):
Second set of comments. Last part tomorrow hopefully.

Detectors/MUON/MCH/Conditions/src/DCSAliases.cxx
aphecetche: yes

Detectors/MUON/MCH/Conditions/include/MCHConditions/Plane.h
aphecetche: indeed

Detectors/MUON/MCH/Conditions/src/DCSAliases.cxx
aphecetche: fixed

Detectors/MUON/MCH/GlobalMapping/include/MCHGlobalMapping/DsIndex.h
aphecetche: yes, fixed

Detectors/MUON/MCH/GlobalMapping/src/DsIndex.cxx
aphecetche: As far as I understand, because of [Argument-dependent lookup](https://en.cppreference.com/w/cpp/language/adl)

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
aphecetche: No, it's not. I hesitated between `findXmin` and `findXmean` and picked the mean in the end... Will remove it.

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
aphecetche: indeed, this function is only used in the `hvAliasToDsIndices` function, so could be more tailored to it. 
I've changed it somewhat. Let me know if you prefer that implementation.

Detectors/MUON/MCH/Conditions/include/MCHConditions/DCSAliases.h
aphecetche: fixed

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
pillot: Yes, that's simpler and keep the internal DsDetId2DsIndex map hidden. And if the performance ever become an issue, the loop over all DS could be done only once and the results stored in a static map DeId/sector --> DsIndices in the function.

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
```diff
@@ -0,0 +1,104 @@
...
+  bool slat = isSlat(chamber);
+  int deId = aliasToDetElemId(alias).value();
+  std::set<int> indices;
+  for (auto dsIndex = 0; dsIndex < o2::mch::NumberOfDualSampas; ++dsIndex) {
+    const auto& dd = getDsDetId(dsIndex);
+    if (dd.deId() != deId) {
+      continue;
+    }
+    if (slat) {
+      indices.emplace(getDsIndex(dd));
```
pillot: you can now use dsIndex here

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
```diff
@@ -0,0 +1,104 @@
...
+      if (xref < x1) {
+        sector = 0;
+      } else if (xref < x2) {
+        sector = 1;
+      } else if (xref < x3) {
+        sector = 2;
+      }
+      const auto number = aliasToNumber(alias);
+      if (sector == number % 10) {
+        indices.emplace(getDsIndex(dd));
```
pillot: here as well

Detectors/MUON/MCH/GlobalMapping/src/Quadrant.cxx
```diff
@@ -0,0 +1,121 @@
...
+          return {Cathode{base + 0, Plane::NonBending}};
+        case 3:
+          return {Cathode{base + 3, Plane::NonBending}};
+        case 1:
+          return {Cathode{base + 3, Plane::Bending}};
+        default:
+          throw std::invalid_argument(fmt::format("wrong alias {}", alias));
+      }
+    }
+  } else {
```
pillot: I would explicitly check if (isStation2(chamber)) and throw an exception in the end of the function to make sure it is never used by mistake for other chambers.

Detectors/MUON/MCH/GlobalMapping/src/Quadrant.cxx
```diff
@@ -0,0 +1,121 @@
...
+std::set<int> solarAliasToDsIndices(std::string_view alias)
+{
+  /** For St12, to get the solar alias to dual sampas relationship
+   * we "just" have to convert the solar crate number to a LV
+   * group (as used in the analog and digital measurements),
+   * and then reuse the LV group logic.
+   */
+  // lvid.type = dcs::MeasurementType::LV_V_FEE_ANALOG;
+  const std::array<int, 4> crateToGroupSt1 = {4, 2, 3, 1};
+  const std::array<int, 4> crateToGroupSt2 = {2, 1, 4, 3};
```
pillot: could be static

pillot (commented):
Should be last set of comments.

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
pillot: you forgot to remove it actually ;-)

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
aphecetche: agreed

Detectors/MUON/MCH/GlobalMapping/src/Quadrant.cxx
aphecetche: agreed

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
aphecetche: agreed

Detectors/MUON/MCH/GlobalMapping/src/Quadrant.cxx
aphecetche: could be ;-)

Detectors/MUON/MCH/GlobalMapping/src/HV.cxx
aphecetche: oops. one more force-push then...

pillot (approved):

aphecetche:
obviously the fjcontrib change is less innocent that what I thought from my tests on my Mac and on a dev FLP... Will investigate. Meanwhile putting this PR in draft mode.

aphecetche:
@TimoWilken fullCI failed two days ago and has not been retried yet : is that normal ?

aphecetche:
@shahor02 the fullCI is now failing due to an unrelated issue, introduced by https://github.com/AliceO2Group/AliceO2/pull/10080 I think ?

PR #10039, opened 2022-10-10
AODProducer: fix typo + small code simplification

sawenzel:
* use correct trackID when filling MC particles for TOF tracks (was typo here)
* simplify code: do not use extra variables


sawenzel:
@nburmaso

PR #10041, opened 2022-10-10
DPL: fix initial delay for timers

ktf:
Long delays between configure and run state could make the randomisation be ineffective.

### DIFF
--- a/Framework/Core/src/LifetimeHelpers.cxx
+++ b/Framework/Core/src/LifetimeHelpers.cxx
@@ -98,20 +98,24 @@ ExpirationHandler::Creator LifetimeHelpers::enumDrivenCreation(size_t start, siz
 
 ExpirationHandler::Creator LifetimeHelpers::timeDrivenCreation(std::chrono::microseconds period)
 {
-  std::random_device r;
-  std::default_random_engine e1(r());
-  std::uniform_int_distribution<uint64_t> dist(0, period.count() * 0.9);
-
-  // We start with a random offset to avoid all the devices
-  // send their first message at the same time, bring down
-  // the QC machine.
-  // We reduce the first interval rather than increasing it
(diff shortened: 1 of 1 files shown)


knopers8 (approved):
thanks!

vascobarroso:
@ktf @knopers8 we have an update window this Wednesday. Is it feasible to include this in tonight's tag ?

ktf:
done

davidrohr:
@ktf @knopers8 : I think this is probably insufficient. The run state the process knows about is only the local run state of itself, not the global one.
It can be that the process goes to RUNNING, but first data arrives only 1 minute later, if e.g. one detector takes quite long to go into RUNNING, which is usually the case in global runs at P2.
So we should initialize the timer the first time runDataProcessing is invoked, not based on state changes.

ktf:
Well, it's a bit of a chicken / egg problem, because we actually process data when the timer expires. Wouldn't the best solution to simply delay the sending, then?

davidrohr:
Can you start the timer when the device receives data for the first time?
Though that would be a problem for source devices....

What do you mean be delay the sending? I guess that could be done, but then the decision whether QC sends something or not would need to be made internally, not relying on timers?

knopers8:
I don't quite understand what is the problem. QC does not wait with using the timer inputs until it gets any data. Also, as far as I understand, all EPNs go to RUNNING at the same time, right?

davidrohr:
yes, all EPNs go to running at the same time, but they will only receive the first data 1-2 minutes later, when all detectors have finished their SOR

knopers8:
OK, but that is really not a problem, as far as I understand. We do not require any data to publish first histograms, as you can see here https://github.com/AliceO2Group/QualityControl/blob/master/Framework/src/TaskRunner.cxx#L193

PR #10042, opened 2022-10-10
Optimization for AOD MC filling (part 1)

sawenzel:
This commit provides a considerable speed gain
for MC-AOD conversion by speeding up the MC particle table filling. The overall speed improvement is ~45% for pp timeframes and ~75% for PbPb timeframes.

This improvement is achieved via:

a) restructuring of the data structure keeping track
   of which MC particles survive in AOD (mToStore):

   - use a simple hashmap per source and event instead of a hash map of combined source,event,track triples. The latter needs to much CPU time in hashing and access.

b) do not loop over all particles uselessly (in some parts of the code):
   - we introduce a new (sorted) container keeping track of relevant particles over which we can loop which get's rid of some filtering

Both measures contribute significantly to the gain. The changes have been validated using the MC release validation procedure on AODs.

Note that the present optimization is step 1 in larger campaign:

https://alice.its.cern.ch/jira/browse/O2-3240


sawenzel:
@nburmaso

PR #10043, opened 2022-10-10
Revert "DPL Analysis: do not use string lookup for finding columns (#9710)"

aalkin:
This reverts commit f5c0177a583f317619011e60429b5c0111868907.

It was found that some Run 2 converted data has order of branches in some trees that is different from the corresponding order of columns in the data model. This restores string lookup to prevent incorrect data access in similar cases.


aalkin:
@ekryshen @jgrosseo @ktf

jgrosseo (approved):

PR #10044, opened 2022-10-10
DPL: Add ability to get symbol name from debug information

ktf:
This will use addr2line and C++ to print the symbols for which we are lacking the proper debug information.

I suspect the actual issue is that we are mixing ld from the system with our own compiler.


ktf:
@Barthelemy do you see elsewhere  the "Could not find the DPL InfoLogger" messages? Where do they come from? Is the timeout related?

PR #10049, opened 2022-10-11
Keep doing random timers until we get some data

ktf:


davidrohr:
I have one technical question: Do you reset the timer every time in the event loop, or only when the timer fires?
We should do the latter.
If we reset very often, we will hit small timer values often, which means that the timer will fire quite often until the first data has arrived.

ktf:
If I understand the question correctly, only after it fires.

davidrohr:
ok, so just to clarify.
You set the random timer.
When the timer fires, you check if data has already arrived, and if not, you set a new random value.
You kepp doing so until the first data arrives, and then you don't set random values any more?

If so I'd merge this, and get it in for tomorrow.

ktf:
Yes, more correctly until one of the data channels sets the oldest possible timeframe for that channel.

PR #10053, opened 2022-10-11
TRD: calibration: vDrift and ExB

f3sch:
This patch series does the following:

1. reusing fit values from the previous slot
2. pulling the initial fit values from the ccdb if there was any which is still valid e.g. ~in the last hour
3. introduced a keyConfig values for more granular control for track selection
4. introduced a new constant for trd
5. some minor touch-ups and doxygen compliant comments

more info in commits.

There were some concerns about the TrackletTransformer however looking through the code vDrift and ExB are only used for calculating the calibrated Dy (where it is a good thing to use updated values anyways) and for caluclation of the Timebin [here](https://github.com/AliceO2Group/AliceO2/blob/3d9ad243b0cff6cfb3112b5d950ca01e69bd7266/Detectors/TRD/base/src/TrackletTransformer.cxx#L139-L155), where an updated vDrift would also be good? How do I test this?

I `tested` it by comparing the fit values before and after for a Run 523677.
I did not have a chance to test if the actual pull for the initial fit values works as I expect, this is somewhat hard for me to test.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
@@ -67,7 +67,6 @@ constexpr int ADCBASELINE = 10;                                       ///< basel
 constexpr int TIMEBINS = 30;            ///< the number of time bins
 constexpr float MAXIMPACTANGLE = 25.f;  ///< the maximum impact angle for tracks relative to the TRD detector plane to be considered for vDrift and ExB calibration
 constexpr int NBINSANGLEDIFF = 25;      ///< the number of bins for the track angle used for the vDrift and ExB calibration based on the tracking
-constexpr double ANODEPLANE = 0.0335;   ///< distance of the TRD anode plane from the drift cathodes in [m]
 constexpr double VDRIFTDEFAULT = 1.546; ///< default value for vDrift
 constexpr double EXBDEFAULT = 0.0;      ///< default value for LorentzAngle
 
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
@@ -67,6 +67,7 @@ constexpr int ADCBASELINE = 10;                                       // baselin
 constexpr int TIMEBINS = 30;           // the number of time bins
 constexpr float MAXIMPACTANGLE = 25.f; // the maximum impact angle for tracks relative to the TRD detector plane to be considered for vDrift and ExB calibration
 constexpr int NBINSANGLEDIFF = 25;     // the number of bins for the track angle used for the vDrift and ExB calibration based on the tracking
+constexpr double ANODEPLANE = 0.0335; // distance of the TRD anode plane from the drift cathodes in [m]
 
 // Trigger parameters
 constexpr double READOUT_TIME = 3000;                  // the time the readout takes, as 30 TB = 3 micro-s.
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
@@ -64,10 +64,12 @@ constexpr float GRANULARITYTRKLSLOPE = 1.f / PADGRANULARITYTRKLSLOPE; ///< granu
 constexpr int ADCBASELINE = 10;                                       ///< baseline in ADC units
 
 // OS: Should this not be flexible for example in case of Kr calib?
-constexpr int TIMEBINS = 30;           ///< the number of time bins
-constexpr float MAXIMPACTANGLE = 25.f; ///< the maximum impact angle for tracks relative to the TRD detector plane to be considered for vDrift and ExB calibration
-constexpr int NBINSANGLEDIFF = 25;     ///< the number of bins for the track angle used for the vDrift and ExB calibration based on the tracking
-constexpr double ANODEPLANE = 0.0335;  ///< distance of the TRD anode plane from the drift cathodes in [m]
+constexpr int TIMEBINS = 30;            ///< the number of time bins
+constexpr float MAXIMPACTANGLE = 25.f;  ///< the maximum impact angle for tracks relative to the TRD detector plane to be considered for vDrift and ExB calibration
+constexpr int NBINSANGLEDIFF = 25;      ///< the number of bins for the track angle used for the vDrift and ExB calibration based on the tracking
+constexpr double ANODEPLANE = 0.0335;   ///< distance of the TRD anode plane from the drift cathodes in [m]
+constexpr double VDRIFTDEFAULT = 1.546; ///< default value for vDrift
+constexpr double EXBDEFAULT = 0.0;      ///< default value for LorentzAngle
 
 // Trigger parameters
 constexpr double READOUT_TIME = 3000;                  ///< the time the readout takes, as 30 TB = 3 micro-s.
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
@@ -22,76 +22,76 @@ namespace trd
 {
 namespace constants
 {
-constexpr int NSECTOR = 18;          // the number of sectors
-constexpr int NSTACK = 5;            // the number of stacks per sector
-constexpr int NLAYER = 6;            // the number of layers
-constexpr int NCHAMBERPERSEC = 30;   // the number of chambers per sector
-constexpr int NHCPERSEC = 60;        // the number of half-chambers per sector
-constexpr int MAXCHAMBER = 540;      // the maximum number of installed chambers
-constexpr int MAXHALFCHAMBER = 1080; // the maximum number of installed half-chambers
-constexpr int NCHAMBER = 521;        // the number of chambers actually installed
-constexpr int NHALFCRU = 72;         // the number of half cru (link bundles)
-constexpr int NLINKSPERHALFCRU = 15; // the number of links per half cru or cru end point.
-constexpr int NLINKSPERCRU = 30;     // the number of links per CRU (two CRUs serve one supermodule)
-constexpr int NCRU = 36;             // the number of CRU we have
-constexpr int NFLP = 12;             // the number of FLP we have.
-constexpr int NCRUPERFLP = 3;        // the number of CRU per FLP
-constexpr int TRDLINKID = 15;        // hard coded link id, specific to TRD
+constexpr int NSECTOR = 18;          ///< the number of sectors
+constexpr int NSTACK = 5;            ///< the number of stacks per sector
+constexpr int NLAYER = 6;            ///< the number of layers
+constexpr int NCHAMBERPERSEC = 30;   ///< the number of chambers per sector
+constexpr int NHCPERSEC = 60;        ///< the number of half-chambers per sector
+constexpr int MAXCHAMBER = 540;      ///< the maximum number of installed chambers
+constexpr int MAXHALFCHAMBER = 1080; ///< the maximum number of installed half-chambers
+constexpr int NCHAMBER = 521;        ///< the number of chambers actually installed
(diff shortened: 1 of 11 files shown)


DataFormats/Detectors/TRD/include/DataFormatsTRD/Constants.h
```diff
@@ -22,74 +22,75 @@ namespace trd
...
+constexpr int MAXDATAPERLINK256 = 1728;               //< max number of linkwords per cru link. (256bit words)
+constexpr int MAXEVENTCOUNTERSEPERATION = 200;        //< how far apart can subsequent mcmheader event counters be before we flag for concern, used as a sanity check in rawreader.
+constexpr int MAXMCMCOUNT = 69120;                    //< at most mcm count maxchamber x nrobc1 nmcmrob
+constexpr int MAXLINKERRORHISTOGRAMS = 10;            //< size of the array holding the link error plots from the raw reader
+constexpr int MAXPARSEERRORHISTOGRAMS = 60;           //< size of the array holding the parsing error plots from the raw reader
+constexpr unsigned int ETYPEPHYSICSTRIGGER = 0x2;     //< CRU Half Chamber header eventtype definition
+constexpr unsigned int ETYPECALIBRATIONTRIGGER = 0x3; //< CRU Half Chamber header eventtype definition
+constexpr int MAXCRUERRORVALUE = 0x2;                 //< Max possible value for a CRU Halfchamber link error. As of may 2022, can only be 0x0, 0x1, and 0x2, at least that is all so far(may2022).
```
martenole: Does this actually make a difference? Is a doxygen doc generated for this namespace? In any case if we change this please use `///<` as in the commenting guidelines: https://rawgit.com/AliceO2Group/CodingGuidelines/master/comments_guidelines.html?showone=Class_Data_Members#Class_Data_Members

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -14,49 +14,48 @@
...
+
 using namespace o2::trd::constants;
 
 namespace o2::trd
 {
 
 double FitFunctor::calculateDeltaAlphaSim(double vdFit, double laFit, double impactAng) const
 {
-
-  double trdAnodePlane = .0335; // distance of the TRD anode plane from the drift cathodes in [m]
```
martenole, line 37: Actually, I think this is a typo. Should be 3.5 cm and this can be taken from the geometry: `o2::trd::GeometryBase::camHght()/(2.f * 100.f)` the 100 to go from cm to m

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -101,13 +100,45 @@ void CalibratorVdExB::initProcessing()
   if (mInitDone) {
     return;
   }
+
+  auto& ccdbmgr = o2::ccdb::BasicCCDBManager::instance();
```
martenole: We cannot use the CCDB manager in any DPL device running in the synchronous stage, so this has to be done differently. I think we should subscribe to the `TRD/Calib/CalVdriftExB` objects and then check the creation time of the object we get to decide whether it should be used or not. The tracklet transformer is already subscribing to it, so you could take some code from there: https://github.com/AliceO2Group/AliceO2/blob/3d9ad243b0cff6cfb3112b5d950ca01e69bd7266/Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx#L153

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -101,13 +100,45 @@ void CalibratorVdExB::initProcessing()
...
+  mFitter.Config().ParSettings(ParamIndex::LA).SetStepSize(.01);
+  mFitter.Config().ParSettings(ParamIndex::VD).SetLimits(0., 3.);
+  mFitter.Config().ParSettings(ParamIndex::VD).SetStepSize(.01);
+  ROOT::Math::MinimizerOptions opt;
+  opt.SetMinimizerType("Minuit2");
+  opt.SetMinimizerAlgorithm("Migrad");
+  opt.SetPrintLevel(0);
+  opt.SetMaxFunctionCalls(1'000);
+  opt.SetTolerance(.001);
+  mFitter.Config().SetMinimizerOptions(opt);
```
martenole, line 127: Thanks, this is indeed better done only once in init

martenole (changes requested):
Thanks a lot for looking into this. I think this the way to go, just needs a few changes (see comments in the code)

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
f3sch: Ok, I have updated the code a bit but I am struggling wrapping my head around how to progress further.
How can I go from dataRequest to checking the objects creation date and extracting the data in VdAndExBCalibDevice::init(...) or do I need to this in run(...)?

martenole: Actually, even with the current state you are not really checking the time the object was created. `latest` will always give you a valid object. The default validity for one vDrift object is 1 hour. In case there was no new object created in the last hour you will get the default object which has infinite validity. So I think you can always take simply the object the framework is providing to you together with the first processed time frame.
Also note that you should not use something like ` auto currTime = std::time(nullptr) * 1000;` in the code. In case we do the processing asynchronously to the data taking that time will be wrong. Instead, you can get the creation time for the TF in question from the data header: see slide 7 of https://docs.google.com/presentation/d/1WPNixo57zAn_gM6vMCIBOHs-DZqcKAkLX0CwUgO_VYg/edit#slide=id.g10e79f44b99_0_0

DataFormats/Detectors/TRD/include/DataFormatsTRD/CalVdriftExB.h
```diff
@@ -34,7 +34,7 @@ class CalVdriftExB
   void setVdrift(int iDet, float vd) { mVdrift[iDet] = vd; }
   void setExB(int iDet, float exb) { mExB[iDet] = exb; }
 
-  float getVdrift(int iDet) const { return mVdrift[iDet]; }
+  float getVdrift(int iDet) const { return (mVdrift[iDet] > 0.0) ? mVdrift[iDet] : constants::VDRIFTDEFAULT; }
```
martenole: This I don't like. I think the CCDB object should always give you what is stored inside. And then it would be up to the consumer to check if the value is 0 or not and possibly use the default value then

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
f3sch: Makes sense with async processing not to take current time like this.
I am still not sure if with `pc.inputs().get<...>` I get a valid object related to that timeframe? there is something like getForTimeStamp for this, right?

martenole: You will get the object which is valid for the current time (creation time of the first TF in flight). This can either be the default object or the last calibration output. The latter one has precedence in case both are present

f3sch: Can I check whether the object is the default or not?

martenole: Not directly. Either by going through the values and checking that everything is at the default. Or we need to add an additional data member to the CCDB object for that. But that would mean that we have to recreate at least the default object for the CCDB. So I would simply check all values. Shouldn't hurt too much, since it needs to be done only once

Detectors/TRD/base/src/TrackletTransformer.cxx
```diff
@@ -58,6 +58,9 @@ float TrackletTransformer::calculateDy(int detector, int slope)
   double padWidth = mPadPlane->getWidthIPad();
 
   float vDrift = mCalVdriftExB->getVdrift(detector);
+  if (vDrift == 0.) {
```
martenole: why would `vDrift` be 0 actually?

f3sch: On my smallish (100ctfs) dataset I test upon, I saw this line after fitting:
image
From this chamber: 
image
Ran it like this:
image


Hence I thought that maybe checking for this, might be worthwhile before we throw we an exception.
I did not investigate why this really happens to fit exactly zero here...

martenole: Ok, so the fit might yield zero. Then I would suggest we change in the calibrator the range for the fit parameter such the minimum value is something like 0.01 or so. We might have a chamber where the HV is strongly reduced or OFF. Then we of course don't want to put just the default vDrift value, but the minimum one instead. And then we can remove this check, or, in case `vDrift` is too close to zero just throw an error

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -101,57 +104,104 @@ void CalibratorVdExB::initProcessing()
...
+{
+  static bool doneOnce = false;
+  if (!doneOnce) {
+    doneOnce = true;
+    mFitFunctor.vdPreCorr.fill(1.546);
+    mFitFunctor.laPreCorr.fill(0.);
+    // We either get a pointer to a valid object in the last error or to the default object
+    // which is always present. The first has precedence over the latter.
+    auto dataCalVdriftExB = pc.inputs().get<o2::trd::CalVdriftExB*>("calvdexb");
+    if (!dataCalVdriftExB) {
```
martenole: In this case you would get a fatal error from the framework, so no need to check this

f3sch: Ok.

Detectors/TRD/base/src/TrackletTransformer.cxx
f3sch: I removed the check and set the fit range to minimum of 0.01.

On another note. Just a thought I had, should we maybe prepare some code to store which chamber has what status e.g. HV reduced, HV off, Capacitor 4.7/removed etc.? Then we could handle these chambers differently in calibration code and it would be easier to see why a chamber behaved badly.

Detectors/TRD/workflow/include/TRDWorkflow/VdAndExBCalibSpec.h
```diff
@@ -116,10 +123,13 @@ DataProcessorSpec getTRDVdAndExBCalibSpec()
   using device = o2::calibration::VdAndExBCalibDevice;
   using clbUtils = o2::calibration::Utils;
 
+  auto dataRequest = std::make_shared<o2::globaltracking::DataRequest>();
```
martenole: I am only noticing this now. Why do you add the dataRequest here? It should not be needed at all I think. Neither is the recoContainer

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -101,57 +104,97 @@ void CalibratorVdExB::initProcessing()
...
   mInitDone = true;
 }
 
+void CalibratorVdExB::retrievePrev(o2::framework::ProcessingContext& pc)
+{
+  static bool doneOnce = false;
+  if (!doneOnce) {
+    doneOnce = true;
+    mFitFunctor.vdPreCorr.fill(1.546);
+    mFitFunctor.laPreCorr.fill(0.);
```
martenole: Now you have nicely added some default parameters to the constants, but you are not using them...
But I think we are converging here :)

Detectors/TRD/workflow/include/TRDWorkflow/VdAndExBCalibSpec.h
f3sch: I think this was some dead code I forgot to delete...

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
f3sch: Good catch :small_airplane:

f3sch:
CI fails seem to be related to O2Physics?

martenole:
Also after I have merged my PR there is a  conflict in the README which needs to be fixed

martenole (approved):

martenole:
Error unrelated, merging

PR #10055, opened 2022-10-11
L1phase correction added; fix L1calibration

peressounko:
Fix in L1phase calibrator;
Add L1phase calibration to reconstruction procedure.


peressounko:
Dear @shahor02, could you please have a look at this PR? Failed test was OK in previous attempt, before I changed to pointer  ctx.inputs().get<std::vector<int>*>("l1phase") and now it looks unrelated to PHOS.

shahor02 (approved):

PR #10056, opened 2022-10-12
DPL: allow ExternalFairMQDeviceProxy to work with multiple input channels

ktf:


ktf:
@davidrohr @chiarazampolli this should be what is needed for the multiple inputs. It does not seem to have any regressions, however I could not extend the test yet to have multiple ones, so I do not know if it actually works in that case. I will not be able to work on this until late Thursday.

chiarazampolli:
Hello @ktf , 
Thanks! As soon as the tests are green, we could test it as an alternative O2 at P2. 
In addition, I wonder whether we should have a separate test for the dcs-proxy (DPs), dcs-config-proxy (configurations) and cpt-proxy, to verify that nothing is broken, @shahor02 .

Chiara

ktf:
@davidrohr @chiarazampolli all tests pass, so I will merge this so that we can try this out as discussed. If there is any issue, the development is self contained and we can revert easily.

PR #10058, opened 2022-10-12
[EMCAL-841, EMCAL-830] General tool for cell-level calibration

mfasDa:
Outsource all calibrations applied in the cell-recalibrator-
workflow into a stand-alone tool that can be used on-the-
fly. CCDB objects must be set from outside. Two methods
applying calibrations are provided, one for a single cell
and one for a collection of cells (both templated in order
to support different objects inplemeting the CellInterface)

### DIFF
--- a/Detectors/EMCAL/calib/include/EMCALCalib/CellRecalibrator.h
+++ b/Detectors/EMCAL/calib/include/EMCALCalib/CellRecalibrator.h
@@ -1,0 +1,205 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+#ifndef ALCEO2_EMCAL_CELLRECALIBRATOR_H
+#define ALCEO2_EMCAL_CELLRECALIBRATOR_H
+
+#include <exception>
+#include <iosfwd>
+#include <optional>
+#include <vector>
+#include <gsl/span>
+
+#include "Rtypes.h"
+
+#include <DataFormatsEMCAL/Constants.h>
+#include <EMCALCalib/BadChannelMap.h>
+#include <EMCALCalib/TimeCalibrationParams.h>
+#include <EMCALCalib/GainCalibrationFactors.h>
+
+namespace o2
+{
+
+namespace emcal
+{
+
+/// \class CellRecalibrator
+/// \brief Tool for recalibration at cell level
+/// \ingroup EMCALcalib
+/// \author Markus Fasel <markus.fasel@cern.ch> Oak Ridge National Laboratory
+/// \since Oct 12, 2022
+///
+/// Applying cell-level calibrations
+/// - bad channel removal
+/// - time shift
+/// - gain calibration
+///
+/// Attention: All calibrations for which calibration objects
+/// are provided are applied. Check for active calibration is
+/// therefore related to the presence of the corresponding
+/// calibration object.
+///
+/// Input can be a single cell (getCalibratedCell) or a
+/// collection of cells (getCalibratedCells). In case of
+/// single cell the result is optional since the cell can
+/// be rejected by the bad channel mask. Only cells of type
+/// high gain or low gain can be calibrated, in case cells
+/// of other types (LEDMON/TRU) are passed to getCalibratedCell
+/// an exception will be thrown.
+///
+/// The calibrator supports all cells of the CellInterface
+/// concept.
+class CellRecalibrator
+{
+ public:
+  /// \class CellTypeException
+  /// \brief Handling of invalid cell types in calibration
+  class CellTypeException : public std::exception
+  {
+   public:
+    /// \brief Constructor
+    CellTypeException() = default;
+
+    /// \brief Destructor
+    ~CellTypeException() noexcept final = default;
+
+    /// \brief Get error message of the exception
+    /// \return Error message
+    const char* what() const noexcept final
+    {
+      return "Only possible to calibrate cells of type high gain or low gain";
+    }
+  };
+
+  /// \brief Constructor
+  CellRecalibrator() = default;
+
+  /// \brief Destructor
+  ~CellRecalibrator() = default;
+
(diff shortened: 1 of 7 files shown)


Detectors/EMCAL/calib/include/EMCALCalib/CellRecalibrator.h
```diff
@@ -0,0 +1,205 @@
...
+  ///
+  /// Applying all calibrations provided based on presence of calibration objects. Only
+  /// cells of type high-gain or low-gain can be calibrated. Since the cell can be rejected
+  /// by the bad channel calibration the return type is std::optional, where an empty optional
+  /// reflects rejected cells.
+  ///
+  /// \param inputcell Cell to calibrate
+  /// \return Calibrated cell (empty optional if rejected by the bad channel map)
+  /// \throw CellTypeException in case the inputcell is neither of type high gain nor low gain
+  template <typename T>
```
shahor02, line 133: Why using template and not std::optional<Cell>? It anyway works only for Cells

Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
```diff
@@ -57,15 +58,11 @@ void CellRecalibratorSpec::run(framework::ProcessingContext& ctx)
       outputtriggers.push_back(nexttrigger);
```
shahor02: can be done directly w/o intermediate object
```
outputtriggers.emplace_back(trg.getBCData(), currentfirst, trg.getNumberOfObjects()).setTriggerBits(trg.getTriggerBits());
```

Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
```diff
@@ -57,15 +58,11 @@ void CellRecalibratorSpec::run(framework::ProcessingContext& ctx)
...
-      }
+    auto calibratedCells = mCellRecalibrator.getCalibratedCells(gsl::span<const o2::emcal::Cell>(inputcells.data() + trg.getFirstEntry(), trg.getNumberOfObjects()));
+    if (calibratedCells.size()) {
+      std::copy(calibratedCells.begin(), calibratedCells.end(), std::back_insert_iterator(outputcells));
     }
-    o2::emcal::TriggerRecord nexttrigger(trg.getBCData(), currentfirst, cellsEvent);
+    o2::emcal::TriggerRecord nexttrigger(trg.getBCData(), currentfirst, calibratedCells.size());
     nexttrigger.setTriggerBits(trg.getTriggerBits());
     outputtriggers.push_back(nexttrigger);
     currentfirst = outputcells.size();
```
shahor02, line 64: same comment

Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
```diff
@@ -86,45 +83,29 @@ void CellRecalibratorSpec::finaliseCCDB(o2::framework::ConcreteDataMatcher& matc
...
-
 void CellRecalibratorSpec::updateCalibObjects()
 {
-  auto badmapOld = mBadChannelMap;
-  auto timeCalibOld = mTimeCalibration;
+  bool hasUpdate = false;
   if (isRunBadChannlCalibration()) {
-    mBadChannelMap = mCalibrationHandler->getBadChannelMap();
-    if (badmapOld != mBadChannelMap) {
+    if (mCellRecalibrator.getBadChannelMap() != mCalibrationHandler->getBadChannelMap()) {
```
shahor02: not sure such an object is 100% reliable (unless the object never changes, for the 1st object this is surely fine): on update, the old cached object is destroyed and instead a new one is created, theoretically it may have the same pointer.

Detectors/EMCAL/calib/include/EMCALCalib/CellRecalibrator.h
mfasDa: The reason for templating the function is that we are planning to modify Florian's pull request to introduce a second uncompressed cell type as transient object with the same interface and the class should support both (the compressed one during the online stage for ED and QC, the transient one starting from the asynchonous stage). Obviously we found a necessity for this step.

Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
mfasDa: OK, will change this.

Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
mfasDa: I was also worrying about that. Anything I can use from the framework to identify whether the object has changed? Once the objects from the online calibration are used they often have a finer-than-run granularity.

shahor02: No, the framework just calls finaliseCCDB on the 1st `pc.inputs().get...` for updated object, the rest is on the author of the device.
To flag object update in the helper I use dedicated data members, e.g. https://github.com/AliceO2Group/AliceO2/blob/cb98b7e5f036024dfce73abf2b6bdb3ad432d1b4/GPU/TPCFastTransformation/CorrectionMapsHelper.h#L77-L83 which I then reser when consuming the object by https://github.com/AliceO2Group/AliceO2/blob/cb98b7e5f036024dfce73abf2b6bdb3ad432d1b4/GPU/TPCFastTransformation/CorrectionMapsHelper.h#L90

mfasDa:
Discussion concerning separation compressed / uncompressed cell class: #10071

Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
mfasDa: I see, should be straight forward to adapt this in the CalibLoader (which I adapted from your CCDB helper classes)

mfasDa: Hi @shahor02 , I have changed the calib loader accordingly. Could you please review? The bitset size is chosen to be 16 in order to allow for other calibration objects to be handled by the calib loader and to make this a general tool for EMCAL CCDB access from DPL workflows.

Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
mfasDa: Done.

Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
mfasDa: Done

shahor02 (approved):

PR #10059, opened 2022-10-12
Limit link loop to maximum number of expected links

wiechula:
* Add mapping of number of links per region and EP in Mapper
* Limit link loop to maximum number of expected links in idc to vector
  - Should fix non proper link mask in raw data


matthias-kleiner:
@davidrohr  @shahor02 can this be merged? It would be really good to have this PR in the nightly build for the next FLP suite. Otherwise we probably cannot use the IDCs due to a bug in the firmware.

wiechula:
The failure is unrelated it seems. cs8 CI might be hanging.

PR #10060, opened 2022-10-12
TRD: calibration: Separated calibration methods

f3sch:
This patch separates the future gain from vdexb calibration clearly. As a consequence it introduces a new interface for o2-calibration-trd (renamed) as a generic binary and gives a bare bones structure for gain calibration.
The various classes introduced are subject to change in the future.

One can then select and configure different calibration methods:

1. `--vdexb` and `--enable-root-input-vdexb` for VDrift and ExB calibration, configurable with `--vdexb-opts`
2. `--gain` and `--enable-root-input-gain` for gain calibration, configurable with `--gain-opts`

This can easily be extended for other possible calibrations in the future.

Signed-off-by: Felix Schlepper <felix.schlepper@cern.ch>

### DIFF
--- a/DataFormats/Detectors/TRD/CMakeLists.txt
+++ b/DataFormats/Detectors/TRD/CMakeLists.txt
@@ -35,6 +35,7 @@ o2_target_root_dictionary(DataFormatsTRD
                        include/DataFormatsTRD/CalibratedTracklet.h
                        include/DataFormatsTRD/DcsCcdbObjects.h
                        include/DataFormatsTRD/AngularResidHistos.h
+                       include/DataFormatsTRD/GainCalibration.h
                        include/DataFormatsTRD/HelperMethods.h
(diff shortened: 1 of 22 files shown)


martenole (changes requested):
Could you please wait for https://github.com/AliceO2Group/AliceO2/pull/9955 and rebase afterwards since I already changed there the tr-calib-workflow.cxx?
And then I would mark this as WIP. Since you are doing the gain calibration implementation yourself I don't see why we cannot directly add the first version here instead of the empty classes? Maybe not everything you have added here is actually needed (for example some readers/writers). That would give a much cleaner and simpler to review PR

f3sch:
Aye.

f3sch:
Closing this since #9955 already introduced the better interface and I am actually not doing the gain calibration after all. This will be now a service task.

PR #10061, opened 2022-10-12
Infologger cleanup

martenole:
- for PHOS see https://alice-qc-shifter.docs.cern.ch/pdp/infologgerMessages/#phos
- for TPC-ITS, also since with the multi-line message due to the `asString` method of the track parameters its not possible to filter out these errors via the text message. And since the data taking can continue I believe a warning should be OK but maybe @shahor02 wants to keep it?
- for TRD @bazinski and @tdietel  I saw that the mismatches are not not coming from incorrectly connected links anymore, but from corrupted headers instead. This I conclude, because you see data from one half-chamber where the header is correct and later on from the same link you get a different half-chamber ID from the TrackletHCHeader. This does not happen for every TF, but often enough to flood the InfoLogger. Now the error `TrackletHCHeaderFailure` is incremented and sent to QC. I will not just take the HCID from the header anymore, since in the worst case one puts data in the wrong half-chamber. In case of a wrongly connected link is still there we should see a spike in the QC plots

### DIFF
--- a/Detectors/GlobalTracking/src/MatchTPCITS.cxx
+++ b/Detectors/GlobalTracking/src/MatchTPCITS.cxx
@@ -1336,8 +1336,9 @@ bool MatchTPCITS::refitTrackTPCITS(int iTPC, int& iITS)
     auto posStart = tracOut.getXYZGlo();
     auto tImposed = timeC * mTPCTBinMUSInv;
     if (std::abs(tImposed - mTPCTracksArray[tTPC.sourceID].getTime0()) > 550) {
-      LOGP(error, "Impossible imposed timebin {} for TPC track time0:{}, dBwd:{} dFwd:{} TB | ZShift:{}, TShift:{} | Trc: {}", tImposed, mTPCTracksArray[tTPC.sourceID].getTime0(),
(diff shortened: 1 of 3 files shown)


Detectors/GlobalTracking/src/MatchTPCITS.cxx
```diff
@@ -1336,7 +1336,7 @@ bool MatchTPCITS::refitTrackTPCITS(int iTPC, int& iITS)
     auto posStart = tracOut.getXYZGlo();
     auto tImposed = timeC * mTPCTBinMUSInv;
     if (std::abs(tImposed - mTPCTracksArray[tTPC.sourceID].getTime0()) > 550) {
-      LOGP(error, "Impossible imposed timebin {} for TPC track time0:{}, dBwd:{} dFwd:{} TB | ZShift:{}, TShift:{} | Trc: {}", tImposed, mTPCTracksArray[tTPC.sourceID].getTime0(),
+      LOGP(alarm, "Impossible imposed timebin {} for TPC track time0:{}, dBwd:{} dFwd:{} TB | ZShift:{}, TShift:{} | Trc: {}", tImposed, mTPCTracksArray[tTPC.sourceID].getTime0(),
```
davidrohr: we could leave only the first line alarm, and make the rest info?

martenole: True, will change that

bazinski (approved):

PR #10063, opened 2022-10-12
MUON: fixed time matching of MFT and MCH tracks

aferrero2707:
- skip empty MCH ROFs when collecting MCH tracks associated to a given MFT ROF
- do not skip MCH ROFs older than the current TF (in future they will be removed in the MCH digits filtering)


aferrero2707:
@rpezzi could you please have a look? Thanks!

rpezzi (approved):
Thanks for spotting and fixing this, Andrea.

PR #10064, opened 2022-10-13
[MFT] Increase MaxCellNeighbours on CA track finder

rpezzi:
Fixes eventual crashes on EPNs


sawenzel (approved):
This is fine for me since it will likely prevent crashes seen during data taking.

In a second step  I would propose to make this more robust in the sense of
having a dynamic memory management for cluster neighbours inside Cells.

This should be as efficient, if cell objects can be made long-lived: Reuse cells object in rows --> This means that cells should not be destructed when a row is reset but rather cells should be cleared and reused.

rpezzi:
Sure, this PR is a temporary solution.
Dynamic memory allocation is more elegant but may add overhead and slow down the tracker. The memory overhead of static array here is minimum. I am a more concerned with the time the tracker spends on such situations, which may block the tracker for a while if there are so many candidates which likely results in bad tracks.

rpezzi:
Hi, the failed CI test is not related (incomplete O2Physics compilation). This should be good to be merged. Thanks.

PR #10065, opened 2022-10-13
MCH: handle data truncation in User Logic

aferrero2707:
Added correct handling of the "truncated data" UserLogic error flag.
This error is set by the UserLogic when it receives too many consecutive data packets from one e-Link, without intermediate SYNC words.

### DIFF
--- a/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/ErrorCodes.h
+++ b/Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/ErrorCodes.h
@@ -31,12 +31,13 @@ enum ErrorCodes {
   ErrorBadClusterSize = 1 << 6,              // 64
   ErrorBadIncompleteWord = 1 << 7,           // 128
   ErrorTruncatedData = 1 << 8,               // 256
-  ErrorUnexpectedSyncPacket = 1 << 9,        // 512
-  ErrorBadELinkID = 1 << 10,                 // 1024
-  ErrorBadLinkID = 1 << 11,                  // 2048
-  ErrorUnknownLinkID = 1 << 12,              // 4096
-  ErrorBadHBTime = 1 << 13,                  // 8192
-  ErrorNonRecoverableDecodingError = 1 << 14 // 16384
+  ErrorTruncatedDataUL = 1 << 9,             // 512
+  ErrorUnexpectedSyncPacket = 1 << 10,       // 1024
+  ErrorBadELinkID = 1 << 11,                 // 2048
+  ErrorBadLinkID = 1 << 12,                  // 4096
+  ErrorUnknownLinkID = 1 << 13,              // 8192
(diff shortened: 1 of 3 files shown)


Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/ErrorCodes.h
```diff
@@ -31,12 +31,13 @@ enum ErrorCodes {
   ErrorBadClusterSize = 1 << 6,              // 64
   ErrorBadIncompleteWord = 1 << 7,           // 128
   ErrorTruncatedData = 1 << 8,               // 256
-  ErrorUnexpectedSyncPacket = 1 << 9,        // 512
-  ErrorBadELinkID = 1 << 10,                 // 1024
-  ErrorBadLinkID = 1 << 11,                  // 2048
-  ErrorUnknownLinkID = 1 << 12,              // 4096
-  ErrorBadHBTime = 1 << 13,                  // 8192
-  ErrorNonRecoverableDecodingError = 1 << 14 // 16384
+  ErrorTruncatedDataUL = 1 << 9,             // 512
```
aphecetche, line 34: I understand the appeal of grouping `TruncatedData` and `TruncatedDataUL` in this definition, but it changes the meaning of previously used error code. Why not simply adding the new error "type" at the end ?

aferrero2707: The numerical values of the error codes are currently not used anywhere in the code, apart from the UL decoder where the errors are emitted and the `errorCodeAsString()` function.

For example:
```
[flp@mchflp1 O2]$ grep -r -n -e ErrorBadHeartBeatPacket Detectors/MUON/
Detectors/MUON/MCH/Raw/Decoder/include/MCHRawDecoder/ErrorCodes.h:29:  ErrorBadHeartBeatPacket = 1 << 4,          // 16
Detectors/MUON/MCH/Raw/Decoder/src/ErrorCodes.cxx:58:  if (ec & ErrorBadHeartBeatPacket) {
Detectors/MUON/MCH/Raw/Decoder/src/UserLogicElinkDecoder.h:210:            sendError(static_cast<int8_t>(mSampaHeader.chipAddress()), static_cast<uint32_t>(ErrorBadHeartBeatPacket));

[flp@mchflp1 QualityControl]$ grep -r -n -e ErrorBadHeartBeatPacket Modules/MUON/
[flp@mchflp1 QualityControl]$ 
```

The QC uses the `errorCodeAsString()` function to convert the numerical values to human-readable text that is then used for the bin labels in the plots.
Having the `TruncatedData*` errors contiguous in value allows to place them automatically in two contiguous bins in the QC plots, which is quite convenient...

aphecetche: Ok. Fair enough.

aphecetche (approved):

PR #10068, opened 2022-10-13
[EMCAL-630] Add static methods for error code names and titles for all reconstruction error classes

mfasDa:
Provide methods for creating error names, titles and
descriptions for all error classes used in the reconstruction.
Monitoring information is nowin one place, together with
the exception it relates to. Other tickets affected:
- EMCAL-534: ALTRO decoding error
- EMCAL-686: Raw fitter error
- EMCAL-629: Page decoding errors

### DIFF
--- a/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/ErrorTypeFEE.h
+++ b/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/ErrorTypeFEE.h
@@ -171,6 +171,20 @@ class ErrorTypeFEE
   /// Helper function, called in the output stream operator for the ErrorTypeFEE
   void PrintStream(std::ostream& stream) const;
 
+  /// \brief Get the number of error types
+  /// \return Number of error types (including undefined)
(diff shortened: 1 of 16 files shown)


mfasDa:
WIP - Don't merge yet

mfasDa:
PR is particularly of relevance for the raw error task in the QC: For the monitoring histograms it defines number of bins and bin labels inside the QC. In case error codes change (i.e. new error codes for new hardware failures are added, or codes are moved) this requires an update in both O2 and QualityControl, which always has to be done in sync. Better have functionality creating names, titles and short description together with the error codes in O2, like this the QC task can use them to initialize the histograms, and if codes change the QC will automatically adapt to this without changing the task.

mfasDa:
All reconstruction error classes are now covered by unit tests. Change ready to merge after successful test.

mfasDa:
CI errors seem unrelated

shahor02:
@mfasDa indeed fullCI failures were unrelated and the reasons are fixed now. Given that your PR is quite large, better have it green in fullCI, could you push some dummy commit to restart it?

mfasDa:
All checks green now

PR #10070, opened 2022-10-13
TPC: Allow manual setting of common mode coupling fraction

wiechula:


wiechula:
It looks like the failing test are unrelated. Can this be merged?

PR #10071, opened 2022-10-13
[WIP][EMCAL-551] improvement of emcal cell energy compression

fjonasALICE:
- this PR is work-in-progress and contains multiple changes related to EMCal cells
- the class Cell.h which was used for CTF decoding and encoding was renamed to CellCompressed.h, including improvements for the energy compression, which is now performed differnetly for HG and LG cells to enhance the effective precision given the number of available bits
- the former Cell class is now replaced with a class that stores the energy etc in float precision (no compression) and is intended to be used from CTF onwards from now on (this change was needed because it was discovered that the energy calibration might cause problems with the enegry range dependent compression)
- the chi2 information was added to CTF encoding and decoding
- a check for the CTF dictionary version was implemented to allow for backward compatibility of old data

### DIFF
--- a/Detectors/EMCAL/base/src/ClusterFactory.cxx
+++ b/Detectors/EMCAL/base/src/ClusterFactory.cxx
@@ -84,6 +84,7 @@ o2::emcal::AnalysisCluster ClusterFactory<InputType>::buildCluster(int clusterIn
   clusterAnalysis.setNCells(inputsIndices.size());
 
   std::vector<unsigned short> cellsIdices;
+  std::vector<float> ampfractions;
 
   for (auto cellIndex : inputsIndices) {
     cellsIdices.push_back(cellIndex);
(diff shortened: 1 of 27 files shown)


Detectors/EMCAL/base/src/ClusterFactory.cxx
```diff
@@ -102,6 +103,21 @@ o2::emcal::AnalysisCluster ClusterFactory<InputType>::buildCluster(int clusterIn
   evalCoreEnergy(inputsIndices, clusterAnalysis);
   evalTime(inputsIndices, clusterAnalysis);
 
+  // set amp fractions
```
fjonasALICE, line 106: this part is unrelated to the cell changes but still made it into this commit. Still useful addition though :-)

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
```diff
@@ -169,8 +206,16 @@ o2::ctf::CTFIOSize CTFCoder::decode(const CTF::base& ec, VTRG& trigVec, VCELL& c
   uint32_t firstEntry = 0, cellCount = 0;
   o2::InteractionRecord ir(header.firstBC, header.firstOrbit);
 
+  CellCompressed cellCompressed;
   Cell cell;
   TriggerRecord trg;
+
+  // perform a check on dict we have old CTF with different energy compression
+  Bool_t isVersion1 = kFALSE;
+  if (header.majorVersion == 0 && header.minorVersion == 1) {
```
fjonasALICE, line 215: we need to validate if this check is correct, this is my guess how it should be done

mfasDa:
The necessity for this is that we found during the energy calibration a small amount of cells with large decalibration (up to factor 4). Once we include the gain type to define energy ranges in order to improve the precision at low E with finer granularity (0 - HGLG transition ~ 16 GeV) we will get a problem with the cell when applying the energy calibration: For those cells which have large decalibration we would force the signal into the saturation range in both directions. We also cannot easily redefine the gain type to deal define energy ranges as it is needed in the time calibration. Like this we must deal with the energy calibration on an uncompressed level. Furthermore it makes also sense to use an uncompressed version at transient level. The compressed cell will then only be used for I/O to/from CTF for en/decoding, while the publishers send the uncompressed version. For the naming convention at long term it makes more sense that the uncompressed Cell class which will be backbone of the user interface gets the name Cell and the compressed version will be renamed to CompressedCell.

PR #10073, opened 2022-10-13
MCH: minor fix in bad channel object metadata

aphecetche:
Instead of indicating the total number of entries seen in the whole detector (that is not really helpful _and_ was overshooting the max int and hence giving negative values), now indicate the mean number of entries per channel.
Should be around 2000 in the current pedestal calibration procedure.
@aferrero2707


aphecetche:
@csuire

aphecetche:
@pillot could you have a look ?

aferrero2707 (approved):
Looks ok for me.

pillot (approved):

PR #10048, opened 2022-10-10
TPC SCD outlier filtering refactoring

martenole:
Outlier filtering is moved from TrackResiduals to TrackInterpolation class in order to allow for setting the maximum number of successfully processed tracks.
Some functionality is removed which allowed to process converted residuals from Run 2. This adds some cleaner separation. TPC cluster residuals are collected in TrackInterpolation. TrackResiduals is for creating the average map from these residuals.
Also added some reader/writer for unbinned residuals to run the aggregator in a standalone mode.

PR #10074, opened 2022-10-13
Add Lumi data members to TPCTransform object

shahor02:


shahor02:
@martenole this can be used to set the average lumi and its error for the produced map.

martenole:
Thanks! I will add the PR with the optional lumi input after my vacation. Don't want to add more to #10048 now

PR #10075, opened 2022-10-13
[EMCAL-623, EMCAL-505] Fix doxygen in clusterizer

mfasDa:
- Move existing documentation from cxx to header file,
  like this it can be also used by modern IDEs
- Add documentation of parameters and return values
  missing so far


mfasDa:
Trivial commit - only doxygen comments affected.

PR #10076, opened 2022-10-14
GPU: Add predicateSum / predicateScan functions in TPC Cluster Finder.

fweig:


davidrohr (dismissed):

davidrohr:
```
ptxas /tmp/tmpxft_000034dc_00000000-6_GPUReconstructionCUDAKernels.ptx, line 131235; error   : Instruction 'vote' without '.sync' is not supported on .target sm_70 and higher from PTX ISA version 6.4
```
I assume the failure might come from the minimum CUDA architecture we are compiling for, which is defined here: https://github.com/AliceO2Group/AliceO2/blob/dev/dependencies/FindO2GPU.cmake#L47

Though, I am seeing the CI sets `-DCUDA_COMPUTETARGET=86`, so it should be 86, not virtual_61 which is failing.

fweig:
> ```
> ptxas /tmp/tmpxft_000034dc_00000000-6_GPUReconstructionCUDAKernels.ptx, line 131235; error   : Instruction 'vote' without '.sync' is not supported on .target sm_70 and higher from PTX ISA version 6.4
> ```
> 
> I assume the failure might come from the minimum CUDA architecture we are compiling for, which is defined here: https://github.com/AliceO2Group/AliceO2/blob/dev/dependencies/FindO2GPU.cmake#L47
> 
> Though, I am seeing the CI sets `-DCUDA_COMPUTETARGET=86`, so it should be 86, not virtual_61 which is failing.

This function was removed in cuda 7.0 . But it's still used in HIP. I've pushed a workaround.

davidrohr:
#10089 might solve the Thrust problem, let's see after that is merged.

PR #10077, opened 2022-10-14
DPL: avoid crash on nullptr in proxy

ktf:
In principle I do not expect this to happen, but just in case.


ktf:
@dennisklein @rbx is there any case in which a fairmq::Parts would contain a unique ptr set to nullptr?

rbx:
You will not have `nullptr` after receiving into empty Parts.
Parts is a wrapper around `std::vector<MessagePtr>`.
But you could of course just put `nullptr` inside yourself. It would also stay there if you receive into this Parts object, since receive appends to the vector.

We also check for `nullptr`s inside of parts on sending and error if we find any. But we don't put `nullptr`s inside ourselves, if I remember right.

So the check is not wrong.

davidrohr:
this doesn't fix the crash we are seeing (tested in environment id 2b8Hxbczvnd)

PR #10082, opened 2022-10-14
Determine PDG properties from G4 ions

benedikt-voelkel:
* add executable which tries to determine unknown ion PDG properties and writes an updated PDG table (based on what was available from O2DatabasO2DatabasePDG)

* add PDG text file which contains already many ion PDGs that were unknown until now (digestable via TDatabasePDG::ReadPDGTable) --> loaded automatically by O2DatabasePDG

* use O2 VMC setup to ensure same physics setup as used in production

* NOTE: The ion properties are determined for the ground states, so no special isomere properties are calculated --> might mean only an approximation for the particle mass for instance.


benedikt-voelkel:
The bad spacing seems to complain about the added `Detectors/gconfig/data/extra_ions_pdg_table.dat`. However, that was automatically created from `TDatabasePDG::WritePDGTable` and I think we should leave it as is.

sawenzel (approved):

PR #10085, opened 2022-10-14
User hook (JITable) for track references during stepping

sawenzel:
This commit implements a generic and customizable mechanism to define and execute a callback during MC stepping. The purpose of the callback is to create TrackReferences for certain conditions, but can be used for other purposes (monitoring etc). The callback just takes the instance to TVirtualMC

The default implementation creates a TrackReference on "IsStop" whenever the current track already left a TrackReference previously.

The feature needs to be switched on via configurable Parameter

`SimCutParams.stepTrackRefHook=1`

Furthermore, the user can provide own callback implementations via a macro file "MyHook.C". The implementation should follow the following pattern:
```
o2::steer::O2MCApplicationBase::TrackRefFcn trackRefHook() {                                                                               return [](TVirtualMC const* vmc) {                                                                                                         auto stack = (o2::data::Stack*)vmc->GetStack();                                                                                          if (vmc->IsTrackStop() && stack->currentTrackLeftTrackRef()) {                                                                              // we add a stopping TrackRef when the current track already                                                                             // registered previous TrackRefs                                                                                                         stack->addTrackReference(o2::TrackReference(*vmc, 0));                                                                                }                                                                                                                                      };                                                                                                                                     }
```

MyHook.C will be activated with

`SimCutParams.mCutParams.stepTrackRefHookFile=PATHTO/MyHook.C`

### DIFF
--- a/Common/SimConfig/include/SimConfig/SimParams.h
+++ b/Common/SimConfig/include/SimConfig/SimParams.h
@@ -23,6 +23,9 @@ namespace conf
 // (mostly used in O2MCApplication stepping)
 struct SimCutParams : public o2::conf::ConfigurableParamHelper<SimCutParams> {
   bool stepFiltering = true; // if we activate the step filtering in O2BaseMCApplication
+  bool stepTrackRefHook = false;                                                                              // if we create track references during generic stepping
+  std::string stepTrackRefHookFile = "${O2_ROOT}/share/Detectors/gconfig/StandardSteppingTrackRefHook.macro"; // the standard code holding the TrackRef callback
+
   bool trackSeed = false;    // per track seeding for track-reproducible mode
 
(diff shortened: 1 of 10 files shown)


miranov25:
Reading the source code the default callback (if enabled) is one
https://github.com/sawenzel/AliceO2/blob/f3de58ee84d12be64f4e32f4627a77675aa7d550/Detectors/gconfig/src/StandardSteppingTrackRefHook.macro

For me it is good default

miranov25:
**The standard hook is fine with me.**
**I suggest turning it on by default so that you only create references when particles have been created before.** The increase in data volume is minimal in this case of the default hook. We will have much less isStop references than the number of particles.

I understand that this might not be the case once the hook changes, but it is definitely an expert action. Also, we will never create more TrackRefernces than there are particles in the stack.

PR #10086, opened 2022-10-14
Alignment fixes + related generic helpers

shahor02:
not fully operational yet, pushing to dev to reduce the time of switching between branches


shahor02:
@bazinski Note that I've modified the Transformer files to make them usable in the alignment code

PR #10087, opened 2022-10-15
Fix clang-tidy

ktf:


ktf:
@shahor02 fullCI fails otherwise. Merging.

PR #10091, opened 2022-10-17
Protects SSE specific code; Make compileable on ARM

sawenzel:


sawenzel:
@davidrohr : That's the only change needed to make O2 compile on aarch64. I couldn't do it in the CMake. The previous fix to filter out "Darwin" is not easily generalizable to ARM on Linux (... or maybe there a way that I missed ...).

davidrohr (dismissed):

PR #10092, opened 2022-10-17
Update for MC-only AOD producer

nburmaso:
* Add initialization from list of kinematics: e.g. `--mckine-fnames "bkg,sgn_1"`
* Remove dependence from collisioncontext
* Do not write BCs table

### DIFF
--- a/Detectors/AOD/src/AODMcProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODMcProducerWorkflowSpec.cxx
@@ -9,14 +9,12 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-/// @file   AODProducerWorkflowSpec.cxx
+/// @file   AODMcProducerWorkflowSpec.cxx
 
 #include "AODProducerWorkflow/AODMcProducerWorkflowSpec.h"
 #include "AODProducerWorkflow/AODProducerHelpers.h"
-#include "CommonUtils/NameConf.h"
 #include "MathUtils/Utils.h"
 #include "CCDB/BasicCCDBManager.h"
-#include "CommonDataFormat/InteractionRecord.h"
 #include "DataFormatsGlobalTracking/RecoContainer.h"
 #include "Framework/AnalysisDataModel.h"
 #include "Framework/ControlService.h"
@@ -26,15 +24,11 @@
 #include "Framework/TableBuilder.h"
 #include "Framework/TableTreeHelpers.h"
 #include "Framework/CCDBParamSpec.h"
-#include "ReconstructionDataFormats/VtxTrackIndex.h"
 #include "SimulationDataFormat/MCTrack.h"
 #include "SimulationDataFormat/MCUtils.h"
 #include "O2Version.h"
 #include "TMath.h"
 #include "TString.h"
-#include "TObjString.h"
-#include <map>
-#include <string>
 #include <vector>
 
 using namespace o2::framework;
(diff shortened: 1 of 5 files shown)


Detectors/AOD/src/AODMcProducerWorkflowSpec.cxx
```diff
@@ -223,102 +194,66 @@ void AODMcProducerWorkflowDPL::init(InitContext& ic)
...
   auto& mcCollisionsBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "MCCOLLISION"});
   auto& mcParticlesBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "MCPARTICLE_001"});
   auto& originTableBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "ORIGIN"});
 
-  auto bcCursor = bcBuilder.cursor<o2::aod::BCs>();
   auto mcCollisionsCursor = mcCollisionsBuilder.cursor<o2::aod::McCollisions>();
   auto mcParticlesCursor = mcParticlesBuilder.cursor<o2::aod::StoredMcParticles_001>();
   auto originCursor = originTableBuilder.cursor<o2::aod::Origins>();
 
-  auto mcReader = std::make_unique<o2::steer::MCKinematicsReader>("collisioncontext.root");
```
sawenzel, line 246: I think there was a use case when taking collisioncontext was justified: When signals are embedded in background events.
Is the new code still doing this correctly?

(If not I would offer 2 options: One with just kinematics and one where collisioncontext is available with the information of which events are embedded onto which other events).

nburmaso: Hi Sandro

This update was requested by @ddobrigk. If I understand correctly, there is no need in collisioncontext and embedding for MC-only producer, but maybe I misunderstood something.

nburmaso: Hi @sawenzel

I have added an option to use collisioncontext to process embedded events properly. Could you please take a look?

ddobrigk: To be clear, I asked for an optional path in which the collision context isn't needed, but I think this should be an option - there is a use case in which this could be switched on. Thanks a lot, Nazar!!

Steer/src/MCKinematicsReader.cxx
```diff
@@ -175,3 +175,33 @@ bool MCKinematicsReader::initFromKinematics(std::string_view name)
 
   return true;
 }
+
+bool MCKinematicsReader::initFromKinematics(std::vector<std::string> names)
```
sawenzel: is this new function really needed?

nburmaso: Hi @sawenzel

If I understand correctly, in case when digitization context is not used it is not possible to load events from multiple sources using [MCKinematicsReader::initFromKinematics(std::string_view name)](https://github.com/AliceO2Group/AliceO2/blob/720c1efc4713578d2835242318a365cd4ed1ef5e/Steer/src/MCKinematicsReader.cxx#L163-L177 ). To initialize MC reader one needs to access private fields, so I introduced an overloaded function.

sawenzel: In my opinion loading from multiple sources does not make so much sense when not using collisioncontext.

nburmaso: I have removed the overloaded function and MC reader is now initialized from only one file when embedding is not used

sawenzel (approved):

Steer/src/MCKinematicsReader.cxx
sawenzel: Thanks!

PR #10093, opened 2022-10-17
TPC sac: implementation in O2

chiarazampolli:
Goes with https://github.com/AliceO2Group/O2DPG/pull/681
@davidrohr , @matthias-kleiner , @rmunzer

### DIFF
--- a/prodtests/full-system-test/aggregator-workflow.sh
+++ b/prodtests/full-system-test/aggregator-workflow.sh
@@ -51,6 +51,7 @@ if [[ "0$GEN_TOPO_VERBOSE" == "01" ]]; then
   echo "CALIB_TPC_VDRIFTTGL = $CALIB_TPC_VDRIFTTGL" 1>&2
   echo "CALIB_TPC_IDC = $CALIB_TPC_IDC" 1>&2
   echo "CALIB_TPC_IDC_BOTH = $CALIB_TPC_IDC_BOTH" 1>&2
+  echo "CALIB_TPC_SAC = $CALIB_TPC_SAC" 1>&2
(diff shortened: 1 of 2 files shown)


davidrohr (changes requested):
I would propose we merge the aggregator workflows for IDCs and SACs in one. No need to deploy this as 2 collections.
Background is a bit: If we want this TPC stuff on an isolated EPN calib node, which is currently enforced by requesting 128 cores for the collection, we cannot split it in 2 collections.

chiarazampolli:
Ciao @davidrohr , 
Ok, I will change this. 
Chiara

chiarazampolli:
Note, though, that currently there is the possibility to have IDC_A and IDC_B separately. Should this also disappear? Or should I add the SAC to the `AGGREGATOR_TASKS == TPCIDC_BOTH`?

davidrohr:
> Note, though, that currently there is the possibility to have IDC_A and IDC_B separately. Should this also disappear? Or should I add the SAC to the `AGGREGATOR_TASKS == TPCIDC_BOTH`?

I would add it for TPCIDC_BOTH, since that is what we'll use in the long run. Once we have the TPCIDC running, we should also remove the obsolete individual IDC_A and IDC_B. Perhaps for now we can leave them for testing.

chiarazampolli:
@davidrohr , can you check my second commit? Note that I added also some changes to calib-workflow.sh, because I think the output proxy should be added to the EPNs only if we are not calibrating via FLPs.

davidrohr:
didn't try, but looks correct

chiarazampolli:
Hello @davidrohr , 
Should we merge this, or wait?

PR #10097, opened 2022-10-17
Fix fake overlap between ITS wrapper volume and MFT patch panel

mario6829:
An overlap between the ITS wrapper volume 2 and the MFT patch panel has been fixed by slightly reducing the wrapper Z dimension. The overlap is fake, because it was not between two real volumes: the ITS wrapper volume is simply an air-filled container holding the two Outer Barrel staves.


mario6829:
@iouribelikov @mconcas this is the PR that fixes the fake overlap between the ITS wrapper volume and the MFT. Please let me know in case of problems. Thanks and regards

mconcas (approved):

mconcas:
Thanks @mario6829 !

PR #10108, opened 2022-10-18
fix in loading calib object, prototype of using new TOF method

shahor02:


shahor02:
Hi @timo, the fillCI has failed due to the missing alien-token...

Merging since the modified code is anyway not tested in the fullCI

PR #10109, opened 2022-10-18
Further devs for unknown ions

benedikt-voelkel:
* add O2DatabasePDG::Mass(int pdg) interface

* implement logic to try to determine potential isomere state based on ground state

* slightly extend currently known ions

* fix mass in extracted ion table from MeV (units in G4) to GeV (units in ROOT)


benedikt-voelkel:
Again, the trailing white spaces come from the particle table `extra_ions_pdg_table.dat`...

run/g4DetermineUnknownPdgProperties.cxx
```diff
@@ -120,8 +120,10 @@ int main(int argc, char** argv)
       continue;
     }
     auto name = particle->GetParticleName();
-    // but in the end we must add the actual - potentially isomeric - PDG
-    auto ret = pdgDB->AddParticle(name.c_str(), name.c_str(), particle->GetPDGMass(), particle->GetPDGStable(), particle->GetPDGWidth(), particle->GetPDGCharge(), "Ion", pdg);
+    // but in the end we must add the actual - potentially isomeric - PDG, make sure to divide mass by 1000 to go from G4 units (MeV) to ROOT units (GeV)
+    auto ret = pdgDB->AddParticle(name.c_str(), name.c_str(), particle->GetPDGMass() / 1000., particle->GetPDGStable(), particle->GetPDGWidth(), particle->GetPDGCharge(), "Ion", pdg);
+    // We also add the ground state (if that is the same as before, it will not be added anyways)
+    pdgDB->AddParticle(name.c_str(), name.c_str(), particle->GetPDGMass() / 1000., particle->GetPDGStable(), particle->GetPDGWidth(), particle->GetPDGCharge(), "Ion", pdgToLookFor);
```
ihrivnac: In Geant4 VMC this is implemented in [TG4ParticlesManager::AddParticleToPdgDatabase](https://github.com/vmc-project/geant4_vmc/blob/2a8ca929ed5ebf3ba68a98c78e8062c84125aacd/source/physics/src/TG4ParticlesManager.cxx#L97); the method is actually private so it cannot be used directly without changing the code.

Note, that  `particle->GetPDGWidth()` is also in energy units , so it needs to be converted; and the charge passed to TDatabasePDG has to be multiplied by `3` (this is a convention in Root). Also as you include Geant4 VMC classes, you can better use `TG4G3Units::Energy()` instead of 1000.

benedikt-voelkel: Thanks a lot for your comments, I will take that into account for an update to this PR!

ihrivnac: Thanks for the update; the properties added to TDatabasePDG look fine now.

DataFormats/simulation/include/SimulationDataFormat/O2DatabasePDG.h
```diff
@@ -47,9 +49,41 @@ class O2DatabasePDG
...
+      return MassImpl(db->GetParticle(pdg), success);
+    }
+    if (auto particle = db->GetParticle(pdg)) {
+      // see if this ion can be found
+      return MassImpl(particle, success);
+    }
+    // if we are here, try one last time to look for ground state of potential isomere state
+    auto pdgAsString = std::to_string(pdg);
+    pdgAsString.back() = '0';
+    return MassImpl(db->GetParticle(std::stoi(pdgAsString)), success);
```
ihrivnac: Instead of going via a string you can do:
```cpp
auto newPdg = pdg/10 * 10;
```

sawenzel (approved):

PR #10110, opened 2022-10-18
[ITS] Factorise description of ITS inner barrel for ITS3 geometry implementation

fgrosa:
@mconcas this PR factorizes the inner barrel description from the geometry, in order to enable different versions needed for ITS3 studies. I still have to move part of the code from `Detector.cxx` to `DescriptorInnerBarrel.cxx` (such as services, alignment objects), but you can start to have a look and let me know if this makes sense to you. 
I already tested the current version and it gives exactly the same result as the old implementation.

### DIFF
--- a/Detectors/ITSMFT/ITS/base/CMakeLists.txt
+++ b/Detectors/ITSMFT/ITS/base/CMakeLists.txt
@@ -12,9 +12,11 @@
 o2_add_library(ITSBase
                SOURCES src/GeometryTGeo.cxx src/ContainerFactory.cxx
                        src/MisalignmentParameter.cxx
+                       src/DescriptorInnerBarrel.cxx
                PUBLIC_LINK_LIBRARIES O2::DetectorsBase O2::ITSMFTBase)
 
 o2_target_root_dictionary(ITSBase
                           HEADERS include/ITSBase/GeometryTGeo.h
                                   include/ITSBase/ContainerFactory.h
-                                  include/ITSBase/MisalignmentParameter.h)
+                                  include/ITSBase/MisalignmentParameter.h
+                                  include/ITSBase/DescriptorInnerBarrel.h)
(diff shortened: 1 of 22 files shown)


mconcas:
Thanks @fgrosa! I'm cc-ing also @iouribelikov, for revision.
I think we could also maybe setup a brief discussion at one of next wp2s just to for everyone to be on the same page.

iouribelikov:
@fgrosa Thank you very much, Fabrizio.   To me, it looks good.
But the changes are numerous.  Let's quickly ( 10min ) discuss them at the ITS2 soft meeting next Tuesday, as was suggested by @mconcas.  Especially, because we expect Mario ( @mario6829 ) to be connected.

fgrosa:
> @fgrosa Thank you very much, Fabrizio. To me, it looks good. But the changes are numerous. Let's quickly ( 10min ) discuss them at the ITS2 soft meeting next Tuesday, as was suggested by @mconcas. Especially, because we expect Mario ( @mario6829 ) to be connected.

Hi @iouribelikov, thanks a lot for your feedback! Indeed I think that discussing it at the next wp2 meeting is a good idea, there is also no hurry to merge this PR.

fgrosa:
@mconcas @iouribelikov after the last checks, I confirm that this code gives exactly the same result as the old one, so I would say that is ready for review!

iouribelikov (dismissed):
@fgrosa Thank you very much, Fabrizio.  For me, it works fine.
(But, of course, I was able to test the ITS2 part only)

fgrosa:
> @fgrosa Thank you very much, Fabrizio. For me, it works fine. (But, of course, I was able to test the ITS2 part only)

Hi @iouribelikov thanks a lot for testing it! There is nothing for ITS3 yet, so if you tested ITS2 it's good enough.

fgrosa:
After few fixes, finally the CI is green @mconcas @iouribelikov. Do you think we can merge it? Thanks!

mconcas (approved):
Fine with me

PR #10116, opened 2022-10-19
MCH: add some missing include guards

aphecetche:
Spotted one by chance and the others using : 

```
find Detectors/MUON/MCH -name '*.h' -exec grep -L "#ifndef O2_" {} \; | grep -v LinkDef
```


Detectors/MUON/MCH/Constants/DetectionElements.h
```diff
@@ -32,3 +35,5 @@ inline bool isValidDetElemId(int deId)
   return std::find(begin(deIdsForAllMCH), end(deIdsForAllMCH), deId) != deIdsForAllMCH.end();
 }
 } // namespace o2::mch::constants
+
+#endif
```
pillot: not sure if new line at end of file is mandatory, but there is one missing here

pillot (approved):

PR #10117, opened 2022-10-19
Fix wrapper volume extrusions due to previous fix of overlaps with MFT

mario6829:
The fix to the overlaps between the ITS wrapper volume 2 and the MFT patch panels (see PR #10097) was too simplistic: shortening the Z length of the wrapper volume while cured the overlap on the C side at the same time produced an extrusion of child volumes on the A side (namely, the rings of the OB support wheels). In fact, the wrapper volumes are carefully shaped polycones just to avoid fake overlaps and extrusions.
This further fix restores the correct wrapper length on the A side while keeping it shorter on the C side.


mario6829:
Hi @iouribelikov @mconcas here it is the further fix to the wrapper volume shape to avoid the extrusions of the support wheel rings. Cheers

mconcas (approved):

iouribelikov (approved):
@mario6829 It's perfect now. Thank you a lot, Mario.

PR #10122, opened 2022-10-20
DPL: fix clang-tidy autocorrect mistake.

ktf:


ktf:
Checked on hyperloop. Solves current crash. Merging.

ktf:
modernize-return-braced-init-list is simply broken.

ktf:
The issue is that it's not true that `std::vector<std::string>({})` is equivalent to `{{}}`, because the former invokes the copy constructor for an empty vector, while the latter creates a vector with one empty string inside.

ktf:
or at least both clang 14 and GCC 12 generate the same code.

PR #10124, opened 2022-10-20
AOD: add CPV table

sevdokim:
Hello! This commit is for adding CPV table into AOD as discussed on WP4+14 meeting on 17 Oct.
The remaining issue is to handle MC labels.

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -17,6 +17,8 @@
 #include "DataFormatsGlobalTracking/RecoContainer.h"
 #include "DataFormatsCTP/Digits.h"
 #include "DataFormatsCTP/Configuration.h"
+#include "DataFormatsCPV/Cluster.h"
+#include "DataFormatsCPV/TriggerRecord.h"
 #include "DataFormatsITS/TrackITS.h"
 #include "DataFormatsMCH/ROFRecord.h"
 #include "DataFormatsMCH/TrackMCH.h"
@@ -1298,8 +1300,10 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
     mMcParticleW = 0xFFFFFFFF;
     mMcParticlePos = 0xFFFFFFFF;
     mMcParticleMom = 0xFFFFFFFF;
-    mCaloAmp = 0xFFFFFFFF;  // todo check which truncation should actually be used
-    mCaloTime = 0xFFFFFFFF; // todo check which truncation should actually be used
+    mCaloAmp = 0xFFFFFFFF;
+    mCaloTime = 0xFFFFFFFF;
+    mCPVPos = 0xFFFFFFFF;
+    mCPVAmpl = 0xFFFFFFFF;
     mMuonTr1P = 0xFFFFFFFF;
     mMuonTrThetaX = 0xFFFFFFFF;
     mMuonTrThetaY = 0xFFFFFFFF;
@@ -1357,6 +1361,10 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
 
   auto caloPHOSCells = recoData.getPHOSCells();
   auto caloPHOSCellsTRGR = recoData.getPHOSTriggers();
+
+  auto cpvClusters = recoData.getCPVClusters();
+  auto cpvTrigRecs = recoData.getCPVTriggers();
+
   auto ctpDigits = recoData.getCTPDigits();
   const auto& tinfo = pc.services().get<o2::framework::TimingInfo>();
   std::vector<o2::ctp::CTPDigit> ctpDigitsCreated;
@@ -1372,6 +1380,8 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   LOG(debug) << "FOUND " << fddRecPoints.size() << " FDD rec. points";
   LOG(debug) << "FOUND " << caloEMCCells.size() << " EMC cells";
   LOG(debug) << "FOUND " << caloEMCCellsTRGR.size() << " EMC Trigger Records";
+  LOG(debug) << "FOUND " << cpvClusters.size() << " CPV clusters";
+  LOG(debug) << "FOUND " << cpvTrigRecs.size() << " CPV trigger records";
 
   LOG(info) << "FOUND " << primVertices.size() << " primary vertices";
   auto& bcBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "BC"});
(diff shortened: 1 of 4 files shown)


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1803,6 +1813,33 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
...
+        LOG(fatal) << "Error: could not find a corresponding BC ID for a CPV Trigger Record; BC = " << bc;
+      }
+      for (int iClu = cpvEvent.getFirstEntry(); iClu < cpvEvent.getFirstEntry() + cpvEvent.getNumberOfObjects(); iClu++) {
+        auto& clu = cpvClusters[iClu];
+        clu.getLocalPosition(posX, posZ);
+        cpvClustersCursor(0,
+                          bcID,
+                          posX,
+                          posZ,
+                          clu.getEnergy(),
```
jgrosseo: Please use the truncate float fraction method to reduce to the expected resolution

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -665,6 +665,22 @@ DECLARE_SOA_TABLE(CaloTriggers, "AOD", "CALOTRIGGER", //! Trigger information fr
...
 
+namespace cpvcluster
+{
+DECLARE_SOA_INDEX_COLUMN(BC, bc);                                  //! BC index
+DECLARE_SOA_COLUMN(PosX, posX, float);                             //! X position in cm
+DECLARE_SOA_COLUMN(PosZ, posZ, float);                             //! Z position in cm
+DECLARE_SOA_COLUMN(Amplitude, amplitude, float);                   //! Signal amplitude
+DECLARE_SOA_COLUMN(PadMult, padMult, uint8_t);                     //! Multiplicity of pads in cluster
+DECLARE_SOA_COLUMN(ModuleNumber, moduleNumber, uint8_t);           //! CPV module number (2, 3 or 4)
+DECLARE_SOA_COLUMN(LocalMaximaNumber, localMaximaNumber, uint8_t); //! Number of local maxima in cluster
```
jgrosseo: What is the range of this value? Can we combine the previous two into one value and then use dynamical columns to expand them?

jgrosseo (commented):
Thanks for this work. Apart from my two comments directly in the code, 

- have you executed `sim_challenge.sh`? 
- Is the AOD output as you expect it? 
- Have you tried to analyze it with a task locally?

Many thanks!

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
sevdokim: All right! Is there any description of truncateFloatFraction() function?

Framework/Core/include/Framework/AnalysisDataModel.h
sevdokim: Sorry, I didn't understand the question? Just to clarify: Amplitude is total cluster amplitude, PadMult is number of pads in cluster, ModuleNumber is CPV module number in which the cluster is found, LocalMaximaNumber is number of local maxima in the cluster. I don't think you can combine any of this value out of the rest. Let's say the variables construct orthogonal basis in space of cpv clusters.

jgrosseo: The comment was on LocalMaximaNumber and ModuleNumber
ModuleNumber needs 2 bits
How many bits are needed by LocalMaximaNumber?

jgrosseo: I.e. what is the highest expected number of local maxima?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
jgrosseo: I don't think so, but you can look at the usage in the same file and the function itself is also very short

sevdokim:
@jgrosseo I ran full_sustem_test.sh to produce AO2D file, also ran clustering manually and checked that the number of clusters, coordinate and amplitude distribution are same in AO2D file and in clusters written to file by clusterizer. 
I didn't try to analyze AO2D file with O2Physics.

Framework/Core/include/Framework/AnalysisDataModel.h
sevdokim: Ah, I see what you mean. I'll check the highest expected number of local maxima with @peressounko 
Don't think it is too high.

sevdokim: So I checked and found that number of local maxima is fictional name. It is 1 if cluster is unfolded and is 2 if cluster is 
not unfolded. So actually it is only one bit. And PadMult, ModuleNumber and LocalMaximaNumber can be actually packed into 8 bits. Thanks for digging in this direction. I'll modify the code accordingly.

jgrosseo: Great! You can then add dynamic columns (look for examples in the same file) to make related analysis code more readable.

jgrosseo:
> @jgrosseo I ran full_sustem_test.sh to produce AO2D file, also ran clustering manually and checked that the number of clusters, coordinate and amplitude distribution are same in AO2D file and in clusters written to file by clusterizer. I didn't try to analyze AO2D file with O2Physics.

Thanks! Then it makes sense to also write a simple task which subscribes to the new table and prints a few values, just to make sure that reading the data works fine as well.

sevdokim:
@jgrosseo 
I have a question regarding MC labels. I see that MC true info is connected with MCParticle by introducing additional table e.g. mcfwdtracklabel. It connects MCParticle with particular labels. However looking at MCParticle, it is constructed from tracks and some labels from cpv can be missing in it. Do you have any suggestions how to organize mc information from cpv in aod?

sevdokim:
hi @jgrosseo 
I checked generated AO2D file with o2-analysis-qa-cpv workflow (which will be commited to O2Physics) and the AOD info is read properly and is in agreement with clusterizer output. So I think we can proceed with this PR.

jgrosseo (approved):

jgrosseo:
> @jgrosseo I have a question regarding MC labels. I see that MC true info is connected with MCParticle by introducing additional table e.g. mcfwdtracklabel. It connects MCParticle with particular labels. However looking at MCParticle, it is constructed from tracks and some labels from cpv can be missing in it. Do you have any suggestions how to organize mc information from cpv in aod?

I think it is better to follow the calolabel example. 
https://github.com/AliceO2Group/AliceO2/blob/dev/Framework/Core/include/Framework/AnalysisDataModel.h#L1018

jgrosseo:
(This needs additional approvals from Detector/AOD it seems...)

davidrohr (approved):

PR #10127, opened 2022-10-20
DPL: fix proxy for QC

ktf:


ktf:
Not clear to me why, but the previous change breaks QC (but not the proxy in general, apparently).

ktf:
Tested on flptest1 and currently broken for QC, so I merge.

ktf:
@vascobarroso @Barthelemy

PR #10129, opened 2022-10-20
TRD add --disable-root-output to data reader

bazinski:
TRD had --enable-root-output, it was removed for reasons unknown.
The logic is now reversed and re put in at @shahor02 request.


shahor02 (approved):
thanks!

martenole:
Hi @bazinski , you had removed it in https://github.com/AliceO2Group/AliceO2/pull/9675
Not sure we need that output flag in the `OptionBits`, or is that used somewhere?

PR #10132, opened 2022-10-20
DPL: fix input proxy for QC

ktf:
channels is in general different from fChannels.


davidrohr (approved):

ktf:
I tested this with the flptest1 setup and it fixes the issue. We did not spot it when running the IDC tests because of course there channels != 1.

ktf:
Merging to get it in the build.

PR #10139, opened 2022-10-21
Replace static array by dynamic vector on MFT CA track finder

rpezzi:
Also adding debug messages to MFT tracker


rpezzi:
Proper solution for https://github.com/AliceO2Group/AliceO2/pull/10064

shahor02 (approved):

PR #10142, opened 2022-10-22
Fix it TRD global tracking

shahor02:
Reset N of refitted ITS clusters after outward ITSAB refit, otherwhise all AB tracks are rejected due to the wrong nClRefit count


f3sch (approved):
LGTM! :thumbsup:

martenole:
Thanks a lot for the fix! Not sure why I did not notice that before. Probably I never tested with ITS AB tracks :/

PR #10148, opened 2022-10-23
I/O for CTP reconstruction results

shahor02:
* add LumiInfo to CTP digits writer and reader
* add LumiInfo to InputHelper and RecoContainer
* add digit writer to `ctp-reco-workflow` with usual option `--disable-root-output` and account for it in the dpl-workflow.sh

### DIFF
--- a/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
+++ b/DataFormats/Detectors/GlobalTracking/include/DataFormatsGlobalTracking/RecoContainer.h
@@ -26,6 +26,7 @@
 #include "SimulationDataFormat/MCCompLabel.h"
 #include "SimulationDataFormat/MCTruthContainer.h"
 #include "SimulationDataFormat/ConstMCTruthContainer.h"
+#include "DataFormatsCTP/LumiInfo.h"
 #include <gsl/span>
 #include <memory>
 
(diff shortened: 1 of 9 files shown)


lietava:
HI @shahor02 , I am looking on 526638, checking (in)consistencies.
I found missing input mask for trigger mask (= missing expected BC in Inputs record, 8 missing out of 900 trigger masks). Either I have bug in analysis or there is bug in reconstruction or there is bug in fw.
Is there some tool which can dump tf file ? Is tf binary if I would need to write something myself ?
Cheers, Roman.

shahor02:
Hi @lietava 

Yes, the TF is binary, it is what DataDistribution TfBuilder sees (in absence of the processing on the FLP: just the RDH blocks from the readout.exe) wrapped with some headers. 
I don't know of any standalone dumping tool except hexdump, but why don't you simply dump the payload here: https://github.com/AliceO2Group/AliceO2/blob/4b30ec6930a6039e7718e95dd6808722af19bf1f/Detectors/CTP/workflow/src/RawDecoderSpec.cxx#L107

Cheers,
 Ruben

lietava:
@shahor02 : I found two bugs. I have fixes but cannot push to your repository.

lietava:
@shahor02 : I pushed to mine: https://github.com/lietava/AliceO2/tree/pr_ctpIO

shahor02:
Hi @lietava 

Could you simply open a PR? What was the effect of the bugs? Comparing results of original separate workflows with that of unified one I did not notice any difference.

lietava:
The first bug was that last interaction in orbit was missing so you can not see it easily.
It was fixed:
https://github.com/lietava/AliceO2/blob/24a93edb6b9d6d779884a26c8ee8e8ce1dbb60cd/Detectors/CTP/workflow/src/RawDecoderSpec.cxx#L158
Second was that when in counting lumi wrong bits were used:
https://github.com/lietava/AliceO2/blob/24a93edb6b9d6d779884a26c8ee8e8ce1dbb60cd/Detectors/CTP/workflow/src/RawDecoderSpec.cxx#L139
I am not sure why it looked ok before - I can have a look.
Now I am using for test run 526638 and I get ~500 kHz.
I will create PR.

PR #10150, opened 2022-10-24
ZDC Preparing output files for EOS to enable tuning of calibration

cortesep:


Detectors/ZDC/calib/src/BaselineCalibSpec.cxx
```diff
@@ -115,6 +128,40 @@ void BaselineCalibSpec::sendOutput(o2::framework::DataAllocator& output)
   output.snapshot(Output{o2::calibration::Utils::gDataOriginCDBWrapper, "ZDCBaselinecalib", 0}, info);         // root-serialized
   // TODO: reset the outputs once they are already sent (is it necessary?)
   // mWorker.init();
+
+  o2::zdc::CalibParamZDC& opt = const_cast<o2::zdc::CalibParamZDC&>(CalibParamZDC::Instance());
```
shahor02: The ConfigKeyValue object must be const ref, why do you constast it?
Please use  `const auto&  opt = CalibParamZDC::Instance();`, also in other places.

Detectors/ZDC/calib/src/BaselineCalibSpec.cxx
```diff
@@ -115,6 +128,40 @@ void BaselineCalibSpec::sendOutput(o2::framework::DataAllocator& output)
...
+  if (opt.rootOutput == true) {
+    mOutputDir = opt.outputDir;
+    if (mOutputDir.compare("/dev/null")) {
+      mHistoFileName = mOutputDir + fmt::format("{}_{}.root", fn, mRunNumber);
```
shahor02: this is error prone since mOutputDir is not guaranteed to have '/' in the end. Usually I do like
```
mHistoFileName = fmt::format("{}{}{}_{}.root", mOutputDir,  mOutputDir.back()=='/' ? "":"/" fn, mRunNumber);
```

Detectors/ZDC/calib/src/BaselineCalibSpec.cxx
```diff
@@ -115,6 +128,40 @@ void BaselineCalibSpec::sendOutput(o2::framework::DataAllocator& output)
...
+        LOG(error) << "Cannot create output file " << mHistoFileName;
+        return;
+      }
+      std::string metaFileDir = opt.metaFileDir;
+      if (metaFileDir.compare("/dev/null")) {
+        mHistoFileMetaData->fillFileData(mHistoFileName);
+        mHistoFileMetaData->type = "calib";
+        mHistoFileMetaData->priority = "high";
+        std::string metaFileNameTmp = metaFileDir + fmt::format("{}_{}.tmp", fn, mRunNumber);
+        std::string metaFileName = metaFileDir + fmt::format("{}_{}.done", fn, mRunNumber);
```
shahor02: same comment here

Detectors/ZDC/calib/src/BaselineCalibSpec.cxx
```diff
@@ -85,6 +89,13 @@ void BaselineCalibSpec::run(ProcessingContext& pc)
     mTimer.Reset();
     mTimer.Start(false);
   }
+  if (mRunStartTime == 0) {
+    mHistoFileMetaData = std::make_unique<o2::dataformats::FileMetaData>();
```
shahor02: Why do you create a metadata as a data member, moreover w/o even checking if it is needed at all? Instead you can create a local object instance in the sendOutput if it is needed

shahor02 (commented):
Hi @cortesep 

See comments below. General question: the files for the epn2eos are created on the aggregator node only, right?

cortesep:
Review comments taken into account. Thanks.

shahor02 (approved):

PR #10151, opened 2022-10-24
Add filtering options to MFT tracker

rpezzi:
This PR add two filtering options to reduce EPN load during sync reconstruction: 

- irFramesOnly: Reconstruct only IRFrames selected by ITS
- Multiplicity filter: selects ROFs matching a range of cluster multiplicity

### DIFF
--- a/Detectors/ITSMFT/MFT/workflow/src/RecoWorkflow.cxx
+++ b/Detectors/ITSMFT/MFT/workflow/src/RecoWorkflow.cxx
@@ -20,6 +20,7 @@
 #include "ITSMFTWorkflow/DigitReaderSpec.h"
 #include "MFTWorkflow/MFTAssessmentSpec.h"
 #include "MFTWorkflow/TracksToRecordsSpec.h"
+#include "GlobalTrackingWorkflowReaders/IRFrameReaderSpec.h"
 
 namespace o2
 {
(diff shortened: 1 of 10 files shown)


rpezzi:
Hi, all tests passed. Looks good to be merged. Thanks.

Detectors/ITSMFT/MFT/workflow/src/RecoWorkflow.cxx
```diff
@@ -51,6 +52,11 @@ framework::WorkflowSpec getWorkflow(
   if (!disableRootOutput) {
     specs.emplace_back(o2::mft::getClusterWriterSpec(useMC));
   }
+  auto& trackingParam = MFTTrackingParam::Instance();
+  if (trackingParam.irFramesOnly) {
+    specs.emplace_back(o2::globaltracking::getIRFrameReaderSpec("ITS", 0, "its-irframe-reader", "o2_its_irframe.root"));
+  }
+
```
shahor02, line 59: Hi @rpezzi 
Not sure how is this supposed to work: with `trackingParam.irFramesOnly` you are unconditionally starting IRFrameReader and request the ITS/IRFrames in the tracker. But in the global (sync or async) workflow the IRFrames are created on the fly by the ITS tracker, and there is no o2_its_irframe.root written.

rpezzi: Hi Ruben, thanks for spotting this. So, there is no need of these request for sync and async reco? Then I can replace the the test at line 56 by an command line option `--readIRFrames`. Would this be ok?

shahor02: Hi Rafael,

Since your workflow reads from the files only in case of `!(upstreamDigits || upstreamClusters)`, then loading of the IR-frames from the file should be allowed only in that case, otherwise it should get it from the upstream.
Reconstruction from the mftdigits.root is normally used for the MC reco, I assume in this case we don't need prescaling.
If you still want to used e.g. to reconstruct from digits obtained in the raw data, you can probably add an option to enforce loading o2_its_irframe.root on top of the trackingParam.irFramesOnly condition.

Cheers,
 Ruben

rpezzi: Perfect. Thanks. Here it is: https://github.com/AliceO2Group/AliceO2/pull/10172

PR #10154, opened 2022-10-25
DPL: fix signature for callback

ktf:
CallbackRegistry silently ignores invokations which do not match the actual callback signature...


ktf:
Currently broken. Merging.

PR #10181, opened 2022-10-29
Pr ctp io

lietava:
The first bug was that last interaction in orbit was missing so you can not see it easily.
It was fixed:
https://github.com/lietava/AliceO2/blob/24a93edb6b9d6d779884a26c8ee8e8ce1dbb60cd/Detectors/CTP/workflow/src/RawDecoderSpec.cxx#L158
Second was that when in counting lumi wrong bits were used:
https://github.com/lietava/AliceO2/blob/24a93edb6b9d6d779884a26c8ee8e8ce1dbb60cd/Detectors/CTP/workflow/src/RawDecoderSpec.cxx#L139
In the bug the first 12 bits are BC and I was taking bit 0x4 from there. So if the BC are randomly distributed in Orbit this rate is would be factor 2 smaller while I see only 0.87 reduction - maybe to the filling scheme.
Now I am using for test run 526638 and I get ~500 kHz.
=> 0.87 reduction  is because IR is OR of all inputs and other inputs fire more often than TVX.


shahor02:
@lietava thanks but your PR contains my and others commits, could you rebase to dev and force-push?

lietava:
@shahor02 : I rebased. It seems error in checks is not on my side ?

shahor02 (approved):
failures are unrelated

PR #10182, opened 2022-10-30
More multithreading changes

ktf:


ktf:
Tested on hyperloop as well.

PR #10184, opened 2022-10-30
Some more multithreading developments

ktf:


ktf:
Trivial, merging.

@TimoWilken it looks like something is wrong with the build machine for o2-dataflow-cs8

image

Could you have a look tomorrow?

TimoWilken:
@ktf looks like the problem was that aliBuild was hanging forever when trying to update the mesos git repo. Unfortunately this was part of the build, so it had a 10h timeout. We're apparently already tracking this as a bug, though it hadn't been observed in a while, so I've bumped the importance.

PR #10186, opened 2022-10-31
ITS: Add per-ROF OMP multithreading to seeding tracklet finding and selection

mconcas:
Hi @shahor02: this PR add the parallelisation of the "tracklet finding" part of the seeding vertexer.
Next will be the tracklet validation.

Hereafter the elapsed time for the `computeTracklets()` call as a function of #threads. This i got from a quick test on my M2 Mac (4 physical cores).

To test it: `--configKeyValues='ITSVertexerParam.nThreads=<n>'`

| # threads | trackleting elapsed time (ms) |
|-----------|-------------------------------|
| 1         | 222.4                         |
| 2         | 137.3                         |
| 3         | 91.6                          |
| 4         | 76.0                          |

As embarrassing parallelism is applied at ROF level, I expect the single TF processing time should theoretically scale better over more threads, on a more capable machine. I am also aware we will not allocate 32 threads for this.

I'll do a more extended test on the other workstation once I'll add the tracklet selection parallelisation.

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/include/ITStracking/Configuration.h
+++ b/Detectors/ITSMFT/ITS/tracking/include/ITStracking/Configuration.h
@@ -106,6 +106,8 @@ struct VertexingParameters {
   int maxTrackletsPerCluster = 2e3;
   int phiSpan = -1;
   int zSpan = -1;
+
+  int nThreads = 1;
 };
 
 struct VertexerHistogramsConfiguration {
--- a/Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
+++ b/Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
@@ -161,62 +161,68 @@ void VertexerTraits::updateVertexingParameters(const VertexingParameters& vrtPar
 
 void VertexerTraits::computeTracklets()
 {
-#pragma omp parallel for num_threads(mNThreads)
-  for (int rofId = 0; rofId < mTimeFrame->getNrof(); ++rofId) {
-    trackleterKernelHost<TrackletMode::Layer0Layer1, true>(
-      mTimeFrame->getClustersOnLayer(rofId, 0),
-      mTimeFrame->getClustersOnLayer(rofId, 1),
-      mTimeFrame->getIndexTable(rofId, 0).data(),
-      mVrtParams.phiCut,
-      mTimeFrame->getTracklets()[0],
-      mTimeFrame->getNTrackletsCluster(rofId, 0),
-      mIndexTableUtils,
-      rofId,
-      0,
-      mVrtParams.maxTrackletsPerCluster);
-    trackleterKernelHost<TrackletMode::Layer1Layer2, true>(
-      mTimeFrame->getClustersOnLayer(rofId, 2),
-      mTimeFrame->getClustersOnLayer(rofId, 1),
-      mTimeFrame->getIndexTable(rofId, 2).data(),
-      mVrtParams.phiCut,
-      mTimeFrame->getTracklets()[1],
-      mTimeFrame->getNTrackletsCluster(rofId, 1),
-      mIndexTableUtils,
-      rofId,
-      0,
-      mVrtParams.maxTrackletsPerCluster);
-    mTimeFrame->getNTrackletsROf(rofId, 0) = std::accumulate(mTimeFrame->getNTrackletsCluster(rofId, 0).begin(), mTimeFrame->getNTrackletsCluster(rofId, 0).end(), 0);
-    mTimeFrame->getNTrackletsROf(rofId, 1) = std::accumulate(mTimeFrame->getNTrackletsCluster(rofId, 1).begin(), mTimeFrame->getNTrackletsCluster(rofId, 1).end(), 0);
-  }
 
-  mTimeFrame->computeTrackletsScans(mNThreads);
-  mTimeFrame->getTracklets()[0].resize(mTimeFrame->getTotalTrackletsTF(0));
-  mTimeFrame->getTracklets()[1].resize(mTimeFrame->getTotalTrackletsTF(1));
+#pragma omp parallel num_threads(mNThreads)
+  {
+#pragma omp for schedule(dynamic)
+    for (int rofId = 0; rofId < mTimeFrame->getNrof(); ++rofId) {
+      trackleterKernelHost<TrackletMode::Layer0Layer1, true>(
+        mTimeFrame->getClustersOnLayer(rofId, 0),
+        mTimeFrame->getClustersOnLayer(rofId, 1),
+        mTimeFrame->getIndexTable(rofId, 0).data(),
+        mVrtParams.phiCut,
+        mTimeFrame->getTracklets()[0],
+        mTimeFrame->getNTrackletsCluster(rofId, 0),
+        mIndexTableUtils,
+        rofId,
+        0,
+        mVrtParams.maxTrackletsPerCluster);
+      trackleterKernelHost<TrackletMode::Layer1Layer2, true>(
+        mTimeFrame->getClustersOnLayer(rofId, 2),
+        mTimeFrame->getClustersOnLayer(rofId, 1),
+        mTimeFrame->getIndexTable(rofId, 2).data(),
+        mVrtParams.phiCut,
+        mTimeFrame->getTracklets()[1],
+        mTimeFrame->getNTrackletsCluster(rofId, 1),
+        mIndexTableUtils,
+        rofId,
+        0,
+        mVrtParams.maxTrackletsPerCluster);
+      mTimeFrame->getNTrackletsROf(rofId, 0) = std::accumulate(mTimeFrame->getNTrackletsCluster(rofId, 0).begin(), mTimeFrame->getNTrackletsCluster(rofId, 0).end(), 0);
+      mTimeFrame->getNTrackletsROf(rofId, 1) = std::accumulate(mTimeFrame->getNTrackletsCluster(rofId, 1).begin(), mTimeFrame->getNTrackletsCluster(rofId, 1).end(), 0);
+    }
+#pragma omp single
+    mTimeFrame->computeTrackletsScans(mNThreads);
+#pragma omp single
+    mTimeFrame->getTracklets()[0].resize(mTimeFrame->getTotalTrackletsTF(0));
+#pragma omp single
+    mTimeFrame->getTracklets()[1].resize(mTimeFrame->getTotalTrackletsTF(1));
 
-#pragma omp parallel for num_threads(mNThreads)
-  for (int rofId = 0; rofId < mTimeFrame->getNrof(); ++rofId) {
-    trackleterKernelHost<TrackletMode::Layer0Layer1, false>(
-      mTimeFrame->getClustersOnLayer(rofId, 0),
-      mTimeFrame->getClustersOnLayer(rofId, 1),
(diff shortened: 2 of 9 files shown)


mconcas:
~Please do not merge yet, it looks like I have a genuine bug in the CI~

mconcas:
> Please do not merge yet, it looks like I have a genuine bug in the CI

This should be fixed. IF fullCI is green we can merge

Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
```diff
@@ -144,35 +156,74 @@ void VertexerTraits::updateVertexingParameters(const VertexingParameters& vrtPar
...
+        mVrtParams.maxTrackletsPerCluster);
+      mTimeFrame->getNTrackletsROf(rofId, 0) = std::accumulate(mTimeFrame->getNTrackletsCluster(rofId, 0).begin(), mTimeFrame->getNTrackletsCluster(rofId, 0).end(), 0);
+      mTimeFrame->getNTrackletsROf(rofId, 1) = std::accumulate(mTimeFrame->getNTrackletsCluster(rofId, 1).begin(), mTimeFrame->getNTrackletsCluster(rofId, 1).end(), 0);
+    }
+#pragma omp single
+    mTimeFrame->computeTrackletsScans(mNThreads);
+#pragma omp single
+    mTimeFrame->getTracklets()[0].resize(mTimeFrame->getTotalTrackletsTF(0));
+#pragma omp single
+    mTimeFrame->getTracklets()[1].resize(mTimeFrame->getTotalTrackletsTF(1));
```
shahor02, line 199: These lines are outside of the MT-ed block, why do you need `omp single`?

Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
```diff
@@ -144,35 +156,74 @@ void VertexerTraits::updateVertexingParameters(const VertexingParameters& vrtPar
...
-      mTimeFrame->getNTrackletsCluster(rofId, 1),
-      mIndexTableUtils,
-      rofId,
-      mVrtParams.maxTrackletsPerCluster);
-    mTimeFrame->getNTrackletsROf(rofId, 0) = std::accumulate(mTimeFrame->getNTrackletsCluster(rofId, 0).begin(), mTimeFrame->getNTrackletsCluster(rofId, 0).end(), 0);
-    mTimeFrame->getNTrackletsROf(rofId, 1) = std::accumulate(mTimeFrame->getNTrackletsCluster(rofId, 1).begin(), mTimeFrame->getNTrackletsCluster(rofId, 1).end(), 0);
+
+#pragma omp parallel num_threads(mNThreads)
+  {
+#pragma omp for schedule(dynamic)
```
shahor02, line 167: please prefer direct action `#pragma omp parallel for num_threads(mNThreads)`

Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
```diff
@@ -144,35 +156,74 @@ void VertexerTraits::updateVertexingParameters(const VertexingParameters& vrtPar
...
+      mTimeFrame->getNTrackletsROf(rofId, 1) = std::accumulate(mTimeFrame->getNTrackletsCluster(rofId, 1).begin(), mTimeFrame->getNTrackletsCluster(rofId, 1).end(), 0);
+    }
+#pragma omp single
+    mTimeFrame->computeTrackletsScans(mNThreads);
+#pragma omp single
+    mTimeFrame->getTracklets()[0].resize(mTimeFrame->getTotalTrackletsTF(0));
+#pragma omp single
+    mTimeFrame->getTracklets()[1].resize(mTimeFrame->getTotalTrackletsTF(1));
+
+#pragma omp for schedule(dynamic)
```
shahor02, line 201: also here: `#pragma omp parallel for num_threads(mNThreads)`

Detectors/ITSMFT/ITS/tracking/include/ITStracking/Configuration.h
```diff
@@ -106,6 +106,8 @@ struct VertexingParameters {
   int maxTrackletsPerCluster = 2e3;
   int phiSpan = -1;
   int zSpan = -1;
+
+  int nThreads = 1;
```
shahor02, line 110: could you point me on where this template is instantiated?

shahor02 (changes requested):

Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
mconcas: My understanding was that when you enter an omp parallel section it spanws the threads and keep them alive until it ends then tries to assign the instruction included in the block to available threads according to scheduling.

shahor02: But your parallel block is the `for` loop only, afterwards OPM does nothing

mconcas: but I have two of them

Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
mconcas: I actually thought that we might keep the threads alive between the two for loops so that we save spawning time (see above comment). I saw a slight perfromance improvement in doing so.
On the other hand, to spawn few threads does no really imply lot of overhead.
If we want to keep it simple it's fine.

Detectors/ITSMFT/ITS/tracking/include/ITStracking/Configuration.h
mconcas: Currently we have in the `*TrackerSpec.cxx` a call to `getGlobalConfiguration()` defined here: https://github.com/AliceO2Group/AliceO2/blob/dfae99cd3b675aa6abd45a6383c776862b8e0bb5/Detectors/ITSMFT/ITS/tracking/src/Vertexer.cxx#L56
That resets config values to `VertexerParamConfig` ones (this are the one that can be overridden by `--configKeyValues`).
Then during vertexerTraits initialisation the vertexer is configured again on the same values, this should be probably removed.
I'll have a look in next iteration.

Vertexer/Tracker configuration is a bit cumbersome, probably because of the GPU compatibility (ConfigParams depends on ROOT things).
I will improve that.

Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
shahor02: Not sure what you mean. https://github.com/AliceO2Group/AliceO2/blob/ed25a1e47c5fd105013061949b6fc2582396ab1c/Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx#L165 just reserves N threads which are applied only to https://github.com/AliceO2Group/AliceO2/blob/ed25a1e47c5fd105013061949b6fc2582396ab1c/Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx#L167-L193 block then, independently, to https://github.com/AliceO2Group/AliceO2/blob/ed25a1e47c5fd105013061949b6fc2582396ab1c/Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx#L201-L226
The linear code outside of the `for` loops is not affected by the OMP, so no `#pragma omp single` is needed

Detectors/ITSMFT/ITS/tracking/include/ITStracking/Configuration.h
shahor02: sorry, I put the link on the wrong line, I meant this template: https://github.com/AliceO2Group/AliceO2/blob/ed25a1e47c5fd105013061949b6fc2582396ab1c/Detectors/ITSMFT/ITS/tracking/include/ITStracking/Configuration.h#L34-L48

mconcas: Ah, that is not something I am aware of, according to git blame it's 4 years old. Probably dead code (?).
Will check and dispose of that, in case.

Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
mconcas: > The linear code outside of the `for` loops is not affected by the OMP, so no `#pragma omp single` is needed

Ok, I might be wrong, my understanding was that when I am inside a `#pragma omp parallel { .... }` block (quoting [this](https://www.ibm.com/docs/en/zos/2.4.0?topic=processing-pragma-omp-parallel) reference):

```
When a parallel region is encountered, a logical team of threads is formed. 
Each thread in the team executes all statements within a parallel region except for work-sharing constructs. 
Work within work-sharing constructs is distributed among the threads in a team.
```
Since we are still in the `omp parallel` block, I'd expect to have the threads to be alive and doing the same things.
This is why I explicitly constrained a single thread to do that (I had crashes due to racing conditions).

shahor02: I thought the `omp parallel` just allocates the threads, w/o assigning them any work, but reading https://stackoverflow.com/questions/1448318/omp-parallel-vs-omp-parallel-for (last comment particularly) it seems that it does. Still, not sure having a large parallel block covering  2 loops is useful, perhaps @davidrohr can suggest.

davidrohr: @shahor02 @mconcas 
I think Matteo understood correctly, since you have 2 parallel blocks:
```
#pragma omp parallel // from here on we run with multiple threads
{
#pragma omp for
  for (int i = 0; i < 10; i++) {
    [...] // This for loop runs with the threads declared in the outer block
  }
#pragma omp single
  {
    foo(); // These functions are only called by one of the threads in the outer block
    bar();
  }
#pragma omp for
  for (int i = 0; i < 12; i++) {
    [...] // Again running multi-threaded, using the same threads as above.
  }
}
```

The diference to:
```
#pragma omp parallel for
for (int i = 0; i < 10; i++) {
  [...]
}
foo();
bar();
#pragma omp parallel for
for (int i = 0; i < 12; i++) {
  [...]
}
```
is that in the first case, the threads are not synchronized up until the end of the parallel block.
In the second case, threads are synchronized when exiting the first parallel block, and before the main thread calls `foo()`

Note the difference between `#pragma omp for` and `#pragma omp parallel for`. The first uses threads of an existing parallel block, while the later would spawn new threads (or take existing threads from a thread pool).
E.g.
```
#pragma omp parallel num_threads(4)
{
#pragma omp parallel for num_threads(4)
  for (int i = 0; i < 10; i++) {
    [...]
  }
}
```
Would use 16 threads to run the full loop 4 times, each time 4-fold multi-threaded.

shahor02: OK, then I withdraw my comment. Actually, from https://stackoverflow.com/questions/1448318/omp-parallel-vs-omp-parallel-for I understand that w/o `#pragma omp single` all threads (not only the main one) will execute the the code outside of the `for` loops but inside the `omp parallel` block

davidrohr: yes, exactly

shahor02 (approved):

PR #10188, opened 2022-10-31
Update codeowners

benedikt-voelkel:


benedikt-voelkel:
@ktf This is what I did and as I was saying, it doesn't know me.

@sawenzel could you please also x-check if the places I put my name make sense?

sawenzel (approved):

PR #10190, opened 2022-11-01
TPC: N2 density update for STP, clang-format

wiechula:
Update N2 density for STP as suggested by Marek Kowalski


wiechula:
can this be merged?

PR #10192, opened 2022-11-01
Added possibility to set maximum size of output files

pbuehler:
Two items have been added to the aod-writer json:
maxfilesize (default value: -1): the approximate maximum size an output file shall reach in MBytes.
resdir (default value: "."): is the directory into which the output files are saved.

Before a DF is saved to disk, the size of all output files is checked to be below _maxfilesize_. If one of the files is larger than _maxfilesize_ MBytes, then all output files are closed and new ones are created. The files are saved into directories _resdir_/xxx, where xxx is an increasing counter starting at 1. The test of the file size is disabled if _maxfilesize_ is <= 0. In that case all DFs are saved into one set of files in directory _resdir_. Directory _resdir_ as well as its subdirectories _resdir_/xxx  are created if they do not exist already.

### DIFF
--- a/Framework/Core/include/Framework/DataOutputDirector.h
+++ b/Framework/Core/include/Framework/DataOutputDirector.h
@@ -90,13 +90,20 @@ struct DataOutputDirector {
   // get the matching TFile
   FileAndFolder getFileFolder(DataOutputDescriptor* dodesc, uint64_t folderNumber, std::string parentFileName);
 
+  // check file sizes
+  bool checkFileSizes();
+  // close all files
   void closeDataFiles();
 
+  // setters
+  void setResultDir(std::string resDir);
   void setFilenameBase(std::string dfn);
+  void setMaximumFileSize(float maxfs);
 
   void printOut();
 
  private:
+  std::string mresultDirectory;
   std::string mfilenameBase;
   std::string* const mfilenameBasePtr = &mfilenameBase;
   std::vector<DataOutputDescriptor*> mDataOutputDescriptors;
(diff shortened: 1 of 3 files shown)


pbuehler:
Hi @ddobrigk, @jgrosseo
This is the PR which allows to set a maximum size of the output AODs. Please have a look and if possible test it in a realistic setting. I only had the chance to test it locally on a small scale.

jgrosseo (commented):
Hi Paul,
Thanks for this PR. I wonder if it is best to check the file size directly on the file system or if we could also keep a count of the size which is incremented on each tree write. I do it this way in the o2-aod-merger in O2Physics. What do you think?

pbuehler:
Dear Jan,
To manage the output directories I need to use something like _filesystem_. So it seemed natural to me to use the _filesystem_ methods to check the file size. But indeed, keeping track of the size could also be done with counters. Are you worried about the performance of file_size method?

jgrosseo:
It was about having less file system calls but I think you are right that it is not a big overhead. So I am fine with this. Waiting also for @ktf to comment.

ktf:
As discussed privately, the only problem I see with this approach is if we do not write to disk, but to e.g. some XRootd endpoint directly. That said, I would merge this and then we improve, if needed.

jgrosseo:
@pbuehler @ktf Did we forget to merge this?

ktf:
Apparently so. I assume @pbuehler rebased dev for some other development. Anyways, recovered in:

https://github.com/AliceO2Group/AliceO2/pull/10232

PR #10202, opened 2022-11-02
[ITS] Fix chip thickness type

fgrosa:


iouribelikov (changes requested):
@fgrosa Thank you very much, Fabrizio.  Let's make the thickness "double".

But this thickness must be just one, with the value of 50 micron !   
(The chip is just the same for both IB and OB)
I would try to add it to this PR.
@mario6829  Changing the OB chip thickness from 100 to 50 micron will not break anything else in the geometry ?  Right ?

mario6829:
Hi @iouribelikov @fgrosa if I remember correctly the chip whole thickness is different on IB and OB (50 and 100 microns respectively). Then from the abstract point of view of the code changing the chip thickness should not produce any error since everything is parametrized, and volume dimensions are compared to avoid extrusions.

iouribelikov (approved):
@fgrosa @mario6829  OK then. Sorry for my "noise".  Let's merge it.

mconcas (approved):
@shahor02 anytime

shahor02:
trivial, merging

PR #10203, opened 2022-11-02
Disabling delta IDCs in pp by default

chiarazampolli:
@matthias-kleiner , @davidrohr , @ehellbar , @wiechula , @shahor02

### DIFF
--- a/prodtests/full-system-test/aggregator-workflow.sh
+++ b/prodtests/full-system-test/aggregator-workflow.sh
@@ -71,7 +71,7 @@ fi
 
 # special settings for aggregator workflows
 if [[ "0$CALIB_TPC_SCDCALIB_SENDTRKDATA" == "01" ]]; then ENABLE_TRACK_INPUT="--enable-track-input"; fi
-if [[ -z "$RESIDUAL_AGGREGATOR_AUTOSAVE" ]];   then RESIDUAL_AGGREGATOR_AUTOSAVE=0; fi
+if [[ -z "$RESIDUAL_AGGREGATOR_AUTOSAVE" ]]; then RESIDUAL_AGGREGATOR_AUTOSAVE=0; fi
 
 # Calibration workflows
 if ! workflow_has_parameter CALIB_LOCAL_INTEGRATED_AGGREGATOR; then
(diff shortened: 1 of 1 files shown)


prodtests/full-system-test/aggregator-workflow.sh
```diff
@@ -221,11 +221,14 @@ fi
...
 
 if ! workflow_has_parameter CALIB_LOCAL_INTEGRATED_AGGREGATOR; then
   if [[ $CALIB_TPC_IDC == 1 ]] && [[ $AGGREGATOR_TASKS == TPCIDC_A || $AGGREGATOR_TASKS == TPCIDC_C || $AGGREGATOR_TASKS == TPC_IDCBOTH_SAC || $AGGREGATOR_TASKS == ALL ]]; then
     add_W o2-tpc-idc-distribute "--crus ${crus} --timeframes ${nTFs} --output-lanes ${lanesFactorize} --send-precise-timestamp true --condition-tf-per-query ${nTFs}"
-    add_W o2-tpc-idc-factorize "--input-lanes ${lanesFactorize} --crus ${crus} --timeframes ${nTFs} --nthreads-grouping 8 --nthreads-IDC-factorization 8 --sendOutputFFT true --enable-CCDB-output true --enablePadStatusMap true --use-precise-timestamp true --add-offset-for-CCDB-timestamp true" "TPCIDCGroupParam.groupPadsSectorEdges=32211"
+    add_W o2-tpc-idc-factorize "--input-lanes ${lanesFactorize} --crus ${crus} --timeframes ${nTFs} --nthreads-grouping 8 --nthreads-IDC-factorization 8 --sendOutputFFT true --enable-CCDB-output true --enablePadStatusMap true --use-precise-timestamp true --add-offset-for-CCDB-timestamp true" "TPCIDCGroupParam.groupPadsSectorEdges=32211 $IDC_DELTA"
```
matthias-kleiner: The  $IDC_DELTA should be between the first set of quotation marks since the second quotation marks are for the configKeyValues, I think
    
    add_W o2-tpc-idc-factorize "--input-lanes ${lanesFactorize} --crus ${crus} --timeframes ${nTFs} --nthreads-grouping 8 --nthreads-IDC-factorization 8 --sendOutputFFT true --enable-CCDB-output true --enablePadStatusMap true --use-precise-timestamp true --add-offset-for-CCDB-timestamp true $IDC_DELTA" "TPCIDCGroupParam.groupPadsSectorEdges=32211"

davidrohr: yes, since $IDC_DELTA is a command line option and not a configKeyValue, then you are right.

prodtests/full-system-test/aggregator-workflow.sh
```diff
@@ -221,11 +221,14 @@ fi
 lanesFactorize=6
 nTFs=1000
 nTFs_SAC=1000
+IDC_DELTA="--disable-IDCDelta true" # by default, deltas are off
```
davidrohr: I have to say I don't understand this ENABLE_IDC_DELTA=empty string. That is a bit confusing to use. The default in the code is to have --disable-IDCDelta disabled. If I understand correctly, we want to enable it by default for PbPb, or upon a special setting. I'd simply do:
```
IDC_DELTA=""
[[ $BEAMTYPE != "PbPb" && "0$ENABLE_IDC_DELTA" != "01" ]] && IDC_DELTA="--disable-IDCDelta true"
```
That is more intuitive to me.

prodtests/full-system-test/aggregator-workflow.sh
chiarazampolli: Indeed.

prodtests/full-system-test/aggregator-workflow.sh
chiarazampolli: This was my first way. Then I thought that if at P2 they just write `ENABLE_IDC_DELTA=true` or `ENABLE_IDC_DELTA=` or `ENABLE_IDC_DELTA=1`, which are all quite clear to me, they should all set it.  
If you think that only the `1` should do it, then I can f course change it. 
I don't have a strong preference, I have to say, but since you do, I will modify it.

davidrohr: indeed I do :), We do not support `=true` / `=false` anywhere, so I would not add another syntax, which is not uniform. That would just create confusion.
For the `=[empty]`: indeed, we have that, but it comes from some previous expert settings, which I actually don't like, and we shouldn't do it that way. It is also failure prone, since it is not always propagated correctly to subshells. Perhaps we should even get rid of it.

chiarazampolli: BTW, the idea is to enable it by default in PbPb, I think this was clear (I point it out since you wrote `we want to enable it by default for pp`).

davidrohr: ok..., that was a typo indeed. Fixed it now. But the code I posted was correct :)

chiarazampolli: Please check the new commit. It is different from what you proposed, but no empty variables, just checking whether the variables are `0` or `1`.

davidrohr (dismissed):
fine with me, should be squashed when merged.

chiarazampolli:
Yes, I wanted first to show you the change in a more explicit way. 
I will squash now. Then let's wait for the CI.

PR #10204, opened 2022-11-03
DPL: enable lossy dispatcher

ktf:
This will drop data after 10ms for the dispatcher. I am not using 0 for now to make sure there is not some weird behaviour which drops data immediately at 88Hz.

### DIFF
--- a/Framework/Core/src/SendingPolicy.cxx
+++ b/Framework/Core/src/SendingPolicy.cxx
@@ -28,8 +28,22 @@ std::vector<SendingPolicy> SendingPolicy::createDefaultPolicies()
             .name = "dispatcher",
             .matcher = [](DeviceSpec const& spec, ConfigContext const&) { return spec.name == "Dispatcher" || DeviceSpecHelpers::hasLabel(spec, "Dispatcher"); },
             .send = [](FairMQDeviceProxy& proxy, fair::mq::Parts& parts, ChannelIndex channelIndex, ServiceRegistryRef registry) {
-              auto *channel = proxy.getOutputChannel(channelIndex);
-              channel->Send(parts, -1); }},
+              OutputChannelInfo const& info = proxy.getOutputChannelInfo(channelIndex);
+              OutputChannelState& state = proxy.getOutputChannelState(channelIndex);
+              // Default timeout is 10ms.
+              // We count the number of consecutively dropped messages.
+              // If we have more than 10, we switch to a completely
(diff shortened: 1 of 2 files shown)


ktf:
@davidrohr @shahor02 @knopers8 @Barthelemy

knopers8 (dismissed):

ktf:
Updated as suggested by @davidrohr. 10 times in a row failing to send and the timeout goes to 0ms, to then go back to 10ms as soon as we succeed sending.

davidrohr (dismissed):
Looks good.
Just one thing I am not sure: non-zero return value signals success? But I guess you know the API better than I do.

ktf:
It's supposed to be the number of messages sent, but indeed i am not sure about a negative return value. Inverted the logic so that it's correct in any case.

ktf:
Also avoiding the (unlikely) issue if we are dropping more than INT64_MAX messages. I guess if we have a run that long, we have other kind of problems to worry about... :-)

Framework/Core/src/SendingPolicy.cxx
```diff
@@ -28,8 +28,22 @@ std::vector<SendingPolicy> SendingPolicy::createDefaultPolicies()
...
-              auto *channel = proxy.getOutputChannel(channelIndex);
-              channel->Send(parts, -1); }},
+              OutputChannelInfo const& info = proxy.getOutputChannelInfo(channelIndex);
+              OutputChannelState& state = proxy.getOutputChannelState(channelIndex);
+              // Default timeout is 10ms.
+              // We count the number of consecutively dropped messages.
+              // If we have more than 10, we switch to a completely
+              // non-blocking approach.
+              int64_t timeout = 10;
+              if (state.droppedMessages > 10) {
```
knopers8: I like the proposition by David, but perhaps a warning could be added when we switch to fully non-blocking?
```suggestion
              int64_t timeout = 10;
              if (state.droppedMessages == 10 + 1) {
                LOGP(warning) << "Failed to send 10 messages with 10ms timeout in a row, switching to completely non-blocking mode"; 
              } 
              if (state.droppedMessages > 10) {
```

davidrohr (dismissed):

ktf:
Ok, merging this since it passes all the tests. Please be aware of the change when you deploy, as it might mean that QC will start to be lossy. @knopers8 @Barthelemy @davidrohr @shahor02

PR #10205, opened 2022-11-03
ZDC - Adjust EOV from CalibParam object

cortesep:


shahor02 (commented):
One can do in this way, but why don't you simply set the EoV to e.g. `run-end + o2::ccdb::CcdbObjectInfo::YEAR` in the calibrators directly, e.g. in the InterCalib::endOfRun() ?

It will be automatically readjusted when loading the next object.

cortesep:
@shahor02 Thanks for the tip. Now CalibParamZDC.eovTune defaults to 1 year.

shahor02 (approved):

PR #10207, opened 2022-11-03
Allow to set different thresholds for MID dead and noisy channels

dstocco:
Currently, the noisy channels were computed only when occurring in the same BC of a calibration trigger.
But the calibration run occurs without beam, so all hits (except the answer to the Front-end test) are noisy.
With this PR, the noisy channels are defined as all hits occuriring outside of a FET event.
Also, different thresholds can be applied to the define noisy channels (in Hz) and to define a dead channel (as a fraction of the number of FET events).
The PR consist of 2 commits:
- the first one merges self-triggered events with events occurring during a calibration trigger
- the second one allows to set different thresholds for noisy and dead channels, computing the masks separately for the two kind of events and then merging them.

### DIFF
--- a/Detectors/MUON/MID/Workflow/src/CalibDataProcessorSpec.cxx
+++ b/Detectors/MUON/MID/Workflow/src/CalibDataProcessorSpec.cxx
@@ -9,12 +9,12 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-/// \file   MID/Workflow/src/FetToDeadSpec.cxx
-/// \brief  Device to convert the FEE test event into dead channels
+/// \file   MID/Workflow/src/CalibDataProcessorSpec.cxx
+/// \brief  Device to convert the calibration data into a list of bad channel candidates
 /// \author Diego Stocco <Diego.Stocco at cern.ch>
-/// \date   21 February 2022
+/// \date   25 October 2022
 
-#include "MIDWorkflow/FetToDeadSpec.h"
+#include "MIDWorkflow/CalibDataProcessorSpec.h"
 
 #include <array>
 #include <vector>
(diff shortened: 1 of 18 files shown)


Detectors/MUON/MID/Workflow/src/CalibDataProcessorSpec.cxx
```diff
@@ -74,31 +74,39 @@ class FetToDeadDeviceDPL
...
+  std::vector<ROFRecord> mDeadROF;  ///< Merged dead ROFs
 
-  std::pair<std::vector<ColumnData>, std::vector<ROFRecord>> getDeadChannels(gsl::span<const ColumnData> fetData, gsl::span<const ROFRecord> fetDataRof, gsl::span<const ColumnData> selfTrigData, gsl::span<const ROFRecord> selfTrigDataRof)
+  void mergeChannels(gsl::span<const ColumnData> fetData, gsl::span<const ROFRecord> fetDataRof, gsl::span<const ColumnData> selfTrigData, gsl::span<const ROFRecord> selfTrigDataRof)
   {
+    // This method selects the self-triggered events that are close in time with a FET event and merges them with the fet event.
+    // This is needed since the detector answer is not perfectly aligned in time.
+    // Since calibration runs occur with no beam, all other self-triggers are actually noise.
+
     int64_t maxDiff = 1;
```
aphecetche: Don't you want `minDiff` and `maxDiff` to be either parameters of this `mergeChannel` or data member of the class ? Seems strange to hardcode it within the function ?

aphecetche:
@dstocco I'd assume the `Workflow/README.md` calibration section would need an update also ?

Detectors/MUON/MID/Workflow/src/CalibDataProcessorSpec.cxx
dstocco: I doubt it will change, but you're right that it is better to allow for the possibility.

dstocco:
Indeed: added now.

aphecetche (approved):

PR #10215, opened 2022-11-04
Protections added in DCS DPs processing

gvolpe79:


Detectors/HMPID/calibration/src/HMPIDDCSProcessor.cxx
```diff
@@ -483,16 +482,18 @@ bool HMPIDDCSProcessor::evalCorrFactor(double dRefArgon, double dCellArgon,
 
   // evaluate 15 mm of thickness C6F14 Trans
 
-  aConvFactor = 1.0 - 0.3 / 1.8;
+  // ef: check if all are not nullpointers
+  if (&dRefArgon == nullptr || &dCellArgon == nullptr || &dRefFreon == nullptr || &dRefFreon == nullptr) {
```
shahor02: The macOS CI complaints are genuine: the address of the reverence cannot be nullptr. You should probably check a specific value (passing a pointer seems to be meaningless since the method is called with the reference of double.

Detectors/HMPID/calibration/include/HMPIDCalibration/HMPIDDCSProcessor.h
```diff
@@ -177,6 +138,94 @@ class HMPIDDCSProcessor
     arNmean.clear();
   }
 
+  void checkEntries(const std::vector<TF1>& arQthresh,
```
shahor02: This method just prints a warning on detecting the default object, w/o any effect on the output, is this what you want?

shahor02 (changes requested):

Detectors/HMPID/calibration/src/HMPIDDCSProcessor.cxx
gvolpe79: modified!

Detectors/HMPID/calibration/include/HMPIDCalibration/HMPIDDCSProcessor.h
gvolpe79: Yes, indeed this method is not really useful, I removed it.

shahor02 (approved):

PR #10216, opened 2022-11-04
ctpdev: ctp readout checking macros

lietava:
Two macros:
CheckCTPDigitts.C: checks if every Trigger Class Record (TCR) has corresponding Input Record.
CheckAOD2CTPDigits.C: checks if all TCR in CTP digits propagated to AOD
Miscellaneous:
time  added to output of scaler rates.


lietava:
PLease, merge.

PR #10224, opened 2022-11-06
Add CTP lumi information to TPC SCD calibration

martenole:
Hi @chiarazampolli with https://github.com/AliceO2Group/AliceO2/commit/a33a3b55ccc874e92a62890e451db15e1b8679fd I am changing a bit the interface of the TimeSlotCalibration. The change is fully backwards compatible and allows the user to define a `fill()` method for their container with as many inputs as wanted. I am using this for the residual aggregator to also process the lumi information, if it is available. Also I am making the `Container` type the same as the `Input` type by default, because in some cases they are identical (see https://github.com/AliceO2Group/AliceO2/commit/f4ecaa65b94844b3afca1cc73a1e54f254e1e364)
@shahor02 in order to be able to forward the lumi information I am using different subspecs on the processing EPNs, because otherwise the `o2-tpc-scdcalib-interpolation-workflow` would depend on itself.

The averaging for a given calibration interval I still need to add. Will be done in a follow-up PR. Also this needs a change to O2DPG to become effective.

### DIFF
--- a/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
+++ b/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h
@@ -36,7 +36,7 @@ class ProcessingContext;
 namespace calibration
 {
 
-template <typename Input, typename Container>
+template <typename Input, typename Container = Input>
 class TimeSlotCalibration
 {
  public:
@@ -104,8 +104,8 @@ class TimeSlotCalibration
   const Slot& getLastSlot() const { return (Slot&)mSlots.back(); }
   const Slot& getFirstSlot() const { return (Slot&)mSlots.front(); }
 
-  template <typename DATA>
-  bool process(const DATA& data);
+  template <typename... DATA>
+  bool process(const DATA&... data);
   virtual void checkSlotsToFinalize(TFType tf, int maxDelay = 0);
   virtual void finalizeOldestSlot();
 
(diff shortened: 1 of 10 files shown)


shahor02:
@martenole, thanks, looks fine. But why do you need to forward the lumi info via o2-tpc-scdcalib-interpolation-workflow ? On the epn side you could simply add it with subspec 0 to the output proxy and it would forward it from whatever primary source (ctf-reader of ctf-decoder) provides it.

martenole:
> @martenole, thanks, looks fine. But why do you need to forward the lumi info via o2-tpc-scdcalib-interpolation-workflow ? On the epn side you could simply add it with subspec 0 to the output proxy and it would forward it from whatever primary source (ctf-reader of ctf-decoder) provides it.

That is a good point.. Do some writer/reader already exist for the lumi information? In case one wants to run the calibration from ROOT files directly?

shahor02:
@martenole sure, the CTP digits/lumi are read by the o2-global-track-cluster-reader via InputHelper if CTP is requested (calling `O2/Detectors/CTP/workflowIO/src/DigitReaderSpec.cxx`) and added to the RecoContainer if CTP is present either in `tracks` or `clusters` request masks.

chiarazampolli (dismissed):
Thanks @martenole , fine with me.

martenole:
Hi @shahor02 
would it be fine with you if I merge this?
Cheers,
Ole

shahor02 (approved):

PR #10229, opened 2022-11-08
TRD: Tracker: Add padrow crossing information

f3sch:
This patch adds to the GPUTRDTrack one new members:
     1. mIsPadrowCrossing:
        unsigned char -> bits 0 to 5 indicate if in a layer the track
        crossed a padrow and bit 6 indicate whether a neighborign was found.
        Rational: Charge information is lost in these tracklets and
        one should possibly even discard those Tracks for gain
        calibration and PID.

Additionally, this was added to TrackQC.

### DIFF
--- a/Detectors/TRD/qc/include/TRDQC/Tracking.h
+++ b/Detectors/TRD/qc/include/TRDQC/Tracking.h
@@ -29,6 +29,7 @@
 #include "Rtypes.h"
 
 #include <gsl/span>
+#include <bitset>
 
 namespace o2
 {
@@ -42,15 +43,16 @@ namespace trd
 {
 
 struct TrackQC {
-  int type;          ///< 0 TPC-TRD track; 1 ITS-TPC-TRD track
-  int nTracklets;    ///< number of attached TRD tracklets
-  int nLayers;       //< Number of Layers of a Track in which the track extrapolation was in geometrical acceptance of the TRD
-  float chi2;        ///< total chi2 value for the track
-  float reducedChi2; ///< chi2 total divided by number of layers in which track is inside TRD geometrical acceptance
-  float p;           ///< the total momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
-  float pt;          ///< the transverse momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
-  float ptSigma2;    //< Sigma2 of pt
-  float dEdxTotTPC;  //< raw total dEdx information for seeding track in TPC
+  int type;                        ///< 0 TPC-TRD track; 1 ITS-TPC-TRD track
+  int nTracklets;                  ///< number of attached TRD tracklets
+  int nLayers;                     //< Number of Layers of a Track in which the track extrapolation was in geometrical acceptance of the TRD
+  float chi2;                      ///< total chi2 value for the track
+  float reducedChi2;               ///< chi2 total divided by number of layers in which track is inside TRD geometrical acceptance
+  float p;                         ///< the total momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
+  float pt;                        ///< the transverse momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
+  float ptSigma2;                  ///< Sigma2 of pt
+  float dEdxTotTPC;                ///< raw total dEdx information for seeding track in TPC
+  std::bitset<6> isPadrowCrossing; ///< indicate if track crossed a padrow in that layer
 
   // layer-wise information for seeding track and assigned tracklet (if available)
   std::array<bool, constants::NLAYER> findable{};  ///< flag if track was in geometrical acceptance
@@ -80,7 +82,7 @@ struct TrackQC {
   std::array<int, constants::NLAYER> trackletMcm{};                                      ///< the MCM number of the tracklet
   std::array<float, constants::NLAYER> trackletChi2{};                                   ///< estimated chi2 for the update of the track with the given tracklet
   std::array<std::array<int, constants::NCHARGES>, constants::NLAYER> trackletCharges{}; ///< charges of tracklets
-  ClassDefNV(TrackQC, 2);
+  ClassDefNV(TrackQC, 3);
 };
 
 class Tracking
--- a/Detectors/TRD/qc/include/TRDQC/Tracking.h
+++ b/Detectors/TRD/qc/include/TRDQC/Tracking.h
@@ -53,6 +53,7 @@ struct TrackQC {
   float ptSigma2;                  ///< Sigma2 of pt
   float dEdxTotTPC;                ///< raw total dEdx information for seeding track in TPC
   std::bitset<6> isPadrowCrossing; ///< indicate if track crossed a padrow in that layer
+  bool hasNeighbor;                ///< indicate if track crossed a padrow in that layer
 
   // layer-wise information for seeding track and assigned tracklet (if available)
   std::array<bool, constants::NLAYER> findable{};  ///< flag if track was in geometrical acceptance
--- a/Detectors/TRD/qc/include/TRDQC/Tracking.h
+++ b/Detectors/TRD/qc/include/TRDQC/Tracking.h
@@ -43,17 +43,18 @@ namespace trd
 {
 
 struct TrackQC {
-  int type;                        ///< 0 TPC-TRD track; 1 ITS-TPC-TRD track
-  int nTracklets;                  ///< number of attached TRD tracklets
-  int nLayers;                     //< Number of Layers of a Track in which the track extrapolation was in geometrical acceptance of the TRD
-  float chi2;                      ///< total chi2 value for the track
-  float reducedChi2;               ///< chi2 total divided by number of layers in which track is inside TRD geometrical acceptance
-  float p;                         ///< the total momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
-  float pt;                        ///< the transverse momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
-  float ptSigma2;                  ///< Sigma2 of pt
-  float dEdxTotTPC;                ///< raw total dEdx information for seeding track in TPC
-  std::bitset<6> isPadrowCrossing; ///< indicate if track crossed a padrow in that layer
-  bool hasNeighbor;                ///< indicate if track crossed a padrow in that layer
+  int type;                          ///< 0 TPC-TRD track; 1 ITS-TPC-TRD track
+  int nTracklets;                    ///< number of attached TRD tracklets
+  int nLayers;                       //< Number of Layers of a Track in which the track extrapolation was in geometrical acceptance of the TRD
+  float chi2;                        ///< total chi2 value for the track
+  float reducedChi2;                 ///< chi2 total divided by number of layers in which track is inside TRD geometrical acceptance
+  float p;                           ///< the total momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
+  float pt;                          ///< the transverse momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
+  float ptSigma2;                    ///< Sigma2 of pt
+  float dEdxTotTPC;                  ///< raw total dEdx information for seeding track in TPC
+  std::bitset<6> isCrossingNeighbor; ///< indicate if track crossed a padrow and/or had a neighboring tracklet in that layer
+  bool hasNeighbor;                  ///< indicate if a track had a tracklet with a neighboring one e.g. potentailly split tracklet
+  bool hasPadrowCrossing;            ///< indicate if track crossed a padrow
 
   // layer-wise information for seeding track and assigned tracklet (if available)
   std::array<bool, constants::NLAYER> findable{};  ///< flag if track was in geometrical acceptance
--- a/GPU/GPUTracking/DataTypes/GPUTRDTrack.cxx
+++ b/GPU/GPUTracking/DataTypes/GPUTRDTrack.cxx
@@ -34,6 +34,7 @@ GPUd() void GPUTRDTrack_t<T>::initialize()
   mRefGlobalTrackId = 0;
   mCollisionId = -1;
   mFlags = 0;
+  mIsPadrowCrossing = 0;
   for (int i = 0; i < kNLayers; ++i) {
     mAttachedTracklets[i] = -1;
   }
@@ -69,6 +70,7 @@ GPUd() void GPUTRDTrack_t<T>::ConvertTo(GPUTRDTrackDataRecord& t) const
   for (int i = 0; i < kNLayers; i++) {
     t.fAttachedTracklets[i] = getTrackletIndex(i);
   }
+  t.mIsPadrowCrossing = getIsPadrowCrossing();
 }
 
 template <typename T>
@@ -81,6 +83,7 @@ GPUd() void GPUTRDTrack_t<T>::ConvertFrom(const GPUTRDTrackDataRecord& t)
   setRefGlobalTrackIdRaw(t.fTPCTrackID);
   mChi2 = 0.f;
   mFlags = 0;
+  mIsPadrowCrossing = 0;
   mCollisionId = -1;
   for (int iLayer = 0; iLayer < kNLayers; iLayer++) {
     mAttachedTracklets[iLayer] = t.fAttachedTracklets[iLayer];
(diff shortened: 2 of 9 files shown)


GPU/GPUTracking/DataTypes/GPUTRDTrack.h
```diff
@@ -106,13 +107,15 @@ class GPUTRDTrack_t : public T
...
   // conversion to / from HLT track structure (only for AliRoot)
   GPUd() void ConvertTo(GPUTRDTrackDataRecord& t) const;
   GPUd() void ConvertFrom(const GPUTRDTrackDataRecord& t);
 
  protected:
   float mChi2;                      // total chi2.
+  unsigned char mIsPadrowCrossing;  // bits 0 to 5 indicate wheter track was in between two padrows
   unsigned int mRefGlobalTrackId;   // raw GlobalTrackID of the seeding track (either ITS-TPC or TPC)
   int mAttachedTracklets[kNLayers]; // indices of the tracklets attached to this track; -1 means no tracklet in that layer
   short mCollisionId;               // the collision ID of the tracklets attached to this track; is used to retrieve the BC information for this track after the tracking is done
```
shahor02: Could you move mCollisionID before mAttached..., to not lose 2 bytes on the alignment?

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
```diff
@@ -657,6 +657,22 @@ GPUd() bool GPUTRDTracker_t<TRDTRK, PROP>::FollowProlongation(PROP* prop, TRDTRK
...
+          prop->getPropagatedYZ(trkWork->getX() - 3.f, projYEntry, projZEntry);
+          // Get Padrow number for Exit&Entry and compare. If is not equal mark
+          // as padrow crossing. While simple, this comes with the cost of not
+          // properly handling Tracklets that hit a different Pad but still get
+          // attached to the Track. It is estimated that the probability for
+          // this is low.
+          auto padrowEntry = pad->GetPadRowNumber(projZEntry);
+          auto padrowExit = pad->GetPadRowNumber(trkWork->getX());
+          if (padrowEntry != padrowExit) {
+            trkWork->setIsPadrowCrossing(iLayer);
```
shahor02: Should not this be be validated by the presence of a tracklet in the neighbouring padrow?

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
```diff
@@ -657,6 +657,22 @@ GPUd() bool GPUTRDTracker_t<TRDTRK, PROP>::FollowProlongation(PROP* prop, TRDTRK
...
+          float projZEntry, projYEntry;
+          // Get Z for Entry of Track
+          prop->getPropagatedYZ(trkWork->getX() - 3.f, projYEntry, projZEntry);
+          // Get Padrow number for Exit&Entry and compare. If is not equal mark
+          // as padrow crossing. While simple, this comes with the cost of not
+          // properly handling Tracklets that hit a different Pad but still get
+          // attached to the Track. It is estimated that the probability for
+          // this is low.
+          auto padrowEntry = pad->GetPadRowNumber(projZEntry);
+          auto padrowExit = pad->GetPadRowNumber(trkWork->getX());
```
martenole:
```suggestion
          auto padrowExit = pad->GetPadRowNumber(trkWork->getZ());
```

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
```diff
@@ -657,6 +657,22 @@ GPUd() bool GPUTRDTracker_t<TRDTRK, PROP>::FollowProlongation(PROP* prop, TRDTRK
             GPUWarning("Track parameter for track %i, x=%f at chamber %i x=%f in layer %i cannot be retrieved", iTrk, trkWork->getX(), currDet, mR[currDet], iLayer);
           }
         }
+        // check if track crosses padrows
+        if (iDet == 0) {
+          float projZEntry, projYEntry;
+          // Get Z for Entry of Track
+          prop->getPropagatedYZ(trkWork->getX() - 3.f, projYEntry, projZEntry);
```
martenole: The drift length of 3cm you should be able to find in the geometry

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
martenole: Yes, when thinking about it this would make more sense. We would probably also want the information about pad row crossing only in case a tracklet is attached. So then you could move this check a bit down to when a tracklet is actually attached to the track. And in that case do in addition the check whether another tracklet is present on the neighboring pad row. Since the tracklets are not sorted within one chamber this probably requires a loop over all tracklet of the given chamber

GPU/GPUTracking/DataTypes/GPUTRDTrack.h
```diff
@@ -86,6 +86,7 @@ class GPUTRDTrack_t : public T
   GPUd() short getCollisionId() const { return mCollisionId; }
   GPUd() int getNtracklets() const;
   GPUd() float getChi2() const { return mChi2; }
+  GPUd() unsigned char getIsPadrowCrossing() const { return mIsPadrowCrossing; }
```
martenole: Maybe you could also add a getter `bool getIsPadrowCrossing(int layer)` for convenience?

GPU/GPUTracking/DataTypes/GPUTRDTrack.cxx
```diff
@@ -31,6 +31,7 @@ GPUd() void GPUTRDTrack_t<T>::initialize()
 {
   // set all members to their default values (needed since in-class initialization not possible with AliRoot)
   mChi2 = 0.f;
+  mIsPadrowCrossing = 0;
```
martenole: There are more functions in `GPUTRDTrack` where the `mPadRowCrossing` variable should be initialized. It should in addition be added to `GPU/GPUTracking/TRDTracking/GPUTRDTrackData.h`

Detectors/TRD/qc/include/TRDQC/Tracking.h
```diff
@@ -42,15 +42,16 @@ namespace trd
...
+  int type;                       ///< 0 TPC-TRD track; 1 ITS-TPC-TRD track
+  int nTracklets;                 ///< number of attached TRD tracklets
+  int nLayers;                    //< Number of Layers of a Track in which the track extrapolation was in geometrical acceptance of the TRD
+  float chi2;                     ///< total chi2 value for the track
+  float reducedChi2;              ///< chi2 total divided by number of layers in which track is inside TRD geometrical acceptance
+  float p;                        ///< the total momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
+  float pt;                       ///< the transverse momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
+  float ptSigma2;                 ///< Sigma2 of pt
+  float dEdxTotTPC;               ///< raw total dEdx information for seeding track in TPC
+  unsigned char isPadrowCrossing; ///< bits 0 to 5 indicate if track crossed a padrow in that layer
```
martenole: In here you could also use a bitset, since this data type will not be used on the GPU

GPU/GPUTracking/DataTypes/GPUTRDTrack.h
f3sch: Sure, did not know that alignment was a concern here. Sorry.
But from my very limited understanding of the topic, the situation is like this:

```
// size = 36, alignment = 4
// before:  +----+----+-*6-+----+
// members: |aaaa|bbbb|c**c|dde.|
// . represents a byte of padding
// a=mChi2, b=mRefGlobalTrackId, c=mAttachedTracklets, d=mCollisionId, e=mFlags
```

```
// size = 40, alignment = 4
// mine:    +----+----+----+-*6-+----+
// members: |aaaa|p...|bbbb|c**c|dde.|
// p=mIsPadrowCrossing
```

```
// size = 44, alignment = 4
// yours:   +----+----+----+----+-*6-+----+
// members: |aaaa|p...|bbbb|dd..|c**c|e....|
```

Maybe better is:
```
// size = 36, alignment = 4
// mine:    +----+----+-*6-+----+
// members: |aaaa|bbbb|c**c|ddep|
```
e.g. moving mIsPadrowCrossing after mFlags.

At least that what I get from testing in an interactive root session and defining
struct foo {...} and playing with alignof() and sizeof() on x86, not sure if this is valid since this is run on a GPU?

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
f3sch: Missed that thanks.

GPU/GPUTracking/DataTypes/GPUTRDTrack.h
martenole: I think this was just a typo. Putting `mIsPadrowCrossing` after `mFlags` is fine

martenole: Just not to put a 1 byte member between 2 4 byte members

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
tdietel: I think the check for tracklets in neighboring padrows is complementary: it would detect if more information is available that should be merged into the already attached tracklet. A similar argument could actually be made for tracklets in the same padrow, especially at low pT and large inclination angle.
The flagging based on tracking only would indicate if some signal was deposited on a neighboring padrow, even it is below the threshold required to generate a tracklet. Even in such a case, the quality of at least one time slice would be compromised, but it would not be detectable by looking for neighboring tracklets.

GPU/GPUTracking/DataTypes/GPUTRDTrack.h
f3sch: Done.

GPU/GPUTracking/DataTypes/GPUTRDTrack.cxx
f3sch: Think I got them all...

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
f3sch: Ok.

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
martenole: > A similar argument could actually be made for tracklets in the same padrow, especially at low pT and large inclination angle.

Are you sure about this? I think low pT tracks spanning over more than 4 pads would rather not be found at all. And for the ones we find (>~700MeV) we should not loose charge information in column direction.

The problem I see when relying purely on the information from the track parameters is that especially for TPC tracks the estimate for the z position might be quite unprecise while eta should be rather well described. This means you can shift the track to some extent along z. And when doing that you can of course always find some position where the track crossed pad rows, but also a position where it did not. Checking if there really were 2 tracklets might be a clearer indication. But that would require a good efficiency for the online tracklets of course. Probably it would be good to check in Run 3 data how often this is the case that for a track two neighboring tracklets are in its vicinity. Could be done both in column and in row direction

tdietel: > Are you sure about this? I think low pT tracks spanning over more than 4 pads would rather not be found at all. And for the ones we find (>~700MeV) we should not loose charge information in column direction.

I don't have numbers for this. I remember that a long time ago, we looked into tracklet splitting, and found that a significant number of tracklets (I remember 10-20%) had duplicates, i.e. were reconstructed twice in neighboring channels. But I do not know how how big the differences for the tracklet charges were.

tdietel: > The problem I see when relying purely on the information from the track parameters is that especially for TPC tracks the estimate for the z position might be quite unprecise while eta should be rather well described. This means you can shift the track to some extent along z. And when doing that you can of course always find some position where the track crossed pad rows, but also a position where it did not.

That's a good point. Do you think this would also be a problem for ITS-TPC-TRD tracks? I guess in this case, the timing information should be good enough to fix z?

And to which degree does the TRD fix the z-position of the TPC-TRD tracks? I think we only use 

> Checking if there really were 2 tracklets might be a clearer indication. But that would require a good efficiency for the online tracklets of course. Probably it would be good to check in Run 3 data how often this is the case that for a track two neighboring tracklets are in its vicinity. Could be done both in column and in row direction

It seems that both checks test slightly different things, and both could be useful in different scenarios, e.g. ITS-TPC-TRD tracks without padrow crossings for gain calibration, where you prefer high-quality tracks, and a check for neighboring tracklets for PID, where you are more interested in efficiency.

I would therefore implement both.

f3sch: > ITS-TPC-TRD tracks without padrow crossings for gain calibration, where you prefer high-quality tracks

Do you(@tdietel) want to generally throw away tracks if they have a padrow crossing for gain calibration or would you like to have this layerwise?

Detectors/TRD/qc/include/TRDQC/Tracking.h
```diff
@@ -42,15 +43,16 @@ namespace trd
...
+  int type;                        ///< 0 TPC-TRD track; 1 ITS-TPC-TRD track
+  int nTracklets;                  ///< number of attached TRD tracklets
+  int nLayers;                     //< Number of Layers of a Track in which the track extrapolation was in geometrical acceptance of the TRD
+  float chi2;                      ///< total chi2 value for the track
+  float reducedChi2;               ///< chi2 total divided by number of layers in which track is inside TRD geometrical acceptance
+  float p;                         ///< the total momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
+  float pt;                        ///< the transverse momentum of the track at the point of the innermost ITS cluster (ITS-TPC-TRD) or at the inner TPC radius (TPC-TRD)
+  float ptSigma2;                  ///< Sigma2 of pt
+  float dEdxTotTPC;                ///< raw total dEdx information for seeding track in TPC
+  std::bitset<6> isPadrowCrossing; ///< indicate if track crossed a padrow in that layer
```
martenole: Maybe you could directly add the global track ID as well?

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
```diff
@@ -821,6 +821,33 @@ GPUd() bool GPUTRDTracker_t<TRDTRK, PROP>::FollowProlongation(PROP* prop, TRDTRK
...
+        continue; // we can only mark padrow crossings once, hence the next loop is pointless
+      }
+      auto currDet = tracklets[mHypothesis[iUpdate + hypothesisIdxOffset].mTrackletId].GetDetector();
+      // Mark tracklets as Padrow crossing if they have a neighboring tracklet.
+      for (int trkltIdx = glbTrkltIdxOffset + mTrackletIndexArray[trkltIdxOffset + currDet]; trkltIdx < glbTrkltIdxOffset + mTrackletIndexArray[trkltIdxOffset + currDet + 1]; ++trkltIdx) {
+        // skip orig tracklet
+        if (mHypothesis[iUpdate + hypothesisIdxOffset].mTrackletId == trkltIdx) {
+          continue;
+        }
+        if (GPUCommonMath::Abs(tracklets[mHypothesis[iUpdate + hypothesisIdxOffset].mTrackletId].GetZbin() - tracklets[trkltIdx].GetZbin())) {
```
martenole: Should you not also check the distance along pad column direction? And the check the check in row direction also does not look correct. You should check that the distance in row direction is exactly 1 pad

f3sch: Good catch!

Detectors/TRD/qc/include/TRDQC/Tracking.h
f3sch: I am still figuring the TOF id out. Let us not intermix features/request across PRs and first get this merged.

martenole: Well it would be just adding one more integer to the QC struct but as you wish ;)

f3sch:
I could reproduce your numbers for the crossings and neighboring tracklets, after running the tracking:
image
So the code should now be ready to be merged.

martenole:
And what is the fraction now with the code as is in this PR? Should be less than 40% now, right?

f3sch:
The numbers above are with the current PR, my original numbers were a combination of two things 1. without checking the distances in padrow and 2. me not recompiling O2PDPSuite... hence not linking to the new O2 compilation (I think...)

martenole:
so now its 11% with the combined condition `padRowEntry != padRowExit || neighbouringTrackletAvailable`?

GPU/GPUTracking/TRDTracking/GPUTRDTracker.cxx
```diff
@@ -821,6 +821,35 @@ GPUd() bool GPUTRDTracker_t<TRDTRK, PROP>::FollowProlongation(PROP* prop, TRDTRK
...
+      // Get Padrow number for Exit&Entry and compare. If is not equal mark
+      // as padrow crossing. While simple, this comes with the cost of not
+      // properly handling Tracklets that hit a different Pad but still get
+      // attached to the Track. It is estimated that the probability for
+      // this is low.
+      const auto padrowEntry = pad->GetPadRowNumber(projZEntry);
+      const auto padrowExit = pad->GetPadRowNumber(trkWork->getZ());
+      if (padrowEntry != padrowExit) {
+        trkWork->setIsPadrowCrossing(iLayer);
+        continue; // we can only mark padrow crossings once, hence the next loop is pointless
```
martenole: I guess with the `hasNeighbor` check one should still do the check below, right?

f3sch: yes, will fix and report back with new numbers.

f3sch:
Finally got the numbers done, here in matrix form:
|          | Crossings (%) | Neighbors (%) |
|----------|--------------|--------------|
| Both     | 11.9         | 9.5          | 
| Crossings | 11.4         | 0.0          |
| Neighbors | 0.0          | 8.8          |

-> there is of course overlap show in Both
from 100 TF  with 522829 tracklets

martenole (approved):

PR #10230, opened 2022-11-08
EMCAL: drop direct infologger access

ktf:


ktf:
InfoLogger should not be touched by the tasks, since it might or might not be there. In particular it will not be there when running asynchronously on the grid. Given that on the FLP we already have a mechanism to attach the correct detector name to log entries, I think we can simply drop a few lines. If this is really needed, we should protect the access by checking with `active<InfoLoggerContext>()`, before doing the `get`.

mfasDa:
If there are other ways to set the detector field fine, otherwise we would object, in particular in global runs with many detectors sending messages it might be tricky to find the relevant ones and we are using the detector field to filter ours.

ktf:
Do you mean on the EPN? Because the FLPs have already another way to set the detector field.

One could also argue that if you filter by "dpl/emc" the facility, you will get all of your messages. IMHO it's better to solve this with a policy on the names of the dataprocessors, rather than with yet another field.

Anyway, I can add a protection instead. Keep in mind that on the grid in any case you will not get infologger and your processes will still pay the price for it (e.g. one extra spurious thread).

mfasDa:
> Do you mean on the EPN? Because the FLPs have already another way to set the detector field.

Yes, it was used on the EPN.

ktf:
Ok, this should be updated to check for the InfoLogger being active before using it.

ktf:
@TimoWilken this is pending at 3 since quite a while.

TimoWilken:
@ktf That builder keeps running out of memory and getting killed while compiling O2Physics for https://github.com/AliceO2Group/O2DPG/pull/717. I've moved the O2DPG check to the O2Physics-dedicated CI builders, which have more memory allocated, in https://github.com/alisw/ali-bot/pull/1216.

ktf:
This should maintain the old behaviour, yet allowing disabling infologger completely. Merging.

PR #10237, opened 2022-11-09
Do not cache DataProcessingContext & TimingInfo

ktf:
Spotted by some syntetic run. The copy was taking 10% of the CPU utilization.


mconcas (approved):

ktf:
Notice that you probably want to do the whole dance only once during the processing, but I will leave you to it.

iravasen:
Thanks @ktf for the modification!

PR #10239, opened 2022-11-09
TPC Clusterizer: Fix comparison between packed and unpacked charge in noisy pad filter.

fweig:


davidrohr:
thx! and thx for spotting!

davidrohr (approved):

PR #10241, opened 2022-11-09
TPC: IDC processing change fatal to warning, add possibility to dump bad data

wiechula:


wiechula:
The failing test seems unrelated. Can this be merged?

PR #10245, opened 2022-11-10
Residual aggregator output dir specified automatically

martenole:
To enable automatic transfer of calibration output to EOS

### DIFF
--- a/Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/include/TPCInterpolationWorkflow/TPCResidualAggregatorSpec.h
+++ b/Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/include/TPCInterpolationWorkflow/TPCResidualAggregatorSpec.h
@@ -62,7 +62,7 @@ class ResidualAggregatorDevice : public o2::framework::Task
       storeMetaFile = true;
     }
     if (!mWriteOutput && (outputDirConf != "none" || storeMetaFile)) {
-      LOGF(alarm, "File output is disabled, but output directory %s was specified and meta file storage is set to %i", outputDirConf, storeMetaFile);
+      LOGF(info, "File output is disabled, but output directory %s was specified and meta file storage is set to %i. No output file will be written", outputDirConf, storeMetaFile);
       storeMetaFile = false;
     }
     mAggregator = std::make_unique<o2::tpc::ResidualAggregator>(minEnt);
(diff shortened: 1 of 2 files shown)


martenole:
This does actually not distinguish between a synthetic and a physics run. Can I check the `RUNTYPE` variable and in case it is not `PHYSICS` disable the output completely? Or should something else be checked to distinguish the run types for the synch. reco?

davidrohr:
I think `$RUNTYPE` is the right thing to check.

martenole:
I tested this locally with default RUNTYPE=Standalone and with RUNTYPE=SYNTHETIC. Only in SYNTHETIC I disable the file writing, because otherwise when the run type is not PHYSICS the calibration does not run at all.
Since we have only one meta file directory for all kinds of data maybe we could in a followup PR change CTF_METAFILES_DIR to METAFILES_DIR and define that centrally in setenv.sh instead of only in dpl-workflow.sh as it is now. But I don't want to change that now so that hopefully this PR can be merged today.

chiarazampolli:
Hi @martenole , 
Which other run types (apart from PHYSICS) need this calibration? 
Chiara

davidrohr:
I think the METAINFO_DIR should go to gen_topo.sh, since it should be set centrally for the epns. All EPN settings should go there, not to setenv.sh.

martenole:
Hi Chiara, hi David,

> Which other run types (apart from PHYSICS) need this calibration?

this calibration is only needed for PHYSICS.

> I think the METAINFO_DIR should go to gen_topo.sh, since it should be set centrally for the epns. All EPN settings should go there, not to setenv.sh.

I agree it should be set in gen_topo.sh to the directory used on the EPNs. But in setenv.sh it should also be defined to `/dev/null` in case it was not previously set. The `/dev/null` is needed when not running on the EPNs

chiarazampolli:
Ciao Ole, 

If it is to be run only in PHYSICS (as I thought), I don't understand the comment from before:
```
Only in SYNTHETIC I disable the file writing, because otherwise when the run type is not PHYSICS the calibration does not run at all.
```
Sorry if it is trivial.
Chiara

martenole:
Ok I also did not say it correctly. In synthetics it should run as well, just to exercise the machinery. But in that case it should not create any ROOT output. In physics runs some files will be written and then transferred to EOS.

martenole:
build/O2/o2 is restarting all the time. Will merge soon to have it in tomorrow's tag

PR #10246, opened 2022-11-10
DPL: reduce level of messages which go to InfoLogger

ktf:
This moves all the messages to the support (6-10) level and below, apart those which are fatal.

Moreover, the first message in one of the levels between (6-10) is reported, with an additional comment to the shifter on where the rest of the messages will go.

### DIFF
--- a/Framework/DataTakingSupport/src/Plugin.cxx
+++ b/Framework/DataTakingSupport/src/Plugin.cxx
@@ -62,9 +62,14 @@ auto createInfoLoggerSinkHelper(InfoLogger* logger, InfoLoggerContext* ctx)
 {
   return [logger,
           ctx](const std::string& content, const fair::LogMetaData& metadata) {
+    if (!logger) {
+      return;
+    }
+
     // translate FMQ metadata
     InfoLogger::InfoLogger::Severity severity = InfoLogger::Severity::Undefined;
     int level = InfoLogger::undefinedMessageOption.level;
+    static int lastReportedToOps = 10;
 
     if (metadata.severity_name == fair::Logger::SeverityName(fair::Severity::nolog)) {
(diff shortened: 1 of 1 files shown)


ktf:
I received feedback that too many messages appear in the OPS (1-5) level, providing excessive logs to be parsed / understood to the shifter.

My proposal is that we report all the messages (but the first one) at a the level > 6 (i.e. support and beyond) for all the severities (but fatal).

This basically will mean that what the shifter will see is something along the lines of:

```
[WARNING]: some warning
[WARNING]: First message for severity WARNING encountered. The rest will be reported to support level only.
[ERROR]: some error
[ERROR]: First message for severity ERROR encountered. The rest will be reported to support level only.
[FATAL]: a fatal error
```

alternatively we could reset the counter every (say) 10 seconds.

davidrohr:
why do we need this? We can set the infologger-threshold with the `--infologger-severity` command line option?
On the EPNs we have `--infologger-severity important`. If this is due to the complaint about too many messages on the FLPs, I would propose we use the same setting in the FLP workflows.

I don't think it makes much sense to filter in FLPs and EPNs in different ways.

I mean, I am not principally against changing the levels, but I would not do it in a way that they are generally suppressed.

davidrohr:
and I do not agree in throttling important, alarm, and error severities.
They should not spam, in case they do it should be fixed in the code, but not on InfoLogger level.

ktf:
I think the message I got is that the OPS level should really be left for messages targeted to shifters, with clear indications of what should be the action to be performed. E.g. "Call PDP on call". This was an attempt to reduce the verbosity, but I agree, it's not actually what is needed. Let's discuss in the SRC meeting as you proposed.

PR #10249, opened 2022-11-10
IDCs: Disable offset for validity range when storing IDCs in the CCDB

matthias-kleiner:
- GainMap using tracks: Store also the RMS map in the CCDB


wiechula (approved):

matthias-kleiner:
can this be merged? The error seems unrelated

martenole:
Failure unrelated, merging

PR #10250, opened 2022-11-10
GPU: Fix compilation issue with HIP+LOG(): workaround

mconcas:
@davidrohr 
I restricted the issue down to two similar offending lines `LOGP(warning, "Anomalous track, target X:{:f}", xk);`. 
Using `LOG(...)` works that around. Probably the underlying format causes this odd behaviour with compiler.
I did not investigate very much the details, my guess is that only the methods that have the lines are used on GPU, and that's why it fails at linking stage.
Otherwise I would not explain why all the other calls across the `TrackParametrization*.*` files don't raise this issue.

There are some leftover formatting that my editor spotted and fixed, hope it does not bother.

NB: I just checked that it compiles and my printouts worked, but this might have some impact on GPU tracking. Worth to check, maybe.

### DIFF
--- a/DataFormats/Reconstruction/src/TrackParametrization.cxx
+++ b/DataFormats/Reconstruction/src/TrackParametrization.cxx
@@ -216,7 +216,7 @@ GPUd() bool TrackParametrization<value_T>::propagateParamTo(value_t xk, const di
   }
   // Do not propagate tracks outside the ALICE detector
   if (math_utils::detail::abs<value_T>(dx) > 1e5 || math_utils::detail::abs<value_T>(getY()) > 1e5 || math_utils::detail::abs<value_T>(getZ()) > 1e5) {
-    LOGP(warning, "Anomalous track, target X:{:f}", xk);
+    LOG(warning) << "Anomalous track, traget X:" << xk;
     return false;
   }
   value_t crv = getCurvature(b[2]);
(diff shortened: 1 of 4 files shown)


davidrohr:
@mconcas : Could you rebase onto https://github.com/AliceO2Group/AliceO2/pull/10252 (which removes some more __HIPCC__ workarounds than you do).

Also, could you then do it as 2 commits, first the revert of the old workaround as 1 commit, and then your workaround on top, so once it is fixed in ROCm we can easily revert only this one.

In general, it would have been good to really create a reproduce for AMD to fix it, but ok, I think your workaround is certainly an improvement, since the logging will at least work again. But we still should follow up that the root cause is fixed.

mconcas:
> Also, could you then do it as 2 commits, first the revert of the old workaround as 1 commit, and then your workaround on top, so once it is fixed in ROCm we can easily revert only this one.

I certainly will do it.
But I don't think it will be necessary to revert my changes, if you mean the actual `LOGP` -> `LOG` changes. They give the same output, just using a different mechanism.

davidrohr:
ok, true, we don't need to revert your changes since you just switched to a different LOG macro. But anyway, I'd separate undoing the old workaround and your changes. And also I'd like to try to revert all of the previous workaround.

mconcas:
> ok, true, we don't need to revert your changes since you just switched to a different LOG macro. But anyway, I'd separate undoing the old workaround and your changes. And also I'd like to try to revert all of the previous workaround.

rebase is done, let me know in case it is not what you meant.

PR #10251, opened 2022-11-10
multiple fixes for barrel alignment

shahor02:


shahor02:
@timo @ktf the licence checker complains about incorrect/missing licence header but does not tell the filename. I see that it writes these filenames to $GITHUB_STEP_SUMMARY but it is not clear how to retrieve this file. Any hint?

aalkin:
@shahor02 the issue is shown inline in `Files changed`, see `Check failure on line 10 in Detectors/Align/macro/algconf.C`. However it is a macro, so it should not be checked, I assume?

PR #10256, opened 2022-11-11
add helper to get nominal field

jgrosseo:


jgrosseo:
@shahor02 This is the same as in GRPObject.h. Do you have objections?

shahor02 (approved):

PR #10257, opened 2022-11-11
TRD raw reader skips links with corrupt digit HC headers

martenole:


bazinski:
This breaks backwards compatibility for byteswapping

martenole:
But the byte swapping was not only in the digit HC headers, was it? In which case it's broken anyways

bazinski:
I went back, it was taken out in #PR9647 not sure why this one stayed, but do we care about that backwards compatibility ?

martenole:
I thought the data for which this was required is anyhow not used anymore or has already been deleted even. Or did I misunderstand something?

bazinski:
Data is still there on tape somewhere, the only thing that *might* be of consequence is Kr but we redid Kr so not sure if we care. I *think* only the Kr raw data was stored, and a noise run, but again we have redone the noise run I think 3 times now, maybe more.

martenole:
a noise run we can always repeat. for Kr I am not sure whether it needs to be reprocessed. maybe @tdietel has an opinion on this? if yes, then I would need some data of that run and then I could add the byte swapping back.
Since at the moment its anyways not working it is better to remove it completely I think

bazinski:
If we ignore the byteswapping, we can lose the --fixforoldtrigger option as well which was fixed, I thinked a week prior to pilot beam, or there abouts.

martenole:
Closing, since superseded by https://github.com/AliceO2Group/AliceO2/pull/10278

PR #10261, opened 2022-11-13
Provisionary fix for ITS fit with B=0

shahor02:
Hi @mpuccio @davidrohr 

The ITS tracking with B=0 produces a weird q/pT spectrum instead of fixed (modulo e.loss) "most probable" q2pt: 

itsQ2Pt

The reason is that in the https://github.com/AliceO2Group/AliceO2/blob/d8699251d53b85f8a6fb2d175116065b3a04007c/Detectors/ITSMFT/ITS/tracking/src/TrackerTraits.cxx#L913 you are initializing the cov.matrix with non-0 non-diagonal elements while with B=0 all correlations with q/pT must be 0. 
This PR attempts to fix this, but I don't understand the way you build the cov. matrix in general: should not it correspond to the matrix updated only at `cluster3`, i.e. only diag. errors in Y and Z set to resolution^2, other diagonal errors "as large as possible" and non-diagonal ones set to 0? 
With this fix the result of the fit becomes:
itsQ2Pt_new
and the final covariances of q/pT with other parameters are 0 within `~10^-7`

@davidrohr, for the TPC I see something different: the mean q/pt corresponds to conventional "most probable" but then there are huge tails on both sides
tpcQ2Pt
and also I see a significant correlation between the q/pT error and e.g. y, this cannot happen with B=0.
tpcCorrYPt

Could you please check if you are not initializing with the covariance with some non-diagonal elements? 

It seems that currently all fits with B=0 are wrong.

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/src/TrackerTraits.cxx
+++ b/Detectors/ITSMFT/ITS/tracking/src/TrackerTraits.cxx
@@ -912,7 +912,8 @@ bool TrackerTraits::trackFollowing(TrackITSExt* track, int rof, bool outward, co
 /// a TrackParCov
 track::TrackParCov TrackerTraits::buildTrackSeed(const Cluster& cluster1, const Cluster& cluster2, const Cluster& cluster3, const TrackingFrameInfo& tf3, float resolution)
 {
-  const float ca = std::cos(tf3.alphaTrackingFrame), sa = std::sin(tf3.alphaTrackingFrame);
+  float ca, sa;
+  o2::math_utils::sincos(tf3.alphaTrackingFrame, sa, ca);
   const float x1 = cluster1.xCoordinate * ca + cluster1.yCoordinate * sa;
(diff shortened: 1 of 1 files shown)


mpuccio:
Hi @shahor02, it looks good to me. I copied the code from the Cooked matrix long ago, so I cannot exclude that I did some mistakes in the process. However, we can even remove this part of the code and do a full fit of the candidate by providing a reasonable set of parameters and a very large covariance as starting conditions.

shahor02:
@mpuccio starting from large diagonal cov matrix would be the right choice later you fit through the all clusters, but I was not sure if you do this: I see that you make a piecewise fit https://github.com/AliceO2Group/AliceO2/blob/4d497280a90190efc2a16e48924658a20360bc9e/Detectors/ITSMFT/ITS/tracking/src/TrackerTraits.cxx#L537-L544
Does it fit through all clusters contiguously, including the `cluster3` ?

shahor02:
Superseded by https://github.com/AliceO2Group/AliceO2/pull/10271

PR #10263, opened 2022-11-13
Rename nThreads option for TPC encoder

martenole:
To avoid clash with the MFT reco workflow option with the same name


martenole:
Confirmed that this fixes the observed issue and the change is trivial. So merging now in order to include it in a new build

PR #10264, opened 2022-11-13
Don't truncate float calorimeter energies to int in EventDisplay

pnwkw:


EventVisualisation/View/src/EventManager.cxx
```diff
@@ -330,7 +330,7 @@ void EventManager::displayCalorimeters(VisualisationEvent& event, const std::str
         return std::hash<float>()(pair.first + 1000.0 * pair.second);
       }
     };
-    std::unordered_map<std::pair<float, float>, int, pair_hash> map; // sum up entries for the same tower
+    std::unordered_map<std::pair<float, float>, float, pair_hash> map; // sum up entries for the same tower
     for (const auto& calo : event.getCalorimetersSpan()) {
       auto key = std::make_pair(calo.getEta(), calo.getPhi());
       map.try_emplace(key, 0);
```
shahor02, line 336: why not simply `map[std::pair<float,float>{calo.getEta(), calo.getPhi()}] += calo.getEnergy();` ? Currently you have 3 map searches instead of 1.

shahor02 (approved):
merging as trivial, but please check the comment below.

PR #10267, opened 2022-11-14
Optimize calorimeter energy aggregation in EventDisplay

pnwkw:


EventVisualisation/View/src/EventManager.cxx
```diff
@@ -332,9 +330,7 @@ void EventManager::displayCalorimeters(VisualisationEvent& event, const std::str
     };
     std::unordered_map<std::pair<float, float>, float, pair_hash> map; // sum up entries for the same tower
     for (const auto& calo : event.getCalorimetersSpan()) {
-      auto key = std::make_pair(calo.getEta(), calo.getPhi());
-      map.try_emplace(key, 0);
-      map[key] = map[key] + calo.getEnergy();
+      map[std::make_pair(calo.getEta(), calo.getPhi())] += calo.getEnergy();
```
sawenzel, line 333: is this guaranteed to be correct for uninitialized/new keys? 

(besides, I think that the name "map" as a variable name might be misleading as there is a type std::map in C++ which is often abbreviated as "map" in code).

sawenzel: @pnwkw

davidrohr: std::map [] operator: `Returns a reference to the value that is mapped to a key equivalent to key, performing an insertion if such key does not already exist. `

`map` should indeed be renamed, but it was there before, and the compiler would catch such an error.

PR #10269, opened 2022-11-14
IDC deltas always on and default scaling for tracks for TimeGain calib

chiarazampolli:
@davidrohr , @martenole , @matthias-kleiner , @rmunzer , @wiechula

### DIFF
--- a/prodtests/full-system-test/aggregator-workflow.sh
+++ b/prodtests/full-system-test/aggregator-workflow.sh
@@ -219,8 +219,8 @@ nTFs=1000
 nTFs_SAC=1000
 nBuffer=100
 IDC_DELTA="--disable-IDCDelta true" # off by default
-# we switch on deltas if explicitly requested; in PbPb it is on by default, unless we switch it off;
-if [[ "0$ENABLE_IDC_DELTA" == "01" ]] || [[ $BEAMTYPE == "PbPb" && "0$ENABLE_IDC_DELTA" != "00" ]]; then IDC_DELTA=""; fi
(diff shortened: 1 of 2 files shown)


davidrohr:
As we have changed the defaults for the scaling meanwhile, perhaps we should also put them here.

chiarazampolli:
Yes, I am waiting for @rmunzer  to confirm that the ones he used lately are fine.

martenole:
But in general we would not need this now for the PbPb test, right? As we can do all the changes with env vars

davidrohr:
Yes, and anyway we might need different settings in Pb-Pb than in pp.

chiarazampolli:
> But in general we would not need this now for the PbPb test, right? As we can do all the changes with env vars

Yes.

chiarazampolli:
Hello @rmunzer , @wiechula , 
What should be the default after the tests? What about PbPb?
Thanks!

wiechula:
@chiarazampolli , I suggest to use for now the defaults we tested:
```
SCALEEVENTS_TPC_TIMEGAIN=10
SCALETRACKS_TPC_TIMEGAIN=1000
```
They should also be fine for low rate PbPb

chiarazampolli:
Hi @wiechula , 
Done. On Friday, the variables will have to be passed from command line, as there is no update of O2 foreseen. Then we can use synthetic to tune for high IR PbPb. 
If there are no further comments, @davidrohr , can this be reviewed?
Chiara

davidrohr (approved):

PR #10270, opened 2022-11-14
Fix: TRD CTF decoder --correct-trd-trigger-offset had no effect

shahor02:


Detectors/TRD/reconstruction/include/TRDReconstruction/CTFCoder.h
```diff
@@ -224,7 +224,7 @@ o2::ctf::CTFIOSize CTFCoder::decode(const CTF::base& ec, VTRG& trigVec, VTRK& tr
   uint32_t trkCount = 0, digCount = 0, adcCount = 0;
   uint32_t orbit = header.firstOrbit, orbitPrev = 0, orbitPrevGood = mFirstTFOrbit;
   uint16_t bc = header.firstBC;
-  bool checkIROK = (mBCShift == 0); // need to check if CTP offset correction does not make the local time negative ?
+  bool checkIROK = (mBCShift != 0); // need to check if CTP offset correction does not make the local time negative ?
```
martenole: Should this not stay as it was? In case here a BC shift is set, then it would not check anymore the condition below `ir.differenceInBC({0, mFirstTFOrbit}) >= mBCShift`, right?

shahor02: Right, I've overdone the fix, thanks for spotting!

shahor02:

PR #10271, opened 2022-11-14
Fit the full candidate with KF

mpuccio:
Dear @shahor02,

this is an alternative fix to what you proposed. However, with this I get slightly less good tracks in the MC but a slightly better good/fake ratio.
In the data, I get a bit more tracks at lower momenta and a bit less tracks at higher momenta.
In case you need it fast, I would propose to merge your PR. I would need tomorrow at least to understand why we do get such a difference.

Cheers,
Max﻿

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/src/TrackerTraits.cxx
+++ b/Detectors/ITSMFT/ITS/tracking/src/TrackerTraits.cxx
@@ -516,19 +516,17 @@ void TrackerTraits::findTracks()
       }
     }
 
-    /// Track seed preparation. Clusters are numbered progressively from the outermost to the innermost.
-    const auto& cluster1_glo = mTimeFrame->getUnsortedClusters()[lastCellLevel + 2].at(clusters[lastCellLevel + 2]);
+    /// Track seed preparation. Clusters are numbered progressively from the innermost going outward.
+    const auto& cluster1_glo = mTimeFrame->getUnsortedClusters()[lastCellLevel].at(clusters[lastCellLevel]);
     const auto& cluster2_glo = mTimeFrame->getUnsortedClusters()[lastCellLevel + 1].at(clusters[lastCellLevel + 1]);
-    const auto& cluster3_glo = mTimeFrame->getUnsortedClusters()[lastCellLevel].at(clusters[lastCellLevel]);
+    const auto& cluster3_glo = mTimeFrame->getUnsortedClusters()[lastCellLevel + 2].at(clusters[lastCellLevel + 2]);
+    const auto& cluster3_tf = mTimeFrame->getTrackingFrameInfoOnLayer(lastCellLevel + 2).at(clusters[lastCellLevel + 2]);
 
-    const auto& cluster3_tf = mTimeFrame->getTrackingFrameInfoOnLayer(lastCellLevel).at(clusters[lastCellLevel]);
-
-    /// FIXME!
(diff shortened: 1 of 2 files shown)


mpuccio:
Never mind, I found the issue: I did not swap the order of the clusters when computing the radii and the centres of the circles. Now in the data, I obtain the following:
image

Detectors/ITSMFT/ITS/tracking/src/TrackerTraits.cxx
```diff
@@ -923,27 +925,20 @@ track::TrackParCov TrackerTraits::buildTrackSeed(const Cluster& cluster1, const
...
-  float cy = 1.e15f;
-  if (std::abs(getBz()) > o2::constants::math::Almost0) {
-    cy = (math_utils::computeCurvature(x1, y1, x2, y2 + resolution, x3, y3) - crv) /
-         (resolution * getBz() * o2::constants::math::B2C) *
-         20.f; // FIXME: MS contribution to the cov[14] (*20 added)
-  }
-  const float s2 = resolution;
-
   return track::TrackParCov(tf3.xTrackingFrame, tf3.alphaTrackingFrame,
                             {y3, z3, crv * (x3 - x0), 0.5f * (tgl12 + tgl23),
```
shahor02: Why `crv` is still used with B=0?
Could you comment the reason of the clusters permutations above? Most important: before you were setting tight errors in the seed cov.matrix, also to slopes, which makes me to think that at least 2 updates were accounted already on the seed level. Now they are relaxed, is this related to the permutations above?

shahor02 (commented):
thanks @mpuccio ! I can wait a couple of days, anyway need to sort out some other problems before I start a new reco. So, better to have a tested code. See question below.

Detectors/ITSMFT/ITS/tracking/src/TrackerTraits.cxx
shahor02: Got your update after my comment. I assume the plot is for B!=0 ? Do you have it for B=0?

mpuccio: Hi @shahor02, here is the same plot before and after this PR as it is now for B=0.
image
In the old code, two-track updates were approximated, that was the Cooked Matrix initialisation by @iouribelikov.
The permutation of the clusters is done just to keep the same signature of the function, that in the future might also disappear.

shahor02: ok, thanks, looks good, just see my question above about using crv for snp estimate also in case of B=0.

mpuccio: Right, I can also avoid it for B=0. Is `DeltaY / std::hypot(DeltaX, DeltaY)` enough for the sinPhi?

shahor02: it would be more precise:
```
auto tgp = std::atan2(DeltaY, DeltaX);
snp = tgp / std::sqrt(1 + tgp*tgp);
```
BTW, I guess one should add a fix for the GPU as well?

mpuccio:
Hi @shahor02, thanks a lot. I removed the curvature in the B=0 case and applied your formula (with mine, it was not working). For the GPU-related fix, we ported the fit to GPU with David long ago, but now that part is not enabled.

PR #10275, opened 2022-11-15
Move MC stack to DetectorsBase

sawenzel:
This is a restructuring commit to
reduce library coupling and to avoid cycling dependencies for future developments:

a) The VMC stack implementation belongs more logically
   to Detectors/Base as an implementation of a Virtual Monte
   Carlo functionality and less as a fundamental O2 simulation data layout

b) remove some uneccessary includes


sawenzel:
Merging since non-functional change which is required for upcoming urgent PRs.

PR #10278, opened 2022-11-15
Some fixes and small improvements for TRD

martenole:


DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
```diff
@@ -89,6 +89,7 @@ class Tracklet64
   GPUd() int getFormat() const { return ((mtrackletWord & formatmask) >> formatbs); };     // no units 0..15
   GPUd() int getHCID() const { return ((mtrackletWord & hcidmask) >> hcidbs); };           // no units 0..1079
   GPUd() int getPadRow() const { return ((mtrackletWord & padrowmask) >> padrowbs); };     // pad row number [0..15]
+  GPUd() float getPadCol() const;                                                          // estimate the pad column number from the tracklet offset (can be off by +-1 pad)
```
f3sch: I would have expected a function like this to output an integer number like getPadRow(),
Is there a reason for this that I do not get?

martenole: Ah right, the `float` was just in there for testing, I forgot to take it out in the end. Thanks for spotting

f3sch (approved):
Cannot comment on all the Digit and GPU settings, but looks ok.

PR #10285, opened 2022-11-16
Fix MFT tracker for B=0 

rpezzi:
This PR fix a bug introduced by the multithreaded MFT tracker which produces no tracks for B=0 reconstruction. 
This commit also ensures Linear track model is used when B=0

I hijack this PR to also disable MFT Tracker IR filter for cosmic runs during sync reco


rpezzi:
Hi. FullCI tests passed. Failures on `build/O2/o2-cs8` seems unrelated.

PR #10287, opened 2022-11-16
[FOCAL-9] Write decoded ASIC header to tree

mfasDa:
Instead of writing header word write header information
(header, BC, counters, trailer) separately.


mfasDa:
CI error unrelated

shahor02 (approved):

PR #10288, opened 2022-11-16
Make HepMC and native status available for MC gen

benedikt-voelkel:
* HepMC and Gen status codes handled/entangled via Union as one single integer

* common functionality to entangle/disentangle full encoding

* be backwards-compatible wrt to HepMC status code which previously was used to decide whether or not primaries should be transported

* set status code correctly for generator form kinematics

### DIFF
--- a/Generators/src/GeneratorFromFile.cxx
+++ b/Generators/src/GeneratorFromFile.cxx
@@ -13,6 +13,7 @@
 #include "Generators/GeneratorFromO2KineParam.h"
 #include "SimulationDataFormat/MCTrack.h"
 #include "SimulationDataFormat/MCEventHeader.h"
+#include "SimulationDataFormat/MCGenStatus.h"
 #include <fairlogger/Logger.h>
 #include <FairPrimaryGenerator.h>
 #include <TBranch.h>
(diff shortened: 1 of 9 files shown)


Generators/include/Generators/Generator.h
```diff
@@ -17,6 +17,7 @@
 #include "FairGenerator.h"
 #include "TParticle.h"
 #include "Generators/Trigger.h"
+#include "SimulationDataFormat/MCStatus.h"
```
sawenzel: probably don't need to include that here but rather in the source file

Generators/src/GeneratorFromFile.cxx
```diff
@@ -245,8 +246,8 @@ bool GeneratorFromO2Kine::importParticles()
 
       LOG(debug) << "Putting primary " << pdg;
 
-      mParticles.push_back(TParticle(pdg, wanttracking, m1, m2, d1, d2, px, py, pz, e, vx, vy, vz, vt));
-      mParticles.back().SetUniqueID((unsigned int)t.getProcess()); // we should propagate the process ID
```
sawenzel, line 249: what about the physics process?

benedikt-voelkel: that was kicked out by accident, back in now

benedikt-voelkel:
Taking 3 bits now instead of 2 for the `MCGenStatusEncoding::isEncoded`

sawenzel (approved):

benedikt-voelkel:
It seems the Mac CIs are unrelated. I see them in other CIs of PRs as well

jgrosseo:
So you decided against using the flag member to flag this encoding? 

How you assure that the 3 "id bits" are not set by accident?

benedikt-voelkel:
We need the backward-compatibility only when generating from "previous" MCKine files or when reading from "previous" AODs.

At the moment we choose the 3 leftmost bits of the status code to make it more unlikely that they are set in the way we want by accident.

To ensure this strictly, we would need
1. a flag in `MCTrack`
1. extend the enum/flag for AOD

Do we need 100% ensured backward-compatibility? We could have that in for a certain transition phase.

I could also contact the analyses that use the status (there are not many as I mentioned in the email from yesterday) and x-check with them how crucial it is for them to decide whether we need it.

benedikt-voelkel:
@jgrosseo indeed, the HepMC status codes are positive. So there can't be a previous status code with the leftmost bits set.

jgrosseo:
OK!

PR #10292, opened 2022-11-16
Use LOG(alarm) instead of std::cerr

dstocco:


dstocco:
@davidrohr @martenole : this should solve the issue with the info logger.

davidrohr (approved):

ktf:
macOS error unrelated. Merging.

PR #10293, opened 2022-11-16
Allow to set the MID HV from CCDB

dstocco:
Once this is approved, I will use it to upload the default object in the CCDB and only after that I will do a second PR to systematically recover the values from the CCDB in the digitisation.

### DIFF
--- a/Detectors/MUON/Common/src/dcs-ccdb.cxx
+++ b/Detectors/MUON/Common/src/dcs-ccdb.cxx
@@ -12,6 +12,8 @@
 #include "CCDB/CcdbApi.h"
 #include "DetectorsDCS/DataPointIdentifier.h"
 #include "DetectorsDCS/DataPointValue.h"
+#include "DetectorsDCS/DataPointCreator.h"
+#include "DetectorsDCS/DataPointCompositeObject.h"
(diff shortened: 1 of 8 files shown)


Detectors/MUON/Common/src/dcs-ccdb.cxx
```diff
@@ -170,13 +172,42 @@ void makeCCDBEntryForDCS(const std::string ccdbUrl, uint64_t timestamp)
...
 
   api.storeAsTFileAny(&dpid2DataDesc, CcdbDpConfName(), md, timestamp, o2::ccdb::CcdbObjectInfo::INFINITE_TIMESTAMP);
 }
 
+void makeDefaultCCDBEntry(const std::string ccdbUrl)
+{
+  DPMAP dpMap;
+  std::string ccdb = fmt::format("{}/Calib/HV", o2::muon::subsysname());
+#if defined(MUON_SUBSYSTEM_MCH)
+  auto aliases = o2::mch::dcs::aliases();
```
aphecetche, line 187: note to self :  will have to implement this part for MCH in another PR....

aphecetche (approved):

PR #10299, opened 2022-11-16
Fix mask of MFTMCHMID

shahor02:
`GlobalTrackID::MFTMCHMID` was inserted between `TPCTRDTOF` and `ITSTPCTRDTOF` while the related masks were defined in the `TPCTRDTOF, ITSTPCTRDTOF, MFTMCHMID` order. Fortunately, the masks are not stored in the persistent data so we can simply swap them (and `MFTMCHMID` is not used though we should keep it at its current position for backward compatibility).


shahor02:
trivial, merging

PR #10300, opened 2022-11-16
TRD: QC: Add global track id to TrackQC

f3sch:
This patch allows for offline extraction of additional TOF PID. And will allow a cleaner cuts for the TRD PID model building.

As @martenole already mentioned in #10229:
> Well it would be just adding one more integer to the QC struct but as you wish ;)

It turned out that adding GlobalTrackId suffices for matching to TOF tracks...

### DIFF
--- a/Detectors/TRD/qc/include/TRDQC/Tracking.h
+++ b/Detectors/TRD/qc/include/TRDQC/Tracking.h
@@ -21,6 +21,7 @@
 #include "DataFormatsTRD/CalibratedTracklet.h"
 #include "DataFormatsTRD/Constants.h"
 #include "ReconstructionDataFormats/TrackTPCITS.h"
+#include "ReconstructionDataFormats/GlobalTrackID.h"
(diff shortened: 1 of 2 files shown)


Detectors/TRD/qc/include/TRDQC/Tracking.h
```diff
@@ -44,6 +44,7 @@ namespace trd
 
 struct TrackQC {
   int type;                          ///< 0 TPC-TRD track; 1 ITS-TPC-TRD track
+  unsigned int refGlobalTrackId;     ///< raw GlobalTrackID of the seeding track (either ITS-TPC or TPC)
```
martenole: why are you not storing it directly as `o2::dataformats::GlobalTrackID`?

f3sch: No particular reason actually..

martenole (approved):

PR #10306, opened 2022-11-17
CollisionContext: Option to ensure that a timeframe has at least one …

sawenzel:
…collision

### DIFF
--- a/Steer/src/CollisionContextTool.cxx
+++ b/Steer/src/CollisionContextTool.cxx
@@ -41,6 +41,7 @@ struct Options {
   int tfid = 0;          // tfid -> used to calculate start orbit for collisions
   int orbitsPerTF = 256; // number of orbits per timeframe --> used to calculate start orbit for collisions
   bool useexistingkinematics = false;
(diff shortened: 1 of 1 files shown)


Steer/src/CollisionContextTool.cxx
```diff
@@ -245,12 +247,15 @@ int main(int argc, char* argv[])
...
       o2::InteractionTimeRecord record;
-      int count = 0;
+      // this loop makes sure that we have at least one collision within the number of orbits asked (if noEmptyTF is enabled)
       do {
+        sampler.setFirstIR(o2::InteractionRecord(0, orbitstart));
+        sampler.init();
         record = sampler.generateCollisionTime();
+      } while (options.noEmptyTF && usetimeframelength && record.orbit >= orbitstart + options.orbits);
+      int count = 1;
+      do {
```
davidrohr, line 258: not fully sure this logic is correct. What if we ask for 0 collisions in the TF? Which was probably not working before anyway. But if `mcnumberasked = 1`, it will generate 2 right now, doesn't it?

sawenzel: no ... since this only applies when I am in "usetimeframelength" mode. In this case, the number of events should be determined dynamically and mcnumberasked neglected.

PR #10307, opened 2022-11-17
TRD: QC: Move root output

f3sch:
This patch allows to reproduce the actual online calibration which runs every 15 minutes. Every entry in the output tree represents a slot and contains the residual profiles and fit results.

Also I added the update for the configKeyValues, which I previously forgot...

### DIFF
--- a/Detectors/TRD/calibration/src/CalibratorVdExB.cxx
+++ b/Detectors/TRD/calibration/src/CalibratorVdExB.cxx
@@ -126,6 +126,20 @@ void CalibratorVdExB::initProcessing()
   opt.SetTolerance(.001);
   mFitter.Config().SetMinimizerOptions(opt);
 
+  // set tree addresses
+  if (mEnableOutput && mOutTree) {
+    mOutTree->Branch("lorentzAngle", &mFitFunctor.laPreCorr);
+    mOutTree->Branch("vDrift", &mFitFunctor.vdPreCorr);
+    for (int iDet = 0; iDet < MAXCHAMBER; ++iDet) {
+      mOutTree->Branch(fmt::format("residuals_{:d}", iDet).c_str(), mFitFunctor.profiles[iDet].get());
+    }
+  }
+
+  // get parameters
+  auto& params = TRDCalibParams::Instance();
+  mMinEntriesChamber = params.minEntriesChamber;
+  mMinEntriesTotal = params.minEntriesTotal;
+
   mInitDone = true;
 }
 
@@ -155,10 +169,6 @@ void CalibratorVdExB::retrievePrev(o2::framework::ProcessingContext& pc)
       mFitFunctor.laPreCorr[iDet] = dataCalVdriftExB->getExB(iDet);
       mFitFunctor.vdPreCorr[iDet] = dataCalVdriftExB->getVdrift(iDet);
     }
-
-    auto& params = TRDCalibParams::Instance();
-    mMinEntriesChamber = params.minEntriesChamber;
-    mMinEntriesTotal = params.minEntriesTotal;
   }
 }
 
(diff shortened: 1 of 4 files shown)


Detectors/TRD/workflow/include/TRDWorkflow/VdAndExBCalibSpec.h
```diff
@@ -76,11 +88,20 @@ class VdAndExBCalibDevice : public o2::framework::Task
     LOG(info) << "Finalizing calibration";
     mCalibrator->checkSlotsToFinalize(o2::calibration::INFINITE_TF);
     sendOutput(ec.outputs());
+    if (mOutFile && mOutTree) {
+      LOG(info) << "Writing calibration to file";
+      mOutFile->cd();
+      mOutTree->Write();
```
f3sch: I am not sure if there is a better way to do this...

Detectors/TRD/workflow/src/trd-calib-workflow.cxx
```diff
@@ -37,6 +38,7 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
 WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
 {
   auto enableRootInp = configcontext.options().get<bool>("enable-root-input");
+  o2::conf::ConfigurableParam::updateFromString(configcontext.options().get<std::string>("configKeyValues"));
```
f3sch, line 40: I only tested the configurable params with the tracking workflow which already called this

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -126,6 +126,18 @@ void CalibratorVdExB::initProcessing()
...
+  if (mOutTree) {
+    mOutTree->Branch("lorentzAngle", &mFitFunctor.laPreCorr);
+    mOutTree->Branch("vDrift", &mFitFunctor.vdPreCorr);
+    //mOutTree->Branch("residuals", &mFitFunctor.profiles);
+  }
+
+  // Retrieve configuration values
+  auto& params = TRDCalibParams::Instance();
+  mMinEntriesChamber = params.minEntriesChamber;
+  mMinEntriesTotal = params.minEntriesTotal;
```
f3sch: I initial put this in the 'wrong' function. Programmatically it did matter and the result would have been the same.
But putting this here is saner.

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
```diff
@@ -126,6 +126,18 @@ void CalibratorVdExB::initProcessing()
   opt.SetTolerance(.001);
   mFitter.Config().SetMinimizerOptions(opt);
 
+  // Set tree branches
+  if (mOutTree) {
+    mOutTree->Branch("lorentzAngle", &mFitFunctor.laPreCorr);
+    mOutTree->Branch("vDrift", &mFitFunctor.vdPreCorr);
+    //mOutTree->Branch("residuals", &mFitFunctor.profiles);
```
f3sch: ROOT does not support writing this _yet_.
Any idea how to get this in the tree without making a completely new object?

Detectors/TRD/workflow/include/TRDWorkflow/VdAndExBCalibSpec.h
```diff
@@ -47,7 +50,16 @@ class VdAndExBCalibDevice : public o2::framework::Task
     auto enableOutput = ic.options().get<bool>("enable-root-output");
     auto slotL = ic.options().get<uint32_t>("sec-per-slot");
     auto delay = ic.options().get<uint32_t>("max-delay");
-    mCalibrator = std::make_unique<o2::trd::CalibratorVdExB>(enableOutput);
+    if (enableOutput) {
```
martenole: I would prefer to move this block completely into `CalibratorVdExB.cxx`. And here inside the spec you could just call a function e.g. `mCalibrator->createOutputFile()` in case ROOT output is enabled. Then you could also leave the default constructor as it is now.
Inside the `CalibratorVdExB` you could then also switch to smart pointers which are preferred: `mOutFile = std::make_unique<TFile>("filename", "recreate")`

Detectors/TRD/workflow/include/TRDWorkflow/VdAndExBCalibSpec.h
martenole: similar here just a call to `mCalibrator->closeOutputFile()`. Besides that there is nothing wrong with the way you are writing

Detectors/TRD/workflow/src/trd-calib-workflow.cxx
martenole: no problem, happens

Detectors/TRD/calibration/src/CalibratorVdExB.cxx
martenole: Hm, `mFitFunctor.profiles` is an array of unique pointers to TProfile objects. I don't think that ROOT supports this. Maybe it would work with a loop, adding one branch per TProfile. Something like `mOutTree->Branch(fmt::format("residuals_{}", iDet), &(mFitFunctor.profiles[iDet]));`

martenole (approved):
Thanks! Will merge once the CI is through

PR #10308, opened 2022-11-17
DPL: fix a few trivial warnings in the HistogramRegistry

ktf:


ktf:
This actually accounts for quite some copies on startup.

PR #10309, opened 2022-11-17
HistogramRegistry: avoid unneded copies and allocations

ktf:


ktf:
Given the map does not change, I am not sure what is the need for it. A switch case is probably much cleaner / better.

PR #10312, opened 2022-11-18
Fix order of run type names in ECSDataAdapters

pnwkw:


shahor02:
Already was fixed by https://github.com/AliceO2Group/AliceO2/pull/10314

PR #10317, opened 2022-11-18
ctpdev: scalers reply

lietava:


lietava:
@shahor02 : can you, please, pull this ?

shahor02 (approved):

PR #10327, opened 2022-11-19
fix in TOF matching (time) to account for integrated times

noferini:
This is a fix for the time window used to match tracks.
- track time errors are now get from createTracksVariadic for all kinds of tracks
- track length are used (beta=1) to shift track time to TOF radius (more efficient than do the opposite)
- 100 ns tolerance added on the maxTrkTime to avoid to cut on slow tracks (relevant for *-TRD tracks)
I would expect to recover almost 50% efficiency in LHC22m/... for ITS-TPC-* tracks because we should recover all ITS-TPC-TRD tracks (now almost complety cut)
Tested on Pb-Pb data (full reco output) by re-running TOF matching

@chiarazampolli , @njacazio , @ercolessi

### DIFF
--- a/Detectors/GlobalTracking/src/MatchTOF.cxx
+++ b/Detectors/GlobalTracking/src/MatchTOF.cxx
@@ -210,13 +210,13 @@ bool MatchTOF::prepareTPCData()
       if (std::abs(trk.getQ2Pt()) > mMaxInvPt) {
         return true;
       }
-      this->addTPCSeed(trk, gid);
+      this->addTPCSeed(trk, gid, time0, terr);
     }
     if constexpr (isTPCITSTrack<decltype(trk)>()) {
       if (trk.getParamOut().getX() < o2::constants::geom::XTPCOuterRef - 1.) {
         return true;
       }
-      this->addITSTPCSeed(trk, gid);
+      this->addITSTPCSeed(trk, gid, time0, terr);
     }
     if constexpr (isTRDTrack<decltype(trk)>()) {
       this->addTRDSeed(trk, gid, time0, terr);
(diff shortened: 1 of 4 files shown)


noferini:
Hi @shahor02 ,
for TPC unconstrained tracks I restored the usage of track_time&_err from TPC track since it seems to me that the values from crateTracksVariadic don't reproduce the proper behavior.

In any case for the track_time I need to use the value stored in the track (not the central value between min_time and max_time) since it is directly connected to the Z coordinate -> used to match TOF simultaneously in space&time.

shahor02:
Hi @noferini 

The time/error provided by createTracksVariadic for the TPC is just a way to get a time interval for fast TOF clusters selection w/o going on TPC track specific data members. Sure, to explore time vs Z dependence one should use TrackTPC.getTime0()

noferini:
o2-cs8 check failing unrelated to this PR
"c++: fatal error: Killed signal terminated program cc1plus" in O2Physics compilation

noferini:
Hi @shahor02 ,
can we merge this PR?
It would be nice to have the change available for the next reconstructions

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -496,12 +507,14 @@ void MatchTOF::doMatching(int sec)
     auto& trackWork = mTracksWork[type][cacheTrk[itrk]];
     auto& trefTrk = trackWork.first;
     auto& intLT = mLTinfos[type][cacheTrk[itrk]];
+    float timeShift = intLT.getL() * 33.35641; // integrated time for 0.75 beta particles in ps, to take into account the t.o.f. delay with respect the interaction BC
```
shahor02, line 510: in principle, you could use `intLT.getTOF(2)` for pion, or better directly `intLT.getTOF(_tr.getPID())` (which at the moment will assume pion anyways.

shahor02 (approved):
Hi @noferini 
Merging, but please see 1 comment below.

PR #10330, opened 2022-11-21
DPL: set TimingInfo field accordingly on new Run

ktf:


ktf:
@shahor02 @chiarazampolli as we discussed

PR #10331, opened 2022-11-21
TRD fix digit phase comments

bazinski:
Correct inconsistent comments in Digit class.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/Digit.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/Digit.h
@@ -41,7 +41,7 @@ using ArrayADC = std::array<ADC_t, constants::TIMEBINS>;
 //    Pre Trigger phase:
 //        LHC clock runs at 40.08MHz, ADC run at 1/4 or 10MHz and the trap runs at 120MHz or LHC*3.
 //        The trap clock can therefore be in 1 of 12 positions relative to the ADC clock.
-//        Only 4 of those positions are valid, hence there are 4 values for pre-trigger phase: 0,3,6,9.
+//        Only 4 of those positions are valid, dependent on the TRAP assembly program timing offset, as of 11/2022 pre-trigger phase: 0,3,6,9.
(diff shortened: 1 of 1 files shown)


DataFormats/Detectors/TRD/include/DataFormatsTRD/Digit.h
```diff
@@ -41,7 +41,7 @@ using ArrayADC = std::array<ADC_t, constants::TIMEBINS>;
 //    Pre Trigger phase:
 //        LHC clock runs at 40.08MHz, ADC run at 1/4 or 10MHz and the trap runs at 120MHz or LHC*3.
 //        The trap clock can therefore be in 1 of 12 positions relative to the ADC clock.
-//        Only 4 of those positions are valid, hence there are 4 values for pre-trigger phase: 0,3,6,9.
+//        Only 4 of those positions are valid, dependent on the TRAP assembly program timing offset, as of 11/2022 pre-trigger phase: 0,3,6,9.
```
martenole, line 44: why as of 11/2022? will this change in the future?

DataFormats/Detectors/TRD/include/DataFormatsTRD/Digit.h
```diff
@@ -100,7 +100,7 @@ class Digit
   /// starting from ClassDef version 4, mDetector keeps both the chamber number and the trigger phase
   ///
   /// bits 0-11 contain the chamber number (valid range from 0-539)
-  /// bits 12-15 contain the trigger phase obtained from digit HC header (valid values 0, 3, 7 and 11)
```
martenole, line 103: thanks, this I missed

DataFormats/Detectors/TRD/include/DataFormatsTRD/Digit.h
bazinski: If the assembler program changes, then the timing changes and then possibly a new set of values with offset of 4.

martenole (approved):
Trivial, merging

PR #10335, opened 2022-11-22
Optionally apply BC shift to triggered detectors in CTF decoding

shahor02:
The `TriggerOffsetsParam` `configurableParam` got an array of custom corrections per detector (0 by default) which can be applied to triggered detectors IR as an *additive* correction during CTF decoding. Note that this convention is opposite to that for the `LM_L0` correction which is subtracted from the encoded IR (at the moment only for the TRD if an option `--correct-trd-trigger-offset` was provided to the `ctf-reader`). Therefore, for the TRD, if this option is ON, the effect of the correction is 
```
ir_corrected = ir_decoded - TriggerOffsetsParam::LM_L0 + TriggerOffsetsParam::customOffset[TRD]; 
```
while for other EMC,PHS,CPV and HMP (and TRD in absence of `--correct-trd-trigger-offset`) it is simply 
```
ir_corrected = ir_decoded + TriggerOffsetsParam::customOffset[<det>];
```
In case the correction moves the IR before the TF 1st orbit, the trigger is discarded.

Update: ZDC in pbpb2022 was resetting the orbit internally to 1 at the SOR and had BC shifted by -1 like for RORC detectors: implemented the shift also for it.

### DIFF
--- a/DataFormats/Detectors/CTP/include/DataFormatsCTP/TriggerOffsetsParam.h
+++ b/DataFormats/Detectors/CTP/include/DataFormatsCTP/TriggerOffsetsParam.h
@@ -22,9 +22,10 @@ namespace o2
 namespace ctp
 {
 struct TriggerOffsetsParam : public o2::conf::ConfigurableParamHelper<TriggerOffsetsParam> {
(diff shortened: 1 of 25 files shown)


shahor02:
merging after a trivial conflict resolution, the CIs were green

PR #10336, opened 2022-11-22
Fixed a bug and modified DCS DPs to CCDB workflow. The bug in the mid…

syano0822:
…lle time value and the largest change value caluculation have been fixed. New workflow sends information about the number of entries in a given time interval and duration time.

### DIFF
--- a/Detectors/ITSMFT/MFT/condition/include/MFTCondition/MFTDCSProcessor.h
+++ b/Detectors/ITSMFT/MFT/condition/include/MFTCondition/MFTDCSProcessor.h
@@ -39,25 +39,33 @@ using DPVAL = o2::dcs::DataPointValue;
 using DPCOM = o2::dcs::DataPointCompositeObject;
 
 struct MFTDCSinfo {
+
   std::pair<uint64_t, double> firstValue; // first value seen by the MFT DCS processor
   std::pair<uint64_t, double> lastValue;  // last value seen by the MFT DCS processor
+  std::pair<uint64_t, double> meanValue;  // mean value seen by the MFT DCS processor
+  std::pair<uint64_t, double> stddevValue; // standard deviation value seen by the MFT DCS processor
   std::pair<uint64_t, double> midValue;   // mid value seen by the MFT DCS processor
   std::pair<uint64_t, double> maxChange;  // maximum variation seen by the MFT DCS processor
+  std::pair<uint64_t, double> summary;    // number of entries by the MFT DCS processor
+
   MFTDCSinfo()
   {
     firstValue = std::make_pair(0, -999999999);
     lastValue = std::make_pair(0, -999999999);
+    meanValue = std::make_pair(0, -999999999);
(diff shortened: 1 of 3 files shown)


rpezzi (dismissed):
I understand this was already tested at P2. Thanks, Satoshi.

Detectors/ITSMFT/MFT/condition/src/MFTDCSProcessor.cxx
```diff
@@ -114,25 +107,26 @@ int MFTDCSProcessor::processDP(const DPCOM& dpcom)
...
-                                                                       // that we did not get the
-                                                                       // same timestamp as the
-                                                                       // latest one
-      dvect.push_back(val);
-    }
+  // if (type == DPVAL_DOUBLE) {
+  // for these DPs, we will store the first, last, mid value, plus the value where the maximum variation occurred
+  auto& dvect = mDpsdoublesmap[dpid];
+
+  LOG(info) << "Processing DP = " << dpcom << ", with value = " << o2::dcs::getValue<double>(dpcom) << ", # is " << dvect.size();
```
shahor02: could you suppress or move this to verbose mode?

Detectors/ITSMFT/MFT/condition/include/MFTCondition/MFTDCSProcessor.h
```diff
@@ -41,19 +41,26 @@ using DPCOM = o2::dcs::DataPointCompositeObject;
...
+    summary = std::make_pair(0, -999999999);
   }
   void makeEmpty()
   {
-    firstValue.first = lastValue.first = midValue.first = maxChange.first = 0;
-    firstValue.second = lastValue.second = midValue.second = maxChange.second = -999999999;
+    firstValue.first = lastValue.first = midValue.first = meanValue.first = stddevValue.first = maxChange.first = summary.first = 0.;
+    firstValue.second = lastValue.second = midValue.second = meanValue.second = stddevValue.second = maxChange.second = summary.second = -999999999;
   }
   void print() const;
```
shahor02, line 66: Since the persistent class layout was changed, you should increment the version in the ClassDefNV below

shahor02 (changes requested):

Detectors/ITSMFT/MFT/condition/include/MFTCondition/MFTDCSProcessor.h
syano0822: Hi Ruben,
Thanks a lot for your review.
I didn't know that the number in ClassDefNV should be modified.
I have a question about the rule.
Does the "persistent class" mean the class used in storing objects?
Satoshi

davidrohr: @shahor02 @syano0822 : What is the status of this PR? Can we address @shahor02 's comment, or shall we close it?

PR #10339, opened 2022-11-22
MCH: added time information to Track class

aferrero2707:
The track time is computed as the average of the time of the associated digits.
If digits are not forwarded to the track finder, the track time is set as the start of the corresponding ROF.

### DIFF
--- a/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
+++ b/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
@@ -110,9 +110,7 @@ class TrackMCH
   /// set the number of the clusters attached to the track and the index of the first one
   void setClusterRef(int firstClusterIdx, int nClusters) { mClusRef.set(firstClusterIdx, nClusters); }
 
-  /// get the interaction record associated to this track
-  InteractionRecord getIR(uint32_t refOrbit) const;
-
+  /// mean time associated to this track, with error
   const Time& getTimeMUS() const { return mTimeMUS; }
   Time& getTimeMUS() { return mTimeMUS; }
   void setTimeMUS(const Time& t) { mTimeMUS = t; }
@@ -121,6 +119,9 @@ class TrackMCH
     mTimeMUS.setTimeStamp(t);
     mTimeMUS.setTimeStampError(te);
   }
+
+  /// interaction record corresponding to the mean track time
+  InteractionRecord getMeanIR(uint32_t refOrbit) const;
 
  private:
   static constexpr int SNParams = 5;  ///< number of track parameters
@@ -151,7 +152,7 @@ class TrackMCH
   double mZAtMID = 0.;                 ///< z position on the MID side where the parameters are evaluated
   double mParamAtMID[SNParams] = {0.}; ///< 5 parameters: X (cm), SlopeX, Y (cm), SlopeY, q/pYZ ((GeV/c)^-1)
   double mCovAtMID[SCovSize] = {0.};   ///< reduced covariance matrix of track parameters, formated as above
-  Time mTimeMUS;                       ///< associated time in microseconds from the TF start
+  Time mTimeMUS{};                     ///< associated time in microseconds from the TF start
 
   ClassDefNV(TrackMCH, 3);
 };
--- a/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
+++ b/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
@@ -21,6 +21,8 @@
 #include <iosfwd>
 
 #include "CommonDataFormat/RangeReference.h"
+#include "CommonDataFormat/InteractionRecord.h"
+#include "CommonDataFormat/TimeStamp.h"
 
 namespace o2
 {
@@ -31,6 +33,7 @@ namespace mch
 class TrackMCH
 {
   using ClusRef = o2::dataformats::RangeRefComp<5>;
+  using Time = o2::dataformats::TimeStampWithError<float, float>;
 
  public:
   TrackMCH() = default;
@@ -107,6 +110,18 @@ class TrackMCH
   /// set the number of the clusters attached to the track and the index of the first one
   void setClusterRef(int firstClusterIdx, int nClusters) { mClusRef.set(firstClusterIdx, nClusters); }
 
+  /// get the interaction record associated to this track
+  InteractionRecord getIR(uint32_t refOrbit) const;
+
+  const Time& getTimeMUS() const { return mTimeMUS; }
+  Time& getTimeMUS() { return mTimeMUS; }
+  void setTimeMUS(const Time& t) { mTimeMUS = t; }
+  void setTimeMUS(float t, float te)
+  {
+    mTimeMUS.setTimeStamp(t);
+    mTimeMUS.setTimeStampError(te);
+  }
+
  private:
   static constexpr int SNParams = 5;  ///< number of track parameters
   static constexpr int SCovSize = 15; ///< number of different elements in the symmetric covariance matrix
@@ -136,8 +151,9 @@ class TrackMCH
   double mZAtMID = 0.;                 ///< z position on the MID side where the parameters are evaluated
   double mParamAtMID[SNParams] = {0.}; ///< 5 parameters: X (cm), SlopeX, Y (cm), SlopeY, q/pYZ ((GeV/c)^-1)
   double mCovAtMID[SCovSize] = {0.};   ///< reduced covariance matrix of track parameters, formated as above
+  Time mTimeMUS;                       ///< associated time in microseconds from the TF start
 
-  ClassDefNV(TrackMCH, 2);
+  ClassDefNV(TrackMCH, 3);
 };
 
 std::ostream& operator<<(std::ostream& os, const TrackMCH& t);
--- a/DataFormats/Detectors/MUON/MCH/src/TrackMCH.cxx
+++ b/DataFormats/Detectors/MUON/MCH/src/TrackMCH.cxx
@@ -17,7 +17,6 @@
 #include "DataFormatsMCH/TrackMCH.h"
 
 #include <cmath>
-#include <limits>
 #include <fmt/format.h>
 #include <string>
 #include <iostream>
@@ -88,25 +87,11 @@ void TrackMCH::setCovariances(const TMatrixD& src, double (&dest)[SCovSize])
 }
 
 //__________________________________________________________________________
-InteractionRecord TrackMCH::getIR(uint32_t refOrbit) const
+InteractionRecord TrackMCH::getMeanIR(uint32_t refOrbit) const
 {
-  // track time converted in bunch-crossing units
-  auto fbc = mTimeMUS.getTimeStamp() / o2::constants::lhc::LHCBunchSpacingMUS;
-  // wrap-up one full orbit if the track time is negative,
-  // the first TF orbit is also decreased by one to compensate
-  if (fbc < 0) {
-    fbc += o2::constants::lhc::LHCMaxBunches;
-    refOrbit -= 1;
-  }
-
-  // get the bc-in-orbit value
-  uint64_t trackBCinTF = uint64_t(std::round(fbc));
-  uint16_t trackBC = uint16_t(trackBCinTF % o2::constants::lhc::LHCMaxBunches);
-  // get the absolute track orbit by adding the first orbit of the current TF
-  uint32_t trackOrbit = uint32_t(trackBCinTF / o2::constants::lhc::LHCMaxBunches) + refOrbit;
-
-  // build the track interaction record
-  return {trackBC, trackOrbit};
+  InteractionRecord startIR(0, refOrbit);
+  auto trackBCinTF = std::llround(mTimeMUS.getTimeStamp() / constants::lhc::LHCBunchSpacingMUS);
+  return startIR + trackBCinTF;
 }
 
 std::ostream& operator<<(std::ostream& os, const o2::mch::TrackMCH& t)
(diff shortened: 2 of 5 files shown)


aferrero2707:
@pillot @aphecetche could you have a look? Thanks!

DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
```diff
@@ -107,6 +110,18 @@ class TrackMCH
   /// set the number of the clusters attached to the track and the index of the first one
   void setClusterRef(int firstClusterIdx, int nClusters) { mClusRef.set(firstClusterIdx, nClusters); }
 
+  /// get the interaction record associated to this track
+  InteractionRecord getIR(uint32_t refOrbit) const;
```
pillot: This might be misleading as the track cannot be associated one single IR. If you want to keep this function, maybe use getMeanIR or getAverageIR or getMostProbableIR?

DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
```diff
@@ -136,8 +151,9 @@ class TrackMCH
   double mZAtMID = 0.;                 ///< z position on the MID side where the parameters are evaluated
   double mParamAtMID[SNParams] = {0.}; ///< 5 parameters: X (cm), SlopeX, Y (cm), SlopeY, q/pYZ ((GeV/c)^-1)
   double mCovAtMID[SCovSize] = {0.};   ///< reduced covariance matrix of track parameters, formated as above
+  Time mTimeMUS;                       ///< associated time in microseconds from the TF start
```
pillot: add a default initializer: Time mTimeMUS{};

DataFormats/Detectors/MUON/MCH/src/TrackMCH.cxx
```diff
@@ -87,6 +87,28 @@ void TrackMCH::setCovariances(const TMatrixD& src, double (&dest)[SCovSize])
   }
 }
 
+//__________________________________________________________________________
+InteractionRecord TrackMCH::getIR(uint32_t refOrbit) const
```
pillot: this can be simplified by using the existing InteractionRecord facilities, e.g.:
InteractionRecord TrackMCH::getIR(const InteractionRecord& startIR) const
{
  auto trackBCinTF = std::llround(mTimeMUS.getTimeStamp() / constants::lhc::LHCBunchSpacingMUS);
  return startIR + trackBCinTF;
}

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -22,6 +22,7 @@
 #include <stdexcept>
 #include <string>
 #include <filesystem>
+#include <limits>
```
pillot: not needed anymore

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -153,12 +155,87 @@ class TrackFinderTask
   }
 
  private:
+  std::pair<InteractionRecord, bool> trackBC2IR(double trackBCinTF, const ROFRecord& trackROF, uint32_t firstTForbit) const
```
pillot: this function is not needed anymore

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -153,12 +155,87 @@ class TrackFinderTask
...
+      LOG(warning) << fmt::format("track time incompatible with track ROF. TRACK: BC={}, IR={},{}  ROF: start={},{}, end={},{}  FIRST_TF_ORBIT: {}",
+                                  trackBCinTF, trackIR.orbit, trackIR.bc, rofStart.orbit, rofStart.bc, rofEnd.orbit, rofEnd.bc, firstTForbit);
+      isOk = false;
+    }
+
+    return {trackIR, isOk};
+  }
+
+  //_________________________________________________________________________________________________
+  void setTrackTimeFromROF(TrackMCH& track, const ROFRecord& trackROF, uint32_t firstTForbit) const
```
pillot: remove this function and use the one from the ROFRecord instead, to do not duplicate the code.

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -153,12 +155,87 @@ class TrackFinderTask
...
+    float bcDiff = ir.differenceInBC(startIR);
+    bcDiff += rofHalfWidth;
+    float tMean = o2::constants::lhc::LHCBunchSpacingMUS * bcDiff;
+    // the time resolution is estimated as the half-width of the track ROF
+    float tErr = o2::constants::lhc::LHCBunchSpacingMUS * rofHalfWidth;
+    track.setTimeMUS(tMean, tErr);
+  }
+
+  //_________________________________________________________________________________________________
+  void setTrackTime(TrackMCH& track, const ROFRecord& trackROF, uint32_t firstTForbit, const gsl::span<const Cluster> usedClusters, const gsl::span<const Digit> usedDigits) const
```
pillot: trackROF --> clusterROF

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -153,12 +155,87 @@ class TrackFinderTask
...
+    }
+
+    // set the track IR from the computed average digits time
+    if (nDigits > 0) {
+      float tMean = o2::constants::lhc::LHCBunchSpacingMUS * trackBCinTF;
+      float tErr = o2::constants::lhc::LHCBunchSpacingMUS * mTrackTime3Sigma;
+      track.setTimeMUS(tMean, tErr);
+    } else {
+      // if no digits are found, compute the time directly from the track's ROF
+      LOG(warning) << "no digits found when computing the track mean time";
```
pillot: With the current code this should never happen. This check can be kept anyway as a sanity check, but I think it should produce a LOG(fatal).

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -153,12 +155,87 @@ class TrackFinderTask
...
+      setTrackTimeFromROF(track, trackROF, firstTForbit);
```
pillot: use e.g. track.setTimeMUS(clusterROF.getTimeMUS({0, firstTForbit}).first); instead.

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -198,10 +275,18 @@ class TrackFinderTask
           cluster.firstDigit = digitLoc.first->second;
         }
       }
+
+      // compute the track time
+      if (mDigits && usedDigits) {
+        setTrackTime(mchTracks.back(), clusterROF, firstTForbit, usedClusters, *usedDigits);
+      } else {
+        setTrackTimeFromROF(mchTracks.back(), clusterROF, firstTForbit);
```
pillot: use e.g. mchTracks.back().setTimeMUS(clusterROF.getTimeMUS({0, firstTForbit}).first); instead.

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -198,10 +275,18 @@ class TrackFinderTask
           cluster.firstDigit = digitLoc.first->second;
         }
       }
+
+      // compute the track time
+      if (mDigits && usedDigits) {
```
pillot: it's enough to check mDigits

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -153,12 +155,87 @@ class TrackFinderTask
...
+    float tMean = o2::constants::lhc::LHCBunchSpacingMUS * bcDiff;
+    // the time resolution is estimated as the half-width of the track ROF
+    float tErr = o2::constants::lhc::LHCBunchSpacingMUS * rofHalfWidth;
+    track.setTimeMUS(tMean, tErr);
+  }
+
+  //_________________________________________________________________________________________________
+  void setTrackTime(TrackMCH& track, const ROFRecord& trackROF, uint32_t firstTForbit, const gsl::span<const Cluster> usedClusters, const gsl::span<const Digit> usedDigits) const
+  {
+    float trackBCinTF = 0;
```
pillot: float --> double

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -153,12 +155,87 @@ class TrackFinderTask
...
+    int nDigits = 0;
+
+    // loop over digits and compute the average track time
+    for (const auto& cluster : usedClusters.subspan(track.getFirstClusterIdx(), track.getNClusters())) {
+      for (const auto& digit : usedDigits.subspan(cluster.firstDigit, cluster.nDigits)) {
+        nDigits += 1;
+        trackBCinTF += (float(digit.getTime()) - trackBCinTF) / nDigits;
```
pillot: float --> double. Even if in the end to time is saved as a float, the computation must be done in double precision to avoid rounding issues (the TF duration in ns is of the order of the significand precision of a float).

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
```diff
@@ -153,12 +155,87 @@ class TrackFinderTask
...
+      }
+    }
+
+    // set the track IR from the computed average digits time
+    if (nDigits > 0) {
+      float tMean = o2::constants::lhc::LHCBunchSpacingMUS * trackBCinTF;
```
pillot: One must add 1.5 BC to trackBCinTF to get the correct mean value, because the digit time t0 is set to the first BC of an ADC sample which last 4BC. So the correct digit time can be t0 + 0, + 1, + 2 or + 3, which is t0 + 1.5 in average.

DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
aferrero2707: I propose to remove the function, and call it `getMeanIR()`

DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
aferrero2707: Done

DataFormats/Detectors/MUON/MCH/src/TrackMCH.cxx
aferrero2707: Ok, unless we completely remove the function

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: Removed

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: removed

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: Indeed, removed

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: Ok

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: Ok. I have added MCH in the message to better specify which detector is generating the fatal error

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: Done

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: done

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: ok

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: Done

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: Good point!

Detectors/MUON/MCH/Workflow/src/TrackFinderSpec.cxx
aferrero2707: Good point!

DataFormats/Detectors/MUON/MCH/src/TrackMCH.cxx
```diff
@@ -17,7 +17,6 @@
 #include "DataFormatsMCH/TrackMCH.h"
 
 #include <cmath>
-#include <limits>
```
pillot: this library is not used in TrackFinderSpec, but it is used in the current code of TrackMCH.

pillot (dismissed):

pillot (approved):

aphecetche:
Oups. Forgot to squash the commits. Sorry about that.

PR #10346, opened 2022-11-23
Revert "DPL: avoid race condition when resetting output proxy"

ktf:
Reverts AliceO2Group/AliceO2#9778


davidrohr:
I have tried a SYNTHETIC run with the default software (environment 2c1B337W6GZ) and with this PR (environment 2c1BWhsAhRB). In both cases we have the same errors
```
E | 2 | 14:01:40.074 |   | DPL | tpc-calibrator-dEdx | 2c1BWhsAhRB | 529706 | Dropping TPC/MIPS/0 Lifetime::0 data in slot 0 with timestamp 73 < 114.
-- | -- | -- | -- | -- | -- | -- | -- | --
```
at the end of run.

PR #10347, opened 2022-11-23
Replace CTF_METAFILES_DIR by EPN2EOS_METAFILES_DIR

martenole:
And configure the meta files directory and output directory for PHOS and ZDC calibration workflows

### DIFF
--- a/Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/include/TPCInterpolationWorkflow/TPCResidualAggregatorSpec.h
+++ b/Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/include/TPCInterpolationWorkflow/TPCResidualAggregatorSpec.h
@@ -40,7 +40,7 @@ namespace calibration
 class ResidualAggregatorDevice : public o2::framework::Task
 {
  public:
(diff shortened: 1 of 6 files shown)


martenole:
Requires https://github.com/AliceO2Group/O2DPG/pull/749 plus an update of the gen_topo.sh script locally on the EPNs

davidrohr:
Perhaps we should set the default /dev/null CALIB_DIR in setenv.sh not i  start_tmux.sh, would be more universal?

martenole:
I don't have a strong preference. In start_tmux.sh its only set for our FST. 
And in the aggregator-workflow.sh I do it simultaneously to how it is done in dpl-workflow.sh: CTF_DIR and EPN2EOS_METAFILES_DIR are also defined there and not in setenv.sh
But per default I could of course also set it to /dev/null instead of FILEWORKDIR

davidrohr:
Well, we coukd set the default to /dev/null in setenc.sh, and set it in other places only if different to /dev/null. Woukd be cleaner in my opinion. And avoids duplication to some extent.

martenole:
Ok, I have moved the settings to setenv.sh in https://github.com/AliceO2Group/O2DPG/pull/749 and in https://github.com/AliceO2Group/AliceO2/pull/10347/commits/73f73023c840363d7cdfcd04c913f4261f1e79d2 I have added one block to avoid file writing in case we are not in a PHYSICS run. To be able to test the different compression levels also in synthetics I allow writing calibration output to a local file in case `FORCE_LOCAL_CALIBRATION_OUTPUT=1` is set which should be done only for short tests where I am watching in order to not fill up the disks... Not sure if this is such an ideal setup.

And sorry for all the noise with the multiple pushs..

davidrohr (approved):
seems all good to me now

davidrohr:
why is this green if https://github.com/AliceO2Group/O2DPG/pull/749 is still read, and you write we require 749? Anyway, I'll let you merge it once you are sure you are done.

martenole:
Yes, this must be merged first now, in order for the O2DPG PR to pass the CI. I am merging now

PR #10348, opened 2022-11-23
Special treatment for shifts of ZDC triggers at BC=0,3563

shahor02:
These special triggers can be shifted only orbit-wise but the BC must remain unchanged


shahor02:
Failure is unrelates, tested locally, merging

PR #10350, opened 2022-11-23
Fix TPC CTF IO test

shahor02:
Compare data-members explicitly to avoid padding problems in the blockwise comparison


shahor02:
Tested locally on mac, merging as the current unit test is broken.

PR #10352, opened 2022-11-24
ITS: Implement AlpideResponseSimulation, on-the-fly root file generation, and more.

mconcas:
Dear @iouribelikov @AlexBigO,

TL;DR: proposed solution is to use a dedicated executable, that does the job of the previous macro, as you can see in the code.

Could you please check if this works for you, also for the digitisation, please?

Longer explanation:
 Macros are really painful to run during the installation, essentially because it's hard to satisfy all the recursive `INCLUDE_PATH`s of all the involved headers, not only the one you have finally in the macro.
There is no clean way to do so, and this makes also sense as multiple targets are being built in parallel in a non-predictable way (at least the uncorrelated ones).

As a personal comment, I would prefer to store all of these .txt files in a dedicated repository in AliceO2, and fetch them with the executable, so that we don't store 800+ more files in this code repository.
I'll let maybe comment @shahor02 and @ktf on this.


Detectors/ITSMFT/common/simulation/include/ITSMFTSimulation/DigiParams.h
```diff
@@ -110,9 +110,9 @@ class DigiParams
   int mNSimSteps = 7;                      ///< number of steps in response simulation
   float mEnergyToNElectrons = 1. / 3.6e-9; // conversion of eloss to Nelectrons
 
-  float mVbb = 3.0;   ///< back bias absolute value for MFT (in Volt)
-  float mIBVbb = 3.0; ///< back bias absolute value for ITS Inner Barrel (in Volt)
-  float mOBVbb = 3.0; ///< back bias absolute value for ITS Outter Barrel (in Volt)
+  float mVbb = 0.0;   ///< back bias absolute value for MFT (in Volt)
+  float mIBVbb = 0.0; ///< back bias absolute value for ITS Inner Barrel (in Volt)
+  float mOBVbb = 0.0; ///< back bias absolute value for ITS Outter Barrel (in Volt)
```
shahor02, line 115: perhaps one can get rid of duplication of all these parameters already defined in the DPLDigitizerParam  (legacy of O2 early days) and use this latter directly?

shahor02 (commented):
Hi @mconcas 

Looks fine. For the dedicated repository for the response files, I agree that having it as a dependency of the O2 (or O2Sim) would be cleaner and would also facilitate having different versions of response files as different branches in this repo.

mconcas:
> Hi @mconcas
> 
> Looks fine. For the dedicated repository for the response files, I agree that having it as a dependency of the O2 (or O2Sim) would be cleaner and would also facilitate having different versions of response files as different branches in this repo.

Then I'd create a dedicated repo for the response files. Maybe generic enough to host also the ones for ITS 3 (when they will be available), so to accommodate both.
Then with @AlexBigO we will restructure his contribution accordingly.

mconcas:
This requires: https://github.com/alisw/alidist/pull/4701

Detectors/ITSMFT/common/data/AlpideResponseData/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+o2_add_executable(alpide-response-generator
+                  SOURCES AlpideResponse.cxx
+                  PUBLIC_LINK_LIBRARIES O2::ITSMFTSimulation
+                                        ROOT::Core
+                  TARGETVARNAME targetName)
+
+set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${CMAKE_SOURCE_DIR}/Detectors/ITSMFT/common/data/AlpideResponseData/AlpideResponse.cxx)
+
+add_custom_command(TARGET O2exe-alpide-response-generator POST_BUILD
+                   COMMAND ${CMAKE_BINARY_DIR}/stage/bin/o2-alpide-response-generator -i  $ENV{ITSRESPONSE_ROOT}/response/AlpideResponseData/ -o ${CMAKE_SOURCE_DIR}/Detectors/ITSMFT/common/data/AlpideResponseData/
```
iouribelikov: The ROOT file with responses should be installed under $O2_ROOT/share.  (Not under $CMAKE_SOURCE_DIR)

iouribelikov (changes requested):
Thank you very much, @mconcas !
The generation of ROOT file with responses works. 
But :
1) The file gets installed under the source directory, whereas the digitiser tries to get it from under the $O2_ROOT/share/.
2) Now, this PR needs to be rebased on dev.

Detectors/ITSMFT/common/data/AlpideResponseData/CMakeLists.txt
mconcas: > Thank you very much, @mconcas ! The generation of ROOT file with responses works. But :
> 
> 1. The file gets installed under the source directory, whereas the digitiser tries to get it from under the $O2_ROOT/share/.
> 2. Now, this PR needs to be rebased on dev.

Thanks, this I saw, and fixed in my local fork I was waiting to have the package in alidist.
I'll update the pr these days.

mconcas:
@iouribelikov could you try with this?

iouribelikov (dismissed):
For me, everything works now perfectly fine.
Thanks a lot, @mconcas.

mconcas:
> For me, everything works now perfectly fine. Thanks a lot, @mconcas.

great, I fixed another hardcoded path in Digitizer.h, if you approve I can merge when/if it becomes green.

iouribelikov (dismissed):
@mconcas  It looks like both of us are working on a Mac !  :-)

mconcas:
> @mconcas It looks like both of us are working on a Mac ! :-)

True, but I have to tell that full_system_test script that runs in fullCI failed on mac too. So case insentitivity is not always that sloppy.
Hope this is last iteration on this... :)

mconcas:
> @mconcas It looks like both of us are working on a Mac ! :-)

@iouribelikov, Sorry for dismissing review again, previos copy_file  command was creating issues with permission on file, making it impossible to delete it as normal user. Now should be fixed.

iouribelikov (approved):

mconcas:
Hi @iouribelikov @shahor02, I think we miss either @sawenzel 's approval or a more powerful merger here.
I cannot merge without all required approvals.

iouribelikov:
@mconcas @shahor02 @sawenzel  Yes. I think we could merge it.

shahor02:
Hi, the fullCI has failed this morning because the ccdb server was down. 
Was the CI fully green before the power cut?

mconcas:
Hi, yes it was. But I don't mind to wait another iteration.

aphecetche:
@mconcas would it be possible to add a more explicit dependency on the new package `ITSResponse` at the cmake level on the O2 side ? Because as it is now the dependency is a bit hidden and only comes through alidist (and thus compilations not using alibuild only notice the dependency very late in the build...)

e.g. a `FindITSResponse.cmake` and the corresponding `find_package(ITSResponse)` ?

mconcas:
> @mconcas would it be possible to add a more explicit dependency on the new package `ITSResponse` at the cmake level on the O2 side ? Because as it is now the dependency is a bit hidden and only comes through alidist (and thus compilations not using alibuild only notice the dependency very late in the build...)
> 
> e.g. a `FindITSResponse.cmake` and the corresponding `find_package(ITSResponse)` ?

@aphecetche: ok, I see. I'll try to set that up.

PR #10354, opened 2022-11-24
When root-outputs are requested, write also TRD tracklets/digits

shahor02:


bazinski:
Is this not a repeat of this ? 

https://github.com/AliceO2Group/AliceO2/blob/a1133a1e30312159a5f43fec6c52f616ac9da28a/Detectors/TRD/reconstruction/src/DataReader.cxx#L100

PR #10359, opened 2022-11-25
Use DataProcessorContext and StreamContext to handle callbacks

ktf:
* Introduce a StreamContext:
  * Used to hold Stream specific state and information
  * Used to hold callbacks which need to be executed on a per stream basis
  * I had to change the DataTakingContext to be a stream service since it implements a per stream callback
* Move callbacks which are intended to be executed on a per dataprocessor basis to the DataProcessorContext.


ktf:
Trying this without the MessageBackend changes.

ktf:
Confused about  full CI working but not the rest. Maybe unrelated? Restarting all the tests.

ktf:
Remaining issue with the CI was a wrong assert. Tested on hyperloop and all the tests pass with no visibile regression in  terms of performance / memory usage. Merging.

PR #10364, opened 2022-11-25
TOF: Add unfied container for parameters on ccdb

njacazio:
@noferini

- add container to store pass and parameters in a map scheme

### DIFF
--- a/DataFormats/Detectors/TOF/CMakeLists.txt
+++ b/DataFormats/Detectors/TOF/CMakeLists.txt
@@ -16,6 +16,7 @@ o2_add_library(DataFormatsTOF
                        src/CalibLHCphaseTOF.cxx
                        src/CalibTimeSlewingParamTOF.cxx
                        src/CTF.cxx
+                       src/ParameterContainers.cxx
                        src/CalibInfoCluster.cxx
                        src/CosmicInfo.cxx
                        src/Diagnostic.cxx
@@ -33,6 +34,7 @@ o2_target_root_dictionary(DataFormatsTOF
                                   include/DataFormatsTOF/RawDataFormat.h
                                   include/DataFormatsTOF/CompressedDataFormat.h
                                   include/DataFormatsTOF/CTF.h
+                                  include/DataFormatsTOF/ParameterContainers.h
                                   include/DataFormatsTOF/CalibInfoCluster.h
                                   include/DataFormatsTOF/CosmicInfo.h
                                   include/DataFormatsTOF/TOFFEElightInfo.h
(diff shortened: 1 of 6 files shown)


njacazio:
@noferini I had to rework a bit the code and the naming of the member functions could you confirm that it works with your macro?

noferini:
> @noferini I had to rework a bit the code and the naming of the member functions could you confirm that it works with your macro?

I sent you the macro

noferini (dismissed):

shahor02 (changes requested):
The error 
```
    LOG(debug) << "Changing parametrization corresponding to key " << key << " from size " << mParameters[key].size() < " to " << p.GetName() << " of size " << p.size();
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
is genuine.

njacazio:
Hi @shahor02 indeed, thanks for spotting this

shahor02 (commented):
@njacazio rechecking the code: don't you need to add a dictionary (ClassDefN and entry in the LinkDef.h) for the Parameters class? 
Note that for the templated classes the template parameter change is not supported by the root schema evolution, i.e. a legacy `Parameter<A>` will not readable with `Parameter<B>` even with class version change.

njacazio:
@shahor02 thanks a lot for the comment!
Indeed I think it will be fine, in principle we don't need the schema evolution. The `Parameter<int>` class is just a way to define sets of parameters with defined names (e.g. in https://github.com/AliceO2Group/AliceO2/pull/10364/files#diff-ca1417bbfb70e51406dde90ebbaf76c473f283d1c82548356e2093828d9228a4R64) that will make it easy to distribute the parameters taken from the `ParameterCollection` expected to be stored on the ccdb.
I hope I understood your point correctly and this clears your doubts.

shahor02:
@njacazio OK for template, but since this is persistent object, I still think you need a ClassDefNF + entry in the LinkDef.

noferini:
Hi, shall we merge this PR?

shahor02:
Hi,
I can merge it, but did you check that you can write and read back the object? As I wrote above, the Parameters class, which is supposed to be persistent, has not ClassDefNV

noferini:
Hi @shahor02,
I didn't have any problem to write and read.

  TFile algFile("ccdb.root", "recreate");
  algFile.WriteObjectAny(&tof, "o2::tof::ParameterCollection", "ccdb_object");
  algFile.Close();


TFile *f = new TFile("ccdb.root");
o2::tof::ParameterCollection *a = (o2::tof::ParameterCollection *) f->Get("ccdb_object");
---

shahor02:
OK, thanks, apparently it is able to use the ClassDef of the derived class. Merging

PR #10372, opened 2022-11-28
Possibility to add EMC recalib for async

chiarazampolli:
@davidrohr : this goes with https://github.com/AliceO2Group/O2DPG/pull/760
Since both touch scripts used online, please check.
@mfasDa


chiarazampolli:
Hello @davidrohr , 
Do you think the errors are related to this PR? I don't think so, the multinode test must be QC related.

PR #10375, opened 2022-11-28
PHOS calibrations check file path in device init

martenole:
We should not check file paths in the `defineDataProcessing` method. Otherwise we fail to create a topology on the infrastructure EPNs where the directories in `/data` are not available.
Therefore I moved the calls to `rectifyDirecetory` into the `init` function of the devices.


martenole:
Failure unrelated, merging now

PR #10381, opened 2022-11-29
Make sure calo table is sorted by BC

nburmaso:


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1207,55 +1207,97 @@ uint8_t AODProducerWorkflowDPL::getTRDPattern(const o2::trd::TrackTRD& track)
...
+                                           const o2::globaltracking::RecoContainer& data)
 {
-  uint64_t globalBC = 0;    // global BC ID
-  uint64_t globalBCRel = 0; // BC id reltive to minGlBC (from FIT)
+  // get calo information
+  auto caloEMCCells = data.getEMCALCells();
+  auto caloEMCCellsTRGR = data.getEMCALTriggers();
+
+  auto caloPHOSCells = data.getPHOSCells();
+  auto caloPHOSCellsTRGR = data.getPHOSTriggers();
```
shahor02, line 1248: Look fine, just 1 comment: in this way, if at least one of the calorimeters are requested, both will be filled. To avoid this you can add here 
```
if (!mInputSources[GIndex::PHS]) {
  caloPHOSCells = {};
  caloPHOSCellsTRGR = {};
}
if (!mInputSources[GIndex::EMC]) {
  caloEMCCells = {};
  caloEMCCellsTRGR = {};
}

nburmaso: Thank you for the comment. I have added these checks after L1248

shahor02 (approved):

PR #10382, opened 2022-11-29
[WIP][EMCAL-687] Discard EMCAL calibration triggers

mfasDa:
EMCAL calibration triggers must not enter the AOD,
as at AOD level there is no way to mark them as
calibration triggers. Therefore they would spoil the
event normalization.

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -36,6 +36,7 @@
 #include "CCDB/BasicCCDBManager.h"
 #include "CommonConstants/PhysicsConstants.h"
 #include "CommonDataFormat/InteractionRecord.h"
+#include "CommonConstants/Triggers.h"
 #include "DataFormatsTRD/TrackTRD.h"
 #include "DataFormatsTRD/TrackTriggerRecord.h"
 #include "DataFormatsGlobalTracking/RecoContainer.h"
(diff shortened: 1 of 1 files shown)


mfasDa:
In principle there are two ways for EMCAL to filter calibration triggers, in the cell-recalibrator-spec and in the aod-producer-spec. Instead they should be streamed to a different file (with a rate of ~1 Hz or less it should be acceptable to keep them all). Having the filter in the aod-producer-spec would have the advantage that it doesn't rely on other workflows running. The drawback is that PHOS does not propagate the trigger bits via there event handler, so a dedicated handling via std::is_same is needed to run the rejection only in the EMCAL case. Having the rejection in the cell converter would simplify the aod-producer-spec as no special treatment in PHOS is needed, on the cost of being discarded in th async QC or significant modifications for the async QC. Nevertheless they must not enter the AODs: For phyics analyses due to the event selection task they will probably not harm, but I would see no way explicitly accessing them from the AOD for analysis without specifying the BC as dedicated EMCAL calib trigger.

shahor02 (commented):
@mfasDa this will clash with the Nazar's https://github.com/AliceO2Group/AliceO2/pull/10381

mfasDa:
@shahor02 I will set the title to WIP to give priority to Nazar's PR. I am in any case not fully sure whether it wouldn't be better to filter them in the recalibrator workflow.

mfasDa:
We will do the separation between physics and calib trigger in the recalibrator, doing this in the aod producer will become too complicated since extra handling for EMCAL in a common function with PHOS would be needed. Furthermore we would need a selector for LED events running in parallel to the AOD producer in any case.

PR #10384, opened 2022-11-29
tof digitizer reads params from ccdb and cleanup

noferini:
Digitization parameters are updated from ccdb

--ccdb-tof-sa flag was already obsolete -> some cleanup (param left for consistency with the script -> remapped in use-ccdb-tof)

### DIFF
--- a/Detectors/GlobalTracking/src/MatchTOF.cxx
+++ b/Detectors/GlobalTracking/src/MatchTOF.cxx
@@ -57,6 +57,10 @@ ClassImp(MatchTOF);
 //______________________________________________
 void MatchTOF::run(const o2::globaltracking::RecoContainer& inp)
 {
+  if (!mMatchParams) {
+    mMatchParams = &o2::globaltracking::MatchTOFParams::Instance();
+  }
+
   ///< running the matching
   mRecoCont = &inp;
   mStartIR = inp.startIR;
@@ -1160,7 +1164,7 @@ void MatchTOF::selectBestMatches()
       flags = flags | o2::dataformats::CalibInfoTOF::kMultiHit;
     }
 
-    if (matchingPair.getChi2() < 3) { // extra cut in Chi2 for storing calib info
+    if (matchingPair.getChi2() < mMatchParams->calibMaxChi2) { // extra cut in ChiSquare for storing calib info
       mCalibInfoTOF.emplace_back(mTOFClusWork[matchingPair.getTOFClIndex()].getMainContributingChannel(),
                                  mTimestamp / 1000 + int(mTOFClusWork[matchingPair.getTOFClIndex()].getTimeRaw() * 1E-12), // add time stamp
                                  deltat,
--- a/Detectors/GlobalTracking/src/MatchTOF.cxx
+++ b/Detectors/GlobalTracking/src/MatchTOF.cxx
@@ -1160,7 +1160,7 @@ void MatchTOF::selectBestMatches()
       flags = flags | o2::dataformats::CalibInfoTOF::kMultiHit;
     }
 
-    if (matchingPair.getChi2() < 3) { // extra cut in Chi2
+    if (matchingPair.getChi2() < 3) { // extra cut in Chi2 for storing calib info
       mCalibInfoTOF.emplace_back(mTOFClusWork[matchingPair.getTOFClIndex()].getMainContributingChannel(),
                                  mTimestamp / 1000 + int(mTOFClusWork[matchingPair.getTOFClIndex()].getTimeRaw() * 1E-12), // add time stamp
                                  deltat,
--- a/Detectors/GlobalTracking/src/MatchTOF.cxx
+++ b/Detectors/GlobalTracking/src/MatchTOF.cxx
@@ -1160,10 +1160,12 @@ void MatchTOF::selectBestMatches()
       flags = flags | o2::dataformats::CalibInfoTOF::kMultiHit;
     }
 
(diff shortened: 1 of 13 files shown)


noferini:
Can we merge this PR?

At the last time I added also a minor change in the digit reader to implement (if requested) a delay in providing TF.
It is needed from QC in MC in order not to loose TF.

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -1160,10 +1160,12 @@ void MatchTOF::selectBestMatches()
       flags = flags | o2::dataformats::CalibInfoTOF::kMultiHit;
     }
 
-    mCalibInfoTOF.emplace_back(mTOFClusWork[matchingPair.getTOFClIndex()].getMainContributingChannel(),
-                               mTimestamp / 1000 + int(mTOFClusWork[matchingPair.getTOFClIndex()].getTimeRaw() * 1E-12), // add time stamp
-                               deltat,
-                               mTOFClusWork[matchingPair.getTOFClIndex()].getTot(), mask, flags);
+    if (matchingPair.getChi2() < 3) { // extra cut in Chi2
```
shahor02: can we avoid hardcoded cuts, using e.g. ConfigurableParam instead?

Detectors/TOF/workflowIO/src/DigitReaderSpec.cxx
```diff
@@ -35,6 +35,9 @@ void DigitReader::init(InitContext& ic)
   LOG(debug) << "Init Digit reader!";
   auto filename = o2::utils::Str::concat_string(o2::utils::Str::rectifyDirectory(ic.options().get<std::string>("input-dir")),
                                                 ic.options().get<std::string>("tof-digit-infile"));
+
+  mDelayInMuSec = atoi(ic.options().get<std::string>("reader-delay").c_str()) * 1E6;
```
shahor02: `reader-delay` option is dynamically added by the HBFUtilsInitalizer : https://github.com/AliceO2Group/AliceO2/blob/db2b72e6a7868cedbb7ec0d3f6ad7d1568d3d0dc/Detectors/Raw/src/HBFUtilsInitializer.cxx#L79 and triggers a delay in *seconds* between successive injections of TFs (but not before 1st one):
https://github.com/AliceO2Group/AliceO2/blob/db2b72e6a7868cedbb7ec0d3f6ad7d1568d3d0dc/Detectors/Raw/src/HBFUtilsInitializer.cxx#L152-L153
Could you clarify why do you need to add a delay *before* the 1st (and the only one in case of digitization) TF?,

shahor02 (commented):
Hi @noferini 
I did not understand your comment about losing the TF w/o delay. See the comment below.
Cheers,
 Ruben

Detectors/TOF/workflowIO/src/DigitReaderSpec.cxx
noferini: because I found that the reader often send the 1st (unique in MC) TF before the QC task is initialized (the reader is too fast).
This means that the task is loosing the single MC TF in many of the subjobs.
Since a MC process has only 1 TF we loose the majority of TFs in TOF digit QC.

By the way, is delay in your initializer in second?
usleep(delay); -> mus?

Cheers,
Francesco

noferini: also discussing here
https://alice.its.cern.ch/jira/browse/O2-3442

shahor02: I thought this problem (sending the 1st TF before the full workflow is initialized) is gone since long, better @ktf comments.
But if the delay needs to be added, you better add it via separate option, e.g. `--delay-1st-tf` and apply it to 1st TF only.
The `--reader-delay` option is passed as float in seconds and internally converted to \mus to be used with `usleep`.

noferini: Ok. DONE

noferini:
o2-cs8 failing in O2Physics compilation
This is not related to this PR

Detectors/GlobalTracking/src/MatchTOF.cxx
noferini: Can I do it in a separate PR? Just to speed up the merging of this one (I missed this comment).

Detectors/TOF/workflowIO/src/DigitReaderSpec.cxx
```diff
@@ -35,6 +35,8 @@ void DigitReader::init(InitContext& ic)
   LOG(debug) << "Init Digit reader!";
   auto filename = o2::utils::Str::concat_string(o2::utils::Str::rectifyDirectory(ic.options().get<std::string>("input-dir")),
                                                 ic.options().get<std::string>("tof-digit-infile"));
+  mDelayInMuSec1TF = atoi(ic.options().get<std::string>("delay-1st-tf").c_str()) * 1E6;
```
shahor02: No sure this will work: you say it is a delay in seconds, but convert the string to int (so values <1 will make no sense) then multiply it by 10^6. You should either make option to directly provide time in \mus or as float in seconds to be converted to \mus by scaling by 10^6

Detectors/GlobalTracking/src/MatchTOF.cxx
noferini: Done

Detectors/TOF/workflowIO/src/DigitReaderSpec.cxx
noferini: move atoi to atof. Thanks for spotting this. -> Done

PR #10395, opened 2022-11-30
[EMCAL-830] Options to drop or redirect cells from LED events

mfasDa:
Recalibrator can either keep LED events in
the standard cell output, drop them entirely
or redirect them to EMC/CELLS and
EMC/CELLSTRGR on subspec 10


mfasDa:
@shahor02 I have now hard coded the output spec to 10 for the LED cells in the redirect mode, I hope this does not create any problems. Up to now we were only using subspec 1 and 0, so 10 should be safe. I would like to keep the number of parameters compact for which I now did not make the subspec for the LED events settable.

shahor02:
They will simply go to the sink if nobody expects them.

PR #10396, opened 2022-11-30
MFT-MCH refit : Removing remnants of old implementation and adding options

mcoquet642:
Removing remnants of old implementation of the refit for matched MFT-MCH tracks, and adding options to tune MFT cluster and track propagation in Kalman filter.


mcoquet642:
Hello, is this unsuccessful test related to the PR or something else ? Thanks.

shahor02 (approved):

PR #10399, opened 2022-11-30
[MCH] adapt to variable TF size (for MC)

pillot:
This allows the time clustering to run in MC, where the TF has not a fixed number of orbits.
The algorithm itself is not changed and I checked that it gives the same results in data.
@aferrero2707 can you have a look please?


aferrero2707:
@pillot it looks ok for me.

aphecetche (approved):

PR #10400, opened 2022-11-30
prototype for the CTP inputs storage in AOD

shahor02:
@pzhristov @ddobrigk here is a draft for accounting CTP inputs in the BCs. Could you please check if the version tricks are correct?
@roman could you confirm that this is what you need?

### DIFF
--- a/Framework/Core/include/Framework/AnalysisDataModel.h
+++ b/Framework/Core/include/Framework/AnalysisDataModel.h
@@ -45,7 +45,7 @@ DECLARE_SOA_TABLE_VERSIONED(BCs_001, "AOD", "BC", 1, o2::soa::Index<>, //! Root 
                             bc::RunNumber, bc::GlobalBC,
                             bc::TriggerMask, bc::TriggerInputs);
 
-using BCs = BCs_001; //! this defines the current default version
+using BCs = BCs_000; //! this defines the current default version
 using BC = BCs::iterator;
 
 namespace timestamp
--- a/Framework/Core/include/Framework/AnalysisDataModel.h
+++ b/Framework/Core/include/Framework/AnalysisDataModel.h
@@ -1017,6 +1017,7 @@ namespace soa
 {
 DECLARE_EQUIVALENT_FOR_INDEX(aod::StoredMcParticles_000, aod::StoredMcParticles_001);
 DECLARE_EQUIVALENT_FOR_INDEX(aod::StoredTracks, aod::StoredTracksIU);
+DECLARE_EQUIVALENT_FOR_INDEX(aod::BCs_001, aod::BCs_000);
 } // namespace soa
 
 namespace aod
--- a/Framework/Core/include/Framework/AnalysisDataModel.h
+++ b/Framework/Core/include/Framework/AnalysisDataModel.h
@@ -31,14 +31,21 @@ DECLARE_SOA_STORE();
 
 namespace bc
 {
-DECLARE_SOA_COLUMN(RunNumber, runNumber, int);          //! Run number
-DECLARE_SOA_COLUMN(GlobalBC, globalBC, uint64_t);       //! Bunch crossing number (globally unique in this run)
-DECLARE_SOA_COLUMN(TriggerMask, triggerMask, uint64_t); //! CTP trigger mask
+DECLARE_SOA_COLUMN(RunNumber, runNumber, int);              //! Run number
+DECLARE_SOA_COLUMN(GlobalBC, globalBC, uint64_t);           //! Bunch crossing number (globally unique in this run)
+DECLARE_SOA_COLUMN(TriggerMask, triggerMask, uint64_t);     //! CTP trigger mask
+DECLARE_SOA_COLUMN(TriggerInputs, triggerInputs, uint64_t); //! CTP inputs
 } // namespace bc
 
-DECLARE_SOA_TABLE(BCs, "AOD", "BC", o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing
+DECLARE_SOA_TABLE(BCs_000, "AOD", "BC", o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing, version 000
                   bc::RunNumber, bc::GlobalBC,
                   bc::TriggerMask);
+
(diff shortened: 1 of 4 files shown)


jgrosseo:
As just discussed in the WP14, this needs a BC_000 --> BC_001 converter. @ddobrigk will have a look. I move this to draft for now.

aalkin:
This will also require
```cpp
namespace soa 
{
DECLARE_EQUIVALENT_FOR_INDEX(aod::BCs_001, aod::BCs_000);
}
```
for the index connection to work with both versions. See `AnalysisDataModel.h` lines 1009-1013. 
@shahor02 please also not that versioning of tables does not change the Description part of the message, versioning is done by subspec.

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -1017,6 +1017,7 @@ namespace soa
 {
 DECLARE_EQUIVALENT_FOR_INDEX(aod::StoredMcParticles_000, aod::StoredMcParticles_001);
 DECLARE_EQUIVALENT_FOR_INDEX(aod::StoredTracks, aod::StoredTracksIU);
+DECLARE_EQUIVALENT_FOR_INDEX(aod::BCs_001, aod::BCs_000);
```
shahor02, line 1020: @aalkin Thanks, here ?

aalkin: Yes, this is the place, thank you.

shahor02:
@ddobrigk @pzhristov shall I merge this? What about the the transformer you mentioned at the meeting?

ddobrigk:
Hi Ruben, please do not merge. Merging this without the converter will render all existing ao2ds useless. I have not had the time today to write the converter for you as I was busy with meetings the whole day but will look at this tomorrow. Thanks!

jgrosseo:
Could we keep it draft until the converter is ready?

jgrosseo:
In fact, it is even more complicated. In order to make the converter one needs to split this PR: Introduce the BC_001 table but keep the BC_000 the default (so no adjustments in the AOD producer). This can be merged then. Then the converter can be created in O2Physics and after that BC_001 is made default.

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -31,14 +31,21 @@ DECLARE_SOA_STORE();
...
-DECLARE_SOA_TABLE(BCs, "AOD", "BC", o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing
+DECLARE_SOA_TABLE(BCs_000, "AOD", "BC", o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing, version 000
                   bc::RunNumber, bc::GlobalBC,
                   bc::TriggerMask);
+
+DECLARE_SOA_TABLE_VERSIONED(BCs_001, "AOD", "BC", 1, o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing, version 001
+                            bc::RunNumber, bc::GlobalBC,
+                            bc::TriggerMask, bc::TriggerInputs);
+
+using BCs = BCs_001; //! this defines the current default version
```
ddobrigk: Maybe we don't need to split any PR: @shahor02, can we just change this line to say `using BCs = BCs_000;` and merge? This will be a fully transparent change and then I will do the O2Physics PR for the converter. Changing one line to `using BCs = BCs_001` later, once the converter is in O2Physics.

jgrosseo: But then the aod producer does not compile...

ddobrigk: Inspecting #10424, which is quite similar, the AOD producer compiles but there seems to be a binding check error stating that `Collisions_000` does not match `Collisions`. The error appears when testing and not when compiling, though. I guess we need further adjustments as `using xxx = xxx_000` does not seem enough to make this transparent.

jgrosseo: It is different because one additional member is added here. In your PR, the members are just swapped.

jgrosseo: Can you send me the error you get on mattermost?

ddobrigk: @shahor02, everyone, how about we just roll back the AO2D producers ([like this](https://github.com/shahor02/AliceO2/pull/34)) until we get the converter in place? Thanks!

shahor02:
@ddobrigk I've merged your PR directly on top of mine, will merge after CI is passed

jgrosseo:
I think something went wrong here because it actually merged the part in the producer as well (producing version 001). All the nightly tests of the produced MC failed:
https://alimonitor.cern.ch/train-workdir/tests/0004/00047699/stdout.log

@ddobrigk could you quickly make a PR to revert the producer part...

shahor02:
Shall I revert at the moment the whole PR?

chiarazampolli:
Hello, 
We are starting the apass2 (@catalinristea), if I understand, we cannot use the latest nightly, right? But we need it for some changes in O2DPG. 
Let us know, this is quite urgent... 
Chiara

chiarazampolli:
The issue should be just in reading, for production it means that we cannot have the Analysis QC running. It is ok for now, we might need to run the analysis QC with an updated tag (now it is a separate job).

ddobrigk:
Hi everyone, I am compiling the bc converter task right now and will have an O2Physics PR in a moment with it... But reverting part of this is also fine with me if you think that's more convenient.

ddobrigk:
So: 
https://github.com/AliceO2Group/AliceO2/pull/10470 should fix the AOD producer 
https://github.com/AliceO2Group/O2Physics/pull/1636 is the converter task
Could you all please check to see if I missed something? Thanks!

jgrosseo:
> Hello, We are starting the apass2 (@catalinristea), if I understand, we cannot use the latest nightly, right? But we need it for some changes in O2DPG. Let us know, this is quite urgent... Chiara

Please don't run with this intermediate version!

ddobrigk:
Okay, let's revert this PR then and go piecewise - my O2 PR might not fix it all completely. @shahor02 , if it helps, once this is reverted I can take a calmer look at what needs to be done incrementally without urgent mode :-)

shahor02:
reverted

PR #10403, opened 2022-12-01
Avoid using deprecated sprintf

ktf:


ktf:
Previous check was fine. Only formatting changed.

PR #10405, opened 2022-12-01
[EMCAL-830] Stream EMCAL LED events to dedicated ROOT files

mfasDa:
- Exclude LED events from stream entering the AODs,
  and redirect them to dedcated subspec 10
- Add cell writer for LED events listening on
  EMC/CELLS/10 and EMC/CELLSTRGR/10

### DIFF
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -379,7 +379,9 @@ has_detectors_reco MFT MCH && has_detector_matching MFTMCH && add_W o2-globalfwd
 
 # ---------------------------------------------------------------------------------------------------------------------
 # Reconstruction workflows applying detector-specific calibrations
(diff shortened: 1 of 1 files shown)


mfasDa:
Based on #10395: This uses the redirection of LED events in the cell-recalibrator-spec introduced in the previous commits and streams the LED cells (rate ~ 0.5 Hz) to a dedicated output file `emcledcells.root`.

The LED cells need to be streamed to a different output than the AOD event for various reasons, among them that calibration events are not marked in the AOD, which makes separation complicated (though they should not harm physics analysis once the event selection wagon is included).

mfasDa:
I have tested offline the chain redirecting physics and calib triggers to different outputs and got the number of events expected in both outputs. The chain can be considered as working.

shahor02 (approved):

PR #10406, opened 2022-12-01
Add option to quit on successful sending

wiechula:


wiechula:
The error looks unrelated. @shahor02 , could you merge if the change is fine for you?

PR #10397, opened 2022-11-30
Updated TimeSlotCalibration interface

martenole:
Since we do not require to specify the Input type anymore for the TimeSlot-based calibrations I removed it from the interface. In the past the Input type was used for the `fill()` method of the container which was handling the input, but by now this has been changed to achieve a better flexibility (see https://github.com/AliceO2Group/AliceO2/commit/00ca6ef5c32b5fbca9be7d37777b8b6f535dc478#diff-a347bacc31898ff9b134e8e454419707090f8ba2a195ff25044d1a472938b990 and some subsequent commits which changed that class).
The `fill()` method accepts an arbitrary number of inputs now and they do not have to be specified as template parameter for the calibration class itself.
I have updated `Detectors/Calibration/README.md` with more details.

PR #10411, opened 2022-12-01
[WIP] Prepare TPC residual aggregator for production

martenole:
Will merge this only after https://github.com/AliceO2Group/AliceO2/pull/10397 as the changes here are on top of that.

Main changes:
- allow configuration of the calibration via configKeyValues @ehellbar this I should have done earlier. Now you can set e.g. the cuts for voxel validation more easily via `--configKeyValues "scdcalib.maxSigY=1.5;scdcalib.maxSigZ=1."`
- add the slot boundaries and first and last processed TF of each slot to the output file name
- add branch with information on the reference tracks to the binned residuals. This allows to select e.g. only ITS-TPC-TRD-TOF tracks also afterwards when the map is created
- prepare limiting the maximum number of processed tracks per TF. This is still WIP. In case there are more tracks available then needed I will implement a sampling to avoid processing only a certain eta range for example

### DIFF
--- a/Detectors/CPV/calib/include/CPVCalibration/GainCalibrator.h
+++ b/Detectors/CPV/calib/include/CPVCalibration/GainCalibrator.h
@@ -70,7 +70,7 @@ class GainCalibData
 }; // end GainCalibData
 //=============================================================================
 using GainTimeSlot = o2::calibration::TimeSlot<GainCalibData>;
-class GainCalibrator final : public o2::calibration::TimeSlotCalibration<Digit, GainCalibData>
+class GainCalibrator final : public o2::calibration::TimeSlotCalibration<GainCalibData>
 {
  public:
   GainCalibrator();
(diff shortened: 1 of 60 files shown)


martenole:
Hi @wiechula and @ehellbar 
Chiara told me that you were discussing the naming scheme of the output files today. Sorry I missed the discussion.
With the changes I have in here the output files are named for example:
```
o2tpc_residuals_368991_421703_394984_404776.root
```
This is for a calibration interval of 10 minutes which is translated into 52713 TFs. The first number 368991 is the beginning of the calibration slot in number of TFs (in this case 368991 / 52713 = 7 so it is the 7th calibration slot of 10 minutes for given run) and the next number 421703 is the last TF that would fall into that slot.
The following two numbers 394984 and 404776 are the last TFs which were actually processed in this job which were added to the job.
Now one could do the merging based on the defined slot boundaries 368991 and 421703.

Of course I could also instead of the absolute number of TFs put there the time in ms corresponding to the TF index. That would make the file names a bit longer though.

What is your preference?

Cheers,
Ole

chiarazampolli:
Hello @martenole , @wiechula , @ehellbar , 
In principle, it does not make any difference, but if we store indices instead of timestamps in the filename, that at the moment when we create the maps (CCDB objects) we need to translate them in timestamps. Knowing the run number should be possible. 
Chiara

martenole:
I am going to close and reopen another PR without the first commit. That way less people are bothered with notifications..

PR #10412, opened 2022-12-01
Switch from degrees to radians, protection against bad covariances

shahor02:


shahor02:
Failure is unrelated, merging

PR #10413, opened 2022-12-02
DPL: avoid regex recompilation

ktf:
The fact that the regex was created in the lambda caused its optimization to be done on a per invokation basis. While this is not a big deal when the metrics are dumped at the end or for small workflows, in case of processing on the EPN with the remote GUI enabled, the sheer amount of metrics which are collected are enough to cause unrecoverable latency in the GUI, making it feel sluggish.


ktf:
Tested on hyperloop as well and the metrics are still there.

ktf:
See for example https://alimonitor.cern.ch/hyperloop/wagon-test/46335

PR #10414, opened 2022-12-02
Prepare residual aggregator for production

martenole:
To continue the discussion from https://github.com/AliceO2Group/AliceO2/pull/10411


ehellbar:
Hello @martenole , @chiarazampolli , @wiechula , @matthias-kleiner ,
as Chiara wrote and after discussion with @matthias-kleiner , at the level of residuals file name both the number if TF w.r.t. run start and the full timestamp in ms would be fine. We could calculate the timestamp at the stage of the ccdb object creation. 

@matthias-kleiner @wiechula  We should in any case propagate the information to the final voxelResiduals file with the results of the residual fits and stay consistent with what is used in the residual trees?

@martenole @chiarazampolli @wiechula What comes to mind is that it could be easier to correlate other data, e.g. QC data or IDCs, to a given residuals / voxelResiduals file via the timestamps if they are already in the file name.

@shahor02 Not directly related to this, but did I understand correctly that the correction object (TPCFastTransform) in the CCDB should have the start and end timestamp of the calibration interval also as data member, the same way as AliTPCChebCorr?

martenole:
After merging https://github.com/AliceO2Group/O2DPG/commit/8343e66deca47355b279121e2b54732a193d4a2b the calib output proxy expects track references which are not yet sent with the current O2@dev. For the EPN software update tomorrow I would either need to revert the commit in O2DPG or merge these changes here now.
I think its easiest to merge this here now as is and then continue the work in a followup PR. This will contain:
- adding time stamps to the output file name
- sampling of tracks in case there are more tracks available per TF then required to have sufficient statistics for the calibration slot

PR #10417, opened 2022-12-02
DPL: move parseConfig / parseControl API to use std::string_view

ktf:
This provides better experience when running the remote GUI with very large workflows like the one running on the EPNs.


ktf:
Tested on hyperloop as well.

PR #10418, opened 2022-12-02
[EMCAL-798] Add input subspecification to offline calib workflow

jokonig:
- During reconstruction, the offline calibrator needs to get uncalibrated cells. Therefore it needs to listen to subspecification 1 instead of 0
- Additionally changed the makeCellIDTimeEnergy default argument to false, as it is a bool and cannot be set to false in the workflow options


mfasDa (dismissed):
Looks good

chiarazampolli (dismissed):

mfasDa (approved):

shahor02 (approved):

PR #10420, opened 2022-12-02
TPC: adding workflow for integrated cluster currents

matthias-kleiner:
Adding workflow for integrating the cluster charge qMax. The integrated cluster charges (ICCs) are similar to the IDCs and can be used for QA, validation of IDCs and/or for calibration in case IDCs are not available.
The format of the ICCs is currently the same as for the IDCs (`std::vector<float>` per CRU), which allows to use the same workflows for processing of the ICCs. The writing of the ICCs to CCDB is disabled for now.

Example from pp MC:
ICCs for A Side: [ICC_A.pdf](https://github.com/AliceO2Group/AliceO2/files/10143385/ICC_A.pdf)
IDCs for A Side: [IDC_A.pdf](https://github.com/AliceO2Group/AliceO2/files/10143397/IDC_A.pdf)

### DIFF
--- a/Detectors/TPC/workflow/CMakeLists.txt
+++ b/Detectors/TPC/workflow/CMakeLists.txt
@@ -200,6 +200,11 @@ o2_add_executable(calib-gainmap-tracks
                   SOURCES src/tpc-calib-gainmap-tracks.cxx
                   PUBLIC_LINK_LIBRARIES O2::TPCWorkflow O2::GlobalTracking)
 
+o2_add_executable(integrate-clusters
+                  COMPONENT_NAME tpc
+                  SOURCES src/tpc-integrate-cluster-currents
+                  PUBLIC_LINK_LIBRARIES O2::TPCWorkflow)
+
 o2_add_test(workflow
             COMPONENT_NAME tpc
             LABELS tpc workflow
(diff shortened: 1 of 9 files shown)


miranov25:
We discussed the issue in the chat.
The difference between the IDC and the ICC in the PDF attached above is probably due to the estimator used.
The Qtot estimator is more suitable for charge estimation than the Qmax estimator.
This estimator is less dependent on the track inclination and thus less radially dependent.

miranov25:
image

miranov25:
image

Detectors/TPC/workflow/CMakeLists.txt
```diff
@@ -200,6 +200,11 @@ o2_add_executable(calib-gainmap-tracks
                   SOURCES src/tpc-calib-gainmap-tracks.cxx
                   PUBLIC_LINK_LIBRARIES O2::TPCWorkflow O2::GlobalTracking)
 
+o2_add_executable(integrate-clusters
+                  COMPONENT_NAME tpc
+                  SOURCES src/tpc-integrate-cluster-currents
```
f3sch, line 205: Sorry to necrobump this PR, just saw that the file ending was missed here.
Now CMake produces the following warning:
image

matthias-kleiner: Thank you for noticing, I will fix it.

PR #10421, opened 2022-12-03
[EMCAL-798] Fix: Argument was not in quotation marks

jokonig:
- Arguments to emc offline calib workflow in calib workflow was no in quotation marks


mfasDa:
This should fix the crashes observed in the test run of LHC22o (double quotes were missing around the newly introduced subspec statement)

mfasDa (approved):

shahor02 (approved):
trivial, merging

PR #10422, opened 2022-12-03
DPL: minor fixes / optimisations

ktf:


ktf:
Formatting will be reapplied in a subsequent PR.

PR #10424, opened 2022-12-04
Update AnalysisDataModel.h

ddobrigk:


ddobrigk:
This is required to swap the PV cov matrix XZ and YY elements following the discussion in the WP4+WP14 meeting (more info [here](https://indico.cern.ch/event/1219192/contributions/5128952/attachments/2542060/4376840/DDChinellato-MondayMeeting-7Nov2022-1.pdf))

jgrosseo:
This looks good. 

To make the transition more smooth, I propose to first merge this and have still Collision_000 as the default and then after the O2Physics PR is merged, make another PR which changes the default to _001.

aalkin:
You will also need to add `DECLARE_EQUIVALENT_FOR_INDEX` to make sure both versions are interchangeable.

ddobrigk:
Thanks, now done: Collisions_000 is the default and the index equivalency has been declared. Actually like this we could already merge, so if there are no objections, I will change to ready for review shortly.

ddobrigk:
I think the test that failed was just a timeout, likely due to the system problems today. I guess this can be merged now in case we agree... thanks!

jgrosseo (approved):

PR #10428, opened 2022-12-05
TRD add reading/usage of T0 calibration value

luisabergmann:
To read and use the (average) TRD T0 calibration value in the TRD Tracklet Transformer from the (test-)CCDB.

### DIFF
--- a/Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
+++ b/Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
@@ -16,6 +16,7 @@
 #include "DataFormatsTRD/Tracklet64.h"
 #include "DataFormatsTRD/CalibratedTracklet.h"
 #include "DataFormatsTRD/CalVdriftExB.h"
+#include "DataFormatsTRD/CalT0.h"
 
 namespace o2
 {
@@ -43,13 +44,15 @@ class TrackletTransformer
   void setCalVdriftExB(const CalVdriftExB* cal) { mCalVdriftExB = cal; };
   void setApplyXOR() { mApplyXOR = true; }
 
+  void setCalT0(const CalT0* cal) { mCalT0 = cal; }
+
   float calculateY(int hcid, int column, int position, const PadPlane* padPlane) const;
 
   float calculateZ(int padrow, const PadPlane* padPlane) const;
 
   float calculateDy(int hcid, int slope, const PadPlane* padPlane) const;
 
-  float calibrateX(double x) const;
+  float calibrateX(int detector, float x) const;
 
   std::array<float, 3> transformL2T(int hcid, std::array<double, 3> spacePoint) const;
 
@@ -67,6 +70,7 @@ class TrackletTransformer
   float mXtb0;
 
   const CalVdriftExB* mCalVdriftExB{nullptr};
+  const CalT0* mCalT0{nullptr};
 };
 
 } // namespace trd
--- a/Detectors/TRD/base/src/TrackletTransformer.cxx
+++ b/Detectors/TRD/base/src/TrackletTransformer.cxx
@@ -75,10 +75,11 @@ float TrackletTransformer::calculateDy(int detector, int slope, const PadPlane* 
   return calibratedDy;
 }
 
-float TrackletTransformer::calibrateX(double x) const
+float TrackletTransformer::calibrateX(int detector, float x) const
 {
-  // Hard-coded value will be replaced once t0 calibration is available from CCDB
-  float t0Correction = -0.279;
+  // NOTE: Calibration value is average over all chambers, stored in PHOS hole (chamber 435)
+  //       option to use chamber-wise values will be implemented in the future
+  float t0Correction = mCalT0->getT0(435);
   return x += t0Correction;
 }
 
@@ -123,7 +124,7 @@ CalibratedTracklet TrackletTransformer::transformTracklet(Tracklet64 tracklet, b
 
   float dy = calculateDy(detector, slope, padPlane);
 
-  float calibratedX = calibrateX(x);
+  float calibratedX = calibrateX(detector, x);
 
   // NOTE: Correction to y position based on x calibration NOT YET implemented. Need t0.
   if (trackingFrame) {
--- a/Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
+++ b/Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
@@ -127,6 +127,8 @@ void TRDTrackletTransformerSpec::updateTimeDependentParams(ProcessingContext& pc
     mTransformer.init();
   }
   pc.inputs().get<o2::trd::CalVdriftExB*>("calvdexb"); // just to trigger the finaliseCCDB
+
+  pc.inputs().get<o2::trd::CalT0*>("calt0"); // just to trigger the finaliseCCDB, added for T0 calibration
 }
 
 void TRDTrackletTransformerSpec::finaliseCCDB(ConcreteDataMatcher& matcher, void* obj)
@@ -137,6 +139,11 @@ void TRDTrackletTransformerSpec::finaliseCCDB(ConcreteDataMatcher& matcher, void
   if (matcher == ConcreteDataMatcher("TRD", "CALVDRIFTEXB", 0)) {
     LOG(info) << "CalVdriftExB object has been updated";
     mTransformer.setCalVdriftExB((const o2::trd::CalVdriftExB*)obj);
+    return;
+  }
+  if (matcher == ConcreteDataMatcher("TRD", "CALT0", 0)) {
+    LOG(info) << "CalT0 object has been updatet";
+    mTransformer.setCalT0((const o2::trd::CalT0*)obj);
     return;
   }
 }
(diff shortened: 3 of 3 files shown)


martenole:
Hi Luisa, you can use clang-format to do the formatting automatically as described here: https://github.com/AliceO2Group/CodingGuidelines
Basically running `git clang-format HEAD~2` inside your O2 folder should be sufficient. Afterwards you need to create a new commit for the changes.
And could you afterwards also squash your commits into a single one?

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
```diff
@@ -139,6 +141,15 @@ void TRDTrackletTransformerSpec::finaliseCCDB(ConcreteDataMatcher& matcher, void
     mTransformer.setCalVdriftExB((const o2::trd::CalVdriftExB*)obj);
     return;
   }
+<<<<<<< HEAD
+  if (matcher == ConcreteDataMatcher("TRD", "CALT0", 0)) {
+=======
+  if(matcher == ConcreteDataMatcher("TRD", "CALT0", 0)){
+>>>>>>> 232fd18031eb964f909dd3198ae0faf3b3ae960f
```
f3sch: This will not built. There is a merge artifact left.

luisabergmann: Thanks, I deleted the obsolete lines accordingly

luisabergmann:
Hi Ole, 
thanks for the comment. I ran the clang-format and squashed the commits

Detectors/TRD/base/src/TrackletTransformer.cxx
```diff
@@ -75,10 +75,11 @@ float TrackletTransformer::calculateDy(int detector, int slope, const PadPlane*
...
 
-float TrackletTransformer::calibrateX(double x) const
+float TrackletTransformer::calibrateX(int detector, double x) const
 {
-  // Hard-coded value will be replaced once t0 calibration is available from CCDB
-  float t0Correction = -0.279;
+  // NOTE: Calibration value is average over all chambers, stored in PHOS hole (chamber 435)
+  //       option to use chamber-wise values will be implemented in the future
+  float t0Correction = mCalT0->getT0(435);
   return x += t0Correction;
```
f3sch: I think the CI might be stuck, the first 6 checks usually run within a minute.
Maybe you could change 'double x' to 'float x' so that we do get rid of the narrowing warning, commit that and push.
This restarts the CI. Two birds, one stone :).

shahor02: It is not stuck, 1st-time committer should be approved to trigger the CI

shahor02 (dismissed):
conditionally approving to start CI

Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
```diff
@@ -67,6 +69,7 @@ class TrackletTransformer
   float mXtb0;
 
   const CalVdriftExB* mCalVdriftExB{nullptr};
+  const CalT0* mCalT0{nullptr};
```
martenole, line 73: seems your are missing an include for `CalT0`

Detectors/TRD/base/src/TrackletTransformer.cxx
```diff
@@ -75,10 +75,11 @@ float TrackletTransformer::calculateDy(int detector, int slope, const PadPlane*
...
 }
 
-float TrackletTransformer::calibrateX(double x) const
+float TrackletTransformer::calibrateX(int detector, double x) const
 {
-  // Hard-coded value will be replaced once t0 calibration is available from CCDB
-  float t0Correction = -0.279;
+  // NOTE: Calibration value is average over all chambers, stored in PHOS hole (chamber 435)
+  //       option to use chamber-wise values will be implemented in the future
+  float t0Correction = mCalT0->getT0(435);
```
martenole, line 82: What is the value for the average T0 of all chambers? Does it spread too much for all chambers or why should we not directly use `detector` here instead? I don't remember now whether you had shown this in the past, sorry

Detectors/TRD/base/src/TrackletTransformer.cxx
luisabergmann: Just to make sure we don't run into precision errors with the tracklets, is float enough for x? Or should I rather change the t0Correction to double?

martenole: float is more than sufficient for all calculations in the tracklet transformer

Detectors/TRD/base/src/TrackletTransformer.cxx
luisabergmann: The average is 1.223 with a std. dev. of ~0.16 across chambers that don't 1) have a "0" as correction (no PH spectrum or failed fit) or 2) have a bad fit (ie. can be easily identified when looking at the fits). A few chambers deviate a lot farther, mostly due to non-ideal fits because of lacking statistics. With @tdietel we discussed keeping the chamber-wise values for all successfull fits and use the average value for all other chambers; but it was decided to leave 0 for failing fits etc., store the average value in the PHOS hole, and use the average value for all chambers (for now, with the option to maybe later use chamber-wise values)

Detectors/TRD/base/src/TrackletTransformer.cxx
luisabergmann: Ok thanks, I changed the type accordingly

Detectors/TRD/base/include/TRDBase/TrackletTransformer.h
luisabergmann: Thanks, I added the include

Detectors/TRD/base/src/TrackletTransformer.cxx
martenole: OK, but which unit is that? Probably time bins, right? Because then we would still need to multiply this with the vDrift to get the correction for the radial position.
With the current change you would shift all TRD space points more than 1.5cm towards larger radii compared to what we are doing now.

bazinski: Would it not be more elegant to simply add an average value to the CalT0 class, than hide an average value in the PHOS hole.

tdietel: Can the mean be calculated from the other values in the CalT0 class? All the values are there, but are there any criteria that are no longer accessible at this point that are used to include/exclude a value in the calculation of the mean? I am mainly thinking about quality of the fit, i.e. chi2. 
If this is not possible, I would agree with Sean's comment that a separate member variable in CalT0 is a cleaner implementation.

tdietel: @martenole, I do not think that T0 should be multiplied with vDrift, as a change of vDrift does not change the start time of the signal. But we have to make sure T0 corresponds to the correct position within the chamber. We agreed that the T0 calibration will calculate the position of the maximum of the peak. This is necessary, because we do not see the actual start of the signal (unless we can find something in the high-resolution pulse height plots that are now possible with the pretrigger phases). The question is now what this T0 value corresponds to. My guess would be that this is the beginning of the drift region, i.e. just at the time when all the charge from the amplification region has been collected.

f3sch:
Not sure if this is related, but what about this section here:

```
double TrackletTransformer::getTimebin(int detector, double x) const
{
  // calculate timebin from x position within chamber
  float vDrift = mCalVdriftExB->getVdrift(detector);
  double t0 = 4.0; // time (in timebins) of start of drift region
   //                 |=> should the line above also be replace with mCalT0->getT0(detector); ?
  double timebin;
```

Detectors/TRD/base/src/TrackletTransformer.cxx
martenole: @luisabergmann what is the unit of your fit results?

luisabergmann: @martenole The unit is time bins
@tdietel At the moment chi2 of the fits is at this time no longer accessible. We could calculate the average in the TrackletTransformer.cxx if we set the t0 value of bad fits to eg. 0 earlier in the automated process (and as I did manually right now) and then only use the non-zero chambers

luisabergmann: Of course we could also add members to the `CalT0` class to either store the average directly or to keep chi2 etc to do the selection in `TrackletTransformer.cxx`.

Detectors/TRD/workflow/src/TRDTrackletTransformerSpec.cxx
```diff
@@ -139,6 +141,11 @@ void TRDTrackletTransformerSpec::finaliseCCDB(ConcreteDataMatcher& matcher, void
     mTransformer.setCalVdriftExB((const o2::trd::CalVdriftExB*)obj);
     return;
   }
+  if (matcher == ConcreteDataMatcher("TRD", "CALT0", 0)) {
+    LOG(info) << "CalT0 object has been updatet";
```
tdietel, line 145:
```suggestion
    LOG(info) << "CalT0 object has been updated";
```

Detectors/TRD/base/src/TrackletTransformer.cxx
tdietel: I would put the the calculation of averages and the decision if a chamber-specific or average value should be used in CalT0.
The TrackletTransformer should only use one of two functions: either `getT0(int detector)` or `getAverageT0()` that returns the average over all 'good' chambers (whatever that means). We should  check at some point which option gives the smaller residuals, but probably only after alignment has been done.
You mention three classes of fits: good, empty or failed, and bad (or should we call them the good, the bad and the ugly? ;-). Failed fits are stored as 0, and for good one, we just have the (positive) value. We could earmark bad fits by storing them as a negative value: we probably do not want to use such a value in the tracklet transformer, but it might useful in the monitoring of the calibration to have access to these results. QC could e.g. report on the fraction of good, failed and bad fits and show the distribution of T0 values for good (and maybe failed) fits.

tdietel: Did you check the stability of the fit procedure? This could be done by looking at the correlation of T0 values for all chambers for two calibration time windows. Basically, the question is to which degree the T0 values are limited by the resolution of the calibration algorithm, and to which degree it's differences between the chambers. This also informs if we should use the average or chamber-specific values.

tdietel (changes requested):

PR #10431, opened 2022-12-05
add passName in DigiParams

noferini:
This is to allow passing passName via configurable parameter (PR to update O2PDG script will follow)

### DIFF
--- a/Steer/DigitizerWorkflow/src/TOFDigitizerSpec.cxx
+++ b/Steer/DigitizerWorkflow/src/TOFDigitizerSpec.cxx
@@ -33,6 +33,7 @@
 #include "TOFBase/Geo.h"
 #include "Framework/CCDBParamSpec.h"
 #include "DataFormatsTOF/ParameterContainers.h"
+#include "SimConfig/DigiParams.h"
 
 using namespace o2::framework;
 using namespace o2::dataformats;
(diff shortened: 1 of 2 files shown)


Steer/DigitizerWorkflow/src/TOFDigitizerSpec.cxx
```diff
@@ -48,6 +49,7 @@ class TOFDPLDigitizerTask : public o2::base::BaseDPLDigitizer
 {
  public:
   TOFDPLDigitizerTask(bool useCCDB, std::string ccdb_url, int timestamp) : mUseCCDB{useCCDB}, mCCDBurl(ccdb_url), mTimestamp(timestamp), o2::base::BaseDPLDigitizer(o2::base::InitServices::FIELD | o2::base::InitServices::GEOM), mPass("unanchored"){};
+  // mPass("unanchored") to be replaced with mPass(o2::conf::DigiParams::Instance().passName) once ccdb ready
```
sawenzel, line 52: I think you could do this immediately and this would be backward compatible. (I am also not sure this will ever come from CCDB.)

noferini: I cannot because we have now only the default ccdb object which provides only an "unanchored" entry.
If we run on anchored MC it will produce a Fatal since the  entry is not ready (we have still to work a bit on that).

noferini:
fullCI error in O2Physics compilation (it is a frequent error not related to this PR)
/sw/SOURCES/O2Physics/10431-slc8_x86-64/0/Common/TableProducer/PID/pidTOF.cxx

noferini:
Hi @sawenzel ,
is the change fine with you?

PR #10435, opened 2022-12-05
DPL Analysis: Use AddAt rather than Get+Set

ktf:
Should save at least one virtual method invokation (maybe 2 if final works as hoped).

### DIFF
--- a/Framework/Core/include/Framework/StepTHn.h
+++ b/Framework/Core/include/Framework/StepTHn.h
@@ -27,6 +27,11 @@ class TArrayF;
 class TArrayD;
 class TCollection;
 
+struct StepTHnFillOp {
+  int64_t bin;
+  double weight;
+};
+
 class StepTHn : public TNamed
 {
  public:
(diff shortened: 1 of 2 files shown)


ktf:
When running under  IgProf on hyperloop this seems to remove completely the contribution of GetAt+SetAt. The bulk of the time is now in FindBin.

jgrosseo (approved):
Looks good to me. We can make a 1-1 comparison of the output to be sure

ktf:
yes, please.

Framework/Core/include/Framework/StepTHn.h
```diff
@@ -105,13 +113,40 @@ class StepTHnT : public StepTHn
...
+    } else if constexpr (std::is_same_v<TemplateArray, TArrayD>) {
+      auto* values = (TArrayD*)mValues[iStep];
+      values->AddAt(op.weight, op.bin);
+      if (mSumw2[iStep]) {
+        ((TArrayD*)mSumw2[iStep])->AddAt(op.weight, op.bin);
+      }
+    } else {
+      mValues[iStep]->SetAt(mValues[iStep]->GetAt(op.bin) + op.weight, op.bin);
+      if (mSumw2[iStep]) {
+        mSumw2[iStep]->SetAt(mSumw2[iStep]->GetAt(op.bin) + op.weight, op.bin);
```
ktf, line 138: btw, is this actually correct?

PR #10437, opened 2022-12-05
First prototype of strangeness tracking workflow

fmazzasc:
- matching between ITS track + V0 or Cascade
- re-assignement of fake clusters + final 3 body refit (w/ DCA fitter, to be improved) 
- output file w/ updated ITS track and reference to the decay daughters
- test macro added
- first performance study presented in: https://indico.cern.ch/event/1180716/


Detectors/StrangenessTracking/tracking/src/StrangenessTracker.cxx
```diff
@@ -0,0 +1,364 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+// StrangenessTracker
```
mpuccio: Ciao @fmazzasc, can you remove this line?

shahor02 (approved):

PR #10443, opened 2022-12-06
Fix status code usage in MCUtils

benedikt-voelkel:
* fix bug which leads to wrong primary assignment in AOD producer

* change status code type in MCTrack to Union o2::mcgenstatus::MCGenStatusEncoding

### DIFF
--- a/DataFormats/simulation/include/SimulationDataFormat/MCGenStatus.h
+++ b/DataFormats/simulation/include/SimulationDataFormat/MCGenStatus.h
@@ -23,6 +23,7 @@ constexpr int isEncodedValue{5};
 
 // internal structure to allow convenient manipulation of properties as bits on an int to (dis)entangle HepMC and specific generator status codes
 union MCGenStatusEncoding {
+  MCGenStatusEncoding() : fullEncoding(0) {}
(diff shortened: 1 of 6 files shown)


sawenzel (approved):

benedikt-voelkel:
Locally, I can compile the full chain, namely `O2sim`, so I don't see any problem with `O2Physics` as indicated here by the `O2` build.
In addition, I checked again some tests, including `o2sim`. Also there I don't see any problems as indicated by the `macOS-arm` test.
Do you have any insight @TimoWilken ?

Also, most of the formatting checks seem to hang?

TimoWilken:
Hi @benedikt-voelkel, we're investigating the problem with the Mac builder; there are some network issues with it.

I'm not sure what happened with the formatting checks; investigating.

TimoWilken:
@benedikt-voelkel seems like the formatting checks didn't run due to a GitHub hiccup. I've pushed an empty commit to force them to re-run (as there didn't seem to be another way to restart them), and they're green now.

For the previous commit, all our CI checks (including Mac) were green, so shall I merge this?

PR #10444, opened 2022-12-07
[FOCAL-10] Basic FOCAL raw decoding workflow

mfasDa:
- Port pixel decoding from QC to O2 (temporarily)
- Define event data format for FOCAL-E
- Add spec for raw decoding of pad and pixel event including merging of pad and pixel data
- Add executable for raw decoding workflow
- Add tree writer for FOCAL events


mfasDa:
CI error unrelated

shahor02 (commented):
@mfasDa just glancing: looks fine. I would suggest to have eventually a global numbering scheme for sensors.
Shall I merge it?

mfasDa:
@shahor02 yes, please merge. For what concerns the global number scheme this might come in a second iteration, we are working on it, for the beginning we might have it only for the pixel layers.

shahor02 (approved):

PR #10445, opened 2022-12-07
o2-sim: Generalized vertex sampling

sawenzel:
This commits brings 2 important new features:

a) Ability to sample an interaction vertex with
   z-dependent x and y positions according to
   MeanVertexObject.

   This is achieved by decoupling the vertex smearing
   process from FairPrimaryGenerator and by generating
   it in our space before calling FairPrimaryGenerator::generateEvent.

b) The ability to (pre-)fix interation vertices in the timeframe
   collision context. This will allow much more consistent application
   of vertices to all parts of a collision (background, signal) and scale
   better to many signal parts. It is a foundation to simplify embedding
   treatment later on.

https://alice.its.cern.ch/jira/browse/O2-3318

### DIFF
--- a/DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
+++ b/DataFormats/Calibration/include/DataFormatsCalibration/MeanVertexObject.h
@@ -69,6 +69,9 @@ class MeanVertexObject : public VertexBase
   void print() const;
   std::string asString() const;
 
+  /// sample a vertex from the MeanVertex parameters
+  math_utils::Point3D<float> sample() const;
+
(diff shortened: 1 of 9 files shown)


sawenzel:
Screenshot 2022-12-08 at 12 37 52

@shahor02 : Z-dependent x and y vertices (linear) in place. This is with the following command line:

`o2-sim --noGeant -n 10000 -g boxgen -o o2sim-boxgen --configKeyValues="Diamond.width[1]=3;Diamond.width[0]=3;Diamond.width[2]=10;Diamond.slopeX=2;Diamond.slopeY=2"`

PR #10446, opened 2022-12-07
[EMCAL-798] Apply gain calibration to cell energies in offline calib

jokonig:
- Energy calibration shifts cell energies. This can influence the final result of the bad channel calibration as it cuts on the mean cell energy for each cell
- application of the energy calibration can be switched on with the worflow argument applyGainCalib
- Similar implementation to what was one in the emcal cell recalibration workflow

### DIFF
--- a/Detectors/EMCAL/workflow/include/EMCALWorkflow/OfflineCalibSpec.h
+++ b/Detectors/EMCAL/workflow/include/EMCALWorkflow/OfflineCalibSpec.h
@@ -11,6 +11,8 @@
 
 #include <string>
 #include "Framework/DataProcessorSpec.h"
+#include "EMCALWorkflow/CalibLoader.h"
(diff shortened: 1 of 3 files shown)


mfasDa (approved):
Looks good

mfasDa:
@shahor02 Could you please review and merge? The change in the calibrator was tested offline.

The change does the following: The bad channel calibration is performed on raw cells in the CTF and the decision is made based on the energy spectrum. Hot/Cold cells might distort here since the spectral shape is affected by the gain. However the gain calibration take care of that. Without application of the gain calibration this leads to an overmasking. The gain calibration was not applied since it was still in the analysis when the calibration code was developed but is now final and in the CCDB, so we should take it into account for in the bad channel calibration.

shahor02 (approved):
will merge after CI

PR #10448, opened 2022-12-07
TPC SCD calib: Ues time stamps for filenames and some fixes for mergi…

martenole:
…ng slots

For async processing the fix for merging slots is not important, since here we want to have a fixed slot length and rather merge afterwards the output files. But for synch. processing this is important of course.

Output file has the format
`o2tpc_residuals_1667421310532_1667421910519_263565_316277.root`
The first two numbers are the time stamps of the beginning and end of the slot in ms, the last two numbers are the slot boundaries in number of TFs for the given run.

### DIFF
--- a/Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
+++ b/Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
@@ -27,12 +27,14 @@ using namespace o2::tpc;
 ResidualsContainer::~ResidualsContainer()
 {
   // trees must be deleted before the file is closed, otherwise segfaults
+  LOGP(debug, "Deleting ResidualsContainer with {} entries. File name is {}", getNEntries(), fileName);
   treeOutResidualsUnbinned.reset();
   treeOutTrackData.reset();
   treeOutResiduals.reset();
   treeOutStats.reset();
   treeOutRecords.reset();
   if (fileOut) {
+    LOG(debug) << "Removing output file of discarded slot";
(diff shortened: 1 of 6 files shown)


Detectors/TPC/calibration/SpacePoints/src/ResidualAggregator.cxx
```diff
@@ -305,6 +314,12 @@ void ResidualsContainer::merge(ResidualsContainer* prev)
   treeOutTrackData = std::move(prev->treeOutTrackData);
   treeOutResidualsUnbinned = std::move(prev->treeOutResidualsUnbinned);
 
+  // since we want to continue using the TTrees of the previous slot, we must
+  // avoid that ROOT deletes them when the TFile of the previous slot is erased
+  treeOutResiduals->SetDirectory(fileOut.get());
```
martenole, line 319: This took me some time to find, since I had not considered the ROOT object ownership which is not automatically transferred with the move above.

chiarazampolli:
Hello @shahor02 , @martenole , 
This is needed for the processing in December, right? Due to the changes to the filename.
Chiara

martenole:
I think we also need https://github.com/AliceO2Group/AliceO2/pull/10448/commits/7d517af98bca8e94b144e040e5881cc972851773 for the async processing. With that we can limit the total number of tracks which we accept and therefore the maximum output size per slot.
I am checking now again the output file size when processing 100 TFs with this. It should be much smaller than before, since we will discard most of the shorter ITS-TPC-TOF and ITS-TPC-TRD tracks

PR #10450, opened 2022-12-08
Adopt FairRoot 18.8

dennisklein:
To prepare O2 for the upcoming FairRoot v18.8 release (probably this year still). (See also [alisw/alidist@4687](https://github.com/alisw/alidist/pull/4687))

Preliminary Changelog: https://github.com/FairRootGroup/FairRoot/blob/v18.8_patches/CHANGELOG.md

Although we will release a first CMake package in v18.8, I prefer to keep O2's `FindFairRoot.cmake` for now and remove it earliest with the adoption of FairRoot v19 (probably early next year) to give the CMake package another iteration to mature (I will propose another PR then).

### DIFF
--- a/dependencies/FindFairRoot.cmake
+++ b/dependencies/FindFairRoot.cmake
@@ -27,17 +27,32 @@ find_library(FairRoot_ParBase ParBase)
 find_library(FairRoot_GeoBase GeoBase)
 find_library(FairRoot_Base Base)
 find_library(FairRoot_Gen Gen)
+find_program(FairRoot_CONFIG_EXECUTABLE NAMES fairroot-config)
 
 set(CMAKE_PREFIX_PATH ${OLD_CMAKE_PREFIX_PATH})
 
+if(FairRoot_CONFIG_EXECUTABLE)
+  execute_process(COMMAND ${FairRoot_CONFIG_EXECUTABLE} --major_version
+                  OUTPUT_VARIABLE FairRoot_VERSION_MAJOR
+                  OUTPUT_STRIP_TRAILING_WHITESPACE)
+  execute_process(COMMAND ${FairRoot_CONFIG_EXECUTABLE} --minor_version
+                  OUTPUT_VARIABLE FairRoot_VERSION_MINOR
+                  OUTPUT_STRIP_TRAILING_WHITESPACE)
+  execute_process(COMMAND ${FairRoot_CONFIG_EXECUTABLE} --patch_version
+                  OUTPUT_VARIABLE FairRoot_VERSION_PATCH
+                  OUTPUT_STRIP_TRAILING_WHITESPACE)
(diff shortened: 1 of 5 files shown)


dependencies/FindFairRoot.cmake
```diff
@@ -27,17 +27,32 @@ find_library(FairRoot_ParBase ParBase)
 find_library(FairRoot_GeoBase GeoBase)
 find_library(FairRoot_Base Base)
 find_library(FairRoot_Gen Gen)
+find_program(FairRoot_CONFIG_EXECUTABLE NAMES fairroot-config)
 
 set(CMAKE_PREFIX_PATH ${OLD_CMAKE_PREFIX_PATH})
 
+if(FairRoot_CONFIG_EXECUTABLE)
```
ktf, line 34: What's the point of using an external executable to find out what should already be there in some CMake Config file?

dennisklein: As I wrote, this PR is all still old pre config mode world. Once we adopt FairRoot v19 this goes away.

ktf: Ok, but why do you need this now, since what we currently have is also pre-config?

dennisklein: `${FairRoot_VERSION}` is used by 3db63ca0fd003d7a11c9bbaf26876a23b9cc7e9a or what do you mean?

ktf: Ok, I missed that, sorry about that. Couldn't you do some SFINAE trick instead? I'd really prefer to not have executables invoked from within Cmake, because in my experience those have the tendency to result in unneeded / missing rebuilds which take long to debug / require wiping the build area.

dennisklein: Hmm, not sure how to reflect a deprecated member variable. Another idea comes to mind: `fairroot-config` is just a small bash skript, I could read the file content and extract the version with some string/regex filter perhaps? Byt not sure if this would help regarding the behaviour you want to avoid. What do you think?

ktf: How about you add an include called "FairRootV18.8.h" and you use `__has_include()` to assert its existence?

ChristianTackeGSI: I am not 100% sure for any FairRoot versions prior to 18.6. But in 18.6 setting `fCreator` is unneeded (that's why we want to get rid of it and deprecated it for now).

Ideas:

1. Just remove setting `fCreator` completely. Even for older versions. No need for any optional/conditional stuff.

or

2. Keep the `fCreator` but calm the warning surgically:

   ```cpp
   #pragma GCC diagnostic push
   #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
     fCreator = this;
   #pragma GCC diagnostic pop
   ```

   and remove this, once 18.8 is fully adopted and no backward compatibility with 18.6 is needed.

PR #10451, opened 2022-12-08
FIT calibration: new features and fixes

afurs:
FT0 calibration: new time/amp spectra info objects defined

FIT calibration: fix for different input/output specs, FT0 new calib object

FT0 calibration: deprecated ChannelOffsetTask is removed


afurs:
Looks like unrelated issue

shahor02:
Hi @afurs 
the FullCI failure is still from the Friday test, meanwhile, the problem was fixed. could you please push a dummy commit to restart the test. BTW, the new code will already expect the new object, right? And I assume it will fail due to the absence of default object. Could you create one and open upload request?

afurs:
Hi @shahor02 , ok, I will restart CICD. No, new code doesn't require it for now. Only in my next PR, before that time JIRA ticket with default object will be ready.

shahor02 (approved):

PR #10453, opened 2022-12-09
Improve simple source

ktf:


ktf:
Should allow to mimick the IDCS / SACS setup.

PR #10455, opened 2022-12-09
remove one print if not verbose mode

noferini:
@preghenella


noferini:
fullCI error: not understood the reason but it shouldn't be related to this trivial PR

noferini:
All checks passed

PR #10456, opened 2022-12-09
[EMCAL-798] Add gain calib flag to offline calibrator

jokonig:
- EMCal energy calibration should be applied during filling of histograms in offline calibrator. Therefore, added workflow argument introduced in pull/10446
- Fix: Change Log(info) to LOG(debug)


mfasDa (approved):
Looks good

jokonig:
@shahor02 could you merge this PR?

shahor02 (approved):

PR #10463, opened 2022-12-09
Fix clang-tidy reported issues

ktf:


ktf:
Force merging since this is breaking the tests. @mfasDa please cross check, but it should be safe.

mfasDa:
Thanks for fixing! Looks fine

PR #10464, opened 2022-12-09
Refine gen status code usage

benedikt-voelkel:
* move bits to signal encoding to the very left

* slightly change MCTrack::getStatusCode interface

* add unit test for status code


benedikt-voelkel:
I have the suspicion that the failure of the test was due to some kind of race condition. The ROOT file used in the `MCTrack` and `testMCGenStatus` have the same name.
I will change that and we shall see

sawenzel (approved):

PR #10468, opened 2022-12-11
add method to get alignable volume by its label

shahor02:


shahor02:
tested locally, merging

PR #10469, opened 2022-12-12
[FOCAL-10] Move FOCAL code to Detectors/Upgrade

mfasDa:
Move FOCAL code to Detectors/Upgrade and
corresponding data formats for Dataformats/
Detectors/Upgrade


mfasDa:
Only move of code into upgrade directory - no change in logics

mconcas (dismissed):
Thanks a lot for this @mfasDa, sorry for the bothering!

mfasDa:
I hope upgrades are not disabled in the dataflow defaults. Because we are using the decoding code of both pads and pixels in the QC task for FOCAL-E on the FLP in our testbeam setup.

mconcas:
> I hope upgrades are not disabled in the dataflow defaults. Because we are using the decoding code of both pads and pixels in the QC task for FOCAL-E on the FLP in our testbeam setup.

Would this fail the related dataflow CI?

mfasDa:
> > I hope upgrades are not disabled in the dataflow defaults. Because we are using the decoding code of both pads and pixels in the QC task for FOCAL-E on the FLP in our testbeam setup.
> 
> Would this fail the related dataflow CI?

I am not 100% sure how the CI is configured, if it tests QC as well then probably yes, since the FOCAL QC library has no option which could exclude it from the build, for which it is always built. Since it uses the decoding code in O2 that should lead to a crash.

mconcas:
> > > I hope upgrades are not disabled in the dataflow defaults. Because we are using the decoding code of both pads and pixels in the QC task for FOCAL-E on the FLP in our testbeam setup.
> > 
> > 
> > Would this fail the related dataflow CI?
> 
> I am not 100% sure how the CI is configured, if it tests QC as well then probably yes, since the FOCAL QC library has no option which could exclude it from the build, for which it is always built. Since it uses the decoding code in O2 that should lead to a crash.

Ok, because in the dataflow defaults this is quite clear: https://github.com/alisw/alidist/blob/a5ff369c04660c80c2a805195054eeffb5eab99c/defaults-o2-dataflow.sh#L49

So, either we enable the upgrades in the defaults or this is probably required to be enabled in any case.

mfasDa:
> > > > I hope upgrades are not disabled in the dataflow defaults. Because we are using the decoding code of both pads and pixels in the QC task for FOCAL-E on the FLP in our testbeam setup.
> > > 
> > > 
> > > Would this fail the related dataflow CI?
> > 
> > 
> > I am not 100% sure how the CI is configured, if it tests QC as well then probably yes, since the FOCAL QC library has no option which could exclude it from the build, for which it is always built. Since it uses the decoding code in O2 that should lead to a crash.
> 
> Ok, because in the dataflow defaults this is quite clear: https://github.com/alisw/alidist/blob/a5ff369c04660c80c2a805195054eeffb5eab99c/defaults-o2-dataflow.sh#L49
> 
> So, either we enable the upgrades in the defaults or this is probably required to be enabled in any case.

Thanks! This is what I was afraid of. And now also the dataflow test crashed in QC. The QC might still be needed for the near future for which we don't want to drop it, so either the upgrades become enabled in the dataflow defaults or we need to move the libraries back to their original place.

mconcas:
> Thanks! This is what I was afraid of. And now also the dataflow test crashed in QC. The QC might still be needed for the near future for which we don't want to drop it, so either the upgrades become enabled in the dataflow defaults or we need to move the libraries back to their original place.

Ok, since as said this is not super critical we can leave it as it is.
Closing this PR would leave the repo in a state which is usable for you, if I understood correctly.
In the future I will then probably consider a more granular toggling of the various upgrades as it is also reasonable not to cover all of them under the same `enable-every-upgrade` flag.
That way we can add only the requried modules to the dataflow defaults or similar cases.
This was convenient when we had few modules but indeed is not going to scale.

I will probably be back to you when this happen, with a proposing PR.
Please just let me know in case you are going to drop the qc case before that happen.

Let me know if you agree, thanks anyways for your time.

mfasDa:
Thanks for your understanding! Dropping QC for the moment is rather unlikely as there might be further testbeam campaigns where the QC might be needed. So we would prefer keeping the code under regular detector code until ugrades will be available on the FLP.

PR #10470, opened 2022-12-12
Downgrade producer to produce o2*bc*

ddobrigk:


jgrosseo:
I think this is not enough. If I check https://github.com/AliceO2Group/AliceO2/pull/10400/files
I see lots of changes in producer-related files (just look for all BC_001). And this is not reverted here.
Maybe @shahor02 is right to revert the other and start from 0 is the easiest...

PR #10472, opened 2022-12-12
[FOCAL-10] Enable pixel mapping in build

mfasDa:
Pixel mapper provides the mapping between
pixel chip identifier and position in pixel layer
of the two testbeam setups used in the SPS
testbeam Nov 22.


mfasDa:
CI error unrelated, @shahor02 could you please merge?

(as discussed on https://github.com/AliceO2Group/AliceO2/pull/10469 we cannot move the FOCAL code to upgrades since it is needed for our QC task and upgrades are not built in dataflow mode, so we keep the code as regular detector code and continue development from there unless upgrades are built also in defaults dataflow).

PR #10474, opened 2022-12-12
[WIP] - ZDC - Updated ZDC AOD format

cortesep:
Updated ZDC data format correcting filling bug and adding signal amplitudes

### DIFF
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -264,11 +264,6 @@ class AODProducerWorkflowDPL : public Task
 
   BunchCrossings mBCLookup;
 
-  // zdc helper maps to avoid a number of "if" statements
-  // when filling ZDC table
-  map<string, float> mZDCEnergyMap; // mapping detector name to a corresponding energy
-  map<string, float> mZDCTDCMap;    // mapping TDC channel to a corresponding TDC value
-
   std::vector<uint16_t> mITSTPCTRDTriggers; // mapping from TRD tracks ID to corresponding trigger (for tracks time extraction)
   std::vector<uint16_t> mTPCTRDTriggers;    // mapping from TRD tracks ID to corresponding trigger (for tracks time extraction)
   std::vector<uint16_t> mITSROFs;           // mapping from ITS tracks ID to corresponding ROF (for SA ITS tracks time extraction)
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -1379,12 +1379,6 @@ void AODProducerWorkflowDPL::init(InitContext& ic)
     mT0Amplitude = 0xFFFFFFFF;
   }
 
-  // initialize zdc helper maps
-  for (auto& ChannelName : o2::zdc::ChannelNames) {
-    mZDCEnergyMap[(string)ChannelName] = 0;
-    mZDCTDCMap[(string)ChannelName] = 999;
-  }
-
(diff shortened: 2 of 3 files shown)


jgrosseo:
The ZDC format cannot be changed like this. The current format is used by the Run 2 converted data. You need to provide a version 1 of the table and provide a converter between the old and new format.
@ddobrigk can give you more details...

cortesep:
Do you want to convert RUN2 ZDC data to RUN3 data format?

If this is the case one should just drop the channels that have no 
meaningful data with a +/- 12.5 ns cut on the TDCs.

The amplitudes were not present in RUN2 but and one could just copy the 
ADC values. Which class makes this conversion?

To perform the reverse conversion from RUN3 to RUN2 one needs to pick 
random pedestals in any case because of the different readout strategy.

cortesep:
The proposed solution is the one that uses less space. If we prefer a "flat ntuple" solution it can be done by inserting dummy values like -infinity for all the channels that have no data.

ddobrigk:
Hi @cortesep, so, Jan Fiete's concern is that we should try and make this change in the ZDC format be transparent when looking into currently existing AO2Ds. In order to ensure that, we need the following: 

* we need to keep track of two versions of the table: the old ZDC table (used by existing AO2Ds) and the new ZDC table (the one that you propose) in our data model file. A little example is the v0 table (the decay, not the detector), which has a version without a collision index and one with a collision index: [see here](https://github.com/AliceO2Group/AliceO2/blob/d487bc5f98d6da00d98c548b83717a90d7020333/Framework/Core/include/Framework/AnalysisDataModel.h#L814). 
* we need to add a 'converter' task. This is easier to do than it sounds. The converter task is just a simple analysis task that takes a certain table in a certain, maybe first version (denoted by appending the usual name with `_000`) and creates another table with the usual name appended with `_001`. An example is the corresponding task that creates version 001 of the v0s table [here](https://github.com/AliceO2Group/O2Physics/blob/master/Common/TableProducer/weakDecayIndices.cxx).

So, these 'converter' tasks have to be written on a case-by-case basis, and I suppose this is what we could also consider in this ZDC change. Then the puzzle becomes one of just keeping, as much as possible, the useful information from Run 2 (as it exists already in the Run 2 converted) in the newer format of the table (which will be generated on-the-fly by the converter task. 

If you want, I would be happy to have a dedicated discussion to try and help further!

jgrosseo:
Thanks @ddobrigk for the detailed response. 
Just one clarification for @cortesep : The Run 1 and 2 data **has already been** converted, all of it: https://twiki.cern.ch/twiki/bin/viewauth/ALICE/AliceO2WP14AF
The format used is the present one in the data model.

cortesep:
Hi @jgrosseo and @ddobrigk ,
thanks for all the explanations. I didn't know about the converted AODs.

The idea of a converter task is very interesting and could be useful to address a few issues:
- the differences in data content that is present between the two data sets 
- to provide a simple and consistent interface for the analyzer (that was my intial concern). Probably this would also help the analyzer to avoid to use information that is present in RUN2 but is not present in RUN3 (in RUN3 there is more information but some is missing due to the readout strategy and the pedestal events are not reconstructed at the moment).
- to calibrate on-fly the ZN towers for RUN2 data. I just noticed that the tower intercalibration that I prepared some years ago apparently was not used in reconstruction

I'm happy to accept the help by @ddobrigk because I haven't worked on analysis tasks since quite a long time.

ddobrigk:
Hi @cortesep, shall we set up a zoom meeting at your earliest convenience? I would be happy to walk you through the details required for writing the converter. As I said, it should not be a major headache at all :-D my e-mail: david.dobrigkeit.chinellato@cern.ch ... can you write me and include also other interested people in cc? Thanks a lot!

cortesep:
Will proceed with separate pull requests

PR #10475, opened 2022-12-12
Provide trivial MC engine and application

benedikt-voelkel:
* O2TrivialMC VMC engine with a minimal implementation of interfaces. It does simply calls the hook for primary generation and pops them for consistency. They are not transported.

  Useful, when only generator level (aka MCKinematics) are needed since it has minimal initialisation overhead.

  Cannot be used for physics simulation, o2-sim will abort.

* O2TrivialMCApplciation VMC application with minimal implementation of interfaces. For the geometry it only constructs a simple box with a vacuum. Non primary generation or anything else.

* o2-sim --noGeant automatically runs with the O2trivialMC engine to ensure minimal initialisation overhead. Only CAVE is constructed, all other modules are omitted.

### DIFF
--- a/macro/o2sim.C
+++ b/macro/o2sim.C
@@ -53,7 +53,6 @@ void check_notransport()
     LOG(info) << "Initializing without Geant transport by applying very tight geometry cuts";
     o2::conf::ConfigurableParam::setValue("SimCutParams", "maxRTracking", 0.0000001);    // 1 nanometer of tracking
     o2::conf::ConfigurableParam::setValue("SimCutParams", "maxAbsZTracking", 0.0000001); // 1 nanometer of tracking
-    // TODO: disable physics processes for material sitting at the vertex
   }
 }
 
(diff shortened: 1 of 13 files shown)


benedikt-voelkel:
Structural comments and naming suggestions very welcome!

I will address the formatting PR after potential discussion.

Detectors/TrivialMC/include/DetectorsTrivialMC/O2TrivialMCApplication.h
```diff
@@ -0,0 +1,65 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
```
sawenzel: I would maybe not call the include dir "DetectorsTrivialMC". I think Detectors is misleading. What about "O2TrivialVMC" ?

macro/o2sim.C
```diff
@@ -49,11 +49,11 @@ void check_notransport()
   // be done on top.
   // This is merely offered for user convenience as it can be done from outside as well.
   auto& confref = o2::conf::SimConfig::Instance();
+  auto& configData = confref.getConfigData();
```
sawenzel: doesn't seem to be needed

Detectors/TrivialMC/include/DetectorsTrivialMC/O2TrivialMCApplication.h
benedikt-voelkel: Should it still be in the `Detectors` directory in the source? Or would you propose to move it somewhere else?
Or do you mean I should put it in the top directory under `O2TrivialVMC`?

Actually, I propose not to use `V` in the name since it is not virtual anymore?

sawenzel: I think "Detectors" itself may be fine. There is already some other VMC stuff in there.

benedikt-voelkel: I did some renaming. Does it make sense?

sawenzel (approved):

PR #10476, opened 2022-12-13
improve error message

jgrosseo:


jgrosseo:
@iarsene This should help to get the tree name in the case you reported

PR #10479, opened 2022-12-13
TPC SCD map creation from unbinned residuals

martenole:
This allows to specify the track type(s) which should be used for the extraction of the distortion map by e.g.

```
o2-tpc-static-map-creator -b --start-from-unbinned --track-sources ITS-TPC-TRD --residuals-infiles o2tpc_residuals.root
```

I have tested this with an input file with around 500k tracks which was not sufficient to create a map since most bins were masked. Will try tomorrow with the most recent output from Catalin.

Pinging @ehellbar and @matthias-kleiner

### DIFF
--- a/Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/include/TPCInterpolationWorkflow/TPCResidualReaderSpec.h
+++ b/Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/include/TPCInterpolationWorkflow/TPCResidualReaderSpec.h
@@ -15,8 +15,10 @@
 #define O2_TPC_RESIDUALREADER
 
 #include "Framework/DataProcessorSpec.h"
+#include "ReconstructionDataFormats/GlobalTrackID.h"
 
 using namespace o2::framework;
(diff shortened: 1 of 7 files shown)


martenole:
I do not understand the formatting check failures. Locally clang-format is happy

ehellbar:
Hi @martenole ,
thank you a lot for implementing this. Do you think one could add another option to also store the new binned residuals tree after the track source selection in the same file as the voxResTree for debug purposes?
Regards,
Ernst

martenole:
Hi Ernst,
yes, that's a good point. I will add that later (today there are a lot of meetings).
Cheers,
Ole

martenole:
> Hi @martenole , thank you a lot for implementing this. Do you think one could add another option to also store the new binned residuals tree after the track source selection in the same file as the voxResTree for debug purposes? Regards, Ernst

Actually, the new binned residuals were already stored automatically in the same output file because the TTree was assigned to the same TFile. Now by default the new binned residuals will not be written, unless `--store-binned` is specified.
Tested locally for run 527033. This is ready to be merged after CI has passed

PR #10483, opened 2022-12-14
CcdbDownloader implementation, which doesn't affect CcdbApi

TrifleMichael:
This PR is adds a version of the CCDB Downloader, which is not implemented into CcdbApi, in order to test it within O2. The CCDB Downloader is used to optimize the number of SSL handshakes and add additional functionality, like parallel downloads and asynchronous downloads.

### DIFF
--- a/CCDB/CMakeLists.txt
+++ b/CCDB/CMakeLists.txt
@@ -11,6 +11,7 @@
 
 o2_add_library(CCDB
                SOURCES  src/CcdbApi.cxx
+                        src/CCDBDownloader.cxx
                         src/BasicCCDBManager.cxx
                         src/CCDBTimeStampUtils.cxx
         src/IdPath.cxx src/CCDBQuery.cxx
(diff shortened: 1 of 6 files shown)


ktf (changes requested):
PR itself is fine with me, but please, fix the clang-format.

sawenzel:
I have difficulties understanding how this new class is supposed to be used? Is it targeting users or is it a helper ... later to be used by CcdbApi? Some more explanations in the PR message would be appreciated.

TrifleMichael:
This PR is adds a version of the CCDB Downloader, which is not implemented into CcdbApi, in order to test it within O2. The CCDB Downloader is used to optimize the number of SSL handshakes and add additional functionality, like parallel downloads and asynchronous downloads.

ktf (dismissed):

costing (dismissed):

ktf:
Ciao, can you please fix the fullCI?

sawenzel (approved):

sawenzel:
Approving to retrigger a build.

PR #10484, opened 2022-12-14
[WIP] fix for TPC-TRD track length

noferini:


noferini:
This is needed to fix a problem in the track length computation for TPC-TRD tracks.
Since TPC tracker doesn't integrated track lengths one piece of track is missing.
Applying in this case the fast computation as done for TPC only tracks before to start matching.

For ITS-TPC-TRD there is not problem since track length is integrated from the beginning (ITS).

PR #10485, opened 2022-12-14
[MCH] normalize the propagation of digits along the reconstruction chain

pillot:
The digits are now needed in the tracking workflows to compute the track time. Therefore I changed a bit the decision of propagating (or not) the digits along the reconstruction chain and in the output files, and I tried to make it more consistent between the various devices:

- by default, the digits associated to the clusters are written to / read from the cluster output files, unless the option --no-digits is specified in the corresponding workflows.

- by default, the track time is explicitly computed from the associated digits time (and thus the cluster digits are needed as input) in the tracking workflows, unless the option --disable-time-computation is specified (in which case the ROF time is used).

- by default, the digits associated to tracks are not written to / read from the track output root file, unless the option --digits is specified in the corresponding workflows.

I updated the README accordingly.

### DIFF
--- a/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
+++ b/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/TrackMCH.h
@@ -33,12 +33,13 @@ namespace mch
 class TrackMCH
 {
   using ClusRef = o2::dataformats::RangeRefComp<5>;
-  using Time = o2::dataformats::TimeStampWithError<float, float>;
 
  public:
+  using Time = o2::dataformats::TimeStampWithError<float, float>;
+
   TrackMCH() = default;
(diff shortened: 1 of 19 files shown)


aphecetche:
@pillot do you know what's the impact of the change of defaults to the output size of a typical `reco-workflow` run w/o `disable-root-output` ?

pillot:
> @pillot do you know what's the impact of the change of defaults to the output size of a typical `reco-workflow` run w/o `disable-root-output` ?

- if the option --digits is used, there is no difference.
- if the option --digits is not used, there is no difference for mchtracks.root (the digits are not written in this case as before), but the digits are still written in mchclusters.root, which represents ~70% of the size of this file for pp@500kHz.

I wanted to decouple the writing of tracks' digits, driven by --digits, from the writing of clusters' digits. But I can easily change the reco-workflow to get back to the previous behaviour, or add another option to disable the writing of the clusters's digits.

aphecetche (approved):

PR #10488, opened 2022-12-15
[ITS3, ITS] Add possibility to switch ITS3 version with configKeyValues

fgrosa:


Detectors/ITSMFT/ITS/simulation/src/Detector.cxx
```diff
@@ -145,7 +145,10 @@ Detector::Detector(Bool_t active, TString name)
     mDescriptorIB.reset(new DescriptorInnerBarrelITS2(3));
   } else if (name == "IT3") {
 #ifdef ENABLE_UPGRADES
-    mDescriptorIB.reset(new DescriptorInnerBarrelITS3(DescriptorInnerBarrelITS3::ThreeLayersNoDeadZones));
+    mDescriptorIB.reset(new DescriptorInnerBarrelITS3());
+    if (its3Version != "") {
+      dynamic_cast<DescriptorInnerBarrelITS3*>(mDescriptorIB.get())->setVersion(its3Version.Data());
```
mconcas: Out of curiosity, why a `dynamic_cast` here?

mconcas (dismissed):

Detectors/ITSMFT/ITS/simulation/src/Detector.cxx
fgrosa: There is no specific reason indeed, so I changed it (and the others) and tested it that it works properly using simple casts as well.

iouribelikov (dismissed):

mconcas:
@shahor02 error seems unrelated and not genuine, if you could merge, pls.

mconcas (dismissed):

fgrosa:
@shahor02 all the checks are now green and @mconcas and @iouribelikov approved already, so can it be merged? Many thanks!

PR #10497, opened 2022-12-16
DPL: add check for missing outputs

ktf:
This will check for missing outputs of lifetime Timeframe and complain if that's actually the case. While this should work as expected and across outputs sent over multiple channels, there might still be some false positives. In particular:

* With a custom completion policy, it might be ok to consume input as they come and produce only the outputs associated to those inputs.
* You could be producing QA data with a timer and timeframe data on real inputs. When creating the QA data, this will complain.

### DIFF
--- a/Framework/Core/include/Framework/DataSender.h
+++ b/Framework/Core/include/Framework/DataSender.h
@@ -37,6 +37,16 @@ class DataSender
              SendingPolicy const& policy);
   void send(fair::mq::Parts&, ChannelIndex index);
   std::unique_ptr<fair::mq::Message> create(RouteIndex index);
+  /// Reset the datasender to a clean state
+  /// so that we can track what whill be sent in the next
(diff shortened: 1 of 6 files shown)


ktf:
@davidrohr this should do what you requested. As I wrote in the commit, there might still be false positives, and for that reason this just produced a "WARNING" message. I am now checking the tests which trigger this behaviour to see if I missed  some other corner case or if they are effectively not producing all the outputs.

davidrohr:
We need to test with the fst. For false positives: i'd disable it when there is no lifetime timeframe input, and for custom completion policies.

Kind Regards
David Rohr

ktf:
Yes, although most likely those are the places were we actually have troubles right now. Maybe I should have a better display in the GUI for the data processors which do not use the default completion policy...

ktf:
This entered in #10238. Closing.

PR #10500, opened 2022-12-17
TPC: adding number of clusters as type for integration

matthias-kleiner:
- Using the number of clusters instead of the cluster charge, the resulting factorized currents are independent of the gain (possible gain fluctuations)

- setting severity to debug

- adding default move constructor


wiechula (approved):

matthias-kleiner:
Can this be merged?

PR #10501, opened 2022-12-18
[RFC] TRD: adding PID workflow

f3sch:
This introduces a new workflow 'o2-trd-pid-workflow', which provides PID information based currently on ML models using the ONNXRuntime.

Should not get merged before at least the default model is in the ccdb.

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/PID.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/PID.h
@@ -1,0 +1,71 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// @file  PID.h
+/// @author Felix Schlepper
+
+#ifndef TRD_PID_H
+#define TRD_PID_H
+
+#include <array>
+#include <unordered_map>
+#include <string>
+
+namespace o2
+{
+namespace trd
+{
+
+/// Option for available PID policies.
+enum class PIDPolicy : unsigned int {
+  // Classical Algorithms
+  LQ1D = 0, ///< 1-Dimensional Likelihood model
+  LQ3D,     ///< 3-Dimensional Likelihood model
+
+  // ML models
+  XGB, ///< XGBOOST
+
+  // Do not anything after this!
+  NMODELS,        ///< Count of all models
+  Test,           ///< Load object for testing
+  DEFAULT = Test, ///< The default option
+};
+
+/// Transform PID policy from string to enum.
+static const std::unordered_map<std::string, PIDPolicy> PIDPolicyString{
+  // Classical Algorithms
+  {"LQ1D", PIDPolicy::LQ1D},
+  {"LQ3D", PIDPolicy::LQ3D},
+
+  // ML models
+  {"XGB", PIDPolicy::XGB},
+
+  // General
+  {"TEST", PIDPolicy::Test},
+  // Default
+  {"default", PIDPolicy::DEFAULT},
+};
+
+/// Transform PID policy from string to enum.
+static const char* PIDPolicyEnum[] = {
+  "LQ1D",
+  "LQ3D",
+  "XGBoost",
+  "SIZE",
+  "Test",
+  "default(=TODO)"};
+
+using PIDValue = float;
+
+} // namespace trd
+} // namespace o2
+
+#endif
--- a/Detectors/AOD/src/aod-producer-workflow.cxx
+++ b/Detectors/AOD/src/aod-producer-workflow.cxx
@@ -36,6 +36,7 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
     {"disable-root-output", o2::framework::VariantType::Bool, false, {"disable root-files output writer"}},
     {"disable-mc", o2::framework::VariantType::Bool, false, {"disable MC propagation"}},
     {"disable-secondary-vertices", o2::framework::VariantType::Bool, false, {"disable filling secondary vertices"}},
+    {"disable-trd-pid", o2::framework::VariantType::Bool, true, {"disable trd pid"}},
     {"info-sources", VariantType::String, std::string{GID::ALL}, {"comma-separated list of sources to use"}},
     {"configKeyValues", VariantType::String, "", {"Semicolon separated key=value strings ..."}},
     {"combine-source-devices", o2::framework::VariantType::Bool, false, {"merge DPL source devices"}},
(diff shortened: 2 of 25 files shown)


Detectors/AOD/src/aod-producer-workflow.cxx
```diff
@@ -36,6 +36,7 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
     {"disable-root-output", o2::framework::VariantType::Bool, false, {"disable root-files output writer"}},
     {"disable-mc", o2::framework::VariantType::Bool, false, {"disable MC propagation"}},
     {"disable-secondary-vertices", o2::framework::VariantType::Bool, false, {"disable filling secondary vertices"}},
+    {"disable-trd-pid", o2::framework::VariantType::Bool, true, {"disable trd pid"}},
```
shahor02, line 39: The default of bool option should be false, please change this to `enable-trd-pid`

shahor02 (changes requested):

DataFormats/Detectors/TRD/include/DataFormatsTRD/PID.h
```diff
@@ -0,0 +1,71 @@
...
+  {"XGB", PIDPolicy::XGB},
+
+  // General
+  {"TEST", PIDPolicy::Test},
+  // Default
+  {"default", PIDPolicy::DEFAULT},
+};
+
+/// Transform PID policy from string to enum.
+static const char* PIDPolicyEnum[] = {
```
martenole, line 58: I wonder why this is needed. Don't you have the information in the map above?

Detectors/TRD/workflow/src/trd-pid-workflow.cxx
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  // input
+  auto maskClusters = GTrackID::getSourcesMask("TRD");
+  auto maskTracks = srcTRD | GTrackID::getSourcesMask("TPC-TRD"); // minimum TRD track
+  if (GTrackID::includesDet(GTrackID::DetID::ITS, srcTRD)) {
+    maskTracks |= GTrackID::getSourcesMask("ITS-TPC-TRD");
+  }
+  auto maskMatches = GTrackID::getSourcesMask(GTrackID::NONE);
```
martenole, line 96: don't you use the maskMatches in the InputHelper?

Detectors/GlobalTrackingWorkflow/helpers/src/InputHelper.cxx
```diff
@@ -207,3 +208,18 @@ int InputHelper::addInputSpecsIRFramesITS(const o2::framework::ConfigContext& co
...
   return 0;
 }
+
+// attach trd pid reader
+int InputHelper::addInputSpecsTRDPID(const o2::framework::ConfigContext& configcontext, o2::framework::WorkflowSpec& specs, GID::mask_t maskMatches, bool mc)
+{
+  if (configcontext.options().get<bool>("disable-root-input")) {
+    return 0;
+  }
+  if (maskMatches[GID::TPCTRD]) {
```
martenole, line 218: Maybe better to use `maskTracks` instead of `maskMathes`? See comment below

Detectors/TRD/pid/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+# This software is distributed under the terms of the GNU General Public
+# License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+#
+# In applying this license CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+o2_add_library(TRDPID
+               SOURCES src/PIDBase.cxx
+               SOURCES src/ML.cxx
```
martenole, line 14: no need for `SOURCES` in each line, only in the first

Detectors/TRD/pid/include/TRDPID/ML.h
```diff
@@ -0,0 +1,97 @@
...
+  /// Return the electron likelihood.
+  /// Different models have different ways to return the probability.
+  virtual PIDValue getELikelihood(const std::vector<Ort::Value>& tensorData) const noexcept = 0;
+
+  /// Fetch a ML model from the ccdb via its binding
+  std::string fetchModelCCDB(o2::framework::ProcessingContext& pc, const char* binding) const;
+
+  /// Calculate pid value
+  template <bool isTPCTRD>
+  PIDValue calculate(const TrackTRD& trkTRD);
```
martenole, line 54: Is there a reason to use a template parameter instead of a function one?

Detectors/TRD/pid/macros/ccdbModelUpload.C
```diff
@@ -0,0 +1,81 @@
...
+void ccdbModelUpload(std::string inFileName, std::string ccdbPath = "TRD_test/PID/xgb")
+{
+
+  o2::ccdb::CcdbApi ccdb;
+  // ccdb.init("http://alice-ccdb.cern.ch");
+  // ccdb.init("http://localhost:8080");
+  ccdb.init("http://ccdb-test.cern.ch:8080");
+  // ccdb.init("http://o2-ccdb.internal");
+  std::map<std::string, std::string> metadata;
+  metadata["UploadedBy"] = "felix";
```
martenole, line 41: Better to use full name + mail address

Detectors/TRD/pid/macros/ccdbModelUpload.C
```diff
@@ -0,0 +1,81 @@
...
+  metadata["Description"] = "onnx model for trd pid";
+  metadata["default"] = "true"; // tag default objects
```
martenole, line 43: Better set to `false`, or adjust time range below (should be 0..infinite for default objects)

Detectors/TRD/pid/src/ML.cxx
```diff
@@ -0,0 +1,202 @@
...
+    LOG(error) << "Error running model inference, using defaults: " << e.what();
+    // fill with negative elikelihood means no information
+    return -1.f;
+  }
+}
+
+template <bool isTPCTRD>
+std::vector<float> ML::prepareModelInput(const TrackTRD& trkTRD)
+{
+  // input is [slope0, slope1, ..., slope5, charge0.0, charge0.1, charge0.2, charge1.0, ..., charge5.3, p]
```
martenole, line 143:
```suggestion
  // input is [slope0, slope1, ..., slope5, charge0.0, charge0.1, charge0.2, charge1.0, ..., charge5.2, p]
```

Detectors/TRD/pid/src/ML.cxx
```diff
@@ -0,0 +1,202 @@
...
+    auto trklt = mTrackletsRaw[trkltId];
+    auto slope = mTrackletsRaw[trkltId].getSlopeBinSigned();
+    auto q0 = mTrackletsRaw[trkltId].getQ0();
+    auto q1 = mTrackletsRaw[trkltId].getQ1();
+    auto q2 = mTrackletsRaw[trkltId].getQ2();
+
+    // TODO handel padrow crossing e.g. z-row merging
+
+    in[iLayer] = slope;
+    in[NLAYER + iLayer * 3 + 0] = q0;
```
martenole, line 176: I think we have the number of charge windows in our constants

martenole (commented):
Some small comments beforehand. Now its better we talk :)

martenole:
Ok, so as we just discussed it is probably better to add the PID workflow device directly to Detectors/TRD/workflow/src/trd-tracking-workflow.cxx and have it there as an option. The track writer can then optionally also put the vector with PID values into the output.
And it would be good to add a dummy PID workflow (that does not even need to rely on any CCDB object) which just puts -1 for all tracks. Then this could already be enabled in the async reco in case there are problems with the ML model.

PR #10505, opened 2022-12-20
ITS Calib: compatibility with data produced with IB commissioning tools

iravasen:
This PR allows to process data on disk obtained with CRU_ITS -> IB commissioning tools (https://gitlab.cern.ch/alice-its-wp10-firmware/CRU_ITS/-/blob/production/software/py/ib_tools/README.md). 

The main changes here are due to the different format of the calibration word which is generated with those tools and to the different number of injections per charge (25). 

Flag `--enable-cru-its` allow to activate the processing of this data while the option `--ninj` allow to set the number of injections per charge point which is 50 by default.


iravasen:
Hi @shahor02, can I kindly ask you to merge this PR when you have 2s, please? Thanks a lot in advance.

mconcas (approved):

PR #10506, opened 2022-12-20
[QC-898] Rename RAW channel type into RAWFMQ

Barthelemy:
RAW was clashing with ioctl_compat.h on Mac, in certain cases.

If you try to open the file I attach with a `TBrowser` and navigate down inside the detector directory, it crashes on Mac.
[results.root.zip](https://github.com/AliceO2Group/AliceO2/files/10267051/results.root.zip)


Barthelemy:
@ktf  could you review ?

ktf:
Note for self, the issue is that ioctl_compat.h redefines (via #define) the "RAW" keyword. Such a behavior breaks things even if I am using enum struct.

PR #10508, opened 2022-12-20
TRD: Adding PID

f3sch:
This adds PID calculations directly to the trd-tracking workflow without creating other DPL devices.
As discussed with @shahor02 the PID value is now part of TrackTRD.
A Dummy object is provided which returns always -1.f as suggested by @martenole.
This is currently the default since no objects are currently in alice-ccdb.

Supersedes #10501

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/PID.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/PID.h
@@ -1,0 +1,74 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// @file  PID.h
+/// @author Felix Schlepper
+
+#ifndef TRD_PID_H
+#define TRD_PID_H
+
+#include <array>
+#include <unordered_map>
+#include <string>
+
+namespace o2
+{
+namespace trd
+{
+
+/// Option for available PID policies.
+enum class PIDPolicy : unsigned int {
+  // Classical Algorithms
+  LQ1D = 0, ///< 1-Dimensional Likelihood model
+  LQ3D,     ///< 3-Dimensional Likelihood model
+
+  // ML models
+  XGB, ///< XGBOOST
+
+  // Do not add anything after this!
+  NMODELS,         ///< Count of all models
+  Test,            ///< Load object for testing
+  Dummy,           ///< Dummy object outputting -1.f
+  DEFAULT = Dummy, ///< The default option
+};
+
+/// Transform PID policy from string to enum.
+static const std::unordered_map<std::string, PIDPolicy> PIDPolicyString{
+  // Classical Algorithms
+  {"LQ1D", PIDPolicy::LQ1D},
+  {"LQ3D", PIDPolicy::LQ3D},
+
+  // ML models
+  {"XGB", PIDPolicy::XGB},
+
+  // General
+  {"TEST", PIDPolicy::Test},
+  {"DUMMY", PIDPolicy::Dummy},
+  // Default
+  {"default", PIDPolicy::DEFAULT},
+};
+
+/// Transform PID policy from string to enum.
+static const char* PIDPolicyEnum[] = {
+  "LQ1D",
+  "LQ3D",
+  "XGBoost",
+  "NMODELS",
+  "Test",
+  "Dummy",
+  "default(=TODO)"};
+
+using PIDValue = float;
+
+} // namespace trd
+} // namespace o2
+
+#endif
--- a/Detectors/TRD/pid/src/ML.cxx
+++ b/Detectors/TRD/pid/src/ML.cxx
@@ -1,0 +1,202 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file ML.cxx
+/// \author Felix Schlepper
+
+#include "TRDPID/ML.h"
+#include "DataFormatsTRD/Constants.h"
+#include "DataFormatsTRD/Tracklet64.h"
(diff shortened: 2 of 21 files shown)


DataFormats/Detectors/TRD/include/DataFormatsTRD/PID.h
```diff
@@ -0,0 +1,74 @@
...
+/// Option for available PID policies.
+enum class PIDPolicy : unsigned int {
+  // Classical Algorithms
+  LQ1D = 0, ///< 1-Dimensional Likelihood model
+  LQ3D,     ///< 3-Dimensional Likelihood model
+
+  // ML models
+  XGB, ///< XGBOOST
+
+  // Do not anything after this!
```
martenole:
```suggestion
  // Do not add anything after this!
```

DataFormats/Detectors/TRD/include/DataFormatsTRD/PID.h
```diff
@@ -0,0 +1,74 @@
...
+  // Default
+  {"default", PIDPolicy::DEFAULT},
+};
+
+/// Transform PID policy from string to enum.
+static const char* PIDPolicyEnum[] = {
+  "LQ1D",
+  "LQ3D",
+  "XGBoost",
+  "SIZE",
```
martenole:
```suggestion
  "NMODELS",
```

Detectors/TRD/pid/src/ML.cxx
```diff
@@ -0,0 +1,202 @@
...
+
+template <bool isTPCTRD>
+std::vector<float> ML::prepareModelInput(const TrackTRD& trkTRD, const o2::globaltracking::RecoContainer& inputTracks)
+{
+  // input is [slope0, slope1, ..., slope5, charge0.0, charge0.1, charge0.2, charge1.0, ..., charge5.2, p]
+  std::vector<float> in(mInputShapes[0][1]);
+  const auto& trackletsRaw = inputTracks.getTRDTracklets();
+  // std::fill(in.begin(), in.end(), 1.f);
+  auto id = trkTRD.getRefGlobalTrackId();
+  in[24] = trkTRD.getP();
```
martenole:
```suggestion
  in.back() = trkTRD.getP();
```
I just don't like the 24 :)

Detectors/TRD/pid/src/TRDPIDLinkDef.h
```diff
@@ -0,0 +1,22 @@
...
+
+#ifdef __CLING__
+
+#pragma link off all globals;
+#pragma link off all classes;
+#pragma link off all functions;
+
+#pragma link C++ class o2::trd::PIDBase + ;
+#pragma link C++ class o2::trd::TRDPIDParams + ;
+#pragma link C++ class o2::conf::ConfigurableParamHelper < o2::trd::TRDPIDParams> + ;
```
martenole, line 22: I think you are missing LinkDefs for `ML` and `Dummy`

Detectors/TRD/workflow/io/include/TRDWorkflowIO/TRDTrackReaderSpec.h
```diff
@@ -15,6 +15,7 @@
 /// @file   TRDTrackReaderSpec.h
 
 #include "DataFormatsTRD/TrackTRD.h"
+#include "DataFormatsTRD/PID.h"
```
martenole: why is this needed?

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -21,11 +21,12 @@
 #include "DataFormatsTRD/CalibratedTracklet.h"
 #include "DataFormatsTRD/TriggerRecord.h"
 #include "DataFormatsTRD/Constants.h"
+#include "TRDPID/PIDBase.h"
```
martenole: this is already in the header

Detectors/TRD/workflow/src/trd-tracking-workflow.cxx
```diff
@@ -42,14 +43,16 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
 {
   // option allowing to set parameters
   std::vector<o2::framework::ConfigParamSpec> options{
-    {"disable-mc", o2::framework::VariantType::Bool, false, {"Disable MC labels"}},
-    {"disable-root-input", o2::framework::VariantType::Bool, false, {"disable root-files input readers"}},
-    {"disable-root-output", o2::framework::VariantType::Bool, false, {"disable root-files output writers"}},
-    {"enable-trackbased-calib", o2::framework::VariantType::Bool, false, {"enable calibration devices which are based on tracking output"}},
-    {"enable-qc", o2::framework::VariantType::Bool, false, {"enable tracking QC"}},
+    {"disable-mc", VariantType::Bool, false, {"Disable MC labels"}},
+    {"disable-pid", VariantType::Bool, false, {"Disable PID"}},
```
martenole: could you change this here to `enable-pid` leaving the default `false` please? That way we don't need to update all steering scripts which generate workflows and add the `--disable-pid` flag there. When we want to have this in production we can then just add `--enable-pid`.

GPU/GPUTracking/DataTypes/GPUTRDTrack.h
```diff
@@ -131,6 +133,7 @@ class GPUTRDTrack_t : public T
 
  protected:
   float mChi2;                       // total chi2.
+  float mSignal{-1.f};               // electron Likelihood for track
```
martenole, line 136: Please add `mSignal` also to the relevant methods in GPUTRDTrack.cxx. You can search for `mChi2` and add it everywhere where that is

martenole (commented):
Thanks, looks good! I have mostly minor comments

Detectors/TRD/workflow/io/include/TRDWorkflowIO/TRDTrackReaderSpec.h
f3sch: It is not, slipped by.

Detectors/TRD/pid/src/ML.cxx
f3sch: Sure, although keep in mind that everything under pid/ should be regarded as preliminary as I do not yet know what the input _really_ will be.
This is mostly all to abstract away from more relevant code.

martenole (approved):

martenole:
The changes in the AOD producer are minor (utilizing 2 unused bits in the TRD pattern as flags and setting the TRD signal which is still dummy but was before not set at all).
Are there any objections? If not, I would merge this tomorrow

bazinski (approved):

jgrosseo:
The changes in the data model are fine for me (I cannot judge the rest). I propose to add a TRD enum to DataTypes.h which has values for the different bits. Also 2 dynamic columns which query then this information would be useful. As example you can check line 238 (in the diff): https://github.com/AliceO2Group/AliceO2/pull/10508/files#diff-6410cfe670c62cda55ab26854e52f05a3b75d325cb663e7852d79df3d4bad58eL238

This does not have to be included in this PR for simplicity, if you could do it right away in another PR...

Let me know.

jgrosseo (approved):

martenole:
Merging now. I see that Felix already worked on @jgrosseo request in https://github.com/AliceO2Group/AliceO2/pull/10553

PR #10509, opened 2022-12-20
Fix TRD refit LT-integral calculation and material accounting

shahor02:


shahor02:
Dependence of the L-integral on the outer param X:
before the fix:
trdNoFix
after:
trdFix

noferini:
Hi @shahor02,
many thanks!
Francesco

martenole (dismissed):
Thanks a lot!

GPU/GPUTracking/Definitions/GPUSettingsList.h
```diff
@@ -97,6 +97,7 @@ AddOptionRTC(trkltResRPhiIdeal, float, 1.f, "", 0, "Optimal tracklet rphi resolu
 AddOptionRTC(applyDeflectionCut, unsigned char, 0, "", 0, "Set to 1 to enable tracklet selection based on deflection")
 AddOptionRTC(stopTrkAfterNMissLy, unsigned char, 6, "", 0, "Abandon track following after N layers without a TRD match")
 AddOptionRTC(useExternalO2DefaultPropagator, unsigned char, 0, "", 0, "Use the default instance of the o2::Propagator, instead of the GPU Reconstruciton one with GPU B field")
+AddOptionRTC(matCorrType, int, 2, "", 0, "Material correction to use: 0 - none, 1 - TGeo, 2 - matLUT")
```
martenole: Ah, I think it would be better to use an `unsigned char` here instead, or move the option up above the first `char` one. Maybe @davidrohr can confirm, but if I remember correctly then the order of the options here is not arbitrary because of the alignment in memory.

shahor02: sure, changed to uint.

martenole (approved):

PR #10514, opened 2022-12-27
CTP: trigger offsets

lietava:
Trigger offsets are to be applied both for interaction record (inputs) and trigger class record.
For LM inputs offset should be zero (although in PbPB some inputs eg TVX were shifted by 1)
For L0/L1 inputs we need to agree if to shift them to LM or do nothing , e.g. leave it for analysers
For trigger class to be shifted to LM , the shift should be LM-L0 + L0-L1 + internal offset.
Internal offsets is due s to CTP firmware . It is couple of BC - I will find it.

For test run:
o2-raw-tf-reader-workflow -b --onlyDet CTP --input-data list0_19.txt  | o2-ctp-reco-workflow -b  --use-verbose-mode  --configKeyValues "TriggerOffsetsParam.customOffset[16]=-1"

### DIFF
--- a/DataFormats/common/include/CommonDataFormat/InteractionRecord.h
+++ b/DataFormats/common/include/CommonDataFormat/InteractionRecord.h
@@ -198,7 +198,7 @@ struct InteractionRecord {
     // bc self-addition operator, no check for orbit wrap
     auto l = toLong() + dbc;
     if (l < 0) {
-      l += InteractionRecord(3564, 0xffffffff).toLong();
+      l += InteractionRecord(DummyBC, MaxGlobalBCs + 1ll).toLong();
     }
     bc = l % o2::constants::lhc::LHCMaxBunches;
     orbit = l / o2::constants::lhc::LHCMaxBunches;
--- a/DataFormats/common/include/CommonDataFormat/InteractionRecord.h
+++ b/DataFormats/common/include/CommonDataFormat/InteractionRecord.h
@@ -197,6 +197,9 @@ struct InteractionRecord {
   {
     // bc self-addition operator, no check for orbit wrap
     auto l = toLong() + dbc;
+    if(l < 0) {
+      l += InteractionRecord(3564,0xffffffff).toLong();
+    }
     bc = l % o2::constants::lhc::LHCMaxBunches;
     orbit = l / o2::constants::lhc::LHCMaxBunches;
     return *this;
--- a/DataFormats/common/include/CommonDataFormat/InteractionRecord.h
+++ b/DataFormats/common/include/CommonDataFormat/InteractionRecord.h
@@ -198,7 +198,7 @@ struct InteractionRecord {
     // bc self-addition operator, no check for orbit wrap
     auto l = toLong() + dbc;
     if (l < 0) {
-      l += InteractionRecord(DummyBC, MaxGlobalBCs + 1ll).toLong();
+      l += MaxGlobalBCs;
     }
     bc = l % o2::constants::lhc::LHCMaxBunches;
     orbit = l / o2::constants::lhc::LHCMaxBunches;
--- a/DataFormats/common/include/CommonDataFormat/InteractionRecord.h
+++ b/DataFormats/common/include/CommonDataFormat/InteractionRecord.h
@@ -197,8 +197,8 @@ struct InteractionRecord {
   {
     // bc self-addition operator, no check for orbit wrap
     auto l = toLong() + dbc;
-    if(l < 0) {
-      l += InteractionRecord(3564,0xffffffff).toLong();
+    if (l < 0) {
+      l += InteractionRecord(3564, 0xffffffff).toLong();
     }
     bc = l % o2::constants::lhc::LHCMaxBunches;
     orbit = l / o2::constants::lhc::LHCMaxBunches;
--- a/Detectors/Base/include/DetectorsBase/CTFCoderBase.h
+++ b/Detectors/Base/include/DetectorsBase/CTFCoderBase.h
@@ -34,7 +34,7 @@
 
 namespace o2
 {
-namespace frameworkT
+namespace framework
 {
 class ProcessingContext;
 }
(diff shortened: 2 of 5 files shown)


lietava:
@shahor02 : Using TriggerOffsetsParams I factorised offset tp LM-L0+L0+L1 part and customOffset part . customOffsetPart 
 is 'only' for correction, i.e. for inputs it should be 0 (if CTP correctly configures), for trigger classes let us see.

Detectors/Base/include/DetectorsBase/CTFCoderBase.h
```diff
@@ -34,7 +34,7 @@
 
 namespace o2
 {
-namespace framework
+namespace frameworkT
```
shahor02: leftover from some test?

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
```diff
@@ -257,29 +257,45 @@ int RawDecoderSpec::addCTPDigit(uint32_t linkCRU, uint32_t triggerOrbit, gbtword
...
+  int32_t bcid = (diglet & bcidmask).to_ulong();
   // LOG(info) << bcid << "    pld:" << pld;
   o2::InteractionRecord ir;
-  ir.orbit = triggerOrbit;
-  ir.bc = bcid;
-  digit.intRecord = ir;
+  int32_t BCShiftCorrection = o2::ctp::TriggerOffsetsParam::Instance().customOffset[o2::detectors::DetID::CTP];
   if (linkCRU == o2::ctp::GBTLinkIDIntRec) {
     LOG(debug) << "InputMaskCount:" << digits[ir].CTPInputMask.count();
+    if (bcid > BCShiftCorrection) {
```
shahor02: what if BCShiftCorrection is negative and `bcid - BCShiftCorrection` exceeds 3564?
Use simply 
```
if (it.toLong()>BCShiftCorrection) {
  ir -= BCShiftCorrection;
} else {
  ir = {0,0}; // normally should not happen
}
```

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
```diff
@@ -257,29 +257,45 @@ int RawDecoderSpec::addCTPDigit(uint32_t linkCRU, uint32_t triggerOrbit, gbtword
...
-        LOG(debug) << bcid << " inputs bcid vase 1 orbit " << triggerOrbit << " pld:" << pld;
+        LOG(debug) << bcid << " inputs bcid case 1 orbit " << triggerOrbit << " pld:" << pld;
       } else {
         LOG(error) << "Two CTP IRs with the same timestamp:" << ir.bc << " " << ir.orbit;
       }
     } else {
       LOG(error) << "Two digits with the same rimestamp:" << ir.bc << " " << ir.orbit;
     }
   } else if (linkCRU == o2::ctp::GBTLinkIDClassRec) {
+    uint32_t offset = BCShiftCorrection + o2::ctp::TriggerOffsetsParam::Instance().LM_L0 + o2::ctp::TriggerOffsetsParam::Instance().L0_L1 - 1;
```
shahor02, line 286: Idem for the checks.

shahor02 (changes requested):
Hi @lietava 
The `customOffset` was actually meant for fixups only. See some comments below.

Detectors/Base/include/DetectorsBase/CTFCoderBase.h
lietava: Just accidental typo. Didnt notice when pushing, sorry.

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
lietava: I see your point although not sure what do you suggest.
I implemented now general formula.

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
lietava: yes

lietava:
> Hi @lietava The `customOffset` was actually meant for fixups only. See some comments below.



> Hi @lietava The `customOffset` was actually meant for fixups only. See some comments below.

comments are ok, I fixed. 
'customOffset' I also understand only for fixups which is PbPB case.
For correctly aligned CTP I expect customOffset = 0.

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
shahor02: I mean the InteractionRecord class has += and -= operators (the argument is BC unit), one should just watch that when adding negative number the result (as ulong abs BC) remain positive since otherwise it will wrap to large value

lietava: ok, I see now. So I used +=. 
Also I added correction for negative orbit/bcid to:
 InteractionRecord& operator+=(int64_t dbc)
https://github.com/lietava/AliceO2/blob/58055cd4486be4cef7125b0330582ef0644a51b8/DataFormats/common/include/CommonDataFormat/InteractionRecord.h#L200
I know in practice it may never happen. So it can be removed if there are other reasons not to have it.

lietava:
@shahor02 : do you have any comments here ? Can I go for review ?

shahor02:
Yes, I've removed the `draft` tag

lietava:
@shahor02 : there is error in ctp-entropy-encoder

DataFormats/common/include/CommonDataFormat/InteractionRecord.h
```diff
@@ -197,6 +197,9 @@ struct InteractionRecord {
   {
     // bc self-addition operator, no check for orbit wrap
     auto l = toLong() + dbc;
+    if (l < 0) {
+      l += MaxGlobalBCs;
+    }
```
shahor02, line 202: this is not correct, please suppress this patch, I've opened a PR https://github.com/AliceO2Group/AliceO2/pull/10528 with a proper check of wrapping for all operations.

shahor02 (changes requested):

DataFormats/common/include/CommonDataFormat/InteractionRecord.h
lietava: PLease, see my comment in 10528.

PR #10516, opened 2022-12-30
DPL Analysis: simplify Join and update Concat

aalkin:
* simplifies Join type computation and makes it universal
* updates Concat type computation and makes it nestable
* both Concat and Join can now be nested within each other


ktf:
Also tested on hyperloop.

PR #10521, opened 2023-01-03
add general/static methods to deal with matching pair candidate

noferini:
3 methods moved to static to be used also outside for development studies (in principle no changes in the current workflow).
2 new static metods added (mainly grouppingMatch) but not yet used.

The purpose is to provide some tools to investigate potential optimization in the matching fot a high occupancy environment: not selecting only accordingly to the Chi2 of a single matching pair (track, TOF cluster) but also considering correlation between tracks (clusters) sharing different matchable clusters(tracks).

WIP till the reconstruction for skimming is frozen...

### DIFF
--- a/Detectors/GlobalTracking/include/GlobalTracking/MatchTOF.h
+++ b/Detectors/GlobalTracking/include/GlobalTracking/MatchTOF.h
@@ -171,7 +171,7 @@ class MatchTOF
   }
 
   void setFIT(bool value = true) { mIsFIT = value; }
-  int findFITIndex(int bc);
+  static int findFITIndex(int bc, const gsl::span<const o2::ft0::RecPoints>& FITRecPoints);
 
   void checkRefitter();
   bool makeConstrainedTPCTrack(int matchedID, o2::dataformats::TrackTPCTOF& trConstr);
@@ -194,6 +194,9 @@ class MatchTOF
   void setTS(unsigned long creationTime) { mTimestamp = creationTime; }
   unsigned long getTS() const { return mTimestamp; }
 
+  static void grouppingMatch(std::vector<o2::dataformats::MatchInfoTOFReco> origin, std::vector<std::vector<o2::dataformats::MatchInfoTOFReco>>& groupped);
+  static void printGroupping(const std::vector<o2::dataformats::MatchInfoTOFReco>& origin, const std::vector<std::vector<o2::dataformats::MatchInfoTOFReco>>& groupped);
+
  private:
   bool prepareFITData();
   int prepareInteractionTimes();
@@ -209,7 +212,8 @@ class MatchTOF
   void doMatching(int sec);
   void doMatchingForTPC(int sec);
   void selectBestMatches();
-  void selectBestMatchesHP();
+  static void BestMatches(std::vector<o2::dataformats::MatchInfoTOFReco>& matchedTracksPairs, std::vector<o2::dataformats::MatchInfoTOF>* matchedTracks, std::vector<int>* matchedTracksIndex, int* matchedClustersIndex, const gsl::span<const o2::ft0::RecPoints>& FITRecPoints, const std::vector<Cluster>& TOFClusWork, const std::vector<matchTrack>* TracksWork, std::vector<o2::dataformats::CalibInfoTOF>& CalibInfoTOF, unsigned long Timestamp, bool MCTruthON, const o2::dataformats::MCTruthContainer<o2::MCCompLabel>* TOFClusLabels, const std::vector<o2::MCCompLabel>* TracksLblWork, std::vector<o2::MCCompLabel>* OutTOFLabels, float calibMaxChi2);
+  static void BestMatchesHP(std::vector<o2::dataformats::MatchInfoTOFReco>& matchedTracksPairs, std::vector<o2::dataformats::MatchInfoTOF>* matchedTracks, std::vector<int>* matchedTracksIndex, int* matchedClustersIndex, const gsl::span<const o2::ft0::RecPoints>& FITRecPoints, const std::vector<Cluster>& TOFClusWork, std::vector<o2::dataformats::CalibInfoTOF>& CalibInfoTOF, unsigned long Timestamp, bool MCTruthON, const o2::dataformats::MCTruthContainer<o2::MCCompLabel>* TOFClusLabels, const std::vector<o2::MCCompLabel>* TracksLblWork, std::vector<o2::MCCompLabel>* OutTOFLabels);
   bool propagateToRefX(o2::track::TrackParCov& trc, float xRef /*in cm*/, float stepInCm /*in cm*/, o2::track::TrackLTIntegral& intLT);
   bool propagateToRefXWithoutCov(o2::track::TrackParCov& trc, float xRef /*in cm*/, float stepInCm /*in cm*/, float bz);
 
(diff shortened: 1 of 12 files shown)


noferini:
Ready for the review (not urgent)

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -207,6 +235,86 @@ int MatchTOF::prepareInteractionTimes()
...
+  for (const auto& gr : grouped) {
+    ngroups++;
+    printf("Group %d\n", ngroups);
+    for (const auto& seed : gr) {
+      printf("Pair: track=%d TOFcl=%d\n", seed.getIdLocal(), seed.getTOFClIndex());
+    }
+  }
+}
+//______________________________________________
+void MatchTOF::groupingMatch(std::vector<o2::dataformats::MatchInfoTOFReco> origin, std::vector<std::vector<o2::dataformats::MatchInfoTOFReco>>& grouped, std::vector<std::vector<int>>& firstEls, std::vector<std::vector<int>>& secondEls)
```
shahor02: You are passing `origin` by copy, i.e. the results of its modification are lost. Is this what you want?
This may create some overheads.

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -207,6 +235,86 @@ int MatchTOF::prepareInteractionTimes()
...
+    auto& trkId = firstEls.emplace_back(dummyInt);
+    auto& cluId = secondEls.emplace_back(dummyInt);
+
+    // first element is the seed
+    const auto& seed = origin[0];
+    trkId.push_back(seed.getIdLocal());
+    cluId.push_back(seed.getTOFClIndex());
+    // remove element added to the current group
+    grouped[pos].push_back(seed);
+    origin.erase(origin.begin());
```
shahor02: `vector.erase` is very expensive method, can't you avoid it, e.g. by using some flags? If not, then, since you are always erasing the 1st element, it is better to use std::dequeu, which has pop_front method.

Detectors/GlobalTracking/include/GlobalTracking/MatchTOF.h
```diff
@@ -194,6 +196,11 @@ class MatchTOF
   void setTS(unsigned long creationTime) { mTimestamp = creationTime; }
   unsigned long getTS() const { return mTimestamp; }
 
+  static void groupingMatch(std::vector<o2::dataformats::MatchInfoTOFReco> origin, std::vector<std::vector<o2::dataformats::MatchInfoTOFReco>>& grouped, std::vector<std::vector<int>>& firstEls, std::vector<std::vector<int>>& secondEls);
```
shahor02: passing `origin` by copy, see more comments in the cxx file.

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -1065,17 +1173,17 @@ void MatchTOF::doMatchingForTPC(int sec)
...
+  if (FITRecPoints.size() == 0) {
     return -1;
   }
 
   int index = -1;
   int distMax = 5; // require bc distance below 5
 
-  for (int i = 0; i < mFITRecPoints.size(); i++) {
-    const o2::InteractionRecord ir = mFITRecPoints[i].getInteractionRecord();
+  for (int i = 0; i < FITRecPoints.size(); i++) {
```
shahor02: this will lead to a warning about comparing signed and unsigned.

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -1093,29 +1201,34 @@ int MatchTOF::findFITIndex(int bc)
...
+}
+//______________________________________________
+void MatchTOF::BestMatches(std::vector<o2::dataformats::MatchInfoTOFReco>& matchedTracksPairs, std::vector<o2::dataformats::MatchInfoTOF>* matchedTracks, std::vector<int>* matchedTracksIndex, int* matchedClustersIndex, const gsl::span<const o2::ft0::RecPoints>& FITRecPoints, const std::vector<Cluster>& TOFClusWork, const std::vector<matchTrack>* TracksWork, std::vector<o2::dataformats::CalibInfoTOF>& CalibInfoTOF, unsigned long Timestamp, bool MCTruthON, const o2::dataformats::MCTruthContainer<o2::MCCompLabel>* TOFClusLabels, const std::vector<o2::MCCompLabel>* TracksLblWork, std::vector<o2::MCCompLabel>* OutTOFLabels, float calibMaxChi2)
+{
   ///< define the track-TOFcluster pair per sector
 
   // first, we sort according to the chi2
-  std::sort(mMatchedTracksPairs.begin(), mMatchedTracksPairs.end(), [this](o2::dataformats::MatchInfoTOFReco& a, o2::dataformats::MatchInfoTOFReco& b) { return (a.getChi2() < b.getChi2()); });
+  std::sort(matchedTracksPairs.begin(), matchedTracksPairs.end(), [](o2::dataformats::MatchInfoTOFReco& a, o2::dataformats::MatchInfoTOFReco& b) { return (a.getChi2() < b.getChi2()); });
```
shahor02: better declare `(const o2::dataformats::MatchInfoTOFReco& a, const o2::dataformats::MatchInfoTOFReco& b)`

Detectors/GlobalTracking/src/MatchTOF.cxx
```diff
@@ -1210,41 +1323,41 @@ void MatchTOF::selectBestMatchesHP()
   std::vector<o2::dataformats::MatchInfoTOFReco> tmpMatch;
 
   // first, we sort according to the chi2
-  std::sort(mMatchedTracksPairs.begin(), mMatchedTracksPairs.end(), [this](o2::dataformats::MatchInfoTOFReco& a, o2::dataformats::MatchInfoTOFReco& b) { return (a.getChi2() < b.getChi2()); });
+  std::sort(matchedTracksPairs.begin(), matchedTracksPairs.end(), [](o2::dataformats::MatchInfoTOFReco& a, o2::dataformats::MatchInfoTOFReco& b) { return (a.getChi2() < b.getChi2()); });
```
shahor02: const args?

Detectors/GlobalTracking/src/MatchTOF.cxx
noferini: Yes, since I need to modify it (e.g. erasing object) without affecting the original vector at least for the moment which is not in production.
In future I could release this condition if I will use the output of the grouping instead of the original one for the next steps.

noferini: I will pass origin as a reference since I changed accordingly to the next comment. Thanks

Detectors/GlobalTracking/src/MatchTOF.cxx
noferini: using flags (unfortunately erase was called also in an internal loop to delete object in random positions)

Detectors/GlobalTracking/include/GlobalTracking/MatchTOF.h
noferini: Sure. Moving to const reference

Detectors/GlobalTracking/src/MatchTOF.cxx
noferini: move to "unsigned int i"

Detectors/GlobalTracking/src/MatchTOF.cxx
noferini: Done, thanks

Detectors/GlobalTracking/src/MatchTOF.cxx
noferini: done

noferini:
@shahor02, thanks for the comments 
All changes were implemented but I need to re-validate the code.
More news asap.

noferini:
One more fix added. I got the same results after all changes are applied.
Waiting for passing checks

shahor02:
Failure is unrelated, merging

PR #10526, opened 2023-01-06
CMake: fix FindO2GPU syntax

aphecetche:
@davidrohr looks like there's a missing parenthesis (at least my CMake of my M1 Mac complains about it ?).


davidrohr:
weird, should not be necessary, since NOT should have precedence over AND/OR.
Perhaps you have additional cmake warnings enabled globally?
Anyway, this doesn't change the syntax, so we can just merge it.

PR #10527, opened 2023-01-06
Fix: revert ITS/MFT cluster dict. hashing to original murmur2 f-n

shahor02:
@mconcas absence of breaks in the switch of hashing f-n was part of the algorithm :)


mconcas:
Was this part of the compiler's warning cleanup?

shahor02:
I guess so: https://github.com/AliceO2Group/AliceO2/pull/8908, anyway, there are no grave consequences, only some extra space is lost on explicit patterns since the hash definition differs from what dictionaries expect.

PR #10528, opened 2023-01-07
Check/fix for wraps in InteracitonRecord arithmetic

shahor02:


lietava (commented):
Hi @shahor02 , it seems to me that my original proposal for wraping was ok  (maybe some 1 somewhere to fix)
but we have different definitions of what 'warping' means.
I tried to explain on example operator +=(int64_t dbc), please, see.

DataFormats/common/include/CommonDataFormat/InteractionRecord.h
```diff
@@ -168,90 +172,112 @@ struct InteractionRecord {
...
+      }
     }
     return tmp;
   }
 
   InteractionRecord& operator+=(int64_t dbc)
   {
-    // bc self-addition operator, no check for orbit wrap
-    auto l = toLong() + dbc;
+    // bc self-addition operator, avoid wrapping
```
lietava, line 214: It seems we have different view what wraping orbit means. For me if dbc>=0 then nothing is to be done: just adding and calculating new id by / and %.
The motivation is that if somebody tells me any time in BC from given event, I calculate correct event id while for your formula it will give me always MaxGlobalBC.
Practical, still academic , example may be if fill last longer than MaxGlobalBC
The same for negative dbc.
Maybe the simplest example is dbc = -1
{0,0}-1 = {0,0}  (you)
{0,0}-1 = {0xffffffff,3563} (me)
Please, let me know if you have any reason for the way it is defined here

shahor02: Hi @lietava, the typical case when the +/- operation is used and has a risk of overflow is adding some margins to the InteractionRecord. Therefore I consider as an error getting a smaller value due to the positive increment (and vice-versa). But also for the case of shifting the IR due to the CTP bias: consider the standalone runs where the orbit/bc starts from 0/0: there one could easily get situation like `{0,0}-1 = {0xffffffff,3563}`, which is also wrong.
On top of that: the signed integer overflow in c++ is undefined behaviour, so I added in all +/- operations check for the overflow before the actual operation.

lietava: Hi @shahor02 , thanks, we are converging.   Then should not attempt {0,0}-1 and {0xffffffff,3563}+1 (and similar) be classified as error/warning ?

shahor02: Depends: when the increment is done just to define the margin, this is not an error, one should simply assign limit value (i.e. 0 or maxBC). For the case when the increment is supposed to correct the trigger, this would is indeed an error, I think the triggers should be simply discarded, as we do, e.g. in 
https://github.com/AliceO2Group/AliceO2/blob/547f8667b02b723ab9abd536844400d72f406fad/Detectors/TRD/reconstruction/src/CruRawReader.cxx#L458-L470 
or 
https://github.com/AliceO2Group/AliceO2/blob/547f8667b02b723ab9abd536844400d72f406fad/Detectors/TRD/reconstruction/include/TRDReconstruction/CTFCoder.h#L251

lietava: @shahor02 : ok, I am not sure what you mean by margins but ok.
Apart of that in your example
in line https://github.com/AliceO2Group/AliceO2/blob/547f8667b02b723ab9abd536844400d72f406fad/Detectors/TRD/reconstruction/src/CruRawReader.cxx#L458-L470  even with your IR wrap it would lead to wrong BC only if orbit = 0  which would never practically happen.
So I am not sure why you need to discard it ?
I understand orbit would jump by -1 in that case but why we can not keep it ?

shahor02: For the margins: in many places, e.g. AOD skimming, we need to select IRs belonging to IR0-M:IR0+M interval.

For the https://github.com/AliceO2Group/AliceO2/blob/547f8667b02b723ab9abd536844400d72f406fad/Detectors/TRD/reconstruction/src/CruRawReader.cxx#L458-L470: generally speaking orbit starts from 0 in all standalone runs. But in this case, the concern is not 0 but going below the start of the TF: only in this case the trigger is discarded.

lietava: It concerns only RORC detectors and ctp readout. Why we can not save in current TF shifted collision from previous TF ?

shahor02: > It concerns only RORC detectors and ctp readout. 
Only for legitimate CTP shifts, in general it may concern any type of data.

> Why we can not save in current TF shifted collision from previous TF ?
Why do you need it? E.g. in case of TRD data, these trigger cannot be related to other barrel detectors.
In general, the reconstruction refers to all objects time as an offset wrt the start of the TF, I find it weird if these offsets are negative. The same is true for the AOD

lietava: Fine.  How shall I proceed with ctp reconstruction ? Closing #10514 and opening new one ?

shahor02: I would suggest to suppress your modification of InteractionRecord  and add a check to not produce corrected orbit less than the 1st orbit of the TF (discarding the trigger in  this case).

PR #10534, opened 2023-01-09
Simplify code to get tables

ktf:


ktf:
@aalkin as discussed. This is in preparation to have Enumerate as a possible input of the process function.

ktf:
@mconcas could you have a look at why it fails with:
```
312/469 Test #238: Detectors/ITSMFT/common/simulation/test/testAlpideSimResponse.cxx ....................***Failed    2.31 sec
Running 1 test case...
Error in <TFile::TFile>: file /sw/slc8_x86-64/O2/10534-slc8_x86-64-local1/share/Detectors/ITSMFT/data/AlpideResponseData/AlpideResponseData.root does not exist
```

ktf:
Tested on hyperloop. Errors unrelated.

PR #10537, opened 2023-01-10
Allow usage of additional track information for TPC SCD map creation

martenole:
Since we have stored also additional track data with the residuals during the YETS production we can use that to tighten some cuts for the map creation. E.g. in order to use only the residuals which belong to tracks that have 7 ITS clusters one can do
```
o2-tpc-static-map-creator -b --start-from-unbinned --residuals-infiles o2tpc_residuals_1666669765254_1666670365241_1159686_1212398.root --configKeyValues "scdcalib.useTrackData=1;scdcalib.minITSNCls=7"
```
Pinging @aschmah


wiechula (approved):

wiechula:
The errors seem unrelated, can this be merged?

martenole:
Yes, they were unrelated, so I merged now

PR #10538, opened 2023-01-10
PDG Database unittest

ktf:
This is a unit test which makes sure that our own instance of the TDatabasePDG, as used by a dummy analysis task, is actually populated by one of our customisations. I will use this to validate the fact that the plugin provides the same behaviour.


ktf:
@TimoWilken this has been running for over 4 hours:

```
![image](https://user-images.githubusercontent.com/10544/211634603-4e68f97d-029c-4d8f-a56f-9528ef21c04a.png)
```

tomorrow we should check what is going on.

PR #10539, opened 2023-01-10
Refine and add utilities for MCGenStatus

benedikt-voelkel:


benedikt-voelkel:
Do not yet merge. An update of O2DPG is required to guarantee correct functioning of custom generators. I will provide that shortly.

benedikt-voelkel:
Related O2DPG PR is https://github.com/AliceO2Group/O2DPG/pull/814

The failed dataflow CI seems to be unrelated.

benedikt-voelkel:
I propose we merge both this and the linked one at the same time so they go into the same tag.

PR #10540, opened 2023-01-10
Discard TRD tracks with bad quality

martenole:
The minimum number of TRD tracklets per track is configurable and set to 3. This avoids that e.g. ITS-TPC tracks are matched to a single tracklet (which is rather likely to be fake) and then don't have a chance anymore to be matched to TOF.
Also the minimum track quality can be specified, but before restricting this I would like to do some more tests. Therefore max chi2 value is set here to a very high value.

This also implies that downstream we will not have any TRD tracks with less than `nTracketsMin` tracklets, e.g. for vDrift calibration. But I think we are not using tracks with <= 2 tracklets in any case there so it should be no problem.


shahor02 (commented):
Thanks @martenole 
but why not use existing GPUSettingsRecTRD ? In the end, the new params are in the same spirit as `mTracker->Param().rec.trd.chi2StrictCut`

martenole:
True, I think I made it too complicated. I thought that this is not strictly a GPU setting, but still it belongs to the TRD reconstruction of course. So I will move it to the GPUSettingsList

PR #10541, opened 2023-01-11
[O2-3489] DPL Analysis: correctly account for offset when accessing sliced boolean column

aalkin:
@saganatt @ktf 
This should guarantee correct access.


saganatt:
Hi, I confirm that with this fix I get correct results.

PR #10542, opened 2023-01-11
Temporarily remove alpide test and improve ITSRESPONSE location

mconcas:
This is temporary to workaround some sporadic CI tests failures.

It will be reverted after some fixups in the O2 recipe for the AlpideResponse package.

### DIFF
--- a/Detectors/ITSMFT/common/data/AlpideResponseData/CMakeLists.txt
+++ b/Detectors/ITSMFT/common/data/AlpideResponseData/CMakeLists.txt
@@ -18,8 +18,16 @@ o2_add_executable(alpide-response-generator
 
 set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${CMAKE_SOURCE_DIR}/Detectors/ITSMFT/common/data/AlpideResponseData/AlpideResponse.cxx)
 
+if(ITSRESPONSE)
+  message(STATUS "ITSRESPONSE option provided, setting ITSRESPONSE_DIR from it: " ${ITSRESPONSE})
+  set(ITSRESPONSE_DIR ${ITSRESPONSE})
+else()
+  message(STATUS "ITSRESPONSE option not provided, setting ITSRESPONSE_DIR from environment ITSRESPONSE_ROOT: " $ENV{ITSRESPONSE_ROOT})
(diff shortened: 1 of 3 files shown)


mconcas:
@ktf this should be part of what we discussed today concerning O2.
 Can it work?

Detectors/ITSMFT/common/data/AlpideResponseData/CMakeLists.txt
```diff
@@ -18,8 +18,16 @@ o2_add_executable(alpide-response-generator
...
+  message(STATUS "ITSRESPONSE option provided, setting ITSRESPONSE_LOCATION from it: " ${ITSRESPONSE})
+  set(ITSRESPONSE_LOCATION ${ITSRESPONSE})
+else()
+  message(STATUS "ITSRESPONSE option not provided, setting ITSRESPONSE_LOCATION from environment ITSRESPONSE_ROOT: " $ENV{ITSRESPONSE_ROOT})
+  set(ITSRESPONSE_LOCATION $ENV{ITSRESPONSE_ROOT})
+endif()
+
 add_custom_command(TARGET O2exe-alpide-response-generator POST_BUILD
-                   COMMAND ${CMAKE_BINARY_DIR}/stage/bin/o2-alpide-response-generator -i $ENV{ITSRESPONSE_ROOT}/response/AlpideResponseData/ -o ${CMAKE_SOURCE_DIR}/Detectors/ITSMFT/common/data/AlpideResponseData/
+                   COMMAND ${CMAKE_BINARY_DIR}/stage/bin/o2-alpide-response-generator -i ${ITSRESPONSE_LOCATION}/response/AlpideResponseData/ -o ${CMAKE_SOURCE_DIR}/Detectors/ITSMFT/common/data/AlpideResponseData/
```
ktf: I think the common pattern is to call those variables _DIR. @aphecetche ?

aphecetche: if intended to be used with e.g. `find_package(ITSResponse)` `_ROOT` would even be better I believe, as `_DIR` are expected to be env variable, as far as I can tell. But would work as well. See https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure (in particular the part after "The set of installation prefixes is constructed using the following steps" listing 9 steps).

PR #10543, opened 2023-01-11
Print message when ROOT is used to create messages

ktf:


ktf:
@davidrohr @shahor02 this is what we discussed in the meeting. Please do not merge it because it will pollute the processing with warning.

davidrohr:
thx, can you also do a PR for the debug message if env variable is set printing the message size?

shahor02:
@ktf, I run with this PR reco of 1 CTF and did not get any message "being serialised with ROOT..". To me this looks strange, since in the CCDB objects there should be a few non-messageable.

Running with this PR the `stage/tests/o2-test-framework-DataAllocator  -b` I get crash, changing your PR to
```
--- a/Framework/Core/include/Framework/MessageContext.h
+++ b/Framework/Core/include/Framework/MessageContext.h
@@ -386,7 +386,11 @@ class MessageContext
       assert(mParts.Size() == 1);
       TMessageSerializer::Serialize(*mPayloadMsg, mObject.get(), nullptr);
       auto* dh = o2::header::get<o2::header::DataHeader*>(mParts.At(0).get());
-      LOGP(warning, "{} is being serialised with ROOT. Consider making the message messageable.", *dh);
+      if (dh) {
+       LOGP(warn, "{} is being serialised with ROOT. Consider making the message messageable.", *dh);
+      } else {
+       LOGP(warn, "Failed to extract the header");
+      }
       mParts.AddPart(std::move(mPayloadMsg));
       return ContextObject::finalize();
     }
```
I see that the header is not fetched:
```
shahoian@alicers05:~/alice/sw/BUILD/O2-latest/O2$ stage/tests/o2-test-framework-DataAllocator  -b --run | grep 'WARN'
[581286:source]: [01:50:09][WARN] Failed to extract the header
[581286:source]: [01:50:09][WARN] Failed to extract the header
[581286:source]: [01:50:09][WARN] Failed to extract the header
```

ktf:
there's is one * too many in the template argument. sorry about that.

shahor02:
> there's is one * too many in the template argument. sorry about that.

sorry, did not understand which extra `*` you mean.

ktf:
Should be fixed. GetData() not get()...

PR #10544, opened 2023-01-11
Include EMCAL and PHOS/CPV in fake CTP readout

mfasDa:
Procedure similar to the implementation for TRD.
In addition for EMCAL:
- Use class mask for EMCAL class
- Reject calibration triggers

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -83,6 +83,7 @@
 #include "TObjString.h"
 #include <map>
 #include <unordered_map>
+#include <set>
 #include <string>
 #include <vector>
 #include <thread>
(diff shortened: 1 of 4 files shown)


mfasDa:
Include EMCAL in fake CTP readout introduced in #9393.

@lietava Please have a look!

lietava:
@mfasDa : thanks. Have you tested it ?

mfasDa:
@lietava I have some comments on your original pull request:
```c++
  const auto& ft0RecPoints = recoData.getFT0RecPoints();
  for (auto& ft0RecPoint : ft0RecPoints) {
    auto t0triggers = ft0RecPoint.getTrigger();
    if (t0triggers.getVertex()) {
      uint64_t globalBC = ft0RecPoint.getInteractionRecord().toLong();
      uint64_t classmask = ctpcfg->getClassMaskForInputMask(0x4);
      // std::cout << "class mask:" << std::hex << classmask << std::dec << std::endl;
      bcsMapT0triggers[globalBC] = classmask;
    }
  }
```
This part looks to me that you determine the BCs with TVX and cache the trigger class for TRD, so that you later can assign it to the CTP digits. Why do you need to obtain it in the loop for each BC? The function getClassMaskForInputMask relies on the trigger descriptor. I had a look at the CTP config from an example run of LHC22m which was reconstructed in apass2 (523148) and for the CTP config of that run in the CCDB the descriptors in the class were all nullptr, consequently I got class mask 0 for all input masks, so I am not sure the implementation is going to work for the runs we need it for. I think that the determination of the class mask via the trigger cluster is more robust, and it needs to be done only once per trigger and function - consequently bcsMapT0triggers needs to store only the global BCs of TVX triggers and could therefore be a simple set.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -151,6 +154,50 @@ void AODProducerWorkflowDPL::createCTPReadout(const o2::globaltracking::RecoCont
...
+
+  // EMCAL:
+  cntwarnings = 0;
+  orbitPrev = 0;
+  bcPrev = 0;
+  // Get class mask for EMCAL:
+  uint64_t classMaskEMCAL = 0;
+  for (const auto& trgclass : ctpcfg->getCTPClasses()) {
+    if (trgclass.cluster->getClusterDetNames().find("EMC")) {
+      classMaskEMCAL = trgclass.classMask;
```
shahor02: should not one break here?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -151,6 +154,50 @@ void AODProducerWorkflowDPL::createCTPReadout(const o2::globaltracking::RecoCont
...
+    }
+  }
+  for (auto& emcrec : caloEMCCellsTRGR) {
+    auto orbitPrevT = orbitPrev;
+    auto bcPrevT = bcPrev;
+    bcPrev = emcrec.getBCData().bc;
+    orbitPrev = emcrec.getBCData().orbit;
+    if (orbitPrev < orbitPrevT || bcPrev >= o2::constants::lhc::LHCMaxBunches || (orbitPrev == orbitPrevT && bcPrev < bcPrevT)) {
```
shahor02: how is this possible?

shahor02 (commented):
@mfasDa, looks fine for me (see some comments below), would be good if also @lietava checks this.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mfasDa: What I had (still have) in mind is to combine all 3 triggers into a single lambda. From what I read in the other PR TRD needs this, and for EMCAL and PHOS it should be not needed but also hurt, as commented in the code, since I am not aware of spurious triggers sent by us. I kept it there in the preparation of merging.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
mfasDa: Ideally I would like to add PHOS and TRD into the same loop, then I can't break but need to check all cases separately, then it could be passes as argument to a lambda function which is generic for all detectors. As I wrote previously I see some issue with the CCDB objects which might lead to incorrect trigger classes in the TRD case. However I first would like to test it with EMCAL alone. While doing the test the PR I will label the PR as [WIP].

lietava:
> @lietava I have some comments on your original pull request:
> 
> ```c++
>   const auto& ft0RecPoints = recoData.getFT0RecPoints();
>   for (auto& ft0RecPoint : ft0RecPoints) {
>     auto t0triggers = ft0RecPoint.getTrigger();
>     if (t0triggers.getVertex()) {
>       uint64_t globalBC = ft0RecPoint.getInteractionRecord().toLong();
>       uint64_t classmask = ctpcfg->getClassMaskForInputMask(0x4);
>       // std::cout << "class mask:" << std::hex << classmask << std::dec << std::endl;
>       bcsMapT0triggers[globalBC] = classmask;
>     }
>   }
> ```
> 
> This part looks to me that you determine the BCs with TVX and cache the trigger class for TRD, so that you later can assign it to the CTP digits. Why do you need to obtain it in the loop for each BC?
=>  yes, you may be right it looks like it can be outside of loop.
 The function getClassMaskForInputMask relies on the trigger descriptor. I had a look at the CTP config from an example run of LHC22m which was reconstructed in apass2 (523148)
=> I printed this config using something like in macro https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/CTP/macro/GetScalers.C  (comment scalers) and it printed ok including descriptors.
 and for the CTP config of that run in the CCDB the descriptors in the class were all nullptr, consequently I got class mask 0 for all input masks, so I am not sure the implementation is going to work for the runs we need it for. I think that the determination of the class mask via the trigger cluster is more robust, and it needs to be done only once per trigger and function - consequently bcsMapT0triggers needs to store only the global BCs of TVX triggers and could therefore be a simple set.

mfasDa:
@lietava I have changed the code in a more generic way. Please have a look whether this looks reasonable.
@peressounko With the generic digit creator including PHOS/CPV became trivial. Please have a look whether this looks OK for you as well.

shahor02 (approved):
looks fine to me, but I would ask @lietava also to validate it

mfasDa:
I have tested on a couple of files from the run mentioned above. I see the CTP digits with the expected trigger masks. For BCs with overlapping triggers the class masks get merged as expected.

lietava:
@mfasDa : indeed using descriptors is not needed at all as you did. (I will fix descriptors although not relevant now here.)
If you tested as commented above 
> I have tested on a couple of files from the run mentioned above. I see the CTP digits with the expected trigger masks. For BCs with overlapping triggers the class masks get merged as expected.

I suggest to pull.

mfasDa:
@shahor02 For merging could you please squash-merge and keep only the commit message of the first commit?

mfasDa:
Mask_523148
@lietava @shahor02 attached is the mask reconstructed from my local test (5 CTFs) for run 523148 in the AO2D.root. In the run the masks based on the CTP configuration were
```
CTP Class:minbias_TVX Hardware mask:1 Cluster index:0 Desc index:0 Downscale:1 BCM: Cluster:cru
CTP Class:minbias_TVX Hardware mask:2 Cluster index:1 Desc index:1 Downscale:1 BCM: Cluster:trd
CTP Class:minbias_TVX_L0 Hardware mask:4 Cluster index:2 Desc index:2 Downscale:1 BCM: Cluster:phscpv
CTP Class:minbias_TVX_L0 Hardware mask:16 Cluster index:3 Desc index:3 Downscale:1 BCM: Cluster:emc
```
We see entries for the single classes firing as well as the permutations of various overlaps. I would say this makes sense, but would be good if we could get a full run reconstructed (as we did for CTP mask from readout) for a final validation.

PR #10545, opened 2023-01-11
[FDD] Now propagate avg time for all the BCs in Rec 

arvindkhuntia:


andreasmolander:
Should we store average time for _all_ events in the same place as the time for events in TVDC?

arvindkhuntia:
> Should we store average time for _all_ events in the same place as the time for events in TVDC?

We can store in the same place but for the background events the size of the channel data should be zero.

andreasmolander (approved):

arvindkhuntia:
Could you please merge this @shahor02?

shahor02 (approved):

PR #10546, opened 2023-01-11
DPL GUI: add compatibility with DebugGUI v0.7.0

ktf:


ktf:
Hopefully the ImPlot we ship already had the new API available, but I am not sure 100% of the changes (most notably the lambda argument swap) were backward compatible. I guess I will add ifdefs otherwise...

@davidrohr you will need this and https://github.com/AliceO2Group/DebugGUI/pull/43. The macOS version seems to work (I am still testing), so maybe it's not so bad after all. I also added the vulkan backend to our distribution, so maybe we can switch to it everywhere.

PR #10547, opened 2023-01-11
Tracking QC: Add Efficiency vs Phi, Eta and Pt

jloemker:
Hey @chiarazampolli and @njacazio,

I fixed a small typo in the previously implemented (but not yet approved) PhiVsPt and added the same histograms for EtaVsPt. 
I got a bit lost in the testing documentation and hope this will run without any troubles.

Please let me know if there is anything else I should do !


chiarazampolli:
Hello @jloemker , 
Thanks! At a first look, all seems ok to me, but you need to fix the editing (see error reported in the PR).
Chiara

PR #10548, opened 2023-01-11
FileFetcher run alien_cp in debug mode

shahor02:
* Set debug env.vars for alien_cp.

* Option `--fetch-failure-threshold <float>` of o2-ctf-reader and o2-raw-tf-reader allows to produce fatal if
abs. number (option value <0) or fraction of fetches wrt total number of files (value > 0) have failed.
E.g. `--fetch-failure-threshold -2.5` will produce fatal on 3rd fetch failure, while `--fetch-failure-threshold 0.1`
will produce fatal after the failure of 10% of input files fetches.


shahor02:
Failures unrelated

PR #10549, opened 2023-01-11
[WIP] Additional cuts for TPC SCD creation

martenole:
Adding some more options to the map creation from unbinned residuals
- cuts on track quality
- cut on DCA
- start and end time can be provided in ms for runs during which TPC settings changed (ping @ehellbar )

Currently WIP since the workflow gets stuck at the start and I don't see why. Will follow up tomorrow. Probably I did something wrong adding the GRPGeomHelper

### DIFF
--- a/Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/src/TPCResidualReaderSpec.cxx
+++ b/Detectors/GlobalTrackingWorkflow/tpcinterpolationworkflow/src/TPCResidualReaderSpec.cxx
@@ -27,6 +27,7 @@
 #include "SpacePoints/TrackInterpolation.h"
 #include "SpacePoints/SpacePointsCalibConfParam.h"
 #include "DetectorsBase/Propagator.h"
+#include "DetectorsBase/GRPGeomHelper.h"
 #include "CommonUtils/StringUtils.h"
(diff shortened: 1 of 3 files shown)


martenole:
Hi @shahor02 
as soon as I add any CCDB input to the TPCResidualReaderSpec the workflow is just stuck without any error. The device does not have any input or output defined. Here the reader is loading its input from a ROOT file within the run method and dumping the results without being connected to any other DPL device.
In the workflow I only have 3 processes: tpc-residual-reader, internal-dpl-ccdb-backend and internal-dpl-injected-dummy-sink.
Can it be that the CCDB input does not work for such a workflow? Then I probably need to separate the reading part from the processing part and inside the processing part request the CCDB input, right?

martenole:
The current workflow looks like this
workflow

shahor02:
Hi @martenole 
The TPCResidualReaderSpec cannot get any input from the CCDB fetcher, since the latter needs some input to extract the TF info from the header and such input can be provided only by the reader itself. 
But why do you need CCDB object in the reader? It is supposed to only read the data and push it to other devices.

martenole:
Ok thanks, this is what I feared. In the beginning I did not separate the reading of residuals from the map creation. It all happens in the reader device, that's why I need the object in the reader.
But I will just create separate devices now, I anyhow should have done this earlier

PR #10563, opened 2023-01-13
Changes for CRU suppression of zeroes

preghenella:
by default the decoder does not expect to have zeroes in words 2 and 3 of the CRU payload. this is going to be the future mode for the readout. a function is defined to restore the old behaviour.

### DIFF
--- a/Detectors/TOF/compression/include/TOFCompression/Compressor.h
+++ b/Detectors/TOF/compression/include/TOFCompression/Compressor.h
@@ -74,6 +74,11 @@ class Compressor
     mDecoderCONET = val;
     mDecoderNextWordStep = val ? 0 : 2;
   };
+ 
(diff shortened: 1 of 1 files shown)


noferini:
Hi @preghenella ,
thanks for the changes.
I took your changes and, if you agree, I will open a new PR (to have the control on it, running some tests and in case applying fix)
I guess I have to change also raw encoding from MC as well
Is it fine with you?

noferini:
I open this new PR
https://github.com/AliceO2Group/AliceO2/pull/10601

shahor02:
@preghenella @noferini should one close this?

noferini:
Hi @shahor02,
yes we can close this (I guess only @preghenella can do it).
Cheers,
Francesco

PR #10565, opened 2023-01-13
[FOCAL-10] Add option to process multiple files in pad rootifier

mfasDa:
- Add option to process multiple files in the pad rootifier
- Add option to switch readout card type


mfasDa:
Option to run on multiple files is to process runs as single run which are split over multiple files

Readout card type switching is for testing and debugging

mfasDa:
CI error unrelated

shahor02 (approved):

PR #10566, opened 2023-01-13
MCH: properly configure the slot finalization

aferrero2707:
The call to `setSlotLength(o2::calibration::INFINITE_TF)` does not seem to be appropriate in order to trigger the slot finalization at EoS, because it does not set the `mFinalizeWhenReady` data member of the `TimeSlotCalibration` base class. This is needed in order to properly trigger the slot finalization when calling `checkSlotsToFinalize(o2::calibration::INFINITE_TF)`.

An explicit call to `setFinalizeWhenReady()` is instead needed. Internally this also calls `setSlotLength(o2::calibration::INFINITE_TF)`.

### DIFF
--- a/Detectors/MUON/MCH/Calibration/include/MCHCalibration/PedestalChannel.h
+++ b/Detectors/MUON/MCH/Calibration/include/MCHCalibration/PedestalChannel.h
@@ -16,21 +16,23 @@
 
 namespace o2::mch::calibration
 {
-/** 
- * @class PedestalChannel 
+/**
+ * @class PedestalChannel
  * @brief Pedestal mean and sigma for one channel
  *
  * A PedestalChannel stores the mean and sigma of the pedestal of one MCH channel,
  * as well as the number of entries (digits) used to compute those values.
  */
 struct PedestalChannel {
-  int mEntries{0};         // number of entries used so far for the mean and variance
-  double mPedestal{0};     // mean
-  double mVariance{0};     // variance
-  DsChannelId dsChannelId; // identifier of the channel
+  int mEntries{0};         ///< number of entries used so far for the mean and variance
+  double mPedestal{0};     ///< mean
+  double mVariance{0};     ///< variance
+  DsChannelId dsChannelId; ///< identifier of the channel
 
   /** return the RMS of the pedestal */
   double getRms() const;
+
+  bool isValid() const; ///< true if the channel is associated to a detector pad
 
   std::string asString() const;
   friend std::ostream& operator<<(std::ostream&, const PedestalChannel&);
--- a/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
+++ b/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
@@ -72,16 +72,24 @@ void BadChannelCalibrationDevice::logStats(size_t dataSize)
 
 void BadChannelCalibrationDevice::run(o2::framework::ProcessingContext& pc)
 {
+  if (mSkipData) {
+    return;
+  }
+
   o2::base::GRPGeomHelper::instance().checkUpdates(pc);
   o2::base::TFIDInfoHelper::fillTFIDInfo(pc, mCalibrator->getCurrentTFInfo());
   mTimeStamp = std::min(mTimeStamp, mCalibrator->getCurrentTFInfo().creation);
   auto data = pc.inputs().get<gsl::span<o2::mch::calibration::PedestalDigit>>("digits");
   mCalibrator->process(data);
 
-  std::string reason;
-  if (mCalibrator->readyToSend(reason)) {
-    LOGP(info, "We're ready to send output to CCDB ({})", reason);
-    sendOutput(pc.outputs(), reason);
+  if (!BadChannelCalibratorParam::Instance().onlyAtEndOfStream) {
+    std::string reason;
+    if (mCalibrator->readyToSend(reason)) {
+      mHasEnoughStat = true;
+      LOGP(info, "We're ready to send output to CCDB ({})", reason);
+      sendOutput(pc.outputs(), reason);
+      mSkipData = true;
+    }
   }
   logStats(data.size());
 }
@@ -100,6 +108,10 @@ ccdb::CcdbObjectInfo createCcdbInfo(const T& object, uint64_t timeStamp, std::st
 void BadChannelCalibrationDevice::endOfStream(o2::framework::EndOfStreamContext& ec)
 {
   if (BadChannelCalibratorParam::Instance().onlyAtEndOfStream) {
+    std::string reason;
+    if (mCalibrator->readyToSend(reason)) {
+      mHasEnoughStat = true;
+    }
     sendOutput(ec.outputs(), "End Of Stream");
   }
 }
@@ -157,28 +169,35 @@ void BadChannelCalibrationDevice::sendOutput(o2::framework::DataAllocator& outpu
   }
 
   LOGP(info, "sendOutput: {}", reason_with_entries);
-  mCalibrator->checkSlotsToFinalize(o2::calibration::INFINITE_TF);
+  mCalibrator->finalize();
 
-  // send regular bad channel object to subspec 0. This regular object
-  // is meant for O2 consumption (in reconstruction and/or simulation)
-  const auto& badChannels = mCalibrator->getBadChannelsVector();
-  auto info = createCcdbInfo(badChannels, mTimeStamp, reason_with_entries);
+  // the bad channels table is only updated if there is enough statistics
+  if (mHasEnoughStat) {
+    // send regular bad channel object to subspec 0. This regular object
+    // is meant for O2 consumption (in reconstruction and/or simulation)
+    const auto& badChannels = mCalibrator->getBadChannelsVector();
+    auto info = createCcdbInfo(badChannels, mTimeStamp, reason_with_entries);
 
-  sendCalibrationOutput(output, 0, &badChannels, &info);
+    sendCalibrationOutput(output, 0, &badChannels, &info);
 
-  // send also a simplified (in comma-separated values format) version
-  // of the bad channel list to subspec 1.
-  // This simplified version is meant for DCS usage
-  // (to populate the Oracle online DB for electronics configuration)
+    // send also a simplified (in comma-separated values format) version
+    // of the bad channel list to subspec 1.
+    // This simplified version is meant for DCS usage
+    // (to populate the Oracle online DB for electronics configuration)
 
-  TObjString badChannelsPOD(toCSV(badChannels).c_str());
-  auto infoPOD = createCcdbInfo(badChannelsPOD, mTimeStamp, reason_with_entries);
+    TObjString badChannelsPOD(toCSV(badChannels).c_str());
+    auto infoPOD = createCcdbInfo(badChannelsPOD, mTimeStamp, reason_with_entries);
 
-  sendCalibrationOutput(output, 1, &badChannelsPOD, &infoPOD);
+    sendCalibrationOutput(output, 1, &badChannelsPOD, &infoPOD);
 
-  LOGP(info, "csv={}", badChannelsPOD.String().Data());
+    LOGP(info, "csv={}", badChannelsPOD.String().Data());
+  } else {
+    LOGP(error, "CCDB not updated: {}", reason_with_entries);
+  }
+
   // and finally send also the data used to compute the bad channel map
   // on a separate channel (for QC mainly)
+  // this is sent also if statistics is too low, for diagnostics purposes
   output.snapshot(o2::framework::Output{"MCH", "PEDESTALS", 0},
                   mCalibrator->getPedestalsVector());
 }
--- a/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrator.cxx
+++ b/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrator.cxx
@@ -34,14 +34,12 @@ void BadChannelCalibrator::initOutput()
 bool BadChannelCalibrator::readyToSend(std::string& reason) const
 {
   reason = "";
-  if (BadChannelCalibratorParam::Instance().onlyAtEndOfStream) {
-    return false;
-  }
 
   // let's check our hypothesis about this object (nslots=1) is actually true
   auto nslots = getNSlots();
   if (nslots != 1) {
-    LOGP(fatal, "nslots={} while it is expected to be 1", nslots);
+    LOGP(error, "nslots={} while it is expected to be 1", nslots);
+    return false;
   }
 
   auto& slot = getFirstSlot();
@@ -53,6 +51,18 @@ bool BadChannelCalibrator::readyToSend(std::string& reason) const
   return statIsEnough;
 }
 
+void BadChannelCalibrator::finalize()
+{
+  // let's check our hypothesis about this object (nslots=1) is actually true
+  auto nslots = getNSlots();
+  if (nslots != 1) {
+    LOGP(fatal, "nslots={} while it is expected to be 1", nslots);
+  }
+
+  auto& slot = getSlot(0);
+  finalizeSlot(slot);
+}
+
 bool BadChannelCalibrator::hasEnoughData(const Slot& slot) const
 {
   const int minNofEntries = BadChannelCalibratorParam::Instance().minRequiredNofEntriesPerChannel;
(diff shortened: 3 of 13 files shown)


aferrero2707:
@aphecetche this PR is expected to fix the issues we are having with the empty bad channels calibrations.

@chiarazampolli @shahor02 could you please have a look and confirm that the fix is correct?
There is also something I do not fully understand in the [TimeSlotCalibration](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h) code, in particular in the calls to [setSlotLength()](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h#L54) and [setFinalizeWhenReady()](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h#L92): the two functions seem to call each other, but with a different logic regarding the value of the `TFType` parameter.

shahor02:
@aferrero2707, 
(1) with  `setFinalizeWhenReady` you tell to calibrator to finalize the slot and to stop the processing once wanted stat. is accumulated (w/o waiting for the EoR). It forces the slot length to infinity just to make sure that everything goes to 1 slot.
(2) In opposite, simply setting the slot length to infinity means that the slot will be filled for every TF w/o attempting to be finalized. 
The finalization will happen at the endOfStream call of the calibrator.

If the behaviour you want is (1), then the your PR is fine.

`setSlotLength(0)`, which calls the  `setFinalizeWhenReady` is just the way to set this option via dummy slot length.

aferrero2707:
@shahor02 we actually want to keep both behaviors, although the processing at End-of-Stream is the preferred one. However, we want the possibility to create the calibration objects once a given statistics is reached, in case the EoS handling is broken for some reason.

However, when `mFinalizeWhenReady = false` and `checkSlotsToFinalize()` is called with `tf = INFINITE_TF`, as we are doing [here](https://github.com/AliceMCH/AliceO2/blob/mrrtf-204-fix-eos-in-calibrator/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx#L160), the slot finalization never happens because [this check](https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h#L317) is always false (`tf` and `tfLim` are both equal to `INFINITE_TF`). Could this be a bug?

At this point I have the impression that we should modify/simply our calibrator logic:
* we should call `setUpdateAtTheEndOfRunOnly()` only when using the EoS method. In this case we should not call `checkSlotsToFinalize()` in our calibrator code, but instead directly call `finalizeSlot()` passing the one and only active slot (since the slot length is set to `INFINITE_TF`)
* when we do not rely on EoS and we require a minimum statistics instead, we should not call `setUpdateAtTheEndOfRunOnly()` and let the calibrator's `process()` function handle the slot finalization

Does this make sense?

aferrero2707:
@shahor02 one more question, as I am still trying to understand the details of the underlying logic:

> with setFinalizeWhenReady you tell to calibrator to finalize the slot and to stop the processing once wanted stat. is accumulated (w/o waiting for the EoR).

How the processing is supposed to be stopped? Should it be manually stopped in the calibrator device code, or is it supposed to happen automatically?

As far as I can see, if the processing is not explicitly stopped in the DPL device code, then a new slot is inserted and starts to be filled...

shahor02:
@aferrero2707 sorry, rechecking the code I see that what I wrote before is wrong: the FinalizeWhenReady first defines a slot with an infinite end of validity and once this slot gets sufficient statistics it is finalized, its end of validity is set to the highest TF seen so far. Then a new slot is defined, from where the previous has ended till `INFINITE_TF` and the same processing continues. Untill the endOfStream calls the `checkSlotsToFinalize` with `TF = INFINITE_TF`, meaning that the processing is over.
So, 
1) if you want to have multiple calibration slots within the run with the boundaries defined according to the collected statistics, then the setFinalizeWhenReady is right way to go. 

2) If you want calibration slots of fixed duration, you just set setSlotLength (or setSlotLengthInSeconds or setSlotLengthInOrbits).

3) If you want to process the data of the whole run in a single calib. slot which is finalized at the endOfStream, then you should  `setSlotLength(o2::calibration::INFINITE_TF-1)`. This will ensure finalization in the endOfStream call of `checkSlotsToFinalize(o2::calibration::INFINITE_TF)` (or simply `checkSlotsToFinalize()`).
Note that in this case, you can set `setUpdateAtTheEndOfRunOnly()` to not spend time on the checks for slots to finalize in the run(..) method.

aferrero2707:
@shahor02 in our case we want to process the data of the whole run in a single calib. slot, which is finalized either at the EndOfStream or as soon as a target statistics is reached (the two modes are controlled by the `BadChannelCalibratorParam::Instance().onlyAtEndOfStream` configuration option).

Given that we only have one slot per run, and that we were already explicitly checking the slot statistics in the `run()` method, I considered that we do not really need the complexity of the `checkSlotsToFinalize()` function, and that we can easily handle the slot finalization explicitly.

Hence I modified/simplified the code in the following way:
1. in the `run()` method, we check the value of the `onlyAtEndOfStream` flag (see [here](https://github.com/AliceMCH/AliceO2/blob/mrrtf-204-fix-eos-in-calibrator/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx#L87)):
    * if `false` we check the slot statistics ([here](https://github.com/AliceMCH/AliceO2/blob/mrrtf-204-fix-eos-in-calibrator/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx#L89)) and we trigger the slot finalization if the statistics is sufficient (see [here](https://github.com/AliceMCH/AliceO2/blob/mrrtf-204-fix-eos-in-calibrator/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx#L174) and [here](https://github.com/AliceMCH/AliceO2/blob/mrrtf-204-fix-eos-in-calibrator/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrator.cxx#L54) for the details). Further data is then skipped once the slot is finalized and the CCDB populated.
    * if `true` the statistics is not checked in the `run()` function and everything is postponed to the EndOfStream
2. in the `sendOutput()` method we directly call the slot finalization (through the [BadChannelCalibrator::finalize(https://github.com/AliceMCH/AliceO2/blob/mrrtf-204-fix-eos-in-calibrator/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrator.cxx#L54)]() helper method) instead of going through the `checkSlotsToFinalize()`, because we only have one single slot to finalize and there is no need to start a new one afterward

Do you have objections to this approach, or suggestions to improve the code?

Thanks!

Detectors/MUON/MCH/Calibration/include/MCHCalibration/PedestalChannel.h
```diff
@@ -16,18 +16,19 @@
...
 struct PedestalChannel {
-  int mEntries{0};         // number of entries used so far for the mean and variance
-  double mPedestal{0};     // mean
-  double mVariance{0};     // variance
-  DsChannelId dsChannelId; // identifier of the channel
+  int mEntries{0};         ///< number of entries used so far for the mean and variance
+  double mPedestal{0};     ///< mean
+  double mVariance{0};     ///< variance
+  DsChannelId dsChannelId; ///< identifier of the channel
+  bool mValid{false};      ///< true if the channel is associated to a detector pad
```
aphecetche: do we really need an extra flag here ? Can't a test on DsChannelId  give the same information ?

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
```diff
@@ -35,8 +35,8 @@ void BadChannelCalibrationDevice::init(o2::framework::InitContext& ic)
   mLoggingInterval = ic.options().get<int>("logging-interval") * 1000;
 
   mCalibrator = std::make_unique<o2::mch::calibration::BadChannelCalibrator>();
-  mCalibrator->setSlotLength(o2::calibration::INFINITE_TF);
   mCalibrator->setUpdateAtTheEndOfRunOnly();
+  mCalibrator->setSlotLength(o2::calibration::INFINITE_TF);
```
aphecetche: here you changed the order the calls to `setUpdateAtTheEndOfRunOnly` and `setSlotLength` : is it on purpose ? in which case would be good to add a comment right there in the code to say so.
If not, just forget this comment ;-)

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
```diff
@@ -72,16 +72,26 @@ void BadChannelCalibrationDevice::logStats(size_t dataSize)
 
 void BadChannelCalibrationDevice::run(o2::framework::ProcessingContext& pc)
 {
+  static bool skip = false;
```
aphecetche: This might not matter much, but the usage of this static means that the same ECS environment can *not* be reused to perform two pedestal runs, right ?

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrator.cxx
```diff
@@ -34,14 +34,12 @@ void BadChannelCalibrator::initOutput()
 bool BadChannelCalibrator::readyToSend(std::string& reason) const
 {
   reason = "";
-  if (BadChannelCalibratorParam::Instance().onlyAtEndOfStream) {
-    return false;
-  }
 
   // let's check our hypothesis about this object (nslots=1) is actually true
   auto nslots = getNSlots();
   if (nslots != 1) {
```
aphecetche, line 40: I'm not sure we still need this check here as it is now duplicated in the `finalize` method which is anyways called, right ?

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrator.cxx
```diff
@@ -61,14 +71,13 @@ bool BadChannelCalibrator::hasEnoughData(const Slot& slot) const
   const int requiredChannels = static_cast<int>(BadChannelCalibratorParam::Instance().minRequiredCalibratedFraction * nofChannels);
 
   auto nofCalibrated = std::count_if(pedData->cbegin(), pedData->cend(),
-                                     [&](const PedestalChannel& c) { return c.mEntries > minNofEntries; });
+                                     [&](const PedestalChannel& c) { return (c.mValid && (c.mEntries > minNofEntries)); });
```
aphecetche: actually, instead of testing for `mValid` (see also the earlier comment about whether or not we really need this extra boolean) here and in line 103, the skipping of the non existing channel could be baked in the iterator directly.

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
shahor02: The order should not matter, but you should set 
```
setSlotLength(o2::calibration::INFINITE_TF);
setMaxSlotsDelay(0);
```
to guarantee that the endOfStream call with `tf = INFINITE_TF`) passes the https://github.com/AliceO2Group/AliceO2/blob/b0f70e90f824acf876f63ca28f8ddeffb6b9523b/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h#L315-L317 check.

Detectors/MUON/MCH/Calibration/include/MCHCalibration/PedestalChannel.h
aferrero2707: Agreed. I have replaced the `mValid` member by a `isValid()` function that actually checks the value of the `dsChannelId`.

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
aferrero2707: No, it was not done on purpose, reverted.

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrator.cxx
aferrero2707: I propose to keep it in order to make sure that the subsequent call to `hasEnoughData()` takes the expected slot as input. In any case this is a fatal condition that we should never encounter.

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrator.cxx
aferrero2707: I am not sure to see how to implement this in the iterator... As explained above, I have instead replaced the `mValid` member by a `isValid()` function that actually checks the value of the `dsChannelId`.

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
aferrero2707: Good point! I have replaced the static variable by a member function, to avoid leaving some herd-to-debug checks in the code.

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
shahor02: @aferrero2707 a typo slipped to my previous comment, should be 
```
setSlotLength(o2::calibration::INFINITE_TF - 1);
setMaxSlotsDelay(0);
```
to guarantee that the endOfStream call with tf = INFINITE_TF) passes the https://github.com/AliceO2Group/AliceO2/blob/b0f70e90f824acf876f63ca28f8ddeffb6b9523b/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h#L315-L317

aferrero2707: @shahor02 I have to confess that I still do not fully understand the logic.. 
Calling `setMaxSlotsDelay(0)` will set `mFinalizeWhenReady` to `true`. If I understand correctly, this means that the program will not go through the lines you quote, instead it will reach this portion of code:
https://github.com/AliceO2Group/AliceO2/blob/b0f70e90f824acf876f63ca28f8ddeffb6b9523b/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h#L279-L286
In this case, the slot is indeed finalized as soon as the desired statistics is reached, but then the slot is removed and a new one is inserted. However, we do not really need this new slot insertion, because we only use one slot for the full run...

shahor02: > Calling `setMaxSlotsDelay(0)` will set `mFinalizeWhenReady` to `true`. 

No, you are confusing it with `setSlotLength(0)` behaviour,  the mMaxSlotsDelay (the unit is a fraction of the slot length) defines maximum difference `tf_current - slot[I]_end_tf` after which the slot `#I` is allowed to be checked for finalization (i.e. no more TFs contributing to it are expected). 
Though after rechecking the code I see that it will work also w/o `setMaxSlotsDelay(0)` setting, the important part is  `setSlotLength(INFINITE_TF-1)`:
since the endOfStream calls  `checkSlotsToFinalize` with `tf==INFINITE_TF`, and the slot can be finalized only if 
`"current" tf >  slot->getTFEnd()`, the `slot->getTFEnd()` must be `<INFINITE_TF`.

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrator.cxx
aphecetche: for the iterator, see https://github.com/AliceMCH/AliceO2/pull/34

Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx
aferrero2707: @shahor02 my bad, I completely mixed things up! You are absolutely correct, and `setSlotLength(INFINITE_TF-1)` is ok to trigger the slot finalization.

Nevertheless, I would still be in favor of not using the `checkSlotsToFinalize()` function in our code, for the following reasons:
* we initialize the calibrator with `setUpdateAtTheEndOfRunOnly()`, therefore we do not take advantage of the `checkSlotsToFinalize()` call at each TF in the data processing routine
* we anyhow check explicitly the available statistics in the `run()` method (see [here](https://github.com/AliceMCH/AliceO2/blob/9008f679c2ca05695ecbd439dad9d8d6db2ed413/Detectors/MUON/MCH/Calibration/src/BadChannelCalibrationDevice.cxx#L87)) when we do not rely on the EoS
* when the EoS is enabled, we finalize the slot even if the statistics is not enough, because we want to always send the output to the QC for diagnostics purposes - AFAIK this is not possible with the `checkSlotsToFinalize()` method

Do you see any drawbacks?

shahor02: @aferrero2707 no, I don't see drawbacks.

aphecetche:
@shahor02 @chiarazampolli as far as I can tell the failure has nothing to do with this PR. Could it be merged so we'll be able to test it at Pt2 later on ?

shahor02 (approved):

PR #10567, opened 2023-01-15
ctp: trigger offsets and config tests

lietava:


lietava:
I do not see ctp problem ?

shahor02 (approved):

PR #10568, opened 2023-01-15
ITS: Improve seeding with low multiplicity events

mconcas:


mconcas:
This PR essentially adds:

- Fix a minor bug in cluster merging in vertexing.
- Softening of the hard cut on contributors, provided that the associated vertex is close to the beamline. It is disabled by default.
- Option to enable 1-tracklet vertexing by using beamline as a second line to find vertex candidate.

@shahor02 @iouribelikov

Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
```diff
@@ -374,6 +374,19 @@ void VertexerTraits::computeVertices()
         }
       }
     }
+    if (mVrtParams.allowSingleContribClusters) {
+      auto beamLine = Line{{mTimeFrame->getBeamX(), mTimeFrame->getBeamY(), -50.f}, {mTimeFrame->getBeamX(), mTimeFrame->getBeamY(), 50.f}}; // use beam position as contributor
```
shahor02, line 378: I don't see where do you see the beam position, should not you read the MeanVertex object, like here: 
https://github.com/AliceO2Group/AliceO2/blob/c0375b3d61d15e0ca7a4fc0da82af9dcdfd41696/Detectors/GlobalTrackingWorkflow/src/PrimaryVertexingSpec.cxx#L240
https://github.com/AliceO2Group/AliceO2/blob/c0375b3d61d15e0ca7a4fc0da82af9dcdfd41696/Detectors/GlobalTrackingWorkflow/src/PrimaryVertexingSpec.cxx#L180-L183
https://github.com/AliceO2Group/AliceO2/blob/c0375b3d61d15e0ca7a4fc0da82af9dcdfd41696/Detectors/GlobalTrackingWorkflow/src/PrimaryVertexingSpec.cxx#L211
etc...

mconcas: In ITS reconstruction we keep track of a rough beam position estimation by using previous vertices transverse positions, weighted on they contributors.
This is reset at every timeframe and have been reliable enough so far...
Making the seeding depending from global reconstruction might be an overkill.

shahor02: Does this mean that the beginning of every TF uses 0,0 mean vertex which progressively goes to its correct value?

mconcas: Yes, but after finding the first seeding vertex, usually we have a sufficient resolution for the tracker tolerances.
With Max we are also planning to make this a persistent information, but will require some additional work to make the TimeFrame object not to be reset at each load, like it is currently done instead.

shahor02: Well, I don't see why not using the MeanVertex, at least in the async mode, which is calibrated exactly for such cases. With current approach the very low intensity data, e.g. lhc22s PbPb will always use 0/0.

mconcas: I see your point. I can implement a flag to enable the usage of meanvertex in case global reconstruction is available, being sync/async.
Not sure about the timescale, depends on when we are going to start apass3.

I would anyways keep also the current option as the default so to be able not to depend from all the rest of the reconstruction whenever we want run ITS reco workflow.

shahor02: Agree, optionally one could require constraining by own previous vertices

iouribelikov (dismissed):

Detectors/ITSMFT/ITS/tracking/src/VertexerTraits.cxx
mconcas: Okay, I'm still missing something; this does not work anymore after new changes. So do not merge it, please.

mconcas:
@shahor02 
Now mean vertex can be used to override the ITS standalone beam position calculation.
This can be enabled by passing the `--ccdb-meanvertex-seed` parameter.

shahor02 (dismissed):
thanks!

mconcas:
@shahor02: the test is probably stuck; besides that, I just rebased it. If you could re-approve, thanks.

mconcas:
Hello @TimoWilken! Test build/O2/o2 has been yellow for quite some time (~days). Could you kindly have a look?
Thanks

TimoWilken:
Hi @mconcas, sorry about that, the builder was stuck due to a now-fixed problem. I've unstuck it and it should pick up this PR soon.

shahor02 (approved):

PR #10572, opened 2023-01-16
Misc. unrelated small fixes.

shahor02:


shahor02:
@davidrohr are these guards really needed https://github.com/AliceO2Group/AliceO2/blob/eab86294739633342fc1dbf18289316061e9bc8a/DataFormats/Reconstruction/include/ReconstructionDataFormats/PID.h#L129-L135 ?

davidrohr:
> @davidrohr are these guards really needed
> 
the recursive function could be a problem, should be rewritten into a loop.

And the pointers to arrays in constexpr address space could lead to unsupported address space casts, but not 100% sure without trying. You can just try to remove it and see what the FullCI says. If needed, this can be solved if it doesn't compile as is.

shahor02:
thanks, I was referring only to `getName` method, which does not use recursive `sameString`.

shahor02:
failures are unrelated

PR #10573, opened 2023-01-16
Update for average TPC space charge distortion correction

martenole:


martenole:
Merging now, o2 check was green before

martenole:
Actually, I thought I could squash manually when merging, but it is possible only to merge all commits into one.

I squashed locally now and the checks were passed before, so merging now

PR #10574, opened 2023-01-16
Consistently use the systematics alignment variance

mpuccio:


Detectors/ITSMFT/ITS/tracking/src/TimeFrame.cxx
```diff
@@ -247,7 +247,7 @@ void TimeFrame::initialise(const int iteration, const TrackingParameters& trkPar
       mClusters[iLayer].resize(mUnsortedClusters[iLayer].size());
       mUsedClusters[iLayer].clear();
       mUsedClusters[iLayer].resize(mUnsortedClusters[iLayer].size(), false);
-      mPositionResolution[iLayer] = std::hypot(trkParam.LayerMisalignment[iLayer], trkParam.LayerResolution[iLayer]);
+      mPositionResolution[iLayer] = std::sqrt(trkParam.SystErrorZ2[iLayer] + trkParam.LayerResolution[iLayer] * trkParam.LayerResolution[iLayer]);
```
shahor02: to be sure: mPositionResolution is responsible both for the rphi and Z, right? If so and if you want to use single parameter, perhaps one should use e.g `(trkParam.SystErrorY2[iLayer]+trkParam.SystErrorZ2[iLayer])/2.` ?

mpuccio: Good point, updated and force-pushed

mconcas (approved):

PR #10577, opened 2023-01-17
Fix RDH error bits decoding

fapfap69:
Fix the HMPID Error bits decoding into the RDH.


fapfap69:
@shahor02 Hi Ruben, the PR fails, but I'm not able to understand where the error is ...

shahor02 (approved):
Failures unrelated (common problem with fullCI).

fapfap69:
Thank you !

PR #10579, opened 2023-01-17
[EMCAL-565] Add energy interval above 4GeV for bad channel calib

jokonig:
- To capture cases where cells are bad only above 4GeV (the current upper limit), a new energy interval above 4GeV is added. The upper limit arises from the upper limit of the input histogram.
- The energy/hit distribution is affected by outliers, if only a limited number of hits is measured. (If on average only 5 hits in the energy interval above 4GeV are registered, a hit with 30GeV will flag the cell as bad, which is not wanted) Therefore, a minimum number of hits is needed to cut on the energy/hit distribution. This value is stored in the calib params


mfasDa (approved):
Looks good!

mfasDa:
CI error unrelated

PR #10580, opened 2023-01-17
Mch hit improvements

mwinn2:
This is pull request improving the position of the MCH hit w.r.t. to the MCH chamber geometry and some preparatory works to include also b-field and angle effects. 
This PR is not yet locally compiled, since the author doesn´t manage to compile O2 locally. Hence, the nightlies will be used for tests for the time being.


mwinn2:
@pillot and @aphecetche : I cannot test it although it should compile now. Could you have a look? 
Do you think that the change is transparent enough with the hit position to merge it? 
Let me know.

PR #10582, opened 2023-01-17
Fix detection of CCDB production server in CcdbApi

shahor02:


costing (dismissed):
Thanks a lot Ruben!

davidrohr:
out of curiosity: why did the old code not work? should be the same, or do I miss something?
Or ok, if the string would start with spaces, it would break?

shahor02:
@davidrohr because the host is not necessarily precisely `http://alice-ccdb.cern.ch`, e.g. for CCDBBasicManager it was failing since by default it sets the `host` to `http://alice-ccdb.cern.ch/`. Or it can be e.g. `http://alice-ccdb.cern.ch/Users/x/xyz/` ...

davidrohr:
ok, makes sense.
doesn't really matter but a minor thing: rfind searches from the right, while what we are searching for should normally be on the left side of the string, so I would use find instead of rfind.

sawenzel (approved):

PR #10595, opened 2023-01-19
treat MC QED background labels as not valid

sawenzel:
Currently the AOD conversion crashes when encountering QED labels. I believe we can simply treat them as noise and therefore invalid labels in the true MC sense.

This commit then fixes the crash in the AOD conversation https://alice.its.cern.ch/jira/browse/O2-3508.

In case, special QED information needs to be included into AODs, we would need to adjust the AOD converter.


sawenzel:
Merged as hotfix to stabilize tests. Could take alternative approaches later.

PR #10596, opened 2023-01-19
TPC IDC: optimizing outlier filtering

matthias-kleiner:
Outlier map:
- 1. check for obvious outlier (IDCs around median +- nSigma)
- 2. check FECs: masking FECs which are out of sync
- 3. check neighbouring outliers (in case not all outlier for larger areas are recognized)

Other changes

IDCCCDBHelper
- add function to get the mean value for IDC0

IDC Drawing
- draw plots in current canvas if one is created
- draw sector informations

IDCFactorize
- parallelize loading of large objects from file
- add function to normalize IDCZero by gain map and/or median per stack
- Moving dump to TTree functionality to CCDBHelper implementation

TPC Mapper
- mapper make function constexpr


wiechula (approved):

matthias-kleiner:
@shahor02 @davidrohr can this be merged? The error seems unrelated

PR #10600, opened 2023-01-19
ITS: Add protection from bogus vtx coordinate

mconcas:
@chiarazampolli @mpuccio 
This is obviously a temporary workaround while I investigate the source of that weird vertex.
Testing the rebase offline, I'll update this asap


chiarazampolli:
Thanks!

chiarazampolli:
I tried https://github.com/AliceO2Group/AliceO2/pull/10600, which is the proper fix, and everything is fine with that. This PR was not tried explicitly, but it will of course not harm.
Thanks Matteo!

mconcas:
@shahor02 this should be a trivial protection that I'd like to keep for production anyways.

PR #10601, opened 2023-01-19
Changes for CRU suppression of zeroes

noferini:


noferini:
@shahor02 ,
this PR is needed because of a change in TOF raw format request by FLP team.
With the zero word suppression in TOF raw format TOF GBTword is not guaranteed to be 16
Then I needed to change also RawFileWriter
Changes validated with MC
encoding + deconding/TOF compressor
oldformat is still recoverable with a special flag (both in encoding and decoding, also tested)

shahor02:
Hi @noferini 
I see that you are steering new/old format via the command line option. Since the RDH will have a flag telling if the padding is on or off, I am going to extend the RDH class, so it can be detected from the data. Will try to do this over the weekend. I would propose to put this PR on hold until this detection is available.

noferini:
> Hi @noferini I see that you are steering new/old format via the command line option. Since the RDH will have a flag telling if the padding is on or off, I am going to extend the RDH class, so it can be detected from the data. Will try to do this over the weekend. I would propose to put this PR on hold until this detection is available.

Yes, sure. Thanks for taking care about it.

noferini:
In the meanwhile I simulated 7500 pp events in a TF (650 kHz) with both configurations: zero word suppression and old one

These are the sizes I got:

RAW -> 10.5 MB (zero word suppression/NEW) vs. 18.1 MB (OLD) -> 40% gain

digits ->0.462 MB (both cases)



For Pb-Pb@50kHz

RAW ->  13.6 MB (zero word suppression/NEW) vs. 28.4 MB (OLD) -> 50% gain

digits -> 3.5 MB (both cases)


For all simulations I used "o2-sim -m TOF" -> only TOF in simulation to speedup the process

-> absolute value of the sizes underestimated (30-50%?) but gain (ratios) should be reliable

noferini:
All changes applied and tested.
Now it runs without padding (encoding and decoding) by default.
To enable padding in encoding from MC and padding auto discovery ->  setting of RDHV7 as default is needed (TOF is ready for that).
Still WIP since depending on https://github.com/AliceO2Group/AliceO2/pull/10612

shahor02:
> All changes applied and tested. Now it runs without padding (encoding and decoding) by default. To enable padding in encoding from MC and padding auto discovery -> setting of RDHV7 as default is needed (TOF is ready for that). Still WIP since depending on #10612

I think it will take some time until we switch to default v7. What will be the RDH version assumed for old data and MC->raw until that?

noferini:
> > All changes applied and tested. Now it runs without padding (encoding and decoding) by default. To enable padding in encoding from MC and padding auto discovery -> setting of RDHV7 as default is needed (TOF is ready for that). Still WIP since depending on #10612
> 
> I think it will take some time until we switch to default v7. What will be the RDH version assumed for old data and MC->raw until that?

With V6 as default (now) I have to assume padding since dataformat returns always zero.
The default is then oldFormat also in encoding from MC.
I added a check on the RDH version. When we will switch to V7 the default will be automatically "no-padding".
This means that we cannot switch to no-padding at P2 without V7 becomes the default

noferini:
FullCI failing due to
[30491:MFTDigitizer]:     o2-sim-digitizer-workflow() [0x4cf8c0]:     std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string() at basic_string.h:658
[30490:ITSDigitizer]:     o2-sim-digitizer-workflow() [0x4d0c70]:     std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string() at basic_string.h:658
[30491:MFTDigitizer]:     o2-sim-digitizer-workflow() [0x4d182d]:     o2::detectors::DetID::getName(int) at DetID.h:140
[30490:ITSDigitizer]:     o2-sim-digitizer-workflow() [0x4d234a]:     o2::itsmft::ITSMFTDPLDigitizerTask::run(o2::framework::ProcessingContext&) at ITSMFTDigitizerSpec.cxx:62 (discriminator 1)

in digitization --> not related to this PR

noferini:
@shahor02 can we merge this? (all checks now passed)

shahor02 (approved):

PR #10604, opened 2023-01-19
account for pileup in TRD timing estimate

shahor02:
Effect of closest pile-up interaction vs distance to it (from the trigger) on the TRD tracklet multiplicity:
trd_pileup

Original results (blue) of [Nov.17 sample](https://alice.its.cern.ch/jira/browse/O2-2761?focusedCommentId=295224&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-295224) and 
accTRDPileup60 and the same data reconstructed with TRD pileup accounted in at most 60 BCs after trigger (red).

### DIFF
--- a/Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
+++ b/Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
@@ -507,8 +507,8 @@ bool TRDGlobalTracking::refitITSTPCTRDTrack(TrackTRD& trk, float timeTRD, o2::gl
   auto detRefs = recoCont->getSingleDetectorRefs(trk.getRefGlobalTrackId());
   int nCl = -1, clEntry = -1, nClRefit = 0, clRefs[14];
   float chi2Out = 0, timeZErr = 0.;
-  int pileUpDist = trk.getPileUpDistance(); // distance to farthest collision within the pileup integration time
-  if (pileUpDist) {
+  bool pileUpOn = trk.hasPileUpInfo(); // distance to farthest collision within the pileup integration time is set
+  if (pileUpOn) {
     timeTRD += trk.getPileUpTimeShiftMUS(); // shift to average pileup position
   }
   auto geom = o2::its::GeometryTGeo::Instance();
@@ -623,8 +623,8 @@ bool TRDGlobalTracking::refitTPCTRDTrack(TrackTRD& trk, float timeTRD, o2::globa
   auto detRefs = recoCont->getSingleDetectorRefs(trk.getRefGlobalTrackId());
   outerParam = trk;
   float chi2Out = 0, timeZErr = 0.;
-  int pileUpDist = trk.getPileUpDistance(); // distance to farthest collision within the pileup integration time
-  if (pileUpDist) {
+  bool pileUpOn = trk.hasPileUpInfo(); // distance to farthest collision within the pileup integration time is set
+  if (pileUpOn) {
     timeTRD += trk.getPileUpTimeShiftMUS(); // shift to average pileup position
   }
   int retVal = mTPCRefitter->RefitTrackAsTrackParCov(outerParam, mTPCTracksArray[detRefs[GTrackID::TPC]].getClusterRef(), timeTRD * mTPCTBinMUSInv, &chi2Out, true, false); // outward refit
@@ -632,7 +632,7 @@ bool TRDGlobalTracking::refitTPCTRDTrack(TrackTRD& trk, float timeTRD, o2::globa
     LOG(debug) << "TPC refit outwards failed";
     return false;
   }
-  if (pileUpDist) { // account pileup time uncertainty in Z errors
+  if (pileUpOn) { // account pileup time uncertainty in Z errors
     timeZErr = mTPCVdrift * trk.getPileUpTimeErrorMUS();
     outerParam.updateCov(timeZErr, o2::track::CovLabels::kSigZ2);
   }
@@ -653,7 +653,7 @@ bool TRDGlobalTracking::refitTPCTRDTrack(TrackTRD& trk, float timeTRD, o2::globa
     LOG(debug) << "TPC refit inwards failed";
     return false;
   }
-  if (pileUpDist) { // account pileup time uncertainty in Z errors
+  if (pileUpOn) { // account pileup time uncertainty in Z errors
     trk.updateCov(timeZErr, o2::track::CovLabels::kSigZ2);
   }
 
--- a/Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
+++ b/Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
@@ -32,6 +32,7 @@
 #include "ITStracking/IOUtils.h"
 #include "ITSBase/GeometryTGeo.h"
 #include "DataFormatsITSMFT/Cluster.h"
+#include "DataFormatsFT0/RecPoints.h"
 #include "ITSReconstruction/RecoGeomHelper.h"
 #include "ITSMFTReconstruction/ClustererParam.h"
 // GPU header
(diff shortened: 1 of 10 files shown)


prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -136,6 +136,7 @@ fi
 
 
 workflow_has_parameter CALIB && [[ $CALIB_TRD_VDRIFTEXB == 1 ]] && TRD_CONFIG+=" --enable-trackbased-calib"
+! has_detector && TRD_CONFIG+=" --disable-ft0-pileup-tagging"
```
martenole:
```suggestion
! has_detector FT0 && TRD_CONFIG+=" --disable-ft0-pileup-tagging"
```

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -370,6 +371,35 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
   std::iota(trackIdxArray.begin(), trackIdxArray.end(), 0);
   std::sort(trackIdxArray.begin(), trackIdxArray.end(), [tracksOutRaw](int lhs, int rhs) { return tracksOutRaw[lhs].getCollisionId() < tracksOutRaw[rhs].getCollisionId(); });
 
+  std::vector<uint16_t> pileUpDist;
+  bool ft0Seen = false;
+  if (mTrkMask[GTrackID::FT0]) { // pile-up tagging was requested
+    auto ft0recPoints = inputTracks.getFT0RecPoints();
+    auto trdTriggers = inputTracks.getTRDTriggerRecords();
```
martenole:
```suggestion
    auto trdTriggers = tmpInputContainer->mTriggerRecords;
```

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -370,6 +371,35 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
...
+    long maxDiff = mTracker->Param().rec.trd.pileupNBC;
+    ft0Seen = ft0recPoints.size() > 0;
+    pileUpDist.resize(trdTriggers.size(), 0);
```
martenole: Ok, so you set it to 0 by default. That way, in case FT0 is available but no pileup collision was seen we use the nominal very small error for the TRD time.
In case FT0 is not available (and only in that case) you set the error to `mTracker->Param().rec.trd.pileupNBC`.
This looks good to me

GPU/GPUTracking/DataTypes/GPUTRDInterfaceO2Track.h
```diff
@@ -74,6 +75,11 @@ class trackInterface<o2::track::TrackParCov> : public o2::track::TrackParCov
 
   GPUdi() bool CheckNumericalQuality() const { return true; }
 
+  GPUdi() uint16_t getPileUpDistance() const { return getUserField(); }
+  GPUdi() void setPileUpDistance(uint16_t d) { return setUserField(d); }
+  GPUdi() float getPuleUpTimeShiftMUS() const { return getPileUpDistance() * o2::constants::lhc::LHCBunchSpacingMUS * 0.5; }
+  GPUdi() float getPuleUpTimeErrorMUS() const { return getPileUpDistance() * o2::constants::lhc::LHCBunchSpacingMUS / 3.4641016; }
```
martenole: typos in the 2 lines above: getPuleUpTime... -> getPileUpTime
and why is the error divided by 3.46?

martenole (commented):
Thanks, looks good. I only have some minor comments. And I have not run it yet

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
shahor02: yes, and I am increasing (in the update to come) the nominal TRD error to 5 ns (since it is used as a gaussian error in TOF matching).

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
shahor02: ok

GPU/GPUTracking/DataTypes/GPUTRDInterfaceO2Track.h
shahor02: Not sure what do you mean by typos, these are conversions of NBC range to time in \mus and its error (3.4641016 = sqrt(12)). There was a typo in the `void setPileUpDistance` (return not needed).

prodtests/full-system-test/dpl-workflow.sh
shahor02: thanks for spotting!

GPU/GPUTracking/DataTypes/GPUTRDInterfaceO2Track.h
martenole: The `return` I overlooked. I just meant that you put "PuleUp" instead of "PileUp", so really just typo

shahor02:
@martenole ok thanks, fixed + the new version also accounts for the pre-trigger pilep (by default allowing +80 BC window, can be varied via configurables `GPU_rec_trd.pileupBwdNBC` (pre-trigger) and `GPU_rec_trd.pileupFwdNBC` (post-trigger).

@noferini I noticed that for the TOF matching you were taking 30 std.dev of the seeding track time resolution. This is certainly an overkill: ITS-TPC resolution and bias are limited to 1ms, TRD now will have 5 ns (w/o pile-up) or detected pile-up span.
I've added `MatchTOF.nsigmaTimeCut` configurable with default value 4 (sigma)

----
For the record: adding the dependence of the NTRDTracklets on the pre-trigger pileup distance:
https://mattermost.web.cern.ch/api/v4/files/58ecwjuuyby6ugytwhiesqaata/preview

noferini:
Hi @shahor02 , It looks strange. I'll check, now I am travelling and I cannot

noferini:
Hi @shahor02 ,
I checked and you are right. I was quite sure to move it to a lower value but I was wrong.
Your propsal is fine.
Thanks for looking into that.

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -370,6 +371,43 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
...
+      for (size_t ft0id = curFT0; ft0id < ft0recPoints.size(); ft0id++) {
+        const auto& f0rec = ft0recPoints[ft0id];
+        if (f0rec.getTrigger().getVertex()) {
+          auto bcdiff = trig.getBCData().toLong() - f0rec.getInteractionRecord().toLong();
+          if (bcdiff > maxDiffBwd) {
+            curFT0 = ft0id + 1;
+            continue;
+          }
+          if (bcdiff > 0) { // pre-trigger pileup, maxDiffBwd is guaranteed to be < max uint8_t
+            if (bwd == 0) {
```
martenole, line 396: Should this condition not be removed? Such that in case there are multiple pre-trigger pileup collision the closest one is chosen and not only the first one?

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -370,6 +371,43 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
...
+              bwd = uint8_t(bcdiff);
+            }
+          } else {
+            if (bcdiff < -maxDiffFwd) {
+              break;
+            }
+            fwd = uint8_t(-bcdiff); // post-trigger pileup, maxDiffFwd is guaranteed to be < max uint8_t
```
martenole, line 403: For the post-trigger collisions we should have the above check `if (fwd == 0)` and set it only once, no?

martenole: Also, here and in some other places the code checker complains about missing parentheses for the casting, so you probably need either `(uint8_t)(...)` or `static_cast<uint8_t>(...)`

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
shahor02: No, we want to find the pre- and post- trigger pile-ups with the largest acceptable distance to the trigger.
The FT0 recpoints are orderer so when accounting pre-trigger pile-up and when the `bwd` is not modified yet (==0) the 1st encountered FT0 recpoint within the `maxDiffBwd` from the trigger gives the pre-trigger pile-up we need, we don't need to update `bwd` further.

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
shahor02: > For the post-trigger collisions we should have the above check `if (fwd == 0)` and set it only once, no?

Why? Like above, since the FT0 is ordered and we want to find the furthest acceptable post-trigger collision, I keep updating fwd until the FT0 exceeds the acceptable range.

shahor02: > 

Looks like it did not line uintXXX_t times, changed to `unsigned char/short`.

martenole: Right, sorry, for the time error we are interested in the furthest and not the closest collisions in the window of course

martenole (approved):

shahor02:
fullCI failure is unrelated, validated locally

PR #10606, opened 2023-01-20
[FOCAL-11] Provide simple base analysis task for TB events

mfasDa:
Simple analysis base task for analysis of o2::focal::Event created in the reconstruction of FOCAL-E data. Internally iterating over events using the o2::focal::EventReader. Users need to implement the functions
- init
- process
- terminate Functionality will stay limited considering the conditions of TB data (i.e. runs will be in the same event tree - no need for chaining, no train infrastructure).


mfasDa:
CI errors unrelated

PR #10608, opened 2023-01-21
TOF: Adding workflow for integrated TOF currents

matthias-kleiner:
The 1D integrated TOF currents (~1ms) are similar as the TPC 1D-IDCs and can be used for QA and calibration.

The clusters are integrated in the `o2-tof-integrate-cluster-workflow` device and aggregated + merged to longer contiguous time intervals (1000TFs default ~ 11s) in the `o2-tof-merge-integrate-cluster-workflow` device using the `TimeSlotCalibration`. The workflow can be used using the CTFs as input:
```
o2-ctf-reader-workflow  --copy-cmd no-copy   --onlyDet TOF  --ctf-input "ctf.list"   --severity warning \
| o2-tof-reco-workflow --input-type compressed-clusters-ctf   --output-type "clusters,disable-writer" --disable-mc --disable-root-output \
| o2-tof-integrate-cluster-workflow | o2-tof-merge-integrate-cluster-workflow --debug true \
|  o2-calibration-ccdb-populator-workflow --ccdb-path "ccdb-test.cern.ch:8080" -b
```

The output will be stored in the CCDB or in local files using the `o2-tof-merge-integrate-cluster-workflow --debug true` option.


matthias-kleiner:
@chiarazampolli 
This is the PR for the integrated cluster currents for the TOF. I already tested it with a list of input CTFs and it was running fine. There is also some output http://ccdb-test.cern.ch:8080/browse/TOF/Calib/ITOFC. The file size seems 60kb for 11 seconds

matthias-kleiner:
Example plot for around 3 minutes for Run 523141
22m_Run523141

chiarazampolli:
Hello @matthias-kleiner ,

Thanks! We currently don’t do calibration from grid jobs since we have no aggregator, so we need to use the option to store to local files, right?

Cheers,

Chiara

noferini:
Dear @matthias-kleiner,
I have a couple of (minor) comments concerning the workflow you reported.

In your tof-workflow config you have
o2-tof-reco-workflow --input-type compressed-clusters-ctf   --output-type "clusters,disable-writer" 

The input-type you are asking is not defined. It works because ctf already provides digits but it is misleading.
I would put "--input-type none" for clarity

Also disable-writer in the output-type doesn't have any meaning. You have --disable-root-output and it is enough (in any case for data tof-workflow is already defined in O2DPG scripts so you will inherit it from the global process)

I also did't understand if and how this is supposed to run in async reco.

Detectors/TOF/workflow/src/TOFIntegrateClusterSpec.cxx
```diff
@@ -0,0 +1,121 @@
...
+      if (sliceInTF < mNSlicesTF) {
+        ++iTOFCNCl[sliceInTF];                          // count number of cluster
+        iTOFCqTot[sliceInTF] += clusters[iCl].getTot(); // integrated charge
+      } else {
+        LOGP(info, "slice in TF of ICC {} is larger than max slice {} with nTSPerSlice {}", sliceInTF, mNSlicesTF, nSlices);
+      }
+    }
+
+    // normalize currents to integration time
+    const double sliceWidthMS = sliceWidthPS * std::pow(10, -9);
```
noferini: std::pow(10, -9) -> 1.E-9

Detectors/TOF/workflow/src/TOFIntegrateClusterSpec.cxx
```diff
@@ -0,0 +1,121 @@
...
+    const int nSlices = maxTimeTF / sliceWidthPS;       // number of slices a TF is divided into
+
+    // storage for integrated currents
+    o2::pmr::vector<float> iTOFCNCl(nSlices);
+    o2::pmr::vector<float> iTOFCqTot(nSlices);
+
+    const auto clusters = pc.inputs().get<gsl::span<o2::tof::Cluster>>("tofcluster");
+    for (size_t iCl = 0; iCl < clusters.size(); ++iCl) {
+      const double timePS = clusters[iCl].getTime();
+      const unsigned int sliceInTF = timePS / sliceWidthPS;
```
noferini: Can you define as a constant
sliceWidthPSinv = 1./sliceWidthPS
and then replace division with multiplication (much faster)?

Detectors/TOF/workflow/src/TOFIntegrateClusterSpec.cxx
noferini: Why you don't define it before of the loop when you define sliceWidthPS?

Detectors/TOF/workflow/src/TOFIntegrateClusterSpec.cxx
```diff
@@ -0,0 +1,121 @@
...
+        ++iTOFCNCl[sliceInTF];                          // count number of cluster
+        iTOFCqTot[sliceInTF] += clusters[iCl].getTot(); // integrated charge
+      } else {
+        LOGP(info, "slice in TF of ICC {} is larger than max slice {} with nTSPerSlice {}", sliceInTF, mNSlicesTF, nSlices);
+      }
+    }
+
+    // normalize currents to integration time
+    const double sliceWidthMS = sliceWidthPS * std::pow(10, -9);
+    const float norm = 1. / float(sliceWidthMS);
```
noferini: I would move also this before of the loop
(you could also define sliceWidthMSinv for clarity, as you prefer)

noferini (changes requested):

matthias-kleiner:
> Dear @matthias-kleiner, I have a couple of (minor) comments concerning the workflow you reported.
> 
> In your tof-workflow config you have o2-tof-reco-workflow --input-type compressed-clusters-ctf --output-type "clusters,disable-writer"
> 
> The input-type you are asking is not defined. It works because ctf already provides digits but it is misleading. I would put "--input-type none" for clarity
> 
> Also disable-writer in the output-type doesn't have any meaning. You have --disable-root-output and it is enough (in any case for data tof-workflow is already defined in O2DPG scripts so you will inherit it from the global process)
> 
> I also did't understand if and how this is supposed to run in async reco.

Hello @noferini 
thanks for your comments. I havent really used the TOF workflows before, so I am not to familiar with them.

> I also did't understand if and how this is supposed to run in async reco.

I am not sure how exactly the async reco is running, but I suppose if an input list of ctfs are used which are ordered in time then it should be fine? We want to run it for the apass 3 to have the currents which we can use for our space-charge distortions studies, but maybe we can also just run it on the GSI and produce the files before apass3.
I think the procedure would be run in a similar way as the space-charge correction map workflows are running, but @chiarazampolli can comment maybe better how it could run during the async reco or if there are potential problems, which I do not see yet.

noferini:
> > Dear @matthias-kleiner, I have a couple of (minor) comments concerning the workflow you reported.
> > In your tof-workflow config you have o2-tof-reco-workflow --input-type compressed-clusters-ctf --output-type "clusters,disable-writer"
> > The input-type you are asking is not defined. It works because ctf already provides digits but it is misleading. I would put "--input-type none" for clarity
> > Also disable-writer in the output-type doesn't have any meaning. You have --disable-root-output and it is enough (in any case for data tof-workflow is already defined in O2DPG scripts so you will inherit it from the global process)
> > I also did't understand if and how this is supposed to run in async reco.
> 
> Hello @noferini thanks for your comments. I havent really used the TOF workflows before, so I am not to familiar with them.
> 
> > I also did't understand if and how this is supposed to run in async reco.
> 
> I am not sure how exactly the async reco is running, but I suppose if an input list of ctfs are used which are ordered in time then it should be fine? We want to run it for the apass 3 to have the currents which we can use for our space-charge distortions studies, but maybe we can also just run it on the GSI and produce the files before apass3. I think the procedure would be run in a similar way as the space-charge correction map workflows are running, but @chiarazampolli can comment maybe better how it could run during the async reco or if there are potential problems, which I do not see yet.

Dear @matthias-kleiner,
if you want to merge and then run the calibrator task with timeslots a posteriori you need to set also a reader which is able to inject the time info in the proper format as stored in o2_tfidinfo.root.
For TOF aasync calibration I encoded tfinfo also in out output object so that it is merged with the calibration infos.

matthias-kleiner:
@chiarazampolli @noferini 
I added a writer device, so we are able to merge the currents afterwards with the `o2_tfidinfo.root` manually.
I think for the async it would be then sufficient to run with 
```
o2-ctf-reader-workflow --copy-cmd no-copy --onlyDet TOF --ctf-input "ctfs.list" --severity warning \
| o2-tof-reco-workflow --input-type none --output-type "clusters" --disable-mc --disable-root-output \
| o2-tof-integrate-cluster-workflow -b
```

noferini:
> @chiarazampolli @noferini I added a writer device, so we are able to merge the currents afterwards with the `o2_tfidinfo.root` manually. I think for the async it would be then sufficient to run with
> 
> ```
> o2-ctf-reader-workflow --copy-cmd no-copy --onlyDet TOF --ctf-input "ctfs.list" --severity warning \
> | o2-tof-reco-workflow --input-type none --output-type "clusters" --disable-mc --disable-root-output \
> | o2-tof-integrate-cluster-workflow -b
> ```

Hi @matthias-kleiner ,
it is very good to habe a writer as separate spec. Thanks for providing it.
Merging o2_tfidinfo.root will be not very comfortable for you since it doesn't scale very well to large stats.

Just in case it can help for TOF we do the following:
- adding TFIDInfo to the calibration object (see DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h)
- filling it in the spec O2/Detectors/TOF/workflow/src/EntropyDecoderSpec.cxx (TFIDInfo is provided by the framework)  : lines   80-82
- create a calib reader injecting both calib data and TFIDInfo: O2/Detectors/TOF/workflowIO/src/CalibInfoReaderSpec.cxx  lines 54 and  82-87 for TFIDInfo

This allows us to merge all calib info as any root file and then just running on it with the reader + the workflow as in sync

matthias-kleiner:
> > @chiarazampolli @noferini I added a writer device, so we are able to merge the currents afterwards with the `o2_tfidinfo.root` manually. I think for the async it would be then sufficient to run with
> > ```
> > o2-ctf-reader-workflow --copy-cmd no-copy --onlyDet TOF --ctf-input "ctfs.list" --severity warning \
> > | o2-tof-reco-workflow --input-type none --output-type "clusters" --disable-mc --disable-root-output \
> > | o2-tof-integrate-cluster-workflow -b
> > ```
> 
> Hi @matthias-kleiner , it is very good to habe a writer as separate spec. Thanks for providing it. Merging o2_tfidinfo.root will be not very comfortable for you since it doesn't scale very well to large stats.
> 
> Just in case it can help for TOF we do the following:
> 
>     * adding TFIDInfo to the calibration object (see DataFormats/Detectors/TOF/include/DataFormatsTOF/Diagnostic.h)
> 
>     * filling it in the spec O2/Detectors/TOF/workflow/src/EntropyDecoderSpec.cxx (TFIDInfo is provided by the framework)  : lines   80-82
> 
>     * create a calib reader injecting both calib data and TFIDInfo: O2/Detectors/TOF/workflowIO/src/CalibInfoReaderSpec.cxx  lines 54 and  82-87 for TFIDInfo
> 
> 
> This allows us to merge all calib info as any root file and then just running on it with the reader + the workflow as in sync

@noferini Many thanks for your help. I am currently trying to implement the reader, but I am struggling to inject the TFIDInfo data. I am doing it like here https://github.com/AliceO2Group/AliceO2/blob/e1a87c1a066727ebdd2468f25bb421811c4cc991/Detectors/TOF/workflowIO/src/CalibInfoReaderSpec.cxx#L83 , but the `o2-tof-merge-integrate-cluster-workflow` which receives the data gets for each `run` call only the same TFInfos i.e. only the first value I injected in the loop. Is there something I am missing, like some update method or something?

I was also looking how it is done for the `HBFUtilsInitializer` and they are setting some `CallbacksPolicy`: 
https://github.com/AliceO2Group/AliceO2/blob/fdfd559a5b550a222d7d9f87b32741f5dc3fc122/Detectors/Raw/src/HBFUtilsInitializer.cxx#L122

I was testing and implemented some dummy `CallbacksPolicy` which seems to work and correctly updates the TFInfo for each `run` call, but I am not sure if something like this would be the appropriate solution for injecting the TFInfos
```
void customize(std::vector<o2::framework::CallbacksPolicy>& policies)
{
  // o2::raw::HBFUtilsInitializer::addNewTimeSliceCallback(policies);
  policies.push_back(o2::framework::CallbacksPolicy{
    [](o2::framework::DeviceSpec const& spec, o2::framework::ConfigContext const& context) -> bool {
      return true;
    },
    [](o2::framework::CallbackService& service, o2::framework::InitContext& context) {
      uint32_t delay = 10;
      service.set(o2::framework::CallbackService::Id::NewTimeslice,
                  [offset = 10, startTime = 20, orbitFirst = 30, runNumber = 40, delay](o2::header::DataHeader& dh, o2::framework::DataProcessingHeader& dph) {
                    int increment = 10;
                    dh.firstTForbit = offset + increment * dh.tfCounter;
                    dh.runNumber = runNumber;
                    dph.creation = startTime + (dh.firstTForbit - orbitFirst);
                    usleep(delay);
                  });
    }});
}
```

matthias-kleiner:
I think the problem is, that I was sending several messages per `run` function. I will try later with only one message (entry in the `TTree`) per `run` call. I assume then it will work

matthias-kleiner:
Hello @noferini I added the reader and tested it. 
So for the async it would be something like 
`o2-ctf-reader-workflow --copy-cmd no-copy --onlyDet TOF --ctf-input list.txt --severity warning | o2-tof-reco-workflow --input-type none --output-type "clusters" --disable-mc --disable-root-output | o2-tof-integrate-cluster-workflow -b`

and for merging the output files 
`o2-tof-integrate-cluster-reader-workflow --tofcurrents-infiles list.txt | o2-tof-merge-integrate-cluster-workflow --debug true -b`

I think this would be fine? Do you have any more suggestions?

noferini (approved):
Hi @matthias-kleiner ,
many thanks!
You were very fast.
Cheers,
Francesco

matthias-kleiner:
> Hi @matthias-kleiner , many thanks! You were very fast. Cheers, Francesco

Hi @noferini okay thanks. Should I squash the commits or is it fine like it is?

noferini:
Hi @matthias-kleiner , it is fine with me.

FullCI fails with the error
[ERROR] pid 30048 (internal-dpl-ccdb-backend) crashed with 128

I think this is not related to this PR... pinging @shahor02 and @ktf . Can you merge it?

matthias-kleiner:
Hello @shahor02 @ktf can you merge this PR?

PR #10610, opened 2023-01-22
TRD raw reader improve RDH check

martenole:


martenole:
Tested on a few TFs from a physics run where usually single HBFs don't exceed 8kB and did not see any RDH errors.
But also for the recent noise run 529969 (where the HBFs are larger) there were not any occurrences of not increasing packet counters in consecutive RDHs.

bazinski (approved):

martenole:
fullCI error unrelated, merging

PR #10611, opened 2023-01-22
AOD stores tracks parameters at calorimeters

shahor02:
@mfasDa @peressounko 

Please check. What it does is:
1) propagates tracks to emcal/dcal X = 440 cm in the proper sector
2) if not in the sector occupied by emcal/dcal (phos) or |eta_position|>0.75 
3) if in the sector with phos AND eta_pos (at emcal X) < 0.178266f (i.e. the mean between the dcal and phos acceptance, @peressounko note that this corresponds to 82.4 cm instead of 83cm you asked), then propagate to phos X=460.

If not failed on any of the previous stages, assigns position eta and phi to already existing trackEta parameters fTrackEtaEMCAL and fTrackPhiEMCAL. Otherwise -999 is assigned to both parameters.

### DIFF
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -232,6 +232,7 @@ class AODProducerWorkflowDPL : public Task
   int mNThreads = 1;
   bool mUseMC = true;
   bool mEnableSV = true;             // enable secondary vertices
+  bool mFieldON = false;
   const float cSpeed = 0.029979246f; // speed of light in TOF units
 
   GID::mask_t mInputSources;
@@ -456,7 +457,7 @@ class AODProducerWorkflowDPL : public Task
                            GIndex trackID, const o2::globaltracking::RecoContainer& data, int collisionID, std::uint64_t collisionBC, const std::map<uint64_t, int>& bcsMap);
 
   TrackExtraInfo processBarrelTrack(int collisionID, std::uint64_t collisionBC, GIndex trackIndex, const o2::globaltracking::RecoContainer& data, const std::map<uint64_t, int>& bcsMap);
-
+  void extrapolateToCalorimeters(TrackExtraInfo& extraInfoHolder, const o2::track::TrackPar& track);
   void cacheTriggers(const o2::globaltracking::RecoContainer& recoData);
 
   // helper for track tables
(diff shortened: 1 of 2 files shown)


ktf:
@shahor02 @davidrohr @sawenzel all the PRs are now failing with:

```
[378:TPCDigitizer_9]: [23:18:34][INFO] TPC: Flushed 7613 digits, 8512 labels and 64 common mode entries
[ERROR] pid 363 (TPCDigitizer_0) crashed with 128
[ERROR] pid 364 (TPCDigitizer_1) crashed with 128
```

in the fullCI. Do you have any idea of what is going on?

shahor02:
Yes, I've seen this but could not reproduce it locally. I asked @sawenzel , he thinks might be resource problem. 
We have this situation since ~1 week, initially some PRs were passing fullCI, now they seem to fail at 100% (always in TPC digitization).

ktf:
I can confirm they get killed by the OOM reaper. Not yet sure why. Maybe related to the increase we also see with some analysis tasks? Did anything particularly relevant change in the last few days?

peressounko:
Dear @shahor02, approach looks reasonable, will test on real/MC AODs once will be committed.

mfasDa:
One thing a bit unfortunate is that the detectors use different radii for propagation (PHOS 460 cm, EMCAL 440 cm). In the overlap region between DCAL and PHOS how can we figure out which track was propagated to which detector?

shahor02:
> One thing a bit unfortunate is that the detectors use different radii for propagation (PHOS 460 cm, EMCAL 440 cm). In the overlap region between DCAL and PHOS how can we figure out which track was propagated to which detector?

The `eta` almost does not change on propagation and there is a good margin to both detectors' acceptance. But, if you want, I can add e.g. 10 to eta in case of propagation to the PHOS.

mfasDa:
> > One thing a bit unfortunate is that the detectors use different radii for propagation (PHOS 460 cm, EMCAL 440 cm). In the overlap region between DCAL and PHOS how can we figure out which track was propagated to which detector?
> 
> The `eta` almost does not change on propagation and there is a good margin to both detectors' acceptance. But, if you want, I can add e.g. 10 to eta in case of propagation to the PHOS.

For the eta I was expecting this, and we can deal here with the eta cut in the correction task as you applied in the track extrapolation. I was more worried about the the two small DCAL supermodules in case the tracks are sufficiently curved. However the effect here is probably small.

peressounko:
Hi @mfasDa @shahor02, one indeed can imagine strongly curved tracks with |eta|<0.1 which will hit both DCal and PHOS. I think, one then should select candidate tracks extrapolated to some radii e.g. 440 with loose cuts and extrapolate once more to a calorimeter under study?

shahor02:
Such tracks are certainly possible but most of such tracks are the secondaries created in the TPC and only a fraction of them will be reconstructible in the TPC. If they are matched in TRD or TOF, then one can trust more or less their eta (though such matches will have a lot of fakes), if they are not matched (and low-pT tracks usually do not produce TRD tracklets) then the time/Z assignment of such tracks is purely conventional so their tagging for being in the EMCAL/PHOS acceptance is anyway random.
Given this irreducible uncertainty, not sure the identification of a small fraction of low-energy tracks hitting both detectors is relevant.

shahor02:
@mfasDa thanks for spotting the problem with the sector definition, there are too many misleading pictures around... Fixed.

shahor02:
@peressounko @mfasDa shall I merge this?

peressounko (approved):
Looks fine, lets test with real data and MC

mfasDa:
Hi @shahor02 , I am also fine with merging!

PR #10612, opened 2023-01-23
[O2-3525] Support for unpadded raw data, RDHv7

shahor02:
* Account for RDH v6 modification: padding status
  + getDataFormat and setDataFormat for RDHUtils and RawDataWriter helpers.

* Support unpadded GBT words in ITS/MFT raw data decoding and encoding

### DIFF
--- a/DataFormats/Headers/include/Headers/RAWDataHeader.h
+++ b/DataFormats/Headers/include/Headers/RAWDataHeader.h
@@ -42,6 +42,102 @@ namespace header
 /// assume that this is the only type the software has to support (based on
 /// experience with previous systems)
 ///
+/// RAWDataHeaderV7
+/// V7 introduces a detector source ID field
+/// Description of the fields can be found here
+/// https://gitlab.cern.ch/AliceO2Group/wp6-doc/-/blob/master/rdh/RDHv7.md
+//
+///
+///       63     56      48      40      32      24      16       8       0
+///       |---------------|---------------|---------------|---------------|
+///
+///       | reserved              | priori|               |    header     |
+/// 0     | reserve zero  |Source | ty bit|    FEE id     | size  |version|
+///
+/// 1     |ep | cru id    |pcount|link id |  memory size  |offset nxt pack|
+///
+/// 2     |                orbit          | reserved         |bunch cross |
+///                                                               | data
+/// 3     |                          reserved                     | format|
+///
+/// 4     |  zero | stop  |   page count  |      trigger type             |
+///
+/// 5     |                          reserved                             |
+///
+/// 6     |      zero     | detector par  |         detector field        |
+///
+/// 5     |                          reserved                             |
+///
+struct RAWDataHeaderV7 {
+  union {
+    // default value
+    uint64_t word0 = 0x00000000ffff4007;
+    //                       | |     | version 67
+    //                       | |   | 8x64 bit words = 64 (0x40) byte
+    //                       | | invalid FEE id
+    //                       | priority bit 0
+    struct {
+      uint64_t version : 8;        /// bit  0 to  7: header version
+      uint64_t headerSize : 8;     /// bit  8 to 15: header size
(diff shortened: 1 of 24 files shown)


noferini:
Hi @shahor02 ,
I'm trying to run some tests with this PR (RDHV7)
I temporary needed to fix
DataFormats/Headers/include/Headers/RDHAny.h
to support for V7 when moving currentHeader -> V7 (trivial fix)
Cheers,
Francesco

shahor02:
> I temporary needed to fix DataFormats/Headers/include/Headers/RDHAny.h to support for V7 when moving currentHeader -> V7 (trivial fix) Cheers, Francesco

@noferini if you refer to the `static_assert` in the RDHAny, which broke CIs, I've just fixed it. Otherwise, please let me know what was your fix.

noferini:
> > I temporary needed to fix DataFormats/Headers/include/Headers/RDHAny.h to support for V7 when moving currentHeader -> V7 (trivial fix) Cheers, Francesco
> 
> @noferini if you refer to the `static_assert` in the RDHAny, which broke CIs, I've just fixed it. Otherwise, please let me know what was your fix.

I referred to that. Thanks for fixing it.
Cheers

noferini:
Hi @shahor02 ,
one more thing. I am fighting with the encoding digit -> raw and try to set nopadding option in the dataformat.
Unfortunately I didn't find any way to propagate the setting to the RawWriter (and than to written RDH)
I have
  mFileWriter.useRDHVersion(RDHUtils::getVersion<o2::header::RAWDataHeader>());

and then
      mFileWriter.addData(Geo::getFEEid(icrate), Geo::getCRUid(icrate), Geo::getCRUlink(icrate), Geo::getCRUendpoint(icrate), mIR, gsl::span(mBuffer[icrate], nbyte));


I cannot pass the padding option to the writer.
Is there a way or has it to be implemented?
Cheers,
Francesco

shahor02:
@noferini see what I did for the its/mft mc->raw: https://github.com/AliceO2Group/AliceO2/blob/c08c9bc2679c028c7ec63ca866ca58832f27d2d0/Detectors/ITSMFT/ITS/simulation/src/digi2raw.cxx#L160-L166

```
  m2r.getWriter().useRDHVersion(rdhV); // version ID to use
  m2r.getWriter().useRDHDataFormat(enablePadding ? 0 : 2);  // format value to be added (only in case v7 is used).
                  // for TOF it should be "enablePadding ? 0 : 1" (in case you want to optionally allow padding even with v7, 
                  // I've added this as a command line option)

  if (!enablePadding) { // CRU page alignment padding is used only if no GBT word padding is used
    m2r.getWriter().setAlignmentSize(o2::itsmft::GBTLink::CRUPageAlignment);
    m2r.getWriter().setAlignmentPaddingFiller(0xff);
  }
```
The lines in the `if` above tell to writer to automatically pad the CRU page (when closing) by 0xff, aligning it to 32 bytes.
This is done [here](https://github.com/AliceO2Group/AliceO2/blob/c08c9bc2679c028c7ec63ca866ca58832f27d2d0/Detectors/Raw/src/RawFileWriter.cxx#L479-L483) and [here](https://github.com/AliceO2Group/AliceO2/blob/c08c9bc2679c028c7ec63ca866ca58832f27d2d0/Detectors/Raw/src/RawFileWriter.cxx#L514-L517)

For TOF it would require `setAlignmentSize(16)` but since you are filling not with a single symbol but by 8 bytes long special word, you cannot use this method directly. You can either use the call-back hooks to align the page:
https://github.com/AliceO2Group/AliceO2/blob/c08c9bc2679c028c7ec63ca866ca58832f27d2d0/Detectors/Raw/README.md?plain=1#L101-L170
or I can add a `special` filler word option (apart from the setAlignmentPaddingFiller) which will be preferred if set. Let me know what do you prefer.

noferini:
It works!
Thanks a lot
Francesco

PR #10616, opened 2023-01-23
TPC SAC: Changing name of output

matthias-kleiner:
- adding check for empty data before writing to CCDB

### DIFF
--- a/Detectors/TPC/workflow/include/TPCWorkflow/TPCFactorizeSACSpec.h
+++ b/Detectors/TPC/workflow/include/TPCWorkflow/TPCFactorizeSACSpec.h
@@ -78,7 +78,7 @@ class TPCFactorizeSACSpec : public o2::framework::Task
   }
 
   static constexpr header::DataDescription getDataDescriptionSAC1() { return header::DataDescription{"SAC1"}; }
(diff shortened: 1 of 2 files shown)


Detectors/TPC/workflow/include/TPCWorkflow/TPCFactorizeSACSpec.h
```diff
@@ -78,7 +78,7 @@ class TPCFactorizeSACSpec : public o2::framework::Task
   }
 
   static constexpr header::DataDescription getDataDescriptionSAC1() { return header::DataDescription{"SAC1"}; }
-  static constexpr header::DataDescription getDataDescriptionTimeStamp() { return header::DataDescription{"FOURIERTS"}; }
+  static constexpr header::DataDescription getDataDescriptionTimeStamp() { return header::DataDescription{"FOURIERTSSAC"}; }
   static constexpr header::DataDescription getDataDescriptionFourier() { return header::DataDescription{"FOURIER"}; }
```
davidrohr: should we add SAC for `getDataDescriptionFourier` as well?

matthias-kleiner: The `getDataDescriptionFourier` is a left over from an older version. It is not used at the moment. I will remove it for the IDCs and SACs

PR #10617, opened 2023-01-23
Revert "Add metadata to AOD writer inputs (#10531)"

chiarazampolli:
This reverts commit f779caf4b986834174514fe310d26e62b3b1477b.
@nburmaso , the AOD producer gets stuck with the PR above. I also tried to include https://github.com/AliceO2Group/AliceO2/pull/10556, but unless I made a mistake, nothing changed. Since we need to use the latest tag for apass2, I think we should revert for now f779caf4b986834174514fe310d26e62b3b1477b, then we can check better for apass3. 

@shahor02 , can you please take a look?


chiarazampolli:
Pinging also @catalinristea @noferini since this is needed to continue apass2.

PR #10618, opened 2023-01-23
Fix outlier rejection for B=0 data

martenole:
- optionally allow skipping outlier rejection completely

@wiechula I tested this for run 520495 and with that most tracks can be interpolated successfully

### DIFF
--- a/Detectors/TPC/calibration/SpacePoints/src/TrackInterpolation.cxx
+++ b/Detectors/TPC/calibration/SpacePoints/src/TrackInterpolation.cxx
@@ -321,7 +321,7 @@ void TrackInterpolation::interpolateTrack(int iSeed)
   trackData.clAvailTOF = gidTable[GTrackID::TOF].isIndexSet() ? 1 : 0;
 
   TrackParams params; // for refitted track parameters and flagging rejected clusters
-  if (validateTrack(trackData, params, clusterResiduals)) {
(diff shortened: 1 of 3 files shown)


wiechula (approved):
@martenole , thanks for the quick implementation!

Detectors/TPC/calibration/SpacePoints/src/TrackInterpolation.cxx
```diff
@@ -437,7 +437,7 @@ bool TrackInterpolation::validateTrack(const TrackData& trk, TrackParams& params
     LOG(debug) << "Skipping track too far from helix approximation";
     return false;
   }
-  if (fabsf(params.qpt) > mParams->maxQ2Pt) {
+  if (fabsf(mBz) > 0.01 && fabsf(params.qpt) > mParams->maxQ2Pt) {
```
shahor02, line 440: What is the reason of this protection? For B=0 the track is supposed to have default pT~0.6 GeV (modulo energy loss corrections). Asking just to be sure that you have not seen too large momentum tracks at B=0.

martenole: This is just because the `params.qpt` here is recalculated within the `compareToHelix` method. And that does not work for B=0. So I have to avoid this check

martenole:
Just for the record here the pT spectrum I see for run 520495
ptSpectrumB0

martenole:
CI failure unrelated, tested locally, merging

PR #10619, opened 2023-01-24
Improve profiling support

ktf:


dependencies/O2CompileFlags.cmake
```diff
@@ -122,6 +122,13 @@ elseif(UNIX)
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-undefined") # avoid undefined in our libs
 endif()
 
+if(ENABLE_TIME_TRACE)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ftime-trace")
```
davidrohr: I would use `set(CMAKE_CXX_FLAGS_${CMAKE_BUILD_TYPE} "${CMAKE_CXX_FLAGS_${CMAKE_BUILD_TYPE}} -ftime-trace")`, which is shorter and more generic

ktf:
Done, also cleaned up some other similar thing. Not sure it's 100% the same, but it should be ok.

PR #10620, opened 2023-01-24
DPL: fix inputs in GUI

ktf:


ktf:
@davidrohr @shahor02 here is the fix for the missing inputs.

ktf:
For the record, the GUI broke due to 05ca63298d98df124d9d295ad4a3c639f3e005e9 which commented out the metric sending.

davidrohr:
> For the record, the GUI broke due to [05ca632](https://github.com/AliceO2Group/AliceO2/commit/05ca63298d98df124d9d295ad4a3c639f3e005e9) which commented out the metric sending.

uhm, I am very sorry. Not sure how that commenting out made it into that PR.....
Could it be that back then you had asked me to disable the metric sending, since it was crashing?

ktf:
No worries.  I suspect it was just the first one of things which broke with the new fmt, it got commented out, and then we forgot to fix it like it was actually done for other similar issues in the same PR.

PR #10622, opened 2023-01-24
More work on space charge distortion correction maps

martenole:
Please don't merge yet. I am still testing, but that takes rather long mostly due to slow access to CCDB objects.

- voxel is defined by cluster position, not track interpolation position
- allow to apply correction map to cluster residuals in order to check its consistency
- some improvements on the accounting in the residual aggregator


martenole:
For comparison these are the distortions in Y extracted for one 10 minute interval of run 527976:
referenceMap
And here the results for the same time interval when I apply the reference map to the residuals before processing them:
consistencyCheck

martenole:
fullCI failure is unrelated. Shall I merge this @wiechula ?
The largest change is the change from using the track interpolation point to find the voxel bin to using the TPC cluster position instead

PR #10624, opened 2023-01-25
Fix deprecation warning

ktf:


davidrohr (dismissed):
thx :)

ktf:
Maybe we should have some function which returns the longest filename for a given system, since 1024 is quite conservative.

ktf:
I forgot one. Updated.

ktf:
Error unrelated. Merging.

PR #10625, opened 2023-01-25
Small explanation on boxgen

sarahherrmann:


sawenzel (approved):
Documentation improvement really appreciated! Note that we are also setting up a new page here `https://aliceo2group.github.io/simulation/`. Will need to make sure to transport all content from here to there.

PR #10628, opened 2023-01-25
DPL: Fixes in timers

knopers8:
See the commit messages for details. I can split the PR into three if that's preferred.

### DIFF
--- a/Framework/Core/include/Framework/DeviceState.h
+++ b/Framework/Core/include/Framework/DeviceState.h
@@ -18,7 +18,6 @@
 #include <string>
 #include <map>
 #include <utility>
-#include <set>
 
 typedef struct uv_loop_s uv_loop_t;
 typedef struct uv_timer_s uv_timer_t;
@@ -91,7 +90,7 @@ struct DeviceState {
   // The list of active timers which notify this device.
   std::vector<uv_timer_t*> activeTimers;
   // The list of timers fired in this loop
-  std::set<uv_timer_t*> firedTimers;
+  std::vector<uv_timer_t*> firedTimers;
   // The list of pollers for active input channels
   std::vector<uv_poll_t*> activeInputPollers;
   // The list of pollers for active output channels
--- a/Framework/Core/include/Framework/DeviceState.h
+++ b/Framework/Core/include/Framework/DeviceState.h
@@ -18,6 +18,7 @@
 #include <string>
 #include <map>
 #include <utility>
+#include <set>
 
 typedef struct uv_loop_s uv_loop_t;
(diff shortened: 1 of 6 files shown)


Framework/Core/src/DeviceSpecHelpers.cxx
```diff
@@ -92,7 +92,8 @@ struct ExpirationHandlerHelpers {
       auto* timer = (uv_timer_t*)(malloc(sizeof(uv_timer_t)));
       timer->data = &state;
       uv_timer_init(state.loop, timer);
-      uv_timer_start(timer, detail::timer_callback, period / 1000, period / 1000);
+      constexpr size_t resolution = 100;
+      uv_timer_start(timer, detail::timer_callback, period / 1000 / resolution, period / 1000 / resolution);
```
ktf: I am not sure about this. Do you mean that you want to fire a timer 100 times per second in order to satisfy a 1 second timer? Maybe resolution should be a function of the period / smaller number, no? A lapping timer does not come for free.

knopers8: > Do you mean that you want to fire a timer 100 times per second in order to satisfy a 1 second timer?

Yes, exactly. Not ideal, I agree, though I checked that the CPU usage with this was still below 1%. However, I cannot defend the ugliness of this approach ;)

> Maybe resolution should be a function of the period / smaller number, no?

Probably 10 would be still fine for the problem with QC on EPNs, but my benchmarks, which count on the cycle period being as requested at least on average, would still produce bad results.

I will then play with another idea to solve it, but this might take a few more hours.

ktf: maybe you could fire it at 90% of the interval the first time (or whatever you think the the jitter will be) and then some smaller intervals.

Framework/Core/include/Framework/DeviceState.h
```diff
@@ -89,6 +90,8 @@ struct DeviceState {
   uv_loop_t* loop = nullptr;
   // The list of active timers which notify this device.
   std::vector<uv_timer_t*> activeTimers;
+  // The list of timers fired in this loop
+  std::set<uv_timer_t*> firedTimers;
```
ktf: no sets in headers, please. Use an std vector

knopers8: sure, done

knopers8:
@ktf can you have a look again please?

ktf:
Merging this since it passes all the tests and I need to have a parallel development to allow for the variable timers. We can take it from here.

PR #10629, opened 2023-01-25
[FOCAL-11] Drop empty timeframes in reader

mfasDa:
In case of uncompressed data the empty timeframes
must be ignored in the reader, otherwise it leads to a crash. The data can however also close with empty
timeframes. In this case the reader would at the end return an empty event, for which an initialize flag is used to determine the init status. Uninitialized events are discarded by the analysis manager.


mfasDa:
CI errors unrelated

PR #10630, opened 2023-01-25
Add TTree aliases for scd map results

martenole:


martenole:
https://github.com/AliceO2Group/AliceO2/pull/10630/commits/585c1f56db55dc6d2e046acdaa2517204328df18 adds the possibility to do the vDrift calibration when creating the map from unbinned residuals. Currently only added to the macro for testing purposes. When validated this will be added to the production workflow. 
But probably the fit will then be done to the binned residuals, since the residual aggregator which does the binning is already on the edge in terms of processing time

PR #10634, opened 2023-01-26
Possibility to set MeanVtx calib parameter from outside

chiarazampolli:
@shahor02 , needed for async calib.


shahor02 (approved):
It is fine, but it will not prevent from creating a new time-slot *before* the one created for the 1st TF seen if the following TF happens to be prior to the 1st one:
https://github.com/AliceO2Group/AliceO2/blob/6818bb14d8113766b4d03302cd47de6a930a6ce7/Detectors/Calibration/include/DetectorsCalibration/TimeSlotCalibration.h#L387

The `DELAY..` acts only on the end of the timeslot, not its start.

chiarazampolli:
For reference: needed for https://github.com/AliceO2Group/O2DPG/pull/835

PR #10636, opened 2023-01-26
Attempt at reducing memory usage in CI

ktf:


ktf:
@shahor02 @davidrohr @sawenzel I tried to reduce the number of TPC lanes, but it then dies in the MFT part anyway. Shall we simply revert the MFT changes?

PR #10638, opened 2023-01-26
[FOCAL-10, FOCAL-11] Fixes in pixel decoding

mfasDa:
- Typo in payload size determination was leading to
  chopped payload sent to the pixel decoder
- New fileds added in PixelChip (and corresponding
  record):
  + feeID, needed to distinguish chips with same
    lane and chip ID
  + status code for propagating busy status


mfasDa:
CI errors unrelated

shahor02 (approved):

PR #10559, opened 2023-01-13
Upgrade MFT track finder

rpezzi:
This commit improves the geometrical considerations used to sort clusters in R,Phi bins on each layer and related allocation of memory.

Geometry optimization changes:
- Cluster sorting binning defined by a cylindrical volume around Z axis, configurable by tracking parameters
- Variable radial bin size for each layer

Memory and CPU optimization changes:
- Move cluster sorting information from ROFrame to tracker class
- Reduce memory footprint of MFT tracker
- Cluster sorting moved to individual threads
- Combines LTF and CA track finding for each ROF (was TF)
- Relations between cluster bins now static member of tracker
- Single stopwatch for track-finding (LTF and CA now run together)

Other changes:
- Configurable parameter for true/fake track flag threshold

PR #10639, opened 2023-01-26
Temporary revert recent MFT tracking changes to check if they are responsible for fullCI failures

shahor02:
@rpezzi there is an [indication](https://alice.its.cern.ch/jira/browse/O2-3520) that https://github.com/AliceO2Group/AliceO2/pull/10559 actually increases the memory consumption by 1 GB which makes fullCI to fail due to the OOM on the builder (as you remember, it was merged only after your private test). 
Could you please fix this large memory allocation?

I've spent ~2 h trying to remove the dependence of the `o2-sim-digitizer-workflow` on the `O2::MFTTracking` but failed. please check https://alice.its.cern.ch/jira/browse/O2-3521


shahor02:
@rpezzi so, the MFT PR was the reason, merging. @sawenzel , thanks for spotting.

PR #10643, opened 2023-01-27
Complain if more than output route

ktf:
Not sure how this can possibly work with pipelining.


ktf:
For the record, the TOF Compressor also triggers my test for possible missing timeframe data on output (still to be merged). What is preventing this to use regular DPL to send the data?

ktf:
@noferini could you explicitly agree? I see no way this can work with more than one output route and it will fail in a mostly silent way.

noferini:
> For the record, the TOF Compressor also triggers my test for possible missing timeframe data on output (still to be merged). What is preventing this to use regular DPL to send the data?

In principle nothing, in practice I need to speak with @preghenella which is at the moment in vacation. This is something we could plan, if he agree, but probably not in a short time scale

noferini:
> @noferini could you explicitly agree? I see no way this can work with more than one output route and it will fail in a mostly silent way.
I run a local test with your change and it works without any problem. For me it is fine to add this protection.

PR #10645, opened 2023-01-27
Flag to allow duplicate processing

martenole:
If requested, when e.g. an ITS-TPC-TRD track is processed also its ITS-TPC seed is processed (if it survives the cuts for ITS-TPC only tracks)
ping @aschmah


martenole:
All checks passed :) Merging now

PR #10646, opened 2023-01-27
Clusterization workflow updeated

gvolpe79:


shahor02:
Hi @gvolpe79 
The CI errors are genuine: 
```
/sw/SOURCES/O2/10646-slc8_x86-64/0/Detectors/HMPID/workflow/src/digits-to-clusters-workflow.cxx:92: undefined reference to `o2::hmpid::getClustersToRootWriter()'
```
But apart from that, could you modify the PR to follow the same modularity / options as other devices? The DigitsToCluster device should not open itself the digit file, instead, it must always get the data as input provided by digit reader device initialized by the workflow. The conventions are: with option `--disable-root-input` data comes from upstream (no digits reader is initialized). Similarly, `--disable-root-output` should not initialize cluster writer. By default both reader and writer are initialized. see e.g. 
O2/Detectors/FIT/FT0/workflow/src/ft0-reco-workflow.cxx

Detectors/HMPID/workflow/CMakeLists.txt
```diff
@@ -79,10 +79,11 @@ o2_add_executable(digits-to-root-workflow
               SOURCES src/digits-to-root-workflow.cxx
               PUBLIC_LINK_LIBRARIES O2::HMPIDWorkflow)
 
-o2_add_executable(clusters-to-root-workflow
-              COMPONENT_NAME hmpid
-              SOURCES src/clusters-to-root-workflow.cxx
-              PUBLIC_LINK_LIBRARIES O2::HMPIDWorkflow)
+# ef : should be removed:
```
shahor02: @gvolpe79 what was the reason to suppress the standalone cluster-writer?

shahor02:
@gvolpe79 the fullCI error seems to be genuine, though from the log I don't understand what is the problem. Will need to fetch your branch and run fullC locally...

shahor02:
Closing to open a fixed version

PR #10650, opened 2023-01-30
Add drift time offset to VDrift calibration object

shahor02:


wiechula (dismissed):
Thanks @shahor02 for the quick implementation!

Detectors/TPC/simulation/src/Digitizer.cxx
```diff
@@ -106,7 +106,7 @@ void Digitizer::process(const std::vector<o2::tpc::HitGroup>& hits,
           LOG(warning) << "Skipping electron with driftTime " << driftTime << " from hit at time " << hitTime;
           continue;
         }
-        const float absoluteTime = eleTime + (mEventTime - mOutputDigitTimeOffset); /// in us
+        const float absoluteTime = eleTime + mTDriftOffset + (mEventTime - mOutputDigitTimeOffset); /// in us
```
shahor02, line 109: Hi @wiechula could you please confirm that this is enough to impose the drift time offset (defined in \mus) in the simulation?

wiechula: Hi @shahor02 , yes this should be the only place.

PR #10654, opened 2023-01-31
 [MFT] Reduce MaxRPhiBins to minimize memory size 

rpezzi:
Tentative quickfix to  #10559

### DIFF
--- a/Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
+++ b/Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
@@ -23,6 +23,9 @@ namespace o2
 {
 namespace mft
 {
+
+using namespace constants::mft;
+
 class TrackerConfig
 {
  public:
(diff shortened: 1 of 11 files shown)


Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
```diff
@@ -46,27 +55,41 @@ class TrackerConfig
...
+  Float_t mRCutAtZmin;
   Bool_t mLTFConeRadius;
   Bool_t mCAConeRadius;
-
   /// Special track finder for TED shots and cosmics, with full scan of the clusters
   bool mFullClusterScan = false;
+  Float_t mTrueTrackMCThreshold; // Minimum fraction of correct MC labels to tag True tracks
 
-  ClassDefNV(TrackerConfig, 2);
+  static Float_t mPhiBinSize;
```
shahor02, line 70: @rpezzi what is the reason to have still relatively large arrays as static? In this case they are eating up memory in the every process the MFT tracking is linked to (and at the moment a lot of devices). Cannot you init them during the tracking initializing?

rpezzi: Hi @shahor02 , these arrays are static so that every tracker thread can use the same, but indeed these could be initialized at runtime by the tracker. Unfortunately I am away and won't be able to work on that.

Another point, `mBins` and `mBinsS` are in fact identical with the current tracker. Therefore one can be suppressed.

shahor02 (approved):
@rpezzi ok, thanks, at the moment merging as is, since tests are passed. Later will modify to dynamic creation of `mBins` arrays.

PR #10655, opened 2023-01-31
[EMCAL-830] Handling of MC labels

mfasDa:
MC labels need to be adapted when applying
bad channel masking in case of simulation:
- CellRecalibrator must return the indices of
  kept cells (assume range-based iterator to
  keep order)
- Labels are filled with the same index as
  corresponding cells, therefore the labels
  with the relative kept indices are added to the
  output MC container

### DIFF
--- a/Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
@@ -23,7 +23,7 @@
 
 using namespace o2::emcal;
 
-CellRecalibratorSpec::CellRecalibratorSpec(uint32_t outputspec, LEDEventSettings ledsettings, bool badChannelCalib, bool timeCalib, bool gainCalib, std::shared_ptr<o2::emcal::CalibLoader> calibHandler) : mOutputSubspec(outputspec), mLEDsettings(ledsettings), mCalibrationHandler(calibHandler)
+CellRecalibratorSpec::CellRecalibratorSpec(uint32_t outputspec, LEDEventSettings ledsettings, bool badChannelCalib, bool timeCalib, bool gainCalib, bool isMC, std::shared_ptr<o2::emcal::CalibLoader> calibHandler) : mOutputSubspec(outputspec), mIsMC(isMC), mLEDsettings(ledsettings), mCalibrationHandler(calibHandler)
 {
   setRunBadChannelCalibration(badChannelCalib);
   setRunTimeCalibration(timeCalib);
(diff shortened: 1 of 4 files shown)


mfasDa:
Relevant for anchored simulations: In anchored simulations the bad channel calibration must be running as well (only one of the 3) to remove cells in towers which are masked in data. Since this affects the cell container the corresponding MCTruthContainer with the labels must be adapted as well, which means labels of cells which are masked have to be removed. Since indices in the cell container match indicies in the MC truth container it is sufficient to return the local cell indices which are kept (assuming forward iteration in the range-based iterator used in the calibrator). All labels of kept indices are then copied to the new MCTruthContainer. Like this correspondece between cells and MC labels is preserved.

mfasDa:
CI errors unrelated.

Detectors/EMCAL/workflow/src/CellRecalibratorSpec.cxx
```diff
@@ -102,6 +115,10 @@ void CellRecalibratorSpec::run(framework::ProcessingContext& ctx)
   // send recalibrated objects
   ctx.outputs().snapshot(o2::framework::Output{o2::header::gDataOriginEMC, "CELLS", mOutputSubspec, o2::framework::Lifetime::Timeframe}, outputcells);
   ctx.outputs().snapshot(o2::framework::Output{o2::header::gDataOriginEMC, "CELLSTRGR", mOutputSubspec, o2::framework::Lifetime::Timeframe}, outputtriggers);
+  if (outputMCLabels.has_value()) {
+    LOG(info) << "Timeframe: " << inputMCLabels->getIndexedSize() << " label entries read, " << outputMCLabels->getIndexedSize() << " label entries kept";
```
shahor02, line 119: isn't this too verbose? Would not reconstruction of MC data be swamped by these messages?

mfasDa: It has the same granularity as for cells itself, once message per run call (timeframe)
``` c++
  LOG(info) << "Timeframe: " << inputcells.size() << " cells read, " << outputcells.size() << " cells kept";
  if (mLEDsettings == LEDEventSettings::REDIRECT) {
    LOG(info) << "Redirecting " << ledcells.size() << " LED cells from " << ledtriggers.size() << " LED triggers";
  }

  // send recalibrated objects
  ctx.outputs().snapshot(o2::framework::Output{o2::header::gDataOriginEMC, "CELLS", mOutputSubspec, o2::framework::Lifetime::Timeframe}, outputcells);
  ctx.outputs().snapshot(o2::framework::Output{o2::header::gDataOriginEMC, "CELLSTRGR", mOutputSubspec, o2::framework::Lifetime::Timeframe}, outputtriggers);
  if (outputMCLabels.has_value()) {
    LOG(info) << "Timeframe: " << inputMCLabels->getIndexedSize() << " label entries read, " << outputMCLabels->getIndexedSize() << " label entries kept";
    ctx.outputs().snapshot(o2::framework::Output{o2::header::gDataOriginEMC, "CELLSMCTR", mOutputSubspec, o2::framework::Lifetime::Timeframe}, outputMCLabels.value());
  }
  ```
  I can reduce the level to detail if desired. I currently keep one message at start and one at end of run.

shahor02 (approved):
ok, thanks, misunderstood the code.

PR #10658, opened 2023-02-01
[WIP] TPC QC: Update to Cluster Task

tklemenz:
This adds an occupancy plot to the cluster task

### DIFF
--- a/Detectors/TPC/qc/src/Clusters.cxx
+++ b/Detectors/TPC/qc/src/Clusters.cxx
@@ -22,8 +22,10 @@
 #include "TPCBase/ROC.h"
 #include "TPCBase/CRU.h"
 #include "TPCBase/Mapper.h"
+#include "TPCBase/ParameterElectronics.h"
 #include "DataFormatsTPC/ClusterNative.h"
 #include "DataFormatsTPC/KrCluster.h"
+#include "CommonConstants/LHCConstants.h"
 
 ClassImp(o2::tpc::qc::Clusters);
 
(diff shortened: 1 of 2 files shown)


tklemenz:
@wiechula I added [WIP] but it should be fine. Can you have a look please?

Detectors/TPC/qc/src/Clusters.cxx
```diff
@@ -119,7 +121,11 @@ void Clusters::normalize()
   mSigmaPad /= mNClusters;
   mTimeBin /= mNClusters;
 
+  mOccupancy = mNClusters;
+  mOccupancy /= float(mProcessedTFs * (o2::constants::lhc::LHCMaxBunches * 128) / float(o2::tpc::ParameterElectronics::TIMEBININBC));
```
wiechula: There is a dicussion to change the TF length to something shorter, so it would make sense to not hardcode this value here. Better make it a member, which can be set in the workflow.
We will need to check with Ruben what would be the proper way.

wiechula (changes requested):

Detectors/TPC/qc/src/Clusters.cxx
martenole: The device should subscribe to the GRP information from ECS and then its method `GRPECSObject::getNHBFPerTF()` can be used. The subscription is in the example on slides 3-5 here: https://docs.google.com/presentation/d/1WPNixo57zAn_gM6vMCIBOHs-DZqcKAkLX0CwUgO_VYg/edit#slide=id.g10e79f44b99_0_7

chiarazampolli: Hello, 
If this is a QC task, for now the fetcher is not used. You can check other detectors QC code to see how they access CCDB, e.g.
Modules/EMCAL/src/ClusterTask.cxx
Chiara

tklemenz: Ok, thanks for the hints. I'll have a look asap.

PR #10659, opened 2023-02-01
DPL: Avoid exposing boost / DeviceInfo in Services

ktf:


ktf:
10% faster compilation of the framework on my box.

PR #10664, opened 2023-02-01
TPC: Improve cluster and track filter

wiechula:
- add different options to store clusters
  * inside the track
  * separate tree
  * separate tree and file
- add option to store clusters not associated to tracks
- add functions to apply corrections


wiechula:
The errors look unrelated, can this be merged?

PR #10665, opened 2023-02-01
[MFT] Move static arrays to be initialized before the processing starts

robincaron13:
To follow PR#10654, here the arrays are kept as static member and are moved to the tracker class to be initialized before the processing starts (not sure if it was what you were thinking @shahor02)

### DIFF
--- a/Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
+++ b/Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
@@ -25,6 +25,7 @@ namespace mft
 {
 
 using namespace constants::mft;
+using BinContainer = std::array<std::array<std::array<std::vector<Int_t>, constants::index_table::MaxRPhiBins>, (constants::mft::LayersNumber - 1)>, (constants::mft::LayersNumber - 1)>;
 
 class TrackerConfig
 {
@@ -39,7 +40,13 @@ class TrackerConfig
   const Int_t getPhiBinIndex(const Float_t phi) const;
   const Int_t getBinIndex(const Int_t rIndex, const Int_t phiIndex) const;
 
+  static void initBins(const BinContainer* mBinContainer);
+  static void initBinContainers();
+
   const std::pair<Int_t, Int_t>& getClusterBinIndexRange(Int_t layerId, Int_t bin) const { return mClusterBinIndexRange[layerId][bin]; }
+
+  static std::unique_ptr<BinContainer> mBins;
+  static std::unique_ptr<BinContainer> mBinsS;
 
  protected:
   // tracking configuration parameters
@@ -63,10 +70,13 @@ class TrackerConfig
   bool mFullClusterScan = false;
   Float_t mTrueTrackMCThreshold; // Minimum fraction of correct MC labels to tag True tracks
 
+  static std::mutex sTCMutex;
+
   static Float_t mPhiBinSize;
   static Float_t mInversePhiBinSize;
   static std::array<Float_t, constants::mft::LayersNumber> mInverseRBinSize;
-
+  static std::array<Int_t, constants::mft::LayersNumber> mPhiBinWin;
+  static std::array<Float_t, constants::mft::LayersNumber> mRBinSize;
   std::array<std::array<std::pair<Int_t, Int_t>, constants::index_table::MaxRPhiBins>, constants::mft::LayersNumber> mClusterBinIndexRange;
 
   ClassDefNV(TrackerConfig, 3);
@@ -75,6 +85,10 @@ class TrackerConfig
 inline Float_t TrackerConfig::mPhiBinSize;
 inline Float_t TrackerConfig::mInversePhiBinSize;
 inline std::array<Float_t, constants::mft::LayersNumber> TrackerConfig::mInverseRBinSize;
+inline std::array<Int_t, constants::mft::LayersNumber> TrackerConfig::mPhiBinWin;
+inline std::array<Float_t, constants::mft::LayersNumber> TrackerConfig::mRBinSize;
+inline std::unique_ptr<BinContainer> TrackerConfig::mBins;
+inline std::unique_ptr<BinContainer> TrackerConfig::mBinsS;
 
 inline const Int_t TrackerConfig::getRBinIndex(const Float_t r, const Int_t layer) const
 {
--- a/Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
+++ b/Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
@@ -26,6 +26,9 @@ namespace mft
 
 using namespace constants::mft;
 using BinContainer = std::array<std::array<std::array<std::vector<Int_t>, constants::index_table::MaxRPhiBins>, (constants::mft::LayersNumber - 1)>, (constants::mft::LayersNumber - 1)>;
+using RArray = std::array<Float_t, constants::mft::LayersNumber>;
+using PhiArray = std::array<Int_t, constants::mft::LayersNumber>;
+using InverseRArray = std::array<Float_t, constants::mft::LayersNumber>;
 
 class TrackerConfig
 {
@@ -40,13 +43,9 @@ class TrackerConfig
   const Int_t getPhiBinIndex(const Float_t phi) const;
   const Int_t getBinIndex(const Int_t rIndex, const Int_t phiIndex) const;
 
-  static void initBins(const BinContainer* mBinContainer);
   static void initBinContainers();
 
   const std::pair<Int_t, Int_t>& getClusterBinIndexRange(Int_t layerId, Int_t bin) const { return mClusterBinIndexRange[layerId][bin]; }
-
-  static std::unique_ptr<BinContainer> mBins;
-  static std::unique_ptr<BinContainer> mBinsS;
 
  protected:
   // tracking configuration parameters
@@ -74,9 +73,12 @@ class TrackerConfig
 
   static Float_t mPhiBinSize;
   static Float_t mInversePhiBinSize;
-  static std::array<Float_t, constants::mft::LayersNumber> mInverseRBinSize;
-  static std::array<Int_t, constants::mft::LayersNumber> mPhiBinWin;
-  static std::array<Float_t, constants::mft::LayersNumber> mRBinSize;
+
+  static std::unique_ptr<InverseRArray> mInverseRBinSize;
+  static std::unique_ptr<PhiArray> mPhiBinWin;
+  static std::unique_ptr<RArray> mRBinSize;
+  static std::unique_ptr<BinContainer> mBins;
+  static std::unique_ptr<BinContainer> mBinsS;
   std::array<std::array<std::pair<Int_t, Int_t>, constants::index_table::MaxRPhiBins>, constants::mft::LayersNumber> mClusterBinIndexRange;
 
   ClassDefNV(TrackerConfig, 3);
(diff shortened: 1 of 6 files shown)


shahor02 (commented):
Hi @robincaron13 
Thanks for taking care, but this not what I meant: moving the static array from the base class to the derived one will change nothing: it will still be allocated w/o the creation of the actual object, i.e. in any process which links to this library. 
Instead, I would define in the `TrackerConfig.h`

```
using BinContainer = std::array<std::array<std::array<std::vector<Int_t>, constants::index_table::MaxRPhiBins>, (constants::mft::LayersNumber - 1)>, (constants::mft::LayersNumber - 1)>;

static std_unique_ptr<BinContainer> mBins;
static std_unique_ptr<BinContainer> mBinsS;
static initBinContainers() { // implementation can go to cxx file
  if (!mBins) {
    mBins = std::make_unique<BinContainer>();
    initBins(mBins.get()); // init bins with static method. If not possible to do using static method, then make this method non-static and lock the code by mutex.
   //
  }
  // same for mBinsS
}
```
In the TrackerSpec call this `initBinContainers` *before* instantiating the trackers (e.g. in its init method (or in the beginning of updateTimeDependentParams) if this method can be static.
Otherwise, put the non-static method (protected by mutex) in the Tracker constructed. 
Same for destroying arrays via `mBins.reset()` : do this either in the Tracker destructor of the protecting the call with mutex or in the TrackerSpec destructor.

For the usage of the mutex see e.g. how the https://github.com/AliceO2Group/AliceO2/blob/816da2514cd86b7d1e4b2177a553b6312002c672/Detectors/Base/include/DetectorsBase/GeometryManager.h#L140 is used in https://github.com/AliceO2Group/AliceO2/blob/816da2514cd86b7d1e4b2177a553b6312002c672/Detectors/Base/src/GeometryManager.cxx#L37-L48 and below.

robincaron13:
Hi @shahor02, 

Thanks a lot for your answer and your proposal. 
Currently I'm making the changes but I have a question regarding the `initBins`. What should contain this function exactly? Because from my understanding an arrays is intialized at the moment we define its size but here it's now about `std::unique_ptr`, also I tried different way like for example using `std::unique_ptr<BinContainer>(new BinContainer)` and no success... Sorry I'm not an expert in the use of some functionalities in c++.

shahor02:
> I have a question regarding the `initBins`. What should contain this function exactly? 
Hi @robincaron13 
The initialization of the uniqe_ptr will just allocate the memory but not assign the values you need.
The initBins should contain what https://github.com/AliceO2Group/AliceO2/blob/50d85663b2976cec58da3651daf6ef93cc210f1c/Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx#L198-L214 does: assignment of the values.
I did not check if https://github.com/AliceO2Group/AliceO2/blob/50d85663b2976cec58da3651daf6ef93cc210f1c/Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx#L109 does anything else but setting `mBins(S)` values, if this function should be called only once before  the processing starts,  then this is actually the "initBins" I meant.

robincaron13:
Hi @shahor02,

Thanks for your answer. Some modifications are pushed according to your comments. 
For the moment I still kept a dummy static `initBins` in the `TrackerConfig.cxx` to do nothing, also basically the `initBinConntainer` is just doing `mBins = std::make_unique<BinContainer>();` and it should be called once in the init of the `TrackerSpec.cxx`. The `mBins.reset()` is called through the new destructor called in `TrackerSpec.cxx`. The mutex is also added to the `intializeFinder` funnction where the assignment of the `mBins` values are perfomed. 
Let me know if it goes in the good direction, or there are still modifications to do.

Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
```diff
@@ -67,25 +70,25 @@ class TrackerConfig
   bool mFullClusterScan = false;
   Float_t mTrueTrackMCThreshold; // Minimum fraction of correct MC labels to tag True tracks
 
+  static std::mutex sTCMutex;
+
   static Float_t mPhiBinSize;
   static Float_t mInversePhiBinSize;
+  static std::array<Float_t, constants::mft::LayersNumber> mInverseRBinSize;
```
shahor02: in principle, this one also can be defined as 
```
static std::unique_ptr<std::array<Float_t, constants::mft::LayersNumber>>  mInverseRBinSize;
```
and instantiated only once together with mBins.

Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
```diff
@@ -39,12 +40,14 @@ class TrackerConfig
...
-  // tracking configuration parameters
-  const auto& getBinsS() { return mBinsS; }
-  const auto& getBins() { return mBins; }
+  static void initBins(const BinContainer* mBinContainer);
+  static void initBinContainers();
 
   const std::pair<Int_t, Int_t>& getClusterBinIndexRange(Int_t layerId, Int_t bin) const { return mClusterBinIndexRange[layerId][bin]; }
 
+  static std::unique_ptr<BinContainer> mBins;
+  static std::unique_ptr<BinContainer> mBinsS;
```
shahor02: what is the reason to make them public?

Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
```diff
@@ -113,6 +115,8 @@ void Tracker<T>::initializeFinder()
     return;
   }
 
+  std::lock_guard<std::mutex> guard(o2::mft::TrackerConfig::sTCMutex);
```
shahor02: The lock will prevent executing the code below at the same time for different tracker copes (one will wait for other), but as it is now, each copy may still assign to mBins. To avoid this, you can check right after the guard if the the mBins is already created and proceed only if it is not. In this logic, you should move the `initContainers` here:

```
std::lock_guard<std::mutex> guard(o2::mft::TrackerConfig::sTCMutex);
if (mBins) {
  return;
}
initBinContainers();
...

```

Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
```diff
@@ -195,7 +199,7 @@ void Tracker<T>::initializeFinder()
               }
 
               binIndex2S = getBinIndex(binR, binPhiS);
-              mBinsS[layer1][layer2 - 1][binIndex1].emplace_back(binIndex2S);
+              mBinsS.get()[0][layer1][layer2 - 1][binIndex1].emplace_back(binIndex2S);
```
shahor02:
```
(*mBinsS.get())[layer1][layer2 - 1][binIndex1].emplace_back(binIndex2S);
```

Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
```diff
@@ -211,7 +215,7 @@ void Tracker<T>::initializeFinder()
               }
 
               binIndex2 = getBinIndex(binR, binPhi);
-              mBins[layer1][layer2 - 1][binIndex1].emplace_back(binIndex2);
+              mBins.get()[0][layer1][layer2 - 1][binIndex1].emplace_back(binIndex2);
```
shahor02:
```
(*mBins.get())[layer1][layer2 - 1][binIndex1].emplace_back(binIndex2);
```

Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
```diff
@@ -288,7 +292,7 @@ void Tracker<T>::findTracksLTF(ROframe<T>& event)
       clsInLayer1 = it1 - event.getClustersInLayer(layer1).begin();
 
       // loop over the bins in the search window
-      for (const auto& binS : getBinsS()[layer1][layer2 - 1][cluster1.indexTableBin]) {
+      for (const auto& binS : mBinsS.get()[0][layer1][layer2 - 1][cluster1.indexTableBin]) {
```
shahor02: same here and below, please change `mBinsS.get()[0]` to `(*mBinsS,get())`

Detectors/ITSMFT/MFT/tracking/src/TrackerConfig.cxx
```diff
@@ -42,4 +42,29 @@ void o2::mft::TrackerConfig::initialize(const MFTTrackingParam& trkParam)
...
+void o2::mft::TrackerConfig::initBins(const BinContainer* mBinContainer)
+{
+  // if it's a non-static method: mutex could be used here
+  // also to protect a BinContainer shared between multiple threads
+
+  // TODO: do nothing at this moment
+  // init the BinContainer
+  // the function initializeFinnder in tracker.cxx is doing this job
+}
+
```
shahor02: this is not needed anymore as all initializations are done in the `initializeFinder`

Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx
```diff
@@ -55,6 +56,8 @@ void TrackerDPL::init(InitContext& ic)
 
   // tracking configuration parameters
   auto& trackingParam = MFTTrackingParam::Instance(); // to avoid loading interpreter during the run
+
+  TrackerConfig::initBinContainers();
```
shahor02: this is moved to `initializeFinder`

Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx
```diff
@@ -427,5 +430,13 @@ DataProcessorSpec getTrackerSpec(bool useMC, int nThreads)
     Options{}};
 }
 
+///_______________________________________
+TrackerDPL::~TrackerDPL()
+{
+  // Deallocate the memory that was previously reserved for these arrays.
```
shahor02: also here the mutex should be used to avoid race condition:
```
std::lock_guard<std::mutex> guard(o2::mft::TrackerConfig::sTCMutex);
TrackerConfig::mBins.reset();
TrackerConfig::mBinsS.reset();
```

shahor02 (changes requested):
Thanks @robincaron13 
In general ok, but please see some comments below.

Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
robincaron13: Ok I will change it to a `std::unique_ptr` 
I imagine for the others arrays below `mPhiBinWin` and `mRBinSize` I can do the same right?

Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
robincaron13: No reason, just my mistake

Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
robincaron13: Ok thanks to clarify, I will change it accordingly to this

Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
robincaron13: Thanks I will fix it

Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
robincaron13: same

Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
robincaron13: yes

Detectors/ITSMFT/MFT/tracking/src/TrackerConfig.cxx
robincaron13: Ok I will remove it

Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx
robincaron13: Yes

Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx
robincaron13: Ok thanks! So I keep the destructor here in `TrackerSpec.cxx` even if I move the `initBinContainer` to `Tracker.cxx`?

Detectors/ITSMFT/MFT/tracking/include/MFTTracking/TrackerConfig.h
shahor02: They are a few bytes long, does not worth it

Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx
shahor02: Ah, no, sorry, overlooked that. This should be done in the destructor of the Tracker. The trackers will be destructed by the automatic destructor of the TrackerSpec.

robincaron13: Ok I will do it

robincaron13:
Hi @shahor02 
Let me know if the last changes added were fine with you. Thanks.

shahor02 (approved):
thanks, looks good

PR #10669, opened 2023-02-02
DPL: allow users to specify multiple timers

ktf:
The idea is that the user attaches as metadata to the timer inputs something like:

  InputSpec{... ,  Lifetime::Timer, timerIntervals({{<timer period>, <time duration>}, {<timer period>, <timer duration>}})}

For example:

  timerIntervals({{1000000, 10}, {10000000, -1}})

to have a timer expiring every 1'000'000 microseconds for the first 10 seconds and then every 10'000'000 microseconds for the rest of the time.

### DIFF
--- a/Framework/Core/src/LifetimeHelpers.cxx
+++ b/Framework/Core/src/LifetimeHelpers.cxx
@@ -98,12 +98,12 @@ ExpirationHandler::Creator LifetimeHelpers::enumDrivenCreation(size_t start, siz
   };
 }
 
-ExpirationHandler::Creator LifetimeHelpers::timeDrivenCreation(std::chrono::microseconds period, std::function<bool(void)> hasTimerFired, std::function<void(uint64_t, uint64_t)> updateTimerPeriod)
+ExpirationHandler::Creator LifetimeHelpers::timeDrivenCreation(std::vector<std::chrono::microseconds> periods, std::vector<std::chrono::seconds> intervals, std::function<bool(void)> hasTimerFired, std::function<void(uint64_t, uint64_t)> updateTimerPeriod)
 {
   std::shared_ptr<size_t> last = std::make_shared<size_t>(0);
   std::shared_ptr<bool> stablePeriods = std::make_shared<bool>(false);
   // FIXME: should create timeslices when period expires....
-  return [last, stablePeriods, period, hasTimerFired, updateTimerPeriod](ChannelIndex channelIndex, TimesliceIndex& index) -> TimesliceSlot {
+  return [last, stablePeriods, periods, intervals, hasTimerFired, updateTimerPeriod](ChannelIndex channelIndex, TimesliceIndex& index) mutable -> TimesliceSlot {
     // We start with a random offset to avoid all the devices
     // send their first message at the same time, bring down
     // the QC machine.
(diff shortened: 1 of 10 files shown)


ktf:
@Barthelemy @knopers8 here is what I had in mind for the configurability. This particular version does not yet work, but I should show how the multiple timers could be provided.

ktf:
@Barthelemy @knopers8 this seems to be what we need. I also added a test which works as expected. Please let me know how it works on your side, then we can merge.

Framework/Core/src/LifetimeHelpers.cxx
```diff
@@ -117,24 +117,34 @@ ExpirationHandler::Creator LifetimeHelpers::timeDrivenCreation(std::chrono::micr
...
+      LOG(debug) << "Timer updated to a stable period of " << periods.front().count() << "us";
     }
     // Nothing to do if the time has not expired yet.
     if (timerHasFired == false) {
       auto newOldest = index.setOldestPossibleInput({*last}, channelIndex);
       index.updateOldestPossibleOutput();
       return TimesliceSlot{TimesliceSlot::INVALID};
     }
+    // Get the first time we were invoked.
+    static auto firstTime = getCurrentTime();
```
knopers8, line 138: will that work with multithreading data processors in one process?

knopers8, line 138: also, since you are not updating `firstTime` upon updating the timer period, `intervals` are actually not intervals but they all refer to the time since `firstTime`. Which is fine with me, but perhaps the name could be clearer in such case.

knopers8 (commented):
Thanks, it seems to work with QC, including the randomization of the first cycle. However, it indeed does treat `intervals` as the time since the start.

Framework/Core/src/LifetimeHelpers.cxx
ktf: We want something which is the first time any timer gets executed, no? If yes, than this is the correct behaviour. If not, indeed we would need to keep track about that information at the input level.

ktf: Yes, indeed "intervals" is really "timeSinceStartOfProcessing". I will change that.

ktf:
Ok, I think I fixed the issue with the old timers and adapted a bit the naming convention. Let me know if you see anything else to be changed. We should still discuss how do you want this to be managed in the case of multithreading or in the case of multiple processed being multiplexed on the same device.

ktf:
@Barthelemy @knopers8 this is now ready to be merged. I will be away for part of next week. I am happy to merge this if it is early, but keep in mind that  before wed the only support I can give for it is "git revert".

Framework/Core/src/LifetimeHelpers.cxx
knopers8: Yes, indeed. Once `intervals` are actually `validities`, it makes sense to me, thanks.

knopers8: Thanks. I think you skipped `intervals` in this file, but `validity` is clearer indeed.

knopers8:
I tried again, it seems to work, thanks! I don't think it is urgent enough to merge it before you come back, but then it is fine with me.

PR #10670, opened 2023-02-02
[EMCAL-888] Add extra table to map multiple MC labels to single calocell

mhemmer-cern:
- Changed implementation from using an extra inbtween table to using and INDEX variable size array for the MC Particle reference and a std::vector<float> for the amplitude fraction that ech particle contributed to the total energy in the MC Calo Cell (changes suggested by @jgrosseo)

### DIFF
--- a/Framework/Core/include/Framework/AnalysisDataModel.h
+++ b/Framework/Core/include/Framework/AnalysisDataModel.h
@@ -1074,9 +1074,9 @@ DECLARE_SOA_COLUMN(McMask, mcMask, uint16_t);                   //! Bit mask to 
 DECLARE_SOA_COLUMN(AmplitudeA, amplitudeA, std::vector<float>); //! Energy fraction deposited by a particle inside this calo cell.
 } // namespace mccalolabel
 
-DECLARE_SOA_TABLE(McCaloLabels_000, "AOD", "MCCALOLABEL", //! Table joined to the calo table containing the MC index (version 000)
+DECLARE_SOA_TABLE(McCaloLabels_000, "AOD", "MCCALOLABEL", //! Table joined to the calo table containing the MC index (version 000, Run 2 format)
                   mccalolabel::McParticleId, mccalolabel::McMask);
-DECLARE_SOA_TABLE_VERSIONED(McCaloLabels_001, "AOD", "MCCALOLABEL", 1, //! Table joined to the calo table containing multiple MC indeces and the amplitude fraction (version 001)
+DECLARE_SOA_TABLE_VERSIONED(McCaloLabels_001, "AOD", "MCCALOLABEL", 1, //! Table joined to the calo table containing multiple MC indices and the amplitude fraction (version 001)
                             mccalolabel::McParticleIds, mccalolabel::McMask, mccalolabel::AmplitudeA);
 using McCaloLabels = McCaloLabels_000;
 using McCaloLabel = McCaloLabels::iterator;
--- a/Framework/Core/include/Framework/AnalysisDataModel.h
+++ b/Framework/Core/include/Framework/AnalysisDataModel.h
@@ -1068,13 +1068,17 @@ using McFwdTrackLabel = McFwdTrackLabels::iterator;
 
 namespace mccalolabel
 {
+DECLARE_SOA_INDEX_COLUMN(McParticle, mcParticle);               //! MC particle
 DECLARE_SOA_ARRAY_INDEX_COLUMN(McParticle, mcParticle);         //! Array of MC particles that deposited energy in this calo cell
 DECLARE_SOA_COLUMN(McMask, mcMask, uint16_t);                   //! Bit mask to indicate detector mismatches (bit ON means mismatch). Bit 15: indicates negative label
 DECLARE_SOA_COLUMN(AmplitudeA, amplitudeA, std::vector<float>); //! Energy fraction deposited by a particle inside this calo cell.
 } // namespace mccalolabel
 
-DECLARE_SOA_TABLE(McCaloLabels, "AOD", "MCCALOLABEL", //! Table joined to the calo table containing the MC index
-                  mccalolabel::McParticleIds, mccalolabel::McMask, mccalolabel::AmplitudeA);
+DECLARE_SOA_TABLE(McCaloLabels_000, "AOD", "MCCALOLABEL", //! Table joined to the calo table containing the MC index (version 000)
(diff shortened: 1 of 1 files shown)


mfasDa:
We need to change the MC label tables for the calorimeters. Reason is that several MC particles can contribute to a single tower. The most prominent examples are:

1) Shower merging, for mesons to two photons with small opening angles (high-pt)
2) High-occupancy, i.e. central PbPb

In particular for point 1 it is crucial to store the labels from contributing particles as they are needed for selection of true merged showers. A simple 1-1 correspondence cell to particle is not feasible for the calorimeters. The existing MC table has to become and index table with 1-1 correspondence to cells, while the actual MC labels need to be moved to a separate table referenced by the index table. This is supposed to mimic the behaviour of the MC truth container.

@peressounko Should be OK for PHOS as well.

sawenzel:
@jgrosseo @pzhristov: Is the Analysis WP ok with this change?

jgrosseo:
This is backward incompatible. Do we have produced MC with this information already?

Why not use a INDEX VLA for the MC labels? In this way the label table stays joinable instead of having the additional lookup table?

@pzhristov @ddobrigk I think this should be discussed on a WP14 meeting in either case.

jgrosseo:
Following the discussion at the WP14, you need to define two columns:

An VLA index, e.g. like this:
`DECLARE_SOA_ARRAY_INDEX_COLUMN(Collision, collisions);`

A VLA for float, e.g. like this:
`DECLARE_SOA_COLUMN(AmplitudeA, amplitudeA, std::vector<float>);`

Once that works, you can implement the versioning. You can take as example the v0s table:
```
DECLARE_SOA_TABLE(V0s_000, "AOD", "V0", //! Run 2 V0 table (version 000)
                  o2::soa::Index<>,
                  v0::PosTrackId, v0::NegTrackId);
DECLARE_SOA_TABLE_VERSIONED(V0s_001, "AOD", "V0", 1, //! Run 3 V0 table (version 001)
                            o2::soa::Index<>, v0::CollisionId,
                            v0::PosTrackId, v0::NegTrackId);

using V0s = V0s_001; //! this defines the current default version
```
Let me know if you have questions.

jgrosseo:
I make it draft for now. We need to sync merging with the O2Physics converter...

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -1068,12 +1068,17 @@ using McFwdTrackLabel = McFwdTrackLabels::iterator;
...
+DECLARE_SOA_INDEX_COLUMN(McParticle, mcParticle);               //! MC particle
+DECLARE_SOA_ARRAY_INDEX_COLUMN(McParticle, mcParticle);         //! Array of MC particles that deposited energy in this calo cell
+DECLARE_SOA_COLUMN(McMask, mcMask, uint16_t);                   //! Bit mask to indicate detector mismatches (bit ON means mismatch). Bit 15: indicates negative label
+DECLARE_SOA_COLUMN(AmplitudeA, amplitudeA, std::vector<float>); //! Energy fraction deposited by a particle inside this calo cell.
 } // namespace mccalolabel
 
-DECLARE_SOA_TABLE(McCaloLabels, "AOD", "MCCALOLABEL", //! Table joined to the calo table containing the MC index
+DECLARE_SOA_TABLE(McCaloLabels_000, "AOD", "MCCALOLABEL", //! Table joined to the calo table containing the MC index (version 000)
                   mccalolabel::McParticleId, mccalolabel::McMask);
+DECLARE_SOA_TABLE_VERSIONED(McCaloLabels_001, "AOD", "MCCALOLABEL", 1, //! Table joined to the calo table containing multiple MC indeces and the amplitude fraction (version 001)
```
```suggestion
DECLARE_SOA_TABLE_VERSIONED(McCaloLabels_001, "AOD", "MCCALOLABEL", 1, //! Table joined to the calo table containing multiple MC indices and the amplitude fraction (version 001)
```

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -1068,12 +1068,17 @@ using McFwdTrackLabel = McFwdTrackLabels::iterator;
...
-DECLARE_SOA_INDEX_COLUMN(McParticle, mcParticle); //! MC particle
-DECLARE_SOA_COLUMN(McMask, mcMask, uint16_t);     //! Bit mask to indicate detector mismatches (bit ON means mismatch). Bit 15: indicates negative label
+DECLARE_SOA_INDEX_COLUMN(McParticle, mcParticle);               //! MC particle
+DECLARE_SOA_ARRAY_INDEX_COLUMN(McParticle, mcParticle);         //! Array of MC particles that deposited energy in this calo cell
+DECLARE_SOA_COLUMN(McMask, mcMask, uint16_t);                   //! Bit mask to indicate detector mismatches (bit ON means mismatch). Bit 15: indicates negative label
+DECLARE_SOA_COLUMN(AmplitudeA, amplitudeA, std::vector<float>); //! Energy fraction deposited by a particle inside this calo cell.
 } // namespace mccalolabel
 
-DECLARE_SOA_TABLE(McCaloLabels, "AOD", "MCCALOLABEL", //! Table joined to the calo table containing the MC index
+DECLARE_SOA_TABLE(McCaloLabels_000, "AOD", "MCCALOLABEL", //! Table joined to the calo table containing the MC index (version 000)
```
```suggestion
DECLARE_SOA_TABLE(McCaloLabels_000, "AOD", "MCCALOLABEL", //! Table joined to the calo table containing the MC index (version 000, Run 2 format)
```

jgrosseo (commented):
This looks good to me. I just made 2 cosmetic suggestions which you can accept in github

jgrosseo:
If the others agree (@pzhristov @ddobrigk), we can already merge the table. Then you could write the converter (in O2Physics, you can have a look at o2-collision-converter to get an idea) and prepare the code changes for producing in version 1 in the AOD producer. Here as agreed first see how large the data will be.

jgrosseo (approved):

ddobrigk:
Sorry for the late reaction, but just to confirm this is fine with me as well! Thanks! 

I see for now we proceed using `000` as the base version, right? I guess one can do the converter in `O2Physics` once this is merged and then move to `001`.

mhemmer-cern:
Yes, so from my understanding the plan was to merge this, then implement a converter task similar to the collision convert in `O2Physics` where we would convert the `000` to `001` and later switch to the 001 as standard. @ddobrigk

PR #10672, opened 2023-02-02
TRD: allow compilation w/o ONNXRuntime

aphecetche:
This is a proposal to allow compilation of the TRD libs even with the ONNXRunTime library is not available. 

Another option (simpler maybe but less convenient IMHO) is to make it actually required in the `find_package(ONNXRuntime CONFIG)` statement in the top level O2 `CMakeLists.txt`, where currently it is not.

Full disclosure : I've not tested that the built library actually makes complete sense w/o ONNXRuntime, but at least it does not prevent O2 to compile and others libs to used (tested only the MCH reco though...)

### DIFF
--- a/Detectors/TRD/pid/CMakeLists.txt
+++ b/Detectors/TRD/pid/CMakeLists.txt
@@ -11,23 +11,33 @@
 
 o2_add_library(TRDPID
                SOURCES src/PIDBase.cxx
-                       src/ML.cxx
                        src/PIDParameters.cxx
                PRIVATE_LINK_LIBRARIES O2::TRDBase
-                                     O2::DataFormatsTRD
-                                     O2::DataFormatsGlobalTracking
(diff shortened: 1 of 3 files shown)


martenole:
@f3sch could you check if these changes are OK for you?

f3sch (approved):
They seem fine to me. ONNX is only really needed if ML models would be used for PID.
If @aphecetche wants to decouple this dependency, thats fine.
I cannot comment on the CMakeList changes, due to unfimiliarity.
Did you @aphecetche built with and without, e.g. testing both branches? If so, go merge.

aphecetche:
@f3sch I did test on my Mac w/o ONNXRuntime (and w/o aliBuild) and (I assume that) the various CIs are with ONNXRuntime, so I guess both cases compile, yes.

martenole:
Thanks, merged

PR #10674, opened 2023-02-02
CODEOWNERS: Fix ITS3 directory

fgrosa:


ktf:
@fgrosa I have also added both you an @arossi81 as writers to the repository.

fgrosa:
> @fgrosa I have also added both you an @arossi81 as writers to the repository.

Thanks!

PR #10682, opened 2023-02-03
Fix race condition when creating a folder on mac

ktf:


ktf:
The first time always fails (while still creating the folder). Not sure if it's a defect, or it's actually posix compliance (need to check the wording of trailing slash...).

ktf:
Ok, it's apparently some issue with libc++ / clang / MSVS. More information at:

https://stackoverflow.com/questions/60130796/return-value-of-stdfilesystemcreate-directories-on-paths-with-trailing-sla

I would say my fix should be good enough.

PR #10684, opened 2023-02-03
Adding protections needed for QC

chiarazampolli:
@ktf , @shahor02 , @knopers8 
This fixes the issue when the QC uses the GeomHelper, but the timer triggers QC before any CCDB was fetched.


knopers8 (approved):
thank you

PR #10685, opened 2023-02-03
STDERR monitor: filter TInterpreter warnings

martenole:
Hoping to avoid flooding of the infologger with these messages:
Screenshot from 2023-02-03 13-56-20


davidrohr (approved):
thx, makes sense.
We should also have the PR with the proper fix to get rid of these warnings today, but good to have this anyway for the future

martenole:
Error unrelated, merging

PR #10686, opened 2023-02-03
fix command line option for o2-mch-digits-writer-workflow

pillot:


prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -410,7 +410,7 @@ has_detector_reco EMC && ( [[ -z "$DISABLE_ROOT_OUTPUT" ]] || needs_root_output
 has_detector_reco PHS && ( [[ -z "$DISABLE_ROOT_OUTPUT" ]] || needs_root_output o2-phos-cell-writer-workflow ) && add_W o2-phos-cell-writer-workflow "$DISABLE_MC"
 has_detector_reco FV0 && ( [[ -z "$DISABLE_ROOT_OUTPUT" ]] || needs_root_output o2-fv0-digits-writer-workflow ) && add_W o2-fv0-digits-writer-workflow "$DISABLE_MC"
 has_detector_reco MID && ( [[ -z "$DISABLE_ROOT_OUTPUT" ]] || needs_root_output o2-mid-decoded-digits-writer-workflow ) && add_W o2-mid-decoded-digits-writer-workflow "--mid-digits-tree-name o2sim" "" 0
-has_detector_reco MCH && ( [[ -z "$DISABLE_ROOT_OUTPUT" ]] || needs_root_output o2-mch-digits-writer-workflow ) && add_W o2-mch-digits-writer-workflow "--outfile mchdigits.root" "" 0
```
davidrohr, line 413: Are you sure we can remove the --outfile? It was writing to a different file before I added that.

pillot: yes, the --outfile option is only used when the output is a binary or an ascii file. When enabling the writing in a root file, the file name is fixed to mchdigits.root.

(I just fixed the trailing whitespace error)

PR #10689, opened 2023-02-03
Remove DCAFitter from DetectorsVertexing

ddobrigk:
Final step in cleaning up libraries: removes DCAFitter entirely from DetectorsVertexing. Should suppress ROOT warning of duplication.


shahor02:
@ddobrigk thanks, could you please suppress the tabs?

ddobrigk:
> @ddobrigk thanks, could you please suppress the tabs?

Yes, of course, done! Sorry for that. Was editing via a terminal on a remote server quickly and must have messed up, now all good I think.

PR #10690, opened 2023-02-03
ITS3: Base: Make ALPIDE pixel size configurable

f3sch:
In preparation for evaluating digits and different pixel sizes.


mconcas (dismissed):

fgrosa (changes requested):
@f3sch the code looks good, thanks a lot! I would just ask you to rename "Alpide" to "SuperAlpide" to be cosistent with the `SegmentationSuperAlpide` (this will not be the final name of the ITS3 chip, but still).

fgrosa (approved):

PR #10691, opened 2023-02-03
Increase TRD dead time from 0.2 to 11 us

martenole:
Hi @tdietel and @pachmay 
the currently implemented hard-coded busy time of the TRD is 3.2 micro seconds. This means that in MC simulations the TRD is read out with more than 300kHz which is not what we have in reality. In 2022 the readout rate was something between 60 and 70 kHz max. With this change it would be limited to ~70kHz also for simulations.

This should maybe be changed to be a configurable parameter in the future, not sure if we can still improve the readout rate by masking some slow chambers for example.


martenole:
Also pinging to @shahor02

tdietel:
I remember that Jorge set the dead time unrealistically low on purpose. This allow to produce TRD data for a larger fraction of the events and therefore reduce simulation times. However, the downside is of course that the trigger efficiency is not modeled correctly.
It would be good to have this value configurable.

pachmay:
Hi @martenole and @tdietel,

for data in Run 3, we have to increase the readout of TRD. Masking is one option, the other is to revisit the CTP settings for TRD. I agree having this configurable for MC is good.

PR #10694, opened 2023-02-03
More equivalency declarations

ddobrigk:
This is a suggested change following a conversation with @aalkin (thanks a lot!)

These equivalency declarations have an important role in avoiding that track pre-selections and MC pre-selections depend unnecessarily on the track propagation workflow (or on the data being from Run 2 or Run 3). By declaring `TracksExtra` and `McTrackLabels` to be equivalent to the base `Tracks`/`TracksIU` tables for indexing purposes, one can de-reference an index to the base track tables straight to the `TracksExtra` or `McTrackLabels` tables without going through the `Tracks` or `TracksIU` tables and populate other tables that can still be joined with the base `Tracks`/`TracksIU` tables later on and filtered out as necessary. 

As an example use case, this will greatly simplify the V0 and cascade candidate pre-selectors, effectively keeping the number of `process` functions down to only four (instead of 8 or more to account for various combinations of `Tracks`/`TracksIU`). This has already been tested to work very well locally.


ddobrigk:
Hm, these errors seem totally unrelated to the change done in the PR... Not sure if they are worrisome or not. Maybe (when convenient, not necessarily during the weekend!) @ktf @TimoWilken could you comment? Thanks a lot!!

ddobrigk:
Not sure if someone did something, but I guess so: the errors disappeared :-) - thanks a lot!!

jgrosseo (approved):

PR #10695, opened 2023-02-03
DPL: make adding distsubtimeframe & co stable

ktf:
Previous version could have failed in case of multiple source devices as in the case of multiple readers.

This PR aims at making sure the device picked to attach the DISTSUBTIMEFRAME is always the same.


ktf:
@chiarazampolli @shahor02 @davidrohr this should be the solution to the problem we discussed today.

chiarazampolli:
Thanks, I will try it.

davidrohr:
works for me, thx!

PR #10705, opened 2023-02-06
Introduce DataAllocator::cookDeadBeef method

ktf:


noferini:
Hi @ktf ,
this is the only change you need or do you plan to other changes?
I could run some tests locally once you think it is ready

ktf:
No, there is one more thing which actually does not need any change at the level of the framework. I will do the following:

* Make a PR which introduces the DataAllocator::cookDeadBeef(), without the TOF changes.
* Provide a separate PR, where a new executable is created which uses cookDeadBeef and which creates the payload via DataAllocator::makeVector, which should basically cut the code in half and make it compatibile with START-STOP-START.

noferini:
Ok, let me know when I can run some tests

ktf:
Do you have any test I can run myself? It would probably simplify
iterations.


Ciao,
Giulio

ktf:
Formatting error actually spurious, merging.

PR #10709, opened 2023-02-06
TRD MC->raw uses default RDH unless overridden

shahor02:


martenole:
Ah, I missed that hard coded option.
Thanks!

martenole (approved):

PR #10712, opened 2023-02-06
Better memory handling for sparse pp

mpuccio:
Dear @shahor02

This should fix the issue with the timing you found on the grid.
We were over-allocating for the LUTs﻿.
In the sparse dataset you pointed out, this brings factor 300 speedup to the vertexer and factor 8 for the tracker.
It needs to be checked for a pp high rate sample, as there is a trade-off with higher occupancies, at least for the tracker with these new settings.
I open the PR for the checks anyway.

Cheers,
Max


shahor02:
Thanks @mpuccio,   @mconcas, shall I merge this or you are still testing?

mpuccio:
Tested, working and faster even in high IR

shahor02 (approved):

PR #10717, opened 2023-02-07
Adapt TOF compressor to use DPL

ktf:


ktf:
@noferini this should be all it's needed to make the TOF compressor be DPL compliant. Could you have a look and let me know how it works with your tests?

noferini:
> @noferini this should be all it's needed to make the TOF compressor be DPL compliant. Could you have a look and let me know how it works with your tests?

I'm going to check. Thanks

ktf:
(Notice also the --old option for the old behaviour).

noferini:
I recompiled with your changes.
I got immediately this error

[O2/latest] ~/alice/O2DPG $> o2-tof-compressor --help
[FATAL] error while setting up workflow in o2-tof-compressor: missing option: old (No such node (old))
For later analysis we write a core dump to core_dump_721752

noferini:
I guess this line is missing
  workflowOptions.push_back(old);

at line 47 of tof-compressor.cxx

noferini:
The line I reported in the previous message fixes the error.
I checked the new task and it doesn't work properly /with --old it works).
I'm trying to understand where is the problem (it seems to me that the message from compressor is not received/digested by the next dpl).
Cheers,
Francesco

ktf:
Sorry, I forgot to squash one commit. Updated.

ktf:
Tested to work together with @noferini. Old version remains available via the `--old` switch.

PR #10719, opened 2023-02-07
Adding script to load functions, and possibility to add more wf

chiarazampolli:
@noferini , @catalinristea @davidrohr , @martenole , @shahor02 

Goes with https://github.com/AliceO2Group/O2DPG/pull/855
- allows adding arbitrary number of extra WFs
- splits the definition of general functions from setenv.sh

### DIFF
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -3,6 +3,7 @@
 # ---------------------------------------------------------------------------------------------------------------------
 # Get this script's directory and load common settings (use zsh first (e.g. on Mac) and fallback on `readlink -f` if zsh is not there)
 MYDIR="$(dirname $(realpath $0))"
+source $MYDIR/gen_topo_helper_functions.sh
 source $MYDIR/setenv.sh
 
 if [[ $EPNSYNCMODE == 0 && $DPL_CONDITION_BACKEND != "http://o2-ccdb.internal" && $DPL_CONDITION_BACKEND != "http://localhost:8084" && $DPL_CONDITION_BACKEND != "http://127.0.0.1:8084" ]]; then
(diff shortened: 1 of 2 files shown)


prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -492,7 +493,12 @@ if [[ ! -z "$EXTRA_WORKFLOW" ]]; then
 fi
 
 if [[ ! -z "$ADD_EXTRA_WORKFLOW" ]]; then
-  add_W $ADD_EXTRA_WORKFLOW
+  OLD_IFS=$IFS
```
davidrohr, line 496: I think to be on the safe side, one should use OLD_IFS="$IFS", same below

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -492,7 +493,12 @@ if [[ ! -z "$EXTRA_WORKFLOW" ]]; then
...
+  IFS=','
+  for wf in $ADD_EXTRA_WORKFLOW; do
+    add_W $wf
```
davidrohr: I'd add `[[ ! -z "$wf" ]] && add_W $wf`.
your loop will not strip commas at the beginning and end of the string, i.e. something like
`ADD_EXTRA_WORKFLOW=",o2-tpc-reco-workflow"` would try to add an empty workflow.

davidrohr (changes requested):
fine for me, just some comments below.
I'd merge the O2DPG one first, and then this one.

prodtests/full-system-test/dpl-workflow.sh
chiarazampolli: Ok, it was working also before, but ok with me. The loop already removes the "," since it is used as a delimiter:
```
[O2PDPSuite/latest] ~/work/O2/LHC22m_apass2/20230119 $> ADD_EXTRA_WORKFLOW=ppp,qqq
[O2PDPSuite/latest] ~/work/O2/LHC22m_apass2/20230119 $> OLD_IFS=$IFS
[O2PDPSuite/latest] ~/work/O2/LHC22m_apass2/20230119 $>   IFS=','
[O2PDPSuite/latest] ~/work/O2/LHC22m_apass2/20230119 $>   for wf in $ADD_EXTRA_WORKFLOW; do
> echo $wf
> done
ppp
qqq
```
Or am I missing your doubt?

davidrohr: It will remove the `,`, but if it starts with the comma, I assume it might use the empty part before the comma as an empty string.

chiarazampolli:
> fine for me, just some comments below. I'd merge the O2DPG one first, and then this one.

Hi @davidrohr , 
Thanks! I am waiting for the CI to pass in O2DPG, but I am not sure what it runs, maybe it will even fail since setenv.sh changed?

PR #10721, opened 2023-02-07
Retrieve number of orbits in TF from GRP

dstocco:
Copes with a future possible modification of the TF length


dstocco:
@shahor02 @chiarazampolli : I tested that this code works in async mode.
However, this workflows is meant to run in sync mode, during a calibration run. How do I retrieve this information in a synchronous run?

martenole:
Hi Diego,
there is nothing more to be done :)
You already have the GRPGeomRequest in your workflow and the GRPECS object is requested. With `o2::base::GRPGeomHelper::instance().getGRPECS()->getNHBFPerTF()` you will then also in synchronous mode get the correct number of HBFs per TF, since that information is provided from the DPL CCDB mechanism together with the first data input to your run() method.
Cheers,
Ole

dstocco:
Hi @martenole , well that's a very good news! Thanks for the info!

aphecetche (approved):

PR #10733, opened 2023-02-08
Fix in stopping track accumulation when stat is ok

shahor02:
@martenole


martenole:
Thanks @shahor02 but actually one should not break here in case additional ITS-TPC tracks are requested. The segfault occurred, because with the default settings ITS-TPC tracks are not requested. And in any case there must be a protection in case no ITS-TPC tracks are available for a given TF.
But if there are, then I don't stop even if a sufficient amount of statistics is already reached. Instead, I still process a certain number of ITS-TPC only tracks to get a more uniform phi coverage

PR #10735, opened 2023-02-08
Add T0 offset correction for average distortion maps

martenole:


martenole:
Tested locally (ctests, fst and map creation using the macro for different runs)
Will merge soon to have it in the next nightly

PR #10736, opened 2023-02-08
Install average map creation macro

martenole:


martenole:
Not yet tested, need to rebuild O2 almost completely for this change... The map creation on the grid should use this macro therefore its needed to install it @chiarazampolli

chiarazampolli:
Hello @martenole , 
This is the script that we use on the grid:
https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/configurations/asyncCalib/createCorrectionMap.sh
Can you modify it accordingly? Maybe with some flags to switch between the approaches?

PR #10738, opened 2023-02-08
DPL: specify balancing in the CompletionPolicy

ktf:
Some completion policies, e.g. Discard or Consume, do not need to have the input channels balanced because they do not starve the queues in any case. For these cases, we allow disabling at the policy level the balancing.

By default the behaviour is unchanged.

As part of this commit I add a test which reproduces the behavior of the SACs / IDCs race condition, which gets corrected by the fact the Consume policy is now set to have balanceChannels = false.

### DIFF
--- a/Framework/Core/CMakeLists.txt
+++ b/Framework/Core/CMakeLists.txt
@@ -333,6 +333,7 @@ foreach(w
         ParallelPipeline
         ParallelProducer
         SlowConsumer
+        SlowProducerWithWildCard
         SimpleDataProcessingDevice01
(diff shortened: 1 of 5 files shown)


ktf:
@davidrohr @chiarazampolli this development fixes the behaviour of my SACs/IDCs lookalike example. I have not done any real test beyond the Framework unit tests, but I suspect it might correct other issues, including the actual SACs/IDCs problem.

ktf:
Tests were failing on linux due to the fact I did not notice a case inconsistency. Retesting but it otherwise seems good.

PR #10739, opened 2023-02-09
TPC: Adding streamer for FastTransform in reconstruction

matthias-kleiner:
SpaceCharge: adding option to use analytical formula for corrections and distortions
 - adding simple default analytical distortions/corrections
 - add streamer in digitizer

DebugStreamer: Allow  writing out streams in parallel

### DIFF
--- a/Common/Utils/CMakeLists.txt
+++ b/Common/Utils/CMakeLists.txt
@@ -25,7 +25,7 @@ o2_add_library(CommonUtils
                        src/IRFrameSelector.cxx
                        src/DebugStreamer.cxx
                PUBLIC_LINK_LIBRARIES ROOT::Hist ROOT::Tree Boost::iostreams O2::CommonDataFormat O2::Headers
-                                     FairLogger::FairLogger O2::MathUtils)
+                                     FairLogger::FairLogger O2::MathUtils TBB::tbb)
 
 o2_target_root_dictionary(CommonUtils
                           HEADERS include/CommonUtils/TreeStream.h
--- a/Common/Utils/include/CommonUtils/DebugStreamer.h
+++ b/Common/Utils/include/CommonUtils/DebugStreamer.h
@@ -21,6 +21,7 @@
 #include "CommonUtils/ConfigurableParamHelper.h"
 #if defined(DEBUG_STREAMER)
 #include "CommonUtils/TreeStreamRedirector.h"
+#include <tbb/concurrent_unordered_map.h>
 #endif
 #endif
 
@@ -29,10 +30,12 @@ namespace o2::utils
 
 /// struct defining the flags which can be used to check if a certain debug streamer is used
 enum StreamFlags {
-  streamdEdx = 1 << 0,         ///< stream corrections and cluster properties used for the dE/dx
-  streamDigitFolding = 1 << 1, ///< stream ion tail and saturatio information
-  streamDigits = 1 << 2,       ///< stream digit information
-  streamITCorr = 1 << 4,       ///< stream ion tail correction information
+  streamdEdx = 1 << 0,          ///< stream corrections and cluster properties used for the dE/dx
+  streamDigitFolding = 1 << 1,  ///< stream ion tail and saturatio information
+  streamDigits = 1 << 2,        ///< stream digit information
+  streamFastTransform = 1 << 3, ///< stream tpc fast transform
+  streamITCorr = 1 << 4,        ///< stream ion tail correction information
+  streamDistortionsSC = 1 << 5, ///< stream distortions applied in the TPC space-charge class (used for example in the tpc digitizer)
 };
 
 #if !defined(GPUCA_GPUCODE) && !defined(GPUCA_STANDALONE)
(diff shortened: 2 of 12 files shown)


davidrohr:
What is sizeof(o2::utils::DebugStreamer)?
Is it guaranteed to be the same on CPU and GPU?
Otherwise it will break GPU reconstruction.

matthias-kleiner:
> What is sizeof(o2::utils::DebugStreamer)? Is it guaranteed to be the same on CPU and GPU? Otherwise it will break GPU reconstruction.

The class is only defined when DEBUG_STREAMER was set during compiling o2. If it was not specified then the class should be always empty and the same size for CPU and GPU.
If o2 was compiled with DEBUG_STREAMER and one runs on GPU, then I guess it would be of different size, but I will check what happens.

davidrohr:
Ok, thx. Then it should be fine.

However, I propose we add a check at the initialization of the GPUReconstruction class when the DEBUG STREAMER define is set, and if so we throw a fatal if the backend is not CPU, to avoid confusing crashes if accidentally they are used together.

wiechula (changes requested):
@matthias-kleiner , thanks for the implementation!
Concerning the streamers, could you please also add then here:
https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/TPCFastTransformation/TPCFastTransform.h#L657
https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/TPCFastTransformation/TPCFastTransform.h#L668
These functions are also called in reconstruction. @davidrohr  are there others needed?

matthias-kleiner:
> Ok, thx. Then it should be fine. However, I propose we add a check at the initialization of the GPUReconstruction class when the DEBUG STREAMER define is set, and if so we throw a fatal if the backend is not CPU, to avoid confusing crashes if accidentally they are used together.

This would be good. Can you implement this?

shahor02 (commented):
@matthias-kleiner did you check if the streaming to trees (even different ones) works in multithreaded code? The root I/O is in general not thread safe, see e.g. https://root-forum.cern.ch/t/root-i-o-and-thread-safety/5857

davidrohr:
Sure, will take care

matthias-kleiner:
> @matthias-kleiner did you check if the streaming to trees (even different ones) works in multithreaded code? The root I/O is in general not thread safe, see e.g. https://root-forum.cern.ch/t/root-i-o-and-thread-safety/5857

Yes I checked this and it is working fine for me. If each thread gets its unique TFile then it is working without problems.

matthias-kleiner:
**Just some information on how to use it, in case someone wants to use it**
1. Create the space-charge object which can be used in the digitizer and the fast transform which can be used in the reconstruction (default analytical formulas are used, in case of custom formulas they should be set in the `createTPCSpaceChargeCorrectionAnalytical()` function)
```
.L O2/Detectors/TPC/reconstruction/macro/createTPCSpaceChargeCorrection.C+
createTPCSpaceChargeCorrectionAnalytical("tpctransform_ana.root");
```

Output of debug spline object:
```
r fastTransformUnitTest_debug1_gridsize180-129-129.root
fastTransform -> Draw("((lx*50-254.5)/49-lx) : lx","","colz");
```
[corrections local X](https://user-images.githubusercontent.com/48915672/217817735-9c560bdb-3919-43b9-bb85-357f7143e13e.png)
[corrections local Y](https://user-images.githubusercontent.com/48915672/217817955-b28d5b21-31aa-4ea2-91d8-76ad4a2d0bcf.png)
[corrections local Z](https://user-images.githubusercontent.com/48915672/217818022-c3bc34f6-29fc-4d24-9124-fd4d6d45031f.png)
[formula - corrections local X](https://user-images.githubusercontent.com/48915672/217818385-adcb0489-d071-48ef-806e-0e9832d43e08.png)

2. Running the digitizer with the analytical distortions and create debug output
```
o2-sim-digitizer-workflow --onlyDet TPC --distortionType 2 --readSpaceCharge "distortions_analytical.root" --configKeyValues "DebugStreamerParam.StreamLevel=32;" --tpc-lanes 4 -b
```
3. Running the reconstruction
```
mkdir -p TPC/Calib/CorrectionMap
cp tpctransform_ana.root TPC/Calib/CorrectionMap/snapshot.root
o2-tpc-reco-workflow --output-type "tracks" --infile "tpcdigits.root" --condition-remap "file:///data/mkleiner/tmp/anasc/sim/=TPC/Calib/CorrectionMap" --configKeyValues "DebugStreamerParam.StreamLevel=8;" -b
```

4. In case debug output was created merge the TTrees
```
hadd debug_fasttransform.root  debug_fasttransform*
root.exe
o2::utils::DebugStreamer::mergeTrees("debug_fasttransform.root", "debug_fasttransform_merged.root");
```
5. Comparing output corrections with expectation:
```
root.exe debug_fasttransform_merged.root
```
[corrections local X - formula:`tree_0 -> Draw("dx - ((lx*50-254.5)/49-lx)  : lx","","");`](https://user-images.githubusercontent.com/48915672/217819650-ab697b57-d71c-4b17-acb5-c8082f2e2fb9.png)
[corrections local Y - formula: `tree_0 -> Draw("(lyT -ly) - (-(2 + 0.01 * (lx + (lx * 50 - 254.5) / 49 - lx)))  : lx","","*")`](https://user-images.githubusercontent.com/48915672/217819787-ed005883-f832-42a3-9dbd-f669ccb998a0.png)
[corrections local Z - formula: `tree_0 -> Draw("(lzT - lz) - (lz * 50 / 51 - lz)  : lz","","*")`](https://user-images.githubusercontent.com/48915672/217819948-f1d16f46-cfb3-4661-aa89-b728c4d70873.png)

matthias-kleiner:
> @matthias-kleiner , thanks for the implementation! Concerning the streamers, could you please also add then here: https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/TPCFastTransformation/TPCFastTransform.h#L657 https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/TPCFastTransformation/TPCFastTransform.h#L668 These functions are also called in reconstruction. @davidrohr are there others needed?

Thanks, I just added them

wiechula:
> > @matthias-kleiner , thanks for the implementation! Concerning the streamers, could you please also add then here: https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/TPCFastTransformation/TPCFastTransform.h#L657 https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/TPCFastTransformation/TPCFastTransform.h#L668 These functions are also called in reconstruction. @davidrohr are there others needed?
> 
> Thanks, I just added them

My idea was to add streamers directly in the functions above, as in the `Transform` to see the values which are called.

matthias-kleiner:
> > > @matthias-kleiner , thanks for the implementation! Concerning the streamers, could you please also add then here: https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/TPCFastTransformation/TPCFastTransform.h#L657 https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/TPCFastTransformation/TPCFastTransform.h#L668 These functions are also called in reconstruction. @davidrohr are there others needed?
> > 
> > 
> > Thanks, I just added them
> 
> My idea was to add streamers directly in the functions above, as in the `Transform` to see the values which are called.

Now I understand, I added this

PR #10740, opened 2023-02-09
Fix metadata messages in AOD

nburmaso:


chiarazampolli:
Hello @nburmaso , 
I could not test this PR, but I propose to merge it. The error:
```
475/475 Test #432: o2sim_hepmc ..........................................................................***Timeout 400.02 sec
```
seems unrelated. @shahor02 , @davidrohr , what do you think? We would benefit from having metadata in AO2Ds in apass3. And we will have the RelVal to verify that it works fine.

PR #10743, opened 2023-02-09
Reset counter at new file

jgrosseo:


jgrosseo:
Thanks @costing for noticing.

PR #10747, opened 2023-02-10
Fix bin initialization for MFT tracker

martenole:
@shahor02 
in our tests `mFullClusterScan` is set to false, but at P2 its set to true. That's why the CI did not catch it.

### DIFF
--- a/Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
+++ b/Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
@@ -110,11 +110,6 @@ void Tracker<T>::configure(const MFTTrackingParam& trkParam, bool firstTracker)
 template <typename T>
 void Tracker<T>::initializeFinder()
 {
-
-  if (mFullClusterScan) {
-    return;
-  }
-
   // The lock will prevent executing the code below at the same time for different tracker copes (one will wait for other)
   std::lock_guard<std::mutex> guard(TrackerConfig::sTCMutex);
   if (mBins) {


Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx
```diff
@@ -110,18 +110,17 @@ void Tracker<T>::configure(const MFTTrackingParam& trkParam, bool firstTracker)
...
   // The lock will prevent executing the code below at the same time for different tracker copes (one will wait for other)
   std::lock_guard<std::mutex> guard(TrackerConfig::sTCMutex);
   if (mBins) {
     return;
   }
   TrackerConfig::initBinContainers();
 
+  if (mFullClusterScan) {
+    return;
+  }
```
martenole: Although this would probably still crash. I don't understand this if statement. It seems the lookup tables must still be created properly, since they are used also if `mFullClusterScan==true`

shahor02: @martenole I am preparing a proper fix. Part of the static arrays pointers (initialized dynamically for the 1st tracker instance) is needed regardless of mFullClusterScan value but was initialized only if `mFullClusterScan==false`. Conversely, with your fix large arrays will be created even when not needed.

martenole: Thanks, I'll close this then. But then at least we found a real bug which was also there before, but was hidden by the fact that the arrays were in any case initialized with zero, right? As far as I can tell this only affected cosmic running though

shahor02:
Yes, the bug existed since https://github.com/AliceO2Group/AliceO2/pull/10654

PR #10748, opened 2023-02-10
MFTTracker: fix initialization of precalculated arrays

shahor02:
Hi @robincaron13 
I've made some of small static arrays non-pointers and decoupled their initialization from the [check](https://github.com/shahor02/AliceO2/blob/4b02a90d8b588415cc5149282e74a726812a0682/Detectors/ITSMFT/MFT/tracking/src/Tracker.cxx#L143) since they are always needed.
But testing the reconstruction with `--configKeyValues "MFTTracking.FullClusterScan=true"` in lhc22m (500 kHz) I find it tremendously slow. With the timing benchmarking I added (need [this](https://github.com/shahor02/AliceO2/blob/4b02a90d8b588415cc5149282e74a726812a0682/Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx#L47 ) uncommented) I get:
```
 o2-ctf-reader-workflow --loop 2 --ctf-input ctf/o2_ctf_run00523308_orbit0363341084_tf0000345682_epn015.root --onlyDet MFT --shm-segment-size 16000000000 | o2-mft-reco-workflow   --shm-segment-size 16000000000 --clusters-from-upstream --disable-mc --nThreads 1 --pipeline mft-tracker:1  --configKeyValues "MFTTracking.FullClusterScan=true"
```
[2044635:mft-tracker]: [15:20:58][INFO] launchTrackFinder| tracker:0 did 1-th ROF in 620802 mus: 338 clusters -> 7 tracks
[2044635:mft-tracker]: [15:20:59][INFO] launchTrackFinder| tracker:0 did 2-th ROF in 326679 mus: 276 clusters -> 4 tracks
...
[2044635:mft-tracker]: [15:21:15][INFO] launchTrackFinder| tracker:0 did 54-th ROF in 224566 mus: 226 clusters -> 0 tracks
[2044635:mft-tracker]: [15:21:15][INFO] launchTrackFinder| tracker:0 did 55-th ROF in 283832 mus: 247 clusters -> 2 tracks
[2044635:mft-tracker]: [15:21:16][INFO] launchTrackFinder| tracker:0 did 56-th ROF in 1040722 mus: 395 clusters -> 5 tracks
[2044635:mft-tracker]: [15:21:20][INFO] launchTrackFinder| tracker:0 did 57-th ROF in 3220702 mus: 579 clusters -> 6 tracks
...
```
More than 1 s per ROF with a few 100 clusters is too much. At the moment, the FullClusterScan is used for cosmic only, but even in this case some noisy event may make the workflow to stuck. Is it possible to optimize the combinatorics?

### DIFF
--- a/Detectors/ITSMFT/MFT/tracking/include/MFTTracking/Tracker.h
+++ b/Detectors/ITSMFT/MFT/tracking/include/MFTTracking/Tracker.h
@@ -77,8 +77,9 @@ class Tracker : public TrackerConfig
   bool fitTracks(ROframe<T>&);
   void computeTracksMClabels(const std::vector<T>&);
 
-  void configure(const MFTTrackingParam& trkParam, bool firstTracker);
+  void configure(const MFTTrackingParam& trkParam, int trackerID);
   void initializeFinder();
+  int getTrackerID() const { return mTrackerID; }
 
  private:
   void findTracksLTF(ROframe<T>&);
@@ -143,6 +144,7 @@ class Tracker : public TrackerConfig
   void addCellToCurrentTrackCA(const Int_t, const Int_t, ROframe<T>&);
   void addCellToCurrentRoad(ROframe<T>&, const Int_t, const Int_t, const Int_t, const Int_t, Int_t&);
 
+  int mTrackerID = 0;
   Float_t mBz;
   std::vector<MCCompLabel> mTrackLabels;
   std::unique_ptr<o2::mft::TrackFitter<T>> mTrackFitter = nullptr;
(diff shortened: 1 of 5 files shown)


robincaron13:
Hi @shahor02, thanks a lot for reporting this and providing the changes. I see the problem, indeed FullClusterScan is only used for cosmics now. I will try to reproduce it, to find why it is very slow and a way to optimize this process.

robincaron13:
Hi @shahor02, I tested the command line on CTFs from lhc22o (also 500 kHz, and better for MFT data), below on the plot: the time used per ROF as function of the number of clusters. From what I understood the last changes and PR that modify the tracking did not touch to the track finder of the FullClusterScan so if there is something slower than previously this mean the initialization of bins container impact the FullClusterScan. I re-tested the workflow to add small changes however I did not see a clear time improvement for now. I'm not sure if I can find a quick solution but I will continue to investigate this. 
TimePerNclusters

shahor02:
Hi @robincaron13 
thanks for checking, I don't think it is now slower than before, I just wondering if it is normal that it takes seconds to process ROF with a few 100 hits, even if this mode is used currently only for cosmic. Anyway, at the moment this is not critical (until we hit problem).

robincaron13:
Hi @shahor02 , I don't have the exact time used previously in mind (around ~100x longer for the total time LTF+CA) but yes enable the full cluster scan is taking really long to process if we compared to the default case.

PR #10750, opened 2023-02-10
gen_topo_helper_functions.sh is sourced from setenv.sh

martenole:


chiarazampolli:
Hello @martenole , 
Just to explain why I kept it outside the setenv: only after sourcing the setenv.sh, one would have had the functions available, but then other env variables are set, and in async reco it does not work. 
Since in the end in async I need these functions early, I added the sourcing in async_pass.sh (https://github.com/AliceO2Group/O2DPG/blob/2d9ec70f8452de1e550743659aea21163314c640/DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh#L170), so your change will be fine.
Chiara

martenole:
Hi Chiara,
sure, I understand why the functions and the env vars should be separated and that is completely fine. Just the path had to be fixed: https://github.com/AliceO2Group/O2DPG/pull/866
But what is not clear yet is why the topology generation silently failed and did not produce an error message. Today, when they started runs at P2 the EPNs were just idling since the workflows were completely empty.
This here is just a cleanup, since online the helper functions are sourced from inside setenv anyways.
Cheers,
Ole

chiarazampolli:
Hello @martenole , 
Thanks, and apologies if there were troubles at P2, what I tested was apass, and there everything was working fine.
Chiara

martenole:
Yes, the fullCI also could not catch it, because there we use the scripts from inside O2, where the path is correctly set: https://github.com/AliceO2Group/AliceO2/blob/6cc09ae1189469401ecc9c3de89248b4aebc6a73/prodtests/full-system-test/gen_topo_helper_functions.sh#L8
But whenever we would used dpl-workflow.sh or aggregator-workflow.sh from O2DPG/DATA/production/ directory like we do in the topology generation on the EPNs it would fail. Its a bit convoluted with these scripts.

PR #10751, opened 2023-02-11
Example macro to convert filling scheme files *.csv into *.root

pbuehler:


pbuehler:
Added an example to CreateBCPattern.C which shows how a filling scheme file (*.csv) from https://lpc.web.cern.ch/cgi-bin/filling_schemes.py can be converted into a *.root file. The *.root files can be used in MC simulations to set a given filling scheme.

macro/CreateBCPattern.C
```diff
@@ -1,6 +1,7 @@
 #if !defined(__CLING__) || defined(__ROOTCLING__)
 #include "CommonDataFormat/BunchFilling.h"
 #include "CommonConstants/LHCConstants.h"
+#include "PWGUD/Core/UDFSParser.h"
```
sawenzel, line 4: Where does this come from? It does not seem to be within O2. I would suggest to move this macro to some other place (O2DPG/utils) if it depends on more than just O2.

pbuehler: Yes indeed this depends on O2Physics. So maybe better to put it there somewhere. On the other this PR is an extension of the existing macro CreateBCPattern.C and I think it could be of general interest.
I see three options: 1. make a copy of the modified  CreateBCPattern.C in O2Physics and drop the changes in O2. 2. Move PWGUD/Core/UDFSParser.h to O2. UDFSParser.h allows to parse filling scheme files *.csv. 3. keep the modified CreateBCPattern.C privat.

pbuehler:
I close this PR and move the  changes of CreateBCPattern.C to a macro in O2Physics/PWGUD/macro.

PR #10753, opened 2023-02-12
Optionally expand TPC tracks cov.matrix in global tracking/refits

shahor02:
New configurable param TrackTuneParams allows expanding cov.matrix of TPC track used in global tracking.
If `trackTuneParams.tpcCovInnerType` and `trackTuneParams.tpcCovOuterType` is not `TrackTuneParams::AddCovType::Disable`, then the content of float[5] arrays `trackTuneParams.tpcCovInner` and `trackTuneParams.tpcCovOuterType` respectively will be used to augment TPC inner or outer param before matching to ITS, TRD, TOF. The addition will preserve the cov. matrix correlations if the `tpcCovXXXType` is set to WithCorrelations, otherwise just the diagonal elements will be augmented, see [updateCovCorr](https://github.com/shahor02/AliceO2/blob/5ffee6555a0cbbb32aa8680c849ddccae1c45bd7/DataFormats/Reconstruction/include/ReconstructionDataFormats/TrackParametrizationWithError.h#L406) and [updateCov](https://github.com/shahor02/AliceO2/blob/5ffee6555a0cbbb32aa8680c849ddccae1c45bd7/DataFormats/Reconstruction/include/ReconstructionDataFormats/TrackParametrizationWithError.h#L388) methods respectively.
Modification of TPC tracks cov.matrix is done when creating matching seed and also when the TPC tracks needs to be refitted (with constrained time) for the following refit in external detector, i.e. 
1) TPC-ITS matching : in the inward refit of TPC track is refitted and its cov. matrix augmented at the TPC inner radius
2) TRD matching: in the outward refit TPC tracks is augmented before refit in the TRD.

The configurable param can be provided to `o2-tpcits-match-workflow`, `o2-trd-global-tracking` and `o2-tof-matcher-workflow`

### DIFF
--- a/Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
+++ b/Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
@@ -35,6 +35,8 @@
 #include "DataFormatsFT0/RecPoints.h"
 #include "ITSReconstruction/RecoGeomHelper.h"
 #include "ITSMFTReconstruction/ClustererParam.h"
+#include "DataFormatsGlobalTracking/TrackTuneParams.h"
+
 // GPU header
 #include "GPUReconstruction.h"
 #include "GPUChainTracking.h"
@@ -314,6 +316,7 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
   }
 
   // load input tracks
+  const auto& trackTune = o2::globaltracking::TrackTuneParams::Instance();
   LOG(debug) << "Start loading input seeds into TRD tracker";
   int nTracksLoadedITSTPC = 0;
   int nTracksLoadedTPC = 0;
@@ -325,6 +328,9 @@ void TRDGlobalTracking::run(ProcessingContext& pc)
     trkAttribs.mTimeAddMax = trkITSTPC.getTimeMUS().getTimeStampError() * mRec->GetParam().rec.trd.nSigmaTerrITSTPC + mRec->GetParam().rec.trd.addTimeRoadITSTPC;
     trkAttribs.mTimeSubMax = trkITSTPC.getTimeMUS().getTimeStampError() * mRec->GetParam().rec.trd.nSigmaTerrITSTPC + mRec->GetParam().rec.trd.addTimeRoadITSTPC;
     GPUTRDTrack trkLoad(trkITSTPC);
+    if (trackTune.tpcCovOuterType != o2::globaltracking::TrackTuneParams::AddCovType::Disable) {
+      trkLoad.updateCov(trackTune.tpcCovOuter, trackTune.tpcCovOuterType == o2::globaltracking::TrackTuneParams::AddCovType::WithCorrelations);
+    }
     auto trackGID = GTrackID(iTrk, GTrackID::ITSTPC);
     if (mTracker->LoadTrack(trkLoad, trackGID.getRaw(), true, &trkAttribs)) {
       continue;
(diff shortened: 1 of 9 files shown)


Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -683,19 +704,32 @@ bool TRDGlobalTracking::refitTPCTRDTrack(TrackTRD& trk, float timeTRD, o2::globa
...
   if (inwards) {
     trkParam = &trk;
     tofL = &trk.getLTIntegralOut();
   } else {
     trkParam = &trk.getOuterParam();
     trkParam->setUserField(trk.getUserField()); // pileup timing info
+
+    const auto& trackTune = o2::globaltracking::TrackTuneParams::Instance();
+    if (trackTune.tpcCovOuterType != o2::globaltracking::TrackTuneParams::AddCovType::Disable) {
+      if (!propagator->PropagateToXBxByBz(*trkParam, o2::constants::geom::XTPCOuterRef, o2::base::Propagator::MAX_SIN_PHI, o2::base::Propagator::MAX_STEP, matCorr)) {
```
martenole, line 724: Should this not only be done for the outwards refit and not for the inwards refit, as there we are starting at the outermost TRD point?
Since the covariance matrix is reset for the inwards fit in any case I think it does not matter, but this would just add unnecessary back and forth propagation, no?

martenole (commented):
Should the covariance not be updated also for the TPC-TRD tracks at the inner TPC radius?
And just out of curiosity: why is there no correction added for Z in https://github.com/AliceO2Group/O2DPG/pull/869 ?

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
shahor02: Hi @martenole 

> Should the covariance not be updated also for the TPC-TRD tracks at the inner TPC radius?

I am expanding the TPC cov matrix only before a fit/matching in the external detector, to fix the matching. The TPC cov. itself (w/o ITS update) is not used at the moment in the analysis (unless the V0 finder with TPC-only tracks is used, in which case the error will propagate to V0 error), but in any case, I  would prefer to expand track errors right before using them for the refit: the only purpose of this operation is to improve the matching eff. Otherwise, I would need to expand bare TPC track errors at their reconstruction level.

>  And just out of curiosity: why is there no correction added for Z in https://github.com/AliceO2Group/O2DPG/pull/869 ?

Z is not used in the matching and I could not calibrate the sigma to add from the difference of TPC and ITS tracks at the matching point, since in Z TPC track is simply moved to match the ITS. Still, I could add the same sigma for the symmetry reasons as for Y.

> Should this not only be done for the outwards refit and not for the inwards refit, as there we are starting at the outermost TRD point? Since the covariance matrix is reset for the inwards fit in any case I think it does not matter, but this would just add unnecessary back and forth propagation, no?

For the inwad refit w/o ITS update see above. For the outward: the point is that the track refitted in the TPC has distortion on both edges and these distortions are not accounted for by the track cov. matrix. So, (for the TRD refit in this case) I manipulate the TPC cov. matrix at the outer boundary of the TPC in order to not affect the TRD outer param by TPC distortions. Same I do to minimize their effect on the [matching](https://github.com/shahor02/AliceO2/blob/3a23e785dbd7cf523eec2c3dbe7292725f5037f8/Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx#L331). That said, at the moment the updateCov at TPC outer radius is a placeholder: I did not activate `trackTune.tpcCovOuterType`.

martenole: Ah, sorry, I misread the changes in github and see it better in the code directly now. In case of the TRD refit it is only done before the outwards refit and not before the inwards fit.
Thank you for the clarifications

PR #10754, opened 2023-02-12
[MRRTF-208] Migrate MCH digit I/O to IO and DevIO libs

aphecetche:
Also did quite a bit of renaming in DevIO, to stick to the following convention :

- a sampler/sink is reading/writing from/to a binary MCH custom file format
- a reader/writer is reading/writing from/to a Root file

The former is for debug and/or Run2 references, while the latter is what's used for Run3 production (data and simulation).

@pillot this is a first step towards what we discussed. In the end, I decided to split the PRs per object type (digits, reclusters, clusters, tracks), in order to have a (more) manageable piece to review... This first PR is for digits.

I've tried to make a number of tests to check I've not broken too many things (see details in https://alice.its.cern.ch/jira/browse/MRRTF-208) but might of course have overlooked some use cases. You'll let me know ;-)


aphecetche:
@sawenzel please note that your review/approval is needed for the two changes in `Steer/DigitizerWorkflow/CMakeLists.txt` and `Steer/src/SimpleDigitizerWorkflow.cxx` which reflect the move of the MCH digit writer to the new MCH IO library (so we have only one digit writer for our digits in the end).

pillot (approved):

aphecetche:
@shahor02 could this one be merged ? (I have 3 others in the pipeline that needs to be done sequentially and the next one depends on this one). Thanks !

PR #10765, opened 2023-02-14
DPL: add callback when exit was requested

ktf:


Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -1056,6 +1056,7 @@ void DataProcessingDevice::Run()
         auto timeout = deviceContext.exitTransitionTimeout;
         if (timeout != 0 && state.streaming != StreamingState::Idle) {
           state.transitionHandling = TransitionHandlingState::Requested;
+          ref.get<CallbackService>()(CallbackService::Id::ExitRequested, ref);
```
ktf, line 1059: @aalkin I do not understand your comment from earlier today about the callback not being actually invoked. AFAICT it is, no?

ktf:
Error in o2Sim unrelated.

PR #10769, opened 2023-02-15
Mft dcs config workflow deadmap

syano0822:
DCS Config file transfer to the CCDB workflow has been modified.

### DIFF
--- a/Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
+++ b/Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
@@ -35,9 +35,6 @@ class DCSConfigUtils
   {
     init();
   }
-
-  ~DCSConfigUtils() = default;
-
   void init()
   {
     initDictionary();
@@ -47,34 +44,34 @@ class DCSConfigUtils
   {
   }
 
-  auto getAddress(const std::string name, const std::string type)
+  int getAddress(std::string name, std::string type)
   {
-    return static_cast<int>((mNameDictType[type])[name]);
+    return (mNameDict[type])[name];
   }
 
-  const std::string getName(int add, const std::string type)
+  std::string getName(int add, std::string type)
   {
-    return static_cast<const std::string>((mAddressDictType[type])[add]);
+    return (mAddressDict[type])[add];
   }
 
-  std::map<int, const std::string>& getAddressMap(const std::string type)
+  std::map<int, std::string> getAddressMap(std::string type)
   {
-    return mAddressDictType[type];
+    return mAddressDict[type];
   }
 
-  std::map<int, const std::string>& getAddressMap(int type)
+  std::map<int, std::string> getAddressMap(int type)
   {
-    return mAddressDictType[getTypeName(type)];
+    return mAddressDict[getTypeName(type)];
   }
 
-  std::map<const std::string, int>& getNameMap(const std::string type)
+  std::map<std::string, int> getNameMap(std::string type)
   {
-    return mNameDictType[type];
+    return mNameDict[type];
   }
 
-  std::map<const std::string, int>& getNameMap(int type)
+  std::map<std::string, int> getNameMap(int type)
   {
-    return mNameDictType[getTypeName(type)];
+    return mNameDict[getTypeName(type)];
   }
 
   const std::string getTypeName(int type)
@@ -93,7 +90,7 @@ class DCSConfigUtils
     }
   }
 
-  const int getType(const std::string type)
+  const int getType(std::string type)
   {
     if (type == "RU") {
       return 0;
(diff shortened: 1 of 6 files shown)


Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
```diff
@@ -47,34 +48,36 @@ class DCSConfigUtils
   {
   }
 
-  auto getAddress(const std::string name, const std::string type)
+  int getAddress(std::string name, std::string type)
   {
-    return static_cast<int>((mNameDictType[type])[name]);
+    std::map<std::string, int> dict = mNameDict[type];
+    return static_cast<int>(dict[name]);
```
shahor02: why do you need to create a copy of `mNameDict[type]` to query the "name"? Use a reference or do directly
```
return (mNameDict[type])[name];
```

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
```diff
@@ -47,34 +48,36 @@ class DCSConfigUtils
...
   }
 
-  const std::string getName(int add, const std::string type)
+  std::string getName(int add, std::string type)
   {
-    return static_cast<const std::string>((mAddressDictType[type])[add]);
+    std::map<int, std::string> dict = mAddressDict[type];
+    return static_cast<std::string>(dict[add]);
```
shahor02: idem.

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
```diff
@@ -47,34 +48,36 @@ class DCSConfigUtils
...
-    return mNameDictType[type];
+    return mNameDict[type];
   }
 
-  std::map<const std::string, int>& getNameMap(int type)
+  std::map<std::string, int> getNameMap(int type)
   {
-    return mNameDictType[getTypeName(type)];
+    return mNameDict[getTypeName(type)];
   }
```
shahor02, line 75: All these methods return copies of the maps, are you sure this is what you want and not references?

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
```diff
@@ -109,171 +112,153 @@ class DCSConfigUtils
...
 
-  std::map<const std::string, std::map<const std::string, int>> mNameDictType;
-  std::map<const std::string, int> mNameDictRu;
-  std::map<const std::string, int> mNameDictAlpide;
-  std::map<const std::string, int> mNameDictUbb;
-  std::map<const std::string, int> mNameDictDeadMap;
+  std::map<std::string, std::map<std::string, int>> mNameDict;
+  std::map<std::string, int> mNameDictRU;
+  std::map<std::string, int> mNameDictALPIDE;
+  std::map<std::string, int> mNameDictUBB;
```
shahor02, line 162: as far as I can see, these maps are used just to fill the content of the `mNameDict`, are they needed as data members rather than being local to `initDictionary()` (or even just a references extracted there from the `mNameDict`).

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
```diff
@@ -109,171 +112,153 @@ class DCSConfigUtils
...
 
-  std::map<const std::string, std::map<int, const std::string>> mAddressDictType;
-  std::map<int, const std::string> mAddressDictRu;
-  std::map<int, const std::string> mAddressDictAlpide;
-  std::map<int, const std::string> mAddressDictUbb;
-  std::map<int, const std::string> mAddressDictDeadMap;
+  std::map<std::string, std::map<int, std::string>> mAddressDict;
+  std::map<int, std::string> mAddressDictRU;
+  std::map<int, std::string> mAddressDictALPIDE;
+  std::map<int, std::string> mAddressDictUBB;
```
shahor02, line 167: idem.

shahor02 (changes requested):
Hi @syano0822 

Please see some comments below. Also note that the fullCI failure is genuine:
```
/sw/SOURCES/O2/10769-slc8_x86-64/0/Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h:38:3: error: use '= default' to define a trivial destructor [modernize-use-equals-default]

```

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
syano0822: DONE.

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
syano0822: DONE

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
syano0822: This class is used to get information that can be understood by a human (just an interpreter). In the initiDictionaly, the dictionaries (maps) are created and then it is used to translate machine-like information, e.g. addresses, to human-kind information, e.g. variable name. So, this is what I want to do.

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
syano0822: The mNameDicts are used in member functions in this class. So they are necessary to be data members.

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
shahor02: I mean `std::map<std::string, int> getNameMap(int type)` creates/returns a copy of the `mNameDict[getTypeName(type)]` at each call instead of returning a reference on the existing map. Sorry, I don't understand what it has to do with the type of information. Anyway, this does not prevent the class from working, can be changed later.

Detectors/ITSMFT/MFT/condition/include/MFTCondition/DCSConfigUtils.h
shahor02: I was referring on mNameDictRU, mNameDictALPIDE and mNameDictUBB, which are filled and used only in the initialization.

syano0822:
The PR doesn't pass the build/AliceO2/O2/o2/macOS-arm test, but it seems not directly related to the PR. Is it right?

PR #10771, opened 2023-02-15
Fix the RDH V7

fapfap69:


Detectors/HMPID/reconstruction/src/HmpidDecoder.cxx
```diff
@@ -89,8 +89,8 @@ HmpidDecoder::~HmpidDecoder()
 /// Init all the members variables.
 void HmpidDecoder::init()
 {
-  mRDHAcceptedVersion = 6;
-  mRDHSize = sizeof(o2::header::RAWDataHeaderV6) / sizeof(uint32_t);
+  mRDHAcceptedVersion = 7;
+  mRDHSize = sizeof(o2::header::RAWDataHeaderV7) / sizeof(uint32_t);
```
shahor02: I don't see where mRDHAcceptedVersion is used, but we should be able to work also with V6. Better use generic `o2::header::RAWDataHeader` which is an alias to default RDH in use. In case you need the nominal default version, you can use
```
o2::header::RAWDataHeader rdh;
mRDHAcceptedVersion = rdh.version;
mRDHSize = rdh.headerSize;
```

Detectors/HMPID/reconstruction/src/HmpidDecoder.cxx
```diff
@@ -359,7 +359,7 @@ bool HmpidDecoder::decodeHmpidError(int ErrorField, char* outbuf)
 int HmpidDecoder::decodeHeader(uint32_t* streamPtrAdr, int* EquipIndex)
 {
   uint32_t* buffer = streamPtrAdr; // Sets the pointer to buffer
-  o2::header::RAWDataHeaderV6* hpt = (o2::header::RAWDataHeaderV6*)buffer;
+  o2::header::RAWDataHeaderV7* hpt = (o2::header::RAWDataHeaderV7*)buffer;
```
shahor02: Idem, please use `o2::header::RAWDataHeader` which for you should be the same for all versions >= 5.

Detectors/HMPID/reconstruction/src/HmpidDecoder2.cxx
```diff
@@ -95,8 +95,8 @@ void HmpidDecoder2::init()
 {
   mVerbose = 0;
 
-  mRDHAcceptedVersion = 6;
-  mRDHSize = sizeof(o2::header::RAWDataHeaderV6) / sizeof(uint32_t);
+  mRDHAcceptedVersion = 7;
+  mRDHSize = sizeof(o2::header::RAWDataHeaderV7) / sizeof(uint32_t);
```
shahor02: idem

Detectors/HMPID/reconstruction/src/HmpidDecoder2.cxx
```diff
@@ -376,7 +376,7 @@ bool HmpidDecoder2::decodeHmpidError(int ErrorField, char* outbuf)
 int HmpidDecoder2::decodeHeader(uint32_t* streamPtrAdr, int* EquipIndex)
 {
   uint32_t* buffer = streamPtrAdr; // Sets the pointer to buffer
-  o2::header::RAWDataHeaderV6* hpt = (o2::header::RAWDataHeaderV6*)buffer;
+  o2::header::RAWDataHeaderV7* hpt = (o2::header::RAWDataHeaderV7*)buffer;
```
shahor02: idem

shahor02 (changes requested):

shahor02 (approved):

PR #10773, opened 2023-02-15
Add possibility to filter the collision BCs and merge digits in adjacent BCs

dstocco:
The filtering is not activated by default. This possibility is mainly useful for testing purposes.

### DIFF
--- a/Detectors/MUON/MID/Filtering/src/FiltererBC.cxx
+++ b/Detectors/MUON/MID/Filtering/src/FiltererBC.cxx
@@ -1,0 +1,113 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file   MID/Filtering/src/FiltererBC.cxx
+/// \brief  BC filterer for MID
+/// \author Diego Stocco <Diego.Stocco at cern.ch>
+/// \date   23 January 2023
+
+#include "MIDFiltering/FiltererBC.h"
+
+namespace o2
+{
+namespace mid
+{
+
+int FiltererBC::matchedCollision(int bc)
+{
+  // First we check the simple case to save time
+  if (mBunchFilling.testInteractingBC(bc)) {
+    return bc;
+  }
+
+  // Then we check for neighbor BCs
+  // We split the check in two so that we start the check from the input bc
+  // and we move farther from it.
+
+  int matchedBC = std::numeric_limits<int>::max();
+
+  // We start on the upper direction
+
+  // The allowed BC are o2::constants::lhc::LHCMaxBunches.
+  // Numeration starts from 0, so o2::constants::lhc::LHCMaxBunches is not allowed.
+  // This ensures that we do not check out of boundaries
+  int end = bc + mBCDiffHigh;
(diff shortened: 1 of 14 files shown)


dstocco:
@aphecetche : errors seem unrelated. I guess you can safely start the review

Detectors/MUON/MID/Workflow/README.md
```diff
@@ -117,6 +117,32 @@ o2-mid-reco-workflow --change-local-to-BC <value>
...
 
+## MID BC filtering
+
+The MID time resolution is better than 25 ns, allowing it to distinguish between different BCs.
+However, a spread of the signal was observed, probably due an insufficient equalization of the delays across the various detection element.
+In order to study and possibly correct for any inefficiency arising from this spread, it is possible to select the BCs corresponding to a collision and merge the digits in a configurable window around this BC.
+To this aim, it is enough to run the reconstruction with the option:
+
+```bash
+o2-mid-reco-workflow --filter-BC <value>
```
aphecetche: actually the filter-BC being a bool I think you can’t specify a value, can you ? 
also, seems most of the boolean options we have in O2 follow the naming convention `--enable-xxx` or `disable-xxx`, so I would follow it here, e.g. `--enable-filter-bc` (as the default is false).

Detectors/MUON/MID/Workflow/src/FilteringBCSpec.cxx
```diff
@@ -0,0 +1,123 @@
...
+    mFiltererBC.setBCDiffHigh(FiltererBCParam::Instance().maxBCDiffHigh);
+    mFiltererBC.setSelectOnly(FiltererBCParam::Instance().selectOnly);
+  }
+
+  void finaliseCCDB(of::ConcreteDataMatcher matcher, void* obj)
+  {
+    if (o2::base::GRPGeomHelper::instance().finaliseCCDB(matcher, obj)) {
+      LOG(info) << "Update Bunch Filling for MID";
+      mFiltererBC.setBunchFilling(o2::base::GRPGeomHelper::instance().getGRPLHCIF()->getBunchFilling());
+      return;
```
aphecetche: that's a detail but the return is superfluous

Detectors/MUON/MID/Workflow/src/FilteringBCSpec.cxx
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  void run(of::ProcessingContext& pc)
+  {
+    updateTimeDependentParams(pc);
+
+    auto data = specs::getData(pc, "mid_filter_BC_in", EventType::Standard);
+    auto inROFRecords = specs::getRofs(pc, "mid_filter_BC_in", EventType::Standard);
+
+    std::unique_ptr<const o2::dataformats::MCTruthContainer<MCLabel>> inMCContainer = mUseMC ? specs::getLabels(pc, "mid_filter_BC_in") : nullptr;
```
aphecetche: any particular reason you're not using the simpler `auto inMCContainer = ...` like for the other inputs ?

Detectors/MUON/MID/Filtering/src/FiltererBC.cxx
```diff
@@ -0,0 +1,113 @@
...
+  // First we check the simple case to save time
+  if (mBunchFilling.testInteractingBC(bc)) {
+    return bc;
+  }
+
+  // Then we check for neighbor BCs
+  // We split the check in two so that we start the check from the input bc
+  // and we move farther from it.
+
+  int matchedBC = 3 * o2::constants::lhc::LHCMaxBunches;
```
aphecetche: why `3*` ?

dstocco: I just need a value larger than o2::constants::lhc::LHCMaxBunches. The factor does not really matter.

dstocco: Changed into: `int matchedBC = std::numeric_limits<int>::max();`

aphecetche (approved):

PR #10781, opened 2023-02-15
DPL: typesafe callback API

ktf:
Incremental commit migrating from the soon to be deprecated runtime API to a new typesafe version.

Correct a couple of mis-invocations while doing so (which could have affected start / stop / start).


ktf:
This is just the first part, introducing the new API, so that I can adapt QC as well without disruption. A followup PR will introduce the reimplemented / simplified CallbackRegistry which will be much faster to compile and less error prone.

ktf:
macOS ARM issue unrelated (o2Sim.C failing). Merging.

PR #10783, opened 2023-02-16
Add RDHv7 option

peressounko:
according to jira O2-3537

### DIFF
--- a/Detectors/PHOS/reconstruction/src/RawReaderMemory.cxx
+++ b/Detectors/PHOS/reconstruction/src/RawReaderMemory.cxx
@@ -40,6 +40,8 @@ o2::header::RDHAny RawReaderMemory::decodeRawHeader(const void* payloadwords)
     return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV5*>(payloadwords));
   } else if (headerversion == 6) {
     return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV6*>(payloadwords));
+  } else if (headerversion == 7) {
+    return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV7*>(payloadwords));
   }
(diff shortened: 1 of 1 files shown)


peressounko:
Dear Ruben, after this fix PHOS should be ready for RDHv7 tests

Detectors/PHOS/reconstruction/src/RawReaderMemory.cxx
```diff
@@ -40,6 +40,8 @@ o2::header::RDHAny RawReaderMemory::decodeRawHeader(const void* payloadwords)
     return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV5*>(payloadwords));
   } else if (headerversion == 6) {
     return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV6*>(payloadwords));
+  } else if (headerversion == 7) {
+    return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV7*>(payloadwords));
   }
```
shahor02, line 45: This will work but in principle, you don't need all this `if`'s (in the function which is called very often) and header copy, you can do simply 
```
const o2::header::RDHAny* RawReaderMemory::decodeRawHeader(const void* payloadwords) const
  auto rdh =  reinterpret_cast<const o2::header::RDHAny*>(&payloadwords);
  // in case you want to check the header version:
  auto v = RDHDecoder::getVersion(*rdh);
  if (v < RDHDecoder::getVersion<o2::header::RDHLowest>() || v>RDHDecoder::getVersion<o2::header::RDHHighest>()) {
     LOG(error) << "Wrong header version " << v;
     throw RawDecodingError::ErrorType_t::HEADER_DECODING;
  }
  return rdh;
```

shahor02:
@peressounko I've opened a commit with uniform fix for EMC, PHS and CPV: https://github.com/AliceO2Group/AliceO2/pull/10794, closing this one.

PR #10792, opened 2023-02-16
TRD: QC: Write corrected charges

f3sch:
TrackingQC produces now also the corrected charges (for now baring gain calibration) including z-row merging

Tagging @jcrkovsk for approval, as this pulls the krypton calibrations from the ccdb and corrects (hopefully) the charges by multiplication of the local gain factor (correct?).

### DIFF
--- a/Detectors/TRD/qc/include/TRDQC/Tracking.h
+++ b/Detectors/TRD/qc/include/TRDQC/Tracking.h
@@ -18,6 +18,7 @@
 
 #include "DataFormatsTRD/TrackTRD.h"
 #include "DataFormatsTRD/Tracklet64.h"
+#include "TRDBase/PadCalibrationsAliases.h"
 #include "DataFormatsTRD/CalibratedTracklet.h"
 #include "DataFormatsTRD/Constants.h"
 #include "ReconstructionDataFormats/TrackTPCITS.h"
@@ -25,9 +26,9 @@
 #include "DataFormatsTPC/TrackTPC.h"
 #include "DetectorsBase/Propagator.h"
 #include "TRDBase/RecoParam.h"
-#include "TH1.h"
 
 #include "Rtypes.h"
+#include "TH1.h"
 
 #include <gsl/span>
 #include <bitset>
@@ -83,12 +84,14 @@ struct TrackQC {
   std::array<int, constants::NLAYER> trackletPositionSigned{}; ///< the raw position from Tracklet64 (signed integer)
   std::array<int, constants::NLAYER> trackletDet{};            ///< the chamber of the tracklet
   // some tracklet details to identify its global MCM number to check if it is from noisy MCM
-  std::array<int, constants::NLAYER> trackletHCId{};                                     ///< the half-chamber ID of the tracklet
-  std::array<int, constants::NLAYER> trackletRob{};                                      ///< the ROB number of the tracklet
-  std::array<int, constants::NLAYER> trackletMcm{};                                      ///< the MCM number of the tracklet
-  std::array<float, constants::NLAYER> trackletChi2{};                                   ///< estimated chi2 for the update of the track with the given tracklet
-  std::array<std::array<int, constants::NCHARGES>, constants::NLAYER> trackletCharges{}; ///< charges of tracklets
-  ClassDefNV(TrackQC, 4);
+  std::array<int, constants::NLAYER> trackletHCId{};                                          ///< the half-chamber ID of the tracklet
+  std::array<int, constants::NLAYER> trackletRob{};                                           ///< the ROB number of the tracklet
+  std::array<int, constants::NLAYER> trackletMcm{};                                           ///< the MCM number of the tracklet
+  std::array<float, constants::NLAYER> trackletChi2{};                                        ///< estimated chi2 for the update of the track with the given tracklet
+  std::array<std::array<int, constants::NCHARGES>, constants::NLAYER> trackletCharges{};      ///< charges of tracklets
+  std::array<std::array<float, constants::NCHARGES>, constants::NLAYER> trackletCorCharges{}; ///< corrected charges of tracklets
+
+  ClassDefNV(TrackQC, 5);
 };
 
 class Tracking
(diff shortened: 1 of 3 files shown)


Detectors/TRD/base/include/TRDBase/PadParameters.h
```diff
@@ -74,10 +74,10 @@ int PadParameters<T>::getChannel(int col, int row) const
     return row + col * constants::NROWC0;
   } else if (mData.size() == constants::NROWC1 * constants::NCOLUMN) {
     return row + col * constants::NROWC1;
-  } else {
-    LOG(error) << "Wrong number of channels set: " << mData.size();
   }
-  return -1;
+
+  LOG(error) << "Wrong number of channels set: " << mData.size();
```
martenole: Why the change here? In case the class is not initialized you would now get two different error messages

Detectors/TRD/qc/src/Tracking.cxx
```diff
@@ -117,19 +124,25 @@ void Tracking::checkTrack(const TrackTRD& trkTrd, bool isTPCTRD)
...
-      mTrackletsRaw[trkltId].getQ2(),
+      tracklet.getQ0(),
+      tracklet.getQ1(),
+      tracklet.getQ2(),
+    };
+    auto correction = mLocalGain.getValue(tracklet.getHCID() / 2, tracklet.getPadCol(), tracklet.getPadRow());
+    qcStruct.trackletCorCharges[iLayer] = {
+      static_cast<float>(tracklet.getQ0()) * correction,
+      static_cast<float>(tracklet.getQ1()) * correction,
+      static_cast<float>(tracklet.getQ2()) * correction,
```
martenole: don't think the `static_cast`s are needed here, or?

Detectors/TRD/qc/src/Tracking.cxx
```diff
@@ -37,6 +38,11 @@ void Tracking::setInput(const o2::globaltracking::RecoContainer& input)
   mTracksTPCTRD = input.getTPCTRDTracks<TrackTRD>();
   mTrackletsRaw = input.getTRDTracklets();
   mTrackletsCalib = input.getTRDCalibratedTracklets();
+
+  if (!mCCDBInit) {
+    mCCDBInit = true;
+    mLocalGain = *(pc.inputs().get<o2::trd::LocalGainFactor*>("localgainfactors"));
```
martenole: I would prefer to set the CCDB object from the TRDGlobalTrackingQCSpec, e.g. via a new method `Tracking::setLocalGainFactors()`. This could be called e.g. from inside `TRDGlobalTrackingQC::updateTimeDependentParams()`. Then you don't need the init flag and the additional includes

Detectors/TRD/base/include/TRDBase/PadParameters.h
f3sch: Right missed the missing return in the if

Detectors/TRD/qc/src/Tracking.cxx
f3sch: probably not

Detectors/TRD/qc/include/TRDQC/Tracking.h
```diff
@@ -83,12 +86,14 @@ struct TrackQC {
...
-  std::array<int, constants::NLAYER> trackletMcm{};                                      ///< the MCM number of the tracklet
-  std::array<float, constants::NLAYER> trackletChi2{};                                   ///< estimated chi2 for the update of the track with the given tracklet
-  std::array<std::array<int, constants::NCHARGES>, constants::NLAYER> trackletCharges{}; ///< charges of tracklets
-  ClassDefNV(TrackQC, 4);
+  std::array<int, constants::NLAYER> trackletHCId{};                                          ///< the half-chamber ID of the tracklet
+  std::array<int, constants::NLAYER> trackletRob{};                                           ///< the ROB number of the tracklet
+  std::array<int, constants::NLAYER> trackletMcm{};                                           ///< the MCM number of the tracklet
+  std::array<float, constants::NLAYER> trackletChi2{};                                        ///< estimated chi2 for the update of the track with the given tracklet
+  std::array<std::array<int, constants::NCHARGES>, constants::NLAYER> trackletCharges{};      ///< charges of tracklets
```
bazinski, line 91: Has the whole rule of not doing arrays of arrays been dropped?

Detectors/TRD/qc/src/Tracking.cxx
f3sch: Thanks for the review. Please check if these changes work for you.

Detectors/TRD/qc/include/TRDQC/Tracking.h
tdietel: I thought the problem were variable-size containers, e.g. vectors. Arrays as fixed size containers should be ok, and I think our digits actually contain an array.

Detectors/TRD/qc/include/TRDQC/Tracking.h
```diff
@@ -118,13 +123,21 @@ class Tracking
   // Make output accessible to DPL processor
   std::vector<TrackQC>& getTrackQC() { return mTrackQC; }
 
+  // Set the local gain factors with values from the ccdb
+  void setLocalGainFactors(o2::framework::ProcessingContext& pc)
```
martenole:
```suggestion
  void setLocalGainFactors(const o2::trd::LocalGainFactor* padGain)
```
Then you can drop the Framework includes above

Detectors/TRD/base/include/TRDBase/PadParameters.h
```diff
@@ -77,7 +77,8 @@ int PadParameters<T>::getChannel(int col, int row) const
   } else {
     LOG(error) << "Wrong number of channels set: " << mData.size();
   }
-  return -1;
+
+  return 0;
```
martenole: I would just leave this as is, or why do you prefer a valid return value here?

Detectors/TRD/qc/include/TRDQC/Tracking.h
bazinski: I dont remember hence my query.

Detectors/TRD/qc/include/TRDQC/Tracking.h
f3sch: Ok

Detectors/TRD/base/include/TRDBase/PadParameters.h
f3sch: dropped

martenole (dismissed):
Thanks. It would of course also work the way you had it before, but I would like to keep all the framework stuff out of the detector code where it is not needed.

martenole (approved):

PR #10794, opened 2023-02-17
Adapt EMC, PHS and CPV to RDHv7

shahor02:
This just makes decoders to accept RDHv7 and optimizes the way RDH is extracted. 

@sevdokim, since the CPV is a CRU detector, it has to implement also the processing of the data w/o 128-bit padding. This is NOT handled by this PR.

### DIFF
--- a/Detectors/PHOS/reconstruction/src/RawReaderMemory.cxx
+++ b/Detectors/PHOS/reconstruction/src/RawReaderMemory.cxx
@@ -34,15 +34,11 @@ void RawReaderMemory::setRawMemory(const gsl::span<const char> rawmemory)
 o2::header::RDHAny RawReaderMemory::decodeRawHeader(const void* payloadwords)
 {
   auto headerversion = RDHDecoder::getVersion(payloadwords);
-  if (headerversion == 4) {
-    return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV4*>(payloadwords));
-  } else if (headerversion == 5) {
-    return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV5*>(payloadwords));
-  } else if (headerversion == 6) {
-    return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV6*>(payloadwords));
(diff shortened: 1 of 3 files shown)


Detectors/PHOS/reconstruction/src/RawReaderMemory.cxx
```diff
@@ -34,15 +34,11 @@ void RawReaderMemory::setRawMemory(const gsl::span<const char> rawmemory)
...
-  } else if (headerversion == 5) {
-    return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV5*>(payloadwords));
-  } else if (headerversion == 6) {
-    return o2::header::RDHAny(*reinterpret_cast<const o2::header::RAWDataHeaderV6*>(payloadwords));
+  if (headerversion < RDHDecoder::getVersion<o2::header::RDHLowest>() || headerversion > RDHDecoder::getVersion<o2::header::RDHHighest>()) {
+    LOG(error) << "Wrong header version " << headerversion;
+    throw RawDecodingError::ErrorType_t::HEADER_DECODING;
   }
-  LOG(error) << "Wrong header version " << int(headerversion);
-  throw RawDecodingError::ErrorType_t::HEADER_DECODING;
```
peressounko, line 45: Should this be removed??? I think this is copy-past error?

shahor02: Well, I've just preserved your error message and exception I case of a wrong header version. Do you want to change this behaviour?

peressounko: I think Log and through should be within if starting in L37 and should not appear in normal flow

shahor02: But they are within the `if` checking for the wrong version

peressounko: Agree, I looked a rose background, sorry

mfasDa (approved):

Detectors/PHOS/reconstruction/src/RawReaderMemory.cxx
shahor02: @peressounko do you still have some doubts or I should merge it?

peressounko (approved):

PR #10800, opened 2023-02-17
Additional robustness for topology generation

martenole:
To accompany https://github.com/AliceO2Group/O2DPG/pull/893


davidrohr (dismissed):
without checing everything in detail, looks good

prodtests/full-system-test/workflow-setup.sh
```diff
@@ -1,140 +1,8 @@
...
-done
-if [[ $SYNCMODE == 1 ]]; then # Add default steps for synchronous mode
-  add_comma_separated WORKFLOW_EXTRA_PROCESSING_STEPS ENTROPY_ENCODER
-else # Add default steps for async mode
-  for i in $LIST_OF_ASYNC_RECO_STEPS; do
-    has_detector_reco $i && add_comma_separated WORKFLOW_EXTRA_PROCESSING_STEPS ${i}_RECO
-  done
-  add_comma_separated WORKFLOW_EXTRA_PROCESSING_STEPS TPC_DEDX
+if [ "0$O2DPG_ROOT" == "0" ]; then
+    echo O2DPG environment not loaded
```
davidrohr: please add `1>&2`

prodtests/full-system-test/aggregator-workflow.sh
```diff
@@ -1,16 +1,22 @@
 #!/bin/bash
 
+# Non-zero exit code already if one command in a pipe fails
+set -o pipefail
```
davidrohr, line 4: you can combine it to set -uo pipefail

martenole: Actually, I changed it now so that one could disable the check just in case I have missed something and otherwise we would need a new O2 for the topology generation to work. After a few weeks we can add it all to one line

prodtests/full-system-test/workflow-setup.sh
martenole: Done and also in the other scripts where it was missing where I took that snippet from

martenole:
Tested locally with FST, topology generation and default async reco. In case I have still forgotten something the check for unbound variables can be disabled via `IGNORE_UNBOUND_VARIABLES=1`.

@chiarazampolli and @shahor02 whenever now dpl-workflow or aggregator-workflow scripts are used to create a workflow/topology any variable which is not set will trigger an error. This should help us to spot typos when we set some variable and then use it without checking if it is actually set to something or not.

Whenever we don't know if a variable is set externally or not, we must use parameter expansion
```
[[ ! -z $INPUT_FILE_LIST ]]
```
becomes
```
[[ ! -z ${INPUT_FILE_LIST:-} ]]
```
If we expect the variable to be set we omit the expansion of course, that way triggering error messages when we do typos.

In oder to set a variable to a default value when we don't know if it has been set externally or not we can do
```
: ${AGGREGATOR_TASKS:=ALL}
```
if AGGREGATOR_TASK is set to something already then it is not changed, otherwise it is set to "ALL". So it is equivalent of doing
```
if [[ -z ${AGGREGATOR_TASKS:-} ]]; then AGGREGATOR_TASKS=ALL; fi
```

The parameter expansions are nicely summarized here: https://stackoverflow.com/a/16753536

martenole:
Merging to have it in tomorrow's tag. The macOS error is new, but appears also in other PRs (e.g. https://github.com/AliceO2Group/AliceO2/pull/11152).
Maybe related to the changes for python 3.11 in https://github.com/alisw/alidist/commit/85bce8114959767b2e4eeea6cb469fbda306d245 @TimoWilken ?
```
## sw/BUILD/Python-modules-latest/log
--
      INFO: compile options: '-c'
      INFO: clang: test_program.c
      clang: warning: no such sysroot directory: '/Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk' [-Wmissing-sysroot]
      test_program.c:1:10: fatal error: 'stdio.h' file not found
      #include <stdio.h>
               ^~~~~~~~~
      1 error generated.
--
      clang -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk -fPIC -O2 -DNPY_NO_DEPRECATED_API=0 -I./pandas/_libs -Ipandas/_libs/src/klib -I/private/tmp/pip-build-env-gy0yug9m/overlay/lib/python3.11/site-packages/numpy/core/include -I/System/Volumes/Data/build/alice-ci-workdir/o2/sw/INSTALLROOT/2a17dc977325f0778e7bb78b115d78d70d3b1f93/osx_x86-64/Python-modules/1.0-local1/share/python-modules/include -I/usr/local/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/include/python3.11 -c pandas/_libs/algos.c -o build/temp.macosx-11-x86_64-cpython-311/pandas/_libs/algos.o -Wno-error=unreachable-code
      clang: warning: no such sysroot directory: '/Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk' [-Wmissing-sysroot]
      In file included from pandas/_libs/algos.c:45:
      /usr/local/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/include/python3.11/Python.h:23:12: fatal error: 'stdlib.h' file not found
      #  include <stdlib.h>
                 ^~~~~~~~~~
      1 error generated.
--
```

PR #10805, opened 2023-02-20
ctpdev: padding raw data

lietava:
1 in digi2raw you can choose padding: o2-ctp-digi2raw -padding 0
2 in reconstruction padding is selected according rdh
Note: there are code changes in Digitizer.cxx not relatig to padding - sorry.

### DIFF
--- a/Detectors/CTP/simulation/src/Digits2Raw.cxx
+++ b/Detectors/CTP/simulation/src/Digits2Raw.cxx
@@ -184,7 +184,7 @@ std::vector<char> Digits2Raw::digits2HBTPayload(const gsl::span<gbtword80_t> dig
   bool valid;
   for (auto const& dig : digits) {
     valid = makeGBTWord(dig, gbtword, size_gbt, Npld, gbtsend);
-    LOG(info) << Npld << " digit:" << dig << " " << (dig.to_ulong() & 0xfff) << " " ;
+    LOG(info) << Npld << " digit:" << dig << " " << (dig.to_ulong() & 0xfff) << " ";
     LOG(info) << "gbt  :" << gbtsend << " valid:" << valid;
     if (valid == true) {
       for (uint32_t i = 0; i < NGBT; i += 8) {
@@ -196,7 +196,7 @@ std::vector<char> Digits2Raw::digits2HBTPayload(const gsl::span<gbtword80_t> dig
         char c = w;
         toAdd.push_back(c);
       }
-      if(mPadding) {
+      if (mPadding) {
         // Pad zeros up to 128 bits
         uint32_t NZeros = (o2::raw::RDHUtils::GBTWord128 * 8 - NGBT) / 8;
         for (uint32_t i = 0; i < NZeros; i++) {
@@ -223,7 +223,7 @@ std::vector<char> Digits2Raw::digits2HBTPayload(const gsl::span<gbtword80_t> dig
       toAdd.push_back(c);
     }
     // Pad zeros up to 128 bits
-    if(mPadding) {
+    if (mPadding) {
       uint32_t NZeros = (o2::raw::RDHUtils::GBTWord128 * 8 - NGBT) / 8;
       for (uint32_t i = 0; i < NZeros; i++) {
         char c = 0;
@@ -231,13 +231,13 @@ std::vector<char> Digits2Raw::digits2HBTPayload(const gsl::span<gbtword80_t> dig
       }
     }
   }
-  if(mPadding == false) {
+  if (mPadding == false) {
     uint32_t NZeros = 0;
     if (countBytes % 16) {
       NZeros = 16 - (countBytes % 16);
     }
     LOG(info) << "Adding bytes:" << NZeros;
-    for(uint32_t i = 0; i < NZeros; i++) {
+    for (uint32_t i = 0; i < NZeros; i++) {
       char c = 0;
       toAdd.push_back(c);
     }
--- a/Detectors/CTP/simulation/src/Digits2Raw.cxx
+++ b/Detectors/CTP/simulation/src/Digits2Raw.cxx
@@ -27,7 +27,6 @@ using namespace o2::ctp;
 
 void Digits2Raw::init()
 {
-
   //
   // Register links
   //
@@ -36,6 +35,8 @@ void Digits2Raw::init()
     outd += '/';
   }
   LOG(info) << "Raw outpud dir:" << mOutDir;
+  //
+  LOG(info) << "Raw Data padding:" << mPadding << " Alignmant size:" << mAlignmentSize;
   // Interaction Record
   int ilink = 0;
   uint64_t feeID = getFEEIDIR();
@@ -48,6 +49,7 @@ void Digits2Raw::init()
   mWriter.registerLink(feeID, mCruID, ilink, mEndPointID, outFileLink1);
   // ilink = 2: HBMap, Counters - tbd
   mWriter.setEmptyPageCallBack(this);
+  mWriter.setAlignmentSize(mAlignmentSize);
 }
 void Digits2Raw::processDigits(const std::string& fileDigitsName)
 {
@@ -56,7 +58,7 @@ void Digits2Raw::processDigits(const std::string& fileDigitsName)
     LOG(fatal) << "Failed to open input digits file " << fileDigitsName;
     return;
   }
-  LOG(info) << "Processing digits to raw";
+  LOG(info) << "Processing digits to raw file:" << fileDigitsName;
   TTree* digiTree = (TTree*)digiFile->Get("o2sim");
   if (!digiTree) {
     LOG(fatal) << "Failed to get digits tree";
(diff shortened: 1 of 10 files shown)


lietava:
@shahor02 : here is the draft.
Questions/comments:
1.) With padding reconstruction is tested with chunk:
/alice/data/2022/LHC22t/529690/raw/0010/o2_rawtf_run00529690_tf00290160_epn136.tf
2.) No padding it is tested with MC. 
How can I specify RDH version in difi2raw ?
3.) in reconstruction to decide if padd or not I use only RDH . Should I use alsi reserved bit ?
4.) There are somewhere raw data without padding to test ?

shahor02:
> How can I specify RDH version in difi2raw ?

You have an option https://github.com/AliceO2Group/AliceO2/blob/03f1f73549cc06a8bdcfae538a0f1b103c659023/Detectors/CTP/simulation/src/digi2raw.cxx#L50, by default, it will take the RDH version linked to RAWDataHeader, currently 6, in future : 7.

You can check how I process options for ITS: https://github.com/AliceO2Group/AliceO2/blob/03f1f73549cc06a8bdcfae538a0f1b103c659023/Detectors/ITSMFT/ITS/simulation/src/digi2raw.cxx#L63-L64 , 
https://github.com/AliceO2Group/AliceO2/blob/03f1f73549cc06a8bdcfae538a0f1b103c659023/Detectors/ITSMFT/ITS/simulation/src/digi2raw.cxx#L114-L117 and do the same for CTP.

> 3.) in reconstruction to decide if padd or not I use only RDH . Should I use alsi reserved bit ?

Not sure what do you mean. The [`rdh.dataFormat`](https://github.com/AliceO2Group/AliceO2/blob/03f1f73549cc06a8bdcfae538a0f1b103c659023/DataFormats/Headers/include/Headers/RAWDataHeader.h#L110) should be enough.

> 4.) There are somewhere raw data without padding to test ?

One should ask Pippo.

Detectors/CTP/simulation/src/Digits2Raw.cxx
```diff
@@ -82,7 +84,7 @@ void Digits2Raw::processDigits(const std::string& fileDigitsName)
     std::vector<gbtword80_t> hbfIR;
     std::vector<gbtword80_t> hbfTC;
     for (auto const& ctpdig : CTPDigits) {
-      LOG(debug) << ctpdig.intRecord.bc << " bc all orbit " << ctpdig.intRecord.orbit;
+      LOG(info) << ctpdig.intRecord.bc << " bc all orbit " << ctpdig.intRecord.orbit;
```
shahor02: isn't this too verbose?

Detectors/CTP/simulation/src/Digits2Raw.cxx
```diff
@@ -174,46 +177,66 @@ void Digits2Raw::emptyHBFMethod(const header::RDHAny* rdh, std::vector<char>& to
...
+        toAdd.push_back(c);
+      }
+    }
+  }
+  if (mPadding == false) {
+    uint32_t NZeros = 0;
+    if (countBytes % 16) {
+      NZeros = 16 - (countBytes % 16);
+    }
+    LOG(info) << "Adding bytes:" << NZeros;
```
shahor02: is this supposed to guarantee the alignment to 16 bytes? If so, it will be taken care of by the RawFileWriter, which you already notify about the alignment size: https://github.com/AliceO2Group/AliceO2/blob/1523fb14112c639069ccb5e1157cea0922fee567/Detectors/CTP/simulation/src/Digits2Raw.cxx#L52
Note that the alignment filler is `0xff`.

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
```diff
@@ -119,49 +121,134 @@ void RawDecoderSpec::run(framework::ProcessingContext& ctx)
       size_gbt = 0;
       orbit0 = triggerOrbit;
     }
-    // LOG(info) << "payload size:" << payload.size();
+    // Create 128 bit words
+    gsl::span<const uint8_t> payload(it.data(), it.size());
+    gbtword128_t gbtWord128 = 0;
+    int wordCount = 0;
+    std::vector<gbtword128_t> gbtwords128;
+    LOG(info) << "payload size:" << payload.size();
```
shahor02: I guess all these logs were for the debugging, please reduce the verbosity in the final version

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
```diff
@@ -81,6 +81,12 @@ void RawDecoderSpec::run(framework::ProcessingContext& ctx)
     auto triggerOrbit = o2::raw::RDHUtils::getTriggerOrbit(rdh);
     uint32_t stopBit = o2::raw::RDHUtils::getStop(rdh);
     uint32_t packetCounter = o2::raw::RDHUtils::getPageCounter(rdh);
+    uint32_t version = o2::raw::RDHUtils::getVersion(rdh);
+    if (version == 7) {
+      mPadding = 0;
+    }
```
shahor02: you can simply do `mPadding = o2::raw::RDHUtils::getDataFormat(rdh) == 0`, w/o checking the version (will be done internally).

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
```diff
@@ -119,49 +121,134 @@ void RawDecoderSpec::run(framework::ProcessingContext& ctx)
...
+        std::string str = word.to_string();
+        str = str.substr(48);
+        w80 = std::bitset<80>(str);
+        LOG(info) << "w80:" << w80;
+        gbtwords80.push_back(w80);
+      }
+    } else {
+      int count = 0;
+      for (auto word : gbtwords128) {
+        int countmod = count % 5;
```
shahor02: I don't understand this part, in the case of no-padding, should not you use gbtWord128 as a consecutive chain of 10-byte words (until you hit the alignment filler)?

Detectors/CTP/simulation/src/Digits2Raw.cxx
lietava: I am still debugging, I comment/info->debug before removing draft pr.

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
lietava: 1 ennablePadding - I will follow your example
2. rdh.dataformat - ok.

Detectors/CTP/simulation/src/Digits2Raw.cxx
lietava: I align TF to 128 bits even with no padding. I see your point. I play with it .

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
lietava: yes

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
lietava: > I don't understand this part, in the case of no-padding, should not you use gbtWord128 as a consecutive chain of 10-byte words (until you hit the alignment filler)?
OK, I may see your point.
My approach is 
1 I create 128 bit words in both cases (padding /no padding)
2 if padding I just take lower 80 bits of 128 bits
3 if not padding I am shifting bits of 128 bit words to cut them to 80 bits appropriately  - I do it brutally as 128 x 5 = 8 x 80, i.e. there are only 5 options.

But I think I understand what you wrote above, I think about  it, please, wait.

lietava:
@shahor02 
1 decoding simplified 
2 when trying 
o2-ctp-digi2raw -r 7
with commented lines https://github.com/lietava/AliceO2/blob/733cdff8198d05e613ecda2ef36cfd2d704c5f98/Detectors/CTP/simulation/src/Digits2Raw.cxx#L234
 I got error 
 what():  payload size is not mutiple of GBT word size

shahor02:
> with commented lines https://github.com/lietava/AliceO2/blob/733cdff8198d05e613ecda2ef36cfd2d704c5f98/Detectors/CTP/simulation/src/Digits2Raw.cxx#L234
> I got error
> what(): payload size is not mutiple of GBT word size

By default in case of CRU detector the RawFileWriter expects 128 aligned data and throws if this is not respected. You have to tell him that the data is not 128-bit padded using the `useRDHDataFormat` method, e.g. 
 https://github.com/AliceO2Group/AliceO2/blob/988ca16b95beaaf96eff4d8203032a67a6ec9ab0/Detectors/ITSMFT/ITS/simulation/src/digi2raw.cxx#L162

lietava:
@shahor02 : I implemented all your suggestions. Remaining: remove debug messages and test on data from Ivan.

lietava:
@shahor02 : I checked on MC , compared dev versus ctpdev on real data . I am removing debug messages.

lietava:
@shahor02 : please, merge

shahor02 (approved):

PR #10808, opened 2023-02-20
ATO-609: switch on track references for the ITS layers and every 15 c…

marslandALICE:
ATO-609: switch on track references for the ITS layers and every 15 cm for TPC volume


Detectors/TPC/simulation/src/Detector.cxx
```diff
@@ -104,6 +104,8 @@ Bool_t Detector::ProcessHits(FairVolume* vol)
 {
   mStepCounter++;
   auto& gasParam = ParameterGas::Instance();
+  const Int_t   kMaxDistRef =15;     // maximal difference between 2 stored references - the parameter should be 15-30 cm default
+  static Double_t lastReferenceR=0;  /// using static global varaible - in Run1,2 it was OK - in Run3 use some other mechanism
```
sawenzel: it would be better to use a class member variable for this. It also keeps state and has no threading downsides.

mconcas (approved):
Fine for the ITS part

sawenzel (approved):

wiechula (approved):
Fine for TPC

PR #10809, opened 2023-02-21
ITS: Use a single TimeFrame for reco instance

mconcas:
We get TimeFrame object/ptr only once from GPUChain. Then we reset it at each TF processing.
To be carefully evaluated, I tested it locally, it works.
I will do some more tests (e.g. w/w.out meanvertex override).

cc-ing @mpuccio for review.

Edit: as a second minor change, I removed some intermediate translation of the found vertices into lightVertex ones. 
It will save some resources.

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/include/ITStracking/Vertexer.h
+++ b/Detectors/ITSMFT/ITS/tracking/include/ITStracking/Vertexer.h
@@ -89,11 +89,6 @@ class Vertexer
   TimeFrame* mTimeFrame = nullptr;   /// Observer pointer, not owned by this class
 };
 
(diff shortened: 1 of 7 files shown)


mconcas:
I did some more tests. It also remains stable for multiple CTFs. 
With and without meanvertex overriding.
If you agree, this can be merged @iouribelikov @mpuccio

mpuccio (dismissed):

iouribelikov (dismissed):

mconcas:
Hi, @shahor02: can we merge this? I re-triggered the CI yersterday but it randomly failed again. Thanks

shahor02 (approved):

PR #10810, opened 2023-02-21
getRunduration: more flexible error handling

sawenzel:
so far solely fatal, now there is the possibility
to switch this off.


sawenzel:
@mfasDa

PR #10814, opened 2023-02-21
Add lumi and time info to unbinned residuals

martenole:


martenole:
Error unrelated, merging. (The change of the output size for the unbinned residuals tree is negligible)

PR #10817, opened 2023-02-22
ITS: install AlpideResponseData.root with explicit rule

mconcas:
@davidrohr : can this work? (Asking FullCI, actually)

### DIFF
--- a/Detectors/ITSMFT/common/data/AlpideResponseData/CMakeLists.txt
+++ b/Detectors/ITSMFT/common/data/AlpideResponseData/CMakeLists.txt
@@ -27,8 +27,8 @@ else()
 endif()
 
 add_custom_command(TARGET O2exe-alpide-response-generator POST_BUILD
-                   COMMAND ${CMAKE_BINARY_DIR}/stage/bin/o2-alpide-response-generator -i ${ITSRESPONSE_DIR}/response/AlpideResponseData/ -o ${CMAKE_SOURCE_DIR}/Detectors/ITSMFT/common/data/AlpideResponseData/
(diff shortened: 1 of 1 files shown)


Detectors/ITSMFT/common/data/AlpideResponseData/CMakeLists.txt
```diff
@@ -27,8 +27,8 @@ else()
 endif()
 
 add_custom_command(TARGET O2exe-alpide-response-generator POST_BUILD
-                   COMMAND ${CMAKE_BINARY_DIR}/stage/bin/o2-alpide-response-generator -i ${ITSRESPONSE_DIR}/response/AlpideResponseData/ -o ${CMAKE_SOURCE_DIR}/Detectors/ITSMFT/common/data/AlpideResponseData/
+                   COMMAND ${CMAKE_BINARY_DIR}/stage/bin/o2-alpide-response-generator -i ${ITSRESPONSE_DIR}/response/AlpideResponseData/ -o ${CMAKE_CURRENT_BINARY_DIR}/
                    DEPENDS ${CMAKE_SOURCE_DIR}/Detectors/ITSMFT/common/data/AlpideResponseData/AlpideResponse.cxx
 )
 
-o2_data_file(COPY ${CMAKE_SOURCE_DIR}/Detectors/ITSMFT/common/data/AlpideResponseData DESTINATION Detectors/ITSMFT/data)
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/AlpideResponseData.root" DESTINATION "${CMAKE_INSTALL_PREFIX}/share/Detectors/ITSMFT/data/AlpideResponseData/")
```
davidrohr, line 34: I think this is fine.
In principle you could also stick to using `o2_data_file`, which will do the install.
But you should state the file explicitly, as you do in the PR, and not refer only to the folder.

PR #10820, opened 2023-02-22
EMCAL fix segfault in raw decoder

martenole:
Hi @mfasDa and @shahor02 
Unfortunately, I don't fully understand the reason for the crash observed on Monday. But with this patch the decoding of the raw data for run 531419 goes through and valgrind does not show any invalid read/write anymore. The invalid reads were pointing to this lambda, but I don't know why it makes a problem now, since the decoder was not changed for some time. Is it possible that this appears now that we have switched to gcc 12?
I also tried changing `DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Cell.h` to make sure it is move assignable and move constructible as required by std::sort https://en.cppreference.com/w/cpp/algorithm/sort, but that did not help. Only removing the references worked for me.
Maybe someone has a better understanding of what is going on?
Cheers,
Ole

### DIFF
--- a/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
+++ b/Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
@@ -548,7 +548,7 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
     if (cells->size()) {
       LOG(debug) << "Event has " << cells->size() << " cells";
       // Sort cells according to cell ID
-      std::sort(cells->begin(), cells->end(), [](const RecCellInfo& lhs, const RecCellInfo& rhs) { return lhs.mCellData.getTower() < rhs.mCellData.getTower(); });
+      std::sort(cells->begin(), cells->end(), [](const RecCellInfo lhs, const RecCellInfo rhs) { return lhs.mCellData.getTower() < rhs.mCellData.getTower(); });
       for (const auto& cell : *cells) {
         if (cell.mIsLGnoHG) {
           // Treat error only in case the LG is above the noise threshold


mfasDa:
@martenole I would also not understand the issue, references in the predicate for the std::sort should be fine. The only thing I could imagine is that there would be a bug in gcc-12. 

I am a bit afraid the copy-by-value could make the code a bit slow.

martenole:
Maybe this also just masks the underlying issue instead of fixing it.

When running with valgrind the error does not appear, but at least you get some error messages (Invalid read/write). Unfortunately, these messages don't point to the lines in the code, just to the library in general.

```
[6530:EMCALRawToCellConverterSpec]: ==6530==    at 0x495FDF7: void std::__unguarded_linear_insert<__gnu_cxx::__normal_iterator<o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo*, std::vector<o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo, std::allocator<o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo> > >, __gnu_cxx::__ops::_Val_comp_iter<o2::emcal::reco_workflow::RawToCellConverterSpec::run(o2::framework::ProcessingContext&)::{lambda(o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo const&, o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo const&)#2}> >(__gnu_cxx::__normal_iterator<o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo*, std::vector<o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo, std::allocator<o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo> > >, __gnu_cxx::__ops::_Val_comp_iter<o2::emcal::reco_workflow::RawToCellConverterSpec::run(o2::framework::ProcessingContext&)::{lambda(o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo const&, o2::emcal::reco_workflow::RawToCellConverterSpec::RecCellInfo const&)#2}>) [clone .constprop.0] (in /home/oschmidt/alice/sw/ubuntu2004_x86-64/O2/residuals-local4/lib/libO2EMCALWorkflow.so)
```

So I did a debug O2 build and there the issue did not appear at all, also without using valgrind.

So I am a bit lost what else to do.

shahor02:
I have also noticed that with debug build the invalid reads are not reported. And, to increase the entropy: they are not reported also with the last year raw data, regardless on the build type...

davidrohr:
I would not remove the references. First it will slow down the processing to some extent, but also we should understand why it failed. This would only hide some underlying problem.

PR #10828, opened 2023-02-23
TRD: QC: correct calculation of tracklet length

f3sch:
Fixes wrong calculation of tracklet length in PR #10792 thanks to Gauthier Legras.
We do not care about exact value but only about the variations.

### DIFF
--- a/Detectors/TRD/qc/src/Tracking.cxx
+++ b/Detectors/TRD/qc/src/Tracking.cxx
@@ -133,8 +133,13 @@ void Tracking::checkTrack(const TrackTRD& trkTrd, bool isTPCTRD)
       tracklet.getQ2(),
     };
 
-    // Corrected Tracklets
-    auto cor = mLocalGain.getValue(tracklet.getHCID() / 2, tracklet.getPadCol(), tracklet.getPadRow()) * (std::abs(tracklet.getSlopeBinSigned()) + 1.f);
+    /// Corrected Tracklets
+    // To correct for longer drift lengths in the drfit chamber and hence more energy deposition,
+    // we calculate the variation to reference for the tracklets (exrapolated from the track fit).
+    // \sqrt{ (dx/dx)^2 + (dy/dx)^2 + (dz/dx)^2}
+    auto tphi = trk.getSnp() / std::sqrt((1.f - trk.getSnp()) * (1.f + trk.getSnp()));
+    auto trackletLength = std::sqrt(1.f + tphi * tphi + trk.getTgl() * trk.getTgl());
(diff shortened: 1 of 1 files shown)


Detectors/TRD/qc/src/Tracking.cxx
```diff
@@ -134,7 +134,9 @@ void Tracking::checkTrack(const TrackTRD& trkTrd, bool isTPCTRD)
     };
 
     // Corrected Tracklets
-    auto cor = mLocalGain.getValue(tracklet.getHCID() / 2, tracklet.getPadCol(), tracklet.getPadRow()) * (std::abs(tracklet.getSlopeBinSigned()) + 1.f);
+    auto tphi = std::tan(std::asin(tracklet.getSnp()));
```
martenole:
```suggestion
    auto tphi = tracklet.getSnp() / std::sqrt((1. - tracklet.getSnp()) * (1. + tracklet.getSnp()));
```
To avoid expensive std::asin operation. Also, I think you need to check the track instead of the tracklet

Detectors/TRD/qc/src/Tracking.cxx
```diff
@@ -134,7 +134,9 @@ void Tracking::checkTrack(const TrackTRD& trkTrd, bool isTPCTRD)
...
+    auto trackletLength = std::sqrt(1.f + tphi * tphi + tracklet.getTgl() * tracklet.getTgl());
```
martenole: Can you please add a comment here what is for you the tracklet length? There is no `getTgl()` for tracklets I think. In general I am not sure how meaningful it would be to include the tracklet direction as a correction factor for the charge.. The track direction should be better

Detectors/TRD/qc/src/Tracking.cxx
```diff
@@ -145,7 +147,9 @@ void Tracking::checkTrack(const TrackTRD& trkTrd, bool isTPCTRD)
         }
         if (std::abs(tracklet.getPadCol() - trklt.getPadCol()) == 1 && std::abs(tracklet.getPadRow() - trklt.getPadRow()) == 0) {
           // Add charge information
-          auto cor = mLocalGain.getValue(trklt.getHCID() / 2, trklt.getPadCol(), trklt.getPadRow()) * (std::abs(trklt.getSlopeBinSigned()) + 1.f);
+          auto tphi = std::tan(std::asin(trklt.getSnp()));
```
martenole: same as above

martenole (changes requested):

davidrohr:
@f3sch : Could you please address the comments of Ole?

f3sch:
@davidrohr Did I not in 3e7ee63?

davidrohr:
sorry, didn't see that...

PR #10830, opened 2023-02-23
New callback registry

ktf:


ktf:
@AliceO2Group/muon-experts I think the end of stream callback silently ignored so far. Could you confirm this was the case?

PR #10832, opened 2023-02-24
Catch exception from stol in CCDBManager

sawenzel:


sawenzel:
@shahor02 : I have seen crashes with exceptions originating from this line ... but they are only caught by some upper framework routine. Probably better to catch here and report.

PR #10833, opened 2023-02-24
[MRRTF-211] Migrate MCH track I/O to IO and DevIO libs

aphecetche:
The `tracks-reader-workflow` also gains a new `--subspec` option to allow reading two different `mchtracks.root` files. Can be used e.g. in a comparison workflow.


pillot (commented):
- The track reader/writer/dumper workflows are missing the IO/CMakeLists.txt
- I think tracks-dumper-workflow.cxx is still WIP. It has unnecessary includes, unused dump() function and missing or unused options.
- I am not sure if TrackTreeReader is supposed to be ported or replaced by the dumper, but TrackTreeReader.cxx is ported and TrackTreeReader.h is not, and it seems nothing uses it anymore.

aphecetche:
@pillot thanks for the review.

> * The track reader/writer/dumper workflows are missing the IO/CMakeLists.txt

Indeed 😞 . Fixed.

> * I think tracks-dumper-workflow.cxx is still WIP. It has unnecessary includes, unused dump() function and missing or unused options.

I've changed it a bit to make it less WIP.

> * I am not sure if TrackTreeReader is supposed to be ported or replaced by the dumper, but TrackTreeReader.cxx is ported and TrackTreeReader.h is not, and it seems nothing uses it anymore.

TrackTreeReader is indeed no longer needed now that the tracks-file-dumper as been removed.

Detectors/MUON/MCH/IO/src/tracks-dumper-workflow.cxx
```diff
@@ -9,24 +9,40 @@
...
+#include <fmt/format.h>
+#include <gsl/span>
+#include <iostream>
+#include <vector>
+
+void dump(std::ostream& os, const o2::mch::TrackMCH& t)
+{
+  auto pt = std::sqrt(t.getPx() * t.getPx() + t.getPy() * t.getPy());
+  os << fmt::format("({:s}) p {:7.2f} pt {:7.2f} nclusters: {} \n", t.getSign() == -1 ? "-" : "+", t.getP(), pt, t.getNClusters());
+}
```
pillot, line 27: The dump function and most of the includes are not needed here.
Also I don't think the HBFUtils stuffs are needed here as the dumper is more like a writer than a reader.
But these are details that can be fixed (or not) later so I approve the PR anyway.

pillot (approved):

Detectors/MUON/MCH/IO/src/tracks-dumper-workflow.cxx
aphecetche: thanks.

aphecetche:
@shahor02 could you merge this one please ? Once more it's mostly about moving things around in MCH libs, but with a change in `GlobalTrackingWorkflow` `CMakeLists.txt`

shahor02:
@aphecetche the fullCI is currently broken due to the codechecker.  Could you run the FST locally if you have not done so?

PR #10834, opened 2023-02-24
Fix O2 build using arrow v11

TimoWilken:
This fixes https://github.com/alisw/alidist/pull/4797, since in Arrow v11, Gandiva's CMake files are now installed in a separate directory from the Arrow ones.

### DIFF
--- a/dependencies/O2Dependencies.cmake
+++ b/dependencies/O2Dependencies.cmake
@@ -34,6 +34,14 @@ find_package(Gandiva CONFIG)
 set_package_properties(Arrow PROPERTIES TYPE REQUIRED)
 set_package_properties(Gandiva PROPERTIES TYPE REQUIRED)
 
+if (NOT TARGET Arrow::arrow_shared)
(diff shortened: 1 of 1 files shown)


TimoWilken:
Unfortunately it seems that this PR doesn't work without the alidist PR https://github.com/alisw/alidist/pull/4797. I'm not sure what the best strategy is for getting both tested and merged together.

aphecetche:
I guess it’s not too much work to make the gandiva find_package line depends on Arrow_VERSION_MAJOR and thus decorrelate the two PRs ?

ktf:
@davidrohr and myself came up with: https://github.com/AliceO2Group/AliceO2/pull/10834/commits/d20e1936c5aba2846c08b6b3d853d1dfc41b9f67. Why wouldn't that work?

ktf:
Nevermind, we see the problem. fix coming.

davidrohr:
errors are no only due to unrelated problems, merging

PR #10843, opened 2023-02-27
Improve headings level scheme

pbuehler:
Changed internal links


pbuehler:
Hi @vkucera, this is the PR which will adapt the heading levels in the 'data model reference'

vkucera:
> Hi @vkucera, this is the PR which will adapt the heading levels in the 'data model reference'

Thanks @pbuehler

PR #10845, opened 2023-02-27
[EMCAL-904] Port correction for energy nonlinearity

mfasDa:
Correction for energy nonlinearity applied at cluster
level, ported from AliEMCALRecoUtils in AliPhysics


mfasDa:
@FriederikeBock @nschmidtALICE

PR #10850, opened 2023-02-28
TPC: add abs v drift to laser qc

tklemenz:
Adds abs v drift to the laser tracks qc output.


tklemenz:
@wiechula, I split this from the other PR (https://github.com/AliceO2Group/AliceO2/pull/10658), so this can be merged already.
This is ready to be merged, I think.

PR #10851, opened 2023-02-28
ITS-GPU: remove synchronisation point

mconcas:
- Remove synchronisation at the end of chunk processing
- Restore vertex saving

In previous approach we had the three threads to be joined after they terminated to process a chunk.
Usually we have more than one chunk to process per TF, this then caused all threads to be as slow as 
the slowest one.

Now the synchronisation happens outside the re-scheduling of chunks data, hence it happen only once


mconcas:
@davidrohr we can go with this, thanks.

PR #10852, opened 2023-02-28
TRD: PID: Added LQND, PyTorch policy support and charge correction

f3sch:


Detectors/TRD/pid/include/TRDPID/Dummy.h
```diff
@@ -34,13 +34,13 @@ class Dummy final : public PIDBase
...
  public:
-  ~Dummy() final = default;
+  ~Dummy() = default;
 
   /// Do absolutely nothing.
   void init(o2::framework::ProcessingContext& pc) final{};
 
   /// Everything below 0.f indicates nothing available.
-  PIDValue process(const TrackTRD& trk, const o2::globaltracking::RecoContainer& input, bool isTPC) final
+  [[nodiscard]] float process(const TrackTRD& trk, const o2::globaltracking::RecoContainer& input, bool isTPCTRD) const final
```
martenole: Why the `[[nodiscard]]` here and in the other places below? I have never seen it used in O2 and I don't know if it is of value here. Could you please elaborate? I don't think it improves the readability of the code

Detectors/TRD/pid/src/TRDPIDLinkDef.h
```diff
@@ -20,5 +20,6 @@
 #pragma link C++ class o2::trd::Dummy + ;
 #pragma link C++ class o2::trd::TRDPIDParams + ;
 #pragma link C++ class o2::conf::ConfigurableParamHelper < o2::trd::TRDPIDParams> + ;
+#pragma link C++ class std::vector < TGraph> + ;
```
martenole, line 27: is this really needed?

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -120,8 +120,12 @@ void TRDGlobalTracking::updateTimeDependentParams(ProcessingContext& pc)
 
     /// Get the PID model if requested
     if (mWithPID) {
-      mBase = getTRDPIDBase(mPolicy);
+      mBase = getTRDPIDPolicy(mPolicy);
       mBase->init(pc);
+
+      // calibration
+      auto localGain = *(pc.inputs().get<o2::trd::LocalGainFactor*>("localgainfactors"));
```
martenole: why creating a copy here and not just pass the pointer of the CCDB object?

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -835,21 +839,32 @@ DataProcessorSpec getTRDGlobalTrackingSpec(bool useMC, GTrackID::mask_t src, boo
 
   // Request PID policy data
   if (withPID) {
+    // request policy
     switch (policy) {
       case PIDPolicy::LQ1D:
-        // inputs.emplace_back("LQ1D", "TRD", "LQ1D", 0, Lifetime::Condition, ccdbParamSpec("TRD/ppPID/LQ1D"));
+        inputs.emplace_back("lq1dlut", "TRD", "LQ1D", 0, Lifetime::Condition, ccdbParamSpec("TRD_test/PID/LQ1D"));
```
martenole, line 842: since PID is disabled by default it should be OK to request the object from the test CCDB for now

DataFormats/Detectors/TRD/include/DataFormatsTRD/PID.h
```diff
@@ -30,44 +31,69 @@ enum class PIDPolicy : unsigned int {
...
+{
+  std::string name;
+  switch (policy) {
+    case PIDPolicy::LQ1D:
+      name = "LQ1D";
+      break;
+    case PIDPolicy::LQ3D:
+      name = "LQ3D";
+      break;
+#ifdef TRDPID_WITH_ONNX
```
martenole, line 56: In case O2 is compiled without ONNX and then you try to put the policy `XGB` for example then without an error the default dummy option would be taken instead, right? I think it would be better to have a dummy policy, but to throw an error by default if the selected policy does not exist

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -835,21 +839,32 @@ DataProcessorSpec getTRDGlobalTrackingSpec(bool useMC, GTrackID::mask_t src, boo
...
         break;
       case PIDPolicy::LQ3D:
-        // inputs.emplace_back("LQ3D", "TRD", "LQ3D", 0, Lifetime::Condition, ccdbParamSpec("TRD/ppPID/LQ3D"));
+        inputs.emplace_back("lq3dlut", "TRD", "LQ3D", 0, Lifetime::Condition, ccdbParamSpec("TRD_test/PID/LQ3D"));
         break;
+#ifdef TRDPID_WITH_ONNX
```
martenole, line 847: would there be any harm here in case one of these policies is loaded without ONNX present? Probably not, right, and then you would not need this `#ifdef` here

martenole (commented):
I did not go through the models themselves in details, just added some general comments

Detectors/TRD/pid/src/TRDPIDLinkDef.h
f3sch: Yes, root only provides native support for st::array<TGraph> not for this.

Detectors/TRD/pid/include/TRDPID/Dummy.h
f3sch: I think it makes the code more readable and says that this value should not be ignored, but I will delete it anyways.

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
f3sch: Does the ccdb object stay in memory during execution. I was unsure about this.

martenole: yes, the DPL CCDB fetcher takes care of it

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
f3sch: Since I mask them out in the enum with the #ifdef, I also have to put it here. Otherwise the case would not be there

f3sch:
I convert this back into a draft, wait until I can make the LUTs and final models and upload them to the alice-ccdb. Then I test and convert back.

PR #10853, opened 2023-02-28
TPC tracking: Add streamers for update track and reject clusters

matthias-kleiner:


GPU/GPUTracking/Merger/GPUTPCGMTrackParam.h
```diff
@@ -224,12 +225,13 @@ class GPUTPCGMTrackParam
...
-  float mC[15];    // the covariance matrix for Y,Z,SinPhi,..
-  float mChi2;     // the chi^2 value
-  int mNDF;        // the Number of Degrees of Freedom
+  float mX;                           // x position
+  float mTZOffset;                    // Z offset with early transform, T offset otherwise
+  float mP[5];                        // 'active' track parameters: Y, Z, SinPhi, DzDs, q/Pt
+  float mC[15];                       // the covariance matrix for Y,Z,SinPhi,..
+  float mChi2;                        // the chi^2 value
+  int mNDF;                           // the Number of Degrees of Freedom
+  o2::utils::DebugStreamer mStreamer; ///<! streamer for writing out track update
```
davidrohr: Instead of add a DebugStreamer data member everywhere, can't we do something like
`static o2::utils::DebugStreamer streamer("debug_InterpolateReject", "UPDATE")`
in the place where we use them? In that way, we don't need to include the DebugStreamer header in all the header data files? This could be done in the way I do it for my GPUROOTDump, which  has very similar functionality, e.g. https://github.com/davidrohr/AliceO2/blob/cf859182f75967b35c46bc643ef44222ddd284c5/GPU/GPUTracking/qa/GPUQA.cxx#L1716

GPU/GPUTracking/Merger/GPUTPCGMTrackParam.cxx
```diff
@@ -308,6 +308,20 @@ GPUd() bool GPUTPCGMTrackParam::Fit(GPUTPCGMMerger* GPUrestrict() merger, int iT
         }
 #endif
         retVal = prop.Update(yy, zz, clusters[ihit].row, param, clusterState, rejectChi2, &interpolation.hit[ihit], refit);
+
+        if (o2::utils::DebugStreamer::checkStream(o2::utils::StreamFlags::streamUpdateTrack)) {
+          mStreamer.setStreamer("debug_update_track", "UPDATE");
+          mStreamer.getStreamer() << mStreamer.getUniqueTreeName("tree_update_track").data()
```
davidrohr: Can we remove the newlines and put alll the `<<` in a single line, to reduce the number of lines used by the debug code?

GPU/GPUTracking/Merger/GPUTPCGMPropagator.cxx
```diff
@@ -651,6 +651,23 @@ GPUd() int GPUTPCGMPropagator::Update(float posY, float posZ, int iRow, const GP
       rejectChi2 = 1;
     } else {
       int retVal = InterpolateReject(posY, posZ, clusterState, rejectChi2, inter, err2Y, err2Z);
+
+      if (o2::utils::DebugStreamer::checkStream(o2::utils::StreamFlags::streamRejectCluster)) {
+        mStreamer.setStreamer("debug_InterpolateReject", "UPDATE");
+        mStreamer.getStreamer() << mStreamer.getUniqueTreeName("tree_InterpolateReject").data()
```
davidrohr: I see several times a setStreamer and then getStreamer call. Can we unify this in a single function taking all arguments. Would reduce one line of code. Sorry for being a bit picky. I really think the functionality is OK, but I am just trying to write it in an as short as possible form, to keep the tracking code as compact as possible. In any case it is pretty long already ... :(

GPU/GPUTracking/Merger/GPUTPCGMTrackParam.h
matthias-kleiner: Thanks, I will check this. Would be indeed better if this works...

GPU/GPUTracking/Merger/GPUTPCGMPropagator.cxx
matthias-kleiner: No problem, thanks for your suggestions :) I will take a look and see if it possible

GPU/GPUTracking/Merger/GPUTPCGMTrackParam.h
miranov25: It is not possible in this way. We use **local variables** that you cannot access in the header. **The ability to stream local variables is the main advantage of the streaming approach.**
This is essential for understanding the algorithm and the data itself.

davidrohr: But why would you need to access something in the header? My point was to just move the mStreamer data member of the class to a local static variable. Since you anyway call setStreamer before you do the actual streaming, I don't see why you need the mStreamer as a data member.

miranov25:
Relates to -  https://alice.its.cern.ch/jira/browse/PWGPP-730 - PWGPP-730

-  https://alice.its.cern.ch/jira/browse/PWGPP-730?focusedCommentId=297569&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-297569

GPU/GPUTracking/Merger/GPUTPCGMTrackParam.h
miranov25: If you can make guarantee the same logging functionality (including local variables) then it is OK for me. In my previous respond, I assumed in your solution it will be not possible.

matthias-kleiner:
@davidrohr I implemented your suggestions. It is much cleaner and also faster now. Is there anything else I should change?

matthias-kleiner:
Now I get errors when compiling for GPU
`/scratch/mkleiner/alice/sw/SOURCES/O2/debug_streamer/0/GPU/GPUTracking/dEdx/GPUdEdx.h:185:97: error: const_cast from 'const __private int *' to '__generic int *' is not allowed
                                                                              << "region=" << (*const_cast<int*>(&region))`

@davidrohr Do you know if there is something like `const_cast<>` for the GPU?

davidrohr:
> Now I get errors when compiling for GPU `/scratch/mkleiner/alice/sw/SOURCES/O2/debug_streamer/0/GPU/GPUTracking/dEdx/GPUdEdx.h:185:97: error: const_cast from 'const __private int *' to '__generic int *' is not allowed << "region=" << (*const_cast<int*>(&region))`
> 
> @davidrohr Do you know if there is something like `const_cast<>` for the GPU?

ok, but this can be solved. Don't have time tonight, but will try to check tomorrow.

miranov25:
I checked the content of the streamer and we have missing information - sector, respectively track rotation.
Is it possible to add that information in?
```
  KEY: TTree	tree_InterpolateReject_0;1	tree_InterpolateReject_0
======> EVENT:10
 posY            = -6.3375
 posZ            = 243.69
 clusterState    = 0
 rejectChi2      = 2
 interpolationhit = (o2::gpu::gputpcgmmergertypes::InterpolationErrorHit*)0xb54a580
 posY            = -6.463365
 posZ            = 244.102890
 errorY          = 0.001180
 errorZ          = 0.016472
 refit           = 1
 retVal          = 0
 err2Y           = 0.000963301
 err2Z           = 0.0235767
 track           = (o2::gpu::GPUTPCGMTrackParam*)0xb9c52f0
 mX              = 94.224998
 mTZOffset       = 42076.765625
 mP[5]           = -6.463365 , 244.102890 , 0.051068 , -0.893345 , 3.035392 

 mC[15]          = 0.001180 , 0.000000 , 0.016472 , 0.000433 , 0.000000 , 
                    0.000216 , 0.000000 , 0.002980 , 0.000000 , 0.000655 , 
                    0.049782 , 0.000000 , 0.029686 , 0.000000 , 4.745279 

 mChi2           = 37.754051
 mNDF            = 15
```

miranov25:
We were looking for the direction of propagation (tree_InterpolateReject_0).  In the output, we see that the fit is done per track, propagating out and later in. Is there a way to get a direction? We can guess in our analysis code, but I prefer to use it if available as a variable.  

image

miranov25:
For the downsampling, we were considering using the track pt. But the track pt is not well defined, as it scales with a lever arm, which is at the beginning small.
Do we have track parameters with a full lever arm? I do not think so. If not we will need the global ID of the track, so we can make downsampling later, knowing full track properties.

### Question for David

Is the track ID already defined at that moment, and could we access it?


image

davidrohr:
I am not sure what you mean by track ID. Do you mean the index in the final track array? This is only defined at a later stage, and not known during seeding and also not during refit.

Track parameters at outermost point are stored as OuterParam after the outward fit.

There is the `iWay`, which is 0 for first inward fit, 1 for outward fit, 2 for inward refit.

we don't have the sector as variable, but you can get the current angle of the track via `prop.GetAlpha()`

miranov25:
Yes. I would like to get the global index of the last track so that we can use it in the analysis.
As far as I know (based on your explanation), it does not exist at this point, the track will be registered later. We can find a way to assign the index beforehand. If this is not possible in the tracking code, I will instruct Matthias to assign us a "unique" number in the streamer himself (creating a unique ID by the first point of the track)

davidrohr:
Hi Marian, this is not possible by design. There is a filter after the refit rejecting tracks, and the reduce operation that creates a linear array of output tracks runs only after all tracks have been refitted, and also that reduce operation is based on atomic_max, so it is not reproducible.

However, the variable `iTrk` is unique inside the track fit. Not sure if that is enough. But it does not relate to the final track index in the tpctracks.root.

miranov25:
Hello @davidrohr 

This is good. In the streamer, we can use iTrk, which you pointed out. I understand that some tracks are rejected based on certain criteria.
That would indeed be a good place for another streamer in the decision algorithm. We can get the relationship iTrk- > track global ID (maybe) if it is accepted, and we get undefined - if it is rejected, so we can test/analyze this track rejection/acceptance algorithm as well.

matthias-kleiner:
> Hi Marian, this is not possible by design. There is a filter after the refit rejecting tracks, and the reduce operation that creates a linear array of output tracks runs only after all tracks have been refitted, and also that reduce operation is based on atomic_max, so it is not reproducible.
> 
> However, the variable `iTrk` is unique inside the track fit. Not sure if that is enough. But it does not relate to the final track index in the tpctracks.root.

Hi @davidrohr ,
thanks, I added the `iTrk` for one debug streamer. Is this variable or a similar variable also available for the other streamer at https://github.com/AliceO2Group/AliceO2/blob/9414f0166abd3c830dd44ce4059cba20d8130100/GPU/GPUTracking/Merger/GPUTPCGMPropagator.cxx#L656?

GPU/GPUTracking/Merger/GPUTPCGMPropagator.cxx
```diff
@@ -651,6 +652,9 @@ GPUd() int GPUTPCGMPropagator::Update(float posY, float posZ, int iRow, const GP
       rejectChi2 = 1;
     } else {
       int retVal = InterpolateReject(posY, posZ, clusterState, rejectChi2, inter, err2Y, err2Z);
+      if (o2::utils::DebugStreamer::checkStream(o2::utils::StreamFlags::streamRejectCluster)) {
+        o2::utils::DebugStreamer::instance()->getStreamer("debug_InterpolateReject", "UPDATE") << o2::utils::DebugStreamer::instance()->getUniqueTreeName("tree_InterpolateReject").data() << "mAlpha=" << mAlpha << "iRow=" << iRow << "posY=" << posY << "posZ=" << posZ << "clusterState=" << clusterState << "rejectChi2=" << rejectChi2 << "interpolationhit=" << *inter << "refit=" << refit << "retVal=" << retVal << "err2Y=" << err2Y << "err2Z=" << err2Z << "track=" << mT << "\n";
```
miranov25, line 656: Hello @Mathias.
here I would like to keep also return value. it is not in the current streamer

miranov25: Indeed in session we found it is already there

davidrohr: You can use `(size_t)mT` as unique track ID

GPU/GPUTracking/dEdx/GPUdEdx.h
```diff
@@ -177,45 +176,32 @@ GPUdnii() void GPUdEdx::fillCluster(float qtot, float qmax, int padRow, unsigned
...
-                            << "threshold=" << thresholdTmp
-                            << "qTotIn=" << qTotIn
-                            << "qMaxTopologyCorr=" << qMaxTopologyCorrTmp
-                            << "qTotTopologyCorr=" << qTotTopologyCorrTmp
-                            << "qMaxResidualCorr=" << qMaxResidualCorrTmp
-                            << "qTotResidualCorr=" << qTotResidualCorrTmp
-                            << "residualGainMapGain=" << residualGainMapGainTmp
-                            << "fullGainMapGain=" << fullGainMapGainTmp
-                            << "\n";
+    o2::utils::DebugStreamer::instance()->getStreamer("debug_dedx", "UPDATE") << o2::utils::DebugStreamer::instance()->getUniqueTreeName("tree_dedx").data()
```
davidrohr, line 182: You can use `(size_t)this` as track ID here. It will be the same as in the propagator, so you can correlate the different streamer.

davidrohr (commented):
Track rejection happens here: https://github.com/AliceO2Group/AliceO2/blob/1c19c23b09bbc6a0d0e13f47e528a738b5c32793/GPU/GPUTracking/Merger/GPUTPCGMTrackParam.cxx#L375
You can also use `(size_t)this` as track ID here.

GPU/GPUTracking/Merger/GPUTPCGMPropagator.cxx
matthias-kleiner: I am just looking at `(size_t)mT` and it seems to be the same for every track. There are only differences if I compare the `(size_t)mT` for different CPU threads. Are you sure it can be used as unique track ID?

davidrohr:
Hi @matthias-kleiner : Indeed you are right. What I said before was not correct, sorry. We have moved that to a local variable as a GPU optimization in order not to use global memory.
Please use the iTrk for now where available.
I'll add some DEBUG defines that will allow propagation of the iTrk inside the propagators only for debug builds.

miranov25:
Hello @davidrohr  and @matthias-kleiner 

I am already very happy with the new streamers. They have helped us a lot in understanding the calibration and reconstruction. It would be good if we could also get a map of the track indices for the current - > final track.

If this can be done quickly, I suggest we do it. If not, can we merge the code as is and add the indices in the next pull request?

miranov25:
@matthias-kleiner , are the latest changes that you used to produce the latest streamer output the already in the github?

davidrohr:
Hi @miranov25 : We cannot merge right now, since it breaks the GPU build.
I have to fix that before.

So you mean you want a map of the iTrk used during the tracking to the track index in tpctracks.root?

matthias-kleiner:
> Hi @miranov25 : We cannot merge right now, since it breaks the GPU build. I have to fix that before.

By breaking you mean the problem with the `const_cast` or is there something else? I just fixed this problem

> So you mean you want a map of the iTrk used during the tracking to the track index in tpctracks.root?

Yes, I think this would be good

matthias-kleiner:
> @matthias-kleiner , are the latest changes that you used to produce the latest streamer output the already in the github?

No, I will commit the code, but I also agree that it would be good to merge this...

matthias-kleiner:
@davidrohr It seems that all the tests have passed except one, which seems to be unrelated. Is there anything else I should change in this PR?

davidrohr:
looks good to me. So you have solved the cast problem for OpenCL yourself? Since I don't see the FullCI failing any more?

matthias-kleiner:
> looks good to me. So you have solved the cast problem for OpenCL yourself? Since I don't see the FullCI failing any more?

Yes, I just added a const_cast function in the debugstreamer class https://github.com/AliceO2Group/AliceO2/blob/23bdd97f8b3ca377531d8220a2173953207a35ec/Common/Utils/include/CommonUtils/DebugStreamer.h#L160 which is just a dummy for the GPU

davidrohr:
ok, I see. But you are casting away the const property. Can't this be avoided?

matthias-kleiner:
> ok, I see. But you are casting away the const property. Can't this be avoided?

Yes it is casted away, but this cannot be avoided. The objects are not changed themselves, so it is no problem. It is somewhere in ROOT where a function is not `const` I think. I could also try to return a non const copy of the object, but I think this wont work, since it is then only a temporary object...

davidrohr:
hm, ok, since it is only for debug code, which is normally not compiled, I don't really mind. Still, would be better to avoid const casts. Usually indicates a broken design in my opinion.

matthias-kleiner:
> hm, ok, since it is only for debug code, which is normally not compiled, I don't really mind. Still, would be better to avoid const casts. Usually indicates a broken design in my opinion.

I also dont like it. Maybe I can try to find a solution with @miranov25 as he was using the `TreeStreamRedirector` a lot in the past

miranov25:
@matthias-kleiner, @davidrohr 

I do not like it either. When I wrote streamer classes ~2002, I did not have many options:
* We can ask the root team if they can make a tree filling with a const pointer. Since they have to set the branch address to this object, they can't make it constant easily. So I did not even ask them.
* We can make a copy of the object, but that will be slow and besides, not all classes have implemented a proper copy constructor. So I already had some problems with this approach

shahor02:
Hi,

Attaching an address to a root tree branch cannot go via const pointer since the idea is that the root creates itself an object and attaches it to passed pointer if the latter is nullptr. I think the const_cast is the cheapest solution. 
But I would consider to declare all `TreeStream& operator<<(...)` methods to `TreeStream& operator<<(const ...)` and use the `constcast` in these methods rather than expose it to the user code.

Cheers,
 Ruben

miranov25:
Hello @shahor02 
I agree with your proposal.  Like that we can hide it. Will you make a modification in the class, or should we do it?
Marian

shahor02:
Hello,

I can do this next week, once back to CERN.

Cheers
  Ruben

On Mar 11, 2023 18:36, Marian Ivanov ***@***.***> wrote:

Hello @shahor02<https://github.com/shahor02>
I agree with your proposal. Like that we can hide it. Will you make a modification in the class, or should we do it?
Marian

PR #10854, opened 2023-02-28
o2-sim: Support for (ALICE3) field prototyping

sawenzel:
Possibility to instantiate a field class during
simulation that differs from o2::field::MagneticField.

This should allow to play with configurations for ALICE3, or other, without disturbing Run3.

Activate this with `export ALICE3_SIM_FIELD=ON`.

Moreover, this commit gives the possibility to take the actual TVirtualMagField field implementation from a ROOT macro with just-in-time compilation. Use this via:

`export ALICE3_MAGFIELD_MACRO=field.C`

where field.C, contains code returning a lambda, for instance

```
std::function<void(const double*, double*)> field() {
  return [](const double*, double* b) {
    static int count=0;
    if (count < 5) {
      std::cerr << "field call\n";
      count += 1;
    }
    b[0] = 1.;
    b[1] = 1.;
    b[2] = 2.;
  };
}
```

This allows to change the field setup without recompiling O2.

### DIFF
--- a/Common/Field/CMakeLists.txt
+++ b/Common/Field/CMakeLists.txt
@@ -16,7 +16,8 @@ o2_add_library(Field
                        src/MagFieldParam.cxx
                        src/MagneticField.cxx
                        src/MagneticWrapperChebyshev.cxx
-                       PUBLIC_LINK_LIBRARIES O2::MathUtils FairRoot::Base)
+                       src/ALICE3MagneticField.cxx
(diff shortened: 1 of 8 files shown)


sawenzel:
@Cas1997 : This PR provides rudimentary support for new fields in o2-sim. Could you see if this would be enough to get you going? Concretely, you would either

* complete the implemantation of the field function in ALICE3MagneticField.cxx
* or use the macro mechanism for rapid prototyping

sawenzel:
@mconcas : f.y.i

PR #10864, opened 2023-03-01
[MRRTF-163] Reject bad channels in DigitFiltering

aphecetche:
This PR contains a few things needed to actually reject bad channels during digit filtering :   

- a [statusMap creator workflow](https://github.com/aphecetche/AliceO2/blob/mrrtf-163-ter/Detectors/MUON/MCH/Conditions/src/StatusMapCreatorSpec.cxx) that emits a `MCH/STATUSMAP/0` message (containing a ... [StatusMap](https://github.com/aphecetche/AliceO2/blob/mrrtf-163-ter/Detectors/MUON/MCH/Conditions/include/MCHConditions/StatusMap.h) object). 
- a new [DigitFilter](https://github.com/aphecetche/AliceO2/blob/6dc35b989fd8735336018f11770530963042977e/Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx#L111) to reject pads based on a mask applied to the statusMap

Note that the creator can for the moment use information from `MCH/Calib/BadChannel` (produced during pedestal runs, existing) and/or `MCH/Calib/RejectList` ("manual" reject list, to be added to the CCDB). Usage of HV/LV values will be done later.

Configurable params are used to steer both the  [statusmap creator](https://github.com/aphecetche/AliceO2/blob/mrrtf-163-ter/Detectors/MUON/MCH/Conditions/include/MCHConditions/StatusMapCreatorParam.h) and the [digit filter](https://github.com/aphecetche/AliceO2/blob/6dc35b989fd8735336018f11770530963042977e/Detectors/MUON/MCH/DigitFiltering/include/MCHDigitFiltering/DigitFilterParam.h#L32).

Note also the introduction of a new class to identify a pad : `DsChannelDetId`, in addition to the existing `DsChannelId`. Both are fitting the relevant ids to `uint32_t`, but one is more readout oriented (the existing [DsChannelId](https://github.com/aphecetche/AliceO2/blob/mrrtf-163-ter/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/DsChannelId.h) ~ {solar,group,index}), while the other (the new [DsChannelDetId](https://github.com/aphecetche/AliceO2/blob/mrrtf-163-ter/DataFormats/Detectors/MUON/MCH/include/DataFormatsMCH/DsChannelDetId.h) ~ {de,ds,channel}) is more detector oriented.

Some doc has been added to [MCHConditions](https://github.com/aphecetche/AliceO2/blob/mrrtf-163-ter/Detectors/MUON/MCH/Conditions/README.md) directory.

### DIFF
--- a/Detectors/MUON/MCH/Conditions/src/StatusMap.cxx
+++ b/Detectors/MUON/MCH/Conditions/src/StatusMap.cxx
@@ -1,0 +1,77 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include "MCHConditions/StatusMap.h"
+
+#include "Framework/Logger.h"
+#include "MCHMappingInterface/Segmentation.h"
+#include "MCHRawElecMap/DsDetId.h"
+#include "MCHRawElecMap/Mapper.h"
+
+ClassImp(o2::mch::StatusMap);
+
+namespace o2::mch
+{
+
+void StatusMap::add(gsl::span<const DsChannelId> badchannels, uint32_t mask)
+{
+  for (auto id : badchannels) {
+    mStatus[raw::convert(id)] |= mask;
+  }
+}
+
+void StatusMap::add(gsl::span<const DsChannelDetId> badchannels, uint32_t mask)
+{
+  for (auto id : badchannels) {
+    mStatus[id] |= mask;
+  }
+}
+
+uint32_t StatusMap::status(const DsChannelDetId& id) const
+{
+  auto s = mStatus.find(id);
+  if (s != mStatus.end()) {
+    return s->second;
+  }
+  return 0;
+}
+
+std::map<int, std::vector<int>> applyMask(const o2::mch::StatusMap& statusMap,
+                                          uint32_t mask)
+{
+  std::map<int, std::vector<int>> rejectList;
+
+  std::unique_ptr<o2::mch::mapping::Segmentation> seg{nullptr};
+  int previousDeId{-1};
+
+  for (const auto& status : statusMap) {
+    auto channelId = status.first;
+    auto deId = channelId.getDeId();
+    if (deId != previousDeId) {
+      previousDeId = deId;
+      seg = std::make_unique<o2::mch::mapping::Segmentation>(deId);
+    }
+    auto dsId = channelId.getDsId();
+    auto channel = channelId.getChannel();
+    auto padId = seg->findPadByFEE(dsId, channel);
+    if (seg->isValid(padId)) {
+      if (mask && (mask & status.second)) {
+        rejectList[deId].emplace_back(padId);
+      }
+    } else {
+      LOGP(warning, "Got an invalid pad in bad channel map DE {} DS {} CH {} !",
+           deId, dsId, channel);
+    }
+  }
+  return rejectList;
+}
+
+} // namespace o2::mch
--- a/Detectors/MUON/MCH/Conditions/src/bad-channels-ccdb.cxx
+++ b/Detectors/MUON/MCH/Conditions/src/bad-channels-ccdb.cxx
@@ -1,0 +1,161 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include "CCDB/CcdbApi.h"
+#include <boost/program_options.hpp>
+#include <ctime>
+#include <numeric>
+#include <regex>
+#include <string>
+#include <vector>
+#include "DataFormatsMCH/DsChannelId.h"
+#include "MCHRawElecMap/Mapper.h"
+#include "MCHMappingInterface/Segmentation.h"
+
+namespace po = boost::program_options;
+
+using BadChannelsVector = std::vector<o2::mch::DsChannelId>;
+
+std::string ccdbPath(const std::string badChannelType)
+{
+  return fmt::format("MCH/Calib/{}", badChannelType);
+}
+
+void queryBadChannels(const std::string ccdbUrl,
+                      const std::string badChannelType, uint64_t timestamp, bool verbose)
+{
+  o2::ccdb::CcdbApi api;
+  api.init(ccdbUrl);
+  std::map<std::string, std::string> metadata;
+  auto source = ccdbPath(badChannelType);
+  auto* badChannels = api.retrieveFromTFileAny<BadChannelsVector>(source.c_str(), metadata, timestamp);
+  std::cout << "number of bad channels = " << badChannels->size() << std::endl;
+  if (verbose) {
+    for (const auto& badChannel : *badChannels) {
+      std::cout << badChannel.asString() << "\n";
+    }
+  }
+}
+
+void uploadBadChannels(const std::string ccdbUrl,
+                       const std::string badChannelType,
+                       uint64_t startTimestamp,
+                       uint64_t endTimestamp,
+                       std::vector<uint16_t> solarsToReject,
+                       bool makeDefault)
+{
+  BadChannelsVector bv;
+
+  auto det2elec = o2::mch::raw::createDet2ElecMapper<o2::mch::raw::ElectronicMapperGenerated>();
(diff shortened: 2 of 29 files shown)


Detectors/MUON/MCH/Conditions/src/StatusMap.cxx
```diff
@@ -0,0 +1,77 @@
...
+
+  std::unique_ptr<o2::mch::mapping::Segmentation> seg{nullptr};
+  int previousDeId{-1};
+
+  for (const auto& status : statusMap) {
+    auto channelId = status.first;
+    auto deId = channelId.getDeId();
+    if (deId != previousDeId) {
+      previousDeId = deId;
+      seg = std::make_unique<o2::mch::mapping::Segmentation>(deId);
```
pillot, line 60: This will create/destroy the segmentation every time it is called, and this function is called for every TF. Why not using the mapping::segmentation(deId) interface instead, which create the segmentation only once as far as I understand?

Detectors/MUON/MCH/Conditions/src/StatusMap.cxx
```diff
@@ -0,0 +1,77 @@
...
+    }
+    auto dsId = channelId.getDsId();
+    auto channel = channelId.getChannel();
+    auto padId = seg->findPadByFEE(dsId, channel);
+    if (seg->isValid(padId)) {
+      if (mask && (mask & status.second)) {
```
pillot, line 66: the case (mask == 0) could be tested only once at the beginning and return an empty map if true.

Detectors/MUON/MCH/Conditions/src/bad-channels-ccdb.cxx
```diff
@@ -0,0 +1,161 @@
...
+                       bool makeDefault)
+{
+  BadChannelsVector bv;
+
+  auto det2elec = o2::mch::raw::createDet2ElecMapper<o2::mch::raw::ElectronicMapperGenerated>();
+
+  for (auto solar : solarsToReject) {
+    auto ds = o2::mch::raw::getDualSampas<o2::mch::raw::ElectronicMapperGenerated>(solar);
+    for (const auto& dsDetId : ds) {
+      o2::mch::mapping::Segmentation seg(dsDetId.deId());
```
pillot, line 62: Also here you could use the mapping::segmentation(deId) interface to avoid re-creating the segmentation at every iterations, no?

Detectors/MUON/MCH/Conditions/src/statusmap-creator-workflow.cxx
```diff
@@ -0,0 +1,39 @@
...
+
+WorkflowSpec defineDataProcessing(const ConfigContext& cc)
+{
+  o2::conf::ConfigurableParam::updateFromString(cc.options().get<std::string>("configKeyValues"));
+
+  if (o2::mch::StatusMapCreatorParam::Instance().useBadChannels ||
+      o2::mch::StatusMapCreatorParam::Instance().useRejectList) {
+    return {o2::mch::getStatusMapCreatorSpec("mch-statusmap-creator")};
+  } else {
+    return {};
```
pillot, line 37: This is redundant with the protection in the spec, which produces "mch-dummy-statusmap-creator" in this case. Thus the behaviour (create a dummy spec or nothing) will differ between this workflow and the reco-workflow. Not really a problem but I am wondering if it is desirable.

Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
```diff
@@ -96,7 +98,39 @@ o2::mch::DigitFilter createSelectSignal()
...
+       rejectList.size(), rejectList.size() > 1 ? "s" : "", statusMask);
+}
+
+DigitFilter createBadChannelFilter(const StatusMap& statusMap,
+                                   uint32_t statusMask)
+{
+  auto rejectList = applyMask(statusMap, statusMask);
+  report(rejectList, statusMask);
+
+  return [rejectList, statusMask](const o2::mch::Digit& digit) -> bool {
```
pillot, line 117: statusMask is not used in the lambda.

Detectors/MUON/MCH/Workflow/CMakeLists.txt
```diff
@@ -126,15 +126,16 @@ o2_add_executable(
             src/reco-workflow.cxx
         COMPONENT_NAME mch
         PUBLIC_LINK_LIBRARIES
+            O2::MCHConditions
             O2::MCHDigitFiltering
             O2::MCHGeometryTransformer
             O2::MCHIO
+            O2::MCHIO
```
pillot, line 133: O2::MCHIO appears twice

pillot (commented):
I think the code as it is is working fine. My detailed comments into the code are mostly related to performances and could probably be done in a next step if relevant.

I also have 2 generic comments, still related to performances, although it might not be an issue for the filtering:
- if the maps are not updated too often, using unordered_map could be more efficient (at the moment it might be interesting to have the keys ordered per DE in applyMask() but maybe not so much with the changes I propose)
- The status map is produced using the new DsChannelDetId pad identifier, but then it is converted at every TF in the filtering in a list of deId/padId, which requires Segmentation::findPadByFEE to retrieve the padId. I am wondering if it would not be more efficient to produce the status map directly with deId/padId, which is the final pad identifier used in the reconstruction? A map of map with both Ids separated might be even better but I suspect it would not be messageable.

Detectors/MUON/MCH/Conditions/src/StatusMap.cxx
aphecetche: Well, it depends on the number of detection elements that do contain bad channels... the first call to `mapping::segmentation` triggers the creation of 156 segmentations. So the hope was that on average we have less than 156 DE with bad channels. But I admit I've not checked carefully, e.g. with the `MUON/Calib/BadChannels` we have from last year, whether that hope is backed up by numbers. I suspect it's the case but I've not checked.

Detectors/MUON/MCH/Conditions/src/StatusMap.cxx
aphecetche: good point. will change.

Detectors/MUON/MCH/Conditions/src/bad-channels-ccdb.cxx
aphecetche: same rationale as above.

Detectors/MUON/MCH/DigitFiltering/src/DigitFilter.cxx
aphecetche: indeed. will fix

Detectors/MUON/MCH/Workflow/CMakeLists.txt
aphecetche: thanks. will fix.

aphecetche:
@pillot Thanks for the review. 

>  I am wondering if it would not be more efficient to produce the status map directly with deId/padId, which is the final pad identifier used in the reconstruction? 

Good point. One option would be to "bite the bullet" and accept to get the DsChannelDetId to be 64 bits and then add the padid to it... 
I agree it would be nice to avoid having to resort to segmentation calls that often...
Will try...

Detectors/MUON/MCH/Conditions/src/StatusMap.cxx
pillot: Maybe I miss something, but even if not all DE have bad channel, their segmentation will be created and destroyed every time this function is called, which for the moment is for every TF, so this number will be multiplied by the number of TF processed (which I admit I don't how much it can be...).

Detectors/MUON/MCH/Conditions/src/bad-channels-ccdb.cxx
pillot: here it is created/destroyed inside a double loop, so likely more often than the number of DE involved. But this executable is "one shot" and performances are certainly not an issue here so there is no point in changing this I agree.

pillot:
> @pillot Thanks for the review.
> 
> > I am wondering if it would not be more efficient to produce the status map directly with deId/padId, which is the final pad identifier used in the reconstruction?
> 
> Good point. One option would be to "bite the bullet" and accept to get the DsChannelDetId to be 64 bits and then add the padid to it... I agree it would be nice to avoid having to resort to segmentation calls that often... Will try...

Is it needed to keep dsId and channel? Couldn't it be replaced by padId (which fits inside a uint16_t)?

Detectors/MUON/MCH/Conditions/src/StatusMap.cxx
aphecetche: no, you're right...

aphecetche:
if the `DsChannelDetId` becomes a (deid,padid) (which is indeed possible as this PR introduces that class, so that's the right moment to decide) it does fit in 32 bits, yes, but then to use it/interpret you might need again the segmentation... Not in the DigitFilter, but e.g. in the quality control (which is the next thing I'll work on, as I'd like to have a few plots of the evolution of the channels we reject during reconstruction). Whether it's an issue is yet to be seen.

All this discussion would be less relevant if there was a way for the DigitFilter to be updated only when the StatusMap actually changes, and not at each TF...

Detectors/MUON/MCH/Conditions/src/statusmap-creator-workflow.cxx
aphecetche: you're right, I can simply return the spec which turns itself into a no-op if needed.

aphecetche:
@pillot I'm closing this one, and will split it in more manageable pieces for the next steps.

PR #10865, opened 2023-03-01
DPL: more tests moved to catch2

ktf:


ktf:
More accumulated cleanups while working on other stuff...

PR #10866, opened 2023-03-02
TRD fix ranges of plots in macro

bazinski:
fix the ranges of the charge windows in the plots.
q0=q1 range 256
q2 range 64 was 256

This dates back to the confusion of what was Q0 and Q2.

### DIFF
--- a/Detectors/TRD/reconstruction/macros/checkTrackletCharges.C
+++ b/Detectors/TRD/reconstruction/macros/checkTrackletCharges.C
@@ -1,3 +1,20 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file checkTrackletCharges.C
+/// \brief Simple macro to display the tracklet charge information.
+//
+//  outputs a png
+//
+
 #if !defined(__CLING__) || defined(__ROOTCLING__)
 // ROOT header
 #include <TROOT.h>
@@ -23,9 +40,9 @@ using namespace o2::trd;
 std::vector<TH1F*> createTrdChargeHists()
 {
   std::vector<TH1F*> hCharges;
-  hCharges.push_back(new TH1F("Charge0", "Charge window 0;charge;counts", 64, 0, 63));
-  hCharges.push_back(new TH1F("Charge1", "Charge window 1;charge;counts", 256, 0, 255));
-  hCharges.push_back(new TH1F("Charge2", "Charge window 2;charge;counts", 256, 0, 255));
+  hCharges.push_back(new TH1F("Charge0", "Charge window 0;charge;counts", 128, -0.5, 127.5));
+  hCharges.push_back(new TH1F("Charge1", "Charge window 1;charge;counts", 128, -0.5, 127.5));
+  hCharges.push_back(new TH1F("Charge2", "Charge window 2;charge;counts", 64, -0.5, 63.5));
   return hCharges;
 }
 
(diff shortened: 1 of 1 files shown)


Detectors/TRD/reconstruction/macros/checkTrackletCharges.C
```diff
@@ -23,9 +23,9 @@ using namespace o2::trd;
 std::vector<TH1F*> createTrdChargeHists()
 {
   std::vector<TH1F*> hCharges;
-  hCharges.push_back(new TH1F("Charge0", "Charge window 0;charge;counts", 64, 0, 63));
+  hCharges.push_back(new TH1F("Charge0", "Charge window 0;charge;counts", 256, 0, 255));
   hCharges.push_back(new TH1F("Charge1", "Charge window 1;charge;counts", 256, 0, 255));
-  hCharges.push_back(new TH1F("Charge2", "Charge window 2;charge;counts", 256, 0, 255));
+  hCharges.push_back(new TH1F("Charge2", "Charge window 2;charge;counts", 64, 0, 63));
```
f3sch: Would it make more sense to also set the ranges for Q0/1 to their respective range i.e. 0...127?

`
 hCharges.push_back(new TH1F("Charge0", "Charge window 0;charge;counts", 127, 0, 127));
 hCharges.push_back(new TH1F("Charge1", "Charge window 1;charge;counts", 127, 0, 127));
 hCharges.push_back(new TH1F("Charge2", "Charge window 2;charge;counts", 63, 0, 63));
`

bazinski: I thought that as well, but then its hard to see the spike on the right.

f3sch: Makes sense, the same argument would then hold for Q2 since we also see a spike there.

bazinski: yip, and your numbers above cut a bin though. Anyway apparently I need to add a copyright, not sure why only now.

f3sch (dismissed):

f3sch:
LGTM!

tdietel:
The charges in Tracklet64 are 8 bits, so one could argue that a range up to 256 would be failsafe. But in practice we will only get 7/7/6 bits from the FEE, so the restricted ranges are also fine.

Detectors/TRD/reconstruction/macros/checkTrackletCharges.C
```diff
@@ -23,9 +40,9 @@ using namespace o2::trd;
 std::vector<TH1F*> createTrdChargeHists()
 {
   std::vector<TH1F*> hCharges;
-  hCharges.push_back(new TH1F("Charge0", "Charge window 0;charge;counts", 64, 0, 63));
+  hCharges.push_back(new TH1F("Charge0", "Charge window 0;charge;counts", 256, 0, 255));
   hCharges.push_back(new TH1F("Charge1", "Charge window 1;charge;counts", 256, 0, 255));
-  hCharges.push_back(new TH1F("Charge2", "Charge window 2;charge;counts", 256, 0, 255));
+  hCharges.push_back(new TH1F("Charge2", "Charge window 2;charge;counts", 128, 0, 127));
```
tdietel: I think that's incorrect: the ranges should be 256, 0.0, 256.0. Or, actually, I usually prefer 256, -0.5, 255.5 to have the labels in the bin centers.

bazinski: oui oui, they can all potentially be 8 bit :-( However this is just plotting them, not manipulating data, so its more a question of clear easily readable graphs, no? and given that most of the time its going to be 7 7 6 when not using the scaled q values, is it not clearer to h ave 255.5 255.5 127.5 ? giving an area at the right with nothing and clearly showing the spike? Yes I definitely agree about the -0.5 to -???.5

davidrohr: @tdietel : could you approve, then we'Ll merge it

tdietel: Restricting it to 7/7/6 bits is fine. Then the ranges should be 
128, -0.5, 127.5
64, -0.5, 63.5
Please update

tdietel (changes requested):

bazinski:
@tdietel  done.

tdietel (approved):

PR #10867, opened 2023-03-02
Re-indexing "working" copies of MCH tracks (and correct ROF entries for MFT tracks)

robincaron13:
To follow https://alice.its.cern.ch/jira/browse/O2-3572 @shahor02:
- re-indexing of "working" copies of MCH tracks (mMCHWork) using `std::vector<int> mMCHID2Work` 
- (could be suggestion) fixing the ROF entries for a loop which is producing mMFTWork.size() with zero, using copy of mMFTTrackROFRec with `gsl::span<const o2::itsmft::ROFRecord> mMFTTrackROFRecCopyUp`

### DIFF
--- a/Detectors/GlobalTracking/src/MatchGlobalFwd.cxx
+++ b/Detectors/GlobalTracking/src/MatchGlobalFwd.cxx
@@ -257,8 +257,9 @@ bool MatchGlobalFwd::prepareMFTData()
   for (int irof = 0; irof < nROFs; irof++) {
     const auto& rofRec = mMFTTrackROFRec[irof];
     int irofUp = irof + 1;
-    if (irofUp == nROFs)
+    if (irofUp == nROFs) {
       irofUp = irof;
+    }
     const auto& rofRecCopyUp = mMFTTrackROFRecCopyUp[irofUp];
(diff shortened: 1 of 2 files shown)


Detectors/GlobalTracking/src/MatchGlobalFwd.cxx
```diff
@@ -306,7 +309,8 @@ void MatchGlobalFwd::loadMatches()
     auto MCHId = match.getMCHTrackID();
     LOG(debug) << "     ==> MFTId = " << MFTId << " MCHId =  " << MCHId << std::endl;
 
-    auto& thisMCHTrack = mMCHWork[MCHId];
+    auto& thisMCHTrack = mMCHWork[mMCHID2Work[MCHId]];
+    LOG(info) << " MFTId: " << MFTId << " MCHId: " << MCHId << " --> mMCHID2Work[MCHId]:" << mMCHID2Work[MCHId];
```
shahor02: I guess this is ready to merge, could you please just remove this info, it will produce lot of output.

shahor02 (approved):
CI was green, merging after suppressing excessive logging line.

PR #10868, opened 2023-03-02
Add strangeness tracking to AO2Ds

fmazzasc:
Add optional tables into the AO2Ds with strange track info + track indexes. @mpuccio


fmazzasc:
Hi @jgrosseo , may I ask you to have a look into this P.R.? You told me on Monday that you have few additional suggestions.  Thanks in advance!

jgrosseo:
Sorry, wrong click. I will have a look and let you know.

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -860,6 +860,110 @@ DECLARE_SOA_TABLE(Decays3Body, "AOD", "DECAY3BODY", //! Run 2 cascade table
...
+DECLARE_SOA_INDEX_COLUMN_FULL(Decay3Body, decay3BodyIdx, int, Decays3Body, ""); //! Decay 3 body index
+
+DECLARE_SOA_COLUMN(DecayX, decayX, float);             //! X coordinate of decay vertex
+DECLARE_SOA_COLUMN(DecayY, decayY, float);             //! Y coordinate of decay vertex
+DECLARE_SOA_COLUMN(DecayZ, decayZ, float);             //! Z coordinate of decay vertex
+DECLARE_SOA_COLUMN(XiMass, xiMass, float);             //! Xi mass
+DECLARE_SOA_COLUMN(OmegaMass, omegaMass, float);       //! Omega mass
+DECLARE_SOA_COLUMN(H3Lmass, h3Lmass, float);           //! H3L mass
+DECLARE_SOA_COLUMN(H4Lmass, h4Lmass, float);           //! H4L mass
+DECLARE_SOA_COLUMN(He4Lmass, he4Lmass, float);         //! He4L mass
```
jgrosseo, line 877: During the WP14 meeting you did not show all these masses but only two of them if I am not mistaken. Is there no way to some of these from each other?

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -860,6 +860,110 @@ DECLARE_SOA_TABLE(Decays3Body, "AOD", "DECAY3BODY", //! Run 2 cascade table
...
+                       track::CYY,
+                       track::CZY, track::CZZ,
+                       track::CSnpY, track::CSnpZ, track::CSnpSnp,
+                       track::CTglY, track::CTglZ, track::CTglSnp, track::CTglTgl,
+                       track::C1PtY, track::C1PtZ, track::C1PtSnp, track::C1PtTgl, track::C1Pt21Pt2,
+                       track::Px<track::Signed1Pt, track::Snp, track::Alpha>,
+                       track::Py<track::Signed1Pt, track::Snp, track::Alpha>,
+                       track::Pz<track::Signed1Pt, track::Tgl>,
+                       track::Rapidity<track::Signed1Pt, track::Tgl>,
+                       track::Sign<track::Signed1Pt>);
```
jgrosseo, line 908: What is your finding from trying to add a new track type and a reference to the track table?

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -860,6 +860,110 @@ DECLARE_SOA_TABLE(Decays3Body, "AOD", "DECAY3BODY", //! Run 2 cascade table
...
```
jgrosseo, line 958: If we keep the cov matrix here (and not in the other table (see above)), then something is not ok I think. There should be a FULL and EXTENDED table...

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -860,6 +860,110 @@ DECLARE_SOA_TABLE(Decays3Body, "AOD", "DECAY3BODY", //! Run 2 cascade table
 using Decays3Body = Decays3Body; //! this defines the current default version
 using Decay3Body = Decays3Body::iterator;
 
+namespace strangenesstracking
+{
+DECLARE_SOA_INDEX_COLUMN_FULL(ITSIdx, Idx, int, Tracks, "");                    //! ITS index
```
```suggestion
DECLARE_SOA_INDEX_COLUMN_FULL(ITSTrack, itsTrack, int, Tracks, "");                    //! ITS index
```

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -860,6 +860,110 @@ DECLARE_SOA_TABLE(Decays3Body, "AOD", "DECAY3BODY", //! Run 2 cascade table
...
+DECLARE_SOA_INDEX_COLUMN(Cascade, cascadeIdx);                                  //! Cascade index
+DECLARE_SOA_INDEX_COLUMN(V0, v0Idx);                                            //! V0 index
```
```suggestion
DECLARE_SOA_INDEX_COLUMN(Cascade, cascade);                                  //! Cascade index
DECLARE_SOA_INDEX_COLUMN(V0, v0);                                            //! V0 index
```
Nowhere else we use `idx` in the column names

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -860,6 +860,110 @@ DECLARE_SOA_TABLE(Decays3Body, "AOD", "DECAY3BODY", //! Run 2 cascade table
...
+DECLARE_SOA_INDEX_COLUMN_FULL(Decay3Body, decay3BodyIdx, int, Decays3Body, ""); //! Decay 3 body index
```
jgrosseo: Why FULL? Isn't DECLARE_SOA_INDEX_COLUMN enough?

jgrosseo (changes requested):
Here some comments just on the data model.

I would prefer to keep this draft until it is overall agreed. @ddobrigk what do you think?

Could you also fix the comment in line 857

Framework/Core/include/Framework/AnalysisDataModel.h
ddobrigk: > Here some comments just on the data model.
> 
> I would prefer to keep this draft until it is overall agreed. @ddobrigk what do you think?
> 
> Could you also fix the comment in line 857

Well, I do not mind much that we keep it WIP or not - I understand you are worried that maybe it gets merged before we converge. However, the more important thing is that we converge quickly, because it would be very important to have this merged! 

For the "Idx" in the index columns, I agree that it's best to omit that because AFAIK the index column returns immediately the object if queried (and that's why we have chosen to do it like this everywhere - no 'idx'). 

Regarding adding these tracks to the track table, the pros and cons are: 
* pro: we have all tracks (with central barrel parametrization) in one table
* pro: we can use standard utilities like the track propagator to possibly calculate DCA 
* con: we need to make sure users check the track type (also in things like standard PID) 
* con: we'll have to fill all auxiliary tables with dummy values to keep joinability (that includes tracksExtra, PID, etc...). But that's probably not an issue given the small data size...

To be honest I don't have a very strong preference either way (track table or separate table)... 

One more comment (to be provocative, not for this PR but for the future): has anybody given any thought to how to possibly store cluster size information for PID in AO2Ds? Stuff we can strangeness-track probably leaves a burning trail compared to pions :-D

jgrosseo: To my understanding at the moment it is anyway incomplete because we need the compressed (stored) and uncompressed (through the spawner table). TracksCov vs. StoredTracksCov. These are a lot of additions to the data model, so I vote for having it in the Tracks table.

The track selection already checks the track type and yes, other service tasks, should do too (and then fill empty values for the entries skipped).

ddobrigk: Ah, I realize I forgot to comment on the individual masses being present: I believe this is a reasonable thing to do. A recalculation would require the daughter tracks, which would blow up the size much more than just storing a few masses... Anyway the exact recalculation of the masses also needs to have its full effect studied and documented later but the data model change with the additional masses looks very reasonable to me!

fmazzasc: Ok, thanks a lot for your comments. I drafted the P.R. again and I am now looking for the best way to modify the AODProducer to add the strange tracks into the full track table

PR #10872, opened 2023-03-02
Improve metrics handling

ktf:


ktf:
This is a separate PR with the developments I am doing to fix the multithreading behaviour. In principle the other developments work well as well, but maybe it a good idea to proceed step by step.

ktf:
@TimoWilken I thought you fixed the issue with the copyright?

TimoWilken:
@ktf I fixed it in the ["copyright headers" workflow](https://github.com/alisw/ali-bot/blob/master/.github/workflows/c%2B%2B-code-formatting.yml#L145), but not in the fullCI check -- I didn't realise that had the same check. Is it worth maintaining both?

ktf:
no

ktf:
Missing tests were already passing, only thing which changed was the number of metrics. Tested also on hyperloop with my local tag and the usual tests work, with no visibile change in performance.

PR #10874, opened 2023-03-03
[ITS, ITS3] Create topology-dictionary routine for ITS3

lbariogl:
All the new classes inherit from ITS2 version.

Main points:
- in the new dictionary format, the position of the COG within the bounding box is given in pixels and not in cm. This change is needed because the same topology can be in the inner barrel or in the outer barrel, which have different segemtations.
- `its3::BuildTopologyDictionary` fills the entries of the dictionary with coordinates in pixels.
- In `its3::TopologyDictionary`, the `GetClusterCoordinates` method returns the position of the cluster in cm (conversion is made internally).
-  It was necessaty to make `itsmft::TopologyDictionary` private data members public, to make them accessible to `its3::BuildTopologyDictionary` (different namespace and `itsmft::BuildTopologyDictionary` already present). This is compatible with previous versions of the code.
- A macro for creating the dictionaries was added in `Detectors/Upgrades/ITS3/macros/test`. It was modulated from the its2 version.

The routine, up to the creation of the ITS3 dictionary, has been tested on simulated Pb-Pb collisions

### DIFF
--- a/Detectors/Upgrades/ITS3/workflow/include/ITS3Workflow/TrackerSpec.h
+++ b/Detectors/Upgrades/ITS3/workflow/include/ITS3Workflow/TrackerSpec.h
@@ -47,7 +47,7 @@ class TrackerDPL : public framework::Task
   void run(framework::ProcessingContext& pc) final;
   void endOfStream(framework::EndOfStreamContext& ec) final;
   void finaliseCCDB(framework::ConcreteDataMatcher& matcher, void* obj) final;
-  void setClusterDictionary(const o2::itsmft::TopologyDictionary* d) { mDict = static_cast<const o2::its3::TopologyDictionary*>(d); } //TODO: remove casting from Run3 to RUn2 dictionary
+  void setClusterDictionary(const o2::itsmft::TopologyDictionary* d) { mDict = static_cast<const o2::its3::TopologyDictionary*>(d); } // TODO: remove casting from Run3 to RUn2 dictionary
 
  private:
   void updateTimeDependentParams(framework::ProcessingContext& pc);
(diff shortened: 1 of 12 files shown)


Detectors/Upgrades/ITS3/workflow/include/ITS3Workflow/TrackerSpec.h
```diff
@@ -47,7 +47,7 @@ class TrackerDPL : public framework::Task
   void run(framework::ProcessingContext& pc) final;
   void endOfStream(framework::EndOfStreamContext& ec) final;
   void finaliseCCDB(framework::ConcreteDataMatcher& matcher, void* obj) final;
-  void setClusterDictionary(const o2::itsmft::TopologyDictionary* d) { mDict = static_cast<const o2::its3::TopologyDictionary*>(d); }
+  void setClusterDictionary(const o2::itsmft::TopologyDictionary* d) { mDict = static_cast<const o2::its3::TopologyDictionary*>(d); } // TODO: remove casting from Run3 to RUn2 dictionary
```
fgrosa: This is needed for the vertexer, which is currently working. Could you please adapt it?

fgrosa (changes requested):
Ciao @lbariogl thanks a lot for the work!
The only comment I have is that the cast from `o2::itsmft::TopologyDictionary` to `o2::its3::TopologyDictionary` should be fixed in the `TrackerSpec` (otherwise I fear that the vertexer would not work anymore) and in the `CheckClusters.C` macro, which makes fail the tests otherwise.

lbariogl:
Hi @fgrosa, you are right and I have implemented your suggestions. I have also solved the problem with the test of `CreateDictionariesITS3.C` (or at least when testin macros also in local, let's se here). Perhaps it is better to test the full chain (including vertexing and tracking) in local before merging the PR and eventually uploading a `its3::TopologyDictionary` object to the CCDB. I can mark the PR as WIP in case.

fgrosa (dismissed):

lbariogl:
Hi @fgrosa @mconcas, I have tested the changes using `CheckClustersITS3.C` and the results (in attachment) look reasonable.

[it3clusters_dx_vs_dz.pdf](https://github.com/AliceO2Group/AliceO2/files/10896274/it3clusters_dx_vs_dz.pdf)
[it3clusters_y_vs_x_vs_z.pdf](https://github.com/AliceO2Group/AliceO2/files/10896276/it3clusters_y_vs_x_vs_z.pdf)

fgrosa (approved):

mconcas (approved):
Seems fine for the ITS part

PR #10875, opened 2023-03-03
Fix aod producer segfault at emcal propagation

shahor02:


davidrohr (dismissed):
thx, this fixes the crash I have been seeing

davidrohr:
Checking the C++ spec, C++14 explicitly declares std::exp as non-constexpr (for whatever reason). So the clang error on macos is legitimate... We should make that non-constexpr.

And this seems a reasonable warning:
```
Detectors/AOD/src/AODProducerWorkflowSpec.cxx:2204:7: error: logical not is only applied to the left hand side of this comparison [-Werror,-Wlogical-not-parentheses]
      !std::abs(outTr.getZAt(xtrg, 0)) > ZEMCALMAX ||
      ```

PR #10881, opened 2023-03-03
TPC FastSpaceChargeCorrection: add missing member to cloneFromObject

matthias-kleiner:


matthias-kleiner:
@davidrohr 
This fixes the problem we discussed. It was actually done correctly in the past when no correction was applied, and only appears now with https://github.com/AliceO2Group/AliceO2/pull/10871 

@sgorbuno Can you confirm, that the member was just forgotten to be copied in the `cloneFromObject` function?

davidrohr:
@matthias-kleiner : thx for the fix. Does this fix also the problem that removing the setNoCorrection yields a max drift length of 0?

matthias-kleiner:
> @matthias-kleiner : thx for the fix. Does this fix also the problem that removing the setNoCorrection yields a max drift length of 0?
Yes this fixes the problem. The member containing the drift length was not copied when `clonefromobject` was called and then the max drift length was reset to 0

PR #10882, opened 2023-03-03
DPL GUI: avoid extra allocation to display metrics

ktf:


ktf:
Superseded by #10921.

PR #10892, opened 2023-03-06
[ITS3]: Add a file to customize material budget LUT input for ITS3 IB

ChunzhengLab:


Detectors/Upgrades/ITS3/macros/test/buildMatBudLUT.C
```diff
@@ -0,0 +1,411 @@
...
+  // Layer 3
+  lrData.emplace_back(LrData(lrData.back().rMax, lrData.back().rMax + drStepChip, zSpanH, zBin, rphiBin));
+  // Air
+  do {
+    lrData.emplace_back(LrData(lrData.back().rMax, lrData.back().rMax + drStep, zSpanH, zBin, rphiBin));
+  } while (lrData.back().rMax < 6.0f - drStep + kToler);
+  lrData.emplace_back(LrData(lrData.back().rMax, 6.0f, zSpanH, zBin, rphiBin));
+
+  // Layer 4
+  lrData.emplace_back(LrData(lrData.back().rMax, lrData.back().rMax + drStepChip4Layer, zSpanH, zBin, rphiBin));
```
fgrosa: Maybe we can have an argument that allows us to turn on/off the addition of the 4th layer to have the possibility to build the LUT both for the 3-layer and the 4-layer versions

fgrosa (changes requested):
Hi @ChunzhengLab thanks a lot for the work! I only have one comment, see below

Detectors/Upgrades/ITS3/macros/test/buildMatBudLUT.C
ChunzhengLab: Many thanks!
With your method, it is now able to automatically determine configure with the 3-layer or the 4-layer version based on the read-in geometry.

fgrosa (dismissed):

mconcas (approved):

PR #10897, opened 2023-03-06
Stream operators for RDH v6 and v7

mfasDa:
- Add streaming operators for RDH v6 and v7
- Move streaming operators from EMCALReconstruction to DetectorsRaw


mfasDa:
Has to go in sync with https://github.com/AliceO2Group/QualityControl/pull/1662

shahor02:
To avoid the QC syncing lol problem, we usually go in 2 steps: you can define methods in the raw and alias them to emcal dir, so that the old methods are still available

mfasDa:
In order to keep compatibility with QC for the transition period I added a file RawHeaderStream.h in EMCALReconstruction which just includes RawHeaderStream.h from DetectorsRaw. In local tests this worked. The file in EMCALReconstruction can be removed once both PRs are merged.

PR #10898, opened 2023-03-06
Allow to run the MID chamber efficiency workflow with matched MCH-MID tracks

dstocco:
The workdflow was moved to Workflow/MUON since it also depends on the matching now.
Notice that one can run on all MID tracks as before if no option is provided


aphecetche (approved):

dstocco:
Errors are unrelated. @ktf is there a way to re-trigger the test without having to re-do the review?

dstocco:
@shahor02 @davidrohr @ktf : the CI keeps failing on an unrelated error since about 1 week. Is this a known issue? Can this be force-merged?

shahor02 (approved):

PR #10905, opened 2023-03-07
[MRRTF-216] new ChannelCode class

aphecetche:
The ChannelCode class encodes in a 64 bits integer the following information about a MCH channel :

- detection element identifier (and index)
- pad index within the detection element
- dual sampa identifier (and index)
- solar identifier (and index)
- elink identifier (or index, those are the same)
- channel number

This class serves the same purposes as DsChannelId, but using two different "coordinate systems" to reference the elements within the spectrometer, while DsChannelId uses only one.  DsChannelId is readout/online oriented, while ChannelCode is both reconstruction/offline _and_ readout/online oriented. But at a cost of wice the size.


aphecetche:
@pillot this one is green, ready to be reviewed :wink:

pillot (approved):

PR #10912, opened 2023-03-07
Avoid keeping large tree in memory when creating distortion maps

martenole:
In case `treeBinnedResiduals` is kept in memory, because the binned residuals should not be stored, the memory usage of the script grows above 6GB.
To avoid this a temporary file is created where the binned residuals are flushed during the map creation.

### DIFF
--- a/Detectors/TPC/calibration/SpacePoints/macro/staticMapCreator.C
+++ b/Detectors/TPC/calibration/SpacePoints/macro/staticMapCreator.C
@@ -32,6 +32,7 @@
 #include <TH2.h>
 #include <TF1.h>
 
+#include <filesystem>
 #include <fstream>
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


wiechula (dismissed):
Failures look unrelated, can this be merged?

martenole:
Actually, I wanted to avoid leaving the temporary buffer file around. But when I call `gSystem->Unlink(tmpFileName.c_str());` then valgrind complaints and I cannot test it anymore.

wiechula:
@martenole , did you try to use https://en.cppreference.com/w/cpp/filesystem/remove

martenole:
Same error from valgrind with that. Was already reported to ROOT: https://sft.its.cern.ch/jira/browse/ROOT-10698
But I assume it is a valgrind issue. Anyway, I monitored the memory usage with top and it was OK. So will merge this once the checks are through

PR #10926, opened 2023-03-08
TRD pileup simulation fix

martenole:
- for each collision the TRD trigger time instead of the actual collision time was set
- for pileup collisions which occured within the dead time of the TRD the labels of each signal were added to the digits even if there was no actual signal contribution

When @shahor02 is back from vacation I will check with his macros if this observes all the issues he had seen. The checks I was doing are all passing fine now.


martenole:
Bug was there since https://github.com/AliceO2Group/AliceO2/commit/ca981855bbda9fff43ebda2d727c8f8a0107b07e and I missed it when I checked another issue with the pileup implementation some time ago :(

pachmay (approved):
Thanks

bazinski (approved):

martenole:
Failure unrelated, merging

PR #10930, opened 2023-03-10
[QC-384] flexible cycle duration

Barthelemy:


Barthelemy:
In MergerConfig `publicationDecision` is replaced by `publicationDecisionNew`. 

`publicationDecision` remains to avoid failures in building QC and get this PR merged. I will then use `publicationDecisionNew` in QC.

Barthelemy:
error is unrelated

Utilities/Mergers/src/MergerBuilder.cxx
```diff
@@ -96,8 +97,13 @@ framework::DataProcessorSpec MergerBuilder::buildSpec()
     merger.algorithm = framework::adaptFromTask<FullHistoryMerger>(mConfig, subSpec);
   }
 
-  merger.inputs.push_back({"timer-publish", "TMR", mergerDataDescription(mName), mergerSubSpec(mLayer, mId), framework::Lifetime::Timer});
-  merger.options.push_back({"period-timer-publish", framework::VariantType::Int, static_cast<int>(mConfig.publicationDecision.param * 1000000), {"timer period"}});
+  // Create the TimerSpec for cycleDurations
+  std::vector<o2::framework::TimerSpec> timers;
+  for (auto& [cycleDuration, period] : mConfig.publicationDecisionNew.param) {
```
knopers8: `period` is rather confusing, can you use sth like `durationSinceStart`?

Utilities/Mergers/include/Mergers/MergerConfig.h
```diff
@@ -71,6 +71,7 @@ struct MergerConfig {
   ConfigEntry<InputObjectsTimespan> inputObjectTimespan = {InputObjectsTimespan::FullHistory};
   ConfigEntry<MergedObjectTimespan, int> mergedObjectTimespan = {MergedObjectTimespan::FullHistory};
   ConfigEntry<PublicationDecision> publicationDecision = {PublicationDecision::EachNSeconds, 10};
+  ConfigEntry<PublicationDecision, std::vector<std::pair<size_t, size_t>>> publicationDecisionNew = {PublicationDecision::EachNSeconds, {{10, 1}}};
```
knopers8: Do you intend to move `publicationDecisionNew` to `publicationDecision` once we are done with migration on the QC side? It seems like a lot of coordination to do.

It would add some code, which would be anyway dropped later, but if we want a smoother transition, you could try creating a class which has constructors for both single `int` and `std::vector<std::pair<size_t, size_t>>`. Then we could do the transition in 3 PRs, one here, one in QC, one here.

knopers8: Can you document the new interface?

knopers8 (changes requested):

Utilities/Mergers/src/MergerBuilder.cxx
knopers8: or `validity`, as in the DPL interface

Barthelemy: ok for validity

Utilities/Mergers/include/Mergers/MergerConfig.h
Barthelemy: fine with me, I have made an extra commit.

knopers8 (approved):
thank you

Barthelemy:
I think that the errors are unrelated. Could this be merged please ?

shahor02 (approved):

PR #10932, opened 2023-03-10
[EMCAL-688] EMCal/ClusterFactory: Add Exotic Cluster Check

mhemmer-cern:
- Add functionality to flag exotic clusters
- Variables for this functionality can be set in EMCal Correction Task in O2Physics, PR for this will follow once this is merged!

### DIFF
--- a/Detectors/EMCAL/base/src/ClusterFactory.cxx
+++ b/Detectors/EMCAL/base/src/ClusterFactory.cxx
@@ -73,13 +73,19 @@ o2::emcal::AnalysisCluster ClusterFactory<InputType>::buildCluster(int clusterIn
   // First calculate the index of input with maximum amplitude and get
   // the supermodule number where it sits.
 
-  auto [inputIndMax, inputEnergyMax, cellAmp] = getMaximalEnergyIndex(inputsIndices);
+  auto [inputIndMax, inputEnergyMax, cellAmp, shared] = getMaximalEnergyIndex(inputsIndices);
+
+  short towerId = mInputsContainer[inputIndMax].getTower();
+
+  float exoticTime = mInputsContainer[inputIndMax].getTimeStamp();
+
+  clusterAnalysis.setIsExotic(isExoticCell(towerId, inputEnergyMax, exoticTime));
 
   clusterAnalysis.setIndMaxInput(inputIndMax);
 
   clusterAnalysis.setE(cellAmp);
 
-  mSuperModuleNumber = mGeomPtr->GetSuperModuleNumber(mInputsContainer[inputIndMax].getTower());
+  mSuperModuleNumber = mGeomPtr->GetSuperModuleNumber(towerId);
 
   clusterAnalysis.setNCells(inputsIndices.size());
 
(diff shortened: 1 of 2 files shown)


Detectors/EMCAL/base/src/ClusterFactory.cxx
```diff
@@ -551,24 +555,145 @@ void ClusterFactory<InputType>::evalElipsAxis(gsl::span<const int> inputsIndices
...
+    if (towerId2 == iTowerId) {
+      ecell1 = mInputsContainer[cellIndex].getEnergy();
+    }
+    if (towerId3 == iTowerId) {
+      ecell1 = mInputsContainer[cellIndex].getEnergy();
+    }
+    if (towerId4 == iTowerId) {
+      ecell1 = mInputsContainer[cellIndex].getEnergy();
+    }
+  }
```
mhemmer-cern: Might add a lookup table to get to O(1) for this part at a later point

Detectors/EMCAL/base/src/ClusterFactory.cxx
```diff
@@ -551,24 +555,145 @@ void ClusterFactory<InputType>::evalElipsAxis(gsl::span<const int> inputsIndices
...
+{
+  if (ecell < mExoticCellMinAmplitude) {
+    return false; // do not reject low energy cells
+  }
+
+  float eCross = getECross(towerId, ecell);
+
+  if (1 - eCross / ecell > mExoticCellFraction) {
+    LOG(debug) << Form("EXOTIC CELL id %d, eCell %f, eCross %f, 1-eCross/eCell %f",
+                       towerId, ecell, eCross, 1 - eCross / ecell);
```
mfasDa: Should not hold up the merging but for logging you should either use the stream operators directly or in case you want a format string `LOGF`. Using ROOTs `Form` together with a stream does not make too much sense, neither is it very performant. Note that in o2 we rather use the fmt library in case we would need format strings.

Detectors/EMCAL/base/src/ClusterFactory.cxx
```diff
@@ -551,24 +555,145 @@ void ClusterFactory<InputType>::evalElipsAxis(gsl::span<const int> inputsIndices
...
+    }
+    if (towerId2 == iTowerId) {
+      ecell1 = mInputsContainer[cellIndex].getEnergy();
+    }
+    if (towerId3 == iTowerId) {
+      ecell1 = mInputsContainer[cellIndex].getEnergy();
+    }
+    if (towerId4 == iTowerId) {
+      ecell1 = mInputsContainer[cellIndex].getEnergy();
+    }
```
mfasDa: I think this implementation is not correct. mCellIndicies are the cells indices only of all clustered cells, while the Fcross cut takes the neightboring cells of the leading cell in order to calculate the Fcross, no matter whehter a cell is clustered. You have to loop over all entries in mInputContainer and check whether the tower index of the cell is among one of the 4 neighbors.

Detectors/EMCAL/base/src/ClusterFactory.cxx
```diff
@@ -551,24 +555,145 @@ void ClusterFactory<InputType>::evalElipsAxis(gsl::span<const int> inputsIndices
...
+  }
+
+  if (ecell1 < mExoticCellInCrossMinAmplitude || w1 <= 0)
+    ecell1 = 0;
+  if (ecell2 < mExoticCellInCrossMinAmplitude || w2 <= 0)
+    ecell2 = 0;
+  if (ecell3 < mExoticCellInCrossMinAmplitude || w3 <= 0)
+    ecell3 = 0;
+  if (ecell4 < mExoticCellInCrossMinAmplitude || w4 <= 0)
+    ecell4 = 0;
```
mfasDa: The original implementation has also a condition that the time difference is within a certain range.

mfasDa (changes requested):
Needs some fixes

Detectors/EMCAL/base/src/ClusterFactory.cxx
mhemmer-cern: Also it should not always be ecell1 that is getting set. :sweat_smile:

mfasDa (dismissed):
Looks good now!

mfasDa:
CI errors unrelated.

PR #10916, opened 2023-03-07
[MRRTF-217] introduce StatusMap object and statusmap-creator-workflow

aphecetche:

PR #10933, opened 2023-03-11
Revert "[MRRTF-217] introduce StatusMap object and statusmap-creator-workflow"

davidrohr:
Reverts AliceO2Group/AliceO2#10916

@aphecetche @pillot : This PR breaks the software validation using the FST for the EPN farm. The workflow gets simply stuck and doesn't process any more. (Did a bisec between O2/dev, and 2022-03-09, which was a working tag, and this PR starts to break it)

I am not fully sure why it passes the full CI. What I have seen is that it fails only when I enable also the calib workflow. FST with reconstruction.

Unfortunately, the remote debug GUI is still broken, so it is not so simple to investigate (https://alice.its.cern.ch/jira/browse/O2-3598?filter=-2).

For you to check, I am uploading logs with and without the PR
[logs.tar.gz](https://github.com/AliceO2Group/AliceO2/files/10948820/logs.tar.gz)

Merging this to return the software to a working state at P2.


aphecetche:
@davidrohr can you show me the command to run (locally) the FST with calib workflow ? 

One point to note : with the default reco params the statusmap creator that is created is a dummy one : 

```
 if (input.empty()) {
    return {"mch-dummy-statusmap-creator",
            {},
            {},
            AlgorithmSpec::dummyAlgorithm()};
  }
```

(I did not want to switch on the real statusmap creator until a next PR with its usage to reject digits in the digit filtering has been done)

Have no idea if that can explain anything (in my local tests the statusmap, either dummy or real, was working fine), but I thought I would mention it in case.

davidrohr:
@aphecetche : What was failing for me on the EPN is:
```
SEVERITY=info FST_TMUX_LOGPREFIX=log WORKFLOW_EXTRA_PROCESSING_STEPS="TPC_DEDX,MFT_RECO,MID_RECO,MCH_RECO,MATCH_MFTMCH,MATCH_MCHMID,MUON_SYNC_RECO,ZDC_RECO" NTIMEFRAMES=2000 DPL_REPORT_PROCESSING=1 TFDELAY=3 $O2_ROOT/prodtests/full-system-test/start_tmux.sh dd
```
But it will use the 8 GPUs in the EPN by default. I have not tried to reproduce it locally.

For running locally, you should add `FST_TMUX_NO_EPN=1 NUM_DPL_WORKFLOWS=1 SHMSIZE=32000000000 DDSHMSIZE=32000000000` (with proper memory sizes that work for you).

Please note that I will be absent from Thursday afternoon on, so for testing a potential fix on the EPNs, please refer to @martenole .

PR #10935, opened 2023-03-11
[EMCAL-688] ClusterFactory: Add lookup table between towerId and cell…

mhemmer-cern:
…index

This should speed up the check for exotics especially in Pb-Pb collisions when we expect a high occupancy


mfasDa (approved):
Looks good.

PR #10937, opened 2023-03-13
Removing /dev/shm files for current session.

chiarazampolli:
Needed in case an old session with the same ID as the current was left uncleaned.


prodtests/full-system-test/run-workflow-on-inputlist.sh
```diff
@@ -30,6 +30,8 @@ LOG_PREFIX="log_$(date +%Y%m%d-%H%M%S)_"
 [[ -z $OVERRIDE_SESSION ]] && export OVERRIDE_SESSION=default_$$_$RANDOM
 [[ -z $INRAWCHANNAME ]] && export INRAWCHANNAME=tf-builder-$$-$RANDOM
 
+SESSION_ID=`fairmq-shmmonitor --get-shmid --session`
```
davidrohr: But this command lacks the actual session name?

chiarazampolli: You are right, and it is even wrong. Testing did not spot it. I will fix it.

chiarazampolli: Should be fixed now.

PR #10938, opened 2023-03-13
Possibility to use CCDBDownloader in CcdbApi

sawenzel:
This commit introduces a new function CURL_perform in CcdbApi, with the purpose to have a single place where we can take the decision to dispatch CURL perform requests to "curl_easy_perform" or the new "CCDBDownloader".

The most important curl_easy_perform have been replaced to go via this new function.

In turn, with this commit we can make experimental use of CCDBDownloader within CcdbApi whenever an environment variable ALICEO2_ENABLE_MULTIHANDLE_CCDBAPI is defined. (Later on we can make this the default route). This new multi-handle route is tested via new unit test.

Further changes are:
* clarification in doxygen documentation
* some optimization by using const& instead of vector copy

### DIFF
--- a/CCDB/src/CcdbApi.cxx
+++ b/CCDB/src/CcdbApi.cxx
@@ -286,7 +286,7 @@ int CcdbApi::storeAsBinaryFile(const char* buffer, size_t size, const std::strin
       curl_easy_setopt(curl, CURLOPT_URL, fullUrl.c_str());
 
       /* Perform the request, res will get the return code */
-      res = curl_easy_perform(curl);
+      res = CURL_perform(curl);
       /* Check for errors */
       if (res != CURLE_OK) {
         LOGP(alarm, "curl_easy_perform() failed: {}", curl_easy_strerror(res));
@@ -566,7 +566,7 @@ bool CcdbApi::receiveObject(void* dataHolder, std::string const& path, std::map<
       string fullUrl = getFullUrlForRetrieval(curlHandle, path, metadata, timestamp, hostIndex);
       curl_easy_setopt(curlHandle, CURLOPT_URL, fullUrl.c_str());
 
-      curlResultCode = curl_easy_perform(curlHandle);
+      curlResultCode = CURL_perform(curlHandle);
 
       if (curlResultCode != CURLE_OK) {
         LOGP(alarm, "curl_easy_perform() failed: {}", curl_easy_strerror(curlResultCode));
@@ -854,7 +854,7 @@ void* CcdbApi::navigateURLsAndRetrieveContent(CURL* curl_handle, std::string con
 
   curlSetSSLOptions(curl_handle);
 
-  auto res = curl_easy_perform(curl_handle);
+  auto res = CURL_perform(curl_handle);
   long response_code = -1;
   void* content = nullptr;
   bool errorflag = false;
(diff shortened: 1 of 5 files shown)


sawenzel:
Benefit can be seen immediately from `strace`:

```
# current production case:
strace -e socket,connect,close,write,read ./stage/tests/o2-test-ccdb-CcdbApi | grep "connect" | wc # --> getting 81
# new experimental case:
export ALICEO2_ENABLE_MULTIHANDLE_CCDBAPI=1
strace -e socket,connect,close,write,read ./stage/tests/o2-test-ccdb-CcdbApi | grep "connect" | wc # --> getting 4
```

sawenzel:
@davidrohr

CCDB/src/CcdbApi.cxx
```diff
@@ -1726,4 +1726,12 @@ void CcdbApi::logReading(const std::string& path, long ts, const std::map<std::s
   LOGP(info, "ccdb reads {}{}{} for {} ({}, agent_id: {}), ", mUrl, mUrl.back() == '/' ? "" : "/", upath, ts < 0 ? getCurrentTimestamp() : ts, comment, mUniqueAgentID);
 }
 
+CURLcode CcdbApi::CURL_perform(CURL* handle) const
+{
+  if (getenv("ALICEO2_ENABLE_MULTIHANDLE_CCDBAPI")) {
```
davidrohr: Usually I prefer to cache this is an static, and also check that it is set to non 0.
```
static bool enableMultiHandle = getenv("ALICEO2_ENABLE_MULTIHANDLE_CCDBAPI") && atoi(getenv("ALICEO2_ENABLE_MULTIHANDLE_CCDBAPI"));
```

davidrohr:
Looks good to me. Let's get this merged, and then test on the EPNs with "extra env variable". If you want, you can address my comment above beforehand.

costing:
@TrifleMichael for information. Michal, this is needed quickly in the Online environment so it will conditionally be enabled there. What is the default timeout to close the connections and how can we control that?

Please make sure this is not used anywhere else but on EPNs as the Offline servers have a limited number of allowed connections (200 connections with a 10s timeout) so this can potentially create a denial of service.

costing (dismissed):

davidrohr:
@costing : thx for checking.
It is steered by an ENV variable, and the plan at the moment is to enable it only on the EPNs for sync processing.
Once it is fully validated and we all timeouts under controll, we can discuss to enable it by default.

CCDB/src/CcdbApi.cxx
sawenzel: sounds reasonable. Will fix asap.

sawenzel:
> @TrifleMichael for information. Michal, this is needed quickly in the Online environment so it will conditionally be enabled there. What is the default timeout to close the connections and how can we control that?

There is a function we can call on the downloader object. We could easily configure the timeout via env variables which we access in the constructor of CcdbApi until we find a reasonable value.

TrifleMichael:
@costing I'm sorry for the late reply. The default timeout is defined in CCDBDownloader.h with the variable mSocketTimeoutMS, which is set currently at 4 seconds. The value can be changed either in the header or by calling the setSocketTimoutTime() method, after an instance of CCDBDownloader is created.

PR #10941, opened 2023-03-13
MCH: read number of orbits in TF from CCDB

aphecetche:
@aferrero2707 here are the mods to read the number of orbits per TF from the CCDB. 
I’ve tested briefly from a TF of last year (and I indeed retrieve 128 orbits per TF). I did not find any TF for this year on alien though, so could not try on recent data. Would it be possible to have a few TFs written for a run containing MCH ?


aferrero2707:
Yes, we can collect some TF with 32 orbits, hopefully they will already come today.

davidrohr:
thx for this PR! Please let me know once you tesated with 32 orbits, so we can merge

davidrohr (approved):

aphecetche:
@aferrero2707 I just tested on a standalone MCH run taken yesterday evening (thanks @csuire), 532688, and I do get the expected : 
```
[42258:mch-data-decoder]: [10:44:50][INFO] Setting number of orbits per TF to 32
```
which tells the information was correctly read from the CCDB.

PR #10944, opened 2023-03-14
TRD config events implementation

bazinski:
TRD trap configs.

This is just the storage portion.


pachmay (dismissed):
Thanks, can you still please fix the clang format etc issues

f3sch:
Could we mark this as draft so we all do not get notified until the code is ready?

bazinski:
ok not draft anymore the files that crept in from the other branch are removed.

pachmay (dismissed):
thanks for modifying

DataFormats/Detectors/TRD/CMakeLists.txt
```diff
@@ -64,3 +66,12 @@ o2_add_test(RawData
...
 
+o2_add_test(TrapConfig
+            COMPONENT_NAME trd
+            PUBLIC_LINK_LIBRARIES O2::DataFormatsTRD
+            SOURCES test/testTrapConfig.cxx
+            ENVIRONMENT O2_ROOT=${CMAKE_BINARY_DIR}/stage
+            LABELS trd
+)
+
+
```
f3sch, line 77: delete

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
```diff
@@ -156,6 +156,20 @@ struct HelperMethods {
     int sector = hcid / constants::NHCPERSEC;
     return getORIinSuperModule(hcid) + constants::NHCPERSEC * sector;
   }
+
+  static int getMCMId(int sector, int stack, int layer, int rob, int mcm)
+  {
+    // return the mcmid of the indexed mcm
+    int mcmid = (sector * constants::NSTACK * constants::NLAYER + stack * constants::NLAYER + layer) * constants::NROBC1 * constants::NMCMROB + rob * constants::NMCMROB + mcm;
```
f3sch, line 163: We already have this nice getDetector function here, why not use it?
```suggestion
    int mcmid = getDetector(sector, stack, layer) * constants::NROBC1 * constants::NMCMROB + rob * constants::NMCMROB + mcm;
```

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+#include "DataFormatsTRD/Constants.h"
+
+#include <string>
+#include <map>
+#include <bitset>
+
+namespace o2::trd
+{
+
+static const int kTrapRegistersSize = 151; // size of our compressed trap registers in units of 32 bits
```
f3sch, line 27: Shouldn't this be in Constants.h?
Also maybe constexpr it.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+
+// enum of all TRAP registers, to be used for access to them
+// registers in the order they come in.
+enum TrapReg {
+
```
f3sch, line 32: delete newline

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+  std::string getName() { return mName; }
+  unsigned short getAddr() { return mAddr; }
+  unsigned short getNbits() { return mNbits; }
+  unsigned int getBase() { return mBase; }
+  unsigned int getWordNumber() { return mWordNumber; }
+  unsigned int getDataWordNumber() { return mDataWordNumber; }
+  unsigned int getShift() { return mShift; }
+  uint32_t getMask() { return mMask; }
+  uint32_t getMax() { return mMax; }
+  bool getIgnoreChange() { return mIgnoreChange; }
```
f3sch, line 491: Generally, it would feel right to mark all getters as const.
Also in TrapConfig.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+  kC10CPU3,
+  kC11CPU3,
+  kNES,
+  kNTP,
+  kNCUT,
+  kPASACHM,
+  kLastReg
+};
+
+class TrapRegInfo
```
f3sch, line 469: Maybe move TrapReg(Info) to its own header, feels cleaner...

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+  int32_t getRegIndexByAddr(unsigned int addr);
+  int32_t getRegIndexByName(const std::string& name);
+
+  // get a registers address by index or name
+  int32_t getRegAddrByIdx(unsigned int regidx);
+  int32_t getRegAddrByName(const std::string& name);
+
+  // retrieve the trap register info for a particular register, no need to index into det/rob/mcm as these are constant over a config
+  o2::trd::TrapRegInfo* getTrapRegInfoByAddr(uint32_t addr);
+  o2::trd::TrapRegInfo* getTrapRegInfoByIdx(uint32_t idx);
```
f3sch, line 564: Feels like an antipattern to return a pointer to a class member because of semantics like 'who owns the memory afterward'

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+ private:
+  std::array<o2::trd::TrapRegInfo, kLastReg> mTrapRegisters;                              //!< store of layout of each block of mTrapRegisterSize, populated via initialiseRegisters
+  std::bitset<o2::trd::constants::MAXMCMCOUNT> mMCMDataHasChanged;                        //!< does the mcm actually contain data.
+  std::array<uint32_t, kTrapRegistersSize * o2::trd::constants::MAXMCMCOUNT> mConfigData; //!< one block of data per mcm.
+  std::map<uint16_t, uint16_t> mTrapRegistersAddressIndexMap;                             //!< map of address into mTrapRegisters, populated at the end of initialiseRegisters
+  void initialiseRegisters();
+
+  uint32_t mTrapConfigNumber;       //!< the version number coming from the config that is written to the traps, from config register C09CPU01
+  uint32_t mTrapConfigVersion;      //!< the version of the coming from the config that is written to the traps, from DigitHCHeader (svn version number)
+  uint16_t mTrapConfigSavedVersion; //!< the version that is saved, for the ability to later save the config differently.
```
f3sch, line 614: I might be misremembering this, but wasnt there talk to convert the svn repo to git?

f3sch (changes requested):

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: That is a whole nother discussion. This is just reflecting the current state of the outside code.

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
bazinski: forgot about it

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: Should migrate into the class, its a constant in so far as the packing does not change, and is effectively linked to the mTrapConfigSavedVersion

martenole:
Could you please explain how we can get this in production? Would you like to have this merged and then add the actual parsing? But then what about the currently implemented TrapConfig? Why do you call this here TrapConfig3 actually?
In general I would prefer to add in one go together with the actual parsing. But if you want we can also do it in multiple steps. Its just not really clear to me what all the steps are

bazinski:
Yes I will put all the other stuff in the following pr, parsing/trapsim/etc. , I am trying to finalise the qc problem, will come back to this later today.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+  // which do not need to be stored on a per mcm basis
+  std::string mName;        //!< Name of the register
+  uint16_t mAddr;           //!< Address in GIO of TRAP
+  uint16_t mNbits;          //!< Number of bits, from 1 to 32
+  uint32_t mBase;           //!< base of this registers block, i.e. TFL?? will have 0 and is in the range [0,kTrapRegistersSize]
+  uint32_t mWordNumber;     //!< word number offset, of size Nbits, in the block of registers
+  uint32_t mDataWordNumber; //!< offset, into "compressed" 32 bit words for the 32 bit word containing this register, offset from the base;
+  uint32_t mMask;           //!< mask to extract register from the word identified by WordNumber
+  uint32_t mMax;            //!< max is not the same as the mask, some values come in as 15 bit mask but are only 12 or 13 bits for max, IRQ values for example
+  uint32_t mShift;          //!< shift to extract the register
```
tdietel, line 520: I cannot understand from these comments how the config event data structure looks like and how these register descriptions link to it.
It would be good to link to some more documentation that defines the meaning of all of these pieces of information.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+  bool mIgnoreChange;       //!< we are not concerned with this value changing for purposes of the differential comparison.
+  ClassDefNV(TrapRegInfo, 1);
+};
+
+class TrapConfig3
+{
+  // class that is actually stored in ccdb.
+  // it holds a compressed version of what comes in the config events
+ public:
```
tdietel, line 529: The public interface feels very big. Are all the getters/setters actually used?

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
```diff
@@ -156,6 +156,20 @@ struct HelperMethods {
     int sector = hcid / constants::NHCPERSEC;
     return getORIinSuperModule(hcid) + constants::NHCPERSEC * sector;
   }
+
+  static int getMCMId(int sector, int stack, int layer, int rob, int mcm)
```
tdietel, line 160: Ole has a similar function to calculate a global ADC channel index. That could be made simpler by using this MCM index.
Ole's function actually avoids that gap in C0 chambers, where we only have 96 instead of 128 MCMs that send data.

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
tdietel: In this definition, I do not see space for the board merger and half-chamber merger MCMs. While they don't process signals, their settings can indicate if certain MCMs are excluded from the readout tree.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+  bool setRegisterValueByAddr(uint32_t data, uint32_t addr, int mcmidx);
+  uint32_t getRegisterValueByName(const std::string& name, int mcmidx);
+  bool setRegisterValueByName(uint32_t data, const std::string& regname, int mcmidx);
+
+  // get a config register value by index, addr, and name, via sector/stack/layer/rob/mcm
+  // no setters for sector/stack/layer/rob/mcm as writing, we only write from retrieved config events.
+  uint32_t getRegisterValueByIdx(uint32_t regix, int sector, int stack, int layer, int rob, int mcm);
+  uint32_t getRegisterValueByAddr(uint32_t regaddr, int sector, int stack, int layer, int rob, int mcm);
+  uint32_t getRegisterValueByIdx(uint32_t regix, int detector, int rob, int mcm);
+  uint32_t getRegisterValueByAddr(uint32_t regaddr, int detector, int rob, int mcm);
```
tdietel, line 548: It seems these methods could be removed, and the code that uses them could be written as `getRegister...(regix, getMCMId(sector, stack, layer, rob, mcm))`

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+
+  // get a config register value by index, addr, and name, via mcm index
+  uint32_t getRegisterValue(uint32_t regidx, int mcmidx);
+  bool setRegisterValue(uint32_t data, uint32_t regidx, int mcmidx);
+  uint32_t getRegisterValueByIdx(uint32_t regidx, int mcmidx);
+  bool setRegisterValueByIdx(uint32_t data, uint32_t regidx, int mcmidx);
+  uint32_t getRegisterValueByAddr(uint32_t addr, int mcmidx);
+  bool setRegisterValueByAddr(uint32_t data, uint32_t addr, int mcmidx);
+  uint32_t getRegisterValueByName(const std::string& name, int mcmidx);
+  bool setRegisterValueByName(uint32_t data, const std::string& regname, int mcmidx);
```
tdietel, line 541: Are all of these methods actually used? I would assume that the user should always refer to the registers using the enum values above (e.g. kADCMSK), and the name and addresses in MCM GIO address space are mostly a convencience for the reader.

Also, what is the difference between getRegisterValue() and getRegisterValueByIdx()?

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+  void getAllMCMByName(std::string registername, std::array<uint32_t, o2::trd::constants::MAXMCMCOUNT>& mcms);
+
+  // return all the mcm for a particular register address
+  void getAllMCMByAddress(int registeraddress, std::array<uint32_t, o2::trd::constants::MAXMCMCOUNT>& mcms);
+
+  // return all the config data in an unpacked array.
+  void getAll(std::array<uint32_t, kLastReg * o2::trd::constants::MAXMCMCOUNT>& configdata);
+
+  // return the full unpacked config event.
+  bool printRegister(int addr, int det, int rob, int mcm);
```
tdietel, line 585: I think you would want to call this with a register index, i.e.:
```
tc.printRegister(kADCMSK, getMCMId(det,rob,mcm));
```
In that case, a single method would be sufficient.

DataFormats/Detectors/TRD/src/TrapConfig3.cxx
```diff
@@ -0,0 +1,818 @@
...
+void TrapConfig3::getAll(std::array<uint32_t, kLastReg * o2::trd::constants::MAXMCMCOUNT>& configdata)
+{
+  for (int mcm = 0; mcm < constants::MAXMCMCOUNT; ++mcm) {
+    for (int reg = 0; reg < kLastReg; ++reg) {
+      configdata[mcm * kLastReg + reg] = getRegisterValueByIdx(reg, mcm);
+    }
+  }
+}
+
+bool TrapConfig3::printRegister(TrapRegInfo* reg, int det, int rob, int mcm)
```
tdietel, line 712: This doesn't seem to do anything?

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+
+  // for compliance with the same interface to o2::trd::TrapConfig and its run1/2 inherited interface:
+  // only these 2 are used in the simulations.
+  uint32_t getDmemUnsigned(uint32_t address, int detector, int rob, int mcm);
+  uint32_t getTrapReg(uint32_t index, int detector, int rob, int mcm);
+
+ private:
+  std::array<o2::trd::TrapRegInfo, kLastReg> mTrapRegisters;                              //!< store of layout of each block of mTrapRegisterSize, populated via initialiseRegisters
+  std::bitset<o2::trd::constants::MAXMCMCOUNT> mMCMDataHasChanged;                        //!< does the mcm actually contain data.
+  std::array<uint32_t, kTrapRegistersSize * o2::trd::constants::MAXMCMCOUNT> mConfigData; //!< one block of data per mcm.
```
tdietel, line 608: Wouldn't an array of arrays make the indexing simpler.
Also note that as discussed elsewhere, this seems to reserve data for non-existent MCMs (C0 chambers in stack 2 with only 6 ROBs). It also seems that BM/HCM MCMs are not stored at all.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
```diff
@@ -0,0 +1,619 @@
...
+
+  const std::array<o2::trd::TrapRegInfo, kLastReg>& getTrapRegisters() { return mTrapRegisters; }
+
+  // population pending
+  void setConfigVersion(uint32_t version) { mTrapConfigVersion = version; }           // these must some how be gotten from git or wingdb.
+  void setConfigNumber(uint32_t number) { mTrapConfigNumber = number; }               // these must be gotten from git or wingdb.
+  void setConfigSavedVersion(uint16_t version) { mTrapConfigSavedVersion = version; } // the version that is saved, for the ability to later save the config differently.
+  uint32_t getConfigVersion() { return mTrapConfigVersion; }                          // these must some how be gotten from git or wingdb.
+  uint32_t getConfigName() { return mTrapConfigNumber; }                              // these must be gotten from git or wingdb.
+  uint16_t getConfigSavedVersion() { return mTrapConfigSavedVersion; }                // the version that is saved, for the ability to later save the config differently.
```
tdietel, line 596: The config version and number can technically differ for different chambers. They should therefore not be a global property of the TRAPconfig. This feature led to the mixed configs in run 2, and should be avoided.
Also, the values should be retrieved from the registers for all the MCMs - at that point, you can check if they all agree, or if you have a mixed config.
What needs a lookup in git or wingDB is to translate the config number to the full name of the config, and the version to a git commit (at the moment it's just the svn release number).

tdietel (commented):
I have a number of comments. At the moment, I am still not sure how the TRAPconfig3 will be filled, nor how it will actually be used.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: Will put in some ascii art, and it will be in that other bigger document more prettily

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: I just put in all the options, the all reduce to access by index, in a singular function.

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
bazinski: Chatted to him, will do.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: It was there for debugging, done differently now, removed.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: Should actually be inside the trapconfig class.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: why does clang not pick this up?

DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
bazinski: Took a while to verify that, but now that I cant find them, it seems rather obvious why :-( they dont produce digit data and hence of course not in the data stream.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: I thought mixed config was an error condition, that is handled on parsing and creating this class. Are you saying that we must allow for mixed configs? so configversion and name will be a vector of chambers long?

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: BM/HCM comment above. Yes array of array would be simpler, and was the reason I asked about this in another PR a while back, have not changed it yet.

DataFormats/Detectors/TRD/src/TrapConfig3.cxx
bazinski: it was used for debugging purposes only.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: sure

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
tdietel: I think here we have to clarify what we want to do with the TRAPconfigs. And right now I can think of three things:
- Make sure we get the correct settings into TRAPconfig. For this, mixed configs should be used as mixed configs, i.e. each MCM might have different settings.
- Issue a warning and mark the run when the config is mixed. This should be part of QC.
- Provide information what FEE config is/was active for a run. This should include the majority config, and which MCMs differed and what how their config was different.

bazinski: 1. MCM can have different settings within a single upload, mixed configs (as I understand it) is different configs (from wingdb) loaded into different MCM. 
2. Yes this is done on parsing and triggers a new object in CCDB, and additionally could send a flag to qc to react on but qc is only listening to 0.5% of the data so will almost surely miss the flagging. So would have to test for a new object in ccdb in anycase.
3. Well yes, differing objects are written to the ccdb, and a script or calling the class to figure out where and what differs.

Maybe a chat sometime soon about this. This pr will get closed soon as in ref to an earlier comment by Ole above, there is no reason to keep trapconfig and trapconfig3, it will all get changed at the same time.

martenole:
Just to clarify: we can keep two versions of the TrapConfig class in 
parallel for a short time, it does not have to be moved completely in 
one PR. But what I would like to see is how the new TrapConfig class 
actually is used and this should be included here. Otherwise we have no 
way to review your PR because we don't know how the functions are 
supposed to be used. There is already quite some discussion here and 
seeing the usage implementation would clarify some things IMO

pachmay:
I think the problem with the FED configurations that we had in run 2, we should catch very early on to avoid the problem. In the past we only had a pop up window for the TRD panel in case it was open. We should have this as a clear error in DCS at the beginning of the run. It is of course good to know for the reco and simulation if a given chamber has a different configuration but we will not always be able to simulate the corresponding condition and thus will have to mask the chamber as off.

DataFormats/Detectors/TRD/include/DataFormatsTRD/TrapConfig3.h
bazinski: done.

PR #10946, opened 2023-03-14
[FOCAL-10] Add DetID for FOCAL

mfasDa:


mfasDa:
As discussed in https://github.com/AliceO2Group/AliceO2/pull/10469 code for upgrades is not built in case of o2-dataflow, consequently whatever is under ENABLE_UPGRADES cannot be used on the FLP. Since FOCAL already uses the FLP to readout FOCAL-E we would like to have the code available on the FLP, which consequently requires FOCAL to enter under the list of regular detectors.

shahor02 (approved):
Looks fine. @mconcas note that if there are upgrade GRPs generated of MC data reconstructed for current DetID layout, this PR will invalidate them.

mconcas:
> Looks fine. @mconcas note that if there are upgrade GRPs generated of MC data reconstructed for current DetID layout, this PR will invalidate them.

Ok, thanks, aknowledged. No issues in this respect.

mfasDa:
CI error unrelated

PR #10948, opened 2023-03-14
TransitionState getters in ProcessingContext

ktf:


ktf:
@shahor02 @chiarazampolli @davidrohr as discussed. It needs to be a method of the context because eventually we want this to not change in the middle of the processing of one timeframe (as it could happen with multithreading).

PR #10951, opened 2023-03-14
GLO QC: Switch to TH1D for pT and extend Chi2 range

jloemker:
I fixed the merge conflicts and reduced the PR to the very important changes (TH1F to TH1D for pT plots and Chi2 range).
The previous issues from setEfficiency are gone as I excluded that part !

### DIFF
--- a/Detectors/GlobalTracking/include/GlobalTracking/MatchITSTPCQC.h
+++ b/Detectors/GlobalTracking/include/GlobalTracking/MatchITSTPCQC.h
@@ -55,8 +55,8 @@ class MatchITSTPCQC
   void finalize();
   void reset();
 
-  TH1F* getHistoPt() const { return mPt; }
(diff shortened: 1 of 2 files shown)


jloemker:
@shahor02 , would you approve it one last time ?
(the previous error can not occur anymore and there are also no more merge conflicts).
Many thanks !

shahor02 (dismissed):

jloemker:
@shahor02, 
I had a look at the Details from the two tests that failed (build/O2/fullCI and build/O2/o2) and there was nothing from the qc-task. I forgot to include the TH1D header (this is why the tests are running again), but I think the fails been caused by something unrelated to this as a missing include should raise an error while building.

shahor02 (dismissed):

shahor02 (approved):

PR #10954, opened 2023-03-15
Drop need for libuv

ktf:


ktf:
@davidrohr @shahor02 this should eliminate the need to expose libuv to root.
@costing @sawenzel there is quite a few places where I am not actually sure CCDBDownloader is not leaking memory. Maybe the code should be further reviewed before it enters production.

davidrohr:
ok, testing this on the EPNs.
Also fixed the compilation of the tests

davidrohr (approved):
I confirm the fix works

PR #10964, opened 2023-03-16
[FOCAL-10] Add missing streamer

mfasDa:
Streamer for new class TriggerWindow for the PAD trigger data was missing in order to stream it as part of the PAD event to the FOCAL event tree.


mfasDa:
CI errors unrelated

shahor02 (approved):

PR #10967, opened 2023-03-16
[EMCAL-688] Fix size of the lookup table

mfasDa:


mfasDa:
@mpuccio This fixes the crash in the correction task in O2-3450

shahor02 (approved):
trivial, merging

PR #10970, opened 2023-03-17
Catch bad topologies

ktf:


ktf:
@shahor02 this is what we discussed yesterday. It should catch all the cases where we have the same (Lifetime::Timeframe) output twice. Could you test with your larger workflow as well?

shahor02:
@ktf thanks, works as expected:
```
[ERROR] invalid workflow in o2-ft0-flp-dpl-workflow: Found duplicate outputs FT0/DIGITSBC/0 in device tf-reader and ft0-datareader-dpl
```

shahor02:
@ktf but something is wrong: the local FST fails with
```
[ERROR] invalid workflow in o2-gpu-reco-workflow: Found duplicate outputs TPC/CLUSTERNATIVE/36 in device gpu-reconstruction and gpu-reconstruction
```

shahor02:
@ktf now it prints 
```
[ERROR] invalid workflow in o2-gpu-reco-workflow: Found duplicate outputs TPC/CLUSTERNATIVE/36 in device gpu-reconstruction (8) and TPC/CLUSTERNATIVE/36 in gpu-reconstruction (8)
```

ktf:
Actually I think this is a genuine problem with the gpu-reconstruction outputs (which is probably harmless, because duplicate outputs from the same device are not an issue). The issue seems to appear at:

GPU/Workflow/src/GPUWorkflowSpec.cxx:937

and following lines, where `TPC/CLUSTERNATIVE/36` is clearly added twice if `mSpecConfig.sendClustersPerSector` is false.

ktf:
@shahor02 it now complains about ITS/VERTICES/0 in its-tracker. Could you check this one as well?

ktf:
Error on macOS unrelated. @sawenzel @costing I sporadically see now:

```
 98/481 Test  #97: CcdbApi-MultiHandle ..................................................................***Failed    2.82 sec
```

is this an known issue on the backend?

PR #10973, opened 2023-03-17
ZDC data model change 

ddobrigk:
@cortesep @coppedis @jgrosseo up for discussion / approval! Thanks :-D

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -1512,11 +1512,11 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   auto& ambigMFTTracksBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "AMBIGUOUSMFTTR"});
   auto& ambigFwdTracksBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "AMBIGUOUSFWDTR"});
   auto& v0sBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "V0_001"});
-  #ifdef O2_ZDC_NEWDATAMODEL
+#ifdef O2_ZDC_NEWDATAMODEL
   auto& zdcBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "ZDC_001"});
-  #else
+#else
   auto& zdcBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "ZDC"});
-  #endif
+#endif
   auto& caloCellsBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "CALO"});
   auto& caloCellsTRGTableBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "CALOTRIGGER"});
   auto& cpvClustersBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "CPVCLUSTER"});
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -1512,7 +1512,11 @@ void AODProducerWorkflowDPL::run(ProcessingContext& pc)
   auto& ambigMFTTracksBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "AMBIGUOUSMFTTR"});
   auto& ambigFwdTracksBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "AMBIGUOUSFWDTR"});
   auto& v0sBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "V0_001"});
+  #ifdef O2_ZDC_NEWDATAMODEL
+  auto& zdcBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "ZDC_001"});
+  #else
   auto& zdcBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "ZDC"});
+  #endif
   auto& caloCellsBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "CALO"});
   auto& caloCellsTRGTableBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "CALOTRIGGER"});
   auto& cpvClustersBuilder = pc.outputs().make<TableBuilder>(Output{"AOD", "CPVCLUSTER"});
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -2573,7 +2573,11 @@ DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src, bool enableSV, boo
   outputs.emplace_back(OutputLabel{"O2ambiguousMFTtrack"}, "AOD", "AMBIGUOUSMFTTR", 0, Lifetime::Timeframe);
   outputs.emplace_back(OutputLabel{"O2ambiguousFwdtrack"}, "AOD", "AMBIGUOUSFWDTR", 0, Lifetime::Timeframe);
   outputs.emplace_back(OutputLabel{"O2v0_001"}, "AOD", "V0_001", 0, Lifetime::Timeframe);
+#ifdef O2_ZDC_NEWDATAMODEL
+  outputs.emplace_back(OutputLabel{"O2zdc_001"}, "AOD", "ZDC_001", 0, Lifetime::Timeframe);
+#else
   outputs.emplace_back(OutputLabel{"O2zdc"}, "AOD", "ZDC", 0, Lifetime::Timeframe);
+#endif
   outputs.emplace_back(OutputLabel{"O2caloCell"}, "AOD", "CALO", 0, Lifetime::Timeframe);
   outputs.emplace_back(OutputLabel{"O2caloCellTRGR"}, "AOD", "CALOTRIGGER", 0, Lifetime::Timeframe);
   outputs.emplace_back(OutputLabel{"O2cpvCluster"}, "AOD", "CPVCLUSTER", 0, Lifetime::Timeframe);
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -2574,7 +2574,7 @@ DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src, bool enableSV, boo
   outputs.emplace_back(OutputLabel{"O2ambiguousFwdtrack"}, "AOD", "AMBIGUOUSFWDTR", 0, Lifetime::Timeframe);
   outputs.emplace_back(OutputLabel{"O2v0_001"}, "AOD", "V0_001", 0, Lifetime::Timeframe);
 #ifdef O2_ZDC_NEWDATAMODEL
-  outputs.emplace_back(OutputLabel{"O2zdc_001"}, "AOD", "ZDC_001", 0, Lifetime::Timeframe);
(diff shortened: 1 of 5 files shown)


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -2522,7 +2573,11 @@ DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src, bool enableSV, boo
   outputs.emplace_back(OutputLabel{"O2ambiguousMFTtrack"}, "AOD", "AMBIGUOUSMFTTR", 0, Lifetime::Timeframe);
   outputs.emplace_back(OutputLabel{"O2ambiguousFwdtrack"}, "AOD", "AMBIGUOUSFWDTR", 0, Lifetime::Timeframe);
   outputs.emplace_back(OutputLabel{"O2v0_001"}, "AOD", "V0_001", 0, Lifetime::Timeframe);
+#ifdef O2_ZDC_NEWDATAMODEL
+  outputs.emplace_back(OutputLabel{"O2zdc_001"}, "AOD", "ZDC_001", 0, Lifetime::Timeframe);
```
aalkin: This is incorrect, it should be `"ZDC", 1`, the version for AOD tables is encoded in subspec. 
Update: the same goes for V0 table above, the only reason this works is that AOD producer's output is not used directly by tasks.

jgrosseo: OK, this is wrong for all of them. See v0, cascade, collision etc. So it seems to work but is not good of course...

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -708,9 +711,8 @@ DECLARE_SOA_TABLE(CPVClusters, "AOD", "CPVCLUSTER", //! CPV clusters
                   cpvcluster::ModuleNumber<cpvcluster::ClusterStatus>, cpvcluster::IsUnfolded<cpvcluster::ClusterStatus>);
 using CPVCluster = CPVClusters::iterator;
 
-namespace zdc
+namespace zdclegacy // needed to have the new dynamic columns with the same name
```
```suggestion
namespace zdc_000 // needed to have the new dynamic columns with the same name
```

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -727,13 +729,319 @@ DECLARE_SOA_COLUMN(TimeZNA, timeZNA, float);                    //!
...
+{
+DECLARE_SOA_INDEX_COLUMN(BC, bc); //! BC index
+// New summarized table, minimal disk footprint, per channel like other detectors
+DECLARE_SOA_COLUMN(Energy, energy, std::vector<float>);       //! Energy of non-zero channels. The channel IDs are given in ChannelE (at the same index)
+DECLARE_SOA_COLUMN(ChannelE, channelE, std::vector<uint8_t>); //! Channel IDs which have reconstructed energy. There are at maximum 26 channels.
+DECLARE_SOA_COLUMN(Amplitude, amplitude, std::vector<float>); //! Amplitudes of non-zero channels. The channel IDs are given in ChannelT (at the same index)
+DECLARE_SOA_COLUMN(Time, time, std::vector<float>);           //! Times of non-zero channels. The channel IDs are given in ChannelT (at the same index)
+DECLARE_SOA_COLUMN(ChannelT, channelT, std::vector<uint8_t>); //! Channel IDs which had non-zero amplitudes. There are at maximum 26 channels.
+// Dynamic columns to take into account packed information; replace old getters
+DECLARE_SOA_DYNAMIC_COLUMN(EnergyZEM1, energyZEM1, //!
```
jgrosseo: Can we add comments for all those as well?

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -727,13 +729,319 @@ DECLARE_SOA_COLUMN(TimeZNA, timeZNA, float);                    //!
...
+                             return thisamplitude;
+                           });
 } // namespace zdc
 
-DECLARE_SOA_TABLE(Zdcs, "AOD", "ZDC", //! ZDC information
-                  o2::soa::Index<>, zdc::BCId, zdc::EnergyZEM1, zdc::EnergyZEM2,
-                  zdc::EnergyCommonZNA, zdc::EnergyCommonZNC, zdc::EnergyCommonZPA, zdc::EnergyCommonZPC,
-                  zdc::EnergySectorZNA, zdc::EnergySectorZNC, zdc::EnergySectorZPA, zdc::EnergySectorZPC,
-                  zdc::TimeZEM1, zdc::TimeZEM2, zdc::TimeZNA, zdc::TimeZNC, zdc::TimeZPA, zdc::TimeZPC);
+DECLARE_SOA_TABLE(Zdcs_000, "AOD", "ZDC", //! ZDC information
```
jgrosseo, line 1018:
```suggestion
DECLARE_SOA_TABLE(Zdcs_000, "AOD", "ZDC", //! ZDC information, version 0
```

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -727,13 +729,319 @@ DECLARE_SOA_COLUMN(TimeZNA, timeZNA, float);                    //!
...
+                  o2::soa::Index<>, zdc::BCId, zdclegacy::EnergyZEM1, zdclegacy::EnergyZEM2,
+                  zdclegacy::EnergyCommonZNA, zdclegacy::EnergyCommonZNC, zdclegacy::EnergyCommonZPA, zdclegacy::EnergyCommonZPC,
+                  zdclegacy::EnergySectorZNA, zdclegacy::EnergySectorZNC, zdclegacy::EnergySectorZPA, zdclegacy::EnergySectorZPC,
+                  zdclegacy::TimeZEM1, zdclegacy::TimeZEM2, zdclegacy::TimeZNA, zdclegacy::TimeZNC, zdclegacy::TimeZPA, zdclegacy::TimeZPC);
+
+DECLARE_SOA_TABLE_VERSIONED(Zdcs_001, "AOD", "ZDC", 1, //! Full ZDC information, std::vector format
```
```suggestion
DECLARE_SOA_TABLE_VERSIONED(Zdcs_001, "AOD", "ZDC", 1, //! ZDC information, version 1, std::vector format
```

jgrosseo (commented):
I have posted a few comments for cosmetic changes but the rest is fine with me

Framework/Core/include/Framework/AnalysisDataModel.h
ddobrigk: thanks, taken :-)

Framework/Core/include/Framework/AnalysisDataModel.h
ddobrigk: yes let me do that now ...

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
ddobrigk: Shall we change this everywhere? Or we keep doing as is? What do you think?

aalkin: Probably better to do the rest in a separate PR, for bookkeeping reasons.

Framework/Core/include/Framework/AnalysisDataModel.h
ddobrigk: Added, thanks!

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
ddobrigk: You mean the rest or all of them? What do you think? :-D

aalkin: ZDC should be done correctly from the start here, I think. The rest can be fixed elsewhere so we do not mix them with ZDC-specific change.

coppedis: So we include in this PR all the fixes for the ZDC and then another request will be open for the remaining issues, correct?

ddobrigk: Okay, the latest PR does as you suggested, @aalkin ! Can you please check if I did no do anything major wrong? (technically, though, it's in the part of the producer that is as-of-yet disabled, but I want to be sure it's ok already). If this is fine I would suggest we go ahead and merge so I can do a PR for the O2Physics part. Thanks!!!

ddobrigk:
Marked as ready to review to trigger all checks already (while also waiting for @aalkin 's green light). Thanks!

jgrosseo (dismissed):

jgrosseo:
Is the PR for the converter already there (as draft)?

ddobrigk:
> Is the PR for the converter already there (as draft)?

I don't think I PR-ed it but I have it in [my fork here](https://github.com/AliceO2Group/O2Physics/compare/master...ddobrigk:O2Physics:zdcdevel). I will do a PR as soon as this is merged; anyway now I see some CI errors that I have to look into for O2Physics. More soon...

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -728,12 +731,317 @@ DECLARE_SOA_COLUMN(TimeZNC, timeZNC, float);                    //!
 DECLARE_SOA_COLUMN(TimeZPA, timeZPA, float);                    //!
 DECLARE_SOA_COLUMN(TimeZPC, timeZPC, float);                    //!
 } // namespace zdc
+namespace zdc_001 // Revised table, required for dynamic in-place replacements
```
aalkin: Please use the same namespace, the versioning should only be done at the table level, otherwise you will have to rename the namespace later.

ddobrigk: Hi Anton, as explained in the comment inside the code, I need to have different namespaces to have different types of columns with the same names: we are swapping out a regular column with a dynamic column in the new version of the table, and these have the same exact name. This cannot be done keeping the same namespace, unfortunately...

aalkin: Since the columns will have different type, they may as well have a different name. In fact, you can rename old columns (only the type names, not the getters). This will be transparent for the users, since they are only using the getters.

ddobrigk: The latest commit should take care: it uses only the `zdc` namespace but changes the column _names_ for the dynamic columns while keeping the same _getters_. Can you please take a look @aalkin ? Thanks a lot!

ddobrigk:
Okay, so the latest CI errors seem unrelated and the code does compile in the regular tests... Now I would say we just need a confirmation with @aalkin to see if everything is in order and then we can merge and proceed to the converter part: I will check if there is maybe any small update that may be required on that side and prepare a PR. Thanks a lot, everyone!!

aalkin:
I have no additional comments, @ddobrigk. Although you may want to produce some new format data files first and test them with O2Physics.

ddobrigk:
> I have no additional comments, @ddobrigk. Although you may want to produce some new format data files first and test them with O2Physics.

Yes, this is a good idea. We did it already with the previous version, but we can redo it. For now, because the data model changes will only kick in after the `define` is uncommented (or, better, the code cleaned up from the previous versions), I suggest we merge and then do the test simultaneously; then the necessary condition to move to the new table in O2 (after also the O2Physics converter PR is merged) is to have the latest code tested.

jgrosseo (approved):

jgrosseo:
I think the CI is unrelated. @ktf can you confirm and then merge?

ddobrigk:
All tests are successful and we should really move on - @shahor02 or @ktf , could you please help us out? I think this is ready to go and agreed upon. Thanks!

shahor02:
I'll run the fullCI locally, will merge if ok.

shahor02:
passed fullCI locally, merging

ddobrigk:
Thank you very much @shahor02 ! I will prepare the O2Physics PR shortly. Have a nice easter break!

PR #10974, opened 2023-03-17
Implement TRFs to choose from for TRD simulation

pestratm:
Add different Time Response Functions to choose from for TRD simulations
Original O2 TRF, TRD TDR TRF, no TRF, Landau distribution TRF (with variable parameters)
@pachmay @tdietel


Detectors/TRD/simulation/src/SimParam.cxx
```diff
@@ -26,6 +26,9 @@ using namespace o2::trd;
 SimParam::SimParam()
 {
   mGasMixture = TRDSimParams::Instance().gas;
+  mTRF = TRDSimParams::Instance().trf;
+  mMu = TRDSimParams::Instance().trf_landau_mu;
+  mSigma = TRDSimParams::Instance().trf_landau_sigma;
```
f3sch, line 31: Maybe we could move these to an initializer list?

Detectors/TRD/simulation/include/TRDSimulation/TRDSimParams.h
```diff
@@ -37,6 +37,10 @@ struct TRDSimParams : public o2::conf::ConfigurableParamHelper<TRDSimParams> {
   float maxMCStepSize = 0.1;                              ///< maximum size of MC steps
   bool doTR = true;                                       ///< switch for transition radiation
   SimParam::GasMixture gas = SimParam::GasMixture::Xenon; ///< the gas mixture in the TRD
+  // TRF parameters
+  int trf = 0;                  // Sampled TRF function. 0: default TRF, 1: TRF described in TRF TDR, 2: No TRF, 3: Landau dist as TRF (parameters specified below)
```
f3sch, line 41: Maybe could include also an enum for this? Like with the GasMixture:
```
enum class TRF: int {
 Sampled = 0,
 TDR,
 Delta,
 Landau,

 // Dont add
 SIZE,
 Default = Sampled
};
```

I think this would make it more verbose than just using this integer but this probably personal taste :smile:

f3sch (commented):
LGTM, just some small nitpicks.

pachmay (approved):
Hi Peter, thanks for the commit. You could still consider the enum suggested by Felix, but otherwise all good.

PR #10978, opened 2023-03-17
ITS-tracking: Fix missing memory clear in case of cosmics + dangerous resize

mconcas:


mconcas:
Hi @shahor02, 
as discussed, we fixed the reset of the vertices access table in case the seeding vertexer is not run (e.g. cosmics).

In the meanwhile, I also spotted a bug in the clearing of the buffer of found vertices. This was not expected to happen during the `TimeFrame::initialise` call done by the tracker, but only by vertexer.
The expected behaviour is now implemented.

mconcas:
@shahor02: HI, I don't know about the hanging CI, but I think this can be merged.

shahor02 (approved):

PR #10981, opened 2023-03-20
store ambiguous tracks only for tracks without collision

jgrosseo:


jgrosseo:
See today's WP14 and JIRA: https://alice.its.cern.ch/jira/browse/O2-3601

PR #10982, opened 2023-03-20
TRD noise calibration update

martenole:
No need for complicated workflow using aggregator, we can simply do a run with a single EPN only. In order to test locally with a few raw TFs:

```
export ARGS_ALL="--session default --severity info --shm-segment-id 0 --shm-segment-size 32000000000  --early-forward-policy noraw --fairmq-rate-logging 0 --timeframes-rate-limit 2 --timeframes-rate-limit-ipcid 0"

o2-raw-tf-reader-workflow $ARGS_ALL --input-data list.txt --onlyDet TRD | o2-trd-datareader $ARGS_ALL | o2-calibration-trd-workflow $ARGS_ALL --noise --configKeyValues="TRDCalibParams.minNumberOfDigits=46458460" | o2-calibration-ccdb-populator-workflow $ARGS_ALL --ccdb-path "http://ccdb-test.cern.ch:8080"
```

What is not yet included is the possibility to compare the results from 2 different noise runs. This Archita will look into.

Example for noisy channel:
channel_143891

Mean ADC values for all channels of chamber 0 with MCM grid overlay:
adcMeanChamber0
One can see the one masked channel on the very left and two masked channels on the very right with a mean ADC value of 10 (and not visible here RMS=0)


f3sch (commented):
Nice work!
My main question would be how can we use this channel info in the other calibrations.
For example in VdriftExB it would probably be a good idea to start excluding the tracklets from these noisy channels from filling the residual histograms.
In tracklets we just have pad col and pad row, so would it be a good idea to exclude tracklets which come from a pad that has a noisy channel?
If so can we integrate this with `PadStatus`?

Detectors/TRD/workflow/include/TRDWorkflow/NoiseCalibSpec.h
```diff
@@ -42,19 +44,67 @@ class TRDNoiseCalibSpec : public o2::framework::Task
...
+      if (!mHaveSentOutput) {
+        LOGP(important, "Enough data received after {} TFs seen, finalizing noise calibration", mNTFsProcessed);
+        mCalibrator.collectChannelInfo();
+        sendOutput(pc.outputs());
+        mHaveSentOutput = true;
+      } else {
+        if ((mNTFsProcessed % 1000) == 0) {
+          LOGP(important, "Not processing anymore. Seen {} TFs in total. Run can be stopped", mNTFsProcessed);
+        }
+      }
```
tdietel, line 69: I am not sure what statistics we want to collect, and if we want to restrict ourselves to a single noise calib object with a fixed size per run. 
At the moment, we require 100 billion digits. With approx. 1 M channels, that's 100k events. That's plenty of statistics, and at several 100Hz readout rate, this should be done in less than 10min.
That's probably a good setting for running online - if a single EPN can handle this.

If we process a run offline, where we have limited statistics, how would this work? We might not have the high statistics on (local) disk.

martenole:
I think the goal is to mask the noisy channels in the MCMs directly. That way we would not even get these tracklets in the first place.
It's not so easy to integrate this with the pad status, because that has only one value per pad and not per channel. Due to the shared pads there are more readout channels than physical readout pads. And I have seen cases where the pad itself has normal values on one channel and not so good values on its shared one (and vice versa I believe).

DataFormats/Detectors/TRD/include/DataFormatsTRD/NoiseCalibration.h
```diff
@@ -27,6 +27,53 @@ namespace trd
...
+  int getFlags() const { return mFlags; }
+
+  void setMean(float mean) { mMean = mean; }
+  void setRMS(float rms) { mRMS = rms; }
+  void setFlags(uint8_t flags) { mFlags = flags; }
+
+ private:
+  float mMean{0.f};
+  float mRMS{0.f};
+  uint8_t mFlags{0};
```
tdietel: When I thought about this struct, I considered to include the number of seen digits for each pad. This adds to the data volume, but might indicate data transfer problems. But it might not be worth it.

Detectors/TRD/calibration/src/CalibratorNoise.cxx
```diff
@@ -14,17 +14,86 @@
...
+        isOK = false;
+      }
+      if (ch.getRMS() > mParams.rmsAdcHigh) {
+        flags |= (1 << ChannelInfo::Flag::HighRMS);
+        isOK = false;
+      }
+      if (ch.getRMS() < mAlmostZero && std::abs(ch.getMean() - mParams.adcMasked) < mAlmostZero) {
+        flags |= (1 << ChannelInfo::Flag::IsMasked);
+        isOK = false;
+      }
```
tdietel: This classification is static, and I wonder what happens when we revisit the criteria for flagging channels?
Relating to Felix' comment about where and how we use the calibration, we might even want different criteria in different use cases.
The data for the classification is always there, so it could be done on the fly.

tdietel (commented):
Thanks a lot, Ole. This is very useful and should get us started with easily taking noise runs.
I have a few minor comments below, but I think it would work as it is.

tdietel:
I have one more comment to add. I noticed a flag for masked channels - but we should note that even if we mask out noisy pads, we would probably not apply those masks for noise runs in order to see if these channels change (recover) over time.

Detectors/TRD/workflow/include/TRDWorkflow/NoiseCalibSpec.h
martenole: But why would we want to create 2 CCDB objects for a single run? I envisage that we have the possibility to start from AliECS a run with "trdnoise" config and then we get as output an up-to-date channel status in the CCDB.
For running offline: I put this in my comment above. You can add `--configKeyValues="TRDCalibParams.minNumberOfDigits=46458460"` which is how I tested it (with 2GB of raw files).

DataFormats/Detectors/TRD/include/DataFormatsTRD/NoiseCalibration.h
martenole: I use one of the flags just to indicate whether at least one digit was seen from one channel. But we could also add a counter there

Detectors/TRD/calibration/src/CalibratorNoise.cxx
martenole: True, currently it does not make sense to have it all. If we would want to go for a very slim object we could keep only the flags and store the settings like what is HighRMS etc (which are also configurable). In the CCDB object which keeps a vector with information for each channel. But since I don't see that we would use this object in the reconstruction but instead only to mask channels and compare to previous objects we shouldn't concentrate on reducing the size of the object

martenole:
> I have one more comment to add. I noticed a flag for masked channels - but we should note that even if we mask out noisy pads, we would probably not apply those masks for noise runs in order to see if these channels change (recover) over time.

I thought these are the channels from which we get a constant ADC value of 10? These are the ones for which I set the "masked" flag

f3sch:
> I think the goal is to mask the noisy channels in the MCMs directly. That way we would not even get these tracklets in the first place.

Good point, so probably skipping them here https://github.com/AliceO2Group/AliceO2/blob/99a50a1aa0004a60b129b02f9a342b5a270eb9fc/Detectors/TRD/reconstruction/src/CruRawReader.cxx#L930-L931 would give us this. So we need to think carefully what we actually want to mask then.

martenole:
> > I think the goal is to mask the noisy channels in the MCMs directly. That way we would not even get these tracklets in the first place.
> 
> Good point, so probably skipping them here
> 
> https://github.com/AliceO2Group/AliceO2/blob/99a50a1aa0004a60b129b02f9a342b5a270eb9fc/Detectors/TRD/reconstruction/src/CruRawReader.cxx#L930-L931
> would give us this. So we need to think carefully what we actually want to mask then.

No, masking them in the MCMs (in hardware) would mean we don't need to filter them anywhere in O2 ;)

f3sch:
I see, would that also mean that would not get these pads in the digits anymore?

bazinski:
@f3sch correct, the mcm channel is switched off.

tdietel:
> I see, would that also mean that would not get these pads in the digits anymore?
Yes, these channels (strictly speaking, as we might only mask one channel for shared pads), will no longer appear in the data stream when masked.
We should therefore make sure that we read them out during noise runs, to confirm if they are indeed still noisy.

bazinski:
@tdietel one thing to consider outside this PR is that the compilation of the trap code should pull in the masks, or is this going to be an error-prone manual thing?

martenole:
> We should therefore make sure that we read them out during noise runs, to confirm if they are indeed still noisy.

But in our noise runs I see some channels which have a constant ADC value of 10. Didn't you say these were the masked ones? Should we then create a new noise config where no channels are masked? Should we maybe have a short chat about this?

f3sch:
> @f3sch correct, the mcm channel is switched off.

Ok, for my understanding how would we update the ccdb object then:
Is this calibration for every run or only every let’s say 3 months?
Do these noisy channels ever start working normally again?
Because without these channels in the digits we can only ever exclude more channels but never activate channels again, right?

pachmay:
In Run 1 +2 we only performed the noise runs when the active chamber status (FED) changed as only this information was used for the analysis. Now we can actually get this information from the FED cfg ever X event that is read out with digits. In one of the last TRD meetings, the idea came up to run a noise run before each physics data taking (at the beginning of a fill); however I fear as we also have problems in configuring the FED that some chambers might not get the reconfiguration command for physics. Thus I think it would be better to perform the noise runs on a regular basis, but under controlled conditions.

Detectors/TRD/workflow/include/TRDWorkflow/NoiseCalibSpec.h
tdietel: The 2 CCDB objects are not really necessary -  I just thought about some obsucre test scenarios. 
I overlooked the possibility to pass the config key. That resolves this issue.

tdietel (approved):
Thanks. Looks good.

f3sch (approved):

martenole:
Failures unrelated, merging

PR #10983, opened 2023-03-20
Simplify TRD tracking QC and avoid crash in synch mode

martenole:
@f3sch with this change I try to simplify a bit the tracking QC dataformat. Please check that all the relevant information you need is still available (it should). For example the eta angle does not need to be stored when we keep the full track parameters. Then you can simply get it from e.g. `trackTRD.getEta()`.
In the beginning I wanted to have this class compatible to what was done in AliRoot, but now I anyhow have to redo the scripts I used back then and therefore I wanted to clean up.
Also it is now possible to disable the PID information, because the workflow otherwise crashes if only ITS-TPC-TRD tracks are used as input (like it is done in the synchronous mode for example).

### DIFF
--- a/Detectors/TRD/calibration/macros/makeDeflectionCorrelation.C
+++ b/Detectors/TRD/calibration/macros/makeDeflectionCorrelation.C
@@ -61,7 +61,7 @@
 #define CHAMBER 200
 #define NTRACKLETS 3
 #define FIT_MIN 1000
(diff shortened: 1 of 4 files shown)


f3sch (approved):
LGTM!

martenole:
Failures unrelated (tested locally), merging

PR #10985, opened 2023-03-20
Dump debugging information for global tracking from interpolation workflow

martenole:
Opening a draft PR for discussion.

@shahor02 when I have the ITS clusters stored inside `mITSClustersArray` which is of the type `std::vector<o2::BaseCluster<float>>` how can I convert the position into the global coordinate frame? In BaseCluster.h it says the position is stored in tracking frame so I think it is rotated to the angle of the stave which I can get from the Sensor ID, right?
My goal is to dump (optionally) all the relevant global tracking information into a single output file of the TPC interpolation workflow, so all track points and the different track parameterizations (ITS, ITS-TPC, ITS-TPC-TRD, ...).


shahor02:
Hi @martenole 

the positions in the tracking frame can be obtained per cluster as 
```
auto glo = gman->getMatrixT2GRot(c.getSensorID())^c.getXYZ();
```
where the singleton gman is to be initialized once as:
```
auto* geom = o2::its::GeometryTGeo::Instance();
geom->fillMatrixCache(o2::math_utils::bit2Mask(o2::math_utils::TransformType::T2GRot);
```

PR #10990, opened 2023-03-21
Fix error introduced by adding FOCAL to DetID

fgrosa:
This PR fixes the modification introduced in https://github.com/AliceO2Group/AliceO2/pull/10946, which breaks the simulation of upgrade detectors (the DetID has been shifted by 1 unity, but not in the Hit names). @mfasDa could you please have a look that it is fine with you?


fgrosa:
@shahor02 @mfasDa given that is trivial, can we merge it even if the tests are red (unrelated errors) since the current code is broken for ITS3 simulations? Thanks!

shahor02 (approved):

mfasDa:
Hi @fgrosa , thanks a lot for fixing! Looks fine.

PR #10998, opened 2023-03-23
Track interpolation skip split clusters also for track extrapolation

martenole:
To be seen as hotfix. Without this the track extrapolation done for ITS-TPC only tracks crashes in case the TPC tracks has more clusters assigned than pad rows.
I need to check if the split clusters have a fixed ordering in which case they might introduce some bias which would need to be corrected. If not, it is fine as it is here. In theory a track can also have more than two clusters assigned per pad row.


wiechula (approved):

martenole:
Error unrelated, merging

PR #10999, opened 2023-03-23
fix for TOF compressor

noferini:
@preghenella @ktf 
This is to fix some mismatch in TOF checkers and to add a protection to payload resize

### DIFF
--- a/Detectors/TOF/compression/src/CompressorTaskOld.cxx
+++ b/Detectors/TOF/compression/src/CompressorTaskOld.cxx
@@ -153,7 +153,8 @@ void CompressorTaskOld<RDH, verbose, paranoid>::run(ProcessingContext& pc)
 
     /** initialise output message **/
     auto bufferSize = mOutputBufferSize >= 0 ? mOutputBufferSize + subspecBufferSize[subspec] : std::abs(mOutputBufferSize);
-    auto payloadMessage = device->NewMessage(bufferSize);
+    auto bufferSizeDouble = bufferSize * 2;
+    auto payloadMessage = device->NewMessage(bufferSizeDouble); // start with double size since later resized (but resize cannot resize more than original one)
     auto bufferPointer = (char*)payloadMessage->GetData();
 
     /** loop over subspec parts **/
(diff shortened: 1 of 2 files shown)


noferini:
Can we merge this PR?

Detectors/TOF/compression/src/CompressorTask.cxx
```diff
@@ -150,6 +150,9 @@ void CompressorTask<RDH, verbose, paranoid>::run(ProcessingContext& pc)
       headerOut.payloadSize += payloadOutSize;
     }
 
+    if (headerOut.payloadSize < 0) {
```
shahor02: `DataHeader::payloadSize` is `uint64_t`, this will be always evaluated to false, even in case of the overflow in the additions above.

Detectors/TOF/compression/src/CompressorTaskOld.cxx
```diff
@@ -180,6 +180,11 @@ void CompressorTaskOld<RDH, verbose, paranoid>::run(ProcessingContext& pc)
     }
 
     /** finalise output message **/
+
+    if (headerOut.payloadSize < 0) {
+      headerOut.payloadSize = 0;
+    }
```
shahor02, line 186: idem

Detectors/TOF/compression/src/CompressorTask.cxx
noferini: You are right, then I have to deeper investigate

noferini: I removed this line in the meanwhile

Detectors/TOF/compression/src/CompressorTaskOld.cxx
noferini: I removed this line in the meanwhile

noferini:
I added a protection to avoid a crash with error message on flp
cannot set used size higher than original.

This is triggered by 
    payloadMessage->SetUsedSize(headerOut.payloadSize);

when payloadSize is greater thano original size (bufferSize).

I set original size to 2 times this value and if compressed size exceeds this value I send out zero payload.

This is the main reason of crashes at P2 (hoping this solve the problem). For the moment applied only to old task (the one used on FLP). If it works I'll port the same also in the new task (which is a bit different)

noferini:
Is it possible to merge this before the next flp suite is created?

shahor02 (approved):

PR #11000, opened 2023-03-23
DPL: workaround limit in number of possible in fly exceptions

ktf:
If we actually reach this limit, it means we are in trouble..


ktf:
Errors unrelated, merging.

PR #11007, opened 2023-03-24
DPL GUI: only keep two strings for the string metrics

ktf:


jgrosseo:
This breaks the Hyperloop measurements which relies on `aod-file-read-info` to see how much data was processed. We need to revert asap.

Off topic: do I read it correctly that this was never correctly measuring on the Grid if a job had more than 32 files?

ktf:
Sorry about that, I did see something wrong at some point, but it was together with something else and upon retrial with only this commit things seemed fine. This is the dip you see in https://alimonitor.cern.ch/hyperloop/wagon-test/71559, but then as you can see in the dev dataset metrics it went back up.

That said, yes, we only keep the last 32 string metrics (1024 for ints and floats) in a ring buffer to have a prefixed cap to the memory used. Doesn't that simply reduce statistics, though?

I will think of a way to keep all of them.

jgrosseo:
It went back up because today I did two additional tests (one more with today's tag and one more with yesterday's tag). If you click on the plot marker it brings you to the test window and you can read of the tag...

It reduces stats on the Grid but we can also use that information to tag slow files etc. So if we cannot keep them in the metrics, then we should find another way to store them...

ktf:
I will be moving the expensive string metrics (i.e. the one in general used for some parts of the GUI) to a different mechanism which does not track history, so we can keep longer buffers where we actually need them (like for the file statistics).

IIRC, the reason why we use the string metrics is to be able to join all the values referring to the same file, no?

PR #11009, opened 2023-03-24
CcdbApi: code reuse; more consistent header treatment for snapshots

sawenzel:
* a single place to update meta information for local snapshot files

* also put meta information when local file is created through storeAsBinaryFile API --> this is important so that queries to the local snapshot can correctly read back the header information


sawenzel:
@shahor02

costing (approved):

PR #11010, opened 2023-03-24
TPC: Allow skipping sector numbers and side name in painter

wiechula:


wiechula:
The error looks unrelated, can this be merged?

PR #11015, opened 2023-03-24
DPL: demote alarm messages when not running online.

ktf:


ktf:
@davidrohr do we really need to have an alarm for the rate limiting also when not running online?

davidrohr:
In general I agree, though we need a better way to detect whether we are running online.
E.g. in the full system test which publishes at a fixed rate it would be good to have these messages.

Can't we introduce a flag somewhere in DPL which we can query, insead of checking multiple env variables in many places.
Perhaps even an integer with states like "no online" = 0, "local online test" = 1, "online at P2" = 2, and then we can check "if state >= 1" etc.

ktf:
Merging this to get it into the build. #11017 should do what you ask?

PR #11023, opened 2023-03-26
Ctpdev: debugging nopadding

lietava:


lietava:
@shahor02 Hi Ruben,
I am testing ctp nopadding with real data using emc as reference. 
I generate triggers as multiples of 60001 so I know what to expect.

In emc I see what I expect while ctp is missing some triggers

Already in RawDecoder I do not see missing orbits (i.e. they have zero payload)

Can something go wrong between CRU and creation of TF file ?

lietava:
You can see in code in code where I am dumping. Already at that point I do not see what is expected, i.e.
I see difference with emc and also sequnce is not multiple of 60001.

lietava:
Data are in /alice/data/2023/LHC23a_EMC/533524/raw/1220/

shahor02:
As I wrote in MM, the raw data is dumped to the TF as it is shipped from the CRU.  Some HBFs might be dropped on CRU level, but if even the period of your generated triggers is not what you expect, does not this point on your FW problems?

lietava:
Yes, that is what I am trying to bracket ... I ll add here more info.

lietava:
Please, see debug info 
[ctpreadout.txt](https://github.com/AliceO2Group/AliceO2/files/11072208/ctpreadout.txt)
in attached file.

lietava:
@shahor02 : I checked in the lab with the same fw version of CRU and CTP as in P2 and I can see in cru dump all bcids as expected.
Then something has to happen between cru and o2-ctp-reco-workflow. As I said and attached above in dump from o2-ctp-reco-workflow I already see some bcids missing. What else can I check ?
Maybe I can check raw data ?

lietava:
HI @shahor02 , fw looks ok. Both in lab and in P2 I see in cru raw data dump all bcs as expected. What can go wrong between flp and creation of tf file ? Or in my code ? It was validated on mc ...

shahor02:
Hi @lietava 
What I see is something different. I've converted 1 TF file to digits as
```
o2-raw-tf-reader-workflow --input-data o2_rawtf_run00533524_tf00000001_epn188.tf --onlyDet CTP,EMC --raw-only-det CTP,EMC | o2-ctp-reco-workflow | o2-ctp-digit-writer | o2-emcal-reco-workflow --input-type raw --output-type cells | o2-emcal-cell-writer-workflow --disable-mc
```
Then dumped for EMC and CTP trigger record and their BC differences:
1) for EMC:
```
TFile *_file0 = TFile::Open("emccells.root");
TTree* osim = (TTree*) _file0->Get("o2sim");
std::vector<o2::emcal::TriggerRecord> vd, *vdp=&vd;
o2::InteractionRecord bcPrev;
o2sim->SetBranchAddress("EMCALCellTRGR", &vdp);
int ntf = o2sim->GetEntries();
.> dmpEMC.txt
size_t cnt = 0; bcPrev.clear();
for (int ie=0;ie<ntf;ie++) {
  o2sim->GetEntry(ie);
  for (size_t i=0;i<vd.size();i++) {
    std::cout << "Trig#" << cnt << " (" << i << " in TF " << ie << ") IR=" << vd[i].getBCData().asString();
    if (!bcPrev.isDummy()) {
      long dif=vd[i].getBCData().differenceInBC(bcPrev);
      std::cout << " distToPrev=" << dif << "(=" << (dif/60001) << "*60001+" << dif%60001 <<")" << "\n";
    } else {
      std::cout << "\n";
    }
    bcPrev = vd[i].getBCData();
    cnt++;
  }
}
.>
```
2) For CTP:
```
TFile *_file0 = TFile::Open("ctpdigits.root");
TTree* osim = (TTree*) _file0->Get("o2sim");
std::vector<o2::ctp::CTPDigit> vd, * vdp = &vd;
o2sim->SetBranchAddress("CTPDigits",&vdp)
int ntf = o2sim->GetEntries();
o2::InteractionRecord bcPrev;
.> dmpCTP.txt
size_t cnt = 0; bcPrev.clear();
for (int ie=0;ie<ntf;ie++) {
  o2sim->GetEntry(ie);
  for (size_t i=0;i<vd.size();i++) {
    std::cout << "Trig#" << cnt << " (" << i << " in TF " << ie << ") IR=" << vd[i].intRecord.asString();
    if (!bcPrev.isDummy()) {
      long dif=vd[i].intRecord.differenceInBC(bcPrev);
      std::cout << " distToPrev=" << dif << "(=" << (dif/60001) << "*60001+" << dif%60001 <<")" << "\n";
    } else {
      std::cout << "\n";
    }
    bcPrev = vd[i].intRecord;
    cnt++;
  }
}
.>
```
The dumps are in [dmp.tar.gz](https://github.com/AliceO2Group/AliceO2/files/11083460/dmp.tar.gz)

1) Given that the TFs are sent to different EPNs in round robin, you should not expect to see in the given TF file consecutive TFs / Triggers
2) In the beginning there is a EMCal trigger which does not fit to your 60001 period scheme
3) Apart from that all distances I see are 60001 both for EMC and CTP (you wrote me that for the CTP this is not the case)
4) CTP indeed loses some triggers seen in the same TF of the EMC, including cases when the EMC sees 2 triggers in the TF while the CTP sees only 1.
 
Are you sure this is not a problem of your raw data decoder? The DD does not touch the payload, I cannot imaging how it can lose part of it.
If you are sure in your decoded, you should probably write some raw data from the readout exe and run it through the `o2-raw-file-reader`.

lietava:
Thanks. 
Please,  remind me how I control frequency in simulation ? 
Is it possible to generate with fixed bc ?

shahor02:
Use `--interactionRate <Hz>`  of the `o2-sim-digitizer (in the sim_challenge script the rate is passed in kHz, run `sim_challenge.sh -h`  for details. For the fixed bc generation I am not sure, should not be possible unless @sawenzel has added it.

lietava:
@shahor02 : back to the weekend data.
I processed FULL runs 533524 (nopadding) and 533533 (with padding).
For emc reference I used files provided by Martin which are consistent with 60001 period.
For ctp I looked directly to dump in my code as I am reading tf:
https://github.com/lietava/AliceO2/blob/c2cd94a505b5766180a5234aea324d3f0a307cd2/Detectors/CTP/workflow/src/RawDecoderSpec.cxx#L148
===> I assume that here I should see everything from tf file, please, confirm 

And I see about 8 % difference  both for padded and not padded case.  See also attached stats.
[533524comp.log](https://github.com/AliceO2Group/AliceO2/files/11088825/533524comp.log)
[533533comp.txt](https://github.com/AliceO2Group/AliceO2/files/11088827/533533comp.txt)

lietava:
@shahor02 : if I do dump on ctp flp in P2  I see no missing triggers with the same period 60001 on sample of 12569 triggers

shahor02:
Is it possible to record the raw data on the flp and rawTF on the EPN at the same time?

lietava:
> Is it possible to record the raw data on the flp and rawTF on the EPN at the same time?

That is the question. I do not know. Pippo is on holiday. Discussing with Martin P. Who else can know ?
I'll check with Vasco.

lietava:
The modificationa of the reader are in [AliceO2Group/AliceO2#11056](https://github.com/AliceO2Group/AliceO2/pull/11056)

shahor02:
@lietava PR11023 was already merged

shahor02:
Note that the problem concerned only reading raw files from CTP FLP. The rawTF files are read as before.

lietava:
> Note that the problem concerned only reading raw files from CTP FLP. The rawTF files are read as before.

HI @shahor02 , well exactly. Similar fix is needed somewhere in tf-file-reader ?
When reading using  o2-raw-tf-reader-workflow I see 684 missing triggers out of 122431.

shahor02:
> HI @shahor02 , well exactly. Similar fix is needed somewhere in tf-file-reader ? When reading using o2-raw-tf-reader-workflow I see 684 missing triggers out of 122431.

No, raw-tf-reader does not check RDH format, ships whatever DD dumped into the file. Could you show an example of the trigger present in the flp raw file and missing in the rawTF for the same TF?

lietava:
HI Ruben, please, see attached file which is comparison of tfreader dump wrt to emc triggers.
EMC triggers are now same as raw reader dump (checked).
tf reader triggers are obtained by print in lines:
https://github.com/lietava/AliceO2/blob/c2cd94a505b5766180a5234aea324d3f0a307cd2/Detectors/CTP/workflow/src/RawDecoderSpec.cxx#L148
For differences of tf triggers , please, see attached file.
In file:
+1 found are okish - we shall discuss them later.
not found :=====> these are NOT found in tf dump
So sanity check is count at the end of file:
found: (112216+9531)= 121747
NOT found: 684
Tot: 122431

[tfreader_emc.txt](https://github.com/AliceO2Group/AliceO2/files/11121968/tfreader_emc.txt)

shahor02:
Hi @lietava 
Not sure what are you comparing and what do you mean saying `EMC triggers are now same as raw reader dump`. 
What I see is the following:

1) rawTF files are:
```
/alice/data/2023/LHC23a_EMC/533619/raw/2120/o2_rawtf_run00533619_tf00000001_epn058.tf
/alice/data/2023/LHC23a_EMC/533619/raw/2120/o2_rawtf_run00533619_tf00000002_epn218.tf
/alice/data/2023/LHC23a_EMC/533619/raw/2120/o2_rawtf_run00533619_tf00000003_epn186.tf
/alice/data/2023/LHC23a_EMC/533619/raw/2120/o2_rawtf_run00533619_tf00000004_epn026.tf
/alice/data/2023/LHC23a_EMC/533619/raw/2120/o2_rawtf_run00533619_tf00000005_epn090.tf
/alice/data/2023/LHC23a_EMC/533619/raw/2120/o2_rawtf_run00533619_tf00005027_epn218.tf
...
```

i.e. TFs from 1 to 5026 are contained only in the 1st 5 files. 

2) Extracting from these files the 1st orbit of each TF for the links `CTP/RAWDATA/64` and `EMC/RAWDATA/0` I check if the EMC and/or CTP STF for given TF_ID was received by any of EPNs. 
Note that here I don't check the trigger content, just the 1st orbit for every received TF.

Results are in [emc_ctp_533619_tfs.txt](https://github.com/AliceO2Group/AliceO2/files/11127847/emc_ctp_533619_tfs.txt)
What I see is that 

a) EMC STF was missing for TFs 2 : 10.

b) TFs 26 - 50 (resp. orbits 243955328 to 243956127) are absent in the rawTF EPN files (but I see them on them in the CTP FLP file you provided). 
The same is true for TFs 62-64. These TFs seem to correspond to your 1st batch of "not found" triggers.

c) I don't see any TF with EMC present and CTF absent.

So, either the missing TFs were recorded not at 100% or they were dropped. But I don't see cases where the EMC is present and CTP is absent. The inverse happens in the very beginning (a).

lietava:
Hi @shahor02 ,
what I mean is following:
1)  I got file from Martin which conatins all triggers seen in EMC readout (I know I can do myself ...)
2)  when I run on raw data after your fixes, I got exactly same triggers as in 1
o2-raw-file-reader-workflow --input-conf CTPraw.cfg --detect-tf0 --configKeyValues "HBFUtils.nHBFPerTF=32" | o2-ctp-reco-workflow -b
3) when I run on TF data it seems I am missing some triggers which are in the list in previous post:
 o2-raw-tf-reader-workflow -b --onlyDet CTP  --input-data list.txt  | o2-ctp-reco-workflow -b  --use-verbose-mode
I am checking only orbits in payload as seen in ctp decoder before decoding as triggers are rare.
List of TF is attached.
[list.txt](https://github.com/AliceO2Group/AliceO2/files/11128754/list.txt)

Maybe we can have chat before or after  RC daily (1630) ?

shahor02:
@lietava but what you describe does not mean that there is a difference between EMC and CTP triggers on the EPNs, it only shows that there is a difference between what comes from the CRU and what is seen on the EPN. And, as far as I could check, your missing triggers correspond to missing TFs in the rawTF files, i.e. the TFs were simply dropped.
You can see it directly on the grafana that there were dropped TFs in the beginning of the run: http://alihlt-gw-prod.cern.ch:3325/d/0PsIiR87z/output-rates-and-sizes-3?orgId=1&from=1680031200000&to=1680032400000

image

Anyway, if you want, we can chat in https://cern.zoom.us/my/shahoian?pwd=M0xzMStUUXczQmVGeUhUZFhSZWFQUT09, just ping me when you are available.

lietava:
@shahor02 : in RAWDataHeader in lines 
https://github.com/lietava/AliceO2/blob/c2cd94a505b5766180a5234aea324d3f0a307cd2/DataFormats/Headers/include/Headers/RAWDataHeader.h#L414
there is triggerOrbit and heartbeatOrbit. I want to be sure what they are.
Referring to RDH:
https://indico.cern.ch/event/1249722/contributions/5250938/attachments/2586209/4461814/RDHV6_V7.pdf
heartbeatOrbit is 32 bit word 5,
triggerOrbit is orbit from trigger type in 32 bit word 8 ?

shahor02:
@lietava you are referring to an obsolete RDHv4. Starting from V5 there is only an orbit (HB or Trigger depending on the CRU or RORC) and BC (used for RORC only).

lietava:
@shahor02 : Data Processors exchange always Time Frames ?

shahor02:
What do you mean?

lietava:
fairmq messages between processes .

shahor02:
There might be some service messages, timers etc.

lietava:
Hi @shahor02 , there was bug in reconstruction - I was throwing away Lm-L1 at the beginning of each HB. 
Please, have a look - it is related to our previous discussion on applying offsets.
In the meantime I would still do more validation.

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
```diff
@@ -332,23 +356,24 @@ int RawDecoderSpec::addCTPDigit(uint32_t linkCRU, uint32_t triggerOrbit, gbtword
   } else if (linkCRU == o2::ctp::GBTLinkIDClassRec) {
     int32_t offset = BCShiftCorrection + o2::ctp::TriggerOffsetsParam::Instance().LM_L0 + o2::ctp::TriggerOffsetsParam::Instance().L0_L1 - 1;
     LOG(debug) << "tcr ir ori:" << ir;
-    // if(0) {
-    if ((int32_t)ir.bc < offset) {
-      // LOG(warning) << "Loosing tclass:" << ir;
+    //if ((int32_t)ir.bc < offset) {
+    if ((ir.orbit <= mTFOrbit ) && ((int32_t)ir.bc < offset)) {
```
shahor02: ok, if I understand correctly, before you were rejecting a trigger with `bc < offset` even if after subtracting the `offset` the `orbit` would still stay within the TF? 
Then the fix is ok.

lietava: yes, I was rejecting always when orbit changed. Now I reject only in the first HB of the TF since orbit jump can not be bigger than 1.

lietava: and also I added print of missing TFs , just for info

lietava: but you are doing the same thing for rorc triggers ?

shahor02: yes

lietava:
Hi @shahor02 , so let us try to pull. As far as I was able to find there was only the bug rejecting for every HB instead of TF.
Still I repeat question: why do I need to reject even at beginning of TF in ctpreadout ? As far as I can see there is no structure in  ctpdigits.root  ?

shahor02:
> Hi @shahor02 , so let us try to pull. As far as I was able to find there was only the bug rejecting for every HB instead of TF. Still I repeat question: why do I need to reject even at beginning of TF in ctpreadout ? As far as I can see there is no structure in ctpdigits.root ?

What do you mean by structure? The reconstruction is done within the TF, all the timestamps in the reconstruction output are wrt 1st orbit of the TF. What is the use of registering triggers outside of the reconstructed time slice (btw, note that also the end of the previous TF which rejected trigger belongs to is effectively lost of analysis due to the TPC drift time).

lietava:
> > Hi @shahor02 , so let us try to pull. As far as I was able to find there was only the bug rejecting for every HB instead of TF. Still I repeat question: why do I need to reject even at beginning of TF in ctpreadout ? As far as I can see there is no structure in ctpdigits.root ?
> 
> What do you mean by structure? The reconstruction is done within the TF, all the timestamps in the reconstruction output are wrt 1st orbit of the TF. What is the use of registering triggers outside of the reconstructed time slice (btw, note that also the end of the previous TF which rejected trigger belongs to is effectively lost of analysis due to the TPC drift time).

I meant that in ctpreadout.root there are absolute event id = {orbit, bcid} . But what do you say is that if correct shift is applied during reconstruction when creating CTF then one can not go outside of TF  ?

shahor02:
Technically, one can go outside of the TF (currently some operations with such objects produce a warning), but what is the of keeping an object which a priory does not belong to any collision which can be reconstructed in the TF?

lietava:
> Technically, one can go outside of the TF (currently some operations with such objects produce a warning), but what is the of keeping an object which a priory does not belong to any collision which can be reconstructed in the TF?

@shahor02 As I said motivation is to not bias eg lumi measurement from ctpreadout. And somehow I feel it is not elegant or how to say it ...  But I agree it is not critical.
The best solution now seems to me not to shift anything during reconstruction 
and do the shifts only in aod producer where are no tfs ?  Do you agree ?

shahor02:
Isn't luminosity provided by scalers? The best solution would be to apply the shift in your firmware, so that the DD chops to TFs already corrected data...

Anyway, if for some reason it is important to have these stray CTP triggers we can keep them since they are not used in the global reconstruction. Will you apply their exclusion in the AODProducer?

lietava:
@shahor02 : you are right most elegant would be to fix it in fw.
I suggest for the moment we keep as it is. 
What do you mean by question: ' Will you apply their exclusion in the AODProducer?' What needs to be excluded ?

shahor02:
I mean you were suggesting to not remove the trigger going out of the TF during the decoding but to suppress it in the AOD creation.

lietava:
> I mean you were suggesting to not remove the trigger going out of the TF during the decoding but to suppress it in the AOD creation.

yes. But in that case AODproducer would do the bc shifts for ctp and rorc readout. What needs to be excluded ?

shahor02:
Sorry, i am lost. You wanted to apply the shifts in the AOD producer, which is still done per TF so the triggers shifted out of the TF need to be discarded. For the RORC the shifts/discarding are applied at raw decoding level.

lietava:
> Sorry, i am lost. You wanted to apply the shifts in the AOD producer, which is still done per TF so the triggers shifted out of the TF need to be discarded. For the RORC the shifts/discarding are applied at raw decoding level.

yes, I know as it is done now. I am just discussing other options for my understanding.
Other option I see NOT to do nothing for ctp and rorcs during reconstruction and do all shifts in AODProducer.
I thought that in aods there is no TF structure so nothing can be rejected. Both for ctp and detectors.

shahor02:
@lietava the AOD creation is part of the async. reconstruction which is always done per TF. Currently there is no possibility to apply shift / outlier rejections in the AOD creation. For RORC detectors this is done on the raw decoding level in the online processing so that the CTF has already corrected triggers. If this is not the case (e.g. for the CTFs created befor July) optinally one can apply shifts/rejections in the CTF decoding.

PR #11027, opened 2023-03-27
DPL: recognize ALARM as a valid level

ktf:


ktf:
@davidrohr @shahor02 the full CI seems to fail everywhere with the same error related to the FT0 configuration, is this a known issue?

martenole:
It requires https://github.com/AliceO2Group/O2DPG/pull/974 in O2DPG. So when it restarts should be fine. Would have probably been better to merge https://github.com/AliceO2Group/AliceO2/pull/11036/ with defaults for the newly introduced variables so that it would not have relied on the O2DPG change

davidrohr:
@TimoWilken @ktf : Can we change the FullCI CI build for AliceO2 and for alidist to O2DPG/master, and not the O2DPG version from alidist?

ktf:
@davidrohr Sure. This https://github.com/alisw/ali-bot/pull/1240 should do it.

PR #11034, opened 2023-03-27
DPL: avoid using a std::set

ktf:


ktf:
@aalkin Am I missing something? Can we simply use sorted vectors in place of sets, given in the end we only insert once?

Framework/Core/include/Framework/AnalysisTask.h
```diff
@@ -148,11 +148,11 @@ struct AnalysisDataProcessorBuilder {
     if constexpr (is_enumeration_v<dT> == false) {
       if constexpr (soa::is_soa_filtered_v<dT>) {
         auto fields = createFieldsFromColumns(typename dT::table_t::persistent_columns_t{});
-        eInfos.push_back({ai, hash, dT::hashes(), std::make_shared<arrow::Schema>(fields), nullptr});
+        eInfos.emplace_back(ExpressionInfo{ai, hash, dT::hashes(), std::make_shared<arrow::Schema>(fields), nullptr});
```
aalkin, line 151:
```cpp
emplace_back(ai, hash, dT::hashes(), std::make_shared<arrow::Schema>(fields), nullptr)
```
should be even better, although it does not matter much for a simple struct.

aalkin:
@ktf indeed, the only reason `std::set` was used is to ensure uniqueness. Still, we should probably test this with a real workflow.

PR #11035, opened 2023-03-27
Initialize run number with -1 to trigger run number change also in ca…

martenole:
…se run number=0


martenole:
Tested with FST locally, merging

PR #11037, opened 2023-03-27
DPL: sort expendable task after non expendable ones

ktf:


ktf:
@Barthelemy @knopers8 @davidrohr this is needed to make sure that expendable tasks are sorted after the non expendable ones. Basically one needs to add:

```C++
.labels = {{"expendable"}}
```

in the definition of the QC `DataProcessorSpecs`. Do we have already a different label for them?

davidrohr:
Not that I am aware of, but QC experts should comment. 
You can mention the syntax also here: https://alice.its.cern.ch/jira/browse/QC-953?filter=-2 , to have it all in one place.

shahor02:
@ktf, the real problem in https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/AliceO2/11037/aa1f02960ce6bf845f782b97f9c25213a9bc62dc/build_O2_fullCI/fullLog.txt was not the 
```
[ERROR] invalid workflow in o2-ft0-reco-workflow: No matching output found for FT0/DIGITSBC/0 as requested by data processor "ft0-reconstructor". Candidates
```
but the 
```
[FATAL] Illegal command-line key/value string: ITSVertexerParam.nThreads=
[FATAL] unknown error while setting up workflow in o2-its-reco-workflow.
```
Apparently, the build started before the needed https://github.com/AliceO2Group/O2DPG/pull/974 was merged?
At least for me with fresh O2DPG and O2 the fst works.

ktf:
Failure on macos unrelated. No-op when no expendable tasks. Merging.

PR #11038, opened 2023-03-27
DPL: support DDS expendable tasks

ktf:


ktf:
@davidrohr @Barthelemy @knopers8 this is the second part, enabling support to export the labels.

PR #11003, opened 2023-03-23
Strangeness tracking in the AO2D, act 2

mpuccio:
It supersedes #10868. @fmazzasc

PR #11040, opened 2023-03-28
Strangeness Tracking in AO2D, act 3

fmazzasc:
W/ @mpuccio , it supersedes #11003 and #10868


Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -855,11 +855,82 @@ DECLARE_SOA_INDEX_COLUMN_FULL(Track2, track2, int, Tracks, "_2"); //! Track 2 in
...
+                       strangenesstracking::DecayX,
+                       strangenesstracking::DecayY,
+                       strangenesstracking::DecayZ,
+                       strangenesstracking::XiMass,
+                       strangenesstracking::OmegaMass,
+                       strangenesstracking::MatchingChi2,
+                       strangenesstracking::TopologyChi2,
+                       strangenesstracking::ITSclsSize);
+
+DECLARE_SOA_TABLE_FULL(TrackedV0s, "TrackedV0", "AOD", "TRACKEDV0", //! Strangeness tracking cascade table
```
jgrosseo: The comment for the three tables is identical. Please explain the difference here :)

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -855,11 +855,82 @@ DECLARE_SOA_INDEX_COLUMN_FULL(Track2, track2, int, Tracks, "_2"); //! Track 2 in
...
 
-using Decays3Body = Decays3Body; //! this defines the current default version
-using Decay3Body = Decays3Body::iterator;
+using Decay3Bodys = Decay3Bodys; //! this defines the current default version
+using Decay3Body = Decay3Bodys::iterator;
+
+namespace strangenesstracking
+{
+DECLARE_SOA_INDEX_COLUMN(Track, track);                                 //! Strange track index
+DECLARE_SOA_INDEX_COLUMN_FULL(ITSTrack, itsTrack, int, Tracks, "_ITS"); //! ITS index
```
ddobrigk, line 867: Thanks a lot for your work! I have a quick question here: do we need two indices? I think we can live with only the line with DECLARE_SOA_INDEX_COLUMN_FULL  (as this provides already the correct track reference) and remove the DECLARE_SOA_INDEX_COLUMN but maybe I am missing something - sorry if so.

fmazzasc: Hi David, one index is for the strange track, which takes the V0/Cascade and propagates into the ITS clusters compatible.
The second index is the reference for the pure ITS track. So both the indices are needed

ddobrigk: I see, makes sense! Thanks for clarifying!

mpuccio (approved):
I am approving (obviously) the ST part. It would be great to merge it if there are no further comments, as we have a few other improvements in the pipeline.

jgrosseo:
The data model changes are fine with me. I think we need someone from reconstruction as well...

mpuccio:
@shahor02 is it fine with you? With this PR we do not enable it... yet. That will come very soon ;)

shahor02 (approved):

PR #11047, opened 2023-03-28
DPL: fix rates

ktf:


ktf:
@awegrzyn @vascobarroso this should bring back the DPL dashboard. Is there a way I can check this?

awegrzyn:
The easiest way is to just run FLP Suite CI. Otherwise I could just test it manually.

ktf:
How do I run the FLP Suite CI?

awegrzyn:
hmm, good questions, if you give me O2 tag or commit I'll do it...

ktf:
50b108aed5969005808f33b7eab1413fae525188 is the commit. How difficult would that be to run the FLP suite CI on each PR?

awegrzyn:
I don't think it's possible atm as we luck resources for that. We could think of something on-demand

awegrzyn:
@ktf I managed to verify it now, it seems all's good

PR #11053, opened 2023-03-29
DPL Analysis: Update combinations to use slice cache

aalkin:
@saganatt this should make the combinations compatible with both old and new API, so that O2Physics can be adapted without interruption.

### DIFF
--- a/Framework/Core/include/Framework/GroupedCombinations.h
+++ b/Framework/Core/include/Framework/GroupedCombinations.h
@@ -15,6 +15,7 @@
 #include "Framework/ASoAHelpers.h"
 #include "Framework/BinningPolicy.h"
 #include "Framework/Pack.h"
+#include "Framework/SliceCache.h"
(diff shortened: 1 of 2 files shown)


saganatt:
Thanks a lot!

1. So, there is still a non-cached sliceByCached version?
2. Why do we prefer raw pointers instead of the smart ones?

aalkin:
1. For compatibility, otherwise the the current O2Physics will not compile
2. It would be better to just copy `SliceCache` because it is a wrapper of a single pointer, but this is the minimal change to keep compatibility.

saganatt:
Ok, I see.

Do you then predict removing the raw pointer once the O2Physics gets updated?

aalkin:
Yes, but it also needs to be done in steps: adding API, adapting O2Physics, removing old API.

saganatt:
I understand, thank you very much

PR #11054, opened 2023-03-29
DPL: allow driver to run under DDS control

ktf:
This introduces a new --driver-mode flag which can be used to modify the DDS configuration so that the dpl driver is part of the topology, communicates with the children via websocket and takes care of the metrics.

### DIFF
--- a/Framework/Core/include/Framework/DriverControl.h
+++ b/Framework/Core/include/Framework/DriverControl.h
@@ -20,9 +20,7 @@
 #include "Framework/DeviceSpec.h"
 #include "Framework/DeviceExecution.h"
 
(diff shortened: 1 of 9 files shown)


ktf:
@davidrohr how can we test this? It will probably need some extra variables to be provided, if we want to have the remote GUI, is it possible to do it from the DDS / ODC / Control interface?

@Barthelemy @knopers8 @vascobarroso this should be easily extendable to the ECS exporter. Are you interested? It would mean that the DPL driver will run as an additional task on each FLP, collect the metrics and allow people to look at the remote GUI (again, with some extra environment variable).

PR #11061, opened 2023-03-30
DPL Analysis: fix memory issue due to creation of empty tables

aalkin:


ktf:
Currently broken, merging.

PR #11062, opened 2023-03-30
[EMCAL-688] Add exception for uninitilised lookup table

mhemmer-cern:
- Add exception class to throw when trying to determine exotic clusters but the neeed look up table is not initilised
	- The class UninitLookUpTableException is neseted within the cluster factory class
	- throw will happen inside of setIsExotic function which is called inside buildCluster function
	- throw will only happen if ecell < mExoticCellMinAmplitude is false
- Add first step to merge set up function for the cluster, cell and index container together with the look up table
	- new setContainer function expects all three objects as span and will set them
	- next step would be to change the current calls of the three (four) setter everywhere with this one

### DIFF
--- a/Detectors/EMCAL/base/include/EMCALBase/ClusterFactory.h
+++ b/Detectors/EMCAL/base/include/EMCALBase/ClusterFactory.h
@@ -312,7 +312,6 @@ class ClusterFactory
   Double_t tMaxInCm(const Double_t e = 0.0, const int key = 0) const;
 
   bool getLookUpInit() const { return mLookUpInit; }
-  void setLookUpInit(bool lookUpInit) { mLookUpInit = lookUpInit; }
 
   bool getCoreRadius() const { return mCoreRadius; }
   void setCoreRadius(float radius) { mCoreRadius = radius; }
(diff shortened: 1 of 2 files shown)


Detectors/EMCAL/base/include/EMCALBase/ClusterFactory.h
```diff
@@ -311,6 +311,9 @@ class ClusterFactory
   /// \param  key: = 0(gamma, default); !=  0(electron)
   Double_t tMaxInCm(const Double_t e = 0.0, const int key = 0) const;
 
+  bool getLookUpInit() const { return mLookUpInit; }
+  void setLookUpInit(bool lookUpInit) { mLookUpInit = lookUpInit; }
```
mfasDa: mLookupInit is a state variable, I would not expose it to the world.

mfasDa (changes requested):
Please protect the state variable mLookupTable from outside access (remove `setLookupInit`), otherwise it is good to go.

Detectors/EMCAL/base/include/EMCALBase/ClusterFactory.h
mhemmer-cern: Done!

mfasDa (approved):
Looks good now!

PR #11066, opened 2023-03-30
[EMCAL-911] Add container class for handling reconstructed cells

mfasDa:
- Event container organizing cells / event (including
  merging of HG/LG cells)
- RecoContainer managing interactions from different
  links
- RecoContainerReader for interation over
  RecoContainer

Including unit test checking features of the
Event- and RecoContainer (access / create events,
merge cells, cell error states, reset)

### DIFF
--- a/Detectors/EMCAL/reconstruction/CMakeLists.txt
+++ b/Detectors/EMCAL/reconstruction/CMakeLists.txt
@@ -19,6 +19,7 @@ o2_add_library(EMCALReconstruction
         src/RecoParam.cxx
         src/RawDecodingError.cxx
         src/ReconstructionErrors.cxx
+        src/RecoContainer.cxx
         src/CaloFitResults.cxx
         src/CaloRawFitter.cxx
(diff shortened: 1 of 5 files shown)


mfasDa:
@shahor02 The class is supposed to enter the RawToCellConverter. There are two purposes for the new class:
- I want to replace with this the internal mapping (map of shared pointers of vectors) which is creating the crashes (however no reason they are appearing now, the last change was Nov. 8 and this part never created problems). Furthermore it simplifies the code.
- The reconstruction of the LEDMON signals is the same as the cells, there is a HG and a LG digitizer, they need to be fit separately and then merged in the same way as cells. As the index range overlaps with the cell range they need to be handled in a separate container. Outsourcing the cell merging part into a separate container makes addig the LEDMON decoding a lot easier as the code for HG/LG merging can be taken from the cells.

I would like to merge this class first and the integration into the RawToCellConverter in a second interation.

shahor02 (approved):

PR #11069, opened 2023-03-30
ITS/MFT decoder registers empty ROFs with --allow-empty-rofs

shahor02:


iravasen:
Hi @shahor02, it works! Thanks a lot. For me it can be merged

PR #11077, opened 2023-03-31
Possibility to trigger endOfStream/shutdown from within data converter

sawenzel:
This provides the possibility to initiate shutdown of the ExternalFairMQDeviceProxy from within the InjectorFunction.

This is useful when the InjectorFunction knows how much data it wants to pass along and can facilitate decisions to stop the DPL part independently on the external source sending EOS.

This used in the MCTrackToDPL proxy which will send endOfStream after a given number of events have been treated. So far, this proxy needed an external entity to kill it. Now this is no longer necessary.

Changes have been made minimally invasive and backward compatibility is ensured.


davidrohr (approved):

ktf:
This is fine with me. We should probably clean up the whole thing to simply pass the ServiceRegistryRef so that one can use the ControlService directly.

knopers8 (approved):

sawenzel:
Got confirmation that this is working nicely in an actual injection into DPL analysis. Merging for now. Can do improvements with `ServiceRegistryRef` in a next step.

PR #11082, opened 2023-03-31
TRD tracking fix pad row crossing check

martenole:
@f3sch found the problem: when using multiple candidates I use the `mCandidates` array as scratch space for the different track parameter hypothesis. This block where I am copying back the track from that array into `t` which points to the output track needs to be at the end of the update loop.
Otherwise setting whether or not the track is pad row crossing is not taken into account. That's why you never saw pad row crossings in the last layer. And I am not sure if the information in the lower layers is reliable unfortunately.


martenole:
Trivial, merging

f3sch:
@martenole Thanks for hunting this down! I must have missed this when I added the code below :+1:

PR #11087, opened 2023-04-01
[MID] require 3/4 chambers hit for both cathodes separately

pillot:
@dstocco here are the changes to enforce the 3/4 condition for both cathodes separately.
I made the changes for the 2 algorithms and the effect in term of loss of efficiency is similar for both.
I also did a bit of clean up at the same time.


dstocco:
It looks ok to me.

aphecetche (approved):

PR #11090, opened 2023-04-02
Workflow to filter out specific TPC tracks and clusters

shahor02:
Example of running:
```
o2-tpc-data-filter-workflow --disable-mc --hbfutils-config o2_tfidinfo.root  --shm-segment-size 16000000000 --accept-track-sources ITS-TPC,ITS-TPC-TRD,TPC-TRD-TOF,ITS-TPC-TOF,ITS-TPC-TRD-TOF | \
o2-tpc-reco-workflow  --disable-mc  --shm-segment-size 16000000000  --input-type pass-through --output-type clusters,tracks,send-clusters-per-sector --filtered-input
```

Additional options to `--min-abs-tgl <> --min-tpc-clusters <>` useful to clean up the cosmic. 
Only clusters of selected tracks will be send to the output and the cluster references of the validated tracks will be redone. 
The "discarded" tracks are not really removed from the output vector (unless `--suppress-rejected` option is passed) in order to not break the references of the global tracks. Instead, they are filled by 0s to be better compressible. When read back they will e.g. return `getNClusters() == 0`.

The output by default is written to `tpctracks_filtered.root` and `tpc-filtered-native-clusters.root` to not overwrite original data.

### DIFF
--- a/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
+++ b/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx
@@ -1383,6 +1383,10 @@ GTrackID RecoContainer::getTPCContributorGID(GTrackID gidx) const
   } else if (src == GTrackID::TPCTOF) {
     const auto& parent0 = getTPCTOFMatch(gidx); // TPC : TOF
     return parent0.getTrackRef();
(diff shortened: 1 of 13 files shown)


shahor02:
run fullCI locally, merging

PR #11093, opened 2023-04-03
DPL Analysis: move sorting order validation code from Kernels.cxx

aalkin:


ktf:
Checked on hyperloop as well. Merging.

PR #11094, opened 2023-04-03
DPL: only publish registered metrics

ktf:


ktf:
@knopers8 this should improve the behaviour of the flushing to only check metrics which were registered. That said, I think the actual issue is that something is triggering the attempt to flush too often. I will investigate that.

PR #11095, opened 2023-04-03
MCH: fix for nm calls

aphecetche:
With recent XCode updates on macOS nm no longer supports single dash options for nm.


pillot (approved):

TimoWilken:
Error unrelated, merged.

PR #11099, opened 2023-04-04
DPL: avoid overflow for the shmId metric

ktf:
Notice this means some metrics might end up being multiplexed. A better solution will come soon.


ktf:
@chiarazampolli as discussed. This will prevent the issue when the segment id is too large (in principle there is only one segment id per job, so the overlap should not be a problem).

PR #11103, opened 2023-04-04
[MRRTF-217] introduce StatusMap object and statusmap-creator-workflow…

aphecetche:
… (take 2)

All the attempts at creating a no-op statusmap-creator device failed, so the fallback solution is to simply not add the statusmap-creator if the would be statusmap-creator has no input (i.e. neither the badchannels nor the rejectlist are requested).
This somehow puts some logic in the reco-workflow that should not be there, but that's the only thing I've found to make it work...


pillot (approved):

aphecetche:
FYI @martenole @davidrohr those changes should fix the "full full" system test.

PR #11105, opened 2023-04-04
[O2-3582] DPL Analysis: update running workflow configuration

aalkin:
@ddobrigk This change updates the option values in running workflow info so that configuration inheritance can work. I've tested this with the pair of Tutorial tasks.


ddobrigk:
Many many thanks @aalkin !

PR #11109, opened 2023-04-04
ZDC - Introduction of RDH V7

cortesep:


Detectors/ZDC/raw/src/RawReaderZDC.cxx
```diff
@@ -32,25 +32,50 @@ void RawReaderZDC::clear()
...
+int RawReaderZDC::processBinaryData(gsl::span<const uint8_t> payload, int linkID, uint8_t dataFormat)
 {
   if (0 <= linkID && linkID < 16) {
     size_t payloadSize = payload.size();
-    for (int32_t ip = 0; ip < payloadSize; ip += 16) {
+    if (dataFormat == 2) {
+      for (int32_t ip = 0; (ip + PayloadPerGBTW) <= payloadSize; ip += PayloadPerGBTW) {
+        // Assign only the actual payload
+        uint32_t gbtw[4] = {0x0, 0x0, 0x0, 0x0};
+        memcpy((void*)gbtw, (const void*)&payload[ip], PayloadPerGBTW);
```
shahor02: Do I understand correctly that for the unpadded data you 1st copy every unpadded GBT word to padded version then decode it? It would be more natural to do the opposite: to extract the payload from the padded version and pass it to decoder expecting pure payload.

Detectors/ZDC/raw/src/ZDCRawParserDPLSpec.cxx
```diff
@@ -104,12 +106,30 @@ void ZDCRawParserDPLSpec::run(ProcessingContext& pc)
...
-          mWorker.processWord((const uint32_t*)&payload[ip]);
+        if (dataFormat == 2) {
+          for (int32_t ip = 0; (ip + PayloadPerGBTW) <= payloadSize; ip += PayloadPerGBTW) {
+            // Assign only the actual payload
+            uint32_t gbtw[4] = {0x0, 0x0, 0x0, 0x0};
+            memcpy((void*)gbtw, (const void*)&payload[ip], PayloadPerGBTW);
+#ifdef O2_ZDC_DEBUG
+            o2::zdc::Digits2Raw::print_gbt_word((const uint32_t*)gbtw);
+#endif
+            if (gbtw[0] != 0xffffffff && gbtw[1] != 0xffffffff && (gbtw[2] & 0xffff) != 0xffff) {
```
shahor02, line 122: same comment

shahor02 (commented):
Hi @cortesep 

Please see comments below. Did you try to generate the raw data from the same MC with and w/o padding and make sure that the decoding results are the same?
Also, Pippo should have a FW with RDHv7 both with and w/o padding, perhaps you could ask for some real data to validate the decoding.

Detectors/ZDC/raw/src/RawReaderZDC.cxx
cortesep: Yes, it avoided to introduce more detailed pointer treatment. Now I removed the unnecessary copy.

cortesep:
Stefan produced two files, 
534496 DataFormat 0x0 :
Data rate 207 MBps, QC ON
RUN 534500 DataFormat 0x2 :
Data rate 138MBps, QC OFF 

The ratio of data rates is 0.67 that is very close to 80/128=0.625
The checks on raw files with DumpRaw class and digigitzation followed by reconstruction do work smoothly.

shahor02 (approved):

shahor02:
local fulCI passed, merging.

PR #11111, opened 2023-04-05
[EMCAL-565, EMCAL-566]: Add start stop functionality to EMCal calib

jokonig:
- Data close to the end of run, or runs that dont have enough events to trigger the calibration, were neglected until now and the data was dropped
- Now, the data is saved at the end of run and loaded in the next run to continue the calibration and to not neglect data that was already recorded
- To store the data, boost histograms are converted to root histograms and for the reading of the histograms its vice-versa. Therefore, the boost histograms have to have variable axes instead of integer axes
- For longer stops, one does not always want to load the old data. Therefore, the parameter setSavedSlotAllowedSOR_EMC in the calib params can be used to avoid loading the old parameters at the SOR.
- To disable the procedure completely, setSavedSlotAllowed_EMC has to be set to false

### DIFF
--- a/Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelCalibrator.h
+++ b/Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelCalibrator.h
@@ -70,6 +70,10 @@ class EMCALChannelCalibrator : public o2::calibration::TimeSlotCalibration<DataI
   void finalizeSlot(Slot& slot) final;
   o2::calibration::TimeSlot<DataInput>& emplaceNewSlot(bool front, TFType tstart, TFType tend) final;
 
+  bool saveLastSlotData(TFile& fl) final;
+
+  bool adoptSavedData(const o2::calibration::TimeSlotMetaData& metadata, TFile& fl) final;
+
   ///\brief Set the testing status.
   void setIsTest(bool isTest) { mTest = isTest; }
   bool isTest() const { return mTest; }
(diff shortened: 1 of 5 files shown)


Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelCalibrator.h
```diff
@@ -70,22 +70,38 @@ class EMCALChannelCalibrator : public o2::calibration::TimeSlotCalibration<DataI
...
+
   ///\brief Set the testing status.
   void setIsTest(bool isTest) { mTest = isTest; }
   bool isTest() const { return mTest; }
 
   const CcdbObjectInfoVector& getInfoVector() const { return mInfoVector; }
   const std::vector<DataOutput>& getOutputVector() const { return mCalibObjectVector; }
 
+  /// \brief get if has enough data should be circumvented at EOR
+  bool getSaveAtEOR() { return mSaveAtEOR; }
```
shahor02: const method

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelCalibrator.h
```diff
@@ -70,22 +70,38 @@ class EMCALChannelCalibrator : public o2::calibration::TimeSlotCalibration<DataI
...
+  /// \brief set if has enough data should be circumvented at EOR
+  void setSaveAtEOR(bool tmp) { mSaveAtEOR = tmp; }
+
+  /// \brief get if has enough data should be circumvented at EOR
+  bool getLoadAtSOR() { return mLoadAtSOR; }
```
shahor02: const method

Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
```diff
@@ -160,6 +174,28 @@ class EMCALChannelCalibDevice : public o2::framework::Task
...
+    static bool firstCall = true;
+    if (firstCall) {
+      firstCall = false;
+      if (mTimeCalibrator) {
+        mTimeCalibrator->loadSavedSlot();
+      } else if (mBadChannelCalibrator) {
+        mBadChannelCalibrator->loadSavedSlot();
+      }
+    }
+
```
shahor02, line 186: Could you make this compatible with ECS stop/start procedure (i.e. starting a new run within the same O2 session, see https://indico.cern.ch/event/1270050/contributions/5335478/attachments/2620040/4530862/EOS_and_SSS.pdf)

shahor02 (changes requested):

Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
jokonig: @shahor02 Are you refering to the TimingInfo::globalRunNumberChanged variable? Is this also needed in case that the accumulation should just continue in the case of the start-stop-start procedure? 
Because for the EMCal we discussed internally that the calibration should not be sensitive to run changes and hence the calibration can just continue with the exact same settings and therefore no new initialization is needed. But maybe I am missing something here.

shahor02: Yes, I was referring to the TimingInfo::globalRunNumberChanged. If you are filling just a single slot and don't need to take any action at EOR/SOR, then you don't need to implement it.
But I see that you have SaveAtEOR and EndOfStream processing, they will be invoked at the EOR (between stop/start), are you going to ignore their results?

jokonig: Hi @shahor02 I discussed with Markus and we concluded the following:
- We want to save the calibration histograms at each end of run, also at the start-stop-start. If the start-stop-start procedure is executed, it would not be necessary to store these objects as the calibration just continues. And we also dont read in the histograms as the fisrtcall should still be set to false. If then the new runs stops, these histograms will be overwritten by the ones accumulated in the new run. SO yes, we are ignoring the objects in case of the start-stop-start procedure.
-  For "normal" runs where no start-stop-start was executed, it should store the calib histograms at the end of run. In the next run where we also call the SOR they should be read in and the calibration can continue
So we think that the implementation as it is now should be sufficient for this.

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelCalibrator.h
jokonig: done

Detectors/EMCAL/calibration/include/EMCALCalibration/EMCALChannelCalibrator.h
jokonig: done

jokonig:
@shahor02 could you have a look if this can be merged now?

shahor02 (approved):

PR #11116, opened 2023-04-05
ZDC: Inclusion of data-driven model for free nucleon spectators

sawenzel:
* A new class encapsulation a parametrized model to calculate free spectators given an impact parameter

* Use of class in Pythia8HI generator. Free spectators are put as part of the MCHeader information.

This is a first step towards having fast parametrized digitization for ZDC such as we already did in AliRoot/Run2.

Co-author: chiara.oppedisano@to.infn.it


sawenzel:
@coppedis

PR #11118, opened 2023-04-05
Fix: pattern iterator passed by reference

fmazzasc:
@shahor02 , this was affecting all the clusters of rare topologies


shahor02 (approved):
Thanks for spotting!

shahor02:
failure unrelated, run fullCI locally

PR #11119, opened 2023-04-05
[O2-3661] Get O2 compiling on M2 Macs

TimoWilken:
This patch is unfortunately not pretty, and may not be valid since it modifies DataPointCompositeObject.h, which is copied from DCS.

The problem on M2 Macs is that we need to use LLVM 15, which has started to throw compiler errors for `std::uniform_int_distribution<char>`.

According to <https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution>, the shortest integral type that is defined as valid for use is `short` (which is typically 16 bits, as opposed to `char`'s 8 bits).

Cc: @shahor02 @ktf @pzhristov

### DIFF
--- a/Detectors/DCS/include/DetectorsDCS/DCSDataPointHint.h
+++ b/Detectors/DCS/include/DetectorsDCS/DCSDataPointHint.h
@@ -33,7 +33,7 @@ using HintType = std::variant<DataPointHint<double>,
                               DataPointHint<float>,
                               DataPointHint<uint32_t>,
                               DataPointHint<int32_t>,
-                              DataPointHint<char>,
+                              DataPointHint<short>,
                               DataPointHint<bool>,
                               DataPointHint<std::string>>;
 
--- a/Detectors/DCS/include/DetectorsDCS/DataPointCompositeObject.h
+++ b/Detectors/DCS/include/DetectorsDCS/DataPointCompositeObject.h
@@ -225,7 +225,7 @@ struct alignas(128) DataPointCompositeObject final {
       double double_value;
       uint32_t uint_value;
       int32_t int_value;
-      char char_value;
+      short char_value;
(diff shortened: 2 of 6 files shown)


chiarazampolli:
Hello, 
I would ping Dominik Voscek and Peter Bond from DCS, maybe they can apply the same patch to https://gitlab.cern.ch/adapos/adapro
Chiara

shahor02:
@TimoWilken why do we need to modify the DCS class? Instead I would simply specialize the `generateRandomDataPoints<char>` as

```
template <>
std::vector<o2::dcs::DataPointCompositeObject> generateRandomDataPoints<char>(const std::vector<std::string>& aliases, char minValue, char maxValue, std::string refDate)
{
  std::vector<o2::dcs::DataPointCompositeObject> dpcoms;
  std::random_device rd;
  std::mt19937 mt(rd());
  std::uniform_int_distribution<short> distrib(short(minValue), short(maxValue));
  auto [seconds, msec] = getDate(refDate);
  for (auto alias : expandAliases(aliases)) {
    dpcoms.emplace_back(o2::dcs::createDataPointCompositeObject(alias, char(dist(mt)), seconds, msec));
  }
  return dpcoms;
}
```

chiarazampolli:
Actually, now that I see better the problem, I agree with Ruben. 
In addition, I am afraid that changing from char to short will not work in combination with DCS code which will send chars (unless they modify their code too, but it might mean to modify also detector code etc). The solution proposed by Ruben seems appropriate to me.

TimoWilken:
Thanks @shahor02, @chiarazampolli, I'll try it out!

pzhristov:
@TimoWilken Probably we do not need any patch until we stay with Xcode 14.2.

PR #11123, opened 2023-04-06
[EMCAL-842, EMCAL-912] Add more calibration objects to CalibLoader, use calib loader in reco workflow

mfasDa:
Added support for calibration objects:
- Temperature calibration
- Time slewing correction (not yet implemented)
- FEE DCS parameters
- BCM scale factors
Added support for CCDB-based configurable params:
- Calibration params (not yet implemented)
- Reconstruction params
- Simulation params

Use CalibLoader in reco workflow for CCDB access of
- RecoParams

Access in the reco workflow is prepared for
- FeeDCS

### DIFF
--- a/Detectors/EMCAL/workflow/include/EMCALWorkflow/CalibLoader.h
+++ b/Detectors/EMCAL/workflow/include/EMCALWorkflow/CalibLoader.h
@@ -21,8 +21,13 @@ namespace o2
 namespace emcal
 {
 class BadChannelMap;
+class FeeDCS;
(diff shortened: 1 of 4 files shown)


mfasDa:
The commit prepares the RawToCellConverter for the access of the FeeDCS params, which will be needed in order to identify inactive links during the reconstruction. At the same step the update of the RecoParam object is already moved to the CalibLoader.

shahor02:
local fullCI passed, merging.

PR #11125, opened 2023-04-06
[EMCAL-688] Fix check if lookup table is initialised

mhemmer-cern:
- `if (!getLookUpInit())` was `if (getLookUpInit())` in isExoticCell, which lead to throwing errors of uninitialised lookup table when the table was in fact initialised.
- Connected to PR #11062


mhemmer-cern:
Tagging @shahor02

mfasDa (approved):
Thanks for fixing!

PR #11129, opened 2023-04-06
ZDC fix formatting of log message

martenole:
ZDC calibration is crashing due to this, see e.g.

```
E | 3 | 06/04/2023 | 20:40:56.459 | epn-calib1 |   | STDERR | stderr/zdc-tdc-calib_calib10_20 | 2eXRdx6XZ13 | 534169 |   |   | stderr: terminate called after throwing an instance of 'fmt::v9::format_error'
```
In general `LOGP(info, "Message {} with number {}", foo, bar)` might be less error prone than the printf syntax `LOGF` which requires type specifiers


martenole:
Trivial, and obviously anyhow not caught by the CI, so merging

martenole:
Well, the formatting should at least be correct...

PR #11131, opened 2023-04-07
Including MCS effects in secondary vertexing for forward tracks

mcoquet642:
I add a simple function to `TrackParCovFwd` to propagate forward tracks to vertex including MCS effects. I then use it in the FwdDCAFitter to propagate forward tracks to the PCA.


Common/DCAFitter/include/DCAFitter/FwdDCAFitterN.h
```diff
@@ -158,6 +160,11 @@ class FwdDCAFitterN
   void setMinParamChange(float x = 1e-3) { mMinParamChange = x > 1e-4 ? x : 1.e-4; }
   void setMinRelChi2Change(float r = 0.9) { mMinRelChi2Change = r > 0.1 ? r : 999.; }
   void setUseAbsDCA(bool v) { mUseAbsDCA = v; }
+  void setMatLUT(o2::base::MatLayerCylSet* m)
```
shahor02: MatLUT comes from the CCDB hence it must be const pointer.

shahor02 (changes requested):
Hi @mcoquet642 

Technically, just one request for a change below. But note that the current MatLut class is
1) optimized for barrel track: it is a set of cylindrical layers with phi/z segmentation with a fast search algorithm. In the long run, MFT/MUON could think about a similar implementation for the forward tracks, i.e. a set of layers normal to the Z axis with X and Y segmentation. 

2) current MatLut is created with the layers R/Z coverage optimized for the ITS: you can see current segmentation in https://github.com/AliceO2Group/AliceO2/blob/4ac6435fc6480e1f0f994923ea0ee55537a9d46a/Detectors/Base/test/buildMatBudLUT.C#L194-L247. Note that for `r<5.2` the region `|z|>20` is not covered, and for `5.2 < r<19` the `|z|>40` is not covered (will contribute 0 mat. budget) and this whole region is covered by a single cyl. slice (i.e. radially uniform mat. budget).

3) To have a reference it is worth having the possibility to switch to a slow but precise mat. budget query via TGeo, as it is done in the https://github.com/AliceO2Group/AliceO2/blob/4ac6435fc6480e1f0f994923ea0ee55537a9d46a/Detectors/Base/src/Propagator.cxx#L674-L689

mcoquet642:
Hi @shahor02, Thanks for the comments. Indeed, it would be good at some point to have an optimized propagation for forward tracks. However for now, I was mainly trying to capture the effects of MCS through the beam pipe. From your comments, it seems to me that these are still included if I use the MatLut class, am I correct ? In any case I added the option to have the full material budget estimate from TGeo, as you suggested. Thanks

Common/DCAFitter/include/DCAFitter/FwdDCAFitterN.h
```diff
@@ -1256,6 +1269,26 @@ void FwdDCAFitterN<N, Args...>::print() const
...
+  // propagate track to vertex including MCS effects if material budget included, simple propagation to Z otherwise
+  if (mUseMatBudget) {
+    float x2x0 = 0;
+    if (mTGeoFallBackAllowed) {
+      auto geoMan = o2::base::GeometryManager::meanMaterialBudget(t.getX(), t.getY(), t.getZ(), p[0], p[1], p[2]);
+      x2x0 = (float)geoMan.meanX2X0;
+    } else {
+      auto mb = mMatLUT->getMatBudget(t.getX(), t.getY(), t.getZ(), p[0], p[1], p[2]);
+      x2x0 = (float)mb.meanX2X0;
+    }
```
shahor02: I think the logic for the fall-back should be the opposite: TGeo must be used if MatLUT is not available:
```
if (mMatLUT) {
    auto mb = mMatLUT->getMatBudget(t.getX(), t.getY(), t.getZ(), p[0], p[1], p[2]);
    x2x0 = (float)mb.meanX2X0;
} else if (TGeoFallBackAllowed) {
    auto geoMan = o2::base::GeometryManager::meanMaterialBudget(t.getX(), t.getY(), t.getZ(), p[0], p[1], p[2]);
    x2x0 = (float)geoMan.meanX2X0;
} 
```

shahor02 (commented):
Hi @mcoquet642 

Yes, the beampipe materials will be accounted when using MatLUT, provided the track is crossing it at |Z|<30 cm. which should be enough for MFT.

Please see a comment below on using of mTGeoFallBackAllowed.

I've noticed that you are maintaining 2 copies of FwdDCAFitterN: in the `Common/DCAFitter` and in the `Detectors/Vertexing`. Since the former was created to provide a standalone library for the Analysis, the copy in the Vertexing can be eliminated (after making sure that O2Physics does not use it. Surely, this can be done in separate PR.

Common/DCAFitter/include/DCAFitter/FwdDCAFitterN.h
mcoquet642: Fixed, thanks

mcoquet642:
@shahor02 Thanks for the comments. I will remove the `Detectors/Vertexing` copy in another PR.

Common/DCAFitter/include/DCAFitter/FwdDCAFitterN.h
```diff
@@ -290,6 +301,7 @@ class FwdDCAFitterN
   float mMinRelChi2Change = 0.98;   // stop iterations is chi2/chi2old > this
   float mMaxChi2 = 100;             // abs cut on chi2 or abs distance
   float mMaxDist2ToMergeSeeds = 1.; // merge 2 seeds to their average if their distance^2 is below the threshold
+  o2::base::MatLayerCylSet* mMatLUT; // use to compute material budget to include MCS effects
```
shahor02: the CI error is genuine, this should be a `const o2::base::MatLayerCylSet* mMatLUT = nullptr;`

mcoquet642: Fixed

mcoquet642:
Is the failing CI still related to this PR?

shahor02:
No, the fullCI fails regularly for unrelated reasons. I'll run it locally and merge once it is passed.

shahor02:
fullCI tested locally.

PR #11132, opened 2023-04-07
[FOCAL-10] Identify ECAL subsystems using source ID

mfasDa:
- E-pixels: Use ITS source ID (32)
- E-pads: Use current source ID (43)


shahor02 (approved):

mfasDa:
Needed as endpoint to identify the two subsystems is unreliable. The assignment  0 - pixels, 1 - pads is unreliable since it was valid only for the first testbeam, where both systems were read out with 1 CRU on different endpoints. Since endpoints changed decoding was no longer working on new data. Using source ID is safer, and also still compatible with raw data from the last testbeam.

PR #11135, opened 2023-04-08
Trigger digits absId fix; time fix

peressounko:


peressounko:
Dear @shahor02 , could you please approve this PR? It fixes problems in MC AO2D production which later cause crash in clusterization. Does not affect real data.

shahor02 (approved):

PR #11136, opened 2023-04-09
DPL: use uv_now rather than uv_hrtime for metrics

ktf:


ktf:
Tested on hyperloop, staging (TOF: FLP + QC) as well. No regressions.

PR #11048, opened 2023-03-28
wip ITS-vertexer: Make cut on XY a configurable param

mconcas:

PR #11142, opened 2023-04-11
ITS-vertexer: Make cut on XY a configurable param

mconcas:
- Make XY cut beam-vertex distance cut
- Add tracklet resolution cut 100um

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/include/ITStracking/Configuration.h
+++ b/Detectors/ITSMFT/ITS/tracking/include/ITStracking/Configuration.h
@@ -104,8 +104,10 @@ struct VertexingParameters {
   float pairCut = 0.04f;
   float clusterCut = 0.8f;
   float histPairCut = 0.04f;
-  float tanLambdaCut = 0.002f; // tanLambda = deltaZ/deltaR
-  float lowMultXYcut2 = 0.01f; // XY cut for low-multiplicity pile up
+  float tanLambdaCut = 0.002f;     // tanLambda = deltaZ/deltaR
+  float lowMultBeamDistCut = 0.1f; // XY cut for low-multiplicity pile up
+  int vertRadiusNsigmaCut = 4;     // N sigma cut for vertex XY
+  float vertRadiusSigma = 0.495f;  // sigma of vertex XY
   float maxZPositionAllowed = 25.f;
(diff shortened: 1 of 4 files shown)


mconcas:
@shahor02 this replaces #11048. From CLI I opened a PR using a branch on AliceO2Group/AliceO2 rather than mconcas/AliceO2 and then I was not able to push further changes.
Sorry for the mess.

I ran on the CTF: `o2_ctf_run00523308_orbit0319097116_tf0000000026_epn224.root`

Here the ntracks distribution when we run on pp with default cuts.
- vertNsigmaCut = 4
- vertRadiusSigma = 500 um
- trackletSigma = 100 um

Screenshot 2023-04-11 at 12 25 03

Seeding XY dist is contained into a 0.5 mm diameter circle around the beam:
Screenshot 2023-04-11 at 12 29 56

shahor02 (approved):

davidrohr:
@mconcas @shahor02 : What is the status of the custom settings? Do I still need to apply some custom configKeyValues in the async reco, or is it meanwhile the default / or added to the async setenv_extra.sh?

mconcas:
> @mconcas @shahor02 : What is the status of the custom settings? Do I still need to apply some custom configKeyValues in the async reco, or is it meanwhile the default / or added to the async setenv_extra.sh?

Defaults are now changed so as to produce the presented results.
No additional parameters need to be passed unless we found these new defaults are not good.

davidrohr:
ok, perfect, thx, then I'll retry once more the async reco benchmark just to be sure.

chiarazampolli:
Hello @mconcas , 
Just to be sure: do we need to change anything in https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/configurations/asyncReco/setenv_extra.sh ? Which settings should one check, in case, to be sure?
Chiara

mconcas:
> Hello @mconcas , Just to be sure: do we need to change anything in https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/configurations/asyncReco/setenv_extra.sh ? Which settings should one check, in case, to be sure? Chiara

Hi, nothing should be changed there for the time being.
I changed the default behaviour, so the modification should come with no further additions to the params.

PR #11144, opened 2023-04-11
[FOCAL-10] Auto-determine whether timeframe has payload from either of the FOCAL-E subsystems

mfasDa:
Determine subsytems for which payload has been
found and use that in the event building. The option
to disable subsystems separately is kept.

### DIFF
--- a/Detectors/FOCAL/workflow/include/FOCALWorkflow/RawDecoderSpec.h
+++ b/Detectors/FOCAL/workflow/include/FOCALWorkflow/RawDecoderSpec.h
@@ -77,6 +77,8 @@ class RawDecoderSpec : public framework::Task
   bool mUsePadData = true;
   bool mUsePixelData = true;
   bool mFilterIncomplete = false;
+  bool mTimeframeHasPadData = false;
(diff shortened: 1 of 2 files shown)


mfasDa:
Information which subsytems are available are needed in the event builder: Since the trigger information is only available from the pixel data the event builder behaves differently in case both subsystems are there, or one is missing:

- Both subsystems are there: Sort events per orbit linearly, map triggers and pad events based on index in array
- Only pixels are there: Connection between trigger and pixel data trivial
- Only pads are there: Use orbit and BC 0 for all pad events in the same orbit.

The event builder does some error checking. In case a subsytem was missing the event builder will raise an error because it cannot match triggers and data. Users needed to specify if one subsystem is missing. In case we run it in the reconstruction chain on the FLP the approach won't work - the decoder needs to determine in an automatic way whether the subsystem was off, and pass this to the event builer. We use the end-of-payload for this. Since decoding of either of the subsystem is kept the option to disable the reconstruction of either of the subsystems is preserved.

mfasDa:
CI errors unrelated

PR #11145, opened 2023-04-11
ITS decoder: fixed bug with getNextChipData

iravasen:
@shahor02 I open this PR to fix a bug introduced with PR11069 (https://github.com/AliceO2Group/AliceO2/pull/11069/files) when the return value of decodeNextTrigger was changed (request from my side)

This in particular fixes an issue we observe at P2 (after this morning's SW upgrade) with ITS QC FHR task where the loop on `decodeNextChipData` is infinite without this fix, i.e. `decodeNextTrigger` never return 0 and hence `decodeNextChipData` never return a `nullptr` (we operate with decodenexttriggerauto = true). 

Here the loop of the FHR task affected, mentioned just above: https://github.com/AliceO2Group/QualityControl/blob/4597ae73b864e7748ddc33da02106523c39502bb/Modules/ITS/src/ITSFhrTask.cxx#L445

### DIFF
--- a/Detectors/ITSMFT/common/reconstruction/src/RawPixelDecoder.cxx
+++ b/Detectors/ITSMFT/common/reconstruction/src/RawPixelDecoder.cxx
@@ -330,7 +330,7 @@ ChipPixelData* RawPixelDecoder<Mapping>::getNextChipData(std::vector<ChipPixelDa
     }
   }
   // will need to decode new trigger
-  if (!mDecodeNextAuto || !decodeNextTrigger()) { // no more data to decode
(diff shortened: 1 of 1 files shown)


iravasen:
I proposed to patch the current FLP suite with this modification.

shahor02 (approved):
@iravasen thanks for the fix, indeed, missed that I use this method in many places. 
Will merge once at least 1 CI is green.

vascobarroso:
@shahor02 if this can be included in tonight's O2 tag we can deploy it tomorrow at P2.

PR #11150, opened 2023-04-12
[EMCAL-548] Configurable output subspecification for DigitsToCellConverter

mfasDa:


mfasDa:
Needed in order to publish on subspec 1, in order to chain the MC reco workflow with the cell-recalibration-workflow for masking of bad channels (and possible MC time calibration)

mfasDa:
CI errors unrelated

PR #11152, opened 2023-04-13
tof fix to recover tof compressor on epn

noferini:
Hi @martenole 
This is to get tof compressor working again on epn.
It now works on my laptop but it should be checked on epn. Can you try it?
Since the problem spotted here
https://alice.its.cern.ch/jira/browse/O2-3681
there is no way at the moment to get tof-compressor and decoder in the same chain if other inputs are required by the decoder (DIST-STF or GRPRequest).
I can easily disable DIST-STF but to disable GRPRequest I need to pass NorbitPerTF as external param, i.e. hardcoded at 32.
This is a temporary solution till we solve the problem with the framework.

### DIFF
--- a/Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
+++ b/Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
@@ -45,7 +45,9 @@ void CompressedDecodingTask::init(InitContext& ic)
 {
   LOG(debug) << "CompressedDecoding init";
 
-  o2::base::GRPGeomHelper::instance().setRequest(mGGCCDBRequest);
+  if (mNorbitsPerTF == -1) {
+    o2::base::GRPGeomHelper::instance().setRequest(mGGCCDBRequest);
+  }
 
   mMaskNoise = ic.options().get<bool>("mask-noise");
   mNoiseRate = ic.options().get<int>("noise-counts");
@@ -141,9 +143,12 @@ void CompressedDecodingTask::run(ProcessingContext& pc)
 {
   mTimer.Start(false);
 
-  o2::base::GRPGeomHelper::instance().checkUpdates(pc);
-  const uint32_t norbits = o2::base::GRPGeomHelper::instance().getGRPECS()->getNHBFPerTF();
-  Utils::setNOrbitInTF(norbits);
+  int norbits = mNorbitsPerTF;
+  if (norbits == -1) {
+    o2::base::GRPGeomHelper::instance().checkUpdates(pc);
+    norbits = o2::base::GRPGeomHelper::instance().getGRPECS()->getNHBFPerTF();
+    mNorbitsPerTF = norbits;
+  }
   mDecoder.setNOrbitInTF(norbits);
 
   static bool isFirstCall = true;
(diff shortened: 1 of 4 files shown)


noferini:
Adding also @preghenella as watcher

Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
```diff
@@ -141,8 +143,12 @@ void CompressedDecodingTask::run(ProcessingContext& pc)
...
-  o2::base::GRPGeomHelper::instance().checkUpdates(pc);
-  const uint32_t norbits = o2::base::GRPGeomHelper::instance().getGRPECS()->getNHBFPerTF();
+  int norbits = mNorbitsPerTF;
+  if (norbits == -1) {
+    o2::base::GRPGeomHelper::instance().checkUpdates(pc);
+    norbits = o2::base::GRPGeomHelper::instance().getGRPECS()->getNHBFPerTF();
+    mNorbitsPerTF = norbits;
+  }
   Utils::setNOrbitInTF(norbits);
   mDecoder.setNOrbitInTF(norbits);
```
martenole, line 152: these two `setNOrbitInTF` methods do the same and the number also needs to be set once in a run. Could you fix this as well?

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -171,7 +171,7 @@ fi
 [[ -z $DISABLE_ROOT_OUTPUT ]] || needs_root_output o2-gpu-reco-workflow && GPU_OUTPUT+=",send-clusters-per-sector"
 
 has_detector_flp_processing CPV && CPV_INPUT=digits
-! has_detector_flp_processing TOF && TOF_CONFIG+=" --ignore-dist-stf"
+! has_detector_flp_processing TOF && TOF_CONFIG+=" --orbits-per-tf 32 --ignore-dist-stf --local-cmp"
```
martenole: could you make the 32 configurable from outside?
E.g.
```suggestion
! has_detector_flp_processing TOF && TOF_CONFIG+=" --orbits-per-tf ${NHBPERTF:-32} --ignore-dist-stf --local-cmp"
```

martenole (commented):
Hi Francesco,
looks good to me, just two minor comments. I will try it on the EPNs and let you know.
Cheers,
Ole

martenole:
Works also on the EPNs @noferini 
No more warnings and the tofdigits have the expected larger size depending on how often I inject the TF in the chain

Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
noferini: I'll fix it. The call is done only once. After the first call I set mNorbitsTF from -1 to the measured value and this will prevent to call it again (and also to do something in the finalizeCCDB call)

prodtests/full-system-test/dpl-workflow.sh
noferini: done

martenole (approved):

martenole:
ran fullCI locally, merging

PR #11155, opened 2023-04-13
TRD preparations for start-stop-start mechanism

martenole:
For details see https://indico.cern.ch/event/1270050/contributions/5335478/attachments/2620040/4530862/EOS_and_SSS.pdf

Also some minor cleanup is included in this PR


bazinski (dismissed):

f3sch (dismissed):
LGTM! This just affects the calibration dpl workflows, right?

martenole:
> LGTM! This just affects the calibration dpl workflows, right?

The logic should stay the same, its only when running online at P2 and we stop and start a new run without tearing down the environment. At some point in the future this might come

martenole:
Failures unrelated, merging

PR #11159, opened 2023-04-14
Fix P2 calculation for Z>1

fmazzasc:
@shahor02


shahor02 (commented):
Thanks for spotting, @fmazzasc 
Trivial, merging.

PR #11160, opened 2023-04-14
[ITS3] Introduce gap in phi and add more customizable parameters

fgrosa:


fgrosa:
Hi @TimoWilken I am not sure why the clang format fails for this PR: it seems that the test is passed, but then the bot fails to push some refs. Could you please have a look? Thanks!

TimoWilken:
Hi @fgrosa, looks like a transient failure... I've restarted the check and this time it worked. Let me know if it fails again!

mconcas (approved):

PR #11165, opened 2023-04-15
Fix matchingChi2 and invariant mass calculations

fmazzasc:
@mpuccio


mconcas (approved):

ddobrigk:
@fmazzasc great, I spotted this too in testing with AO2Ds and with this it's already fixed! Thanks!

PR #11166, opened 2023-04-16
fix in TOF encoder (MC) in eventwords

noferini:
This is to fix a bug when decoding raw data encoded from sim (also in synthetic runs).
In the past behavior almost all events are flagged with mismatched word error preventing clusterization (digits are filtered out in clusterization when that flag is on.
@martenole, this fixes the problem we saw when comparing digits and clusters from sim

It would be nice to update after this fix the synthetic raw data we are re-playing in synthetic runs.
@shahor02, the bug affects only reconstructions which uses TOF clusters

### DIFF
--- a/Detectors/TOF/reconstruction/src/Encoder.cxx
+++ b/Detectors/TOF/reconstruction/src/Encoder.cxx
@@ -228,7 +228,9 @@ void Encoder::encodeTRM(const std::vector<Digit>& summary, Int_t icrate, Int_t i
 
   // set TRM data size
   int neventwords = getSize(trmheader, mUnion[icrate]) / 4 + 1;
-  neventwords -= neventwords / 4 * 2;
+  if (mOldFormat) {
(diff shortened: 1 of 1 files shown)


davidrohr:
@noferini : thx. I anyway wanted to update the SYNTHETIC data set. Do I understand correctly that this fix is in the encoder, i.e. I can generate a new SYNTHETIC data set with this, and we can replay it at P2 without changing the software version there? I.e. we don't have to do it in sync?

noferini:
> @noferini : thx. I anyway wanted to update the SYNTHETIC data set. Do I understand correctly that this fix is in the encoder, i.e. I can generate a new SYNTHETIC data set with this, and we can replay it at P2 without changing the software version there? I.e. we don't have to do it in sync?

Yes, there is no need to update O2 at P2, just the raw files from MC.

martenole:
Thanks Francesco!

PR #11172, opened 2023-04-17
ATO-609: Additional ITS material added (to be MOVED in ITS code)

marslandALICE:
This part of the code should be moved to "ITS" but for the time being it should be OK

### DIFF
--- a/Detectors/Passive/src/Cave.cxx
+++ b/Detectors/Passive/src/Cave.cxx
@@ -31,6 +31,7 @@
 #include "TGeoManager.h"
 #include "TGeoVolume.h"
 #include "TGeoPgon.h"
+#include "TGeoPcon.h"
 #include "TGeoTube.h"
 #include "TGeoCompositeShape.h"
 using namespace o2::passive;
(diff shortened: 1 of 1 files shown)


davidrohr:
@sawenzel : could you have a look how to proceed. This is needed for more realistic TPC number of clusters. Apparently it should be moved into the ITS part. Who would be responsible?
Or could we just merge it as is, and move it later?

sawenzel:
Sorry. This PR escaped my attention. Since this material is ITS, I think we should directly put it as part of the ITS detector.
It should be easy enough to move it directly there. @mconcas will try to do so unless this is hyper-urgent?

mconcas:
> Sorry. This PR escaped my attention. Since this material is ITS, I think we should directly put it as part of the ITS detector. It should be easy enough to move it directly there. @mconcas will try to do so unless this is hyper-urgent?

I am trying right now to append this to some lower-level ITS-related service volumes. Let's say that if this is not going to converge by tomorrow, we will merge it as it is now and re-iterate on it later.

I know this thing has been a long-standing topic that we promised to take care of, and I understand the current urgency. Therefore I would not mind having this merged even earlier if you need that.

davidrohr:
Hi, this is not super urgent. I need it at the moment for local simulations, so I can just merge it locally.
If you can clean it up and merge it some time next week, that would be totally fine.

sawenzel:
@mconcas : I think this was done meanwhile via an ITS PR. Can you confirm and then we can close this I believe.

mconcas:
> @mconcas : I think this was done meanwhile via an ITS PR. Can you confirm and then we can close this I believe.

Yes this was done in the linked PR. I guess we can close indeed.

PR #11174, opened 2023-04-17
TRD noise update default parameter

martenole:
Seen today that 100 billion takes too long, because with only a single EPN the largest fraction of input data is dropped because of backpressure


martenole:
trivial, merging

PR #11175, opened 2023-04-17
Restore collision-sorting for tracks with ST

mpuccio:


mpuccio:
Hi @shahor02, @ddobrigk, the AO2D are currently broken with the ST enabled. It would be great if we could get this merged to continue the development and enable the strangeness tracking in the default reconstruction chain.

shahor02:
Will run the FST locally, then merge

shahor02:
local fullCI passed, merging.

PR #11176, opened 2023-04-17
Workflow to print tracks/vertices matching to paricular timestamps

shahor02:
Example:
```
o2-bc-tracks-dump-workfow --sel-bc 1177488263775,1177678567451,1177680066700,1177680575625,1178253418915,1178257212280 --bc-margin 50 --shm-segment-size 24000000000
```
 to print tracks and vertices matching to global BC IDs passed via --sel-bc within the margin of --bc-margin


shahor02:
code-check passed, FST is irrelevant for this PR.

PR #11179, opened 2023-04-18
Added vertex resolution plots

reynier0611:
I changed the indentation and that's why the two codes seem so different, but in reality the only actual change is the addition of vertex resolution plots.


mconcas:
> I changed the indentation and that's why the two codes seem so different, but in reality the only actual change is the addition of vertex resolution plots.

This is also why the clang-format test fails. :)
A PR here: https://github.com/reynier0611/AliceO2/pull/1 has been automatically opened to fix the format change.

PR #11184, opened 2023-04-18
[EMCAL-840] Add container for EMCal time slew params

jokonig:
- EMCal cell time is slightly dependent on the cell energy
- To take out the residual difference as function of the cell energy, a time slewing correction is needed
- The TimeCalibrationSlewingParams provide a simple container to store a polynomial function of 3rd order that can be evaluated at a certain energy
- The implementation of the actual correction will follow in a second step


jokonig:
@shahor02 could you have a look if this can be merged? The error in build/O2/o2-cs8 is coming from something unrelated in O2Physics.

mfasDa (approved):

chiarazampolli:
Pinging @davidrohr , since Ruben is away. 
this is needed for https://alice.its.cern.ch/jira/browse/O2-3685, right?

PR #11185, opened 2023-04-18
[FOCAL-10] Use FEE ID to distinguish FOCAL-E systems

mfasDa:
Configure based on TB and lab setup:
- Pads: Only 1 FEE (aggregator), ID 0xcafe
- Pixels: All other FEE IDs

### DIFF
--- a/Detectors/FOCAL/workflow/src/RawDecoderSpec.cxx
+++ b/Detectors/FOCAL/workflow/src/RawDecoderSpec.cxx
@@ -48,8 +48,7 @@ void RawDecoderSpec::run(framework::ProcessingContext& ctx)
 
   int inputs = 0;
   std::vector<char> rawbuffer;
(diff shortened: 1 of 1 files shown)


mfasDa:
Replaces #11132: In lab tests we realised that the STF builder will not accept the data with the current source ID reported (32 for pixels. 43 for pads), since it requires the source ID to be the DAQ ID which is 39 for both FOCAL subsystems. Also falling back to CRU endpoints no longer works. We must identify - like other detectors - the detector segment based on the FEE ID. For the moment we can hard code the FEE ID since we have only one FEE for the PADs with the ID 0xcafe, so we know all other FEEs are connected to pixel readout. In a second step we will introduce a mapper which maps the FEE ID to a subsystem and a position within the subsystem, which can take the connection FEE ID - mapping information from the CCDB or a text file so it becomes configurable.

PR #11186, opened 2023-04-18
DPL_REPORT_PROCESSING!=0 value used as downscaling factor for reporting

shahor02:


davidrohr (changes requested):
I think this will not work correctly, setting `DPL_REPORT_PROCESSING=0` on the EPNs will trigger an integer divide by 0.
I think we should initialize prescale = 1 and remove the else branch, and then in `.matcher` we can do
```
static bool report = (CommonServices::getDeploymentMode() == DeploymentMode::OnlineDDS || forceReport) && prescale != 0;
```
This would then also allow us to disable the reporting in online mode if desired.

shahor02:
Thanks for spotting, I've modified in a slightly different way.

PR #11189, opened 2023-04-19
Remove extra factor in front of O2JOBID

chiarazampolli:
Related to https://github.com/AliceO2Group/O2DPG/pull/993

### DIFF
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -46,7 +46,7 @@ if [[ -z ${TIMEFRAME_RATE_LIMIT:-} ]] && [[ $DIGITINPUT != 1 ]]; then
   ! has_detector TPC && TIMEFRAME_RATE_LIMIT=$(($TIMEFRAME_RATE_LIMIT * 4))
   [[ ! -z ${EPN_GLOBAL_SCALING:-} ]] && TIMEFRAME_RATE_LIMIT=$(($TIMEFRAME_RATE_LIMIT * $EPN_GLOBAL_SCALING))
 fi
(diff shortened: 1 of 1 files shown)


davidrohr (changes requested):
I think we should not remove this factor. Perhaps we can reduce it a bit.
The idea behind it is that NUMAID is not guaranteed to be 0, so I multiple the O2JOBID with a factor to make sure that 2 O2JOBIDs cnanot collide, if they run on different NUMAIDs.

However, I think what we could do instead is
```
${O2JOBID:-$NUMAID})
```
i.e. we just use the O2JOBID without the NUMAID if it is set. As long as we don't run 2 workflows in the same O2JOB, that will be fine.

chiarazampolli:
Hi @davidrohr , 
I don't understand, sorry. Multiplying by 10 gives just a higher O2JOBID. I removed the factor otherwise we get a number that does not fit int32. 
I can also implement your suggestion, but I don't see why removing `*10` is not ok (sorry, maybe it is naive).

davidrohr:
If you remove the *10, and you have o2jobid=10 with NUMAID=1, and O2jobid=11 with NUMAID=0 on the same node, they will collide. Thus I proposed to use `${O2JOBID:-$NUMAID})`.

(The original idea was that o2jobid and numaid are orthogonal, and I multiply o2jobid by 10, so we could even have multiple numaid in the same o2jobid. But I guess that won't be needed.)

chiarazampolli:
Ah ok, so the 10 is a kind of shift for the NUMAID. Now I get it. Since I thought that we can only have 1 or 2, I was confused. 
Ok, I will change as you suggested.

davidrohr (approved):

PR #11191, opened 2023-04-19
[FOCAL-12] FOCAL Pad Pedestal DA

mfasDa:
* [FOCAL-12] Add structure for Pad pedestals

  One pedestal value per layer and channel.
  Throwing InvalidChannelException in case of
  invalid channel in filling and access to non-existing
  channel in reading.

* [FOCAL-12] Pedestal calibration DA for Pads

  Pedestal calibration based on ADC distribution
  with random trigger for each layer and channel
  separately. Various methods (Max bin, mean,
  gauss fit) are available for the pedestal determination.

### DIFF
--- a/Detectors/FOCAL/calib/CMakeLists.txt
+++ b/Detectors/FOCAL/calib/CMakeLists.txt
@@ -9,7 +9,12 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-add_subdirectory(base)
-add_subdirectory(calib)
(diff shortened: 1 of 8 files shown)


mfasDa:
Preparation for the pedestal calibration: Pedestal objects are already existing for the last testbeam in text files, they will be ported to PadPedestal objects to be stored in the CCDB (test or CCDB on FOCAL FLP) to be made available at analysis level. For the next testbeam we are planning to subtract pedestals in the pad QC online, requireing access to pedestals via the CCDB.

PR #11194, opened 2023-04-20
Fix processing of nodata=1/internal=1 ITS/MFT triggers

shahor02:


martenole (approved):

martenole:
This yields very verbose debug output. By default `mVerbosity` is set to 0 (`=GBTLink::Verbosity::VerboseErrors`) for the first decoder instance and to -1 (`=GBTLink::Verbosity::Silent`) for the others. Now unfortunately I cannot even disable the verbosity with `ARGS_EXTRA_PROCESS_o2_itsmft_stf_decoder_workflow="--decoder-verbosity -1"` since `if (mVerbosity)` is true for -1.
So one more change is needed.

Detectors/ITSMFT/common/reconstruction/src/RawPixelDecoder.cxx
```diff
@@ -181,6 +181,9 @@ bool RawPixelDecoder<Mapping>::doIRMajorityPoll()
         mIRPoll[link.ir]++;
       } else if (link.status == GBTLink::StoppedOnEndOfData || link.status == GBTLink::AbortedOnError) {
         link.statusInTF = GBTLink::StoppedOnEndOfData;
+        if (mVerbosity) {
```
martenole:
```suggestion
        if (mVerbosity != GBTLink::Verbosity::Silent) {
```

Detectors/ITSMFT/common/reconstruction/src/RawPixelDecoder.cxx
```diff
@@ -194,9 +197,15 @@ bool RawPixelDecoder<Mapping>::doIRMajorityPoll()
   }
   mInteractionRecordHB = mInteractionRecord;
   if (mInteractionRecord.isDummy()) {
+    if (mVerbosity) {
```
martenole:
```suggestion
    if (mVerbosity != GBTLink::Verbosity::Silent) {
```

Detectors/ITSMFT/common/reconstruction/src/RawPixelDecoder.cxx
```diff
@@ -194,9 +197,15 @@ bool RawPixelDecoder<Mapping>::doIRMajorityPoll()
...
+      LOG(info) << "doIRMajorityPoll: did not find any valid IR";
+    }
     return false;
   }
   mInteractionRecordHB.bc = 0;
+  if (mVerbosity) {
```
martenole:
```suggestion
  if (mVerbosity != GBTLink::Verbosity::Silent) {
```

martenole (changes requested):
I cannot commit to this PR. But the verbosity is too high I fear for running online, even if only one decoder does these reports. Would it be OK for ITS to run with `ARGS_EXTRA_PROCESS_o2_itsmft_stf_decoder_workflow="--decoder-verbosity -1"`? My suggestions should be added nevertheless, since otherwise one cannot disable the verbosity

martenole:
For some reason github let me commit now. But given the high verbosity of this we will have to run with `ARGS_EXTRA_PROCESS_o2_itsmft_stf_decoder_workflow="--decoder-verbosity -1"` or, alternatively, move the new info messages from this PR to debug severity. Please ITS experts let me know what you prefer

martenole:
Tested locally, this is needed at P2, so merging

shahor02:
Hi @martenole 

Sorry, I was running with a single lane and forgot about verbosity=-1 in case of pipelining. 
Thanks for the patch, but these logs are actually for debug only, even in case of verbosity==VerboseErrors they are not needed. 
I've just merged https://github.com/AliceO2Group/AliceO2/pull/11196 which makes these logs for `>=VerboseHeaders` only.

PR #11199, opened 2023-04-20
Fix ST MC label as well

mpuccio:


mpuccio:
@ddobrigk this should solve the issue you saw

ddobrigk:
Super thanks, @mpuccio ! Right now I am stuck with a simulation problem (@lbariogl is on the same boat...) and trying to debug that but once I get that sorted out I will test immediately...

ddobrigk:
In the meantime, this has been tested to be perfectly fine! @shahor02 can we please go ahead and merge it? Thanks a lot!!

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -476,18 +476,16 @@ void AODProducerWorkflowDPL::fillTrackTablesPerCollision(int collisionID,
   /// Add strangeness tracks to the table
   auto sTracks = data.getStrangeTracks();
   for (auto& collStrTrk : mCollisionStrTrk) {
-    if (collStrTrk.first < collisionID) {
+    if (mVtxToTableCollID.find(collStrTrk.first) == mVtxToTableCollID.end() || mVtxToTableCollID[collStrTrk.first] != collisionID) {
```
shahor02: @mpuccio @ddobrigk if I get it right, for every `collisionID` and every strange track the whole `mVtxToTableCollID` is scanned, which creates `Ncollisions^2 * Nstrange / 2` combinatorics. Isn't it better to fill in advance a
`unordered_map<int collID, std::vector strange_track IDs>` ?

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -1111,6 +1110,19 @@ void AODProducerWorkflowDPL::fillMCTrackLabelsTable(const MCTrackLabelCursorType
       }
     }
   }
+
+  // filling the tables with the strangeness tracking labels
+  auto sTrackLabels = data.getStrangeTracksMCLabels();
+  for (auto& collStrTrk : mCollisionStrTrk) {
+    if (mVtxToTableCollID.find(collStrTrk.first) == mVtxToTableCollID.end() || mVtxToTableCollID[collStrTrk.first] != vertexId) {
```
shahor02: same question about excessive mVtxToTableCollID check. Additionally, the check is still done even for the "orphan" vertexID=-1, though by construction there are no strange tracks in this case, right?

mpuccio:
Hi @shahor02, I implemented your request.

mpuccio:
Hi @shahor02, could we merge this?

PR #11201, opened 2023-04-20
[WIP][FOCAL-12] Pedestal calibration DA for Pads

mfasDa:
Pedestal calibration based on ADC distribution
with random trigger for each layer and channel
separately. Various methods (Max bin, mean,
gauss fit) are available for the pedestal determination.


mfasDa:
CCDB handling according to CPV Pedestal calibration.

Pedestal determination according to https://gitlab.cern.ch/alice-focal-readout/cru_decoder/-/blob/master/analysis/padData.C

mfasDa:
Merged into #11191

PR #11124, opened 2023-04-06
[EMCAL-889] Add filling of McCaloLabels in AODProducerWorkflowSpec

mhemmer-cern:
**AODProducerWorkflowSpec**:
- Add TMCCaloLabelCursor to addToCaloTable and fillCaloTable to fill the McCaloLabels
- Add new ConfigParamSpec `emc-select-leading` which will set the member variable `mEMCselectLeading`. If this flag is set to true, we will only save the MC Particle with the largest AmplitudeFraction for each cell. Otherwise we will save all MC Particles inside the table, except noise which we will never save.
- Moved fillCaloTable function call down after the fillMCParticlesTable so the MCParticlesTable are filled
- Moved `bcsMap.clear();` behind the filling of the MCParticle table

**DataFormatsPHOS/EventData**, **DataFormatsPHOS/EventHandler** and **DataFormatsPHOS/MCLabel**:
- Adjusted PHOS EventData class and MCLabel class to contain the needed members and functionalities, so that this will compile. Compare to https://github.com/AliceO2Group/AliceO2/pull/11117 where this was done for the EMCal.
- Increment ClassDefNV

**AnalysisDataModel**:
- Remove of Mask column for McCaloLabel table. After careful consideration we concluded that we do not need the mask column in Run3 and thus we decided to drop it.

**emc-reco-workflow**:
- Addition of inputsubspec option for emc-reco-workflow. Before the subspec could only be set for in AND output, now this is separate, which was needed to test the AODProducer chain with the recalib workflow.

- Connected to PR #11117 and PR #10670

PR #11205, opened 2023-04-21
[EMCAL-888] Removal of mask column in McCaloLabel table

mhemmer-cern:
- Remove of Mask column for McCaloLabel table. After careful consideration we concluded that we do not need the mask column in Run3 and thus we decided to drop it.


mhemmer-cern:
Needed for https://github.com/AliceO2Group/AliceO2/pull/11124

mhemmer-cern:
Tagging @jgrosseo @shahor02 @mfasDa

jgrosseo (approved):

PR #11208, opened 2023-04-21
fixing memory leak in tof compressor (epn mode)

noferini:
@martenole @ktf ,

I found the memory increases (mem leak) on EPNs and new tof compressor.
I reproduced the problem locally and I was able to solve with the change in this PR (tested by replaying with infinite loop over a tf from a cosmics run).

### DIFF
--- a/Detectors/TOF/compression/src/CompressorTask.cxx
+++ b/Detectors/TOF/compression/src/CompressorTask.cxx
@@ -156,7 +156,7 @@ void CompressorTask<RDH, verbose, paranoid>::run(ProcessingContext& pc)
     }
 
     v.resize(headerOut.payloadSize);
-    pc.outputs().adoptContainer(output, std::move(v), true, header::gSerializationMethodNone);
+    pc.outputs().adoptContainer(output, std::move(v));
   }
 }
 
--- a/Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
+++ b/Detectors/TOF/workflow/src/CompressedDecodingTask.cxx
@@ -212,7 +212,6 @@ void CompressedDecodingTask::decodeTF(ProcessingContext& pc)
     }
     contDeadBeef = 0; // if good data, reset the counter
   }
-
   /** loop over inputs routes **/
 
   std::vector<InputSpec> sel{InputSpec{"filter", ConcreteDataTypeMatcher{"TOF", "CRAWDATA"}}};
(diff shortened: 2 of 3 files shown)


ktf:
Looks good to me.

noferini:
I also added a change in the decoder to be compliant with dist subtimreframe (otherwise digits QC is not filled).
This allowed also to recover GRP from CCDB.
The solution I implemented consists in moving CRAWDATA inputs from optional to timeframe (since now compressor is running in front of the decoder input should be guaranteed).
It works fine locally, I would port also this change to be tested as soon as possible in the real life

davidrohr:
> I also added a change in the decoder to be compliant with dist subtimreframe (otherwise digits QC is not filled). This allowed also to recover GRP from CCDB. The solution I implemented consists in moving CRAWDATA inputs from optional to timeframe (since now compressor is running in front of the decoder input should be guaranteed). It works fine locally, I would port also this change to be tested as soon as possible in the real life

this depends. If the compressor runs on the same node as the decoder, the input is guaranteed. 
But if the compressor is on the FLPs and the decoder on the EPNs, then input is not guaranteed and it must be optional.
Perhaps you can also make this configurable?

noferini:
> > I also added a change in the decoder to be compliant with dist subtimreframe (otherwise digits QC is not filled). This allowed also to recover GRP from CCDB. The solution I implemented consists in moving CRAWDATA inputs from optional to timeframe (since now compressor is running in front of the decoder input should be guaranteed). It works fine locally, I would port also this change to be tested as soon as possible in the real life
> 
> this depends. If the compressor runs on the same node as the decoder, the input is guaranteed. But if the compressor is on the FLPs and the decoder on the EPNs, then input is not guaranteed and it must be optional. Perhaps you can also make this configurable?

Indeed it is configurable. This is done only if compressor run on epn. In flp case nothing was changed (see --local-cmp option
)

Detectors/TOF/compression/src/CompressorTask.cxx
```diff
@@ -156,7 +156,7 @@ void CompressorTask<RDH, verbose, paranoid>::run(ProcessingContext& pc)
     }
 
     v.resize(headerOut.payloadSize);
-    pc.outputs().adoptContainer(output, std::move(v), true, header::gSerializationMethodNone);
```
ktf, line 159: Yes, indeed the `true` should only be used for the CCDB object cases, were we keep them in memory for more than one timeframe.

noferini: Thanks for the check, then this explains what I observed

noferini:
If there are no objection I would merge so that we could try to validate the workflow on epn next week (no impact in the current workflow used for data taking).

PR #11209, opened 2023-04-21
[O2-3671] ITS-tracking: make road a template class

mconcas:
@shahor02: this reduces the global size of the Road class from 64 to 32 bytes, lowering the footprint of the tracker.

EDIT: After the latest commit current Road size is now 24Bytes.

CC-ing: @sawenzel

### DIFF
--- a/Detectors/ITSMFT/ITS/tracking/include/ITStracking/Road.h
+++ b/Detectors/ITSMFT/ITS/tracking/include/ITStracking/Road.h
@@ -13,8 +13,8 @@
 /// \brief
 ///
 
-#ifndef TRACKINGITSU_INCLUDE_ROAD_H_
-#define TRACKINGITSU_INCLUDE_ROAD_H_
+#ifndef TRACKINGCA_INCLUDE_ROAD_H
+#define TRACKINGCA_INCLUDE_ROAD_H
 
 #ifndef GPUCA_GPUCODE_DEVICE
 #include <array>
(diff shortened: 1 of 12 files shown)


Detectors/ITSMFT/ITS/tracking/include/ITStracking/Road.h
```diff
@@ -41,30 +42,68 @@ class Road final
...
+  }
+
+  // static constexpr int maxRoadSize = 13;
 
  private:
-  int mCellIds[mMaxRoadSize];
+  int mCellIds[maxRoadSize];
   int mRoadSize;
   int mLabel;
   bool mIsFakeRoad;
```
ktf, line 66: while you are at it why not changing mRoadSize to a char, swapping the order with mLabel, and so that this bool does not take 32 bits due to padding? I.e.:

```c++
int mLabel;
char mRoadSize;
bool isFakeRoad;
```

which is 8 bytes or even better if the number of labels is small:

```c++
short mLabel; // max 2^16 labels
char mRoadSize;
bool mIsFakeRoad;
```

which is just 4 bytes.

mconcas: > while you are at it why not changing mRoadSize to a char, swapping the order with mLabel, and so that this bool does not take 32 bits due to padding? I.e.:
> 
> ```c++
> int mLabel;
> char mRoadSize;
> bool isFakeRoad;
> ```
> 
> which is 8 bytes or even better if the number of labels is small:
> 
> ```c++
> short mLabel; // max 2^16 labels
> char mRoadSize;
> bool mIsFakeRoad;
> ```
> 
> which is just 4 bytes.

Indeed `mRoadSize` can be changed to char or unsigned char and swapped, to further reduce the size of `Road<5>`, good catch.
This would shrink the total size from 32 to 28. I am also checking whether we still use the label. This would reduce the total size down to 24, leaving just 2 bytes to padding.
Updates after local test.

mconcas: Tested with local simulation, no problems detected, worth a round of CI.

mconcas:
CI error seems not genuine, merging.

PR #11212, opened 2023-04-22
DPL: make sure caching strategy is provided explicitly

ktf:


ktf:
Just refactoring a bool to an enum. full CI works. Merging.

PR #11213, opened 2023-04-22
Fix deprecated sprintf

ktf:


ktf:
@davidrohr I am just not sure if the constexpr is allowed for the GPU code.

GPU/GPUTracking/Global/GPUChainTrackingClusterizer.cxx
```diff
@@ -145,13 +145,14 @@ std::pair<unsigned int, unsigned int> GPUChainTracking::TPCClusterizerDecodeZSCo
...
+          constexpr size_t bufferSize = 3 * std::max(sizeof(*rdh), sizeof(*hdr)) + 1;
+          char dumpBuffer[bufferSize];
           for (size_t i = 0; i < sizeof(*rdh); i++) {
-            sprintf(dumpBuffer + 3 * i, "%02X ", (int)((unsigned char*)rdh)[i]);
+            snprintf(dumpBuffer + 3 * i, bufferSize, "%02X ", (int)((unsigned char*)rdh)[i]);
           }
           GPUAlarm("RDH of page: %s", dumpBuffer);
           for (size_t i = 0; i < sizeof(*hdr); i++) {
-            sprintf(dumpBuffer + 3 * i, "%02X ", (int)((unsigned char*)hdr)[i]);
+            snprintf(dumpBuffer + 3 * i, bufferSize, "%02X ", (int)((unsigned char*)hdr)[i]);
```
davidrohr: sorry, but this is wrong. You should use in both cases `snprintf(dumpbuffer + 3 * i, 4, [...])`. Not all the write may write buffersize characters, that would go out of bounds. But anyway this is safe, since "%02X " is guaranteed to be 3 character + terminating 0.
Otherwise constexpr would not be a problem here.

davidrohr (changes requested):

GPU/GPUTracking/Global/GPUChainTrackingClusterizer.cxx
ktf: Yeah, sorry about that. Fixed.

PR #11214, opened 2023-04-23
[FOCAL-8] Adding debug mode for pad pedestal DA

mfasDa:
In case of the debug mode the extracted pedestals
are written to a file in the same way as if they were
a CCDB object


mfasDa:
Bypass to file particularly necessary in case we want to do i.e. a phase scan where we want to evaluate the pedestals and write them to file for postprocessing, but not send them to the CCDB in order to not mix up with regular pedestal runs.

PR #11215, opened 2023-04-23
[O2-3374] [rANS] ANS Entropy Coder Toolbox  and CTF-Coder Integration

MichaelLettrich:
New Release of the ANS Coder library and integration into CTF-Coders

New Features in librans:
* High throughput rANS entropy coding algorithms based on SSE 4.1or AVX2 instructions with fallback options in case of incompatible CPUs 
* Fast on-the-fly creation of dictionaries from source data
* Fast on-the-fly renorming of dictionaries based on dataset metrics
* Compression and serialization of dictionaries using Elias-Delta coding
* Fast Packing of badly compressible or  incompressible data. 
* Algorithms that decide on-the-fly if to entropy code or pack the source data based on dataset metrics

Integration into CTF-Coders:
*  Bump ANS release version from `0.1` to `1.0`
* `--ans-version` command-line option in CTF-Coders to allow switching between legacy algorithms `compat` or `0.1` and new algorithms `1.0`.
* Improved control flow and data consistency checks in `EncodedBlocks`


davidrohr:
Thx for this PR! One can see a lot of work went into it.

Could you fix: Codechecker error: /sw/SOURCES/O2/11215-slc8_x86-64/0/Utilities/rANS/benchmarks/bench_ransTPC.cxx:86:3: error: use '= default' to define a trivial default constructor [modernize-use-equals-default]

And some commit messages are still bogus, e.g. "wip" and "f types". Do you want to rebase this, or to squash?

The macCI errors seem also genuine.

MichaelLettrich:
thank you @davidrohr ! 
We still have a couple of pending steps till the codebase is in a shape to be reviewed. 
* The macOS errors relate to parallel stl not being implemented for Clang and macOS defining `size_t` differently to `uint64_t` or `uint32_t` which causes ambiguities. I will fix those  
* I will also fix the fullCI issues with arguments, that is genuine as well. 
* @davidrohr and I need to sit down and discuss how to adapt the  CMake to enable/disable hardware features at compile time 
* I will then compile with all warnings on, to check for warnings in my codebase and remove them.
* Then I rework the commit messages and squash parts of the commits to provide a healthy balance of readability and error tracing. 
* then, finally we can run the reviews, validate and merge.

davidrohr:
> * @davidrohr and I need to sit down and discuss how to adapt the  CMake to enable/disable hardware features at compile time

For the CMake: What exactly do we need to do there? Can't we extract the possible features from the compile flags? I think e.g. Vc does this.

Currently, we have 2 compile settings, for the EPNs with `-march=znver2`, and without march for the rest.
I think we can assume SSE4.1 globally, and enable that for all x86 code without explicit march. For AVX2 I would be a bit more conservative.

MichaelLettrich:
Yes, it will be something like this.  There are these options:
* non-x86: no hardware acceleration
* x86: `SSE4.1` or `AVX2` + `BMI2` instructions 

I will have a look how its done with VC and see how we can use the results of the checks that CMake does in the beginning of the process. Once I have that, I will ideally pass by or talk to you on zoom quickly to review if that's enough and if there is more that is needed. sounds good?

davidrohr:
Hi, that sounds reasonable to me, and ideally we would auto-detect all.
- no hardware acceleration if not x86 automatically.
- if we are on x86, check if avx2 is available with current compile flags, if yes use it.
- Now the question is whether we always fall back to sse4, or whether we also check if sse4 is available, and if not we also go without hardware acceleration also on x86.
If we have the check for instruction support anyway, I think we can also just check if sse4 is available.

One question is: Do you need to know this at CMake level, or is it enough to check during compilation? I think during compilation GCC will have some defines that expose whether we have SSE / AVX.
I just checked the Vc CMake, and it does a lot of parsing for compile flags I think. But anyhow, we have Vc as O2 dependency and it checks all these flags, so we could just use the Vc results to make our decision in CMake.

MichaelLettrich:
Dear @shahor02 , dear @davidrohr , all checks pass, the commit history is cleaned up I think we can proceed to review and in-depth compatibility testing.

MichaelLettrich:
@davidrohr  regarding the CMake: 
`librans` enables/disables features using compiler defines which it detects using pre-processor macros, set by the compiler in [include/rANS/internal/common/defines.h](https://raw.githubusercontent.com/MichaelLettrich/AliceO2/feature-rans-simd/Utilities/rANS/include/rANS/internal/common/defines.h). 
The best way  to do this in my opinion is to use the `-march=` command which will specify the target micro architecture to optimize for. 
I have set  the default to this to `native` which will optimize to the current system. The binaries produced this way are not portable, but we only need portability on the systems that build the releases, right? 
The architecture can be changed using [`-DRANS_TARGET_ARCH`](https://github.com/MichaelLettrich/AliceO2/blob/f237dccaefe76751bcfa1a5c8d22b420c41eac82/Utilities/rANS/CMakeLists.txt#L13-L22C1). This should allow us to use any of the micro-architectures defined  by [compilers](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html), eg. `-march=zenver2` for the EPN builds or use one of the generic `x86_64`  defines by the  x86-64 psABI documents for the grid: 

| architecture | features |
|-------------------|--------------|
| `x86-64`  |`CMOV`, `CX8`, `FPU`, `FXSR`, `MMX`, `OSFXSR`, `SCE`, `SSE`, `SSE2` |
|`x86-64-v2`|`CMPXCHG16B`,`LAHF`,`POPCNT`,`SSE3`,`SSE4_1`,`SSE4_2`,`SSSE3`|
|`x86-64-v3`|`AVX`, `AVX2`, `BMI1`, `BMI2`, `F16C`, `FMA`, `LZCNT`, `MOVBE`, `OSXSAVE`|
|`x86-64-v4`|`AVX512F`, `AVX512BW`, `AVX512CD`, `AVX512DQ`, `AVX512VL`|

davidrohr:
> 

Hi @MichaelLettrich , please do not change the -march.
This is set externally via the alidist defaults. And we need portability, since we do not build on the same servers where we run.
Besides, I think the approach to use preprocessor defines and rely on the -march is perfect. We just should not change the default -march.

MichaelLettrich:
@shahor02 and I recently completed a review of the code. 
During extended testing some previously unknown bugs were discovered that will have to be fixed.

shahor02:
@MichaelLettrich there was a conflict with the PR I've merged today, so I've pushed a version rebased to dev and with conflict fixed.

shahor02:
Hi @MichaelLettrich 

I've found the reason of the `o2::rans::internal::HistogramView...` crash: we have 2 kind of external dictionaries: 
1) with EncodedBlocks stored as a TTree, in which case the old format Metadata will be correctly parsed to new format (thanks to the root automatic schema evolution) when the dictionary is loaded from the CCDB
2) dictionary stored as vector<char>, which is just cast to EncodedBlocks, in this case the schema evolution does not work and the Metadata is messed up.
Currently only TPC and CTP are using type (1) dictionaries (paths like http://alice-ccdb.cern.ch/browse/TPC/Calib/CTFDictionaryTree), until I move other detectors to the same format we should test with TPC and CTP only.

I've extended the test script (attached) to accept extra options
```
./test.sh -c <input-ctf-list> [-m max-tf <N>] [--wnew] [--rnew]
```
With `--rnew` defined it assumes that the input data is in the new 1.0 format (otherwise, default compat mode). Similarly, with `--wnew` the output CTFs are requested to be stored in the new format.

With this I still see the following problems:

Running as 
```
WORKFLOW_DETECTORS="CTP" ./test.sh -c inp.lst -m 1
```
it crashes with:
```
[1241917:ctp-entropy-encoder]: o2-ctp-entropy-encoder-workflow: /home/shahoian/alice/sw/SOURCES/O2/dev/0/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h:269: void o2::ctf::Block<W>::storeLiterals(int, const W*) [with W = unsigned int]: Assertion `sz <= registry->getFreeSize()' failed.
```

The gdb shows 
```
#7  0x00007f854b7f8a4d in o2::ctf::Block<unsigned int>::storeLiterals (this=0x7f832631ed20, _nliterals=1, _literals=0x66a5b30)
    at /home/shahoian/alice/sw/SOURCES/O2/dev/0/DataFormats/Detectors/Common/include/DetectorsCommonDataFormats/EncodedBlocks.h:269
269         assert(sz <= registry->getFreeSize());
(gdb) p sz
$1 = 16
(gdb) p registry->getFreeSize()
$2 = 0
```
The problem seems to be in the modified expandStorage method: 
1) sometimes it gets an expansion target in words and sometimes in bytes.
2) one cannot use this-> data members in this method since `this` might not point to a valid object after previous expansion.

I've opened a PR on your branch, https://github.com/MichaelLettrich/AliceO2/pull/10, could you please check particularly the consistency of the expandStorage usage ? 

Once this is fixed, I can run `WORKFLOW_DETECTORS="CTP,TPC" ./test.sh -c inp.lst` w/o problem and it seems that the decoding of the original and of the rewritten (in the default `compat` mode) CTF produces consistent results.

Then, if I run `WORKFLOW_DETECTORS="CTP" ./test.sh -c inp.lst  --wnew`  I don't see any non-0 nDict in any of the block, so it looks lile the `--ans-version 1.0` was ignored:
```
ctf->Scan("CTP_block.0.nDict : CTP_block.0.nData : CTP_block.0.nLiterals");
************************************************
*    Row   * CTP_block *  CTP_bloc *  CTP_bloc *
************************************************
*        0 *         0 *         4 *         1 *
*        1 *         0 *         0 *         0 *
*        2 *         0 *         4 *         1 *
*        3 *         0 *         5 *         1 *
*        4 *         0 *         4 *         0 *
*        5 *         0 *         4 *         1 *
*        6 *         0 *         5 *         1 *
*        7 *         0 *         4 *         1 *
*        8 *         0 *         0 *         0 *
*        9 *         0 *         4 *         1 *
*       10 *         0 *         4 *         0 *
...
```
But the same exercise for TPC crashes: 
```
WORKFLOW_DETECTORS="TPC" ./test.sh -c inp.lst  --wnew
```
leads to:
```
[1333129:tpc-entropy-encoder]: [14:29:14][INFO] TPC CTF Columns Combining ON
[1333129:tpc-entropy-encoder]: [14:29:15][WARN] rescaling Frequency Table for empty message
[1333129:tpc-entropy-encoder]: [14:29:15][WARN] rescaling Frequency Table for empty message
[1333129:tpc-entropy-encoder]: [14:29:17][WARN] rescaling Frequency Table for empty message
[1333129:tpc-entropy-encoder]: [14:29:19][WARN] rescaling Frequency Table for empty message
[1333129:tpc-entropy-encoder]: [14:29:21][WARN] rescaling Frequency Table for empty message
[1333129:tpc-entropy-encoder]: [14:29:21][INFO] Loaded TPC CTF dictionary v1.0.1626472048 (16/07/21 21:47:28 UTC) from CCDB
[1333129:tpc-entropy-encoder]: [14:29:21][INFO] Caching in 139874955374560 ptr to TPC/CTFDICT/0 (0x417a6c0)
[1333129:tpc-entropy-encoder]: [14:29:21][INFO] Estimated output size is 2140877 bytes
[1333129:tpc-entropy-encoder]: [14:29:21][INFO] StoreData 2424 bytes, offs: 2304:4728
[1333129:tpc-entropy-encoder]: [14:29:21][INFO] StoreLiterals 4 bytes, offs: 4728:4732
[1333129:tpc-entropy-encoder]: [14:29:21][INFO] StoreData 476 bytes, offs: 4736:5212
[1333129:tpc-entropy-encoder]: [14:29:21][INFO] StoreData 160 bytes, offs: 5216:5376
[1333129:tpc-entropy-encoder]: o2-tpc-reco-workflow: /home/shahoian/alice/sw/SOURCES/O2/dev/0/Utilities/rANS/include/rANS/internal/encode/SingleStreamEncoderImpl.h:82: std::tuple<typename o2::rans::internal::EncoderImpl<symbol_T, o2::rans::internal::SingleStreamEncoderImplBase<streamingLowerBound_V, symbol_T, derived_T> >::state_type, stream_IT> o2::rans::internal::SingleStreamEncoderImplBase<streamingLowerBound_V, symbol_T, derived_T>::renorm(state_type, stream_IT, typename std::remove_pointer<typename std::remove_cv<typename o2::rans::internal::EncoderImpl<symbol_T, o2::rans::internal::SingleStreamEncoderImplBase<streamingLowerBound_V, symbol_T, derived_T> >::symbol_type>::type>::type::value_type) [with stream_IT = unsigned int*; long unsigned int streamingLowerBound_V = 20; symbol_T = const o2::rans::internal::PrecomputedSymbol*; derived_T = o2::rans::internal::SingleStreamEncoderImpl<20>; typename o2::rans::internal::EncoderImpl<symbol_T, o2::rans::internal::SingleStreamEncoderImplBase<streamingLowerBound_V, symbol_T, derived_T> >::state_type = long unsigned int; state_type = long unsigned int; typename std::remove_pointer<typename std::remove_cv<typename o2::rans::internal::EncoderImpl<symbol_T, o2::rans::internal::SingleStreamEncoderImplBase<streamingLowerBound_V, symbol_T, derived_T> >::symbol_type>::type>::type::value_type = unsigned int; std::remove_pointer_t<typename std::remove_cv<typename o2::rans::internal::EncoderImpl<symbol_T, o2::rans::internal::SingleStreamEncoderImplBase<streamingLowerBound_V, symbol_T, derived_T> >::symbol_type>::type> = const o2::rans::internal::PrecomputedSymbol; typename std::remove_cv<typename o2::rans::internal::EncoderImpl<symbol_T, o2::rans::internal::SingleStreamEncoderImplBase<streamingLowerBound_V, symbol_T, derived_T> >::symbol_type>::type = const o2::rans::internal::PrecomputedSymbol*; typename o2::rans::internal::EncoderImpl<symbol_T, o2::rans::internal::SingleStreamEncoderImplBase<streamingLowerBound_V, symbol_T, derived_T> >::symbol_type = const o2::rans::internal::PrecomputedSymbol*]: Assertion `state < maxState' failed.
[1333129:tpc-entropy-encoder]: *** Program crashed (Segmentation fault, FPE, BUS, ABRT, KILL, Unhandled Exception, ...)
[1333129:tpc-entropy-encoder]: Backtrace by DPL:
[1333129:tpc-entropy-encoder]: ABRT
[1333129:tpc-entropy-encoder]: Executable is /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/o2-tpc-reco-workflow
[1333129:tpc-entropy-encoder]:     /lib/x86_64-linux-gnu/libc.so.6:     ?? ??:0
[1333129:tpc-entropy-encoder]:     /lib/x86_64-linux-gnu/libc.so.6:     ?? ??:0
[1333129:tpc-entropy-encoder]:     /lib/x86_64-linux-gnu/libc.so.6:     ?? ??:0
[1333129:tpc-entropy-encoder]:     /lib/x86_64-linux-gnu/libc.so.6:     ?? ??:0
[1333129:tpc-entropy-encoder]:     /lib/x86_64-linux-gnu/libc.so.6:     ?? ??:0
[1333129:tpc-entropy-encoder]:     /lib/x86_64-linux-gnu/libc.so.6:     ?? ??:0
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: std::tuple<unsigned long, unsigned int*> o2::rans::internal::SingleStreamEncoderImplBase<20ul, o2::rans::internal::PrecomputedSymbol const*, o2::rans::internal::SingleStreamEncoderImpl<20ul> >::renorm<unsigned int*>(unsigned long, unsigned int*, unsigned int)
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: unsigned int* o2::rans::internal::SingleStreamEncoderImpl<20ul>::putSymbols<unsigned int*>(unsigned int*, o2::rans::internal::PrecomputedSymbol const* const&)
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: decltype(auto) o2::rans::Encoder<o2::rans::internal::SingleStreamEncoderImpl<20ul>, o2::rans::SymbolTable<unsigned short, o2::rans::internal::PrecomputedSymbol>, 2ul>::process<unsigned int*, unsigned short*, std::back_insert_iterator<std::vector<unsigned short, std::allocator<unsigned short> > >, true>(unsigned short*, unsigned short*, unsigned int*, std::back_insert_iterator<std::vector<unsigned short, std::allocator<unsigned short> > >) const
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: unsigned int* o2::ctf::ExternalEntropyCoder<unsigned short>::encode<unsigned short*, unsigned int*>(unsigned short*, unsigned short*, unsigned int*, unsigned int*)
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: o2::ctf::CTFIOSize o2::ctf::EncodedBlocks<o2::tpc::CTFHeader, 23, unsigned int>::encodeRANSV1External<unsigned short*, std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> > >(unsigned short*, unsigned short*, int, void const*, std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> >*, double)
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: o2::ctf::CTFIOSize o2::ctf::EncodedBlocks<o2::tpc::CTFHeader, 23, unsigned int>::entropyCodeRANSV1<unsigned short*, std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> > >(unsigned short*, unsigned short*, int, o2::ctf::Metadata::OptStore, std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> >*, void const*, float)
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: o2::ctf::CTFIOSize o2::ctf::EncodedBlocks<o2::tpc::CTFHeader, 23, unsigned int>::encode<unsigned short*, std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> > >(unsigned short*, unsigned short*, int, unsigned char, o2::ctf::Metadata::OptStore, std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> >*, void const*, float)
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: auto o2::tpc::CTFCoder::encode<std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> > >(std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> >&, o2::tpc::CompressedClusters const&, o2::tpc::CompressedClusters const&, std::vector<bool, std::allocator<bool> >*, std::vector<bool, std::allocator<bool> >*, std::vector<bool, std::allocator<bool> >*, std::vector<bool, std::allocator<bool> >*)::{lambda(auto:1, auto:2, o2::tpc::CTF::Slots, unsigned long, std::vector<bool, std::allocator<bool> >*)#1}::operator()<unsigned short*, unsigned short*>(unsigned short*, unsigned short*, o2::tpc::CTF::Slots, unsigned long, std::vector<bool, std::allocator<bool> >*) const
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: o2::ctf::CTFIOSize o2::tpc::CTFCoder::encode<std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> > >(std::vector<unsigned char, boost::container::pmr::polymorphic_allocator<unsigned char> >&, o2::tpc::CompressedClusters const&, o2::tpc::CompressedClusters const&, std::vector<bool, std::allocator<bool> >*, std::vector<bool, std::allocator<bool> >*, std::vector<bool, std::allocator<bool> >*, std::vector<bool, std::allocator<bool> >*)
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: o2::tpc::EntropyEncoderSpec::run(o2::framework::ProcessingContext&)
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: o2::framework::adaptFromTask<o2::tpc::EntropyEncoderSpec, bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&>(bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&)::{lambda(o2::framework::InitContext&)#1}::operator()(o2::framework::InitContext&) const::{lambda(o2::framework::ProcessingContext&)#4}::operator()(o2::framework::ProcessingContext&) const
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: void std::__invoke_impl<void, o2::framework::adaptFromTask<o2::tpc::EntropyEncoderSpec, bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&>(bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&)::{lambda(o2::framework::InitContext&)#1}::operator()(o2::framework::InitContext&) const::{lambda(o2::framework::ProcessingContext&)#4}&, o2::framework::ProcessingContext&>(std::__invoke_other, o2::framework::adaptFromTask<o2::tpc::EntropyEncoderSpec, bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&>(bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&)::{lambda(o2::framework::InitContext&)#1}::operator()(o2::framework::InitContext&) const::{lambda(o2::framework::ProcessingContext&)#4}&, o2::framework::ProcessingContext&)
[1333129:tpc-entropy-encoder]:     /home/shahoian/alice/sw/ubuntu2204_x86-64/O2/dev-local4/bin/../lib/libO2TPCWorkflow.so: std::enable_if<is_invocable_r_v<void, o2::framework::adaptFromTask<o2::tpc::EntropyEncoderSpec, bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&>(bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&)::{lambda(o2::framework::InitContext&)#1}::operator()(o2::framework::InitContext&) const::{lambda(o2::framework::ProcessingContext&)#4}&, o2::framework::ProcessingContext&>, void>::type std::__invoke_r<void, o2::framework::adaptFromTask<o2::tpc::EntropyEncoderSpec, bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&>(bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&)::{lambda(o2::framework::InitContext&)#1}::operator()(o2::framework::InitContext&) const::{lambda(o2::framework::ProcessingContext&)#4}&, o2::framework::ProcessingContext&>(o2::framework::adaptFromTask<o2::tpc::EntropyEncoderSpec, bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&>(bool&, bool&, std::shared_ptr<o2::base::GRPGeomRequest>&)::{lambda(o2::framework::InitContext&)#1}::operator()(o2::framework::InitContext&) const::{lambda(o2::framework::ProcessingContext&)#4}&, o2::framework::ProcessingContext&)
```

 Could you check please?

[test.sh.gz](https://github.com/AliceO2Group/AliceO2/files/11704915/test.sh.gz)

shahor02:
@MichaelLettrich I've imported TTree-based dictionaries and created http://alice-ccdb.cern.ch/browse/<DET>/Calib/CTFDictionaryTree for all detectors, now the problem with messy MetaData should be gone.

MichaelLettrich:
@shahor02  thank you so much for your contribution to this. I would have never found the Metadata bug. I will continue investigation.

MichaelLettrich:
@shahor02  Running the CTF tests with SIMD enabled, I found multiple bugs in the `CTFHelper` iterator classes, that only manifest in lookups with SIMD enabled. They look like they are easy to fix. Would be good if we could review them together once fixed.

shahor02:
Sure, let me know when.

MichaelLettrich:
@shahor02 Thanks for providing the tests and reporting your findings. I fixed the issues you encountered.

The phenomenon you see with no dicts stored inline is expected. Because you are supplying an external dictionary, we are still using an external coder, but with the new algorithm, so no dict is saved into the newly created ctfs.
The crash you observed was the consequence of incorrect use of external dictionaries with the `V1` encoder and has been fixed. In a nutshell:  
* `compat` encoding algorithms could use a dictionary that was renormed to to 31 Bits . That was very resource hungy and those cases usually would not compress well anyways and should be packed ideally. 
* `V1` encoding algorithms requires dicts to be renormed to max 20 Bits. Since the `V1` renorming algorithms are much more efficient internally, and everything that does not compress well should be packed anyways, the 20 Bits are sufficient. 
*  Fix: Using renorming precision calculated by `compat`  and feed it to  `V1`  might cause `V1` encoders internally to produce invalid rANS states. So I added a sanitizer step when reading external dictionaries to ensure `V1` encoders never go past their internal threshold. I also hardened encoders to not accept symbol tables that might produce errors.

I observed that when using librans with SIMD, the `CTFHelpers` iterators were causing issues. This was due to them not complying with the rules of the`std::random_access_iterator` [interface](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator). Fixing the issues resolved the crashes.

Now we should be good to continue testing.

shahor02:
@MichaelLettrich thanks!

I've pushed directly one fix (imposing to read TTree-based dictionaries only) which was missing in my previous PR.
I've also modified the `test.sh` so that with the `--wnew` option the `--ans-version 1.0` is added together with `--ctf-dict none`. In this way indeed per-TF dictionaries are stored. Meanwhile, I still see some blocks with `nDict=0, nData>0 and nLiterals=0`. Is this normal?

With updated branch still see the following problems:
1) building of ZDC decoder from the external dictionary fails:
```
o2-ctf-reader-workflow --timeframes-rate-limit 2 --timeframes-rate-limit-ipcid 0 --ctf-data-subspec 0 --ctf-input inp.lst --allow-missing-detectors -b --onlyDet ZDC
...
[2929826:zdc-entropy-decoder]: [16:32:11][ERROR] Exception caught: Incompatible Frequency table dimensions: Cannot add [65526,65543] to [0, 65535]  
```

2) encoding with `--ans-version 1.0` fails for MID at the 2nd TF (some unhandled exception or memory overrun):
```
WORKFLOW_DETECTORS=MID ./test.sh -c inp.lst  --wnew
...
[2956156:mid-entropy-encoder]: [17:06:00][INFO] MID_CTF: Container of 7 blocks, size: 402041 bytes, unused: 376393
[2956156:mid-entropy-encoder]: [17:06:00][INFO] Block 0 for 13660 message words of 2 bytes | NDictWords: 72 NDataWords: 978 NLiteralWords: 0
[2956156:mid-entropy-encoder]: [17:06:00][INFO] Block 1 for 13660 message words of 4 bytes | NDictWords: 1 NDataWords: 35 NLiteralWords: 0
[2956156:mid-entropy-encoder]: [17:06:00][INFO] Block 2 for 13660 message words of 2 bytes | NDictWords: 1 NDataWords: 61 NLiteralWords: 1
[2956156:mid-entropy-encoder]: [17:06:00][INFO] Block 3 for 13660 message words of 1 bytes | NDictWords: 0 NDataWords: 1 NLiteralWords: 0
[2956156:mid-entropy-encoder]: [17:06:00][INFO] Block 4 for 69125 message words of 2 bytes | NDictWords: 190 NDataWords: 4270 NLiteralWords: 0
[2956156:mid-entropy-encoder]: [17:06:00][INFO] Block 5 for 13825 message words of 1 bytes | NDictWords: 16 NDataWords: 305 NLiteralWords: 0
[2956156:mid-entropy-encoder]: [17:06:00][INFO] Block 6 for 13825 message words of 1 bytes | NDictWords: 3 NDataWords: 284 NLiteralWords: 0
[2956156:mid-entropy-encoder]: [17:06:00][INFO] raw input:603020/input to encode:288840/ctf size:124340 bytes in 0 s
[2956156:mid-entropy-encoder]: [17:06:00][INFO] StoreDict 304 bytes, offs: 736:1040
[2956156:mid-entropy-encoder]: [17:06:00][INFO] StoreData 3884 bytes, offs: 1040:4924
[2956157:ctf-writer]: [17:06:00][INFO] MID: Container of 7 blocks, size: 25648 bytes, unused: 0
[2956157:ctf-writer]: [17:06:00][INFO] Block 0 for 13660 message words of 2 bytes | NDictWords: 72 NDataWords: 978 NLiteralWords: 0
[2956157:ctf-writer]: [17:06:00][INFO] Block 1 for 13660 message words of 4 bytes | NDictWords: 1 NDataWords: 35 NLiteralWords: 0
[2956157:ctf-writer]: [17:06:00][INFO] Block 2 for 13660 message words of 2 bytes | NDictWords: 1 NDataWords: 61 NLiteralWords: 1
[2956157:ctf-writer]: [17:06:00][INFO] Block 3 for 13660 message words of 1 bytes | NDictWords: 0 NDataWords: 1 NLiteralWords: 0
[2956157:ctf-writer]: [17:06:00][INFO] Block 4 for 69125 message words of 2 bytes | NDictWords: 190 NDataWords: 4270 NLiteralWords: 0
[2956157:ctf-writer]: [17:06:00][INFO] Block 5 for 13825 message words of 1 bytes | NDictWords: 16 NDataWords: 305 NLiteralWords: 0
[2956157:ctf-writer]: [17:06:00][INFO] Block 6 for 13825 message words of 1 bytes | NDictWords: 3 NDataWords: 284 NLiteralWords: 0
[2956157:ctf-writer]: [17:06:00][INFO] TF#0: wrote CTF{Run:0534152 TF:125 Orbit:39620544 CteationTime:1680795156412 Detectors: MID} of size 25449 to /home/shahoian/Downloads/rANS/o2_ctf_run00534152_orbit0039620544_tf0000000125_alicers05.root in 0.06 s
...
[2956155:mid-entropy-decoder]: [17:06:00][INFO] Decoded {13462,0,0} MID columns for {13236,0,0} ROFRecords, (raw input:585096/input to encode:280668/ctf size:75292 bytes) in 0 s
[ERROR] Workflow crashed - pid 2956156 (mid-entropy-encoder) was killed abnormally with exit code 128, could be out of memory killer, segfault, unhandled exception, SIGKILL, etc...
```

After excluding MID and ZDC I was able to decode old CTF and rewrite it in the new format, then to read it back:
```
WORKFLOW_DETECTORS=FT0,FV0,FDD,ITS,MFT,TPC,TRD,TOF,EMC,PHS,CTP,CPV,HMP ./test.sh -c inp.lst  --wnew
mv o2_ctf_run00534152_orbit0039620544_tf0000000125_alicers05.root o2_ctf_run00534152_orbit0039620544_tf0000000125_alicers05_new.root
WORKFLOW_DETECTORS=FT0,FV0,FDD,ITS,MFT,TPC,TRD,TOF,EMC,PHS,CTP,CPV,HMP ./test.sh -c o2_ctf_run00534152_orbit0039620544_tf0000000125_alicers05_new.root  --wnew
```

Note that to see the workflow generated by `test.sh` you can run it with
```
WORKFLOWMODE=print  ./test.sh ....

```
Updated script is attached: [test.sh.gz](https://github.com/AliceO2Group/AliceO2/files/11714987/test.sh.gz)

MichaelLettrich:
@shahor02  Thank you for updates and for running the checks

 `nDict=0`, `nData>0` and `nLiterals=0` should be the case if data is packed (` OptStore=PACK` ) and not entropy coded because the source data  does not compress well enough. You should be able to see it in the metadata. If you want to ensure all data is entropy coded, you can change the settings in the per-detector `CTFCoder.h` implementations from `EENCODE_OR_PACK` to `EENCODE` or `PACK` only. The rest is definitely bugs and issues. I will investigate.

shahor02:
Hi @MichaelLettrich 
updated with rebased to dev/ conflicts resolved.

shahor02:
Hi @MichaelLettrich 
I've opened a PR to your branch with the changes we discussed. I've done a fast benchmark, for the size the improvement is spectacular, I see factor 2 (but did not yet check the correctness of decoded data for all detectors).
In opposite, for the speed the results are not uniform but mostly worse for the embedded dictionaries.
Especially, for the TPC: factor 57 and 22 for encoding and decoding resp.

timing below is in ns for one of the TFs in the inp.list I sent you (to get the timing printed, run with `export DPL_REPORT_PROCESSING=1` then grep the log file e.g. as `grep wall t1to0.log | grep entropy`.


  | old decoder | new decoder | old encoder | new encoder
-- | -- | -- | -- | --
cpv | 101282 | 117818 | 588404 | 324057
ctp | 74915 | 117434 | 224892 | 146424
emcal | 580894 | 193580 | 741176 | 714720
fdd | 378294 | 77763 | 83517 | 107206
ft0 | 238068 | 352074 | 366206 | 210900
fv0 | 166684 | 102736 | 536814 | 229767
hmpid | 121062 | 100743 | 299433 | 151976
its | 14457667 | 4049845 | 1470993 | 3060042
mch | 6084870 | 16079027 | 4018045 | 17820744
mft | 352746 | 220859 | 822125 | 1013728
mid | 2596880 | 2102485 | 2418398 | 3991128
phos | 351583 | 243156 | 760841 | 683913
tof | 2536694 | 2292998 | 2942621 | 1893644
tpc | 14825776 | 340601269 | 5786255 | 330307148
trd | 259851 | 198233 | 557118 | 7455034
zdc | 54529 | 106777 | 307537 | 114932

MichaelLettrich:
@shahor02 thank you for your PR. I will review it carefully and merge it. Also thanks for measurements - but it seems that there is a typo on the labels though as "new decoder" occurs twice. 
As for the compression improvements - as you can imagine it highly depends on the data, but indeed that sounds very good. Looks like the data deviates a lot from the dictionaries. 
As for the performance:  Did you build with assertions and optimization enabled? SIMD Code suffers disproportionately from enabled assert statements as data is copied around and inlining of functions might be prevented.
Once things work work correctly and robustly, I will do a propper profiling and have an in-depth look into performance.

shahor02:
Hi @MichaelLettrich 

I've fixed the label (the last column is "new encoder".
I've built your branch as is, just checked that no `-O0` was used.

MichaelLettrich:
@shahor02 thank you very much. `-O0` obviously matters a lot, but we build `EncodedBlocks.h` with assertions (`#undef NDEBUG`). You will see a significant speedup if assertions are disabled. There is other factors that contribute as well of course.  I will have to benchmark in detail using a profiler to find out what introduces the performance degradation. In my measurements, I have been significantly faster with SIMD encoding + in-place dicts than Non-SIMD with existing dicts.

shahor02:
OK, do you have something essential to add to this PR or I can start more thorough test (once you merge my PR)?

MichaelLettrich:
@shahor02 I will review tonight and hopefully merge, then I still have to cleanup the commit history with all those patches we added and I have one small change I want to add. I should be able to finish this over the weekend.

MichaelLettrich:
@davidrohr: 
As discussed with you a while ago I adjusted the rANS CMake scripts such that 
* by default we build everything with the global settings 
* If global settings enable `SSE4.1`, `AVX2` or `BMI2`, the compiler will set a couple of defines for which librans checks. If they are found, we selectively enable settings, otherwise we don't. 
* The unit tests in librans are built with `march=native` to run as many tests as possible, since these are not bundled up or cross-compiled if I understand correctly
* Users may use `-DRANS_ARCH=<arch>` to manually add a target architecture to the library `target_compile_options`.

Does that work for you?

davidrohr:
that seems reasonable, and without checking everything in detail, your cmake changes look good

shahor02:
Concerning `stage/tests/o2-test-tpc-CalDet` failing in the previous version (built with `-DRANS_ARCH=native`): it looks like some optimizer bug: it fails in https://github.com/MichaelLettrich/AliceO2/blob/8b07be76f9af4cf666dad0309353df41510a3eb4/Detectors/TPC/base/test/testTPCCalDet.cxx#L34-L37

If I change it to 
```
template <typename T>
bool isEqualAbs(T val1, T val2)
{
  //  return std::abs(val1 - val2) < std::numeric_limits<T>::epsilon();
  auto res = std::abs(val1 - val2) < std::numeric_limits<T>::epsilon();
  if (!res) {
    LOGP(info, "res = {} | {} - {}", res, val1, val2);
  }
  return res;
}
```
then no it always returns true and the test is passed. If I comment put LOGP, it fails. Even it fails, when after `isEqualAbs` returning false I print `std::abs(val1 - val2)` I get exactly 0.  Also, changing binary 
`isEqual &=  isEqualAbs(...)` to boolean `isEqual = isEqual &&  isEqualAbs(...)` does not help, i.e. with the patch like this:

```
diff --git a/Detectors/TPC/base/test/testTPCCalDet.cxx b/Detectors/TPC/base/test/testTPCCalDet.cxx
index 3fcfae6bc3..ae37c90d90 100644
--- a/Detectors/TPC/base/test/testTPCCalDet.cxx
+++ b/Detectors/TPC/base/test/testTPCCalDet.cxx
@@ -233,9 +233,13 @@ BOOST_AUTO_TEST_CASE(CalDet_Arithmetics)
   padCmp = pad;
   padCmp *= number;
 
+  LOGP(info, "Start : isEqual {} | eps {}", isEqual, std::numeric_limits<float>::epsilon());
   for (auto const& arrays : boost::combine(padCmp.getData(), pad.getData())) {
     for (auto const& val : boost::combine(arrays.get<0>().getData(), arrays.get<1>().getData())) {
-      isEqual &= isEqualAbs(val.get<0>(), val.get<1>() * number);
+      isEqual = isEqual && isEqualAbs(val.get<0>(), val.get<1>() * number);
+      if (!isEqualAbs(val.get<0>(), val.get<1>() * number)) {
+       LOGP(error, "{} {} {} diff {} | -> {} / {}", val.get<0>(), val.get<1>() * number, number, val.get<0>() - val.get<1>() * number , isEqual, isEqualAbs(val.get<0>(), val.get<1>() * number));
+      }
     }
   }

```
I get:
```
[INFO] Start : isEqual true | eps 1.1920929e-07
[ERROR] 4.2000003 4.2000003 0.2 diff 0 | -> false / true
[ERROR] 4.6 4.6 0.2 diff 0 | -> false / true
[ERROR] 4.8 4.8 0.2 diff 0 | -> false / true
[ERROR] 5.2000003 5.2000003 0.2 diff 0 | -> false / true
[ERROR] 5.6 5.6 0.2 diff 0 | -> false / true
[ERROR] 6.2000003 6.2000003 0.2 diff 0 | -> false / true
```

@MichaelLettrich @davidrohr, any idea?

davidrohr:
> * The unit tests in librans are built with `march=native` to run as many tests as possible, since these are not bundled up or cross-compiled if I understand correctly

On second thought, even that will break the CI. We have different build machines, which might do incremental builds. But after the build, all will run the full ctests. I.e., I believe it could be that some system runs the ctests compiled by another one.
@TimoWilken : could you comment if this is possible, or whether I am mistaken.

shahor02:
@MichaelLettrich the MacOS CI error seems to be genuine, could you guard https://github.com/MichaelLettrich/AliceO2/blob/8b07be76f9af4cf666dad0309353df41510a3eb4/Utilities/rANS/include/rANS/internal/containers/DenseHistogram.h#L530 by 
`#ifdef RANS_OPENMP` / `#endif` ?

MichaelLettrich:
@shahor02 thank you very much for digging. 

* The error  related to `stage/tests/o2-test-tpc-CalDet` looks like a weird error related to optimization. if i surround the file with

 ```cpp
 #pragma GCC push_options
#pragma GCC optimize("O0")

//your code

#pragma GCC pop_options
 ```
 then the error disappears.
 @davidrohr @shahor02  Would it be legitimate to propose this as a fix?
 
*  for the macos-arm error it looks like libc++ implements reverse iterators differently than libstdc++. It's actually great the the compiler is able to catch it, because that one would be impossible to find. I will fix it. 
 
*  The `OMP unroll partial (n)` statement is relatively new (OMP 5.1). Not sure if its implemented outside of GCC. Replacing it by `#pragma GCC unroll n` should fix it. this is both understood by gcc and clang.

shahor02:
@MichaelLettrich for the `stage/tests/o2-test-tpc-CalDet`, let @davidrohr decides, I don't feel myself qualified to judge what is the best solution.

For the openMP: we disable it on macOS (alidist/o2.sh has `"openmp:(?!osx.*)"`), so it should be guarded by ifdefs.

davidrohr:
I don't understand why would `o2-test-tpc-CalDet` fai? Your PR doesn't modify it, does it?

shahor02:
@davidrohr see above, this happens when native optimization is on

davidrohr:
> @davidrohr see above, this happens when native optimization is on

but native optimization should be on only for the ctf test, right? So why does it affect tpc cal det test?

Althoug, still, one should understrand what goes wrong with the tpc cal det test with native optimization.
I can have a look these days...

MichaelLettrich:
@davidrohr I am not sure either why this happens. I will rebuild o2 with the default setup today on my machine to see if the flags get propagated somehow.

MichaelLettrich:
actually, it looks like the build cache was not cleared out by the pr builders. so now with `-march=native` only active for the unit tests that need it, the error disappears.

MichaelLettrich:
@ktf It looks like there is a collision between `o2::framework` and `absl` on macos. Is this something you are aware of? and what are the implications for the use of absl as part of o2? 
`THREAD_ANNOTATION_ATTRIBUTE__' macro redefined [-Werror,-Wmacro-redefined]`

davidrohr:
> @ktf It looks like there is a collision between `o2::framework` and `absl` on macos. Is this something you are aware of? and what are the implications for the use of absl as part of o2? `THREAD_ANNOTATION_ATTRIBUTE__' macro redefined [-Werror,-Wmacro-redefined]`

Just as an idea, if we cannot easily solve this, you said changing the hash backend is a one-liner. So worst case we could use std::unordered_map on macos, and abseil otherwise?

MichaelLettrich:
Yes, that would work. just verified it.

davidrohr:
> Yes, that would work. just verified it.

ok, for me that would be a viable temporary workaround, if @ktf  doesn't have a better idea soon.

ktf:
It should be easy to fix. that's boilerplate for some thread checking stuff.

ktf:
https://github.com/AliceO2Group/AliceO2/pull/11832 should make it compile also on macos.

MichaelLettrich:
thank you very much !

shahor02:
Results of [benchmarking](https://docs.google.com/spreadsheets/d/1lCGKOQVEZgTVNMnuUTOU3rwUc9U7fEtB8kfMguBBSvU/edit#gid=661088214) with 500 kHz data (LHC23zc_537963)

image

shahor02:
@MichaelLettrich now this
```
                 from /sw/SOURCES/O2/11215-slc7_x86-64/0/Framework/Core/include/Framework/InitContext.h:14,
                 from /sw/SOURCES/O2/11215-slc7_x86-64/0/Detectors/Base/include/DetectorsBase/CTFCoderBase.h:34,
                 from /sw/SOURCES/O2/11215-slc7_x86-64/0/Detectors/Base/src/CTFCoderBase.cxx:16:
/sw/SOURCES/O2/11215-slc7_x86-64/0/Framework/Core/include/Framework/ServiceRegistry.h:46:19: error: expected unqualified-id before string constant
   46 | struct CAPABILITY("mutex") MutexLock {
```
appeared in the CIs, do you understand the reason?

What about 
```
/Volumes/build/build/alice-ci-workdir/o2/sw/SOURCES/O2/11215/0/Framework/Foundation/include/Framework/ThreadSafetyAnalysis.h:24:9: error: 'THREAD_ANNOTATION_ATTRIBUTE__' macro redefined [-Werror,-Wmacro-redefined]
#define THREAD_ANNOTATION_ATTRIBUTE__(x) // no-op
```
on macOS?

MichaelLettrich:
@ktf can this be related to 

> @MichaelLettrich now this
> 
> ```
>                  from /sw/SOURCES/O2/11215-slc7_x86-64/0/Framework/Core/include/Framework/InitContext.h:14,
>                  from /sw/SOURCES/O2/11215-slc7_x86-64/0/Detectors/Base/include/DetectorsBase/CTFCoderBase.h:34,
>                  from /sw/SOURCES/O2/11215-slc7_x86-64/0/Detectors/Base/src/CTFCoderBase.cxx:16:
> /sw/SOURCES/O2/11215-slc7_x86-64/0/Framework/Core/include/Framework/ServiceRegistry.h:46:19: error: expected unqualified-id before string constant
>    46 | struct CAPABILITY("mutex") MutexLock {
> ```
> 
> appeared in the CIs, do you understand the reason?
> 
> What about
> 
> ```
> /Volumes/build/build/alice-ci-workdir/o2/sw/SOURCES/O2/11215/0/Framework/Foundation/include/Framework/ThreadSafetyAnalysis.h:24:9: error: 'THREAD_ANNOTATION_ATTRIBUTE__' macro redefined [-Werror,-Wmacro-redefined]
> #define THREAD_ANNOTATION_ATTRIBUTE__(x) // no-op
> ```
> 
> on macOS?

@ktf the macOS part should be fixed by #11832 and I think the macOS CI was not re-run since then. Could you clear the macOS builders of the PR? 
Also the new mutex issues seen on linux appear to be related to thread sanitizers as well if I see that correctly ?

MichaelLettrich:
@ktf thanks for restarting the macOS builds. It looks like now both macOS and linux report the same issues.

MichaelLettrich:
squashing completed, code is ready to be merged from my side.

MichaelLettrich:
@shahor02 any ideas why on macos we have these tests failing? looks like mostly root macros to me?

shahor02:
@MichaelLettrich macros test on macOS is failing in all PRs since ~2 weeks, you can ignore it.

MichaelLettrich:
@ktf  @shahor02  @davidrohr, looks like everything that can pass at the moment passes. from my side this PR can (finally) be merged.

shahor02:
Good, thanks!

PR #11217, opened 2023-04-23
[FOCAL-13] Bad channel map for FOCAL-pads

mfasDa:
Bad channels can be in any of the channels in any
of the layers.


mfasDa:
CI errors unrelated.

PR #11218, opened 2023-04-23
Reduce states overhead

ktf:


ktf:
Further reduce the overhead of capturing device state if the state itself is not used (e.g. the GUI is not needed).

PR #11221, opened 2023-04-24
Revert "[O2-3582] DPL Analysis: update running workflow configuration"

aalkin:
Reverts AliceO2Group/AliceO2#11105 as it apparently breaks array configurable.


ddobrigk:
@ktf @pzhristov @shahor02 @jgrosseo it's important that this be merged ASAP to recover functionality in time for the tutorial - then we study and understand the problem later. Thanks, @aalkin , for the work and reversal!

PR #11223, opened 2023-04-24
Fix wrong uniform_int_distribution usage

ktf:


ktf:
@chiarazampolli @shahor02 any objections to this?

chiarazampolli:
Fine with me, @TimoWilken also had tried, and in his PR there was a suggestion by @shahor02 : https://github.com/AliceO2Group/AliceO2/pull/11119#issuecomment-1497540895

ktf:
I think the comment of @shahor02 is actually accounted in my PR. I do not change the type of the class, which I agree would be dangerous, but I simply eliminate the dependency of the distribution on the actual storage type. If you need chars between (e.g.) 10 and 20, you do not care if the distribution is actually done using long long, and then downcasted or with the actual type. As long as the boundaries are compatible with the type T, you are fine.

ktf:
Do we have any way to validate this change?

ktf:
@chiarazampolli @shahor02 shall I merge this and then we think about possible issues?

shahor02:
Fine for me.

PR #11224, opened 2023-04-24
[O2-1787] First (rough) implementation of ITS services

mario6829:
A first rough implementation of ITS services (cables, wires and the like) was introduced in the geometry. In this version the mean material budget is taken into account but not the details of each single element. (The original code was developed by A.Morsch)

### DIFF
--- a/Detectors/ITSMFT/ITS/simulation/include/ITSSimulation/Detector.h
+++ b/Detectors/ITSMFT/ITS/simulation/include/ITSSimulation/Detector.h
@@ -286,6 +286,10 @@ class Detector : public o2::base::DetImpl<Detector>
   /// \param motherVolume the TGeoVolume owing the volume supports
   void createOuterBarrelSupports(TGeoVolume* motherVolume);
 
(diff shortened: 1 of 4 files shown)


mario6829:
Dear @mconcas @iouribelikov here it is a very first implementation of ITS services, based on @amorsch code. Please let me know in case of problems.

mconcas:
Hi @mario6829 , thanks for this PR!
Eventually this is going to supersede https://github.com/AliceO2Group/AliceO2/pull/11172, once I'll figure out the errors in CI.

EDIT: there was just a hiccup in the Mixture and Medium definition for the polymer.
Now should be working, if you want to check the change.

Thanks a lot!

mario6829:
many thanks @mconcas that's perfect!

mconcas (approved):

sawenzel (approved):

sawenzel:
pinging also @amorsch for comments

mconcas:
As the update came on JIRA https://alice.its.cern.ch/jira/browse/O2-1787, this can be merged.

PR #11225, opened 2023-04-25
tof bug fix in filling diagnostic words

noferini:
This is to fix a bug due to a wrong condition preventing to fill TRM errors when loading diagnostic words.
After the fix diagnostic words are loaded properly when reading ctf (tested locally on one ctf)


noferini:
Can we merge this PR (it would be nice to have it in before of next week since it impacts on sync calib)

martenole:
Hi @noferini would this change be needed for the O2 running on the FLPs or are the processes affected by your changes only running on the EPNs?

noferini:
This change affects the calbiration so it runs only on the calib node (epn partition)

noferini:
actually it runs in the epns (TOF decoder task).
In any case, flp are not touched

martenole:
Thanks, then I'll see that we get this in next week (in case we have a sw update slot)

PR #11227, opened 2023-04-25
[ITS3] Update default parameters and adapt code to work with fourth layer

fgrosa:
This PR adapt the ITS3 code to work with an additional layer. @mconcas @mpuccio I had to make two minor modifications to the `IndexTableUtils.h`, `TimeFrame.h`, and `TrackerTraits.h` files to make it work. They should be harmless, but please have a look that they are fine for you. Thanks a lot!


fgrosa:
> This PR adapt the ITS3 code to work with an additional layer. @mconcas @mpuccio I had to make two minor modifications to the `IndexTableUtils.h`, `TimeFrame.h`, and `TrackerTraits.h` files to make it work. They should be harmless, but please have a look that they are fine for you. Thanks a lot!

Moved the ITS-related code in https://github.com/AliceO2Group/AliceO2/pull/11234

mconcas (approved):

PR #11228, opened 2023-04-25
FT0Calibration: changing TFs per slot with respect to 32 HBFs

afurs:


chiarazampolli (approved):

chiarazampolli:
Hello @afurs , 

Do you need to change here too?
https://github.com/AliceO2Group/O2DPG/blob/1c11e45bfbd78de4392a4ca7f53e5187d8cd6828/DATA/production/calib/ft0-timeoffset-aggregator.sh#L14
Chiara

afurs:
Hello @chiarazampolli yes, please. Or I can change it later.

chiarazampolli:
Please, change it in O2DPG. 
Consider also changing it in a more universal way, in number of orbits or seconds, to not depend on the TF length.

afurs:
Okay, thanks!

afurs:
Hello @chiarazampolli I changed it in standalone, or do you mean var preparation which will be used here? if it is possible, could you please merge it asap, because right now we have 5/4 min slot length which is too short for our calibration :)

chiarazampolli:
Hello @afurs ,

i meant to change it such that it is in number of orbits or seconds, so that it does not depend on the TF length in orbits (if it gets changed), see mail somr time ago to wp13.

pinging @shahor02 , @davidrohr to merge, i cannot.

chiara

afurs:
@chiarazampolli ok, I see

martenole:
I can merge as it is now, but it would indeed be good if we could get rid of hard-coded numbers that depend on the TF length. See README here: https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/Calibration#configuration-of-the-timeslot
And for example here it is used https://github.com/AliceO2Group/AliceO2/blob/4c316f8658ffa0396657fa5602ab53dbf2ba6fc9/Detectors/TRD/workflow/include/TRDWorkflow/VdAndExBCalibSpec.h#L50

PR #11229, opened 2023-04-25
DPL: fix expendable tasks for the QC case

ktf:
This will make sure that if there is a data dependency to an expendable task, like in the case of QC, we do not change the order in which tasks are sorted. This means that it will be responsibility of the QC check to make sure it can survive missing data.


ktf:
Merging this. Has no consequences when the label is not there due to the first check.

PR #11230, opened 2023-04-25
[WIP] Changing granularity in TFs for TOF, considering new TF length

chiarazampolli:
@noferini : Ideally we should change it further, to use a duration in seconds or orbits.


chiarazampolli:
Moving to WIP, I prefer to change it in a more universal way.

PR #11233, opened 2023-04-25
EMCAL-915

mfasDa:
- STU error base class with main error codes
- Integration of STU decoding errors into
  ErrorTypeFEE


mfasDa:
Error handling for trigger reconstruction.

mfasDa:
Errors unrelated

PR #11238, opened 2023-04-26
[EMCAL-834] Checking active links per trigger

mpoghos:


Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -167,6 +181,12 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
       auto feeID = raw::RDHUtils::getFEEID(header);
       auto triggerbits = raw::RDHUtils::getTriggerType(header);
 
+      ULong64_t BCcoll = ULong64_t(triggerBC) + ULong64_t(triggerOrbit * 3564);
```
mfasDa: Please use instead
``` c++
 ULong64_t BCcoll  = currentIR.toLong();
```
instead.

Detectors/EMCAL/workflow/src/RawToCellConverterSpec.cxx
```diff
@@ -338,6 +358,21 @@ void RawToCellConverterSpec::run(framework::ProcessingContext& ctx)
     mOutputTriggerRecords.emplace_back(interaction, currentevent.getTriggerBits(), eventstart, ncellsEvent + nLEDMONsEvent);
   }
 
+  std::unordered_map<ULong64_t, std::bitset<46>>::iterator p;
+  for (p = bcFreq.begin(); p != bcFreq.end(); p++) {
+    if (p->second != bitSetActiveLinks) {
```
mfasDa: Could be handled with range-based iteration + c++17 structured-binding:
``` c++
for(const auto &[globalBC, activelinks) : bcFreq) {
  if (activelinks != bitSetActiveLinks){
    ...
  }
}
```

mfasDa (changes requested):
Apart from the global BC using the function from o2::InteractionRecords the rest looks good to me.

mfasDa (approved):
Looks good!

mfasDa:
@mpoghos In order to pass the fullCI we need a default CCDB object for EMC/Calib/FeeDCs in the timestamp range of the unanchored MCs.

mfasDa:
Needs https://alice.its.cern.ch/jira/browse/O2-3727

PR #11239, opened 2023-04-26
[EMCAL-840] Add TimeCalibSlewingParasm to LinkDef

jokonig:
- In recent commit https://github.com/AliceO2Group/AliceO2/pull/11184 the time calibration slewing container was introduced. However it was not added to the linkdef which is needed to create an instance of the object.


mfasDa:
Errors unrelated

mfasDa (approved):

PR #11241, opened 2023-04-26
CPV: extend format of CPV raw data trailer

sevdokim:
Hello! CPV raw data trailer format was changed by 1 bit: word counter in trailer now takes 10 bits instead of 9 bits previously. It is needed for nopadding data (dataformat =0x2) where word count per DMA page can be larger than 508 CPV words.


peressounko (approved):

shahor02 (commented):
Is this change backward compatible?

sevdokim:
Hi @shahor02 
Yes the change is backward compatible because previously the newly used bit was reserved to be 0. As dataformat 0x0 ensures wordcount to be less than 508, it will be 0 with new encoding as well.

PR #11245, opened 2023-04-27
[EMCAL-915, FOCAL-12, FOCAL-13] Make exception classes final

mfasDa:
Silencing Wfinal-dtor-non-final-class from clang


mfasDa:
Only silience compiler warnings.

PR #11246, opened 2023-04-27
[EMCAL-562] EMCAL/DCAL STU Decoder

mpoghos:


Detectors/EMCAL/reconstruction/include/EMCALReconstruction/StuDecoder.h
```diff
@@ -0,0 +1,153 @@
...
+  int32_t mFrameReceived = 0;        ///<
+  int32_t mCFGword16 = 0;            ///<
+
+  STUtype_t mSTU = STUtype_t::ESTU;
+  bool mIsFullPayload = true; ///<
+
+  void init();
+  void decodeL1JetPatchIndices(const uint32_t* buffer);
+  void decodeL1GammaPatchIndices(const uint32_t* buffer);
+  void decodeFastOrADC(const uint32_t* buffer);
```
mfasDa, line 124: Buffer should be passed as `const gsl::span<const uint32_t>`

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/StuDecoder.h
```diff
@@ -0,0 +1,153 @@
...
+
+  STUtype_t getSTUtype() const { return mSTU; }
+  int getFeeID() const { return STUparam::FeeID[mSTU]; }
+  int getNumberOfTRUs() const { return STUparam::NTRU[mSTU]; }
+
+  std::vector<int16_t> getL1JetHighPatchIndices() const { return mL1JetHighPatchIndex; }
+  std::vector<int16_t> getL1JetLowPatchIndices() const { return mL1JetLowPatchIndex; }
+  std::vector<int16_t> getL1GammaHighPatchIndices() const { return mL1GammaHighPatchIndex; }
+  std::vector<int16_t> getL1GammaLowPatchIndics() const { return mL1GammaLowPatchIndex; }
+  std::vector<int16_t> getFastOrADCs() const { return mFastOrADC; }
```
mfasDa, line 64: Return type should be `gsl::span<const int16>`. Do we really expect the indices to be signed?

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/StuDecoder.h
```diff
@@ -0,0 +1,153 @@
...
+  int getNumberOfL1JetLowPatches() const { return mL1JetLowPatchIndex.size(); }
+  int getNumberOfL1GammaHighPatches() const { return mL1GammaHighPatchIndex.size(); }
+  int getNumberOfL1GammaLowPatches() const { return mL1GammaLowPatchIndex.size(); }
+
+  int16_t getIndexOfL1JetHighPatch(int id) const { return mL1JetHighPatchIndex[id]; }
+  int16_t getIndexOfL1JetLowPatch(int id) const { return mL1JetLowPatchIndex[id]; }
+  int16_t getIndexOfL1GammaHighPatch(int id) const { return mL1GammaHighPatchIndex[id]; }
+  int16_t getIndexOfL1GammaLowPatch(int id) const { return mL1GammaLowPatchIndex[id]; }
+
+  int16_t getFastOrADC(int iTRU, int iCh) const { return mFastOrADC[iTRU + getNumberOfTRUs() * iCh]; }
```
mfasDa, line 76: Use std::size_t for index in order to prevent access with negative indices. What happens if index >  size?

Detectors/EMCAL/reconstruction/src/StuDecoder.cxx
```diff
@@ -0,0 +1,241 @@
...
+    for (int i = 0; i < offset; i++) {
+      std::cout << std::dec << i << ")  0x" << std::hex << buffer[offset + i] << "   <- Gamma-low-word\n";
+    }
+  }
+
+  const int nTRU = getNumberOfTRUs();
+  const int nTRU2 = getNumberOfTRUs() / 2;
+
+  unsigned short gammaHigh[nTRU][TRUparam::NchannelsOverPhi];
+  unsigned short gammaLow[nTRU][TRUparam::NchannelsOverPhi];
```
mfasDa, line 169: Since the functions return non-const the const here is effectively a lie. The array creation results in undefined behaviour. A std::vector must be used instead. In any case for arrays please use std::array instead of c-arrays.

Detectors/EMCAL/reconstruction/src/StuDecoder.cxx
```diff
@@ -0,0 +1,241 @@
...
+  std::cout << "L1JetLowThreshold   :" << std::dec << getL1JetLowThreshold() << std::endl;
+  std::cout << "Rho                 :" << std::dec << getRho() << std::endl;
+  std::cout << "FrameReceivedSTU    :" << std::dec << getFrameReceivedSTU() << std::endl;
+  std::cout << "RegionEnable        :" << std::hex << "0x" << getRegionEnable() << std::endl;
+  std::cout << "FrameReceived       :" << std::dec << getFrameReceived() << std::endl;
+  std::cout << "ParchSize           :" << std::dec << getParchSize() << std::endl;
+  std::cout << "FWversion           :" << std::hex << "0x" << getFWversion() << std::endl;
+
+  std::cout << "\n";
+}
```
mfasDa, line 241: Ideally one would put the configuration in a struct, create a streaming operator (operator<<) for the struct (where I usually add a function printStream that is called by the streaming operator). Like that the dumpSTUcfg could make use of that, but other streams as well.

Detectors/EMCAL/reconstruction/src/StuDecoder.cxx
mfasDa, line 169: Furthermore, if they are intended to serve as bitsets, use std::bitset instead.

Detectors/EMCAL/reconstruction/src/StuDecoder.cxx
```diff
@@ -0,0 +1,241 @@
...
+
+  int nSubregionEta = getSubregionsEta();
+  int nSubregionPhi = getSubregionsPhi();
+  bool isFired = false;
+
+  int jetSize = 2 + getParchSize(); // 2 for 2x2-patch, 4 for 4x4=patch
+
+  for (Int_t ix = 0; ix < nSubregionEta - (jetSize - 1); ix++) {
+    uint32_t row_JetHigh = buffer[ix];
+    uint32_t row_JetLow = buffer[ix + offset];
```
mfasDa, line 136: Also those two I assume are bitsets in reality ...

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/StuDecoder.h
```diff
@@ -0,0 +1,153 @@
...
+  /// Performing various consistency checks on the RCU trailer
+  /// In case of failure an exception is thrown.
+  //  void checkRCUTrailer();
+
+  RawReaderMemory& mRawReader; ///< underlying raw reader
+
+  std::vector<int16_t> mL1JetHighPatchIndex;
+  std::vector<int16_t> mL1JetLowPatchIndex;
+  std::vector<int16_t> mL1GammaHighPatchIndex;
+  std::vector<int16_t> mL1GammaLowPatchIndex;
```
mfasDa, line 100: Only from the code one realises that this a bitset which is absolutely confusing. I recommend using bitfields as used in various places, i.e. https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/EMCAL/base/include/EMCALBase/RCUTrailer.h

An implementation would be 
```c++
struct GammaPatchIndex {
  union {
    uint16_t mDataWord;
    struct {
       uint16_t mPhiIndex: 5;
       uint16_t mEtaIndex: 5;
       uint16_t mTRUindex: 5;
    };
  };
};
```
... and then:
``` c++
std::vector<GammaPatchIndex> mGammaHighPatchIndex;
```

Similar for jet patch indices.

Detectors/EMCAL/reconstruction/src/StuDecoder.cxx
```diff
@@ -0,0 +1,241 @@
...
+      if (isFired) {
+        mL1JetLowPatchIndex.push_back(((ix << 8) & 0xFF00) | (iy & 0xFF));
+      }
+    }
+  }
+
+  return;
+}
+
+void StuDecoder::decodeL1GammaPatchIndices(const uint32_t* buffer)
```
mfasDa, line 152: In the way I see in the code buffer is a sequence of 32 bit words with 4 8 bit words, a struct can be defined as

```c++
struct GammaPatchWord{
  union {
    uint32_t mData;
    struct {
       uint32_t mWord3: 8;
       uint32_t mWord2: 8;
       uint32_t mWord1: 8;
       uint32_t mWord0: 8;
    };
  };
};
```
See particularly as exampled how the different word types are defined for ITSMFT https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/ITSMFT/common/reconstruction/include/ITSMFTReconstruction/GBTWord.h and adapted in a similar way for FOCAL https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/FOCAL/reconstruction/include/FOCALReconstruction/PadWord.h

you would call the function via
```c++
decodeL1GammaPatchIndices(gsl::span<const GammaPatchWord>(reinterpret_cast<const GammaPatchWord *>(buffer + first, nwords)));
```

mfasDa (changes requested):
In addition to dedicated code remarks nearly all std::couts need to be replaced by LOG(DEBUG) or lower level.

davidrohr:
@mfasDa @mpoghos : what is the status of this PR? It is stall for quite some time. Is it still needed? Should it be merged?

mfasDa:
@davidrohr yes, please merge!

PR #11247, opened 2023-04-27
TRD gain calibration

glegras:


Detectors/TRD/calibration/include/TRDCalibration/CalibratorGain.h
```diff
@@ -0,0 +1,91 @@
...
+  void closeOutputFile();
+
+  const std::vector<o2::trd::CalGain>& getCcdbObjectVector() const { return mObjectVector; }
+  std::vector<o2::ccdb::CcdbObjectInfo>& getCcdbObjectInfoVector() { return mInfoVector; }
+
+  void initProcessing();
+
+  /// Initialize the fit values once with the previous valid ones if they are
+  /// available.
+  void retrievePrev(o2::framework::ProcessingContext& pc);
```
f3sch, line 71: You are not doing that. Why define at all, just initialize with your default value.

Detectors/TRD/calibration/src/CalibratorGain.cxx
```diff
@@ -0,0 +1,202 @@
...
+    if (nEntries < mMinEntriesChamber) {
+      LOGF(debug, "Chamber %d did not reach minimum amount of %d entries for refit", iDet, mMinEntriesChamber);
+      continue;
+    }
+    mdEdxhists[iDet]->Scale(1. / nEntries);
+
+    // Fitting histogram
+    TF1Convolution* fConv = new TF1Convolution("[3] * TMath::Landau(x, [0], [1]) * exp(-[2]*x)", "TMath::Gaus(x, 0, [0])", 0., NBINSGAINCALIB, true);
+    fConv->SetNofPointsFFT(2000);
+    TF1* fitFunction = new TF1("fitConvLandau", *fConv, 0., NBINSGAINCALIB, fConv->GetNpar());
```
f3sch: This seems somewhat wasteful. You are never freeing the memory allocated to these functions and continuously allocating more. How about moving the  fit function to class member, while performing the convolution once during the init?

Detectors/TRD/calibration/src/CalibratorGain.cxx
```diff
@@ -0,0 +1,202 @@
...
+    fitFunction->SetParameters(mdEdxhists[iDet]->GetMean() / 1.25, 15, 0.02, 1, 0.1);
```
f3sch: Btw, how stable are your fits? I struggled somewhat with this and achieved stability with setting parameter limits like these here:
``` C
     fitFunction->SetParLimits(0, 30., 100.0);
      fitFunction->SetParLimits(1, 5.0, 25.0);
      fitFunction->SetParLimits(2, -0.1, 0.5);
      fitFunction->SetParLimits(3, -0.1, 0.5);
      fitFunction->SetParLimits(5, 1.0, 10.0);
```

Detectors/TRD/calibration/src/CalibratorGain.cxx
```diff
@@ -0,0 +1,202 @@
...
+
+    int fitStatus = mdEdxhists[iDet]->Fit(fitFunction, "LQ", "", 1, NBINSGAINCALIB - 4);
```
f3sch: If you use parameter limits, I think you must add "B" to the options here.

Detectors/TRD/calibration/src/CalibratorGain.cxx
```diff
@@ -0,0 +1,202 @@
...
+
+    if (fitStatus != 0) {
+      LOGF(info, "Fit for chamber %i failed, nEntries: %d", iDet, nEntries);
```
f3sch:
```suggestion
      LOGF(warn, "Fit for chamber %i failed, nEntries: %d", iDet, nEntries);
```

Detectors/TRD/workflow/include/TRDWorkflow/GainCalibSpec.h
```diff
@@ -0,0 +1,154 @@
...
+    mCalibrator->checkSlotsToFinalize(o2::calibration::INFINITE_TF);
+    mCalibrator->closeOutputFile();
+    sendOutput(ec.outputs());
+  }
+
+  void stop() final
+  {
+    mCalibrator->closeOutputFile();
+  }
+
```
f3sch: This mechanism has changed upstream since you last looked, for reference see #11155.

Detectors/TRD/workflow/io/src/TRDCalibWriterSpec.cxx
```diff
@@ -14,6 +14,7 @@
 #include "Framework/InputSpec.h"
```
f3sch, line 14: @martenole Doe this warrant its own writer/reader can we not merge the for the vdrift calib into one, i.e., having one file like trdcalib.root?

Detectors/TRD/calibration/src/TrackBasedCalib.cxx
```diff
@@ -77,6 +90,78 @@ void TrackBasedCalib::calculateAngResHistos()
...
+
+      const auto& tracklet = mTrackletsRaw[trkWork.getTrackletIndex(iLayer)];
+
+      float Q0 = tracklet.getQ0();
+      float Q1 = tracklet.getQ1();
+      float Q2 = tracklet.getQ2();
+      // TODO: Should we also correct for local gain here?
+      float localGainCorr = 1.;
+      if (mLocalGain)
+        localGainCorr = mLocalGain->getValue(tracklet.getHCID() / 2, tracklet.getPadCol(), tracklet.getPadRow());
```
f3sch: @pachmay, @tdietel  should the gain calibration use the krypton calibration, it seemed to me that they were independent?

Detectors/TRD/calibration/src/TrackBasedCalib.cxx
```diff
@@ -77,6 +90,78 @@ void TrackBasedCalib::calculateAngResHistos()
   // mAngResHistos.print();
 }
 
+int TrackBasedCalib::filldEdx(gsl::span<const TrackTRD>& tracks, bool isTPCTRD)
+{
+  auto& params = TRDCalibParams::Instance();
```
f3sch, line 88: How about moving this to a class member and removing all the calls to Instance()?

Detectors/TRD/calibration/src/TrackBasedCalib.cxx
```diff
@@ -77,6 +90,78 @@ void TrackBasedCalib::calculateAngResHistos()
...
+      // TODO: Should we also correct for local gain here?
+      float localGainCorr = 1.;
+      if (mLocalGain)
+        localGainCorr = mLocalGain->getValue(tracklet.getHCID() / 2, tracklet.getPadCol(), tracklet.getPadRow());
+
+      float dEdx = (Q0 + Q1 + Q2) / trkLength / localGainCorr;
+
+      if (Q0 == 0 || Q1 == 0 || Q2 == 0 || Q0 >= 127 || Q1 >= 127 || Q2 >= 62) {
+        continue;
+      }
```
f3sch, line 141: I think you should move this check before applying the corrections.

f3sch (dismissed):
Nice work :+1: ! I added some general comments, questions and I am sure others will review much more thoroughly.

martenole:
Hi, was there a presentation with the details on the gain calibration somewhere? Could you point me to that or elaborate here a bit more what you plan to do? In general its of course very good to see progress here, so thank you for your work.

Detectors/TRD/calibration/src/TrackBasedCalib.cxx
martenole: why do you actually want to ignore such values here?

Detectors/TRD/calibration/src/TrackBasedCalib.cxx
```diff
@@ -77,6 +90,78 @@ void TrackBasedCalib::calculateAngResHistos()
...
+
+    for (int iLayer = NLAYER - 1; iLayer >= 0; --iLayer) {
+      if (trkWork.getTrackletIndex(iLayer) == -1) {
+        continue;
+      }
+      if (mNoiseCalib && mNoiseCalib->isTrackletFromNoisyMCM(mTrackletsRaw[trkWork.getTrackletIndex(iLayer)])) {
+        // ignore tracklets which originate from noisy MCMs
+        continue;
+      }
+      if (propagateAndUpdate(trkWork, iLayer, false)) {
```
martenole: I don't think a full propagation is necessary here. This would duplicate a lot of work done anyhow for the vDrift and ExB calibration so it should either be combined or you just use a simple `getSnpAt(radius, b-field)` to obtain the track inclination at given radius. The eta should be constant enough in any case so you can get that from the outer param of the TRD track directly

Detectors/TRD/calibration/src/TrackBasedCalib.cxx
```diff
@@ -77,6 +90,78 @@ void TrackBasedCalib::calculateAngResHistos()
...
+        break;
+      }
+
+      float trkPhi = o2::math_utils::asin(trkWork.getSnp());
+      float trkLength = sqrt(1 + tan(trkPhi) * tan(trkPhi) + trkWork.getTgl() * trkWork.getTgl());
```
martenole: the trigonometric functions and especially the `asin` operation are rather slow, so you should check for an alternative in places which are called often like here for every track. Instead of `tan (phi)^2` you can substitute `sin (phi)^2 / (1 - sin(phi)^2)` so you can use the parameter stored in the track directly

Detectors/TRD/workflow/io/src/TRDCalibWriterSpec.cxx
martenole: I am not sure if we need reader/writer at all?

martenole (dismissed):
Some comments already now and the (**dummy!**) approval to get the CI checks started for first time contributors. I think we should discuss this, for example  at the next TRD meeting in a bit more detail

glegras:
Hi, I think that indeed, it is better to discuss it at next TRD meeting in more detail.

Let me also remind the context: We want to have automatized what I did locally (among others this presentation at this TRD meeting in January: https://indico.cern.ch/event/1248149/contributions/5244607/attachments/2584311/4457974/TRDGainCalibration.pdf )

And a description (duplicate of the mail I sent to some of you):  
For the code, I just started from the code on drift velocity calibration as a base. During global tracking within the enable-trackbased-calib option, I collect for each tracklet (after cut on ntracklets, p, and dEdxTPC of the track), the dedx in the chamber (which is just sum of the charges, corrected by local gain and track length in the chamber). The output is for each timestamp an array with number of counts in each dedx bin and for each chamber.
Then, in the calibration workflow, the fit is done (with Landau*exp convoluted with Gaussian) for every chamber and we write the Most Probable Value.

Detectors/TRD/workflow/io/src/TRDCalibWriterSpec.cxx
pachmay: Hi @martenole, I am wondering: the gain calibration should run on the tracklet charges. If we restrict it to the digits, don't we then reduce the statistics a lot. When running on tracklets in the synchronous reco, we could obtain more statistics; however we need to merge the split charges before.

Detectors/TRD/calibration/src/TrackBasedCalib.cxx
pachmay: the final tracklet charges used for PID should be corrected for chamber wise gain and local gain (krypton)

we could also correct the charges with the local gain before the chamber wise gain calibration, but I assume that there is not much difference

Detectors/TRD/workflow/io/src/TRDCalibWriterSpec.cxx
martenole: I think in this implementation the digits are not used at all and the tracklet charges are used instead. This is how it was planned, right?

pachmay: Maybe I misunderstood your earlier comment. Indeed the tracklet charges should be used; thus to get more statistics, it might be better to not restrict ourselves to the digit events and therefore we do not need/use the writer and reader for the drift velocity; is that technically possible?

pachmay: Hi Ole, maybe we could just do something similar as for the t0 calibration; we wait until enough statistics is accumulated and the fits are very successful and then proceed with the gain calibration. Would that work?

f3sch:
@glegras could you make this for now a draft, then we would now when you want to review and merge this?

glegras:
Hi, this is kind of my final version with your comments implemented. So you can start to review now.

Detectors/TRD/workflow/include/TRDWorkflow/GainCalibSpec.h
```diff
@@ -0,0 +1,168 @@
...
+{
+  using device = o2::calibration::GainCalibDevice;
+  using clbUtils = o2::calibration::Utils;
+
+  std::vector<OutputSpec> outputs;
+  outputs.emplace_back(ConcreteDataTypeMatcher{o2::calibration::Utils::gDataOriginCDBPayload, "MEANDEDX"}, Lifetime::Sporadic);
+  outputs.emplace_back(ConcreteDataTypeMatcher{o2::calibration::Utils::gDataOriginCDBWrapper, "MEANDEDX"}, Lifetime::Sporadic);
+  std::vector<InputSpec> inputs;
+  inputs.emplace_back("input", "TRD", "GAINCALIBHISTS");
+  // inputs.emplace_back("calgain", "TRD", "CALIBGAIN", 0, Lifetime::Condition, ccdbParamSpec("TRD/Calib/CalibGain"));
```
glegras: Here, the line is supposed to be used for retrieving the latest results from the ccdb (used by the function retrievePrev in CalibratorGain.cxx). However, since the ccdb entry is not yet created, it returns an error for now. So I assume it should be left commented for now, and get uncommented later. (the same goes for the function retrievePrev)

DataFormats/Detectors/TRD/include/DataFormatsTRD/GainCalibHistos.h
```diff
@@ -0,0 +1,55 @@
...
+class GainCalibHistos
+{
+ public:
+  GainCalibHistos() = default;
+  GainCalibHistos(const GainCalibHistos&) = default;
+  ~GainCalibHistos() = default;
+  void reset();
+  bool addEntry(float dEdx, int chamberId);
+  float getHistogramEntry(int index) const { return mdEdxEntries[index]; }
+  // int getBinCount(int index) const { return mNEntriesPerBin[index]; }
```
martenole: if you don't need this function, could you remove it?

DataFormats/Detectors/TRD/src/GainCalibHistos.cxx
```diff
@@ -0,0 +1,64 @@
...
+
+bool GainCalibHistos::addEntry(float dEdx, int chamberId)
+{
+  // add entry for given dEdx
+  // returns 0 in case of success
+  int chamberOffset = chamberId * NBINSGAINCALIB;
+  int iBin = (int)dEdx;
+  ++mdEdxEntries[chamberOffset + iBin];
+  ++mNEntriesTot;
+  return 0;
```
martenole: this function will always return zero. What happens in case iBin is out of bounds?

Detectors/TRD/calibration/include/TRDCalibration/CalibratorGain.h
```diff
@@ -0,0 +1,94 @@
...
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file CalibratorGain.h
+/// \brief TimeSlot-based calibration of gain
+/// \author Ole Schmidt, Gauthier Legras
```
martenole:
```suggestion
/// \author Gauthier Legras
```

Detectors/TRD/calibration/src/CalibratorGain.cxx
```diff
@@ -0,0 +1,206 @@
...
+      msg = "Previous Object";
+      break;
+    }
+  }
+  LOG(info) << "Calibrator: From CCDB retrieved " << msg;
+
+  // Here we set each entry regardless if it is the default or not.
+  for (int iDet = 0; iDet < MAXCHAMBER; ++iDet) {
+    fitResults[iDet] = dataCalGain->getMPVdEdx(iDet);
+  }*/
```
martenole: We can upload a default object of the gain calibration before we merge your PR. So you can uncomment these lines already now.

Detectors/TRD/calibration/src/CalibratorGain.cxx
```diff
@@ -0,0 +1,206 @@
...
+{
+  if (mInitDone) {
+    return;
+  }
+  for (int iDet = 0; iDet < MAXCHAMBER; ++iDet) {
+    mdEdxhists[iDet] = std::make_unique<TH1F>(Form("hdEdx%d", iDet), "dEdx", NBINSGAINCALIB, 0., NBINSGAINCALIB);
+  }
+
+  // Init fitting functions
+  fConv = new TF1Convolution("[3] * TMath::Landau(x, [0], [1]) * exp(-[2]*x)", "TMath::Gaus(x, 0, [0])", 0., NBINSGAINCALIB, true);
```
martenole: all class member variables must start with `m`, so `mFconv` for example would be valid. Also, I can you replace the plain `new` and use smart pointers instead?

Detectors/TRD/calibration/src/CalibratorGain.cxx
```diff
@@ -0,0 +1,206 @@
...
+  fConv->SetNofPointsFFT(2000);
+  fitFunction = new TF1("fitConvLandau", *fConv, 0., NBINSGAINCALIB, fConv->GetNpar());
```
martenole: same as above

Detectors/TRD/calibration/src/CalibratorGain.cxx
martenole: should be OK without as long as no predefined function is used for fitting: https://root.cern.ch/doc/master/classTH1.html#a7e7d34c91d5ebab4fc9bba3ca47dabdd

Detectors/TRD/calibration/src/CalibratorGain.cxx
```diff
@@ -0,0 +1,206 @@
...
+  CalGain calObject;
+  for (int iDet = 0; iDet < MAXCHAMBER; ++iDet) {
+    // For Chambers which did not have the minimum amount of entries in this slot e.g. missing, empty chambers.
+    // We will reuse the previous one. This would have been read either from the ccdb or come from a previous successful fit.
+    calObject.setMPVdEdx(iDet, fitResults[iDet]);
+  }
+  auto clName = o2::utils::MemFileHelper::getClassName(calObject);
+  auto flName = o2::ccdb::CcdbApi::generateFileName(clName);
+  std::map<std::string, std::string> metadata; // TODO: do we want to store any meta data?
+  long startValidity = slot.getStartTimeMS() - 10 * o2::ccdb::CcdbObjectInfo::SECOND;
```
martenole: there is no need for the extra protection and -10 seconds

Detectors/TRD/calibration/src/CalibratorGain.cxx
```diff
@@ -0,0 +1,206 @@
...
+  mInfoVector.emplace_back("TRD/Calib/CalibGain", clName, flName, metadata, startValidity, startValidity + o2::ccdb::CcdbObjectInfo::HOUR);
```
martenole: I would set the end validity to + 3 days. By default, the objects will be uploaded to the CCDB with adjustable end of validity. This means that the object you upload will in the beginning have a long validity, but the moment a new object is uploaded the end validity of the previous object is set to the start validity of the new one. So the validities don't overlap and you avoid gaps. I should do the same for the vD and ExB calibration

Detectors/TRD/calibration/src/CalibratorGain.cxx
martenole: Actually, what do you plan to do with the previous values which you try to retrieve here?

Detectors/TRD/workflow/include/TRDWorkflow/GainCalibSpec.h
martenole: we can upload a default object already now

Detectors/TRD/workflow/include/TRDWorkflow/GainCalibSpec.h
```diff
@@ -0,0 +1,168 @@
...
+  void init(o2::framework::InitContext& ic) final
+  {
+    o2::base::GRPGeomHelper::instance().setRequest(mCCDBRequest);
+    auto slotL = ic.options().get<uint32_t>("sec-per-slot");
+    auto delay = ic.options().get<uint32_t>("max-delay");
+    mCalibrator = std::make_unique<o2::trd::CalibratorGain>();
+    mCalibrator->setSlotLengthInSeconds(slotL);
+    mCalibrator->setMaxSlotsDelay(delay);
+    if (ic.options().get<bool>("enable-root-output")) {
+      mCalibrator->createOutputFile(); // internally checks if output is enabled
```
martenole: the comment does not make sense here

Detectors/TRD/workflow/include/TRDWorkflow/GainCalibSpec.h
```diff
@@ -0,0 +1,168 @@
...
+
+} // namespace calibration
+
+namespace framework
+{
+
+DataProcessorSpec getTRDGainCalibSpec()
+{
+  using device = o2::calibration::GainCalibDevice;
+  using clbUtils = o2::calibration::Utils;
```
martenole: seems you are not using these aliases, could you please remove them?

Detectors/TRD/workflow/src/TrackBasedCalibSpec.cxx
```diff
@@ -70,7 +70,10 @@ void TRDTrackBasedCalibDevice::run(ProcessingContext& pc)
   updateTimeDependentParams(pc); // Make sure this is called after recoData.collectData, which may load some conditions
   mCalibrator.setInput(recoData);
   mCalibrator.calculateAngResHistos();
+  mCalibrator.calculateGainCalibObjs();
```
martenole: You will need to add some logic here. So far the track based calibration was only important for the vDrift and ExB calibration. Now you are adding gain, but we don't have both calibrations enabled always. So the track based calibration needs to be aware which calibration should be run and input/output and processing should be configured accordingly

Detectors/TRD/calibration/src/TrackBasedCalib.cxx
```diff
@@ -77,6 +82,85 @@ void TrackBasedCalib::calculateAngResHistos()
...
+      if (Q0 == 0 || Q1 == 0 || Q2 == 0 || Q0 >= 127 || Q1 >= 127 || Q2 >= 62) {
+        continue;
+      }
+
+      // Correction for tracklet angle
+      const auto& trackletCalib = mTrackletsCalib[trkltId];
+      float tgl = trkIn.getTgl();
+      float snp = trkIn.getSnpAt(o2::math_utils::sector2Angle(trkltSec), trackletCalib.getX(), bz);
+
+      float trkLength = sqrt(1 + snp * snp / (1 - snp * snp) + tgl * tgl);
```
martenole, line 151: can you please add a comment here how you calculate the track length and double check that this is correct?

Detectors/TRD/calibration/src/TrackBasedCalib.cxx
```diff
@@ -77,6 +82,85 @@ void TrackBasedCalib::calculateAngResHistos()
...
+
+      float localGainCorr = 1.;
+      if (mLocalGain) {
+        localGainCorr = mLocalGain->getValue(tracklet.getHCID() / 2, tracklet.getPadCol(), tracklet.getPadRow());
```
martenole:
```suggestion
        localGainCorr = mLocalGain->getValue(tracklet.getDetector(), tracklet.getPadCol(), tracklet.getPadRow());
```

martenole (changes requested):
Thanks, I think we are on a good track here. The only larger task which is still missing is to add some logic to the TrackBasedCalib class. Let me know in case some comments are not clear

Detectors/TRD/calibration/src/CalibratorGain.cxx
glegras: I was thinking that uploading the previous value when not enough statistics should be better that uploading the default value.
But in fact, we could also decide to do an interpolation, so maybe it is better to let it commented (or even remove it and let the default value) and discuss it later.

Detectors/TRD/calibration/src/CalibratorGain.cxx
glegras: What do you exactly want me to do with smart pointers?

Detectors/TRD/workflow/src/TrackBasedCalibSpec.cxx
glegras: I would think by by default we will always doing both calibrations during reconstruction?
But okay, I can still add the options enable-vdexb-calib and enable-gain-calib
In this case, should I let the option enable-trackbased-calib for post compatibility, or can I remove it?

Detectors/TRD/calibration/src/CalibratorGain.cxx
martenole: you should avoid a plain "new" in your code. Here for example you are never deleting fConv and fitFunction. Even though both are created in the init function which is called only once this is bad practice and should be avoided

Detectors/TRD/calibration/src/CalibratorGain.cxx
martenole: I would not complicate things with an interpolation. Taking the value from the previous calibration is OK. As said, we can upload a default value beforehand and then you can uncomment this block already now

Detectors/TRD/workflow/src/TrackBasedCalibSpec.cxx
martenole: By default we definitely want the vDrift and ExB calibration during the synchronous reco, since they have an impact on the tracklets which are used for global tracking and the TPC calibration. During the asynchronous reco we would want both to be active.
The enable-trackbased-calib option should stay. And you should add another option to the trd-tracking-workflow named enable-gain-calib for example. This boolean flag (false by default) should be passed to the track based calib device which then either adds the gain calibration or not. 
So in case trackbased calibrations are enabled we will always have vD and ExB, but gain will be optional

Detectors/TRD/calibration/src/CalibratorGain.cxx
glegras: Hi @martenole, in this case can you upload the object already now or tell me how to do it, so that I can uncomment and test the code before updating the PR?

martenole: Sure. To what should I set the default? 42?

martenole: Hi @glegras can you check if this is OK? http://ccdb-test.cern.ch:8080/browse/TRD/Calib/CalibGain

glegras: Hi @martenole, can you change the start validity date of the default object to beginning of 2022?

martenole (dismissed):
DON'T MERGE YET, approving so that the CI starts

Detectors/TRD/workflow/src/TrackBasedCalibSpec.cxx
```diff
@@ -133,14 +154,16 @@ DataProcessorSpec getTRDTrackBasedCalibSpec(o2::dataformats::GlobalTrackID::mask
                                                               o2::base::GRPGeomRequest::Aligned, // geometry
                                                               inputs,
                                                               true);
+  outputs.emplace_back(o2::header::gDataOriginTRD, "GAINCALIBHISTS", 0, Lifetime::Timeframe);
```
martenole: The changes you do here to the DPL device should depend on whether the gain calibration was enabled or not. If not, then no additional inputs/outputs should be requested

martenole (commented):
Hi Gauthier, this looks very good and we can almost merge it as is. You will only need to change the `Detectors/TRD/workflow/src/trd-tracking-workflow.cxx` and add the option `enable-gain-calib` there instead of to the TrackBasedCalib.
Because from that workflow you can pass the option down to the TrackBasedCalib device and then decide there which input/output should be requested.
Then I think we are good to go

martenole (approved):
Approving so that the CI checks are started.
@glegras I had to add some changes on top, especially because the large `std::array<int, nChambers*nBins>` was not feasible in the `GainCalibHistos` class. Now it should be OK. I did a small test reconstruction just to see that the workflow does not crash. But I did not check yet whether the output is meaningful.
In any case, when the CI is green we can merge it as is and then you can continue to commission the gain calibration.
Cheers,
Ole

PR #11249, opened 2023-04-28
Refactor libuv contexts

ktf:


ktf:
Formatting will be fixed in a separate PR. Merging to do progress on the uv_write / uv_read on the main thread.

PR #11256, opened 2023-04-29
DPL: always run the first loop

ktf:
This is for the case in which a device has no inputs and therefore it should not wait for incoming data.


ktf:
Attempt at fixing case in which the workflow does not start. What is probably happening is that before the changes for 100% CPU usage the triggering of the event loop was the reason why it started computing. Fixed that, the source devices which do not have an out of band input (e.g. the readers) can happen to miss the initial "kick". I will try a better solution later today, but this should be already enough.

PR #11264, opened 2023-05-02
DPL Analysis: update running workflow configuration

aalkin:
This updates RunningWorkflowInfo with the current configuration, so it can be examined by all tasks. Additionally fixes assigning vectors to Variant so that size is preserved.

@ddobrigk once again, the configuration can be inherited
@vkucera the issue with the array size resetting to 1 is fixed

### DIFF
--- a/Framework/Core/src/runDataProcessing.cxx
+++ b/Framework/Core/src/runDataProcessing.cxx
@@ -1611,6 +1611,91 @@ int runStateMachine(DataProcessorSpecs const& workflow,
           }
 
           // This should expand nodes so that we can build a consistent DAG.
+
+          // This updates the options in the runningWorkflow.devices
+          for (auto& device : runningWorkflow.devices) {
+            // ignore internal devices
+            if (device.name.find("internal") != std::string::npos) {
+              continue;
+            }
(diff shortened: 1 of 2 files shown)


vkucera:
> This updates RunningWorkflowInfo with the current configuration, so it can be examined by all tasks. Additionally fixes assigning vectors to Variant so that size is preserved.
> 
> @ddobrigk once again, the configuration can be inherited @vkucera the issue with the array size resetting to 1 is fixed

Thanks @aalkin

Framework/Core/src/runDataProcessing.cxx
```diff
@@ -1611,6 +1611,91 @@ int runStateMachine(DataProcessorSpecs const& workflow,
           }
 
           // This should expand nodes so that we can build a consistent DAG.
+
+          // This updates the options in the runningWorkflow.devices
+          for (auto& device : runningWorkflow.devices) {
+            // ignore internal devices
+            if (device.name.find("internal") != std::string::npos) {
```
ktf, line 1618: `rfind("internal", 0)`, no?

aalkin: I do not think we have cases where a device would have "internal" more than once in its name. We can probably simplify this, by adding a bool flag to the DeviceSpec.

PR #11265, opened 2023-05-02
DPL Analysis: consistently use accessor method

aalkin:
@mhemmer-cern fix for the PresliceUnsorted


mhemmer-cern:
Thank you Anton!

PR #11267, opened 2023-05-02
Remove TPC starting row cut

fmazzasc:
@shahor02 , this removes the cut on the starting row of the TPC track in the matching. This helps to recover ~ 10% of the secondary track missing matchings. The number of fakes looks still negligible in the MC. In the plot the algorithm efficiency is defined as: (# of true matched candidates) / (# of candidates with both ITS and TPC tracks)
c


shahor02 (approved):

ddobrigk:
Thanks a lot @fmazzasc ! 

For reference, this was discussed in the AOT tracks meeting earlier today: 
https://indico.cern.ch/event/1281698/contributions/5387798/attachments/2640630/4569917/Strangeness-Reco-04.pdf

A specific point is that, while of course the PR is the best way to remove the cut altogether, the same effect can be achieved by setting this cut to a large value (e.g. 200). It could therefore be incorporated in `apass4` reconstruction of pp as a parameter change and not necessarily as a new tag / ported code, which may simplify matters a bit.

fmazzasc:
@shahor02 , could you merge this?

PR #11268, opened 2023-05-02
DPL: another attempt at fixing a race condition in readers

ktf:
Drop obsolete comment as well.


ktf:
Actually it's not this. It should be handled correctly by the current code. I am starting to suspect some ABI issue...

PR #11270, opened 2023-05-02
[FOCAL-13] Make BadChannelMap streamable

mfasDa:
enumeration datatype was not streamable. Using
uint8_t as underlying datatype.


mfasDa:
CI errors unrelated

PR #11271, opened 2023-05-02
dev: creating decoding method which can be used also in QC ctp

lietava:
Hi @shahor02 , we are doing code for CTP QC. I realised that it would be useful to have function in decoding which can be used also in:
https://github.com/AliceO2Group/QualityControl/blob/master/Modules/CTP/src/RawDataQcTask.cxx
Do you see any problem with that ?
Any other comments ?
Cheers, R.

### DIFF
--- a/Detectors/CTP/reconstruction/include/CTPReconstruction/RawDataDecoder.h
+++ b/Detectors/CTP/reconstruction/include/CTPReconstruction/RawDataDecoder.h
@@ -41,9 +41,9 @@ class RawDataDecoder
   void setDoLumi(bool lumi) { mDoLumi = lumi; }
   void setDoDigits(bool digi) { mDoDigits = digi; }
   void setVerbose(bool v) { mVerbose = v; }
-  uint32_t getIRRejected() { return mIRRejected; }
-  uint32_t getTCRRejected() { return mTCRRejected; }
-  std::vector<uint32_t> getTFOrbits() { return mTFOrbits; }
+  uint32_t getIRRejected() const { return mIRRejected; }
+  uint32_t getTCRRejected() const { return mTCRRejected; }
+  std::vector<uint32_t>& getTFOrbits() { return mTFOrbits; }
   int init();
 
  private:
--- a/Detectors/CTP/reconstruction/include/CTPReconstruction/RawDataDecoder.h
+++ b/Detectors/CTP/reconstruction/include/CTPReconstruction/RawDataDecoder.h
@@ -1,0 +1,69 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file RawDataDecoder.h
+/// \brief Digits  tw Raw translation
+/// \author Roman Lietava
+
+#ifndef ALICEO2_CTP_RAWDATADECODER_H_
+#define ALICEO2_CTP_RAWDATADECODER_H_
+
+#include <vector>
+#include <map>
+#include <deque>
+#include "DPLUtils/DPLRawParser.h"
+#include "Framework/InputRecord.h"
+#include "Framework/InputRecord.h"
+#include "DataFormatsCTP/Digits.h"
+#include "DataFormatsCTP/LumiInfo.h"
+#include "DataFormatsCTP/TriggerOffsetsParam.h"
+
+namespace o2
+{
+namespace ctp
+{
+class RawDataDecoder
+{
+ public:
+  RawDataDecoder() = default;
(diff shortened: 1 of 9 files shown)


shahor02:
The decoding function itself is fine but using RawDecoderSpec as an include for other processor is not good, may mess up dependencies. It is better to put it into a separate class, even as having only this static method.
Also, to pass the CI, we will need 1st to merge O2 with this method, then open PR for QC.

lietava:
@shahor02 : thanks. I'll create new class. In workflow or in reconstruction ?
Yes, we shall wait with QC for this. btw what is 'CI' ?

shahor02:
I would put it in the reconstruction, with all framework classes forward declared in the header and included in the implementation.
CI is for the continuous integration (the build test running for every PR)

lietava:
Hi @shahor02 , created separate class. Please, have a look.

lietava:
@shahor02  build/O2/o2-dataflow failed. But I can not see any CTP related error ?

Detectors/CTP/reconstruction/include/CTPReconstruction/RawDataDecoder.h
```diff
@@ -0,0 +1,70 @@
...
+  RawDataDecoder() = default;
+  ~RawDataDecoder() = default;
+  static void makeGBTWordInverse(std::vector<gbtword80_t>& diglets, gbtword80_t& GBTWord, gbtword80_t& remnant, uint32_t& size_gbt, uint32_t Npld);
+  int addCTPDigit(uint32_t linkCRU, uint32_t triggerOrbit, gbtword80_t& diglet, gbtword80_t& pldmask, std::map<o2::InteractionRecord, CTPDigit>& digits);
+  int decodeRaw(o2::framework::InputRecord& inputs, std::map<o2::InteractionRecord, CTPDigit>& digits, std::vector<LumiInfo>& lumiPointsHBF1);
+  void setDoLumi(bool lumi) { mDoLumi = lumi; }
+  void setDoDigits(bool digi) { mDoDigits = digi; }
+  void setVerbose(bool v) { mVerbose = v; }
+  uint32_t getIRRejected() { return mIRRejected; }
+  uint32_t getTCRRejected() { return mTCRRejected; }
```
shahor02: these are const getters

Detectors/CTP/reconstruction/include/CTPReconstruction/RawDataDecoder.h
```diff
@@ -0,0 +1,70 @@
...
+  std::vector<uint32_t> getTFOrbits() { return mTFOrbits; }
```
shahor02: Here you return a copy of the vector but as far as I can see you never modify it. I guess it can be
```
const std::vector<uint32_t>& getTFOrbits() const { return mTFOrbits; }
```

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
```diff
@@ -26,28 +25,32 @@ void RawDecoderSpec::init(framework::InitContext& ctx)
...
+  mDecoder.setVerbose(mVerbose);
+  mDecoder.setDoLumi(mDoLumi);
+  mDecoder.setDoDigits(mDoDigits);
   LOG(info) << "CTP reco init done";
 }
 void RawDecoderSpec::endOfStream(framework::EndOfStreamContext& ec)
 {
-  std::sort(mTFOrbits.begin(), mTFOrbits.end());
-  size_t l = mTFOrbits.size();
+  std::vector<uint32_t> TFOrbits = mDecoder.getTFOrbits();
```
shahor02: if you change the `getTFOrbits()` in the way I've suggested, then here there should be 
```
const auto& TFOrbits = mDecoder.getTFOrbits();
```

Detectors/CTP/reconstruction/include/CTPReconstruction/RawDataDecoder.h
lietava: yes

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
lietava: but I modify TFOrbits in the line 36 ?

shahor02: OK, did not notice the `sort`. But then before it was applied to the copy rather than to the data member. No it is ok.

shahor02 (dismissed):

Detectors/CTP/workflow/src/RawDecoderSpec.cxx
lietava: yes, I understand, thanks.

lietava:
Hi @shahor02 , this can be merged ?

shahor02:
@lietava it says ` branch cannot be rebased due to conflicts`, can you check?

lietava:
I do not see anything on my web interface . I solved conflict yesterday - see commits - last merge  ?

shahor02:
image
Can you update your dev and try to do `git rebase dev` from your branch ?

lietava:
I pushed by force. Now I have conflict:
image

lietava:
I resolve the conflict now.

lietava:
Conflict resolved.

lietava:
sorry, I didnt see this
"Can you update your dev and try to do git rebase dev from your branch ?"
I will do.

PR #11272, opened 2023-05-03
Allow using catch2 from QC

ktf:


ktf:
@Barthelemy I think this might work...

ktf:
macOS ARM error unrelated. Merging.

PR #11275, opened 2023-05-03
DPL: drop old assertion

ktf:


ktf:
@sawenzel thank you for your report.

PR #11278, opened 2023-05-03
AOD Producer: reserve table before filling

ktf:


ktf:
@davidrohr @ddobrigk @pzhristov @nburmaso I suspect the reason the AOD producer is so expensive is that it does not reserve the memory before filling it, with the net result it keeps reallocating things, just like push_back on an empty vector. I think this change (and similar ones for all the other tables) should solve the problem.

ktf:
@sawenzel @benedikt-voelkel how can we test this? Does it run in the sim tests?

ktf:
Ok, the issue was an unrelated ROOT v6-28-04 one. Now this with the correct label seems to work fine. We probably need to do the reserve for all the big tables.

ktf:
Errors unrelated. I explicitly checked that the number of tracks is the same before and after. I suspect this might give some improvement already. If someone could do the same for all the other tables that would be great.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -2593,7 +2596,7 @@ DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src, bool enableSV, boo
   outputs.emplace_back(OutputLabel{"O2mcparticle_001"}, "AOD", "MCPARTICLE_001", 0, Lifetime::Timeframe);
   outputs.emplace_back(OutputLabel{"O2mctracklabel"}, "AOD", "MCTRACKLABEL", 0, Lifetime::Timeframe);
   outputs.emplace_back(OutputLabel{"O2mfttrack"}, "AOD", "MFTTRACK", 0, Lifetime::Timeframe);
-  outputs.emplace_back(OutputLabel{"O2track_iu"}, "AOD", "TRACK_IU", 0, Lifetime::Timeframe);
+  outputs.emplace_back(OutputLabel{"Tracks_IU"}, "AOD", "TRACK_IU", 0, Lifetime::Timeframe);
```
ktf, line 2599: One should probably use `o2::aod::MetadataTrait<o2::aod::StoredTracksIU>::metadata::tableLabel()` and the the various `o2::aod::MetadataTrait<o2::aod::StoredTracksIU>::metadata::origin()` `o2::aod::MetadataTrait<o2::aod::StoredTracksIU>::metadata::description()`.

PR #11279, opened 2023-05-03
Fix and report decreasing chips order on the cable

shahor02:
@iravasen new decoding error type is added for the ordering problem related to https://ali-bookkeeping.cern.ch/?page=log-detail&id=49886


shahor02:
Don't merge until approved by the ITS

PR #11282, opened 2023-05-04
Demote TextDriverClient messages

ktf:


ktf:
@Barthelemy @davidrohr this should get rid of the messages CONTROL_ACTION...

Barthelemy:
@ktf Thank you

PR #11284, opened 2023-05-04
Do not expose boost to new ROOT

ktf:


ktf:
@sawenzel I have not yet tried with the full stuff, however the new root / cling seems not to like boost and the missing auto we have. Do you see any problem having the above?

ktf:
For the record, I now get:

```
179 QEDGenParam.xSectionHad : 8             [ CODE ]
180 [^[[01;36m22:59:27^[[0m][^[[01;32mINFO^[[0m] Info: QED background generation parameters stored to qedgenparam.ini
181 (o2::eventgen::GeneratorTGenerator *) 0x159d125a0
182 [^[[01;36m22:59:27^[[0m][^[[01;31mFATAL^[[0m] Failed to retrieve 'extgen': problem with configuration
183 libc++abi: terminating due to uncaught exception of type fair::FatalException
```

any idea of what I am doing wrong? This is with:

```
NEvents=5 NEventsQED=100 $O2_ROOT/prodtests/full_system_test.sh
```

ktf:
Ok, issue understood. fg must actually be a static somewhere so it works when running the full CI (rather than just importing the macro). That said the issue with boost is genuine and it's fixed by this PR. Merging.

PR #11285, opened 2023-05-04
TPC corr.map scaling options

shahor02:
All workflows which need TPC correction maps got an extra option `--require-ctp-lumi`, 
if active it will request CTP LumiInfo which will be used to rescale the TPC corrections
maps a la Run2.
Additionally, every device using TPC corr.maps got extra options
`--corrmap-lumi-mean <float>` and `--corrmap-lumi-inst <float>` which allow to
override the mean lumi characterizing the map and instantaneous lumi for rescaling
(normally expected from CTP LumiInfo).

### DIFF
--- a/Detectors/TPC/workflow/src/tpc-reco-workflow.cxx
+++ b/Detectors/TPC/workflow/src/tpc-reco-workflow.cxx
@@ -68,6 +68,7 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
     {"configKeyValues", VariantType::String, "", {"Semicolon separated key=value strings (e.g.: 'TPCHwClusterer.peakChargeThreshold=4;...')"}},
     {"configFile", VariantType::String, "", {"configuration file for configurable parameters"}},
     {"filtered-input", VariantType::Bool, false, {"Filtered tracks, clusters input, prefix dataDescriptors with F"}},
+    {"require-ctp-lumi", o2::framework::VariantType::Bool, false, {"require CTP lumi for TPC correction scaling"}},
     {"select-ir-frames", VariantType::Bool, false, {"Subscribe and filter according to external IR Frames"}}};
   o2::raw::HBFUtilsInitializer::addConfigOption(options);
   std::swap(workflowOptions, options);
@@ -137,7 +138,7 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   std::vector<int> laneConfiguration = tpcSectors; // Currently just a copy of the tpcSectors, why?
   auto nLanes = cfgc.options().get<int>("tpc-lanes");
   auto inputType = cfgc.options().get<std::string>("input-type");
-
+  auto requireCTPLumi = cfgc.options().get<bool>("require-ctp-lumi");
   // depending on whether to dispatch early (prompt) and on the input type, we
   // set the matcher. Note that this has to be in accordance with the OutputSpecs
   // configured for the PublisherSpec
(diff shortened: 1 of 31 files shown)


Detectors/TPC/workflow/src/tpc-reco-workflow.cxx
```diff
@@ -137,7 +138,7 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
   std::vector<int> laneConfiguration = tpcSectors; // Currently just a copy of the tpcSectors, why?
   auto nLanes = cfgc.options().get<int>("tpc-lanes");
   auto inputType = cfgc.options().get<std::string>("input-type");
-
+  auto requireCTPLumi = cfgc.options().get<bool>("require-ctp-lumi");
```
wiechula, line 141: Shouldn't this be used in the `getWorkflow` call below?

wiechula (changes requested):

Detectors/TPC/workflow/src/tpc-reco-workflow.cxx
shahor02: Thanks for spotting, @wiechula , indeed, the option was checked but not passed to the getWorkflow

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -220,6 +220,7 @@ void GPURecoWorkflowSpec::init(InitContext& ic)
 
     // initialize TPC calib objects
     initFunctionTPC();
+    mFastTransformHelper->init(ic);
```
davidrohr: can you move this to initFunctionsTPC? Would make sense thematically.

GPU/Workflow/src/GPUWorkflowSpec.cxx
```diff
@@ -879,7 +889,8 @@ Inputs GPURecoWorkflowSpec::inputs()
     inputs.emplace_back("tpctopologygain", gDataOriginTPC, "TOPOLOGYGAIN", 0, Lifetime::Condition, ccdbParamSpec(CDBTypeMap.at(CDBType::CalTopologyGain)));
     inputs.emplace_back("tpcthreshold", gDataOriginTPC, "PADTHRESHOLD", 0, Lifetime::Condition, ccdbParamSpec("TPC/Config/FEEPad"));
     o2::tpc::VDriftHelper::requestCCDBInputs(inputs);
-    o2::tpc::CorrectionMapsLoader::requestCCDBInputs(inputs);
+    Options optsDummy;
+    mFastTransformHelper->requestCCDBInputs(inputs, optsDummy, mSpecConfig.requireCTPLumi); // option filled here is lost
```
davidrohr, line 892: not that it changes much, but you can save one line by just writing `optsDummy()` in the call to `requestCCDBI`.

davidrohr (commented):
Fine with me, just some minor comments.
One thing we have to follow up thoug: If I understand correctly, we'll have to update calib objects for GPU tracking with every time frame?
Currently this update is rather slow, and with 32 orbit time frames, it creates a significant performance overhead. Particularly without the multi-threading available.

GPU/Workflow/src/GPUWorkflowSpec.cxx
shahor02: ok, but this required to pass `ic` to initFunctionTPC.

GPU/Workflow/src/GPUWorkflowSpec.cxx
shahor02: I had it like that initially, but it complained about assignment to rvalue

shahor02:
> Currently this update is rather slow, and with 32 orbit time frames, it creates a significant performance overhead. Particularly without the multi-threading available.

If the rescaling is requested via `--corrmap-lumi-inst  l>0` option, which overrides inst. lumi coming from the CTP, then the map will be updated only when a one is provided (at least 1 min period). In case we rely on the CTP lumi (most of 2022 data does not have it) then indeed, for every TF we will likely rescale the map. To avoid this we may add some margin, so that the map is updated only when the lumi change exceeds it. 
Though, it would be better to be able to change the lumi settings in the transform w/o recreating the whole transform object.

PR #11291, opened 2023-05-05
Add missing includes for gcc13.1

wiechula:


martenole (dismissed):

martenole:
Looks good to me, but I have no idea why the cstdint header cannot be found from the Tracklet64.h in the fullCI
edit: its included from GPU code where that header is not available, so the include needs to be protected

DataFormats/Detectors/TRD/include/DataFormatsTRD/Tracklet64.h
```diff
@@ -27,6 +27,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#include <cstdint>
```
martenole, line 31:
```suggestion
#ifndef GPUCA_GPUCODE_DEVICE
#include <cstdint>
#endif
```

wiechula:
Thanks @martenole !

wiechula:
The error looks unrelated, can this be merged?

martenole (approved):

PR #11293, opened 2023-05-06
Add FOCAL to rawTF reader

mfasDa:


shahor02 (commented):
I would avoid requiring Focal by default in the raw reader. You may have it allowed, but then the default should be not 'all' but 'all'-FOC

mfasDa:
this formatting clash gets really annoying ...

mfasDa:
Re-formatted with ancient clang-format-13, now formatting checks pass. Incredible.

Detectors/Raw/TFReaderDD/src/tf-reader-workflow.cxx
```diff
@@ -25,7 +25,7 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
   // option allowing to set parameters
   std::vector<ConfigParamSpec> options;
   options.push_back(ConfigParamSpec{"input-data", VariantType::String, "", {"input data (obligatory)"}});
-  options.push_back(ConfigParamSpec{"onlyDet", VariantType::String, "all", {"list of dectors"}});
+  options.push_back(ConfigParamSpec{"onlyDet", VariantType::String, "all-FOC", {"list of dectors"}});
```
shahor02, line 28: this will not work, the all-FOC should be implemented as: 
```
o2::detectors::DetID::getMask("all") & ~o2::detectors::DetID::getMask("FOC")
```

shahor02 (changes requested):

mfasDa:
Withdrawing PR

PR #11295, opened 2023-05-06
McTracks to o2::McParticles

DanielSamitz:
Suggestion for a converter McEventHeader/McTracks --> o2::McCollisions/o2::McParticles_001 for on-the-fly MC analysis. The idea is to be able to use the same analysis task for on-the-fly MC analysis and for running on AO2Ds, and also to be able to connect the on-the-fly analysis to tables produced by other standard o2 analysis tasks.

WIP, can probably be done in a more elegant way (ideally without the need for any extra specification of the InputSpec of the analysis task and the --on-the-fly flag)

### DIFF
--- a/run/CMakeLists.txt
+++ b/run/CMakeLists.txt
@@ -111,6 +111,16 @@ o2_add_executable(kine-publisher
                   SOURCES o2sim_kine_publisher.cxx
                   PUBLIC_LINK_LIBRARIES O2::Framework O2::SimulationDataFormat O2::Steer)
 
+o2_add_executable(mctracks-to-aod
+                  COMPONENT_NAME sim
+                  SOURCES o2sim_mctracks_to_aod.cxx
+                  PUBLIC_LINK_LIBRARIES O2::Framework O2::SimulationDataFormat)
+
+o2_add_dpl_workflow(mctracks-to-aod-simple-task
+                    SOURCES SimExamples/McTracksToAOD/mctracks_to_aod_simple_task.cxx
+                    PUBLIC_LINK_LIBRARIES O2::Framework
+                    COMPONENT_NAME Analysis)
+
 o2_data_file(COPY o2simtopology_template.json DESTINATION config)
 
 # * # add a complex simulation as a unit test (if simulation was enabled)
--- a/run/SimExamples/McTracksToAOD/mctracks_to_aod_simple_task.cxx
+++ b/run/SimExamples/McTracksToAOD/mctracks_to_aod_simple_task.cxx
@@ -13,16 +13,10 @@
 #include "Framework/HistogramRegistry.h"
 #include "Framework/AnalysisDataModel.h"
 #include "Framework/ConfigParamSpec.h"
+#include "Framework/runDataProcessing.h"
 
 using namespace o2;
 using namespace o2::framework;
-
-void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
-{
-  workflowOptions.push_back(ConfigParamSpec{"on-the-fly", o2::framework::VariantType::Bool, false, {"set the inputs for on-the-fly analysis"}});
-}
-
-#include "Framework/runDataProcessing.h"
 
 struct SimpleTask {
 
@@ -51,13 +45,6 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
 {
   WorkflowSpec specs;
   DataProcessorSpec dSpec = adaptAnalysisTask<SimpleTask>(cfgc, TaskName{"mctracks-to-aod-simple-task"});
-  if (cfgc.options().get<bool>("on-the-fly")) {
-    std::vector<InputSpec> inputs;
-    inputs.emplace_back("McCollisions", "AOD", "MCCOLLISION", 0, Lifetime::Timeframe);
-    inputs.emplace_back("McParticles", "AOD", "MCPARTICLE_001", 0, Lifetime::Timeframe);
-    inputs.emplace_back("McParticles_001Extension", "AOD", "MCPARTICLE_001E", 0, Lifetime::Timeframe);
-    dSpec.inputs = inputs;
-  }
   specs.emplace_back(dSpec);
   return specs;
 }
--- a/run/SimExamples/McTracksToAOD/mctracks_to_aod_simple_task.cxx
+++ b/run/SimExamples/McTracksToAOD/mctracks_to_aod_simple_task.cxx
@@ -1,0 +1,63 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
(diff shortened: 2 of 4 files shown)


aalkin:
There should be no need to replace analysis task input specs... However, there is a known bug in AOD producer that uses incorrect specifications for the messages it creates with the tables. The only reason it works as it is now, is that those tables are written in the AOD file with correct tree names.

@ddobrigk we do need to fix the versioning specs in AODProducer before we miss things like this get merged.

DanielSamitz:
> There should be no need to replace analysis task input specs... However, there is a known bug in AOD producer that uses incorrect specifications for the messages it creates with the tables. 

Yes, I noticed that when I try to do this only for the McCollisions, then it works "out of the box" without changing anything at all in the analysis task. Only when adding also the McParticles somehow the aod-reader comes in and gives an error `Found duplicate outputs TFF/TFFilename/0 in device internal-dpl-aod-reader (1) and TFF/TFFilename/0 in mctracks-to-aod (3)`, and that's why I had to do it the way it is now for the moment.

aalkin:
Well, my suggestion is not to follow AODProducer in this and make this code consistent with data model from the start. Please note that TFNumber is *not* an event number, but a timeframe number, as the tables (including particles table) are accumulated for a several hundreds of events before being sent, I recommend doing the same thing here. Please also note how particles table is defined in AnalysisDataModel.h, it is an extended table, meaning it has two pieces, that you need to create *separately*. The "Stored" version and the extension (with additional columns *only*, and not the full list). The framework will combine them automatically for the analysis task. In the same data model you can also see that definitions for particles table have desc "MCPARTICLE", and the version 0 or 1 is encoded as subspec. However, instead of hardcoding, it is much better to use the data model defined metadata (see, for example, AnalysisHelpers.h, a method define in lines 111-114).

DanielSamitz:
Hi @aalkin , thank you for your help.
Now I managed to get it to run without changing any InputSpecs of the analysis task, by using `Produces` and `Spawns` for the tables.
The collisions are now collected 200 hundred per timeframe for the output, i.e. when I write out any tables produced by the analysis task there are 200 events per DF_xxx subdirectory. But I think they are still sent one by one. So if I understood correctly I should collect a few hundred of them in some vector in mctracks_to_aod and only then fill the tables for McCollisions and McParticles with all these events at once.Then clear the vector and start over again until I reach this number of events again. Is this correct?

aalkin:
No need to specifically collect and then fill, the tables are filled row-by-row anyway. But you may want to use `.reserve(N)` with the filling cursor for collisions to make it preallocate the memory. 

Regarding how to fill a certain number, the inputs MCTRAKCS and MCHEADER are supposed to be timeframe-based, do they really contain info for a single collision?

ddobrigk:
I will take a look at the actual code shortly (busy preparing UW talk material), but in fact, we should also have the data frame aggregation be configurable here. It will act as a sort of rate limiting, and probably merging a few events could improve a bit the processing. Does this make sense @aalkin ? What do you think @DanielSamitz ? Thanks a lot for this crucial development!!!

aalkin:
I'm not very familiar with the simulation, so my question is, in fact, exactly about that. Does aggregation happen at some earlier stage so that MCHEADER and MCTRACKS are already aggregated, or are those single for a single simulated collision? In the first case one can combine several TFs into one, in the second case one can combine an arbitrary number of collisions in a TF by using a counter. 

@DanielSamitz as I understand from re-reading your comment, those inputs are indeed for a single collision. Then the solution is simple - you add a counter, you keep filling the table until it reaches a certain number. Then you do the snapshot and reset the filling cursor.

ddobrigk:
Ah I just looked at the file and the number of collisions per data frame (in this case better really call it data frame as opposed to TF :-D ) is already configurable. Thanks a lot, @DanielSamitz , and sorry for the spam!

DanielSamitz:
Hi @aalkin and @ddobrigk ,
what I mean is that if I do

```
void process(aod::McCollisions const& mcCollisions)
  {
    int size = mcCollisions.size();
    LOGP(info, "Number of McCollisions in dataframe: {}", size);
  }
```
and I run the task on an AO2D, I get a few times something like "Number of McCollisions in dataframe: 2861"

If I run with a generator and the `o2-sim-mctracks-proxy` and my `o2-sim-mctracks-to-aod`, I get each time "Number of McCollisions in dataframe: 1".

In the output, e.g. when I write out a tree from the task, there are 200 collisions (or whatever is set via `--collisionsPerTimeframe`) collected together into one DF_xxx directory in the ROOT file, but it seems they are still being sent one by one.

I'm not sure if this is how it is supposed to be. Should the collisions somehow be collected into a few hundred collisions per dataframe already in the generator, the mctracks-proxy or in the mctrakcs-to-aod?

aalkin:
To clarify, the analysis framework is designed to operate on timeframe inputs, so using `Produces` will make the task send out the outputs after each incoming "timeframe" portion that is given to the task by the proxy. The aggregation should happen before the analysis task, so either in the proxy or in converter. I would suggest the proxy, but I suspect there are already pieces of code that depend on it sending collisions one-by-one, so that leaves your mctrakcs-to-aod as the target. You will have to make create and manage both cursor and output manually, however. You do not need any intermediate arrays, you can just fill cursors until the counter reaches a threshold and then send the messages and reset the cursor for the next batch.

DanielSamitz:
Ok, thanks Anton.
@sawenzel can this be done in the mctracks_proxy or should I try to implement it in the mctracks_to_aod converter?

ddobrigk:
Hi @DanielSamitz @aalkin, this would be super cool to have merged even if there might be still some work for some next iterations: if we can play soon, that's nice already. Is there something super significant missing? Also tagging @sawenzel @mconcas as interested parties :-) Super thanks!

sawenzel:
> Hi @DanielSamitz @aalkin, this would be super cool to have merged even if there might be still some work for some next iterations: if we can play soon, that's nice already. Is there something super significant missing? Also tagging @sawenzel @mconcas as interested parties :-) Super thanks!

We should discuss this in some meeting as I don't understand all aspects of this dicussion. For now, I don't understand why this aggregation needs to be done at all. Why can one not define the "dataframe" at the same granularity as a single "MC event"? (I understand that for performance reasons aggregation makes sense of course). I'll have to see if the DPL proxy actually allows to send out stuff at different granularity than it receives things. If this is possible, it should be easy to do. 

Note that orthogonal to this we were anyway thinking of restructuring the proxy to implement load balancing and to make it more hyperloop friendly.

aalkin:
I second the idea to have a meeting about this, I believe this issue requires some architectural decisions to be made. The aggregation, specifically, is an essential performance requirement for analysis - the arrow tables consumed by an analysis task can be processed in bulk, and the analysis framework is designed with this in mind. So it needs to happen somewhere before the analysis tasks in this chain,  and it can be done either in the proxy or in the converter, both options are fine. In my opinion, however, having the aggregation done in the proxy allows us to make the converter considerably simpler and possibly modular. We should definitely discuss this.

PR #11296, opened 2023-05-07
DPL Analysis: filling vector columns with vector or span input

aalkin:
This change permits using both `std::vector` and `gsl::span` as an input for filling variable size array columns in arrow tables with DPL cursor. This can be useful, for example, when a skimmed table is filled, so that the span returned from the original table can be filled directly, without copying its content to a vector first.

@ktf do you have any comments on the implementation? Probably it can be improved.
@pbuehler

### DIFF
--- a/Framework/Core/include/Framework/AnalysisHelpers.h
+++ b/Framework/Core/include/Framework/AnalysisHelpers.h
@@ -78,10 +78,13 @@ struct WritingCursor<soa::Table<PC...>> {
 
  private:
   template <typename T>
-  static decltype(auto) extract(T const& arg)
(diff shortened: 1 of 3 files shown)


jgrosseo:
This is very nice and would also speed up the track propagation for those tracks which do not need to be propagated.

@ddobrigk

ktf:
As discussed offline, I think a better approach would be to return a object which has a statically overloaded `operator()`, one for each fill type that needs to be supported. This would have the advantage that there is no runtime check to perform, typesafety is still enforced and we can specialise the methods to avoid copies in all cases. One could even play smart and have something using uniform_apply_ref to fill directly from C++ structs.

PR #11298, opened 2023-05-08
GPURecoWorkflowSpec inits CorrMapsLoader only if tracks are requested

shahor02:


shahor02:
merging since w/o this O2DPG/MC sctips fail.

PR #11299, opened 2023-05-08
Avoid setting the memory lock when we have no ulimit

chiarazampolli:
@davidrohr , this fixes running the async reco. 
@martenole , if you agree, it would be very useful to have it merged. 

@catalinristea, FYI.

### DIFF
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -36,7 +36,7 @@ source $GEN_TOPO_MYDIR/getCommonArgs.sh || { echo "getCommonArgs.sh failed" 1>&2
 workflow_has_parameter CALIB && { source $O2DPG_ROOT/DATA/common/setenv_calib.sh; [[ $? != 0 ]] && echo "setenv_calib.sh failed" 1>&2 && exit 1; }
 
 [[ -z ${SHM_MANAGER_SHMID:-} ]] && ( [[ $EXTINPUT == 1 ]] || [[ $NUMAGPUIDS != 0 ]] ) && ARGS_ALL+=" --no-cleanup"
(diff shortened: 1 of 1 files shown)


davidrohr:
Fine with me. But we should still disable setting this no_ulimit flag on the epns.

chiarazampolli:
Hello @davidrohr , 
I don't know if this is set on the EPNs, but without this fix, the async reco is broken, both locally and on the grid, so could it be merged if it seems ok?
Thanks!

chiarazampolli:
Hello @martenole , 
Can you merge this? Async reco is broken without it.
Chiara

martenole:
Ok, I will merge it. But I think what David means is that for the async jobs which are running on the EPNs one should not set `SETENV_NO_ULIMIT=1`. I see it is not set anywhere for synchronous processing, so I don't see an issue there. But in the async scripts for example in https://github.com/AliceO2Group/O2DPG/blob/4b76fbbe8e9a6f4acde10b0734436d61d0e880f8/DATA/production/configurations/asyncReco/setenv_extra.sh#L6 that export should be protected by something so that it is not deployed for productions on the EPNs. I don't know how this can be checked though

chiarazampolli:
> Ok, I will merge it. But I think what David means is that for the async jobs which are running on the EPNs one should not set `SETENV_NO_ULIMIT=1`. I see it is not set anywhere for synchronous processing, so I don't see an issue there. But in the async scripts for example in https://github.com/AliceO2Group/O2DPG/blob/4b76fbbe8e9a6f4acde10b0734436d61d0e880f8/DATA/production/configurations/asyncReco/setenv_extra.sh#L6 that export should be protected by something so that it is not deployed for productions on the EPNs. I don't know how this can be checked though

Ah ok, now I understand. I can protect it, so that it is disabled when we use GPUs. 
Do you (or anyone) remember why we had it there? Is it needed at all?

martenole:
I don't know, sorry

PR #11300, opened 2023-05-08
Drop boost usage in staticMapCreator.C

martenole:


ktf:
fullCI passes. Merging.

PR #11305, opened 2023-05-09
TPC ZS Decoder: Fix in-kernel multithreading.

fweig:
Fixes an issue when using in-kernel multithreading with the Dense-Link Decoder.

@davidrohr, I took the liberty to add some more error checking / sanity checks for the ZS Decoding. Let me know if you want any changes there.


davidrohr (approved):
thx for the fix, indeed not so easy to spot !

PR #11307, opened 2023-05-09
[ATO-581] Speed up of createDigitsFromSim()

mhemmer-cern:
- Merged two for loops two increase performance
- `localPadRow` in createCluster is now a const unsigned int as it should be when comparing with `Mapper::ROWSPERREGION[region]`
- remove variables that are only used once directly afterwards
- move constructor of digitContainer outside of loop and add reset call inside loop. The reset call is very expansice but meassures with TStopwatch show that it is still faster.

### DIFF
--- a/Detectors/TPC/simulation/macro/toyCluster.C
+++ b/Detectors/TPC/simulation/macro/toyCluster.C
@@ -44,6 +44,7 @@
 #include "TPCBase/Mapper.h"
 #include "TPCBase/ParameterDetector.h"
 #include "TPCBase/ParameterElectronics.h"
+#include "TPCBase/CDBInterface.h"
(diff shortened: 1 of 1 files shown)


mhemmer-cern:
@matthias-kleiner could you please check if this looks okay to you?

matthias-kleiner:
> @matthias-kleiner could you please check if this looks okay to you?

Hello @mhemmer-cern , this is fine for me

sawenzel:
macro fails CI test

mhemmer-cern:
Yes but this should be unrelated to the changes since this happens to many o2 and O2physics PRs. Cl probably needs to get updated and Timo was already pinged on multiple PRs because of this.

PR #11316, opened 2023-05-10
Send MID calibration data at periodic intervals

dstocco:
The MID calibration data are produced in a dedicated run. So the output was built using the full statistics of the run, computing the calibration data only at the end.
However, this requires the EOS or stop mechanisms to work properly. This was the case, but some recent development seems to have broken the synchronisation with the QC.
The bad channels are correctly generated and stored in the CCDB, but the corresponding QC plot is empty.
Since the treatment of objects produced at EOS is quite sensitive, I decided to generate a list of bad channels when enough statistics is reached (typically at half of the calibration run with the current rate of 1kHz and duration of 2 minutes).
In this way we should be less dependent on modification/optimisation of the EOS processes.

### DIFF
--- a/Detectors/MUON/MID/Calibration/include/MIDCalibration/ChannelCalibrator.h
+++ b/Detectors/MUON/MID/Calibration/include/MIDCalibration/ChannelCalibrator.h
@@ -92,7 +92,7 @@ class ChannelCalibrator final : public o2::calibration::TimeSlotCalibration<Cali
   void setThreshold(double threshold) { mThreshold = threshold; }
 
  private:
-  std::vector<ColumnData> mBadChannels; /// List of bad channels
+  std::vector<ColumnData> mBadChannels{}; /// List of bad channels
   double mThreshold = 0.9;              /// dead channels threshold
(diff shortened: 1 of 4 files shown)


aphecetche (dismissed):

dstocco:
@TimoWilken : the PR formatting is `Expected` since yesterday. Is it a known issue?

martenole:
Happened also with https://github.com/AliceO2Group/AliceO2/pull/11318 and https://github.com/AliceO2Group/AliceO2/pull/11317. Could you please push a dummy commit as Ruben did here: https://github.com/AliceO2Group/AliceO2/commit/e56e6008f064bbbaf848ec96a5ba7335eb94401b which should trigger the checks? 
If you just append the commit to this PR without any squashing or force-pushing I can squash it out while merging and we don't need to go through the full CI again, since all other checks are green now.
Timo it at CHEP currently

dstocco:
Hi @martenole ,
actually I found a small issue in my PR, so I will integrate it with a real commit, not a dummy one

aphecetche (approved):

PR #11317, opened 2023-05-10
Add FOCAL to rawTF reader

mfasDa:
- Add FOCAL to the list of detectors in order to be able to decode FOCAL data
- Remove FOCAL from default detectors in order to prevent setting up FOCAL raw data by default
- Add option for FOCAL DPL output (in case FOCAL raw decoding workflow is running on the FLP)


shahor02 (dismissed):
Looks fine but for some reason the formatting checks were stuck, pushing dummy commit to retrigger

PR #11318, opened 2023-05-10
use tighter MCH tracking limits in sync mode

pillot:
The MCH reconstruction parameters (not only the tracking) have been tuned to optimise (physics) performances for asynchronous reconstruction.
These tighter cuts for the tracking should only be used for synchronous reconstruction now.


pillot:
for some reason the formatting checks did not started, but the rest is green :)

martenole:
@pillot  Not sure why the formatting checks did not run. But this script is anyhow not checked so I can merge it as is.
Just to be sure: As far as I can see you still keep the same tracking limits in sync mode which you had before. But with this PR the limits will not be applied in async. Was this your intention?

pillot:
> @pillot Not sure why the formatting checks did not run. But this script is anyhow not checked so I can merge it as is.
> Just to be sure: As far as I can see you still keep the same tracking limits in sync mode which you had before. But with this PR the limits will not be applied in async. Was this your intention?

Hi @martenole , yes, this is the intention.

martenole:
Ok, thanks for confirming. For me "use tighter MCH tracking limits **only** in sync mode" would have been clearer then :)

PR #11320, opened 2023-05-11
[WIP] Use shm-mlock-segment-on-creation only if in EPNSYNCMODE

chiarazampolli:
@martenole , as discussed (it worked).
@davidrohr : without this, the async reco is broken on the EPNs...


davidrohr:
How does the async reco fail on the EPNs? EPNs should be configured to allow the mlocking, and it should be used. We can merge this temporarily as a workaround and then fix and revert it, but this is certainly not the correct final solution.

Kind Regards
David Rohr

chiarazampolli:
Don't merge, yet. The problem might be something else.

chiarazampolli:
The issue is something else, closing

PR #11321, opened 2023-05-12
[WIP] Tpc idc enable pad status map writeout

rmunzer:


davidrohr:
@rmunzer : what is the status of this? Do we want to merge it? If not, we should mark it as WIP so it does not get merged accidentally.

PR #11322, opened 2023-05-12
[FOCAL-10] Allow event writer to run without dedicated subspecification

mfasDa:
- No-subspec mode needed i.e. to subscribe to data
  from raw-tf-reader-workflow
- Option to require dedicated subspec should be
  preserved for i.e. future recalibration


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/11317

I would first like to get the other PR merged, then I will rebase this one and the PR should be trivial.

PR #11325, opened 2023-05-12
TPC: Improve SAC decoding

wiechula:
* Parallel decoding option
* Add missing initialization
* Add option to re-align data stream in case packets are missing
* Add option to inject dummy data in case packets are missing
* Add option to accumulate data over several TFs before decoding and sending


wiechula:
Can this be merged?

PR #11326, opened 2023-05-12
Revert "Remove TPC starting row cut"

chiarazampolli:
This commit has a negative impact on the ITSTPC matching efficiency, increasing it at low pT (from fakes?) and decreasing it at high pT.

This reverts commit d1be5112b790985f502185840b25a8085d8c59af.

@fmazzasc , @ddobrigk , @noferini , @wiechula , @aschmah

Attached some plots, showing the effect of the PR that we would like to revert. We would like to recover the behaviour of the tag on 4 May.

image

image

image

image

### DIFF
--- a/Detectors/GlobalTracking/include/GlobalTracking/MatchTPCITSParams.h
+++ b/Detectors/GlobalTracking/include/GlobalTracking/MatchTPCITSParams.h
@@ -40,6 +40,7 @@ struct MatchTPCITSParams : public o2::conf::ConfigurableParamHelper<MatchTPCITSP
   float minTPCTrackR = 50.; ///< cut on minimal TPC tracks radius to consider for matching, 666*pt_gev*B_kgaus/5
   float minITSTrackR = 50.; ///< cut on minimal ITS tracks radius to consider for matching, 666*pt_gev*B_kgaus/5
   int minTPCClusters = 25; ///< minimum number of clusters to consider
+  int askMinTPCRow = 15;   ///< disregard tracks starting above this row
 
   float cutMatchingChi2 = 30.f; ///< cut on matching chi2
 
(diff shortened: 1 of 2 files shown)


ddobrigk:
Hi @chiarazampolli , we had explicitly checked that the fake rate was under control, but it is true that we didn't check for primaries. I think we could maybe do that study as well. In any case, @shahor02 also mentioned this cut was there almost accidentally, which is why @fmazzasc did a PR with it... I am also unsure how it can be that the matching efficiency *decreases* with the removal of a selection. I will try to run some further checks from my side...

shahor02:
Hi,

I don't remember why I had this cut originally, certainly not to improve high pt matching. 
The only reason I can imagine why this happens is that this increases the amount of low-pt candidates which increases the number of fakes, this stealing candidates for proper high pt match. Bit i will not be able to check anything before Tuesday.

noferini:
Dear all,
since the cut is a configurable it is better in any case to keep it (especially in the tag we will use for apass4).
I would revert the commit as soon as possible in order to freeze the reco tag

ddobrigk:
Just to say here too (as in the mail exchange): yes let's revert asap since it will still be configurable after the reversal. Thanks and have a nice weekend!

chiarazampolli:
> Dear all, since the cut is a configurable it is better in any case to keep it (especially in the tag we will use for apass4). I would revert the commit as soon as possible in order to freeze the reco tag

I fully agree with Francesco, if we could merge this as soon as possible, it would be good.
But then, of course, we need to decide what value to use for such cut. If it was there accidentally, it seems indeed that it was cleaning fakes at low pT, allowing more matches at high pT - this is what we concluded, as also Ruben commented.
For sure, more studies would be needed, but I am afraid we have no time for apass4. So we just need to decide what value we want.

martenole (approved):

PR #11327, opened 2023-05-12
Add current integration workflows for FIT and TOF

martenole:
goes together with https://github.com/AliceO2Group/O2DPG/pull/1068

### DIFF
--- a/prodtests/full-system-test/calib-workflow.sh
+++ b/prodtests/full-system-test/calib-workflow.sh
@@ -16,6 +16,12 @@ if workflow_has_parameters CALIB_LOCAL_INTEGRATED_AGGREGATOR CALIB_PROXIES; then
   exit 2
 fi
 
+if [[ -z ${TPC_IDCCALIB_NSLICESPERTF:-} ]]; then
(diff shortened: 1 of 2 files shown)


prodtests/full-system-test/calib-workflow.sh
```diff
@@ -16,6 +16,12 @@ if workflow_has_parameters CALIB_LOCAL_INTEGRATED_AGGREGATOR CALIB_PROXIES; then
   exit 2
 fi
 
+if [[ -z ${NSLICESPERTF:-} ]]; then
+  NSLICESPERTF=$(echo "scale=2;$NHBPERTF/128*11" | bc)
```
davidrohr: could you make this variable name more specific? like TPC_IDCCALIB_SLICESPERTF?

martenole: sure

martenole:
If there are no more comments I would merge this as is.
By default the FV0 and FDD reco workflows are not run in the synch phase. They have to be added explicitly as extra processing steps in order to have the calibration workflow running.

PR #11328, opened 2023-05-15
DPL Analysis: avoid iterating over value cache if the argument is larger than the largest value in the cache

aalkin:


Framework/Core/src/ArrowTableSlicingCache.cxx
```diff
@@ -25,6 +25,14 @@ std::pair<int64_t, int64_t> SliceInfoPtr::getSliceFor(int value) const
   if (values.empty()) {
     return {offset, 0};
   }
+  auto p = values.size() - 1;
+  while (values[p] < 0) {
```
ktf, line 30:
```suggestion
  while (values[p] < 0) {
    assert(p > 0);
```

Framework/Core/src/ArrowTableSlicingCache.cxx
```diff
@@ -25,6 +25,14 @@ std::pair<int64_t, int64_t> SliceInfoPtr::getSliceFor(int value) const
   if (values.empty()) {
     return {offset, 0};
   }
+  auto p = values.size() - 1;
```
ktf, line 29:
```suggestion
  assert(!values.empty());
  auto p = values.size() - 1;
```

ktf:
@aalkin thanks. I would add a couple of assert just to make sure your assumptions about the vectors not being empty are always satisfied.

Framework/Core/src/ArrowTableSlicingCache.cxx
aalkin: All-negative values of indices would indicate a malformed table. I can add a check for that, but it should be on cache creation, rather then here.

Framework/Core/src/ArrowTableSlicingCache.cxx
aalkin: Non-empty table would never result in an empty values array. This is checked at the stage of cache creation.

Framework/Core/src/ArrowTableSlicingCache.cxx
ktf: The assert is there exactly to indicate preconditions which the reader might or might not know about. I think it's good practice to have those kind of asserts, so that one does not have to keep into account the inner details in order to review the PR.

PR #11329, opened 2023-05-15
DPL: fix the cpu_usage_fraction metric

ktf:
We need to add up time, we cannot consider the absolute value like we do for the transmitted bytes.

Notice also that the value is in general between 0 and 1000 but it can be > 1000 due to rounding errors (i.e. some processing time from the previous interval assigned to the current one).

### DIFF
--- a/Framework/Core/src/DataProcessingDevice.cxx
+++ b/Framework/Core/src/DataProcessingDevice.cxx
@@ -2037,7 +2037,8 @@ bool DataProcessingDevice::tryDispatchComputation(ServiceRegistryRef ref, std::v
     states.updateState({.id = short((int)ProcessingStateId::DATA_RELAYER_BASE + action.slot.index), (int)(record.size() + buffer - relayerSlotState), relayerSlotState});
     uint64_t tEnd = uv_hrtime();
     stats.updateStats({(int)ProcessingStatsId::LAST_ELAPSED_TIME_MS, DataProcessingStats::Op::Set, (int64_t)(tEnd - tStart)});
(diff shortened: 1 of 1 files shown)


ktf:
@awegrzyn as promised. I was simply not summing up things correctly. This fixes it, I tested it with a simple example and I get the expected numbers.

ktf:
Trivial and tested locally. Merging.

davidrohr:
Just for my undestanding: Why do we have this assignment to the next (apparently wrong) time interval. Can't we fix that?

ktf:
Because I need to calculate (tEnd - tStart) and tEnd is obtained only when the computation ends so you might end up sending a metric before the last interval is finished, especially when multithreaded.

I could indeed split an interval whenever I publish metrics, but I need to check how to do that, because the accounting might not be trivial. I will check what it takes.

davidrohr:
So this means the intervals are asynchronous to the processing? I.e. if one processing overlaps  the interval border, tEnd doesn't exist yet when metrics for the previous intervals are sent, so the full (tEnd - tStart) is assigned to the first interval?

In that case splitting of the time would be trivial. You just take tEnd = now() and reset tStart = now().
But splitting for other quantities like data rate would not be so trivial.
For multithreading, you could do the same thing, or sum up the metrics synchronously per thread (that would also avoid the splitting), store and update them temporarily during the interval, and then merge the values from the threads together and send them at the end of the interval.

ktf:
Data rate should actually be easy / already correct, because transferred bytes are a monotonically increasing quantity, so you just sample it at a given point and do the calculations wrt the previous given point. In this case we need to sum up smaller contributions to a given interval which requires splitting all the open intervals and summing them up. Which is for sure doable, but requires a bit more than just a one liner. Notice this is needed regardless of accounting them per thread or not, because the average is over two publishing samples.

PR #11330, opened 2023-05-15
CPV: clusterize digits with common side

sevdokim:
Hello! In this commit the clustering algorithm is changed from clustering digits with common vertex to clustering digits with common side. It was checked on real pp data and showed approximately same performance and resources consumption. However It is important for PbPb data where detector occupancy is higher and probability of cluster overlap is greater: no need to merge several neighboring clusters in one.


peressounko (approved):

martenole:
Looks good to me and I can merge it as is. But since you commented out the common vertex option instead of replacing it - would it make sense to have this configurable via either an option to the DPL device or a configKeyValue?

peressounko:
This is rather fundamental change in reconstruction, I do not think we will play with it, so not need to make switch.

martenole:
Ok, thanks. Then I'll merge it as is

PR #11332, opened 2023-05-16
DPL: make sending policy per channel

ktf:
This avoids having expendable / non-critical devices blocking the data flow since any device can now drop (non forwarded) data when sending to one.

Notice that this means that non-critical devices must keep into account they might not see as much data as expected.

### DIFF
--- a/Framework/Core/include/Framework/ChannelInfo.h
+++ b/Framework/Core/include/Framework/ChannelInfo.h
@@ -66,11 +66,14 @@ struct InputChannelInfo {
   int pollerIndex = -1;
 };
 
+struct SendingPolicy;
+
 /// Output channel information
 struct OutputChannelInfo {
   std::string name = "invalid";
   ChannelAccountingType channelType = ChannelAccountingType::DPL;
   fair::mq::Channel& channel;
+  SendingPolicy const* policy;
 };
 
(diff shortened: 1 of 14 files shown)


ktf:
@Barthelemy @knopers8 @davidrohr as discussed. Notice the last comment in the commit. A non-critical device must take into account the fact that its inputs might or might not come, depending on how fast it was at accepting the incoming data.

davidrohr:
I don't understand what you mean by "Notice that this means that non-critical devices must keep into account they might not see as much data as expected."?
If I remember correctly, we said in the discussion (minutes in https://alice.its.cern.ch/jira/browse/QC-953?filter=-2) that all outputs of expendable tasks are overwritten to be sporadic, and all devices that consume this data must have their input configured as sporadic, otherwise the topology generation should fail.

In this case it should just come naturally. By definition the tasks must properly handle that a sporadic input is not present.
Or do you refer to something else? I just want to avoid that there is another implicit requirement for the consumers.

ktf:
no, indeed. if we switch the output to be sporadic it will be ok. i need to check if the policies are matched early enough.

ktf:
Merging this so that we can go forward. I will ad the check in a later PR.

PR #11333, opened 2023-05-16
[EMCAL-565, EMCAL-566]: Switch to BasicCCDBManager in EMCal calibDB

jokonig:
- Switch to BasicCCDBManager in order to chache objects as ccdb api does not support caching


mfasDa (approved):
Looks good

shahor02 (approved):

PR #11334, opened 2023-05-16
DPL GUI: fix referencing object on stack

ktf:
Without this we have a copy of the timestamps, which is both stupid and wrong. The reference brings back sanity.


ktf:
Tested locally and obviously wrong. Merging.

PR #11337, opened 2023-05-17
DPL Analysis: method to return only filtered iterators from array index

aalkin:
In many cases when getting an array of rows using index from a filtered table, only those rows that pass the filter are required. The default `<getter>_as<soa::Filtered<T>>` returns all iterators, while newly added `filtered_<getter>_as<soa::Filtered<T>>` will return only those that also pass the filters, that are applied to `T`.


jgrosseo:
That is quite cool!

Can we already write out the same table which we had in the input or is that impossible with the current topology building?

aalkin:
I'm not sure I understand the question. The table is unchanged, this getter just skips the rows that are not present in the list of selected rows, when creating a vector of iterators.

PR #11341, opened 2023-05-17
DPL GUI: fix axis range

ktf:


ktf:
Trivial and tested with the GUI locally.

PR #11348, opened 2023-05-18
Add macro to check TRD noise runs

tdietel:
This PR adds an updated version of the macro that I used for the presentation in the TRD weekly meeting on 17 April to look at noise runs.

Only a macro is added, and no compiled code is affected.

The macro is installed in share/Detectors/TRD/macro/ to make it available on in central O2 installations on lxplus without downloading the source code.


martenole:
Thanks for the macro @tdietel

PR #11351, opened 2023-05-19
Open DCA cut in SVertexer for He3 and He4

fmazzasc:


fmazzasc:
Hi @shahor02 , this P.R. allows heavily ionising particles to pass the DCA to PV cut in the SVertexer. This increases by a factor ~ 2 the reconstruction efficiency of low Q-value hypernuclei (see plot).
The ideal solution would be to have PID in tracking but for the time being a rectangular cut on the TPC dEdx looks to me the best workaround.
- One drawback is that in this way we have to load in any case the TPC tracks: I added a flag to exclude them from the combinatorial part. 
- I tested the P.R. on one ctf (/alice/data/2022/LHC22m/523308/apass3_relval_cpu2_debug/0040): the number of V0s found is the same and the time increase is negligible (91 vs 90 seconds)

What do you think? 

hyp_eff

Detectors/GlobalTrackingWorkflow/src/secondary-vertexing-workflow.cxx
```diff
@@ -79,13 +79,18 @@ WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
   auto requireCTPLumi = configcontext.options().get<bool>("require-ctp-lumi");
 
   GID::mask_t src = allowedSources & GID::getSourcesMask(configcontext.options().get<std::string>("vertexing-sources"));
+
+  // don't loop over TPC tracks if not needed, but load them anyway for dEdx evaluation
+  bool excludeTPCtracks = !(src & GID::getSourceMask(GID::TPC)).any();
```
shahor02: for the record: it works also with `src[GID::TPC]`

Detectors/GlobalTrackingWorkflow/src/secondary-vertexing-workflow.cxx
```diff
@@ -79,13 +79,18 @@ WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
...
+  src = src | GID::getSourceMask(GID::TPC);
```
shahor02: Unconditional request of the TPC input will lead to a crash if for some reason the TPC is not in the run. 
I think it is better to make the `excludeTPCtracks` a data member of the SVertexerParams while keeping the TPC DetID request as it is now. Then, if the TPC is requested, it will be used for your trick and  `excludeTPCtracks` will steer the combinatorics. 
Once we will have the TPC PID assignment propagated via global tracks, no change in the DetID src will be needed.

and in case it is f

Detectors/Vertexing/src/SVertexer.cxx
```diff
@@ -343,7 +350,19 @@ void SVertexer::buildT2V(const o2::globaltracking::RecoContainer& recoData) // a
...
+      if (tpcGID.isIndexSet()) {
+        auto& tpcTrack = recoData.getTPCTrack(tpcGID);
+        float dEdxTPC = tpcTrack.getdEdx().dEdxTotTPC;
+        if (dEdxTPC > mSVParams->minTPCdEdx && trc.getP() > mSVParams->minMomTPCdEdx) // accept high dEdx tracks (He3, He4)
+        {
+          ionisingTrack = true;
+        }
+      }
+
+      if (!acceptTrack(tvid, trc) && !ionisingTrack) {
```
shahor02: Instead of disabling the dca cut, should not we in general pass the info about PID to the acceptTrack so that it is accounted in the (propagation for the ) DCA calculation?

shahor02 (commented):
Hi @fmazzasc 

In general looks good, but please see a few comments below. 
We should discuss with TPC people if they can suggest a reco-time PID at least for the particles with large dEdX.

Detectors/Vertexing/src/SVertexer.cxx
fmazzasc: The issue is not due to the wrong PID employed for computing the DCA; it is more the fact that He3 and He4 daughters point into the direction of the mother as they inherit most of the initial momentum. For these particles the DCA cut should be removed/released

shahor02: OK, understood, looks fine for me, I think you can remove the draft tag to trigger CI.

fmazzasc: Done! Thanks

fmazzasc:
Hi @shahor02 , could we merge this?

PR #11352, opened 2023-05-20
DPL Analysis: add support for fixed size int8_t array and bitmap helper

aalkin:
@ekryshen @ddobrigk


ktf:
New feature, merging.

ekryshen:
@aalkin Thanks a lot! Just one question on the interface. As far as I can see, users are supposed to access bit values in bitmask using row.boolArray_bit(i)
So I guess we would have to change the interface to evsel alias and selection arrays in all user tasks anyway. One way to avoid it is to implement row.boolArray()[i] interface if doesn't take too much effort. What do you think?

aalkin:
@ekryshen adding `operator[]` is not very straightforward, unfortunately. Also it is better, in my opinion, to have a clear distinction between interfaces for arrays and bitmaps.

ekryshen:
@aalkin Ok, no problem. I guess sed replacement will do the job.

PR #11354, opened 2023-05-21
TPC IDCs: Adding function to get number of outliers for QC

matthias-kleiner:
- do not overwrite previous status flags


wiechula (approved):
Can be merged from my side.

PR #11360, opened 2023-05-22
MFT assessment must get B-field from CCDB instead of local file

shahor02:


robincaron13 (approved):
Thanks to open it, fine from my side.

PR #11361, opened 2023-05-22
ZDC: Prevent overflow/segfault problems due to very slow particles

sawenzel:
Simulations with slow neutrons lead to segfaults in ZDC digitizer. This is fixing the issue by skipping over such very slow hits.


sawenzel:
@davidrohr

davidrohr (approved):
Could ZDC experts comment?
Can we merge this?

PR #11363, opened 2023-05-22
temporary trick to circumvent EMC memory corruption

shahor02:
@mpoghos


shahor02:
does not work, superseded by https://github.com/AliceO2Group/AliceO2/pull/11364

PR #11365, opened 2023-05-23
DPL: use sporadic for expendable devices output

ktf:


ktf:
@Barthelemy do you have expendable devices already in the tests?

ktf:
So the error must be unrelated as it happened in another PR as well. I think there is either something fishy on the machine (@TimoWilken could you please have a look?) or something in the loading of the libraries can eventually slow down processing.

Barthelemy:
@ktf No I don't but I have a branch to play with it.

ktf:
Ok, as discussed the testActivity test which was failing was indeed due to cold cache, bloated dependencies and slow CI VM.

PR #11366, opened 2023-05-23
Cutoff very slow particles in TOF digitization

sawenzel:


sawenzel:
@davidrohr

noferini:
Hi @sawenzel,
the current condition cut on the time if the hit goes in the next readout window (ROW length = 30 mus).
If we are close to the border we can cut too much.
In general I would avoid to apply a cut which depends on the position of the collision within the ROW.
I would suggest simply to cut on the hit time if it exceeds a given threshold so that we know a priori what is done (only based on track kinematics and anything else).

if(  hit.GetTime() > 1000) { // 1 mus?
  continue;
}

sawenzel:
Yes, totally fine for me. I'll change it accordingly.

davidrohr (dismissed):
works for me.
I'd propose to fix the formatting and then merge it.

PR #11375, opened 2023-05-23
[O2-1787] Add side bar in the Cage side panels. Fix small overlap with services

mario6829:
Two side bars belonging to the rails inside the lateral panels of the ITS Cage were added: this should improve the material budget description with respect to real data. These new elements were inserted in the V3Cage class.
At the same time a small overlap between the panels and the rough elements describing the ITS services (passed unnoticed in the first implementation) was fixed in the V3Services class.


mario6829:
Hi @mconcas @iouribelikov this PR includes the implementation of side bars in the Cage panels as suggested by @amorsch . Please let me know in case of problems. Cheers

mconcas (approved):

PR #11379, opened 2023-05-24
Adapting MFT assessment histograms ranges and replacing one variable

robincaron13:
Adapting the histogram ranges for histograms and replacing tanl variable to cotl (cotangent lambda) only at both MFT and GlobalFwd assessment workflow level (to be decided later if we apply this to track parameters)


robincaron13:
Hi @shahor02, these are trivial changes, already tested.

shahor02 (approved):

PR #11384, opened 2023-05-24
By default avoid MD5 calc. for epn2eos metadata

shahor02:
As it takes > 2 s/GB, becoming too slow for CTF file. The MD5 can be still calculated by reguesting it explicitly  FileMetaData::fillFileData(const std::string& fname, bool md5)


shahor02:
Don't merge, waiting for a green light from epn2eos team

PR #11385, opened 2023-05-25
ZDC - Fix bug in data loss reporting

cortesep:


cortesep:
This is a small fix in error reporting code and introduction of a missing check. The affected program is usually not run in production. However the same functionality has to be corrected in QC code. It will be propagated soon.

PR #11391, opened 2023-05-25
AOD-Calo: Crash avoidance and error diagnostic

sawenzel:
MC error diagnostic during CaloTable filling
    
    This commit is related to https://alice.its.cern.ch/jira/browse/O2-3799
    
    During filling of the Calo table it:
    
    a) checks if a trackID is stored in mToStore
    b) if not, emits diagnostic information about that track
       for further analysis of the problem
    
    In result, O2-3799 is no longer critical since it shouldn't crash anymore.
    However, the true origin of the problem still needs to be addressed.


sawenzel:
@mhemmer-cern @nburmaso @benedikt-voelkel

PR #11393, opened 2023-05-25
Fix bogus wrong bunches diff message

martenole:
ping @shahor02 
For a proper warning we would need to know the number of HBFs per TF. For this I think we need to pass this as parameter from every device which is using this helper and these devices in turn would need to fetch that info from CCDB if they don't do it anyhow already. But I am not sure if this would be important to have actually?


martenole:
Closing, as this was done like that on purpose. We compare to the start IR, to the difference should never be negative. The cast to unsigned gives a very large number for negative values and this then triggers this alarm

PR #11395, opened 2023-05-26
[EMCAL-889] AOD-Calo: Avoidance of crash during emcCorrectionTask

mhemmer-cern:
- The fix provided by @sawenzel in PR #11391 results in `particleIds` and `amplitudeFraction` being of unequal lenght which would lead to crashes in the emcCorrectionTask. So for now just fill -1 as particleId to the `particleIds` to make sure the two vectors have equal lenght and we can check for negative indices in the emcCorrectionTask if we need to.

### DIFF
--- a/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
+++ b/Detectors/AOD/src/AODProducerWorkflowSpec.cxx
@@ -1408,6 +1408,7 @@ void AODProducerWorkflowDPL::addToCaloTable(const TCaloHandler& caloHandler, con
               if (iter != trackStore->end()) {
                 particleIds.emplace_back(iter->second);
               } else {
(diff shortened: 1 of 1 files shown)


mhemmer-cern:
Tagging @mfasDa @nburmaso @benedikt-voelkel @sawenzel

sawenzel:
Fine for me ... but wouldn't it be easier to also just not fill amplitudes when the track is invalid?
This would avoid having to check for `-1` later.

mhemmer-cern:
I thought the amplitudes are valid in that case, just the particleID is not saved, but the energy of the mc particle should still be known and valid when filling the table. So we would only lose the reference for the particle, but at least still have the information of the energy fraction for all mc particles in a given cell.

PR #11396, opened 2023-05-26
Reserve memory for tables before filling

nburmaso:
In continuation of #11278

### DIFF
--- a/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
+++ b/Detectors/AOD/include/AODProducerWorkflow/AODProducerWorkflowSpec.h
@@ -52,6 +52,23 @@ using DataRequest = o2::globaltracking::DataRequest;
 namespace o2::aodproducer
 {
 
+template <typename T>
+void addTableToOutput(std::vector<OutputSpec>& o, const std::string& ver = "", header::DataHeader::SubSpecificationType t = 0)
+{
+  std::string d = o2::aod::MetadataTrait<T>::metadata::description();
+  if (!ver.empty()) {
+    d += "_";
+    d += ver;
+  }
+  char chDesc[16];
+  std::strcpy(chDesc, d.c_str());
+  o.emplace_back(OutputLabel{o2::aod::MetadataTrait<T>::metadata::tableLabel()},
+                 o2::aod::MetadataTrait<T>::metadata::origin(),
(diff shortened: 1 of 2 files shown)


nburmaso:
Dear @ddobrigk, @ktf, @pzhristov, @sawenzel,
Could you please check if the changes are fine with you?

ktf:
Hi, without actually trying the PR, your changes look correct to me.

sawenzel:
@nburmaso : Have you tested this in an actual (MC) workflow?

nburmaso:
Hi @sandro, yes I did test this modified version with the sim_challenge.sh MC. Should I test it with one of the DPG production scripts as well?

nburmaso:
Fixed some errors in tables versions.

It seems that it is not possible to use predefined table descriptions from the datamodel (as Giulio proposed in #11278) 

e.g. [for McCollisionLabels the description is MCCOLLISLABEL](https://github.com/AliceO2Group/AliceO2/blob/bf993cac605003b3ccb5563bfe6a93173be6d88a/Framework/Core/include/Framework/AnalysisDataModel.h#L1461), so the name in AO2D file would be O2mccollislabel, not O2mccollisionlabel. For the moment this table is added to outputs like this `outputs.emplace_back(OutputLabel{"McCollisionLabels"}, "AOD", "MCCOLLISIONLABEL", 0, Lifetime::Timeframe)`

Nazar

ktf:
Interesting, but then again maybe the correct thing to do is to change the name of the branch to match the description? Anyways, this we should discuss separately. if you are happy with the result of this PR, I do not thing we should delay further.

ddobrigk:
Thanks a lot for your work, @nburmaso ! 

So, if I remember well, the issue of the description being `[MCCOLLISLABEL](https://github.com/AliceO2Group/AliceO2/blob/bf993cac605003b3ccb5563bfe6a93173be6d88a/Framework/Core/include/Framework/AnalysisDataModel.h#L1461)` is that the description has a maximal number of characters which is easily exceeded by typical table names such as McCollisionLabel - and likely the description was shortened because of that. This happened to me already a few times, so I surmise that's the reason for the awkward "MCCOLLISLABEL"... In any case, I think we should not have that stop the merging of a PR that is about calling `.reserve` but we should rather discuss this tuning separately - I agree with @ktf on that!

PR #11398, opened 2023-05-26
DPL: fix data relayer state display

ktf:


ktf:
error unrelated. verified visually. merging.

PR #11404, opened 2023-05-30
[FOCAL-10] Remove temporary mapper class

mfasDa:


mfasDa:
Probably needs to wait until the new QualityControl is tagged and deployed

PR #11406, opened 2023-05-31
Fix new fairmq

ktf:


ktf:
Minimal set of changes, it will produce tons of deprecation warnings due to the change between fChannels and GetChannels(). Hopefully no platform will treat those as fatal.

@dennisklein @rbx Thank you for making sure the old API is maintained in the new release. As a cherry on the cake, it would be good if the introduction of new APIs and the deprecation of older ones happens in two subsequent releases, so that we do not get spammed by warnings.

dennisklein:
> As a cherry on the cake, it would be good if the introduction of new APIs and the deprecation of older ones happens in two subsequent releases, so that we do not get spammed by warnings.

So, I just do two releases on every release, the first without deprecation warnings and the second with? Just, that you don't have to fix your downstream workflow? I am really unhappy about this request :( Can you give an example of any other projects that pre-announces every single deprecation in an extra release?

davidrohr:
If it is only a deprecation warning, can't we bump fmq and disable the warnings.
Then we update O2 
Then we update the fmq recipe again and reenable the warning?

Although admittedly, deprecating something for which you just now provide the alternative is a bit fast :)

dennisklein:
> If it is only a deprecation warning, can't we bump fmq and disable the warnings.

Since disabling all deprecation warnings may be too coarse, what about I provide you with additional flags in upstream, so you can more selectively steer the deprecation warnings?

> is a bit fast

Can you define "fast". As I said, would it help, if I just do two releases 5 min apart with the only difference having one which pre-announces the deprecation? Or do you mean releases should happen with a certain amount of wall time in between?

ktf:
Everyone ;-) For example Apple introduced Metal way earlier than deprecating OpenGL, not too mention other projects like ROOT which basically does not deprecate APIs for a long time (ever), despite being de-facto replaced by something new.

I personally think it is a good practice to separate addition of features / API from the deprecation of old ones and then again from the actual removal of the features (thank you again since you did it this time).

We do not need to be to strict on this, and when the migration is trivially possible in a backward compatible way (which is not for the fChannels -> GetChannels() one) we should not bother. However the net result right now is that for the next couple of weeks or so, until we are fully sure that 1.5.0 is ok and we do not need to quickly rollback to 1.4.56, everyone will be swamped by deprecation warnings / errors (if -Werror is used).

dennisklein:
Proposal:

When a release adds deprecation warnings (I don't care about the `-Werror` because tbh that is your own choice and consequences to bear if you redefine what a warning is...) and a flag which let's you "acknowledge" it by effectively disabling them. This switch could be one or more macros - depending on the changes - per release, e.g. `FAIRMQ_SILENCE_DEPRECATIONS_1_5_0`. Would this help?

ktf:
> If it is only a deprecation warning, can't we bump fmq and disable the warnings.
> Then we update O2
> Then we update the fmq recipe again and reenable the warning?

Yes, everything can be done, I even have 50 lines of templated code which allows me to call GetChannels() or fChannels depending on which one is present. The point is that deprecating things **immediately** (not fast) is adding nuisance / extra work to everyone you depend on for very little benefit (we lived for years with fChannels not being private, we can live one more month).

ktf:
> Proposal:
> When a release adds deprecation warnings (I don't care about the `-Werror` because tbh that is your own choice and consequences to bear if you redefine what a warning is...) and a flag which let's you "acknowledge" it by effectively disabling them. This switch could be one or more macros - depending on the changes - per release, e.g. `FAIRMQ_SILENCE_DEPRECATIONS_1_5_0`. Would this help?

"There's no problem in Computer Science that can't be solved by adding another layer of abstraction to it, except for the too many layers of abstractions one". As I said, one could even fix this with 50 lines of template metaprogramming, the problem is keeping things simple and with as less nuisances for the downstream users (whose compiler options, BTW, you do not control). Anyways, this time we will survive, for next time, I would greatly appreciate the staggering of the deprecation warnings (just like I appreciated heartily the fact that the removal of the old API does not happen immediately).

dennisklein:
But can we please clarify the "timing" topic as well, if we are at it.

It is very likely, there will be a FairMQ v1.6 release next week. The diff will be a new feature for @knopers8 and the actual removal of APIs deprecated in 1.5 or earlier. Nothing else, small release. In your perception 1.5 and 1.6 would then be very quickly followed upon each other, because you just today started to adopt 1.5 even though it was released way before (Given that @knopers8 wants to use the new feature soon which creates pressure to adopt 1.6 also soonish, right?). And then, even another scenario, let's say 1.6 would have been released already last week. But because you were still on 1.4, would you then today really adopt 1.5 first before you move to 1.6? No, you would jump to the latest and the whole pre-announcing/staggering of deprecations would go out of the window anyways? So, what do you expect from 1.6 next week? No removal of recently deprecated APIs yet. If yes, please define "recently" months? number of releases?

ktf:
The use case is what I illustrated above: I have a release which is validated in production and one which is not (regardless of what they are called). I need to be able to support both so that I can quickly rollback. The moment I get told by run coordination the new release is ok and there are no regressions I move to the new APIs and if there are no troubles, you can deprecated the old API. This usually happens within 1 / 2 weeks from the adoption of the new release, unless there are issues. If you do two releases one after the other and you deprecate and remove APIs within two weeks, please hold the deprecation / removal of old APIs or at least make sure the change can be simply detected with `#if __has_include` or similar tricks.

dennisklein:
I wrote something down as a conclusion of this discussion. Feel free to comment, we can always refine the wording, add missing aspects, etc: https://github.com/FairRootGroup/FairMQ/discussions/474.

ktf:
I commented there. Thank you.

PR #11408, opened 2023-05-31
Allow setting material correction type for track interpolation

martenole:
The TPC track interpolation was actually never using material correction. Thanks to @aschmah for spotting


martenole:
Failure unrelated, merging

PR #11409, opened 2023-05-31
Add Chi_c1(3872) in the PDG database

vkucera:


ddobrigk:
Many thanks, @vkucera ! I think this looks very reasonable. Are these three particles all that is currently being used as new or revised in terms of settings from the HF side? If so, then we can hopefully move on soon and then do the second step of removing the `TDatabasePDG` from `RecoDecay.h` (more info [here](https://alice.its.cern.ch/jira/browse/O2-3782))

vkucera:
@sawenzel @benedikt-voelkel Ready for merge.

benedikt-voelkel (approved):

PR #11410, opened 2023-05-31
discard covariances between V0 position and momentum DOFs

shahor02:


shahor02:
will be superseded by other PR

PR #11412, opened 2023-06-01
[MFT,MUON] Fix tracker and fwd matcher workflow for zero B field

robincaron13:
Hi @shahor02, this is the fix needed here https://alice.its.cern.ch/jira/browse/O2-3759 for running the global-fwd-matching workflow on zero B field data.

### DIFF
--- a/Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx
+++ b/Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx
@@ -263,9 +263,9 @@ void TrackerDPL::run(ProcessingContext& pc)
     }
 
   } else {
-    LOG(info) << "Field is off! ";
+    LOG(debug) << "Field is off! ";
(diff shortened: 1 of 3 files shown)


Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx
```diff
@@ -263,24 +263,34 @@ void TrackerDPL::run(ProcessingContext& pc)
     }
 
   } else {
+    LOG(info) << "Field is off! ";
     std::vector<std::vector<o2::mft::ROframe<TrackLTFL>>> roFrameVec(mNThreads); // One vector of ROFrames per thread
+    LOG(info) << "Reserving ROFs ";
```
shahor02: could you change these logs to debug level?

shahor02 (commented):
looks fine, could you just remove excessive logs, see below.

Detectors/ITSMFT/MFT/workflow/src/TrackerSpec.cxx
robincaron13: done, thanks.

shahor02 (approved):

PR #11417, opened 2023-06-01
[MCH] add plots + print statistics + compare if ROFs are different

pillot:


Detectors/MUON/MCH/Evaluation/src/CompareTask.h
```diff
@@ -39,21 +39,24 @@ class CompareTask
...
             const std::list<ExtendedTrack>& tracks1,
             const std::list<ExtendedTrack>& tracks2);
 
-  std::list<ExtendedTrack> getExtendedTracks(const ROFRecord& rof,
-                                             gsl::span<const TrackMCH> tfTracks,
-                                             gsl::span<const Cluster> tfClusters);
+  void getExtendedTracks(const ROFRecord& rof,
+                         gsl::span<const TrackMCH> tfTracks,
+                         gsl::span<const Cluster> tfClusters,
+                         std::list<ExtendedTrack>& tracks);
```
aphecetche: I'm not so sure to understand why you changed the signature of this function ? Generally I much prefer a `getXXX` function to actually return what it is supposed to "get" instead of using a function parameter for that...

pillot: To avoid copying the list of tracks from one function to the other then into the new list created outside. I am not sure the compiler elide them, in particular the last copy.

aphecetche: well, I could be wrong, but I doubt the memory management of that particular list is the bottleneck of the comparison code :wink: (and if performance was an issue, we would also/instead have to challenge the choice of a list versus a vector for instance) 
But now the function is no longer pure (there's a side effect on tracks) and so it opens up new questions that I can only answer looking at the function code, e.g. is the tracks list reset by the function ? So IMO the function is less easy to reason about.
(and it would also complicate its usage in a multithreading situation, but ok, that's not really our business anyways :wink:)
Unless you (strongly) object, I would prefer to revert that change.

pillot: I don't strongly object ;')

aphecetche (approved):

PR #11419, opened 2023-06-02
Move away from deprecated FairMQ API

ktf:


ktf:
Keep it draft until we are happy with FairMQ 1.5.0

ktf:
I think FLP yesterday asked to hold on this. Did I miss something?

davidrohr:
Ah, didn't see that, perhaps...
Sorry...
But we are updating O2 today anyway, hopefully that will work out. And we are using it in production on the EPNs already. Worst case we can revert it :)

ktf:
> Worst case we can revert it :)

Yeah, I was hoping to avoid some conflicts I have with something in my PR pipeline...

PR #11425, opened 2023-06-02
correction of the  M46J carbon density

frmanso:
Decrease of the M46J carbon density from 1.84 to 1.48 (bug writting the density).


robincaron13 (approved):

shahor02:
trivial, merging

PR #11428, opened 2023-06-02
TPC interpolation store seed BEFORE changing it

martenole:
I want to store the track parameters of the seeding track (ITS outer parameter), so this must be done before the seed is manipulated of course


martenole:
Failure unrelated, merging

PR #11430, opened 2023-06-02
Avoid ROOT warning in EMCAL calib

martenole:
We see some errors in the IL which are actually ROOT warnings written to stderr and therefore picked up by our stderr monitor:

E	3	00:07:47.506		STDERR	stderr/calib-emcalchannel-badch	2fd1hDtUD6h	537421	stderr: Warning in <TFile::Append>: Replacing existing TH1: hist (Potential memory leak).

This is because o2::utils::TH2FFromBoost assigns by default the name "hist" to the histogram it returns.

That was not a real issue. This simply fixes the warning.


mfasDa:
@jokonig Please review! I also think this should not be a problem, since it is streaming the current state to disk replacing the old histogram is fine (it should be overwritten)

jokonig (approved):

martenole:
Failure unrelated, merging

PR #11432, opened 2023-06-02
Enable using CCDB VDrift in TPC Digitizer

shahor02:
@chiarazampolli @sawenzel let me know if this should be merged to avoid conflicts with ongoing/planned MCs.
Note that although the vdrift provided from the command line as `TPCGasParam.DriftV` ConfigurableParam overrides the vdrift from the CCDB, the CCDB object is still always loaded and its new part, the time offset, is applied unless it is in turn overridden from the command line by `TPCDetParam.DriftTimeOffset` ConfigurableParam.

### DIFF
--- a/Steer/DigitizerWorkflow/src/TPCDigitizerSpec.cxx
+++ b/Steer/DigitizerWorkflow/src/TPCDigitizerSpec.cxx
@@ -246,7 +246,7 @@ class TPCDPLDigitizerTask : public BaseDPLDigitizer
     auto& cdb = o2::tpc::CDBInterface::instance();
     cdb.setUseDefaults(!mUseCalibrationsFromCCDB);
     // whatever are global settings for CCDB usage, we have to extract the TPC vdrift from CCDB for anchored simulations
-    //    mTPCVDriftHelper.extractCCDBInputs(pc);
+    mTPCVDriftHelper.extractCCDBInputs(pc);
     if (mTPCVDriftHelper.isUpdated()) {
       const auto& vd = mTPCVDriftHelper.getVDriftObject();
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Hello @shahor02 ,

Thanks! i will try this on Monday. 
for the time offsets: for now 22cdef take the vdrift as config param, so for those we should switch off the offset in digitization, right? Or is the behavior consistent in sim and reco?
as a remark, the BC shift is never applied in MC, of course. 

in general, this PR should not affect any ongoing MC, since it is sw, and we don’t change version while running.

chiara

shahor02:
I've tested, it seems to work as expected. With the tag containing this PR  both the VDrift proper and the t0 offset will be loaded from CCDB object. As I wrote, `TPCGasParam.DriftV=...` passed from the command line will override only VDrift, but the t0 from the CCDB object will be used unless it is also overridden via `TPCDetParam.DriftTimeOffset=0`. That said, if one provides from the command line only  `TPCGasParam.DriftV=...` *both* in simulation and reconstruction, the t0 offset from the CCDB object will be used consistently in both of them.

Conversely, with current tags if you pass the DriftV from the command line in sim/rec. it will use default t0=0 in sim, but in the rec. it will use whatever t0 it gets from the CCDB object. This is irrelevant for 22cdef since they don't have t0-calibrated objects, but for the 2022 periods with VDrift provided by Alex we should either use the new tag or to override `TPCDetParam.DriftTimeOffset=0` both in sim and rec.

chiarazampolli:
Hello @shahor02 , 
I think we should merge this and start testing it in view of 500 kHz MC.
Thanks!
Chiara

PR #11434, opened 2023-06-03
PHOS Geometry : fix local to rel/abs cell id

dsekihat:
PHOS Geometry : fix local to rel/abs cell id

### DIFF
--- a/Detectors/PHOS/base/include/PHOSBase/Geometry.h
+++ b/Detectors/PHOS/base/include/PHOSBase/Geometry.h
@@ -134,7 +134,7 @@ class Geometry
   const TGeoHMatrix* getAlignmentMatrix(int mod) const { return &(mPHOS[mod]); }
 
  private:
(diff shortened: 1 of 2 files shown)


peressounko:
Dear Daiki, 
Thank you for looking at this bug, but I do not approve these changes:
1. actual step of PHOS cell (including still walls is 2.25 cm) - no need to change it
2. local coordinates should point to the center of a cell, not at the border, so it should be shifted by 0.5 w.r.t. sell number as we have even number of cells and point (0.,0.) in the center of PHOS module.
Empty lines at the edge of a module in the cluster occupancy plot  are related to depth correction. I fixed this in lines 446-447 of caloClusterProducer in PR https://github.com/AliceO2Group/O2Physics/pull/2729. Unfortunately, this PR stalled for a few days already.
    Best regards, Dmitri

dsekihat:
Dear Dmitri, Thank you for reply. I agree with you. I close this PR.

PR #11437, opened 2023-06-05
Fix rate limiting

ktf:


ktf:
This fixes the rate limiting in analysis for me. Most of the changes are actually unrelated, but somehow intertwined. Most of the changes are actually in other PRs already, so I will proceed to merge them as they pass tests, so that this one will finally become much simpler.

ktf:
One more attempt. This fixes the reproducer I have on my laptop.

ktf:
@ddobrigk @jgrosseo @aalkin @pzhristov this seems to finally fix the issue with the rate limiting. The PR is quite large, although it includes other cleanups and improvements. I tested on hyperloop using my build and a few of the usual workflows (Correlations, Histograms, SpectraTPC*). Could you verify a few of your own workflows, just to make sure I did not miss anything?

davidrohr (dismissed):
This looks ok to me, particularly the changes in RateLimiter.cxx, though I am sure they were not the problem, since that code path is used only in async reco, but anyway they seem correct.

PR #11443, opened 2023-06-06
DPL: hardened consumeWhenAny logic

ktf:
If there is no conditions at all, simply return Consume.

Wait for all conditions if they are missing.

Compared with the previous implemementation, this should force consumption for the case in which it is a non-condition input which is still missing. Not sure how that could happen, though.

### DIFF
--- a/Framework/Core/src/CompletionPolicyHelpers.cxx
+++ b/Framework/Core/src/CompletionPolicyHelpers.cxx
@@ -200,10 +200,14 @@ CompletionPolicy CompletionPolicyHelpers::consumeWhenAny(const char* name, Compl
 {
   auto callback = [](InputSpan const& inputs, std::vector<InputSpec> const& specs) -> CompletionPolicy::CompletionOp {
     bool canConsume = false;
-    // iterate on all specs and all inputs simultaneously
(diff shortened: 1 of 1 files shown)


ktf:
@sawenzel @benedikt-voelkel this is the only thing I can think of. Do you know if empty messages have a special meaning for that particular workflow?

benedikt-voelkel:
Maybe @davidrohr or @shahor02 can answer that?

shahor02:
In the TPCDigitizerSpec code I don't see any special check for empty payload, but I am not sure such messages are expected at all.

davidrohr:
> @sawenzel @benedikt-voelkel this is the only thing I can think of. Do you know if empty messages have a special meaning for that particular workflow?

@ktf : just for reference, I think empty messages can have a special meaning. E.g. if clusters are send as a vector, and empty message will represent 0 clusters.

ktf:
By empty you mean empty payload? The cornercase here is with empty header, which I think it's not possible. I can reopen this in any case. It should not hurt.

davidrohr:
sure, I meant empty payload.
No idea if empty headers can appear or not, but I do not see how.

PR #11444, opened 2023-06-06
make init of QEDepem more robust

sawenzel:
retry a few times in case of initialization errors


sawenzel:
Needs https://github.com/AliceO2Group/AEGIS/pull/16 and bump of AEGIS before merging.

PR #11447, opened 2023-06-06
DPL: enable early forwarding when only conditions are involved

ktf:
In this particular case, the optional mechanism is not triggered in any case, so we can safely do the early forwarding.

### DIFF
--- a/Framework/Core/src/DataProcessingDevice.cxx
+++ b/Framework/Core/src/DataProcessingDevice.cxx
@@ -1000,22 +1000,34 @@ void DataProcessingDevice::fillContext(DataProcessorContext& context, DeviceCont
   /// We must make sure there is no optional
   /// if we want to optimize the forwarding
   context.canForwardEarly = (spec.forwards.empty() == false) && mProcessingPolicies.earlyForward != EarlyForwardPolicy::NEVER;
+  bool onlyConditions = true;
+  bool overriddenEarlyForward = false;
   for (auto& forwarded : spec.forwards) {
+    if (forwarded.matcher.lifetime != Lifetime::Condition) {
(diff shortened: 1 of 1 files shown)


ktf:
@TimoWilken alibuild23 seems to be sluggish. Can you have a look if anything can be improved there?
@Barthelemy any progress with reducing the dependencies of QC?

Barthelemy:
@ktf not yet. I fear that in the end it will require changes in O2 because we need a couple of things that are in `GlobalTracking` and this pulls 80 libraries....

https://alice.its.cern.ch/jira/browse/QC-976

TimoWilken:
Hm, I'm not sure what's causing the slowdown there. alibuild23 already has a fast "io1" virtual disk attached.

Framework/Core/src/DataProcessingDevice.cxx
```diff
@@ -991,23 +991,35 @@ void DataProcessingDevice::fillContext(DataProcessorContext& context, DeviceCont
   /// We must make sure there is no optional
   /// if we want to optimize the forwarding
   context.canForwardEarly = (spec.forwards.empty() == false) && mProcessingPolicies.earlyForward != EarlyForwardPolicy::NEVER;
+  bool onlyConditions = false;
+  bool overriddenEarlyForward = false;
   for (auto& forwarded : spec.forwards) {
+    if (forwarded.matcher.lifetime == Lifetime::Condition) {
+      onlyConditions = true;
```
davidrohr: isn't this logic for detecting onlyConditions wrong, since it seems inverted?

ktf: you are right, fixed.

PR #11452, opened 2023-06-07
ITS-Studies: add skeleton of an entrypoint for generic ITS-related studies

mconcas:
The idea is copied from @shahor02 ITSOffset study.

I would like to gather a set of tasks, each performing some specific study related to ITS reco/tracking output, rather than spreading custom macros.

Current schema would be to have a single executable where one can choose which analysis to perform (if not all, by default), will see if this scales.

This way I hope to factorise the struggle related to the understanding of the DPL I/O and configuration for new/temporary contributors.
I also would like to maintain a single codebase for a tool to run systematic test/checks on data/MC given reconstruction objects.

In current PR I also included the skeleton of the impact parameter study that @apalasciano is going to integrate.


mconcas:
Errors unrelated, merging.

PR #11454, opened 2023-06-07
DPL: early release of buffers in aod reader

ktf:
This introduces the concept of LifetimeHolder, which allows getting rid of intermediate objects as soon as they get out of scope, rather than waiting for the processing to be done.

In particular this PR uses it for the TreeToTable codepath, effectively cutting in half the common needs of the AOD reader.


ktf:
Tested also in standalone analysis and on hyperloop.

PR #11456, opened 2023-06-08
FT0Digitizer: fixing misleading option --use-ccdb-ft0

afurs:
Misleading option `--use-ccdb-ft0` causes miscalibration for FT0 digitizer based on deprecated calibration object `FT0/Calib/ChannelTimeOffset`. Digitizer is not suppose to use it. This is preliminary suppression of the option. Later remaining peace of related code will be also deleted, due to no use anymore.


sawenzel:
@noferini

noferini:
Thanks

PR #11457, opened 2023-06-08
Residual aggregator finalizes slots upon state transition and does no…

martenole:
…t wait for eos

This hopefully reduces the occurrences of leftover residual files which cannot be transferred to EOS because they belong to slots which are not finalized.


martenole:
Putting a screenshot here to link it from jira on a slightly related topic...
valgrind-residual-aggregator

PR #11459, opened 2023-06-08
No need to create rootmap files since ROOT 6.16

ktf:


ktf:
Apparently still needed...:-/

PR #11465, opened 2023-06-09
Improvements to CTP-FT0 triggering in MC

sawenzel:
* CTP digitizer: apply trigger input mask only when input test succeeded
* CTP digitizer: Only create non-trivial CTPdigits (whenever mask collector is non-zero)

* add function to FT0 trigger to see easily see if it was vertex
* minor spelling fixes and code-reformating for better understandability

This is the result of some debugging. I noticed that CTP digitizer emits too many CTPdigits in comparison to FT0 vertex triggers, which shouldn't be the case for the default config which only triggers on T0-MTVX.

### DIFF
--- a/Detectors/CTP/simulation/src/Digitizer.cxx
+++ b/Detectors/CTP/simulation/src/Digitizer.cxx
@@ -27,15 +27,16 @@ std::vector<CTPDigit> Digitizer::process(const gsl::span<o2::ctp::CTPInputDigit>
   std::map<o2::detectors::DetID::ID, std::vector<CTPInput>> det2ctpinp = mCTPConfiguration->getDet2InputMap();
   // To be taken from config database ?
   std::map<std::string, uint64_t> detInputName2Mask =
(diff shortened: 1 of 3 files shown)


sawenzel:
@lietava @shahor02

Detectors/CTP/simulation/src/Digitizer.cxx
```diff
@@ -27,58 +27,71 @@ std::vector<CTPDigit> Digitizer::process(const gsl::span<o2::ctp::CTPInputDigit>
...
-    data.intRecord = hits.first;
     std::bitset<CTP_NINPUTS> inpmaskcoll = 0;
     for (auto const inp : hits.second) {
       switch (inp->detector) {
         case o2::detectors::DetID::FT0: {
           // see dummy database above
           for (auto const& ctpinp : det2ctpinp[o2::detectors::DetID::FT0]) {
             uint64_t mask = (inp->inputsMask).to_ullong() & detInputName2Mask[ctpinp.name];
-            inpmaskcoll |= std::bitset<CTP_NINPUTS>(ctpinp.inputMask);
+            if (mask) {
```
sawenzel, line 47: for the record, this is the main proposed change: inputsMask is the input from the FT0 detector ... but this could have been a non-vertex trigger and the `mask`  could be 0. Still, in this case the trigger was taken by CTP.

afurs (dismissed):

Detectors/CTP/simulation/src/Digitizer.cxx
lietava: @sawenzel : yes, you are right. That is why mask was not used.

PR #11466, opened 2023-06-09
DPL: fix plugin appending

ktf:


ktf:
Full CI passes. Obviously currently wrong. Merging.

PR #11467, opened 2023-06-09
DPL: support more than one callback

ktf:
For the moment this will not allow replacing callbacks, but in general it should not be a problem.


ktf:
@shahor02 what we discussed.

shahor02 (approved):
Works fine, thanks, please merge it if you consider it final

PR #11468, opened 2023-06-09
FST TMUX make CALIB_TASKS configurable

martenole:
@davidrohr with the default options now the FST goes OOM. I can of course adjust FST_TMUX_MEM_OVERRIDE so that it works again, but I think it is still useful to be able to test the different calibration tasks selectively. In principle we have for sync stage now BARREL_[TF/SPORADIC] FORWARD_[BARREL/SPORADIC] and CALO_TF


davidrohr:
fine with me, but what has changed that we need so much more memory suddenly?
We should check that...

martenole:
For this I probably need to bisect, I have no idea. For now merging this as is, since not checked by CI in any case and I have tested it locally

martenole:
False alarm - after clearing the disk caches and restarting the ccdb-memory service on epn000 the FST runs again smoothly also with the default settings. I believe we are simply close to the edge in terms of memory usage

PR #11471, opened 2023-06-09
Distribution of CTP trigger within CollisionContext

sawenzel:
This gives the (preliminary) possibility to distribute CTPdigits (momentarily, trigger decisions for FT0 MVTX) for digitizers.

This will work for MC that runs digitization in stages: 
*) First FT0 + CTP
*) Then digitizers that trigger on FT0 MVTX

In a future PR, this can be done with proper DPL-inputs, but this PR should already unblock work, where detectors need access to CTP trigger records.

There is a small example in HMPID, showing how the CTP triggers can be accessed.

### DIFF
--- a/DataFormats/Detectors/CTP/CMakeLists.txt
+++ b/DataFormats/Detectors/CTP/CMakeLists.txt
@@ -17,8 +17,9 @@ o2_add_library(DataFormatsCTP
           src/TriggerOffsetsParam.cxx
   PUBLIC_LINK_LIBRARIES O2::CommonDataFormat
                         O2::Headers
-                        O2::SimulationDataFormat
                         O2::CommonUtils
+                        O2::DetectorsCommonDataFormats
+                        O2::DataFormatsParameters
                         O2::CommonConstants)
 o2_target_root_dictionary(DataFormatsCTP
                           HEADERS include/DataFormatsCTP/Digits.h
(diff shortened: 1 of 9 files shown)


ktf:
Myself and @Barthelemy are now getting:

```
ninja: error: dependency cycle: stage/lib64/G__O2SimulationDataFormat_rdict.pcm -> stage/lib64/libO2DataFormatsCTP.so -> DataFormats/Detectors/CTP/CMakeFiles/O2lib-DataFormatsCTP.dir/src/Digits.cxx.o -> cmake_object_order_depends_target_O2lib-DataFormatsCTP -> DataFormats/Detectors/CTP/G__O2DataFormatsCTP.cxx -> stage/lib64/G__O2SimulationDataFormat_rdict.pcm
```

I am trying to rebuild from scratch, but maybe there is actually something wrong?

martenole:
Also seen by Chiara here: https://github.com/AliceO2Group/O2DPG/pull/1102#issuecomment-1590700483

sawenzel:
I am not seeing any problems. Just did a fresh compile and it passes without problems. Maybe this is related to some caching issue.

shahor02:
I had the same problem, had to run alibuild from scratch

ktf:
I think the issue is that our cmake macros are not able to understand when you drop a dependency and you readd it in a way it would cause a circular dependency when considering the original one. Indeed the only solution I could think of is rebuilding from scratch. I am not sure if there is any easy way we can correct this. At some point @davidrohr @aphecetche gave some thoughts about it, but I think the conclusion was there is nothing possible.

PR #11476, opened 2023-06-11
ITS: removed EoS / non EoS option + fit for pulse shape 2D scan

iravasen:
This includes:
- removal of EoS / non-EoS option: now we always use the internal data flags to ship the data to aggregator when the run is finished (before EoS arrival). This is a workaround for data drops at end-of-stream. 
- for pulse shape 2D scan, we include the option to make linear fit calculation for slope and intercept extraction (for IB ToT calibration).


iravasen:
Hello @mconcas, I'm done with this so for me as soon as all tests are passed it can be merged.

mconcas (approved):

PR #11482, opened 2023-06-12
CCDBDownloader: Added timeout settings

TrifleMichael:
Added better default timeout values, and the setters for them, based on Costin's suggestions.


TrifleMichael:
Sorry for the multiple commits - at first I didn't notice `setSocketTimoutTime` was referenced from outside the downloader

PR #11483, opened 2023-06-12
CCDBDownloader: Added timeout defaults and setters

TrifleMichael:


CCDB/src/CCDBDownloader.cxx
```diff
@@ -395,11 +424,14 @@ int CCDBDownloader::startTimeout(CURLM* multi, long timeout_ms, void* userp)
...
 {
   curl_easy_setopt(handle, CURLOPT_PRIVATE, data);
-
   curl_easy_setopt(handle, CURLOPT_CLOSESOCKETFUNCTION, closesocketCallback);
   curl_easy_setopt(handle, CURLOPT_CLOSESOCKETDATA, this);
   curl_easy_setopt(handle, CURLOPT_OPENSOCKETFUNCTION, opensocketCallback);
   curl_easy_setopt(handle, CURLOPT_OPENSOCKETDATA, this);
+
+  curl_easy_setopt(handle, CURLOPT_TIMEOUT_MS, mRequestTimeoutMS);
+  curl_easy_setopt(handle, CURLOPT_CONNECTTIMEOUT_MS, mConnectionTimeoutMS);
```
sawenzel, line 433: so with this, curl with automatically time out when no answer received within a certain time?

TrifleMichael: Yes. Request timeout is the time for the entire request to be completed, while connection timeout is only for establishing connection. Connection timeout is shorter if the client receives information that destination cannot be reached (for example from their router).

sawenzel (approved):

sawenzel:
Did you test this? Could you setup a system with Costins local server in which the timeouts can be forced?

TrifleMichael:
> Did you test this? Could you setup a system with Costins local server in which the timeouts can be forced?

I tested both connection timeouts and request timeouts using a local instance of CCDBDownloader (without Costin's server).

For connection timeouts I tried to connect to a valid but non existent IP.
For request timeouts I started downloading data from alicdb2 and then turned off the network on my computer.
Both of those timeouts work as they should.

I have not tested the happy eyeballs algorithm.

I will setup a local server and perform more tests there. I'll update you with the results.

TrifleMichael:
> Did you test this? Could you setup a system with Costins local server in which the timeouts can be forced?

After further testing using the ccdb-local and iptable rules I am certain all three timeouts work properly.

On localhost the happy eyeballs algorith will not engage, as it requires the DNS to resolve ipv6. Instead it will connect over ipv4 without any delay, regardless of set happy eyeballs timout. I've tested this algorithm on remote servers and it works well.

sawenzel:
Thanks. Merging.

PR #11485, opened 2023-06-12
Add TRD pulse height DPL device (dummy)

martenole:
A starting point to move the creation of the pulse height spectra from QC into O2, such that
a) we avoid the sampling of digits and get more statistics in the plots
b) the PH plots can be used downstream for example for the T0 calibration

This is currently a dummy.
TODO:
- move the algorithm for creating the PH plot from QC into `PulseHeight::process()`
- add a writer to write the output to a file
- initialization and reset methods (if needed, depending on the output type chosen)


f3sch (approved):
Since this is a dummy anyways, I will just approve without looking thoroughly :+1:

martenole:
Failures unrelated, tested locally, merging

PR #11486, opened 2023-06-12
Some uncorrelated improvements

martenole:
Important is the change to the residual aggregator in order to query the CCDB objects only once at SOR, since no updated objects are required


martenole:
Trivial changes, fullCI passed and the change to the residual aggregator should be deployed at P2. So merging

PR #11489, opened 2023-06-13
CCDBDownloader: More error checks

TrifleMichael:
Added result check for all libuv functions that return an error code. Libuv functions that lack this check are the ones returning void.

Even though almost all libuv function errors result in segmentation faults, this change may help with debugging the downloader in the future.


TrifleMichael:
I've added 4 types of error checks: Libuv, CurlMulti, CurlEasy and Sockets.

There is an additional check for whether the uvloop closes prematurely. That check uses the mIsClosing variable.

Because sockets are opened in only one line, and closed in two lines I didn't create a seperate function for testing them. Additionaly not creating that function makes it easier to differentiate whether closing or opening failed.

I've tested to make sure that no errors are logged on the current version. If an error causes a segmentation fault the thread will crash **without** calling LOG(error).

davidrohr:
Could you rebase without merge commits, and squash the fixup commits?

PR #11490, opened 2023-06-13
[EMCAL-565] Add fucntionality to only select specific trigger in calib

jokonig:
- When EMCal gamma or jet triggers are used during data taking, these events have to be rejected as otherwise the energy spectrum would have turn-on curves, which might result in a wrong calibration of time and bad channels. Therefore, only specific triggers are allowed in the analysis.
- The selection of triggers can be configures via the calib params. A string of : speparated trigger names can be defined there. Only those triggers will be selected.
- The trigger names are checked at the initialization in finaliseCCDB (each time the CTP config changes) and mapped to the corresponding trigger bit
- For each emcal trigger, the bc from the ctp is matched with the bc of the emcal event and it is then checked if one of the selected triggers is contained in the current event

### DIFF
--- a/Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
+++ b/Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
@@ -33,6 +33,8 @@
 #include "CommonUtils/NameConf.h"
 #include "DetectorsBase/GRPGeomHelper.h"
 #include "Framework/CCDBParamSpec.h"
+#include "DataFormatsCTP/Digits.h"
+#include "DataFormatsCTP/Configuration.h"
 
 // for time measurements
 #include <chrono>
@@ -50,7 +52,7 @@ class EMCALChannelCalibDevice : public o2::framework::Task
   using EMCALCalibParams = o2::emcal::EMCALCalibParams;
 
  public:
-  EMCALChannelCalibDevice(std::shared_ptr<o2::base::GRPGeomRequest> req, bool params, std::string calibType, bool rejCalibTrg) : mCCDBRequest(req), mLoadCalibParamsFromCCDB(params), mCalibType(calibType), mRejectCalibTriggers(rejCalibTrg) {}
+  EMCALChannelCalibDevice(std::shared_ptr<o2::base::GRPGeomRequest> req, bool params, std::string calibType, bool rejCalibTrg, bool rejL0Trig) : mCCDBRequest(req), mLoadCalibParamsFromCCDB(params), mCalibType(calibType), mRejectCalibTriggers(rejCalibTrg), mRejectL0Triggers(rejL0Trig) {}
 
   void init(o2::framework::InitContext& ic) final
   {
@@ -87,13 +89,37 @@ class EMCALChannelCalibDevice : public o2::framework::Task
     if (matcher == ConcreteDataMatcher("EMC", "EMCALCALIBPARAM", 0)) {
       LOG(info) << "EMCal CalibParams updated";
       EMCALCalibParams::Instance().printKeyValues(true, true);
-      return;
     }
     if (matcher == ConcreteDataMatcher("EMC", "SCALEFACTORS", 0)) {
       if (mBadChannelCalibrator && EMCALCalibParams::Instance().useScaledHisto_bc) {
         LOG(info) << "Configuring scale factors for bad channel map";
         mBadChannelCalibrator->getCalibExtractor()->setBCMScaleFactors(reinterpret_cast<o2::emcal::EMCALChannelScaleFactors*>(obj));
         mScaleFactorsInitialized = true;
+      }
+    }
+    if (matcher == ConcreteDataMatcher("CTP", "CTPCONFIG", 0)) {
+      // clear current class mask and prepare to fill in the updated values
+      // The trigger names are seperated by a ":" in one string in the calib params
+      mSelectedClassMasks.clear();
+      std::string strSelClassMasks = EMCALCalibParams::Instance().selectedClassMasks;
+      std::string delimiter = ":";
+      size_t pos = 0;
+      std::vector<std::string> vSelMasks;
+      while ((pos = strSelClassMasks.find(delimiter)) != std::string::npos) {
+        vSelMasks.push_back(strSelClassMasks.substr(0, pos));
+        strSelClassMasks.erase(0, pos + delimiter.length());
+      }
+      vSelMasks.push_back(strSelClassMasks);
+
+      auto ctpconf = reinterpret_cast<o2::ctp::CTPConfiguration*>(obj);
+
+      for (auto& cls : ctpconf->getCTPClasses()) {
+        LOG(debug) << "CTP class: " << cls.name << "\t " << cls.classMask;
+
+        if (std::find(vSelMasks.begin(), vSelMasks.end(), cls.name) != vSelMasks.end()) {
+          mSelectedClassMasks.push_back(cls.classMask);
+          LOG(info) << "Setting selected class mask " << cls.name << " to bit " << cls.classMask;
+        }
       }
     }
   }
@@ -122,11 +148,16 @@ class EMCALChannelCalibDevice : public o2::framework::Task
       pc.inputs().get<o2::emcal::EMCALChannelScaleFactors*>("EMC_Scalefactors");
     }
 
+    // prepare CTPConfiguration such that it can be loaded in finalise ccdb
+    pc.inputs().get<o2::ctp::CTPConfiguration*>(getCTPConfigBinding());
+
     if (!mIsConfigured) {
       // configure calibrators (after calib params are loaded from the CCDB)
       configureCalibrators();
       mIsConfigured = true;
     }
+
+    auto ctpDigits = pc.inputs().get<gsl::span<o2::ctp::CTPDigit>>(getCTPDigitsBinding());
 
     // reset EOR behaviour
     if (mTimeCalibrator) {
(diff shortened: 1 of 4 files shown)


mfasDa (commented):
In general looks good, only I am not sure I like too much the idea of adding a dependency to O2::GlobalTracking for adding the RecoContainer just for getting the CTP digits - dependencies should be as lightweight as possible. If the GlobalTracking bucket depends on SimulationDataFormat the creates a pile of dependencies.

mfasDa:
@shahor02 Please review! Since we now start enabling the L0 triggers we must distinguish between the two types particularly in the bad channel calib, since hit distribution and spectral shape looks very different in triggered events compared to min. bias events.

mfasDa:
@lietava Please check the trigger selection! Ideally CTP would provide a common helper to get the trigger class from the CTP digits via the CTPConfiguration - this might be needed also in other parts in the synchronous stage (i.e. QC).

jokonig:
@mfasDa I removed the dependency to the global tracking. Thanks for the suggestion to directly access the CTP digits!

lietava:
hi all, comments to procedure, if I understand well: Trigger class names are kept constant.
For every run one finds from class names the position if these classes in trigger class mask and then select/deselect them.
Correct ?
(I'll check the code later I am travelling at the moment.)

lietava:
I looked at ctp part of the code. Logically looks ok.
Was it tested ?

@mfasDa : do you mean something like 
uint64_t CTPConfiguration::getClassMask(std::string& classname)
or
bool CTPConfiguration::IsClassFired(CTPDigit& digit, std::string& classname)
?

mfasDa (dismissed):
Looks good!

mfasDa:
The full CI fails for this PR as there is no default object for the unanchored MC range (@lietava).

mfasDa:
> I looked at ctp part of the code. Logically looks ok. Was it tested ?
> 
> @mfasDa : do you mean something like uint64_t CTPConfiguration::getClassMask(std::string& classname) or bool CTPConfiguration::IsClassFired(CTPDigit& digit, std::string& classname) ?

Yes, this is what was thinking about. Since the CTPConfiguration is in DataformatsCTP not sure whether a utility function would be needed - depending on whether content from higher level libraries is needed. Such functionality is needed for the synchronous stage and having all detectors implementing it themselves is in my opinion not a useful long-term solution.

lietava:
> The full CI fails for this PR as there is no default object for the unanchored MC range (@lietava).

https://github.com/AliceO2Group/AliceO2/blob/a18d4725b2052ed582b886d7a0d76d5fb08be283/Detectors/CTP/simulation/src/Digitizer.cxx#LL133C3-L138C45
This one is not good ?

shahor02:
> > The full CI fails for this PR as there is no default object for the unanchored MC range (@lietava).
> 
> https://github.com/AliceO2Group/AliceO2/blob/a18d4725b2052ed582b886d7a0d76d5fb08be283/Detectors/CTP/simulation/src/Digitizer.cxx#LL133C3-L138C45 This one is not good ?

@lietava the lines you refer to use hardcoded timestamp 1546300800000, this is in principle not good but will work since there is a matching object: http://alice-ccdb.cern.ch/browse/CTP/Config/Config/1546300800000

But the fullCI error comes from the DPL CCDB fetcher as it is requested by the AODProducer: https://github.com/AliceO2Group/AliceO2/blob/74ef0705c240e5818f17fe19af4f4b26e9bc94e2/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L2653
The request is done with the runNumber in the metadata, and this makes the query fail even though there is an object matching by timestamp (but not runnumber): 
```
reco_NOGPU.log-[66462:internal-dpl-ccdb-backend]: [06:45:58][ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch/CTP/Config/Config/1550600800000/runNumber=310000/
```

mfasDa:
@lietava It crashes with
```
reco_NOGPU.log-[62203:internal-dpl-ccdb-backend]: [23:48:24][ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch/CTP/Config/Config/1550600800000/runNumber=310000/
reco_NOGPU.log-[62203:internal-dpl-ccdb-backend]: [23:48:24][FATAL] Unable to find object CTP/Config/Config/1550600800000
reco_NOGPU.log:[62203:internal-dpl-ccdb-backend]: [23:48:24][ERROR] Exception caught: Fatal error 
```
Indicating that the CCDB object for the unanchored MC range is not found. However in the CCDB I see for CTP/Config/Config an object with validity range 1451606461000 to 1623565274588 that should cover the request. @shahor02 any idea? Could it be that the query for the run number fails, because I don't see this as meta data for the CCDB object.

lietava:
@mfasDa : 
 If I understand well Ruben comment above than the runnumber is to be changed.
Validity range is as described https://alice.its.cern.ch/jira/browse/O2-3101
Then second point is that this CTPConfig probably does not contain emcal classes with emcal inputs (checking) which you need ?

mfasDa:
@lietava I think changing the run number will not help, since there might be a whole run range for unanchored MC. So changing for one case might break another case. So either the run number is removed from the request or you need to create dedicated objects for the entire run range of the unanchored MC.

Then indeed we need the EMCAL readout trigger, otherwise the test becomes meaningless. As you know we are asking for this since a while, not only for the full CI but more because the simulations done at the moment are completely useless for EMCAL and other triggered detectors I expect.

Detectors/EMCAL/calibration/testWorkflow/EMCALChannelCalibratorSpec.h
```diff
@@ -328,25 +392,34 @@ DataProcessorSpec getEMCALChannelCalibDeviceSpec(const std::string calibType, co
...
     inputs.emplace_back("EMC_CalibParam", o2::header::gDataOriginEMC, "EMCALCALIBPARAM", 0, Lifetime::Condition, ccdbParamSpec("EMC/Config/CalibParam"));
   }
   if (calibType.find("badchannel") != std::string::npos) {
     inputs.emplace_back("EMC_Scalefactors", o2::header::gDataOriginEMC, "SCALEFACTORS", 0, Lifetime::Condition, ccdbParamSpec(CalibDB::getCDBPathChannelScaleFactors()));
   }
+  // inputs.emplace_back("EMC_BadChannelMap", o2::header::gDataOriginEMC, "BADCHANNELMAP", 0, Lifetime::Condition, ccdbParamSpec("EMC/Calib/BadChannelMap"));
+
+  // data request needed for rejection of EMCal trigger
+  bool CTPConfigPerRun = true;
+  inputs.emplace_back(device::getCTPConfigBinding(), "CTP", "CTPCONFIG", 0, Lifetime::Condition, ccdbParamSpec("CTP/Config/Config", CTPConfigPerRun));
```
shahor02: Hi, as I wrote above, the fullCI fails in the CCDB request of the CTP config to be matched both by timestamp and the runNumber, while the MC CCDB object matching by time has no runNumber associated.

But in opposite to what I wrote, this is not due to the CTP config request from the AOD producer (there the runNumber request is optional and is off by default, see https://github.com/jokonig/AliceO2/blob/7520e708096e97c836a6675670275837c133a849/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L113) but from these lines, where the runNumber is requested unconditionally.

Note that requesting the same object with different requirements from different processes is not a well-defined procedure, perhaps one should always require runNumber for CTP config but be more forgiving if the object has no such a tag.
@costing would it be possible to retrieve the object if the only matching one by timestamp has no runNumber in the metadata ?

costing: Not really, a metadata constraint is applied directly in the query so the database will not return anything if no object with the indicated run number is found. If the client is ok with a non-matching run number, how about repeating the query without that constraint as a fallback, only for this particular usecase?

lietava:
@shahor02 : I need to create new CTP Config .
It is run independent therefore I suggest to have it unanchored with run in the full run range 300000-309999 as before.
Is it ok ?

shahor02:
With the code of this PR: not really. At the moment if you require runNumber in the object metadata, then it is strictly required.

lietava:
HI @shahor02 , I understand there is problem getting CTPConfig from database. But I can not understand from the above discussion where exactly it is.
I was trying to add to this PR the exception catching to understand better, see https://github.com/lietava/AliceO2/tree/EMCAL-565, draft PR11512 (it seems I do not have permission to push to this PR).

shahor02:
@lietava this PR declares requires a run-dependent CCDB CTP config (i.e. having `runNumber=xxx` in the metadata). Since the CTP config matching to the run number used in fullCI does not have runNumber associated, it's fetching fails.

lietava:
@shahor02 :thanks. 
Still why it was working before ?
There is only one line you highlighted where ctp config is used or am I missing other place ?
Then for real data config has to be run dependent.
For MC I see 3 options:
-  run validity in database is over all runs - is it possible ?
- if run config is not found default from code is taken 
- run dependent ctp configs are created for MC
-

shahor02:
@lietava when before? This unconditional request for run dependent CTP config is introduced in this PR. There's a similar line in the AODproducer but there the run tag is optional and is off in MC.

For the solution, let's see what @costing says about the possibility to still fetch the object if it has no any run tag.
We have thousands of run slots for non-anchored MC, defining config for all of them would be an overkill.

costing:
I don't think we can change the query logic while guaranteeing consistent behavior for all other workflows. To me if the client doesn't actually need an object for a particular run number, why not drop the constraint upfront? Or as a fallback query (i.e. try first with the run number, if nothing matches, try without).

shahor02:
@costing, the problem is that in general the CTP configs are run-specific, and this should be respected also when doing anchored MC. But we have this long range of dummy run-numbers for non-anchored MC and their objects are mostly default or field/collision type specific. Would not it be useful in general to consider an object with the metadata field XXX matched (but with a lower priority than an exact match)  if the field value is wildcarded, by e.g. "*"?

shahor02:
@roman @mfasDa we discussed with @costing , even if the general solution is possible it may take some time. Meanwhile, could you open an upload request for CTP config for runs 310000 (pbpb) and 300000, 30300 (pp) ?

jokonig:
Hi, 
I now changed the code such that the bool ctpcfgperrun is a workflow argument and is off by default (same way as in the AODProducerWorkflow)

shahor02:
> Hi, I now changed the code such that the bool ctpcfgperrun is a workflow argument and is off by default (same way as in the AODProducerWorkflow)

OK, but in the long run this is not a viable solution, also for the AODProducerWorkflow: a wrong object may be picked for the MC anchored to a run corresponding to a timestamp with multiple CTP configs matching.

lietava:
> @roman @mfasDa we discussed with @costing , even if the general solution is possible it may take some time. Meanwhile, could you open an upload request for CTP config for runs 310000 (pbpb) and 300000, 30300 (pp) ?

@shahor02 : I created ticket for creating configs: https://alice.its.cern.ch/jira/browse/O2-3866

shahor02:
@lietava ok, but I don't see any object attached in the upload ticket.

lietava:
> 
No, sorry, I am busy. Let you know when ready.

jokonig:
@shahor02 could you merge this PR? Or is there anything that still needs to be done?

shahor02:
@mfasDa should this be already merged or CTP needs to upload something before?

mfasDa:
From how I understand it should be ready to merge, no more changes are needed on the CTP side.

PR #11492, opened 2023-06-13
Optionally prescale query for objects with non-default ccdb-query-rate

shahor02:
Option `--condition-tf-per-query-multiplier N` will downscale by N the query rate for objects whose `CCDBParamSpec` was defined with non-trivial `ccdb-query-rate M` option (e.g. request query at every TF for `M=1`).
The check for such an object will be done once the previously checked `timingInfo.TFCounter has ID < N*M` (note the change: before the check would be done only for (timingInfo.timeSlice % M) == 0.

Also, the `--condition-tf-per-query M` will trigger a check not for `(timingInfo.timeSlice % M) == 0` but once previously checked TFCounter is more than M TFs old.

@martenole e.g. with ` --condition-tf-per-query-multiplier 360` the workflow will query objects line VDrift or BField only once per `~ 1 second` instead of every TF.

### DIFF
--- a/Framework/CCDBSupport/src/CCDBHelpers.cxx
+++ b/Framework/CCDBSupport/src/CCDBHelpers.cxx
@@ -34,6 +34,7 @@ struct CCDBFetcherHelper {
     size_t cacheHit = 0;
     size_t minSize = -1ULL;
     size_t maxSize = 0;
+    int lastCheckedTF = 0;
   };
 
   struct RemapMatcher {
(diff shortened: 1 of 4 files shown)


martenole:
Thanks @shahor02 !
So how do I change the aggregator workflow to take into account this option? Should each workflow get the option `--internal-dpl-ccdb-backend "--condition-tf-per-query-multiplier 360"`? Or is giving this option once enough? In the end we anyways only have a single ccdb backend, right?

shahor02:
@martenole  https://github.com/AliceO2Group/AliceO2/pull/11494

PR #11493, opened 2023-06-14
DPL Analysis: Table builder rework to reduce usage of `std::index_sequence`

aalkin:


ktf:
Nice. Is this ready to go?

aalkin:
@ktf yes, I have tested it with local analysis.

PR #11495, opened 2023-06-14
GlobalTrackClusterReader: rate limiting with an external device

ktf:
This injects a new device in the topology, with the purpose of implementing rate limiting with multiple devices.


ktf:
@chiarazampolli @shahor02 this should be a viable stop gap solution while we reduce the memory and / or we have proper rate limiting without the need of an extra device.

PR #11497, opened 2023-06-14
ALICE3-TRK: Rewrite TRK simulation from scratch

mconcas:
- Reduce complexity down to almost bare minimum
- Remove dead code
- Add possibility to configure/dump TRK layout onto .txt file


mconcas:
CI error unrelated. Merging.

PR #11500, opened 2023-06-15
Improve simulation reproducibility with Geant4

sawenzel:
This achieves reproducible Geant4 simulations with multicore, sub-event parallel, o2-sim (when o2-sim is started with the same seed).


sawenzel:
Needs https://github.com/vmc-project/geant4_vmc/pull/45 during runtime. Will mark it as WIP once compilation confirmed.

sawenzel:
Should be ok after https://github.com/alisw/alidist/pull/5065 has been merged

PR #11501, opened 2023-06-15
DPL: warn if we are unable to forward on a channel

ktf:
Also printout an error if there was an error while forwarding.

### DIFF
--- a/Framework/Core/src/DataProcessingDevice.cxx
+++ b/Framework/Core/src/DataProcessingDevice.cxx
@@ -590,7 +590,16 @@ static auto forwardInputs = [](ServiceRegistryRef registry, TimesliceSlot slot, 
     auto channel = proxy.getForwardChannel(ChannelIndex{fi});
     LOG(debug) << "Forwarding to " << channel->GetName() << " " << fi;
     // in DPL we are using subchannel 0 only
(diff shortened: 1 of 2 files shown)


ktf:
@davidrohr as discussed.

davidrohr:
> @davidrohr as discussed.

thx, will you add the check of the zmq return value for the receive as well?

ktf:
I have just added it. I am not sure what will happen on STOP, though. @teo will you tear down the channels on the RUNNING->READY transition?

teo:
> I have just added it. I am not sure what will happen on STOP, though. @teo will you tear down the channels on the RUNNING->READY transition?

Not sure I understand the question. I only request the `STOP` transition, then it's up to FairMQ to do any channel magic it deems necessary. I don't push any properties nor perform any other call.

davidrohr:
Let's merge this and see what happens. In principle, our stop timeout and EoS, which delays stop, should make sure that we don't get errors in case of a sane shutdown.

davidrohr (approved):

ktf:
@teo I vaguely remembered that it was the plugin to create the channels, but apparently it's not. Maybe it was ODC only behaviour. Nevermind.

PR #11505, opened 2023-06-15
[O2-1787] Implementation of the Beam Pipe Support inside the ITS Cage on the A Side

mario6829:
The Beam Pipe Support structure anchored to the ITS Cage and placed on the A Side was implemented following the blueprints with pretty a lot of details. The actual implementation is inside the V3Cage class. To avoid fake overlaps with the rough description of the ITS services (cables and the like) inside the Cage, these volumes were slightly modified (in the V3Services class): two holes were drilled on the sides of the service shapes through which the supporting bars can pass.

### DIFF
--- a/Detectors/ITSMFT/ITS/simulation/include/ITSSimulation/V3Cage.h
+++ b/Detectors/ITSMFT/ITS/simulation/include/ITSSimulation/V3Cage.h
@@ -91,23 +91,23 @@ class V3Cage : public V11Geometry
 
   /// Creates the Titanium lower part of the collar supporting the beam pipe
   /// \param mgr  The GeoManager (used only to get the proper material)
-  TGeoCompositeShape* createBPSuppLowerCollar(const TGeoManager* mgr = gGeoManager);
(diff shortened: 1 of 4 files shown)


mario6829:
Hi @mconcas @iouribelikov here it is the implementation of the beam pipe support inside the cage. The services were slightly modified to avoid any overlap. Please let me know in case of problems.

mconcas (approved):

martenole:
This breaks the fullCI with the warnings from [above](https://github.com/AliceO2Group/AliceO2/pull/11505#issuecomment-1596117306)
I wonder why the fullCI still got green here. Can we remove the redundant void arguments @mconcas ?

martenole:
> This breaks the fullCI with the warnings from [above](https://github.com/AliceO2Group/AliceO2/pull/11505#issuecomment-1596117306) I wonder why the fullCI still got green here. Can we remove the redundant void arguments @mconcas ?

I have removed them here: https://github.com/AliceO2Group/AliceO2/pull/11537

PR #11506, opened 2023-06-15
Add debug output to TPC residual aggregator

martenole:


martenole:
Trivial, merging because of https://alice.its.cern.ch/jira/browse/O2-3859

PR #11509, opened 2023-06-15
Introduce proper generator IDs

benedikt-voelkel:
* 3 values forseen
  * global ID assigend to a o2::eventgen::PrimaryGenerator
  * cocktail constituent ID in case a o2::eventgen::Generator consists
    of multiple cocktail constituents. If a specific constituent is used
    for an event, that event will be flagged with that ID
  * source ID to mark source in emebdding scenarios

* all 3 values are encoded into a single short, passed to mcCollision
  table at the end

* provide decoding via helper functions

* Global ID (and short description)
  * can be set via
    o2-sim --confKeyValues \
     "PrimaryGenerator.id=3;PrimaryGenerator.description=a specific gen"

* Each cocktail constituent must first be registered via
  o2::eventgen::Generator::addCocktailConstituent(int, std::string)

  then, each egnerated event must set a valid ID during
  o2::eventgen::Generator::GenerateEvent
  or
  o2::eventgen::Generator::importParticles

  if cocktail constituents are set but no valid ID is given, no event
  will be generated

### DIFF
--- a/Framework/Core/include/Framework/AnalysisDataModel.h
+++ b/Framework/Core/include/Framework/AnalysisDataModel.h
@@ -19,7 +19,7 @@
 #include "CommonConstants/PhysicsConstants.h"
 #include "CommonConstants/GeomConstants.h"
 #include "CommonConstants/ZDCConstants.h"
-#include "SimulationDataFormat/MCGenStatus.h"
+#include "SimulationDataFormat/MCGenProperties.h"
 
 using namespace o2::constants::math;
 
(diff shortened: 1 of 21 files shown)


benedikt-voelkel:
Tested with https://github.com/benedikt-voelkel/O2DPG/tree/cocktail-test

`${O2DPG_ROOT}/MC/run/PWGHF/run_pp_HFtriggers_ccbar_gaptrigger.sh`

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -1278,6 +1278,13 @@ DECLARE_SOA_COLUMN(PosZ, posZ, float);                       //! Z vertex positi
 DECLARE_SOA_COLUMN(T, t, float);                             //! Collision time relative to given bc in ns
 DECLARE_SOA_COLUMN(Weight, weight, float);                   //! MC weight
 DECLARE_SOA_COLUMN(ImpactParameter, impactParameter, float); //! Impact parameter for A-A
+DECLARE_SOA_DYNAMIC_COLUMN(GetGenId, getGenId,               //! The global generator ID which might have been assigned by the user
```
jgrosseo: Could you remove the comment in line 1273 and add a comment behind the //! in line 1274 that the information in GeneratorsID is encoded with o2::mcgenid::getCocktailId?
Thanks!

benedikt-voelkel: I added a comment to use the disentangled dynamic columns to access generator IDs. Do you think the comment makes sense?

jgrosseo: I think the comment is good. Now at second read, I wonder though why you call it `GetGenId` and not `GetGeneratorId`. I think the longer version would read better but I don't have a strong opinion.

Shouldn't we also remove the "// TODO enum to be added" because this does not make sense anymore now (at least not for the global column but only for the sub-ids...

benedikt-voelkel: While addressing your comment, I also did some re-naming a few days ago. In particular, I changed "cocktail" to "subGenerator".
 Because what we named "cocktail" previously in this PR did not align with what we understood of that in Run1 and Run2.

jgrosseo: Thanks!

PR #11513, opened 2023-06-16
[FOCAL-12] Fixes in PadPedestal object for CCDB access

mfasDa:
- Add dictionary for ChannelID
- Add operator== for object comparison


mfasDa:
After CCDB upload when reading back the object I get number of Channels = 0, and a warning about a missing dictionary entry for ChannelID. The dictionary entry is added explicitly, together with the comparison operator used for object checks.

mfasDa:
CI errors unrelated

PR #11517, opened 2023-06-16
Allow early release tables

ktf:


ktf:
@ddobrigk @njacazio @nburmaso Once merged, this should allow you to call ".release()" on `TableBuilder` / `Produces<>` instances to signal that you are done filling for the timeframe. This should allow you to significantly reduce the peak memory used in e.g. PID or the AOD producer.

ddobrigk:
@ktf That's great, thank you very very much! I will test soon and let you know!

ktf:
Ok, looks like we need a small patch in O2Physics with some compilers due to some missing optimisation. Will do it in a bit and retry.

PR #11518, opened 2023-06-16
Adding SOX and EOX to RCT CCDB

chiarazampolli:
@shahor02 , @noferini , @sawenzel , @catalinristea , @benedikt-voelkel , @gconesab : this should be used for anchoring.


shahor02 (approved):
trivial, merging

PR #11523, opened 2023-06-18
Revert "Condtion ctf*root.part -> ctf.root renaming by creation of me…

martenole:
…tafile"

This reverts commit 8a7d1ba4436b1b75d0e955dc99758c4cadd7a07d.

@shahor02 I thought about it and I don't think that it is possible to write the meta file before the file is available with its final name. Even with some tricks, we would not want the epn2eos tool to transfer the part file, so we can write the meta file only after the data file is final.

But please check if you have another idea.


shahor02:
Thanks @martenole , https://github.com/AliceO2Group/AliceO2/pull/11524 fixes the problem, closing this one.

PR #11524, opened 2023-06-18
Fix creation of CTF metadata file

shahor02:
tested locally with and w/o metadata creation.


Detectors/CTF/workflow/src/CTFWriterSpec.cxx
```diff
@@ -616,14 +621,14 @@ void CTFWriterSpec::closeTFTreeAndFile()
           metaFileOut << ctfMetaData;
           metaFileOut.close();
           if (!TMPFileEnding.empty()) {
-            std::filesystem::rename(o2::utils::Str::concat_string(mCurrentCTFFileNameFull, TMPFileEnding), mCurrentCTFFileNameFull);
+            std::filesystem::rename(actualFileName, mCurrentCTFFileNameFull);
```
martenole, line 624: Ah ok, I see. So now if the CTF writer would crash in the worst case after the renaming in this line then we would at least at that point have the temporary meta file and then we could more easily rename that meta file if we would want to (and if this would actually ever happen of course).
I thought first that the logic to write the meta file first and then the data file could never work, but the way you have it now is the best we can do I think. Thanks @shahor02 !

PR #11528, opened 2023-06-20
DPL Analysis: handle fully unassigned case

aalkin:
Correctly handles cases where there are no assigned (i.e. with positive index) groups can be found in slice cache.


ktf:
Currently broken. Merging.

PR #11529, opened 2023-06-20
FT0Digitizer: hotfix for default value in --use-ft0-ccdb option

afurs:
Options defined like `ConfigParamSpec{"option", o2::framework::VariantType::Bool, true, {"option"}} ` always returns `true` via access `configcontext.options().get<bool>("option")` whether option provided or not. So, default value `true` for boolean types always suppresses option usage.

### DIFF
--- a/Steer/DigitizerWorkflow/src/SimpleDigitizerWorkflow.cxx
+++ b/Steer/DigitizerWorkflow/src/SimpleDigitizerWorkflow.cxx
@@ -192,7 +192,7 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
   workflowOptions.push_back(ConfigParamSpec{"ccdb-tof-sa", o2::framework::VariantType::Bool, false, {"enable access to ccdb tof calibration objects via CCDBManager (obsolete remap to use-ccdb-tof)"}});
 
   // option to use/not use CCDB for FT0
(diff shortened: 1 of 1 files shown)


shahor02:
@afurs effectively, you are switching off using CCDB in the digitization, is it ok?

afurs:
@shahor02 actually, this will make option usage correct. If `-use-ft0-ccdb` provided, CCDB will be used in FT0 digitization. Currently it doesn't mater because for miscalibration all objects have zero values and covers all period(please see https://alice.its.cern.ch/jira/browse/O2-3811), so there will be no effect anyway.

shahor02:
yes, but if the default behaviour should be to used the ccdb, then it makes sense to change this option to e.g.
```
ConfigParamSpec{"ignore-ccdb-ft0", o2::framework::VariantType::Bool, false, {"disabable access to ccdb ft0 calibration objects"}}
```
...
```
auto useCCDB = !configcontext.options().get<bool>("ignore-ccdb-ft0");
```

noferini:
Dear all,
I agree with Ruben but since we are going to restart a test (tomorrow?) I would suggest to merge this PR so that we can include the fix in that test.
Then I would ask @afurs  to open a new PR following the suggestion of Ruben (at the same time we will include in O2DPG the ignore flag till the de-calibrations are ready (as done for reconstruction).
Does it sound good for you?
Cheers,
Francesco

PR #11531, opened 2023-06-20
[QC-889] Retry curl call in case of failure

Barthelemy:
Configurable. Off by default. The delay, in microseconds, can also be specified. 

We will use it in QC.

### DIFF
--- a/CCDB/include/CCDB/CcdbApi.h
+++ b/CCDB/include/CCDB/CcdbApi.h
@@ -556,7 +556,7 @@ class CcdbApi //: public DatabaseInterface
   mutable TGrid* mAlienInstance = nullptr;                       // a cached connection to TGrid (needed for Alien locations)
   bool mNeedAlienToken = true;                                   // On EPN and FLP we use a local cache and don't need the alien token
   static std::unique_ptr<TJAlienCredentials> mJAlienCredentials; // access JAliEn credentials
-  int mCurlRetries = 1;
+  int mCurlRetries = 3;
   int mCurlDelayRetries = 100000; // in microseconds
 
   ClassDefNV(CcdbApi, 1);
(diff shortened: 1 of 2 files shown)


Barthelemy:
@shahor02 FYI

CCDB/include/CCDB/CcdbApi.h
```diff
@@ -545,6 +556,8 @@ class CcdbApi //: public DatabaseInterface
   mutable TGrid* mAlienInstance = nullptr;                       // a cached connection to TGrid (needed for Alien locations)
   bool mNeedAlienToken = true;                                   // On EPN and FLP we use a local cache and don't need the alien token
   static std::unique_ptr<TJAlienCredentials> mJAlienCredentials; // access JAliEn credentials
+  int mCurlRetries = 1;
```
costing: Can we start with 3 by default ?

CCDB/src/CcdbApi.cxx
```diff
@@ -1788,7 +1788,11 @@ CURLcode CcdbApi::CURL_perform(CURL* handle) const
   if (mIsCCDBDownloaderEnabled) {
     return mDownloader->perform(handle);
   }
-  return curl_easy_perform(handle);
+  CURLcode result;
+  for (int i = 1; i <= mCurlRetries && (result = curl_easy_perform(handle)) != CURLE_OK; i++) {
+    usleep(mCurlDelayRetries);
```
costing: maybe `mCurlDelayRetries * i` ?

CCDB/include/CCDB/CcdbApi.h
Barthelemy: yes, if that's ok with everyone ? I wanted to have it disabled at first but if you think that it can be useful immediately then 3 seems like a good number.

CCDB/src/CcdbApi.cxx
Barthelemy: good idea, i will change it.

PR #11533, opened 2023-06-21
Removing thread safety breaches from CCDBDownloader

TrifleMichael:
Creating and initializing uv_handles has been moved from the constructor into the uv_loop thread, so to avoid the existing thread safety violations.

Additionally a lock was added as to prevent the constructor from returning before the uv_loop started. This protects the uv_loop from receiving asynchronous handles before it was initialized, which would cause it to crash.

Moreover, the current mechanism for reusing an existing uv_loop (which can be passed via constructor) is not thread safe, and will be removed and reworked in following PRs.

### DIFF
--- a/CCDB/include/CCDB/CCDBDownloader.h
+++ b/CCDB/include/CCDB/CCDBDownloader.h
@@ -218,6 +218,16 @@ class CCDBDownloader
   std::mutex mHandlesQueueLock;
 
   /**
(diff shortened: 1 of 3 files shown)


TrifleMichael:
I found a mistake in the code of this PR - in the case that uv_handles don't close immediately the uv_loop thread may be joined before closing the uv_loop properly.

I'll add a fix after it passes my local tests.

PR #11536, opened 2023-06-21
wip: ITS-Studies: Add proto-MCKineReaderSpec task, TrackChecker skeleton

mconcas:
This PR adds:

- Additional interface to read o2sim_Kine Files on a TF base rather than event one
- Skeleton for Reco/MCSim tracks comparison (i.e. CheckTracksCA.C idea)
- Fixes for unneeded requests such as HBInfos (for now)

This PR needs:

- Better understanding of how we can read MC Kine info and integrate in a DPL-TimeFrame-based approach
- Better understanding and treatment of non-messageable MCHeaders

### DIFF
--- a/Detectors/ITSMFT/ITS/postprocessing/studies/src/MCKinematicReaderSpec.cxx
+++ b/Detectors/ITSMFT/ITS/postprocessing/studies/src/MCKinematicReaderSpec.cxx
@@ -1,0 +1,58 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include <ITSStudies/MCKinematicReaderSpec.h>
+#include <Framework/ControlService.h>
+
+namespace o2
+{
+namespace steer
+{
+using namespace o2::framework;
+
+void KinematicReader::init(o2::framework::InitContext& ic)
+{
+  mMCKinReader->initFromKinematics(std::string(mKineFileName));
+  mNEvents = mMCKinReader->getNEvents(0);
+  LOGP(info, "Initialised KinematicReader");
+}
+
+void KinematicReader::run(ProcessingContext& pc)
+{
+  std::vector<o2::dataformats::MCEventHeader> tfMCHeaders;
+  std::vector<o2::MCTrack> tfMCTracks;
+  for (int iEvent{0}; iEvent < mNEvents; iEvent++) { // Single TF loaded from File, might extend it to use source
+    auto mcHeader = mMCKinReader->getMCEventHeader(0, iEvent);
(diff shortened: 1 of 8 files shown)


Detectors/ITSMFT/ITS/postprocessing/studies/src/MCKinematicReaderSpec.cxx
```diff
@@ -0,0 +1,58 @@
...
+#include <ITSStudies/MCKinematicReaderSpec.h>
+#include <Framework/ControlService.h>
+
+namespace o2
+{
+namespace steer
+{
+using namespace o2::framework;
+
+void KinematicReader::init(o2::framework::InitContext& ic)
```
sawenzel, line 21: Hi @mconcas : 

As you know, there is an existing DPL device that is 99% the same. So in principle I would want to avoid duplicating code and rather add new features there.

Detectors/ITSMFT/ITS/postprocessing/workflow/standalone-postprocessing-workflow.cxx
```diff
@@ -63,12 +63,13 @@ WorkflowSpec defineDataProcessing(ConfigContext const& configcontext)
   o2::globaltracking::InputHelper::addInputSpecsPVertex(configcontext, specs, useMC);
   o2::globaltracking::InputHelper::addInputSpecsSVertex(configcontext, specs);
 
+  // additional sources
+  specs.emplace_back(o2::steer::getMCKinematicReaderSpec());
```
sawenzel, line 67: is it a hard requirement that you need to include devices likes this? Maybe it could be sufficient to do this at runtime using DPL pipes.

mconcas:
Hi @sawenzel , PR is in WIP state for the reason I wanted to discuss with you about having a `framework::task` to do that.

> As you know, there is an existing DPL device that is 99% the same.

If you are referring to: https://github.com/AliceO2Group/AliceO2/blob/dev/run/o2sim_kine_publisher.cxx : it seemed to me that it publishes data relative to the single events with the time granularity of a timeframe, I need to have all events generated for the single timeframe in the same time slice so that I can really compare reconstructed data with simulated ones. I guess this could be easily implemented as a toggleable option in the reader.

> is it a hard requirement that you need to include devices likes this?

I personally find it more convenient, as this is a workflow I give to students/contributors to the ITS studies to launch as a single exe. I already implicitly included other readers to the specs for reconstructed data by using `InputHelper` and `DataRequest` from global tracking, I'd like to stay consistent if possible. 

Perhaps it could be optimal to keep both approaches like it is done for other readers, by splitting the executable from the DPL task and make the executable instantiating that.

In case you would agree, I can also take care of moving these files under `/Steer` and re-adjust the corresponding publisher.

aalkin:
Hello @mconcas, as a matter of fact I'm implementing timeframe aggregation for o2sim_mctracks_proxy with intent to do the same for the kine_publisher. Currently we have several independent pieces of code to go from o2sim to mcheader/mctracks and from that to AOD. We should consolidate this by implementing the conversion step as a library, that can be used to make both an attachable workflow and standalone binaries. Ideally this should work regardless if the data is read from file or injected directly from the simulation.

mconcas:
I am now closing this as I moved to a case-specific usage of the MCKineReader, not requiring a DPL task, after discussing it with @sawenzel.
As @aalkin is taking over with a more comprehensive approach, I'll stick to my simplified one, keeping an eye on the progress and considering it for possible future developments.

PR #11543, opened 2023-06-22
Improved thread safety in construtor and destructor of the downloader

TrifleMichael:
All handle and loop initialization and deletion has been moved from the constructor and destructor to the loop thread.

Additionally, all mechanisms regarding the use of an external uv_loop have been removed, as their use would cause more thread safety violations.

### DIFF
--- a/CCDB/include/CCDB/CCDBDownloader.h
+++ b/CCDB/include/CCDB/CCDBDownloader.h
@@ -114,8 +114,7 @@ class CCDBDownloader
    */
   int mMaxHandlesInUse = 3;
 
-  // CCDBDownloader(uv_loop_t uv_loop);
-  CCDBDownloader(uv_loop_t* uv_loop = nullptr);
+  CCDBDownloader();
   ~CCDBDownloader();
 
   /**
@@ -144,6 +143,9 @@ class CCDBDownloader
    * @param completionFlag Should be set to false before passing it to this function. Will be set to true after all transfers finish.
    */
   std::vector<CURLcode> batchAsynchPerform(std::vector<CURL*> const& handleVector, bool* completionFlag);
+
+  void sendCloseLoopHandle();
+  static void closeLoop(uv_async_t* handle);
 
   /**
    * Limits the number of parallel connections. Should be used only if no transfers are happening.
@@ -182,12 +184,6 @@ class CCDBDownloader
 
  private:
   /**
-   * Indicates whether the loop that the downloader is running on has been created by it or provided externally.
-   * In case of external loop, the loop will not be closed after downloader is deleted.
-   */
-  bool mIsExternalLoop;
-
-  /**
    * Used in debug to detect whether uv loop closed prematurely.
    */
   bool mIsClosing = false;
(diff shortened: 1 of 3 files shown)


ktf (changes requested):
I am not sure I understand the changes. Why do you need a separate thread and a condition variable to initialise the thing?

ktf:
Also, I am completely against the removal of a the external uv_loop, which is a requirement for having this integrated in the multithreaded framework.

TrifleMichael:
> I am not sure I understand the changes. Why do you need a separate thread and a condition variable to initialise the thing?

uv_loop runs in a seperate thread to allow the main thread to continue executing the CcdbApi code. The seperate thread has been there since the first version of the downloader.

The condition variable is used to stop the downloader constructor from returning before the uv_loop is initialized. This protects it from receiving asynchronous handles (via download scheduling) too early. Not doing so would lead to crashes.

TrifleMichael:
> Also, I am completely against the removal of a the external uv_loop, which is a requirement for having this integrated in the multithreaded framework.

The removal of external uv_loop is temporary, as the previous implementation both had multiple thread safety violations, and it required the uv_loop to not be initialized, effectively making it useless.

After this PR is finished I'll start working on adding proper external uv_loop handling.

ktf:
> uv_loop runs in a seperate thread to allow the main thread to continue executing the CcdbApi code. The seperate thread has been there since the first version of the downloader.

I think you should consider running everything on the main thread, using UV_RUN_ONCE / uv_task to avoid blocking whatever CCDB API code you need to run in parallel. I am pretty sure it will result in much simpler and more robust behaviour.

sawenzel:
I would suggest to take these improvements as they are now with the objective to make EPN operations stable. Then one can attack next-level milestones such as external uv_loop.

sawenzel (approved):

ktf:
The problem is not the external uv_loop, the issue is that having uv_loop on a separate thread and synchronising the main thread is a recipe for more headaches. I would remove the mThreadLoop, make sure the uv_loop is always running on the main thread and invoke uv_run(UV_NO_WAIT) whenever you have a chance in the CCDB downloader API. When you need to block to get an actual result, you simply invoke uv_run(UV_NO_WAIT) in a loop, until the handle you are interested in is completed. I think you can probably remove half of the current CCDBDownloader code and synchronization points if you structure it like that.

ktf (changes requested):
I would suggest to fix the design for good, rather than postponing it.

davidrohr:
Unfortunately, this fix did not work. Run 538751 with it shows still the same symptoms:
```
E | 2 | 19:37:47.010 | epn117 |   | DPL | internal-dpl-ccdb-backend | 2g3hoQAnW2V | 538751 | CCDBDownloader: uvloop closed prematurely
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
```
So I think we should either redesign the libuv usage in a safer way, probably what @ktf commented makes sense.
Alternatively, as discussed, we actually don't need the threading at the moment, so if we add an option to just run everything synchronously, only keeping the sockets and tcp connections open with some timeout, that should be ok as well.

PR #11544, opened 2023-06-22
Improve failure propagation/handling between sim service and control …

sawenzel:
…client

https://alice.its.cern.ch/jira/browse/O2-3875


sawenzel:
@costing : All the CCDB tests go banana ...

costing:
The check is reverted so this should be working again.

PR #11545, opened 2023-06-22
Introducing iCell selections

mpuccio:
Remove cells as early as possible in the ITS tracking, reducing the
number of candidate tracks to be fitted


mpuccio:
This will decrease by 15% the computing time for the tracker on the CPU, and it is showing better efficiency and lower fakes in the MC. From this PR we can further optimise the fitting by avoiding refitting common roots of the candidate tree.

mconcas (approved):

PR #11546, opened 2023-06-22
Fix user agent in CCDB Downloader

ktf:


davidrohr (dismissed):
looks reasonableto me.

ktf:
I fixed the conflict.

@costing this should be what is missing for the UserAgent. I am surprised you say it only comes from CI, though. I would have expected everywhere. Did we turn off the CCDB downloader?

PR #11549, opened 2023-06-23
[MCGen] Only complain about status if real primary

benedikt-voelkel:


sawenzel (approved):

benedikt-voelkel:
Error in checks not related to changes. I think this is good to be merged @sawenzel

PR #11556, opened 2023-06-26
DPL Analysis: update AOD producers

aalkin:
@nburmaso @jgrosseo @ddobrigk 
I've added a helper struct that can be used to simplify defining outputs for AOD producers according to data model. I've also cleaned up the producers' code a bit. 

This also should enable the usage of MC AOD producer for on-the-fly simulation. Otherwise AOD creation is not affected. However, @nburmaso, it would be great if you can test and validate this change.


Detectors/AOD/src/AODProducerWorkflowSpec.cxx
```diff
@@ -2697,42 +2694,43 @@ DataProcessorSpec getAODProducerWorkflowSpec(GID::mask_t src, bool enableSV, boo
...
-  addTableToOutput<Cascades>(outputs, 1);
-  addTableToOutput<Collisions>(outputs, 1);
-  addTableToOutput<Decay3Bodys>(outputs);
-  addTableToOutput<FDDs>(outputs, 1);
-  addTableToOutput<FT0s>(outputs);
-  addTableToOutput<FV0As>(outputs);
-  addTableToOutput<StoredFwdTracks>(outputs);
-  addTableToOutput<StoredFwdTracksCov>(outputs);
-  addTableToOutput<McCollisions>(outputs);
+  outputs.push_back(OutputForTable<BCs>::spec());
```
ktf: how about `std::vector<OutputSpec> outputs = {...}`?

Framework/Core/include/Framework/AnalysisHelpers.h
```diff
@@ -101,44 +101,35 @@ struct WritingCursor<soa::Table<PC...>> {
...
     return OutputRef{metadata::tableLabel(), metadata::version()};
   }
 };
 
+/// This helper class allows you to declare things which will be created by a
+/// given analysis task. Notice how the actual cursor is implemented by the
+/// means of the WritingCursor helper class, from which produces actually
+/// derives.
+template <typename T>
+struct Produces : WritingCursor<typename soa::PackToTable<typename T::table_t::persistent_columns_t>::table>, OutputForTable<T> {
```
ktf: As discussed, I am not sure the double inheritance is a good idea, but indeed as we said it is not actually needed and you can use the helper directly in the AnalysisManager where appropriate.

Detectors/AOD/src/AODProducerWorkflowSpec.cxx
aalkin: Done.

Framework/Core/include/Framework/AnalysisHelpers.h
aalkin: Done.

Detectors/AOD/src/AODMcProducerWorkflowSpec.cxx
```diff
@@ -80,7 +72,7 @@ void AODMcProducerWorkflowDPL::fillMCParticlesTable(o2::steer::MCKinematicsReade
         }
       }
       // enumerate saved mc particles and their relatives to get mother/daughter relations
-      for (int particle = 0; particle < mcParticles.size(); particle++) {
+      for (auto particle = 0U; particle < mcParticles.size(); ++particle) {
```
jgrosseo, line 75: I don't mind the ++ vs ++ but doesn't the compiler optimize this away anyway when you do not use the lvalue?

jgrosseo:
Overall it looks OK but there are lots of changes, so we should check indeed if the output is binary equivalent. @aalkin have you already tried it with `prodtests/sim_challenge.sh` ? You can run the last step twice (old and new code) and see if the produced AO2D is identical...

Detectors/AOD/src/AODMcProducerWorkflowSpec.cxx
aalkin: This is a cosmetic change, of course, as it does not matter for the simple type.

aalkin:
@jgrosseo I've tested with `sim_challenge.sh` and the output seems to be identical, as expected.

jgrosseo:
> @jgrosseo I've tested with `sim_challenge.sh` and the output seems to be identical, as expected.

Thanks! For me it is fine then but I think others should also give their blessing. :-)

ktf:
Fine with me as well.

Waiting for @nburmaso to also be on board.

nburmaso:
Fine with me too :)

PR #11557, opened 2023-06-26
Add Mass function wrapper to O2DatabasePDG service

benedikt-voelkel:


benedikt-voelkel:
Thanks to @ktf this is what we came up with to offer a particle mass interface in the `O2DatabasePDG` service

sawenzel (dismissed):

Framework/PhysicsSupport/include/Framework/O2DatabasePDGPlugin.h
```diff
@@ -12,11 +12,21 @@
 #define O2_FRAMEWORK_O2DATABASEPDG_H_
 
 #include "Framework/Plugins.h"
+#include "SimulationDataFormat/O2DatabasePDG.h"
```
ktf: I suspect the issue is that we should hide this in the .cc.

benedikt-voelkel: Thanks, your are right. Done

ktf:
Failure unrelated. Merging.

PR #11558, opened 2023-06-26
ITS-GPU: put standalone version at state of the art

mconcas:
@davidrohr: this PR will update the "standalone gpu" to latest developments (basically what was presented to CHEP).

This code works when we use the former ` TrackerSpec` approach.
I ran it using `GPURecoWorkflowSpec` and indeed the standalone library and the gpu reco compete for managing the GPU memory.

My proposal is that we merge this to put in place state of the art of gpu tracking for ITS, then I later will start to extend it such that I can interface with `GPURecoWorkflowSpec` and integrate with that one piece at the time.


davidrohr:
Fine for me to merge it as is for now. We should just keep the ossibility to run it inside the gpuworkflow operational.
Temporarily, for testing you could simply enforce a small gpu mempry for the gpu reconstruction part, via GPUMEMSIZE env var to dpl-workflow.sh.
And then we have to see that we clean up the memory management asap.

Kind Regards
David Rohr

martenole:
Hi @ktf 
for this PR the fullCI has passed although in the logs I do see some errors:
```
/sw/SOURCES/O2/11558-slc8_x86-64/0/Detectors/ITSMFT/ITS/tracking/GPU/cuda/TrackerTraitsGPU.cu:58:52: warning: use nullptr [modernize-use-nullptr]
  GpuTimer(const int offset, cudaStream_t stream = 0) : mOffset(offset)
                                                   ^
                                                   nullptr
```
Now, other PRs fail the fullCI with the same error: https://github.com/AliceO2Group/AliceO2/pull/11562#issuecomment-1609466194

Do you know how this can happen?

ktf:
I am not sure. Since when do we use the code checker also for the cuda code?

PR #11561, opened 2023-06-27
Lumi scaling options belong to workflow and not to device

shahor02:


martenole:
@shahor02 is this change required to enable the lumi scaling at P2? I am asking because I just said in the RC weekly that we could enable the scaling already now.
And for enabling it I would simply add `TPC_CORR_SCALING="--require-ctp-lumi"` to the env vars

shahor02:
@martenole yes, unfortunately, as a result of a bug the scaling was applied in all tasks using TPC tracks except the gpu-reconstruction...
Note that the fix is on the `dpl-workflow.sh` level only.

martenole:
Ok, thanks. I'll check with RC how they want to proceed (if they agree we can simply overwrite the script from custom O2DPG)

PR #11562, opened 2023-06-27
TRD fix failing transformation test

bazinski:
Remove offending mac arm assert, replace with if statements.
Will investigate properly once build finishes.

[O2-3888](https://alice.its.cern.ch/jira/browse/O2-3888)

### DIFF
--- a/Detectors/TRD/base/src/PadPlane.cxx
+++ b/Detectors/TRD/base/src/PadPlane.cxx
@@ -144,7 +144,14 @@ double PadPlane::getPadRow(double z) const
   }
 
   // sanity check: is the padrow coordinate reasonable?
-  assert(!(padrow < 0.0 || padrow > double(mNrows)));
+  // assert(!(padrow < 0.0 || padrow > double(mNrows)));
+  if (padrow < 0.0) {
+    padrow = 0;
+  } else {
+    if (padrow > double(mNrows)) {
+      padrow = mNrows;
+    }
+  }
 
   return padrow;
 }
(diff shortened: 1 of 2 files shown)


Detectors/TRD/base/src/PadPlane.cxx
```diff
@@ -175,7 +175,7 @@ double PadPlane::getPad(double y, double z) const
     tiltOffsetPad = tiltOffsetY * mInverseWidthIPad;
     pad += tiltOffsetPad;
   }
-
+  LOGP(info, " pad : {} ", pad);
   assert(!(pad < 0.0 || pad > double(mNcols)));
```
martenole: why not just remove the assert and force it to the boundaries? In any case this function seems to be used only in the test macro which fails and causes CI problems now

bazinski: just waiting for a mac arm to finish building, my concern is the error is somewhere else and this is just the manifestation of it. Yes, the plan was if its not obvious will just move pad to the which ever boundary its closest to

bazinski: assert changed to if statements, and will investigate properly when my arm build of O2 finishes.

martenole:
On my laptop this macro always runs without issues. No matter if started in non-compiled mode, compiled mode, or compiled mode with debug symbols:
```c++
#include "TRDBase/PadPlane.h"
#include "TRDBase/Geometry.h"
#endif
void check()
{
  auto mGeo = o2::trd::Geometry::instance();
  mGeo->createPadPlaneArray();
  auto padPlane = mGeo->getPadPlane(0,0);
  auto p = padPlane->getPad(-1000.,0.);
  std::cout << p << std::endl;
}
```
It prints a negative number which should not be the case. So the assert is a no-op. Probably on arm this is not the case.

sawenzel:
I am confused. Is this supposed to fix the test o2-test-trd-Transform? In my case, the test never actually executed because it was missing the ALICE TGeo geometry in the first place. So maybe, in addition, #11563 is needed (but I may have looked into the wrong file).

bazinski:
@martenole agreed with the  < 0, but the tests are just testing the boundaries and should be valid. Why do those tests fail  and only on a mac arm. We should not be getting a negative pad.

PR #11563, opened 2023-06-27
Fix TRD geom transformation test

sawenzel:
Test was not actually testing anything since
TGeo geometry was not loaded.

### DIFF
--- a/Detectors/TRD/base/test/testCoordinateTransforms.cxx
+++ b/Detectors/TRD/base/test/testCoordinateTransforms.cxx
@@ -26,6 +26,7 @@
 #include "TRDBase/PadPlane.h"
 #include "DetectorsBase/GeometryManager.h"
 #include "TRDBase/Geometry.h"
+#include "CCDB/BasicCCDBManager.h"
 
 namespace o2
 {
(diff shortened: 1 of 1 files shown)


Detectors/TRD/base/test/testCoordinateTransforms.cxx
```diff
@@ -43,7 +44,13 @@ void testRCPoint(double calculatedPoint, double predictedPoint)
...
 {
+  // init ALICE geometry --> needed to instantiate TRD Geometry instance
+  auto& ccdbmgr = o2::ccdb::BasicCCDBManager::instance();
+  ccdbmgr.get<TGeoManager>("GLO/Config/GeometryAligned");
+
   auto mGeo = o2::trd::Geometry::instance();
+  BOOST_CHECK(mGeo);
+
   mGeo->createPadPlaneArray();
   mGeo->createClusterMatrixArray();
```
martenole, line 55: Actually, just this line here should be removed which causes the error to be printed. The cluster matrix array is not used in this test so this is not the issue

sawenzel: Yes, I agree. Will close this PR.

sawenzel: TRD should then remove the line `mGeo->createClusterMatrixArray`

bazinski: its removed in #11562

PR #11564, opened 2023-06-27
TPC: Updates in painter

wiechula:
* change namespace to struct for simpler handling in ROOT
* add default colors and markers
* add function to make sparse multigraphs from trees, e.g. to plot vs. run number


wiechula:
can this be merged?

PR #11565, opened 2023-06-27
TRD fix pad coordinate test

martenole:
There are some borderline test cases which expect extremely small values and compare them to zero or values exactly equal to the number of columns. Depending on the precision this can lead to the assert triggered. See p5 (which is 4.4408921e-16 for example), p6, p18 and p19.
I assume the assert is normaly ignored, but not on arm. See comment in https://github.com/AliceO2Group/AliceO2/pull/11562#issuecomment-1609306469
In any case I don't see why we would want to have the assert there in any case so I would remove it.


bazinski:
why does this pr exist given 11562 ?

PR #11572, opened 2023-06-27
Pass --require-ctp-lumi as global, --corrmap-lumi-mean/inst as local opts.

shahor02:


chiarazampolli:
Hello @shahor02 ,

can this be forced-merged? I don’t think it should be a problem.

chiara

shahor02:
Errors unrelated, tested locally

chiarazampolli:
Thanks!

PR #11575, opened 2023-06-28
Making CCDBDownloader single threaded

TrifleMichael:
The CCDBDownloader was modified based on @ktf suggestions.

Detailed changes:
- Removed the worker thread from CCDBDownloader. Downloads are now executed within the perform() function.
- Restored running the Downloader on an external uv_loop. Added appropriate test.
- The socket keepalive mechanism is active only if uv_loop is passed to the constructor.
- Temporarily removed asynchronous downloading functions.
- Temporarily removed asynchronous callbacks.
- Fixed error checking in the test_with_break.

Additional note:
If an external loop is passed to the constructor, then after performing the Downloader cannot be left without either running it's uv_loop after the keepalive time passes, or calling the destructor. Not adhering to that will result in sockets not being closed after appropriate time.


ktf:
Without looking to much into the details, this looks fine for me. The only remaining bit would be to expose runLoop() from the CCDBApi so that I can have the socket closed by running your loop while idling.

davidrohr:
tested locally, and solved trivial merge conflic, merging

PR #11577, opened 2023-06-28
DPL: optimize state flushing

ktf:


ktf:
@knopers8 could you check if this improves the performance of the dispatcher?

PR #11578, opened 2023-06-28
O2sim: Add rate limiting to mctracks proxy

aalkin:
Adds rate limiting to mctracks-proxy. Adds sending EoS to forwarding channel if o2sim quits earlier. Kine forward channel configuration is changed to `pair` to avoid dropping events. Note that proxy sends a single event per "timeframe". Rate limiting needs to be set in terms of number of events that are to be kept in memory by the proxy.

### DIFF
--- a/run/o2sim_parallel.cxx
+++ b/run/o2sim_parallel.cxx
@@ -41,6 +41,8 @@
 #include <unordered_map>
 #include <filesystem>
 #include <atomic>
+#include "Framework/SourceInfoHeader.h"
+#include "Headers/Stack.h"
 
 #include "SimPublishChannelHelper.h"
 #include <CommonUtils/FileSystemUtils.h>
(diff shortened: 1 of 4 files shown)


run/o2sim_parallel.cxx
```diff
@@ -89,6 +91,24 @@ void remove_tmp_files()
...
+    fair::mq::MessagePtr payload(forwardchannel.NewMessage());
+    o2::framework::SourceInfoHeader sih;
+    sih.state = o2::framework::InputChannelState::Completed;
+    auto channelAlloc = o2::pmr::getTransportAllocator(forwardchannel.Transport());
+    auto header = o2::pmr::getMessage(o2::header::Stack{channelAlloc, sih});
+    parts.AddPart(std::move(header));
+    parts.AddPart(std::move(payload));
+    forwardchannel.Send(parts);
+    LOGP(info, "SENDING END-OF-STREAM TO PROXY AT {}", address.c_str());
+  }
```
sawenzel, line 111: couldn't we have this in the HitMerger (which also knows when all
events are done)? Or do you target EOS even in case of failures here?

aalkin: Indeed, this way it will be sent even in case of failures so the proxy will not be stuck if o2sim was killed. We can also potentially avoid event counting in the proxy, but I did not touch that for now.

sawenzel (approved):

PR #11579, opened 2023-06-28
Propagate maximum allowed CPU and memory usage threshold to ECS

knopers8:
...which can then kill a given task when the limits are crossed.


knopers8:
cc @teo

ktf:
The general idea is fine with me. Please check the unhandled switch case.

knopers8:
it should be ok now

knopers8:
@ktf would you mind merging it?

PR #11584, opened 2023-06-29
ctpdev: CTP RunManager class in separate file

lietava:


lietava:
@shahor02 : I think the problem is in qc:
https://github.com/AliceO2Group/QualityControl/blob/a9eab82e920ddf6405dbcf0eb41dbae8926e862b/Modules/CTP/include/CTP/CountersQcTask.h#L99
where the CTPRunManager is used.
Configuration.h which is included in CountersQcTask.h , does not have CTPRunManager anymore.
I see 2 ways how to proceed:
1.) Revert and keep CTPRunManager in Configuration.h
2.) #include RunManager.h in CountersQcTask.h
3.) #include RunManager.h in Configuration.h
PLease, let me know what do you suggest.
Cheers, Roman.

shahor02:
Hi @lietava 

Could you add `#include "DataFormatsCTP/RunManager.h"` at the END of the Configuration.h, e.g. after https://github.com/AliceO2Group/AliceO2/blob/04a7bdd3246032f842f5e2e355450b1622c64c82/DataFormats/Detectors/CTP/include/DataFormatsCTP/Configuration.h#L199
This should satisfy the QC dependencies. If it works and CI passes, we will merge this PR, then you should change the dependency of the QC from Configuration.h to RunManager.h, and once this is merged, remove fake include from the Configuration.h.

lietava:
@shahor02 : now it seems ok. remaining error is not related to ctp ? R.

shahor02 (approved):
@lietava merging this, now you can change the QC dependency to RunManager, and once it is merged, undo the trick with including RunManager.h in the end of the Configuration.h

PR #11586, opened 2023-06-29
DPL: do not publish metrics for the GUI so often

ktf:


davidrohr:
Is this for the remote or the local GUI?
I am not sure if this is ideal, if we update the GUI only twice every second, it might feel less responsive :(

ktf:
Indeed, I do not like the result...

ktf:
I will make sure the metrics are only sent if the GUI is actually active.

davidrohr:
Can we do something more clever, i.e. checking if the value has changed, and publishing only those?
Or could we count how many values we update in total, and adjust the publishing interval accordingly?

ktf:
This (publishing the changed values after 200ms) is already what happens. However for some workflows this still seems to be to much. Hopefully #11577 will be enough for now. Anyway, I will profile a bit and come back at this.

PR #11601, opened 2023-07-03
DPL: move the nextTimeslice to the decongestion service

ktf:
While for the moment this is just reshuffling, it will allow handling the case in which a Lifetime::Timeframe is dropped.

### DIFF
--- a/Framework/Core/src/DecongestionService.h
+++ b/Framework/Core/src/DecongestionService.h
@@ -20,6 +20,8 @@ struct DecongestionService {
   bool isFirstInTopology = true;
   /// Last timeslice we communicated. Notice this should never go backwards.
   int64_t lastTimeslice = 0;
+  /// The next timeslice we should consume, when running in order
+  int64_t nextTimeslice = 0;
   // Task to enqueue the oldest possible timeslice propagation
   // at the end of any processing chain.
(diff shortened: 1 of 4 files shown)


ktf:
@davidrohr I think it's actually easier than we anticipated, no need to have an extra context / callback for the CompletionPolicy. We just move the nextTimeslice to the DecongestionService and get it from the ServiceRegistry. Then we can account as we wish in the .domainInfoUpdated callback.

Framework/Core/src/DecongestionService.h
```diff
@@ -20,6 +20,8 @@ struct DecongestionService {
   bool isFirstInTopology = true;
   /// Last timeslice we communicated. Notice this should never go backwards.
   int64_t lastTimeslice = 0;
+  /// The next timeslice we should consume, when running in order
+  int64_t nextTimeslice = 0;
```
davidrohr, line 24: makes sense to me, I would make it a bit more clear here that this is only for the ordered completion policy.
What I like about this in particular is that we can reset them together for START/STOP/START, and we don't need an extra reset for the completion policy.

ktf: Exactly.

PR #11602, opened 2023-07-03
Feature/o2 monitor occupancy

sidibu:
Implemented the possibility to set a value threshold for the O2 monitor as well as plotting the occupancy for the given signals @wiechula.

### DIFF
--- a/Detectors/TPC/monitor/include/TPCMonitor/SimpleEventDisplay.h
+++ b/Detectors/TPC/monitor/include/TPCMonitor/SimpleEventDisplay.h
@@ -63,6 +63,8 @@ class SimpleEventDisplay : public CalibRawBase
 
   void setPedstals(CalPad* pedestals) { mPedestals = pedestals; }
 
+  void setSignalThreshold(UInt_t signalThreshold) { mSignalThreshold = signalThreshold; }
+  
   TH1D* makePadSignals(Int_t roc, Int_t row, Int_t pad);
 
   /// set time bin range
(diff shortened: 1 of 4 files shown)


Detectors/TPC/monitor/include/TPCMonitor/SimpleEventDisplay.h
```diff
@@ -77,23 +83,26 @@ class SimpleEventDisplay : public CalibRawBase
...
+  Int_t mLastSelSector;    //!< Last sector selected for processing
+  Int_t mCurrentRow;       //!< current row processed
+  Int_t mCurrentPad;       //!< current pad processed
+  Float_t mMaxPadSignal;   //!< maximum bin of current pad
+  Int_t mMaxTimeBin;       //!< time bin with maximum value
+  Bool_t mSectorLoop;      //!< only process one sector
+  Int_t mFirstTimeBin;     //!< first time bin to accept
+  Int_t mLastTimeBin;      //!< last time bin to accept
+  UInt_t mSignalThreshold; //!< minimum adc value
+  Bool_t mShowOccupancy;   //<! true iff occupancy should be calculated, false otherwise
```
wiechula, line 105:
```suggestion
  Bool_t mShowOccupancy;   //!< true iff occupancy should be calculated, false otherwise
```

Detectors/TPC/monitor/src/SimpleEventDisplay.cxx
```diff
@@ -124,26 +127,34 @@ Int_t SimpleEventDisplay::updateROC(const Int_t roc,
...
 
-  if (corrSignal > val) {
+  if (corrSignal > val && corrSignal >= mSignalThreshold) {
     calROC.setValue(row, pad, corrSignal);
     mMaxPadSignal = corrSignal;
     mMaxTimeBin = timeBin;
   }
+
+  CalROC& calROCOccupancy = mPadOccupancy.getCalArray(mCurrentROC);
+  auto occupancy = calROCOccupancy.getValue(row, pad);
```
wiechula, line 152:
```suggestion
  const auto occupancy = calROCOccupancy.getValue(row, pad);
```

wiechula (changes requested):
Just a few minor things.

PR #11606, opened 2023-07-03
DPL: attempt at supporting multiple output proxies

ktf:
The basic idea is that now that we can do early forwarding, we can just use that in the output proxy, and multiple routes matching the same input will simply be shallow copied (if in the same transport) or deep copied (when the transport is different).

### DIFF
--- a/Framework/Core/include/Framework/FairMQDeviceProxy.h
+++ b/Framework/Core/include/Framework/FairMQDeviceProxy.h
@@ -55,7 +55,7 @@ class FairMQDeviceProxy
   /// Retrieve the channel index from a given OutputSpec and the associated timeslice
   [[nodiscard]] ChannelIndex getOutputChannelIndex(OutputSpec const& spec, size_t timeslice) const;
   /// Retrieve the channel index from a given OutputSpec and the associated timeslice
-  [[nodiscard]] ChannelIndex getForwardChannelIndex(header::DataHeader const& header, size_t timeslice) const;
+  void getMatchingForwardChannelIndexes(std::vector<ChannelIndex>& result, header::DataHeader const& header, size_t timeslice) const;
   /// ChannelIndex from a RouteIndex
   [[nodiscard]] ChannelIndex getOutputChannelIndex(RouteIndex routeIndex) const;
   [[nodiscard]] ChannelIndex getInputChannelIndex(RouteIndex routeIndex) const;
--- a/Framework/Core/src/DataProcessingDevice.cxx
+++ b/Framework/Core/src/DataProcessingDevice.cxx
@@ -553,7 +553,7 @@ static auto toBeForwardedHeader = [](void* header) -> bool {
   return true;
 };
 
-static auto toBeforwardedMessageSet = [](ChannelIndex& cachedForwardingChoice,
+static auto toBeforwardedMessageSet = [](std::vector<ChannelIndex>& cachedForwardingChoices,
                                          FairMQDeviceProxy& proxy,
                                          std::unique_ptr<fair::mq::Message>& header,
                                          std::unique_ptr<fair::mq::Message>& payload,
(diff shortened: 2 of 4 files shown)


ktf:
Not sure this is the whole story (i.e. if having multiple matching forward routes will actually work), but worth trying, I guess.

davidrohr:
I just tried this with the global reco+calib workflow, but it seems not to work. The workflow comes up now that the error message is removed, but it seems only 1 of the 2 calib workflows is receiving the CTP/DIGITS.
The TPC aggregator-workflow drops all time frames since CTP/DIGITS are missing, throwing "Dropping lifetime:timeframe... as it can never be completed" messages.

ktf:
I have updated this with a version which works with my simple reproducer. I.e. I see the same number of messages being pushed to both channels. In order for this to work, the channels must be named with the same prefix. I.e. if before you were using "name=downstream" you can now push data to more than one channel by calling it "name:downstream1,...;name=downstream=2" etc. A working example for this is:

```
o2-testworkflows-simple-source --dataspec a:TST/A1/0 | \
o2-dpl-output-proxy --dataspec downstream:TST/A1/0 --dpl-output-proxy "--channel-config 'name=downstream,transport=zeromq,method=bind,type=pub,address=tcp://0.0.0.0:1234,rateLogging=1;name=downstream2,transport=zeromq,method=bind,type=pub,address=tcp://0.0.0.0:1235,rateLogging=1'"  --severity detail
```

If you have troubles we can sit together tomorrow and iron out the remaining issues, if any.

davidrohr:
But wait, this is not exactly what we are doing. We have multiple dpl-output-proxy devices in the merged workflow. They are sending different dataspecs, but the dataspecs are partially overlapping.

ktf:
In principle that should work as well now, albeit I admittedly did not test. Given apparently I broke one of the unit tests, I will start committing the ancillary bits in a separate PR and test also the "multiple output proxy". That said, I wonder if it does not make more sense to simply have a single output proxy redirecting to multiple channels.

ktf:
Ok, I have adapted the code to only support multiple output proxies (not yet multiple output routes, which we keep for a later development) and verified it works with:

```
stage/bin/o2-testworkflows-simple-source --dataspec a:TST/A1/0 | \
stage/bin/o2-dpl-output-proxy --proxy-name out1 --dataspec downstream:TST/A1/0 --out1 "--channel-config 'name=downstream,transport=zeromq,method=bind,type=pub,address=tcp://0.0.0.0:1234,rateLogging=1'" --severity detail |\
stage/bin/o2-dpl-output-proxy --proxy-name out2 --dataspec downstream:TST/A1/0 --out2 "--channel-config 'name=downstream,transport=zeromq,method=bind,type=pub,address=tcp://0.0.0.0:1235,rateLogging=1'" --severity detail
```

martenole:
Hi, all checks seems to have passed, is this ready to be merged? Would it solve both problems we have with CTP at the moment? Meaning (if I understand correctly)
1) on the FLP the raw data is used both by the decoder and by QC. This does not work, since only one device will get the data. That's why currently CTP QC workflow is disabled
2) for the calib workflows we would like several aggregator proxies to receive the CTP digits

davidrohr:
I think we should test it in a local FST with calib before, and check for new error messages in the logs.
Last time it broken TPC calib, and there were new Dropping Lifetime::Timeframe meesages.

davidrohr:
To me, it looks like it is still not working. In the FST with CALIB, I am getting e.g.
```
[1656483:residual-aggregator]: [17:48:14][ERROR] Dropping incomplete GLO/TRKREFS/0 Lifetime::timeframe data in slot 0 with timestamp 12 < 13 as it can never be completed.
```

ktf:
I suspect at the moment we rely on only the first route being used for a matching input spec. With this PR, it would result in a duplicate message. I can have a look in a bit/ tomorrow.

ktf:
With 654c2687e4993545dcd6a7d8d9d0d82eec722790 the behavior should be exactly the same as before (including the fact it does not support multiple routes). Can we verify that?

ktf:
Ok, I have moved the version with the old behaviour in #11788 and this version takes the unique set of all the forwarding channels, so the messages should be forward only once, even if they match multiple routes because of wildcards or alike.

ktf:
Updated. This version will only consider one DPL channel and as many as present non dpl ones.

ktf:
Nevermind, this is the actual fix I meant to do. We can have multiple output routes to the internal topology, of course, but a given set of messages should be forwarded to only one internal route (and as many external ones as requested).

ktf:
This should work also in the case the consumeWhenAll policy is provided for the output proxy. Before I was forcing the early forwarding, resulting in completion policies being effectively ignored.

davidrohr:
ok, makes sense. recompiling and testing

davidrohr:
Still doesn't work, both the TPC residual aggregator task and the EMCAL calib task report "Dropping lifetime::timeframe"

ktf:
This preserves the order in which the channels get forwarded. Not sure why it would matter, though.

ktf:
QC and macos issues unrelated. Merging as agreed during our debug session.

PR #11607, opened 2023-07-04
DPL GUI: add signposts to profiles

ktf:


ktf:
Trivial and not checked in the CI. Merging.

PR #11608, opened 2023-07-04
O2sim: Aggregating mode for mctracks proxy

aalkin:
* FairMQDeviceProxy: send TF counter metrics if Monitoring service is available
* FairMQDeviceProxy: message aggregation function added
* mctracks-proxy: added `--aggregate-timeframe <number>` option to combine `<number>` of incoming events into a single multipart message and require a Monitoring service so that TF counter can be sent and TF-based rate limiting works
* mctracks-to-aod: updated to handle the aggregated messages
* mctracks-to-aod: removed the redundant spawner task (analysis workflows add aod-spawner automatically)
* example script: updated to use `--aggregate-timeframe`

@sawenzel @mconcas

### DIFF
--- a/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
+++ b/Framework/Core/src/ExternalFairMQDeviceProxy.cxx
@@ -425,7 +425,8 @@ DataProcessorSpec specifyExternalFairMQDeviceProxy(char const* name,
                                                    std::vector<OutputSpec> const& outputs,
                                                    char const* defaultChannelConfig,
                                                    InjectorFunction converter,
-                                                   uint64_t minSHM)
+                                                   uint64_t minSHM,
+                                                   bool sendTFcounter)
 {
   DataProcessorSpec spec;
   spec.name = strdup(name);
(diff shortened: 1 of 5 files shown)


sawenzel (dismissed):
Thanks for this development. Looks good to me. Maybe @DanielSamitz can also take a look?

run/o2sim_mctracks_proxy.cxx
```diff
@@ -63,38 +64,88 @@ class ConsumerTask
...
+      headerDH.subSpecification = headerMatcher.subSpec;
+      headerDH.payloadSize = headerSize;
+      headerDH.payloadSerializationMethod = gSerializationMethodROOT;
+      headerDH.splitPayloadParts = nPerTF;
+      headerDH.splitPayloadIndex = *eventCounter;
+      o2::header::Stack hhs{headerDH, hdph};
+
+      DataProcessingHeader tdph{*TFcounter, 0};
+      ConcreteDataMatcher tracksMatcher = DataSpecUtils::asConcreteDataMatcher(specs[1]);
+      tracksDH.dataOrigin = tracksMatcher.origin;
```
ktf, line 122: nitpick: I would use the
```c++
DataHeader tracksDH {
 .dataOrigin = ...,
 .dataDescription = ...
};
```

syntax.

DanielSamitz:
Thank you, this is great. In principle we are missing something similar in the [o2sim-kine-publisher](https://github.com/AliceO2Group/AliceO2/blob/dev/run/o2sim_kine_publisher.cxx), but I'd say that there it is not so important, because there it is not necessary to do the conversion McTracks --> McParticles on the fly.

aalkin:
@DanielSamitz there is `AODMcProducer` that is more useful for creating AOD from kine files. 

I can update the kine-publisher as well, in a separate PR, if it is actually used.

Framework/Core/src/ExternalFairMQDeviceProxy.cxx
```diff
@@ -598,6 +624,9 @@ DataProcessorSpec specifyExternalFairMQDeviceProxy(char const* name,
               timingInfo.creation = dph->creation;
             }
             dataHandler(timingInfo, parts, 0, ci);
+            if (ctx.services().active<o2::monitoring::Monitoring>()) {
```
ktf: Can you please use a boolean flag which defaults to false so that we do not change the behaviour for all the proxies we have? I fear with Monitoring the overhead of sending this message online is too high.

aalkin: Done.

sawenzel (approved):

PR #11610, opened 2023-07-04
DPL: do not bind any port during workflow generation

ktf:


ktf:
@knopers8 this might be what you need. It's still not clear to me why you cannot dump the workflow once and use it everywhere on the cluster like we do for DDS.

knopers8 (approved):
I confirm that I do not see a port bind attempt for the remote GUI anymore with this PR. Thanks!

PR #11612, opened 2023-07-05
Leftovers...from ZDC geometry -> PR needed

coppedis:
I don't know how it could have happened, but I realized today that few fixes for geometry (important to avoid overlaps in the geometry) were not ported! I would like to have them included...if we do not clash with any other important stuff.
Thanks, Chiara


shahor02 (commented):
Hi @coppedis 

I don't understand this PR. The changes for Geometry.h are already in dev, while the diff of Detector.cxx is applied to line which are completely different in the current dev:
https://github.com/AliceO2Group/AliceO2/blob/c29cab0f8c5403bbbc8638b7c2e1c1de90b0d000/Detectors/ZDC/simulation/src/Detector.cxx#L738-L743

Could you please check?

Cheers, 
 Ruben

coppedis:
@shahor02 
You are right...changes are already in...sorry I got confused by the web interface that was telling me that these commit were not merged. I close this PR.

PR #11613, opened 2023-07-05
[FOCAL-12] Fix operator==

mfasDa:
Condition was inverted by accident,
for channel-by-channel check. In
addition adding getter for number
of channels.


mfasDa:
CI error unrelated.

PR #11616, opened 2023-07-05
DPL: use unbelievable value for nOrbitsPerTF if it is not set

knopers8:
Now it is blindly assumed that nOrbitsPerTF is 128, while at P2 we are already running with 32. I would propose to set it to a less believable value, so no one assumes it is correct until it is actually set e.g. via FairMQProperty.

### DIFF
--- a/Framework/Core/include/Framework/DataTakingContext.h
+++ b/Framework/Core/include/Framework/DataTakingContext.h
@@ -33,7 +33,7 @@ struct DataTakingContext {
   /// The current run number
   std::string runNumber{UNKNOWN};
   /// How many orbits in a timeframe
-  uint64_t nOrbitsPerTF = 128;
+  uint64_t nOrbitsPerTF = 0;
   /// The start time of the first orbit in microseconds(!)
   long orbitResetTimeMUS = 0;
   /// The current lhc period
--- a/Framework/Core/src/CommonServices.cxx
+++ b/Framework/Core/src/CommonServices.cxx
@@ -215,7 +215,7 @@ o2::framework::ServiceSpec CommonServices::datatakingContextSpec()
       auto forcedRaw = services.get<RawDeviceService>().device()->fConfig->GetProperty<std::string>("force_run_as_raw", "false");
       context.forcedRaw = forcedRaw == "true";
 
-      context.nOrbitsPerTF = services.get<RawDeviceService>().device()->fConfig->GetProperty<uint64_t>("Norbits_per_TF", 128); },
+      context.nOrbitsPerTF = services.get<RawDeviceService>().device()->fConfig->GetProperty<uint64_t>("Norbits_per_TF", 0); },
     .kind = ServiceKind::Stream};
 }
 


ktf:
I have no objections, but @shahor02 or @davidrohr should probably comment here if they are ok. Maybe using -1 as a value it's better?

davidrohr:
I think in general it is correct to initialize with a bogus value, to avoid using a wrong default. However, there are 2 things I would like to comment:
1. there are nHbfPerTf defaults in many places, all still 128. We should change all of them to 32.
2. Who initializes the value in the DataTakingContext, and who uses it. Normally, nHbfPerTf should come from the GRP during processing, and from HBFUtils for simulation/digitization. Why do we have a 3rd source here? I would propose we initialize it automatically from one of the 2 others?

knopers8:
> I have no objections, but @shahor02 or @davidrohr should probably comment here if they are ok. Maybe using -1 as a value it's better?

Since the value is uint, i fear comparing it to `-1` might pose more risk.

> there are nHbfPerTf defaults in many places, all still 128. We should change all of them to 32.

Actually it was not clear to me that `nHbfPerTf` and `nOrbitsPerTF` are by definition the same until this morning. I am not sure though if changing the default to 32 is the way to go, if we might use a given piece of code to reconstruct old data.

> Who initializes the value in the DataTakingContext, and who uses it. Normally, nHbfPerTf should come from the GRP during processing, and from HBFUtils for simulation/digitization. Why do we have a 3rd source here? I would propose we initialize it automatically from one of the 2 others?

Sounds fair. This one is initialized as a fairmq property and I am OK to modify ECS to propagate it there, but indeed having one less source of this information would be safer.

davidrohr:
yes, hbfPerTf and orbitPerTf are the same.
Probably we should rename it to be unique.

Well, I meant I would change the HBFUtils default, and related stuff, since it is used for digitization / simulation.
Reconstruction should then pick up the value automatically, and not have 32 as default.

OK, now I understand where it comes from. We had this property before we had CCDB and GRP, so we used it to set this setting, but I think this is more legacy code. Ideally we should remove this one I'd say, and take it from GRP.

knopers8:
OK, makes sense.

I do use this one in QC (fortunately only experimentally), so I'll have to switch to using GRP, but then I can make a PR to remove this one. In O2 it is not used anywhere:
```
~/alice/O2$ grep -rn ./ -e 'nOrbitsPerTF'
./Framework/Core/include/Framework/DataTakingContext.h:36:  uint64_t nOrbitsPerTF = 0;
./Framework/Core/src/CommonServices.cxx:218:      context.nOrbitsPerTF = services.get<RawDeviceService>().device()->fConfig->GetProperty<uint64_t>("Norbits_per_TF", 0); },
./Detectors/TPC/workflow/test/test_ft_EPN_Aggregator.cxx:180:  const int nOrbitsPerTF = 128; // number of orbits per TF
./Detectors/TPC/workflow/test/test_ft_EPN_Aggregator.cxx:182:  const int nIDCs = nOrbitsPerTF / nOrbitsPerIDC;
./Detectors/TPC/simulation/include/TPCSimulation/IDCSim.h:56:  /// \param nOrbitsPerTF number of orbits per TF
./Detectors/TPC/simulation/include/TPCSimulation/IDCSim.h:57:  static void setNOrbitsPerTF(const unsigned int nOrbitsPerTF) { mOrbitsPerTF = nOrbitsPerTF; }
```
So it's fine for me if we close this one.

davidrohr:
great, if it is not used in O2 at all, and if you can switch to using the GRP in QC, I agree to close this, and them remove it in a follow-up once QC is switched.
If you want, you can also do a string replace in O2 and QC from OrbitsPerTF to HBFPerTF :)

PR #11617, opened 2023-07-05
CalibLaserTracksSpec: protect for the case you get a condition object,

ktf:
but not the actual tracks


shahor02:
Hm, OK, but this device gets as an input either TPC reco tracks or they filtered subsample, but always at every TF and the input has default Lifetime::Timeframe.

How it can start processing w/o getting the input?

ktf:
i agree it should not happen. there is no custom completion policy,
correct?

Ciao,
Giulio

PR #11622, opened 2023-07-06
DPL: Remove nOrbitsPerTF from DataTakingContext

knopers8:
as agreed in https://github.com/AliceO2Group/AliceO2/pull/11616, the field is removed since it is not actually set and the recommended way is to get it from GRP Helper or CCDB directly

will fail in QC until this is merged and propagated to a version in alidist: https://github.com/AliceO2Group/QualityControl/pull/1873


knopers8:
it should be good to go now

PR #11624, opened 2023-07-06
DPL: do not use arrow service if at P2

ktf:


ktf:
I think the real issue is https://github.com/AliceO2Group/AliceO2/pull/11624/commits/4e8a82e2b279db098ad24e4e854aaa06eadefa2b, i.e. a maxRefreshLatency of 0 will publish the metrics every time 0 ms have passes, even if they are unchanged. I must have changed them to 0 while doing the changes for the analysis rate limiting and forgot / did not think to change them to a more reasonable number. In teory it could even be larger, since those metrics are always published when they get updated.

knopers8 (approved):

ktf:
Tested with the gang of 5 on hyperloop. Merging.

PR #11626, opened 2023-07-07
Change TPC IDC & SAC proxy ports

knopers8:
The old ports are in the range of ephemeral ports on our machines, thus there is a risk that we might try to bind 47900 and 47901 when they are already used.


davidrohr:
this is fine, but we should not merge it now, but when we make the next global sw update at P2, to deploy in parallel with the necessary changes in the FLP workflows.

vascobarroso:
@davidrohr we plan to release a new FLP Suite later this week. Could you ensure that this is part of one of the upcoming O2 tags ? Deadline is Thursday

davidrohr:
sure

PR #11629, opened 2023-07-07
DPL: remove wrong debug assertion

ktf:


ktf:
Was introduced to verify some condition when debugging the rate limiting but it's actually obviously wrong.

PR #11632, opened 2023-07-07
O2sim: update mctracks-proxy and mctrack-to-aod

aalkin:
* add a helper function to create DataHeader for OutputSpec
* handle cases where the total number of events requested is not divisable by the number of events in aggregated timeframe
* correctly fill mc particle flags


aalkin:
Hi @sawenzel, could you please merge this?

PR #11634, opened 2023-07-08
DPL: do not update / flush metrics which are not enabled

ktf:


ktf:
Tested also on hyperloop. Merging because I like the feature in any case. #11633 can either be merged as is, or translated to use this feature.

PR #11639, opened 2023-07-10
Apply MFT tracker high mult protection always in synch mode

martenole:
We see from time to time the MFT tracking being too slow on individual TFs (up to 30 minutes) if it encounters a ROF with many clusters. Currently a protection is only added in case ITS is not included in the run. For the synch. reco I think the protection should always be there.
More details in https://alice.its.cern.ch/jira/browse/O2-3740

pinging @robincaron13

Since I am not sure if I am missing something (it is not clear to me why the protection was added like this in the first place) I'll wait for approval before merging


martenole:
Pinging to @shahor02 
FYI: I hope it makes sense to have the protection for MFT added for all online processing?

shahor02:
Should be fine, though I would prefer a protection on the seeds per cluster or something like that rather than absolute mult.

PR #11640, opened 2023-07-10
[EMCAL-834] Checking active links per trigger

mfasDa:
For each trigger checking from which DDLs the
reconstruction received pages. If we don't receive
pages from all active links the page is discarded.

### DIFF
--- a/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/ErrorTypeFEE.h
+++ b/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/ErrorTypeFEE.h
@@ -57,6 +57,7 @@ class ErrorTypeFEE
     FIT_ERROR,         ///< Raw fit failed
     GEOMETRY_ERROR,    ///< Decoded position outside EMCAL
     GAIN_ERROR,        ///< Error due to gain type
+    LINK_ERROR,        ///< Error due to missing DDL links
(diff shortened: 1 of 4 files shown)


mfasDa:
Re-introducing commit reverted in https://github.com/AliceO2Group/AliceO2/pull/11266. We will need this commit in order to drop timeframes where not all links are present, leading to inconsistent data.

Indeed there might be a prroblem with synthetic data, in case there are no pages created for links where we don't have data (@shahor02) - i this case the link checker is indeed supposed to fail. We might need to force writing empty pages in case of simulated raw data.

The crash reported by David points to a check for non-0 range in the framework:
```
[3679048:calib-emcalchannel-badchannel]: [09:44:47][FATAL] Unhandled o2::framework::runtime_error reached the top of main of o2-calibration-emcal-channel-calib-workflow, device shutting down. Reason: deque::_M_range_check: __n (which is 0)>= this->size() (which is 0)
```
In this case the crash would also appear for true positives since in such cases we send emtpy vectors and would not pass the range check (@ktf). The framework must support also empty containers for such cases.

mfasDa:
I adapted the code in a way that it can be switched off via a config option and by default is switched off in case of simulated raw data. Like this it should be safe for running on synthetic raw.

mfasDa:
Slight modification: Task now sending a raw data error to the QC for each missing link

PR #11641, opened 2023-07-10
[EMCAL-742] Add rejection of EMCal L0 triggered events in calib histo exporter

jokonig:
- For bad channel and time calibration, only minimum bias triggered data should be used
- As the EMCal started taking data with L0 triggers in 2023, L0 triggered events have to be rejected
- Information taken from CTP input (CTPDigits)
- Selected trigger names are settable in the calib params
- This does not effect the online calibration!


mfasDa:
CI errors unrelated

mfasDa (approved):

PR #11642, opened 2023-07-10
DPL: consider dummy-sink dependent on everything so it is always put at the end of the workflow

aalkin:
@ktf as discussed


ktf:
The change itself seems fine, I tested on hyperloop as well with a couple of workflows and they were ok. I could not do the full round of tests because I need to adapt my workflows to the recent changes in the datamodel, but it should be fine.

PR #11644, opened 2023-07-10
TRD add additional helper functions

pestratm:
Added some additional helper functions for TRD that @tdietel provided in his macros
@martenole @pachmay

### DIFF
--- a/DataFormats/Detectors/TRD/src/Tracklet64.cxx
+++ b/DataFormats/Detectors/TRD/src/Tracklet64.cxx
@@ -43,6 +43,35 @@ GPUd() int Tracklet64::getPadCol() const
   return CAMath::Nint(6.f + mcmCol * ((float)NCOLMCM) + padLocal);
 }
 
+GPUd() float Tracklet64::PadPositionMCM() const
+{
+  return 12.0 - (getPositionBinSigned() * GRANULARITYTRKLPOS);
+}
+
+GPUd() float Tracklet64::PadPosition() const
+{
+  // int padLocalBin = tracklet.getPosition() ^ 0x80;
+  float padMCM = PadPositionMCM();
+  int mcmCol = HelperMethods::getMCMCol(getROB(), getMCM());
+  return float((mcmCol + 1) * NCOLMCM) + 2.0 - padMCM;
+}
+
+GPUd() float Tracklet64::UncalibratedPad() const
+{
+  float y = getUncalibratedY();
+  int mcmCol = (getMCM() % NMCMROBINCOL) + NMCMROBINCOL * (getROB() % 2);
+  // one pad column has 144 pads, the offset of -63 is the center of the first MCM in that column
+  // which is connected to the pads -63 - 9 = -72 to -63 + 9 = -54
+  // float offset = -63.f + ((float)NCOLMCM) * mcmCol;
+  float padWidth = 0.635f + 0.03f * (getDetector() % NLAYER);
+  return y / padWidth + 71.0;
+}
+
+GPUd() float Tracklet64::Slope() const
+{
+  return getSlopeBinSigned() * GRANULARITYTRKLSLOPE / ADDBITSHIFTSLOPE;
+}
+
 #ifndef GPUCA_GPUCODE_DEVICE
 void Tracklet64::printStream(std::ostream& stream) const
 {
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/HelperMethods.h
@@ -150,9 +150,9 @@ struct HelperMethods {
 
   static int getLinkIDfromHCID(int hcid)
   {
-    //return a number in range [0:29] for the link related to this hcid with in its respective CRU
-    //lower 15 is endpoint 0 and upper 15 is endpoint 1
-    //a side has 30, c side has 30 to give 60 links for a supermodule
+    // return a number in range [0:29] for the link related to this hcid with in its respective CRU
+    // lower 15 is endpoint 0 and upper 15 is endpoint 1
+    // a side has 30, c side has 30 to give 60 links for a supermodule
     int sector = hcid / constants::NHCPERSEC;
     return getORIinSuperModule(hcid) + constants::NHCPERSEC * sector;
   }
@@ -203,6 +203,12 @@ struct HelperMethods {
     rob = (mcmCol >= constants::NMCMROBINCOL) ? (row / constants::NMCMROBINROW) * 2 + 1 : (row / constants::NMCMROBINROW) * 2;
     mcm = (row % constants::NMCMROBINROW) * constants::NMCMROBINCOL + (mcmCol % constants::NMCMROBINCOL);
     channel = constants::NADCMCM - 1 - ((chamberIndex % constants::NCHANNELSPERROW) % constants::NADCMCM);
+  }
+
+  static int getMCMCol(int irob, int imcm)
+  {
+    // calculate the column number of the MCM within the TRD detector
+    return (imcm % constants::NMCMROBINCOL) + constants::NMCMROBINCOL * (irob % 2);
   }
 };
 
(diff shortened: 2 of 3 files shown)


DataFormats/Detectors/TRD/src/Tracklet64.cxx
```diff
@@ -43,6 +43,35 @@ GPUd() int Tracklet64::getPadCol() const
...
+  float y = getUncalibratedY();
+  int mcmCol = (getMCM() % NMCMROBINCOL) + NMCMROBINCOL * (getROB() % 2);
+  // one pad column has 144 pads, the offset of -63 is the center of the first MCM in that column
+  // which is connected to the pads -63 - 9 = -72 to -63 + 9 = -54
+  // float offset = -63.f + ((float)NCOLMCM) * mcmCol;
+  float padWidth = 0.635f + 0.03f * (getDetector() % NLAYER);
+  return y / padWidth + 71.0;
+}
+
+GPUd() float Tracklet64::Slope() const
```
martenole, line 70: could for this not `getUncalibratedDy(1)` be used? And I think one needs to take into account the pad width for the slope, no?

DataFormats/Detectors/TRD/src/Tracklet64.cxx
```diff
@@ -43,6 +43,35 @@ GPUd() int Tracklet64::getPadCol() const
...
+
+GPUd() float Tracklet64::PadPosition() const
+{
+  // int padLocalBin = tracklet.getPosition() ^ 0x80;
+  float padMCM = PadPositionMCM();
+  int mcmCol = HelperMethods::getMCMCol(getROB(), getMCM());
+  return float((mcmCol + 1) * NCOLMCM) + 2.0 - padMCM;
+}
+
+GPUd() float Tracklet64::UncalibratedPad() const
```
martenole, line 59: isn't this a duplicate of `getUncalibratedY()`?

DataFormats/Detectors/TRD/src/Tracklet64.cxx
```diff
@@ -43,6 +43,35 @@ GPUd() int Tracklet64::getPadCol() const
   return CAMath::Nint(6.f + mcmCol * ((float)NCOLMCM) + padLocal);
 }
 
+GPUd() float Tracklet64::PadPositionMCM() const
+{
+  return 12.0 - (getPositionBinSigned() * GRANULARITYTRKLPOS);
+}
+
+GPUd() float Tracklet64::PadPosition() const
```
martenole, line 51: isn't this a duplicate of `getPadCol()`?

DataFormats/Detectors/TRD/src/Tracklet64.cxx
```diff
@@ -43,6 +43,35 @@ GPUd() int Tracklet64::getPadCol() const
   return CAMath::Nint(6.f + mcmCol * ((float)NCOLMCM) + padLocal);
 }
 
+GPUd() float Tracklet64::PadPositionMCM() const
+{
+  return 12.0 - (getPositionBinSigned() * GRANULARITYTRKLPOS);
```
martenole, line 48: `getPositionBinSigned() * GRANULARITYTRKLPOS` should give the local pad number relative to the MCM center. Why are you subtracting 12 here? Shouldn't it be 10.5? But in any case I am not sure if these functions are needed

martenole (changes requested):
Hi @pestratm thanks, the update to the HelperMethods is fine with me. For the tracklet64 I think the existing functions should be OK. Can you see if they are sufficient for the macros you want to add to the TRD repo?
And please not in general functions in O2 need to start with lower case letter

DataFormats/Detectors/TRD/src/Tracklet64.cxx
pestratm: ```UncalibratedPad()``` gives the uncalibrated pad number corresponding to the tracklet position, while ```getUncalibratedY()``` gives the uncalibrated y-coordinate. That's why for ```UncalibratedPad()```, we divide by the pad width, isn't it?

DataFormats/Detectors/TRD/src/Tracklet64.cxx
pestratm: If we take into account the pad width, then we can use ```getUncalibratedDy(1)```. Tbh, I'm not sure about Tom's reasoning not to take it into account.

DataFormats/Detectors/TRD/src/Tracklet64.cxx
martenole: but the uncalibrated pad number should really be given by `getPadCol()`. I just want to avoid any confusion and that we use consistently the functions we have in all the places. Otherwise it get's quite messy. So maybe you could use the functions in your PR to the TRD repo which are already available and drop the ones which are added to Tracklet64? The additional function in HelperMethods is fine of course

DataFormats/Detectors/TRD/src/Tracklet64.cxx
tdietel: > isn't this a duplicate of `getPadCol()`?

getPadCol() returns an int, but the position of a tracklet is known better than a pad. Therefore I think it makes sense to provide a function that gives the tracklet position with float precision.
I suspect changing getPadCol() to return a float might break some code, hence I introduced another helper function. I am thinking about calling it getPadColF() (or getPadColFloat, suggestions welcome).

DataFormats/Detectors/TRD/src/Tracklet64.cxx
tdietel: It looks like this essentially does the same thing as PadPosition() discussed above, but in a different way (i.e. calculating the pad from the y coordinate, rather than from MCM and position within the MCM). 
I don't remember exactly, but I might have compared the two to make sure I get it right.

DataFormats/Detectors/TRD/src/Tracklet64.cxx
tdietel: The slope is given in units of pad/timebin, while dy is a distance in cm.
For my raw data display, I wanted everything in pads and timebins, hence the extra function.
I find code with getSlope() more readable than writing the below calculation, but that might be personal preference.

DataFormats/Detectors/TRD/src/Tracklet64.cxx
tdietel: From my memory, the reason for 12 is that it matches the digits in the raw data display. At the time, I did not try to figure out why this works.

martenole: in `UncalibratedPad()` there is also a +2 which I don't understand. This effectively turns the 12 into a 10, so closer to the expected 10.5

martenole:
Ok, I'd suggest we add
- `getPositionInPadCol()` to return the TRD pad column number as a float
- `getSlopeInPadsPerTB()` to return the slope as float in units of pads per time bin
These functions should then be used internally by `getPadCol()`, `getUncalibratedY()` and `getUncalibratedDy()`.

But we still have to understand the offset. In the `getPadCol()` I currently set an offset of 8 pads, because this is what I see when matching digits to tracklets on tracks and simply looking for the maximum ADC sum inside the digits. On average this should give me the correct offset for the conversion, no? I probably need to look closer again.

@pestratm do you want to give it a try to implement the above 2 functions to Tracklet64 or should I do it?

pestratm:
> Ok, I'd suggest we add
> 
>     * `getPositionInPadCol()` to return the TRD pad column number as a float
> 
>     * `getSlopeInPadsPerTB()` to return the slope as float in units of pads per time bin
>       These functions should then be used internally by `getPadCol()`, `getUncalibratedY()` and `getUncalibratedDy()`.
> 
> 
> But we still have to understand the offset. In the `getPadCol()` I currently set an offset of 8 pads, because this is what I see when matching digits to tracklets on tracks and simply looking for the maximum ADC sum inside the digits. On average this should give me the correct offset for the conversion, no? I probably need to look closer again.
> 
> @pestratm do you want to give it a try to implement the above 2 functions to Tracklet64 or should I do it?

You mean keeping `PadPosition()` as `getPositionInPadCol()`, and `Slope()` as `getSlopeInPadsPerTB()`?

PR #11645, opened 2023-07-10
DPL: signpost update

ktf:
This is a revamped version of the Signpost tracing support. It's actually now quite usable to either investigate cases in which there are nested debug intervals.

On macOS by default is produces the os_signposts which can then be visualised in instruments. For example single stepping in the test_Signpost results in:

image

While on linux / if defined(O2_FORCE_LOGGER_SIGNPOST) one will get properly indented (according to the number of parallel intervals) printout:

```
test_Signpost 1
id0000000000000003:Test category   S> This is a test signpost
id0000000000000005:Test category   S>  A sepaarate interval
id0000000000000003:Test category   *> An event in an interval
id0000000000000003:Test category   E> End of the first interval
id0000000000000005:Test category   E>  A sepaarate interval
id000000016ba59040:Test category   S> A signpost interval from a pointer
id000000016ba59040:Test category   E> A signpost interval from a pointer
id000000016ba59010:Test category   S> A signpost with an engineering type formatter "d"
id000000016ba59010:Test category   E> A signpost interval from a pointer
id0000000000000003:Test category   S> This is dynamic signpost which you will see, because we turned them on
id0000000000000003:Test category   E> This is dynamic signpost which you will see, because we turned them on
```

Notice how the `DYNAMIC` loggers must be enabled explicitly, either via the associated macro, or in instruments. The STACKTRACE will also attach a stacktrace to each event, but for the moment it only works in instruments.

### DIFF
--- a/Framework/Core/src/CommonServices.cxx
+++ b/Framework/Core/src/CommonServices.cxx
@@ -646,14 +646,16 @@ auto flushStates(ServiceRegistryRef registry, DataProcessingStates& states) -> v
   });
 }
 
+O2_DECLARE_DYNAMIC_LOG(monitoring_service);
+
 /// This will flush metrics only once every second.
 auto flushMetrics(ServiceRegistryRef registry, DataProcessingStats& stats) -> void
 {
-  ZoneScopedN("flush metrics");
+  O2_SIGNPOST_ID_GENERATE(sid, monitoring_service);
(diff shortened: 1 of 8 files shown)


ktf:
Notice also that this is only enabled when you do not have -DNDEBUG or when you define O2_FORCE_LOGGER_SIGNPOST.

PR #11646, opened 2023-07-11
Dataformats/TPC: Add PID Response class

tubagundem:


DataFormats/Detectors/TPC/include/DataFormatsTPC/PIDResponse.h
```diff
@@ -0,0 +1,66 @@
...
+{
+ public:
+  /// default constructor
+  PIDResponse() = default;
+
+  /// default destructor
+  ~PIDResponse() = default;
+
+  /// setters
+  void setBetheBlochParams(std::array<double, 5>& betheBlochParams) { mBetheBlochParams = betheBlochParams; }
```
wiechula:
```suggestion
  void setBetheBlochParams(const std::array<double, 5>& betheBlochParams) { mBetheBlochParams = betheBlochParams; }
```

DataFormats/Detectors/TPC/include/DataFormatsTPC/PIDResponse.h
```diff
@@ -0,0 +1,66 @@
...
+  void setMIP(double mip) { mMIP = mip; }
+  void setChargeFactor(double chargeFactor) { mChargeFactor = chargeFactor; }
+
+  /// getters
+  std::array<double, 5> getBetheBlochParams() { return mBetheBlochParams; }
+  double getMIP() { return mMIP; }
+  double getChargeFactor() { return mChargeFactor; }
```
wiechula:
```suggestion
  std::array<double, 5> getBetheBlochParams() const { return mBetheBlochParams; }
  double getMIP() const { return mMIP; }
  double getChargeFactor() const { return mChargeFactor; }
```

DataFormats/Detectors/TPC/include/DataFormatsTPC/PIDResponse.h
```diff
@@ -0,0 +1,66 @@
...
+
+  /// get expected signal of the track
+  double getExpectedSignal(const TrackTPC& track, const o2::track::PID::ID id) const;
+
+  /// get most probable PID of the track
+  o2::track::PID::ID getMostProbablePID(const TrackTPC& track);
```
wiechula:
```suggestion
  o2::track::PID::ID getMostProbablePID(const TrackTPC& track) const;
```

DataFormats/Detectors/TPC/src/PIDResponse.cxx
```diff
@@ -0,0 +1,47 @@
...
+using namespace o2::tpc;
+
+double PIDResponse::getExpectedSignal(const TrackTPC& track, const o2::track::PID::ID id) const
+{
+  const double bethe = mMIP * o2::tpc::BetheBlochAleph(static_cast<double>(track.getP() / o2::track::pid_constants::sMasses[id]), mBetheBlochParams[0], mBetheBlochParams[1], mBetheBlochParams[2], mBetheBlochParams[3], mBetheBlochParams[4]) * std::pow(static_cast<float>(o2::track::pid_constants::sCharges[id]), mChargeFactor);
+  return bethe >= 0.f ? bethe : -999.f;
+}
+
+// get most probable PID
+o2::track::PID::ID PIDResponse::getMostProbablePID(const TrackTPC& track)
```
wiechula:
```suggestion
o2::track::PID::ID PIDResponse::getMostProbablePID(const TrackTPC& track) const
```

DataFormats/Detectors/TPC/src/PIDResponse.cxx
```diff
@@ -0,0 +1,47 @@
...
+{
+  const int count = 8;
+  const double dEdx = track.getdEdx().dEdxTotTPC;
```
wiechula: I think we should have a check for the case we don't have dEdx:
if (dEdx < 10) {
return o2::track::PID::Pion;
}

DataFormats/Detectors/TPC/src/PIDResponse.cxx
```diff
@@ -0,0 +1,47 @@
...
+double PIDResponse::getExpectedSignal(const TrackTPC& track, const o2::track::PID::ID id) const
+{
+  const double bethe = mMIP * o2::tpc::BetheBlochAleph(static_cast<double>(track.getP() / o2::track::pid_constants::sMasses[id]), mBetheBlochParams[0], mBetheBlochParams[1], mBetheBlochParams[2], mBetheBlochParams[3], mBetheBlochParams[4]) * std::pow(static_cast<float>(o2::track::pid_constants::sCharges[id]), mChargeFactor);
+  return bethe >= 0.f ? bethe : -999.f;
+}
+
+// get most probable PID
+o2::track::PID::ID PIDResponse::getMostProbablePID(const TrackTPC& track)
+{
+  const int count = 8;
```
wiechula: This excludes Alpha. Is this intended?
Perhaps use the count directly in the loop. See below.
```suggestion
```

DataFormats/Detectors/TPC/src/PIDResponse.cxx
```diff
@@ -0,0 +1,47 @@
...
+  const double dEdx = track.getdEdx().dEdxTotTPC;
+
+  double distanceMin = std::abs(dEdx - getExpectedSignal(track, 0));
+  int id = 0;
+
+  // calculate the distance to the expected dEdx signals
```
wiechula:
```suggestion
  // calculate the distance to the expected dEdx signals
  // start from Pion to exlude Muons
```

DataFormats/Detectors/TPC/src/PIDResponse.cxx
```diff
@@ -0,0 +1,47 @@
...
+  for (int i = 2; i < count; i++) {
```
wiechula:
```suggestion
  for (o2::track::PID::ID i = o2::track::PID::Pion; i < o2::track::PID::NIDs; i++) {
```

DataFormats/Detectors/TPC/src/PIDResponse.cxx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// get most probable PID
+o2::track::PID::ID PIDResponse::getMostProbablePID(const TrackTPC& track)
+{
+  const int count = 8;
+  const double dEdx = track.getdEdx().dEdxTotTPC;
+
+  double distanceMin = std::abs(dEdx - getExpectedSignal(track, 0));
+  int id = 0;
```
wiechula:
```suggestion
  auto id = o2::track::PID::Electron;
  double distanceMin = std::abs(dEdx - getExpectedSignal(track, id));
```

DataFormats/Detectors/TPC/src/PIDResponse.cxx
```diff
@@ -0,0 +1,47 @@
...
+
+#include "DataFormatsTPC/PIDResponse.h"
+#include "DataFormatsTPC/BetheBlochAleph.h"
+
+using namespace o2::tpc;
+
+double PIDResponse::getExpectedSignal(const TrackTPC& track, const o2::track::PID::ID id) const
+{
+  const double bethe = mMIP * o2::tpc::BetheBlochAleph(static_cast<double>(track.getP() / o2::track::pid_constants::sMasses[id]), mBetheBlochParams[0], mBetheBlochParams[1], mBetheBlochParams[2], mBetheBlochParams[3], mBetheBlochParams[4]) * std::pow(static_cast<float>(o2::track::pid_constants::sCharges[id]), mChargeFactor);
+  return bethe >= 0.f ? bethe : -999.f;
```
wiechula: Didn't we discuss to put some lower bg value (could also be made configurable)? Also, your function is expected to return double.
Something like:
```suggestion
  const double bg = static_cast<double>(track.getP() / o2::track::pid_constants::sMasses[id]);
  if (bg < 0.05) {
    return -999.;
  }
  const double bethe = mMIP * o2::tpc::BetheBlochAleph(bg, mBetheBlochParams[0], mBetheBlochParams[1], mBetheBlochParams[2], mBetheBlochParams[3], mBetheBlochParams[4]) * std::pow(static_cast<float>(o2::track::pid_constants::sCharges[id]), mChargeFactor);
  return bethe >= 0. ? bethe : -999.;
```

wiechula (changes requested):
A few minor things.

DataFormats/Detectors/TPC/include/DataFormatsTPC/PIDResponse.h
```diff
@@ -0,0 +1,64 @@
...
+#ifndef AliceO2_TPC_PIDResponse_H
+#define AliceO2_TPC_PIDResponse_H
+
+// o2 includes
+#include "ReconstructionDataFormats/PID.h"
+#include "DataFormatsTPC/TrackTPC.h"
+
+namespace o2::tpc
+{
+
```
wiechula: I think TrackTPC can be forward declared:
```suggestion

namespace o2::tpc
{
  class TrackTPC;
```

DataFormats/Detectors/TPC/src/PIDResponse.cxx
```diff
@@ -0,0 +1,55 @@
...
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// @file   PIDResponse.cxx
+/// @author Tuba Gündem, tuba.gundem@cern.ch
+///
+
+#include "DataFormatsTPC/PIDResponse.h"
+#include "DataFormatsTPC/BetheBlochAleph.h"
```
wiechula:
```suggestion
#include "DataFormatsTPC/BetheBlochAleph.h"
#include "DataFormatsTPC/TrackTPC.h"
```

GPU/GPUTracking/Merger/GPUTPCGMO2Output.cxx
```diff
@@ -134,6 +136,10 @@ GPUdii() void GPUTPCGMO2Output::Thread<GPUTPCGMO2Output::output>(int nBlocks, in
     if (merger.Param().par.dodEdx) {
       oTrack.setdEdx(tracksdEdx[i]);
     }
+
+    auto pid = pidResponse.getMostProbablePID(oTrack);
+    oTrack.setPID(pid);
+
     oTrack.setOuterParam(o2::track::TrackParCov(
       outerPar.X, outerPar.alpha,
       {outerPar.P[0], outerPar.P[1], outerPar.P[2], outerPar.P[3], outerPar.P[4]},
```
wiechula: The setting of the PID should be done after `setOuterParam` and should set the PID also for outer params:
```suggestion

    oTrack.setOuterParam(o2::track::TrackParCov(
      outerPar.X, outerPar.alpha,
      {outerPar.P[0], outerPar.P[1], outerPar.P[2], outerPar.P[3], outerPar.P[4]},
      
    const auto pid = pidResponse.getMostProbablePID(oTrack);
    oTrack.setPID(pid);
    oTrack.getOuterParam().setPID(pid);
```

wiechula (changes requested):
A few more minor things.

Also, we need to be able to use the code on the GPU, if I understand correctly. So we'll need the GPU definitions inside. Please add this.

And for the `GPUTPCGMO2Output.cxx` we should check with @davidrohr and @shahor02 , since it would be good to be able to configure the class from outside (e.g. load from ccdb). And I'm not sure what would be the proper way to do that and pass it on.

tubagundem:
Hi @davidrohr and @shahor02,

could you check this PR?

GPU/GPUTracking/Merger/GPUTPCGMO2Output.cxx
```diff
@@ -134,12 +136,18 @@ GPUdii() void GPUTPCGMO2Output::Thread<GPUTPCGMO2Output::output>(int nBlocks, in
...
     }
+
     oTrack.setOuterParam(o2::track::TrackParCov(
       outerPar.X, outerPar.alpha,
       {outerPar.P[0], outerPar.P[1], outerPar.P[2], outerPar.P[3], outerPar.P[4]},
       {outerPar.C[0], outerPar.C[1], outerPar.C[2], outerPar.C[3], outerPar.C[4], outerPar.C[5],
        outerPar.C[6], outerPar.C[7], outerPar.C[8], outerPar.C[9], outerPar.C[10], outerPar.C[11],
        outerPar.C[12], outerPar.C[13], outerPar.C[14]}));
+
+    const auto pid = pidResponse.getMostProbablePID(oTrack);
```
davidrohr: This should not run unconditionally, since we don't necessarily need it in sync reco.
Does it require dEdx? If yes, please move it into the above if stratement `if (merger.Param().par.dodEdx) {`.
Otherwise, we should add an extra flag in the GPU reco settings to enable / disable.

tubagundem: Yes, it needs the dEdx information. I added the if statement below otherwise, `mOuterParam.mPID` is overwritten by `oTrack.setOuterParam(o2::track::TrackParCov(`.

DataFormats/Detectors/TPC/include/DataFormatsTPC/PIDResponse.h
```diff
@@ -0,0 +1,117 @@
...
+  // calculate the distance to the expected dEdx signals
+  // start from Pion to exlude Muons
+  for (o2::track::PID::ID i = o2::track::PID::Pion; i < o2::track::PID::NIDs; i++) {
+    const float distance = o2::gpu::GPUCommonMath::Abs(dEdx - getExpectedSignal(track, i));
+    if (distance < distanceMin) {
+      id = i;
+      distanceMin = distance;
+    }
+  }
+
```
shahor02, line 111: @tubagundem @wiechula : In the aliroot we were overriding the electrons to the closest alternative PID in the bands crossing regions: https://github.com/alisw/AliRoot/blob/f01cd3609506f29aa9b5f0c408b4234c99325a61/STEER/ESD/AliESDpid.cxx#L472-L480. I think we should introduce the same here, perhaps passing the bands and tolerances definitions via configurable params.

GPU/GPUTracking/Merger/GPUTPCGMO2Output.cxx
```diff
@@ -113,6 +114,7 @@ GPUdii() void GPUTPCGMO2Output::Thread<GPUTPCGMO2Output::output>(int nBlocks, in
   const unsigned int flagsRequired = getFlagsRequired(merger.Param().rec);
   TrackTPC* outputTracks = merger.OutputTracksTPCO2();
   unsigned int* clusRefs = merger.OutputClusRefsTPCO2();
+  PIDResponse pidResponse{};
```
davidrohr: Could you move the instantiation of this class also into the if branch. No reason to always have it, and it will simplify compiler optimizations.

DataFormats/Detectors/TPC/include/DataFormatsTPC/PIDResponse.h
tubagundem: If I understood correctly and add some code like below, you would like to pass these `(p > 0.38) && (p < 0.48)`  and `(p > 0.75) && (p < 0.85)` via configurable params and not hardcoded right?

 ```
 // override the electrons to the closest alternative PID in the bands crossing regions
  if (id == o2::track::PID::Electron && useElectronExclusionBands) {
    const float p = track.getP();
    if ((p > 0.38) && (p < 0.48)) {
        id = o2::track::PID::Kaon;
    } else if ((p > 0.75) && (p < 0.85)) {
        id = o2::track::PID::Proton;
    }
  }
  ```
  
Then do I need to add another option to reco workflow below?
https://github.com/AliceO2Group/AliceO2/blob/1af9293cdfd6c8f7005998c7ea2dce3dc7d70dc6/Detectors/TPC/workflow/src/tpc-reco-workflow.cxx#L57

shahor02: Yes, but I would do this via configurable params, e.g. adding after https://github.com/AliceO2Group/AliceO2/blob/1af9293cdfd6c8f7005998c7ea2dce3dc7d70dc6/GPU/GPUTracking/Definitions/GPUSettingsList.h#L58

```
AddOptionRTC(PID_EKrangeMin, float, 0.38, "", 0, "min P of electron/K BB bands crossing")
AddOptionRTC(PID_EKrangeMax, float, 0.48, "", 0, "max P of electron/K BB bands crossing")
AddOptionRTC(PID_EPrangeMin, float, 0.75, "", 0, "min P of electron/p BB bands crossing")
AddOptionRTC(PID_EPrangeMax, float, 0.85, "", 0, "max P of electron/p BB bands crossing")
```
unless @davidrohr may suggest a more appropriate place.
The numeric values later should be tuned for run3.

davidrohr: > unless @davidrohr may suggest a more appropriate place.
> The numeric values later should be tuned for run3.

fine wth me

tubagundem: I reconstructed one ctf file to test it. It looks like below before and after I added the overriding of electrons.

image

shahor02: but why we don't see the bands?

tubagundem: It has a very slight effect. I am assuming that we need to tune momentum range according to data?

image

shahor02: Probably the markers are too large so that the density is visible except on the periphery. Anyway, judging about the crossing bands on the log plot of this scale makes no sense, they should contain ~10% of K and p, which could be lost under the electron hypothesis.

davidrohr (approved):
ok for the GPU part now, should be squashed when merging

wiechula (approved):
Looks fine for me. Perhaps in future we need the possibility to configure the BB params from CCDB.

wiechula:
@shahor02 , is this also fine for you?

shahor02 (approved):

PR #11650, opened 2023-07-12
FITRaw: refactoring + new raw data metric feature

afurs:
FIRRaw refactoring and new raw data metric feature:
1. Raw data metric collects information about raw2digit decoding procedure per each FEE module, i.e. failures.
2. LookupTable is now more common structure for FIT detectors, contains information about detector ID.
3. Digit2raw configuration and preparation is moved into FITRaw lib.

### DIFF
--- a/Detectors/FIT/workflow/include/FITWorkflow/RawReaderFIT.h
+++ b/Detectors/FIT/workflow/include/FITWorkflow/RawReaderFIT.h
@@ -44,6 +44,7 @@ class RawReaderFIT : public RawReaderType
   ~RawReaderFIT() = default;
   typedef RawReaderType RawReader_t;
   typedef typename RawReader_t::DigitBlockFIT_t DigitBlockFIT_t;
+  typedef typename RawReader_t::RawDataMetric_t RawDataMetric_t;
   typedef typename DigitBlockFIT_t::LookupTable_t LookupTable_t;
   typedef typename DigitBlockFIT_t::Digit_t Digit_t;
   typedef typename DigitBlockFIT_t::SubDigit_t SubDigitTmp_t;
@@ -61,10 +62,17 @@ class RawReaderFIT : public RawReaderType
   o2::header::DataOrigin mDataOrigin;
   std::vector<Digit_t> mVecDigit;
   std::vector<DetTrigInput_t> mVecTrgInput;
+  std::vector<RawDataMetric> mVecRawDataMetric;
   SubDigit_t mVecSubDigit;             // tuple of vectors
   SingleSubDigit_t mVecSingleSubDigit; // tuple of vectors
   bool mEnableEmptyTFprotection{false};
   bool mDumpData;
+  void dumpRawDataMetrics() const
+  {
+    for (const auto& entry : mVecRawDataMetric) {
+      entry.print();
+    }
+  }
   void reserveVecDPL(std::size_t nDigits, std::size_t nSubDigits)
   {
     mVecDigit.reserve(nDigits);
@@ -97,6 +105,7 @@ class RawReaderFIT : public RawReaderType
       },
                  mVecSingleSubDigit);
     }
+    mVecRawDataMetric.clear();
   }
   template <std::size_t... IsubDigits, std::size_t... IsingleSubDigits>
   auto callGetDigit(std::index_sequence<IsubDigits...>, std::index_sequence<IsingleSubDigits...>)
@@ -132,6 +141,7 @@ class RawReaderFIT : public RawReaderType
     if (mDumpData) {
       callPrint(IndexesSubDigit{}, IndexesSingleSubDigit{});
     }
+    RawReader_t::getMetrics(mVecRawDataMetric);
   }
   void configureOutputSpec(std::vector<o2::framework::OutputSpec>& outputSpec) const
   {
@@ -151,6 +161,7 @@ class RawReaderFIT : public RawReaderType
     if constexpr (sUseTrgInput) {
       outputSpec.emplace_back(mDataOrigin, DetTrigInput_t::sChannelNameDPL, 0, o2::framework::Lifetime::Timeframe);
     }
+    outputSpec.emplace_back(mDataOrigin, "RawDataMetric", 0, o2::framework::Lifetime::Timeframe);
   }
   void makeSnapshot(o2::framework::ProcessingContext& pc) const
   {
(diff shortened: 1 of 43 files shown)


afurs:
Looks like unrelated issue(failed test at QC and something in O2Physics). @shahor02 shall I restart CICD via dummy commit?

martenole:
Hi @afurs Ruben is away until next week. I did not go through this large PR in detail yet, but I assume this is something you would like to have at P2 before PbPb, right? Is it really just restructuring and adding of metrics? Have you tested it with recorded raw data to see that it gives the expected results? I assume this should fully agree with the current raw decoding version, right?

afurs:
Hi @martenole yes, this is new feature which we need before PbPb. It has full agreement with current decoding procedure.  But I would like to make some tests again for sure. So this PR better to postpone, shall I put WIP int title?

martenole:
Would be good to have all tests done of course. For FIT there is also some processing done on the FLPs, right? For the FLP suite to be installed next week on the 18th O2 will be taken from the next nightly tag I think, so if there are possibly some incompatibilities (I don't know) then we better postpone merging this until after next week and then we could install it at the next windows (latest mid august)

afurs:
@martenole we have decoder(1 % data taking) +QC at FLP and main decoding at EPN. Okay, I finished tests, decoding is fine. Raw data metric emulation only require some hotfixes, but we can live w/o it for now.

Framework/Core/include/Framework/ASoA.h
```diff
@@ -1722,8 +1722,11 @@ std::tuple<typename Cs::type...> getRowData(arrow::Table* table, T rowIterator,
       return (*mColumnIterator & (static_cast<type>(1) << bit)) >> bit;                                                                                                           \
     }                                                                                                                                                                             \
   };                                                                                                                                                                              \
-  static const o2::framework::expressions::BindingNode _Getter_ { _Label_, typeid(_Name_).hash_code(),                                                                            \
-                                                                  o2::framework::expressions::selectArrowType<MAKEINT(_Size_)>() }
+  static const o2::framework::expressions::BindingNode _Getter_                                                                                                                   \
```
martenole: any idea why these 3 Framework/Core files were touched? its only formatting, but they should still not be changed by this PR

Detectors/FIT/workflow/include/FITWorkflow/RawReaderFIT.h
```diff
@@ -151,6 +161,7 @@ class RawReaderFIT : public RawReaderType
     if constexpr (sUseTrgInput) {
       outputSpec.emplace_back(mDataOrigin, DetTrigInput_t::sChannelNameDPL, 0, o2::framework::Lifetime::Timeframe);
     }
+    outputSpec.emplace_back(mDataOrigin, "RawDataMetric", 0, o2::framework::Lifetime::Timeframe);
```
martenole, line 164: should the RawDataMetrics not be sent only when requested and not unconditionally?

afurs: @martenole shall I make raw data metric sending optional for DPL task? Actually I forgot, in case if it sends messages with `Lifetime::Timeframe` should there be always receiver task? Raw data metric objects will be used only for QC

Framework/Core/include/Framework/ASoA.h
afurs: > any idea why these 3 Framework/Core files were touched? its only formatting, but they should still not be changed by this PR

I got formatting issue at the CICD start, so I just used `git clang-format HEAD~1` and `git clang-format origin/dev` (Usually I use `clang-format -style=file -i ...`). Was it wrong decision? Shall I cancel that commit?

martenole: Usyually `git clang-format HEAD~X` with X being the number of commits you did is fully sufficient. I believe that in your local branch you somehow introduced some merging from `dev` and therefore git thought you touched these files which you did not. Since all your changes are basically in a single commit only you could do (while on your working branch)
```
# update your local repo
git fetch --all
# put your changes on top of dev (in case your remote name is e.g. origin instead of upstream adapt the command)
git rebase upstream/dev
# merge all your changes in a single commit, drop the clang format of framework files
git rebase -i HEAD~3
git push -f
```

Detectors/FIT/workflow/include/FITWorkflow/RawReaderFIT.h
martenole: I think I misread. You want to have only the dumping of raw data metrics optionally, but they should always be produced. I think that is OK. If there is no consumer the message should just be dropped by the framework.

Framework/Core/include/Framework/ASoA.h
afurs: Ah, yes, I usually develop my code in the dedicated branch, and then I do for O2:
```
git checkout dev
git pull --rebase
git checkout my_branch
git merge dev
```
I guess this is not good schema, is it? Should I merge `my_branch` -> `dev` and then push it next time?

Detectors/FIT/workflow/include/FITWorkflow/RawReaderFIT.h
afurs: Okay, thanks! Dumping is just for debugging

Framework/Core/include/Framework/ASoA.h
martenole: in O2 there are anyhow no merge commits allowed. The history here is always linear. So I would use rebase instead of merge. Other than that your workflow looks fine

afurs: I see, okay, then no merging anymore :) What shall I do now? In previous message you attached list of git operations. Am I understood correctly these git commands as  solution for current situation?

martenole: this would be the easiest to get rid of the unrelated framework changes I think, so yes please

afurs: @martenole I see that there are still changes in the unrelated framework, maybe I shall use `git revert` for merging and formatting commit?

martenole: yes and for some reason your commit "clang" no contains changes both to framework and to your file. So I fear this one you have to fix manually:
```
git reset --soft HEAD~1
# unstage files you don't want to change
git restore --staged <list of framework files>
# restore their previous state
git checkout <list of framework files>
# commit only the changes you want to keep
git commit -m "clang"
git push
```

afurs:
@martenole okay, now it looks clean, unrelated framework is untouched now

Framework/Core/include/Framework/ASoA.h
afurs: @martenole thanks a lot! Actually I did:
```
git reset --hard main_commit_id
"formatting command"
git commit...
git push --force
```
Now I see only my main commit and formatting of related framework

afurs:
Hotfix for raw data metric implemented. Since currently `MetricStatus_t` is `uint8_t` , `boost::program_options` recognizes only first char.

afurs:
Looks like unrelated issue, QC testing failed. @shahor02 shall I restart CICD via dummy commit?

shahor02 (approved):

PR #11653, opened 2023-07-12
[EMCAL-565, EMCAL-566] Add gain calib factors to online calibration

jokonig:
- Gain calibration influences the cell energy. Hence the bad channel calibration can be influenced by the gain calibration factors
- Gain calib factors are read from ccdb and then passed to the calib data
- Application can be switched off with the workflow argument no-applyGainCalib
- cell energy is increased to 100 MeV for filling of the bad channel histograms as default to speedup the filling (before was 75 MeV)


mfasDa:
CI errors unrelated

mfasDa (approved):
Looks good!

PR #11654, opened 2023-07-13
Hmpidao2d

gvolpe79:


Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -641,20 +641,32 @@ using AmbiguousFwdTrack = AmbiguousFwdTracks::iterator;
...
+                  hmpid::HMPIDXTrack,
+                  hmpid::HMPIDYTrack,
+                  hmpid::HMPIDXMip,
+                  hmpid::HMPIDYMip,
                   hmpid::HMPIDNPhotons,
-                  hmpid::HMPIDQMip);
+                  hmpid::HMPIDQMip,
+                  hmpid::HMPIDClusSize,
+                  hmpid::HMPIDMom,
+                  hmpid::HMPIDPhotsCharge);
```
jgrosseo: This breaks the Run 2 converted data which is written in the current format. 
You have to create a version 1 of the table and provide a converter from version 0 to 1

jgrosseo (changes requested):
This is backward incompatible for Run 2 converted data and needs to be adjusted (see my comment on the data model).

jgrosseo:
Tagging also @ddobrigk

Framework/Core/include/Framework/AnalysisDataModel.h
gvolpe79: Hi @jgrosseo, thanks for the suggestion! Done!

davidrohr:
@gvolpe79 : replying to your e-mail. The FullCI for this PR is failing, since it runs the async reco on some simulated test-data set, and aod producer in async reco fails since it does not receive `HMP/MATCHES/0`.
Which workflow generates the HMPID matches? Currently, the async reco runs only `o2-hmpid-digits-to-clusters-workflow` with the following command line
```
o2-hmpid-digits-to-clusters-workflow --session default --severity info --shm-segment-id 0 --shm-segment-size 8000000000  --early-forward-policy noraw --monitoring-backend no-op:// --fairmq-rate-logging 0 --timeframes-rate-limit 24 --timeframes-rate-limit-ipcid 0 --disable-root-input --disable-root-output --pipeline HMP-Clusterization:1  --configKeyValues "NameConf.mDirGeom=/sw/BUILD/9cb9cc345a40ffd5204e9fc5af582a6b7dd44223/O2-full-system-test/full-system-test-sim;NameConf.mDirGRP=/sw/BUILD/9cb9cc345a40ffd5204e9fc5af582a6b7dd44223/O2-full-system-test/full-system-test-sim;keyval.input_dir=/sw/BUILD/9cb9cc345a40ffd5204e9fc5af582a6b7dd44223/O2-full-system-test/full-system-test-sim;keyval.output_dir=/dev/null;;"
```

gvolpe79:
Thanks @davidrohr, so the code is ok! The workflow that produces the HMPID mathces is the following
o2-hmpid-matcher-workflow

davidrohr:
Can you try with this patch:
```
diff --git a/prodtests/full-system-test/dpl-workflow.sh b/prodtests/full-system-test/dpl-workflow.sh
index 25ef99c2c5..52ca5b920d 100755
--- a/prodtests/full-system-test/dpl-workflow.sh
+++ b/prodtests/full-system-test/dpl-workflow.sh
@@ -431,6 +431,7 @@ has_detector FDD && ! has_detector_from_global_reader FDD && has_processing_step
 has_detector FV0 && ! has_detector_from_global_reader FV0 && has_processing_step FV0_RECO && add_W o2-fv0-reco-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC"
 has_detector ZDC && ! has_detector_from_global_reader ZDC && has_processing_step ZDC_RECO && add_W o2-zdc-digits-reco "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC"
 has_detector HMP && ! has_detector_from_global_reader HMP && has_processing_step HMP_RECO && add_W o2-hmpid-digits-to-clusters-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT --pipeline $(get_N HMP-Clusterization HMP REST 1 HMPCLUS)"
+has_detector HMP && ! has_detector_from_global_reader HMP && has_processing_step HMP_RECO && add_W o2-hmpid-matcher-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT"
 has_detectors_reco MCH MID && has_detector_matching MCHMID && add_W o2-muon-tracks-matcher-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_MC $DISABLE_ROOT_OUTPUT --pipeline $(get_N muon-track-matcher MATCH REST 1)"
 has_detectors_reco MFT MCH && has_detector_matching MFTMCH && add_W o2-globalfwd-matcher-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC --pipeline $(get_N globalfwd-track-matcher MATCH REST 1 FWDMATCH)" "$MFTMCHConf"
 
```

davidrohr:
Not fully sure if that is 100% correct, but the matcher should never create root input, since it will always come from somewhere else already, thus I assume the options are correct like this.

gvolpe79:
@davidrohr done! Let's see!

gvolpe79:
Hi @davidrohr so now the problem is in the hmpid-matcher! It was working when I tested it using tracks and hmpid clusters from file as in sim_challenge.sh, maybe in the dpl-workflow.sh it expects them on stream! Can I run dpl-workflow.sh on my pc?

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -431,6 +431,7 @@ has_detector FDD && ! has_detector_from_global_reader FDD && has_processing_step
 has_detector FV0 && ! has_detector_from_global_reader FV0 && has_processing_step FV0_RECO && add_W o2-fv0-reco-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC"
 has_detector ZDC && ! has_detector_from_global_reader ZDC && has_processing_step ZDC_RECO && add_W o2-zdc-digits-reco "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC"
 has_detector HMP && ! has_detector_from_global_reader HMP && has_processing_step HMP_RECO && add_W o2-hmpid-digits-to-clusters-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT --pipeline $(get_N HMP-Clusterization HMP REST 1 HMPCLUS)"
+has_detector HMP && ! has_detector_from_global_reader HMP && has_processing_step HMP_RECO && add_W o2-hmpid-matcher-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT"
```
shahor02: @gvolpe79, in the matcher you are requesting ITS-TPC matching MC info, which is not available during raw data processing (which FST does). This line should look like
```
has_detector HMP && ! has_detector_from_global_reader HMP && has_processing_step HMP_RECO && add_W o2-hmpid-matcher-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC"
```

gvolpe79: @shahor02 , thanks! I have just modified dpl-workflow.sh!

jgrosseo:
@gvolpe79 Did you see my comments above on the need to change the version and write a converter? This is something you have to provide in this PR (otherwise it cannot be merged).

gvolpe79:
@jgrosseo I saw your comments, and indeed I also replied! I already modified the data model adding version 0 e version 1 of the hmpid table. Modification have been already committed! The converter has to be added in O2Physics, so I have to open an other pull request! Anyway, I'm going to do also that!

jgrosseo:
> @jgrosseo I saw your comments, and indeed I also replied! I already modified the data model adding version 0 e version 1 of the hmpid table. Modification have been already committed! The converter has to be added in O2Physics, so I have to open an other pull request! Anyway, I'm going to do also that!

Yes, sorry I overlooked it.

The version 1 looks good. Please also add a line like this
`DECLARE_EQUIVALENT_FOR_INDEX(aod::Collisions_000, aod::Collisions_001);`
(check line 1402 in the current version).

Yes, the converter has to go to O2Physics. In order to do this cleanly, you have to split this PR:
- first merge the new AnalysisDataModel.h keeping version *0* as default
- next day, merge the O2Physics PR with the converter
- next day, merge the remainder of this PR and the moving from version *0* to *1*

prodtests/full-system-test/dpl-workflow.sh
gvolpe79: hi @shahor02 now the problem is again on the aod-producer-workflow....!!

prodtests/full-system-test/dpl-workflow.sh
```diff
@@ -431,6 +431,7 @@ has_detector FDD && ! has_detector_from_global_reader FDD && has_processing_step
 has_detector FV0 && ! has_detector_from_global_reader FV0 && has_processing_step FV0_RECO && add_W o2-fv0-reco-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC"
 has_detector ZDC && ! has_detector_from_global_reader ZDC && has_processing_step ZDC_RECO && add_W o2-zdc-digits-reco "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT $DISABLE_MC"
 has_detector HMP && ! has_detector_from_global_reader HMP && has_processing_step HMP_RECO && add_W o2-hmpid-digits-to-clusters-workflow "$DISABLE_DIGIT_ROOT_INPUT $DISABLE_ROOT_OUTPUT --pipeline $(get_N HMP-Clusterization HMP REST 1 HMPCLUS)"
+has_detector HMP && ! has_detector_from_global_reader HMP && has_processing_step HMP_RECO && add_W o2-hmpid-matcher-workflow "--disable-root-input $DISABLE_ROOT_OUTPUT $DISABLE_MC"
```
shahor02: please use `$DISABLE_DIGIT_ROOT_INPUT` instead of explicit `--disable-root-input`

shahor02: Your matcher produces `HMP/CONSTRAINED/0` (and `HMP/MC_CONSTRAINED/0`) outputs
https://github.com/AliceO2Group/AliceO2/blob/2b3de23a05d5c195513437fe8ad33ac9249a2e10/Detectors/GlobalTrackingWorkflow/src/HMPMatcherSpec.cxx#L163-L165, which can be consumed only by the writer https://github.com/AliceO2Group/AliceO2/blob/2b3de23a05d5c195513437fe8ad33ac9249a2e10/Detectors/HMPID/workflow/src/HMPMatchedWriterSpec.cxx#L50-L53. 
Then the reader produces the same output under the `HMP/MATCHES/0` (`HMP/MCLABELS/0`) specs which are consumed by the AODProducer via RecoContainer.
In the raw mode the data goes directly from the matcher to the AODProducer and obviously there is a mismatch between specs. You should change all specs above to  `HMP/MATCHES/0` (`HMP/MCLABELS/0`).

shahor02 (changes requested):

prodtests/full-system-test/dpl-workflow.sh
davidrohr: > please use `$DISABLE_DIGIT_ROOT_INPUT` instead of explicit `--disable-root-input`

actually I put --disable-root-input there, since I though the matcher would never read ROOT input, also if other workflows would use ROOT input when running from digits?

But we can also homogeneously use $DISABLE_DIGIT_ROOT_INPUT, I have no strong opinion on this.

shahor02: I've asked to change it just for the homogeneity.

gvolpe79:
Hi @shahor02 now I see an error with the matcher, but it is not clear if it is still when running the dpl-workflow.sh.

shahor02:
Hi @gvolpe79 

As the fullCI log shows, the problem is in the `sim_challenge.sh`:
```
./sim-challenge.log[ERROR] invalid workflow in o2-hmpid-matcher-workflow: No matching output found for HMP/CONSTRAINED/0 as requested by data processor "HMPMatchedWriter". Candidates:
```
This is because you've changed the matcher output to `HMP/MATCHES/0` but the input of the writer is still `HMP/CONSTRAINED/0`.
All inputs/outputs must match each other.
Perhaps you should try to run locally `$O2_ROOT/prodtests/full_system_test.sh` and `$O2_ROOT/sim_challenge.sh` before pushing the modifications.

gvolpe79:
Hi @shahor02 and @jgrosseo , all the checks have been passed so in principle the code could be merged! For the backward compatibility Jan suggested (see message above) to do like this:

- first merge the new AnalysisDataModel.h keeping version 0 as default
- next day, merge the O2Physics PR with the converter
- next day, merge the remainder of this PR and the moving from version 0 to 1

I have a question, to merge just AnalysisDataModel.h should I open a new pull request or this one can be merged partially?

Anyway, I was wondering if instead we could proceed like this: 
I prepare the O2Physics pull request for the converter and then with a bit of coordination between us we merge both at the same time! It should be safe also in this way, but simpler for me.

ddobrigk:
> Hi @shahor02 and @jgrosseo , all the checks have been passed so in principle the code could be merged! For the backward compatibility Jan suggested (see message above) to do like this:
> 
> * first merge the new AnalysisDataModel.h keeping version 0 as default
> * next day, merge the O2Physics PR with the converter
> * next day, merge the remainder of this PR and the moving from version 0 to 1
> 
> I have a question, to merge just AnalysisDataModel.h should I open a new pull request or this one can be merged partially?
> 
> Anyway, I was wondering if instead we could proceed like this: I prepare the O2Physics pull request for the converter and then with a bit of coordination between us we merge both at the same time! It should be safe also in this way, but simpler for me.

(sorry for jumping in on this topic only now, as I was busy traveling) - Hi @gvolpe79 , I am afraid we cannot merge simultaneously - O2Physics is always built against O2 of the day before, and this means we really have to do it staggered in time to make sure the compilation of the PRs go through. Therefore, in practice, I fear the most convenient option is still to separate the PRs - you can just open a new one with only the changes to the `AnalysisDataModel` and we can go from there. Of course, @jgrosseo and @shahor02 please jump up in case I am missing something! Thanks!

shahor02:
Hi, I agree with @jgrosseo and @ddobrigk

gvolpe79:
ok @ddobrigk, @jgrosseo , @shahor02 so I will implement the hmpid converter doing something similar to O2Physics/Common/TableProducer/zdcConverter.cxx. But where this converter is used? Do I need to implement/modify something else?!

ddobrigk:
Hi @gvolpe79 , for now, let's just merge the data model change keeping version 000 as default - then we can calmly work on the converter PR. So, what you need is exactly to mirror the zdc converter (and also add it to the CMakeLists), and that's all, no further action needed from your side for the converter.

What will then start happening once we start using version 001 is that anyone analysing old data will have the old table in AO2D, and the old table has to be brought up to version 001. This is done by analysers via the addition of this hmpid converter... 

If needed, we can iterate on the specifics, but for sure, the first step is the change to the data model since that is needed for the converter to compile. Thanks!!

gvolpe79:
Hi @ddobrigk, so but this is the point, if in O2Physics I have just to implement the converter that is then used only at analysis level, why we cannot do as I suggested! O2Physics with the hmp converter will compile anyway, independently on O2 and the other way around for O2! I don't want to insist, just to understand if I'm missing something :)

gvolpe79:
Hi @ddobrigk, I miss your last raw, the converter needs the new data model, ok is clear now!

gvolpe79:
Hi @shahor02 the converter is done and merged in o2physics, I think that now also this pull request could be merged! I see that build/O2/o2-cs8 fails but it seems that the error is not related to my code!

jgrosseo:
@ddobrigk Will you create the needed service wagon in Hyperloop and announce it?
I am not sure which of the nightly Hyperloop trains will fail because they need the converter. In case you get mails, could you follow up?

ddobrigk:
@jgrosseo , sure. I will create the service wagon & announce but I checked with Giacomo and all the hmpid trees are empty in converted data, so I would not actually expect this to impact active analyser usage. As for the integration tests, these may still crash, you are correct: I will check those. Thanks!

jgrosseo:
Thanks!

Really it is always empty? That is a pity because the code is there:
https://github.com/alisw/AliPhysics/blob/8d5c8f10e396c333bfdf5448c5aa7ae23735fd4f/RUN3/AliAnalysisTaskAO2Dconverter.cxx#L1937C1-L1937C1

PR #11657, opened 2023-07-14
ctpdev: dump raw on error

lietava:


lietava:
@martenole @shahor02 : I want to dump whole TF  at line 
https://github.com/lietava/AliceO2/blob/7d05b9a39bc4bc723ebb7b4d803f16d0d7a10fcd/Detectors/CTP/reconstruction/src/RawDataDecoder.cxx#L303
Is there some way how to use the DPLRawParser or some other tool for that ?
Cheers, Roman.

lietava:
@shahor02 @martenole : I found way how to dump. Let me know, please, if you have some comments. Otherwise I clean the code and 'ready for review'.

Detectors/CTP/reconstruction/src/RawDataDecoder.cxx
```diff
@@ -161,7 +170,7 @@ int RawDataDecoder::decodeRaw(o2::framework::InputRecord& inputs, std::vector<o2
     } else {
       LOG(error) << "Unxpected  CTP CRU link:" << linkCRU;
     }
-    LOG(debug) << "RDH FEEid: " << feeID << " CTP CRU link:" << linkCRU << " Orbit:" << rdhOrbit << " triggerType:" << triggerType;
+    LOG(info) << "RDH FEEid: " << feeID << " CTP CRU link:" << linkCRU << " Orbit:" << rdhOrbit << " triggerType:" << triggerType;
```
shahor02: isn't this too verbose?

Detectors/CTP/reconstruction/src/RawDataDecoder.cxx
```diff
@@ -193,6 +202,7 @@ int RawDataDecoder::decodeRaw(o2::framework::InputRecord& inputs, std::vector<o2
     if (mPadding == 1) {
       wordSize = 16;
     }
+    LOG(info) << ii << " payload size:" << payload.size();
```
shahor02: this too?

Detectors/CTP/reconstruction/src/RawDataDecoder.cxx
```diff
@@ -286,7 +282,20 @@ int RawDataDecoder::decodeRaw(o2::framework::InputRecord& inputs, std::vector<o2
       digits.push_back(dig.second);
     }
   }
-  return 0;
+  ret = 1;
```
shahor02: I guess this was for the test only?

Detectors/CTP/reconstruction/src/RawDataDecoder.cxx
```diff
@@ -286,7 +282,20 @@ int RawDataDecoder::decodeRaw(o2::framework::InputRecord& inputs, std::vector<o2
...
+  if (ret) {
+    if (nwrites < 1) {
+      std::ofstream dumpctp("/tmp/dumpCTP.bin", std::ios::out | std::ios::binary);
```
shahor02: it is worth to check if the file is ok
```
if (!dumpctp.good()) {
  LOGP(error, "Failed to open file {}", "/tmp/dumpCTP.bin");
} else {
  ...
}
```

Detectors/CTP/reconstruction/src/RawDataDecoder.cxx
lietava: yes

Detectors/CTP/reconstruction/src/RawDataDecoder.cxx
lietava: yes

Detectors/CTP/reconstruction/src/RawDataDecoder.cxx
lietava: yes

lietava:
@shahor02 : thanks, yes, it was only draft. I removed debugs and added option for number of errors to print.

davidrohr:
If the code is ok now, would be great to merge it today. Then we could deploy it tomorrow with the software update.

lietava:
There is one error check but it does not seems to be ctp related.

shahor02 (approved):

PR #11660, opened 2023-07-17
Implementing derivative map to O2

ChSonnabend:


ChSonnabend:
Pinging @wiechula

davidrohr (commented):
This looks of for the change in the TPCFastTransform now, since it doesn't change the class layout.
However, I just saw that you added the additional command line option only for gpu-reco-workflow.cxx.
Could you add it also for the tpc-reco-workflow, in Detectors/TPC/workflow?

ChSonnabend:
Hi,
Thanks for the comment, indeed I missed that it is in the gpu workflow but not in the tpc wrokflow. However I'm not fully sure now: I add it as an option in the CorrectionMapsLoader.cxx, so I don't even know if any of the worflows themselves would need it as an input, it would automatically be an option if the correction maps loader is used. So wouldn't it be better to remove it also from the gpu workflow?

GPU/TPCFastTransformation/CorrectionMapsHelper.h
```diff
@@ -38,22 +38,22 @@ class CorrectionMapsHelper
...
   GPUd() void InverseTransformYZtoX(int slice, int row, float y, float z, float& x) const
   {
-    mCorrMap->InverseTransformYZtoX(slice, row, y, z, x, mCorrMapRef, mLumiScale);
+    mCorrMap->InverseTransformYZtoX(slice, row, y, z, x, mCorrMapRef, mLumiScale, mLumiScaleMode);
   }
 
   GPUd() void InverseTransformYZtoNominalYZ(int slice, int row, float y, float z, float& ny, float& nz) const
   {
-    mCorrMap->InverseTransformYZtoNominalYZ(slice, row, y, z, ny, nz); // FIXME: Wrong, but call at least one of the functions to set ny / nz to non-bogus values
+    mCorrMap->InverseTransformYZtoNominalYZ(slice, row, y, z, ny, nz, mCorrMapRef, mLumiScale, mLumiScaleMode); // FIXME: Wrong, but call at least one of the functions to set ny / nz to non-bogus values
```
davidrohr: Did you test if this works correctly? If yes, I'd remove the comment.

Detectors/TPC/calibration/src/CorrectionMapsLoader.cxx
```diff
@@ -64,6 +71,7 @@ void CorrectionMapsLoader::addOptions(std::vector<ConfigParamSpec>& options)
 {
   addOption(options, ConfigParamSpec{"corrmap-lumi-mean", VariantType::Float, 0.f, {"override TPC corr.map mean lumi (if > 0), disable corrections if < 0"}});
   addOption(options, ConfigParamSpec{"corrmap-lumi-inst", VariantType::Float, 0.f, {"override instantaneous CTP lumi (if > 0) for TPC corr.map scaling, disable corrections if < 0"}});
+  addOption(options, ConfigParamSpec{"corrmap-lumi-mode", VariantType::Int, 0, {"scaling mode: (default) 0 = static + scale * full; 1 = full + scale * derivative"}});
```
davidrohr, line 74: I agree, in principle we have the option here. So I don't understand why we need it in the workflow itself as well. @shahor02 : could you clarify why you added the options to the workflow binaries in the first place?

ChSonnabend: As discussed it's probably to set the option separately in different workflows, for now I kept the option only in the CorrectionMapsLoader. If needed in the future we can add it also to the workflows.

GPU/TPCFastTransformation/CorrectionMapsHelper.h
ChSonnabend: It seems to work without problems on local tests

wiechula (approved):
Looks cool for me

ChSonnabend:
Nice! I think the build errors are again unrelated to these changes...

davidrohr:
looks good, will merge once the CI is through

PR #11661, opened 2023-07-17
TRD new structure for Tracklet64

martenole:
@tdietel and @pestratm I tried to restructure a bit the Tracklet64 class. It should contain now all the functionality needed also for the raw data display. Most of the methods are simple one liners re-using the other methods which are already there.

Concerning the shift of 2 pads in `getPadCol()` this we should also discuss together. What I did was to change the PH class a bit to search for each tracklet which was matched to a tracklet for the closest digit which is a local maximum. For that I found an offset of +2.2 pads wrt my expectation.

But this was already implemented before, so this PR only adds some helper methods. All the previously implemented methods remain the same.

trackletPadOffset


martenole:
CI errors unrelated, merging

PR #11662, opened 2023-07-18
Optionally override magnet currents

shahor02:
Magnet currents can be overridden to arbitrary values, in which case the validity check will only report unsupported values (if any) w/o producing fatal. E.g.
`O2_OVERRIDE_L3_CURRENT=0 O2_OVERRIDE_DIPOLE_CURRENT=5123.5`

### DIFF
--- a/Common/Field/src/MagneticField.cxx
+++ b/Common/Field/src/MagneticField.cxx
@@ -558,18 +558,29 @@ MagFieldParam::BMap_t MagneticField::getFieldMapScale(float& l3, float& dip, boo
 
   MagFieldParam::BMap_t map = MagFieldParam::k5kG;
   float sclL3, sclDip;
+  float l3sav = l3, dipsav = dip;
 
   float l3Pol = l3 > 0 ? 1 : -1;
   float diPol = dip > 0 ? 1 : -1;
 
   l3 = TMath::Abs(l3);
   dip = TMath::Abs(dip);
+  static bool overrideL3 = std::getenv("O2_OVERRIDE_L3_CURRENT") != nullptr;
+  static bool overrideDIP = std::getenv("O2_OVERRIDE_DIPOLE_CURRENT") != nullptr;
+  static bool warnL3Done = false, warnDipDone = false, warnPolarityDone = false;
 
   if (TMath::Abs((sclDip = dip / diNominalCurrent) - 1.) > tolerance && !uniform) {
     if (dip <= zero) {
       sclDip = 0.; // some small current.. -> Dipole OFF
     } else {
-      LOG(fatal) << "MagneticField::createFieldMap: Wrong dipole current (" << dip << " A)!";
+      if (!overrideDIP) {
+        LOG(fatal) << "MagneticField::createFieldMap: Wrong dipole current (" << dipsav << " A)!";
+      } else {
+        if (!warnDipDone) {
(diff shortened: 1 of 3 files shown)


davidrohr (changes requested):
In general I am fine to have the override currents configurable, however I would change few things.
- Could we prepent the env variables with `O2_`.
- Could we do the implementation in GRPMagField in the way I did it in my PR, i.e.:
  - Leave the inline function in the header, and have a static flag in there whether to override or not.
  - Static flag is then initialized by a function from the .cxx. The reason for this is that it that the initialization as one expression forces to compiler to implement the initialization in a thread-safe way, so it is generally good practice. And I'd like to have the getenv stuff out of the .h.
- Also in MagneticField.cxx, I would make the boolean flags static.

Common/Field/src/MagneticField.cxx
```diff
@@ -558,18 +558,29 @@ MagFieldParam::BMap_t MagneticField::getFieldMapScale(float& l3, float& dip, boo
...
 
   if (TMath::Abs((sclDip = dip / diNominalCurrent) - 1.) > tolerance && !uniform) {
     if (dip <= zero) {
       sclDip = 0.; // some small current.. -> Dipole OFF
     } else {
-      LOG(fatal) << "MagneticField::createFieldMap: Wrong dipole current (" << dip << " A)!";
+      if (!overrideDIP) {
+        LOG(fatal) << "MagneticField::createFieldMap: Wrong dipole current (" << dipsav << " A)!";
+      } else {
+        if (!warnDipDone) {
```
davidrohr, line 579: shouldn't this be a fatal as well? If it is an unsupported value, we don't have a field map for it, have we?

DataFormats/Parameters/src/GRPMagField.cxx
```diff
@@ -41,3 +41,25 @@ void GRPMagField::print() const
 {
   printf("magnet currents (A) L3 = %.3f, Dipole = %.f; uniformity = %s\n", getL3Current(), getDipoleCurrent(), mUniformField ? "true" : "false");
 }
+
+o2::units::Current_t GRPMagField::checkDipoleOverride()
+{
+  static float v = getenv("O2_OVERRIDE_DIPOLE_CURRENT") ? atof(getenv("O2_OVERRIDE_DIPOLE_CURRENT")) : NOOVERRIDEVAL;
+  static bool alarmShown = false;
+  if (v != NOOVERRIDEVAL && !alarmShown) {
+    LOGP(alarm, "Overriding DIPOLE current to {}", v);
```
davidrohr: One more point I forgot, I would make this an ERROR, just to be sure it is seen by the shift crew. (I think they usually filter out alarms)

davidrohr (commented):
looks good to me now. Just 2 more minor comments:

Common/Field/src/MagneticField.cxx
shahor02: We will not have good map but still reasonable scaling will be applied to existing maps. E.g one can emulate L3 Off / Dipole On field, will not be good for physics but e.g. for ED will work (better then 0/0).

DataFormats/Parameters/src/GRPMagField.cxx
shahor02: ok

PR #11664, opened 2023-07-18
TRD: PID: fixups

f3sch:
@martenole For you to look over.

I am not sure what way would be prefered for the lookup tables, just a vector<tgraph> and hardcode the momentum intervals or the class I included?

Also I did not change the paths to the ccdb yet, we can upload a preliminary object or wait for the results of the service work by leonardo?

And should I pull over the macros from the other repo so that we have the CI?


martenole:
Thanks, did not go through it yet but the CI errors are real, can you please check?
```
/Volumes/build/alice-ci-workdir/o2/sw/SOURCES/O2/11664/0/Detectors/TRD/pid/src/PIDBase.cxx:66:11: error: enumeration value 'NMODELS' not handled in switch [-Werror,-Wswitch]
  switch (policy) {
          ^~~~~~
/Volumes/build/alice-ci-workdir/o2/sw/SOURCES/O2/11664/0/Detectors/TRD/pid/src/PIDBase.cxx:66:11: note: add missing switch cases
```

Detectors/TRD/workflow/src/TRDGlobalTrackingSpec.cxx
```diff
@@ -837,21 +838,27 @@ DataProcessorSpec getTRDGlobalTrackingSpec(bool useMC, GTrackID::mask_t src, boo
...
+        inputs.emplace_back("xgb", "TRD", "XGB", 0, Lifetime::Condition, ccdbParamSpec("TRD_test/PID_new/xgb"));
         break;
+      case PIDPolicy::PY:
+        inputs.emplace_back("py", "TRD", "py", 0, Lifetime::Condition, ccdbParamSpec("TRD_test/PID_new/py"));
+        break;
+#endif
       case PIDPolicy::Dummy:
         break;
-      default:
-        throw std::runtime_error("Unable to load requested PID policy data!");
```
martenole, line 853: @f3sch since the PID policies that require ONNX are not visible to the compiler in case it is not available you cannot remove the `default` here. It triggers a `Wswitch` and since all warnings are treated as errors the compilation on mac fails

Detectors/TRD/pid/include/TRDPID/LQND.h
```diff
@@ -0,0 +1,159 @@
...
+    const auto isNegative = std::signbit(trkSeed.getSign()); // positive and negative charged particles are treated differently since ExB effects the charge distributions
+    const auto& trackletsRaw = input.getTRDTracklets();
+    float lei0{1.f}, lei1{1.f}, lei2{1.f}, lpi0{1.f}, lpi1{1.f}, lpi2{1.f}; // likelihood per layer
+    for (int iLayer = 0; iLayer < constants::NLAYER; ++iLayer) {
+      int trkltId = trkIn.getTrackletIndex(iLayer);
+      if (trkltId < 0) { // no tracklet attached
+        continue;
+      }
+      const auto xCalib = input.getTRDCalibratedTracklets()[trkIn.getTrackletIndex(iLayer)].getX();
+      if (!o2::base::Propagator::Instance()->PropagateToXBxByBz(trk, xCalib, o2::base::Propagator::MAX_SIN_PHI, o2::base::Propagator::MAX_STEP, o2::base::Propagator::MatCorrType::USEMatCorrNONE)) {
```
martenole: I think this rather expensive full propagation is not necessary. The tgl should be rather constant. You could use `getSnpAt(o2::math_utils::sector2Angle(input.getTRDTracklets()[trkIn.getTrackletIndex(iLayer)]), xCalib, bz)`. The `bz` you can get via `o2::base::Propagator::Instance()->getNominalBz()`.

Detectors/TRD/pid/include/TRDPID/LQND.h
```diff
@@ -0,0 +1,159 @@
...
+
+namespace o2
+{
+namespace trd
+{
+namespace detail
+{
+/// Lookup Table class for ccdb upload
+template <int nDim>
+class LUT
```
martenole, line 45: I like the way to have the LUT as a class with a configurable number of dimensions. Would it be possible to make it accept not only 1 and 3D, but also things like 2 or 7D for example? I think we had such LUTs in the past in Run 2

Detectors/TRD/pid/src/ML.cxx
```diff
@@ -135,52 +103,57 @@ PIDValue ML::calculate(const TrackTRD& trkTRD, const o2::globaltracking::RecoCon
...
-      in[NLAYER + iLayer * NCHARGES + 2] = -1.f;
+      // no tracklet attached, fill with default values e.g. charge=-1.,
+      in[iLayer * NCHARGES + 0] = -1.f;
+      in[iLayer * NCHARGES + 1] = -1.f;
+      in[iLayer * NCHARGES + 2] = -1.f;
+      in[18 + iLayer] = -1.f;
       continue;
+    } else {
+      const auto xCalib = input.getTRDCalibratedTracklets()[trkTRD.getTrackletIndex(iLayer)].getX();
+      if (!o2::base::Propagator::Instance()->PropagateToXBxByBz(trk, xCalib, o2::base::Propagator::MAX_SIN_PHI, o2::base::Propagator::MAX_STEP, o2::base::Propagator::MatCorrType::USEMatCorrNONE)) {
```
martenole: as above

martenole (commented):
In general looks all good. Just some minor comments on the track extrapolation and I think it would be nice to have the LUT class a bit more flexible.
For the CCDB objects I would also wait for Leonardo first

Detectors/TRD/pid/include/TRDPID/LQND.h
f3sch: 7D seems nonsensical since we only get 3 windows from the FEE anyways, this was different in Run 2 were we would get 7 windows, I think.
How would you combine the windows in the 2D case (Q0+Q2 and Q1, seems the most resonable to me?)

martenole: ok 7D is probably indeed not needed, but 2D I could imagine (although no idea which windows would be best at the moment). Could you not get rid of the whole `if constexpr` block of the class? For 1D it will anyhow be called with iDim = 0 only and so you could return `mLUTS[index + iDim]` always, no?

martenole (approved):

PR #11665, opened 2023-07-18
Mark MC particles reconstructed in calorimeters for storage

nburmaso:
Marking MC particles reconstructed as signals in EMCAL and PHOS as "reconstructed" to fix issue described in https://alice.its.cern.ch/jira/browse/O2-3799


sawenzel (approved):
Thanks. Looks good to me.

PR #11670, opened 2023-07-19
CCDBDownloader: Restoring asynchronous functions

TrifleMichael:
Added back:
- Asynchronous batch perform
- Asynchronous batch perform with callback
- Tests for both functions

Additional information:
- To ensure asynchronous performs are executed correctly one must run the mUVLoop often enough to ensure proper reading from sockets.
- Asynchronous perform functions return a structure which must be freed after all downloads scheduled via that perform finish. The fields contained in that structure don't have to be freed as they use shared pointers.
- Callbacks are launched via uv_work, which means they will be executed asynchronously on another thread (managed by uv_loop) as long as the mUVLoop is running.

Additionally the code for old tests functions has been revised and compressed.

### DIFF
--- a/CCDB/include/CCDB/CCDBDownloader.h
+++ b/CCDB/include/CCDB/CCDBDownloader.h
@@ -28,6 +28,7 @@ typedef struct uv_poll_s uv_poll_t;
 typedef struct uv_signal_s uv_signal_t;
 typedef struct uv_async_s uv_async_t;
 typedef struct uv_handle_s uv_handle_t;
+typedef struct uv_work_s uv_work_t;
 
 using namespace std;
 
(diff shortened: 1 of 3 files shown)


davidrohr:
Is this done in a way that async batch transfer can be easily enabled / disabled?
I.e., if we deploy this at P2, can we configure it to run in the way we are currently running?

TrifleMichael:
> Is this done in a way that async batch transfer can be easily enabled / disabled? I.e., if we deploy this at P2, can we configure it to run in the way we are currently running?

Asynchronous transfers are scheduled using separate functions from blocking ones. Unless `batchAsynchPerform` or `batchAsynchWithCallback` is used the downloads will be done in a blocking manner as it is done now.

ktf:
Thank you for your PR. I am currently on holidays, so I did not have a deep look into this, however on a first glance 
it's not clear to me why would you need all those shared pointers. Can you do without? To me it seems that you never have a case where there is more than one place where the destruction can happen.

TrifleMichael:
> Thank you for your PR. I am currently on holidays, so I did not have a deep look into this, however on a first glance it's not clear to me why would you need all those shared pointers. Can you do without? To me it seems that you never have a case where there is more than one place where the destruction can happen.

The shared pointers are used only in AsynchronousResults and structures that copy it's fields. There are two advantages of that:
- The user can call `delete` on AsynchronousResults without worrying about deleting the pointers inside said structure.
- If the user deletes the AsynchronousResults too early the program won't break.

Of course this functionality is not obligatory and I can remove it.

ktf:
@TrifleMichael Thank you for your clarification. Do I understand correctly that the idea is that you use the shared pointers to retrive the information about a single element of the batch? Then I think what you have is not thread safe and should be reworked in terms of std::promise / std::future. 

That said, I would start slow and rework the API so that the Callback returns a single "promise like" object which you can wait on for completion. Maybe we can discuss tomorrow?

PR #11676, opened 2023-07-20
Protection in angle2sector conversion

shahor02:


shahor02:
@davidrohr this is the only thing which comes to my mind as a possible reason for the crash in aod-producer you have seen.

PR #11680, opened 2023-07-23
New MFT barrel

frmanso:
Full new version of the central MFT barrel, with new dimensions, new positioning. Added also the conical A-side MFT barrel but commented for the time being.


robincaron13:
Hi @frmanso, thanks a lot for this new version of MFT barrel. Could you please fix the `rin_wireA` size array regarding the build error? Did you do your own checks to ensure that there is no overlaps somewhere or agreement/comparison with previous geometry.

Detectors/ITSMFT/MFT/base/src/Barrel.cxx
```diff
@@ -235,243 +215,390 @@ TGeoVolumeAssembly* Barrel::createBarrel()
...
+    xPosIniA = rin_wireA[l] * TMath::Sin(51.0 * kDegRad);
+    for (Int_t k = 0; k < 5; k++) {
+      xPosIniA = xPosIniA - radiia;
+      Float_t yPosIniA = sqrt(rin_wireA[l] * rin_wireA[l] - xPosIniA * xPosIniA);
+      BarrelVolume->AddNode(BarrelWiresA, 1, new TGeoTranslation(xPosIniA, -yPosIniA, 81.325));
+      BarrelVolume->AddNode(BarrelWiresA, 1, new TGeoTranslation(-xPosIniA, -yPosIniA, 81.325));
+      xPosIniA = xPosIniA - 0.03;
+    }
+  }
+  Double_t xIsol = (rin_wireA[2] - 0.1) * TMath::Sin(50.5 * kDegRad);
```
robincaron13: Could you fix this index array of `rin_wireA`?

Detectors/ITSMFT/MFT/base/src/Barrel.cxx
```diff
@@ -235,243 +215,390 @@ TGeoVolumeAssembly* Barrel::createBarrel()
...
+  Double_t yIsol = sqrt((rin_wireA[2] - 0.1) * (rin_wireA[2] - 0.1) - xIsol * xIsol);
```
robincaron13: same here

frmanso:
Robin,
     I don't see any particular message related to rin_wireA, what's the matter? About the overlaps, I made all verifications to avoid this kind of issue.

robincaron13:
Hi Frank, this is what I see from the build log:
 `/Volumes/build/alice-ci-workdir/o2/sw/SOURCES/O2/11680/0/Detectors/ITSMFT/MFT/base/src/Barrel.cxx:384:26: error: array index 2 is past the end of the array (which contains 2 elements) [-Werror,-Warray-bounds]
  Double_t yIsol = sqrt((rin_wireA[2] - 0.1) * (rin_wireA[2] - 0.1) - xIsol * xIsol);`
  `rin_wireA` is defined with 2 elements but you are trying to access the third one.

frmanso:
That's right, the number used in the array rin_wireA is too large, it should be 0 insteed of 2. I don't understand why I didn't see this error while compiling and performing simulations. I used the usual "aliBuild build O2 --defaults o2" to compile without any warning.

robincaron13:
Hi @frmanso so you confirm it is the right value `rin - 0.82` to be set in the definition of `xIsol` ? (not the reduced one `rin - 0.82 - 2 * radiia`)
I'll approve the changes, thanks

robincaron13 (approved):

PR #11681, opened 2023-07-23
First step in adding CTP inputs to data model

ddobrigk:
@lietava this will make it possible to write a converter from BCs_000 to BCs_001, keeping backwards compatibility (the old data will simply have a dummy value for the input mask that is completely empty. I can take care of writing this converter as it is rather simple (but it has to be built against a newer version of O2 so I guess we have to wait a day for the building to catch up). Keeping as draft for now. 

Then we also have to add the input mask saving to the AOD producer. This I don't know how to do but I am hoping that is easy as there might simply be an appropriate getter already somewhere :-) - let's discuss tomorrow morning!

### DIFF
--- a/Framework/Core/include/Framework/AnalysisDataModel.h
+++ b/Framework/Core/include/Framework/AnalysisDataModel.h
@@ -37,11 +37,19 @@ namespace bc
 DECLARE_SOA_COLUMN(RunNumber, runNumber, int);          //! Run number
 DECLARE_SOA_COLUMN(GlobalBC, globalBC, uint64_t);       //! Bunch crossing number (globally unique in this run)
 DECLARE_SOA_COLUMN(TriggerMask, triggerMask, uint64_t); //! CTP trigger mask
+DECLARE_SOA_COLUMN(InputMask, inputMask, uint64_t);     //! CTP input mask
 } // namespace bc
 
-DECLARE_SOA_TABLE(BCs, "AOD", "BC", o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing
+DECLARE_SOA_TABLE(BCs_000, "AOD", "BC", 
+                  o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing
                   bc::RunNumber, bc::GlobalBC,
                   bc::TriggerMask);
+DECLARE_SOA_TABLE_VERSIONED(BCs_001, "AOD", "BC", 1,
+                  o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing
+                  bc::RunNumber, bc::GlobalBC,
+                  bc::TriggerMask, bc::InputMask);
+
+using BCs = BCs_000; // current version
 using BC = BCs::iterator;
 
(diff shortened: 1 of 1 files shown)


lietava:
@ddobrigk , thanks. AOD producer - I think only couple of  lines need to be added at:
https://github.com/AliceO2Group/AliceO2/blob/ca9197ac4d879f58104813b635ea6250fe02c95d/Detectors/AOD/src/AODProducerWorkflowSpec.cxx#L2067C1-L2096C1
?
e.g. bcToInputMask to be created in analogy with bcToTriggerMask and bcCursosr parameters to be extended ?

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -37,11 +37,19 @@ namespace bc
 DECLARE_SOA_COLUMN(RunNumber, runNumber, int);          //! Run number
 DECLARE_SOA_COLUMN(GlobalBC, globalBC, uint64_t);       //! Bunch crossing number (globally unique in this run)
 DECLARE_SOA_COLUMN(TriggerMask, triggerMask, uint64_t); //! CTP trigger mask
+DECLARE_SOA_COLUMN(InputMask, inputMask, uint64_t);     //! CTP input mask
 } // namespace bc
 
-DECLARE_SOA_TABLE(BCs, "AOD", "BC", o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing
+DECLARE_SOA_TABLE(BCs_000, "AOD", "BC",
+                  o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing
```
```suggestion
DECLARE_SOA_TABLE(BCs_000, "AOD", "BC", //! Root of data model for tables pointing to a bunch crossing
                  o2::soa::Index<>,
```

Framework/Core/include/Framework/AnalysisDataModel.h
```diff
@@ -37,11 +37,19 @@ namespace bc
...
                   bc::RunNumber, bc::GlobalBC,
                   bc::TriggerMask);
+DECLARE_SOA_TABLE_VERSIONED(BCs_001, "AOD", "BC", 1,
+                            o2::soa::Index<>, //! Root of data model for tables pointing to a bunch crossing
```
```suggestion
DECLARE_SOA_TABLE_VERSIONED(BCs_001, "AOD", "BC", 1, //! Root of data model for tables pointing to a bunch crossing, version 1
                            o2::soa::Index<>,
```

lietava:
@ddobrigk : I got the bot message that if no action this PR will be close.  Probably is better to keep it open ?

jgrosseo (approved):

ddobrigk:
Just FYI I checked and indeed this straightforward commit has to be step 1 in the implementation of the CTP input masks in AO2Ds. All good to go! Thanks!

ddobrigk:
So, I admit I do not recognize these failed test errors... I suppose they are unrelated. Maybe @ktf @pzhristov @jgrosseo can take a look? It would be good to merge this so I can proceed with the converter PR. Thanks a lot!

jgrosseo:
mac CI unrelated. Merging.

PR #11692, opened 2023-07-25
newhmpiddatamodel

gvolpe79:


ddobrigk:
Hi @gvolpe79 thanks a lot for taking care of this! I would suggest two things: 
* perhaps change the name of the PR into something more meaningful. It's important for bookkeeping ;-) 
* I am not sure you need to declare the equivalency for index - I think versioned tables are automatically identical. This is more for some other use cases, such as Tracks and TracksIU, which are in principle 'unrelated' tables as far as the framework is concerned; no other versioned tables have been declared identical. But I leave to the experts to comment!

gvolpe79:
Hi @ddobrigk , if you refer to this "DECLARE_EQUIVALENT_FOR_INDEX(aod::HMPID_000, aod::HMPID_001);" Jan Fiete suggested to me to do like this, let me know what shall I do!

ddobrigk:
> 

Hi @gvolpe79 , hmmm the equivalency was declared in the past for `Collisions_000` and `StoredMcParticles_000` but not for `Zdcs_000`, `FDDs_000`, `V0s_000`, `Cascades_000` and `McCaloLabels_000`. I guess the main point is to make it such that one can de-reference indices to one of the versions to the other versions and therefore that may be why one would want it - but I am unsure if that actually comes up in everyday usage. @jgrosseo do we actually need this equivalency? If so, it would be best to declare it for all versioned tables, but I am not sure. Sorry if I miss something. Thanks!

gvolpe79:
Hi @shahor02  build/O2/o2 and build/O2/o2-cs8 are failing but I'm not sure if the errors are genuine and related to the modifications I did. If it is ok can you then merge the code?

PR #11694, opened 2023-07-25
Add STDERRMON filter for histogram division

martenole:
In case we don't have a patch from ITS for next weeks update we could merge this before we build the sw for next Monday.


martenole:
ITS has merged a fix in QC already. Still in the future we might have again histograms which are divided for which one has custom bin labels and the other not and then this produces warnings that appear as errors in the IL while the result is valid. So merging this in any case

PR #11695, opened 2023-07-25
Adding TRD t0 fitting procedure

luisabergmann:
The fits are done both for the PH distribution inclusive over all chambers and (if enough statistics are available) for the individual chambers.

added files:
- TRDPHReaderSpec reads the data produced by the PHData type
- T0FitSpec steers the procedure
- T0Fit performs the fits
- T0FitHistos is a helper class that passes the data from the TRDPHReader to T0Fit

updatet files:
- the option "t0" is added to the calibration workflow
- an option to set the average t0 is added to CalT0.h
- a default value t0 = 1.2 is added to Constants.h
- a missing library is added to PHData.h
- the required minimum entries to perform the fits are added to CalibrationParams.h

### DIFF
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/CalT0.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/CalT0.h
@@ -32,11 +32,34 @@ class CalT0
   ~CalT0() = default;
 
   void setT0(int iDet, float t0) { mT0[iDet] = t0; }
+  void setT0av(float t0) { mT0av = t0; }
 
   float getT0(int iDet) const { return mT0[iDet]; }
+  float getT0av() const { return mT0av; }
+  float calcT0av() const
+  {
+    if (mT0.size() == 0) {
+      return -1;
+    }
+    float sum = 0;
+    int counts = 0;
+    for (int iDet = 0; iDet < constants::MAXCHAMBER; ++iDet) {
+      if (mT0[iDet] > -5) {
+        sum += mT0[iDet];
+        ++counts;
+      }
+    }
+
+    if (counts > 0) {
+      return sum / counts;
+    } else {
+      return -2;
+    }
+  }
 
  private:
   std::array<float, constants::MAXCHAMBER> mT0{}; ///< calibrated T0 per TRD chamber
+  float mT0av{-1};                                ///< calibrated average T0
 
   ClassDefNV(CalT0, 1);
 };
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/CalT0.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/CalT0.h
@@ -35,7 +35,9 @@ class CalT0
   void setT0av(float t0) { mT0av = t0; }
 
   float getT0(int iDet) const { return mT0[iDet]; }
+  //getT0av() returns the average T0 obtained by fitting the data from all chambers combined
   float getT0av() const { return mT0av; }
+  //calcT0av() returns the average T0 from all individual chambers
   float calcT0av() const
   {
     if (mT0.size() == 0) {
@@ -59,7 +61,7 @@ class CalT0
 
  private:
   std::array<float, constants::MAXCHAMBER> mT0{}; ///< calibrated T0 per TRD chamber
-  float mT0av{-1};                                ///< calibrated average T0
+  float mT0av{-1};                                ///< average T0 obtained from fitting the PH data from all chambers combined
 
   ClassDefNV(CalT0, 1);
 };
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/CalT0.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/CalT0.h
@@ -35,9 +35,9 @@ class CalT0
   void setT0av(float t0) { mT0av = t0; }
 
   float getT0(int iDet) const { return mT0[iDet]; }
-  //getT0av() returns the average T0 obtained by fitting the data from all chambers combined
+  // getT0av() returns the average T0 obtained by fitting the data from all chambers combined
   float getT0av() const { return mT0av; }
-  //calcT0av() returns the average T0 from all individual chambers
+  // calcT0av() returns the average T0 from all individual chambers
   float calcT0av() const
   {
     if (mT0.size() == 0) {
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
@@ -34,7 +34,6 @@ class T0FitHistos
   T0FitHistos() = default;
   T0FitHistos(const T0FitHistos&) = default;
   ~T0FitHistos() = default;
-  void init();
   void reset();
   auto getDetector(int index) const { return mDet[index]; }
   auto getTimeBin(int index) const { return mTB[index]; }
@@ -50,7 +49,6 @@ class T0FitHistos
   std::vector<int> mTB{};
   std::vector<int> mADC{};
   size_t mNEntriesTot{0};
-  bool mInitialized{false};
 
   ClassDefNV(T0FitHistos, 1);
 };
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
@@ -1,0 +1,61 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file T0FitHistos.h
+/// \brief Class to store the TRD PH values for each TRD chamber
+
+#ifndef ALICEO2_T0FITHISTOS_H
+#define ALICEO2_T0FITHISTOS_H
+
+#include "DataFormatsTRD/Constants.h"
+#include "DataFormatsTRD/PHData.h"
+#include "Framework/InputRecord.h"
+#include "Rtypes.h"
+#include <vector>
+#include <memory>
+#include <gsl/span>
+
+namespace o2
+{
+namespace trd
+{
+
+class T0FitHistos
+{
+ public:
+  T0FitHistos() = default;
+  T0FitHistos(const T0FitHistos&) = default;
+  ~T0FitHistos() = default;
+  void init();
+  void reset();
+  auto getDetector(int index) const { return mDet[index]; }
+  auto getTimeBin(int index) const { return mTB[index]; }
+  auto getADC(int index) const { return mADC[index]; }
+  auto getNEntries() const { return mNEntriesTot; }
+
+  void fill(const std::vector<o2::trd::PHData> data);
+  void merge(const T0FitHistos* prev);
+  void print();
+
+ private:
+  std::vector<int> mDet{};
+  std::vector<int> mTB{};
+  std::vector<int> mADC{};
+  size_t mNEntriesTot{0};
+  bool mInitialized{false};
+
+  ClassDefNV(T0FitHistos, 1);
+};
+
+} // namespace trd
+} // namespace o2
+
+#endif // ALICEO2_T0FITHISTOS_H
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
@@ -39,7 +39,7 @@ class T0FitHistos
   auto getADC(int index) const { return mADC[index]; }
   auto getNEntries() const { return mNEntriesTot; }
 
-  void fill(const std::vector<o2::trd::PHData> data);
+  void fill(const std::vector<o2::trd::PHData>& data);
   void merge(const T0FitHistos* prev);
   void print();
 
--- a/DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
+++ b/DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
@@ -34,7 +34,6 @@ class T0FitHistos
   T0FitHistos() = default;
   T0FitHistos(const T0FitHistos&) = default;
   ~T0FitHistos() = default;
-  void reset();
   auto getDetector(int index) const { return mDet[index]; }
   auto getTimeBin(int index) const { return mTB[index]; }
   auto getADC(int index) const { return mADC[index]; }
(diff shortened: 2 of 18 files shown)


Detectors/TRD/calibration/src/T0Fit.cxx
```diff
@@ -0,0 +1,270 @@
...
+  for (int i = lowerBoundFit; i <= upperBoundFit; ++i) {
+
+    if (TMath::Abs(y[i]) < 1e-7) {
+      continue;
+    }
+
+    double funcVal = par[0] * TMath::Erf(x[i]) + par[1] * TMath::Landau(x[i], par[2], par[3]);
+
+    sum2 += TMath::Power(y[i] - funcVal, 2) / y[i];
+    // sum2 += TMath::Power((y[i] - funcVal)/yErr[i],2);
```
martenole: could you remove this line in case its not needed?

Detectors/TRD/workflow/include/TRDWorkflow/T0FitSpec.h
```diff
@@ -0,0 +1,177 @@
...
+#include "Framework/ControlService.h"
+#include "Framework/WorkflowSpec.h"
+#include "CCDB/CcdbApi.h"
+#include "CCDB/CcdbObjectInfo.h"
+#include "Framework/CCDBParamSpec.h"
+#include "DetectorsBase/GRPGeomHelper.h"
+#include <chrono>
+
+#include "TH2.h"
+#include "TH1.h"
```
martenole: I don't think the histogram includes are needed here and would remove them

Detectors/TRD/calibration/src/T0Fit.cxx
```diff
@@ -0,0 +1,270 @@
...
+  // set tree addresses
+  if (mEnableOutput) {
+    mOutTree->Branch("t0_chambers", &t0_chambers);
+    mOutTree->Branch("t0_average", &t0_average);
+  }
+
+  mInitDone = true;
+}
+
+void T0Fit::retrievePrev(o2::framework::ProcessingContext& pc)
```
martenole: I don't see where you are using the information from the previous CCDB object and since we want to only monitor T0 I don't think that is needed. I would simply remove this function and as you are already doing it assign a dummy value in case the fit is invalid

Detectors/TRD/calibration/src/T0Fit.cxx
```diff
@@ -0,0 +1,270 @@
...
+
+    mFitter.FitFCN();
+    fitResult = mFitter.Result();
+
+    if (fitResult.IsValid()) {
+      mFuncErfLandau->SetParameters(fitResult.GetParams()[0], fitResult.GetParams()[1], fitResult.GetParams()[2], fitResult.GetParams()[3]);
+      t0_chambers[iDet] = mFuncErfLandau->GetMaximumX();
+    } else {
+      LOG(info) << "t0 fit for chamber " << iDet << " is not valid, set to -5";
+      t0_chambers[iDet] = -5;
```
martenole: Maybe you can add a member to your fit class for the dummy value `int mDummyT0 = -5` and use that everywhere?

Detectors/TRD/calibration/src/T0Fit.cxx
```diff
@@ -0,0 +1,270 @@
...
+#include "TStopwatch.h"
+#include "CCDB/CcdbApi.h"
+#include "CCDB/BasicCCDBManager.h"
+#include "CommonUtils/NameConf.h"
+#include "CommonUtils/MemFileHelper.h"
+#include <TFile.h>
+#include <TTree.h>
+
+#include "TMath.h"
+#include "TCanvas.h"
```
martenole:
```suggestion
```

Detectors/TRD/calibration/include/TRDCalibration/T0Fit.h
```diff
@@ -0,0 +1,111 @@
...
+
+  /// Initialize the fit values once with the previous valid ones if they are
+  /// available.
+  void retrievePrev(o2::framework::ProcessingContext& pc);
+
+ private:
+  bool mInitDone{false};                                         ///< flag to avoid creating output etc multiple times
+  const TRDCalibParams& mParams{TRDCalibParams::Instance()};     ///< reference to calibration parameters
+  size_t mMinEntriesTotal{mParams.minEntriesTotalT0Fit};         ///< minimum total number of angular deviations
+  size_t mMinEntriesChamber{mParams.minEntriesChamberT0Fit};     ///< minimum number of angular deviations per chamber for accepting refitted value
```
martenole: why are you not using directly `mParams.minEntriesChamberT0Fit` instead of creating another member to the class?

DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
```diff
@@ -0,0 +1,61 @@
...
+namespace trd
+{
+
+class T0FitHistos
+{
+ public:
+  T0FitHistos() = default;
+  T0FitHistos(const T0FitHistos&) = default;
+  ~T0FitHistos() = default;
+  void init();
```
martenole: I don't think this `init()` method is needed

DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
```diff
@@ -0,0 +1,61 @@
...
+  void reset();
```
martenole: for `reset()` I did not check, but is it used anywhere?

DataFormats/Detectors/TRD/include/DataFormatsTRD/CalT0.h
```diff
@@ -32,11 +32,34 @@ class CalT0
   ~CalT0() = default;
 
   void setT0(int iDet, float t0) { mT0[iDet] = t0; }
+  void setT0av(float t0) { mT0av = t0; }
 
   float getT0(int iDet) const { return mT0[iDet]; }
+  float getT0av() const { return mT0av; }
+  float calcT0av() const
```
martenole, line 41: here I was a bit confused first that you have in principle two getters for the average T0. Could you please add some comment(s) that mT0av is the average value that you get when you take the PH data from all chambers combined and fit that and `calcT0av` returns the average T0 from all individual detectors?

martenole (commented):
Hi Luisa, thanks a lot this looks already very good. I only have some minor comments (cleanup, documentation) so we should be able to merge this soon

Detectors/TRD/calibration/include/TRDCalibration/T0Fit.h
luisabergmann: no real reason, other than I did the same for the total number of entries. Should I change mMinEntriesTotal as well, or leave that as is for readability (in hasEnoughData())?

DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
luisabergmann: I added it because it was implemented in both the vdrift and gain calibration container classes. But as far as I can tell, it is not used by the base class

DataFormats/Detectors/TRD/include/DataFormatsTRD/CalT0.h
luisabergmann: added the comments

Detectors/TRD/calibration/src/T0Fit.cxx
luisabergmann: removed the commented line

Detectors/TRD/workflow/include/TRDWorkflow/T0FitSpec.h
luisabergmann: removed the histogram includes (here and in T0Fit.h)

Detectors/TRD/calibration/src/T0Fit.cxx
luisabergmann: removed the function

Detectors/TRD/calibration/src/T0Fit.cxx
luisabergmann: added the dummy value

Detectors/TRD/calibration/src/T0Fit.cxx
luisabergmann: removed TCanvas include

DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
luisabergmann: removed init()

DataFormats/Detectors/TRD/include/DataFormatsTRD/CalT0.h
luisabergmann: added the comments

Detectors/TRD/calibration/include/TRDCalibration/T0Fit.h
martenole: I'd do the same also for the other variable

DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
luisabergmann: Correction: There is a reset() method in DetectorCalibration/TimeSlotCalibration.h:

```
virtual void reset()
  { // reset to virgin state (need for start - stop - start)
    mSlots.clear();
    mLastClosedTF = 0;
    mFirstTF = 0;
    mMaxSeenTF = 0;
    mLastCheckedTFInfiniteSlot = 0;
    mWasCheckedInfiniteSlot = false;
    initOutput();
  }
```

So it is used, and at the moment redefined by the derived class.
Should I leave it in or remove and let it use the base method?

Detectors/TRD/calibration/include/TRDCalibration/T0Fit.h
luisabergmann: replaced both variables with the direct instance of the calibration parameters

DataFormats/Detectors/TRD/include/DataFormatsTRD/T0FitHistos.h
martenole: These methods are not related. The `TimeSlotCalibration` is the 'manager class' which handles all given slots and performs the calibration when some slots have enough statistics reached for example. The `reset` handles the case when a run is stopped and a new one started. In that case all the remaining slots are removed. Your `T0FitHistos` class is the 'container class'. Each slot knows its first and last seen TF, some statistics and one container. The container is never reset as far as I can see. You can merge it with other containers or remove it or create new ones. You can check also `Detectors/Calibration/include/DetectorsCalibration/TimeSlot.h` where the required functionality for a container class is defined

luisabergmann: ok, so it's not used and as per the readme only print, merge and fill need to be implemented. I removed reset()

DataFormats/Detectors/TRD/src/T0FitHistos.cxx
```diff
@@ -0,0 +1,54 @@
...
+/// \brief Class to store the TRD PH values for each chamber
+
+#include "DataFormatsTRD/T0FitHistos.h"
+#include <fairlogger/Logger.h>
+#include <algorithm>
+
+using namespace o2::trd;
+using namespace o2::trd::constants;
+
+void T0FitHistos::fill(const std::vector<o2::trd::PHData> data)
```
f3sch:
```suggestion
void T0FitHistos::fill(const std::vector<o2::trd::PHData>& data)
```

DataFormats/Detectors/TRD/src/T0FitHistos.cxx
```diff
@@ -0,0 +1,54 @@
...
+{
+  for (auto ph : data) {
```
f3sch:
```suggestion
  for (const auto& ph : data) {
```

DataFormats/Detectors/TRD/src/T0FitHistos.cxx
```diff
@@ -0,0 +1,54 @@
...
+    mDet.push_back(prev->getDetector(i));
+    mTB.push_back(prev->getTimeBin(i));
+    mADC.push_back(prev->getADC(i));
+  }
+}
+
+void T0FitHistos::print()
+{
+  LOG(info) << "There are " << mNEntriesTot << " entries in the container";
+}
```
f3sch, line 54: This is not used anywhere, right? Seems like a leftover debugging function.

Detectors/TRD/calibration/src/T0Fit.cxx
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  mFitFunctor.lowerBoundFit = 0;
+  mFitFunctor.upperBoundFit = 5;
+
+  double mParamsStart[4];
+  mParamsStart[0] = 100;
+  mParamsStart[1] = 500;
+  mParamsStart[2] = 0.5;
+  mParamsStart[3] = 0.5;
```
f3sch, line 96: These magical constants seem all like they should be moved to be configurable parameter.

Detectors/TRD/calibration/src/T0Fit.cxx
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  mFitter.FitFCN();
+  auto fitResult = mFitter.Result();
+
+  if (fitResult.IsValid()) {
+    mFuncErfLandau->SetParameters(fitResult.GetParams()[0], fitResult.GetParams()[1], fitResult.GetParams()[2], fitResult.GetParams()[3]);
+    t0_average = mFuncErfLandau->GetMaximumX();
+  } else {
+    LOG(info) << "t0 fit for inclusive distribtion is not valid, set to " << mDummyT0;
```
f3sch:
```suggestion
    LOG(warn) << "t0 fit for inclusive distribtion is not valid, set to " << mDummyT0;
```

f3sch (changes requested):
@luisabergmann very nice work, thank you! 
I have some very minor suggestions.

DataFormats/Detectors/TRD/src/T0FitHistos.cxx
luisabergmann: I'm not sure if it's used, but it's one of the methods that's required to be implemented regarding to the README in https://github.com/AliceO2Group/AliceO2/tree/dev/Detectors/Calibration

Detectors/TRD/calibration/src/T0Fit.cxx
luisabergmann: How would I do that? Do I add lines in T0FitSpec.h like for "sec-per-slot" or "max-delay" ?

f3sch: Sorry to be unclear, I meant to put them in CalibrationParams.h, that way you can control them from the outside later on (if the need ever arises)

DataFormats/Detectors/TRD/src/T0FitHistos.cxx
f3sch: Ah good point. Sorry for the noise.

f3sch:
As a general point, it would be _really_ nice if you could document the procedure and the parameters briefly in the readme :)

martenole (approved):
I think it looks very good as it is now. I am going to merge it and further improvements can of course still be added in followup PRs :)

PR #11696, opened 2023-07-25
ctpdev: fix for dumping raw data at epn

lietava:


Detectors/CTP/reconstruction/src/RawDataDecoder.cxx
```diff
@@ -290,7 +290,7 @@ int RawDataDecoder::decodeRaw(o2::framework::InputRecord& inputs, std::vector<o2
     }
   }
   // ret = 1;
-  if (ret) {
+  if (mErrorIR || mErrorTCR) {
     if (nwrites < mErrorMax) {
```
martenole, line 296: fine with me to not reset the error counters so that you can print the statistics at end of stream. But then I would dump only a single error per decoder. Otherwise you would now write the data for a single TF with error to /tmp/dumpCTP0.bin and then unconditionally always write meaningless /tmp/dumpCTP1.bin and /tmp/dumpCTP2.bin until mErrorMax is reached

lietava:
@martenole : I added sticky bit, please, check

martenole (approved):
Thanks, looks good

lietava:
@shahor02 : does not look like CTP errors ?

PR #11703, opened 2023-07-27
[EMCAL-660] Add STU link support in getLinkAssignment

mfasDa:


mfasDa:
STU DDLs are on equipment 44 and 45. Link / C-RORC IDs in the ID region from 40-43 are not used, consequently they are set to -1. Since STU DDLs are read out with different cards on different FLPs the array for the card ID needed to be restructured, since the same card ID for consecutive SRUs (supermodule) starting from 0 does not hold for STU DDLs.

shahor02 (approved):

PR #11704, opened 2023-07-27
O2sim: draft version of HEPMC to MCHEADER/MCTracks publisher

aalkin:
DPL device to inject MCHEADER/MCTracks messages, to be used in conjunction with mctracks-to-aod converter. Supports using FIFO pipes as input files thanks to HepMC reader, so the events can be injected on-the-fly (how it is done with Rivet for example). 

```sh
mkfifo /tmp/fifo.hepm
generator -o /tmp/fifo.hepmc &
o2-sim-hepmc-publisher --hepmc /tmp/fifo.hepmc | o2-sim-mctracks-to-aod | ...
```


ktf:
Only adds an additional workflow. Merging.

PR #11705, opened 2023-07-27
[QC-869] Produce moving windows of selected merged objects

knopers8:


Barthelemy (approved):
Thank you, very good !

As discussed, this is potentially a lot more versions to store in the QCDB. We must think about a rigorous clean up policy for those.

PR #11707, opened 2023-07-28
Add labelled string array

njacazio:


njacazio:
@aalkin what do you think?

aalkin:
It is totally possible, but there is many more places where code needs to be modified (do a `git grep LabeledArray` to get an idea) and you should also add a test in `Framework/Core/test/test_Variants.cxx`.

njacazio:
Hi @aalkin thanks a lot! Indeed that would be very nice to have

njacazio:
@aalkin  I think it should be ok, what do you think?

aalkin:
@ktf could you please check this PR?

Framework/Core/test/test_Variants.cxx
```diff
@@ -218,14 +218,20 @@ TEST_CASE("Array2DTest")
 TEST_CASE("LabeledArrayTest")
 {
   float m[3][4] = {{0.1, 0.2, 0.3, 0.4}, {0.5, 0.6, 0.7, 0.8}, {0.9, 1.0, 1.1, 1.2}};
+  std::string mS[3][4] = {{"a", "b", "c", "d"}, {"e", "f", "g", "h"}, {"i", "l", "m", "n"}};
```
ktf, line 221: Any reason to have this rectangular? Can you modify it to be variable sized or even better add a second case with it? What happens if the arrays are empty? What happens if they are size one? It would be nice to have  the test extended to cover those cases as well.

ktf (approved):

ktf:
Assuming the test would have triggered in case of big issues. Merging. Please update the test to cover more cases.

PR #11710, opened 2023-07-29
minor fix for TOF noisy in sim

noferini:
This is a fix to take into account the threshold for noisy (1kHz, 10 kHz, 100 kHz) set in clusterization to be consistent with data (current default -> level 1 -> 10 kHz).
It has not a big impact (< 0.1% in pp) but it should be corrected.


noferini:
sorry, I opened in the wrong branch... closed

PR #11713, opened 2023-07-30
Split SVs to index and kinematics parts, run str.tracking from SVertexer

shahor02:
V0, Cascade and 3-Body decay objects are split into indices (of prongs, PV...) and optional kinematics objects. By default, only indices will be produced. To have also kinematics filled one should use `o2-secondary-vertexing-workflow` with options e.g. `--configKeyValues svertexer.createFullV0s=true;svertexer.createFullCascades=true;svertexer.createFull3Bodies=true`.

Since the strangeness tracker needs secondary vertices kinematics, by default it will be running from the `o2-secondary-vertexing-workflow`.  One can still run it as a separate `o2-strangeness-tracking-workflow` receiving input from the `o2-secondary-vertexing-workflow` (or root file readers) provided the `o2-secondary-vertexing-workflow` was run with the request of full secondary vertices creation and with extra option `--disable-strangeness-tracker`.

### DIFF
--- a/Common/DCAFitter/CMakeLists.txt
+++ b/Common/DCAFitter/CMakeLists.txt
@@ -9,7 +9,7 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-#add_compile_options(-O0 -g -fPIC)
+# add_compile_options(-O0 -g -fPIC)
 
 o2_add_library(DCAFitter
                TARGETVARNAME targetName
--- a/Common/DCAFitter/CMakeLists.txt
+++ b/Common/DCAFitter/CMakeLists.txt
@@ -9,6 +9,8 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
+#add_compile_options(-O0 -g -fPIC)
+
 o2_add_library(DCAFitter
                TARGETVARNAME targetName
                SOURCES src/DCAFitterN.cxx
--- a/Common/DCAFitter/include/DCAFitter/DCAFitterN.h
+++ b/Common/DCAFitter/include/DCAFitter/DCAFitterN.h
@@ -112,6 +112,9 @@ class DCAFitterN
     return std::array<float, 3>{float(vd[0]), float(vd[1]), float(vd[2])};
   }
 
+  ///< return position of quality-ordered candidate in the internal structures
+  int getCandidatePosition(int cand = 0) const { return mOrder[cand]; }
+
   ///< return Chi2 at PCA candidate (no check for its validity)
   float getChi2AtPCACandidate(int cand = 0) const { return mChi2[mOrder[cand]]; }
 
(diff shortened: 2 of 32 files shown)


shahor02:
@mpuccio @fmazzasc still have to validate, but please have a look at the str.tracker modifications (hopefully minimal). 
In brief, since the existence of full V0s or cascades (with prongs info) is guaranteed only in the svertexer checkV0 and checkCascade methods, I've modified the StrangenessTracker so that its V0 and cascade processing can be done per object with external call. The functionality of the StrangenessTracker in a standalone mode should be unchanged.

fmazzasc:
Hi @shahor02 , thanks! I will test the new version of the workflow tomorrow and let you know!

fmazzasc:
Hi @shahor02 , thanks! I will test the new version of the workflow tomorrow and let you know!

shahor02:
@fmazzasc It seems all necessary fixes were applied, 

@mconcas your `Detectors/ITSMFT/ITS/postprocessing/studies/src/AvgClusSize.cxx` seems to be the only code which requires the kinematics of the secondary vertices. Once this PR is merged, it will not be produced anymore by default (and already produced root files will not be fully readable). So, one would need to either run a special SV reconstruction with the full output (as indicated in the PR description) or to add to your code recalculation of V0/cascade kinematics from prongs indices and various tracks, as it is done in the AOD analysis...

mconcas:
> @fmazzasc It seems all necessary fixes were applied,
> 
> @mconcas your `Detectors/ITSMFT/ITS/postprocessing/studies/src/AvgClusSize.cxx` seems to be the only code which requires the kinematics of the secondary vertices. Once this PR is merged, it will not be produced anymore by default (and already produced root files will not be fully readable). So, one would need to either run a special SV reconstruction with the full output (as indicated in the PR description) or to add to your code recalculation of V0/cascade kinematics from prongs indices and various tracks, as it is done in the AOD analysis...

Hi @shahor02 , thanks for letting me know! Sure we'll sort this out one way or the other. Thanks a lot.

knopers8:
@shahor02 We see dependency errors in O2 in QC CI checks related to this PR:
```
## sw/BUILD/O2-latest/log
--
+ grep stage
+ cut -f1 -d:
+ cmake --build . -- -j32 install
ninja: error: dependency cycle: stage/lib64/G__O2DetectorsVertexing_rdict.pcm -> stage/lib64/libO2StrangenessTracking.so -> Detectors/StrangenessTracking/tracking/CMakeFiles/O2lib-StrangenessTracking.dir/src/StrangenessTracker.cxx.o -> cmake_object_order_depends_target_O2lib-StrangenessTracking -> Detectors/StrangenessTracking/tracking/G__O2StrangenessTracking.cxx -> stage/lib64/G__O2DetectorsVertexing_rdict.pcm
--
```
Could you please have a look? Maybe we could revert this until it is fixed, since it is breaking the CI in QC.

shahor02:
@knopers8 this is not a problem of the PR, but a feature of the alibuild: the dependencies for root dictionaries generation are created by the alibuild and are not updated when cmakefiles are modified. In this case, I had to fix reverse the dependency of StrangenessTracking on Vertexing  (in fact, there was no real dependent before, it was just declared in  the makefile) to the opposite one.
Unfortunately, the only clean way to overcome this is rebuild O2 from scratch, I believe this was done in the CI of this PR.

@TimoWilken do you know why the CI of QC (e.g. https://github.com/AliceO2Group/QualityControl/pull/1910) did not pick rebuilt O2?

PR #11725, opened 2023-08-03
Move StrangenessTracking to Vertexing dir.

shahor02:
Hopefully this will allow to avoid circular dependencies in previously generated cmakefiles


shahor02:
trivial, merging

PR #11726, opened 2023-08-04
Exclude GainCalibHistos from the Root dictionary

pzhristov:
GainCalibHistos triggers inclusion of Boost polymorphic_allocator.h that is not accepted by Cling. See https://alice.its.cern.ch/jira/browse/O2-4002 for detailed discussion.

### DIFF
--- a/DataFormats/Detectors/TRD/CMakeLists.txt
+++ b/DataFormats/Detectors/TRD/CMakeLists.txt
@@ -36,7 +36,6 @@ o2_target_root_dictionary(DataFormatsTRD
                        include/DataFormatsTRD/CalibratedTracklet.h
                        include/DataFormatsTRD/DcsCcdbObjects.h
                        include/DataFormatsTRD/AngularResidHistos.h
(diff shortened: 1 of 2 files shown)


pzhristov:
@tdietel @martenole In case the class has to be present in the Root dictionary, for example because of the Root IO, then we can close this PR and make sure the Root macros are used in compiled mode.

martenole:
Hi Peter,
thanks a lot for your fix which is working fine and of course for digging in the code to find the actual problem. I would prefer https://github.com/AliceO2Group/AliceO2/pull/11734 though where I simply avoided the need to include `InputRecord.h` so that we can keep the ROOT dictionary and i/o possibilities for the classes.
Cheers,
Ole

pzhristov:
> 

Hi Ole,
I completely agree that is we can avoid the inclusion of `InputRecord.h`, we should. Please close this PR in favour of https://github.com/AliceO2Group/AliceO2/pull/11734 
Cheers,
Peter

PR #11734, opened 2023-08-08
Avoid InputRecord include in TRD data formats lib

martenole:
For `T0FitHistos` the include could simply be removed, for `GainCalibHistos` I changed the data type which is sent from the EPNs to the aggregator from a fixed size vector to simply a vector of histogram entries. Since the number of entries is usually smaller than `NBINSGAINCALIB*MAXCHAMBER` this should if at all have a positive performance impact.

Fixes the boost errors reported in https://alice.its.cern.ch/jira/browse/O2-4002 and replaces https://github.com/AliceO2Group/AliceO2/pull/11726


martenole:
TRDSynchReco

PR #11737, opened 2023-08-08
[EMCAL-565, EMCAL-566]: Improvment in speed of EMCAL online calibrations

jokonig:
- Speed of the calibration is not sufficient for Pb--Pb data taking.
- This is caused by variable binned histograms which takes too much time to fill them as the bin index cannot be calculated trivially
- The histograms for the online calibration are now changed to regular (equidistant) binning which makes them significantly faster
- As the possibility of an offline calibration with variable sized root histograms in the same code should still be possible, the root to boost function is templated to allow for both types of binning.


mfasDa (approved):
Looks good

PR #11742, opened 2023-08-08
Write MC event weight into AO2D

nburmaso:
In continuation of #11738


nburmaso:
PR related to JIRA [O2-3683](https://alice.its.cern.ch/jira/browse/O2-3683)

benedikt-voelkel (approved):

PR #11743, opened 2023-08-09
quick fix

lietava:
see: https://alice.its.cern.ch/jira/browse/O2-4028


martenole (approved):
Trivial, merging. Why didn't we get a `-Werror=bool-operation` before?

PR #11746, opened 2023-08-09
Fix missing const in GroupedCombinations

vkucera:
This prevented subscriptions to tables via reference to const.
@saganatt Please check.

### DIFF
--- a/Framework/Core/include/Framework/GroupedCombinations.h
+++ b/Framework/Core/include/Framework/GroupedCombinations.h
@@ -64,7 +64,7 @@ struct GroupedCombinationsGenerator {
     {
     }
     template <typename... T2s>
-    GroupedIterator(const GroupingPolicy& groupingPolicy, const G& grouping, std::tuple<T2s...>& associated, SliceCache* cache_)
+    GroupedIterator(const GroupingPolicy& groupingPolicy, const G& grouping, const std::tuple<T2s...>& associated, SliceCache* cache_)
       : GroupingPolicy(groupingPolicy),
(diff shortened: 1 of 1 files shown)


saganatt:
Thanks a lot! I will check for the build checks. Have you compiled it locally in the DEBUG mode or the default RELWITHDEB?

aalkin:
This was probably needed before the slicing cache was introduced, the objects had to be non-const so their internal cache could be used. Have you tested this also by running the workflows that use combinations?

vkucera:
> This was probably needed before the slicing cache was introduced, the objects had to be non-const so their internal cache could be used. Have you tested this also by running the workflows that use combinations?

Why would restricting the references in the combination code change anything about the workflows that use it?

aalkin:
Filtered combinations may require modifying the internal state of table objects. It should not be the case now, after the external cache was introduced, but just to be on the safe side it is better to test it.

vkucera:
Would O2Physics still compile if it was the case though?
In any case I'm going to compile the O2Physics master against this branch and run a workflow that uses the `Pair`. Would that be sufficient?

aalkin:
Yes, it can still compile. Any workflow with filtered combinations should do.

vkucera:
OK, by filtered combinations, do you mean combinations made out of filtered tables?

aalkin:
Made out of filtered tables and/or applying additional filters.

vkucera:
Tested with `PWGHF/HFC/Tasks/taskFlow.cxx`

vkucera:
All checks passed. Can we merge it?

saganatt (approved):
Looks all good on my side then, if Anton also agrees.

PR #11747, opened 2023-08-09
DPL: avoid ODC waiting indefinitely for the driver

ktf:
When deploying in embedded mode, ODC will wait for the driver to transition through the state machine stetes. However right now the driver itself is not a FairMQ device, nor has any special plugin, so Control will wait for a transition which will never happen and eventually kill the run.

This marks the task as expendable, telling ODC to ignore the state changes for it, hopefully working around the problem.

Being expendable is also probably a good idea, because we do not want that a bug in the GUI crashes the whole workflow.


ktf:
@TimoWilken I have the impression the x86 mac is down. Could you have a look tomorrow?

davidrohr:
Thinking about it, doesn't it mean we wait for a timeout at start of run, until ODC recognizes the driver doesn't react? I think in this case we should make the driver a FMQ device anyway, to avoid such delays.

PR #11749, opened 2023-08-09
Add ClassDefNV to GPU/TPCFastTransformation classes

shahor02:
@sgorbuno was there a special reason to not add ClassDefs for these classes. 
In principle, the ClassDef is optional for non-TObject classes, though the root documentation claims that it improves the IO performance. But I've noticed that for classes with the ClassDef this [problem](https://alice.its.cern.ch/jira/browse/O2-4029?focusedId=304773&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-304773) is more rare.


sgorbuno (approved):

sgorbuno:
@shahor02, 
I think there was no reason, I just forgot to add it.

PR #11750, opened 2023-08-10
MFT chips alignment

frmanso:
Modified MFT geometry to perform the chips alignment, avoiding overlaps. By default this option is not activated (false). To use it, add the "MFTBase.buildAlignment=true" in the o2 simulation command.


shahor02:
Hi, what is the status of this PR? Was it tested in sim/rec. I see a potential problem if it is activated: simulation will create and use geometry with gaps on the fly but reconstruction takes the geometry_aligned from CCDB. Depending how do you shift, the position of the sensors in the geometries used for sim and rec. may be different.

frmanso:
As writted in the PR comment, this feature is not activated by default. One have to use the geometry_aligned from the sim step, not from the CCDB since the geometry is not the same. This modification has not yet been tested but it's the purpose of this PR, to give the possibility to MFT people to test it.

PR #11754, opened 2023-08-11
TRD raw reader update QC statistics

martenole:
- This allows the RawData QC class to get all information from the raw data statistics so it does not need to subscribe to all TRD data anymore. That should allow to disable the sampling of the QC class so that we get the raw parsing statistics for all data.
- As follow up a PR to QC is needed to take all information from the statistics.
- Lastly another PR to O2 will follow removing the huge 165kB `mParsingErrorsByLink` array from the statistics in favor of the `mParsingErrorsLink` vector.

### DIFF
--- a/DataFormats/Detectors/TRD/CMakeLists.txt
+++ b/DataFormats/Detectors/TRD/CMakeLists.txt
@@ -45,7 +45,8 @@ o2_target_root_dictionary(DataFormatsTRD
                        include/DataFormatsTRD/KrCluster.h
                        include/DataFormatsTRD/KrClusterTriggerRecord.h
                        include/DataFormatsTRD/NoiseCalibration.h
-                        include/DataFormatsTRD/PHData.h
+                       include/DataFormatsTRD/PHData.h
+                       include/DataFormatsTRD/RawDataStats.h
                        include/DataFormatsTRD/CTF.h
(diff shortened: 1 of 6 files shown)


bazinski (dismissed):

martenole:
Hm, @ktf are the checks in here run against QC@master or against a tagged version of QC? Because the CI checks are failing with
```
/sw/slc8_x86-64/O2/11754-slc8_x86-64-local1/include/Framework/InputRecord.h:434:21: error: static assertion failed: ROOT objects need to be retrieved by pointer
```
Which has already been changed in https://github.com/AliceO2Group/QualityControl/pull/1922

ktf:
Depends on the tests. For some we use a tagged version, for some we use dev, on purpose.

ktf:
Tests specs are here https://github.com/alisw/ali-bot/tree/master/ci/repo-config/

martenole:
Thanks, although from the files there I don't see the QualityControl version explicitly set. And in the build logs I see that QC v1.113 is used. So unless @knopers8 or @Barthelemy create an intermediate QC tag and update alidist before the actual tag which will likely be done today or tomorrow there is no way to get the changes in O2 before the update which is a bit of a pity. But OK, then there is no much TRD raw QC until the sw update the week after

PR #11755, opened 2023-08-11
ITS Thr Scan bug fix: swap of indices

iravasen:
Forgot that in Thr scan and vreset2d scan the indices in the matrix with the hits are swapped. This PR is to fix that. 

@shahor02 @nicolovalle This will fix the current issue with thr scan at P2 (https://ali-bookkeeping.cern.ch/?page=log-detail&id=64005). I would merge it as soon as you can (when tests are ok). Thanks a lot!


nicolovalle:
Thank you @iravasen for the quick fix. I also suggest to merge as soon as possible. Cheers.

PR #11758, opened 2023-08-13
MRRTF-224: assign severity level to decoding errors

aferrero2707:
The errors produced by the raw decoder are now categorized according to their severity (Warning/Error/Fatal).
The `Fatal` level is currently not assigned to any error code and is added for future use.


aferrero2707:
@aphecetche could you please review this? Changes are rather simple. Thanks!

PR #11762, opened 2023-08-14
MCH: added option to output the time-clustered ROFs without filtering

aferrero2707:
The additional output containing all the time-clustered ROFs without futher filtering can be useful for diagnistics, oir example in online QC. 
The output is disabled by default, and can be enabled by setting `"MCHTimeClusterizer.enableUnfilteredOutput=true"` in the configuration keys.
The default name of the DPL message is `"UTC-F-DIGITROFS"`. It can be customized via the `--output-unfiltered-digitrofs-data-description` command-line option.

### DIFF
--- a/Detectors/MUON/MCH/TimeClustering/include/MCHTimeClustering/TimeClusterFinderSpec.h
+++ b/Detectors/MUON/MCH/TimeClustering/include/MCHTimeClustering/TimeClusterFinderSpec.h
@@ -29,6 +29,7 @@ o2::framework::DataProcessorSpec
                            std::string_view inputDigitDataDescription = "F-DIGITS",
                            std::string_view inputDigitRofDataDescription = "F-DIGITROFS",
                            std::string_view outputDigitRofDataDescription = "TC-F-DIGITROFS",
+                           std::string_view outputUnfilteredDigitRofDataDescription = "UTC-F-DIGITROFS",
                            std::string_view inputIRFrameDataDescription = "ITS/IRFRAMES");
 
 } // end namespace mch
--- a/Detectors/MUON/MCH/TimeClustering/src/digits-to-timeclusters-workflow.cxx
+++ b/Detectors/MUON/MCH/TimeClustering/src/digits-to-timeclusters-workflow.cxx
@@ -29,6 +29,7 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
 {
   workflowOptions.push_back(ConfigParamSpec{"input-digits-data-description", VariantType::String, "F-DIGITS", {"description string for the input digit data"}});
   workflowOptions.push_back(ConfigParamSpec{"input-digitrofs-data-description", VariantType::String, "F-DIGITROFS", {"description string for the input digit rofs data"}});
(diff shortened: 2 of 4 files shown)


aferrero2707:
@aphecetche this PR introduces the possibility to output the time-clustered ROFs without any filtering, something we discussed in the past and that could be useful for online QC. Could you have a look? Thanks!

Detectors/MUON/MCH/TimeClustering/include/MCHTimeClustering/TimeClusterFinderSpec.h
```diff
@@ -28,6 +28,8 @@ o2::framework::DataProcessorSpec
   getTimeClusterFinderSpec(const char* specName = "mch-time-cluster-finder",
                            std::string_view inputDigitDataDescription = "F-DIGITS",
                            std::string_view inputDigitRofDataDescription = "F-DIGITROFS",
+                           bool enableUnfilteredOutput = false,
+                           std::string_view outputUnfilteredDigitRofDataDescription = "UTC-F-DIGITROFS",
```
pillot: I would put the new arguments after "outputDigitRofDataDescription", which is set in reco-workflow.cxx and would now be mixed up with "enableUnfilteredOutput", unless you also change reco-workflow.cxx accordingly.

Detectors/MUON/MCH/TimeClustering/src/digits-to-timeclusters-workflow.cxx
```diff
@@ -39,11 +41,14 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
 WorkflowSpec defineDataProcessing(const ConfigContext& cc)
 {
   o2::conf::ConfigurableParam::updateFromString(cc.options().get<std::string>("configKeyValues"));
+  const auto& param = o2::mch::TimeClusterizerParam::Instance();
```
pillot: I would check this configurable parameter directly inside getTimeClusterFinderSpec() instead of passing it as an argument. This way the saving of the extra ouput would be activable for every workflows that use this spec (i.e. from reco-workflow too) without having to modify each of them to catch this parameter.

pillot (commented):
Laurent will make the final review but in the meantime I have a couple of suggestions if it may help.
Not sure this is what you want but currently this extra output cannot be enabled from the reco-workflow which is used in online reco. Beside the current changes will alter this workflow (see detailed comments).

aphecetche:
I do agree with Philippe comments.

Detectors/MUON/MCH/TimeClustering/include/MCHTimeClustering/TimeClusterFinderSpec.h
aferrero2707: Good point! I have done as you suggested, without changing reco-workflow.cxx

Detectors/MUON/MCH/TimeClustering/src/digits-to-timeclusters-workflow.cxx
aferrero2707: Indeed, done as suggested.

pillot (approved):

aferrero2707:
@ktf there are two CI failures that seem to be un-related to this PR. Would it be possible to re-run those CI jobs? Or should I just push some dummy commit to re-trigger the CI? Thanks!

PR #11766, opened 2023-08-14
[WIP][EMCAL-558] Make EMCAL cell uncompressed at transient level

mfasDa:
- Replace packed bitmap by regular data members
- Move encoding / decoding previously implemented
  in regular getters / setters to dedicated functions
  used by CTF coder

### DIFF
--- a/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Cell.h
+++ b/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Cell.h
@@ -192,6 +192,12 @@ class Cell
   /// \return Encoded bit representation
   uint16_t getCellTypeEncoded() const;
 
+  static uint16_t encodeEnergyV0(float energy);
+  static uint16_t encodeEnergyV1(float energy, ChannelType_t celltype);
+  static uint16_t V0toV1(uint16_t energybits, ChannelType_t celltype);
+  static float decodeEnergyV0(uint16_t energybits);
+  static float decodeEnergyV1(uint16_t energybits, ChannelType_t celltype);
+
  private:
   /// \brief Set cell energy from encoded bit representation (from CTF)
   /// \param energyBits Bit representation of energy
--- a/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Cell.h
+++ b/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Cell.h
@@ -13,6 +13,7 @@
 #define ALICEO2_EMCAL_CELL_H_
 
 #include <bitset>
+#include <cfloat>
 #include "DataFormatsEMCAL/Constants.h"
 
 namespace o2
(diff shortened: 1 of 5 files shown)


mfasDa:
Needed to operate transiently with full precision. In a second step the compression method will be changed to optimise the bit usage to provide higher precision at lower energies.

The idea of the new compression is to make use of the cell type bit to separate the bitmap into two energy ranges, corresponding to the energy ranges of the HG (0-16 GeV) and LG (16 - 250 GeV) hardware digitisers, and take the cell type bit into account in the energy compression to distinguish the two ranges. This allows with the minimum amount of bits used in the CTF to recover the performance of the uncompressed case. A drawback of this compression is however that the energy calibration can move cell energies from the low gain to the high gain region and vice versa, and the cell type cannot be changed since the time calibration depends on it. As consequence the compression which is now used for in-memory representation must be used only when writing to / reading from CTF, which happens before any energy calibration is applied.

DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Cell.h
```diff
@@ -154,38 +143,84 @@ class Cell
...
+  void setTowerIDEncoded(uint16_t towerIDBits);
 
-  CellData* getDataRepresentation() { return reinterpret_cast<CellData*>(mCellWords); }
-  const CellData* getDataRepresentation() const { return reinterpret_cast<const CellData*>(mCellWords); }
+  /// \brief Set cell type from encoded bit representation (from CTF)
+  /// \param channelTypeBits Bit representation of cell type
+  void setChannelTypeEncoded(uint16_t channelTypeBits);
 
-  char mCellWords[6]; ///< data word
+  short mTowerID = SHRT_MAX;                             ///< Tower ID
```
shahor02, line 218: move short after float data members to not lose on the alignment padding

DataFormats/Detectors/EMCAL/src/Cell.cxx
```diff
@@ -19,56 +19,161 @@ using namespace o2::emcal;
...
+    truncatedEnergy = 0.;
+  } else if (truncatedEnergy > EnergyEncoding::v0::ENERGY_TRUNCATION) {
+    truncatedEnergy = EnergyEncoding::v0::ENERGY_TRUNCATION;
+  }
+  return static_cast<int16_t>(std::round(truncatedEnergy / EnergyEncoding::v0::ENERGY_RESOLUTION));
+}
+
+uint16_t Cell::encodeEnergyV1(float energy, ChannelType_t celltype)
 {
   double truncatedEnergy = energy;
```
shahor02, line 117: why double?

DataFormats/Detectors/EMCAL/src/Cell.cxx
```diff
@@ -19,56 +19,161 @@ using namespace o2::emcal;
...
+      resolutionApplied = EnergyEncoding::v1::ENERGY_RESOLUTION_LG;
+      energyOffset = EnergyEncoding::v1::HGLGTRANSITION;
+    }
+    case ChannelType_t::TRU: {
+      resolutionApplied = EnergyEncoding::v1::ENERGY_RESOLUTION_TRU;
+    }
+    case ChannelType_t::LEDMON: {
+      resolutionApplied = EnergyEncoding::v1::ENERGY_RESOLUTION_LEDMON;
+    }
+  }
```
shahor02, line 175: you forgot `break`-s after `case`-s

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
```diff
@@ -188,7 +195,14 @@ o2::ctf::CTFIOSize CTFCoder::decode(const CTF::base& ec, VTRG& trigVec, VCELL& c
...
-      cell.setPacked(tower[cellCount], cellTime[cellCount], energy[cellCount], status[cellCount]);
+      /*
+      uint16_t packedEnergy = energy[cellCount];
+      if (isVersion1) {
+        packedEnergy = o2::emcal::Cell::V0toV1(energy[cellCount], status[cellCount]);
+      }
+      cell.initialiseFromEncoded(tower[cellCount], cellTime[cellCount], packedEnergy, status[cellCount]);
+      */
+      cell.initialiseFromEncoded(tower[cellCount], cellTime[cellCount], energy[cellCount], status[cellCount]);
       cellVec.emplace_back(cell);
```
shahor02, line 206: use directly `cellVec.emplace_back(tower[cellCount], cellTime[cellCount], packedEnergy, status[cellCount]);`

shahor02 (changes requested):
as agreed

DataFormats/Detectors/EMCAL/src/Cell.cxx
mfasDa: Changed to auto

DataFormats/Detectors/EMCAL/src/Cell.cxx
mfasDa: Fixed, thanks for spotting!

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
mfasDa: That would only work if the values (cell time, energy and status) are in uncompressed format (no constructor for compressed information), or?

DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Cell.h
mfasDa: Done

Detectors/EMCAL/reconstruction/include/EMCALReconstruction/CTFCoder.h
shahor02: No, emplace_back calls the constructor of the same signature, just the construction happens in place

mfasDa: But this will then need a dedicated constructor. OK, can in principle be added.

mfasDa:
Close in favour of #11784 to track the PR with the proper JIRA ticket.

PR #11769, opened 2023-08-15
be explicit about signed char

sawenzel:
Fixing a compile error on ubuntu (which complained about -1 not being a char).

In fact, it seems that char needs explicit signed qualifier (in contrast to other integer types).


shahor02 (approved):
thanks, though on ubuntu 22.04 I don't see such an error. Which version do you use?

sawenzel:
@shahor02 : I may be due to a native compiler which I installed with `apt get` (and also this is on ARM hardware).

PR #11770, opened 2023-08-15
DPL Analysis: avoid spurious message at the end of the processing

ktf:
This solves the issue with the dummy extra iteration which is not needed and actually flagged as an error after 336eda18358.

The current logic is nevertheless to be reviewed because it seems to silently skip all the timeframes in a file if the first table cannot be read.


ktf:
@ddobrigk @aalkin @pzhristov @vkucera @pbuehler @jgrosseo: this fixes for good the fact that the extra iteration in the AOD reader is flagged as an error now. We should however discuss at some point if the current behaviour on errors, i.e. stop reading a file if the first table of any dataframe is misssing, is the wanted / correct one and if not what it should be.

vkucera:
Thanks @ktf

ktf:
Tested also on hyperloop, same size output. Merging.

PR #11776, opened 2023-08-16
Use raw data page full size for the CTP raw data dump

shahor02:


lietava:
@shahor02 thanks. So after checks it will be merged ?

shahor02:
actually, the changes are trivial and I've tested them locally, merging.

PR #11782, opened 2023-08-17
[EMCAL-551] Increase Cell ClassDef version

mfasDa:
- Increase ClassDef version of cell to 2, as the type
  of the data word has changed from ushort array to
  char array
- Add method in LinkDef to convert the data words
  from the old format to the new format.


mfasDa:
CI errors unrelated.

shahor02 (approved):

PR #11783, opened 2023-08-17
Added asynchronous functionality

TrifleMichael:
Added back:
- Asynchronous batch perform.
- A getAll() function, that works similar to future.get().
- Test for said perform function.

Additional information:
- The old blocking function is still present, and is the default.
- To ensure asynchronous performs are executed correctly one must run the mUVLoop often enough to ensure proper reading from sockets.
- Asynchronous perform functions return a structure which must be freed after all downloads scheduled via that perform finish and not earlier.

Additionally the code for old tests functions has been revised and compressed.

Previous PR for this functionality (which is not merged and is closed):
https://github.com/AliceO2Group/AliceO2/pull/11670

### DIFF
--- a/CCDB/src/CCDBDownloader.cxx
+++ b/CCDB/src/CCDBDownloader.cxx
@@ -54,6 +54,7 @@ void curlMultiErrorCheck(CURLMcode code)
     LOG(error) << "CCDBDownloader: CURL error - " << curl_multi_strerror(code);
   }
 }
+
 namespace
 {
 std::string uniqueAgentID()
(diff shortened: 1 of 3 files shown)


TrifleMichael:
Only the first commit contains functionality, the rest are formatting fixes. Sorry for that.

CCDB/src/CCDBDownloader.cxx
```diff
@@ -352,22 +353,8 @@ void CCDBDownloader::transferFinished(CURL* easy_handle, CURLcode curlCode)
...
-        // Temporary change before asynchronous calls are reintroduced
-        LOG(error) << "CCDBDownloader: Illegal request type";
-        break;
-      case ASYNCHRONOUS_WITH_CALLBACK:
-        // Temporary change before asynchronous calls will callbacks are reintroduced
-        LOG(error) << "CCDBDownloader: Illegal request type";
-        break;
-    }
-  }
+  *data->transferFinished = true;
```
ktf, line 356: What's the point of transferFinished, if you than delete the data immediately after?

ktf: ah, ok, I missed this is an iterator, nevermind.

ktf:
Thank you for this. To a quick look, this seems fine. Does it touch any API which is currently used? If not, I would suggest we merge this and we try to use it in the CCDB fetcher.

ktf:
@TrifleMichael @costing After our discussion of today, do you want to have this merged or do you plan to change this further?

TrifleMichael:
> @TrifleMichael @costing After our discussion of today, do you want to have this merged or do you plan to change this further?

It would be useful to merge it, as the url navigation will be based on this code.

PR #11784, opened 2023-08-17
[EMCAL- 834] Optimisation of cell-level compression

mfasDa:
- Make EMCAL cell uncompressed at transient level
- New energy compression algorithm based on cell type
- Provide reading directive from packed bitset
- Perpare CTF Coder for V0-V1 switch

### DIFF
--- a/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Cell.h
+++ b/DataFormats/Detectors/EMCAL/include/DataFormatsEMCAL/Cell.h
@@ -192,9 +192,18 @@ class Cell
   /// \return Encoded bit representation
   uint16_t getCellTypeEncoded() const;
 
+  void initializeFromPackedBitfieldV0(const char* bitfield);
+
+  static float getEnergyFromPackedBitfieldV0(const char* bitfield);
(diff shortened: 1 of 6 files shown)


mfasDa:
Needed to operate transiently with full precision. In a second step the compression method will be changed to optimise the bit usage to provide higher precision at lower energies.

The idea of the new compression is to make use of the cell type bit to separate the bitmap into two energy ranges, corresponding to the energy ranges of the HG (0-16 GeV) and LG (16 - 250 GeV) hardware digitisers, and take the cell type bit into account in the energy compression to distinguish the two ranges. This allows with the minimum amount of bits used in the CTF to recover the performance of the uncompressed case. A drawback of this compression is however that the energy calibration can move cell energies from the low gain to the high gain region and vice versa, and the cell type cannot be changed since the time calibration depends on it. As consequence the compression which is now used for in-memory representation must be used only when writing to / reading from CTF, which happens before any energy calibration is applied.

mfasDa:
Sorry, again typo in ticket ID. I however insist on the ticket being tracked by the correct ticket - in the end this helps more than what we loose with one unused PR ID.

